PR #1014, opened 2018-10-05
Use nested object for managing payment type/method matrix

joelochlann:
Precursor to PayPal Recurring

This also switches the behaviour to initialise all payment methods on a given tab when the tab becomes active, rather than initialising each one when its radio button is selected. I'm assuming this will make the UX better for PayPal recurring and Stripe, since these require initialisation of third-party code before the "Contribute" button will work (and in PayPal's case, the button won't even display until we've executed their code & the iframe has loaded)

### DIFF
--- a/assets/pages/new-contributions-landing/components/ContributionForm.jsx
+++ b/assets/pages/new-contributions-landing/components/ContributionForm.jsx
@@ -10,8 +10,15 @@ import { countryGroupSpecificDetails, type CountryMetaData } from 'helpers/inter
 import { type UsState, type CaState } from 'helpers/internationalisation/country';
 import { type CountryGroupId } from 'helpers/internationalisation/countryGroup';
 import { classNameWithModifiers } from 'helpers/utilities';
-import { type PaymentHandler, type PaymentMethod } from 'helpers/checkouts';
-import { config, type Contrib, type Amount } from 'helpers/contributions';
+import { type PaymentHandler } from 'helpers/checkouts';
+import {
+  config,
+  type Contrib,
+  type Amount,
+  type PaymentMatrix,
+  type PaymentMethod,
+  baseHandlers,
+} from 'helpers/contributions';
 import { type CheckoutFailureReason } from 'helpers/checkoutErrors';
 import { openDialogBox } from 'helpers/paymentIntegrations/newPaymentFlow/stripeCheckout';
 import { type PaymentAuthorisation } from 'helpers/paymentIntegrations/newPaymentFlow/readerRevenueApis';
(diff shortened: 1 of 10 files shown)


assets/pages/new-contributions-landing/components/ContributionForm.jsx
```diff
@@ -143,6 +143,40 @@ const getAmount = (props: PropTypes) =>
...
+}
+
+const formHandlersForRecurring: {[PaymentMethod]: (props: PropTypes) => void} = {
+  PayPal: () => { /* TODO PayPal recurring */ },
+  Stripe: openStripePopup,
+  DirectDebit: (props: PropTypes) => { props.openDirectDebitPopUp(); },
+};
+
+const formHandlers: {
+  [Contrib]: {
```
joelochlann: it would be cool if we could tell Flow to require us to provide keys for **all** possible values of `Contrib` - that would give us some extra type-safety when trying to call these functions

assets/pages/new-contributions-landing/components/ContributionPayment.jsx
```diff
@@ -53,37 +53,42 @@ const mapDispatchToProps = {
...
-
-      case 'PayPal':
-        // TODO
-        break;
+  // TODO IMPORTANT: we need to initialise Stripe separately for one-off and recurring
+  // I think this requires us to make
+  // paymentHandlers: { [PaymentMethod]: PaymentHandler | null }
+  // into
+  // paymentHandlers: { [Contrib]: {[PaymentMethod]: PaymentHandler | null }}
+  setupStripeCheckout(onPaymentAuthorisation, contributionType, currency, isTestUser)
```
joelochlann: From my testing it looks like this functionality to re-initialise Stripe with a different key was already missing before this PR (since we just checked `paymentHandler[paymentMethod]` which would be the same for one-off, monthly, annual)

joelochlann: Yes, have confirmed this by testing. Would be good to see if @regiskuckaertz has any thoughts on this

assets/pages/new-contributions-landing/components/ContributionForm.jsx
joelochlann: WIP of this using type parameters here: https://github.com/guardian/support-frontend/compare/npf-refactor-and-comments...jr-add-all-payments-type?expand=1

assets/pages/new-contributions-landing/components/ContributionPayment.jsx
regiskuckaertz: Oh I see, my bad 😵 I guess you have no other choice. Those matrices are coming up in several places now, perhaps it is reasonable to unify them all.

joelochlann: as far as the type is concerned, we've managed to unify the matrix! https://github.com/guardian/support-frontend/pull/1019

joelochlann: and unlike these object index types, it will enforce that all cases are covered so there shouldn't be runtime errors of trying to call a function which isn't set in the matrix

PR #1016, opened 2018-10-05
Drop a cookie to track Awin affiliate referrals

rupertbates:
## Why are you doing this?

Now that we are redirecting subscribe traffic to the support site, the affiliate tracking is getting lost which will lead to errors in attribution. This PR makes sure that we store the referrer in a cookie readable on subs when users first arrive on our site.

[**Trello Card**](https://trello.com/c/vEW5B56Q/1905-add-awin-to-support-frontend)

### DIFF
--- a/assets/helpers/tracking/awin.js
+++ b/assets/helpers/tracking/awin.js
@@ -1,0 +1,17 @@
+// @flow
+
+import { set as setCookie } from 'helpers/cookie';
+
+export default function storeReferrer() {
+  const cookieExpiryDays = 30;
+  const parsedUrl = new URL(window.location.href);
+  const utmSource = parsedUrl.searchParams.get('utm_source');
+  const utmMedium = parsedUrl.searchParams.get('utm_medium');
+  const gclid = parsedUrl.searchParams.get('gclid'); // Google AdWords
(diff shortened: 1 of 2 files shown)


assets/helpers/tracking/awin.js
```diff
@@ -0,0 +1,17 @@
...
+export default function storeReferrer() {
+  const cookieExpiryDays = 30;
+  const parsedUrl = new URL(window.location.href);
+  const utmSource = parsedUrl.searchParams.get('utm_source');
+  const utmMedium = parsedUrl.searchParams.get('utm_medium');
+  const gclid = parsedUrl.searchParams.get('gclid'); // Google AdWords
+  if (utmSource && utmMedium) {
+    setCookie('gu_referrer_channel', `${utmSource}&${utmMedium}`, cookieExpiryDays);
+  } else if (gclid) {
+    setCookie('gu_referrer_channel', 'google&adwords', cookieExpiryDays);
```
JustinPinner: Presumably the `gclid` value itself isn't required here

rupertbates: No, it just indicates that they've come from an adwords ad and hence not from an Awin referral

PR #1018, opened 2018-10-05
Add thankYouPageStageMap type

jranks123:
## Why are you doing this?
This guards us against not dealing with all thank you page stages

### DIFF
--- a/assets/pages/new-contributions-landing/contributionsLandingReducer.js
+++ b/assets/pages/new-contributions-landing/contributionsLandingReducer.js
@@ -24,8 +24,6 @@ export type UserFormData = {
   email: string | null,
 }
 
-
-
 export type ThankYouPageStageMap<T> = {
   setPassword: T,
   thankYou: T,
--- a/assets/pages/new-contributions-landing/contributionsLandingReducer.js
+++ b/assets/pages/new-contributions-landing/contributionsLandingReducer.js
@@ -26,14 +26,14 @@ export type UserFormData = {
 
 
 
-export type ThankYouPageStageMatrix<T> = {
+export type ThankYouPageStageMap<T> = {
   setPassword: T,
   thankYou: T,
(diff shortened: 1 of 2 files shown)


assets/pages/new-contributions-landing/contributionsLandingReducer.js
```diff
@@ -24,7 +24,14 @@ export type UserFormData = {
   email: string | null,
 }
 
-export type ThankYouPageStage = 'setPassword' | 'thankYou' | 'thankYouPasswordSet' | 'thankYouPasswordNotSet';
+export type ThankYouPageStageMap<T> = {
```
tsop14: Why does this have a type parameter? Aren't they always going to be booleans?

jranks123: The purpose of the Map is so that we can create a structure of `ThankYouPageStage ` to something with the confidence that we are dealing with every instance of `ThankYouPageStage`.

For example, in ContributionThankYouContainer, we want to return the correct `React$Element<*>` depending on the stage, so we create a `ThankYouPageStageMap` of type `React$Element<*>`. But this could be used with any type, for example if we wanted a different headline of the page for each `ThankYouPageStage` we could create a `ThankYouPageStage<string>`, and create a structure like this:

```
const headline: ThankYouPageStage<string> = {
    setPassword: "set password",
  thankYou: "some headline:,
  thankYouPasswordSet: "you've made a password, go you",
  thankYouPasswordNotSet: "booo"
}
```

On line 34 you will see that we are using the keys of this Map to define the different possible variations of `ThankYouPageStage`. The benefit of this is that if we decide later to add a new `ThankYouPageStage`, we will add it to the Map, and then we will have to deal with the new stage everywhere were we have used this structure, otherwise we will get a flow error.

PR #1022, opened 2018-10-08
NPF: Fix form handlers for recurring 

jranks123:
## Why are you doing this?
Fixes a spread syntax bug whereby we weren't overriding the base handlers with the form handlers for recurring.


joelochlann:
superseded by #1034

PR #1024, opened 2018-10-09
Link to post deployment test wiki section from Prout message

jacobwinch:
## Why are you doing this?

This should be more helpful than just linking to the test code.


johnduffell:
👍 great!

PR #1025, opened 2018-10-09
Improve SEO on digipack page

walaura:
## Why are you doing this?
Marketing asked us to improve the SEO of the digipack page, this PR adds the following changes:

[**Trello Card**](https://trello.com/c/PQeyeXKJ/1933-improve-support-pages-seo)

## Changes

- Update page title info as per recommendation from Chris
- Update page metadata as per recommendation from Chris
- Add a canonical meta tag
- Add canonical href lanf links

## Screenshots

N/A

### DIFF
--- a/app/config/StringsConfig.scala
+++ b/app/config/StringsConfig.scala
@@ -1,12 +1,13 @@
 package config
 
 import com.typesafe.config.ConfigFactory
+import config.ConfigImplicits._
 
 class StringsConfig {
   val config = ConfigFactory.load("strings.conf")
 
-  val supportLandingDescription = config.getString("supportLanding.description")
-  val contributionsLandingDescription = config.getString("contributionsLanding.description")
-  val subscriptionsLandingDescription = config.getString("subscriptionsLanding.description")
-  val digitalPackLandingDescription = config.getString("digitalPackLanding.description")
+  val supportLandingDescription = config.getOptionalString("supportLanding.description")
+  val contributionsLandingDescription = config.getOptionalString("contributionsLanding.description")
+  val subscriptionsLandingDescription = config.getOptionalString("subscriptionsLanding.description")
+  val digitalPackLandingDescription = config.getOptionalString("digitalPackLanding.description")
 }
--- a/app/config/StringsConfig.scala
+++ b/app/config/StringsConfig.scala
@@ -8,4 +8,5 @@ class StringsConfig {
   val supportLandingDescription = config.getString("supportLanding.description")
   val contributionsLandingDescription = config.getString("contributionsLanding.description")
   val subscriptionsLandingDescription = config.getString("subscriptionsLanding.description")
+  val digitalPackLandingDescription = config.getString("digitalPackLanding.description")
 }
--- a/app/controllers/Subscriptions.scala
+++ b/app/controllers/Subscriptions.scala
@@ -69,7 +69,7 @@ class Subscriptions(
     val js = "digitalSubscriptionLandingPage.js"
     val css = "digitalSubscriptionLandingPageStyles.css"
     val description = Some(stringsConfig.digitalPackLandingDescription)
-    val canonicalLink = Some(s"${supportUrl}/${countryCode}/subscribe/digital")
+    val canonicalLink = Some(buildCanonicalDigitalSubscriptionLink(countryCode))
 
     Ok(views.html.main(title, id, js, css, description, canonicalLink)).withSettingsSurrogateKey
   }
@@ -96,4 +96,8 @@ class Subscriptions(
       val css = "digitalSubscriptionCheckoutPageStyles.css"
       Ok(views.html.main(title, id, js, css)).withSettingsSurrogateKey
     }
+
+  def buildCanonicalDigitalSubscriptionLink(countryCode: String): String =
+    s"${supportUrl}/${countryCode}/subscribe/digital"
+
 }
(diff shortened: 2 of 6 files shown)


app/views/main.scala.html
```diff
@@ -23,6 +24,10 @@
 		<meta property="og:description" content="@description" />
 	}
 
+	@if(canonicalLink.isDefined) {
+		<link rel="canonical" href="@canonicalLink">
```
JustinPinner: Should this be `@canonicalLink.get`?

app/controllers/Subscriptions.scala
```diff
@@ -63,11 +64,14 @@ class Subscriptions(
...
   def digital(countryCode: String): Action[AnyContent] = CachedAction() { implicit request =>
     implicit val settings: Settings = settingsProvider.settings()
-    val title = "Support the Guardian | Digital Subscription"
+    val title = "Support the Guardian | Digital Pack Subscription"
     val id = "digital-subscription-landing-page-" + countryCode
     val js = "digitalSubscriptionLandingPage.js"
     val css = "digitalSubscriptionLandingPageStyles.css"
-    Ok(views.html.main(title, id, js, css)).withSettingsSurrogateKey
+    val description = Some(stringsConfig.digitalPackLandingDescription)
+    val canonicalLink = Some(buildCanonicalDigitalSubscriptionLink(countryCode))
```
LATaylor-guardian: In this instance `canonicalLink` doesn't need to be an `Option[String]` it can simply be of type `String` and you can remove the `Some` that wraps the function call. This will mean you won't have to check it's defined within your template.

app/controllers/Subscriptions.scala
```diff
@@ -63,11 +64,14 @@ class Subscriptions(
 
   def digital(countryCode: String): Action[AnyContent] = CachedAction() { implicit request =>
     implicit val settings: Settings = settingsProvider.settings()
-    val title = "Support the Guardian | Digital Subscription"
+    val title = "Support the Guardian | Digital Pack Subscription"
     val id = "digital-subscription-landing-page-" + countryCode
     val js = "digitalSubscriptionLandingPage.js"
     val css = "digitalSubscriptionLandingPageStyles.css"
-    Ok(views.html.main(title, id, js, css)).withSettingsSurrogateKey
+    val description = Some(stringsConfig.digitalPackLandingDescription)
```
LATaylor-guardian: Description here wants to be wrapped in `Option` rather than `Some`. This is because when we call `stringsConfig.digitalPackLandingDescription` it calls  `config.getString("digitalPackLanding.description")` which can return the String config item we want or if it is not present will throw an exception. See: https://lightbend.github.io/config/latest/api/com/typesafe/config/Config.html#getString-java.lang.String-.

Ideally, this behaviour would be modelled at the type level in `StringConfig` the object interacting with the TypeLevel config object but out of scope for this PR.

app/config/StringsConfig.scala
```diff
@@ -8,4 +8,5 @@ class StringsConfig {
   val supportLandingDescription = config.getString("supportLanding.description")
   val contributionsLandingDescription = config.getString("contributionsLanding.description")
   val subscriptionsLandingDescription = config.getString("subscriptionsLanding.description")
+  val digitalPackLandingDescription = config.getString("digitalPackLanding.description")
```
LATaylor-guardian: `config.getString` returns the config item or will throw an exception (see: https://lightbend.github.io/config/latest/api/com/typesafe/config/Config.html#getString-java.lang.String-).

Ideally we want to wrap in Option here to catch that exception and return None or Some if the config item is present. This prevents you having to wrap with Some below and assuming the config item is there. 

what you do with the Option here if it is None is a design choice. You can default to some other String, fail loudly with sys.error if it's core to the application running (not in this instance) or propagate the Option throughout as you have done.

app/controllers/Subscriptions.scala
LATaylor-guardian: See above.

app/controllers/Subscriptions.scala
walaura: This is a parameter being sent to the main template for all pages, meaning it has to be optional or else all pages would need to incorporate it, right?

app/views/main.scala.html
```diff
@@ -23,6 +24,10 @@
 		<meta property="og:description" content="@description" />
 	}
 
+	@if(canonicalLink.isDefined) {
```
LATaylor-guardian: You may simply`.map` here: `@canonicalLink.map { link => ... }`

app/views/main.scala.html
walaura: I'm using the style from `description.isDefined` from above

app/views/main.scala.html
walaura: I'm using the style `description.isDefined` uses right above, they should match IMO

app/config/StringsConfig.scala
lmath: Good shout. We shall read those strings as Option[String]. The None case is already handled in the template main.scala.html.

app/views/main.scala.html
JustinPinner: `isDefined` makes sure the `Option` is non-empty (not `None`). You still have to to a `get` or `getOrElse` to extract the content, no?

lmath: It works without the get but I don't know why 
image
We get that and I'm not sure why it's not like `<link rel="canonical" href="Some(https://support.thegulocal.com/uk/subscribe/digital)">` or something

Anyway, let's map over them both as @LATaylor-guardian suggested. 😄

app/views/main.scala.html
lmath: I just opted to map over both as I think your suggestion makes it clearer @LATaylor-guardian

PR #1026, opened 2018-10-09
Update css-loader

walaura:
## Why are you doing this?

As part of [the update libraries card](https://trello.com/c/gv9pLixX/1855-update-client-side-libraries), the `css-loader` plugin was the one firing the most alarms. This bumps it up to `1.0.0` and replaces the minifier (no longer part of css-loader as of v1) with `cssnano`, as [recommended](https://github.com/webpack-contrib/css-loader/blob/master/CHANGELOG.md) by its changelog.

[**Trello Card**](https://trello.com/c/gv9pLixX/1855-update-client-side-libraries)

## Screenshots

n/a

### DIFF
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -5,6 +5,7 @@ const ManifestPlugin = require('webpack-manifest-plugin');
 const MiniCssExtractPlugin = require('mini-css-extract-plugin');
 const autoprefixer = require('autoprefixer');
 const pxtorem = require('postcss-pxtorem');
+const cssnano = require('cssnano');
 
 module.exports = (cssFilename, outputFilename, minimizeCss) => ({
   plugins: [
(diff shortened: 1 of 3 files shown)


webpack.common.js
```diff
@@ -94,14 +95,15 @@ module.exports = (cssFilename, outputFilename, minimizeCss) => ({
...
-            },
           },
           {
             loader: 'postcss-loader',
             options: {
-              plugins: [pxtorem({ propList: ['*'] }), autoprefixer()],
+              plugins: [
+                pxtorem({ propList: ['*'] }),
+                autoprefixer(),
+                minimizeCss ? cssnano : false,
```
Ap0c: What does a value of `false` do, does it cause webpack to just ignore the array item?

PR #1028, opened 2018-10-09
Use ideal country code in Apple app store links where possible

JustinPinner:
## Why are you doing this?
Apple's app store defaults to US, and therefore US dollars, unless given a hint at the user's location. We can pass that through (after translating it into one of Apple's known values).

[**Trello Card**](https://trello.com/c/fX63xBuy/2005-add-region-country-into-ios-app-store-link)

## Screenshots
screen shot 2018-10-09 at 16 21 11

screen shot 2018-10-09 at 16 21 33

screen shot 2018-10-09 at 16 21 56

screen shot 2018-10-09 at 16 22 22

screen shot 2018-10-09 at 16 23 24

### DIFF
--- a/assets/helpers/externalLinks.js
+++ b/assets/helpers/externalLinks.js
@@ -36,9 +36,7 @@ export type SubsUrls = {
 const subsUrl = 'https://subscribe.theguardian.com';
 const patronsUrl = 'https://patrons.theguardian.com';
 const defaultIntCmp = 'gdnwb_copts_bundles_landing_default';
-const iOSAppUrl = 'https://itunes.apple.com/app/the-guardian/id409128287?mt=8';
 const androidAppUrl = 'https://play.google.com/store/apps/details?id=com.guardian';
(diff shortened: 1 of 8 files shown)


assets/helpers/externalLinks.js
```diff
@@ -216,14 +214,45 @@ function getDigitalCheckout(
...
 
+function convertCountryGroupIdToAppStoreCountryCode(cgId: CountryGroupId) {
+  const groupFromId = countryGroups[cgId];
+  if (groupFromId) {
+    switch (groupFromId.supportInternationalisationId.toLowerCase()) {
+      case 'uk':
+        return 'gb';
+      case 'int':
+        return 'us';
+      case 'eu':
```
lmath: There isn't an app store with euros?

JustinPinner: No, they support individual country codes, but not a EU according to their list here: https://affiliate.itunes.apple.com/resources/documentation/linking-to-the-itunes-music-store/#CountryCodes

👍for tests

PR #1032, opened 2018-10-10
Update newspaper -> magazine for Weekly

JustinPinner:
## Why are you doing this?
Because it's a glorious magazine now, not a newspaper.

[**Trello Card**](https://trello.com/c/S7PAob3t/2010-update-product-copy-for-gw-on-all-subscriptions-landing-pages)

## Changes

* See that word 'newspaper' on the Guardian Weekly copy? That's 'magazine' now, thank you very much. End of.

## Screenshots
**Before:**
screen shot 2018-10-10 at 16 21 25

**After:**
screen shot 2018-10-10 at 17 44 16


JustinPinner:
So the TC issues are back. Fun times.

PR #1037, opened 2018-10-11
Reduce Highlight Size

Ap0c:
## Why are you doing this?

Reducing the height of the top highlight on the subs landing page.

## Changes

- Reduced the padding and the corresponding `top` gaps.

### Before

header-before

### After

header-after


LATaylor-guardian:
👍

PR #1039, opened 2018-10-12
generate and post a visit token for the recurring new payment flow

johnduffell:
## Why are you doing this?

At the moment we are blocking out duplicate recurring contributions at the identity id level, so that if the step functions retry or the user double clicks/get dodgy wifi, we don't accidentally create two redurring contributions.  This works very well, but when we have the ability to contribute without logging in we can no longer do this at the identity id level as they are not really logged in.

The solution is to block the duplicates at the contrbutions form level rather than at the identity level.

I need to generate a token in the page, submit it to support frontend backend, then push it into the step functions, then check it against a field in the zuora subscription TBD.

This PR is the groundwork for this.  It generates the token and POSTs it to the api endpoint in support frontend.
The next PR will push that over to the step functions and more PRs after that.

For more background see this google doc https://docs.google.com/document/d/14wYvCij4pYZ-0YIY46lO4DZhL4ZfHD2max6WAyLnkpQ/edit?usp=sharing

[**Trello Card**](https://trello.com/c/eJnOL8se/71-npf-make-step-functions-allow-multiple-recurring-contributions-for-one-identity-account)

## Screenshots

```js
{
  "firstName":"aaa",
  "lastName":"aaa",
  "country":"GB",
  "state":null,
  "email":"aaa@gu.com",
  "contribution":{"amount":5,"currency":"GBP","billingPeriod":"Monthly"},
  "paymentFields":{"stripeToken":"tok_asaaa"},
  "ophanIds":{"pageviewId":"aaaa","browserId":null,"visitId":null},
  "referrerAcquisitionData":{},
  "supportAbTests":[],
  "visitToken":"1539362080"
}
```

@joelochlann @jranks123 @tomrf1 @tsop14 see what you think please!


app/controllers/Application.scala
```diff
@@ -146,3 +150,9 @@ class Application(
       Redirect("/" + path, request.queryString, MOVED_PERMANENTLY)
   }
 }
+
+case class VisitToken(value: String) extends AnyVal
+object VisitToken {
+  def apply(seed: LocalDateTime): VisitToken =
+    new VisitToken(s"${seed.toEpochSecond(ZoneOffset.UTC)}")
```
johnduffell: this only changes once a second, meaning it may be predictable.  Should we be using a secure/unpredictable random number?

assets/helpers/paymentIntegrations/newPaymentFlow/visitToken/actions.js
```diff
@@ -0,0 +1,15 @@
+// @flow
```
johnduffell: i pretty much copied this and reducer.js from the csrf equivalent if that's right?

assets/helpers/paymentIntegrations/newPaymentFlow/visitToken/reducer.js
```diff
@@ -0,0 +1,41 @@
...
+
+// ----- Imports ----- //
+
+import type { Action } from './actions';
+
+
+// ----- Types ----- //
+
+export type VisitToken = {
+    token: ?string,
```
johnduffell: I've had to make it optional, this isn't really optional.  I'm not sure how to tell they type system that `window.guardian.visitToken` will always be defined

app/controllers/Application.scala
joelochlann: does this mean two pages served within the same second would have the same `VisitToken`?

assets/helpers/paymentIntegrations/newPaymentFlow/visitToken/actions.js
joelochlann: maybe let's put this under `helpers` rather than `paymentIntegrations`

assets/helpers/paymentIntegrations/newPaymentFlow/visitToken/reducer.js
joelochlann: I think we would need to define a type for `window.guardian`, which we currently don't. The accuracy of this type would be dependent on the serverside logic which plays out that JSON in a `<script>` tag. Maybe something really clever for this would be to generate Flow types from Scala for those links between Twirl & JavaScript... https://github.com/davegurnell/bridges??

app/controllers/Application.scala
johnduffell: I have changed to use SecureRandom.  it now enters a long as hex into the page e.g.
`window.guardian.visitToken = "3ada0bd900da76d3";`

assets/helpers/paymentIntegrations/newPaymentFlow/visitToken/actions.js
johnduffell: done

assets/helpers/paymentIntegrations/newPaymentFlow/visitToken/reducer.js
johnduffell: nice idea. that would be a good thing to do in future.  The writing into the page is effectively a serialisation to string, so all the information is lost.  Having to only respecify this at the boundary wouldn't be a disaster, but if we could not lose it in the first place it would be better.

assets/helpers/visitToken/actions.js
```diff
@@ -0,0 +1,15 @@
+// @flow
+
+// ----- Types ----- //
+
+export type Action = { type: 'SET_VISIT_TOKEN', name: string };
+
+
+// ----- Actions Creators ----- //
+
+export function setToken(name: string): Action {
```
Ap0c: Conventionally this [action creator](https://redux.js.org/basics/actions#action-creators) would take the same name as the action; in this case `setVisitToken`.

assets/helpers/visitToken/actions.js
```diff
@@ -0,0 +1,15 @@
...
+  return { type: 'SET_VISIT_TOKEN', name };
+}
+
+export default setToken;
```
Ap0c: We prefer [named exports](https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export#Description) to default exports for everything except components and reducer files. This is because these types of files conceptually should only ever have one export, whereas other files, such as those that contain actions, can have multiple exports. So in this case:

```jsx
export { setToken };
```

assets/helpers/visitToken/reducer.js
```diff
@@ -0,0 +1,41 @@
...
+
+export default function visitTokenReducer(
+  state: VisitToken = initialState,
+  action: Action,
+): VisitToken {
+
+  switch (action.type) {
+
+    case 'SET_VISIT_TOKEN':
+      return Object.assign({}, state, { token: action.name });
```
Ap0c: This could also be achieved with [object spread](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#Spread_in_object_literals):

```jsx
return { ...state, token: action.name };
```

assets/helpers/visitToken/reducer.js
```diff
@@ -0,0 +1,41 @@
...
+// ----- Imports ----- //
+
+import type { Action } from './actions';
+
+
+// ----- Types ----- //
+
+export type VisitToken = {
+    token: ?string,
+};
```
Ap0c: Does the state here need to be an object? Could it be:

```jsx
type VisitToken = ?string;
```

johnduffell:
better late then never @Ap0c 😆 Thanks for the comments though 💯 
I mostly copied the CSRF one and changed the name, but I will use your comments in the next PR I am working on in this code.

assets/helpers/visitToken/reducer.js
johnduffell: I'm confused now, it seems like the initial state is only used in this function, however the SET_VISIT_TOKEN event is never raised.  However as a side effect of any unrelated event whatsoever being raised, we return the initial state.  This means that we are relying on an arbitrary event being raised in order to fill in the default.  This doesn't seem very accurate design, unless I'm missing something?

Ap0c: Are you saying that this state never changes?

assets/helpers/visitToken/actions.js
joelochlann: What's the rationale for preferring exports at the bottom of the file? I find it easier to see right next to the function body whether it's exported (it's akin to marking a method `public` in other languages), rather than having to the jump to the bottom of the file. It can sometimes make the cmd+clicking in IntelliJ a bit more tortuous

joelochlann:
One thing I would say about this - I think it highlights that data which is set on the server in `window.guardian` and never mutated doesn't belong in the Redux store. A lot of un-needed boilerplate could be removed if we just reference `window.guardian` directly where needed (which we in fact already do in a few places). What do you think @Ap0c? Are there any advantages to putting data like this in Redux?

assets/helpers/visitToken/reducer.js
johnduffell: discussed in person, the answer is yes it never changes, and I've fixed it all in the next PR https://github.com/guardian/support-frontend/pull/1049 😉

assets/helpers/visitToken/reducer.js
johnduffell: fixed in next pr

assets/helpers/visitToken/actions.js
johnduffell: this file has been deleted in the next PR, but interesting debate, thanks for the explanation

Ap0c:
> One thing I would say about this - I think it highlights that data which is set on the server in `window.guardian` and never mutated doesn't belong in the Redux store. A lot of un-needed boilerplate could be removed if we just reference `window.guardian` directly where needed (which we in fact already do in a few places). What do you think @Ap0c? Are there any advantages to putting data like this in Redux?

I think that if we're pulling information out of the `window` object it would be preferable to do that in a single place in the application, perhaps where create the Redux store, and then propagate that information through to where it's needed. That way we constrain our side-effectful code to exist in one place rather than spreading it throughout the application. I realise that we don't do this at the moment, because we pull information out of the `window` in a number of files, including some of the reducers.

If we go with this approach, the question is then how it makes its way down to where it's needed. There are two approaches that I can think of:

- Put it in the Redux store so that we can pull it out using the same mechanism as we do for other bits of state. This will probably require higher-order reducers, although we could probably create a single-use function to handle this, something like:

```jsx
function createConfigReducer<A>(somePieceOfConfig: A) {
  return (state: A = somePieceOfConfig, action: {}): A => state;
}
```

- Pass it down as static component props. This won't require a reducer but it will result in more "pass-through" props, to get the information down to where it's needed in the component tree (you don't get the benefit of connecting components further down the tree because the data isn't in the Redux store).

What do you think @joelochlann @johnduffell ?

johnduffell:
that's a long comment!  I'm in favour of whatever reduces the dependencies most, and what improves the cohesion most.
pass through parameters is always a no-no for me, as it can be dependency injected instead (in scala I would partially apply the function)
I'd rather the javascript loader that kicks off the in page app, got the fields from wherever and just passed them in as parameters to whatever needs it (the `regularPaymentRequestFromAuthorisation` function)  Then the messy bits are at the edge and everything inside no longer cares about where it's coming from.

joelochlann:
Those are interesting ideas. Personally I'm struggling to see what tangible benefits we get from not reading from `window.guardian` directly. From the clientside perspective everything in there is just a constant. So accessing it doesn't make code side-effectful, in my view. A function that accesses a constant is still pure (https://stackoverflow.com/questions/45672603/can-a-pure-function-depend-on-an-external-constant).

I do take the point that nothing in the JS language enforces the immutability of `window.guardian`. But it's a very widespread convention in the department to store read-only config from the server in there, so while not technically constant like `const tipPercentage = 0.15` in the StackOverflow example, I would argue that for all intents and purposes it is.

I'm sympathetic to the argument that reading from the global inside a function body can making testing harder since you can't vary the global by varying the function params. But in JavaScript it's so easy to mock such globals in test setup that I'm not sure this a big issue for me.

PR #1042, opened 2018-10-16
Use new thankyou page for single contributions

tomrf1:
Successful paypal one-off contribution redirects to:
picture 27

### DIFF
--- a/app/controllers/PayPalOneOff.scala
+++ b/app/controllers/PayPalOneOff.scala
@@ -58,9 +58,16 @@ class PayPalOneOff(
     }
   }
 
-  def resultFromPaypalSuccess(success: PayPalSuccess)(implicit request: RequestHeader): Result = {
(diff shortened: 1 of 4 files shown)


app/controllers/PayPalOneOff.scala
```diff
@@ -104,7 +114,7 @@ class PayPalOneOff(
 
     emailForUser(request.user)
       .flatMap(paymentAPIService.executePaypalPayment(paymentJSON, acquisitionData, queryStrings, _, isTestUser))
-      .fold(resultFromPaymentAPIError, resultFromPaypalSuccess)
+      .fold(resultFromPaymentAPIError, success => resultFromPaypalSuccess(success, country))
```
joelochlann: I think we may need to update the error page too eventually but I don't think we've even got a design for that (and it's non-essential for launching the test). I've added a card for it: https://trello.com/c/je1lyCWY/99-npf-follow-up-paypal-error-page-design-for-new-flow

PR #1044, opened 2018-10-16
Stop showing error message after successful paypal payment creation

tomrf1:
## Why are you doing this?
When the one-off paypal contribution form is submitted, it briefly and incorrectly displays an error message before navigating to paypal.

[**Trello Card**](https://trello.com/c/jkv5OUHs/97-npf-an-error-message-is-always-briefly-displayed-when-submitting-the-on-off-payment-form)


johnduffell:
nice spot! 👍

PR #1045, opened 2018-10-17
Update the CSP to allow the Awin master tag to work.

rupertbates:
## Why are you doing this?
The Awin master tag which we need for affiliate referral tracking is currently being blocked by our CSP, this updates it to allow the tag to work.


LATaylor-guardian:
👍

PR #1047, opened 2018-10-17
initialise third party payment libraries

tsop14:
## Why are you doing this?
So the NPF works for Stripe one-off. 

[**Trello Card**](https://trello.com/c/41e5ijeC/76-npf-stripe-does-not-switch-between-recurring-one-off-keys-when-changing-tabs-this-means-stripe-one-off-is-broken)

## Changes

* Renames `paymentHandlers` to `thirdPartyPaymentLibraries` because that is what they are! 
* Adds extra dimension to thirdPartyPaymentLibraries as Stripe one-off is different from recurring. Uses PaymentMatrix to ensure all combinations are covered! 
* Moves initialization of these libraries out of the render function and in to `contributionsLandingInit`.
* Initialises all libraries on page load (rather than when the tab / payment option is selected)
* No longer initialises with the matrix as the majority of combinations did not require action, instead uses a humble if statement to initialise stripe checkout.

### DIFF
--- a/assets/pages/new-contributions-landing/components/ContributionForm.jsx
+++ b/assets/pages/new-contributions-landing/components/ContributionForm.jsx
@@ -86,7 +86,7 @@ type PropTypes = {|
   otherAmount: string | null,
   paymentMethod: PaymentMethod,
   isSignedIn: boolean,
-  paymentHandlers: PaymentMatrix<PaymentHandler | null>,
+  thirdPartyPaymentLibraries: PaymentMatrix<PaymentHandler | null>,
   updateFirstName: Event => void,
   updateLastName: Event => void,
   updateEmail: Event => void,
@@ -120,7 +120,7 @@ const mapStateToProps = (state: State) => ({
   otherAmount: state.page.form.formData.otherAmounts[state.page.form.contributionType].amount,
   paymentMethod: state.page.form.paymentMethod,
   isSignedIn: state.page.user.isSignedIn,
-  paymentHandlers: state.page.form.paymentHandlers,
+  thirdPartyPaymentLibraries: state.page.form.thirdPartyPaymentLibraries,
   contributionType: state.page.form.contributionType,
   checkoutFormHasBeenSubmitted: state.page.form.formData.checkoutFormHasBeenSubmitted,
   isDirectDebitPopUpOpen: state.page.directDebit.isPopUpOpen,
@@ -154,8 +154,8 @@ const getAmount = (props: PropTypes) =>
 // ----- Event handlers ----- //
 
 function openStripePopup(props: PropTypes) {
-  if (props.paymentHandlers[props.contributionType] && props.paymentHandlers[props.contributionType].Stripe) {
-    openDialogBox(props.paymentHandlers[props.contributionType].Stripe, getAmount(props), props.email);
+  if (props.thirdPartyPaymentLibraries[props.contributionType] && props.thirdPartyPaymentLibraries[props.contributionType].Stripe) {
+    openDialogBox(props.thirdPartyPaymentLibraries[props.contributionType].Stripe, getAmount(props), props.email);
   }
 }
 
--- a/assets/pages/new-contributions-landing/components/ContributionForm.jsx
+++ b/assets/pages/new-contributions-landing/components/ContributionForm.jsx
@@ -9,7 +9,7 @@ import { connect } from 'react-redux';
 
 import { type CountryGroupId } from 'helpers/internationalisation/countryGroup';
 import { classNameWithModifiers } from 'helpers/utilities';
-import { type PaymentHandler } from 'helpers/checkouts';
+import { type ThirdPartyPaymentLibrary } from 'helpers/checkouts';
 import {
   config,
   type Contrib,
@@ -60,7 +60,7 @@ type PropTypes = {|
   email: string,
   otherAmount: string | null,
   paymentMethod: PaymentMethod,
-  thirdPartyPaymentLibraries: PaymentMatrix<PaymentHandler | null>,
+  thirdPartyPaymentLibraries: PaymentMatrix<ThirdPartyPaymentLibrary | null>,
   contributionType: Contrib,
   currency: IsoCurrency,
   paymentError: CheckoutFailureReason | null,
--- a/assets/pages/new-contributions-landing/components/ContributionForm.jsx
+++ b/assets/pages/new-contributions-landing/components/ContributionForm.jsx
@@ -86,7 +86,7 @@ type PropTypes = {|
   otherAmount: string | null,
   paymentMethod: PaymentMethod,
   isSignedIn: boolean,
-  paymentHandlers: { [PaymentMethod]: PaymentHandler | null },
+  paymentHandlers: PaymentMatrix<PaymentHandler | null>,
   updateFirstName: Event => void,
   updateLastName: Event => void,
   updateEmail: Event => void,
@@ -154,8 +154,8 @@ const getAmount = (props: PropTypes) =>
 // ----- Event handlers ----- //
 
 function openStripePopup(props: PropTypes) {
-  if (props.paymentHandlers.Stripe) {
-    openDialogBox(props.paymentHandlers.Stripe, getAmount(props), props.email);
+  if (props.paymentHandlers[props.contributionType] && props.paymentHandlers[props.contributionType].Stripe) {
+    openDialogBox(props.paymentHandlers[props.contributionType].Stripe, getAmount(props), props.email);
   }
 }
 
--- a/assets/pages/new-contributions-landing/components/ContributionForm.jsx
+++ b/assets/pages/new-contributions-landing/components/ContributionForm.jsx
@@ -154,8 +154,9 @@ const getAmount = (props: PropTypes) =>
 // ----- Event handlers ----- //
 
 function openStripePopup(props: PropTypes) {
-  if (props.thirdPartyPaymentLibraries[props.contributionType] && props.thirdPartyPaymentLibraries[props.contributionType].Stripe) {
-    openDialogBox(props.thirdPartyPaymentLibraries[props.contributionType].Stripe, getAmount(props), props.email);
+  const paymentLibraries = props.thirdPartyPaymentLibraries[props.contributionType];
+  if (paymentLibraries && paymentLibraries.Stripe) {
+    openDialogBox(paymentLibraries.Stripe, getAmount(props), props.email);
   }
 }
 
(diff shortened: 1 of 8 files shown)


assets/pages/new-contributions-landing/contributionsLandingActions.js
```diff
@@ -93,9 +93,11 @@ const setGuestAccountCreationToken = (guestAccountCreationToken: string): Action
...
   ({ type: 'SET_THANK_YOU_PAGE_STAGE', thankYouPageStage });
 
-const isPaymentReady = (paymentReady: boolean, paymentHandlers: ?{ [PaymentMethod]: PaymentHandler }): Action =>
-  ({ type: 'UPDATE_PAYMENT_READY', paymentReady, paymentHandlers: paymentHandlers || null });
-
+const isPaymentReady = (
+  paymentReady: boolean,
+  paymentHandlers: ?{ [Contrib]: { [PaymentMethod]: PaymentHandler }},
+): Action =>
+  ({ type: 'UPDATE_PAYMENT_READY', paymentReady, thirdPartyPaymentLibraries: paymentHandlers || null });
```
regiskuckaertz: I'm surprised you can override `thirdPartyPaymentLibraries`, there is no guarantee that `paymentHandler` will have the correct shape with the required values, is there?

assets/pages/new-contributions-landing/contributionsLandingActions.js
```diff
@@ -264,7 +266,7 @@ const paymentAuthorisationHandlers: PaymentMatrix<(Dispatch<Action>, State, Paym
 };
 
 const onThirdPartyPaymentAuthorised = (paymentAuthorisation: PaymentAuthorisation) =>
-  (dispatch: Dispatch<Action>, getState: () => State): void => {
+  (dispatch: Function, getState: () => State): void => {
```
regiskuckaertz: 🤔

assets/pages/new-contributions-landing/contributionsLandingInit.js
```diff
@@ -23,12 +25,38 @@ function initialisePaymentMethod(state: State, dispatch: Dispatch<Action>) {
...
 }
 
+function initialiseStripeCheckout(onPaymentAuthorisation, contributionType, currencyId, isTestUser, dispatch) {
+
+  setupStripeCheckout(onPaymentAuthorisation, contributionType, currencyId, isTestUser)
+    .then((handler: PaymentHandler) => dispatch(isPaymentReady(true, { [contributionType]: { Stripe: handler } })));
+}
+
+function initialisePaymentMethods(state: State, dispatch: Function) {
+  const { countryId } = state.common.internationalisation;
```
regiskuckaertz: `const { countryId, currencyId } = state.common.internationalisation`

assets/pages/new-contributions-landing/contributionsLandingInit.js
```diff
@@ -23,12 +25,38 @@ function initialisePaymentMethod(state: State, dispatch: Dispatch<Action>) {
...
+  const { switches } = state.common.settings;
+  const { currencyId } = state.common.internationalisation;
+  const { isTestUser } = state.page.user;
+
+  const onPaymentAuthorisation = (paymentAuthorisation: PaymentAuthorisation) => {
+    dispatch(paymentWaiting(true));
+    dispatch(onThirdPartyPaymentAuthorised(paymentAuthorisation));
+  };
+
+  ['ONE_OFF', 'ANNUAL', 'MONTHLY'].forEach((contribType) => {
```
regiskuckaertz: I would refactor this into an inner loop over the payment methods:
```js
['ONE_OFF', 'ANNUAL', 'MONTHLY'].forEach(ct =>
  ['Stripe', 'PayPal', 'DirectDebit'].forEach(pm =>
    if (isValidPaymentMethod(ct, pm, switches, countryId))
      initPaymentMethod(...)
  )
)
```

assets/pages/new-contributions-landing/contributionsLandingReducer.js
```diff
@@ -148,11 +159,11 @@ function createFormReducer(countryGroupId: CountryGroupId) {
         return { ...state, paymentMethod: action.paymentMethod };
 
       case 'UPDATE_PAYMENT_READY':
-        return action.paymentHandlers
+        return action.thirdPartyPaymentLibraries
           ? {
             ...state,
             paymentReady: action.paymentReady,
-            paymentHandlers: { ...state.paymentHandlers, ...action.paymentHandlers },
+            thirdPartyPaymentLibraries: { ...state.thirdPartyPaymentLibraries, ...action.thirdPartyPaymentLibraries },
```
regiskuckaertz: Unfortunately that won't work. The spread syntax performs a shallow copy, so any element in `action.thirdPartyPaymentLibraries` will overwrite those in `state.thirdPartyPaymentLibraries`, whereas what you want is to merge them together.

assets/pages/new-contributions-landing/contributionsLandingInit.js
tsop14: done

assets/pages/new-contributions-landing/components/PaymentMethodSelector.jsx
```diff
@@ -35,9 +32,9 @@ type PropTypes = {
   currency: IsoCurrency,
   paymentMethod: PaymentMethod,
   onPaymentAuthorisation: PaymentAuthorisation => void,
-  paymentHandlers: { [PaymentMethod]: PaymentHandler | null },
+  thirdPartyPaymentLibraries: PaymentMatrix<PaymentHandler | null>,
   updatePaymentMethod: PaymentMethod => Action,
-  setPaymentIsReady: (boolean, ?{ [PaymentMethod]: PaymentHandler }) => Action,
+  setPaymentIsReady: (boolean, ?{ [Contrib]: { [PaymentMethod]: PaymentHandler }}) => Action,
```
joelochlann: I think we should probably rename this function and remove the boolean since it's a bit misleading (which payment is ready?). Maybe `setThirdPartyPaymentLibrary`

assets/pages/new-contributions-landing/contributionsLandingReducer.js
tsop14: Good point! Have updated, thanks.

assets/pages/new-contributions-landing/contributionsLandingActions.js
joelochlann: You mean because are no types here?

```javascript
function initialiseStripeCheckout(onPaymentAuthorisation, contributionType, currencyId, isTestUser, dispatch) 
```

Could be a weakness in Flow. Since I think untyped arguments get an `any` type, and the implied `contributionType: any` is ultimately being used as a key required to be `Contrib`, I would've thought it would complain

assets/pages/new-contributions-landing/components/ContributionForm.jsx
```diff
@@ -60,7 +60,7 @@ type PropTypes = {|
   email: string,
   otherAmount: string | null,
   paymentMethod: PaymentMethod,
-  paymentHandlers: { [PaymentMethod]: PaymentHandler | null },
+  thirdPartyPaymentLibraries: PaymentMatrix<PaymentHandler | null>,
```
joelochlann: Let's rename `PaymentHandler` type to `ThirdPartyPaymentLibrary`

tsop14: done

assets/pages/new-contributions-landing/components/PaymentMethodSelector.jsx
tsop14: Have renamed & removed boolean, this function is only called once with 'true' and the boolean didn't affect the logic.

assets/pages/new-contributions-landing/contributionsLandingActions.js
tsop14: this isn't declared as a `type Action` and so it cannot have type `Dispatch<Action>`.  Should it be included in `type Action`? I saw that a few other places with similar issue used generic Function type which is why I used same fix here.

assets/pages/new-contributions-landing/contributionsLandingInit.js
tsop14: As we only want to initialise Stripe one off, and Stripe and PayPal recurring, writing out these specific commands rather than looping over all options.

assets/pages/new-contributions-landing/contributionsLandingActions.js
```diff
@@ -100,12 +100,18 @@ const setGuestAccountCreationToken = (guestAccountCreationToken: string): Action
 const setThankYouPageStage = (thankYouPageStage: ThankYouPageStage): Action =>
   ({ type: 'SET_THANK_YOU_PAGE_STAGE', thankYouPageStage });
 
-const setPaymentIsReady = (paymentReady: boolean, paymentHandlers: ?{ [PaymentMethod]: PaymentHandler }): Action =>
-  ({ type: 'UPDATE_PAYMENT_READY', paymentReady, paymentHandlers: paymentHandlers || null });
+const setThirdPartyPaymentLibrary =
```
tsop14: This is over several lines in order to make the linter happy. ✨ Solution thanks to @jranks123

assets/pages/new-contributions-landing/components/ContributionForm.jsx
```diff
@@ -60,7 +60,7 @@ type PropTypes = {|
   email: string,
   otherAmount: string | null,
   paymentMethod: PaymentMethod,
-  paymentHandlers: { [PaymentMethod]: PaymentHandler | null },
+  thirdPartyPaymentLibraries: PaymentMatrix<ThirdPartyPaymentLibrary | null>,
```
joelochlann: should this type be `ThirdPartyPaymentLibraries` now?

assets/pages/new-contributions-landing/components/PaymentMethodSelector.jsx
```diff
@@ -35,9 +32,9 @@ type PropTypes = {
   currency: IsoCurrency,
   paymentMethod: PaymentMethod,
   onPaymentAuthorisation: PaymentAuthorisation => void,
-  paymentHandlers: { [PaymentMethod]: PaymentHandler | null },
+  thirdPartyPaymentLibraries: PaymentMatrix<ThirdPartyPaymentLibrary | null>,
```
joelochlann: and here

johnduffell:
> Renames paymentHandlers to thirdPartyPaymentLibraries because that is what they are!

⭐️

PR #1048, opened 2018-10-17
Ad free banner in DP page

walaura:
## Why are you doing this?
To promote ad-free more fiercely. 

[**Trello Card**](https://trello.com/c/P3ONYbjw/1912-add-ad-free-to-digital-pack-product-page)

## Changes
* Add an ad free promotion component
* Add a 2px top padding to the journalism component below because the text was awfully close to the dark blue strip

# Screenshots

screen shot 2018-10-17 at 3 18 10 pm


lmath:
btw, if you get the latest from master (including https://github.com/guardian/support-frontend/pull/1046) then snyk will calm down

assets/components/adFreeSection/AdFreeSection.jsx
```diff
@@ -0,0 +1,33 @@
...
+  return (
+    <LeftMarginSection modifierClasses={['blue']}>
+      <div className="component-ad-free">
+        <div className="component-ad-free__content">
+          <h3 className="component-ad-free__badge">New</h3>
+          <div className="component-ad-free__wrapper">
+            <div className="component-ad-free__col">
+              <h2 className="component-ad-free__header">Ad-free on all your devices</h2>
+              <p className="component-ad-free__copy">
+                {'Avoid the adverts and read without interruptions when you\'re signed in on your apps and theguardian.com'}
```
walaura: Not sure about this? Eslint wanted me to escape the `'` character as an html entity if I put it as text, but that sounds like a readability nightmare

assets/components/adFreeSection/AdFreeSection.jsx
```diff
@@ -0,0 +1,33 @@
...
+// ----- Imports ----- //
+import React from 'react';
+import GridImage from 'components/gridImage/gridImage';
+import LeftMarginSection from 'components/leftMarginSection/leftMarginSection';
+
+export default function AdFreeSection() {
+
+  return (
+    <LeftMarginSection modifierClasses={['blue']}>
+      <div className="component-ad-free">
```
Ap0c: By convention we make the `className` match the component name, so in this case it would become `component-ad-free-section`. We also try to scope everything in the component within the [BEM "block"](http://getbem.com/naming/), so here `LeftMarginSection` would live inside `<div className="component-ad-free-section">`.

assets/components/adFreeSection/AdFreeSection.jsx
Ap0c: I think that's ok, we do that a lot. It's `&#39;`?

assets/components/adFreeSection/AdFreeSection.jsx
```diff
@@ -0,0 +1,33 @@
...
+        <div className="component-ad-free__content">
+          <h3 className="component-ad-free__badge">New</h3>
```
Ap0c: Because this is a shared component, we don't necessarily know what level of heading it should have, it depends on the context in the page. You could use the `Heading` component and pass down the level of heading in the hierachy to this component (e.g. `1`, `2`, `3` etc.).

assets/components/leftMarginSection/leftMarginSection.scss
```diff
@@ -1,6 +1,13 @@
 .component-left-margin-section {
   display: flex;
 
+  &.component-left-margin-section--blue {
```
Ap0c: Because this class is unique across the codebase I think you could put this rule-set at the top level of this file?

assets/pages/digital-subscription-landing/digitalSubscriptionLanding.jsx
```diff
@@ -13,6 +13,7 @@ import Page from 'components/page/page';
 import countrySwitcherHeaderContainer from 'components/headers/countrySwitcherHeader/countrySwitcherHeaderContainer';
 import CustomerService from 'components/customerService/customerService';
 import Footer from 'components/footer/footer';
+import AdFreeSection from 'components/adFreeSection/AdFreeSection';
```
Ap0c: By convention we usually put filenames in camelCase.

PR #1049, opened 2018-10-17
Use scope token passed from the client side to kick off step functions

johnduffell:
## Why are you doing this?

This PR is part of the work to only allow one recurring contribution per time that the reader brings up the form on their browser.

This is the next stage, when we take the field that is being posted to the backend of this app and post it into the step functions with the new models.

Eventually the step functions will post that into zuora and we can use it in the checks.

This is the previous PR that added the field to the client side, although it's been changed a lot  in this one
https://github.com/guardian/support-frontend/pull/1039

[**Trello Card**](https://trello.com/c/qH5teTnT/86-npf-make-support-frontend-backend-pass-through-browser-form-token-to-step-functions)

@Ap0c as discussed I have updated the redux stuff to not expect there to be changes.
@joelochlann @rupertbates @jacobwinch @jranks123 @tsop14 take a look when you get some time please!


build.sbt
```diff
@@ -64,7 +64,7 @@ libraryDependencies ++= Seq(
   "com.amazonaws" % "aws-java-sdk-s3" % awsVersion,
   "org.typelevel" %% "cats-core" % "1.0.1",
   "com.dripower" %% "play-circe" % "2609.1",
-  "com.gu" %% "support-models" % "0.32",
+  "com.gu" %% "support-models" % "0.33",
```
johnduffell: https://github.com/guardian/support-models/pull/42

app/controllers/Application.scala
```diff
@@ -111,14 +112,14 @@ class Application(
   def newContributionsLanding(countryCode: String): Action[AnyContent] = maybeAuthenticatedAction().async { implicit request =>
     type Attempt[A] = EitherT[Future, String, A]
     implicit val settings: Settings = settingsProvider.settings()
-    val visitToken = VisitToken(new SecureRandom())
+    val scopeToken = CreateScopeToken(new SecureRandom())
```
johnduffell: I decided scopeToken made more sense as this token is used for the scope of access given to the identity id from the user of this page.  Named after where it's going rather than where it came from.  However naming is hard and this isn't instantly understandable .

app/controllers/Application.scala
```diff
@@ -153,8 +154,7 @@ class Application(
   }
 }
 
-case class VisitToken(value: String) extends AnyVal
-object VisitToken {
-  def apply(rand: SecureRandom): VisitToken =
-    new VisitToken(s"${rand.nextLong.toHexString}")
+object CreateScopeToken {
```
johnduffell: the case class is now in the support models.  (Even though at this point it's yet to be serialised, sent to the client side, and sent back again for deserialising.)

app/controllers/RegularContributions.scala
```diff
@@ -113,8 +115,8 @@ class RegularContributions(
     SafeLogger.info(s"[${request.uuid}] User ${fullUser.id} is attempting to create a new ${request.body.contribution.billingPeriod} contribution")
 
     val result = for {
-      user <- identityService.getUser(fullUser)
-      response <- client.createContributor(request.body, contributor(user, request.body), request.uuid).leftMap(_.toString)
+      user <- identityService.getUser(fullUser) // FIXME also need to check they are email verified before using AccessScopeNoRestriction
```
johnduffell: this fixme is logged as https://trello.com/c/sj9HGwtp/98

app/controllers/Application.scala
rupertbates: I find the term 'Scope' confusing here, it doesn't really describe the purpose of the token. Do you think we could come up with something that does? Maybe `deDupingToken` or something along those lines?

johnduffell: jonny and I have come up with session id, and we are going to store it in the local session storage

rupertbates: I like that better, it's a less overloaded term 👍

johnduffell: thanks, see new PR https://github.com/guardian/support-models/pull/43

johnduffell:
ok I have pushed the changes based on people's comments, depends on https://github.com/guardian/support-models/pull/43 and I need to do some testing (it compiles so far) I will do that this afternoon, but please come back and make some comments as it's basically ready

build.sbt
```diff
@@ -64,7 +64,7 @@ libraryDependencies ++= Seq(
   "com.amazonaws" % "aws-java-sdk-s3" % awsVersion,
   "org.typelevel" %% "cats-core" % "1.0.1",
   "com.dripower" %% "play-circe" % "2609.1",
-  "com.gu" %% "support-models" % "0.32",
+  "com.gu" %% "support-models" % "0.34-SNAPSHOT",
```
johnduffell: https://github.com/guardian/support-models/pull/43

johnduffell:
I have tested.  when not logged in, the token is sent to the step functions.  When logged in, the token is sent as null.
I have merged in master and it's ready when it's reviewed!  please take another look @rupertbates @jranks123

app/controllers/Application.scala
```diff
@@ -111,14 +111,13 @@ class Application(
   def newContributionsLanding(countryCode: String): Action[AnyContent] = maybeAuthenticatedAction().async { implicit request =>
     type Attempt[A] = EitherT[Future, String, A]
     implicit val settings: Settings = settingsProvider.settings()
-    val visitToken = VisitToken(new SecureRandom())
```
johnduffell: this token was generated server side, but suddenly we remembered about caching.  We can trust the client side more to produce a reliable random number in that case.

PR #1052, opened 2018-10-18
Ensure that the email is available in the state on the new thank you page 

jranks123:
## Why are you doing this?

We want to guarantee that the email will be available on the thank you page, so that we can use it to set marketing permissions. 

This PR:

. sets a session cookie as the user is typing in their email
. Initialises `state.page.form.formData.email` with the value from the session, if available

This means that even if the user refreshes the thank you page, or if they had to get there via the one off paypal site, the email will be available in the state.

### DIFF
--- a/assets/pages/new-contributions-landing/contributionsLandingActions.js
+++ b/assets/pages/new-contributions-landing/contributionsLandingActions.js
@@ -69,6 +69,9 @@ const updateFirstName = (firstName: string): Action => ({ type: 'UPDATE_FIRST_NA
 const updateLastName = (lastName: string): Action => ({ type: 'UPDATE_LAST_NAME', lastName });
 
 const updateEmail = (email: string): Action => {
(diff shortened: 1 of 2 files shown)


assets/pages/new-contributions-landing/contributionsLandingActions.js
```diff
@@ -68,7 +68,10 @@ const updateFirstName = (firstName: string): Action => ({ type: 'UPDATE_FIRST_NA
 
 const updateLastName = (lastName: string): Action => ({ type: 'UPDATE_LAST_NAME', lastName });
 
-const updateEmail = (email: string): Action => ({ type: 'UPDATE_EMAIL', email });
+const updateEmail = (email: string): Action => {
+  storage.setSession('gu.email', email);
```
joelochlann: can we have a comment here explaining why we're doing this?

PR #1053, opened 2018-10-18
Set New Payment Flow test participation correctly

joelochlann:
The NPF test will be done as a serverside test which is not normally how we run tests.

One option for setting the A/B test participations clientside (so we attribute conversions correctly) is what I've done here. We simply have a test with a single variant on the new and old contributions landing pages. Anyone arriving on either page will get put in the correct variant. This makes it very much a one-off solution tied to our use-case for this particular test.

There might be some edge cases with this that I haven't thought of. Once a browser has a participation set in localStorage they will stay in it for subsequent visits with that browser. As far as I'm aware that mirrors the behaviour of the serverside test (a cookie would be set putting them in that variant as long as the cookie survives). But I need to double-check that.

An alternative would be to write out our serverside participations into the `window.guardian` object. This might be better if we think we'll do serverside tests in future.

Not too many sunk costs (only spent about half an hour on this solution) so be honest about what you think is better! An additional consideration is of course our Wednesday cut-off date for running the test in time to use the new payment flow during the 1 Million campaign.

## Note about the serverside test
I'm envisioning that when we switch the test on, we will have just a single URL (https://support.theguardian.com/contribute) which will serve either the old or the new page.


assets/helpers/abTests/abtestDefinitions.js
```diff
@@ -18,4 +19,39 @@ export const tests: Tests = {
...
   },
+  newPaymentFlow: {
+    variants: ['control'],
+    audiences: {
+      ALL: {
+        offset: 0,
+        size: 1,
+      },
+    },
+    // TODO: turn to false
```
johnduffell: when?

joelochlann: before I merge

joelochlann: it was for my testing

jranks123: probably before merging, if we are to go ahead with this. Then when we launch the test we have a another PR switching it to true.

assets/helpers/abTests/abtestDefinitions.js
```diff
@@ -18,4 +19,39 @@ export const tests: Tests = {
     independent: true,
     seed: 3,
   },
+  newPaymentFlow: {
```
jranks123: won't you need to add an `oldPaymentFlow` test too?

joelochlann: no, one test (`newPaymentFlow`), two variants (`control`, `newPaymentFlow`)

assets/helpers/abTests/abtestDefinitions.js
```diff
@@ -18,4 +19,39 @@ export const tests: Tests = {
...
+      },
+    },
+    // TODO: turn to false
+    isActive: true,
+    independent: true,
+    seed: 5,
+  },
+};
+
+export const allTests: Tests = {
```
joelochlann: whoops, I've just realised this is a bad idea - people landing on any other page on the site would always get put in the `newPaymentFlow` variant (since it overwrites the first) and then be stuck in that even if they were in the control via the serverside test

assets/helpers/abTests/abtestDefinitions.js
```diff
@@ -18,4 +20,40 @@ export const tests: Tests = {
...
+  newPaymentFlow: {
+    // 100% of people will be put in this variant
+    variants: ['control'],
+    audiences: {
+      ALL: {
+        offset: 0,
+        size: 1,
+      },
+    },
+    isActive: false,
```
joelochlann: we can switch these on next week

joelochlann:
Think this can be merged tomorrow if you're OK with - or you can wait for me on Monday

assets/pages/contributions-landing/contributionsLanding.jsx
```diff
@@ -20,7 +19,11 @@ import { createPageReducerFor } from './contributionsLandingReducer';
 
 const countryGroupId: CountryGroupId = detect();
 
-const store = pageInit(createPageReducerFor(countryGroupId));
+const store = pageInit(
+  createPageReducerFor(countryGroupId),
+  false,
```
jranks123: I'm not a fan of passing through pure boolean values to functions, and they feel especially weird in the middle. 

I know the alternatives to this are neither bountiful nor particularly desirable, but maybe you could at least put the boolean to the end?

I don't know, this might be a weird preference I have that no one else has, and feel completely free to ignore.

johnduffell: does JS allow named parameters? that would help.

joelochlann: agreed. JS doesn't have named parameters, except through the "trick" of passing an object instead (you can do object destructuring inside the actual parameter list to get individual variables inside the function body)

going to leave this one since re-ordering args to `pageInit` would affect a lot of places

PR #1055, opened 2018-10-18
load stripe once not thrice

tsop14:
## Why are you doing this?
So the stripe library is loaded once 😎

## Changes

* Removes `loadStripe` from loop

### DIFF
--- a/assets/helpers/paymentIntegrations/newPaymentFlow/stripeCheckout.js
+++ b/assets/helpers/paymentIntegrations/newPaymentFlow/stripeCheckout.js
@@ -60,24 +60,22 @@ function setupStripeCheckout(
   contributionType: Contrib,
   currency: IsoCurrency,
   isTestUser: boolean,
-): Promise<Object> {
-  return loadStripe().then(() => {
-    const handleToken = (token) => {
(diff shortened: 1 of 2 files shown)


assets/helpers/paymentIntegrations/newPaymentFlow/stripeCheckout.js
```diff
@@ -60,24 +60,23 @@ function setupStripeCheckout(
...
-      onPaymentAuthorisation({ paymentMethod: 'Stripe', token: token.id });
-    };
+): Object {
+  const handleToken = (token) => {
+    onPaymentAuthorisation({ paymentMethod: 'Stripe', token: token.id });
+  };
 
-    const stripeKey = getStripeKey(contributionType, currency, isTestUser);
+  const stripeKey = getStripeKey(contributionType, currency, isTestUser);
+  console.log('stripe Key', stripeKey);
```
tomrf1: can this be removed?

tsop14: Should definitely be removed!

PR #1056, opened 2018-10-19
Enable marketing subscription on the new thank you page

tomrf1:
## Why are you doing this?
Enables the 'Sign me up' button on the new thank you page.
Uses the same `confirmOptIn` state as the existing implementation, but in a new scope.

[**Trello Card**](https://trello.com/c/vKWRd715/11-npf-thank-you-page-wire-up-marketing-permissions)

## Screenshots
picture 4
picture 6

### DIFF
--- a/assets/pages/new-contributions-landing/contributionsLanding.scss
+++ b/assets/pages/new-contributions-landing/contributionsLanding.scss
@@ -903,7 +903,7 @@ form {
 }
 
 .button--newsletter {
-  color: #ffffff;
+  color: gu-colour(garnett-neutral-5);
   background-color: #22834D;
 
   &:enabled {
--- a/assets/pages/new-contributions-landing/contributionsLanding.scss
+++ b/assets/pages/new-contributions-landing/contributionsLanding.scss
@@ -908,6 +908,10 @@ form {
 
   &:enabled {
     cursor: pointer;
+
+    &:hover {
+      background-color: #206B44;
+    }
   }
 }
 
--- a/assets/pages/new-contributions-landing/contributionsLanding.scss
+++ b/assets/pages/new-contributions-landing/contributionsLanding.scss
@@ -904,19 +904,19 @@ form {
 
 .button--newsletter {
   color: gu-colour(garnett-neutral-5);
-  background-color: #22834D;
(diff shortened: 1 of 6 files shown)


jranks123:
My only concern is that we aren't telling the user to check their email, as they won't receive the emails until they confirm via the email.

On the current flow, we say `We'll be in touch. Check your inbox for a confirmation link.`. Could we change the `You can stop these at any time` copy to  `We'll be in touch. Check your inbox for a confirmation link.` when the user clicks on the button?

assets/pages/new-contributions-landing/contributionsLanding.scss
```diff
@@ -903,10 +903,20 @@ form {
 }
 
 .button--newsletter {
-  color: #29a05f;
+  color: #ffffff;
```
jranks123: we should use ` gu-colour(garnett-neutral-5);`

assets/pages/new-contributions-landing/contributionsLanding.scss
```diff
@@ -903,10 +903,20 @@ form {
...
+  background-color: #22834D;
```
jranks123: where did you get this colour from? If it is a colour from the guardian palette then we should add it to `colours.scss` and reference it

assets/pages/new-contributions-landing/contributionsLandingReducer.js
```diff
@@ -99,7 +102,7 @@ function createFormReducer(countryGroupId: CountryGroupId) {
   // ----- Initial state ----- //
 
   const initialState: FormState = {
-    contributionType: 'MONTHLY',
```
jranks123: You need to change this back to `MONTHLY`

assets/pages/new-contributions-landing/contributionsLanding.scss
tomrf1: from the link in the trello card

assets/pages/new-contributions-landing/contributionsLandingReducer.js
tomrf1: oops!

PR #1058, opened 2018-10-19
NPF: Post payment flows

jranks123:
## Why are you doing this?
This PR implements all the different post payment flows that can occur after a payment. 

For one off contributions, everyone hits this page after contributing:

| One-off contributions thank you page |
|image|


assets/pages/new-contributions-landing/components/SetPassword/SetPassword.jsx
```diff
@@ -2,153 +2,59 @@
...
-
 // ----- Render ----- //
 
 function SetPassword(props: PropTypes) {
+  if (props.paymentMethod === 'DirectDebit') {
+    return (
+      <div className="set-password__content">
+        <h1 className="header">Thank you for a valuable contribution. Your Direct Debit has been set up.</h1>
+        <section className="confirmation">
+          <p className="confirmation__message">{'Look out for an email confirming your recurring payment. This will appear as \'Guardian Media Group\' on your bank statements.'}</p>
```
tsop14:
```suggestion
          <p className="confirmation__message">{'Look out for an email within three business days confirming your recurring payment . This will appear as \'Guardian Media Group\' on your bank statements.'}</p>
```

assets/pages/new-contributions-landing/components/ButtonWithRightArrow.jsx
```diff
@@ -11,6 +11,7 @@ type PropTypes = {
   type: string,
   accessibilityHintId: string,
   buttonCopy: string,
+  url?: ?string,
```
joelochlann: Do you need to allow both `null` and the property being omitted?

assets/pages/new-contributions-landing/components/ContributionThankYouContainer.jsx
```diff
@@ -2,38 +2,40 @@
...
-
-
 // ----- Render ----- //
 
 function ContributionThankYouContainer(props: PropTypes) {
 
+  const thankYouPageStage: ThankYouPageStageMap<React$Element<*>> = {
+    thankYouSetPassword: (<ContributionThankYouSetPassword />),
+    thankYou: (<ContributionThankYou showDirectDebitCopy={props.paymentMethod === 'DirectDebit'} />),
+    thankYouPasswordDeclinedToSet: (<ContributionThankYou showDirectDebitCopy={false} />),
```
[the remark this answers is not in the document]
jranks123: Yes I agree it's not very clear. The reason is that if a Direct Debit user hits the thank you page having come from the setPassword screen then we don't want to show the Direct Debit copy that we have to show to be compliant. 

On reflection, I think it might be clearer to showDirectDebitCopy to default to false. But i'll also add a comment into `ContributionThankYou.jsx` explaining what situations we would expect to show the copy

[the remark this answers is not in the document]
joelochlann: > The reason is that if a Direct Debit user hits the thank you page having come from the setPassword screen then we don't want to show the Direct Debit copy that we have to show to be compliant.

Is that because it's already been shown on the setPassword screen?

assets/pages/new-contributions-landing/components/ContributionThankYouPasswordSet.jsx
```diff
@@ -0,0 +1,35 @@
...
+          preferences.
+        </p>
+      </section>
+      <MarketingConsent />
+      <ButtonWithRightArrow
+        componentClassName={classNameWithModifiers('confirmation', ['backtothegu'])}
+        buttonClassName={classNameWithModifiers('button', ['wob'])}
+        accessibilityHintId="accessibility-hint-return-to-guardian"
+        type="button"
+        href="https://www.theguardian.com"
```
tomrf1: `url`

assets/pages/new-contributions-landing/components/ContributionThankYouContainer.jsx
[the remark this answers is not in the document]
jranks123: Yes I agree it's not very clear. The reason is that if a Direct Debit user hits the thank you page having come from the setPassword screen then we don't want to show the Direct Debit copy that we have to show to be compliant. 

On reflection, I think it might be clearer to showDirectDebitCopy to default to false. But i'll also add a comment into `ContributionThankYou.jsx` explaining what situations we would expect to show the copy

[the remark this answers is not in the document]
jranks123: Yes, the setPassword screen is the first "thank you" page that they see for those who get offered the chance to set a password. We need to show the copy on the first page they see, which will be `setPassword` or `thankYou`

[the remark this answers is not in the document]
jranks123: I've added a comment now in `ContributionThankyou.jsx` in the `PropTypes` definition

assets/pages/new-contributions-landing/contributionsLanding.scss
```diff
@@ -170,7 +170,7 @@ body {
   color: gu-colour(garnett-neutral-1);
   font: bold 36px/1.15 $gu-headline;
   margin: 0 -10px;
-  padding: 0 10px $gu-v-spacing;
+  padding: 0 10px $gu-v-spacing*1.5 10px;
```
joelochlann: could use `$gu-h-spacing/2` for the left & right

assets/pages/new-contributions-landing/components/ContributionThankYou.jsx
```diff
@@ -4,50 +4,66 @@
...
-import MarketingConsent from '../components/marketingConsent';
+import MarketingConsent from '../components/MarketingConsent';
+import { ButtonWithRightArrow } from '../components/ButtonWithRightArrow';
 
 // ----- Types ----- //
 
 /* eslint-disable react/no-unused-prop-types */
 type PropTypes = {
   contributionType: Contrib,
+  // wil be true when this is the first thank you page a Direct Debit recurring contributor sees
```
joelochlann: will

joelochlann: I'm still struggling to how this logic works. where's the bit that decides it's the first thank you page seen? I only see

```jsx
thankYou: (<ContributionThankYou showDirectDebitCopy={props.paymentMethod === 'DirectDebit'} />),
```

assets/pages/new-contributions-landing/contributionsLanding.scss
joelochlann: could use`$gu-h-spacing/2`

assets/pages/new-contributions-landing/components/ContributionThankYou.jsx
```diff
@@ -2,52 +2,77 @@
...
+}
+
 // ----- Render ----- //
 
 function ContributionThankYou(props: PropTypes) {
+
+  let directDebitHeaderSuffix = '';
+  let directDebitMessageSuffix = '';
+
+  if (props.paymentMethod === 'DirectDebit' && !props.hasSeenDirectDebitThankYouCopy) {
```
joelochlann: ❤️ 
so much clearer

assets/pages/new-contributions-landing/components/ContributionThankYouContainer.jsx
```diff
@@ -2,38 +2,40 @@
...
-  thankYouPasswordNotSet: (<ContributionThankYou />),
-};
-
-
 // ----- Render ----- //
 
 function ContributionThankYouContainer(props: PropTypes) {
 
+  const thankYouPageStage: ThankYouPageStageMap<React$Element<*>> = {
+    thankYou: (<ContributionThankYou showDirectDebitCopy={props.paymentMethod === 'DirectDebit'} />),
```
joelochlann: oh, I think we can get rid of `showDirectDebitCopy={props.paymentMethod === 'DirectDebit'}` now, no?

joelochlann: ... and therefore also remove `paymentMethod` from the props here

jranks123: ah yes good point. 

This has made me think that we should define PropTypes as an [exact object type](https://flow.org/en/docs/types/objects/#toc-exact-object-types) as convention. We already do this in some places, not all. If it had been, flow would have caught this. I will do this in a separate PR.

PR #1059, opened 2018-10-19
Display error message if signup fails

tomrf1:
## Why are you doing this?
If there is a server-side error after a user enters a password, it needs to display an error message in the same page.

[**Trello Card**](https://trello.com/c/Rl2eyHuG/103-npf-show-error-message-if-set-password-fails)

## Screenshots
picture 8


tomrf1:
Now re-implemented after rebasing against master with @jranks123 's changes https://github.com/guardian/support-frontend/pull/1058

PR #1062, opened 2018-10-22
Countdown clock component + implementation

walaura:
## Why are you doing this?
Adds a generic unstyled countdown clock component to be used in the flash sale (as soon as we get designs for it) as well as an implementation in redux for `digitalSubscriptionsLanding`

## Screenshots
N/A


assets/components/countdown/countdown.jsx
```diff
@@ -0,0 +1,89 @@
+// @flow
+
+// ----- Imports ----- //
+import React, { Component } from 'react';
+
+
+// ---- Types ----- //
+type CountdownTime = {
+  t: number,
```
svillafe: What would be `t`?

walaura: the unix time remaining. Should I make the variable clearer?

svillafe: if you could give it a more meaningful name would be great 👍

assets/components/countdown/countdown.jsx
```diff
@@ -0,0 +1,89 @@
...
+  const days = Math.floor(t / (1000 * 60 * 60 * 24));
+
+  return {
+    t, seconds, minutes, hours, days,
+  };
+};
+
+
+// ----- Component ----- //
+export default class Countdown extends Component<PropTypes, StateTypes> {
```
svillafe: We used to avoid stateful components, maybe @Ap0c has an updated point of view about this.

assets/components/countdown/countdown.jsx
```diff
@@ -0,0 +1,89 @@
...
+type PropTypes = {
+  date: number[]
+};
+
+type StateTypes = {
+  countdown: CountdownTime
+}
+
+
+// ---- Helpers ----- //
```
svillafe: Just a small detail here, there is a [global folder for helpers](https://github.com/guardian/support-frontend/tree/master/assets/helpers), if you think that these helpers could be reuse in some other places maybe is better to place them inside the global folder. If these helpers are really specific to this component, it is ok to leave them here. It is up to you.

walaura:
`Date` can be either initialised with a string (unsafe according to mdn), a unix timestamp (which is not timezone agnostic) or a list of parameters forming a date, which fits the bill. A wrapper component could read from the config and apply the props to this component but it makes no sense to tangle a generic countdown with the page config

joelochlann:
Nice work @walaura!

Sorry for being late to the party after you've already re-factored away from stateful component. But for such a self-contained and purely presentational component I'd argue against using redux here.

Here are some guidelines from [this page in redux docs](https://redux.js.org/faq/organizingstate#do-i-have-to-put-all-my-state-into-redux-should-i-ever-use-reacts-setstate):

> Some common rules of thumb for determining what kind of data should be put into Redux:
> 
> - Do other parts of the application care about this data?
> - Do you need to be able to create further derived data based on this original data?
> - Is the same data being used to drive multiple components?
> - Is there value to you in being able to restore this state to a given point in time (ie, time travel debugging)?
> - Do you want to cache the data (ie, use what's in state if it's already there instead of re-requesting it)?

Since as far as I can tell the answer to all of these questions is no in this particular case, it seems to me we shouldn't put it into redux. As you alluded to on Hangouts chat, I think this becomes even clearer when you look at the difference between your original PR (a single self-contained file) and what it looks like now. To my mind, it was far easier to understand before.

I know @Ap0c and @rupertbates have some opinions on this too and this isn't Contributions team work so I'll defer to opinions of those on S&C. But just wanted to add my two cents!

svillafe:
Is it possible to add unit tests for the helpers functions, actions and reducer?

PR #1015, opened 2018-10-05
add contribution survey to thank you pages

johnduffell:
## Why are you doing this?

Add new survey for people after the ycontributed.  This is needed to get plenty of testimonials to understand the motivations of contributors.

This is based pretty much character for character on https://github.com/guardian/support-frontend/pull/829  it is bascially just a repeat of the original with a new link.

[**Trello Card**](https://trello.com/c/XTSgKu0q/23-testimonials-survey-ask-2018)

## Screenshots

recurring pending
image

direct debit
image

one off
image

@ionamckendrick @jranks123 @joelochlann

PR #1064, opened 2018-10-22
Remove survey from thank you page

tomrf1:
## Why are you doing this?
This survey is no longer needed.
This branch can be reverted if the survey is needed again in future.

## Changes

* Change 1
* Change 2

#### Before:
picture 11

#### After:
picture 10


.eslintrc
```diff
@@ -36,6 +36,7 @@
     "import/no-extraneous-dependencies": "off",
     "padded-blocks": "off",
     "import/prefer-default-export": "off",
+    "no-else-return": 0,
```
johnduffell: can we make that "only one return statement on any code path" if it exists instead?  I do agree there should be as few wasys as possible to do something, but I also do disagree with this way round (but maybe it's the only option in eslint)

assets/components/contributionsThankYou/contributionsThankYouPage.jsx
```diff
@@ -52,18 +51,16 @@ export default function ContributionsThankYouPage(props: PropTypes) {
...
-  if (props.directDebit) {
-    return (
-      <div className="component-direct-debit-details__container">
-        <DirectDebitDetails {...props.directDebit} />
-        <ContributionsSurveySection />
-      </div>
-    );
-  }
-  // recurring non-DD
-  return <EmailConfirmation />;
```
johnduffell: https://github.com/guardian/support-frontend/pull/1015/commits
I was having fun with the same file only the other week

assets/components/contributionsThankYou/contributionsThankYouPage.jsx
```diff
@@ -51,16 +51,14 @@ export default function ContributionsThankYouPage(props: PropTypes) {
...
-      return <EmailConfirmation />;
-    }
+  // recurring
+  if (props.directDebit) {
+    return (
+      <div className="component-direct-debit-details__container">
+        <DirectDebitDetails {...props.directDebit} />
+      </div>
+    );
   }
```
tomrf1: I would prefer an `else` block here, but the linter doesn't like it

assets/components/contributionsThankYou/contributionsThankYouPage.jsx
```diff
@@ -52,15 +51,11 @@ export default function ContributionsThankYouPage(props: PropTypes) {
 // ----- Auxiliary Components ----- //
 
 function BodyCopy(props: PropTypes) {
```
tsop14: could this have a clearer name? `RecurringContribBodyCopy`

PR #1067, opened 2018-10-23
Serverside test for new payment flow

joelochlann:
- `/$country<(uk|us|au|eu|int|nz|ca)>/contribute` => old or new flow
- `/$country<(uk|us|au|eu|int|nz|ca)>/contribute.new` => new flow
- `/$country<(uk|us|au|eu|int|nz|ca)>/contribute.old` => old flow

Since the top route is now uncached, I've added very basic serverside A/B test functionality to the backend. It only supports a single concurrent test, to 100% of the audience, in a 50/50 split. Based on a pseudorandom coin toss, it sets a cookie called `gu.serverside.ab.test` whose value is either `Control` or `Variant`. That's about it.

So the top route will serve the new flow if the request has a `gu.serverside.ab.test` cookie with a value of `Variant`, or if the request is placed in that variant (in which the cookie will be set in the response).

As a result I've removed `segment` from the `ExperimentSwitch` type. So these are now just on/off switches, basically. (This switching functionality is still very useful for testing locally).

**BEFORE THIS PR IS MERGED** I need to change the `experiments` part of `settings.json` in S3 from this:
```
    "experiments": {
      "newPaymentFlow": {
        "name": "newPaymentFlow",
        "description": "Redesign of the payment flow UI",
        "segment": "Perc0",
        "state": "On"
      }
    },
```
to this:

```
    "experiments": {
      "newPaymentFlow": {
        "name": "newPaymentFlow",
        "description": "Redesign of the payment flow UI",
        "state": "Off"
      }
    },
```

so the test is not turned on by this PR. Then when we do release the test, we just change the state to "On"


joelochlann:
Based on my exploration of our Fastly VCL snippets, it looks like we've got everything in place for a 50% serverside test. It works like this:

### recv
Sets (or appends to) an `X-GU-Experiment` header to the request. If there is a `GU_Experiment` cookie it will use the value of that, otherwise it will be set to `Perc50-control` or `Perc50-variant` based on a 50/50 random choice.

### fetch
After getting a response from the backend, it sets (or appends to) the `Vary` header on this response, adding `X-GU-Experiment`

### deliver
- In the response, replaces `X-GU-Experiment` in the `Vary` header with `Cookie`, since the browser has no knowledge of the custom header *(will this split the cache by user, if logged in??)*
- If not already set, sets a `GU_Experiment` cookie on the response to either `Perc50-control` or `Perc50-variant` based on a 50/50 random choice. *(why not just use the X-GU-Experiment header instead of duplicating the logic here?)*

So essentially, a `GU_Experiment` cookie is used to pass A/B test info between Fastly and the client, while an `X-GU-Experiment` header is used to pass A/B test info between Fastly and the backend.

Does my understanding seem correct @regiskuckaertz? My biggest source of confusion is the `Vary` header and its implications for caching. What is the effect of setting `Vary` to `X-GU-Experiment` in the `fetch` snippet, but then changing it to `Cookie` in the `deliver` snippet? (I'm also lacking some VCL/Fastly knowledge here which I may have to ask @davidfurey to help me with)

And relatedly, given that the endpoint for this page has to be a `maybeAuthenticatedAction` (and therefore `Cache-Control: Private`), is there any point bothering with a `Vary` header?

joelochlann:
And I'm not sure the `Vary` stuff is working as intended anyway since I don't see `Cookie` in my `Vary` response headers when requesting

joelochlann:
Based on some discussion with @rupertbates, I'm thinking we could just do this test logic in the Scala app rather than Fastly since with this change the `{countryGroup}/contribute` endpoint will no longer be cached. This would be more intelligible than the VCL.

regiskuckaertz:
I think the reason to do it in Fastly is that it automatically forks it cache for all the variants, so the back-end does not get hit all the time. I can't remember the code snippets but can take a look at it tomorrow.

regiskuckaertz:
So happy to see this btw 🎉 🕺

app/controllers/Application.scala
```diff
@@ -96,16 +96,18 @@ class Application(
...
-    Ok(views.html.main(
-      title = "Support the Guardian | Make a Contribution",
-      description = stringsConfig.contributionsLandingDescription,
-      mainId = s"contributions-landing-page-$countryCode",
-      mainJsBundle = "contributionsLandingPage.js",
-      mainStyleBundle = "contributionsLandingPageStyles.css",
-      scripts = views.html.addToWindow("paymentApiPayPalEndpoint", paymentAPIService.payPalCreatePaymentEndpoint)
-    )).withSettingsSurrogateKey
+    val experiments = settings.switches.experiments
+    if (experiments.get("newPaymentFlow").exists(_.isParticipating)) {
```
johnduffell: this settings.switches thing, how does it have an isParticipating thing, surely participation is a user level thing rather than a switch?

app/controllers/Application.scala
```diff
@@ -96,16 +96,18 @@ class Application(
...
+      request.user.traverse[Attempt, IdUser](identityService.getUser(_)).fold(
+        _ => Ok(newContributions(countryCode, None)),
+        user => Ok(newContributions(countryCode, user))
```
johnduffell: to save duplication we can fold it into an option and then `.map(Ok(newContributions(countryCode, _)))`

johnduffell:
I'm all in favour of doing it all in the backend with no fastly.  Caching is a nice idea but the amount of time we spend trying to get it right and still making mistakes, I would say the prize needs to be big to justify.

joelochlann:
> I think the reason to do it in Fastly is that it automatically forks it cache for all the variants, so the back-end does not get hit all the time. I can't remember the code snippets but can take a look at it tomorrow.

Yep, if the page were cacheable this would be useful. Because this is now a `maybeAuthenticatedAction` which ultimately uses `Cache-Control: Private`, the backend will get hit all the time anyway. This is because the new page needs to fetch user data to pre-populate the form if they're signed in.

app/controllers/Application.scala
joelochlann: It's `ExperimentSwitch` that has an `isParticipating` flag. I guess they're under switches because they have an on/off property like other non-A/B-test switches. Maybe that's confusing but it's following the model of dotcom where A/B tests are on the switchboard along with features.

app/controllers/Application.scala
joelochlann: have added the logging. Wasn't quite sure what the other suggestion was, could you write it out?

app/admin/ServersideAbTest.scala
```diff
@@ -0,0 +1,22 @@
...
+import cookies.ServersideAbTestCookie
+import play.api.mvc.RequestHeader
+
+import scala.util.Random
+
+object ServersideAbTest {
+  // Serverside A/B tests currently only support a single concurrent test
+  // running to 100% audience with a 50%/50% split
+  def getParticipation(implicit request: RequestHeader): Participation = {
+    ServersideAbTestCookie.get.fold(computeParticipation)(_.value match {
```
johnduffell: do we need to flatMap with the match statement first and then getOrElse?  otherwise if the cookie is corrupted somehow we will have a match error or something

johnduffell:
```scala
  def getParticipation(implicit request: RequestHeader): Participation = {
    ServersideAbTestCookie.get.flatMap(_.value match {
      case "Control" => Some(Control)
      case "Variant" => Some(Variant)
      case _ => None
    }).getOrElse(computeParticipation)
  }
```

app/admin/Settings.scala
```diff
@@ -79,22 +79,17 @@ object SettingsSource {
...
-  private def inControl(implicit request: RequestHeader): Boolean = checkHeader(segment, Group.Control)
-
-  def canRun(implicit request: RequestHeader): Boolean = isOn
-  def isParticipating(implicit request: RequestHeader): Boolean = canRun && inVariant
-  def isControl(implicit request: RequestHeader): Boolean = canRun && inControl
-  def value(implicit request: RequestHeader): Group = (isParticipating, isControl) match {
-    case (true, _) => Group.Variant
-    case (_, true) => Group.Control
-    case _ => Group.Unknown
+  def canRun: Boolean = isOn
```
johnduffell: why do we need both isOn and canRun ?

app/admin/Settings.scala
```diff
@@ -79,22 +79,17 @@ object SettingsSource {
...
+  def isInVariant(participation: ServersideAbTest.Participation): Boolean = participation match {
```
johnduffell: I think you can replace the whole function with this
```suggestion
  def isInVariant(participation: ServersideAbTest.Participation): Boolean =
  participation == ServersideAbTest.Variant && canRun
```
same for isInControl.

app/admin/Settings.scala
joelochlann: 😄 good point!

app/admin/ServersideAbTest.scala
joelochlann: yes you're quite right!

app/controllers/Application.scala
johnduffell: I'm not totally sure whether there's an EitherT needs unwrapping somewhere, but something like this
```suggestion
      val maybeUser = request.user.traverse[Attempt, IdUser](identityService.getUser(_)).leftMap(
          logger.error(s"Error fetching user from identity: $err")
      ).toOption
      Ok(newContributions(countryCode, maybeUser).withSettingsSurrogateKey.withServersideAbTestCookie
```

app/controllers/Application.scala
```diff
@@ -29,8 +32,9 @@ class Application(
     payPalConfigProvider: PayPalConfigProvider,
     paymentAPIService: PaymentAPIService,
     stringsConfig: StringsConfig,
-    settingsProvider: SettingsProvider
-)(implicit val ec: ExecutionContext) extends AbstractController(components) with SettingsSurrogateKeySyntax {
+    settingsProvider: SettingsProvider,
+    guardianDomain: GuardianDomain
+)(implicit val ec: ExecutionContext) extends AbstractController(components) with SettingsSurrogateKeySyntax with StrictLogging with ServersideAbTestCookie {
```
johnduffell: `with ServersideAbTestCookie` I'm not keen on using traits to import things without a namespace, you can get exactly the same concept just using an object but without these extra links
```scala
object WithServersideAbTestCookie {
  def apply(blahblah) = {
    .....
  }
}
```

app/cookies/ServersideAbTestCookie.scala
```diff
@@ -0,0 +1,44 @@
...
+  def exists(implicit request: RequestHeader): Boolean = request.cookies.exists(_.name == name)
+  def get(implicit request: RequestHeader): Option[Cookie] = request.cookies.get(name)
+
+  def addTo(result: Result)(implicit domain: GuardianDomain, participation: Participation): Result =
+    result.withCookies(create(domain, participation))
+}
+
+trait ServersideAbTestCookie {
+
+  implicit class ResultSyntax(result: Result)(
```
johnduffell: I also struggle with implicit classes - I do use them reluctantly a bit to move boilerplate to the end of for comprehensions (e.g. we have to lift various things into a EitherT[Future[Either[.]]] as part of a for comprehension) but on the whole, a normal function suffices for almost everything.

jacobwinch:
We will want to exclude the post-deployment test runs from these A/B tests (or at least force them to hit the control). Otherwise these tests will start failing 50% of the time (+ standard PayPal failure rate)!

We already do this using a cookie in the client-side code: https://github.com/guardian/support-frontend/blob/master/assets/helpers/abTests/abtest.js#L158-L160

joelochlann:
Great spot @jacobwinch!

app/cookies/ServersideAbTestCookie.scala
joelochlann: Yeah I'm inclined to agree. Though I'm following the pattern of `.withSettingsSurrogateKey` here and I feel like they should at least be consistent - and I'm not prepared to switch over all the `withSettingsSurrogateKey` usages in this PR. I will come back to this in a future PR

app/controllers/Application.scala
joelochlann: Yeah, again this is how `.withSettingsSurrogateKey` works so I may switch them both over to functions in a future PR. Interested if @Mullefa has any additional arguments in favour of the trait + implicit class style here since he introduced it.

joelochlann:
@jacobwinch I've added another route, `/$country<(uk|us|au|eu|int|nz|ca)>/contribute.old`, and set selenium to run its contribution landing page test against that

app/controllers/Application.scala
johnduffell: agree Guy, some things are conceptually different and some things are just two ways to write the same thing.  Furthermore, some things actually overspecify the relationships between different parts of the code.
I think we should be harshest against the third case, but I agree we should try to improve matters in the middle case, as that is a more human readability parsing issue

app/controllers/Application.scala
```diff
@@ -96,16 +100,26 @@ class Application(
...
-      mainJsBundle = "contributionsLandingPage.js",
-      mainStyleBundle = "contributionsLandingPageStyles.css",
-      scripts = views.html.addToWindow("paymentApiPayPalEndpoint", paymentAPIService.payPalCreatePaymentEndpoint)
-    )).withSettingsSurrogateKey
+    implicit val participation: ServersideAbTest.Participation = ServersideAbTest.getParticipation
+    implicit val gd: GuardianDomain = guardianDomain
+
+    val experiments = settings.switches.experiments
+    if (experiments.get("newPaymentFlow").exists(_.isInVariant(participation))) {
+      logger.info("Serving new contribution landing page")
```
johnduffell:
```suggestion
      logger.info("SERVENEW: Serving new contribution landing page")
```

johnduffell:
looks good to me! if you happen to know how to send the cloudwatch event that would be a nice to have, but either way it's great 👍

app/controllers/Application.scala
Mullefa: nice way of putting it John!

PR #1068, opened 2018-10-23
Enforce exact types for PropTypes as convention

jranks123:
## Why are you doing this?
Unless there is a good reason (for example, you are using object spread syntax to read your props so that you don't have to name them in the component), we should use [exact types](https://flow.org/en/docs/types/objects/#toc-exact-object-types) for our PropTypes. This prevents people deleting props from components but neglecting to stop sending those props to the component.

There are multiple instances where this had infact already happened, see 

- `assets/pages/digital-subscription-landing/components/ctaSwitch.jsx`
- `assets/pages/digital-subscription-landing/components/productBlock.jsx` 
- `assets/pages/new-contributions-landing/components/ContributionThankYouContainer.jsx` 

It also guards against making a typeo when sending an optional prop to a component, and not being notified by flow, which is what I think has happened here (@Ap0c can you confirm this? I've made a fix assuming that this was what had happened):

- `assets/components/priceCta/priceCta.jsx`

So this PR enforces exact types _nearly_ everywhere for PropTypes, and fixes any validity errors that this change shone a light on

### DIFF
--- a/assets/components/contribute/contribute.jsx
+++ b/assets/components/contribute/contribute.jsx
@@ -13,12 +13,12 @@ import type { Node } from 'react';
 
 // ----- Types ----- //
 
-type PropTypes = {
+type PropTypes = {|
   copy: string | Node,
   children: Node,
   heading?: string,
   modifierClasses: Array<?string>,
-};
+|};
 
 
 // ----- Component ----- //
--- a/assets/components/adFreeSection/adFreeSection.jsx
+++ b/assets/components/adFreeSection/adFreeSection.jsx
@@ -9,9 +9,9 @@ import Heading, { type HeadingSize } from 'components/heading/heading';
 
 // ---- Types ----- //
 
-type PropTypes = {
+type PropTypes = {|
   headingSize: HeadingSize,
-};
+|};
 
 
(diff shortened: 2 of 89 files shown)


assets/components/contribute/contribute.jsx
```diff
@@ -13,12 +13,12 @@ import type { Node } from 'react';
 
 // ----- Types ----- //
 
-type PropTypes = {
+type PropTypes = {|
   copy: string | Node,
   children: Node,
   heading?: string,
```
tsop14: Is the exact object type declaration subordinate to the optional property type?

tsop14:
👍  but with a question around how it works with optional properties.

assets/components/contribute/contribute.jsx
jranks123: Yes, in as much as if something is defined as optional then you don't have to pass it through, even if it is an exact type. The basic type ensures that everything that should be there, is there. The exact bolt on ensures that nothing extra is passed through. 

I think this line from the docs sums it up:

`it is not valid to pass an object with “extra” properties to an exact object type.`

PR #1069, opened 2018-10-23
Styled flash sale countdown component

walaura:
## Why are you doing this?
Adds a styled component for the flash sale countdown. It's not implemented anywhere on the site yet, this is just the styles in isolation.

There's also a small fix in `Countdown.jsx` to render `00` instead of `0`

## Screenshots
screenshot 2018-10-23 at 4 13 27 pm

### DIFF
--- a/assets/components/countdown/countdown.jsx
+++ b/assets/components/countdown/countdown.jsx
@@ -75,7 +75,7 @@ export default class Countdown extends Component<PropTypes, StateTypes> {
 
     return (
       <time>
-        {units.map(addLeadingZeros).join(':')}
+        {units.map(unit => addLeadingZeros(unit, 2)).join(':')}
       </time>
     );
   }
--- a/assets/components/countdown/countdown.jsx
+++ b/assets/components/countdown/countdown.jsx
@@ -71,11 +71,16 @@ export default class Countdown extends Component<PropTypes, StateTypes> {
       days, hours, minutes, seconds,
     } = this.state.time;
 
-    const units = days > 0 ? [days, hours, minutes] : [hours, minutes, seconds];
+    const units = days > 0 ? [days, hours, minutes, seconds] : [hours, minutes, seconds];
 
     return (
-      <time>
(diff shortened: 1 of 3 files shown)


assets/components/flashSaleCountdown/flashSaleCountdown.scss
```diff
@@ -0,0 +1,20 @@
...
+  font-size: 18px;
+  padding: 6px 10px 4px;
+  display: inline-block;
+}
+
+.component-flash-sale-countdown__chip--time {
+  background: gu-colour(sport-garnett-feature);
+  color: #fff;
+  font-weight: 800;
+  width: 98px;
```
walaura: This fixed width is set to prevent the counter from shifting sizes slightly when the time changes, as the font doesn't have monospaced numbers.

akash1810: `98px` feels very magical. Could we add a comment for it?

walaura: Haha, i just dropped an inline comment here, let me put it in the CSS itself

assets/components/countdown/countdown.jsx
```diff
@@ -71,11 +71,16 @@ export default class Countdown extends Component<PropTypes, StateTypes> {
...
     } = this.state.time;
 
-    const units = days > 0 ? [days, hours, minutes] : [hours, minutes, seconds];
+    const units = days > 0 ? [days, hours, minutes, seconds] : [hours, minutes, seconds];
 
     return (
-      <time>
-        {units.map(addLeadingZeros).join(':')}
+      <time className="component-countdown">
+        {units.map((unit, index) => (
```
JustinPinner: Can we `units.join(':')` this?

JustinPinner: (instead of the lengthy (no pun intended) `map` operation)

JustinPinner: Stupid jsx :(

PR #1071, opened 2018-10-23
Annual contribution amounts test 3

tomrf1:
## Why are you doing this?
A third test for annual contribution amounts.
All users will see the annual contribution option. There are 2 variants.

Some of the code in `annualContributions.js` is quite hacky to support the old payments flow, which we will hopefully remove soon.

In future we will build a tool to make these tests much easier to configure.

[**Trello Card**](https://trello.com/c/gQMjG2Om/92-amounts-test-annual-contribs-round-3-ad-hoc)

#### Old flow, control
picture 14

#### Old flow, variantA
picture 18

#### New flow, control
picture 13

#### New flow, variantA
picture 17

### DIFF
--- a/assets/helpers/abTests/abtestDefinitions.js
+++ b/assets/helpers/abTests/abtestDefinitions.js
@@ -9,7 +9,7 @@ export type AnnualContributionsTestVariant = 'control' | 'annualAmountsA';
 // to browsers landing on pages/contributions-landing/contributionsLanding.jsx
 export const oldContributionsFlowTests: Tests = {
   annualContributionsRoundThree: {
-    variants: ['annualAmountsA', 'annualAmountsB'],
+    variants: ['control', 'annualAmountsA'],
     audiences: {
       ALL: {
         offset: 0,
--- a/assets/helpers/abTests/abtestDefinitions.js
+++ b/assets/helpers/abTests/abtestDefinitions.js
@@ -3,7 +3,7 @@ import type { Tests } from './abtest';
 
 // ----- Tests ----- //
 
-export type AnnualContributionsTestVariant = 'annualAmountsA' | 'annualAmountsB' | 'notintest';
+export type AnnualContributionsTestVariant = 'control' | 'annualAmountsA' | 'notintest';
 
 // Participations in these tests are only assigned
 // to browsers landing on pages/contributions-landing/contributionsLanding.jsx
@@ -39,7 +39,7 @@ export const oldContributionsFlowTests: Tests = {
 // to browsers landing on pages/new-contributions-landing/contributionsLanding.jsx
 export const newContributionsFlowTests: Tests = {
   annualContributionsRoundThree: {
-    variants: ['annualAmountsA', 'annualAmountsB'],
+    variants: ['control', 'annualAmountsA'],
     audiences: {
       ALL: {
         offset: 0,
--- a/assets/helpers/abTests/abtestDefinitions.js
+++ b/assets/helpers/abTests/abtestDefinitions.js
@@ -3,7 +3,7 @@ import type { Tests } from './abtest';
 
 // ----- Tests ----- //
 
-export type AnnualContributionsTestVariant = 'control' | 'annualAmountsA' | 'notintest';
+export type AnnualContributionsTestVariant = 'control' | 'annualAmountsA';
 
 // Participations in these tests are only assigned
 // to browsers landing on pages/contributions-landing/contributionsLanding.jsx
--- a/assets/helpers/abTests/abtestDefinitions.js
+++ b/assets/helpers/abTests/abtestDefinitions.js
@@ -3,7 +3,7 @@ import type { Tests } from './abtest';
 
 // ----- Tests ----- //
 
-export type AnnualContributionsTestVariant = 'control' | 'annualAmountsA';
+export type AnnualContributionsTestVariant = 'control' | 'annualAmountsA' | 'notintest';
 
 // Participations in these tests are only assigned
 // to browsers landing on pages/contributions-landing/contributionsLanding.jsx
--- a/assets/helpers/abTests/abtestDefinitions.js
+++ b/assets/helpers/abTests/abtestDefinitions.js
@@ -3,13 +3,13 @@ import type { Tests } from './abtest';
 
 // ----- Tests ----- //
 
-export type AnnualContributionsTestVariant = 'control' | 'annual' | 'annualHigherAmounts' | 'notintest';
+export type AnnualContributionsTestVariant = 'annualAmountsA' | 'annualAmountsB' | 'notintest';
 
 // Participations in these tests are only assigned
 // to browsers landing on pages/contributions-landing/contributionsLanding.jsx
 export const oldContributionsFlowTests: Tests = {
-  annualContributionsRoundTwo: {
-    variants: ['control', 'annual', 'annualHigherAmounts'],
+  annualContributionsRoundThree: {
+    variants: ['annualAmountsA', 'annualAmountsB'],
     audiences: {
       ALL: {
         offset: 0,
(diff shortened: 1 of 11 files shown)


assets/helpers/abTests/helpers/annualContributions.js
```diff
@@ -2,10 +2,130 @@
...
 import { type Store } from 'redux';
 import { contributionSelectionActionsFor } from 'components/contributionSelection/contributionSelectionActions';
+import type { AnnualContributionsTestVariant } from '../abtestDefinitions';
 
-export function setInitialAmountsForAnnualHigherAmountsVariant(store: Store<*, *, *>) {
-  const annualTestVariant = store.getState().common.abParticipations.annualContributionsRoundTwo;
-  if (annualTestVariant === 'annualHigherAmounts') {
-    store.dispatch(contributionSelectionActionsFor('CONTRIBUTE_SECTION').setAmountForContributionType('ANNUAL', 100));
+/* eslint-disable quote-props */
+const numbersInWords = {
```
tomrf1: Largely copied from `contributions.js` to avoid circular dependencies

assets/helpers/abTests/helpers/annualContributions.js
```diff
@@ -2,10 +2,130 @@
...
+  '250': 'two hundred and fifty',
+  '300': 'three hundred',
+  '500': 'five hundred',
+  '750': 'seven hundred and fifty',
+};
+/* eslint-enable  quote-props */
+
+const VariantA = {
+  defaults: {
+    GBPCountries: '100',
```
tomrf1: the old payment flow logic does not use the `isDefault` field, so this lookup is necessary

jranks123:
Before I look at this, I just wanna say that you should double check as to why the control is different from what the current annual control was before, as it is possible that this is a mistake

tomrf1:
@jranks123 from Jesse - "the control in the next round is made up of the most promising variants in the previous round"

assets/helpers/abTests/helpers/annualContributions.js
tomrf1: also, in future we can write something to auto-generate these strings

assets/helpers/contributions.js
```diff
@@ -498,7 +454,7 @@ function getContributionTypeRadios(
     accessibilityHint: 'Make a regular annual contribution',
   };
 
-  return annualTestVariant === 'annual' || annualTestVariant === 'annualHigherAmounts'
+  return annualTestVariant === 'control' || annualTestVariant === 'annualAmountsA'
```
joelochlann: Since this test has just two variants, does this mean that with this test, Annual Contributions will be present for everyone, regardless of test participation?

Because there's actually a ticket to make Annual Contributions always present on the old flow, before launching the new flow test (because we don't want to compare the new flow with Annual vs the old flow half the time with Annual and half the time without). https://trello.com/c/lWta59KJ/108-npf-annual-contributions-to-100-audience-on-existing-flow.

So it would be a good thing if this is achieved as part of this PR! But we could probably make a few tweaks so that fact is clearer, e.g. remove this ternary condition if it can never evaluate to the right-hand side

assets/pages/contributions-landing/pagesVersions/horizontalLayoutLandingPage.jsx
```diff
@@ -78,8 +78,8 @@ const CountrySwitcherHeader = CountrySwitcherHeaderContainer(
 
 const HorizontalLayoutLandingPage: (PropTypes) => React.Node = (props: PropTypes) => {
   const annualContributeCopy = 'Make a recurring commitment to support The Guardian long term or a single contribution as and when you feel like it – choose the option that suits you best.';
-  const annualTestVariant = props.store && props.store.getState().common.abParticipations.annualContributionsRoundTwo;
-  const copyText = annualTestVariant === 'annual' || annualTestVariant === 'annualHigherAmounts'
+  const annualTestVariant = props.store && props.store.getState().common.abParticipations.annualContributionsRoundThree;
+  const copyText = annualTestVariant === 'control' || annualTestVariant === 'annualAmountsA'
```
joelochlann: again, this ternary operator feels a bit misleading now since if you have knowledge of the test (two variants, 100% of the audience) you know that in practice the condition will always evaluate to true

joelochlann:
Could we get screenshots of both the control and the variant please?

assets/helpers/contributions.js
tomrf1: yes, all users will see the annual button.
If we're not going back, and don't need `notintest`, then you're right I can simplify some logic here

joelochlann: Yes, the model is now: we always have three contribution types (one-off, monthly, annual) which have standard amounts, and we can override the amounts if the user is in some test variant. I **think** (though you could check with Jesse) that you can assume that for any amounts test the control has the standard amounts. It may be that after a successful test we update these standard amounts, affecting both the normal site behaviour and the control for future tests (as in this case). But I don't think we would ever have different control amounts for different tests running at the same time, or have a control which deviated from the normal site behaviour. So amounts tests are always about overriding amounts in the variants.

Anything which makes that clearer in code is helpful. But doesn't necessarily have to be in this PR

assets/pages/new-contributions-landing/contributionsLandingReducer.js
```diff
@@ -86,9 +86,9 @@ export type State = {
 
 function createFormReducer(countryGroupId: CountryGroupId) {
   const amountsForCountry: { [Contrib]: Amount[] } = {
-    ONE_OFF: amounts('notintest').ONE_OFF[countryGroupId],
-    MONTHLY: amounts('notintest').MONTHLY[countryGroupId],
-    ANNUAL: amounts('notintest').ANNUAL[countryGroupId],
+    ONE_OFF: amounts('control').ONE_OFF[countryGroupId],
```
tomrf1: initialise here with the control - this will be updated by `ContributionAmount.jsx` when the variant name is available

tomrf1:
I've added more screenshots and got the a/b test working properly with the new flow now

assets/pages/new-contributions-landing/contributionsLandingReducer.js
jranks123: why did you need to do this?

tomrf1: the form needs to be initialised with something. I've just changed the name of the variant used for the initialisation

tomrf1: I could make the parameter here optional (defaulting to `control`), so that we don't have to pass it in at this point

jranks123: what was wrong with notintest?

tomrf1: ah, I removed it from the list of valid values so it had to be replaced here

assets/helpers/abTests/abtestDefinitions.js
```diff
@@ -38,6 +38,18 @@ export const oldContributionsFlowTests: Tests = {
 // Participations in these tests are only assigned
 // to browsers landing on pages/new-contributions-landing/contributionsLanding.jsx
 export const newContributionsFlowTests: Tests = {
+  annualContributionsRoundThree: {
+    variants: ['control', 'annualAmountsA'],
```
jranks123: how come the variants here are 'control' and 'annaulAmountsA' yet above it's 'annualAmountsA' and 'annualAmountsB'?

assets/pages/new-contributions-landing/contributionsLandingReducer.js
jranks123: hmmmm. I'm pretty sure that the abtest framework can still assign people notintest though, theoretically no? Why did you have to remove it in this scenario?

assets/helpers/abTests/abtestDefinitions.js
tomrf1: ooh, good spot. Happened during the rebase when the tests were broken out into old/new

assets/pages/new-contributions-landing/contributionsLandingReducer.js
tomrf1: ok, I didn't realise `notintest` is a special value. All users will now see the annual contribution option, so I thought it made sense to default to `control`. I can add `notintest` back in if it's better? (and default to the same settings as the `control`)

jranks123: Yes, our usual pattern is to name the variants in the test as control + whatever your variant names are, then change the default behaviour for the variants. I.e you shouldn't really be doing something like `if (variantName === 'control') {...` (which I don't think you are doing). But you should be aware that if the ab testing framework fails to assign someone (usually due to a location constraint or something) then it assigns them as 'notintest'. So it's probably best  to put it in there, even though in this case everyone will be in one test or the other

tomrf1: ok, thanks!

PR #1072, opened 2018-10-23
thank you button fix

tsop14:
## Why are you doing this?
* Fix link to the Gu on thank you page. 

## Changes

* Uses `CtaLink` for link on thank you page
* Removes optional url from `ButtonWithRightArrow`. This elem is used twice on the password setting thank you page. One submits the form if a password is being added, the other uses the onclick function passed through to redirect to the next thank you page. As neither use `url`, this property can be removed. 

screen shot 2018-10-23 at 19 28 36

### DIFF
--- a/assets/pages/new-contributions-landing/components/ButtonWithRightArrow.jsx
+++ b/assets/pages/new-contributions-landing/components/ButtonWithRightArrow.jsx
@@ -20,7 +20,7 @@ function ButtonWithRightArrow(props: PropTypes) {
 
   return (
     <div className={props.componentClassName}>
-      <button
+      <a
         className={props.buttonClassName}
         type={props.type}
         href={props.url}
@@ -30,7 +30,7 @@ function ButtonWithRightArrow(props: PropTypes) {
         {props.buttonCopy}
         <p id={props.accessibilityHintId} className="accessibility-hint">{props.buttonCopy}</p>
         <SvgArrowRight />
-      </button>
+      </a>
     </div>
   );
 }
--- a/assets/pages/new-contributions-landing/components/ButtonWithRightArrow.jsx
+++ b/assets/pages/new-contributions-landing/components/ButtonWithRightArrow.jsx
@@ -11,7 +11,6 @@ type PropTypes = {
   type: string,
   accessibilityHintId: string,
   buttonCopy: string,
-  url: ?string,
   onClick: () => void,
 };
 // ----- Render ----- //
@@ -20,17 +19,16 @@ function ButtonWithRightArrow(props: PropTypes) {
 
   return (
     <div className={props.componentClassName}>
-      <a
+      <button
         className={props.buttonClassName}
         type={props.type}
-        href={props.url}
         aria-describedby={props.accessibilityHintId}
         onClick={props.onClick}
       >
         {props.buttonCopy}
         <p id={props.accessibilityHintId} className="accessibility-hint">{props.buttonCopy}</p>
         <SvgArrowRight />
-      </a>
+      </button>
     </div>
   );
 }
@@ -39,7 +37,6 @@ function ButtonWithRightArrow(props: PropTypes) {
 
 ButtonWithRightArrow.defaultProps = {
   onClick: () => undefined,
-  url: null,
 };
 
 export { ButtonWithRightArrow };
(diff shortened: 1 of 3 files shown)


assets/pages/new-contributions-landing/components/ButtonWithRightArrow.jsx
```diff
@@ -20,7 +20,7 @@ function ButtonWithRightArrow(props: PropTypes) {
 
   return (
     <div className={props.componentClassName}>
-      <button
+      <a
```
walaura: could this have an `aria-role="button"`? Provided it's being used as a button. If it's normally used as a link then it's actually better now!

assets/pages/new-contributions-landing/components/ButtonWithRightArrow.jsx
```diff
@@ -17,20 +17,21 @@ type PropTypes = {
...
-        type={props.type}
-        href={props.url}
-        aria-describedby={props.accessibilityHintId}
-        onClick={props.onClick}
-      >
-        {props.buttonCopy}
-        <p id={props.accessibilityHintId} className="accessibility-hint">{props.buttonCopy}</p>
-        <SvgArrowRight />
-      </button>
+      { props.url ? <a href={props.url}>{buttonElem}</a> : buttonElem }
```
regiskuckaertz: I see what you are doing here, however this is problematic for UAs: `<a><button>...`. I would instead choose either `A` or `BUTTON` and stick to it.

joelochlann: Putting a `<button>` inside an `<a>` isn't actually valid HTML5 so it's behaviour may be unpredictable across browsers (http://w3c.github.io/html/textlevel-semantics.html#the-a-element says there must be no [interactive content](http://w3c.github.io/html/dom.html#interactive-content-2) descendants, which includes `<button>`)

joelochlann: Not to mention the fact that you could pass both an `onClick` to the button and an `href` for the link - not sure what the behaviour would be then!

joelochlann: This component feels a bit overloaded. I think the best option is to create a separate `LinkWithRightArrow` component which takes `url` but not `onClick`. Hopefully you can just apply the class [`button`](https://github.com/guardian/support-frontend/blob/master/assets/pages/new-contributions-landing/contributionsLanding.scss#L927-L937) to it to get the styling, or if not at least re-use most of what's in there

tsop14:
have updated this PR to just deal with broken country select. Will update the button in a separate PR as comments above have suggested a different approach would be better.

assets/pages/new-contributions-landing/components/ButtonWithRightArrow.jsx
tsop14: There is a `ctaLink` that will do nicely for the Thank You page use case. Can then remove the `props.url` .

joelochlann:
I'll accept some blame for the `href` on a `<button>` since I reviewed Jonny's PR #1058 and didn't spot it! 🤕

assets/pages/new-contributions-landing/contributionsLanding.scss
```diff
@@ -917,6 +917,10 @@ form {
   margin-bottom: $gu-v-spacing + 20px;
```
joelochlann: weirdly there's a separate `.confirmation--backtothegu` just above, you could probably merge the two (not directly to do with your PR though)

assets/pages/new-contributions-landing/components/ContributionThankYou.jsx
```diff
@@ -62,14 +61,14 @@ function ContributionThankYou(props: PropTypes) {
...
       <MarketingConsent />
-      <ButtonWithRightArrow
-        componentClassName={classNameWithModifiers('confirmation', ['backtothegu'])}
-        buttonClassName={classNameWithModifiers('button', ['wob'])}
-        accessibilityHintId="accessibility-hint-return-to-guardian"
-        type="button"
-        url="https://www.theguardian.com"
-        buttonCopy="Return to The Guardian&nbsp;"
-      />
+      <div className="confirmation confirmation--backtothegu">
```
joelochlann: I wonder if you could just put the modifier classes directly on the `CtaLink`? then you wouldn't need the direct child selector in the CSS (we tend to want to keep our selectors just flat classes as it makes overriding things easier)

tsop14: Have added `--backtothegu` as class modifier on `component-cta-link` and updated css to set font size on this.

assets/pages/new-contributions-landing/contributionsLanding.scss
tsop14: Merged.

PR #1081, opened 2018-10-25
Timer in showcase pages

walaura:
## Why are you doing this?
Add a sense of urgency to the checkout pages for the flash sale. 

Do not merge! this might be an AB test (need to see if we have the audience) and otherwise, it shouldn't go out until Monday

Related: #1069, #1062

## Screenshots

screenshot 2018-10-24 at 2 13 34 pm
screenshot 2018-10-24 at 2 52 02 pm


LATaylor-guardian:
👍

assets/helpers/flashSale.js
```diff
@@ -85,7 +85,7 @@ function getIntcmp(product: SubscriptionProduct, intcmp: ?string, defaultIntcmp:
   return intcmp || defaultIntcmp;
 }
 
-function getEndTime(): number { return new Date(2018, 9, 24, 14, 30, 0).getTime(); }
+function getEndTime(): number { return new Date(2018, 10, 4, 0, 0, 0).getTime(); }
```
walaura: can we triple check this is correct?

assets/components/countdown/countdown.jsx
```diff
@@ -71,17 +78,29 @@ export default class Countdown extends Component<PropTypes, StateTypes> {
...
     return (
       <time className="component-countdown">
-        {units.map((unit, index) => (
-          <span>
-            <span className="component-countdown__time">{addLeadingZeros(unit, 2)}</span>
-            {index < units.length - 1 && ':'}
+        {days > 0 && chip('days', days)}
+        {chip('hrs', hours)}
+        {chip('mins', minutes)}
+        {chip('secs', seconds)}
```
walaura: In terms of a11y, I checked the [GDS guidance](https://accessibility.blog.gov.uk/2017/02/08/advice-for-creating-content-that-works-well-with-screen-readers/) and they actually encourage _against_ using any type of additional markup for this, just roll with the awkward spellings.

assets/helpers/flashSale.js
rupertbates: Can you use the same date as is used in the `flashSaleIsActive()` function above? Then we'll only need to change the date in one place next time we do this.

walaura: Great point!

assets/helpers/flashSale.js
```diff
@@ -6,13 +6,16 @@ import { getProductPrice } from 'helpers/subscriptions';
 
 import type { SubscriptionProduct } from './subscriptions';
 
+// Get end time
+function getEndTime(): number { return new Date(2018, 10, 5, 0, 0).getTime();; }
```
rupertbates: Can you use the same date as is used in the flashSaleIsActive() function above? Then we'll only need to change the date in one place next time we do this.

walaura: Great point! all set

walaura:
sorry about the weirdness of `63c4899` right at the end, i was messing with my local clock for testing and as github sorting knows it's set 0 days into the future

PR #1082, opened 2018-10-25
Fix the wrapping of the title on the Digital Pack US page

rupertbates:
## Why are you doing this?

I noticed we have a wrapping problem on the US digipack page, this fixes it:
## Before
screen shot 2018-10-25 at 15 05 36
## After
screen shot 2018-10-25 at 15 13 46


walaura:
I think this might be fixed already!! I changed this `height` to `min-height` in my last PR. Defs looks better in 2 lines instead of 3 tho

PR #1083, opened 2018-10-25
Delete unreliable post-deployment test

jacobwinch:
## Why are you doing this?

This test is very unreliable and it wastes a lot of time. 

It fails so regularly that it undermines confidence in the post-deployment tests in general.

We already have fairly sensitive alerting for this functionality (https://github.com/guardian/payment-api/blob/master/src/main/resources/cloud-formation.yaml#L430-L447), so I don't think we introduce too much risk by deleting this test.

We may need to delete the recurring PayPal test too (for the same reasons), but apparently this test is the most flakey, so let's try getting rid of it first.

[**Trello Card**](https://trello.com/c/JXk7kxGp/2021-delete-paypal-one-off-acceptance-test)

## Changes

* Delete unreliable test
* Delete unused PayPal method
* Fix name of remaining one-off scenario

### DIFF
--- a/test/selenium/OneOffContributionsSpec.scala
+++ b/test/selenium/OneOffContributionsSpec.scala
@@ -20,7 +20,7 @@ class OneOffContributionsSpec extends FeatureSpec with GivenWhenThen with Before
 
   feature("Sign up for a one-off contribution") {
 
-    scenario("One-off contribution sign-up with Stripe - GBP") {
+    scenario("One-off contribution sign-up with Stripe - AUD") {
 
       val stripePayment = 22.55
       val currency = "AUD"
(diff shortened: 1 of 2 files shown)


test/selenium/OneOffContributionsSpec.scala
```diff
@@ -20,7 +20,7 @@ class OneOffContributionsSpec extends FeatureSpec with GivenWhenThen with Before
 
   feature("Sign up for a one-off contribution") {
 
-    scenario("One-off contribution sign-up with Stripe - GBP") {
+    scenario("One-off contribution sign-up with Stripe - AUD") {
```
rupertbates: We are happy to leave this one in are we?

Ap0c:
Looks good to me from a code perspective 👍.

test/selenium/OneOffContributionsSpec.scala
jacobwinch: This one should be very reliable (famous last words), as it directly calls the callback rather than interacting with the Stripe UI: https://github.com/guardian/support-frontend/blob/09610f52ec1f416137a30dab23d6dd545d6906ec/assets/components/paymentButtons/stripePopUpButton/stripePopUpButton.jsx#L73-L74

PR #1084, opened 2018-10-25
Remove circles svg from mobile on thank you page

jranks123:
## Why are you doing this?
There was a visual bug whereby the circles were getting in the way of the copy in the header on mobile. To keep things simple (and for expediency), we are going to remove the circles from the header for now.

Before:
image

After:
image

This affects all the pages of the thank you flow (set password and thank you)

### DIFF
--- a/assets/pages/new-contributions-landing/components/ContributionBackground.jsx
+++ b/assets/pages/new-contributions-landing/components/ContributionBackground.jsx
@@ -4,7 +4,6 @@
 
 import React from 'react';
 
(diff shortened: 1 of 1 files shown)


joelochlann:
hang on, doesn't this also remove it from the homepage?

https://github.com/guardian/support-frontend/blob/master/assets/pages/new-contributions-landing/contributionsLanding.jsx#L72

joelochlann:
not that's necessarily a problem

PR #1089, opened 2018-10-25
Rejig the circle-less header on mobile

jranks123:
## Why are you doing this?
This PR hides the header background `div` on mobile, meaning that the body is brought up to the text.

This also makes the header block yellow on mobile, for landing and thank you pages. 

@ionamckendrick what do you think?

|Landing page Before| Landing page after |
|------|------------|
|image|image|

|Thank you page before| Thank you page after |
|------|----------|
|image)|image|

### DIFF
--- a/assets/pages/new-contributions-landing/contributionsLanding.scss
+++ b/assets/pages/new-contributions-landing/contributionsLanding.scss
@@ -239,6 +239,11 @@ header[role="banner"] {
     width: 100%;
   }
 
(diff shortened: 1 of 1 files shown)


tsop14:
👍 for no circles - where have you set the background-color to yellow?

ionamckendrick:
@tsop14 @jranks123 We should *only* use a yellow background on the final thank you page for each flow. All others should be plain white for now.

PR #1092, opened 2018-10-26
Set default annual value in new payment flow

tomrf1:
## Why are you doing this?
Currently the default annual amount is initialised with the control values.
This change updates the selected annual amount once the variant name is known

## Screenshots
control:
picture 1
variant:
picture 2

### DIFF
--- a/assets/helpers/abTests/helpers/annualContributions.js
+++ b/assets/helpers/abTests/helpers/annualContributions.js
@@ -2,7 +2,6 @@
 
 import { type Store } from 'redux';
 import { contributionSelectionActionsFor } from 'components/contributionSelection/contributionSelectionActions';
-import type { AnnualContributionsTestVariant } from '../abtestDefinitions';
 
 /* eslint-disable quote-props */
 const numbersInWords = {
@@ -99,7 +98,7 @@ const VariantA = {
   ],
 };
 
-export const getAnnualAmounts = (annualTestVariant: AnnualContributionsTestVariant) => {
+export const getAnnualAmounts = (annualTestVariant: string) => {
   if (annualTestVariant === 'annualAmountsA') {
     return {
       GBPCountries: VariantA.standard,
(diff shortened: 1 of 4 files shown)


tomrf1:
TODO later - move the setting of the valid amounts into `init` as well, currently it's done [within the component](https://github.com/guardian/support-frontend/blob/master/assets/pages/new-contributions-landing/components/ContributionAmount.jsx#L91)

assets/helpers/abTests/helpers/annualContributions.js
```diff
@@ -99,7 +98,7 @@ const VariantA = {
   ],
 };
 
-export const getAnnualAmounts = (annualTestVariant: AnnualContributionsTestVariant) => {
+export const getAnnualAmounts = (annualTestVariant: string) => {
```
tomrf1: This is far from ideal, but the tests model doesn't make this easy at the moment (everything is just a string).
We can revisit this later
@joelochlann

assets/pages/new-contributions-landing/contributionsLandingInit.js
```diff
@@ -57,12 +59,22 @@ function initialisePaymentMethods(state: State, dispatch: Function) {
   loadPayPalRecurring().then(() => dispatch(setPayPalHasLoaded()));
 }
 
+function initialiseSelectedAnnualAmount(state: State, dispatch: Function) {
+  const { countryGroupId } = state.common.internationalisation;
+  const annualTestVariant = state.common.abParticipations.annualContributionsRoundThree;
```
joelochlann: this should be fine since we know that `annualContributionsRoundThree` is in the test definitions. but for safety let's check on the existence of this property first - in case some, say, removes the test definition and then causes initialisation code to break

assets/helpers/abTests/helpers/annualContributions.js
joelochlann: yeah, what we ultimately need is an `AmountsTest` model and a single function which will get the amounts for all contribution types, based on default amounts overridden by any test participations. but this can wait given current time pressures

PR #1096, opened 2018-10-29
Emails with passwords must log in

joelochlann:
## What works
- On email form field change events, we make a request to check the userType of the email
- We show an element below the email form telling the user whether:
  - request is pending
  - request failed
  - they need to sign in
  - or nothing, if none of the above are true

## Still to do
- [x] prevent the form submitting if any of the following are true: `isSignInRequired`, `isIdentityRequestPending`, `lastIdentityResponse === 'failure'`
- [ ] should we actually wait to send the request until submit is clicked?
- [ ] proper UI design for the `<MustSignIn>` component


jranks123:
` should we actually wait to send the request until submit is clicked?`

No, we have to do it before that in order to be able to invalidate the paypal button. I think we should try and call the API on the `onInput` event for the email input field, but only call the API if the email is valid, meaning there should be only 3 or 4 API calls each time (as `.com` and `.co.uk` will be the most common emails).  We will need to be careful though and make sure that we process each call sequentially, that is we should be wary of, say, the 2nd call to the API returning after the 3rd, which could lead to us to setting `isSignInRequired` to false when it is actually true.

assets/pages/new-contributions-landing/components/ContributionFormFields.jsx
```diff
@@ -92,6 +102,7 @@ function FormFields(props: PropTypes) {
         placeholder="example@domain.com"
         icon={<SvgEnvelope />}
         onInput={props.updateEmail}
+        onChange={props.checkIfEmailHasPassword}
```
jranks123: I would put this in `onInput', and only call the API if the email is a valid email address, for the reasons listed in my general comment on this PR.

assets/pages/new-contributions-landing/components/ContributionFormFields.jsx
```diff
@@ -100,6 +111,11 @@ function FormFields(props: PropTypes) {
         disabled={isSignedIn}
       />
       <Signout isSignedIn />
+      <MustSignIn
```
jranks123: Seeing as all these props are read directly from the state, I'd rather we just read them from the state in `MustSignIn` rather than passing them through here.

assets/pages/new-contributions-landing/contributionsLandingActions.js
```diff
@@ -127,6 +132,47 @@ const setThirdPartyPaymentLibrary =
...
+const setLastIdentityResponse = (lastIdentityResponse: IdentityResponse): Action => ({
+  type: 'SET_LAST_IDENTITY_RESPONSE',
+  lastIdentityResponse,
+});
+
+const checkIfEmailHasPassword = (email: string) =>
+  (dispatch: Dispatch<Action>, getState: () => State): void => {
+    const state = getState();
+    dispatch(setIdentityRequestPending(true));
+    fetchJson(
```
jranks123: We should wrap this in `logPromise`

assets/pages/new-contributions-landing/contributionsLandingActions.js
```diff
@@ -127,6 +132,47 @@ const setThirdPartyPaymentLibrary =
...
+      `/dummy/${encodeURIComponent(email)}`,
+      getRequestOptions('same-origin', state.page.csrf),
+    ).then(({ userType }) => {
+      dispatch(setIdentityRequestPending(false));
+      if (typeof userType !== 'string') {
+        throw new Error('userType string was not present in response');
+      }
+      dispatch(setLastIdentityResponse('success'));
```
jranks123: Do we need a way of ensuring that the last response from identity is the response from the last request to identity?

assets/pages/new-contributions-landing/contributionsLandingReducer.js
```diff
@@ -66,6 +69,9 @@ type FormState = {
   thankYouPageStage: ThankYouPageStage,
   hasSeenDirectDebitThankYouPageCopy: boolean,
   payPalHasLoaded: boolean,
+  isSignInRequired: boolean,
+  isIdentityRequestPending: boolean,
+  lastIdentityResponse: IdentityResponse,
```
jranks123: maybe this would be clearer what this is if it were named `lastIdentityResponseStatus`

assets/pages/new-contributions-landing/contributionsLandingReducer.js
```diff
@@ -66,6 +69,9 @@ type FormState = {
   thankYouPageStage: ThankYouPageStage,
   hasSeenDirectDebitThankYouPageCopy: boolean,
   payPalHasLoaded: boolean,
+  isSignInRequired: boolean,
```
jranks123: tiny naming preference (feel free to ignore) but I think these sort of variables are better named without the `is` prefix, which I don't think adds anything

assets/pages/new-contributions-landing/contributionsLandingActions.js
```diff
@@ -127,6 +132,47 @@ const setThirdPartyPaymentLibrary =
...
+  type: 'SET_LAST_IDENTITY_RESPONSE',
+  lastIdentityResponse,
+});
+
+const checkIfEmailHasPassword = (email: string) =>
+  (dispatch: Dispatch<Action>, getState: () => State): void => {
+    const state = getState();
+    dispatch(setIdentityRequestPending(true));
+    fetchJson(
+      `/dummy/${encodeURIComponent(email)}`,
```
jranks123: change this to ``${routes.getuser-type}?email=${encodeURIComponent(email)}`` to make it work with the new endpoint

assets/pages/new-contributions-landing/components/ContributionFormFields.jsx
joelochlann: we want @tomrf1 to re-use this component, so we'll pass them in

assets/pages/new-contributions-landing/components/MustSignIn.jsx
```diff
@@ -0,0 +1,73 @@
...
+  isSignedIn: boolean,
+  userTypeFromIdentityResponse: UserTypeFromIdentityResponse,
+  contributionType: Contrib,
+  checkoutFormHasBeenSubmitted: boolean,
+|};
+
+
+// ----- Functions ----- //
+
+// Build signout URL from given return URL or current location.
```
tomrf1: signout?

PR #1100, opened 2018-10-29
Add get-user-type endpoint

jranks123:
## Why are you doing this?
This endpoint takes an email address and checks to see whether the provided email address has an associated password


app/controllers/IdentityController.scala
```diff
@@ -52,6 +53,26 @@ class IdentityController(
...
+    maybeEmail.fold {
+      SafeLogger.error(scrub"No email provided")
+      Future.successful(InternalServerError.asInstanceOf[Result])
+    } { email =>
+      identityService
+        .getUserType(email)
+        .fold(
+          err => {
+            SafeLogger.error(scrub"Failed to retrieve user type for ${email}: ${err.toString}")
+            InternalServerError
```
joelochlann: I think if you make both of these `InternalServerError("whatever message")` then you don't need the `.asInstanceOf[Result]`

app/controllers/IdentityController.scala
```diff
@@ -52,6 +53,26 @@ class IdentityController(
         }
       )
   }
+
+  def getUserType(maybeEmail: Option[String]): Action[AnyContent] = PrivateAction.async { implicit request =>
+    maybeEmail.fold {
+      SafeLogger.error(scrub"No email provided")
+      Future.successful(InternalServerError.asInstanceOf[Result])
```
joelochlann: this one should probably be `BadRequest("No email provided")`

PR #1101, opened 2018-10-30
Abstract FlashSaleDigitalPack into a more generic FeaturedProductHero component

walaura:
## Why are you doing this?

In preparation for next week's AB test this splits the DP-related parts of the hero component we will be using from the basic layout itself, this will let us easily plug in the rest of test variants as soon as we get the assets for them.

There's no visible user-facing changes. 

Continues the work from #1098 

[**Trello Card**](https://trello.com/c/BaVEqtHY/2043-create-new-featured-product-variants)


assets/components/featuredProductHero/featuredProductHero.jsx
```diff
@@ -0,0 +1,78 @@
...
+    hasTimer,
+    headingSize,
+    headingText,
+    subheadingText,
+    bodyText,
+    cta,
+    image,
+  } = props;
+
+  const timerClassName = hasTimer ? 'component-featured-product-hero__countdownbox' : 'component-featured-product-hero__countdownbox component-featured-product-hero__countdownbox--hidden';
```
Ap0c: We have a helper in `utilities` that might simplify this a bit:

```jsx
classNameWithModifiers('component-featured-product-hero__countdownbox', hasTimer ? [] : ['hidden'])
```

assets/components/featuredProductHero/featuredProductHero.jsx
```diff
@@ -0,0 +1,78 @@
...
+              {bodyText}
+            </p>
+            {cta}
+          </div>
+        </div>
+        <div className="component-featured-product-hero__image">
+          {image}
+        </div>
+      </div>
+    </section >
```
Ap0c: I think the highlighter is complaining here because there's a space between `section` and `>`.

PR #1102, opened 2018-10-30
Check email is valid before calling identity API, and don't submit form unless email doesn't have a password

jranks123:
- remove dummy endpoint
- use new identity endpoint 
- Check email is valid before calling identity API
- Don't submit form unless email doesn't have a password (all recurring payment types bar paypal) 
- Don't block one-off contributions 
- Call identity API on init 
- Don't show error message until form has been submitted

### DIFF
--- a/assets/pages/new-contributions-landing/contributionsLandingActions.js
+++ b/assets/pages/new-contributions-landing/contributionsLandingActions.js
@@ -2,6 +2,7 @@
 
 // ----- Imports ----- //
 
+import { checkEmail } from 'helpers/formValidation';
 import type { CheckoutFailureReason } from 'helpers/checkoutErrors';
 import { type ThirdPartyPaymentLibrary } from 'helpers/checkouts';
 import { type Amount, logInvalidCombination, type Contrib, type PaymentMethod, type PaymentMatrix } from 'helpers/contributions';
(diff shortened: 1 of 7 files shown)


assets/pages/new-contributions-landing/contributionsLandingActions.js
```diff
@@ -377,16 +383,24 @@ const paymentAuthorisationHandlers: PaymentMatrix<(Dispatch<Action>, State, Paym
     Stripe: (dispatch: Dispatch<Action>, state: State, paymentAuthorisation: PaymentAuthorisation): void => {
       dispatch(executeStripeOneOffPayment(stripeChargeDataFromAuthorisation(paymentAuthorisation, state)));
     },
-    DirectDebit: () => { logInvalidCombination('ONE_OFF', 'DirectDebit'); },
-    None: () => { logInvalidCombination('ONE_OFF', 'None'); },
+    DirectDebit: () => {
+      logInvalidCombination('ONE_OFF', 'DirectDebit');
```
LATaylor-guardian: You may want to store 'ONE_OFF', 'ANNUAL' and 'MONTHLY' as constants somewhere given they are used across files.

jranks123: I think the reason we don't do this is because we use flow types to safeguard against typos. It's a convention we use across the entire codebase but I'd be interested in peoples opinions @joelochlann @Ap0c ?

Anyway, this change isn't even really part of this PR, it was just some auto formatting intelliJ did and I thought I'd keep in as I preferred it!

joelochlann: Flow will guard against this, yes. Storing these strings in an object might make autocomplete slightly easier but not too much in it

PR #1105, opened 2018-10-30
Featured product test AB setup

walaura:
## Why are you doing this?

Continues #1101 . This sets up a component to render the three possible variants for the featured product AB test in `FeaturedProductHero` and makes it change text colours based on the product.

We are missing images & final copy but adding those should be a super minor PR compared to this which is the major code-related work

There are no user-facing changes as the component is not rendered from anywhere (Plan is for it to replace the flash sale hero currently in place) but to throw in a quick visual this is how the current state (missing final copy & images & a designer to go over the colours) looks like for GW if it were to be rendered:

screenshot 2018-10-30 at 1 03 08 pm

### DIFF
--- a/assets/pages/subscriptions-landing/components/featuredProducts.jsx
+++ b/assets/pages/subscriptions-landing/components/featuredProducts.jsx
@@ -6,6 +6,7 @@ import React, { type Node } from 'react';
 import { type SubsUrls } from 'helpers/externalLinks';
 import { getQueryParameter } from 'helpers/url';
 import { type SubscriptionProduct } from 'helpers/subscriptions';
+import { type CountryGroupId } from 'helpers/internationalisation/countryGroup';
 
 import GridPicture from 'components/gridPicture/gridPicture';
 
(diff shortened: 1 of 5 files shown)


assets/pages/subscriptions-landing/components/featuredProducts.jsx
```diff
@@ -0,0 +1,87 @@
...
+    altText="ad-free, live and discover, and the daily edition"
+    fallbackImgType="png"
+  />);
+
+
+const getProducts = (subsLinks: SubsUrls): {DigitalPack: Product, Paper: Product, GuardianWeekly: Product} => ({
+  DigitalPack: {
+    name: 'DigitalPack',
+    headingText: 'Digital Pack',
+    bodyText: 'Read the Guardian ad-free on all devices, plus get all the benefits of the Premium App and Daily Edition iPad app of the UK newspaper',
```
rupertbates: I think we might need to localise this copy. Usually the UK version just says 'the newspaper' and everywhere else says 'the UK newspaper'.

walaura: do we have any libraries in place for localising strings?

walaura:
@rupertbates it is TBD but I'll confirm tomorrow! (We are missing images too)

assets/pages/subscriptions-landing/components/featuredProducts.jsx
Ap0c: [This](https://github.com/guardian/support-frontend/blob/master/assets/helpers/contributions.js#L134) is a pattern we use in a few places.

PR #1109, opened 2018-10-31
Paypal button enables/disables correctly according to validity of form and the User Type

jranks123:
## Why are you doing this?
This PR enables/disables the PayPal button according to the validity of form and the type of account belonging to the email address entered by the user. 

There is only one edge case bug _i think_. This is where the user goes from typing a valid email that doesn't have a password, then clicks directly onto the payment button. In this instance, the user has to double click. But this is unexpected behaviour due to the order of the buttons, and is something we are begrudgingly willing to accept.

### DIFF
--- a/assets/helpers/checkoutForm/checkoutForm.js
+++ b/assets/helpers/checkoutForm/checkoutForm.js
@@ -51,7 +51,6 @@ export const formInputs = (formClassName: string): Array<HTMLInputElement> => {
 export const getForm: string => Object | null =
   (formName: string) => document.querySelector(`.${formName}`);
 
-// Takes either a form element, or the HTML class of the form
 export const formElementIsValid = (formElement: Object | null) => {
   if (formElement && formElement instanceof HTMLFormElement) {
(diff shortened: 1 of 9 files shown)


assets/helpers/checkoutForm/checkoutForm.js
```diff
@@ -45,14 +48,33 @@ export const formInputs = (formClassName: string): Array<HTMLInputElement> => {
...
-export const formIsValid = (formClassName: string) => {
-  const form = document.querySelector(`.${formClassName}`);
-  if (form && form instanceof HTMLFormElement) {
-    return form.checkValidity();
+// Takes either a form element, or the HTML class of the form
+export const formIsValid = (form: Object | string) => {
+
+  let formElement;
+
+  if (typeof form === 'string') {
```
tomrf1: might be nicer to have `formIsValid` only take a form element, and query for the element before calling it

jranks123: 👍

PR #1110, opened 2018-10-31
Tweaked The Three Headers

Ap0c:
## Why are you doing this?

Some tweaks to the headers that we used to keep them consistent. This affects the country switcher header, the simple header and the new contributions header. They won't be applied yet because the class to apply them doesn't exist yet. This will be added in a future PR.

cc @JustinPinner 

## Changes

- Tweaked the roundel position.
- Updated some colours and padding.


assets/stylesheets/gu-sass/gu-sass.scss
```diff
@@ -8,4 +8,85 @@
...
+  }
+
+  // new-contributions-landing/contributionsLanding.scss
+  #new-contributions-landing-page-uk,
+  #new-contributions-landing-page-eu,
+  #new-contributions-landing-page-us,
+  #new-contributions-landing-page-au,
+  #new-contributions-landing-page-int,
+  #new-contributions-landing-page-nz,
+  #new-contributions-landing-page-ca {
```
JustinPinner: Can we just say `#new-contributions-landing-page` for these?

PR #1111, opened 2018-10-31
add .editorconfig file

akash1810:
## Why are you doing this?
IntelliJ's default is to indent js files with 4 spaces. It looks like we've standadised on 2 in the repo, so add .editorconfig to tell the IDE what to use.

More info here - https://editorconfig.org/

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,11 @@
+# editorconfig.org
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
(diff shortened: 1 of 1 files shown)


akash1810:
> are the rest of options in line with the repo?

Yep 😄

tsop14:
Nice! Making life easier two spaces at a time 🙌

johnduffell:
that's great, thanks for sorting that annoying issue! 👍

PR #1114, opened 2018-10-31
enable devtools in DEV

akash1810:
## Why are you doing this?
The react dev tools allow you to view your app's DOM as components, allowing you to see the props etc.

Really useful to help determine what component is responsible for what part of the page.

See:
- https://github.com/developit/preact/pull/339#issuecomment-252769393
- https://github.com/facebook/react-devtools


assets/helpers/page/page.js
```diff
@@ -36,6 +36,12 @@ import storeReferrer from 'helpers/tracking/awin';
 
 import { type Action } from './pageActions';
 
+// TODO: ideally the `stage` will be injected via config
```
walaura: `webpack.prod.js` is already defining `process.env.NODE_ENV='production'` in PROD. We could use this already or add it to `webpack.dev.js`

akash1810: 👍 Great shout! Done!

akash1810:
DNM - failing in CI.

akash1810:
Tested in CODE and there's no difference in the bundle size, so this is good to go! 🎉 

img

PR #1115, opened 2018-10-31
Footer for digital pack checkout

lmath:
## Why are you doing this?
We need to include info about terms & conditions, the privacy policy and FAQs for the digital pack checkout page. 

[**Trello Card**](https://trello.com/c/jZGDNvMV/1851-checkout-footer-update)

## Changes

* Moved FAQ into a subscriptionFaq component (I don't think needing customer service details implies we also want FAQ details. Also we have FAQs that vary based on product.) 
* Added a T&C + privacy policy component for subscriptions
* Added FAQ, T&C, and privacy policy info into the footer for the digital pack checkout.

## Screenshots
image


assets/helpers/legal.js
```diff
@@ -3,21 +3,33 @@
...
+  EURCountries: defaultContributionTermsLink,
+  International: defaultContributionTermsLink,
+  NZDCountries: defaultContributionTermsLink,
+  Canada: defaultContributionTermsLink,
+};
+
+const subscriptionsTermsLinks: {
+  [SubscriptionProduct]: string,
+} = {
+  DigitalPack: 'https://www.theguardian.com/info/2014/aug/06/guardian-observer-digital-subscriptions-terms-conditions',
```
lmath: We are only getting links for Digital Pack at the moment. I just like having a complete match.

assets/components/legal/subscriptionTermsPrivacy/subscriptionTermsPrivacy.scss
```diff
@@ -0,0 +1,13 @@
+.component-subscription-terms-privacy__text {
+  color: gu-colour(neutral-4);
+  display: block;
+}
+
+.component-subscription-terms-privacy__href {
```
akash1810: Worth making this more generic so that all links on the page get the same treatment? It'll also mean we can remove `subscriptionFaq.scss`

assets/components/subscriptionFaq/subscriptionFaq.jsx
```diff
@@ -0,0 +1,69 @@
...
+      return (
+        <div className="component-subscription-faq">
+          <div className="component-subscription-faq__text">
+            You may also find help in our
+            <a className="component-subscription-faq__href" href={faqLink(props)}> Frequently Asked Questions</a>.
+          </div>
+        </div>
+      );
+  }
+}
```
Ap0c: It looks like the wrappers here are quite similar. You could use the React [`children` API](https://reactjs.org/docs/react-api.html#reactchildren):

```jsx
function Wrapper(props: { children: React$Node }) {
  return (
    <div className="component-subscription-faq">
      <div className="component-subscription-faq__text">{props.children}</div>
    </div>
  );
}
```

to make this bit re-usable, and then apply it like this:

```jsx
<Wrapper>
  You may also find help in our
  <a className="component-subscription-faq__href" href={faqLink(props)}> Frequently Asked Questions</a>.
</Wrapper>
```

This is just an option if you find it useful, I'm not sure it makes it any simpler 🤔.

assets/components/legal/subscriptionTermsPrivacy/subscriptionTermsPrivacy.jsx
```diff
@@ -0,0 +1,37 @@
+// @flow
+
+// ----- Imports ----- //
+
+import React from 'react';
+
+import { subscriptionsTermsLinks, privacyLink } from 'helpers/legal';
+import { type SubscriptionProduct } from '../../../helpers/subscriptions';
```
walaura: nitpick: can this be an absolute import?

assets/pages/digital-subscription-checkout/digitalSubscriptionCheckout.jsx
```diff
@@ -16,6 +16,8 @@ import CustomerService from 'components/customerService/customerService';
 
 import reducer from './digitalSubscriptionCheckoutReducer';
 import CheckoutStage from './components/checkoutStage';
+import SubscriptionTermsPrivacy from '../../components/legal/subscriptionTermsPrivacy/subscriptionTermsPrivacy';
+import SubscriptionFaq from '../../components/subscriptionFaq/subscriptionFaq';
```
Ap0c: I think you can drop the leading `../../` from these.

assets/pages/digital-subscription-landing/digitalSubscriptionLanding.jsx
```diff
@@ -19,6 +19,7 @@ import IndependentJournalismSection from './components/independentJournalismSect
 import ProductBlock from './components/productBlock';
 import PromotionPopUp from './components/promotionPopUp';
 import promotionPopUpReducer from './components/promotionPopUpReducer';
+import SubscriptionFaq from '../../components/subscriptionFaq/subscriptionFaq';
```
Ap0c: `../../` as above.

assets/components/subscriptionFaq/subscriptionFaq.jsx
```diff
@@ -0,0 +1,69 @@
+// @flow
+
+// ----- Imports ----- //
+
+import React from 'react';
+import { subscriptionsTermsLinks } from 'helpers/legal';
+import type { SubscriptionProduct } from '../../helpers/subscriptions';
```
walaura: same here!
```suggestion
import type { SubscriptionProduct } from 'helpers/subscriptions';
```

assets/components/legal/subscriptionTermsPrivacy/subscriptionTermsPrivacy.scss
```diff
@@ -0,0 +1,13 @@
...
+  color: gu-colour(neutral-4);
```
walaura: You can also use `color: inherit`in this one so it'll inherit it from the parent (which has the same text colour)

assets/components/legal/subscriptionTermsPrivacy/subscriptionTermsPrivacy.scss
lmath: I thought about putting the styles for links into `footer.scss` instead and then adding `className=component-footer__href` to the links but I wasn't sure if that's where those styles should live. Is that what you're suggesting, or something else?

lmath: Either way, I think I'm going to merge this tomorrow morning, but possibly followed by a little PR to genericise after.

PR #1117, opened 2018-10-31
Block recurring contributions for emails with a password on old guest checkout

jranks123:
## Why are you doing this?
This is preparation for the NPF test.

A few things:

- Mirrors the behaviour of the new payment flow for recurring contributions in regards to stopping the flow if we determine that the email address you have entered has a password and you aren't signed in 
- Fixes a styling bug w.r.t displaying errors
- Does away with the individual `shouldValidate` function for each field, and instead adopting the convention on the NPF of `checkoutFormHasBeenSubmitted`. 
- Deletes the cumbersome and unnecessary checkout form state tests. Because we a) aren't implementing this level of unit testing on the new flow and b) I couldn't be bothered to fix broken tests for a bit of state we are planning on removing from the codebase anyway. Sorry. 
- Adds the `MustSignIn` component to just below the email field, although this needs some visual sprucing up

| OPF Guest checkout with Must Sign In component present |
|-----------|
|image|


assets/pages/regular-contributions/components/formFields.jsx
```diff
@@ -161,29 +168,40 @@ function NameForm(props: PropTypes) {
...
+              modifierClasses={['email']}
+              showError={shouldShowError(props.email, props.checkoutFormHasBeenSubmitted)}
+              errorMessage="Please enter a valid email address."
+              type="email"
+              pattern={emailRegexPattern}
+              required
+            />
+            {/* TODO: style this properly */ }
+            <MustSignIn
+              returnUrl=""
```
joelochlann: should we just remove this property? why would we ever want it to not be the current URL?

assets/pages/regular-contributions/components/formFields.jsx
```diff
@@ -46,10 +44,15 @@ type PropTypes = {|
...
   setLastName: (string) => void,
   setEmail: (string) => void,
+  checkIfEmailHasPassword: (string) => void,
   countryGroup: CountryGroupId,
   country: IsoCountry,
   isSignedIn: boolean,
   stateField: UsState | CaState,
+  checkoutFormHasBeenSubmitted: boolean,
+  userTypeFromIdentityResponse: UserTypeFromIdentityResponse,
+  checkoutFormHasBeenSubmitted: boolean,
```
joelochlann: whoops, `checkoutFormHasBeenSubmitted` is duplicated

assets/pages/regular-contributions/components/contributionsGuestCheckoutContainer.jsx
```diff
@@ -4,48 +4,50 @@
...
+  ) => {
+    if (checkoutFormShouldSubmit(
+      contributionType,
+      isSignedIn,
+      userTypeFromIdentityResponse,
+      getForm(formClassName),
+    )) {
+      dispatch(setStage('payment'));
+    }
+    dispatch(setCheckoutFormHasBeenSubmitted());
```
joelochlann: probably not for this PR, but I increasingly think that this would be better called `setShouldShowFormErrors`

PR #1120, opened 2018-10-31
mobile invalid field error message

tsop14:
## Why are you doing this?
Users with small mobile screens might click the submit button, nothing will happen and they could be confused. This can happen if the form is invalid because of a field which is off the page view. This PR adds an error message to inform the user that action needs to be taken before they try again.

[**Trello Card**](https://trello.com/c/4mshyMeF/122-npf-follow-up-bug-fields-with-errors-dont-autofocus-on-mobile)

## Changes

* Adds an action that sets `prop.formIsValid`. 
* Action dispatched as part of `onSubmit`so error message will only display once the user has tried to submit the form. The message will stay even if the user corrects a field (and the field specific error message disappears). 
* Adds styling and displays only on mobile.

@ionamckendrick 

## Screenshots
screen shot 2018-10-31 at 17 57 23

mobile-error2


jranks123:
I think we also need to show this message when we are going to force them to sign in, which means that the message needs to accommodate that

jranks123:
Also, I think we should keep with the green error message rather than red, to keep in line with the rest of the error messaging

jranks123:
Maybe we could only show this if the email field is not visible on the screen (if it is on the screen then we know that all fields are visible)

tsop14:
> I think we also need to show this message when we are going to force them to sign in, which means that the message needs to accommodate that

This logic currently will only show if the form has an invalid field (and is on a small mobile screen e.g. iphone5). A further PR can add the error message if they need to sign in.

tsop14:
> Also, I think we should keep with the green error message rather than red, to keep in line with the rest of the error messaging

This is now consistent but everything is red. @ionamckendrick @zeftilldeath

assets/components/paymentFailureMessage/paymentFailureMessage.scss
```diff
@@ -10,10 +10,21 @@
...
 }
 
+.component-payment-failure-message--invalid_form_mobile {
+  padding-top: $gu-v-spacing;
+  margin-bottom: -$gu-v-spacing;
+  color: gu-colour(news-garnett-main-1);
+  background-color: gu-colour(news-faded);
+
+  // point at you can see the all the individual invalid field error messages
+  @include mq($from: 361px) {
```
joelochlann: do we need to be so aggressive with this value? it seems like it's not too huge an issue if we show it on some larger phones, whereas if we don't show it where it's needed it would be annoying. maybe we could bump it to `phablet`? or at least `mobileMedium`

jranks123: I agree with the principle here, to only show it when you can't see the topmost element error message. The danger is though that as we change the form, this value becomes inaccurate. 

So I think either we do something clever where we detect whether the topmost element is visible, or we play it safer and allow the message to show on larger screens to be safe.

tsop14: Have used `mobile medium`.

walaura: hey sorry i'm super late to this party but there are vertical media queries (based on screen height) maybe that would be a batter fit for this?

PR #1121, opened 2018-10-31
Track users who must log in

joelochlann:
This fires ComponentEvents to Ophan and click events to GA, to track clicks on the form submit or recurring PayPal button which were successful, or were blocked, and if blocked, why. This is so we can monitor how many people coming through the flow are forced down the sign-in route.


assets/pages/new-contributions-landing/components/ContributionForm.jsx
```diff
@@ -169,14 +170,24 @@ function onSubmit(props: PropTypes): Event => void {
     // Causes errors to be displayed against payment fields
     props.setCheckoutFormHasBeenSubmitted();
     event.preventDefault();
+    const componentId = `${props.paymentMethod}-${props.contributionType}-submit`;
```
joelochlann: Got very sidetracked here by a strange bug. Basically, if you click the submit button rapidly after changing the email, the `onChange` on the email field gets fired but the `onSubmit` never does.

If you *hold down* the mouse for a period of time, you consistently get both events - the change event when the mouse goes down (and the email field loses focus), and the submit event when it comes back up

So there's a strange race condition here which I haven't fully figured out. It is actually a bit worse than the bug we've already accepted, because it means that the circumstances in which a user needs to click the button twice are more common than we thought. The button click could fail after changing the email in any way - not just from a "sign-in required" email to a "no sign-in required" one.

Merits a bit more thought. But probably still acceptable for go-live.

PR #1004, opened 2018-09-28
Do not Merge - Implement TIP tool for paypal

michaelwmcnamara:
The "Test In Production" tool  (TIP tool for short) is a way of visualising the status of a systems critical path in production without the need for running tests.

This branch sets up the wiring for TIP and sets up tracking of one-off payments from paypal. 

image

PR #1123, opened 2018-11-01
Implement monitoring for one-off PayPal payments using TiP

jacobwinch:
## Why are you doing this?

This PR adds production monitoring for one-off contributions, using the [TiP library](https://github.com/guardian/tip#testing-in-production-tip---). As a first pass (to see if this is useful), I've only implemented this for one-off PayPal contributions. 

After each deploy, a dashboard is automatically generated. This dashboard is updated (based on real user behaviour) to show which critical journeys have been completed:

image

Obviously this dashboard looks a bit sparse when we're only tracking 3 events, but hopefully it provides a reasonable demonstration of how this would work for us.

Thanks to @michaelwmcnamara, who spiked adding this library here: https://github.com/guardian/support-frontend/pull/1004. This PR is largely based on his work.

## Changes

* Add TiP library, basic configuration and wiring
* Track successful one-off PayPal journeys

### DIFF
--- a/app/controllers/PayPalOneOff.scala
+++ b/app/controllers/PayPalOneOff.scala
@@ -12,9 +12,10 @@ import cats.data.EitherT
 import cats.implicits._
 import monitoring.SafeLogger
 import admin.{Settings, SettingsProvider, SettingsSurrogateKeySyntax}
-
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.Try
+import services.{IdentityService, PaymentAPIService, TestUserService}
+import com.gu.tip.Tip
 
 class PayPalOneOff(
     actionBuilders: CustomActionBuilders,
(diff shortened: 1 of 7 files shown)


jacobwinch:
> so TIP will add a verified in PROD label to each PR once this is merged?

Right now it will just create/update the dashboards - we can add the label too (just needs a config change) if people think this would be useful?

johnduffell:
I think this PR might have caused a small issue, there are log messages appearing now
`2018-11-02 18:02:14,559 [default-akka.actor.default-dispatcher-8] ERROR com.gu.tip.TipFactory$$anon$1 - Unrecognized path name: DE One-off PayPal contribution`

jacobwinch:
Thanks @johnduffell - I'll fix this.

PR #1124, opened 2018-11-01
Promote GW on subs landing when flash sale ends

walaura:
## Why are you doing this?
Currently, after the flash sale is over the promotional hero block will go away - This will make it display a promo for GW (below) instead.

2 weeks after the sale ends we'll use this block to do a 3 variant AB test between different products. This PR contains all of those blocks and some testing infra, some extra work in those aspects, not related to this PR might be needed later on.

it also polishes up the CSS so the header image and text always fit instead of potentially "escaping" the frame. This is more important now that there are four different banners (flash sale + ab tests) in this block.

[**Trello Card**](https://trello.com/c/BaVEqtHY/2043-create-new-featured-product-variants)

## Screenshots
  
screenshot 2018-11-01 at 10 34 12 am
screenshot 2018-11-01 at 10 34 16 am


assets/pages/subscriptions-landing/components/featuredProducts.jsx
```diff
@@ -42,10 +44,28 @@ const dpImage = (
...
     fallbackSize={500}
-    altText="ad-free, live and discover, and the daily edition"
+    altText=""
     fallbackImgType="png"
   />);
 
+const weeklyImage = (
+  <GridImage
+    gridId="guardianWeeklyHero"
+    srcSizes={[140, 500, 1000, 1080]}
```
akash1810: I'm not sure that extra 80px is worth it.

Additionally, can we use check Grid has created crops of these widths? For more context see [here](https://github.com/guardian/support-frontend/pull/1107), the patron image didn't have a 1000px wide crop as the source image was ~800px.

For reference, the standard Grid widths are defined [here](https://github.com/guardian/grid/blob/master/cropper/app/lib/CropperConfig.scala#L29-L30).

walaura: it totally isn't! it exists (that's the original image) but yeah it's not necessary

PR #1127, opened 2018-11-01
Route for new GW landing page

walaura:
## Why are you doing this?

Humble beginnings

[**Trello Card**](https://trello.com/c/SGDARHeu/2055-gw-product-page-skeleton)

## Screenshots
screenshot 2018-11-01 at 1 20 43 pm


akash1810:
This is a great PR to demonstrate how to add a new route to the site. May be worth adding a note to the wiki somewhere for future generations?

PR #1128, opened 2018-11-01
Do not set cookie if test switched off

joelochlann:
Fixes a bug whereby even if the serverside test was switched off in `settings.json`, the server would still set a cookie and the client would still set an A/B test participations (although thankfully it only ever served the existing page!).

So I've
- renamed the test from  `newPaymentFlow` -> `newFlow` so we don't get derailed by the dud data we've been pumping in for a few days
- removed the `newFlow` test from `abtestDefinitions.js` because this was misleading (none of the settings in there, not even `isActive`, actually had any effect on test since this was all determined serverside)
- made sure the client takes account of the serverside switch when figuring out A/B test participation
- made sure the server only drops a cookie if the test switch is on

### DIFF
--- a/app/controllers/Application.scala
+++ b/app/controllers/Application.scala
@@ -106,7 +106,8 @@ class Application(
     implicit val gd: GuardianDomain = guardianDomain
 
     val experiments = settings.switches.experiments
-    if (experiments.get("newPaymentFlow").exists(_.isInVariant(participation))) {
+    val npf = experiments.get("newPaymentFlow")
+    if (npf.exists(_.isInVariant(participation))) {
       SafeLogger.info("SERVENEW: Serving new contribution landing page")
       request.user.traverse[Attempt, IdUser](identityService.getUser(_)).fold(
         err => {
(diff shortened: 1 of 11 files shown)


app/controllers/Application.scala
```diff
@@ -106,7 +106,8 @@ class Application(
     implicit val gd: GuardianDomain = guardianDomain
 
     val experiments = settings.switches.experiments
-    if (experiments.get("newPaymentFlow").exists(_.isInVariant(participation))) {
+    val npf = experiments.get("newPaymentFlow")
```
jranks123: newFlow

conf/DEV.public.conf
```diff
@@ -25,4 +25,4 @@ settingsSource.s3 {
 }
 
 # Uncomment the line below if you want to override admin settings locally
-# settingsSource.local.path="~/.gu/support-frontend-admin-console/settings.json"
+settingsSource.local.path="~/.gu/support-frontend-admin-console/settings.json"
```
jranks123: DId you mean to uncomment this?

joelochlann: NO! thanks ;-)

app/controllers/Application.scala
joelochlann: thank you!

assets/helpers/abTests/abtestDefinitions.js
```diff
@@ -4,7 +4,7 @@ import type { Tests } from './abtest';
 // ----- Tests ----- //
 
 export type AnnualContributionsTestVariant = 'control' | 'annualAmountsA' | 'notintest';
-export type NewPaymentFlowTestVariant = 'control' | 'newPaymentFlow';
+export type NewFlowTestVariant = 'control' | 'newFlow';
```
tsop14: If the test isn't defined here, could move the enum to abtest.js

joelochlann: actually once I moved it, I realised it was only being used in one place where it wasn't adding any value, so have ditched it

PR #1129, opened 2018-11-01
NCF: Make input focus ring persist on hover

walaura:
## Why are you doing this?

The input fields in the NCF weren't getting their blue focus state until the mouse was outside of them, which I suspect is a bug and in my criteria looked a bit confusing and took away a lot of the satisfaction of clicking down on the input field.

Luckily the [components cheatsheet](https://interactive.guim.co.uk/embed/article-embeds/HiZef/embed.html) keeps the bright focus ring on hover too, so I can just change the ncf behaviour to match the cheatsheet and be happy about it! 😸 

## Screenshots
2018-11-01 2_43_39 pm

### DIFF
--- a/assets/pages/new-contributions-landing/contributionsLanding.scss
+++ b/assets/pages/new-contributions-landing/contributionsLanding.scss
@@ -733,7 +733,7 @@ form {
   color: gu-colour(media-garnett-main-1);
 }
 
-.form__input:focus {
+.form__input:enabled:focus {
   box-shadow: 0 0 0 3px gu-colour(sport-garnett-media-main-1);
   outline: none;
 }


assets/pages/new-contributions-landing/contributionsLanding.scss
```diff
@@ -733,7 +733,7 @@ form {
   color: gu-colour(media-garnett-main-1);
 }
 
-.form__input:focus {
+.form__input:enabled:focus {
```
walaura: `:enabled` here just makes the selector as [specific](https://specificity.keegan.st/) as the `:enabled:hover` selector on top, meaning focus can take priority

PR #1134, opened 2018-11-01
Form Library

Ap0c:
## Why are you doing this?

The beginnings of a standardised library of form components, intended for the checkouts. One reason for this is to standardise design language across the forms on the site. Another is to make components more composable, i.e. combine a collection of simple components to get a new component with augmented functionality, rather than a monolithic component that handles many different circumstances.

This PR introduces a `forms` folder in the `components` directory composed of the following:

- `standardFields` - Unmodified HTML form fields. These are functionally the same as the vanilla form fields. They exist to apply standardised styling across the forms on the site. You can choose to use them and get this styling, or roll your own and apply custom styles.
- `customFields` - Modified HTML form fields. These come with extra functionality when compared to the vanilla form fields, and corresponding styling. Again, you can choose to use these or roll your own.
- `formHOCs` - [Higher-order components](https://reactjs.org/docs/higher-order-components.html) that apply additional functionality to form fields. They can be applied to the form components from this library or custom ones created elsewhere.

cc @JustinPinner 

## Changes

- Added `button`, `fieldset`, `input`, `legend` and `select` standard components.
- Added `radioInput` as a custom field because the design language calls for additional properties.
- Added `withError` and `withLabel` HOCs.


assets/components/forms/standardFields/button.scss
```diff
@@ -0,0 +1,15 @@
+.component-button {
```
walaura: should this sync up in any way with the cta component?

assets/components/forms/standardFields/input.scss
```diff
@@ -0,0 +1,10 @@
+@import './formFields';
+
+.component-input:not([type="radio"]) {
+  @include form-field;
+
+  height: 38px;
+  padding-left: 12px;
+  padding-right: 12px;
+  width: 314px;
```
walaura: Could this be max-width instead? Just to allow it some flexibility in smaller cases

assets/components/forms/standardFields/input.scss
```diff
@@ -0,0 +1,10 @@
+@import './formFields';
+
+.component-input:not([type="radio"]) {
+  @include form-field;
+
+  height: 38px;
```
walaura: does this have to be explicitly set? could it be a combo of `min-height`, `padding`, and `line-height`?

assets/components/forms/standardFields/input.scss
```diff
@@ -0,0 +1,10 @@
...
+  padding-left: 12px;
```
walaura: can this be `$gu-h-spacing`? do we need those extra 2 px?

PR #1135, opened 2018-11-02
SEO Updates to GW page, checkout ty page

walaura:
## Why are you doing this?

Bit of a grabbag PR containing

- An updated `headingBlock` element with a yellow label on top for its `h1` (see screenies) that will be used as the heading for all pages going on, including `thank you`.
- SEO updates on the Scala side (including the h1) for the GW page, which I still consider to be unstarted but has gained a couple of CSS and HTML that needs to be converted to React components. (At this stage I'm still figuring out how to best lay the header out and the flexibility of raw HTML is a godsend here)

[**Trello Card**](https://trello.com/c/CPPgEPtM/2058-gw-page-update-headingblock-so-it-acts-as-a-reusable-component-for-all-pages)
[**Trello Card**](https://trello.com/c/6p3aLUyc/1979-seo-requirements-for-guardian-weekly-product-page)

## Screenshots

screenshot 2018-11-02 at 9 35 06 am
screenshot 2018-11-02 at 9 41 56 am

## A note on IE11
The header uses a bit of positioning and `transform` fanciness to have the yellow tag break outside its box yet support any height. This is done so the hero images on top can sit behind it without it being too much of a fuss to have them overlap.

This is all good and works in all our supported browsers. ** However** the impact of this trick not working is quite drastic (the yellow block would appear on top of the black one, making it unreadable) so I've wrapped this whole behaviour inside a `supports(){}` query which is not supported by IE11.

Long story short, very old browsers including IE11 will instead see a yellow bar filling the screen (like the black one below it) and browsers that can do both `support()` and `transform()` (Every browser we support except IE11) will see the normal header in the screenshots above

### DIFF
--- a/app/controllers/Subscriptions.scala
+++ b/app/controllers/Subscriptions.scala
@@ -141,6 +141,6 @@ class Subscriptions(
     s"${supportUrl}/${countryCode}/subscribe/digital"
 
   def buildCanonicalWeeklySubscriptionLink(countryCode: String): String =
-    s"${supportUrl}/${countryCode}/subscribe/guardian-weekly"
+    s"${supportUrl}/${countryCode}/subscribe/weekly"
 
 }
--- a/app/controllers/Subscriptions.scala
+++ b/app/controllers/Subscriptions.scala
@@ -104,7 +104,7 @@ class Subscriptions(
     Ok(views.html.main(title, id, js, css)).withSettingsSurrogateKey
   }
 
-  def weeklyGeoRedirect: Action[AnyContent] = geoRedirectAllMarkets("subscribe/guardian-weekly")
+  def weeklyGeoRedirect: Action[AnyContent] = geoRedirectAllMarkets("subscribe/weekly")
 
   def weekly(countryCode: String): Action[AnyContent] = CachedAction() { implicit request =>
     implicit val settings: Settings = settingsProvider.settings()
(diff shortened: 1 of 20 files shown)


assets/components/headingBlock/headingBlock.scss
```diff
@@ -1,34 +1,62 @@
...
-    width: 500px;
+    font-size: 36px;
   }
 }
 
 .component-heading-block__banner {
-  border-top: 1px solid gu-colour(news-garnett-highlight);
-  padding: 8px ($gu-h-spacing / 2) $gu-v-spacing;
+  @extend %component-heading-block__card;
+  border-top: 1px solid #ffffff;
```
jranks123: use `garnett-neutral-5`

app/controllers/Subscriptions.scala
```diff
@@ -108,11 +108,22 @@ class Subscriptions(
 
   def weekly(countryCode: String): Action[AnyContent] = CachedAction() { implicit request =>
     implicit val settings: Settings = settingsProvider.settings()
-    val title = "Support the Guardian | The Guardian Weekly"
+    val title = "The Guardian Weekly Subscriptions | The Guardian"
```
Ap0c: Should this be "Subscription"?

assets/pages/weekly-subscription-landing/weeklySubscriptionLanding.scss
```diff
@@ -7,3 +7,33 @@
...
+.weekly-subscription-landing-header {
+  background: gu-colour(garnett-neutral-3);
+  overflow: hidden;
+  padding-top: 300px;
+  .component-left-margin-section__content {
+    border-left: 1px solid gu-colour(garnett-neutral-4);
+  }
+}
+
+.weekly-subscription-landing-header img {
```
Ap0c: Prefer styling on classes.

PR #1137, opened 2018-11-02
Add click events since conversions aren't working on new flow

joelochlann:
For new flow, we just send clicks (since conversions aren't working)

For old flow, I've added in clicks in addition to conversions (since conversions are working)


joelochlann:
actually it looks like `onPaymentResult` gets called after recurring payments too. plus we won't get PayPal one-off with this approach. so closing

PR #1139, opened 2018-11-02
Send session id on old flow

joelochlann:
and turn guest checkout back on

This sessionId is required by the backend when trying to create a guest account for a new email address. We supply it in the new flow but forgot to send it for the old flow


johnduffell:
👍 we could check on monday in case anyone made a double contribution by mistake, as if the browser double sends it might cause that issue

PR #1140, opened 2018-11-05
Digital Checkout Details

Ap0c:
## Why are you doing this?

Added the details fields to the digital subs checkout.

[**Trello Card**](https://trello.com/c/i6owjmJJ/1507-personal-details-in-checkout)

cc @JustinPinner 

## Changes

- Added a `CheckoutForm` component with four fields.
- Added `firstName`, `lastName`, `country` and `telephone` to the state, along with their corresponding actions and reducer entries (I'm trialling putting actions and reducers in the same file to see if it's any simpler).
- Added a utility helper to compare two strings.
- Added an `asControlled` HOC to handle controlled components.


assets/pages/digital-subscription-checkout/components/checkoutForm.jsx
```diff
@@ -0,0 +1,76 @@
...
+import { Select } from 'components/forms/standardFields/select';
+import { withLabel } from 'components/forms/formHOCs/withLabel';
+import { asControlled } from 'components/forms/formHOCs/asControlled';
+
+import { type State, actions } from '../digitalSubscriptionCheckoutReducer';
+
+
+// ----- Form Fields ----- //
+
+function Input1(props: { id: string, label: string, value: string, setValue: string => void }) {
```
walaura: is `Name1` a convention we use more broadly?

Ap0c: I believe @regiskuckaertz used it in places while working on the contributions checkout. It's meant to signify "Input transformed", and I don't think JavaScript allows `'` in variable names. Alternatives would be `InputAsControlledWithLabel`, `NewInput` or `Input_`.

As a side note, I was hoping this intermediate component wouldn't be necessary and that we'd be able to do:

```jsx
const FirstName = compose(
  connect(
    (state: State) => ({ value: state.page.firstName }),
    { setValue: actions.setFirstName },
  ),
  asControlled,
  withLabel,
)(Input);
```

all in one. However, the prop types seem to get lost somewhere in the call to `connect`, so this intermediate function is needed mainly to provide Flow with the type definition.

assets/helpers/utilities.js
```diff
@@ -13,6 +13,13 @@ function descending(a: number, b: number): number {
   return a < b ? 1 : 0;
 }
 
+// String comparison for use with Array.prototype.sort.
+function compareString(a: string, b: string) {
```
regiskuckaertz: I think `a.localeCompare(b)` should work?

assets/pages/digital-subscription-checkout/components/checkoutForm.jsx
regiskuckaertz: > However, the prop types seem to get lost somewhere in the call to connect

Sad 😕You can probably add type annotations to solve that, though I don't know that it's clearer.

assets/components/forms/formHOCs/asControlled.jsx
```diff
@@ -0,0 +1,33 @@
...
+import React from 'react';
+
+
+// ----- Types ----- //
+
+type InitialProps<V> = {
+  value: V,
+};
+
+type AugmentedProps<Props> = Props & {
```
rupertbates: What is the idea behind the types in this file?

assets/pages/digital-subscription-checkout/components/checkoutForm.jsx
akash1810: Not a huge fan of `Input1` either. `InputField`?

assets/pages/digital-subscription-checkout/components/checkoutForm.jsx
```diff
@@ -0,0 +1,75 @@
...
+    <div>
+      <FirstName id="first-name" label="First name" type="text" />
+      <LastName id="last-name" label="Last name" type="text" />
+      <Country id="country" label="Country">
+        {Object.keys(countries)
+          .sort((a, b) => countries[a].localeCompare(countries[b]))
+          .map(iso => <option value={iso}>{countries[iso]}</option>)
+        }
+      </Country>
+      <Telephone id="telephone" label="Telephone (optional)" type="tel" />
```
akash1810: other than the label, how do we signal this field is optional?

assets/pages/digital-subscription-checkout/components/checkoutForm.jsx
```diff
@@ -0,0 +1,75 @@
...
+)(Select1);
+
+
+// ----- Component ----- //
+
+function CheckoutForm() {
+
+  return (
+    <div>
+      <FirstName id="first-name" label="First name" type="text" />
```
akash1810: A slightly nicer API would be for these fields to know their type. That is, `FirstName` knows its `text` and `Telephone` a `tel` as I the `type` wouldn't change.

Ap0c: True. I could do that by creating another intermediate component, partially applying the prop. However, wouldn't the id and label also be specific to this field in that case too?

assets/pages/digital-subscription-checkout/components/checkoutForm.jsx
LATaylor-guardian: @regiskuckaertz haha I literally saw this naming convention in the `Doobie` source code yesterday with `Update0`.

assets/pages/digital-subscription-checkout/components/checkoutForm.jsx
rupertbates: Yes I think you could very easily just subsume the id and label into the component

Ap0c:
Closing temporarily to fix an issue with re-rendering.

app/controllers/Subscriptions.scala
```diff
@@ -128,14 +128,18 @@ class Subscriptions(
...
     authenticatedAction(recurringIdentityClientId) { implicit request =>
-      implicit val settings: Settings = settingsProvider.settings()
-      val title = "Support the Guardian | Digital Subscription"
-      val id = "digital-subscription-checkout-page-" + countryCode
-      val js = "digitalSubscriptionCheckoutPage.js"
-      val css = "digitalSubscriptionCheckoutPageStyles.css"
-      Ok(views.html.main(title, id, js, css)).withSettingsSurrogateKey
+      if (displayCheckout == "true") {
+        implicit val settings: Settings = settingsProvider.settings()
+        val title = "Support the Guardian | Digital Subscription"
```
LATaylor-guardian: We should declare these constants at object level so that they are created once only  throughout the lifecycle of the application. At the moment they are created everytime this function is invoked.

app/controllers/Subscriptions.scala
```diff
@@ -128,14 +128,18 @@ class Subscriptions(
 
   def premiumTierGeoRedirect: Action[AnyContent] = geoRedirect("subscribe/premium-tier")
 
-  def displayForm(countryCode: String): Action[AnyContent] =
+  def displayForm(countryCode: String, displayCheckout: String): Action[AnyContent] =
```
LATaylor-guardian: displayCheckout should be a `Boolean`.

app/controllers/Subscriptions.scala
```diff
@@ -128,14 +128,18 @@ class Subscriptions(
...
     authenticatedAction(recurringIdentityClientId) { implicit request =>
-      implicit val settings: Settings = settingsProvider.settings()
-      val title = "Support the Guardian | Digital Subscription"
-      val id = "digital-subscription-checkout-page-" + countryCode
-      val js = "digitalSubscriptionCheckoutPage.js"
-      val css = "digitalSubscriptionCheckoutPageStyles.css"
-      Ok(views.html.main(title, id, js, css)).withSettingsSurrogateKey
+      if (displayCheckout == "true") {
```
LATaylor-guardian: As above.

app/controllers/Subscriptions.scala
```diff
@@ -128,14 +128,18 @@ class Subscriptions(
...
+        implicit val settings: Settings = settingsProvider.settings()
```
LATaylor-guardian: Does this need to be implicit? I'd prefer if we passed this on explicitly. It makes the code easier to read and reason about.

JustinPinner:
As discussed (offline) this is narrow-focussed enough that we can go ahead and live with it as-is to get us going. If we find problems later, we'll worry about them then. 👍

PR #1145, opened 2018-11-06
Scale up to 6 instances

joelochlann:
To handle any increase in traffic during the US midterms

Since we made the `contribute` routes uncacheable (they now vary by user login cookie), we're being extra cautious


joelochlann:
should do, because of this: https://github.com/guardian/support-frontend/blob/master/conf/riff-raff.yaml#L4-L12

PR #1146, opened 2018-11-06
Add meta description to NPF

jranks123:
## Why are you doing this?

Our SEO score from Google has dropped from 100% to 93% with NPF. This lil PR will sort it out. 

image


jranks123:
@joelochlann updated robots.txt, can you give this another once over?

PR #1149, opened 2018-11-07
Turn off annual test

jranks123:
## Why are you doing this?
I had started to remove all the code, but then realised how much there was and thought it seemed unnecessary when we are very likely to resume amounts testing as soon as the NPF test is complete.


joelochlann:
As mentioned via hangouts - I think @jessewilkins wanted updated amounts for the control but I'd suggest merging this now and dealing with that separately

johnduffell:
this PR description is not quite clear I guess this was an amounts test rather than annual overall?

PR #1150, opened 2018-11-07
Weekly page design updates

walaura:
## Why are you doing this?

To match amo's latest design.

Very locally-scoped changes here except for importing the title piece font. This will be used more extensively across the board in the new pages so it made sense to just import it globally.

[**Trello Card**](https://trello.com/c/ILBIN5Mm/2061-gw-page-skeleton)

## Screenshots
screenshot 2018-11-07 at 1 49 02 pm

### DIFF
--- a/assets/components/svgs/weeklyHeroCircles.jsx
+++ b/assets/components/svgs/weeklyHeroCircles.jsx
@@ -1,0 +1,11 @@
+// @flow
+
+import React from 'react';
+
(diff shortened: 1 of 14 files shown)


walaura:
@JustinPinner ha! that'll change, amo is looking at options. i'm thinking merging this first thing tomorrow morning as a sort of happy checkpoint, then throwing in all the final tweaks & copy tomorrow

PR #1151, opened 2018-11-07
NPF: Add extra form validation and amounts selection tracking + couple of tracking fixes

jranks123:
## Why are you doing this?
We have been seeing an unexpectedly high rate of form errors on the new payment flow. This PR does a few things around this:

- Adds the form validation error tracking to the old flow so we can see if this is a new issue we've created
- Adds the reasons for the form validation error to the event

and example of what the form validation error event looks like: 

`blocked-because-form-not-valid-contributionEmail-pattern-mismatch-contributionFirstName-value-missing-contributionLastName-value-missing-contributionState-value-missing`

Yes it's rather unwieldy, but it gets the job done.

It also adds click tracking on the amounts and contribution type selections on the new and oldflow

Examples of the events:
NPF change amount: `npf-contribution-amount-toggle-GBPCountries-MONTHLY-10`
NPF change contribution type: `npf-contribution-type-toggle-GBPCountries-MONTHLY`

OPF change amount: `opf-contribution-amount-toggle-GBPCountries-100`
OPF change contribution type: `opf-contribution-type-toggle-GBPCountries-ONE_OFF`

You'll notice that the OPF change amount doesn't have the contribution type. This is because both the amounts toggle and the contribution type toggle come from the same generic `RadioToggle` class which doesn't have knowledge of the context it is in with regards to other instances of Radio Toggle. We could get around this, but as this flow will likely be turned off very soon I think this will suffice for now

Finally, it puts the NPF recurring conversion tracking in the right place

@guardian/contributions

### DIFF
--- a/assets/helpers/checkoutForm/checkoutForm.js
+++ b/assets/helpers/checkoutForm/checkoutForm.js
@@ -63,7 +63,8 @@ export const invalidReason = (form: Object | null): string => {
   let invalidReasonString = '';
   if (form instanceof HTMLFormElement) {
     [...form.elements].forEach((element) => {
-      if (element instanceof HTMLInputElement && !element.checkValidity()) {
+      if ((element instanceof HTMLInputElement || element instanceof HTMLSelectElement)
+        && !element.checkValidity()) {
         invalidReasonString += `-${element.id}${getInvalidReason(element.validity)}`;
       }
     });
--- a/assets/helpers/checkoutForm/checkoutForm.js
+++ b/assets/helpers/checkoutForm/checkoutForm.js
@@ -60,16 +60,22 @@ const getInvalidReason = (validityState: ValidityState) => {
 };
 
 export const invalidReason = (form: Object | null): string => {
-  let invalidReasonString = '';
-  if (form instanceof HTMLFormElement) {
-    [...form.elements].forEach((element) => {
-      if ((element instanceof HTMLInputElement || element instanceof HTMLSelectElement)
-        && !element.checkValidity()) {
-        invalidReasonString += `-${element.id}${getInvalidReason(element.validity)}`;
-      }
-    });
+  try {
+    let invalidReasonString = '';
+    if (form instanceof HTMLFormElement) {
+      [...form.elements].forEach((element) => {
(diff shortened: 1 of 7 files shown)


joelochlann:
could we add clicks on the tabs too?

assets/helpers/checkoutForm/checkoutForm.js
```diff
@@ -50,6 +50,29 @@ export const formInputs = (formClassName: string): Array<HTMLInputElement> => {
...
+  if (validityState.valueMissing) {
+    return '-value-missing';
+  } else if (validityState.patternMismatch) {
+    return '-pattern-mismatch';
+  }
+  return '';
+};
+
+export const invalidReason = (form: Object | null): string => {
+  let invalidReasonString = '';
```
joelochlann: maybe we should wrap this whole thing in a try/catch and also log an event on error, just so we know when the whole constraint validation API is missing in a particular browser.

assets/helpers/checkoutForm/checkoutForm.js
```diff
@@ -50,6 +50,29 @@ export const formInputs = (formClassName: string): Array<HTMLInputElement> => {
...
+        invalidReasonString += `-${element.id}${getInvalidReason(element.validity)}`;
+      }
+    });
+  }
+  return invalidReasonString;
+};
+
+
 export const formElementIsValid = (formElement: Object | null) => {
   if (formElement && formElement instanceof HTMLFormElement) {
```
joelochlann: we could also add some super-paranoid events in here, for things that really shouldn't be happening. for instance, if the `if` condition is false. and another try/catch around `.checkValidity()`

assets/helpers/checkoutForm/checkoutForm.js
jranks123: done

jranks123:
@joelochlann added tab clicks, and did it for OPF too. And I've added try/catch based off your first comment

jranks123:
@joelochlann was just writing a reply to that effect. 

Thanks for the approval. No need to rush it through now as it's just gone 6PM and we will see the events live in GA as soon as it's merged anyway, so we can have a proper think about it tomorrow

PR #1155, opened 2018-11-08
Remove client-side GA conversion tracking

tomrf1:
## Why are you doing this?
GA conversion tracking is now done server-side by support-workers and payment-api.

### DIFF
--- a/assets/helpers/tracking/conversions.js
+++ b/assets/helpers/tracking/conversions.js
@@ -10,10 +10,6 @@ export default function trackConversion(
   participations: Participations,
   currentRoute: string,
 ) {
-  if (!doNotTrack()) {
-    // Fire GTM conversion events
-    successfulConversion(participations);
-  }
   // Send an Ophan pageview. Because this function is used to track page views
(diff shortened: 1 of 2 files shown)


assets/helpers/tracking/conversions.js
```diff
@@ -1,19 +1,13 @@
 // @flow
 
-import { doNotTrack } from 'helpers/page/page';
 import { getAbsoluteURL } from '../url';
 import { pageView } from './ophanComponentEventTracking';
-import { successfulConversion } from './googleTagManager';
 import type { Participations } from '../abTests/abtest';
 
 export default function trackConversion(
```
joelochlann: I'd actually go as far as to say we should remove all calls to `trackConversion`. We definitely don't use that Ophan pageview to track conversions - if anything it will just add noise to our pageview data. The only thing to check would be if `trackConversion` is called in any subscriptions-related code, whether S&C team need it. But from a Contributions perspective, I think we can ditch this entire function

joelochlann: @jranks123 would you agree?

PR #1157, opened 2018-11-08
Add submit tracking for paypal button + refactor

jranks123:
## Why are you doing this?
Previously, we were not sending events to GA + ophan on form submission when the PayPal recurring button was clicked. 

This PR fixes this issue and centralises the logic. 

In addition, this PR allows the benefits of [Emmas PR](https://github.com/guardian/support-frontend/pull/1120) to show the more obvious error message on mobile for PayPal recurring 

| Form validation box showing now for PayPal recurring on mobile |
|--------|
|image|


assets/helpers/checkoutForm/checkoutForm.js
```diff
@@ -78,14 +79,48 @@ export const invalidReason = (form: Object | null): string => {
...
 
-
 export const formElementIsValid = (formElement: Object | null) => {
   if (formElement && formElement instanceof HTMLFormElement) {
     return formElement.checkValidity();
   }
   return false;
 };
 
+type FlowType = 'opf' | 'npf';
```
joelochlann: this term is a bit overloaded, what with Flow types! maybe `OldOrNewFlow` or something

assets/helpers/checkoutForm/checkoutForm.js
```diff
@@ -78,14 +79,48 @@ export const invalidReason = (form: Object | null): string => {
...
+
+export const onFormSubmit = (
+  flowPrefix: FlowType,
```
joelochlann: with this many params, an object makes sense to give the params names at the call sites and make it more readable. you can actually destructure them into local variables right in the parameter list:

```javascript
const x = ({a, b, c}) => {console.log(a);}
```

jranks123: I'm not sure I agree entirely in this situation, as most of them are passed in already named very clearly, i,e `contributionType`, `paymentMethod` etc. But I take your point on the flow prefix. A good compromise would be to store all the parameters as variables before passing them through (which nearly all of them already are anyway), that way you don't have to add loads of boilerplate at the other end. What do you think?

jranks123: then, to use ContributionForm as an example, you call it with something like this: 

`
   const flowPrefix = 'npf';
    const form = event.target;
    const handlePayment = () => formHandlers[props.contributionType][props.paymentMethod](props);
    onFormSubmit(
      flowPrefix,
      props.paymentMethod,
      props.contributionType,
      form,
      props.isSignedIn,
      props.userTypeFromIdentityResponse,
      props.setFormIsValid,
      props.setCheckoutFormHasBeenSubmitted,
      handlePayment,
    );
`

jranks123: actually, I agree with you more looking at how it's called in `ContributionsLandingActions`. I'll give it a go and see how it feels.

assets/pages/regular-contributions/components/contributionsGuestCheckoutContainer.jsx
```diff
@@ -32,29 +30,25 @@ const mapDispatchToProps = (dispatch: Dispatch<CheckoutAction>) => ({
...
+    onFormSubmit(
+      'opf',
+      paymentMethod,
+      contributionType,
+      getForm(formName),
+      isSignedIn,
+      userTypeFromIdentityResponse,
+      () => {}, // we don't need this for the old flow
+      () => { dispatch(setCheckoutFormHasBeenSubmitted()); },
+      () => { dispatch(setStage('payment')); },
```
joelochlann: here's an example where having names would add value - for these last three functions

for those where the names are the same as the parameter names (`paymentMethod`, `contributionType`), you wouldn't have any repetition as you can just use the shorthand object syntax.

assets/helpers/checkoutForm/checkoutForm.js
joelochlann: but then don't you have boilerplate (i.e. storing them in variables) at every call site? vs boilerplate once in the function definition?

I take the point in general about the variable names but I flagged an example in `assets/pages/regular-contributions/components/contributionsGuestCheckoutContainer.jsx` where I think param names would be helpful (e.g., what are we passing an empty function for?)

assets/helpers/checkoutForm/checkoutForm.js
```diff
@@ -78,14 +79,48 @@ export const invalidReason = (form: Object | null): string => {
...
+  setFormIsValid: boolean => void,
+  setCheckoutFormHasBeenSubmitted: () => void,
+  handlePayment?: () => void,
+) => {
+  const componentId = `${paymentMethod}-${contributionType}-submit`;
+  const formIsValid = formElementIsValid(form);
+  const userType = isSignedIn ? 'signed-in' : userTypeFromIdentityResponse;
+  const canContribute = canContributeWithoutSigningIn(contributionType, isSignedIn, userTypeFromIdentityResponse);
+  if (formIsValid) {
+    setFormIsValid(true);
```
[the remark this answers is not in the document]
joelochlann: I think it's fine as is

assets/helpers/checkoutForm/checkoutForm.js
jranks123: You were right, it's much nicer now. Have a look and see what you think

PR #1158, opened 2018-11-08
Fix recurring selenium test for old payment flow

jranks123:
## Why are you doing this?

We need to write tests for the old flow, but for now, I've fixed the ones we broke before.

Annoyingly, the second time around, the email field autofills and won't clear. I looked into it and it's something to do with `autocomplete` being on which Selenium has trouble with ( [see here](https://www.google.co.uk/search?q=selenium+webelement+clear+not+working&rlz=1C5CHFA_enGB806GB806&oq=selenium+webelement+clear+not+working&aqs=chrome..69i57.9175j0j7&sourceid=chrome&ie=UTF-8) ) But for now, the second time around we just don't fill in the email and it still works with the test email generated the first time round!

### DIFF
--- a/test/selenium/RecurringContributionsSpec.scala
+++ b/test/selenium/RecurringContributionsSpec.scala
@@ -20,42 +20,32 @@ class RecurringContributionsSpec extends FeatureSpec with GivenWhenThen with Bef
 
   feature("Sign up for a Monthly Contribution") {
 
-    ignore("Monthly contribution sign-up with Stripe - GBP") {
+    scenario("Monthly contribution sign-up with Stripe - GBP") {
 
       val landingPage = ContributionsLanding("uk")
-      val recurringContributionForm = new RecurringContributionForm
+      val testUser = new TestUser(driverConfig)
+
+      val recurringContributionForm = RecurringContributionForm(testUser)
       val recurringContributionThankYou = new RecurringContributionThankYou
 
       Given("that a test user goes to the contributions landing page")
-      val testUser = new TestUser(driverConfig)
       goTo(landingPage)
       assert(landingPage.pageHasLoaded)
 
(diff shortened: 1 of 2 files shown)


test/selenium/RecurringContributionsSpec.scala
```diff
@@ -20,42 +20,32 @@ class RecurringContributionsSpec extends FeatureSpec with GivenWhenThen with Bef
...
+      assert(recurringContributionForm.detailsPageHasLoaded)
 
-      Given("that the user fills in their personal details correctly")
-      registerPageTwo.fillInPersonalDetails()
+      Given("The user fills in their details correctly")
+      recurringContributionForm.fillInPersonalDetails()
 
-      When("they submit the form to create their Identity account")
-      registerPageTwo.submit()
+      Given("They select next")
```
jacobwinch:
```suggestion
      When("They select next")
```

jacobwinch:
> it still works with the test email generated the first time round!

That's slightly surprising, shouldn't it complain that the email is already linked to an identity account (due to the first scenario)?

jacobwinch:
Probably beyond the scope of this PR, but if we are running with a control which supports guest checkout then we should consider testing:

1. Authenticated identity user buys monthly contribution
2. Unauthenticated user buys monthly contribution

At the moment I think we are testing 2 twice(?), so we could gain some coverage here.

jranks123:
@jacobwinch

> That's slightly surprising, shouldn't it complain that the email is already linked to an identity account (due to the first scenario)?

No, because we never actually create the account in the first instance, so there is no password attatched to the account, which is the only situation where we demand that you sign in. 


And yes you are totally right, I am going to start writing the new flow selenium tests now and I will cover all of those situations. This PR is more just fixing what we broke before

PR #1160, opened 2018-11-09
Improve Tip implementation

jacobwinch:
## Why are you doing this?

To improve upon https://github.com/guardian/support-frontend/pull/1123 and address the spurious error spotted by @johnduffell here: https://github.com/guardian/support-frontend/pull/1123#issuecomment-435499777

## Changes

* Only attempt to verify the path if we have chosen to monitor the user's region / journey
* Use types to make path verification more robust (in preparation for adding this tracking to the recurring flows).


app/monitoring/Tip.scala
```diff
@@ -22,3 +24,38 @@ object TipFromConfig {
...
+
+  type VerifyPath = String => Future[TipResponse]
+
+  sealed trait MonitoredRegion
+  case object AU extends MonitoredRegion
+  case object GB extends MonitoredRegion
+  case object US extends MonitoredRegion
+
+  sealed trait MonitoredProduct
+  case object OneOffContribution extends MonitoredProduct { override def toString: String = "One-off contribution" }
```
johnduffell: overriding toString is my least favourite thing

johnduffell: if you do really want to do it, and want to make it less boilerplatey you may be able to use a sealed abstract class instead of trait, and make it a parameter called toString.

jacobwinch: @johnduffell - I can make all of these vals (tipString?), but it's unclear to me how/why that's any better than overriding toString in this case. 

Would you mind explaining why overriding is a bad idea in this case? Happy to discuss this in person if it's easier.

johnduffell: explained in person, basically toString exists on everything and it's not clear from the name what the purpose of it us (other than turn into some kind of string for unknown purpose) so would lose type safety of knowing that it's (probably) the right object.  `tipString` is a nice idea, or `value` - although the more we use `value`, the more that argument against it would apply there too!
Also I suggested using a sealed abstract class instead of trait, then the string can be a constructor param rather than an override with all the associated boiler plate.

jacobwinch: Thanks for summarising our discussion @johnduffell - I had a look at using abstract classes but decided I preferred sealed traits for this case.

conf/tip.yaml
```diff
@@ -1,20 +1,8 @@
...
   description: Successful one-off contribution in USD (using PayPal)
 
-- name: GB One-off PayPal contribution
+- name: UK One-off contribution with PayPal
   description: Successful one-off contribution in GBP (using PayPal)
 
-- name: AU One-off PayPal contribution
+- name: AU One-off contribution with PayPal
   description: Successful one-off contribution in AUD (using PayPal)
-
```
jacobwinch: These additional paths were added in: https://github.com/guardian/support-frontend/pull/1199.

Since we should now cope with all of these cases cleanly (rather than logging spurious errors), I'm removing them again (I'd prefer use TiP for the most important combinations only).

johnduffell:
👍 great work,!

PR #1164, opened 2018-11-09
Switch to NPF

tomrf1:
## Why are you doing this?
The NPF did well, we're moving all users over.

Note - there is a lot of cleanup to be done later

## Changes

* Change 1
* Change 2

### DIFF
--- a/app/controllers/SiteMap.scala
+++ b/app/controllers/SiteMap.scala
@@ -21,6 +21,7 @@ class SiteMap(
     Ok(sitemapResponse)
   }
 
+  // TODO: more cleanup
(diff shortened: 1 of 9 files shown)


app/controllers/SiteMap.scala
```diff
@@ -21,6 +21,7 @@ class SiteMap(
     Ok(sitemapResponse)
   }
 
+  // TODO: more cleanup
```
joelochlann: I added this comment but this is actually orthogonal to all of this stuff so you can delete it

tomrf1:
I'll merge this on Monday

PR #1165, opened 2018-11-12
Fix selenium test - remove old url

tomrf1:
## Why are you doing this?
This got missed in the PR to move to NPF. Causes selenium tests to fail


johnduffell:
if the tests are expecting the old payment flow, they are going to fail anyway even if they see the new payment flow
you can go for it as it's not going to make anything worse, but we could always add the old payment flow route back just to keep it testing at least the backend part of the app👍

PR #1170, opened 2018-11-12
Fix old contributions tests

tomrf1:
## Why are you doing this?
The selenium tests are currently failing because we've removed the routes to the old payment flow.
Now the only way to use the old payment flow is to visit /uk.

- Commit 1 - moves these old selenium tests into `/contributions/old`
- Commit 2 - fixes them (can now only use old flow for UK).

New selenium tests for the new payment flow will follow asap


johnduffell:
so the old flow is still there it's just on /uk instead of .old?
if it works that's good, this can all be updated as part of the cleaning up of old payment flow stuff
👍

PR #1172, opened 2018-11-12
Remove the `internationalSubscribePages` switch

rupertbates:
## Why are you doing this?

When we first implemented the international subscribe pages they were behind a switch in case we wanted to turn them off again. This is no longer needed so I've removed it.

This has been tested on CODE with both the old settings file and the new updated one and works fine so the deployment order should be: Merge this PR, wait till it deploys then remove the switch from `settings.json`

[**Trello Card**](https://trello.com/c/1YNXWVNb/2056-remove-old-international-non-uk-subs-landing-page-redirection)


LATaylor-guardian:
👍

PR #1173, opened 2018-11-13
Handle missing parameter in PayPal NVP responses better

rupertbates:
## Why are you doing this?

While I was investigating a high 500 rate on support-frontend I noticed that one of them was because we were not handling an unexpected response from PayPal properly. This a PR that copes with this and also adds a bit more logging so we can hopefully diagnose the problem a bit more easily.

### DIFF
--- a/app/controllers/PayPalRegular.scala
+++ b/app/controllers/PayPalRegular.scala
@@ -37,15 +37,21 @@ class PayPalRegular(
         returnUrl = routes.PayPalRegular.returnUrl().absoluteURL(secure = true),
         cancelUrl = routes.PayPalRegular.cancelUrl().absoluteURL(secure = true)
       )(paypalBillingDetails)
-    }.map { response =>
-      Ok(Token(response).asJson)
+    }.map { maybeString =>
+      maybeString
+        .map(s => Ok(Token(s).asJson))
+        .getOrElse(BadRequest("We were unable to set up a payment for this request (missing PayPal token)"))
     }
   }
 
   def createAgreement: Action[Token] = maybeAuthenticatedAction().async(circe.json[Token]) { implicit request =>
     withPaypalServiceForRequest(request) { service =>
       service.createBillingAgreement(request.body)
-    }.map(token => Ok(Token(token).asJson))
+    }.map { maybeString =>
+      maybeString
+        .map(s => Ok(Token(s).asJson))
+        .getOrElse(BadRequest("We were unable to create an agreement for this request (missing PayPal token)"))
+    }
   }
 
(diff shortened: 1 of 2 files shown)


app/controllers/PayPalRegular.scala
```diff
@@ -37,15 +37,21 @@ class PayPalRegular(
         returnUrl = routes.PayPalRegular.returnUrl().absoluteURL(secure = true),
         cancelUrl = routes.PayPalRegular.cancelUrl().absoluteURL(secure = true)
       )(paypalBillingDetails)
-    }.map { response =>
-      Ok(Token(response).asJson)
```
johnduffell: good old toString strikes again

app/services/PayPalNvpService.scala
```diff
@@ -56,9 +57,14 @@ class PayPalNvpService(apiConfig: PayPalConfig, wsClient: WSClient) extends Touc
 
   // Takes an NVP response and retrieves a given parameter as a string.
   private def retrieveNVPParam(response: QueryString, paramName: String) =
-    response.paramMap(paramName).head
+    response.paramMap(paramName).headOption match {
+      case None =>
+        SafeLogger.warn(s"Parameter $paramName was missing from the NVP response - $response")
+        None
+      case Some(value) => Some(value)
```
johnduffell: not sure if it's the case here, but for deterministicness sake I tend to avoid headOption if I expect exactly one item, usually I do a pattern match on a head :: Nil and then send everything else to the bin

johnduffell: also you are doing a pattern match and returning exactly the same thing which is a bit surprising - could just put the overall headOption into a val and have a `if (val.isEmpty)  logger.warn` and then return the val at the end

app/controllers/PayPalRegular.scala
```diff
@@ -37,15 +37,21 @@ class PayPalRegular(
...
+        .map(s => Ok(Token(s).asJson))
+        .getOrElse(BadRequest("We were unable to set up a payment for this request (missing PayPal token)"))
     }
   }
 
   def createAgreement: Action[Token] = maybeAuthenticatedAction().async(circe.json[Token]) { implicit request =>
     withPaypalServiceForRequest(request) { service =>
       service.createBillingAgreement(request.body)
-    }.map(token => Ok(Token(token).asJson))
+    }.map { maybeString =>
```
johnduffell: these maybeString are actually maybeTokenString

johnduffell:
thanks for looking at this 👍 I made a few post merge comments but nothing to affect the functionality just general comments!

PR #1176, opened 2018-11-13
Flash sale november 18

rupertbates:
## Why are you doing this?

To support the November flash sale on Print, Print+ & Digital Pack.

There are quite a few changes because we haven't done a flash sale on Print and Print+ since we moved the landing pages from subscribe to support so I've had to make the 'SubscriptionBundle' components and the launched the Paper featured product header 'flash sale aware'. This does mean that the next time we run one of these sales it should be a lot less work.

[**Trello Card**](https://trello.com/c/rlQle5hY/2068-print-print-and-dp-flash-sale)

## Changes

* Set dates and promo codes for next flash sale
* Added `Paper` & `PaperAndDigital` to the `includedProducts` array
* Made the `paper` and `paperDigital` bundle components vary their copy when the flash sale is on
* Made the Paper featured product header vary its copy when the flash sale is on
* Made Paper the default featured product for the UK while the flash sale is on

### New paper copy in feature header
screen shot 2018-11-13 at 16 33 43

### Updated bundle copy
screen shot 2018-11-13 at 16 27 31
screen shot 2018-11-13 at 16 27 15

### Flash sale copy on the Digital pack page
screen shot 2018-11-13 at 16 27 49


assets/components/subscriptionBundles/paperDigital.jsx
```diff
@@ -3,33 +3,46 @@
...
-  countryGroupId: CountryGroupId,
   abTest: ComponentAbTest | null,
   gridId: 'paperDigitalCirclePink' | 'paperDigitalCircleOrange'
 |};
 
 
 // ----- Component ----- //
 
+function getCopy() {
+  if (flashSaleIsActive('Paper')) {
```
JustinPinner: Should this be `flashSaleIsActive('PaperAndDigital')` perhaps?

assets/pages/subscriptions-landing/components/featuredProducts.jsx
```diff
@@ -95,6 +96,22 @@ const paperImage = (
     fallbackImgType="png"
   />);
 
+const getDigitalPackSubHeadingAndBody = (countryGroupId: CountryGroupId) => {
+  if (flashSaleIsActive('DigitalPack')) {
+    return {
+      subheadingText: 'Save 50% for three months',
+      bodyText: `Read the Guardian ad-free on all devices, including the Premium App and Daily Edition iPad app. ${discountedDisplayPrice('DigitalPack', countryGroupId)} for your first three months.`,
```
JustinPinner: I wish we would say '... ad-free on desktop and mobile devices ...' here. Not a problem for this PR but something to raise elsewhere I guess.

assets/components/subscriptionBundles/paperDigital.jsx
rupertbates: Good spot!

PR #1177, opened 2018-11-13
Show survey on NPF thank you pages

tomrf1:
## Why are you doing this?
We need to show the survey on the new payment flow thank you pages

picture 19
picture 20
picture 21

Users will not be shown the button if we ask them to sign up (but on the next page they will see it)
picture 22


assets/components/headers/countrySwitcherHeader/countrySwitcherHeaderContainerWithTracking.js
```diff
@@ -7,7 +7,7 @@ import { connect } from 'react-redux';
 import CountrySwitcherHeader from 'components/headers/countrySwitcherHeader/countrySwitcherHeader';
 
 import { countryGroups, type CountryGroupId } from 'helpers/internationalisation/countryGroup';
-import type { CommonState } from 'helpers/page/page';
+import type { CommonState } from 'helpers/page/commonReducer';
```
tomrf1: unrelated to this PR, but needed to fix the build

PR #1186, opened 2018-11-15
Extract out GW page components

walaura:
## Why are you doing this?

Extracts out `weeklyCta`, `weeklyFeatureList`, and `weeklyForm` as reusable components so they can be used in the upcoming print subs page.

[**Trello Card**](https://trello.com/c/Rjq5SmCR/2057-print-product-page-v0)

## Changes

* `weeklyCta` 👉 `productPageButton`
* `weeklyFeatureList` 👉 `productPageFeatures`
* `weeklyForm` 👉 `productPagePeriodForm`

This last one is the most involved one as it now has its own scoped redux container. The form component itself has become a bit simpler as well as the brunt of the logic is not handled by a wrapper component.

## Screenshots

Nothing has changed visually!

screenshot 2018-11-15 at 11 59 58 am


assets/components/productPage/productPagePeriodForm/productPagePeriodForm.jsx
```diff
@@ -0,0 +1,88 @@
...
+
+export default function ProductPagePeriodForm<P:string>({
+  periods, selectedPeriod, onSubmitAction, setPeriodAction,
+}: PropTypes<P>) {
+
+  const keys = Object.keys(periods);
+
+  return (
+    <form
+      className="component-product-page-period-form-wrap"
```
Ap0c: The top-level `className` takes the name of the component, in this case `component-product-page-period-form`.

assets/components/productPage/productPagePeriodForm/productPagePeriodFormActions.js
```diff
@@ -0,0 +1,19 @@
+// @flow
+import { sendTrackingEventsOnClick, type SubscriptionProduct } from 'helpers/subscriptions';
+
+export type Action<P> = { type: 'SET_PERIOD', period: P, scope: string };
+
+function productPagePeriodFormActionsFor<P>(scope: string, product: SubscriptionProduct | null) {
+  function setPeriod(period: string): (dispatch: Function) => Action<P> {
```
Ap0c: Dispatch has the type `Dispatch<A>`, where `A` is the action type. This can be imported from `redux`.

assets/components/productPage/productPagePeriodForm/productPagePeriodFormLabel.jsx
```diff
@@ -0,0 +1,44 @@
...
+
+
+// ----- Render ----- //
+
+export default ({
+  type, title, offer, children, checked, onChange,
+}: PropTypes) => {
+  const id = uuidv4();
+  return (
+    <label onChange={onChange} htmlFor={id} className="component-product-page-period-form-label-wrap">
```
Ap0c: The top-level `className` takes the name of the component, in this case `component-product-page-period-form-label`.

assets/pages/weekly-subscription-landing/components/weeklyForm.jsx
```diff
@@ -2,90 +2,69 @@
...
-});
+const mapStateToProps = (state: State): StatePropTypes<WeeklyBillingPeriod> => {
+  const periods = {};
+  Object.keys(billingPeriods).forEach((k) => {
+    periods[k] = {
+      title: billingPeriods[k].title,
+      copy: billingPeriods[k].copy(state.common.internationalisation.countryGroupId),
+      offer: billingPeriods[k].offer || null,
+    };
+  });
```
Ap0c: Optionally, if you wanted to do this without creating an intermediate mutable object I think `reduce` may work:

```jsx
const periods = Object.keys(billingPeriods).reduce((ps, k) => ({
  ...ps,
  [k]: {
    title: billingPeriods[k].title,
    copy: billingPeriods[k].copy(state.common.internationalisation.countryGroupId),
    offer: billingPeriods[k].offer || null,
  },
}), {});
```

assets/components/productPage/productPagePeriodForm/productPagePeriodFormReducer.js
```diff
@@ -0,0 +1,26 @@
+// @flow
+// ----- Imports ----- //
+import type { Action } from './productPagePeriodFormActions';
+
+export type State<Period> = {
+  period: Period | null;
+};
+
+// ----- Reducer ----- //
+function productPagePeriodFormReducerFor<Period>(scope: string, initialState: State<Period>) {
```
Ap0c: I think you could also pass in `period` as the second argument here, and then define the initial state in this function:

```jsx
function productPagePeriodFormReducerFor<Period>(scope: string, period: Period) {

  const initialState = { period };
  ...
}
```

That way the initial state is defined alongside the reducer and its `State` type. I don't know if this is better, it's just an alternative.

assets/pages/weekly-subscription-landing/components/weeklyForm.jsx
walaura: omg this is great!

assets/components/productPage/productPagePeriodForm/productPagePeriodFormReducer.js
walaura: your suggestion makes the whole module a bit more portable considering the state object is always encapsulated outside of this, I like it! 👍

PR #1190, opened 2018-11-15
Digital Subscription Checkout Validation

Ap0c:
## Why are you doing this?

To apply validation rules to the digital subscription checkout, and to make the validation code generic enough to use on other checkouts.

This introduces a new `validate` function, which takes a list of rules and corresponding error messages. It then returns a series of error messages based on these rules. Also included is a `FormError` type that combines a form field with an error message, which can be used in conjunction with this validation framework (or not, any error message format should work). The resultant error messages are then used in combination with the `withError` HOC to display error messages on the page (see the screenshot below).

[**Trello Card**](https://trello.com/c/yH3GBpu7/1782-mandatory-fields-incomplete-error-messaging)

cc @JustinPinner 

## Changes

- Tweaked `withError` to use `error` as a prop instead of `message`.
- Added a new `validation` helper.
- Added a new `Option` type and corresponding `headOption` helper for arrays.
- Updated the `CheckoutForm` components to use the `withError` HOC.
- Updated the checkout reducer and actions to run validation rules and update the state with any errors.

## Screenshots

form-errors

### DIFF
--- a/assets/components/forms/formHOCs/withError.jsx
+++ b/assets/components/forms/formHOCs/withError.jsx
@@ -7,23 +7,56 @@ import React from 'react';
 
 // ----- Types ----- //
 
-type AugmentedProps<Props> = Props & {
-  message: string | null,
+type Option<A> = A | null;
+
+export type FormError<FormField> = {
+  field: FormField,
+  message: string,
+};
+
+type AugmentedProps<Props, FormField> = Props & {
+  errors: FormError<FormField>[],
+  fieldName: FormField,
 };
 
 type In<Props> = React$ComponentType<Props>;
-type Out<Props> = React$ComponentType<AugmentedProps<Props>>;
+type Out<Props, A> = React$ComponentType<AugmentedProps<Props, A>>;
+
+
+// ----- Functions ----- //
+
+function headOption<A>(arr: Array<A>): Option<A> {
+  return arr[0] || null;
+}
+
+function firstError<FormField>(field: FormField): FormError<FormField>[] => Option<string> {
+
+  return (errors) => {
+    const msgs = errors.filter(err => err.field === field).map(err => err.message);
+    return headOption(msgs);
+  };
+
+}
 
 
 // ----- Component ----- //
 
-function withError<Props: { id: string }>(Component: In<Props>): Out<Props> {
-  return ({ message, ...props }: AugmentedProps<Props>) => (
-    <div>
-      <Component {...props} />
-      {message ? <label className="component-error" htmlFor={props.id}>{message}</label> : null}
-    </div>
-  );
+function withError<Props: { id: string }>(Component: In<Props>): Out<Props, *> {
+
+  function WrapperComponent<FormField>({ errors, fieldName, ...props }: AugmentedProps<Props, FormField>) {
+
+    const error = firstError(fieldName)(errors);
+
+    return (
+      <div>
+        <Component {...props} />
+        {error ? <label className="component-error" htmlFor={props.id}>{error}</label> : null}
+      </div>
+    );
+
+  }
+
+  return WrapperComponent;
 }
 
 
--- a/assets/components/forms/formHOCs/withError.jsx
+++ b/assets/components/forms/formHOCs/withError.jsx
@@ -4,59 +4,30 @@
 
 import React from 'react';
 
+import { type Option } from 'helpers/types/option';
+
 
 // ----- Types ----- //
 
-type Option<A> = A | null;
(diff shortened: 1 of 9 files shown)


assets/components/forms/formHOCs/withError.jsx
```diff
@@ -4,11 +4,13 @@
 
 import React from 'react';
 
+import { type Option } from 'helpers/types/option';
```
rupertbates: !

assets/components/forms/formHOCs/withError.jsx
```diff
@@ -4,11 +4,13 @@
...
+
 
 // ----- Types ----- //
 
 type AugmentedProps<Props> = Props & {
```
rupertbates: What do you think about calling this type `ValidatedProps`?

assets/helpers/types/option.js
```diff
@@ -0,0 +1,17 @@
+// @flow
+
+// ----- Type ----- //
+
+export type Option<A> = A | null;
```
rupertbates: Is it potentially confusing that this `Option` is not really very much like Scalas?
I think I might expect to be able to map over it for instance.

assets/helpers/subscriptionsForms/__tests__/validationTest.js
```diff
@@ -0,0 +1,100 @@
...
+  notNull,
+  firstError,
+  formError,
+  validate,
+} from '../validation';
+
+
+// ----- Tests ----- //
+
+describe('validation', () => {
```
rupertbates: 👍

assets/components/forms/formHOCs/withError.jsx
Ap0c: Universally across all the `formHOCs`?

assets/helpers/types/__tests__/optionTest.js
```diff
@@ -0,0 +1,36 @@
...
+
+    it('retrieves the first item from an array', () => {
+      expect(headOption([1, 2, 3])).toBe(1);
+    });
+
+    it('returns null when the array is empty', () => {
+      expect(headOption([])).toBeNull();
+    });
+
+    it('returns null when the array contains null', () => {
```
akash1810: Could we fold this, and the next two tests into `retrieves the first item from an array` above?

assets/helpers/types/option.js
Ap0c: Fair point. I introduced this because I was writing `A | null` a lot and I wanted to formalise a type to handle that. I went with `Option` because I thought people who have seen that in Scala would be familiar with it. However, if you think it might cause confusion there are some alternatives:

1. Use a different name, e.g. `Maybe` which is used in Haskell. This is also the name Flow gives to `?A`; I don't know whether that would be more or less confusing.

2. Write a couple of helpers to bring the functionality closer to Scala, starting with `map`.

assets/helpers/types/__tests__/optionTest.js
Ap0c: 👍 done.

assets/helpers/types/option.js
rupertbates: Happy to go with it as is for now and add in helpers when they become useful. It would be great to try this out as well: https://medium.com/inside-rimeto/optional-chaining-in-typescript-622c3121f99b
but obviously requires TS

assets/components/forms/formHOCs/withError.jsx
rupertbates: If validation is the only 'augmentation' I think it would make more sense yes

Ap0c: `asControlled`, `canShow` (coming soon), `withError` and `withLabel` all have this concept of `AugmentedProps`. The name could be updated to make it clearer, but I'd prefer it if it were one that worked across the different HOCs to keep them consistent.

assets/pages/digital-subscription-checkout/components/checkoutForm.jsx
```diff
@@ -76,7 +91,9 @@ function CheckoutForm(props: PropTypes) {
         type="tel"
         value={props.telephone}
         setValue={props.setTelephone}
+        error={firstError('telephone', props.errors)}
       />
+      <button onClick={() => props.submitForm()}>Submit</button>
```
walaura: can we make this a regular `<form>` element using `onSubmit`? That way we get some a11y goodness plus you can the enter key after the last field

PR #1192, opened 2018-11-15
Fix paypal one-off GA events

tomrf1:
## Why are you doing this?
Currently when sending data to payment-api, the gaId value is always `1` - get the `_ga` cookie value instead.


joelochlann:
Nice one!

PR #1193, opened 2018-11-15
Test shrinking the header on mobile

joelochlann:
- Removes dead A/B test code from the NPF test
- Adds a test for making more of the screen space on mobile
- Renames `customSegmentCondition` to `canRun` and lets it actually remove a test participation if not true (previously once the participation had made it into localStorage, `canRun` being false could not remove it)
- Only runs the test when coming from the epic or banner

## Screenshots
These are on Chrome's iPhone 5 emulator but on a real phone there is even less space due to the address bar at the top and navigation buttons at the bottom. On my iPhone you can't even see the amounts - just the single/monthly/annual tabs

| `shrink_no-blurb_no-header` | `shrink_no-blurb` | `shrink` | `control` |
| ------------ | ----------------| -----| -----|
| support thegulocal com_uk_contribute iphone 5_se 4 | support thegulocal com_uk_contribute iphone 5_se 3 | support thegulocal com_uk_contribute iphone 5_se 2 | support thegulocal com_uk_contribute iphone 5_se 1 |

@ionamckendrick


jranks123:
I think you will need to update the wiki to change references to `customSegmentCondition` to `canRun`, and maybe a couple of other things https://github.com/guardian/support-frontend/wiki/Client-side-AB-Testing#implementation-of-a-test

PR #1194, opened 2018-11-16
Send user-agent to payment-api

tomrf1:
## Why are you doing this?
I'm still not seeing paypal one-offs in GA. Perhaps it's because Play is currently sending `AHC/2.0` for the user-agent


tomrf1:
follow up - this does fix it. GA is filtering out this user-agent

joelochlann:
Good spot!

PR #1197, opened 2018-11-16
Selenium tests for NPF

tomrf1:
## Why are you doing this?
Adds post-deploy selenium tests for the new payment flow.
This PR only covers the same cases as the old payment flow tests. There are lots more test cases to be covered by future PR(s)


test/selenium/contributions/new_flow/OneOffContributionsSpec.scala
```diff
@@ -0,0 +1,68 @@
...
+  override def beforeAll(): Unit = {
+    Config.printSummary(driverConfig.sessionId)
+    Dependencies.dependencyCheck
+  }
+
+  override def afterAll(): Unit = { driverConfig.quit() }
+
+  feature("Sign up for a one-off contribution") {
+
+    scenario("One-off contribution sign-up with Stripe - AUD") {
```
joelochlann: wow, did we really only test Australia for one-offs on the old flow?

PR #1204, opened 2018-11-19
Lower password max-length to 20

tomrf1:
## Why are you doing this?
Making the password rules consistent with [identity-api](https://github.com/guardian/identity/blob/master/identity-api/src/main/scala/com/gu/identity/api/postgres/PostgresPasswordHashRepository.scala#L20)

Invalid password entry has been the cause of _some_ of the 500 errors on support-frontend lately - it's not surprising that some users try it repeatedly given that the message is incorrect here.

## Screenshots
picture 6

For the record, these are logged by support-frontend as:
```
Failed to set password using guest account registration token...
{ "status": "error", "errors": [ { "message": "Invalid password", "description": "The password provided is invalid" } ] }
```

### DIFF
--- a/assets/pages/new-contributions-landing/components/SetPassword/SetPasswordForm.jsx
+++ b/assets/pages/new-contributions-landing/components/SetPassword/SetPasswordForm.jsx
@@ -126,7 +126,7 @@ function SetPasswordForm(props: PropTypes) {
           value={props.password}
           onInput={props.updatePassword}
           pattern={'^.{6,20}$'}
-          isValid={props.password.length >= 6 && props.password.length <= 72}
+          isValid={props.password.length >= 6 && props.password.length <= 20}
           formHasBeenSubmitted={props.passwordHasBeenSubmitted}
           errorMessage="Please enter a password between 6 and 20 characters long"
           required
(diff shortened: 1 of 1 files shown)


tomrf1:
@jranks123 ah, that must be incorrect as well...unless the validation in identity-api needs to be updated

joelochlann:
Thanks for investigating this @tomrf1! Worth checking with @Mullefa whether it's the placeholder message or the identity-api validation that needs updating

assets/pages/new-contributions-landing/components/SetPassword/SetPasswordForm.jsx
```diff
@@ -125,10 +125,10 @@ function SetPasswordForm(props: PropTypes) {
           autoComplete="off"
           value={props.password}
           onInput={props.updatePassword}
-          pattern={'^.{6,72}$'}
+          pattern={'^.{6,20}$'}
           isValid={props.password.length >= 6 && props.password.length <= 72}
```
joelochlann: oops, you need to change this line too!

joelochlann: `props.password.length <= 72`

tomrf1: ah, thanks! I assumed `pattern` was used for the check

jranks123:
@tomrf1 So I've just checked, and via `https://profile.theguardian.com` you can definitely set your password up to 72 characters long. So I guess identity should probably change their API to be in line with this? But for now, I'd go ahead with this

PR #1209, opened 2018-11-19
Refactor form submission blocking logic into one place

jranks123:
## Why are you doing this?
Previously, the way we blocked a PayPal recurring contribution at the point of clicking the payment button was different from how we blocked all other types of payment. The PayPal Express Checkout button forces us to be pre-active:  has to be validated or invalidated on the changing of the form: it can't happen on the click of the button. So every time something on the form changed, we were enabling or disabling the PayPal button. But for everything else, we were being reactive: we would check the state of the form variables on the payment button being clicked. 

This mixture of blocking strategies meant that the same logic was duplicated in multiple places, and was a bit tricky to get your head around unless you knew the codebase well. 

The new approach goes with the reactive approach: when a form field variable is changed, we determine whether or not the form should be submittable, then set a new bit of state called `formIsSubmittable` at the same time as enabling/disabling the PayPal button. This means that all of our logic about whether the form should submit now sits in one place.

Because we now want some state changes in the `UserState` to change a variable in the page state for `ContributionsLanding` (`isRecurringContributor` and `isSignedIn` both affect the value of `formIsSubmittable`), I've made it so that each page that uses the user reducer provides its own set of User Actions, that way everyone else can use the existing ones and we can use the same set, with `isRecurringContributor` and `isSignedIn` modified to enable or disable the form at the same time as setting the value in the user state.


assets/helpers/checkoutForm/checkoutForm.js
```diff
@@ -87,15 +84,22 @@ export const formElementIsValid = (formElement: Object | null) => {
...
-  form: Object | null,
-) {
-  return formElementIsValid(form)
-    && canContributeWithoutSigningIn(contributionType, isSignedIn, userTypeFromIdentityResponse)
+  canContributeWithoutSigningIn: boolean,
+  formValid: boolean,
+|}
+
+
+export function checkoutFormShouldSubmit(params: CheckoutFormShouldSubmitParameters) {
```
jranks123: This function is now the one place where we decide whether the form should submittable or not.

assets/pages/new-contributions-landing/contributionsLandingActions.js
```diff
@@ -193,10 +179,43 @@ const sendFormSubmitEventForPayPalRecurring = () =>
...
+
     dispatch(setStateValue(value));
-    dispatch(togglePayPalButton());
+
+    const state = getState();
+
+    // TODO: use the actual form state rather than re-fetching from DOM
+    const formIsValid = formElementIsValid(getForm('form--contribution'));
+
+    dispatch(setFormIsValid(formIsValid));
```
jranks123: In a future PR, I am going to get rid of the setFormIsValid action, and make `setFormIsSubmittable` take a list of reasons (which can be the form is invalid, or they are an existing recurring contributor, or we are waiting on finding out their User Type from identity, etc).

jranks123: This will lessen the number of side effects here, and centralise/simplify the error message display logic, which I think would be nice. I like the idea of the complicated logic happening in once place (here), then the individual components not having to do too much work to figure out what do display.

assets/pages/new-contributions-landing/contributionsLandingActions.js
```diff
@@ -193,10 +179,43 @@ const sendFormSubmitEventForPayPalRecurring = () =>
...
+      state.page.form.contributionType,
+      state.page.user.isSignedIn,
+      state.page.form.userTypeFromIdentityResponse,
+    ),
+    isRecurringContributor: state.page.user.isRecurringContributor,
+  });
+
+  if (shouldEnable && window.enablePayPalButton) {
+    dispatch(setFormIsSubmittable(true));
+    window.enablePayPalButton();
```
joelochlann: could we move these inside the actions? so that whenever the state is changed the PayPal button will be in sync

assets/pages/new-contributions-landing/contributionsLandingActions.js
```diff
@@ -193,10 +179,43 @@ const sendFormSubmitEventForPayPalRecurring = () =>
...
+  } else if (window.disablePayPalButton) {
+    dispatch(setFormIsSubmittable(false));
+    window.disablePayPalButton();
+  }
+}
+
+/*
+  Set the value of a form variable, then set the form validity and submitability
+*/
+function setValue<T>(setStateValue: T => Action, value: T) {
```
joelochlann: maybe call this `setFormFieldValue`? `setValue` is a bit general

assets/pages/new-contributions-landing/contributionsLandingActions.js
joelochlann: Great idea!

assets/pages/new-contributions-landing/contributionsLandingActions.js
```diff
@@ -193,10 +179,43 @@ const sendFormSubmitEventForPayPalRecurring = () =>
     onFormSubmit(formSubmitParameters);
   };
 
-function setValueAndTogglePayPal<T>(setStateValue: T => Action, value: T) {
-  return (dispatch: Function): void => {
+
+function enableOrDisableForm(formIsValid: boolean, state: State, dispatch: Function) {
+
+  const shouldEnable = checkoutFormShouldSubmit({
```
joelochlann: maybe this function can be called `checkoutFormIsSubmittable` to match the new language

assets/helpers/checkoutForm/checkoutForm.js
joelochlann: since this function is now beautifully concise and only called in one place, you could simply inline it in `enableOrDisableForm`, hopefully discouraging people from calling it themselves in new places

assets/pages/new-contributions-landing/contributionsLandingActions.js
jranks123: I have just realised why this is called `setValue`. As well as being used to set the state value of all the form fields, it is also used to set the state value of `userTypeFromIdentityResponse`, which isn't a form field. T

assets/helpers/render.js
```diff
@@ -3,11 +3,11 @@
 import ReactDOM from 'react-dom';
 import { logException } from 'helpers/logger';
 
-const renderPage = (content: Object, id: string) => {
+const renderPage = (content: Object, id: string, callBack?: () => void) => {
```
jranks123: We have to do this because before we render we set some state variables in `ContributionsLandingInit`, and this was invalidating the form because it was trying to access the DOM elements which didn't exist yet. We can remove this once we stop using the constraints validation API all together, which we have planned.

assets/pages/new-contributions-landing/components/ContributionErrorMessage.jsx
```diff
@@ -46,7 +46,7 @@ function ContributionErrorMessage(props: PropTypes) {
 
   if (props.paymentError) {
     return (<PaymentFailureMessage checkoutFailureReason={props.paymentError} />);
-  } else if (!props.formIsValid) {
+  } else if (!props.formIsValid && props.checkoutFormHasBeenSubmitted) {
```
jranks123: We only want this error to show once the form has been submitted once

assets/helpers/checkoutForm/onFormSubmit.js
```diff
@@ -11,25 +10,21 @@ export type FormSubmitParameters = {
...
-  form: Object | null,
   isSignedIn: boolean,
   userTypeFromIdentityResponse: UserTypeFromIdentityResponse,
-  setFormIsValid: boolean => void,
+  formIsValid: boolean,
   setCheckoutFormHasBeenSubmitted: () => void,
   handlePayment?: () => void,
-  isRecurringContributor: boolean,
+  formIsSubmittable: boolean,
+  formIsValid: boolean,
```
tsop14: Repeat of line 15.

assets/helpers/checkoutFormIsSubmittableActions.js
```diff
@@ -0,0 +1,64 @@
+// @flow
+
+// ----- Imports ----- //
+
+import { canContributeWithoutSigningIn } from 'helpers/identityApis';
+import type { State } from '../pages/new-contributions-landing/contributionsLandingReducer';
```
joelochlann: we shouldn't have helpers depending on page-specific code

joelochlann: since page-specific code often depends in turn on helper functions, this kind of thing will make circular dependencies more likely.

jranks123: Ok, really I just want it to be able to be called by both the UserActions and ContributionsLandingActions. Could I move this to the same level as `ContributionsLandingActions`? What do you think makes sense?

joelochlann: ohhhhh I see.... hmmm

assets/helpers/render.js
joelochlann: Nice of `ReactDom.render` to accept a callback argument!

assets/helpers/user/userActions.js
```diff
@@ -18,54 +15,17 @@ export type Action =
...
-  return { type: 'SET_TEST_USER', testUser };
-}
-
-export function setPostDeploymentTestUser(postDeploymentTestUser: boolean): Action {
-  return { type: 'SET_POST_DEPLOYMENT_TEST_USER', postDeploymentTestUser };
-}
-
-export function setGnmMarketing(preference: boolean): Action {
-  return { type: 'SET_GNM_MARKETING', preference };
+export type UserSetStateActions = {
```
joelochlann: can we make this an exact type?

assets/pages/new-contributions-landing/setUserStateActions.js
```diff
@@ -0,0 +1,23 @@
...
+  (dispatch: Function): void => {
+    dispatch(setFormSubmissionDependentValue(() => ({ type: 'SET_IS_SIGNED_IN', isSignedIn })));
+  };
+
+const setIsRecurringContributor = (): ((Function) => void) =>
+  (dispatch: Function): void => {
+    dispatch(setFormSubmissionDependentValue(() => ({ type: 'SET_IS_RECURRING_CONTRIBUTOR' })));
+  };
+
+const setUserStateActions = {
```
joelochlann: re: the exact type (my gut was telling me it was a good idea but I was trying to think of a specific example!). here, if you had a typo in your override property, it would type-check if it wasn't an exact type. but with the exact type, flow would pick this up

joelochlann:
OK, two things I spotted that aren't quite right about the behaviour. (I did a mega-PR-review, @jranks123 style, and testing it locally)

- Previously, in Chrome, it would remember an email that had been previously entered into the form, even after a refresh (maybe this is something to do with the email being stored in session storage? so maybe not actually intentional functionality. but it's actually pretty nice to have. maybe we could extend it to name too?? so if you leave the form and come back, everything's pre-populated)
- If you enter email, first name, last name, and then click on "Other" but don't fill anything in the amount box, the form is submittable (popups come up, etc)

jranks123:
@joelochlann Great spot with "Other"!

PR #1211, opened 2018-11-20
Only set one-off contribution cookie for one-off contributions

jranks123:
## Why are you doing this?

Previously, we were setting the one-off contribution cookie (that `gu.com` uses to prevent displaying the epic or banner) for recurring contributors hitting the thank you page. There is an equivalent recurring contributions cookie, which is set server side, so we don't need to set anything here for recurring contributors. 

The impact of this bug is that everyone who has given a recurring contribution in the past few weeks will have both cookies. There's nothing we can really do about this, other than fix it going forwards

### DIFF
--- a/assets/pages/new-contributions-landing/contributionsLanding.jsx
+++ b/assets/pages/new-contributions-landing/contributionsLanding.jsx
@@ -12,6 +12,8 @@ import { init as pageInit } from 'helpers/page/page';
 import { renderPage } from 'helpers/render';
 import { detect, countryGroups, type CountryGroupId } from 'helpers/internationalisation/countryGroup';
 import * as user from 'helpers/user/user';
+import * as storage from 'helpers/storage';
+import type { Contrib } from 'helpers/contributions';
 import { set as setCookie } from 'helpers/cookie';
 import Page from 'components/page/page';
 import Footer from 'components/footer/footer';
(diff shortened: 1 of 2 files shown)


assets/pages/new-contributions-landing/contributionsLanding.jsx
```diff
@@ -81,10 +89,10 @@ const router = (
           exact
           path="/:countryId(uk|us|au|eu|int|nz|ca)/thankyou"
           render={() => {
-            setCookie(
-              ONE_OFF_CONTRIBUTION_COOKIE,
-              currentTimeInEpochMilliseconds.toString(),
-            );
+            // we set the recurring cookie server side
```
johnduffell: good comment

assets/pages/new-contributions-landing/contributionsLandingInit.js
```diff
@@ -84,6 +85,9 @@ const init = (store: Store<State, Action, Function>) => {
   selectDefaultPaymentMethod(state, dispatch);
   initialisePaymentMethods(state, dispatch);
   initialiseSelectedAnnualAmount(state, dispatch);
+  storage.setSession('contributionType', state.page.form.contributionType);
+  storage.setSession('paymentMethod', state.page.form.paymentMethod);
```
johnduffell: what are these for?

johnduffell:
good spot, thanks for fixing 👍

assets/pages/new-contributions-landing/contributionsLandingInit.js
jranks123: It's used in various places. We already set these when they change, this just sets them on initialisation too

PR #1212, opened 2018-11-20
Handle 'FailState' result from step functions

tomrf1:
## Why are you doing this?
If the step functions reach the `FailState`:
picture 9
then the user is not told that something failed - they are instead redirected to the thank you page.
This change ensures that the standard error is displayed.

This was discovered because a user managed to submit a US contribution with no state selected.

## Screenshots

picture 8

### DIFF
--- a/app/services/stepfunctions/RegularContributionsClient.scala
+++ b/app/services/stepfunctions/RegularContributionsClient.scala
@@ -169,7 +169,7 @@ object StepFunctionExecutionStatus {
     val searchForFinishedCheckout: Option[StatusResponse] = detailedHistory.collectFirst {
       case detailsAttempt if detailsAttempt.map(_.getName) == Success("CheckoutSuccess") =>
         StatusResponse(Status.Success, trackingUri, None)
-      case detailsAttempt if detailsAttempt.map(_.getName) == Success("CheckoutFailure") =>
+      case detailsAttempt if detailsAttempt.map(_.getName) == Success("SucceedOrFailChoice") =>
         StatusResponse(Status.Failure, trackingUri, getFailureDetails(stateWrapper, detailsAttempt.get))
     }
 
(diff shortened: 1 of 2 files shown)


app/services/stepfunctions/RegularContributionsClient.scala
```diff
@@ -169,7 +169,7 @@ object StepFunctionExecutionStatus {
     val searchForFinishedCheckout: Option[StatusResponse] = detailedHistory.collectFirst {
       case detailsAttempt if detailsAttempt.map(_.getName) == Success("CheckoutSuccess") =>
         StatusResponse(Status.Success, trackingUri, None)
-      case detailsAttempt if detailsAttempt.map(_.getName) == Success("CheckoutFailure") =>
```
rupertbates: We are still going to need to handle cases where we reach the `CheckoutFailure` state as this is the more common one caused by card processing type errors

johnduffell: we do, this state history is a list and this state comes before both so it will behave the same as before for the old state

johnduffell:
thanks tom for spotting that we don't tell the user that it didn't work, I think this is a big improvement 👍 
the original error about state I think we should only look into if it happens more times.

app/services/stepfunctions/RegularContributionsClient.scala
tomrf1: adding to what @johnduffell said, the state history does not contain the 'FailState` even if it does reach it. This is presumably because it has `Type: Fail`
So `SucceedOrFailChoice` allows us to catch both

rupertbates: 👍

rupertbates: They might well still end up seeing a thank you page if the execution takes more than the timeout period, but this is definitely and improvement

jacobwinch:
Thanks for fixing this 👍 

For the record, it looks like I introduced this bug here: https://github.com/guardian/support-frontend/pull/820.

PR #1213, opened 2018-11-20
Stripe Apple Pay verification file to confirm domain ownership

rjbarbour-g:
## Why are you doing this?

Stripe needs to verify domain ownership to activate Apple Pay for our domain. This file, provided by Stripe, will be checked by them to confirm this. It is expected to be at the path of the new route.


joelochlann:
We actually have some metrics in GA for how many users have the Payment Request API available with a payment method set up. I'll show you tomorrow. It will be really interesting to see if that percentage goes up after this change!

joelochlann:
Here's Stripe's documentation on their Payment Request API button: https://stripe.com/docs/stripe-js/elements/payment-request-button

PR #1214, opened 2018-11-20
Correct id on contribution tab

joelochlann:
Fixes id changed in #1207


joelochlann:
thanks @tomrf1, sorry for semi-deliberately breaking your post-deployment tests... they work like a charm!

PR #1215, opened 2018-11-20
Fetch paper subs prices from S3

walaura:
## Why are you doing this?
Fetch paper subs prices from S3 and format them in a client-side safe format. Reviews are hugely welcome as the method for adding up prices contains lots of assumptions.

### Why does the format look like this?

We have different types of pricing types, some are discounts (linked to other prices), some are a lump 1 time payment (6 for 6), some are a subscription period, some are something different altogether. For the scope of print subs we are only interested in the payment per period but we want the shape of this object to be able to accomodate growth as it will be used more extensively in the future.

As such, `pricePerPeriod` is named extensively instead of being just `price` in case we want to incorporate these aditional pricing types in the future. The price itself is returned as an array of currencies+prices so we can do currency switching. (Even though right now it's hardcoding GBP)

## Screenshots
screenshot 2018-11-21 at 1 25 29 pm


app/codecs/CirceDecoders.scala
```diff
@@ -119,6 +120,27 @@ object CirceDecoders {
   implicit val payPalErrorBodyDecoder: Decoder[PayPalError] = deriveDecoder
   implicit val payPalSuccessDecoder: Decoder[PayPalSuccess] = deriveDecoder
 
+  implicit val pricingDecoder: Decoder[Pricing] = deriveDecoder
```
LATaylor-guardian: Circe will be able to do this with automatic derivation so the decoders and encoders shouldn't be needed. Happy to show you how.

walaura: would love to know!

app/controllers/Subscriptions.scala
```diff
@@ -124,6 +126,12 @@ class Subscriptions(
     Ok(views.html.main(title, id, js, css, description)).withSettingsSurrogateKey
   }
 
+  def paperPrices(): Action[AnyContent] = NoCacheAction() { implicit request =>
```
rupertbates: I would think it's worth making this response cacheable

walaura: Good point! two things on that
1. There's a standing question of how stale can we allow the prices to get (@paulbrown1982 ?)
2. Should we be okay to make this cacheable we can include the json within the paper subs page vs making an endpoint altogether

app/models/ZuoraCatalog/ZuoraCatalog.scala
```diff
@@ -0,0 +1,72 @@
+package models.ZuoraCatalog
+
+case class ZuoraCatalogPricing(
```
rupertbates: I don't think you need the `ZuoraCatalog` prefix on all of these classes, it's enough that they're in the `ZuoraCatalog` package. In the medium term I'd like to move these classes into the support-models library so that they are accessible to other systems, some of them already exist there without the prefix so it would be good to keep things consistent.

walaura: I mostly added those to prevent the names being extremely generic (Product, Price, and so on) since the main use of these classes is pure JSON parsing rather than any actual data modeling within the code!

app/controllers/Subscriptions.scala
rupertbates: I think including it in the page would be preferable, otherwise we have to cater for the possibility of the pricing call failing. I can't see staleness being an issue as we only cache for 1 minute

app/models/ZuoraCatalog/ZuoraCatalog.scala
```diff
@@ -0,0 +1,72 @@
...
+    name: Option[String],
+    id: String,
+    productRatePlanCharges: List[ZuoraCatalogCharge]
+) {
+  def price: Double = {
+    productRatePlanCharges
+      .filter(_.endDateCondition.contains("Subscription_End"))
+      .map(
+        _.pricing
+          .filter(_.currency == "GBP")
```
paulbrown1982: I don't understand why there's a restriction to GBP?

app/services/ZuoraCatalogService.scala
```diff
@@ -0,0 +1,41 @@
...
+import services.aws.AwsS3Client.{fetchJson, s3}
+import com.amazonaws.services.s3.model.GetObjectRequest
+import com.gu.support.config.{Stage, Stages}
+import io.circe.generic.auto._
+import io.circe.syntax._
+import models.ZuoraCatalog._
+
+object ZuoraCatalogService {
+
+  val paperCollectionProductRatePlanIds = List(
```
paulbrown1982: Why did we decide to hardcode the rate plan IDs and not use the Product Benefit tag on the Product Rate Plan Charge to map the products like what happens in Membership-Common?

app/models/ZuoraCatalog/ZuoraCatalog.scala
walaura: is `pricing` always an array of currencies? right now I'm adding all GBP ones, but if it's always 1 price per currency I can map it to the final result

app/services/ZuoraCatalogService.scala
walaura: because I didn't knew about it! will have a look

app/models/ZuoraCatalog/ZuoraCatalog.scala
```diff
@@ -0,0 +1,72 @@
...
+)
+
+case class ZuoraCatalogPricePlan(
+    id: String,
+    name: Option[String],
+    pricePerPeriod: List[ZuoraCatalogPricing]
+)
+
+object ZuoraCatalogPricePlan {
+  def build(ZuoraCatalog: ZuoraCatalog, plans: List[String]): List[ZuoraCatalogPricePlan] = {
```
QuarpT: this function could be refactored into a for comprehension
you could find an alternative to using `.head`

walaura: how would you go about doing that?

walaura: How would you go about doing this?

PR #1220, opened 2018-11-21
tabs CSS

walaura:
## Why are you doing this?

We need tabs, this puts tabs in. 

There's a wrapper component using react's state instead of redux. The tabs are nonfunctional (other than flipping states around) and this component is using a mock implementation of the actual wrapper component because the actual integration task for the tabs is somewhat involved (new reducer, state changes, move URLs around, trigger refocusing, actually put in two separate pages). 

In short, a second PR will come with the functionality for the tabs, while this is purely for the HTML & CSS side of it. I tried to follow some a11y best practices **but** as our tabs are closer to links in behaviour I decided to make them so you can trigger them as buttons, and clicking one will toggle focus to the content area with all the dynamic content, which should work beautifully with screen readers!

[**Trello Card**](https://trello.com/c/1VUQ3qjk/1954-print-product-page-design-v0)

## Screenshots

screenshot 2018-11-21 at 4 24 35 pm


assets/pages/paper-subscription-landing/components/tabs.jsx
```diff
@@ -0,0 +1,33 @@
...
+
+import { outsetClassName } from 'components/productPage/productPageContentBlock/productPageContentBlock';
+import ProductPageTabs from 'components/productPage/productPageTabs/productPageTabs';
+
+class Tabs extends Component<any, any> {
+  state = {
+    active: 0,
+  }
+  onChange(active: any) {
+    this.setState({ active });
```
akash1810: 👍

assets/components/productPage/productPageTabs/productPageTabs.jsx
```diff
@@ -0,0 +1,40 @@
...
+  onChange: (id: number) => void
+|};
+
+
+// ----- Render ----- //
+
+const ProductPageTabs = ({
+  tabs, active, onChange,
+}: PropTypes) => (
+  <div role="tablist" className="component-product-page-tabs">
```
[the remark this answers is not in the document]
walaura: bit more semantic yes, good shout!

assets/components/productPage/productPageTabs/productPageTabs.jsx
```diff
@@ -0,0 +1,40 @@
...
+
+import React from 'react';
+import { classNameWithModifiers } from 'helpers/utilities';
+
+import './productPageTabs.scss';
+
+
+// ---- Types ----- //
+type Tab = {|
+  name: string,
```
walaura: this is left open to potentially expand in the future with `title` or `href` or anything else that might crop up

PR #1221, opened 2018-11-21
Fix recurring cookie logic

jranks123:
## Why are you doing this?
This ensures that only those with the cookie set to `true` will be blocked

### DIFF
--- a/assets/helpers/user/user.js
+++ b/assets/helpers/user/user.js
@@ -51,7 +51,7 @@ const init = (dispatch: Function) => {
     dispatch(setPostDeploymentTestUser(true));
   }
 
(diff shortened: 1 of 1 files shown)


assets/helpers/user/user.js
```diff
@@ -51,7 +51,7 @@ const init = (dispatch: Function) => {
     dispatch(setPostDeploymentTestUser(true));
   }
 
-  if (getCookie('gu_recurring_contributor')) {
+  if (getCookie('gu_recurring_contributor') === 'true') {
```
joelochlann: this model feels a bit error-prone. could we wrap this cookie check in its own function? `isRecurringContributor`? maybe with a corresponding setter function?

PR #1226, opened 2018-11-22
Fix styling & standardise buttons

tsop14:
## Why are you doing this?
Fix styling on set password thank you page.
Use consistent component throughout NPF and share styling on using `button` class where possible. 

In the future it would be nice to adopt the HOC @Ap0c is creating, so all the buttons on NPF had the same base component and styling, then the form submit buttons could be composed to add required behaviour. 

[**Trello Card**](https://trello.com/c/Zagb8j3S/171-fix-button-styling-on-set-password-screen)

## Changes

* Uses ButtonWithRightArrow for all buttons which are links / form submits on NPF
* Moves reused styling to `button` class to be shared by all buttons on NPF. 
* Buttons on thank you page black on white --> white on black to be consistent with rest of site and button on set password page.

## Screenshots
screen shot 2018-11-23 at 15 20 12
screen shot 2018-11-23 at 13 56 52


assets/pages/new-contributions-landing/contributionsLanding.scss
```diff
@@ -785,39 +784,6 @@ form {
   text-decoration: underline;
 }
 
-.form__submit {
```
tsop14: These style rules are moved not removed. As they are shared across landing and thank you pages they've been put in a new section.

assets/pages/new-contributions-landing/components/SetPasswordForm.jsx
```diff
@@ -131,19 +132,20 @@ function SetPasswordForm(props: PropTypes) {
...
-          type="submit"
-          buttonCopy="Create an account"
-        />
-        <ButtonWithRightArrow
-          componentClassName={classNameWithModifiers('form__submit', ['no-thanks'])}
-          buttonClassName={classNameWithModifiers('form__submit-button', ['no-thanks'])}
-          accessibilityHintId="accessibility-hint-no-thanks"
-          type="button"
-          buttonCopy="No, thank you"
+        <div className="form__submit form__submit--set-password">
```
tsop14: No longer using a component to make this consistent with landing page. Next step will be to make both use a core component shared across site.

jranks123: I don't think we should move away from using the `ButtonWithRightArrow` class, I don't see the benefit. This is only consistent with the landing page anyway in the sense that we're not using a component, but I don't think that's worth the added noise of writing the markup here. 

Also, the buttons are so similar, I think it's confusing that they have different implementations (one just pure markup, one using `CtaLink`

jranks123: I would just keep `ButtonWithRightArrow` for this PR, then make the landing page use it (or a modified version where the arrow is optional) in the future

jranks123: But I'd be happy with going back to using the component, for now, if you'd prefer

PR #1227, opened 2018-11-22
remove session id generation code

johnduffell:
## Why are you doing this?

from https://github.com/guardian/support-models/pull/48 https://github.com/guardian/support-workers/pull/157
Originally we added deduplication on identity id to prevent multiple contributions in a very watertight way. This was because sometimes we saw duplicates when the step functions retried the create subscription step.
We added the ability to contribute when not logged in, meaning that in a previous PR I added session id so we would deduplicate all requests that came from the same browser when logged in, but didn't deduplicate if the person wasn't logged in. This would prevent as best as possible anyone who ended up double paying without leaking any information.
Unfortunately it turned out that we can only actually deduplicate in that way within a single execution of the step functions, as currently once the zuora step passes, the analytics events are sent regardless.  This would otherwise mean that with an existing subscription, it would skip the zuora subscribe, but still send an analytics event, which would then be associated with no actual contribution.
After this change we will always create a new contribution every time the step functions are called.  So there are no possible data leaks.

[**Trello Card**](https://trello.com/c/4dkYhRxM/191-tech-debt-remove-session-ids-from-support-frontend)

## Changes
This just deletes all the code to generate and submit the session id.

This is tested locally against the CODE support-workers, I will test in CODE too.


app/controllers/RegularContributions.scala
```diff
@@ -115,8 +113,8 @@ class RegularContributions(
     SafeLogger.info(s"[${request.uuid}] User ${fullUser.id} is attempting to create a new ${request.body.contribution.billingPeriod} contribution")
 
     val result = for {
-      user <- identityService.getUser(fullUser) // FIXME also need to check they are email verified before using AccessScopeNoRestriction
-      response <- client.createContributor(request, contributor(user, request.body), request.uuid, AuthenticatedAccess).leftMap(_.toString)
+      user <- identityService.getUser(fullUser)
```
johnduffell: if they are logged in, we will have already asked them on the client side to go and amend their existing contribution.  So by this time we just go ahead anyway.

johnduffell:
looks like this is working, I also deleted the old session id field from prod zuora, and one contribution was successfully created using the updated support workers and support frontend!

PR #1228, opened 2018-11-22
Tabs on print subs page

walaura:
## Why are you doing this?

Real tabs implementation for the paper subs page! With links! And focus states! 

As I said in #1220 this was quite the involved task, have a looksie at the changes

[**Trello Card**](https://trello.com/c/1VUQ3qjk/1954-print-product-page-design-v0)

## Changes

* New reducer/action for the current tab, 
* Added this route to routes.js
* `PaperDeliveryMethod ` in subscriptions to handle navigation
* Proper links in`ProductPageTabs`
* A new multiline element for um...multilines
* Tweaks to the form element so the product list can change
* CSS refactor for `ProductPageContentBlock` (it was growing unwieldy)
* The final copy!!

## Screenshots

screenshot 2018-11-22 at 2 38 41 pm


assets/pages/paper-subscription-landing/components/content.jsx
```diff
@@ -0,0 +1,146 @@
...
+  setTabAction: typeof setTab
+|};
+
+// ----- Render ----- //
+
+class Content extends Component<PropTypes> {
+
+  componentDidUpdate({ selectedTab }) {
+    if (selectedTab !== this.props.selectedTab && this.tabRef) {
+      // $FlowIgnore
```
Ap0c: What error is this ignoring?

assets/pages/paper-subscription-landing/components/content.jsx
```diff
@@ -0,0 +1,146 @@
...
+          </ProductPageTextBlock>
+        </ProductPageContentBlock>
+        <ProductPageContentBlock type="feature">
+          <ProductPageTextBlock title="Now pick your perfect delivery subscription package" />
+          <ProductPageContentBlockDivider />
+          <Form />
+        </ProductPageContentBlock>
+        {faqs}
+      </div>
+    );
```
Ap0c: Can these be broken out into functional components using the "Auxiliary Components" pattern seen elsewhere in the codebase?

assets/pages/paper-subscription-landing/components/content.scss
```diff
@@ -0,0 +1,3 @@
+.paper-subscription-landing-content__focusable {
+  outline: 0;
+}
```
Ap0c: Could this live in the `paperSubscriptionLandingPage.scss` stylesheet?

assets/pages/paper-subscription-landing/paperSubscriptionLandingPageActions.js
```diff
@@ -4,11 +4,21 @@
...
+
+// ----- Types ----- //
+export type TabActions = { type: 'SET_TAB', tab: PaperDeliveryMethod }
+
 
 // ----- Action Creators ----- //
 
 const { setPlan } = ProductPagePlanFormActionsFor<PaperBillingPlan>('Paper', 'Paper');
+const setTab = (tab: PaperDeliveryMethod): TabActions => {
+  window.history.replaceState({}, null, paperSubsUrl(tab));
```
Ap0c: What's the thinking behind this?

assets/pages/paper-subscription-landing/paperSubscriptionLandingPageReducer.js
```diff
@@ -2,22 +2,56 @@
...
+  page: {
+    tab: ActiveTabState,
+    plan: FormState<PaperBillingPlan>,
+  }
 };
 
-
-export default () => {
+export default (initialTab: PaperDeliveryMethod) => {
   const promoInUrl = getQueryParameter('promo');
```
Ap0c: Can this be passed as an argument as is the case for `method/initialTab`?

assets/pages/paper-subscription-landing/components/tabs.jsx
```diff
@@ -2,32 +2,71 @@
...
+    href: paperSubsUrl('delivery'),
+  },
+};
+
+type StatePropTypes = {|
+  selectedTab: number,
+|};
+
+type DispatchPropTypes = {|
+  setTabAction: (PaperDeliveryMethod) => *,
```
Ap0c: Is the return type of this `TabActions`?

assets/pages/paper-subscription-landing/components/tabs.jsx
```diff
@@ -2,32 +2,71 @@
...
+
+type PropTypes = {|
+  ...StatePropTypes,
+  ...DispatchPropTypes,
+|};
+
+// ----- Component ----- //
+
+const Tabs = ({ selectedTab, setTabAction }: PropTypes) => (
+  <div className={outsetClassName}>
```
Ap0c: Can we avoid using component-related classes outside of the components?

assets/components/productPage/productPageContentBlock/productPageContentBlock.scss
```diff
@@ -1,54 +1,66 @@
...
+  & .component-left-margin-section__content,
+  & .component-product-page-features .component-product-page-features__item {
+    border-color: $border;
+  }
+  & .component-product-page-content-block-divider__line {
+    @include multiline(4, $border);
+  }
+  /*div > $selector adds some leeway for react wrapping elements*/
+  & + &,
+  & + div > &  {
```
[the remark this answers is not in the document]
walaura: this needs to apply to a classless `div`, the goal is to style named components that have to be wrapped in a div, in this case, the tab pages inside the tabs. This is until we get `preact.Fragment`

assets/pages/paper-subscription-landing/components/content.jsx
walaura: `focusOptions` which is not part of the flow spec yet but is [defined and documented](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus) by browser vendors and will just be ignored in browsers that don't support it

assets/pages/paper-subscription-landing/paperSubscriptionLandingPageActions.js
walaura: have the url match the selected tab as they both have canonical pages and routes

assets/pages/paper-subscription-landing/paperSubscriptionLandingPageReducer.js
walaura: yes but that would mean using `getQueryParameter` as as side effect in `paperSubscriptionLandingPage`

assets/pages/paper-subscription-landing/components/tabs.jsx
walaura: the idea of this classname is exactly to interplay with the component that exports it (it cancels out the padding, preventing hardcoding it in the sub component) it can become a component however (a div wrapping any children) which is what I've been doing with more involved components like `productPageContentBlockDivider`

assets/pages/paper-subscription-landing/components/content.scss
walaura: Any particular reason? I might be too used to CSS modules but like the idea of keeping it as close as possible to where this behaviour happening to the code, even if it's a three liner

assets/pages/paper-subscription-landing/components/content.jsx
```diff
@@ -0,0 +1,162 @@
...
+            <span>
+            If you live in Greater London (within the M25),
+            you can use The Guardian’s home delivery service. Don’t
+            worry if you live outside this area - you can still {linkToVouchers}.
+            </span>
+          ),
+          'Your paper will arrive before 7am from Monday to Saturday and before 8.30am on Sunday',
+          'Your newspaper deliveries will begin as quickly as five days from you subscribing',
+          'Delivery to individual flats, or apartments within blocks, is unavailable as we require access to your post box to deliver your paper',
+          'You can pause your subscription for up to 36 days a year. So if you’re going away anywhere, you won’t have to pay for the papers that you miss',
```
akash1810: maybe not for this PR: worth making this a link to the manage page?

assets/pages/paper-subscription-landing/components/content.jsx
```diff
@@ -0,0 +1,162 @@
...
+
+  tabRef: ?HTMLDivElement;
+
+  render() {
+    const { selectedTab, setTabAction } = this.props;
+
+    const collectionPage = (
+      <div className="paper-subscription-landing-content__focusable" tabIndex={-1} ref={(d) => { this.tabRef = d; }}>
+        <ContentVoucherFaqBlock />
+        <ContentForm title="Now pick your perfect voucher subscription package" text="When you take out a voucher subscription, we’ll send you a book of vouchers. There’s one for each newspaper in the package you choose. So if you choose a Sixday package, for example, you’ll receive six vouchers for each week, delivered every quarter" />
```
akash1810: minor, not for this PR: IIRC we got feedback that `quarter` wasn't a very accessible term and `every three months` was preferred?

assets/components/productPage/productPageTabs/productPageTabs.jsx
```diff
@@ -27,13 +29,23 @@ const ProductPageTabs = ({
...
-      {tabs.map(({ name }, index) => {
+      {tabs.map(({ name, href }, index) => {
         const isActive = index === active;
         return (
           <li className="component-product-page-tabs__li">
-            <button className={classNameWithModifiers('component-product-page-tabs__tab', isActive ? ['active'] : [])} onClick={() => onChange(index)}>
-              {name}
-            </button>
+            {href ? (
+              <a href={href} className={classNameWithModifiers('component-product-page-tabs__tab', isActive ? ['active'] : [])} onClick={(ev) => { ev.preventDefault(); onChange(index); }}>
```
akash1810: What scenario would we have an `a` over a `button`?

walaura: they look like tabs but they are links to two pages with two different canonical urls, since they have an href they make more sense as a link. I think normally Buttons are for actions rather than navigation but things like tabbars are a bit muddy (actual tabbars actually have their own set of aria roles)

This tabbar is a bit more complex on top of that because it's not really a tabbar it's navigation between two pages (there are more differences than similarities) dressed as a tabbar.

(Plus search engines can follow the links and index both pages!)

PR #1233, opened 2018-11-22
WIP: Display prices and savings

walaura:
## Why are you doing this?

Display is ugly because this is meant to be merged after #1228, all prices are a-ok 👌

## Screenshots

screenshot 2018-11-22 at 5 51 14 pm

Bonus! The new boxes ripple down to the weekly page, here's how they look:

screenshot 2018-11-22 at 5 58 43 pm

### DIFF
--- a/app/controllers/Subscriptions.scala
+++ b/app/controllers/Subscriptions.scala
@@ -143,10 +143,12 @@ class Subscriptions(
     val canonicalLink = Some(buildCanonicalPaperSubscriptionLink())
 
     def pricesToMap(prices: List[ZuoraCatalogPricePlan], prefix: String) = {
-      prices.map { price => (
-        s"$prefix-${price.name.getOrElse(price.id).toLowerCase}",
-        price.pricePerPeriod.filter(_.currency == "GBP").map(_.price).sum.toString
-      )}.toMap
+      prices.map { price =>
+        (
+          s"$prefix-${price.name.getOrElse(price.id).toLowerCase}",
+          price.pricePerPeriod.filter(_.currency == "GBP").map(_.price).sum.toString
+        )
+      }.toMap
     }
 
     val collectionPrices = pricesToMap(getPaperPrices.collection, "collection")
--- a/app/controllers/Subscriptions.scala
+++ b/app/controllers/Subscriptions.scala
@@ -144,7 +144,7 @@ class Subscriptions(
 
     def pricesToMap(prices: List[ZuoraCatalogPricePlan], prefix: String) = {
       prices.map(price => (
-        s"${prefix}-${price.id}",
+        s"$prefix-${price.name.getOrElse(price.id).toLowerCase}",
         price.pricePerPeriod.filter(_.currency == "GBP").map(_.price).sum.toString
       )).toMap
     }
--- a/app/controllers/Subscriptions.scala
+++ b/app/controllers/Subscriptions.scala
@@ -143,10 +143,10 @@ class Subscriptions(
     val canonicalLink = Some(buildCanonicalPaperSubscriptionLink())
 
     def pricesToMap(prices: List[ZuoraCatalogPricePlan], prefix: String) = {
-      prices.map(price => (
+      prices.map { price => (
         s"$prefix-${price.name.getOrElse(price.id).toLowerCase}",
         price.pricePerPeriod.filter(_.currency == "GBP").map(_.price).sum.toString
-      )).toMap
+      )}.toMap
     }
 
     val collectionPrices = pricesToMap(getPaperPrices.collection, "collection")
--- a/app/controllers/Subscriptions.scala
+++ b/app/controllers/Subscriptions.scala
@@ -7,9 +7,12 @@ import com.typesafe.scalalogging.LazyLogging
 import config.StringsConfig
 import play.api.mvc._
 import admin.{Settings, SettingsProvider, SettingsSurrogateKeySyntax, SwitchState}
+import models.ZuoraCatalog.ZuoraCatalogPricePlan
 import utils.RequestCountry._
 import views.html.helper.CSRF
+
 import scala.concurrent.ExecutionContext
+import services.ZuoraCatalogService.getPaperPrices
 
 class Subscriptions(
     actionRefiners: CustomActionBuilders,
@@ -139,7 +142,17 @@ class Subscriptions(
     val css = "paperSubscriptionLandingPage.css"
     val canonicalLink = Some(buildCanonicalPaperSubscriptionLink())
 
-    Ok(views.html.main(title, id, js, css, None, canonicalLink)).withSettingsSurrogateKey
+    def pricesToMap(prices: List[ZuoraCatalogPricePlan], prefix: String) = {
+      prices.map(price => (
+        s"${prefix}-${price.id}",
+        price.pricePerPeriod.filter(_.currency == "GBP").map(_.price).sum.toString
+      )).toMap
+    }
+
+    val collectionPrices = pricesToMap(getPaperPrices.collection, "collection")
+    val deliveryPrices = pricesToMap(getPaperPrices.delivery, "delivery")
+
+    Ok(views.html.main(title, id, js, css, None, canonicalLink, data = collectionPrices ++ deliveryPrices)).withSettingsSurrogateKey
   }
 
   def premiumTierGeoRedirect: Action[AnyContent] = geoRedirect("subscribe/premium-tier")
--- a/app/models/ZuoraCatalog/ZuoraCatalog.scala
+++ b/app/models/ZuoraCatalog/ZuoraCatalog.scala
@@ -42,7 +42,7 @@ case class ZuoraCatalogPricePlan(
 
 object ZuoraCatalogPricePlan {
 
-  def round(price: Double, precision: Int): Double =
+  def round(price: Double): Double =
     BigDecimal(price).setScale(2, BigDecimal.RoundingMode.HALF_DOWN).toDouble
 
   def build(ZuoraCatalog: ZuoraCatalog, plans: List[String]): List[ZuoraCatalogPricePlan] = {
--- a/app/models/ZuoraCatalog/ZuoraCatalog.scala
+++ b/app/models/ZuoraCatalog/ZuoraCatalog.scala
@@ -58,7 +58,7 @@ object ZuoraCatalogPricePlan {
               List(
                 ZuoraCatalogPricing(
                   "GBP",
-                  round(productRatePlan.price, 2)
+                  round(productRatePlan.price)
                 )
               )
             )))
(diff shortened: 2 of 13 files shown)


app/views/main.scala.html
```diff
@@ -94,7 +95,9 @@
     </div>
   </noscript>
 
-	<div id="@mainId"></div>
+	<div id="@mainId"
+    @{data.map{case (key, value) => s"""data-$key=$value """}}
```
walaura: this is extremely fragile

app/controllers/Subscriptions.scala
```diff
@@ -139,7 +142,17 @@ class Subscriptions(
...
+      prices.map(price => (
+        s"$prefix-${price.name.getOrElse(price.id).toLowerCase}",
+        price.pricePerPeriod.filter(_.currency == "GBP").map(_.price).sum.toString
+      )).toMap
+    }
+
+    val collectionPrices = pricesToMap(getPaperPrices.collection, "collection")
+    val deliveryPrices = pricesToMap(getPaperPrices.delivery, "delivery")
+
+    Ok(views.html.main(title, id, js, css, None, canonicalLink, data = collectionPrices ++ deliveryPrices)).withSettingsSurrogateKey
```
walaura: Not sure if we want to send just the prices or all the plan info so we can be a bit more clever instead of hardcoding the plans in both js and scala. I believe @paulbrown1982 has some thoughts on how to no hardcode them on the scala side (email incoming) and he's super right. Might be good to use this PR to refine our fetching of prices

app/models/ZuoraCatalog/ZuoraCatalog.scala
```diff
@@ -41,6 +41,10 @@ case class ZuoraCatalogPricePlan(
 )
 
 object ZuoraCatalogPricePlan {
+
+  def round(price: Double, precision: Int): Double =
+    BigDecimal(price).setScale(2, BigDecimal.RoundingMode.HALF_DOWN).toDouble
```
walaura: This was completely down to trial and error until the prices matched `subscriptions-frontend` and needs a proper investigation into how the prices are rounded

assets/components/productPage/productPagePlanForm/productPagePlanForm.jsx
```diff
@@ -16,7 +17,9 @@ import './productPagePlanForm.scss';
 export type Plan = {|
   title: string,
   copy: string,
-  offer: string | null,
+  offer: Option<string>,
+  price: Option<string>,
```
walaura: might be good to name `price` and `saving` something more data agnostic like `footerHeader` and `footerCopy`?

assets/helpers/subscriptions.js
```diff
@@ -30,7 +30,9 @@ export type ComponentAbTest = {
 export type BillingPeriod = 'sixweek' | 'quarter' | 'year' | 'month';
 export type WeeklyBillingPeriod = 'sixweek' | 'quarter' | 'year';
 
-export type PaperBillingPlan = 'everyday' | 'sixday' | 'weekend' | 'sunday';
+export type PaperBillingPlan =
+  'collectionEveryday' | 'collectionSixday' | 'collectionWeekend' | 'collectionSunday' |
+  'deliveryEveryday' | 'deliverySixday' | 'deliveryWeekend' | 'deliverySunday';
```
walaura: this is replicated from #1233 and will probs result in a merge error, #1233 is the source of truth for all conflicts

app/models/ZuoraCatalog/ZuoraCatalog.scala
```diff
@@ -41,6 +41,10 @@ case class ZuoraCatalogPricePlan(
 )
 
 object ZuoraCatalogPricePlan {
+
+  def round(price: Double, precision: Int): Double =
```
LATaylor-guardian: `precision` isn't currently being used inside this function - is it needed?

app/models/ZuoraCatalog/ZuoraCatalog.scala
LATaylor-guardian: You should use the `precision` value you're passing into the function here or remove the precision val if you don't need it to be configurable.

app/controllers/Subscriptions.scala
```diff
@@ -139,7 +142,17 @@ class Subscriptions(
     val css = "paperSubscriptionLandingPage.css"
     val canonicalLink = Some(buildCanonicalPaperSubscriptionLink())
 
-    Ok(views.html.main(title, id, js, css, None, canonicalLink)).withSettingsSurrogateKey
+    def pricesToMap(prices: List[ZuoraCatalogPricePlan], prefix: String) = {
+      prices.map(price => (
+        s"$prefix-${price.name.getOrElse(price.id).toLowerCase}",
```
LATaylor-guardian: How does the format of the key here relate?

app/controllers/Subscriptions.scala
```diff
@@ -139,7 +142,17 @@ class Subscriptions(
     val css = "paperSubscriptionLandingPage.css"
     val canonicalLink = Some(buildCanonicalPaperSubscriptionLink())
 
-    Ok(views.html.main(title, id, js, css, None, canonicalLink)).withSettingsSurrogateKey
+    def pricesToMap(prices: List[ZuoraCatalogPricePlan], prefix: String) = {
+      prices.map(price => (
```
LATaylor-guardian: Small syntactical thing with Scala:

When using map/filter etc on a single line you would write it as so with brackets:

`foo.map(f => ...)`

Whereas when over multiple lines you would use curly braces like so:

```
foo.map { f => 
...
}
```

app/views/main.scala.html
LATaylor-guardian: Nitpick: space between `map` and `{` and `case`. So: `@{data.map { case`

app/controllers/Subscriptions.scala
walaura: The keys are [hardcoded in JS](https://github.com/guardian/support-frontend/pull/1233/files#diff-5d0077fde2361c7fce60450a4338b000R40). It's a bit ugly but the other parts of this project that fetch things from the scala side use `window.{var}` globals which are also hardcoded and a bit more feeble.

A nicer way to do this code wise could be to fetch this from an api-like endpoint but that involves a network round trip. I'm considering passing the whole object as a JSON in the data attr but the keys would still have to be hardcoded in JS to define the shape of the final object

app/controllers/Subscriptions.scala
rupertbates: I'm looking into how to handle prices at the moment, let's go with this for now and I'll rework it later

app/views/main.scala.html
```diff
@@ -11,9 +13,14 @@
 	hrefLangLinks: Map[String, String] = Map(),
 	styles: Html = Html(""),
 	scripts: Html = Html(""),
+  data: Map[String, String] = Map.empty,
   csrf: Option[String] = None
 )(implicit assets: AssetsResolver, request: RequestHeader, settings: Settings)
 
+@mapToDataset(data: Map[String, String]) = @{data.map { case (key, value) =>
```
rupertbates: Maybe move this function out into the `ViewHelpers` class

assets/helpers/subscriptions.js
rupertbates: Do you think 
```
export type DeliveryBillingPlan = Everyday' | 'Sixday' | 'Weekend' | 'Sunday';
export type CollectionBillingPlan = Everyday' | 'Sixday' | 'Weekend' | 'Sunday';
```
might be better?

assets/pages/paper-subscription-landing/components/form.jsx
```diff
@@ -5,45 +5,91 @@
...
 import { type Action } from 'components/productPage/productPagePlanForm/productPagePlanFormActions';
 import ProductPagePlanForm, { type StatePropTypes, type DispatchPropTypes } from 'components/productPage/productPagePlanForm/productPagePlanForm';
 
 import { type State } from '../paperSubscriptionLandingPageReducer';
 import { setPlan } from '../paperSubscriptionLandingPageActions';
 
 
+// ---- Helpers ----- //
+
+const getPriceStr = (price: Option<number>): Option<string> =>
```
rupertbates: I think you can use the `displayPrice` function from subscriptions.js here

assets/pages/paper-subscription-landing/components/form.jsx
```diff
@@ -5,45 +5,91 @@
...
+  (price ? `From £${price} per month` : null);
+
+const getSavingStr = (subscription: Option<number>, newsstand: Option<number>): Option<string> =>
```
rupertbates: Can this function be generalised for other products? If so then move it to subscriptions.js

assets/pages/paper-subscription-landing/paperSubscriptionLandingPageReducer.js
```diff
@@ -2,22 +2,61 @@
...
 
+import { combineReducers } from 'redux';
 import type { CommonState } from 'helpers/page/commonReducer';
-import { type PaperBillingPlan } from 'helpers/subscriptions';
+import { type Option } from 'helpers/types/option';
 import { getQueryParameter } from 'helpers/url';
+import { type PaperBillingPlan } from 'helpers/subscriptions';
 import { ProductPagePlanFormReducerFor, type State as FormState } from 'components/productPage/productPagePlanForm/productPagePlanFormReducer';
 
+export type PaperPrices = {
```
rupertbates: I think I would probably model this as:
```
PaperPrices = {
    collection {
        everyday: Option<number>,
        sixday: Option<number>,
        weekend: Option<number>,
        sunday: Option<number>,
    }
    ...
```

assets/pages/paper-subscription-landing/components/form.jsx
walaura: I can't as these prices are fed from a different place, `displayPrice` uses the hardcoded list in the helper

assets/pages/paper-subscription-landing/components/form.jsx
walaura: it's ab it of a one-off for now i'd advocate for generalising it when it's actually reused

assets/pages/paper-subscription-landing/paperSubscriptionLandingPageReducer.js
walaura: that is cleaner but both the dataset fetching and setting a default ticked option from the url rely on the price list being flat, it's a pretty involved change for this PR :(

assets/pages/paper-subscription-landing/components/form.jsx
rupertbates: Ok, let's leave it for a later refactor once the price stuff is better integrated 👍

PR #1234, opened 2018-11-22
Don't use the Constraint Validation API to determine whether to submit the form or not

jranks123:
## Why are you doing this?
Using the Constraint Validation API to determine whether the form is valid was causing us some issues. There was a race condition error whereby if you change the amount to other, the change to the DOM doesn't happen in time for the correct valid state to be determined. This was currently only an issue for the PayPal button, as the rest of the payment methods evaluated the form on submit, rather than on form change, but that is changing in [this PR](https://github.com/guardian/support-frontend/pull/1209).  

It was also always the plan to move to state-based form validation anyway because that's what we use to decide whether to show an error or not, and having 2 different validation systems was always going to (and has already) cause problems. 

So we need to switch to state-based form validation. 

Note: you can see why we are moving our form validation logic into one place, as I've had to make the changes twice on this PR, once for PayPal recurring and once for everything else.

### DIFF
--- a/assets/helpers/checkoutForm/checkoutForm.js
+++ b/assets/helpers/checkoutForm/checkoutForm.js
@@ -4,6 +4,16 @@
 import { type ContributionType, contributionTypeIsRecurring } from 'helpers/contributions';
 import { type UserTypeFromIdentityResponse } from 'helpers/identityApis';
 import { canContributeWithoutSigningIn } from 'helpers/identityApis';
+import type { OtherAmounts, SelectedAmounts } from 'helpers/contributions';
+import type { CountryGroupId } from 'helpers/internationalisation/countryGroup';
+import type { CaState, UsState } from 'helpers/internationalisation/country';
+import {
+  checkAmountOrOtherAmount,
+  checkEmail,
+  checkFirstName,
+  checkLastName,
+  checkStateIfApplicable,
+} from 'helpers/formValidation';
 
 // Copied from
 // https://github.com/playframework/playframework/blob/master/framework/src/play/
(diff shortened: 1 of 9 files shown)


assets/pages/new-contributions-landing/contributionsLandingActions.js
```diff
@@ -163,13 +163,25 @@ const setUserTypeFromIdentityResponse = (userTypeFromIdentityResponse: UserTypeF
...
+      lastName: state.page.form.formData.lastName,
+      email: state.page.form.formData.email,
+    };
+
     const shouldEnable = checkoutFormShouldSubmit(
       state.page.form.contributionType,
       state.page.user.isSignedIn,
       state.page.user.isRecurringContributor,
       state.page.form.userTypeFromIdentityResponse,
       // TODO: use the actual form state rather than re-fetching from DOM
```
joelochlann: I guess you can remove this TODO!

assets/helpers/checkoutForm/checkoutForm.js
```diff
@@ -86,18 +96,48 @@ export const formElementIsValid = (formElement: Object | null) => {
 
 export const formIsValid = (formClassName: string) => formElementIsValid(getForm(formClassName));
 
+export type FormIsValidParameters = {
```
joelochlann: I like this name. But could it be even better? `ValuesWhichDecideFormValidity`? `StateWhichDecidesFormValidity`?

assets/helpers/contributions.js
```diff
@@ -80,6 +80,12 @@ type Config = {
   }
 }
 
+export type OtherAmounts = {
```
joelochlann: ❤️

joelochlann: for factoring these out

assets/helpers/formValidation.js
```diff
@@ -5,15 +5,50 @@
...
-export const isValidEmail: string => boolean = input => new RegExp(emailRegexPattern).test(input);
+export const isNotEmpty: (string | null) => boolean = input => !!input && input.trim() !== '';
+export const isValidEmail: (string | null) => boolean = input => !!input && new RegExp(emailRegexPattern).test(input);
 export const isLargerOrEqual: (number, string) => boolean = (min, input) => min <= parseFloat(input);
 export const isSmallerOrEqual: (number, string) => boolean = (max, input) => parseFloat(input) <= max;
 export const maxTwoDecimals: string => boolean = input => new RegExp('^\\d+\\.?\\d{0,2}$').test(input);
 
-export const checkFirstName: string => boolean = isNotEmpty;
-export const checkLastName: string => boolean = isNotEmpty;
+export const checkFirstName: (string | null) => boolean = isNotEmpty;
```
joelochlann: this is in anticipation of removing them from one-off?

assets/helpers/contributions.js
```diff
@@ -80,6 +80,12 @@ type Config = {
...
+  [ContributionType]: { amount: string | null }
```
joelochlann: not really your PR but I wonder whether this `amount` could be of type `Amount`? or maybe better, rename type `Amount` to `PresetAmount` since that's really what it's modelling

assets/helpers/formValidation.js
jranks123: No, it's because the type of firstName is `string | null`, so it seemed clearer to be explicit about what types it could take (it didn't actually throw an error previously, but I prefer this)

PR #1236, opened 2018-11-23
Don't display marketing permissions button without a valid email, and show error on API call failure 

jranks123:
## Why are you doing this?
We've been getting a fair few alerts concerning marketing permissions, which from what I can gather is down to 2 things:

1) We display the marketing permission button even if we haven't got a valid email to send to the Identity Consents API (this can only happen if we can't find their email in the session storage)

2) When the API call fails,  we don't provide any feedback to the user, meaning that they tend to spam the button, causing multiple errors

This PR fixes both of those issues, by not displaying the component if we haven't got a valid email address for the user, and displaying an error if the API call fails. 

| The error message being displayed |
|-------|
|image|


tsop14:
👍  and glad that the name better fits all the use cases. 

One suggestion, might be useful to keep the text "Subscriptions, membership and contributions
Get related news and offers..." and display the error message below to give context.

jranks123:
@tsop14 I like that suggestion!

PR #1241, opened 2018-11-24
scripts to make setup easier

akash1810:
## Why are you doing this?
This is to make the project slightly easier to setup and run, ideally to empower the our ux+d and product colleagues to make changes. Essentially, scripting [these instructions](https://github.com/guardian/support-frontend/wiki/Install).

Ideally the process is:
- clone repo
- run `setup.sh`
- run `fetch-config.sh`
- run `devrun.sh`

Then, going forward, `devrun.sh` is all that's needed to run the project. 🤞

Inspired by https://github.com/guardian/frontend/blob/master/setup.sh.

img

TODO:
- [x] add a note to ensure `include sites-enabled/*` is in the main nginx config
- [x] ensure using the correct version of node by running `nvm use` first
- [x] `sudo chown -R $(whoami):admin /etc/gu` to change ownership of `/etc/gu` to current user
- [x] add a note to update `/etc/hosts` with an entry for `127.0.0.1 support.thegulocal.com`

### DIFF
--- a/devrun.sh
+++ b/devrun.sh
@@ -1,0 +1,4 @@
+#!/usr/bin/env bash
+
+yarn devrun &
+sbt devrun
--- a/fetch-config.sh
+++ b/fetch-config.sh
@@ -1,0 +1,5 @@
+#!/usr/bin/env bash
+
+aws s3 cp s3://membership-private/DEV/support-frontend.private.conf \
+  /etc/gu/support-frontend.private.conf \
+  --profile membership
--- a/fetch-config.sh
+++ b/fetch-config.sh
@@ -1,5 +1,12 @@
 #!/usr/bin/env bash
 
-aws s3 cp s3://membership-private/DEV/support-frontend.private.conf \
-  /etc/gu/support-frontend.private.conf \
-  --profile membership
+CONFIG_DIR=/etc/gu
+
+if [[ -w ${CONFIG_DIR} ]]; then
+  aws s3 cp s3://membership-private/DEV/support-frontend.private.conf \
+    ${CONFIG_DIR}/support-frontend.private.conf \
+    --profile membership
+else
+  echo "ERROR! Cannot write to ${CONFIG_DIR}. Check it exists and you have write permissions to it.";
+  echo "  'sudo chown -R $(whoami):admin ${CONFIG_DIR}' will make you the owner of ${CONFIG_DIR}";
+fi
--- a/setup.sh
+++ b/setup.sh
@@ -1,0 +1,146 @@
+#!/usr/bin/env bash
+set -e
+
+readonly SYSTEM=$(uname -s)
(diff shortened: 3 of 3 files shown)


akash1810:
cc @katyabeer

setup.sh
```diff
@@ -0,0 +1,146 @@
...
+}
+
+install_sbt() {
+  if ! installed sbt; then
+    brew install sbt@1
+  fi
+}
+
+install_yarn() {
+  npm install -g yarn
```
walaura: I think if `brew` is also a dep `brew install yarn --without-node` is the preferred way to go

walaura: i'm not qualified to answer this sadly. I do remember somebody (Justin?) telling me to use brew instead of npm for a reason I can't quite remember either :(

devrun.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
```
walaura: I'm unsure about this last script, not the part about merging sbt & webpack (that's perfect) but rather wouldn't it be better to 'teach to fish' ux&d and get them familiar with yarn? we can do a `yarn full-devrun` command to run both using yarn and in turn get them familiar with using the `yarn` prefix

setup.sh
```diff
@@ -0,0 +1,146 @@
...
+    for i in "${!EXTRA_STEPS[@]}"; do
+      echo "  $((i+1)). ${EXTRA_STEPS[$i]}"
+    done
+  fi
+}
+
+main () {
+  check_encryption
+  create_aws_config
+  install_homebrew
```
JustinPinner: Might be worth checking that brew installed correctly before using it?

setup.sh
```diff
@@ -0,0 +1,146 @@
...
+install_homebrew() {
+  if mac && ! installed brew; then
+    ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
+  fi
+}
+
+install_jdk() {
+  if ! installed javac; then
+    if linux; then
+      sudo apt-get install -y openjdk-7-jdk
```
JustinPinner: Is there a reason for Linux to stay with JDK 7?

akash1810: nope just a bad copy + paste

setup.sh
akash1810: `set -e` should handle this if calls to `brew` fail IIUC?

fetch-config.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+aws s3 cp s3://membership-private/DEV/support-frontend.private.conf \
```
tsop14: Need to ensure you have permission to write to `/etc/gu`.

akash1810: Done.

PR #1242, opened 2018-11-26
Digital Checkout Payment Frequency, Method and Styling

Ap0c:
## Why are you doing this?

Adding payment frequency and payment method radio buttons to the checkout. Also applying styling to the checkout page as a whole.

[**Trello Card**](https://trello.com/c/04Ln5TWR/1506-select-subscription-frequency)

cc @JustinPinner 

## Changes

- Added `CheckoutCopy` and `CheckoutHeading` components for sections with shared styling.
- Updated `RadioInput` to use a custom radio visual.
- Created a `withArrow` HOC to add the right arrow SVG to buttons and links.
- Tweaked the styling of some of the form components (e.g. added a chevron to the `Select` component).
- Added headings to the checkout form; added payment frequency and payment method sections.
- Added heading and free trial copy.
- Refactored some imports.
- Added reducer/action logic and types to handle payment frequency and payment method.

## Screenshots

| Desktop | Mobile |
|-|-|
| checkout-desktop | checkout-mobile |

### DIFF
--- a/assets/components/checkoutCopy/checkoutCopy.jsx
+++ b/assets/components/checkoutCopy/checkoutCopy.jsx
@@ -1,0 +1,30 @@
+// @flow
+
+// ----- Imports ----- //
+
+import React from 'react';
+
+
+// ----- Types ----- //
+
+type PropTypes = {|
+  strong: string,
+  copy: string,
+|};
+
+
+// ----- Component ----- //
+
+function CheckoutCopy(props: PropTypes) {
+  return (
+    <p className="component-checkout-copy">
+      <strong className="component-checkout-copy__strong">{props.strong}</strong>
+      {props.copy}
+    </p>
+  );
+}
+
+
+// ----- Exports ----- //
+
+export default CheckoutCopy;
--- a/assets/components/checkoutHeading/checkoutHeading.scss
+++ b/assets/components/checkoutHeading/checkoutHeading.scss
@@ -19,4 +19,6 @@
 
 .component-checkout-heading__copy {
   @include gu-fontset-body-sans;
+
+  color: gu-colour(news-garnett-highlight);
 }
--- a/assets/components/checkoutHeading/checkoutHeading.scss
+++ b/assets/components/checkoutHeading/checkoutHeading.scss
@@ -11,9 +11,10 @@
 .component-checkout-heading__heading {
   @include gu-fontset-heading-large;
   margin-bottom: $gu-v-spacing * 3;
+  width: 300px;
 
   @include mq($from: tablet) {
-    width: 580px;
+    width: 500px;
   }
 }
 
--- a/assets/components/checkoutHeading/checkoutHeading.scss
+++ b/assets/components/checkoutHeading/checkoutHeading.scss
@@ -11,10 +11,10 @@
 .component-checkout-heading__heading {
   @include gu-fontset-heading-large;
   margin-bottom: $gu-v-spacing * 3;
-  width: 300px;
+  width: gu-span(4);
 
   @include mq($from: tablet) {
-    width: 500px;
+    width: gu-span(6);
   }
 }
 
--- a/assets/components/checkoutHeading/checkoutHeading.scss
+++ b/assets/components/checkoutHeading/checkoutHeading.scss
@@ -1,0 +1,22 @@
+.component-checkout-heading {
+  color: #fff;
+
+  .component-left-margin-section__content {
+    background-color: gu-colour(garnett-neutral-1);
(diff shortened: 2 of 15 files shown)


assets/components/checkoutCopy/checkoutCopy.jsx
```diff
@@ -0,0 +1,30 @@
+// @flow
+
+// ----- Imports ----- //
+
+import React from 'react';
```
walaura: can we use `ProductPageTextBlock` instead? going from the CSS and props they feel similar enough but I'm not sure without screenshots

assets/components/checkoutHeading/checkoutHeading.scss
```diff
@@ -0,0 +1,25 @@
...
+    background-color: gu-colour(garnett-neutral-1);
+    padding: 0 0 4px ($gu-h-spacing / 2);
+  }
+
+}
+
+.component-checkout-heading__heading {
+  @include gu-fontset-heading-large;
+  margin-bottom: $gu-v-spacing * 3;
+  width: 300px;
```
walaura: I put in a `gu-span()` mixin mirroring dotcom that will calculate these sizes (number of columns) more neatly for us, (this would be 4 and 6 columns in this case)

assets/components/forms/formHOCs/withArrow.scss
```diff
@@ -0,0 +1,9 @@
+.component-with-arrow {
+  .svg-arrow-right-straight {
+    height: 45%;
+    display: inline-block;
+    vertical-align: top;
+    margin-left: 16px;
```
walaura: can these be multiples of `$gu-h-spacing`?

assets/components/forms/formHOCs/withArrow.jsx
```diff
@@ -0,0 +1,28 @@
+// @flow
+
+// ----- Imports ----- //
+
+import React from 'react';
+
+import SvgArrowRightStraight from 'components/svgs/arrowRightStraight';
```
walaura: this might be a great point to start using [imported styles](https://github.com/guardian/support-frontend/pull/1183) in this page and these components, which will makes the HOCs more 'magical' as we don't have to remember to import the CSS

assets/components/forms/formHOCs/withError.scss
```diff
@@ -1,9 +1,13 @@
 .component-error {
   display: block;
-  width: 324px;
+  width: calc(100vw - 36px);
```
walaura: where does `36` come from?

assets/components/forms/formHOCs/withArrow.scss
Ap0c: I'm not sure it's worth it here, this would be `$gu-h-spacing * 0.8`.

assets/components/checkoutCopy/checkoutCopy.jsx
Ap0c: As far as I understand it, that component is for headings and content in the sections of the product page? The `CheckoutCopy` component is used for the chunks of text just above the button in the screenshots above.

assets/components/forms/formHOCs/withArrow.jsx
Ap0c: 👍 will look at doing that in a future PR.

assets/components/forms/formHOCs/withError.scss
Ap0c: `20px` of margin + `16px` of padding.

assets/components/checkoutCopy/checkoutCopy.jsx
walaura: ah, I can see it now! `TextBlock` is only concerned with different types of content under a heading and caps out the max width. It's close but not the same! let's leave this like that (with `gu-span` if possible!

assets/components/forms/formHOCs/withArrow.scss
walaura: why is it `16` though? could it it be a rounder 15?

assets/components/forms/formHOCs/withError.scss
walaura: would using `100%` +`box-sizing: border-box` work for this? I'm a bit wary of replicating so many metrics this deep down

Ap0c: You could apply `box-sizing: border-box` to avoid having to apply the reduction for padding, but I think it would still leave you with the margin to account for.

walaura: what about `auto`? as a block this should take all the available space sans margins already right?

PR #1208, opened 2018-11-19
Make the shrink_no-blurb_no-header variant from the smallMobileHeader test the control (EPIC and BANNER) only

jranks123:
## Why are you doing this?

Following [this test ](https://github.com/guardian/support-frontend/pull/1193), we are going to default to the winning variant, `shrink_no-blurb_no-header`, on mobile for those coming from the epic and banner. 

This PR also sets up a new test, which is the same as the test before, but for everyone that didn't come from the epic or banner. 

| `shrink_no-blurb_no-header` | `shrink_no-blurb` | `shrink` | `control` |
| ------------ | ----------------| -----| -----|
| support thegulocal com_uk_contribute iphone 5_se 4 | support thegulocal com_uk_contribute iphone 5_se 3 | support thegulocal com_uk_contribute iphone 5_se 2 | support thegulocal com_uk_contribute iphone 5_se 1 |

PR #1243, opened 2018-11-26
US campaign copy on /contribute page

tomrf1:
## Why are you doing this?
If a user is referred by a US campaign epic, display a different message on the /contribute page. This means users who come from the control epic will see the old copy.

[**Trello Card**](https://trello.com/c/yTsOtQOQ/216-update-copy-on-us-landing-page)

## Changes

* Adds function `isUsCampaignTest`, which uses the `abTest` provided in the query string to decide which copy to show. This will be used for other design changes later.
* Make the text below the headline optional

## Screenshots
picture 18

### DIFF
--- a/assets/helpers/internationalisation/contributions.js
+++ b/assets/helpers/internationalisation/contributions.js
@@ -1,6 +1,7 @@
 // @flow
 
 import { type CountryGroupId } from 'helpers/internationalisation/countryGroup';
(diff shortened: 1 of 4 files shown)


tsop14:
Are these custom messages only for desktop? #1208 means they will be hidden on mobile?

tomrf1:
@tsop14 good point! That does mean mobile users will not see the copy

PR #1244, opened 2018-11-26
Rolling out countdown timer for Digital Pack & Featured product

katyabeer:
Making the countdown timer look more prominent and rolling it out as default for the featured product header - Digital Pack (US, AU, RoW only).

## Why are you doing this?

[**Trello Card**](https://trello.com/c/R64yrpsc)

## Screenshots
**Before:**

screen shot 2018-11-26 at 16 21 28

**After:**
screen shot 2018-11-26 at 16 21 38

**Before:**
screen shot 2018-11-26 at 16 22 45

**After:**
screen shot 2018-11-26 at 16 21 45

### DIFF
--- a/assets/components/featuredProductHero/featuredProductHero.jsx
+++ b/assets/components/featuredProductHero/featuredProductHero.jsx
@@ -14,7 +14,6 @@ import type { HeadingSize } from 'components/heading/heading';
 // ----- Types ----- //
 
 type PropTypes = {
-  hasTimer?: boolean,
   headingSize: HeadingSize,
   headingText: string,
   subheadingText?: ?string,
@@ -27,7 +26,6 @@ type PropTypes = {
 export default function FeaturedProductHero(props: PropTypes) {
 
   const {
-    hasTimer,
     headingSize,
     headingText,
     subheadingText,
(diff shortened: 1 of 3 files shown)


assets/components/flashSaleCountdown/flashSaleCountdown.scss
```diff
@@ -1,5 +1,5 @@
 .component-flash-sale-countdown {
-  font-size: 14px;
+  font-size: 1em;
```
walaura: we normally put font sizes in px, could this be `16px` instead?

assets/components/featuredProductHero/featuredProductHero.jsx
```diff
@@ -37,7 +35,7 @@ export default function FeaturedProductHero(props: PropTypes) {
     product,
   } = props;
 
-  const timerClassName = classNameWithModifiers('component-featured-product-hero__countdownbox', hasTimer ? [] : ['hidden']);
```
walaura: looking at this now perhaps it would be cleaner to pass `product === 'DigitalPack'` as `hasTimer` higher down the line?

katyabeer: hasTimer isn't in use anywhere. So it doesn't make sense to make a prop (says Akash) :)

walaura: it's not anywhere because the timer wasn't showing until this PR :D the idea behind it was to enable/disable the timer from outside the component, just to make the component a bit more reusable

PR #1250, opened 2018-11-27
Show the countdown timer for any product which is in the flash sale

rupertbates:
## Why are you doing this?

We should show a flash sale countdown timer in the featured product header  for any product which is in the flash sale, not just Digital Pack as we were doing previously


JustinPinner:
Nice

PR #1252, opened 2018-11-28
Update promocode for the Print flashsale

katyabeer:
## Why are you doing this?

AS a Marketing Manager
I WANT to link to a page that shows all the Print and Print + Digital options from the featured product header SO THAT customers can see all the available options in one place.

Specifically the print featured product header should use the promo code 'GCA80Z' so the url would be: https://subscribe.theguardian.com/p/GCA80Z 

[**See the Trello ticket here**](https://trello.com/c/NBmyiHc5)

## Question for @rupertbates :

Do we need to create a unique tracking code too, to know the impact of the change?

### DIFF
--- a/assets/helpers/flashSale.js
+++ b/assets/helpers/flashSale.js
@@ -41,7 +41,7 @@ const saleDetails: SaleDetails = {
     intcmp: '',
   },
   Paper: {
-    promoCode: 'GCA80F',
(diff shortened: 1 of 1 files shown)


katyabeer:
@rupertbates We've spoken to Sarah and decided not to proceed with replacing the promo code. There are 2-3 days left until the sale so it's really not worth it!

PR #1253, opened 2018-11-29
Add recurring paths to TiP

jacobwinch:
## Why are you doing this?

This is a follow up to https://github.com/guardian/support-frontend/pull/1123 and https://github.com/guardian/support-frontend/pull/1160. It allows us to track important recurring contributions flows and updates a [TiP dashboard](https://github.com/guardian/tip#testing-in-production-tip---) accordingly. 

This should make it easier to spot broken permutations of the checkout flow in production (based on real user data).

image

Note that there will be a follow up PR which swaps us over to using a centrally hosted TiP service before this dashboard is properly shareable.

## Changes

* Add recurring paths to TiP config
* Verify relevant paths once successful form submission is confirmed
* Refactor response handling (and minor logging changes) to remove duplication

### DIFF
--- a/app/monitoring/Tip.scala
+++ b/app/monitoring/Tip.scala
@@ -1,7 +1,9 @@
 package monitoring
 
 import app.BuildInfo
+import com.gu.i18n.Country
 import com.gu.support.config.Stages.PROD
+import com.gu.support.workers.model.{DirectDebitPaymentFields, PayPalPaymentFields, PaymentFields, StripePaymentFields}
 import com.gu.tip.{Tip, TipConfig, TipFactory, TipResponse}
 import config.Configuration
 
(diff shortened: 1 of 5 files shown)


app/monitoring/Tip.scala
```diff
@@ -54,6 +62,19 @@ object PathVerification {
     case _ => None
   }
 
+  def monitoredRegion(country: Country): Option[MonitoredRegion] = country match {
+    case Country.Australia => Some(AU)
```
akash1810: What's the difference between `Country` and `MonitoredRegion`?

jacobwinch: `Country` is used to model any country which can be selected as a billing or shipping address (makes more sense in the context of subscriptions), whereas `MonitoredRegion` only applies to higher traffic countries/flows we want to include on this dashboard.

PR #1255, opened 2018-12-03
Aggressive caching for assets (from 1 hour to 1 year)

joelochlann:
I created this branch months & months ago and it was just sitting there. I think I got misled/put off by being unable to test this locally (the assets are not served from `/public/compiled-assets/` but dynamically via webpack) or something silly. (Yes, I could've just run `yarn build-prod`...)

Anyway I just tested this on CODE and it seems like it's working!

The idea is, use the standard long-lived asset caching strategy. That is, cache these assets for 1 year because we generate a new hash of their content each time they change. So if they change, the browser/Fastly will re-fetch them, and if they don't, the browser/Fastly will serve from its cache. Currently it's defaulting to 3600 = 1 hour. So this will speed things up a lot for any repeat visitors who return after an hour.

The `yarn.lock` changes are unrelated - I got them after running `yarn install` due to some new depedencies on master

## before
cache-before

## after
cache-after


joelochlann:
I've just realised there's now no `public` in the cache-control header, which there was previously. So I've added that in to make sure I don't accidentally change any other behaviour with this change

joelochlann:
I've tagged a lot of people as reviewers since I want everyone to be aware of this change in case anyone knows of any adverse effects. Sorry if this is just noise for you!

akash1810:
Great! Would we need to update this header in the VCL too? (Assuming its being explicitly set in VCL).

joelochlann:
@akash1810 - good point, I'll check but I don't think we set any cache headers in the VCL, I think they're all set by origin

joelochlann:
@akash1810 yep nothing in Fastly (only VCL snippets are to do with geolocation and some unused ones around serverside A/B tests - which I should delete while I'm there)

PR #1256, opened 2018-12-03
Digital Subscriptions Heading

Ap0c:
## Why are you doing this?

Adding a heading for "Digital Subscriptions" to the subs landing page (UK only).

[**Trello Card**](https://trello.com/c/f31qOfGL/2090-put-digital-subscriptions-title-back-into-the-uk-subs-landing-page)

## Changes

- Added heading text.

### Before

heading-before

### After

heading-after

### DIFF
--- a/assets/components/subscriptionsByCountryGroup/components/digitalSection.jsx
+++ b/assets/components/subscriptionsByCountryGroup/components/digitalSection.jsx
@@ -30,7 +30,7 @@ type PropTypes = {|
 
 function DigitalSection(props: PropTypes) {
   return (
-    <ThreeSubscriptions heading="">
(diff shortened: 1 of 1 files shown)


assets/components/subscriptionsByCountryGroup/components/digitalSection.jsx
```diff
@@ -30,7 +30,7 @@ type PropTypes = {|
 
 function DigitalSection(props: PropTypes) {
   return (
-    <ThreeSubscriptions heading="">
```
walaura: out of curiosity, what is the reason this was set as an empty string rather than `null`?

PR #1258, opened 2018-12-04
Change email signup to disabled after clicking

johnduffell:
## Why are you doing this?
Sometimes when network issues happened, we were getting multiple duplicate requests for the sign up to the emails button.

This was because when they clicked the button it wasn't actually disabled and the visual appearance didn't change, so during slow network conditions they would click it quite a few times.

This PR changes the button to be disabled when they click, until the resonse comes in.
It also changes the button to look different so people know their click was registered.

[**Trello Card**](https://trello.com/c/IC361x2a/198-investigate-500-errors-should-they-be-500s)

before clicking (no change)
image
new - while loading (used to be the same as before clicking)
image
after complete (no change)
image
after error (no change)
image

@jranks123 for tech review and @ionamckendrick for behaviour review please

### DIFF
--- a/assets/pages/regular-contributions/__tests__/__snapshots__/regularContributionsReducersTest.js.snap
+++ b/assets/pages/regular-contributions/__tests__/__snapshots__/regularContributionsReducersTest.js.snap
@@ -30,6 +30,7 @@ Object {
   "marketingConsent": Object {
     "confirmOptIn": null,
     "error": false,
+    "loading": false,
   },
   "regularContrib": Object {
     "amount": 20,
--- a/assets/pages/regular-contributions/__tests__/__snapshots__/regularContributionsReducersTest.js.snap
+++ b/assets/pages/regular-contributions/__tests__/__snapshots__/regularContributionsReducersTest.js.snap
@@ -30,7 +30,7 @@ Object {
   "marketingConsent": Object {
     "confirmOptIn": null,
     "error": false,
-    "loading": false,
+    "requestPending": false,
   },
   "regularContrib": Object {
     "amount": 20,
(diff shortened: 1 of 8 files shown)


johnduffell:
not sure why the build is failing, as it is not true, so some linter help would be great!
```
/Users/jduffell/ws/support-frontend/assets/components/marketingConsentNew/marketingConsent.jsx
  22:9   error  'csrf' PropType is defined but prop is never used     react/no-unused-prop-types
  23:12  error  'onClick' PropType is defined but prop is never used  react/no-unused-prop-types
  24:10  error  'error' PropType is defined but prop is never used    react/no-unused-prop-types
  25:12  error  'loading' PropType is defined but prop is never used  react/no-unused-prop-types
```

johnduffell:
thanks to @jranks123 this should build, and I've tested it again as working! 🌟 ready for a thumbs up.

jranks123:
I'm not sure about `Loading`. Not sure that language fits. Rather, how about a spinner in the button?

assets/pages/regular-contributions/__tests__/__snapshots__/regularContributionsReducersTest.js.snap
```diff
@@ -30,6 +30,7 @@ Object {
   "marketingConsent": Object {
     "confirmOptIn": null,
     "error": false,
+    "loading": false,
```
jranks123: I feel like loading is quite ambiguous here, maybe a better term would be `requestPending`

johnduffell: will do

johnduffell:
I've updated everything to request pending now, plus the button
image

PR #1259, opened 2018-12-04
warn when something bad happens that doesn't affect conversion

johnduffell:
## Why are you doing this?

Currently we have a lot of alerts for stuff that shouldn't happen, but isn't as severe as all that.
At the moment we can either shout loudly that something is broken, or hide it completely.
It would be better to have a category for alerts that don't need immediate attention but should be planned in to a sprint.

This PR adds a warning count in cloudwatch and makes it alert when guest password or email sing up don't work.
These are chosen because they don't affect conversion, however they reflect failures that would influence business metrics.

I'm keen to add an alarm into the cloudformation for PROD, but I would want us to decide on a suitable threshold and also wording to make it clear this is an alert to be planned into a sprint rather than an something is on fire one.

[**Trello Card**](https://trello.com/c/IC361x2a/198-investigate-500-errors-should-they-be-500s)

image

### DIFF
--- a/app/controllers/IdentityController.scala
+++ b/app/controllers/IdentityController.scala
@@ -15,12 +15,14 @@ import models.identity.responses.SetGuestPasswordResponseCookies
 import codecs.CirceDecoders._
 
 import scala.concurrent.{ExecutionContext, Future}
+import scala.util.Try
 
 class IdentityController(
     identityService: IdentityService,
     components: ControllerComponents,
     actionRefiners: CustomActionBuilders,
-    guardianDomain: GuardianDomain
+    guardianDomain: GuardianDomain,
+    warn: () => Try[Unit]
 )(implicit ec: ExecutionContext)
   extends AbstractController(components) with Circe {
 
@@ -34,7 +36,8 @@ class IdentityController(
         Ok
       } else {
         SafeLogger.error(scrub"Failed to send consents preferences email for ${request.body.email}")
-        InternalServerError
+        warn()
+        NotFound
       }
     }
   }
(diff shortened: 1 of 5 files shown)


app/controllers/IdentityController.scala
```diff
@@ -45,7 +48,8 @@ class IdentityController(
       .fold(
         err => {
           SafeLogger.error(scrub"Failed to set password using guest account registration token ${request.body.guestAccountRegistrationToken}: ${err.toString}")
-          InternalServerError
+          warn()
+          NotFound
```
jacobwinch: I'm not sure that NotFound is an appropriate status code in these cases. I guess you are changing this so that these endpoints no longer trigger the High 5XX rate alarm?

If so, perhaps updating the threshold on that alarm (to accommodate the fact that we expect these calls to fail occasionally) would be better?

johnduffell: Trouble is the background is that given the average rate of contribtuions, we could be 5xx every single attempt to sign up and it would barely trigger the alarm as it is.
I certainly agree that it should be a 500 error from a technical point of view, but after long discussion with @jranks123 I came to the conclusion that 5xx itself isn't granular enough given the wide range of things the app does.  We want to treat 5xx differently depending which endpoint caused them which is not possible without separate ELBs or lambdas.

As an alternative compromise, could we increase the threshold as you suggest, and change the "warning" metric I just added to be "ErrorCount" and have a PreCheckout/PostCheckout dimension.   Then we can alarm more urgently on the former.
That does however mean if the server can't even hit cloudwatch for some reason or doesn't get through that code path, we wouldn't actually alert as quickly.

Any thoughts?

jacobwinch: Fair point, perhaps we can try your existing solution first and see if it helps us 👍

johnduffell:
I have added the alarm to the CFN and tested in CODE (I will remove again from CODE though)
image

PR #1260, opened 2018-12-05
Background image for US campaign

blongden73:
## Why are you doing this?
Adding a background image for only US readers. This change only affects tablet -> wide breakpoints, the image does not display on mobile. 

Desktop
screen shot 2018-12-05 at 11 04 10

LeftCol
screen shot 2018-12-05 at 11 03 58

Wide
screen shot 2018-12-05 at 11 05 37


LATaylor-guardian:
@blongden73 Do we know how this will effect conversion? I'm assuming given it's Q3 of breakeven year we will need to test this has the positive effect on conversion on this page that one assumes?

setup.sh
```diff
@@ -0,0 +1,164 @@
...
+    ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
+  fi
+}
+
+install_jdk() {
+  if ! installed javac; then
+    if linux; then
+      sudo apt-get install -y openjdk-8-jdk
+    elif mac; then
+      EXTRA_STEPS+=("Download the JDK from http://www.oracle.com/technetwork/java/javase/downloads/jdk8-downloads-2133151.html")
```
mchv: We should not encourage people to install Oracle JDK, but `OpenJDK`!
People should use `https://adoptopenjdk.net` for instance but not `oracle` website.

akash1810:
@blongden73 I don't think we can make the diff in this PR smaller to only include the image change as the script additions are on another PR already. Should be a matter of cherry-picking the last commit onto another branch. Happy to help with that if needed.

tsop14:
@LATaylor-guardian this is part of the US end of year campaign being worked on by the contributions team going to message you from our chat channel so can discuss.

assets/pages/new-contributions-landing/components/ContributionBackground.jsx
```diff
@@ -3,15 +3,48 @@
...
+
+type PropTypes = {|
+  countryGroupId: CountryGroupId,
+|};
 
 // ----- Render ----- //
 
-function NewContributionBackground() {
+function NewContributionBackground(props: PropTypes) {
+  if (props.countryGroupId !== 'UnitedStates') {
```
joelochlann: could be slightly clearer to swap these round (`if (props.countryGroupId === 'UnitedStates')` first)

PR #1262, opened 2018-12-05
fix location of nginx conf symlnk

akash1810:
Relative symlinks are bad because they're relative to the destination 😄.


johnduffell:
good idea 👍 coincidentally @jranks123 and I were just looking at that exact line about 8 minutes ago

PR #1266, opened 2018-12-06
us ticker on desktop contributions landing page

tsop14:
## Why are you doing this?
Tickers drive contributions. Adds ticker to desktop only for United States.

[**Trello Card**](https://trello.com/c/EnIUufi3/234-implement-ticker-to-the-us-landing-page)

## Changes

* Adds a stateful component for US ticker

## Screenshots

@ionamckendrick @blongden73 
ticker_chrome

screen shot 2018-12-06 at 11 14 47

Have tested on Safari, FF and Edge and looks OK. 

If the fetch fails or the values are not as expected it will display a white space. It doesn't look great but prevents page jumping from setting hide/show.

screen shot 2018-12-06 at 12 27 19


assets/pages/new-contributions-landing/components/ContributionUsTicker.jsx
```diff
@@ -0,0 +1,111 @@
...
+    };
+
+  }
+
+  componentDidMount(): void {
+    const self = this;
+    getInitialTickerValues().then(({ totalSoFar, goal }) => {
+      const initialTotal = totalSoFar * 0.8;
+      this.count = initialTotal;
+      this.totalSoFar = totalSoFar;
```
joelochlann: I don't think there's any distinction between `this` and `self` in this case since arrow functions don't create a new `this` context. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#No_separate_this

assets/pages/new-contributions-landing/components/ContributionUsTicker.jsx
```diff
@@ -0,0 +1,111 @@
...
+          </div>
+
+          <div className="contributions-landing-ticker__goal">
+            <div className="contributions-landing-ticker__count">${Math.floor(this.state.goal).toLocaleString()}</div>
+            <div className="contributions-landing-ticker__count-label contributions-landing-ticker__label">our goal</div>
+          </div>
+        </div>
+        <div className="contributions-landing-ticker__progress-bar">
+          <div className="contributions-landing-ticker__progress">
+            <div ref={(filledProgressBar) => { this.filledProgressBar = filledProgressBar; }} className="js-contributions-landing-ticker__filled-progress contributions-landing-ticker__filled-progress" />
```
joelochlann: I guess we don't need this `js-` class?

assets/pages/new-contributions-landing/contributionsLanding.scss
```diff
@@ -188,6 +188,78 @@ footer[role="contentinfo"] a {
...
+.contributions-landing-ticker__label {
+  font-style: italic;
+}
+
+.contributions-landing-ticker__progress-bar {
+  position: relative;
+}
+
+.contributions-landing-ticker__progress {
+  width: calc(100%);
```
joelochlann: do we need the `calc()` here?

ionamckendrick:
@tsop14 Please could the copy on the LHS read "contributed" instead of "so far". Thanks! Looks great.

assets/pages/new-contributions-landing/components/ContributionFormContainer.jsx
```diff
@@ -98,6 +99,7 @@ function ContributionFormContainer(props: PropTypes) {
     : (
       <div className="gu-content__content">
         <h1 className={headerClasses}>{countryGroupDetails.headerCopy}</h1>
+        { props.countryGroupId === 'UnitedStates' && <ContributionUsTicker /> }
```
joelochlann: this is quite a nice way of conditionally rendering, but having it right next to a different way of achieving the same thing feels a bit jarring. maybe we should stick to one?

assets/pages/new-contributions-landing/components/ContributionUsTicker.jsx
```diff
@@ -0,0 +1,127 @@
...
+      this.setState({ totalSoFar: this.totalSoFar });
+    } else {
+      this.setState({ totalSoFar: this.count });
+      window.requestAnimationFrame(this.increaseCounter);
+    }
+  }
+
+
+  render() {
+    let wrapperClassName = "contributions-landing-ticker--hidden";
```
Ap0c: Should this be `contributions-landing-ticker contributions-landing-ticker--hidden`?

assets/pages/new-contributions-landing/components/ContributionUsTicker.jsx
```diff
@@ -0,0 +1,127 @@
...
+    if (this.state && this.state.totalSoFar && this.state.totalSoFar != 0) {
```
Ap0c: `!==`?

assets/pages/new-contributions-landing/components/ContributionUsTicker.jsx
```diff
@@ -0,0 +1,127 @@
...
+  totalSoFar: number,
+  goal: number,
+|}
+
+type PropTypes = {}
+
+
+// ---- Helpers ----- //
+
+const getInitialTickerValues = () =>
```
Ap0c: You may be able to add an explicit return type to this:

```jsx
Promise<{| totalSoFar: number, goal: number |}>
```

tsop14: Good idea, added.

assets/pages/new-contributions-landing/components/ContributionUsTicker.jsx
tsop14: yes the linter isn't happy either! 🙈 updated.

assets/pages/new-contributions-landing/components/ContributionFormContainer.jsx
tsop14: updated.

tsop14:
@blongden73 
screen shot 2018-12-06 at 14 21 13

assets/pages/new-contributions-landing/components/ContributionFormContainer.jsx
```diff
@@ -98,6 +99,7 @@ function ContributionFormContainer(props: PropTypes) {
     : (
       <div className="gu-content__content">
         <h1 className={headerClasses}>{countryGroupDetails.headerCopy}</h1>
+        { props.countryGroupId === 'UnitedStates' ? <ContributionUsTicker /> : null }
```
akash1810: minor: could we make `UnitedStates` a const in this repo... somewhere...? Just to avoid magic strings.

assets/pages/new-contributions-landing/components/ContributionUsTicker.jsx
```diff
@@ -0,0 +1,120 @@
...
+  goal: number,
+|}
+
+type PropTypes = {}
+
+
+// ---- Helpers ----- //
+
+const getInitialTickerValues = (): Promise<{| totalSoFar: number, goal: number |}> =>
+  fetch('https://interactive.guim.co.uk/docsdata-test/1ySn7Ol2NQLvvSw_eAnVrPuuRnaGOxUmaUs6svtu_irU.json')
```
akash1810: do we need to handle the failure case?

assets/pages/new-contributions-landing/components/ContributionUsTicker.jsx
```diff
@@ -0,0 +1,120 @@
...
+
+      return { totalSoFar, goal };
+    });
+
+const percentageTotalAsNegative = (total: number, goal: number) => {
+  let percentage = ((total / goal) * 100) - 100;
+  if (percentage > 0) {
+    percentage = 0;
+  }
+  return percentage;
```
akash1810: minor: could use a ternary and a const here to avoid the let

```suggestion
const percentage = ((total / goal) * 100) - 100;
return percentage > 0 ? 0 : percentage;
```

assets/pages/new-contributions-landing/components/ContributionUsTicker.jsx
```diff
@@ -0,0 +1,120 @@
...
+      this.setState({ totalSoFar: this.totalSoFar });
+    } else {
+      this.setState({ totalSoFar: this.count });
+      window.requestAnimationFrame(this.increaseCounter);
+    }
+  }
+
+
+  render() {
+    let wrapperClassName = 'contributions-landing-ticker contributions-landing-ticker--hidden';
```
akash1810: could we use https://github.com/guardian/support-frontend/blob/master/assets/helpers/utilities.js#L21-L27 here to add the `--hidden` modifier?

conf/CODE.public.conf
```diff
@@ -16,7 +16,7 @@ support.url="https://support.code.dev-theguardian.com"
...
+play.filters.headers.contentSecurityPolicy = "default-src 'self' www.paypalobjects.com www.paypal.com www.sandbox.paypal.com js.stripe.com pasteup.guim.co.uk ophan.theguardian.com j.ophan.co.uk media.guim.co.uk www.google-analytics.com www.googletagmanager.com tagmanager.google.com polyfill.io www.googleadservices.com googleads.g.doubleclick.net www.google.com www.google.co.uk optimize.google.com static.ads-twitter.com bat.bing.com bid.g.doubleclick.net t.co analytics.twitter.com stats.g.doubleclick.net www.youtube-nocookie.com connect.facebook.net www.facebook.com consumer.krxd.net cdn.krxd.net beacon.krxd.net secure.adnxs.com checkout.stripe.com fonts.googleapis.com ssl.gstatic.com www.gstatic.com fonts.gstatic.com sentry.io www.dwin1.com data: wss: 'unsafe-inline' q.stripe.com payment.code.dev-guardianapis.com https://interactive.guim.co.uk/"
```
akash1810: I wonder if its worth proxying the interactive json call and remove this entry to the csp?

assets/pages/new-contributions-landing/components/ContributionFormContainer.jsx
joelochlann: yes good point. these are already in `countryGroup.js`, and there's an export `countryGroup` with appropriate types. (https://github.com/guardian/support-frontend/blob/master/assets/helpers/internationalisation/countryGroup.js#L33-L86).

in this case you would use it as follows:

```javascript
countryGroup[props.countryGroupId].name === 'United States'
```

which is a bit roundabout, so maybe it's worth replicating the object keys as a property `id` so you could do

```
props.countryGroupId === countryGroup.UnitedStates.id
```

assets/pages/new-contributions-landing/components/ContributionUsTicker.jsx
tsop14: Would be better to have explicit handling but what happens if no data is returned (fetch fails or not in form expected) is the wrapper never gets `--hidden` removed and so a white space appears (see screenshot in PR description).

assets/pages/new-contributions-landing/components/ContributionUsTicker.jsx
```diff
@@ -0,0 +1,118 @@
...
+  totalSoFar: number,
+  goal: number,
+|}
+
+type PropTypes = {}
+
+
+// ---- Helpers ----- //
+
+const getInitialTickerValues = (): Promise<{| totalSoFar: number, goal: number |}> =>
```
jranks123: I think you should be able to put `Promise<StateTypes>` here

assets/pages/new-contributions-landing/components/ContributionUsTicker.jsx
```diff
@@ -0,0 +1,118 @@
...
+  totalSoFar: number;
+
+  animateBar(totalSoFar: number, goal: number) {
+    const progressBarElement = this.filledProgressBar;
+    if (progressBarElement && progressBarElement instanceof HTMLElement) {
+      progressBarElement.style.transform = `translateX(${percentageTotalAsNegative(totalSoFar, goal)}%)`;
+    }
+  }
+
+  increaseCounter() {
```
jranks123: wow, this is a far out way of defining a function! No `const` or `function`. Is this style/syntax specific to stateful components?

conf/CODE.public.conf
tsop14: I think this would be a good idea. But wont change for this PR.

assets/pages/new-contributions-landing/components/ContributionUsTicker.jsx
jranks123: 👍

PR #1272, opened 2018-12-07
Price Type

Ap0c:
## Why are you doing this?

Adding a type to handle prices, contains the amount and currency. To be used on the digital checkout.

## Changes

- Tidied up the currency helper, and added helper functions to retrieve the glyphs for a given `IsoCurrency`.
- Added a new `Price` type and corresponding helper file, with helpers to create a `Price` for a given currency and one to display the price as a string.
- Updated the digital subscriptions prices to use the new `Price` type.

### DIFF
--- a/assets/helpers/internationalisation/price.js
+++ b/assets/helpers/internationalisation/price.js
@@ -22,9 +22,9 @@ const EUR = (value: number): Price => ({ value, currency: 'EUR' });
 const NZD = (value: number): Price => ({ value, currency: 'NZD' });
 const CAD = (value: number): Price => ({ value, currency: 'CAD' });
 
-function showPrice(p: Price, extended: boolean = false): string {
+function showPrice(p: Price, isExtended: boolean = false): string {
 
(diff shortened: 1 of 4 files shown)


assets/helpers/internationalisation/price.js
```diff
@@ -0,0 +1,43 @@
...
+// ----- Functions ----- //
+
+const GBP = (value: number): Price => ({ value, currency: 'GBP' });
+const USD = (value: number): Price => ({ value, currency: 'USD' });
+const AUD = (value: number): Price => ({ value, currency: 'AUD' });
+const EUR = (value: number): Price => ({ value, currency: 'EUR' });
+const NZD = (value: number): Price => ({ value, currency: 'NZD' });
+const CAD = (value: number): Price => ({ value, currency: 'CAD' });
+
+function showPrice(p: Price, extended: boolean = false): string {
```
akash1810: would `isExtended` give more indication of the type?

PR #1275, opened 2018-12-07
Update referrer test names for US campaign

joelochlann:
These have got out of sync. And we have a banner test too now.


joelochlann:
Supersede by other PRs that removed this check

PR #1277, opened 2018-12-10
Harcode paper prices

walaura:
## Why are you doing this?

Replaces the dataset+zuora fetching with hardcoded ones in our JS. Since we are hardcoding them for checkout this makes more sense

### DIFF
--- a/assets/helpers/subscriptions.js
+++ b/assets/helpers/subscriptions.js
@@ -133,6 +133,17 @@ const digitalSubscriptionPrices = {
   },
 };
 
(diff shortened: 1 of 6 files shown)


assets/helpers/subscriptions.js
```diff
@@ -133,6 +133,17 @@ const digitalSubscriptionPrices = {
   },
 };
 
+const PaperSubscriptionPrices = {
```
Ap0c: `paperSubscriptionPrices`

PR #1278, opened 2018-12-10
Document seemingly weird bits of productPage

walaura:
## Why are you doing this?

Some parts of `ProductPage` are not super intuitive at first. Inline documentation to the rescue!

### DIFF
--- a/assets/components/productPage/productPageButton/productPageButton.jsx
+++ b/assets/components/productPage/productPageButton/productPageButton.jsx
@@ -7,6 +7,16 @@ import SvgArrowRightStraight from 'components/svgs/arrowRightStraight';
 
 import './productPageButton.scss';
 
+/*
+This button will be rendered as an <a> element or a <button> element
+depending on whether it has an `onClick` poaremeters or a `href` parameter.
+This helps preserve the semantics of the underlying HTML.
+
+Sometimes it is helpful to have both an `href` and `onClick` on an <a>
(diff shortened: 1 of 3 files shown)


assets/components/productPage/productPageContentBlock/productPageContentBlock.scss
```diff
@@ -18,11 +18,33 @@
...
+  |+--------------+|
+  || another blue ||
+  || block        ||
+  |+--------------+|
+  +----------------+
+
   */
   & + &,
-  & + div > &  {
+  & + div > &:first-child  {
```
walaura: There was a bug here before where the selector would have added a border to all children of the same color instead of just the first one, this is the correct behaviour.

assets/components/productPage/productPageButton/productPageButton.jsx
```diff
@@ -7,6 +7,16 @@ import SvgArrowRightStraight from 'components/svgs/arrowRightStraight';
 
 import './productPageButton.scss';
 
+/*
+This button will be rendered as an <a> element or a <button> element
+depending on whether it has an `onClick` poaremeters or a `href` parameter.
```
akash1810: typo

walaura:
```suggestion
depending on whether it has an `onClick` parameters or a `href` parameter.
```

PR #1281, opened 2018-12-10
Make primary CTA on thank you page solid

johnduffell:
## Why are you doing this?

The main cta on a page should be solid for usability reasons.  The supplementary ones should generally be hollow.  In the thank you page the Share your thoughts is the main one, although the sign up to email one is also kind of main so can remain solid.

[**Trello Card**](https://trello.com/c/8y47X6If/224-correct-colours-of-the-ctas-on-our-thank-you-pages)

Before:
image
After:
image

@ionamckendrick I think I covered the things you described, please let me know if I forgot any of the tweaks!

### DIFF
--- a/assets/pages/new-contributions-landing/contributionsLanding.scss
+++ b/assets/pages/new-contributions-landing/contributionsLanding.scss
@@ -1056,7 +1056,6 @@ form {
   border-bottom: 1px solid gu-colour(garnett-neutral-4);
   margin: 0 -10px;
   padding: 0 10px;
-  color: #121212;
 
   h3 {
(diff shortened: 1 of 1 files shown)


ionamckendrick:
Looks good!

One small/separate, but connected thing to check: the "Please tell us..." text should be the same as the "Subscriptions, membership and contributions". Please could you check type size / line height / weight and make sure they're consistent? @johnduffell

johnduffell:
thanks @ionamckendrick , I have updated the default colour for the whole page to be garnett-neutral-1 i.e. 121212 rather than 333333.  See the edited screenshot in the description above.

PR #1282, opened 2018-12-10
Add useful comment to RegularContributions.scala

jranks123:
At the moment, if the `createContributorAndUser` method returns an `InternalServerError`, rather than a `StatusResponse`, we don't return the GuestAccountRegistrationToken. This PR adds a comment explaining the consequences of this (we could fix it but it's too small an edge case to justify working on it right now)


johnduffell:
great thanks, now we can do away with the trello card! 👍

PR #1285, opened 2018-12-10
us campaign desktop tests

tsop14:
## Why are you doing this?
Adds an ab test for the new us EOY campaign features. 

## Changes

* Set the ticker and background image as variants in a us-only ab test. These will only appear to desktop users.
* The campaign copy is set as the control - everyone in the US sees the campaign copy on desktop and mobile (unless they are also in a `ab-smallMobileHeaderNotEpicOrBanner` variant which does not display it).
* Removes referral logic given everyone sees the campaign copy. 

## Screenshots

Current (not referred from epic / banner)
screen shot 2018-12-10 at 15 19 26

| campaignCopy  | copyAndTicker | copyAndTickerAndBackgroundImage |
| ------------- | ------------- |------------- |
| screen shot 2018-12-10 at 15 53 46 | screen shot 2018-12-10 at 16 36 33 | screen shot 2018-12-10 at 16 42 11 |
| screen shot 2018-12-10 at 16 13 21 |   | screen shot 2018-12-10 at 15 54 41 |

### DIFF
--- a/assets/helpers/abTests/abtestDefinitions.js
+++ b/assets/helpers/abTests/abtestDefinitions.js
@@ -8,30 +8,17 @@ export type AnnualContributionsTestVariant = 'control' | 'annualAmountsA' | 'not
 
 export const tests: Tests = {
 
-  usTickerLandingPage: {
-    variants: ['ticker'],
+  usDesktopEOYCampaign: {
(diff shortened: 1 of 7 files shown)


ionamckendrick:
Looking really good!!

One thing: could we increase the padding under the ticker please? Perhaps so it is the same as between the header text and the first grey line. Is that ok?

PR #1286, opened 2018-12-10
Apple Pay test

jranks123:
## Why are you doing this?
The inaugural Apple Pay test. 

The Apple Pay button will only show for those with ApplePay enabled on their iPhone/on Safari on OSX. 

Still left to do (future PR): 

- Currently, Apple Pay payments are tracked as Stripe payments. Need to add Apple Pay as a payment provider in the ophan acquisition thrift definition and update acquisition event producer
- The test excludes AU, as AU has a different Stripe key and there is a little bit of work to do to set it up for AU. 
- The integration throws a few warnings, which don't affect the functionality of the integration. But it would be good to talk to Stripe to see how we can mitigate against them. It is my suspicion that these warnings are unavoidable when trying to integrate the integration in a react-redux world, and if you want to put some custom form validation in front of the button (one of the warnings is `Do not call show() yourself if you are using the paymentRequestButton Element`, yet the docs say that if you want to do custom form validation, you have to put `ebent.preventDefault` in the click handler then call `show()` yourself. 

Notes for reviewing - You may find yourself getting confused by the Apple Pay integration being referred to in terms of Stripe everywhere. That is because this integration will soon support Google Pay/the Payment Request API, and the integration is actually Stripes.

Here's the integration in action:
applepay

If anyone wants to run it themselves locally, [here](https://docs.google.com/document/d/1IYKN7dk8hdXy3MpX6ErLbxHoCLyIipotd7MQTy3NV3I/edit?usp=sharing) is a handy guide I've put together (soon to be enshrined in the README)


assets/helpers/contributions.js
```diff
@@ -87,6 +87,11 @@ export type OtherAmounts = {
 export type SelectedAmounts = { [ContributionType]: Amount | 'other' };
 
 
+const getAmount = (selectedAmounts: SelectedAmounts, otherAmounts: OtherAmounts, contributionType: ContributionType) =>
```
jranks123: I've moved this here from `ContributionForm`

assets/helpers/contributions.js
```diff
@@ -51,9 +51,9 @@ export const logInvalidCombination = (contributionType: ContributionType, paymen
 };
 
 export type ThirdPartyPaymentLibraries = {
-  ONE_OFF: { Stripe: Object },
-  MONTHLY: { Stripe: Object, PayPal: Object },
-  ANNUAL: { Stripe: Object, PayPal: Object },
+  ONE_OFF: { Stripe: Object | null },
```
tsop14: Could you use [?Object](https://flow.org/en/docs/types/maybe/) here?

jranks123: I prefer the `<type> | null` pattern, as it's more specific. `?<type>` can be `<type>`, `null` or `undefined`

package.json
```diff
@@ -8,7 +8,7 @@
     "flow": "flow",
     "lint": "eslint --ext .js --ext .jsx .",
     "build-dev": "webpack --config webpack.dev.js",
-    "build-prod": "npm-run-all clean validate test 'webpack --config webpack.prod.js --env.prod'",
```
tsop14: Why have these been removed?

assets/pages/new-contributions-landing/components/StripePaymentRequestButton.jsx
```diff
@@ -0,0 +1,237 @@
...
+  setPaymentRequest: Object => void,
+  onPaymentAuthorised: (PaymentAuthorisation, Function) => void,
+  updateEmail: (string) => void,
+  isTestUser: boolean,
+}) {
+  const paymentRequest = props.stripe.paymentRequest({
+    country: props.country,
+    currency: props.currency,
+    total: {
+      label: 'Demo total',
```
tsop14: Should this be updated?

jranks123: Ah yes, good spot

package.json
jranks123: Thanks for spotting! This was for speeding up development, as you have to run `yarn build-prod` to get Apple Pay to work locally

assets/pages/new-contributions-landing/contributionsLanding.scss
```diff
@@ -1159,6 +1159,40 @@ form {
...
+}
+
+.stripe-payment-request-button__divider {
+  line-height: 32px;
+  padding:0;
+  position:relative;
+  overflow:hidden;
+  text-align:center;
+}
+.stripe-payment-request-button__divider:before,.stripe-payment-request-button__divider:after {
```
jranks123: This CSS is to get the lines either side of `or`

assets/helpers/internationalisation/country.js
```diff
@@ -344,14 +344,53 @@ const countries = {
...
-
 // ----- Types ----- //
 
 export type UsState = $Keys<typeof usStates>;
 export type CaState = $Keys<typeof caStates>;
 export type IsoCountry = $Keys<typeof countries>;
 export type StateProvince = UsState | CaState;
 
+// Annoyingly, this isn't Stripe's documentation, but if you try and submit
+// a country that isn't on this list, you get an error
```
johnduffell: how did you work that out? are we going to have to keep this list up to date?

assets/pages/new-contributions-landing/components/StripePaymentRequestButton.jsx
```diff
@@ -0,0 +1,242 @@
...
+function updateUserEmail(data: Object, setEmail: string => void) {
+  const email = data.payerEmail;
+  if (email && isValidEmail(email)) {
+    setEmail(email);
+  } else {
+    const msg = email
+      ? `Failed to set email for stripe payment request user with email: ${email}`
+      : 'Failed to set email: no email in data object';
+    logException(msg);
+  }
```
johnduffell: we are checking `email` is false twice, why not only check once, something like this
```
if (email) {
  if (isValidEmail(email)) {
    setEmail(email);
  } else {
    logException('Failed to set email for stripe payment request user with email: ${email}');
  }
} else {
  logException('Failed to set email: no email in data object');
}
```

assets/pages/new-contributions-landing/components/StripePaymentRequestButton.jsx
```diff
@@ -0,0 +1,242 @@
...
+}
+
+const onComplete = (complete: Function) => (res: PaymentResult) => {
+  if (res.paymentStatus === 'success') {
+    complete('success');
+  } else if (res.paymentStatus === 'failure') {
+    complete('fail');
+  }
```
johnduffell: are there any other statuses we should handle?

assets/pages/new-contributions-landing/components/StripePaymentRequestButton.jsx
```diff
@@ -0,0 +1,242 @@
...
+  return null;
+}
+
+// ----- Auxiliary components ----- //
+
+const PaymentRequestButton = injectStripe(connect(mapStateToProps, mapDispatchToProps)(paymentRequestButton));
+
+// ----- Default props----- //
+
+export default StripePaymentRequestButton;
```
johnduffell: this file is a bit of a monster, just wonder is there any way of splitting either the file or even the main `function paymentRequestButton` down at all?

assets/pages/new-contributions-landing/contributionsLanding.scss
```diff
@@ -1159,6 +1159,40 @@ form {
...
+/* Stripe Payment Request Button */
+
+.stripe-payment-request-button__button {
+  margin-top: $gu-v-spacing * 2;
+  margin-bottom: $gu-h-spacing;
+}
+
+.stripe-payment-request-button__divider {
+  line-height: 32px;
+  padding:0;
```
johnduffell: some inconsistent spacing after the `:` in this file

assets/pages/new-contributions-landing/components/StripePaymentRequestButton.jsx
```diff
@@ -0,0 +1,242 @@
...
+  country: string,
+  currency: string,
+  amount: number,
+  setCanMakeApplePayPayment: (boolean) => void,
+  setPaymentRequest: Object => void,
+  onPaymentAuthorised: (PaymentAuthorisation, Function) => void,
+  updateEmail: (string) => void,
+  isTestUser: boolean,
+}) {
+  const paymentRequest = props.stripe.paymentRequest({
```
johnduffell: it isn't totally clear what a paymentRequest is, espcially when it gets into the redux state and is just an Object.  I'm not sure if it's possible to call the state StripePaymentRequestSession or something, if that's what it is?

assets/pages/new-contributions-landing/contributionsLandingActions.js
```diff
@@ -398,15 +418,25 @@ const recurringPaymentAuthorisationHandlers = {
...
     },
-    Stripe: (dispatch: Dispatch<Action>, state: State, paymentAuthorisation: PaymentAuthorisation): void => {
-      dispatch(executeStripeOneOffPayment(stripeChargeDataFromAuthorisation(paymentAuthorisation, state)));
+    Stripe: (
+      dispatch: Dispatch<Action>,
+      state: State,
+      paymentAuthorisation: PaymentAuthorisation,
+      onComplete?: Function,
+    ): void => {
+      dispatch(executeStripeOneOffPayment(stripeChargeDataFromAuthorisation(paymentAuthorisation, state), onComplete));
```
johnduffell: I noticed we use dispatch for events mostly, but there are some commands like this one too.  Is it just a case of mix and match or are there any rules on this?

johnduffell:
looks good, I've reviewed it mostly for learning purposes!  I notice we have a few very large files  (hundreds of lines).  Normally I try to separate things unless there's a reason to keep them together, rather than the opposite way round, but I am not sure whether this stuff can or should be separated.
Other than that great work!

assets/pages/new-contributions-landing/components/StripePaymentRequestButton.jsx
jranks123: 👍

assets/pages/new-contributions-landing/components/StripePaymentRequestButton.jsx
jranks123: `success` and `fail` are the only two possible values of `res.paymentStatus` (see the `PaymentResult` type)

assets/pages/new-contributions-landing/components/StripePaymentRequestButton.jsx
jranks123: Done

assets/pages/new-contributions-landing/components/StripePaymentRequestButton.jsx
```diff
@@ -0,0 +1,198 @@
...
+
+function initialisePaymentRequest(props: PropTypes) {
+  const paymentRequest = props.stripe.paymentRequest({
+    country: props.country,
+    currency: props.currency,
+    total: {
+      label: 'The Guardian',
+      amount: props.amount,
+    },
+    requestPayerEmail: true,
```
joelochlann: so do we hide the email field on the form if they're using the Payment Request button?

joelochlann: oh I see, we just say "or". I guess you've probably discussed this at length already but it would be interesting to see how it performs if we just hide all the rest. Maybe as a separate variant/later test

assets/pages/new-contributions-landing/components/StripePaymentRequestButton.jsx
```diff
@@ -0,0 +1,198 @@
...
+  });
+  paymentRequest.on('token', ({ complete, token, ...data }) => {
+    // We need to do this so that we can offer marketing permissions on the thank you page
+    updateUserEmail(data, props.updateEmail);
+    const tokenId = props.isTestUser ? 'tok_visa' : token.id;
+    props.onPaymentAuthorised({ paymentMethod: 'Stripe', token: tokenId }, onComplete(complete));
```
joelochlann: Not necessary for this PR, but it would be good to have a re-think and see if we can return a promise from async actions like `onStripePaymentRequestApiPaymentAuthorised `. Then we would need to pass in this `onComplete` callback.

It certainly seems like it's possible, looking at e.g. https://redux.js.org/advanced/async-actions#indexjs

But not sure off the top of my head how we do something similar (might not require major changes)

assets/pages/new-contributions-landing/contributionsLanding.scss
```diff
@@ -1159,6 +1159,40 @@ form {
...
+  position:relative;
+  overflow:hidden;
+  text-align:center;
+}
+.stripe-payment-request-button__divider:before,.stripe-payment-request-button__divider:after {
+  content:" ";
+  position:absolute;
+  top:50%;
+  margin-left:-999em;
```
joelochlann: what's the idea behind these margins? why won't `width: 100%` do the job?

joelochlann: and any reason for `em`?

assets/pages/new-contributions-landing/components/StripePaymentRequestButton.jsx
```diff
@@ -0,0 +1,198 @@
...
+  if (!props.paymentRequest) {
+    initialisePaymentRequest({ ...props });
+    return null;
+  }
+  if (!props.canMakeApplePayPayment) {
+    return null;
+  }
+
+
+  updateAmount(props.amount, props.paymentRequest);
```
joelochlann: I wonder if it's clearer to have this happen in the `onClick`?

assets/pages/new-contributions-landing/components/StripePaymentRequestButton.jsx
```diff
@@ -0,0 +1,198 @@
...
+
+const onComplete = (complete: Function) => (res: PaymentResult) => {
+  if (res.paymentStatus === 'success') {
+    complete('success');
+  } else if (res.paymentStatus === 'failure') {
+    complete('fail');
+  }
+};
+
+function onClick(event, props: PropTypes) {
```
joelochlann: could you put a comment explaining why we need to intercept the click event ourselves?

assets/pages/new-contributions-landing/components/StripePaymentRequestButton.jsx
```diff
@@ -0,0 +1,198 @@
...
+}
+
+
+// ---- Component ----- //
+function PaymentRequestButton(props: PropTypes) {
+  // If we haven't initialised the payment request, initialise it and return null, as we can't insert the button
+  // until the async canMakePayment() function has been called on the paymentRequest object.
+  if (!props.paymentRequest) {
+    initialisePaymentRequest({ ...props });
+    return null;
```
joelochlann: It's quite confusing how there's one `return null;` in the container element, and then some more checks here. Is it possible to put them all in one place?

joelochlann: Or if not, a comment explaining why

assets/pages/new-contributions-landing/components/StripePaymentRequestButton.jsx
jranks123: 👍

assets/pages/new-contributions-landing/components/StripePaymentRequestButton.jsx
jranks123: Before we try and load the Stripe button, we need to check whether the user is in the right country/Stripe has loaded (you can't insert the `<StripeProvider>` tag until it's loaded)/they are in the test. These checks happen at the container level.

Then, once we've decided to load it, we need to check whether a) the `paymentRequest` object has been initialised, and b) whether the browser check has come back true.

So I think the way it currently is laid out makes sense. Basically, loading the `<StripeProvider>` is dependent on some state, and then you can't initialise the `paymentRequest` object until you've injected the `stripe` object into the button (see line 211 in StripePaymentRequestButton), which you get from `StripeProvider`.

assets/pages/new-contributions-landing/components/StripePaymentRequestButton.jsx
jranks123: Yes, great suggestion!

joelochlann:
As discussed, let's just spell out the relationship between Stripe libs:

- `checkout.js`: https://github.com/guardian/support-frontend/blob/jr-apple-pay/assets/helpers/paymentIntegrations/newPaymentFlow/stripeCheckout.js#L27-L43
- `stripe.js`: https://github.com/guardian/support-frontend/blob/jr-apple-pay/app/views/main.scala.html#L102
- and the extra Stripe elements dependency in our webpack bundle

assets/pages/new-contributions-landing/contributionsLandingReducer.js
```diff
@@ -203,6 +216,34 @@ function createFormReducer(countryGroupId: CountryGroupId) {
...
+            canMakeApplePayPayment: action.canMakeApplePayPayment,
+          },
+        };
+
+      case 'SET_PAYMENT_REQUEST':
+        return {
+          ...state,
+          stripePaymentRequestButtonData: {
+            ...state.stripePaymentRequestButtonData,
+            paymentRequest:
```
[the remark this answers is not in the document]
jranks123: 👍

joelochlann:
A related point. If this test is successful (which is highly likely), I think using Stripe elements instead of checkout becomes much more important. Because we're now already including all the libs necessary to use Stripe elements, but we still have the dependency on `checkout.js`. If we could ditch the latter, we'd speed the page up (even though the latter is not render-blocking)

They're similar in size so we'd roughly halve the amount of JS we have to load & parse from Stripe:
```
total 216
drwxr-xr-x  4 joseph_smith wheel    136 Dec 12 12:01 ./
drwxrwxrwt 40 root         wheel   1360 Dec 12 12:01 ../
-rw-r--r--  1 joseph_smith wheel 102062 Dec  7 01:19 checkout.js
-rw-r--r--  1 joseph_smith wheel 114981 Dec 11 21:23 stripe.js
joseph_smith@33937:/tmp/stripe$
```

PR #1288, opened 2018-12-11
update images

akash1810:
[**Trello Card**](https://trello.com/c/AOTkdOHM/1962-daily-edition-image-updates)

### Before
https://support.theguardian.com/uk
image

https://support.theguardian.com/uk/subscribe/digital
image

### After
https://support.thegulocal.com/uk
image

https://support.thegulocal.com/uk/subscribe/digital
image


akash1810:
@walaura should have added a comment about this. There isn't a 4000px wide for this crop. Its 7856px at 5MB! I don't think we should be serving images that big. If quality is a concern, we should invest in migrating to the resize service dotcom uses.

PR #1290, opened 2018-12-11
add extra Ts and Cs for the landing pages

johnduffell:
## Why are you doing this?

New T&C are required in california to clarify cancellation and frequency, so we might as well add them everywhere as it makes things clearer.

[**Trello Card**](https://trello.com/c/Tbv8uop2/226-update-in-line-terms-conditions)

[Main checkout flow](https://support.theguardian.com/uk/contribute)
BEFORE:
image
AFTER:
image

---
[old checkout flow](https://support.theguardian.com/contribute/recurring-guest?contributionValue=5&contribType=MONTHLY&currency=GBP)
BEFORE:
image
AFTER:
image

@ionamckendrick does that look about right?  Those are the only two pages I could find on support-frontend, but there is subscribe-frontend where there are other T&C for the print&digital products

### DIFF
--- a/assets/components/legal/legalSection/legalSectionContainer.js
+++ b/assets/components/legal/legalSection/legalSectionContainer.js
@@ -4,21 +4,14 @@
 
 import { connect } from 'react-redux';
 
-import type { CommonState } from 'helpers/page/commonReducer';
-
 import LegalSection from './legalSection';
-
 
 // ----- State Maps ----- //
 
-function mapStateToProps(state: { common: CommonState }) {
-
-  return {
-    countryGroupId: state.common.internationalisation.countryGroupId,
-  };
-
-}
-
+const mapStateToProps = state => ({
+  countryGroupId: state.common.internationalisation.countryGroupId,
+  contributionType: state.page.form.contributionType,
+});
 
 // ----- Exports ----- //
 
--- a/assets/components/legal/legalSection/legalSectionContainer.js
+++ b/assets/components/legal/legalSection/legalSectionContainer.js
@@ -10,7 +10,7 @@ import LegalSection from './legalSection';
 
 const mapStateToProps = state => ({
   countryGroupId: state.common.internationalisation.countryGroupId,
-  contributionType: state.page.form.contributionType,
+  contributionType: state.page.form ? state.page.form.contributionType : '',
 });
 
 // ----- Exports ----- //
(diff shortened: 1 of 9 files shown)


johnduffell:
with this change, this will also show on one off as @jranks123 pointed out - so I will hide it if you are on one off as it might confuse people who are making a one off.

assets/components/legal/legalSection/legalSectionContainer.js
```diff
@@ -4,21 +4,14 @@
...
 // ----- State Maps ----- //
 
-function mapStateToProps(state: { common: CommonState }) {
-
-  return {
-    countryGroupId: state.common.internationalisation.countryGroupId,
-  };
-
-}
-
```
johnduffell: I'm not sure why the varying format for the mapStateToProps function in different places.  I changed it to this other format bu tI'm not sure whether it loses type safety doing it this way?

assets/components/legal/termsPrivacy/termsPrivacy.jsx
```diff
@@ -22,6 +25,13 @@ function TermsPrivacy(props: PropTypes) {
 
   return (
     <div className="component-terms-privacy">
+      {props.contributionType !== 'ONE_OFF' ?
```
johnduffell: I wasn't sure if theres a constant or I just use the string literal

ionamckendrick:
> maybe need padding! I think @ionamckendrick will have an opinion when she gets to this

@johnduffell Please could we add a little padding between the two, as Tom suggested. Although not loads, as we don't want to push the page down too much on mobile. Would you mind showing me when you tweak it? 

Longer term, we need to work out a more elegant solution – way too much text on this page. However, legal needs must trump for now ...

johnduffell:
I've added and made new screenshots, please take a look @ionamckendrick 

I also spotted a bug I caused in old payment flow which stopped it rendering, so that was lucky I came back to it again!

assets/components/legal/legalSection/legalSectionContainer.js
```diff
@@ -4,21 +4,14 @@
...
+const mapStateToProps = state => ({
+  countryGroupId: state.common.internationalisation.countryGroupId,
+  contributionType: state.page.form ? state.page.form.contributionType : '',
```
johnduffell: the old payment flow doesn't seem to have `state.page.form` so I had to put this check in to let it render, this means the updated T&C will always show.

johnduffell:
@ionamckendrick I have merged to unblock @jranks123 as it's better than nothing from a compliance point of view, but visually of course there's nothing at all stopping us tweaking the padding in another PR, just leave a comment on here or something and we can tweak it no trouble.  probably on monday if it's me doing it.

PR #1293, opened 2018-12-11
remove unnecessary imports

akash1810:
this also fixes the logo's padding in the header of the showcase page

## Screenshots
Before
image

After
image

### DIFF
--- a/assets/pages/showcase/showcase.scss
+++ b/assets/pages/showcase/showcase.scss
@@ -5,16 +5,9 @@
 // ----- Shared Components ----- //
 @import '~components/footer/footer';
 @import '~components/headers/simpleHeader/simpleHeader';
-
 @import '~components/svgs/svg';
-
 @import '~components/headingBlock/headingBlock';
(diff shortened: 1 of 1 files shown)


assets/pages/showcase/showcase.scss
```diff
@@ -5,16 +5,9 @@
 // ----- Shared Components ----- //
 @import '~components/footer/footer';
 @import '~components/headers/simpleHeader/simpleHeader';
-
 @import '~components/svgs/svg';
-
 @import '~components/headingBlock/headingBlock';
 @import '~components/leftMarginSection/leftMarginSection';
```
walaura: Not for this PR but I wonder if we should import this from `productPageContentBlock.js`, as it is a dependency? I guess ideally we'd want it to be in `leftMarginSection.js` itself though…

akash1810: Yeah, the components themselves should def import their own styles rather than the page that composes them together.

PR #1301, opened 2018-12-11
Prefill name fields in the digital pack checkouts

lmath:
## Why are you doing this?
We should make it as easy as possible for users to sign up for a digital pack. So if we have their name, let's fill it in (and let them edit if needed)

[**Trello Card**](https://trello.com/c/sJDnpz0S/2093-pre-populate-checkout-fields)

## Changes

* Adds a digitalSubscription template that saves firstName, lastName, and email into the window object. 
* Adds email to the CheckoutState 
* Removes User Reducer from the digitalSubscriptionCheckoutReducer (we needed it for the email)
* Adds a few tests for the displayForm function in the Subscriptions controller
aaaaaaand
* Sets the initial state for firstName and lastName to the values we fetched and put in the window object (thus pre-filling them in the form)

## Screenshots
It's just the digital pack checkout as before, but the name fields have values 

But, what the heck...
image


assets/pages/digital-subscription-checkout/helpers/user.js
```diff
@@ -0,0 +1,28 @@
...
+  email: Option<string>,
+|};
+
+function getUser(): User {
+
+  if (window && window.guardian && window.guardian.user) {
+    return {
+      firstName: window.guardian.user.firstName || null,
+      lastName: window.guardian.user.lastName || null,
+      email: window.guardian.user.email || null,
```
Ap0c: You may need to do a `typeof a === 'string'` check here to prevent other types making it through.

akash1810:
Looks great!

Do we want some integration tests against the end to end flow? We're currently testing the correct js asset is served, but not that it's doing the right thing.

assets/pages/digital-subscription-checkout/helpers/user.js
lmath: Good idea. I've tried to add this check.

lmath:
That's a good idea @akash1810. I wanted to make sure that there was some test for returning a 500 if the call to identity fails (as we do in the contributions checkouts) since that's new as of this PR but I think an integration test could be in a separate PR.

app/controllers/Subscriptions.scala
```diff
@@ -147,20 +153,32 @@ class Subscriptions(
...
+    val title = "Support the Guardian | Digital Subscription"
+    val id = "digital-subscription-checkout-page-" + countryCode
+    val js = "digitalSubscriptionCheckoutPage.js"
+    val css = "digitalSubscriptionCheckoutPageStyles.css"
+    val csrf = CSRF.getToken.value
+
+    digitalSubscription(title, id, js, css, Some(csrf), idUser)
+  }
+
+  def displayForm(countryCode: String, displayCheckout: String, isCsrf: Boolean = false): Action[AnyContent] = {
```
akash1810: can we make `displayCheckout` a `Boolean`?

lmath: I asked @Ap0c about why it's a String and it's because there's no need to make it possible for a type check to fail when we just need the param to hide the page 
(a la
image )

But apparently we've had this conversation on the team before so I'm going to change it to a Boolean anyway 😅

PR #1304, opened 2018-12-12
Turn on Apple Pay test for all users

jranks123:
## Why are you doing this?
Have checked and everything seems to be working, so I'm going to turn on the test for everyone


tsop14:
👍

PR #1310, opened 2018-12-13
add `/print` routes

akash1810:
## Why are you doing this?
Mainly because I find it difficult to remember its `/paper`, maybe because, at least internally, we call things "print, print+".

### DIFF
--- a/conf/routes
+++ b/conf/routes
@@ -81,11 +81,15 @@ GET  /$country<(uk|us|au|int)>/subscribe/premium-tier   controllers.Subscription
 GET  /subscribe/weekly                            controllers.Subscriptions.weeklyGeoRedirect()
 GET  /$country<(uk|us|au|int|nz|ca|eu)>/subscribe/weekly   controllers.Subscriptions.weekly(country: String)
 
-GET  /subscribe/paper          controllers.Subscriptions.paperMethodRedirect(withDelivery: Boolean = false)
-GET  /subscribe/paper/delivery          controllers.Subscriptions.paperMethodRedirect(withDelivery: Boolean = true)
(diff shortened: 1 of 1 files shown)


conf/routes
```diff
@@ -81,11 +81,15 @@ GET  /$country<(uk|us|au|int)>/subscribe/premium-tier   controllers.Subscription
...
-GET  /uk/subscribe/paper/delivery          controllers.Subscriptions.paper(withDelivery: Boolean = true)
-
+GET  /subscribe/paper                             controllers.Subscriptions.paperMethodRedirect(withDelivery: Boolean = false)
+GET  /subscribe/paper/delivery                    controllers.Subscriptions.paperMethodRedirect(withDelivery: Boolean = true)
+GET  /uk/subscribe/paper                          controllers.Subscriptions.paper(withDelivery: Boolean = false)
+GET  /uk/subscribe/paper/delivery                 controllers.Subscriptions.paper(withDelivery: Boolean = true)
+
+GET  /subscribe/print                             controllers.Subscriptions.paperMethodRedirect(withDelivery: Boolean = false)
+GET  /subscribe/print/delivery                    controllers.Subscriptions.paperMethodRedirect(withDelivery: Boolean = true)
+GET  /uk/subscribe/print                          controllers.Subscriptions.paper(withDelivery: Boolean = false)
```
walaura: we should make one of these routes be redirects so as to not have dupe pages SEO-wise

akash1810: Yeah, I opted against that as it causes another handshake for the client. Though I suspect the `print` endpoints will be used exclusively internally (if not just by me) so it probably doesn't matter.

akash1810:
Closing. This is more an issue with my memory and is solved with bookmarks. Less code = better code!

PR #1319, opened 2018-12-14
Update header and T&C links on weekly page

lmath:
## Why are you doing this?
Keeping things up to date and tidy. 

## Changes

* Update header image
* Remove headline from the product page hero
* GridImage -> Grid Picture
* Update the T&C link copy

## Screenshots
(compare to current state of the page here: https://support.theguardian.com/uk/subscribe/weekly)

| mobile        | tablet           | big wide web screen  |
| ------------- |-------------| -----|
|     |  |  |

### DIFF
--- a/assets/pages/weekly-subscription-landing/weeklySubscriptionLanding.jsx
+++ b/assets/pages/weekly-subscription-landing/weeklySubscriptionLanding.jsx
@@ -152,11 +152,11 @@ const content = (
       </ProductPageContentBlock>
       <ProductPageContentBlock>
         <ProductPageTextBlock title="Promotion terms and conditions">
-          <p>Offer subject to availability. Guardian News and Media Limited (&ldquo;GNM&rdquo;) reserves the right to withdraw this promotion at any time. For full 6 for 6 promotion terms and conditions see <a target="_blank" rel="noopener noreferrer" href={`https://subscribe.theguardian.com/p/WWM99X/terms?country=${subsCountry}`}>here</a>.
+          <p>Offer subject to availability. Guardian News and Media Limited (&ldquo;GNM&rdquo;) reserves the right to withdraw this promotion at any time. For full 6 for 6 promotion terms and conditions, see <a target="_blank" rel="noopener noreferrer" href={`https://subscribe.theguardian.com/p/WWM99X/terms?country=${subsCountry}`}>here</a>.
           </p>
         </ProductPageTextBlock>
         <ProductPageTextBlock title="Guardian Weekly terms and conditions">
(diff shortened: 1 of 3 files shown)


lmath:
The css in `weeklySubscriptionLanding.scss` is mostly copied from `showcase.scss` as the product page hero needs to look really similar.

It makes sense to me that there isn't a gridPicture.scss because grid pictures are used all over the place, but if there is going to be crossover for these header images sometimes, is there somewhere a common style for them should live in this case? (thanks to @akash1810 for suggesting I think about that, but alas, I'm still not sure how to go about it)

walaura:
if we are committed to this type of layout it does make sense! we could wrap `ProductPageHero` to make a `ProductPageHeroWithCovers` that applies this type of treatment. My only worry is the cover images might all have slightly different positioning requirements to the point we won't be able to do much other than removing the left border?

assets/pages/weekly-subscription-landing/weeklySubscriptionLanding.jsx
```diff
@@ -73,18 +73,35 @@ const content = (
       footer={<Footer />}
     >
       <ProductPagehero
-        headline="Become a Guardian Weekly subscriber"
```
walaura: this was the only instance of `headline` being used, worth checking with the designers (@blongden73 commented yesterday that he wanted to tweak its colour too) if this element will reappear again in any headers or we can remove it entirely

assets/pages/weekly-subscription-landing/weeklySubscriptionLanding.jsx
```diff
@@ -135,11 +152,11 @@ const content = (
       </ProductPageContentBlock>
       <ProductPageContentBlock>
         <ProductPageTextBlock title="Promotion terms and conditions">
-          <p>Offer subject to availability. Guardian News and Media Limited (&ldquo;GNM&rdquo;) reserves the right to withdraw this promotion at any time. For full promotion terms and conditions visit <a target="_blank" rel="noopener noreferrer" href={`https://subscribe.theguardian.com/p/WWM99X/terms?country=${subsCountry}`}>subscribe.theguardian.com/p/WWM99X/terms</a>
+          <p>Offer subject to availability. Guardian News and Media Limited (&ldquo;GNM&rdquo;) reserves the right to withdraw this promotion at any time. For full 6 for 6 promotion terms and conditions, see <a target="_blank" rel="noopener noreferrer" href={`https://subscribe.theguardian.com/p/WWM99X/terms?country=${subsCountry}`}>here</a>.
```
walaura: yay! non-url links 😄

PR #1322, opened 2018-12-17
Throw in a .paparazzirc

walaura:
## Why are you doing this?

This lets us use <a href="https://github.com/guardian/paparazzi">paparazzi</a> (`npx @guardian/paparazzi`) to take a burst of screenshots of all routes we desire.

it will also aid us in the development of the tool itself, which right now will outright murder laptops and is probably missing a couple options

### DIFF
--- a/.paparazzirc
+++ b/.paparazzirc
@@ -1,0 +1,32 @@
+{
+	"routes": [
+		"uk",
+		"showcase",
+		"uk/subscribe",
+		"uk/subscribe/paper",
+		"uk/subscribe/digital",
+		"uk/subscribe/weekly",
+		"uk/contribute",
+		"uk/contribute/recurring",
+		"paypal/return"
+	],
+	"out": ".test-screenshots",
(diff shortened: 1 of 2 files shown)


.paparazzirc
```diff
@@ -0,0 +1,32 @@
+{
+	"routes": [
+		"uk",
+		"showcase",
+		"uk/subscribe",
+		"uk/subscribe/paper",
+		"uk/subscribe/digital",
+		"uk/subscribe/weekly",
+		"uk/contribute",
+		"uk/contribute/recurring",
```
joelochlann: I would probably ditch this one since we'll be removing the old payment flow soon - I noticed you've released a new bundle page (maybe you're A/B testing it right now?) which means these routes will soon be unreachable through any flow

walaura: I added it especially because of the old payment flow still being partially there! it's one of the few I'm afraid of accidentally fussing up this week as the header work progresses

walaura:
D: my bad!

PR #1321, opened 2018-12-17
Prepare for DigitalPack flash sale

JustinPinner:
## Why are you doing this?
We're going to run a sale. This PR sets us up to do that.

[**Trello Card**](https://trello.com/c/Hy1lZa32/2095-digital-pack-flash-sale)

## Changes

* Encapsulate all flash sale-related copy, prices etc in a single place
* Components requiring flash sale data call into a common point
* Formulate a data model that allows multiple sales in multiple regions for multiple products to run concurrently, with the possibility of fine-tuning specific copy

## Checklist
- [x] Sale isn't visible before or after the expected dates
- [x] Sale isn't visible outside of expected region (UK)
- [x] Front page displays the correct bundle text
- [x] Subs landing page displays correct featured product and text
- [x] Subs landing page display countdown timer
- [x] Product page displays correct headings
- [x] Product page displays countdown timer
- [x] The correct promo code is sent to the checkout
- [x] The checkout displays the expected product and price
 
## Screenshots
**Sale off (or non-UK region)** - **Landing Page**
screencapture-support-thegulocal-uk-subscribe-flash-sale-inactive-2018-12-17-12_13_23

**Sale off (or non-UK region)** - **Product Page**
screencapture-support-thegulocal-uk-subscribe-digital-flash-sale-inactive-2018-12-17-12_17_38

**Sale on (UK-only)** - **Landing Page**
screencapture-support-thegulocal-uk-subscribe-flash-sale-active-2018-12-17-12_14_44

**Sale on (UK-only)** - **Product Page**
screencapture-support-thegulocal-uk-subscribe-digital-flash-sale-active-2018-12-17-12_16_48

PR #1323, opened 2018-12-17
Prepare for DigitalPack flash sale (all territories)

JustinPinner:
## Why are you doing this?
We're going to run a sale. This PR sets us up to do that, and replaces https://github.com/guardian/support-frontend/pull/1321 because we've now made it run in multiple territories. 

[**Trello Card**](https://trello.com/c/Hy1lZa32/2095-digital-pack-flash-sale)

## Changes

* Encapsulate all flash sale-related copy, prices etc in a single place
* Components requiring flash sale data call into a common point
* Formulate a data model that allows multiple sales in multiple regions for multiple products to run concurrently, with the possibility of fine-tuning specific copy

## Checklist
- [x] Sale isn't visible before or after the expected dates
- [x] Sale isn't visible outside of expected region (UK)
- [x] Front page displays the correct bundle text
- [x] Subs landing page displays correct featured product and text
- [x] Subs landing page display countdown timer
- [x] Product page displays correct headings
- [x] Product page displays countdown timer
- [x] The correct promo code is sent to the checkout
- [x] The checkout displays the expected product and price
- [x] All of the above tested with UK, AU, US & INT regions

## Screenshots
[Screenshots-DP-FlashSale-Global-Dec-2018.zip](https://github.com/guardian/support-frontend/files/2687199/Screenshots-DP-FlashSale-Global-Dec-2018.zip)


assets/helpers/flashSale.js
```diff
@@ -1,84 +1,225 @@
...
+    },
+  };
+  if (flashSaleIsActive(product, countryGroupId)) {
+    const sale = getSales(product, countryGroupId)[0];
+    return sale.saleDetails[countryGroupId].saleCopy;
+  }
+  return emptyCopy;
+}
+
+function getFormattedFlashSalePrice(product: SubscriptionProduct, countryGroupId: CountryGroupId): string {
```
rupertbates: Is `discountedDisplayPrice()` in subscriptions.js unused now this function exists? If so you could remove it.

JustinPinner: It's unused from the POV of flash sales. I'll check to see if that was the only invocation. Thanks.

JustinPinner: Turns out there were a few related things that could also go. So they're gone.
Re-tested (locally) - all ok.

PR #1325, opened 2018-12-18
Use a local settings file if it exists

rupertbates:
## Why are you doing this?

Currently to use local settings with support-frontend you need to uncomment a line in DEV.public.conf. This is not ideal because it means that you then have an uncommitted change before you do anything else and you need to remember to revert it before committing.

Also I prefer to use local settings all the time as it means you can work on large parts of the site (including all of the subs product pages) without Janus credentials (the only reason you need them is to load settings from S3 which are not needed in DEV)

This PR changes the settings loading behaviour so that the app will load from local settings by default in DEV but will fall back to S3 if the local settings file doesn't exist.

### DIFF
--- a/app/admin/Settings.scala
+++ b/app/admin/Settings.scala
@@ -1,14 +1,16 @@
 package admin
 
-import scala.io.{BufferedSource, Source}
-import com.typesafe.config.Config
+import java.io.FileNotFoundException
+import java.nio.file.{Files, Paths}
 
-import play.api.mvc.RequestHeader
-import codecs.CirceDecoders._
-import io.circe.parser._
 import cats.implicits._
+import codecs.CirceDecoders._
 import com.amazonaws.services.s3.AmazonS3
+import com.typesafe.config.Config
+import com.typesafe.scalalogging.LazyLogging
+import io.circe.parser._
 
+import scala.io.{BufferedSource, Source}
 import scala.util.Try
 
 case class Switches(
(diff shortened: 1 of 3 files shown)


app/admin/Settings.scala
```diff
@@ -40,17 +42,15 @@ object Settings {
   def fromLocalFile(source: SettingsSource.LocalFile): Either[Throwable, Settings] =
```
johnduffell: what's the difference between Settings and SettingsSource, as both seem to be able to get from a file?

app/admin/SettingsProvider.scala
```diff
@@ -33,8 +32,9 @@ class LocalFileSettingsProvider private (initialSettings: Settings) extends Sett
 
 object LocalFileSettingsProvider {
 
-  def fromLocalFile(localFile: SettingsSource.LocalFile): Either[Throwable, SettingsProvider] =
+  def fromLocalFile(localFile: SettingsSource.LocalFile): Either[Throwable, SettingsProvider] = {
     Settings.fromLocalFile(localFile).map(new LocalFileSettingsProvider(_))
+  }
```
johnduffell: did you mean to add the braces?

johnduffell:
that's a great idea! although I don't totally follow the code, 👍 for the idea for sure

app/admin/Settings.scala
rupertbates: From looking at it `SettingsSource` provides the location and `Settings` does the actual read. I didn't write the original code so not 100% sure on why it was divided up this way

app/admin/SettingsProvider.scala
rupertbates: Nope, I've removed them

PR #1326, opened 2018-12-18
Upgrade support-models

rupertbates:
## Why are you doing this?
As part of the work to hook up the new Digital Pack checkout to support workers we need to upgrade the support-models dependency to the latest version.

There are quite a few changes for a library upgrade due to some refactoring in support-models to share more code between support-workers and support-frontend, particularly the Circe codecs.

[**Trello Card**](https://trello.com/c/5N8LQ5yU/2060-be-able-to-trigger-the-creation-of-a-digital-pack-subscription-via-support-frontend)


test/codecs/CirceDecodersTest.scala
```diff
@@ -21,76 +21,6 @@ import io.circe.generic.auto._
 
 class CirceDecodersTest extends WordSpec with MustMatchers {
 
-  "referrerAcquisitionDataCodec" should {
```
johnduffell: were these tests supposed to be added in the other pr https://github.com/guardian/support-libraries/pull/6/files ?

johnduffell:
looks good I was just wondering whether the tests were intentially deleted or supposed to be retained in the support-libraries PR
other than that, great! 👍

test/codecs/CirceDecodersTest.scala
rupertbates: They're not needed any more, there are already tests covering this serialisation in support-models

PR #1327, opened 2018-12-18
Track Stripe Apple Pay Payments 

jranks123:
## Why are you doing this?
Previously, we were tracking Apple Pay payments as Stripe payments. This PR adds an extra bit of metadata to `StripeChargeData`,  `stripePaymentMethod`, so that the `payment-api` can track it separately. 

Soon we will be adding Stripe Payment Request Button payments, so this PR also provisions for that future scenario. 

[Corresponding payment api PR](https://github.com/guardian/payment-api/compare/jr-add-apple-pay?expand=1)

### DIFF
--- a/assets/pages/new-contributions-landing/components/StripePaymentRequestButton.jsx
+++ b/assets/pages/new-contributions-landing/components/StripePaymentRequestButton.jsx
@@ -160,7 +160,7 @@ function initialisePaymentRequest(props: PropTypes) {
     // We need to do this so that we can offer marketing permissions on the thank you page
     updateUserEmail(data, props.updateEmail);
     const tokenId = props.isTestUser ? 'tok_visa' : token.id;
-    props.onPaymentAuthorised({ paymentMethod: 'Stripe', token: tokenId })
+    props.onPaymentAuthorised({ paymentMethod: 'Stripe', token: tokenId, stripePaymentMethod: 'StripeApplePay' })
       .then(onComplete(complete));
   });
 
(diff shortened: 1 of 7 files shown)


assets/pages/new-contributions-landing/components/StripePaymentRequestButton.jsx
```diff
@@ -160,7 +160,7 @@ function initialisePaymentRequest(props: PropTypes) {
     // We need to do this so that we can offer marketing permissions on the thank you page
     updateUserEmail(data, props.updateEmail);
     const tokenId = props.isTestUser ? 'tok_visa' : token.id;
-    props.onPaymentAuthorised({ paymentMethod: 'Stripe', token: tokenId })
+    props.onPaymentAuthorised({ paymentMethod: 'Stripe', token: tokenId, stripePaymentMethod: 'StripeApplePay' })
```
johnduffell: do we need to double up on the "stripe" part, the "stripe payment method" is just apple pay.  Also stripe is the payment method, we're overloading the word payment method.  We could have to change to payment processor and payment experience or something.

jranks123: It will end up as `StripeApplePay` in the `PaymentProvider` field in the datalake. It is overloading the field, but we decided that there wasn't enough to gain by splitting out the two parts of the payment

tomrf1: Is the component in this file specifically for Apple pay?

jranks123: Is it currently. When we expand it to the Stripe Payment Request Button, we will have to update this.

PR #1329, opened 2018-12-20
prevent alarming on a 1 minute glitch in the state machine

johnduffell:
## Why are you doing this?

The alarm went off on a one minute outage.
image

We expect it to need 2 before it actually alarms, but it would be missing data the rest of the time.  It last went off in september
image

[According to AWS](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/AlarmThatSendsEmail.html#alarms-and-missing-data) it looks backwards if there's missing data and we need more samples.  But in this case there are no such thing as an OK sample, a missing sample is OK.  So we need to treat missing samples as OK.

### DIFF
--- a/cloud-formation/cfn.yaml
+++ b/cloud-formation/cfn.yaml
@@ -341,5 +341,6 @@ Resources:
       Period: 60
       EvaluationPeriods: 2
       Statistic: Sum
+      TreatMissingData: notBreaching
     DependsOn:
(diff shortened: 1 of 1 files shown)


johnduffell:
this is good in CODE 
image

johnduffell:
looks ok in prod 
image

PR #1330, opened 2018-12-20
Add 10% promo to GW annual subs

JustinPinner:
_Edited_: updated screen grabs

## Why are you doing this?
We want to activate the 10ANNUAL Guardian Weekly promo from support frontend. This was never brought forward from subscriptions-frontend, so it's a bit of a just-get-it-working approach at the moment.

Caveats:
It may not play nicely alongside flash sales - I haven't tested that yet. We don't have any GW flash sales running (AFAIK) but maybe it'd be better to define this as a long-running/open-ended sale for the weekly product?

It'll need a code change and deploy to remove it again if/when the time comes.

[**Trello Card**](https://trello.com/c/wAe1CS6S/2101-add-10-off-annual-back-into-gw-product-page)

## Changes

* Always use the 10ANNUAL code on weekly annual subs for ALL territories

## Screenshots
**UK**
screencapture-support-thegulocal-uk-subscribe-weekly-2018-12-21-11_58_35

**US**
screencapture-support-thegulocal-us-subscribe-weekly-2018-12-21-12_00_41

## Tested on CODE
Yes

### DIFF
--- a/assets/helpers/externalLinks.js
+++ b/assets/helpers/externalLinks.js
@@ -19,7 +19,6 @@ import type {
   SubscriptionProduct,
   WeeklyBillingPeriod,
   PaperBillingPlan,
-  WeeklyPromoCode,
 } from 'helpers/subscriptions';
 
 import { getPromoCode, getIntcmp } from './flashSale';
@@ -100,6 +99,7 @@ const defaultPromos: PromoCodes = {
   DigitalPack: getPromoCode('DigitalPack', 'GBPCountries', 'DXX83X'),
   Paper: getPromoCode('Paper', 'GBPCountries', 'GXX83P'),
   PaperAndDigital: getPromoCode('PaperAndDigital', 'GBPCountries', 'GXX83X'),
+  GuardianWeekly: getPromoCode('GuardianWeekly', 'GBPCountries', '10ANNUAL'),
 };
 
 const customPromos: {
@@ -303,7 +303,7 @@ function getWeeklyCheckout(
   cgId: CountryGroupId,
   nativeAbParticipations: Participations,
   optimizeExperiments: OptimizeExperiments,
-  promoCode?: ?WeeklyPromoCode,
+  promoCode: Option<string>,
 ): string {
   const acquisitionData = deriveSubsAcquisitionData(
     referrerAcquisitionData,
--- a/assets/helpers/externalLinks.js
+++ b/assets/helpers/externalLinks.js
@@ -15,7 +15,12 @@ import { type Option } from 'helpers/types/option';
 import type { Participations } from 'helpers/abTests/abtest';
 import { type OptimizeExperiments } from 'helpers/optimize/optimize';
 import { getBaseDomain } from 'helpers/url';
-import type { SubscriptionProduct, WeeklyBillingPeriod, PaperBillingPlan } from 'helpers/subscriptions';
+import type {
+  SubscriptionProduct,
+  WeeklyBillingPeriod,
+  PaperBillingPlan,
+  WeeklyPromoCode,
+} from 'helpers/subscriptions';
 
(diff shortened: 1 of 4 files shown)


assets/helpers/externalLinks.js
```diff
@@ -298,6 +303,7 @@ function getWeeklyCheckout(
   cgId: CountryGroupId,
   nativeAbParticipations: Participations,
   optimizeExperiments: OptimizeExperiments,
+  promoCode?: ?WeeklyPromoCode,
```
walaura: we've got [this `option` helper](https://github.com/guardian/support-frontend/blob/271db421a952d7a66ab4437d5d556e949e85b2a6/assets/helpers/types/option.js) that wraps the ? in a slightly more typesafe manner

rupertbates:
There is already a `defaultPromos` const in externalLinks.js which has promos for the various products, can't you add it in there?

assets/helpers/subscriptions.js
```diff
@@ -125,6 +125,54 @@ const paperSubscriptionPrices = {
   deliverySunday: GBP(15.12),
 };
 
+export type WeeklyPromoCode = '10ANNUAL';
```
rupertbates: This doesn't look like it should be a new type to me, there is already a `PromoCodes` type in externalLinks.js, surely this is an instance of that?

assets/helpers/externalLinks.js
```diff
@@ -95,6 +99,7 @@ const defaultPromos: PromoCodes = {
   DigitalPack: getPromoCode('DigitalPack', 'GBPCountries', 'DXX83X'),
   Paper: getPromoCode('Paper', 'GBPCountries', 'GXX83P'),
   PaperAndDigital: getPromoCode('PaperAndDigital', 'GBPCountries', 'GXX83X'),
+  GuardianWeekly: getPromoCode('GuardianWeekly', 'GBPCountries', '10ANNUAL'),
```
JustinPinner: We can get away with this because the promo code is the same in all territories

assets/helpers/subscriptions.js
JustinPinner: This was intended to guard against any other value than `10ANNUAL` being used. I've just made that a string type now because where it's used it's always a string, and switching to a `PromoCodes` type requires making deeper changes, which may not be worth it yet as I'm not sure how long this promo will be the default for.

This does mean there's a weakness at https://github.com/guardian/support-frontend/pull/1330/files#diff-7bdd389e4027e883c04ab0d735f2d779R259 because `promoCode` could be anything and may not exist in `subscriptionPromoPricesForGuardianWeekly`.

assets/helpers/externalLinks.js
```diff
@@ -298,6 +303,7 @@ function getWeeklyCheckout(
   cgId: CountryGroupId,
   nativeAbParticipations: Participations,
   optimizeExperiments: OptimizeExperiments,
+  promoCode: Option<string>,
```
rupertbates: `getDigitalCheckout` above just takes this value from the `defaultPromos` object would this work here to keep it consistent?

JustinPinner: `defaultPromos` are always for `GBPCountries` which I _think_ would be ok for now. Ideally I that would be a function that looks up defaults with the active the country code I suppose.

rupertbates: Ok, up to you I'm happy with it. We did discuss getting rid of the '10ANNUAL' code and just implementing this as a price change to the annual product but for now this looks fine.

katyabeer:
That's ace, Justin! Could we please replicate the 'introductory offer' label style in the Annual card please? It should say 'Save 10%'. Thanks a lot :)
Sorry for squeezing this change request in the last minute!!


screen shot 2018-12-21 at 11 42 31

PR #1331, opened 2018-12-20
Add selenium tests for subs

walaura:
## Why are you doing this?

Test that the subs landing pages are not completely broken post-deploy

[**Trello Card**](https://trello.com/c/sYEQ9kbt/2084-post-deploy-test-for-landing-pages)

## Screenshots
screenshot 2018-12-20 at 4 57 47 pm

### DIFF
--- a/test/selenium/subscriptions/LandingPagesSpec.scala
+++ b/test/selenium/subscriptions/LandingPagesSpec.scala
@@ -22,25 +22,31 @@ class LandingPagesSpec extends FeatureSpec with GivenWhenThen with BeforeAndAfte
 
   override def afterAll(): Unit = { driverConfig.quit() }
 
-  feature("Subscription landing pages") {
-
+  feature("Paper landing page") {
     scenario("Basic loading") {
 
       val paperSubsPage = new PaperSubs()
-      val weeklySubsPage = new WeeklySubs()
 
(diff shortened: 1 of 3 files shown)


test/selenium/subscriptions/LandingPagesSpec.scala
```diff
@@ -0,0 +1,46 @@
...
+
+      Given("that a test user goes to the paper subs landing page")
+      goTo(paperSubsPage)
+      Then("the page should display")
+      assert(paperSubsPage.pageHasLoaded)
+
+      Given("that a test user goes to the GW landing page")
+      goTo(weeklySubsPage)
+      Then("the page should display")
+      assert(weeklySubsPage.pageHasLoaded)
```
johnduffell: this should probably be two features I think?  rather than both in one - it should share the before/afterAll functionality so the browser shouldn't restart

johnduffell:
great! 👍 🥇 however my comment is above, have a look before merging

test/selenium/subscriptions/LandingPagesSpec.scala
walaura: okay!

walaura:
will do! never used selenium before >< (still need a github like 👍!)

PR #1332, opened 2018-12-20
Digipack support workers3

rupertbates:
## Why are you doing this?

This is part 3 of the work to hook support-frontend up to support-workers for Digital Pack creation.
The main part of this PR is the new endpoint for the checkout form to post to to create a state machine execution, but to make sure that this works I've also added basic code to post to it from the checkout form. 

This means we can now actually create a Digital Pack subscription end to end from support-frontend!

See the checklist below for next steps.

[**Trello Card**](https://trello.com/c/5N8LQ5yU/2060-be-able-to-trigger-the-creation-of-a-digital-pack-subscription-via-support-frontend)

## Changes
* Create a new controller for digital subscriptions as the `Subscriptions` one was getting pretty big
* Factor out some behaviour from the Subscriptions controller into a trait so that it can be mixed into both controllers
* Create a new `/create` endpoint in the `DigitalSubscriptions` controller to trigger a step function execution.
* Add code to the Digital Pack checkout page to activate Stripe and then post the form to the new `/create` endpoint

## TODO
- [ ] Error handling
- [ ] Progress indicator on form submission
- [ ] Thank you page
- [ ] Direct debit payment method
- [ ] Convert div to a real form

### DIFF
--- a/app/controllers/DigitalSubscription.scala
+++ b/app/controllers/DigitalSubscription.scala
@@ -28,7 +28,7 @@ import views.html.helper.CSRF
 
 import scala.concurrent.{ExecutionContext, Future}
 
-class DigitalPack(
+class DigitalSubscription(
     client: SupportWorkersClient,
     val assets: AssetsResolver,
     val actionRefiners: CustomActionBuilders,
--- a/app/controllers/DigitalSubscription.scala
+++ b/app/controllers/DigitalSubscription.scala
@@ -8,14 +8,12 @@ import cats.data.EitherT
 import cats.implicits._
 import com.gu.identity.play.IdUser
 import com.gu.support.config.{PayPalConfigProvider, StripeConfigProvider}
-import com.gu.support.workers.User
+import com.gu.support.workers.{BillingPeriod, User}
 import com.gu.tip.Tip
 import config.Configuration.GuardianDomain
 import config.StringsConfig
-import cookies.RecurringContributionCookie
 import io.circe.syntax._
 import lib.PlayImplicits._
-import monitoring.PathVerification._
 import monitoring.SafeLogger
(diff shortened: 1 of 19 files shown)


app/controllers/DigitalSubscription.scala
```diff
@@ -0,0 +1,161 @@
...
+      country = request.country,
+      state = request.state,
+      allowMembershipMail = false,
+      allowThirdPartyMail = user.statusFields.flatMap(_.receive3rdPartyMarketing).getOrElse(false),
+      allowGURelatedMail = user.statusFields.flatMap(_.receiveGnmMarketing).getOrElse(false),
+      isTestUser = false // testUsers.isTestUser(user.publicFields.displayName)
+    )
+  }
+
+  protected def respondToClient(
```
jacobwinch: Quite a bit of this logging / monitoring is recurring contributions specific

app/controllers/DigitalSubscription.scala
```diff
@@ -0,0 +1,149 @@
...
+        val billingPeriod = request.body.product.billingPeriod
+        SafeLogger.info(s"[${request.uuid}] User ${request.user.id} is attempting to create a new $billingPeriod digital subscription")
+        val result = for {
+          user <- identityService.getUser(request.user)
+          statusResponse <- client.createSubscription(request, contributor(user, request.body), request.uuid).leftMap(_.toString)
+        } yield statusResponse
+        respondToClient(result, request.body.product.billingPeriod)
+    }
+
+  private def contributor(user: IdUser, request: CreateSupportWorkersRequest) = {
```
jacobwinch: Sorry to be a pain, but we should rename this too!

PR #1335, opened 2018-12-24
Digital subscription checkout

rupertbates:
## Why are you doing this?

This is the next step in wiring up the Digital Pack checkout.

Most of this PR is taken up with a refactor of the way that billing periods are represented in support frontend. Previously we've had 2 separate types representing billing periods; one for contributions and one for subscriptions, this limits code reuse, is different to the back end where there is only one type and is generally a bad idea!

I have now consolidated the `BillingPeriod` types and values in a new file billingPeriod.js and used these throughout the codebase. I will probably do the same for payment methods next as this is similarly fragmented.

This PR also removes hard coded dummy values from the Digital Pack checkout and replaces them with the correct values.

[**Trello Card**](https://trello.com/c/vPu51Rgj/2131-call-the-endpoint-to-create-a-digital-pack-subscription-from-the-checkout)

### DIFF
--- a/assets/helpers/user/user.js
+++ b/assets/helpers/user/user.js
@@ -4,13 +4,20 @@
 
 import { routes } from 'helpers/routes';
 import * as cookie from 'helpers/cookie';
-import { getSession } from 'helpers/storage';
 import { get as getCookie } from 'helpers/cookie';
+import { getSession } from 'helpers/storage';
 import { defaultUserActionFunctions } from 'helpers/user/defaultUserActionFunctions';
 import type { UserSetStateActions } from 'helpers/user/userActions';
 
 
 // ----- Functions ----- //
+
+function isTestUser(): boolean {
+  const isDefined = x => x !== null && x !== undefined;
+  const uatMode = window.guardian && window.guardian.uatMode;
+  const testCookie = cookie.get('_test_username');
+  return isDefined(uatMode) || isDefined(testCookie);
+}
 
 const init = (dispatch: Function, actions: UserSetStateActions = defaultUserActionFunctions) => {
 
@@ -30,8 +37,6 @@ const init = (dispatch: Function, actions: UserSetStateActions = defaultUserActi
   const windowHasUser = window.guardian && window.guardian.user;
   const userAppearsLoggedIn = cookie.get('GU_U');
 
-  const uatMode = window.guardian && window.guardian.uatMode;
-
   function getEmailFromBrowser(): ?string {
     if (window.guardian && window.guardian.email) {
       return window.guardian.email;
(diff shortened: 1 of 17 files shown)


assets/pages/digital-subscription-checkout/helpers/paymentProviders.js
```diff
@@ -37,8 +56,21 @@ function create(state: State, token: string) {
...
+  const { paymentMethod, isTestUser } = state.page.checkout;
+  const price = getDigitalPrice(countryGroupId, state.page.checkout.billingPeriod);
+  switch (paymentMethod) {
+    case 'Stripe':
+      loadStripe()
+        .then(() => setupStripeCheckout((authorisation: StripeAuthorisation) => create(state, authorisation.token), 'REGULAR', currencyId, isTestUser))
+        .then(stripe => openDialogBox(stripe, price.value, state.page.checkout.email));
+      break;
+    case 'DirectDebit':
+      console.log('Direct Debit');
```
jacobwinch: Should these log statements be removed before shipping?

rupertbates: No this is all still work in progress, I've got to implement that next.

jacobwinch: thanks for explaining

assets/helpers/user/user.js
```diff
@@ -4,14 +4,21 @@
...
 import { get as getCookie } from 'helpers/cookie';
+import { getSession } from 'helpers/storage';
 import { defaultUserActionFunctions } from 'helpers/user/defaultUserActionFunctions';
 import type { UserSetStateActions } from 'helpers/user/userActions';
 
 
 // ----- Functions ----- //
 
+function isTestUser(): boolean {
+  const isDefined = x => x !== null && x !== undefined;
```
jranks123: Could we add a type to x? I think it looks a little neater:

`const isDefined = (x: string) => x !== null && x !== undefined;`

assets/helpers/user/user.js
```diff
@@ -41,15 +46,11 @@ const init = (dispatch: Function, actions: UserSetStateActions = defaultUserActi
 
   const emailFromBrowser = getEmailFromBrowser();
 
-  const isUndefinedOrNull = x => x === null || x === undefined;
-
-  const testUserCondition = (isUndefinedOrNull(uatMode) && cookie.get('_test_username')) || uatMode;
-
-  if (testUserCondition) {
+  if (isTestUser()) {
```
jranks123: Nice cleaning up of the logic here

assets/helpers/user/user.js
rupertbates: Argh, just merged! I'll change it for my next PR

PR #1337, opened 2018-12-27
[NPF] CSS cleanup

jranks123:
## Why are you doing this?
This PR refactors the new contributions landing page CSS

Main takeaways:

1) Removes support for grid. It isn't supported properly on Safari, which accounts for > 20% of our traffic, so it isn't worth the added complexity of having 2 sets of CSS. 

2) Removes all instances where we put multiple classes inside a media query, and moves to the pattern of each class having it's own media queries. 

3) Where possible, moves component specific CSS into it's own file, so as not to bloat the main CSS file. 

This has resulted in `contributionsLanding.scss` going from > 1200 lines to ~800 lines, and sets some standards going forwards. 

There should be no visual difference as a result of this PR.


rjbarbour-g:
Has this been tested across major browsers?

assets/pages/new-contributions-landing/contributionsLanding.scss
```diff
@@ -25,19 +32,149 @@ body {
...
+      transform: translate(-50%,0);
+      width: 1400px;
+    }
+  }
+}
+
+.gu-content--shrink {
+  @include mq($from: mobile, $until: tablet) {
+    $header-height: 44px;
+    grid-template-rows: [header-t] $header-height [header-b] 1fr [footer-t] auto [footer-b];
```
joelochlann: leftover grid stuff

assets/pages/new-contributions-landing/contributionsLanding.scss
```diff
@@ -165,472 +328,91 @@ header[role="banner"] {
...
 
-  footer[role="contentinfo"] {
-    grid-row: footer-t / footer-b;
-  }
+/*= FOOTER */
 
-  .gu-content__main {
-    grid-row: header-b / footer-t;
-  }
+footer[role="contentinfo"] {
```
joelochlann: maybe we could get rid of these tag/attribute selectors while we're at it, and replace with classes

joelochlann:
You could try using this to generate screenshots: https://github.com/guardian/paparazzi

PR #1341, opened 2019-01-03
Temporarily ignore scenario which is unstable due to a Redirect Test

jacobwinch:
## Why are you doing this?

This test is currently failing sporadically, because it is not being excluded from an Optimize test (https://github.com/guardian/support-frontend/wiki/AB-Testing-with-google-optimize#avoiding-problems-with-the-post-deploy-tests). 

We can revert this PR on Friday (once the test has finished).


jacobwinch:
Closing this because the test is now over anyway.

PR #1299, opened 2018-12-11
test hiding the name for 50% of users on one off

johnduffell:
## Why are you doing this?

We don't use the first and last name for one offs, but we insist on them being entered.  This PR tests what hpapens if we remove them.

[**Trello Card**](https://trello.com/c/vU7sj6vJ/227-remove-name-fields-test-single-contributions)
BEFORE::
image
AFTER:
image

@jranks123

PR #1345, opened 2019-01-03
remove the name fields completely from single contribtuions

johnduffell:
## Why are you doing this?

This test showed no real change from taking out the name fields, and those fields are not used for anything anyway.  This frees up the space to use for something else in future.
test pr was here: https://github.com/guardian/support-frontend/pull/1299

same trello card:
[**Trello Card**](https://trello.com/c/vU7sj6vJ/227-remove-name-fields-test-single-contributions-make-this-control)

updated screenshot has no change
image

however the recurring still has the fields
image

@jranks123

### DIFF
--- a/assets/pages/new-contributions-landing/checkoutFormIsSubmittableActions.js
+++ b/assets/pages/new-contributions-landing/checkoutFormIsSubmittableActions.js
@@ -89,7 +89,7 @@ const formIsValidParameters = (state: State) => ({
   firstName: state.page.form.formData.firstName,
   lastName: state.page.form.formData.lastName,
   email: state.page.form.formData.email,
-  showOneOffNameFields: state.common.abParticipations.showOneOffNameFields === 'control' || state.page.form.contributionType !== 'ONE_OFF',
+  showOneOffNameFields: state.page.form.contributionType !== 'ONE_OFF',
 });
 
 function enableOrDisableForm() {
--- a/assets/helpers/abTests/abtestDefinitions.js
+++ b/assets/helpers/abTests/abtestDefinitions.js
@@ -54,22 +54,6 @@ export const tests: Tests = {
     seed: 5,
   },
 
-  showOneOffNameFields: {
-    variants: [
-      'control',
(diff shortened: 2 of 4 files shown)


assets/pages/new-contributions-landing/checkoutFormIsSubmittableActions.js
```diff
@@ -89,7 +89,7 @@ const formIsValidParameters = (state: State) => ({
   firstName: state.page.form.formData.firstName,
   lastName: state.page.form.formData.lastName,
   email: state.page.form.formData.email,
-  showOneOffNameFields: state.common.abParticipations.showOneOffNameFields === 'control' || state.page.form.contributionType !== 'ONE_OFF',
+  showOneOffNameFields: state.page.form.contributionType !== 'ONE_OFF',
```
rupertbates: Should this variable be renamed to showNameFields? It seems odd to have a variable name showOneOffNameFields which is only true for non one off contributions!

johnduffell: excellent point, I'll fix that

johnduffell: I'll change to "hasNameFields" because this is actually validating them, not showing them

jranks123: Do you even need this actually? Seeing as you already pass in contributionType. You could change the line on line 74 to say `contributionType === 'ONE_OFF' ?`

johnduffell: thanks guys yeah I misread originally but I have renamed and moved it now, have a look 👍

johnduffell:
should I be fixing this issue? - it looks like we generally use single quotes though
image

johnduffell:
I had a look and found this
image
so I have changed that rule to check for single rather than double
image
and we can see what happens

PR #1346, opened 2019-01-03
Updates to the print product page for the print flash sale.

lmath:
## Why are you doing this?
To support the print flash sale. 

[**Trello Card**](https://trello.com/c/AMzW3SFV/2110-print-sale-updates)

## Changes
During a flash sale
* Updated heading 
* Updated prices displayed
* Once the user selects a plan, we send them to the appropriate checkout page on subs-frontend with the promo code applied
* Details of "Offer is for the first year. Standard subscription rates apply thereafter." added below the plans. 

note: paired with @jacobwinch 

## Screenshots
Compare to when flash sale is not on: https://support.theguardian.com/uk/subscribe/paper

`/uk/subscribe/paper`

| mobile-ish        | tablet-ish           | big desktop screen  |
| ------------- |-------------| -----|
|    |  |  |

`/uk/subscribe/paper/delivery`

| mobile-ish        | tablet-ish           | big desktop screen  |
| ------------- |-------------| -----|
|    |  |  |

### DIFF
--- a/assets/pages/paper-subscription-landing/components/content.jsx
+++ b/assets/pages/paper-subscription-landing/components/content.jsx
@@ -17,6 +17,7 @@ import { sendClickedEvent } from 'helpers/tracking/clickTracking';
 
 import { type State } from '../paperSubscriptionLandingPageReducer';
 import { setTab, type TabActions } from '../paperSubscriptionLandingPageActions';
+import { flashSaleIsActive } from 'helpers/flashSale';
 
 import Form from './form';
 
@@ -41,6 +42,12 @@ const ContentHelpBlock = ({ faqLink, telephoneLink }: {faqLink: Element<string>,
   </ProductPageContentBlock>
 );
 
+function flashSaleAwareProductPageInfoChip(): string {
+  if(flashSaleIsActive('Paper', 'GBPCountries')){
+    return "You can cancel your subscription at any time. Offer is for the first year. Standard subscription rates apply thereafter.";
+  }
+  return "You can cancel your subscription at any time.";
+}
 
 const ContentForm = ({ title, text }: {title: string, text?: Option<string>}) => (
   <ProductPageContentBlock type="feature">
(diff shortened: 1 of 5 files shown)


assets/pages/paper-subscription-landing/components/content.jsx
```diff
@@ -41,6 +42,12 @@ const ContentHelpBlock = ({ faqLink, telephoneLink }: {faqLink: Element<string>,
   </ProductPageContentBlock>
 );
 
+function flashSaleAwareProductPageInfoChip(): string {
```
rupertbates: Justin recently made quite a lot of changes to move all of the copy for flash sales into flashSale.js. We should probably do that with this copy to keep it consistent

rupertbates: Also I think you could just call this function `getInfoChipCopy` to make it more consistent with other files where we do the same thing

assets/pages/paper-subscription-landing/paperSubscriptionLandingPage.jsx
```diff
@@ -49,6 +50,15 @@ const store = pageInit(reducer(method, promoInUrl), true);
 
 // ----- Render ----- //
 
+function flashSaleAwareHeading(): string {
```
rupertbates: I think calling this `getHeading()` is more consistent with other parts of the code

assets/pages/paper-subscription-landing/components/content.jsx
lmath: Yes, good point re: the naming. I will then leave this in `content.jsx` (for now). 

I thought about putting this InfoChip copy in flashSale.js, but since this is the first print flash sale since the new print page, I'm not sure if this is a fundamental part of print flash sale changes that will happen each time yet, and it's also not a change that happens for each product (like e.g. a headline change is).

assets/pages/paper-subscription-landing/paperSubscriptionLandingPage.jsx
lmath: Thanks. I will rename this.

PR #1348, opened 2019-01-04
Disable the countdown timer.

lmath:
## Why are you doing this?
https://github.com/guardian/support-frontend/pull/1346

⬆️ There is a print flash sale starting today, but we'd like the countdown timer to appear later on. TBD if we will raise a PR just to add it back in whenever we want it, or some kind of feature to enable it x days before the end of the sale. Regardless, we should disable it for now. 

[**Trello Card**](https://trello.com/c/NyBumRDL/2134-countdown-timer-should-only-appear-during-the-last-week-of-the-sale)

## Changes

* Disable the countdown timer. There is a trello card about enabling it linked above. 

## Screenshots

| timer on (before this PR)  | timer back off (after this PR)       |
| ------------- |-------------| 
|    |  |

### DIFF
--- a/assets/pages/subscriptions-landing/components/featuredProductAb.jsx
+++ b/assets/pages/subscriptions-landing/components/featuredProductAb.jsx
@@ -83,7 +83,7 @@ function FeaturedProductAb(props: PropTypes) {
       cta={getCta(product)}
       headingSize={headingSize}
       product={product.name}
-      hasTimer={flashSaleIsActive(product.name, countryGroupId)}
+      hasTimer={false}
       countryGroupId={countryGroupId}
     />) : null;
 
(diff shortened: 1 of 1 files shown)


SarahLaughton:
Looks good to me ✅

assets/pages/subscriptions-landing/components/featuredProductAb.jsx
```diff
@@ -83,7 +82,7 @@ function FeaturedProductAb(props: PropTypes) {
       cta={getCta(product)}
       headingSize={headingSize}
       product={product.name}
-      hasTimer={flashSaleIsActive(product.name, countryGroupId)}
+      hasTimer={false}
```
lmath: Here's a function I played around with that works but is a hack in case it's useful as somewhere to start later on

```
function showTimer(product: Product, countryGroupId: countryGroupId): boolean {
  if(flashSaleIsActive(product.name, countryGroupId)) {
    const now = new Date();
    const endTime = getEndTime(product.name, countryGroupId);
    const sevenDaysFromNow = now.setDate(now.getDate() + 7);

    return sevenDaysFromNow - endTime > 0;
  }
  return false;
}
```

lmath: There is now a function to do this added in https://github.com/guardian/support-frontend/pull/1455

PR #1354, opened 2019-01-08
Use correct log level for success messages

jacobwinch:
## Why are you doing this?

I noticed that this log level was incorrect (Sentry emailed me about it whilst I was testing).


jacobwinch:
This looks like a problem with the PayPal sandbox 😒

PR #1358, opened 2019-01-09
Fix contributions link on showcase page on mobile

walaura:
## Why are you doing this?

Tiny fix for an issue where the svg would overlap the button., adds z-indexes so the button is always on "top" of the svg.

While I was here I also fixed the button sizing because they were slightly too large on this page and it's been bugging me for ages.

screenshot 2019-01-09 at 1 08 34 pm


joelochlann:
thank you @walaura!

PR #1359, opened 2019-01-09
Fix header on checkout thank you page

walaura:
## Why are you doing this?
Updates the thank you page header to match the one from the main checkout, this solves an issue with how it displays on mobile and is more closely aligned with the original design:

screenshot 2019-01-09 at 3 02 46 pm


LATaylor-guardian:
👍

PR #1364, opened 2019-01-11
Revert "Revert "[NPF] CSS cleanup""

jranks123:
Reverts guardian/support-frontend#1361, and fixes the bug that was the reason for the original revert


assets/pages/new-contributions-landing/components/SetPasswordForm.jsx
```diff
@@ -18,7 +18,7 @@ import { trackComponentClick } from 'helpers/tracking/ophanComponentEventTrackin
 import { NewContributionTextInput } from './ContributionTextInput';
 import { type ThankYouPageStage } from '../contributionsLandingReducer';
 import { setThankYouPageStage, setPasswordHasBeenSubmitted, setPasswordError, updatePassword, type Action } from '../contributionsLandingActions';
-import { ButtonWithRightArrow } from './ButtonWithRightArrow';
+import { ButtonWithRightArrow } from './ButtonWithRightArrow/ButtonWithRightArrow';
```
walaura: It's Saturday so im not reviewing this just yet BUT this is a self reminder for us frontend devs to have a chat about button reunification! I think the design has settled to the point we can start thinking of how to build 1 True Button with all its pesky edge cases

PR #1366, opened 2019-01-14
Update ty page marketing component

walaura:
## Why are you doing this?

It was rendering quite poorly and looked bad, this fixes it. In doing so it updates the marketing consent container button to use `productPageButton` instead, so to reuse the styles.

[**Trello Card**](https://trello.com/c/XSqJvW5F/2142-fix-checkout-ty-page-layout-issues)

## Screenshots
support thegulocal com_uk_subscribe_digital_checkout_displaycheckout true
screenshot 2019-01-14 at 2 28 20 pm


assets/components/productPage/productPageButton/productPageButton.jsx
```diff
@@ -4,6 +4,7 @@
```
jranks123: What does product page button refer to? Is that the name of the "showcase page"? In which case, maybe we should call this something more general as it's now being used on the contributions flow?

assets/pages/new-contributions-landing/components/ContributionThankYou/ContributionThankYou.scss
```diff
@@ -126,3 +104,21 @@
...
 }
 
+.component-marketing-consent.component-marketing-consent--newsletter {
+  border-bottom: 1px solid gu-colour(garnett-neutral-4);
+}
+
+.component-marketing-consent__message {
+  font-family: $gu-text-egyptian-web;
+  letter-spacing: -0.5px;
+  margin-bottom: 24px;
```
jranks123: You could use ``$gu-v-spacing` * 2` here

assets/components/productPage/productPageButton/productPageButton.jsx
walaura: this is why I wanna have this chat on buttons, to sort out how we wanna unify buttons. At the moment i just did this to unblock the checkout landing so we didn't have to reuse styles

assets/pages/new-contributions-landing/components/ContributionThankYou/ContributionThankYou.scss
walaura: I copypasted the current code :D happy to do this in a subsequent PR

jranks123: Would that not be overkill? A whole PR to change `24px` to `$gu-v-spacing * 2`? Up to you, but I would recommend cleaning up as you go along here!

PR #1367, opened 2019-01-14
Add some server side validation to the digital pack checkout (firstName and lastName)

lmath:
## Why are you doing this?
Most of the time, empty or invalid fields will be caught by client-side validation. However, if the client side validation doesn't run for some reason we know that it's not worth trying to create a subscription for that user if required fields are blank or fields are invalid. 

So, when the user fills in the digital pack checkout and fills in payment details, the checkout will call `POST /subscribe/digital/create`. Now this endpoint will return with a 500 if firstName or lastName are blank, and the checkout will catch error responses and display an error message to the user. 

PR to follow that returns a 400 instead.

[**Trello Card**](https://trello.com/c/tr3TQ4YH/2033-server-side-validation)

## Changes
* If firstName or lastName make it to the server as an empty string, we now return an error from the server response rather than starting an execution of the state machine that will inevitably fail 
* Added another error reason for when the user's name and contact info are invalid
* if the call to `POST /subscribe/digital/create` fails, then we catch that error and display a message to the user

**REVIEWERS:**

To try out these changes locally, you'll need to temporarily disable client side validation here
* https://github.com/guardian/support-frontend/blob/master/assets/pages/digital-subscription-checkout/digitalSubscriptionCheckoutReducer.js#L136

In addition to the usual PR feedback
* Please also comment on this PR what else can be validated (to follow in another PR)
* Please suggest better wording for the error 😄 

## Screenshots

image


app/controllers/DigitalSubscription.scala
```diff
@@ -106,14 +107,20 @@ class DigitalSubscription(
...
+      implicit request: AuthRequest[CreateSupportWorkersRequest] =>
         val billingPeriod = request.body.product.billingPeriod
         SafeLogger.info(s"[${request.uuid}] User ${request.user.id} is attempting to create a new $billingPeriod digital subscription")
-        val result = for {
-          user <- identityService.getUser(request.user)
-          statusResponse <- client.createSubscription(request, createUser(user, request.body), request.uuid).leftMap(_.toString)
-        } yield statusResponse
-        respondToClient(result, request.body.product.billingPeriod)
+
+        if (validationPasses(request.body)) {
```
lmath: This `if/else` is ugly and I'm currently thinking of how to make it less ugly, so if you have any suggestions I am all 👂

lmath:
I'm not validating phone number because I don't think it's actually being sent to `/subscribe/digital/create` currently.

I think it would be sensible to check that there is some value provided for country (since the user could leave it as --) but since payment frequency and payment method are radio buttons with a default I'm not sure how it's possible to send nothing even if client side validation fails. 

Anything else?

assets/pages/digital-subscription-checkout/components/checkoutForm.jsx
```diff
@@ -105,8 +105,10 @@ function statesForCountry(country: Option<IsoCountry>): React$Node {
 
 function CheckoutForm(props: PropTypes) {
 
+  const errorHeading = props.submissionError === 'personal_details_incorrect' ? 'Failed to Create Subscription' :
```
walaura: I'd try to move this to `errorReasons.js` if possible. Do we need a heading at all?

lmath: Ooh, good point. We are currently using the `GeneralErrorMessage` component for a payment-related error so I suppose there are a few options
1) Use the `ErrorMessage` component for both. Then neither have a heading.
2) Put the logic to select an error heading in `generalErrorMessage.jsx` instead. (Is that any better?)
3) Use an `ErrorMessage` for form validation and a `GeneralErrorMessage` for payment errors
4) Rejig general error message to have the errorHeading be optional 
5) Something else?

walaura: The way I'd go about this probably is to update the `appropriateErrorMessage()` function to return the body and a title maybe? So all the wording is in the same place

lmath: I think the error reason `personal_details_incorrect` always implies an error message that is like `Please double check the name and contact details you provided and try again.` but not necessarily always the heading `Failed to Create Subscription`. If I can think of a heading that is more general but not so general as to be meaningless then yes, updating `appropriateErrorMessage()` seems like the best approach.

app/controllers/DigitalSubscription.scala
```diff
@@ -138,10 +145,13 @@ class DigitalSubscription(
     result.fold(
       { error =>
         SafeLogger.error(scrub"[${request.uuid}] Failed to create new $billingPeriod Digital Subscription, due to $error")
-        InternalServerError
+        error match {
+          case "validation-failure" => BadRequest
```
lmath: This is a cry for help.

lmath: I tried to use types like a sane person, but the eitherT/covariance confusion is too much. Do you have time to chat with me about this @jacobwinch ?

assets/helpers/paymentIntegrations/newPaymentFlow/readerRevenueApis.js
```diff
@@ -177,6 +177,9 @@ function postRegularPaymentRequest(
       if (response.status === 500) {
```
jacobwinch: So just to confirm I've understood correctly, when this PR is merged, we'll return:

```
case 'internal_error':
```

And then we'll start displaying this friendlier error after a subsequent PR is merged:

```
case 'personal_details_incorrect':
  return 'Please double check the name and contact details you provided and try again.';
```

Right now the user experience is very poor (I expect they'll see the pending thankyou page, and then get a failure email later), so even this interim change is still a big improvement 👍

lmath: That's right!

PR #1369, opened 2019-01-15
Handle 500s on calls to endpoints that create subscriptions

lmath:
## Why are you doing this?
`POST /subscribe/digital/create `
and 
`POST /contribute/recurring/create`

accept the same request body and both start the process of creating a recurring payment via support-workers. So, `postRegularPaymentRequest()` is used by both the recurring contributions and digital pack checkout. 
Both create endpoints currently may return a 500, but when that happens, this happens:
image
(and the page hangs)

Try it out: 
- force a 500 response
image
- go through the checkout in `https://support.thegulocal.com/uk/contribute` 

(the same applies to https://support.thegulocal.com/uk/subscribe/digital/checkout?displayCheckout=true)

then checkout this branch and instead see an error
image

[**Trello Card**](https://trello.com) - there isn't one

## Changes

* When calling `POST /subscribe/digital/create ` or `POST /contribute/recurring/create`, now we do not immediately convert the response to a json, so that we can handle a 500.

### DIFF
--- a/assets/helpers/paymentIntegrations/newPaymentFlow/readerRevenueApis.js
+++ b/assets/helpers/paymentIntegrations/newPaymentFlow/readerRevenueApis.js
@@ -179,7 +179,8 @@ function postRegularPaymentRequest(
         return ({ paymentStatus: 'failure', error: 'internal_error' });
       }
 
-      return response.json().then(checkRegularStatus(participations, csrf, setGuestAccountCreationToken, setThankYouPageStage));
+      return response.json()
+        .then(checkRegularStatus(participations, csrf, setGuestAccountCreationToken, setThankYouPageStage));
 
     })
     .catch(() => {
--- a/assets/helpers/paymentIntegrations/newPaymentFlow/readerRevenueApis.js
+++ b/assets/helpers/paymentIntegrations/newPaymentFlow/readerRevenueApis.js
@@ -179,7 +179,8 @@ function postRegularPaymentRequest(
         return ({ paymentStatus: 'failure', error: 'internal_error' });
       }
 
-      return checkRegularStatus(participations, csrf, setGuestAccountCreationToken, setThankYouPageStage)(response);
+      return response.json().then(checkRegularStatus(participations, csrf, setGuestAccountCreationToken, setThankYouPageStage));
+
     })
     .catch(() => {
       logException(`Error while trying to interact with ${uri}`);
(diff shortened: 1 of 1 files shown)


lmath:
Hey @jranks123 @joelochlann @johnduffell @tomrf1

Since this affects recurring contributions would love thoughts from someone on your team 😃

assets/helpers/paymentIntegrations/newPaymentFlow/readerRevenueApis.js
```diff
@@ -172,10 +172,20 @@ function postRegularPaymentRequest(
...
 ): Promise<PaymentResult> {
-  return logPromise(fetchJson(
-    uri,
-    requestOptions(data, 'same-origin', 'POST', csrf),
-  ).then(checkRegularStatus(participations, csrf, setGuestAccountCreationToken, setThankYouPageStage)));
+  return logPromise(fetch(uri, requestOptions(data, 'same-origin', 'POST', csrf)))
+    .then((response) => {
+      if (response.status === 500) {
+        logException(`500 Error while trying to post to ${uri}`);
+        return ({ paymentStatus: 'failure', error: 'internal_error' });
```
walaura: The way I'd do this is to  `throw` an error here then catch it later in the promise flow, happy to have a chat!

assets/helpers/paymentIntegrations/newPaymentFlow/readerRevenueApis.js
```diff
@@ -172,10 +172,21 @@ function postRegularPaymentRequest(
...
```
johnduffell: so to confirm my reading of the code - basically if we get a 500, we don't want to assume the content is anything meaningful (JSON), better to just replace whatever content we got back with this hard coded json and then we can deal with it properly later.
If this was Scala I'd want to return an Either where the Right is whatever actually came from the server, and if it's a left then we didn't get anything useful back.  Then we are not having to fake a response that didn't happen.  But I appreciate this is JS where turning everything into more simply typed json is more usual?

johnduffell:
good spot and it looks good to me (with my scala hat on I have detailed my reservation having the format of the json generated also in the client side code) Either wy this is definitely an improvement on before regardless 👍 💯

assets/helpers/paymentIntegrations/newPaymentFlow/readerRevenueApis.js
lmath: I also wished for an Either 😭

PR #1373, opened 2019-01-17
return a bad request if we know the request to /subscribe/digital/create is missing information

lmath:
## Why are you doing this?
In the digital pack checkout, we want to add some server side validation in case the client side validation doesn't run for some reason. 

Currently, if the first name or last name comes through as an empty string in the request, we wish to just short circuit and return an error because it's not even worth trying to create a digital subscription. Some basic request body validation was added here: https://github.com/guardian/support-frontend/pull/1367

But if the validation fails, we should return a 400 rather than a 500.

That way, we can display a more useful error message to the user, prompting them to check the details they gave us in the form, rather than the default "Sorry, something has gone wrong." message. 

That's what this PR is about. 

[**Trello Card**](https://trello.com/c/tr3TQ4YH/2033-server-side-validation)

## Changes

* If a request to `POST /subscribe/digital/create` has a request body that fails the request body validation added in https://github.com/guardian/support-frontend/pull/1367, it will return a 400 (handled on the client side already here: https://github.com/guardian/support-frontend/blob/master/assets/helpers/paymentIntegrations/newPaymentFlow/readerRevenueApis.js#L180)

## Screenshots
 
image

### DIFF
--- a/app/controllers/DigitalSubscription.scala
+++ b/app/controllers/DigitalSubscription.scala
@@ -105,6 +105,16 @@ class DigitalSubscription(
     )
   }
 
+  sealed trait CreateDigitalSubscriptionError {
+    def asString: String
+  }
+  case class ServerError(message: String) extends CreateDigitalSubscriptionError {
+    override def asString: String = message
+  }
+  case class RequestValidationError(message: String) extends CreateDigitalSubscriptionError {
+    override def asString: String = message
+  }
+
   def create: Action[CreateSupportWorkersRequest] =
     authenticatedAction(recurringIdentityClientId).async(circe.json[CreateSupportWorkersRequest]) {
       implicit request: AuthRequest[CreateSupportWorkersRequest] =>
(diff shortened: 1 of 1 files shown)


app/controllers/DigitalSubscription.scala
```diff
@@ -105,20 +105,36 @@ class DigitalSubscription(
...
     authenticatedAction(recurringIdentityClientId).async(circe.json[CreateSupportWorkersRequest]) {
       implicit request: AuthRequest[CreateSupportWorkersRequest] =>
         val billingPeriod = request.body.product.billingPeriod
         SafeLogger.info(s"[${request.uuid}] User ${request.user.id} is attempting to create a new $billingPeriod digital subscription")
 
         if (validationPasses(request.body)) {
-          val result: EitherT[Future, String, StatusResponse] = for {
-            user <- identityService.getUser(request.user)
-            statusResponse <- client.createSubscription(request, createUser(user, request.body), request.uuid).leftMap(_.toString)
+          val userOrError: EitherT[Future, CreateDigitalSubscriptionError, IdUser] = identityService.getUser(request.user).leftMap(ServerError(_))
```
lmath: These are extracted because otherwise the `result` just below causes a compilation failure, expecting 
```
val result: EitherT[Future, ServerError, StatusResponse]
```

app/controllers/DigitalSubscription.scala
```diff
@@ -105,20 +105,36 @@ class DigitalSubscription(
...
   }
 
+  sealed trait CreateDigitalSubscriptionError {
+    def asString: String
+  }
+  case class ServerError(message: String) extends CreateDigitalSubscriptionError {
+    override def asString: String = message
+  }
+  case class RequestValidationError(message: String) extends CreateDigitalSubscriptionError {
+    override def asString: String = message
```
johnduffell: do you need the asString as well as message or can you just call it message (and remove asString from the subclasses)?
or vice versa

app/controllers/DigitalSubscription.scala
johnduffell: yes scala's type inferencing only really works in one direction so it often backs itself into a corner early on.
You can always create a type alias as a middle ground if you find it's tidier to use that instead
```suggestion
          type MyOp[RES] = EitherT[Future, CreateDigitalSubscriptionError, RES]
          val userOrError: MyOp[IdUser] = identityService.getUser(request.user).leftMap(ServerError(_))
```

johnduffell: another option is to pass a type parameter to leftMap so that it doesn't try to infer it (wrongly)
```suggestion
          val userOrError = identityService.getUser(request.user).leftMap[CreateDigitalSubscriptionError](ServerError(_))
```

johnduffell: Incidentally, I found all the same problems with the zuora-auto-cancel stuff about inferring , and to avoid brushing up against so many type parameters I ended up making ApiGatewayOp and ClientFailableOp rather than using the built in Either.  It did end up making things look a bit nicer, but I think the way you are doing it is more correct.

lmath: Thanks John! Good to consider these options. I will see how the code looks if I add a type alias.

app/controllers/DigitalSubscription.scala
```diff
@@ -105,20 +105,32 @@ class DigitalSubscription(
     )
   }
 
+  sealed class CreateDigitalSubscriptionError(message: String)
```
johnduffell:
```suggestion
  sealed abstract class CreateDigitalSubscriptionError(message: String)
```

app/controllers/DigitalSubscription.scala
johnduffell: there are always lots of options in scala!

PR #1378, opened 2019-01-21
Update SimpleHeader to optionally display navigation links

walaura:
## Why are you doing this?

The new unified header!…well, part of it.

This adds desktop navigation links to the header in `support-frontend`. Sort of. Namely, this PR adds **the potential** (under a prop) for the header to show links. It also heavily reworks the CSS to add all required styling and makes the country picker pretty and modern (took me long enough).

Next steps will be to go over this with design, (had to make a couple compromises at this point to not ship 1 humongous PR) then plug all of this into a google optimise AB test.

[**Trello Card**](https://trello.com/c/KYf6b8QN/2154-add-navigation-to-subscriptions-pages-on-support-site-only-desktop)

## Tell us something entertaining

Well, the design calls for the nav to sort of cozily tuck itself under the logo provided it fits. A bit like a cat:

screenshot 2019-01-21 at 6 30 30 pm

To achieve this I'm using [resize observers](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver/ResizeObserver) (Think of them as element-level media queries). 
Now, this is completely unnecessary and a breakpoint would suffice provided the nav element count doesn't change but I really like the flexibility of it and it's been a good learning exercise in terms of what we can do with layout. 

Since there's no polyfill available I've created a helper function to wrap it that'll fall back to a less performant `window.resize` variant if needed. The full way this works is as follows:

1. First it uses a normal `window.resize` if there's no `ResizeObserver` in the browser. This is possible because instead of using the resizeObserver callback i'm getting the widths manually every time it gets called (it's more straightforward in this case, orchestrating 3 elements).
2. Should that fail for any reason (yet everything else is somehow still running) it'll fall back to the widest possible nav as a default. At this point the only way to break it is to overflow the biggest possible nav by throwing a bunch of links. Which is not really fixable with code.

## Screenshots

new, with nav:
screenshot 2019-01-21 at 6 04 10 pm

new without nav:
screenshot 2019-01-21 at 6 26 23 pm

current:
screenshot 2019-01-21 at 6 27 09 pm


assets/components/headers/simpleHeader/simpleHeader.scss
```diff
@@ -3,50 +3,138 @@
...
+}
+
+.component-simple-header-topnav-logo__graun {
   float: right;
   height: 100%;
   flex: 0 1 auto;
 }
+
+.component-simple-header-bottomnav {
+  display: none;
```
joelochlann: why do we have the `display: none;`? isn't it overriden by the very next line?

walaura: yes 😹

assets/components/headers/simpleHeader/simpleHeader.jsx
```diff
@@ -2,37 +2,133 @@
...
+        }
+        >
+          <a className="component-simple-header-bottomnav__link" href={href}>{text}</a>
+        </li>
+    ))}
+    </ul>
+  </nav>
+);
+
+export default class SimpleHeader extends Component<PropTypes, State> {
```
joelochlann: maybe this shouldn't be called `SimpleHeader` anymore! 😛

assets/components/headers/simpleHeader/simpleHeader.jsx
```diff
@@ -2,37 +2,133 @@
...
+);
+
+type HeaderBottomNavigationPropTypes = {|
+  getMenuRef: (?Element) => void
+|};
+
+const HeaderBottomNavigation = ({ getMenuRef }: HeaderBottomNavigationPropTypes) => (
+  <nav className="component-simple-header-bottomnav">
+    <ul className="component-simple-header-bottomnav__ul" ref={getMenuRef}>
+      {links.map(({ href, text }) => (
```
joelochlann: it could be nice if the header accepted links as an argument rather than hardcoding the dependency here - that way the header could easy be used with different nav bars

joelochlann:
RE: accepting links as arguments rather than hard-coding the dependency. After discussion with @walaura it seems the thinking is that locking it down is actually a benefit, since it's assumed it won't change across the site. I'm not so sure though - if we use it on contributions, I'd expect there to be a much more restricted nav (since we've always seen financial gains from reducing noise & choice on the contributions flow). Happy with whatever you decide on that but just flagging up a potential future use case!

assets/components/headers/simpleHeader/simpleHeader.jsx
walaura: haha, true that! will address

walaura:
@joelochlann absolutely agree on letting the linsk change. i think the best path fwd for now is to "hardcode" them but make sure there's a single `links` object referenced at the root and cascading down. Meaning, it would be trivial to turn it into a prop when that need arises

PR #1384, opened 2019-01-22
Make contribution type default to the type specified in the URL parameters, if exists

jranks123:
## Why are you doing this?
If a user is in the contributions type test but the link has a url parameter saying what contribution types to display, we want to respect the url parameters and make sure we don't put that user into the contribution types test. 

Also, in the session storage, change:

`contribType` and `contributionType` -> `selectedContributionType`
`paymentMethod` -> `selectedPaymentMethod`

for consistency

### DIFF
--- a/assets/helpers/abTests/abtestDefinitions.js
+++ b/assets/helpers/abTests/abtestDefinitions.js
@@ -1,5 +1,6 @@
 // @flow
 import { isFromEpicOrBanner } from 'helpers/referrerComponent';
+import { getQueryParameter } from 'helpers/url';
 import type { Tests } from './abtest';
 
 // ----- Tests ----- //
@@ -49,6 +50,7 @@ export const tests: Tests = {
     isActive: true,
     independent: true,
     seed: 5,
+    canRun: () => !getQueryParameter('contributionTypes'),
   },
 
 };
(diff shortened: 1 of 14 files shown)


assets/helpers/abTests/abtestDefinitions.js
```diff
@@ -49,6 +50,7 @@ export const tests: Tests = {
     isActive: true,
     independent: true,
     seed: 5,
+    canRun: () => !getQueryParameter('contributionTypes'),
```
joelochlann: you need to check for `defaultContributionType` too

I notice that we're calling it [`contributionType` in session storage](https://github.com/guardian/support-frontend/blob/master/assets/helpers/checkouts.js#L94) but `defaultContributionType` in the URL. Maybe they should both be `selectedContributionType` for clarity.

jranks123:
@joelochlann good idea, I've also changed `paymentMethod` to `selectedPaymentMethod` for consistency (as well as the legacy `contribType` which is used by the old flow, because why not)

assets/helpers/tracking/googleTagManager.js
```diff
@@ -42,11 +42,11 @@ function getOrderId() {
 }
 
 function getContributionType() {
-  const param = getQueryParameter('contribType');
+  const param = getQueryParameter('selectedContributionType');
```
joelochlann: do you think the param for the old flow should be named the same? it's basically determining the whole page, not just which tab is selected - so maybe its clearer to keep the old name

jranks123: Ok I'll remove. Can I get a conditional approve?

PR #1385, opened 2019-01-23
Make amounts configurable

tomrf1:
## Why are you doing this?
Currently, the amounts (by region/contribution type) are configured in the source code.
This PR makes them configured from s3 (to be managed via the support-admin-console).

I recommend viewing each commit separately - the 1st commit is just moving some things

[**Trello Card**](https://trello.com/c/wO05dMsk/303-support-frontend-make-amounts-configurable)

## Changes

* Server now gets the amounts config from s3, in addition to the switches
* Client receives the amounts in the `window.guardian.settings` object (alongside the switches) and uses this to set the amounts + defaults.
* Removed the annual amounts test (which isn't running)

Tested on new and old payment flows.
The old payment flow still uses defaults hardcoded in the `config` object in `contributions.js`, but it's not worth changing this.


app/admin/settings/Amounts.scala
```diff
@@ -0,0 +1,22 @@
+package admin.settings
+
+import io.circe.{Decoder, Encoder}
+
+case class Amount(value: String, isDefault: Option[Boolean])
```
tomrf1: The value is a string on the client, so for simplicity I've kept it that was here as well.
This is a bit wrong, but I'm not sure it's worth changing

app/admin/settings/Amounts.scala
```diff
@@ -0,0 +1,22 @@
...
+  NZDCountries: List[Amount],
+  Canada: List[Amount]
+)
+case class Amounts(ONE_OFF: AmountsRegions, MONTHLY: AmountsRegions, ANNUAL: AmountsRegions)
+
+object Amounts {
+  import io.circe.generic.auto._
+
+  implicit val amountsEncoder = Encoder[Amounts]
+  implicit val amountsDecoder = Decoder[Amounts]
```
johnduffell: oviously this is an existing project, however I normally use the convention to prepend the class name with "Wire" if it is used to generate a serialiser (e.g. `WireAmounts`, to clearly distinguish it from Domain abstraction level case classes and also make it more clear where renaming is safe.  In many cases we keep the domain model and wire model the same to reduce the amount of code.

conf/DEV.public.conf
```diff
@@ -30,4 +30,10 @@ settingsSource {
       path="~/.gu/support-admin-console/switches.json"
     }
   }
+  amounts {
+    s3 {
```
johnduffell: could be amounts.s3 the same as the others?

johnduffell: personally I'm not a fan of putting so much in config, the only thing that changes is the DEV/PROD part so all the rest could be hard coded.  In my mind config is useful for changing things without a recompile, and we would generally rebuild every time we edit these files anyway so we are just losing type safety and IDE support for nothing..

app/admin/settings/Settings.scala
```diff
@@ -57,13 +47,14 @@ object Settings {
...
 
-case class SettingsSources(switches: SettingsSource)
+case class SettingsSources(switches: SettingsSource, amounts: SettingsSource)
 
 object SettingsSources {
   def fromConfig(config: Config): Either[Throwable, SettingsSources] = {
     for {
       switchesSource <- SettingsSource.fromConfig(config, "switches")
-    } yield SettingsSources(switchesSource)
+      amountsSource <- SettingsSource.fromConfig(config, "amounts")
```
johnduffell: is this reloaded regularly while the app is running or would it only run on startup?  obviously doesn't matter at this stage, but I'm interested.

johnduffell:
personally I would have preferred doing PRs as you went along as this one is rather large.  But having said that the scala looks fine.  I think the client side stuff should be reviewed by someone who knows more!  But overall this will be a great step towards easy configuration so good work. 👍

app/admin/settings/Settings.scala
tomrf1: The `SettingsSource` is loaded once at startup - it tells the program where to find the amounts config. The switches/amounts config *is* then reloaded regularly

conf/DEV.public.conf
tomrf1: good point, will have a think

tomrf1:
@johnduffell I guess the server code could have been a separate PR, though it is a separate commit. Most of the file changes are in the first commit because I moved some files!

conf/DEV.public.conf
tomrf1: I'll revisit this in a follow up PR as this one is big enough already!

johnduffell: that's a sensible approach!

PR #1387, opened 2019-01-23
Install storybook

walaura:
(don't freak out about the line count! it's the yarn lockfile)

## Why are you doing this?
So we can have a canonical place to display, build, and test shared UI components in. The goals of this are:

- To have a unified point of reference for this stuff for dev, visual design and uxd to refer to and avoid rebuilding.
- To allow visual design to propose (or ship) enhancements at a component level (for example @zeftilldeath fixing all the hover states of the buttons, mine are atrocious)
- To allow me, Laura, to play around with setting up storybook and knobs.

[[Read more on storybook here]](https://storybook.js.org/)

Still quite early stages as this is checkout week. Use`yarn storybook` to run.

## Screenshots

screenshot 2019-01-23 at 2 50 38 pm

## Is this a design system

**No** that's a cursed word, it's a tool for UI development that I'm keen for us to start using, let's see how we use it and how it evolves.

### DIFF
--- a/.storybook/addons.js
+++ b/.storybook/addons.js
@@ -1,0 +1,1 @@
+import '@storybook/addon-knobs/register';
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -1,0 +1,7 @@
+import { configure } from '@storybook/react';
+
+function loadStories() {
+  require('../stories/button.jsx');
+}
+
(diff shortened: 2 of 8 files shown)


gtrufitt:
I couldn't choose a single gif https://giphy.com/search/excited

PR #1388, opened 2019-01-23
Align Checkout & Product page styles+components

walaura:
## Why are you doing this?

I've been meaning to do this for ages! But seriously though, the rationalisation for this is:
- For the users, whom we love: There's a stronger visual consistency going through the flow, which makes the whole thing feel more valid, and
- For us developers: This checkout form will be the "golden standard" upon which to develop the others. It is likely they will literally be copypasted as a starting point. Changing stuff _now_ involves significantly less work than changing it down the line times three.
- it wasn't that much work really? It's not like we had two completely different sets of pages, just minor inconsistencies all over. This gets us closer to the ideal checkouts our designers have conceived.

To do this I've created a couple small components under `checkout` to help us (they will all grow with the other 2 checkouts but I didn't wanna codify too many design assumptions at this stage). Also tweaked the CSS so it uses the font and size mixins & variables. Also used as many styles from `productPage` as possible (TBD: rename `productPage` to `page` or something)

[**Trello Card**](https://trello.com/c/GJjUbIfI/2168-style-digital-pack-checkout)

## Todo

- [x] test error message styles

## Screenshots

support thegulocal com_uk_subscribe_digital_checkout_displaycheckout true 4

support thegulocal com_uk_subscribe_digital_checkout_displaycheckout true 2

The TY page looks about the same for the scope of this PR, I'm attaching the screenie to show there's no ugly regressions

### DIFF
--- a/assets/components/checkout/checkout.jsx
+++ b/assets/components/checkout/checkout.jsx
@@ -1,0 +1,19 @@
+// @flow
+
+import React, { type Node } from 'react';
+
+import './checkout.scss';
+
+
+type PropTypes = {
+  children: Node,
+};
+const Checkout = ({ children }: PropTypes) => (
+  <div className="component-checkout">
+    <div className="component-checkout__form">
+      {children}
+    </div>
+  </div>
+);
+
+export default Checkout;
--- a/app/controllers/DigitalSubscription.scala
+++ b/app/controllers/DigitalSubscription.scala
@@ -143,6 +143,7 @@ class DigitalSubscription(
       lastName = request.lastName,
       country = request.country,
       state = request.state,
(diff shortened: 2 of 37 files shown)


assets/components/checkout/checkout.jsx
```diff
@@ -0,0 +1,19 @@
...
+};
+const Checkout = ({ children }: PropTypes) => (
+  <div className="component-checkout">
+    <div className="component-checkout__form">
+      {children}
+    </div>
+  </div>
+);
+
+export default Checkout;
```
walaura: this component makes little sense now but it'll contain way more stuff (it's the checkout layout) down the line so it's got room to grow

PR #1390, opened 2019-01-23
Turn off smallMobileHeaderNotEpicOrBanner test

jranks123:
## Why are you doing this?
The results are in, and the `shrink_no-blurb_no-header` variant is the clear winner. 

We are therefore rolling this out to production.

|landing page before | landing page after |
|-------|------|
|image|image|

|thank you before | thank you page after |
|-------|------|
|image|image|


assets/pages/new-contributions-landing/contributionsLanding.scss
```diff
@@ -127,15 +98,17 @@ header[role="banner"] {
   background-color: gu-colour(header-dark);
   border-bottom: 1px solid gu-colour(garnett-neutral-4);
```
walaura: don't quote me on this but I think the underline for the header should be `header-border` now

assets/pages/new-contributions-landing/contributionsLanding.scss
```diff
@@ -127,15 +98,17 @@ header[role="banner"] {
...
   position: relative;
-  height: 75px;
```
walaura: could this be inferred from the height of the contents (`.glogo`) and the padding instead of explicitly set?

assets/pages/new-contributions-landing/contributionsLanding.scss
```diff
@@ -169,21 +142,25 @@ header[role="banner"] {
...
 .countryGroups {
   font-family: $gu-text-sans-web;
-  margin-top: $gu-v-spacing * 2.5;
   max-width: 355px;
-  margin-left: 10px;
   text-overflow: ellipsis;
   width: 185px;
   white-space: nowrap;
+  margin-left: 0;
+  // TODO: this is basically just vertical centering. use flex-box!
```
walaura: let's use it!!

joelochlann: Since I wrote this comment, I've had a play around with this. What I think you need is:

1. `display: flex` and `align-items: center` on the container (the `<header>` element). This will make sure its children (the logo and the country selector) are vertically centered, which is what we want on mobile.
2. you can then remove any top/bottom margin on the `.glogo` (in fact you have to)
3. from tablet, put `align-self: flex-end` on the `.countryGroups` so it hugs the bottom of the header instead. The logo remains vertically centered
4. and add a `position: absolute;` to `.countryGroups__list` so it can drop down out of the header (I don't quite understand why this wasn't needed right now, but oh well)

I've definitely found it worthwhile getting to grips with flex-box because it makes so many common situations easy without the need for hard-coded specific margins/positions which then are invalidated once the container size changes. The flex-box way means the CSS actually describes what's happening (the children of the header are vertically aligned, then one child is aligned to the bottom), and means this will still be true no matter what the height of the header.

joelochlann: I've done it here: https://github.com/guardian/support-frontend/pull/1390/commits/6b9859f6eaa215abff7a1c3b99d86e968e12ebac

PR #1392, opened 2019-01-24
DO NOT MERGE: Display new digital pack checkout without requiring query param

jacobwinch:
## Why are you doing this?

We're almost ready to drive some real traffic to the new digital pack checkout flow, so we'll need to drop this query param soon.

## Changes

* Drop showCheckout query param
* Drop unused isCsrf query param

## Screenshots
N/A

### DIFF
--- a/app/controllers/DigitalSubscription.scala
+++ b/app/controllers/DigitalSubscription.scala
@@ -67,20 +67,16 @@ class DigitalSubscription(
 
   def digitalGeoRedirect: Action[AnyContent] = geoRedirect("subscribe/digital")
 
-  def displayForm(countryCode: String, displayCheckout: Boolean, isCsrf: Boolean = false): Action[AnyContent] =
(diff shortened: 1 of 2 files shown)


app/controllers/DigitalSubscription.scala
```diff
@@ -67,20 +67,16 @@ class DigitalSubscription(
 
   def digitalGeoRedirect: Action[AnyContent] = geoRedirect("subscribe/digital")
 
-  def displayForm(countryCode: String, displayCheckout: Boolean, isCsrf: Boolean = false): Action[AnyContent] =
+  def displayForm(countryCode: String): Action[AnyContent] =
```
walaura: what was `isCsrf` doing before?

jacobwinch: I'm not sure - when I checked it was unused so I just tidied it up. Perhaps @rupertbates will know why it existed?

PR #1393, opened 2019-01-24
If we collect a telephone number, send it with the request to create a digital pack. 

lmath:
## Why are you doing this?
As of https://github.com/guardian/support-workers/pull/170, we can include a telephone number in the request to support-workers to create a digital pack, and it will get added to the salesforce contact. 

So, if someone in the checkout provides us with the telephone number, we should send it with the request so that it's added to the salesforce contact. 

[**Trello Card**](https://trello.com/c/zcaNKssy/2156-make-sure-telephone-number-if-provided-makes-its-way-to-salesforce)

## Changes

* Use latest version of support-models
* Include telephone number in the request to `/subscribe/digital/create` so that it will be included in the request to support-workers to create a digital pack, and added to the salesforce contact.

## Screenshots
image
⬇️ 
image

🎉 

related PRs:
https://github.com/guardian/support-workers/pull/170
https://github.com/guardian/support-libraries/pull/16

### DIFF
--- a/app/controllers/DigitalSubscription.scala
+++ b/app/controllers/DigitalSubscription.scala
@@ -143,7 +143,7 @@ class DigitalSubscription(
       lastName = request.lastName,
       country = request.country,
       state = request.state,
-      telephoneNumber = request.telephone,
+      telephoneNumber = request.telephoneNumber,
       allowMembershipMail = false,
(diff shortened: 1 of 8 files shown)


LATaylor-guardian:
👍 Nice tests also.

PR #1394, opened 2019-01-24
Remove an unstable test

jacobwinch:
## Why are you doing this?

This test is unstable and has broken the build several times. 

The polling functionality is still being tested here https://github.com/guardian/support-frontend/blob/master/assets/helpers/__tests__/promiseTest.js#L58-L71 (pollUntilPromise calls sleepPromise) so I think this is superfluous anyway.

## Changes

* Remove unstable test

### DIFF
--- a/assets/helpers/__tests__/promiseTest.js
+++ b/assets/helpers/__tests__/promiseTest.js
@@ -4,9 +4,7 @@
 
 import {
   repeatPromise,
-  sleepPromise,
   pollUntilPromise,
-  // logPromise,
   bracketPromise,
 } from '../promise';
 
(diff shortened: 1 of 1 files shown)


jacobwinch:
> do we still want this test around just in case? testing for `expect(t2 - t1).toBeGreaterThanOrEqual(95)` should give us enough precision to make it a worthwhile test.

The [function](https://github.com/guardian/support-frontend/blob/b10d4c09db4ad01653b9551c9a22c2d6df2781a1/assets/helpers/promise.js#L12-L17) under test is:

```
// Runs a promise `i` milliseconds in the future
function sleepPromise<A>(i: number, p: () => Promise<A>): Promise<A> {
  return new Promise((resolve, reject) => {
    setTimeout(() => p().then(resolve, reject), i);
  });
}
```

It felt like we weren't testing much other than setTimeout() with this, so I concluded that removing it altogether was appropriate.

johnduffell:
great, in my world all tests would test a pure function so let's get them one at a time! 👍

PR #1397, opened 2019-01-24
Add tracking to onward links from digital pack checkout page. 

lmath:
Paired with @walaura 

## Why are you doing this?

MORE DATA 

We want to send tracking events to the buttons on digital pack checkout page so that we can understand our users. 

[**Trello Card**](https://trello.com/c/w4xIYuRp/1513-checkout-analytics-phase-1)

## Changes

* Moved returnSection to the scope of the digital pack checkout
* Added an onclick that will send a tracking event to the app store links, return to the guardian link, and FAQ & T&C links in the footer.


lmath:
After playing back the post deploy tests, it seems that on the old flow the paypal window loads up and the test waits find an element/the lack of it. But changes to paypal's sandbox environment aren't necessarily long lasting so I'll leave this for now since this PR hasn't made changes to code that touches paypal.

PR #1413, opened 2019-01-29
Drop PayPal coverage and add annual coverage to post deploy tests

jacobwinch:
## Why are you doing this?

The PayPal tests have been failing more frequently recently, and a bit of time has been wasted investigating these failures. I don't think there is much appetite for investing more time in trying to stabilise these tests, so I'd like to remove them. 

## Changes

* Remove PayPal test code and scenarios
* Test annual with Stripe mock (US), instead of re-testing monthly for new flow (haven't done this with the old flow because it should be going away soon anyway)
* Describe features more clearly (differentiate between old and new flows) for easier debugging


johnduffell:
sounds good as long as we have a suitable alarm to flag up a lack of paypal payments coming thought.  I think we have a No Payments for an hour alarm already which should cover it more or less.  but it would be good to have something more obvious.
👍

joelochlann:
Thanks Jacob. I'm just wondering whether there's any value in keeping them up to `"the PayPal Express Checkout mini-browser should display"`? Do they normally fail in the middle of the PayPal flow? At least that would tell us if we'd broken the PayPal JS in some way

jacobwinch:
> I'm just wondering whether there's any value in keeping them up to `"the PayPal Express Checkout mini-browser should display"`

Good point - this could be a helpful compromise. Recently they have been failing in the middle of the PayPal flow, but it looks like we have had some fun trying to detect the initial page in the past:

```
def initialPageHasLoaded: Boolean = {	
    pageDoesNotHaveElement(id("preloaderSpinner"))	
    pageDoesNotHaveElement(id("spinner"))	
    pageHasUrlOrElement(guestRegistrationUrlFragment, emailInput)	
}
```

I wonder if there is a really simple way we can get confidence that PayPal can be loaded, without caring so much about the specific elements on the PayPal form.

joelochlann:
> I wonder if there is a simple way we can get confidence that PayPal can be loaded, without caring so much about the specific elements on the form (which are liable to change in ways we can't control / don't care about).

If we can just detect that a popup has appeared (regardless of what's in it), that would give me confidence that the PayPal button is working. The logic around blocking the PayPal button when the form is invalid is a bit hairy, and we've had issues in the past where third-party libs haven't loaded properly, so some reassurance that clicking the button actually does something would be the most helpful bit I think!

PR #1414, opened 2019-01-29
Add Stripe Payment Request Button support

jranks123:
## Why are you doing this?
Now that the Stripe contract has been updated to take Google Pay into account, we (almost) can allow the Payment Request Button to be used to take one off payments. Woohoo!

| The Stripe Payment Request Button in all it's glory |
|------------|
|stripe|

### DIFF
--- a/assets/pages/new-contributions-landing/components/StripePaymentRequestButton/StripePaymentRequestButton.jsx
+++ b/assets/pages/new-contributions-landing/components/StripePaymentRequestButton/StripePaymentRequestButton.jsx
@@ -144,6 +144,16 @@ const availablePaymentRequestButtonPaymentMethod: Object => (StripePaymentMethod
   return null;
 };
 
+function setUpPaymentListener(props: PropTypes, paymentRequest: Object, paymentMethod: StripePaymentMethod) {
+  paymentRequest.on('token', ({ complete, token, ...data }) => {
+    // We need to do this so that we can offer marketing permissions on the thank you page
+    updateUserEmail(data, props.updateEmail);
+    const tokenId = props.isTestUser ? 'tok_visa' : token.id;
+    props.onPaymentAuthorised({ paymentMethod: 'Stripe', token: tokenId, stripePaymentMethod: paymentMethod })
+      .then(onComplete(complete));
+  });
+}
+
 function initialisePaymentRequest(props: PropTypes) {
   const paymentRequest = props.stripe.paymentRequest({
     country: props.country,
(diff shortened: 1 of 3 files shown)


assets/pages/new-contributions-landing/components/StripePaymentRequestButton/StripePaymentRequestButton.jsx
```diff
@@ -132,20 +131,29 @@ function onClick(event, props: PropTypes) {
...
-
 // The value of result will either be:
 // . null - browser has no compatible payment method)
 // . {applePay: true} - applePay is available
-// . {applePay: false} - GooglePay or PaymentRequestApi available
-const browserIsCompatible = (result: Object) => {
-  if (weAreSupportingGooglePay) {
-    return result !== null;
+// . {applePay: false} - GooglePay, Microsoft Pay and PaymentRequestApi available
+const availablePaymentRequestButtonPaymentMethod: Object => (StripePaymentMethod | null) = (result: Object) => {
```
joelochlann: don't we normally use the other return type notation?

```javascript
const availablePaymentRequestButtonPaymentMethod = (result: Object): StripePaymentMethod | null =>
...
```

I find this one hard to parse since I'm used to the other one - and it seems you need to notate the `Object` input type twice which I'm not a fan of

jranks123: 👍 sorry about this, not sure what came over me...

assets/pages/new-contributions-landing/components/StripePaymentRequestButton/StripePaymentRequestButton.jsx
```diff
@@ -156,18 +164,13 @@ function initialisePaymentRequest(props: PropTypes) {
...
 
   paymentRequest.canMakePayment().then((result) => {
-    if (browserIsCompatible(result)) {
-      trackComponentClick('apple-pay-loaded');
-      props.setCanMakeStripePaymentRequestPayment(true);
+    const paymentMethod = availablePaymentRequestButtonPaymentMethod(result);
+    if (paymentMethod !== null) {
+      trackComponentClick(`${paymentMethod}-loaded`);
+      props.setPaymentRequestButtonPaymentMethod(paymentMethod);
+      setUpPaymentListener(props, paymentRequest, paymentMethod);
```
joelochlann: I just wonder if to eliminate race conditions it's worth doing `setUpPaymentListener(props, paymentRequest, paymentMethod);` before `props.setPaymentRequestButtonPaymentMethod(paymentMethod);` since I think the latter might be entirely synchronous (i.e. the state update & DOM re-render all happen before the next line)

it's super-paranoid but I'm just wondering if there's a way someone might manage to click the button before the event handler's bound. could be best just to make that impossible

joelochlann: > since I think the latter might be entirely synchronous (i.e. the state update & DOM re-render all happen before the next line)

actually I'm really not sure about that...

joelochlann: I think it might be at the React level that the state changes are decoupled from re-renders: https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous

jranks123: Yeah sure, why not! No harm done

PR #1415, opened 2019-01-29
Route alias for optimize

joelochlann:
@jessewilkins and I were talking about how the results from the Optimize A/A redirect test are very strange.

Optimize redirects to `?test`, which is supposed to be a dummy parameter. Just to eliminate the possibility that something strange is going on with the query params that's affecting the test, I thought we could try and actually different route that just serves the same page. (I've had a quick look and have no idea what could be going on - it's just to eliminate possibilities)


joelochlann:
actually - we're not going to do this

PR #1416, opened 2019-01-30
Add <Dialog> element

walaura:
## Why are you doing this?

We need an accessible (as possible) way to display dialogs that exit the page flow, float over it, and can be dismissed. This component wraps a native `<dialog>` element (which should take care of a11y for us) and polyfills it as much as possible.

There's some barebones styling but the idea is that every dialog client would implement its own styling.

Tested it with voiceover in Chrome, Safari & Firefox, as well as with my own keyboard. A known issue is that the focus trap only works going down, this is for simplicity reasons in this PR, implementing a better focus trap can be done without altering the public interface.

[**Trello Card**](https://trello.com/c/ZvhjtHje/2187-wrap-dialog-for-nav-menus-mobile-nav)

## Screenshots

screenshot 2019-01-30 at 12 30 36 pm

screenshot 2019-01-30 at 12 30 34 pm


rupertbates:
I'm anticipating what Jamie would say if he were here: "Is it not possible to achieve the same thing using stateless components?"

walaura:
@rupertbates I can't think of a way :( FWIW [react's own section on a11y](https://reactjs.org/docs/accessibility.html) asks you to use props and state, notably being the 1 part of the react site that doesn't tell you to _not_ use them

stories/dialog.jsx
```diff
@@ -0,0 +1,71 @@
...
+        with the page under it while it is open
+      </p>
+      <p>You probably do not actually want this,
+        as this dialog does not have the click outside behaviour
+        and makes for a confusing experience for screen readers
+      </p>
+      <ControlledDialogButton modal={false} />
+    </ProductPageTextBlock>
+  </div>
+));
```
SiAdcock: Why make this an option available through props? Can you think of a use case for it?

assets/components/dialog/dialog.jsx
```diff
@@ -0,0 +1,115 @@
...
+  render() {
+    const {
+      open, modal, children, onStatusChange, dismissOnBackgroundClick, ...otherProps
+    } = this.props;
+
+    return (
+      <dialog // eslint-disable-line jsx-a11y/no-redundant-roles
+        className={classNameWithModifiers('component-dialog', [modal ? 'modal' : null, open ? 'open' : null])}
+        aria-modal={modal}
+        aria-hidden={!open}
```
SiAdcock: Should we also include the `open` attribute?

assets/components/dialog/dialog.jsx
```diff
@@ -0,0 +1,115 @@
...
+        tabIndex="-1"
+        role="dialog"
```
SiAdcock: I'm guessing this is needed for browsers that don't support the `<dialog>` element?

assets/components/dialog/dialog.jsx
```diff
@@ -0,0 +1,115 @@
...
+      if (this.ref) {
+        this.ref.focus();
+      }
+    });
+  }
+
+  close() {
+    if (this.ref && this.ref.close) {
+      this.ref.close();
+    }
```
SiAdcock: What happens to the focus after the modal is dismissed? Ideally it would return back to the button that opened it, or perhaps returned to the beginning of the page

assets/components/dialog/dialog.jsx
walaura: yes :(

assets/components/dialog/dialog.jsx
walaura: heya! it gets opened using the [js callbacks](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog) in the lifecycle hooks a bit higher up the component. You can't actually use both. if you have `open="true"` and try to fire up `dialog.show()` that's a javascript warning, makes for a lovely imperative vs declarative debate!

(Also, `open="true"` doesn't support modals at all 😝. I think the usage scenario is for serving markup from a server. Even the MDN demos don't set it via js)

As all `<dialog>` elements support `dialog.show()` I chose to omit the open attr altogether and instead style the polyfill with traditional css

assets/components/dialog/dialog.jsx
```diff
@@ -0,0 +1,115 @@
...
+    const {
+      open, modal, children, onStatusChange, dismissOnBackgroundClick, ...otherProps
+    } = this.props;
+
+    return (
+      <dialog // eslint-disable-line jsx-a11y/no-redundant-roles
+        className={classNameWithModifiers('component-dialog', [modal ? 'modal' : null, open ? 'open' : null])}
+        aria-modal={modal}
+        aria-hidden={!open}
+        tabIndex="-1"
```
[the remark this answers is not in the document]
walaura: So, since browser support for `dialog` is fiddly at best I have decided to implement focus management cheaply by just focusing the whole dialog after it opens – this seems to net sensible results in all tested browsers.

The native behaviour would be to focus on the first interactive element – however that requires more DOM fiddling, i'm not against exploring that solution but I'm not confident it's as portable as focusing the whole thing :(

I found healthy debate [here](https://github.com/whatwg/html/issues/1929) about this all, my takeaways are:
- The guidance refers more about a numberical tabindex than the magical `-1` and `0` (because `dialog`s live outside of the element to a level)
- Some user agents make `dialog` focusable? (thinking about using `tabindex=0`
- Focusing the dialog itself vs focusing the first interactive element is an open discussion?

assets/components/dialog/dialog.jsx
walaura: alas it depends on the implementer. the dialog does fire an event when it closes, so the component that fired it should be able to return the focus. In my testing it would seem that chrome makes a decent job at doing this on its own.

Something we could do could be to tweak the API so there's a nullable but required `returnFocusAfterClose()` prop to make it explicit that if you implement a dialog you have to take care of that? But I want to test default browser behaviour a bit more in case that just adds extra work

walaura:
@SiAdcock not at all! you raise many super valid points and have given me food for thought, will tag you in a subsequent PR

assets/components/dialog/dialog.jsx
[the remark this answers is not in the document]
SiAdcock: Awesome, thanks for sharing the discussion, really interesting! 

The guidance we received from the accessibility consultant:

- if there's an obvious candidate to receive focus in the dialog, use that, OTHERWISE
- provide focusable empty divs at the top and bottom of the dialog and use JavaScript to return focus to the top div when tabbing out of the bottom div (and vice versa). 

I can link you the notes he provided if you need it. But there's more than one way to peel a potato, and as long as the approach is tested, it's all good! 👍

assets/components/dialog/dialog.jsx
SiAdcock: Interesting. And so confusing! I'm sure it's contrary to what is outlined in [the spec](https://html.spec.whatwg.org/multipage//interactive-elements.html#the-dialog-element). Browsers are a law unto themselves 🙄

assets/components/dialog/dialog.jsx
SiAdcock: Cool, sounds like it would work!

As an aside, it's worth testing the following combos of [browsers and screenreaders](https://webaim.org/projects/screenreadersurvey7/#browsercombos), as these are the most commonly used:

- IE11 + JAWS
- Firefox + NVDA
- Firefox + JAWS
- Safari + Voiceover

assets/components/dialog/dialog.jsx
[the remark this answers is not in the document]
walaura: we do have a naively implemented oneline focus trap at the end but not one at the top 😢 I'm not sure how to implement both sides without a lot of code. it's something i want to look into though!

PR #1419, opened 2019-01-30
Render placeholder html to improve PF apparently speed

johnduffell:
## Why are you doing this?

We are concerned about the performance of the PF screen as it's all rendered in JS and we only see the form when the JS has run.
We have decided to do some tests to see where we should focus.  The first test is getting a placeholder on the screen asap so the reader knows what to expect.  We simply cover the placeholder screen with a loading blobs.

We are hoping that the test is good enough that we will be certain whether this is where it's worth spending effort.  We will do a separate test for whether speeding up the actual render of the working page will be fruitful too.  This might be a little more in depth but hopefully more valuable hence why it's worth getting this test our separately.

This change means our time to first paint is 0.9s instead of 2.9s.  The remaining times are pretty much the same even if slightly slower in places.

We are going to improve the placeholder copy before merging!

Perf Before:
image

Perf After:
image

Loading screen (this one is for US, it would have previously been white and blank)
image

[**Trello Card**](https://trello.com/c/rVHEliy6/281-npf-on-desktop-and-mobile-speed-performance-improvements)

## Changes

The optimise hiding has been removed on both variants, this can be put back in if necessary in the control?

@jranks123 @ionamckendrick @jranks123 @tomrf1


app/controllers/Application.scala
```diff
@@ -99,16 +99,44 @@ class Application(
...
+  private def contributionsHtml(countryCode: String, idUser: Option[IdUser], isTest: Boolean)(implicit request: RequestHeader, settings: AllSettings) = {
+    if (isTest) {
+      val annualAmounts: Map[String, (String, String, String, String)] = Map(
+        "uk" -> ("£50", "£100", "£250", "£500"),
+        "us" -> ("$50", "$100", "$250", "$500"),
+        "au" -> ("$100", "$250", "$500", "$750"),
+        "eu" -> ("€50", "€100", "€250", "€500"),
+        "int" -> ("£50", "£100", "£250", "£500"),
+        "nz" -> ("$50", "$100", "$250", "$500"),
+        "ca" -> ("$50", "$100", "$250", "$500")
```
johnduffell: we have all the amounts correctly inserted but the second is selected by default in the placeholder.  in a AU/NZ the first one is selected by default in the real rendered copy, so the reader will see the selection end up different to the placeholder.  I don't think this is an issue but I wanted to flag it just in case.

app/views/newContributionsTest.scala.html
```diff
@@ -0,0 +1,71 @@
+@import assets.AssetsResolver
```
johnduffell: this is just a copy of the non "Test" one with an extra parameter and calling test.scala.html instead of main.scala.html.

walaura:
> When the test finishes we will remove this hack completely and reimplement in a more robust way, as the current way is a horrible hack!

👋 hiii i'm keen to try doing some server side rendering in here!!!! if this test is successful it feels like a perfect excuse ^^

a question: how does google optimize affect this test? AFAIK it also hides the page until js loads, that might interfere with this test?

app/controllers/Application.scala
johnduffell: if we want to do amounts testing we would have to do something smart here.  Since the amounts are hard coded, we don't want to give a worse experience to the people in both tests.  We could spend a bit of time once the test is running to implement a rudimentary server side ab testing allocation to enable the amounts test.

assets/helpers/abTests/abtest.js
```diff
@@ -123,6 +123,27 @@ function getParticipationsFromUrl(): ?Participations {
   return null;
 }
 
+function getParticipationsFromQuery(): ?Participations {
+
+  const hashUrl = (new URL(document.URL)).search;
+  const index = hashUrl.indexOf('ssr');
```
johnduffell: not sure if I should be more careful in case some of the other tracking stuff happens to have 'ssr' in the string.

assets/helpers/abTests/abtestDefinitions.js
```diff
@@ -19,4 +18,17 @@ export const tests: Tests = {
     seed: 3,
   },
 
+  ssr: {
+    variants: ['off', 'on'],
+    audiences: {
+      ALL: {
+        offset: 0,
+        size: 0,
```
johnduffell: this is switched on but noone will be automatically allocated, it will all happen via the url parameters

app/views/test.scala.html
```diff
@@ -0,0 +1,2347 @@
...
+      <link rel="alternate" href="@link" hreflang="@lang" />
+    }
+
+    @settingsScript(settings)
+
+    <title>@title</title>
+    <meta property="og:title" content="@title"/>
+
+    @optimizeScript()
+    <style>
```
joelochlann: could we inline the minified CSS which is delivered to the browser in prod instead? this will be larger than what gets sent as a separate file normally, which seems a wasted opportunity

PR #1421, opened 2019-01-31
Dialog tweaks

walaura:
## Why are you doing this?

Implement some of @SiAdcock's suggestions (the easy high api layer impact ones) in the dialog component #1416  

## Changes

* Don't handle browser focus if it supports `<dialog>`
* Remove `modal` prop
* Add new `styled` prop that adds a modal-like appearance (this was previously tangled together with modal-ness. Most appearances will be custom but this makes it more satisfying to implement. (instant dialogs!)
* New default is for dialogs to not dismiss themselves when clicking out – to make implementers think about adding a close button (needed for keyboard users). the prop for this is now called `blocking` for simplicity
* Got carried away writing stories for this!

## Screenshots

screenshot 2019-01-31 at 10 17 14 am
screenshot 2019-01-31 at 10 17 09 am

### DIFF
--- a/assets/components/button/_sharedButton.jsx
+++ b/assets/components/button/_sharedButton.jsx
@@ -30,6 +30,7 @@ type SharedButtonPropTypes = {|
   icon?: Node,
   appearance: Appearance,
   iconSide: IconSide,
+  getRef?: (?Element) => void,
   modifierClasses: string[],
 |};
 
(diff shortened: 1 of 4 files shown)


SiAdcock:
> New default is for dialogs to not dismiss themselves when clicking out – to make implementers think about adding a close button (needed for keyboard users)

Does the functionality exist to dismiss using the `Escape` key? I think keyboard users would expect it. A close button is better though!

walaura:
> > New default is for dialogs to not dismiss themselves when clicking out – to make implementers think about adding a close button (needed for keyboard users)
> 
> Does the functionality exist to dismiss using the `Escape` key? I think keyboard users would expect it. A close button is better though!

ah! good point. it's in for `<dialog>` not in the polyfill. Will add

PR #1434, opened 2019-02-01
use grey boxes instead of the actual amounts in placeholders

johnduffell:
## Why are you doing this?

Since we are going to test rendering a placeholder before the payment flow appears in https://github.com/guardian/support-frontend/pull/1419
We are rendering the amounts early on so people have something to look at.  However we don't think that will be important and it also makes it a little more difficult to run amounts test (we would need server side test allocation)

This PR changes it to be a grey box which is replaced by the amounts in due course.  I haven't used any paricular grey and it has a 0.6 opacity black cover over it anyway, but this can be tweaked easily.

image

I haven't done the button at the bottom because it's probably outside of where the user will be paying attention anyway, but it would be possible to add that if necessary.

@ionamckendrick @jranks123

### DIFF
--- a/app/views/test.scala.html
+++ b/app/views/test.scala.html
@@ -13,8 +13,7 @@
   styles: Html = Html(""),
   scripts: Html = Html(""),
   data: Map[String, String] = Map.empty,
-  csrf: Option[String] = None,
-  annualAmounts: (String, String, String, String)
+  csrf: Option[String] = None
 )(implicit assets: AssetsResolver, request: RequestHeader, settings: AllSettings)
 
 <!DOCTYPE html>
@@ -1854,6 +1853,12 @@
         cursor: pointer
       }
 
+      .amounts__placeholder {
+        width: 30px;
+        height: 18px;
+        background-color: #eee;
+      }
+
       .form__radio-group--tabs .form__radio-group-list {
         display: -webkit-flex;
         display: flex;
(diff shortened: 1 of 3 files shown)


walaura:
not a designer but for the circles perhaps leaving them empty would work better?

ionamckendrick:
> not a designer but for the circles perhaps leaving them empty would work better?

@walaura Don't worry, I've made changes with John. I wanted fully filled circles, which he's now adding!

johnduffell:
yep thanks for all the suggestions, I have pushed the final screenshots, have a look in the edits of the description!  Let me know if any more tweaks needed.

app/views/test.scala.html
```diff
@@ -2223,12 +2226,38 @@
...
+          Amount</legend><ul class="form__radio-group-list">
+          <li class="form__radio-group-item">
+            <span class="amounts__placeholder">
+            <input id="contributionAmount-50" class="form__radio-group-input" type="radio" name="contributionAmount" value="50">
+            <label for="contributionAmount-50" class="form__radio-group-label" aria-label="50 pounds">
+            </label></span>
+          </li>
+          <li class="form__radio-group-item">
+            <span class="amounts__placeholder">
+            <input id="contributionAmount-100" class="form__radio-group-input" type="radio" name="contributionAmount" value="100" checked>
```
joelochlann: could we delete the inputs and `for` and `aria-label` attributes on the `<label>`? just to express the intentions more clearly - it took me a minute to realise that all of the numeric amounts were meaningless here

ionamckendrick:
> yep thanks for all the suggestions, I have pushed the final screenshots, have a look in the edits of the description! Let me know if any more tweaks needed.

@johnduffell Looking lots better! Two small tweaks from me: 
1) can we remove the arrow from the main CTA (so it's simply a plain yellow lozenge)
2) can we remove the yellow outline on the buttons (it's cleaner if they're all pale grey)

Thanks!

app/views/test.scala.html
walaura: for a11y reasons I'd put yhis entire preload body inside `aria-hidden` so that screen readers don't start reading the skeleton!

johnduffell: good idea, didn't think about that!  maybe we should test with a screen reader!

johnduffell: I left the inputs just for styling simplicity.  It started as a copy and paste of the html and css, but then we started tweaking the appearance a bit thus messing with the css and dom a bit, but I'm loathe to start refactoring too much as we are literally just doing it for the test result and reverting back at the end before we do next steps.  I definitely think we will implement it according to best practices if it produces a positive result.

The main thing is avoiding any kind of broken appearance to end users as that would affect the test result (e.g. confusing readers with screen readers or making the screen flash etc)

johnduffell:
I have updated the screen shot and pushed the latest changes! @ionamckendrick

PR #1442, opened 2019-02-01
Allow the app to access the catalog.json in S3

rupertbates:
## Why are you doing this?

The app needs permission to access the catalog.json file in S3. This changes the cloud formation template to allow it.

### DIFF
--- a/cloud-formation/cfn.yaml
+++ b/cloud-formation/cfn.yaml
@@ -101,7 +101,10 @@ Resources:
             Statement:
               - Effect: Allow
                 Action: s3:GetObject
-                Resource: !Sub arn:aws:s3:::membership-private/${Stage}/support-frontend.private.conf
+                Resource:
+                - !Sub arn:aws:s3:::membership-private/${Stage}/support-frontend.private.conf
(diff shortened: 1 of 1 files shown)


cloud-formation/cfn.yaml
```diff
@@ -101,7 +101,9 @@ Resources:
             Statement:
               - Effect: Allow
                 Action: s3:GetObject
-                Resource: !Sub arn:aws:s3:::membership-private/${Stage}/support-frontend.private.conf
+                Resource:
+                - !Sub arn:aws:s3:::membership-private/${Stage}/support-frontend.private.conf
+                - !Sub arn:aws:s3:::gu-zuora-catalog/PROD/Zuora-${Stage}/catalog.json
```
LATaylor-guardian: Should `PROD` be ${stage} here also?

rupertbates: No we need PROD for all environments there weirdly.

johnduffell: it's not that weird, actually the first PROD says it's the output from the PROD catalog lambda, and that lambda is set up to fetch from all environments.  The second one is the PROD/UAT/DEV backend for zuora.

PR #1447, opened 2019-02-01
Ensure currency/price info displayed to users is consistent with attempted charge

jacobwinch:
## Why are you doing this?

Prior to https://github.com/guardian/support-frontend/pull/1423 it was possible to display one currency on the checkout form but charge users in a different currency (e.g. showing users USD pricing and then attempting to charge them in GBP). This PR fixes that problem.

## What caused this bug?

If a user's identity account has a country stored, we were [pre-selecting](https://github.com/guardian/support-frontend/pull/1312) the country field on the checkout form with it. This turns out to be problematic because the common state which is used to determine the currency/price (i.e. the state that we pass to Stripe) was only being updated on country field update (not on initialisation). Consequently if a user's country in Identity = US and they arrived at /**uk**/subscribe/digital/checkout, we would show them USD pricing _until_ they reached the Stripe popup, when the pricing flips to GBP before being passed to the server (see screenshot). 

## What else is going on in this PR?

As this type of bug is easy to introduce when country is being replicated in two parts of the state, this PR removes country from checkout state entirely. We now rely on [this function](https://github.com/guardian/support-frontend/blob/master/assets/helpers/internationalisation/country.js#L527-L550) to calculate country (in common state), and wire this through. This allows us to work out which country (and payment method) should be pre-selected on the Digital Pack checkout. 

We also now explicitly call setCountry (from commonActions) rather than relying on having two actions with the same name (see description on https://github.com/guardian/support-frontend/pull/1386).

[**Trello Card 1**](https://trello.com/c/GMQmzcBp/2196-investigate-bug-which-allows-users-to-create-digital-pack-subs-with-invalid-country-currency-combinations)
[**Trello Card 2**](https://trello.com/c/5UDL8f1G/2183-bug-setcountry-is-firing-two-different-redux-actions-on-the-checkout-form)

## Changes

* Remove concept of user country / identity country
* Make use of https://github.com/guardian/support-frontend/pull/1457 to wire through common state so this can be used when determining the initial state for the checkout flow
* Add a couple of unit tests

## Screenshots

The problem:

image


assets/pages/digital-subscription-checkout/digitalSubscriptionCheckout.jsx
```diff
@@ -18,12 +19,12 @@ import SubscriptionFaq from 'components/subscriptionFaq/subscriptionFaq';
 
 import { initReducer } from './digitalSubscriptionCheckoutReducer';
 import CheckoutStage from './components/checkoutStage';
-
 import './digitalSubscriptionCheckout.scss';
 
 // ----- Internationalisation ----- //
 
-const countryGroupId: CountryGroupId = detect();
+const countryGroupId: CountryGroupId = detectCountryGroup();
```
jacobwinch: I don't really want to call these functions here. I'd prefer to use the common state in digitalSubscriptionCheckoutReducer, but I am struggling to work out how to wire this up...

assets/pages/digital-subscription-checkout/digitalSubscriptionCheckout.jsx
```diff
@@ -36,7 +33,10 @@ const reactElementId: {
 
 // ----- Redux Store ----- //
 
-const store = pageInit(commonState => initReducer(commonState.internationalisation.countryGroupId), true);
+const store = pageInit(commonState =>
+  initReducer(commonState.internationalisation.countryId, commonState.internationalisation.countryGroupId), true);
+
+const { countryGroupId } = store.getState().common.internationalisation;
```
walaura: what do we use `countryGroupId ` in here for?

assets/pages/digital-subscription-checkout/helpers/paymentProviders.js
```diff
@@ -41,7 +40,7 @@ function buildRegularPaymentRequest(state: State, paymentAuthorisation: PaymentA
   return {
     firstName,
     lastName,
-    country: country || 'GB',
+    country: countryId || 'GB',
```
walaura: do we need ` || 'GB'` in here anymore?

assets/pages/digital-subscription-checkout/helpers/paymentProviders.js
```diff
@@ -103,11 +102,15 @@ function showPaymentMethod(
     case 'DirectDebit':
       dispatch(openDirectDebitPopUp());
       break;
+    case null:
+    case undefined:
+      console.log('Undefined payment method');
+      break;
     default:
       console.log(`Unknown payment method ${paymentMethod}`);
```
walaura: not for this PR but as these are runtime errors we might wanna send them to sentry

assets/pages/digital-subscription-checkout/helpers/paymentProviders.js
jacobwinch: Nope, good spot

assets/pages/digital-subscription-checkout/digitalSubscriptionCheckoutReducer.js
```diff
@@ -68,17 +75,17 @@ export type State = ReduxState<{|
...
-  | { type: 'SET_LAST_NAME', lastName: string }
-  | { type: 'SET_TELEPHONE', telephone: string }
-  | { type: 'SET_COUNTRY', country: string }
-  | { type: 'SET_STATE_PROVINCE', stateProvince: string }
-  | { type: 'SET_BILLING_PERIOD', billingPeriod: DigitalBillingPeriod }
-  | { type: 'SET_PAYMENT_METHOD', paymentMethod: PaymentMethod }
-  | { type: 'SET_FORM_ERRORS', errors: FormError<FormField>[] }
-  | { type: 'SET_SUBMISSION_ERROR', error: ErrorReason }
-  | { type: 'SET_FORM_SUBMITTED', formSubmitted: boolean }
+  | { type: 'SET_DP_CHECKOUT_STAGE', stage: Stage }
```
rupertbates: These could all have a separate `scope` parameter on them rather than the _DP_ infix. We do something similar here: https://github.com/guardian/support-frontend/blob/ab5d62c631c91e6e92c13884025d7963d638dfc3/assets/components/marketingConsent/marketingConsentActions.js#L16

assets/pages/digital-subscription-checkout/digitalSubscriptionCheckoutReducer.js
```diff
@@ -163,11 +189,11 @@ export type FormActionCreators = typeof formActionCreators;
 
 // ----- Reducer ----- //
 
-function initReducer(countryGroupId: CountryGroupId) {
-  const billingPeriodInurl = getQueryParameter('period');
-  const user = getUser(countryGroupId); // TODO: this is unnecessary, it should use the user reducer
-  const initialBillingPeriod: DigitalBillingPeriod = billingPeriodInurl === 'Monthly' || billingPeriodInurl === 'Annual'
-    ? billingPeriodInurl
+function initReducer(initialCountry: IsoCountry, initialCountryGroup: CountryGroupId) {
```
rupertbates: We can derive the CountryGroupId from an IsoCountry so do we need both here?

assets/pages/digital-subscription-checkout/digitalSubscriptionCheckoutReducer.js
walaura: Do we want to do this before we know for sure if the checkout infrastructure will be unified at that level? This reducer is very DP specific at the moment and it might be a better start for print if we just copypaste it as a start rather than abstract the shared logic?

rupertbates: Even if we don't it feels nicer than sticking DP in the middle of everything plus it's easier to copy and paste because you just change the scope rather than all of the names.

walaura: i suppose my concern is that we would still be "firing" 2+ reducers with the same action and trusting the reducer itself to not do anything if it's out of scope, vs implicitly marking it in the action.

I don't feel strongly about this though, and even while writing this im coming around to think it's a better future bet to use `scope` in any case

jacobwinch: I agree that adding a scope param will be a better solution for this. However, I don't want to mix that up with this PR because it's already been sitting around too long and I would like to get the bug fixed.

I'll revisit this in a separate PR, so have restored the old action names for now.

PR #1451, opened 2019-02-01
Fix annual promotion summary and add tests

rupertbates:
## Why are you doing this?

There were some small bugs with the way that the `PromotionSummary` component handled various promotions. This fixes them and adds tests.

[**Trello Card**](https://trello.com/c/YDhuWZgm/2194-add-jest-test-for-promotion-summaries)


walaura:
can't review today! but we have this `Price` objct to display prices+glyphs that might come in handy

PR #1453, opened 2019-02-01
Update year in LICENSE

walaura:
## Why are you doing this?

I dont even know man its super late on a Friday and i got stood up for a show. there is that what you wanted to know? because that's what happened. now i'm here sleep deprived ot wanting to go out and face the snow waiting for Giraffe to go to the pub

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright 2017 Guardian News & Media Ltd
+Copyright 2019 Guardian News & Media Ltd
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -1,4 +1,4 @@
-Copyright 2017 Guardian News & Media Ltd
+Copyright 2019 Guardian News & Media Ltd
```
johnduffell:
```suggestion
Copyright 2017-2019 Guardian News & Media Ltd
```
technically I think we should cover all the years, or at least that's what I've had to do at more highly regulated places than here

johnduffell:
😂 👍 🕥

PR #1455, opened 2019-02-04
Show countdown timer for last 7 days on the DP product page during a flash sale. 

lmath:
## Why are you doing this?

doesn't look terribly urgent

[**Trello Card**](https://trello.com/c/NyBumRDL/2134-countdown-timer-should-only-appear-during-the-last-week-of-the-print-sale)

## Changes

* Adds a function that determines if the now is x days from the end of the sale so that we can use it to determine if we should show the countdown timer
* Use this new function to only show the countdown timer 7 days before the end of the digital pack flash sale

## Screenshots
This is just changing the conditions by which we show the countdown timer or not, so n/a I think

### DIFF
--- a/assets/helpers/flashSale.js
+++ b/assets/helpers/flashSale.js
@@ -419,7 +419,7 @@ function getFormattedFlashSalePrice(
   return fixDecimals(sale.saleDetails[countryGroupId].price);
 }
 
-function countdownTimerActive(flashSaleActive: boolean, showForHowManyDays: number, endTime: number): boolean {
+function countdownTimerIsActive(flashSaleActive: boolean, showForHowManyDays: number, endTime: number): boolean {
   if (flashSaleActive) {
     const now = new Date();
     const daysFromNow = now.setDate(now.getDate() + showForHowManyDays);
@@ -443,5 +443,5 @@ export {
   getDuration,
   getTimeTravelDaysOverride,
   getFlashSaleActiveOverride,
-  countdownTimerActive,
+  countdownTimerIsActive,
 };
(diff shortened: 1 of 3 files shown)


SarahLaughton:
👍 Love it!

assets/helpers/flashSale.js
```diff
@@ -419,6 +419,16 @@ function getFormattedFlashSalePrice(
   return fixDecimals(sale.saleDetails[countryGroupId].price);
 }
 
+function showCountdownTimer(flashSaleActive: boolean, showForHowManyDays: number, endTime: number): boolean {
```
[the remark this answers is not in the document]
lmath: Good point. I will rename it.

assets/helpers/flashSale.js
```diff
@@ -419,6 +419,16 @@ function getFormattedFlashSalePrice(
   return fixDecimals(sale.saleDetails[countryGroupId].price);
 }
 
+function countdownTimerActive(flashSaleActive: boolean, showForHowManyDays: number, endTime: number): boolean {
```
walaura: can we call this `countdownTimerIsActive`  (super pedantic i know)

assets/helpers/flashSale.js
```diff
@@ -419,6 +419,16 @@ function getFormattedFlashSalePrice(
...
+  if (flashSaleActive) {
+    const now = new Date();
```
walaura:
```suggestion
    const timeTravelDays = getTimeTravelDaysOverride();
    const now = timeTravelDays ? Date.now() + (timeTravelDays * 86400000) : Date.now();
```

this'll allow the time travel parameter to work

PR #1456, opened 2019-02-04
Split SSR test audience at backend

jranks123:
## Why are you doing this?
We were splitting the audience for the test on `gu.com`, which in hindsight was unnecessary: we can split it here and get the entire audience into the test. 

There is a PR to stop adding the`ssr` query string parameter on links from gu.com, although once this PR is merged, we will overrride the value of ssr anyway


app/utils/QueryStringUtils.scala
```diff
@@ -0,0 +1,8 @@
+package utils
+
+object QueryStringUtils {
+  def addServerSideRenderingTestParameterQueryString(queryString: Map[String, Seq[String]]): Map[String, Seq[String]] = {
+    val ssrParamterValue = if (math.random() < 0.5) "on" else "off"
```
joelochlann: You could use `Random.nextBoolean` from `scala.util.Random` for this

joelochlann: typo: `ssrParameterValue`

PR #1460, opened 2019-02-04
Track header links

walaura:
😲👆 the header is still hidden in all pages – this pr has no visible impact

## Why are you doing this?

To test the effectiveness of the navigation

- Tracks links from normal / mobile header plus mobile menu opens/closes in this format:

- `header-link - subscriptions:paper - desktop`
- `header-link - subscriptions:paper - mobile`
- `header - menu-open`

Missing the patrons campaign stuff!

[**Trello Card**](https://trello.com/c/o122CMPr/2163-subs-navigation-analytics)


assets/components/headers/header/header.scss
```diff
@@ -46,7 +46,7 @@
       margin-top: 40px;
     }
   }
-  .component-dialog--open .component-mobile-header {
+  .component-dialog--open .component-header-mobile-menu {
```
jesseyuen: Consider making the 0.2s a sass variable? (plus other speeds if we have them?)

PR #1461, opened 2019-02-04
Optimize tracking fixes

rupertbates:
## Why are you doing this?

We don't currently track Optimize tests properly on pages which don't use Redux, this PR fixes this.

[**Trello Card**](https://trello.com/c/4kEtf6ia/2181-ab-test-tracking)


rupertbates:
> LGTM 🚀
> 
> do we want to add redux to all pages? the new header will need it anyway if we do use redux state for the dropdowns

Yes maybe, it would make things less confusing I think. Not for this PR though.

PR #1462, opened 2019-02-05
fix ssr tracking

jranks123:
## Why are you doing this? 
Previously, we were just saving the participation of the SSR test to the storage, but in nearly all cases, not actually tracking people through GA or the datalake (the only cases where we would have tracked it is when people came back to the page after visiting once).

This PR fixes this bug.

### DIFF
--- a/assets/helpers/abTests/abtest.js
+++ b/assets/helpers/abTests/abtest.js
@@ -123,7 +123,7 @@ function getParticipationsFromUrl(): ?Participations {
   return null;
 }
 
-function getParticipationsFromQuery(): ?Participations {
+function getSSRParticipationsFromQuery(): ?Participations {
 
   const hashUrl = (new URL(document.URL)).search;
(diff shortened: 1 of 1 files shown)


app/utils/QueryStringUtils.scala
```diff
@@ -1,10 +1,9 @@
 package utils
-
 import scala.util.Random
 
 object QueryStringUtils {
   def addServerSideRenderingTestParameterQueryString(queryString: Map[String, Seq[String]]): Map[String, Seq[String]] = {
     val ssrParameterValue = if (Random.nextBoolean()) "on" else "off"
-    queryString ++ List("ssr" -> Seq(ssrParameterValue))
+    queryString ++ List("ssrTwo" -> Seq(ssrParameterValue))
```
joelochlann: you could write this as

```scala
queryString + ("ssrTwo" -> Seq(ssrParameterValue))
```

PR #1463, opened 2019-02-05
Enable amounts tests

tomrf1:
## Why are you doing this?
Amounts are configured using an s3 file (soon to be managed from Support Admin Console).
We also need the ability to run amounts tests which override the configured defaults.
For now, test amounts will be configured in the code (client-side).

Note - we are running a new annual amounts test this week, but have not decided on the exact config yet. I'll update assets/helpers/abTests/annualAmountsTest.js when this has been decided.

[**Trello Card**](https://trello.com/c/dWLgtddF/339-enable-amounts-tests-to-override-sac-configured-amounts)

### DIFF
--- a/assets/helpers/abTests/abtestDefinitions.js
+++ b/assets/helpers/abTests/abtestDefinitions.js
@@ -1,8 +1,9 @@
 // @flow
 import type { Tests } from './abtest';
-import { annualAmountsA } from './annualAmountsTest';
+import { annualAmountsLower, annualAmountsFive, annualAmountsOther } from './annualAmountsTest';
 
 // ----- Tests ----- //
+
 
 export const tests: Tests = {
   ssrTwo: {
@@ -26,8 +27,16 @@ export const tests: Tests = {
         id: 'control',
       },
       {
-        id: 'annualAmountsA',
-        amountsRegions: annualAmountsA,
+        id: 'lower',
+        amountsRegions: annualAmountsLower,
+      },
+      {
+        id: 'five',
+        amountsRegions: annualAmountsFive,
+      },
+      {
+        id: 'other',
+        amountsRegions: annualAmountsOther,
       },
     ],
     audiences: {
--- a/assets/helpers/abTests/abtestDefinitions.js
+++ b/assets/helpers/abTests/abtestDefinitions.js
@@ -37,7 +37,7 @@ export const tests: Tests = {
         size: 1,
       },
     },
-    isActive: true,
+    isActive: false,
     independent: true,
     seed: 3,
   },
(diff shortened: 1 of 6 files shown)


assets/helpers/abTests/abtestDefinitions.js
```diff
@@ -17,4 +19,27 @@ export const tests: Tests = {
     seed: 4,
   },
 
+  annualContributionsRoundFour: {
+    type: 'AMOUNTS',
+    variants: [
+      {
+        id: 'control',
+        amountsRegions: annualAmountsControl,
```
joelochlann: I feel like it should be possible to just omit this param if nothing's getting overriden?

tomrf1: good point

assets/helpers/abTests/helpers.js
```diff
@@ -0,0 +1,49 @@
...
+    const { amountsRegions } = variant;
+
+    Object.keys(amountsRegions).forEach((countryGroupId) => {
+
+      Object.keys(amountsRegions[countryGroupId]).forEach((contributionType) => {
+
+        newAmountsRegions[countryGroupId][contributionType] = amountsRegions[countryGroupId][contributionType];
+      });
+    });
+  }
```
joelochlann: maybe we could try this in future? https://www.npmjs.com/package/deepmerge

especially since we'll be doing something similar for Contribution Types and maybe other parameters

joelochlann: in fact, I wonder if a more general solution would be to let a test override the entire `settings` object? (not necessarily for this PR, but as we extend what can be overridden by a test)

open to opinions on this. @rupertbates any thoughts?

assets/helpers/page/page.js
```diff
@@ -74,13 +75,16 @@ function buildInitialState(
     currencyId,
   };
 
+  // Override the default amounts config with any test participations
+  const amountsWithParticipations = overrideAmountsForParticipations(abParticipations, settings.amounts);
```
joelochlann: I find this variable name a bit confusing, maybe `amountsWithParticipationOverrides`?

assets/helpers/abTests/annualAmountsTest.js
```diff
@@ -0,0 +1,129 @@
+// @flow
+
+import { type AmountsRegions } from 'helpers/contributions';
+
+export const annualAmountsControl: AmountsRegions = {
+  // Use the config from server
+};
+
+export const annualAmountsA: AmountsRegions = {
```
joelochlann: if the number of actually overridden amounts is less unwieldy, I'd be up for putting this straight in `abtestDefinitions.js` so it's all in one place (just a single file to change when removing).

pending getting the actual amounts!

assets/helpers/abTests/helpers.js
tomrf1: `deepmerge` looks good.

And good point - this could be very generic. Each variant could instead define a partial `Settings` object, which is used to 'patch' the current `settings`.

tomrf1:
I've updated this branch with the actual amounts tests, so this is now ready to go

PR #1469, opened 2019-02-06
Add animated dots to Story Book and make sure that the dark type is used in the MustSignIn component

jranks123:
## Why are you doing this?
Somewhere along the line, our animated dots underneath the email field (which appear when we are checking to see if the email address has an account already) became illegible (as they are currently white on a near-white background. This PR fixes this 

image

Also, adds animated dots to storyboard


assets/pages/new-contributions-landing/contributionsLanding.scss
```diff
@@ -613,6 +613,10 @@ form {
   a, a:visited {
     color: gu-colour(news-garnett-main-1);
   }
+
+  .component-animated-dots div {
```
joelochlann: any way we can avoid the nested selector?

walaura: could `component-animated-dots` take an `appearance={light|dark}` parameter?

walaura: and add it to storybook? 
face-with-pleading-eyes_1f97a

johnduffell: actually we use the same technique on the dots after we contribute, and the loading screen on the speed test (I copied the former for the latter).
Perhaps I should have had an extra class on each of the 3 "bounce" dots that it could detect? 

Also it's a bit confusing that it is also nested inside the .form_error CSS rule?  But if we want to reuse componenents with styling tweaks there would always be an element of nesting unless we pass a context class down into the component?

stories/animatedDots.jsx
```diff
@@ -0,0 +1,32 @@
...
+
+
+const appearanceKnob = () => radios('Appearance', ['light', 'dark'], 'light');
+
+
+const stories = storiesOf('Animated dots', module).addDecorator(withKnobs);
+
+stories.add('Animated dots', () => (
+  <div style={{
+    background: '#121212',
```
walaura: not blocking, but this background could change based on the knob being light or dark!

PR #1470, opened 2019-02-06
Add address details to digipack checkout

damien-shaw-guardian:
## Why are you doing this?
Go Cardless require an address for billing purposes

[https://trello.com/c/OCErCAEv/2208-add-address-fields-to-digipack-checkout-flow](https://trello.com)

## Changes

* additional address fields have been added to the digi pack payment form

## Screenshots

screen shot 2019-02-06 at 16 47 18


assets/pages/digital-subscription-checkout/digitalSubscriptionCheckoutReducer.js
```diff
@@ -160,6 +190,7 @@ const formActionCreators = {
   setTelephone: (telephone: string): Action => ({ type: 'SET_TELEPHONE', telephone }),
   setBillingCountry: (countryRaw: string) => (dispatch: Dispatch<Action | CommonAction>) => {
     const country = fromString(countryRaw);
+    console.log({ country });
```
walaura: we don't need this!

assets/pages/digital-subscription-checkout/digitalSubscriptionCheckoutReducer.js
```diff
@@ -139,7 +166,10 @@ function getErrors(fields: FormFields): FormError<FormField>[] {
 // ----- Action Creators ----- //
 
 const setStage = (stage: Stage): Action => ({ type: 'SET_STAGE', stage });
-const setFormErrors = (errors: Array<FormError<FormField>>): Action => ({ type: 'SET_FORM_ERRORS', errors });
+const setFormErrors = (errors: Array<FormError<FormField>>): Action => {
+  console.log(errors);
```
walaura: nor this

assets/pages/digital-subscription-checkout/digitalSubscriptionCheckoutReducer.js
```diff
@@ -202,6 +239,12 @@ function initReducer(initialCountry: IsoCountry) {
     email: user.email || '',
     firstName: user.firstName || '',
     lastName: user.lastName || '',
+    addressLine1: '',
+    addressLine2: '',
```
walaura: do we need this to be nullable if it's optional?

LATaylor-guardian: It's handled the same way as telephone below which is also optional in the form.

walaura:
I'm aware this is scope creeping, but would it be possible to only show these fields to users paying via DD? we might be creating unnecessary friction for stripe users, which are all the international users

lmath:
@walaura Last I heard, we're thinking of running an a/b test later on to do just that and find out what the effect of that extra friction is before removing them

PR #1471, opened 2019-02-06
Dedupe CSS in dev

walaura:
## Why are you doing this?

Sorry about the massive PR! it's a lot of moving stuff around for no visual changes

**Our problem**: in dev, the scss files loaded from modules weren't being deduplicated, this meant that for every module loading `gu-sass` to get mixins and stuff, all the global css was coming along for the ride too, applying the same rule over and over again and making browsers _sloooow_.

Alongside that, some css files from old pages were still importing modules again, duplicating their rules in the process.

**The solution** for this is twofold. First of all, `gu-sass` is henceforth split up into `gu-sass` (helper functions that do not output any css) and `skeleton` which is the base css for each page and should be included **once**.

The second part was removing dupe imports manually. As this PR was ludicrous in side already I went ahead and did that for all our pages using webpack's css.

A third part is adding a new sass loader which promises to help with deduplication. Due to the way we load sass (per component) this won't do much, as sass is never aware of these duplications. Every little bit helps though.

As far as production and css output is concerned none of this matters at all, as the css was being deduped by `cssnano` at build time. That said, we should enjoy faster builds from not giving it so much CSS to dedupe in the first place. 

(@ giraffe: this won't affect the SSR test)

[**Trello Card**](https://trello.com/c/ikc6Jljp/2206-css-loading-issue-on-dev-env)

## Screenshots
check 'em out.
screenshot 2019-02-06 at 5 49 35 pm


assets/components/countryGroupSwitcher/countryGroupSwitcher.scss
```diff
@@ -1,3 +1,5 @@
+@import '~stylesheets/gu-sass/gu-sass';
```
walaura: this is not clear at first, it's done so `@extend .gu-content-margin` will work in these files - we want to rework the underlying grid moving forward, so this is an interim solution

jranks123:
Before you merge this, can you check with me or John that this won't affect our server side rendering test?

Hopefully tomorrow we will have something out that means that we don't have to copy + paste the CSS across to our test template manually, but until then we need to be careful as half of the audience on contribute is only being served whatever we last copied + pasted

jranks123:
haha I did not mean to close this PR!!!

PR #1473, opened 2019-02-06
100% new showcase page

walaura:
## Why are you doing this?

We did it team, the old page is gone!! Lots of cleanup to be done yet.

The new page lives in `/uk/support` as it'll get internationalised in the future. there are a ton of new redirects so the old urls (`/uk` and `/showcase`) still work out, and also another teeny redirect from `/support` to `/uk/support` that will eventually become a geo redirect.

**Remember to turn off the optimize ab test before merging**

[**Trello Card**](https://trello.com/c/A0NCo3EA/2167-switch-off-showcase-page-ab-test-and-put-100-traffic-to-the-new-page)

## Screenies

desktop-https-support thegulocal com-uk

For context, this is the **old** showcase page that will go away
support theguardian com_uk_utm_expid 0ctcmyigsgqiufp0emfobw 0 utm_referrer


jranks123:
Wait, I take back my approve, temporarily...

conf/routes
```diff
@@ -10,19 +10,29 @@ GET /unsupported-browser                            controllers.Application.unsu
...
 GET /ca                                             controllers.Application.redirect(location="/ca/contribute")
+GET /uk                                             controllers.Application.redirect(location="/uk/support")
 GET /                                               controllers.Application.geoRedirect()
-GET /showcase                                       controllers.Application.showcase()
 
-GET /sitemap.xml                                    controllers.SiteMap.sitemap
+
+# ----- Bundles Landing Page ----- #
+
+GET /support                                        controllers.Application.redirect(location="/uk/support")
```
jranks123: Wait a second, this will redirect everyone, globally, to the UK showcase page. Do we need to something different for US etc?

walaura: I just made up `/support` for this PR, the goal for our team is to eventually have internationalised showcase pages but right now we are starting with UK. Afaik the rest of redirects (/) and (/{country}) will preserve their rules

PR #1474, opened 2019-02-07
Remove region from Digital Pack checkout URL 

jacobwinch:
## Why are you doing this?

We do not currently use this for anything, as the preselected country is determined by the Internationalisation state which has been set earlier in the flow (presumably for most users this will be happen due to them visiting https://support.theguardian.com/uk/subscribe/digital).

Leaving this around adds confusion and makes setting up redirect tests complicated.

Note that I haven't set up any redirects for the old URLs - I didn't think it was necessary as the old flows have only been tested with a tiny number of users, but let me know if you disagree with this.

[**Trello Card**](https://trello.com/c/HoZPd10P/1514-a-b-test-on-digital-pack-checkout-flow)

## Changes

* Remove internationalisation of the checkout page


assets/pages/digital-subscription-checkout/digitalSubscriptionCheckout.scss
```diff
@@ -34,10 +34,7 @@
 
 // ----- Page-Specific Styles ----- //
 
-#digital-subscription-checkout-page-int,
-#digital-subscription-checkout-page-uk,
-#digital-subscription-checkout-page-us,
-#digital-subscription-checkout-page-au {
```
walaura: i wonder if we gain an advantage from this language-id thing in _any_ page, seems fiddly to me, not sure if we could remove it across the board while keeping the localised url

walaura:
I'd like to have redirects for the old urls because, selfishly, for test purposes i've gotten super used to just adding `/checkout` at the end of the dp url – i think we could just make them permanent redirects and call it a day?

jacobwinch:
> i think we could just make them permanent redirects and call it a day?

@walaura 👍see https://github.com/guardian/support-frontend/pull/1474/commits/f3b8f55b46986d435fd679ecb55e4c847ae8f53f

PR #1475, opened 2019-02-07
When we get an address for a user paying with direct debit, include it in the request to create the subscription

lmath:
## Why are you doing this?

We need to send address information when a new direct debit mandate is created via GoCardless in the digital pack checkout. We also need to allow direct debit to still work without address details because they currently don't need to be collected for recurring contributions. 

This `support-workers` PR (https://github.com/guardian/support-workers/pull/172) ensures that when address details are included in the request to create a subscription, that they are sent with the PaymentMethod when the Zuora subscription is set up. 

[**Trello Card**](https://trello.com/c/W4Hib27m/2192-when-creating-a-direct-debit-if-address-details-are-submitted-in-the-request-to-create-a-digital-pack-they-should-be-submitted-t)

## Changes

This PR ensures that if we have collected address fields when a user is paying via direct debit that 

1. They are formatted the way Zuora expects.
2. They are sent along to `support-workers` in the request to create a subscription. 

## Screenshots
n/a


app/services/stepfunctions/SupportWorkersClient.scala
```diff
@@ -185,3 +193,63 @@ object StepFunctionExecutionStatus {
   }
 
 }
+
+object PaymentFieldsEmbellisher {
```
rupertbates: I'd break this object out into it's own file

app/services/stepfunctions/SupportWorkersClient.scala
```diff
@@ -185,3 +193,63 @@ object StepFunctionExecutionStatus {
...
+    def singleAddressLine(addressLine1: String): AddressLine = {
+      val pattern: Regex = "([0-9]+) (.+)".r
+
+      addressLine1 match {
+        case pattern(streetNumber, streetName) => AddressLine(Some(streetNumber), streetName)
+        case _ => AddressLine(None, addressLine1)
+      }
+    }
+
+    val addressLine1MaybeSplit: Option[AddressLine] = addressLine1.map(singleAddressLine(_))
```
rupertbates: `.map(singleAddressLine(_))` can just be `.map(singleAddressLine)`

app/services/stepfunctions/SupportWorkersClient.scala
```diff
@@ -185,3 +193,63 @@ object StepFunctionExecutionStatus {
...
+    val addressLine2MaybeSplit: Option[AddressLine] = addressLine2.map(singleAddressLine(_))
```
rupertbates: Same as above

lmath:
I see that codacy has complained but I think the `if/else if/else` here is actually clearer than pattern matching in this case

PR #1478, opened 2019-02-07
Serve the CSS inline (apart from in DEV mode)

jranks123:
## Why are you doing this?

As part of the Server Side Rendering test, we are testing inlining the CSS. The current solution was a very hacky copy and paste job, that is unmaintainable going forwards and is blocking other tests. 

This solution removes the need to copy and paste the CSS into the template, by telling the Play app to read the exisiting CSS file directly and insert it into the template.

When in Dev mode, `webpack-dev-server` stores the assets in memory, so they aren't easy to access directly from the Play app. Therefore, when in dev mode, we will serve up the css the old fashioned way, by putting a link with a `href` to the file on the server. 

Tested in CODE and locally


assets/pages/new-contributions-landing/contributionsLanding.scss
```diff
@@ -739,3 +739,66 @@ form {
   padding: 0 $gu-h-spacing $gu-v-spacing;
   margin-top: $gu-v-spacing;
 }
+
+
+/* Server side rendering test loading message*/
```
jranks123: This stuff moved across from the test.scala.html template

assets/pages/new-contributions-landing/contributionsLanding.scss
```diff
@@ -739,3 +739,66 @@ form {
...
+
+.form__radio-group--pills .amounts__placeholder .form__radio-group-input + .form__radio-group-label {
+  border: none;
+}
+
+.amounts__placeholder .form__radio-group-label {
```
walaura: not for this pr but could this be perhaps a higher level body/html class? that way you could also serve the same html

assets/pages/new-contributions-landing/contributionsLanding.scss
jranks123: This stuff moved across from the test.scala.html template

PR #1479, opened 2019-02-07
Remove old contributions flow tests

walaura:
## Why are you doing this?

#1473 has broken the production tests because they are testing against the old showcase page. This removes the old tests and moves the new tests from `/contributions/new_flow` to `/contributions`


jacobwinch:
image 

🥇🥇🥇🥇🥇🥇🥇🥇🥇🥇🥇🥇

PR #1480, opened 2019-02-07
Remove old checkout flow (woohoo!)

jranks123:
## Why are you doing this?
We no longer need this, so we can safely delete it! Woohoo!


assets/helpers/paymentIntegrations/stripeCheckout.js
```diff
@@ -1,66 +1,76 @@
 // @flow
 
-
-// ----- Setup ----- //
-
-let stripeHandler = null;
-
+/**
+ * This module contains a copy of pages/oneoff-contributions/helpers/ajax so that
+ * the new payment flow can use this code without any impact on the existing
```
jranks123: This looks like I've changed the file, but basically I've deleted the old `assets/helpers/paymentIntegrations/stripeCheckout.js` and replaced it with the one from assets/helpers/paymentIntegrations/newPaymentFlow/stripeCheckout.js`, which is the only one we need now

walaura:
I am literally never reviewing 4000 lines of code but if the test pass, lgtm!

PR #1487, opened 2019-02-07
Use a yellow favicon in DEV

walaura:
## Why are you doing this?
Looks like a sun!! Dotcom & identity do this too and it's super convenient if you have a thousand tabs 

[**Trello Card**](https://trello.com/c/HHqobaNU/139-yellow-favicon-in-dev)

## Screenshots

screenshot 2019-02-07 at 4 28 18 pm


app/controllers/Application.scala
```diff
@@ -33,12 +33,13 @@ class Application(
     stringsConfig: StringsConfig,
     settingsProvider: AllSettingsProvider,
     guardianDomain: GuardianDomain,
-    stage: Stage
+    val stage: Stage
```
rupertbates: Is this needed?

app/controllers/PayPalOneOff.scala
```diff
@@ -27,12 +28,14 @@ class PayPalOneOff(
     paymentAPIService: PaymentAPIService,
     identityService: IdentityService,
     settingsProvider: AllSettingsProvider,
-    tipMonitoring: Tip
+    tipMonitoring: Tip,
+  stage: Stage
```
johnduffell: is this indentation alright? looks underindented on github diff

walaura: i think my intellij messed up spaces/tabs, will go over today

johnduffell:
> Seems a lot of code to change a favicon doesn't it?! 😃

agree!... but not a lot of code to change the content of every single page?

although on that topic @walaura another option is to keep it in the standard place, link it to a "real" route rather than an assets path.  then you can put the if statement in the FavIconController instead of every single template.

walaura:
> > Seems a lot of code to change a favicon doesn't it?! 😃
> 
> agree!... but not a lot of code to change the content of every single page?
> 
> although on that topic @walaura another option is to keep it in the standard place, link it to a "real" route rather than an assets path. then you can put the if statement in the FavIconController instead of every single template.

Great idea! i actually tried replacing it via webpack too, this is on that line. I'm not sure if the overhead in production is worth it?

johnduffell:
not sure if the favicon is critical for PROD... and anyway we can put super long cache headers in so fastly will basically keep it for ever, so it would prbably be just as quick if not quicker for the one person who hits it first during/after a deploy!

However I don't want to rain on your great idea and it's not a blocker, I just got thinking about why rupert's comment was true!

walaura:
@johnduffell nono, I agree it's A TON of code plus a new extra global peppered all over the place. This is already such a tough merge that it's gonna take work no matter what I do, I'm happy to try to make the route later today!

PR #1490, opened 2019-02-08
Remove deprecated backend code and redirect deprecated endpoints to /contribute

jranks123:
## Why are you doing this?
We no longer send people to:

`/contribute/recurring`    
`/contribute/recurring-guest`           
`/contribute/recurring/thankyou`                
`/contribute/recurring/existing`             
`/contribute/recurring/pending`                  
`/contribute/one-off `
`/contribute/one-off/thankyou`
`/contribute/one-off/autofill`       

So we can delete the code behind these endpoints, and redirect anyone still coming to the endpoint (probably bots) to the contribute page.


conf/routes
```diff
@@ -45,19 +45,27 @@ GET /event/*eventId                                 controllers.Application.redi
 
 GET  /monthly-contributions                         controllers.Application.contributionsRedirect()
 
+# Deprecated endpoints that we now redirect to the contribute page. Once we have established that no traffic
+# is coming this way, we should remove them
+GET  /contribute/recurring                          controllers.Application.contributeGeoRedirect()
```
joelochlann: I'm not sure if we have any SEO experts in the department but I have a feeling that the best way to tell any search engine bots what has happened to these endpoints is to serve a `301 Moved Permanently` redirect response rather than just serve up a different endpoint

And it does actually look like at least some of these endpoints have been indexed:
https://www.google.com/search?q=the+guardian+contribute+recurring

joelochlann: https://moz.com/learn/seo/redirection

jacobwinch: Good point @joelochlann. @walaura recently added a helper function for this https://github.com/guardian/support-frontend/blob/f472a20b4991eb0235a791410c41b03369ea7422/app/controllers/Application.scala#L81-L83 - perhaps we should be using that instead?

jranks123: Wow that's so weird that the thank you page comes up as the third result!

joelochlann: I know!

PR #1492, opened 2019-02-08
refactor: changed form props in digitalSubscriptionCheckoutReducer to null

damien-shaw-guardian:
**Refactor:** changed form props in digitalSubscriptionCheckoutReducer to null,

This conforms to the backend Type system which is using Option to denote the expected type


lmath:
Thank you for doing this! 🌟

PR #1493, opened 2019-02-08
Use subscriptions client id

jacobwinch:
## Why are you doing this?

We are currently using the wrong identity client id as part of the new Digital Pack checkout flow. Notably, we display the following text (which doesn't make sense in the context of this flow):

> To complete your contribution, please sign in to your Guardian account or quickly create one.
> 
> This will help you manage future payments. Staying signed in means you’ll no longer see the “since you’re here” messages asking you to support our journalism.

This PR fixes the problem by using our very own client id.

## Changes

* Use correct client id

## Screenshots

Before:
image

After:
image


LATaylor-guardian:
👍

PR #1496, opened 2019-02-11
Normalise phone numbers before saving them to Salesforce

lmath:
## Why are you doing this?
https://github.com/guardian/support-frontend/pull/1393 added that when a user provided a phone number, that it would be saved to Salesforce.

However, the digital pack checkout on subscribe.theguardian normalises phone numbers before sending them to Salesforce (we can't expect CSRs to know country codes for every possible country!). So, I've brought in com.google.i18n.phonenumbers and the code in membership-common that does the normalising:

https://github.com/guardian/membership-common/blob/master/src/main/scala/com/gu/memsub/NormalisedTelephoneNumber.scala

https://github.com/guardian/membership-common/blob/master/src/test/scala/com/gu/memsub/NormalisedTelephoneNumberTest.scala

And now when a telephone number is provided, it is normalised (based on the user entered number and the billing country) before it is sent along to salesforce.

[**Trello Card**](https://trello.com/c/IQxt9H0N/2203-normalise-telephone-numbers-before-sending-them-to-salesforce)

## Changes

* Adds `com.google.i18n.phonenumbers` dependency
* Copies over the code in membership-common that is normalising phone numbers in the digital pack checkout on subscribe.theguardian
* Switches telephone number back to an `Input1` component so that we actually send it along in the request to `POST subscribe/digital/create`

## Screenshots
image

⬇️ 

image


app/controllers/DigitalSubscription.scala
```diff
@@ -118,20 +120,24 @@ class DigitalSubscription(
 
         type ApiResponseOrError[RES] = EitherT[Future, CreateDigitalSubscriptionError, RES]
 
-        if (validationPasses(request.body)) {
+        val normalisedTelephoneNumber = NormalisedTelephoneNumber.fromStringAndCountry(request.body.telephoneNumber, Some(request.body.country))
```
LATaylor-guardian: Why does `request.body.country` need to be wrapped in Some here?

LATaylor-guardian:
Nice test coverage @lmath! 👍

app/controllers/DigitalSubscription.scala
lmath: That's a good point! It doesn't. That ended up there because I just brought over some code for NormalisedTelephoneNumber directly from membership-common. I've modified the `fromStringAndCountry` to take a `Country` rather than an `Option[Country]`

lmath:
Thanks for that test coverage @AWare 😄

PR #1498, opened 2019-02-11
Temporarily add a second version of Digital Pack landing page to support Optimize test

jacobwinch:
## Why are you doing this?

The current setup of our Digital Pack checkout experiment is problematic, because it is not recording the behaviour of users who drop out of the flow after hitting the Identity sign in/sign up step. This is likely to unfairly skew the result in favour of the new flow. 

In order to work around this problem, this PR introduces a second version of the Digital Pack landing page. This allows us to set up a new test which uses the Digital Pack landing page as the entry point. In such a test, 50% of users will see the existing landing page <region>/subscribe/digital (which links out to the old checkout flow), and the other 50% will be redirected* to <region>/subscribe/digital/new (which links out to the new checkout flow). 

All users will now be served a page on support.theguardian.com _before_ any redirects to identity occur, which should ensure that a user's session is included in the experiment, even if they drop out of the flow after moving to the identity step. 

_*Unfortunately Optimize will only allow us to specify a single URL for the redirect test, which means that two redirects are actually required to get the user to the landing page. We'll get something like this:_

_User makes request to /uk/subscribe/digital > support-frontend serves the page > Optimize redirects to /subscribe/digital/new > support-frontend serves /uk/subscribe/digital/new._

### DIFF
--- a/app/controllers/DigitalSubscription.scala
+++ b/app/controllers/DigitalSubscription.scala
@@ -78,6 +78,7 @@ class DigitalSubscription(
   }
 
   def digitalGeoRedirect: Action[AnyContent] = geoRedirect("subscribe/digital")
+  def digitalNewGeoRedirect: Action[AnyContent] = geoRedirect("subscribe/digital/new")
 
   def displayForm(): Action[AnyContent] =
     authenticatedAction(subscriptionsClientId).async { implicit request =>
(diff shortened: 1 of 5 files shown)


jacobwinch:
Going to try a different approach.

PR #1500, opened 2019-02-11
Clean up dialogs in storybook

walaura:
## Why are you doing this?

The dialog component was bonkers long and I saw this cool pattern of a component that provides state to its children as props that is super handy for storybook, so I implemented that. Hopefully this makes the actual dialog specific code easier to follow..

In addition there's two new storybook features:

- See the source code of stories inline
- Load stories from `/stories/_index` instead of going to the storybook config

Also!! Storybook itself is a bit prettier now:

screenshot 2019-02-12 at 11 57 48 am


.storybook/util/withState.jsx
```diff
@@ -0,0 +1,19 @@
...
+    this.refs = this.props.refs;
+  }
+  render() {
+    return (
+      this.props.children(this.state, (state) => this.setState(state), this.refs)
+    );
+  }
+}
+
+export default WithState;
```
walaura: I've locked this away in the storybook folder because I don't think live code should have these powers but maybe we _do_ want it? it's a bit like having hooks

.storybook/util/withState.jsx
```diff
@@ -0,0 +1,19 @@
+// @flow
+
+import React, {type Node, Component} from 'react';
+
+class WithState extends Component<{|initialState: {}, children: Function, refs?: any[]|}, {}> {
+  state = {
+    ...this.props.initialState
+  }
+  componentDidMount() {
+    this.refs = this.props.refs;
```
walaura: this is firmly in the i'm not sure if it should work but it works' camp, given this is internal only I think that's fine but I'm happy to investigate

PR #1503, opened 2019-02-12
Use a yellow favicon in DEV (with less code)

walaura:
Rehash of #1487 

----

## Why are you doing this?
Looks like a sun!! Dotcom & identity do this too and it's super convenient if you have a thousand tabs 

[**Trello Card**](https://trello.com/c/HHqobaNU/139-yellow-favicon-in-dev)

## Screenshots

screenshot 2019-02-07 at 4 28 18 pm


johnduffell:
Thanks! a lot simpler than changing the URL based on the stage! 👍also means the assets are outside of the hashing therefore it can be cached much longer.

johnduffell:
did this really work it's not actually fetching it for me in prod?
weirdly it's an image/x-icon from play, but in the html it's down as a image/png

johnduffell:
oh actually it does work, apparently you have to restart chrome to pick up the favicon!  but it's still weird that png thing

PR #1506, opened 2019-02-12
Message for customers who have signed in to get into checkout but already have Digital Pack - Front

walaura:
## Why are you doing this?
AS A customer I WANT to be shown a message when I'm trying to buy Digital Pack but already have it SO THAT I don't end up buying it twice in error.

This PR includes a new stage for the checkout, accessible at `subscribe/digital/checkout/existing` should we do a server side redirect.

[**Trello Card**](https://trello.com/c/06A5kQwD/2226-message-for-customers-who-have-signed-in-to-get-into-checkout-but-already-have-digital-pack-ty-page)

## Screenshots

screenshot 2019-02-12 at 3 06 58 pm

### DIFF
--- a/assets/pages/digital-subscription-checkout/digitalSubscriptionCheckoutReducer.js
+++ b/assets/pages/digital-subscription-checkout/digitalSubscriptionCheckoutReducer.js
@@ -35,7 +35,7 @@ import { showPaymentMethod, onPaymentAuthorised, countrySupportsDirectDebit } fr
 
 // ----- Types ----- //
 
-export type Stage = 'checkout' | 'thankyou' | 'thankyou-pending';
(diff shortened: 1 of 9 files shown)


assets/pages/digital-subscription-checkout/digitalSubscriptionCheckoutReducer.js
```diff
@@ -231,8 +231,11 @@ function initReducer(initialCountry: IsoCountry) {
     : Monthly;
   const { productPrices } = window.guardian;
 
+  const pathName = new URL(window.location).pathname.split('/');
+  const checkoutPathName: string = pathName.splice(pathName.findIndex(c => c === 'checkout')).pop();
```
walaura: 1. Should we do this? (vs a parameter or client side checking for id)
2. If we are to do this, should I extract this into a helper? So it has a more descriptive name

PR #1510, opened 2019-02-13
Remove 6for6 from GW page

walaura:
## Why are you doing this?

Because journalism is in crisis

[**Trello Card**](https://trello.com/c/nWZR6loN/2202-remove-6-for-6-from-gw-product-page)

## Changes

* Remove the 6for6 price tier
* Default to Quarterly (@SarahLaughton !)
* Rename `billingPeriods` to `displayBillingPeriods` so as to imply that this may be a subset of the periods – we are keeping 6for6 in the data model for a bit longer than this removal so we can't eviscerate all of it just yet.

## Screenshots

screenshot 2019-02-13 at 1 36 53 pm


walaura:
I'm not sure if we want to hold onto this by tomorrow?

SarahLaughton:
> I'm not sure if we want to hold onto this until tomorrow?

Yes please - the marketing activity will switch over tomorrow morning.

PR #1513, opened 2019-02-13
TY page if the user has a DP

walaura:
## Why are you doing this?
Second bit of #1506 

Fun stuff: we had to create this page its own route in both scala and react because otherwise it was weirdly entangled with the checkout code.

**Do not merge yet** we are writing tests

[**Trello Card**](https://trello.com/c/06A5kQwD/2226-message-for-customers-who-have-signed-in-to-get-into-checkout-but-already-have-digital-pack-ty-page)

## Screenshots

screenshot 2019-02-12 at 3 06 58 pm

### DIFF
--- a/app/controllers/DigitalSubscription.scala
+++ b/app/controllers/DigitalSubscription.scala
@@ -6,7 +6,7 @@ import admin.settings.{AllSettings, AllSettingsProvider, SettingsSurrogateKeySyn
 import assets.AssetsResolver
 import cats.data.EitherT
 import cats.implicits._
-import com.gu.identity.play.IdUser
+import com.gu.identity.play.{AccessCredentials, IdUser}
 import com.gu.support.catalog.DigitalPack
 import com.gu.support.config.{PayPalConfigProvider, StripeConfigProvider}
 import com.gu.support.pricing.PriceSummaryServiceProvider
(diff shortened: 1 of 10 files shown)


test/controllers/SubscriptionsTest.scala
```diff
@@ -58,6 +62,14 @@ class SubscriptionsTest extends WordSpec with MustMatchers with TestCSRFComponen
       val payPal = mock[PayPalConfigProvider]
       when(payPal.get(any[Boolean])).thenReturn(PayPalConfig("", "", "", "", "", ""))
 
+      import cats.data.EitherT
+      import cats.implicits._
```
walaura: oof we should remove this!

app/controllers/DigitalSubscription.scala
```diff
@@ -72,16 +73,32 @@ class DigitalSubscription(
...
         error => {
           SafeLogger.error(scrub"Failed to display digital subscriptions form for ${request.user.id} due to error from identityService: $error")
-          InternalServerError
+          Future.successful(InternalServerError)
         },
-        user => Ok(digitalSubscriptionFormHtml(user))
-      ).map(_.withSettingsSurrogateKey)
+        user => {
+          userHasDigipack(request.user) map {
+            case true =>  Redirect(routes.DigitalSubscription.displayThankYouExisting().url, request.queryString, status = FOUND)
```
jacobwinch: Minor: I would probably favour if/else here as I think we're just dealing with a Boolean

PR #1514, opened 2019-02-14
Send billing address in request to support-workers

lmath:
## Why are you doing this?

Now that we need to send address details (when provided, so currently just for digital pack)
- along with the direct debit payment method
- to the billing address on the user's salesforce account

I would like to send through address details along with the `User` in `CreateSupportWorkersRequest` and let `support-workers` care about what kind of data Zuora or Salesforce wants. 

And also ensure that country and optional state are still making their way to Zuora in the same way as before.  

See: https://github.com/guardian/support-libraries/pull/23

## Changes

* Send billing address along with User in the request to support-workers

## Screenshots
n/a


build.sbt
```diff
@@ -68,6 +68,7 @@ libraryDependencies ++= Seq(
   "org.typelevel" %% "cats-core" % "1.0.1",
   "com.dripower" %% "play-circe" % "2609.1",
   "com.gu" %% "support-services" % "0.7",
+  "com.gu" %% "support-models" % "0.57-SNAPSHOT",
```
lmath: Depends on https://github.com/guardian/support-libraries/pull/23

app/controllers/DigitalSubscription.scala
```diff
@@ -150,6 +150,7 @@ class DigitalSubscription(
       lastName = request.lastName,
       country = request.country,
```
lmath: Once `support-workers` gets country and state from billing address, I will remove country and state from `CreateSupportWorkersRequest`

PR #1517, opened 2019-02-14
Add NZ & Canada subscribe pages

rupertbates:
## Why are you doing this?

So that customers in New Zealand and Canada have localised subscribe pages and correct prices in their local currency for the Digital Pack.

[**Trello Card**](https://trello.com/c/ZfGLlSyQ/2186-internationalise-digital-pack-product-page)

### DIFF
--- a/assets/pages/digital-subscription-landing/digitalSubscriptionLanding.scss
+++ b/assets/pages/digital-subscription-landing/digitalSubscriptionLanding.scss
@@ -25,7 +25,9 @@
 #digital-subscription-landing-page-uk,
 #digital-subscription-landing-page-us,
 #digital-subscription-landing-page-au,
-#digital-subscription-landing-page-eu {
+#digital-subscription-landing-page-eu,
+#digital-subscription-landing-page-nz,
+#digital-subscription-landing-page-ca {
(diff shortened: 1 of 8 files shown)


assets/pages/digital-subscription-landing/digitalSubscriptionLanding.scss
```diff
@@ -25,7 +25,9 @@
 #digital-subscription-landing-page-uk,
 #digital-subscription-landing-page-us,
 #digital-subscription-landing-page-au,
-#digital-subscription-landing-page-eu {
+#digital-subscription-landing-page-eu,
+#digital-subscription-landing-page-nz,
+#digital-subscription-landing-page-ca {
```
walaura: not for this PR but do we need all these ids? what is the purpose?

rupertbates: I don't think they really serve a purpose, the idea was to allow per country customisation but since we don't really do that I think we should remove them.

walaura: Gonna card this in health

PR #1520, opened 2019-02-15
Stop sending address fields along with payment fields

lmath:
## Why are you doing this?

We will now send always billing address through when starting a new step function execution. So, this makes billing address required. 

Additionally, after https://github.com/guardian/support-frontend/pull/1514 is merged I've moved the responsibility to convert addresses collected from the form -> addresses required by Zuora (eventually GoCardless via the Zuora-GoCardless integration) or Salesforce. So, I can remove that munging code. 

Additionally, after https://github.com/guardian/support-frontend/pull/1514 is merged, `support-workers` will not need address fields in the payment fields, so let's stop sending them.

I will need to remove payment fields, country and state in a future PR.

More info here: https://github.com/guardian/support-libraries/pull/23

## Changes

* Bumped support-services to 0.9 (billing address is now required)
* Stopped sending address fields along with payment fields (after https://github.com/guardian/support-workers/pull/176 is merged support-workers will handle address munging to make fields zuora-friendly)
* Remove address munging code that will live in `support-workers` instead

## Screenshots
n/a


rupertbates:
What about for recurring contributions? They won't have a billing address.

lmath:
They're going to send e.g.
```
Address(
      lineOne = None,
      lineTwo = None,
      city = None,
      state = request.state,
      postCode = None,
      country = request.country
    )
```

in `billingAddresss`

PR #1522, opened 2019-02-15
Display Zuora prices in DP + Unify Price/ProductPrice

walaura:
Brace up buttercups

We used to have `helpers/internationalisation/prices` and `helpers/productPrice/productPrices`. They returned slightly different objects but what I really liked from `internationalisation/prices` is that it returned the currency, so I've added that to `productPrices`.

I've also preemptively split up `productPrices` to contain types and helpers and `digitalProductPrices` to contain the DP specific parts. Bit of an API change here, in broad terms this is how it used to be:
* getPrice => {price, promotion?}
* getPriceToPay => {number with promo applied} 

And now it's
* getRegularPrice => {price, currency}
* getFinalPrice => {price or promotional price if any, currency}
* getPromotion => {promotion?}

And a new helper where if you have a promotion and a regular price you can use `applyPromotion(price, promotion)` to get the final price.

As for `helpers/internationalisation/prices`, since it's only concerned with hardcoded prices, I've removed the last bits of it to avoid confusion. You used to be able to use functions like `GBP(12.34)` to return a `price` but in practice this was vaguely in use and will fade away in a bigger way as we continue fetching prices from zuora.

This PR also adds the new prices to the DP landing page (including promo code support!) but only if there are no flash sales currently active. The reason for this being we have to make a decision on how to send the flash sale promo code to the js app.


app/controllers/DigitalSubscription.scala
```diff
@@ -61,14 +61,18 @@ class DigitalSubscription(
...
     val description = stringsConfig.digitalPackLandingDescription
     val canonicalLink = Some(buildCanonicalDigitalSubscriptionLink("uk"))
+    val promoCode = request.queryString.get("promoCode").flatMap(_.headOption)
     val hrefLangLinks = Map(
       "en-us" -> buildCanonicalDigitalSubscriptionLink("us"),
       "en-gb" -> buildCanonicalDigitalSubscriptionLink("uk"),
       "en-au" -> buildCanonicalDigitalSubscriptionLink("au"),
       "en" -> buildCanonicalDigitalSubscriptionLink("int")
     )
+    val productPrices = priceSummaryServiceProvider.forUser(false).getPrices(DigitalPack, promoCode)
```
rupertbates: You need to pass in true for test users here

walaura: @rupertbates I don't have an iduser on this route :(

app/views/main.scala.html
```diff
@@ -13,7 +16,8 @@
 	styles: Html = Html(""),
 	scripts: Html = Html(""),
   data: Map[String, String] = Map.empty,
-  csrf: Option[String] = None
+  csrf: Option[String] = None,
+  productPrices: Option[ProductPrices] = None,
```
rupertbates: I'm not sure this needs to be an `Option` it's not on the checkout page

walaura: `main` is used for all default responses, most without prices right now and some that will never have prices (like 404 pages or the checkout ty) I'd rather not create a new html template if we can avoid it

app/controllers/DigitalSubscription.scala
rupertbates: You can still check whether they have a `_test_username` cookie. Have a look at TestUserService

walaura: oh!

app/views/main.scala.html
rupertbates: I hadn't noticed that this was in main. It would probably be good to remove the code from `digitalSubscription.scala.html` which does the same thing and delegate to this file.

walaura: good shout!

PR #1524, opened 2019-02-15
bump support-services to 0.10

lmath:
## Why are you doing this?
As of https://github.com/guardian/support-libraries/pull/25, now that country and state are already being sent along with the billing address, we should stop sending them along as fields on `User`.

Note: depends on https://github.com/guardian/support-workers/pull/177

## Changes
* Bump `support-services` to 0.10
* Stop sending country and state fields along with `User` fields

## Screenshots

n/a


app/monitoring/Tip.scala
```diff
@@ -70,7 +70,7 @@ object PathVerification {
   }
 
   def monitoredPaymentMethod(paymentFields: PaymentFields): MonitoredPaymentMethod = paymentFields match {
-    case DirectDebitPaymentFields(_, _, _, _, _, _, _, _) => DirectDebit
```
jacobwinch: 🎉

PR #1526, opened 2019-02-15
Use constants for currencies

rupertbates:
## Why are you doing this?
Same idea as #1521

Introduce a set of constants for currency values which means we don't need to sprinkle string representations throughout the app.

This is better because it is less prone to typos and allows IDEs to do refactoring and code analysis like 'find usages' much more successfully.


rupertbates:
Having problems with the currency tests, closing for now

PR #1527, opened 2019-02-18
Storybook stories for <content>

walaura:
## Why are you doing this?

Did this by the end of friday – changed them to use `appearance` instead of `type` like the rest of elements too

screenshot 2019-02-18 at 10 22 27 am


stories/layout.jsx
```diff
@@ -0,0 +1,108 @@
...
+        You can wrap custom non-text elements that need the same max-width as
+        the text in <code>NarrowContent</code>. It looks just like <code>Text</code> but
+        semantics matter! Don{'\''}t wrap non text elements in <code>Text</code> just to get a max width
+      </NarrowContent>
+    </Content>
+    <Content >
+      <Text title="Feeling adventurous ?" />
+      <Outset>
+        <Text>
+          {/* eslint-disable-next-line */}
```
akash1810: any reason why we don't want eslint here?

walaura: it was complaining about the emoji and the fix was breaking flow and I couldn't run into any easy fixes so I figured since we shouldn't be using emojis in the actual pages anyway this was just not worth fixing further

PR #1535, opened 2019-02-19
Turn off SSR test

tomrf1:
## Why are you doing this?
The server-side rendering test was successful. This change switches 100% of the audience to server-side rendering.

Note - there is some more cleanup to be done but for now let's get this switched off


jranks123:
FYI it had passed teamcity build, just not codacity

PR #1542, opened 2019-02-22
Refactor to use a generic endpoint for creating subscriptions

jacobwinch:
## Why are you doing this?

We are in the process of migrating the creation of all subscriptions products to use support-frontend/workers. Since the call to create a subscription is going to be very similar for all subscriptions products, this PR refactors the existing code to remove any Digital Pack specifics.

As far as I can tell, the main difference we'll need to take into account when creating different products will be the validation rules that we use, so I've also tried to suggest an approach which allows the validation to be made product-specific.

This PR also adds an optional firstDeliveryDate to the CreateSupportWorkersRequest model, in preparation for https://trello.com/c/oRCVNC3s/2248-print-checkout-component-with-a-list-of-start-dates. Note that this PR doesn't wire that date through to the backend yet - this will follow as a separate change but I wanted to get things ready for @walaura).

[**Trello Card**](https://trello.com/c/NHSxiQqe/2254-add-a-create-endpoint-for-paper)

## Changes

* Add new generic controller (and endpoint) for creating subscriptions and wire it up
* Remove all code for creating subscriptions from Digital Pack controller
* Add optional first delivery date so that the interface is complete for clients (although note that this is not currently wired through to support-workers)
* Start calling the new endpoint for creating Digital Pack subscriptions (client side change), but leave the old endpoint in place until we are sure that noone will call it
* Tidy up some unused clutter from Digital / Paper Controllers

### DIFF
--- a/app/controllers/CreateSubscription.scala
+++ b/app/controllers/CreateSubscription.scala
@@ -1,0 +1,127 @@
+package controllers
+
+import actions.CustomActionBuilders
+import actions.CustomActionBuilders.AuthRequest
+import admin.settings.{AllSettingsProvider, SettingsSurrogateKeySyntax}
+import cats.data.EitherT
+import cats.implicits._
+import com.gu.identity.play.IdUser
+import com.gu.support.workers.{Address, BillingPeriod, DigitalPack, User}
+import io.circe.syntax._
+import lib.PlayImplicits._
+import monitoring.SafeLogger
+import monitoring.SafeLogger._
+import play.api.libs.circe.Circe
+import play.api.mvc._
+import services.stepfunctions.{CreateSupportWorkersRequest, StatusResponse, SupportWorkersClient}
+import services.{IdentityService, TestUserService}
+import utils.NormalisedTelephoneNumber
+import utils.NormalisedTelephoneNumber.asFormattedString
+import utils.SimpleValidator._
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class CreateSubscription(
+  client: SupportWorkersClient,
+  val actionRefiners: CustomActionBuilders,
+  identityService: IdentityService,
+  testUsers: TestUserService,
+  components: ControllerComponents,
+  settingsProvider: AllSettingsProvider,
+  val supportUrl: String
+)(implicit val ec: ExecutionContext) extends AbstractController(components) with GeoRedirect with CanonicalLinks with Circe with SettingsSurrogateKeySyntax {
+
+  import actionRefiners._
+
+  sealed abstract class CreateSubscriptionError(message: String)
+  case class ServerError(message: String) extends CreateSubscriptionError(message)
+  case class RequestValidationError(message: String) extends CreateSubscriptionError(message)
+
+  type ApiResponseOrError[RES] = EitherT[Future, CreateSubscriptionError, RES]
+
+  def create: Action[CreateSupportWorkersRequest] =
+    authenticatedAction(recurringIdentityClientId).async(circe.json[CreateSupportWorkersRequest]) {
+      implicit request: AuthRequest[CreateSupportWorkersRequest] =>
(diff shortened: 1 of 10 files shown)


app/controllers/CreateSubscription.scala
```diff
@@ -0,0 +1,127 @@
...
+  sealed abstract class CreateSubscriptionError(message: String)
+  case class ServerError(message: String) extends CreateSubscriptionError(message)
+  case class RequestValidationError(message: String) extends CreateSubscriptionError(message)
+
+  type ApiResponseOrError[RES] = EitherT[Future, CreateSubscriptionError, RES]
+
+  def create: Action[CreateSupportWorkersRequest] =
+    authenticatedAction(recurringIdentityClientId).async(circe.json[CreateSupportWorkersRequest]) {
+      implicit request: AuthRequest[CreateSupportWorkersRequest] =>
+        val validator: CreateSupportWorkersRequest => Boolean = request.body.product match {
```
jacobwinch: This pattern match is obviously pointless at the moment but it gets things ready for adding separate validation for Paper which will be coming soon...

app/controllers/DigitalSubscription.scala
```diff
@@ -21,19 +16,14 @@ import monitoring.SafeLogger._
...
-import views.html.{subscriptionCheckout, main}
+import views.html.subscriptionCheckout
 import views.html.helper.CSRF
-import utils.SimpleValidator._
 
 import scala.concurrent.{ExecutionContext, Future}
 
 class DigitalSubscription(
     priceSummaryServiceProvider: PriceSummaryServiceProvider,
-    client: SupportWorkersClient,
```
jacobwinch: This has been removed as it's only used when creating a sub

app/controllers/DigitalSubscription.scala
```diff
@@ -44,9 +34,7 @@ class DigitalSubscription(
     components: ControllerComponents,
     stringsConfig: StringsConfig,
     settingsProvider: AllSettingsProvider,
-    val supportUrl: String,
-    tipMonitoring: Tip,
```
jacobwinch: This was unused

app/controllers/PaperSubscription.scala
```diff
@@ -26,20 +26,16 @@ import scala.concurrent.{ExecutionContext, Future}
 
 class PaperSubscription(
   priceSummaryServiceProvider: PriceSummaryServiceProvider,
-  client: SupportWorkersClient,
```
jacobwinch: Again, just removing some unused params here

conf/routes
```diff
@@ -94,7 +94,9 @@ GET  /$country<(uk|us|au|eu|int|nz|ca)>/subscribe/digital     controllers.Digita
 GET  /$country<(uk|us|au|eu|int|nz|ca)>/subscribe/digital/checkout  controllers.Application.permanentRedirectWithCountry(country, location="/subscribe/digital/checkout")
 GET  /subscribe/digital/checkout  controllers.DigitalSubscription.displayForm()
 GET  /subscribe/digital/checkout/thankyou-existing  controllers.DigitalSubscription.displayThankYouExisting()
-POST /subscribe/digital/create  controllers.DigitalSubscription.create
+
+# Deprecated, soon all requests to create a subscription should go via /subscribe/create instead
```
jacobwinch: I guess we should leave this here for a short while because some users might have the old JS when this is first deployed(?) and I don't want them to get 404s if they try to pay...

app/controllers/CreateSubscription.scala
lmath: I like this idea! But maybe move this stuff into SimpleValidator too? Then the validation stuff can be together and it would be easier to test.

jacobwinch: Yep good point. I have a follow up PR for Paper specific validation so I will refactor this as part of that change.

PR #1546, opened 2019-02-22
Improve Stripe Payment Request Button UX

jranks123:
## Why are you doing this?

To improve the UX around the Stripe Payment Request Button

|Control|
|------|
|control|

|Variant|
|-----|
|variant|


app/utils/QueryStringUtils.scala
```diff
@@ -3,7 +3,6 @@ package utils
 
 object QueryStringUtils {
   def addFormDesignTestParameterQueryString(queryString: Map[String, Seq[String]]): Map[String, Seq[String]] = {
-    val formDesignTestParameterValue = if (Random.nextFloat <= 0.2) "control" else "variant"
-   queryString + ("formDesignTest" -> Seq(formDesignTestParameterValue))
```
jranks123: This change has actually already happened. annoyingly, I checked out this branch from a branch where I'd previously made this change. Don't worry it won't cause any problems, I promise.

assets/components/button/button.scss
```diff
@@ -7,6 +7,7 @@
   text-decoration: none;
   font-weight: bold;
   height: $gu-cta-height;
+  min-height: $gu-cta-height;
```
paulbrown1982: Thanks for explaining why you needed to do this @jranks123. Maybe just do a smoke test on a few random other pages on the support site.

paulbrown1982:
👍

PR #1551, opened 2019-02-25
Freq tabs ordering test

johnduffell:
## Why are you doing this?
There is a theory that ordering the tabs differently will improve performance of the page.  This test will have the existing variant ordered "S-M-A" as well as the two alternatives with Annual in the middle.  The theory is that it would let people centre on the standard option and allow them to look each way.

There is a separate effort to experiment more deeply with the whole way people are deciding the frequency but that needs some UX input.

[**Trello Card**](https://trello.com/card/board/frequency-tabs-ordering-on-landing-page-test/5bae0284c151920bfc73c13b/5c6fd80d583ae77fac82b568)

|Original:|
|---|
|image|
|New SSR (waiting for page to load)|
|image|

----

| variant MAS:|variant SAM:| control|
|---|---|---|
|image|image|image|

@jranks123


assets/pages/new-contributions-landing/contributionsLanding.scss
```diff
@@ -927,6 +927,11 @@ form {
   background-color: #ededed;
 }
 
+.frequencies__placeholder .form__radio-group-label {
+  background-color: #ffffff;
```
jranks123: You'll wanna use `gu-colour(neutral-100)`

johnduffell: thanks, did that now! 👍

johnduffell:
I had a lot of trouble getting it to pick up the new build even after switching on status check reporting in teamcity for the new build and respinning it.  In the end I just switched off the check.
We can turn it on again later assuming we are still worried about people merging broken things onto master?

PR #1554, opened 2019-02-26
Fix devrun.sh

walaura:
## Why are you doing this?
For my convenience really.

This fixes `devrun.sh`


johnduffell:
👍 
are there any Readme/docs should also be updated?

walaura:
@johnduffell SO MANY I’ll garden the wiki today

PR #1556, opened 2019-02-26
Fix the integration tests

rupertbates:
## Why are you doing this?
Since moving to a mono repo there have been some failing integration tests in the support-workers project. This fixes them.

### DIFF
--- a/support-workers/src/main/scala/com/gu/acquisitions/AcquisitionServiceBuilder.scala
+++ b/support-workers/src/main/scala/com/gu/acquisitions/AcquisitionServiceBuilder.scala
@@ -1,4 +1,4 @@
-package com.gu.ophan
+package com.gu.acquisitions
 
 import com.amazonaws.auth.{AWSCredentialsProviderChain, InstanceProfileCredentialsProvider}
 import com.amazonaws.auth.profile.ProfileCredentialsProvider
@@ -7,9 +7,9 @@ import com.gu.config.Configuration
 import com.gu.okhttp.RequestRunners
 import com.gu.support.config.Stages
 
-object AcquisitionService {
+object AcquisitionServiceBuilder {
 
-  def apply(isTestService: Boolean): com.gu.acquisition.services.AcquisitionService =
+  def build(isTestService: Boolean): com.gu.acquisition.services.AcquisitionService =
     if (isTestService) {
       com.gu.acquisition.services.MockAcquisitionService
     } else {
--- a/support-workers/src/main/scala/com/gu/support/workers/lambdas/SendAcquisitionEvent.scala
+++ b/support-workers/src/main/scala/com/gu/support/workers/lambdas/SendAcquisitionEvent.scala
@@ -18,7 +18,7 @@ import ophan.thrift.{event => thrift}
 import scala.concurrent.ExecutionContext.Implicits.global
 
 class SendAcquisitionEvent(serviceProvider: ServiceProvider = ServiceProvider)
-    extends ServicesHandler[SendAcquisitionEventState, Unit] {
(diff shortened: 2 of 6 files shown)


support-workers/src/main/scala/com/gu/acquisitions/AcquisitionServiceBuilder.scala
```diff
@@ -7,9 +7,9 @@ import com.gu.config.Configuration
 import com.gu.okhttp.RequestRunners
 import com.gu.support.config.Stages
 
-object AcquisitionService {
+object AcquisitionServiceBuilder {
```
rupertbates: Having this object called `AcquisitionService` was confusing because it actually constructs a `com.gu.acquisition.services.AcquisitionService`. I think this name better describes its actual purpose

support-workers/src/main/scala/com/gu/support/workers/lambdas/SendAcquisitionEvent.scala
```diff
@@ -18,7 +18,7 @@ import ophan.thrift.{event => thrift}
 import scala.concurrent.ExecutionContext.Implicits.global
 
 class SendAcquisitionEvent(serviceProvider: ServiceProvider = ServiceProvider)
-    extends ServicesHandler[SendAcquisitionEventState, Unit] {
+    extends ServicesHandler[SendAcquisitionEventState, Unit](serviceProvider) {
```
rupertbates: This class was not passing the `ServiceProvider` it receives in its constructor through to its super class. This made it impossible to mock services in tests.

support-workers/src/main/scala/com/gu/support/workers/lambdas/ServicesHandler.scala
```diff
@@ -9,7 +9,7 @@ import io.circe.{Decoder, Encoder}
 import scala.concurrent.duration._
 import scala.concurrent.{ExecutionContext, Future}
 
-abstract class ServicesHandler[T <: StepFunctionUserState, R](servicesProvider: ServiceProvider = ServiceProvider, d: Option[Duration] = None)(
+abstract class ServicesHandler[T <: StepFunctionUserState, R](servicesProvider: ServiceProvider, d: Option[Duration] = None)(
```
rupertbates: I have removed the default value for the `servicesProvider` constructor parameter to force overriding classes to pass this on and hopefully prevent the same bug happening again.

support-workers/src/main/scala/com/gu/acquisitions/AcquisitionServiceBuilder.scala
johnduffell: scala convention is the apply method in the companion object is the constructor.
However I would agree if it's more than just a constructor (i.e. it's doing some decision making) then it makes sense to be something in its own right.
The latter appears to be the case here by the looks, so your change makes sense! 😄

support-workers/src/main/scala/com/gu/support/workers/lambdas/ServicesHandler.scala
johnduffell: totally separate comment to this PR, but I'm not a massive fan of inheritance and overriding as dependencies are less explicit and harder to not have.  However I do see that it makes a stuff available "for free" without as much boilerplate.  It's the old "is-a vs has-a" debate or "composition vs inheritance".
In this case it's clear that there's very little in the `ServicesHandler` so things are not bad at the moment, as long as it doesn't expand.

johnduffell:
great! thanks for fixing.  just some informational comments 👍

PR #1559, opened 2019-02-26
Landing page copy test

jranks123:
## Why are you doing this?

In this test we are testing some longer, more persuasive copy on the Landing Page

|Control| Variant|
|----|----|
|image|image|

|Control (AUS)| Variant (AUS)|
|----|----|
|image|image|

This also removes the copy from the server-side rendered placeholder, so that we don't have to duplicate the ab test logic there. 

Note: I deleted contributions.js, as all it's contents were only being used in one place (plus I had to change it to a jsx file so that I could make the `contributionCopy` of the type `React.Element`, and this was causing a false error on importing it for some reason that I didn't have the time nor inclination to figure out


ionamckendrick:
Looks great!

support-frontend/assets/helpers/abTests/abtestDefinitions.js
```diff
@@ -5,6 +5,7 @@ import { annualAmountsLower, annualAmountsFive, annualAmountsOther } from './ann
 // ----- Tests ----- //
 
 export type FrequencyTabsTestVariant = 'control' | 'sam' | 'mas' | 'notintest';
+export type LandingPageCopyTestVariant = 'control' | 'help' | 'open' | 'notintest';
```
johnduffell: open isn't needed (as discussed)

support-frontend/assets/pages/new-contributions-landing/components/ContributionFormContainer.jsx
```diff
@@ -79,6 +82,85 @@ const mapDispatchToProps = (dispatch: Function) => ({
...
+  tickerJsonUrl?: string,
+};
+
+const countryGroupSpecificDetails: (variant: LandingPageCopyTestVariant) => {
+  [CountryGroupId]: CountryMetaData
+} = (variant: LandingPageCopyTestVariant) => {
+  switch (variant) {
+    case 'control':
+    case 'notintest':
+    default:
```
johnduffell: it's fine to have default above case 'help' in js?  I've got my "C" brain on otherwise!

support-frontend/assets/pages/new-contributions-landing/contributionsLanding.scss
```diff
@@ -432,6 +436,14 @@ footer[role="contentinfo"] a {
     }
   }
 
+  .gu-content__blurb--help-variant {
+    max-width: 404px;
+    p {
```
johnduffell: when I was doing the css strpping back, I noticed a few where we were having nested element names e.g. h1, p, rather than class names, so I removed.  Is it normal to have them in or not?

johnduffell:
👍 looks good, thanks!
one suggestion is when you raise a PR, delete any parts of the template that are not relevant/used, otherwise it just adds clutter.

support-frontend/assets/pages/new-contributions-landing/components/ContributionFormContainer.jsx
jranks123: I'll switch them around for idiomatic reasons, but this does work!

PR #1561, opened 2019-02-26
Support for creating print subscriptions: part 1

lmath:
## Why are you doing this?
New print checkout 😄 

We will need to be able to create print subscriptions via `support-workers`. 

This PR is the first part of this. This will make it possible to a create print subscription via `support-workers`, but with some assumptions/PRs to follow:
- https://github.com/guardian/support-frontend/pull/1560 <- using the first delivery date properly will follow in a future PR after this one is merged
- thank you and failure emails
- this code currently expects that billing and delivery address are the same
- acquisition events send 0 for the purchase amount

[**Trello Card**](https://trello.com/c/9YE6DIdo/1959-update-support-workers-to-enable-it-to-create-print-subscriptions)

## Changes

* It is now possible to create a paper subscription via support-workers if the suitable information is passed via ProductType (see: https://github.com/guardian/support-frontend/blob/master/support-models/src/main/scala/com/gu/support/workers/Products.scala#L36)

### DIFF
--- a/support-frontend/test/codecs/ProductTypeDecoderTest.scala
+++ b/support-frontend/test/codecs/ProductTypeDecoderTest.scala
@@ -1,0 +1,5 @@
+package codecs
+
+class ProductTypeDecoderTest {
+
+}
--- a/support-frontend/test/codecs/ProductTypeDecoderTest.scala
+++ b/support-frontend/test/codecs/ProductTypeDecoderTest.scala
@@ -1,5 +1,104 @@
 package codecs
 
-class ProductTypeDecoderTest {
+import com.gu.i18n.Currency
+import com.gu.support.catalog.{Collection, Sunday}
+import com.gu.support.workers.ProductType._
+import com.gu.support.workers._
+import io.circe.parser.parse
+import org.scalatest.{MustMatchers, WordSpec}
 
+class ProductTypeDecoderTest extends WordSpec with MustMatchers {
+
+  "ProductTypeDecoder" should {
+    "decode a contribution" in {
+      val json =
+        """
+          |{
+          |  "amount":10,
+          |  "currency":"GBP",
+          |  "billingPeriod": "Monthly"
+          |}
+        """
+          .stripMargin
+
+      val parsedJson = parse(json).toOption.get
+
+      val productType: ProductType = decodeProduct.decodeJson(parsedJson).right.get
+
+      val contributionMaybe = productType match {
+        case c: Contribution => Some(c)
+        case _ => None
+      }
+
+      contributionMaybe.get.amount mustBe 10
+      contributionMaybe.get.billingPeriod mustBe Monthly
+    }
+
+    "decode a paper sub" in {
+      val json =
+        """
+          |{
+          |  "currency":"GBP",
+          |  "billingPeriod": "Monthly",
+          |  "productOptions":"Sunday",
+          |  "fulfilmentOptions": "Collection"
+          |}
+        """
+          .stripMargin
+
+      val parsedJson = parse(json).toOption.get
+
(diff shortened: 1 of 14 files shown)


support-frontend/test/codecs/ProductTypeDecoderTest.scala
```diff
@@ -0,0 +1,104 @@
+package codecs
+
+import com.gu.i18n.Currency
+import com.gu.support.catalog.{Collection, Sunday}
+import com.gu.support.workers.ProductType._
+import com.gu.support.workers._
+import io.circe.parser.parse
+import org.scalatest.{MustMatchers, WordSpec}
+
+class ProductTypeDecoderTest extends WordSpec with MustMatchers {
```
lmath: I added this after I needed to swap the order in `ProductType`'s decoder because my changes to support print subs were initially creating digital packs 🙈

rupertbates: It would be good to use the `SerialisationTestHelpers` trait for these tests. See an example in: https://github.com/guardian/support-frontend/blob/79f05140561d2e1fb571b90871654711a3aec96e/support-models/src/test/scala/com/gu/support/promotions/SerialisationSpec.scala#L12

rupertbates: 💯 for adding a test for this though!

support-workers/src/main/scala/com/gu/support/workers/lambdas/FailureHandler.scala
```diff
@@ -38,8 +38,9 @@ class FailureHandler(emailService: EmailService) extends FutureHandler[FailureHa
   private def sendEmail(state: FailureHandlerState) = {
     val emailFields = state.product match {
       case c: Contribution => FailedContributionEmailFields(email = state.user.primaryEmailAddress, IdentityUserId(state.user.id))
+      //TODO!!! will currently send nothing
```
rupertbates: Is this TODO in the wrong place or is github just displaying things weirdly?

lmath: That's also true (digital pack failure emails not being sent) but I admit not particularly relevant to this PR.

support-frontend/test/codecs/ProductTypeDecoderTest.scala
lmath: Thanks for pointing me to that. Definitely nicer. https://github.com/guardian/support-frontend/pull/1561/commits/5f2d09d550087fdb9d3c54afed5310b8b8b6b8a3

support-workers/src/test/scala/com/gu/zuora/Fixtures.scala
```diff
@@ -54,6 +57,21 @@ object Fixtures {
     Subscription(date, date, date)
   )
 
+  val touchpointEnvironment = TouchPointEnvironments.fromStage(Configuration.stage)
+  def getRatePlans[T <: Product](product: T): Seq[ProductRatePlan[Product]] = product.ratePlans.getOrElse(touchpointEnvironment, Nil)
+  val everydayHDProductRatePlanId = getRatePlans(catalog.Paper).find(x => x.fulfilmentOptions == HomeDelivery && x.productOptions == Everyday) map (_.id)
```
rupertbates: I think you could replace this with `catalog.Paper.getProductRatePlan(touchpointEnvironment, Monthly, HomeDelivery, Everyday)`

lmath: Good idea. Thanks.

PR #1565, opened 2019-02-27
Fix devrun again

walaura:
## Why are you doing this?
it wasn't working before

### DIFF
--- a/support-frontend/devrun.sh
+++ b/support-frontend/devrun.sh
@@ -1,0 +1,5 @@
+#!/usr/bin/env bash
+
+yarn devrun &
+yarn storybook &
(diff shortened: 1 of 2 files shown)


support-frontend/devrun.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+yarn devrun &
+yarn storybook &
+cd ..; sbt "project support-frontend" devrun
```
regiskuckaertz: I'm slightly dubious of the `cd ..` statement, but I'm no bash expert.

walaura: it's intentional (sbt needs to run from the root) but i'm not sure if it can be improved)

PR #1568, opened 2019-02-27
Added click tracking to the required for fields explainer details.

rjbarbour-g:
Added css hover state to payment frequency tabs and amount buttons with custom colour matching the 'Contribute...' button.

## Why are you doing this?

We think that some users don't realise that they can change the payment frequency and that the hover state will provide a clearer affordance.

We want to track how may users read the required fields explainer.

[**Trello Card for hover**](https://trello.com/c/oZEVCAot/328-implement-hover-states-for-product-type-and-amount-buttons)
[**Trello Card for tracking**](https://trello.com/c/148LzjFX/330-put-tracking-code-in-place-for-expandable-required-fields)

## Screenshots

image

### DIFF
--- a/support-frontend/assets/pages/new-contributions-landing/components/ContributionFormFields.jsx
+++ b/support-frontend/assets/pages/new-contributions-landing/components/ContributionFormFields.jsx
@@ -18,7 +18,7 @@ import {
   emailRegexPattern,
 } from 'helpers/formValidation';
 import { type UserTypeFromIdentityResponse } from 'helpers/identityApis';
-
+import {trackComponentClick} from 'helpers/tracking/ophanComponentEventTracking';
 import { NewContributionState } from './ContributionState';
 import { NewContributionTextInput } from './ContributionTextInput';
 import { MustSignIn } from './MustSignIn';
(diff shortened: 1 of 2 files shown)


walaura:
wild idea: have you thought of ab testing underlining the tab titles? not the most aesthetic solution but would be a pretty quick way to prove the hypothesis

ionamckendrick:
@walaura So I already have a second round of testing planned where we highlight the tabs in different ways. Primarily I want them to be more representative of the old school address book tabs (do you know what I mean?) – come chat to me today and I'll show you my designs. Rather than an underline, more of an over line ... but you'll hopefully like what I'm suggesting as it's similar to what you're suggesting!

support-frontend/assets/pages/new-contributions-landing/components/ContributionFormFields.jsx
```diff
@@ -164,8 +164,21 @@ function FormFields(props: PropTypes) {
...
         showRequiredLabel={showRequiredLabel}
       />
-      <details className={showRequiredLabel ? 'form-fields__required-explainer' : 'hidden'}>
-        <summary >
+      <details
+        className={showRequiredLabel ? 'form-fields__required-explainer' : 'hidden'}
+      >
+        <summary
+          onClick={(event) => {
+            if (event.detail !== 0) {
```
jranks123: Could you add a comment here explaining why we are checking this? Maybe with a link to the stack overflow question

PR #1579, opened 2019-02-28
use `yarn stat` to generate a js bundle size chart for your enjoyment

walaura:
## Why are you doing this?
I lifted this from `identity-frontend`

screenshot 2019-02-28 at 5 26 34 pm

What this exactly does is twofold – first, it makes webpack output a json with all the compilation stats, then adds a tool to turn that json into a more actionable viz of what's in each bundle.

### DIFF
--- a/support-frontend/package.json
+++ b/support-frontend/package.json
@@ -13,7 +13,7 @@
     "webpack-dev-server": "webpack-dev-server --config webpack.dev.js --port 9211 --public support.thegulocal.com",
     "devrun": "npm-run-all clean webpack-dev-server",
     "test": "echo 'Running JS tests' && jest",
-    "stat": "webpack-bundle-analyzer target/web/build-npm/stats.json",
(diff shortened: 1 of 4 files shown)


support-frontend/package.json
```diff
@@ -13,6 +13,7 @@
     "webpack-dev-server": "webpack-dev-server --config webpack.dev.js --port 9211 --public support.thegulocal.com",
     "devrun": "npm-run-all clean webpack-dev-server",
     "test": "echo 'Running JS tests' && jest",
+    "stat": "webpack-bundle-analyzer public/compiled-assets/stats.json",
```
walaura:
```suggestion
    "stat": "yarn build-prod && webpack-bundle-analyzer public/compiled-assets/stats.json",
```

johnduffell:
nice, so that measnw e can type "yarn stat" to get the chart?  we should totally add to the readme or wiki!

walaura:
@johnduffell yep!! it's even spin up it's own local server and open chrome it's a oneliner

PR #1580, opened 2019-03-01
Fix styling on contribute page

johnduffell:
## Why are you doing this?

In the [previous PR](https://github.com/guardian/support-frontend/pull/1577) I added back some of the styling needed for the thank you page, but some of the new styling I had to move off to avoid affecting it.  I moved it to the gu-content__content-contributions class but forgot to add that class to the original non thank you page.  That meant the width wasn't limited at higher breakpoints

This PR fixes that.

|before|after|
|---|---|
|image|image|

320
image
480
image
740 (this looks a bit squashed but it doesn't seem to be worse than before)
image
1250
image
2560
image

[**Trello Card**](https://trello.com/c/i23v7T5J/435-bug-blurb-too-far-to-the-right)

@jranks123


johnduffell:
this commit should have also been in this PR.....
https://github.com/guardian/support-frontend/commit/436f08bd75f658ed0e793a8e98978597889efc29

PR #1593, opened 2019-03-04
Get the paper checkout to call the create endpoint to create paper subs.

lmath:
## Why are you doing this?
So that using the paper checkout will result in the ajax call `POST /subscribe/create`, which will in turn create a paper subscription via `support-workers`

[**Trello Card**](https://trello.com/c/s3bTXdKj/2281-get-the-paper-checkout-to-call-the-create-endpoint-to-create-paper-subs)

## Changes
* `onPaymentAuthorised` now makes a regular payment request 

Note: currently copies the same address as both billing and delivery address.

### DIFF
--- a/support-frontend/assets/pages/paper-subscription-checkout/helpers/paymentProviders.js
+++ b/support-frontend/assets/pages/paper-subscription-checkout/helpers/paymentProviders.js
@@ -6,16 +6,76 @@ import {
   setupStripeCheckout,
 } from 'helpers/paymentIntegrations/stripeCheckout';
 import { type IsoCountry } from 'helpers/internationalisation/country';
-import type { PaymentAuthorisation } from 'helpers/paymentIntegrations/readerRevenueApis';
+import { type PaymentAuthorisation } from 'helpers/paymentIntegrations/readerRevenueApis';
 import {
   type PaymentResult,
+  type RegularPaymentRequest,
+  postRegularPaymentRequest, regularPaymentFieldsFromAuthorisation,
 } from 'helpers/paymentIntegrations/readerRevenueApis';
 import { type Dispatch } from 'redux';
 import { openDirectDebitPopUp } from 'components/directDebit/directDebitActions';
 import { finalPrice as paperFinalPrice } from 'helpers/productPrice/paperProductPrices';
+import { routes } from 'helpers/routes';
+import { getQueryParameter } from 'helpers/url';
+import { getOphanIds, getSupportAbTests } from 'helpers/tracking/acquisitions';
 import { type State, setSubmissionError, setFormSubmitted, type Action, setStage } from '../paperSubscriptionCheckoutReducer';
 
-function onPaymentAuthorised(paymentAuthorisation: PaymentAuthorisation, dispatch: Dispatch<Action>) {
+function buildRegularPaymentRequest(state: State, paymentAuthorisation: PaymentAuthorisation): RegularPaymentRequest {
+  const { currencyId, countryId } = state.common.internationalisation;
+  const {
+    firstName,
+    lastName,
+    billingAddressLine1,
+    billingAddressLine2,
+    billingTownCity,
+    billingPostcode,
+    email,
+    telephone,
+  } = state.page.checkout;
+
+  const product = {
+    currency: currencyId,
+    billingPeriod: 'Monthly',
+    fulfilmentOptions: state.page.checkout.fulfilmentOption,
+    productOptions: state.page.checkout.productOption,
+  };
+
+  console.log(product);
+
+  const paymentFields = regularPaymentFieldsFromAuthorisation(paymentAuthorisation);
+
+  return {
+    firstName,
+    lastName,
+    billingAddress: {
+      lineOne: billingAddressLine1,
+      lineTwo: billingAddressLine2,
+      city: billingTownCity,
+      state: null,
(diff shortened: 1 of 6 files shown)


support-frontend/assets/pages/paper-subscription-checkout/helpers/paymentProviders.js
```diff
@@ -6,16 +6,76 @@ import {
...
 import { finalPrice as paperFinalPrice } from 'helpers/productPrice/paperProductPrices';
+import { routes } from 'helpers/routes';
+import { getQueryParameter } from 'helpers/url';
+import { getOphanIds, getSupportAbTests } from 'helpers/tracking/acquisitions';
+import { Monthly } from 'helpers/billingPeriods';
+
 import { type State, setSubmissionError, setFormSubmitted, type Action, setStage } from '../paperSubscriptionCheckoutReducer';
 
-function onPaymentAuthorised(paymentAuthorisation: PaymentAuthorisation, dispatch: Dispatch<Action>) {
+function buildRegularPaymentRequest(state: State, paymentAuthorisation: PaymentAuthorisation): RegularPaymentRequest {
```
rupertbates: I'm a bit concerned at this file being a copy of the one in `/digital-subscription-checkout`. It feels like we should make that file generic enough to handle both products and put it a top level file.

support-frontend/assets/pages/paper-subscription-checkout/helpers/paymentProviders.js
```diff
@@ -30,7 +90,17 @@ function onPaymentAuthorised(paymentAuthorisation: PaymentAuthorisation, dispatc
   };
 
   dispatch(setFormSubmitted(true));
-  Promise.resolve({ paymentStatus: 'success' }).then(handleSubscribeResult);
+
+  const data = buildRegularPaymentRequest(state, paymentAuthorisation);
+
+  postRegularPaymentRequest(
+    routes.digitalSubscriptionCreate,
```
rupertbates: If this route is used by both products (and it's good if it is), then it should change to `subscriptionCreate`

support-frontend/assets/pages/paper-subscription-checkout/helpers/paymentProviders.js
walaura: @rupertbates I'd say for now let's copypaste them and before weekly (that's 3 already) let's work to identify the areas that are common enough to be extracted off. (Same story with the frontend bits such as the postcode finder)

support-frontend/assets/pages/paper-subscription-checkout/helpers/paymentProviders.js
```diff
@@ -6,16 +6,76 @@ import {
...
+    fulfilmentOptions: state.page.checkout.fulfilmentOption,
+    productOptions: state.page.checkout.productOption,
+  };
+
+  const paymentFields = regularPaymentFieldsFromAuthorisation(paymentAuthorisation);
+
+  return {
+    firstName,
+    lastName,
+    billingAddress: {
```
walaura: this is a stylistic nit, but could we extract these as variables a bit higher up the function (like `product` is) this return looks a bit hefty to me

support-frontend/assets/pages/paper-subscription-checkout/helpers/paymentProviders.js
```diff
@@ -30,7 +90,17 @@ function onPaymentAuthorised(paymentAuthorisation: PaymentAuthorisation, dispatc
...
+
+  const data = buildRegularPaymentRequest(state, paymentAuthorisation);
+
+  postRegularPaymentRequest(
+    routes.digitalSubscriptionCreate,
+    data,
+    state.common.abParticipations,
+    state.page.csrf,
+    () => {},
+    () => {},
```
walaura: what are these two noops?

support-frontend/assets/pages/paper-subscription-checkout/helpers/paymentProviders.js
lmath: oh yes I will fix that

support-frontend/assets/pages/paper-subscription-checkout/helpers/paymentProviders.js
lmath: They're both only used by the contributions checkout, something to do with setting a guest account token, and a thank-you page state. This is another candidate for a refactor, but probably not in the scope of this PR.

rupertbates: This is another place where the original code is written with a few contribution specific assumptions which don't apply to subs. The first supports guest checkout and the second sets the thank you page stage. The second should probably be refactored to apply to all checkouts but not in this PR

support-frontend/assets/pages/paper-subscription-checkout/helpers/paymentProviders.js
lmath: I see your concern is about this whole file @rupertbates , but I wanted to stop and think about what could be extracted specifically for this `buildRegularPaymentRequest` that is now in three places since this PR adds the third. But since it's roughly a mapper from the `CheckoutState` for each checkout to a `RegularPaymentRequest` I'm not sure how to do that without enforcing structure onto each `CheckoutState`.

PR #1594, opened 2019-03-04
TY pages

walaura:
## Why are you doing this?

Make users feel warm and fuzzy

[**Trello Card**](https://trello.com/c/TKmfAUMq/2241-print-checkout-ty-page)

## Changes

* TY page
* TY pending page
* Asynchronously load the mkt consents (for fun & profit)

## Screenshots

screenshot 2019-03-04 at 3 58 37 pm

## Wait what do you mean by asyncronously load

Ah yes so I've experimentally copied the islands work from dotcom for us to use. What this means is that instead of bundling the marketing consent (14kb!) in the main JS bundle it will get loaded from a chunk before it renders. 

This means the page can load faster as it has less javascript to download & parse on the first go. Obviously if this works okay we can use it for more parts of the pages that are expensive to download

Compare digital (no splitting) with paper (split):
screenshot 2019-03-04 at 4 09 00 pm
screenshot 2019-03-04 at 4 08 48 pm


support-frontend/assets/pages/paper-subscription-checkout/components-ty/thankYou.jsx
```diff
@@ -0,0 +1,138 @@
...
+      </Content>
+      <Content>
+        <Text>
+          <SansParagraph>
+            You can manage your subscription by visiting our Manage section or accessing it via your Guardian account.
+          </SansParagraph>
+        </Text>
+      </Content>
+      <Content>
+        <Asyncronously loader={import('./marketingConsentContainer')}>
```
rupertbates: Nice!

PR #1596, opened 2019-03-05
Refactor paypal button

rupertbates:
## Why are you doing this?
All the code which deals with PayPal express checkout - the type used by recurring contributions and subscriptions - used to sit in the contribution landing page folder, and was very specific to contributions, for instance the `PayPalRecurringButton` component required a `contributionType` prop from which it would work out the amount and the billing period to pass to PayPal. 
This PR makes the PayPal related code more generic so that it can be used for subscriptions as well,  and also moves the component up to a top level component.


support-frontend/assets/components/paypalExpressButton/PayPalExpressButton.jsx
```diff
@@ -20,13 +20,9 @@ type PropTypes = {|
...
   formClassName: string,
   isTestUser: boolean,
-  contributionType: ContributionType,
-  setupRecurringPayPalPayment: (
-    resolve: string => void,
-    reject: Error => void,
-    IsoCurrency, CsrfState,
-    regularContributionType: ContributionType
-  ) => void,
+  amount: number,
```
rupertbates: Rather than passing the contribution type into this component so that it can work out the amount to charge and the billing period we now pass these in explicitly.

support-frontend/assets/components/paypalExpressButton/PayPalExpressButton.jsx
```diff
@@ -46,7 +42,7 @@ type PropTypes = {|
 // 1. Loading this iframe is an expensive operation which causes an obvious visual re-render
 // 2. We don't want to have to re-bind handlers which interact with the iframe
 //    (e.g. the handler bound in getPayPalOptions)
-export class PayPalRecurringButton extends React.Component<PropTypes> {
+export class PayPalExpressButton extends React.Component<PropTypes> {
```
rupertbates: I've renamed the component to `PayPalExpressButton` to reflect the fact that it is no longer used only by recurring contributions. PayPal Express is the actual PayPal api that we are using.

support-frontend/assets/components/paypalExpressButton/PayPalExpressButton.jsx
walaura: how do we determine the currency?

support-frontend/assets/pages/new-contributions-landing/components/ContributionSubmit.jsx
```diff
@@ -6,21 +6,19 @@ import type { Csrf as CsrfState } from 'helpers/csrf/csrfReducer';
...
 import { hiddenIf } from 'helpers/utilities';
+import { setupRecurringPayPalPayment, type SetupPayPalRequestType } from 'helpers/paymentIntegrations/payPalRecurringCheckout';
+import type { BillingPeriod } from 'helpers/billingPeriods';
 import { type State } from '../contributionsLandingReducer';
-import { PayPalRecurringButton } from './PayPalRecurringButton';
-import {
-  sendFormSubmitEventForPayPalRecurring,
-  setupRecurringPayPalPayment,
-} from '../contributionsLandingActions';
+import { PayPalExpressButton } from '../../../components/paypalExpressButton/PayPalExpressButton';
```
walaura:
```suggestion
import { PayPalExpressButton } from 'components/paypalExpressButton/PayPalExpressButton';
```
nit

support-frontend/assets/components/paypalExpressButton/PayPalExpressButton.jsx
rupertbates: It's a prop, github just isn't showing it.

PR #1597, opened 2019-03-05
Enable ticker for AU

tomrf1:
## Why are you doing this?
We are reviving the ticker for an upcoming AU campaign

## Changes

1. Updated ticker styling
2. Ticker can be enabled by adding `ticker=true` in the url query string
3. Adding `frontline-campaign=true` to the url query string adds custom campaign to the form (below the ticker)

## Screenshots
picture 4


tomrf1:
I've updated this PR with the latest design for the AU campaign. It makes use of 2 recent changes:
https://github.com/guardian/support-frontend/pull/1605
https://github.com/guardian/support-frontend/pull/1602

@ionamckendrick

PR #1600, opened 2019-03-05
optimise newsletter svg and lazy load it

johnduffell:
## Why are you doing this?

We are focussing on performance.  The JS download and execute cost 50% of the page render time.  One of the things padding out the JS is a newsletter SVG that is not even used on the first page.
This is a simple PR to just compress the SVG with https://jakearchibald.github.io/svgomg/ (which almost halves the gzipped size with no perceptible quality difference) and moves it to a file that's downloaded on demand when it's needed.

|unoptimised svg|optimised svg|
|---|---|
|image|image|

file downloaded only when it was rendered with Content-Length: 3849
image

[**Trello Card**](https://trello.com/c/V6TpIxwX/353-js-remove-huge-svg-from-newslettersjsx)

@jranks123


johnduffell:
this has apparently reduced the bundle from 349228 to 331878 a reduction to 95% (103226 to 95772 or 93% for the gzipped size)

johnduffell:
although that figure also includes Rupert's PR by mistake but it shouldn't affect it https://github.com/guardian/support-frontend/pull/1599

johnduffell:
although that figure also includes Rupert's PR by mistake but it shouldn't affect it https://github.com/guardian/support-frontend/pull/1599

PR #1601, opened 2019-03-05
Checkout form refactor

rupertbates:
## Why are you doing this?
This PR starts the process of breaking up the `CheckoutForm` component into multiple sub-components, this will:
* Make these components reusable across all subscriptions checkouts, reduce copy and paste code and mean that we only have to make changes/fix bugs once rather than in multiple files
* Reduce the size of the `CheckoutForm` component for readability and ease of maintenance

The first two components I have refactored are the payment method selector section:
screen shot 2019-03-05 at 12 39 19

and the submit buttons: 
screen shot 2019-03-05 at 16 54 05
screen shot 2019-03-05 at 16 53 39

These both have some relatively complex logic about what they should show and when so we definitely want to make sure that that logic is only defined in one place for all subs checkouts.


support-frontend/assets/components/checkoutForm/checkoutForm.jsx
```diff
@@ -36,7 +36,7 @@ the top level form itself
 */
 
 type FormPropTypes = {
-  children: ChildrenArray<Element<typeof FormSection>>,
+  children: ChildrenArray<Element<any>>,
```
rupertbates: This change is needed to allow us to include `FormSections` inside other components as we do with `PaymentMethodSelector`

support-frontend/assets/pages/digital-subscription-checkout/components/checkoutForm.jsx
```diff
@@ -312,69 +309,30 @@ function CheckoutForm(props: PropTypes) {
...
+                onPaymentAuthorised={props.onPaymentAuthorised}
+                csrf={props.csrf}
+                currencyId={props.currencyId}
+                payPalHasLoaded={props.payPalHasLoaded}
+                formIsValid={props.formIsValid}
+                validateForm={props.validateForm}
+                isTestUser={props.isTestUser}
+                setupRecurringPayPalPayment={props.setupRecurringPayPalPayment}
+                amount={props.amount}
+                billingPeriod={props.billingPeriod}
```
walaura: might be cleaner to spread props here?

support-frontend/assets/components/subscriptionCheckouts/subscriptionSubmitButtons.jsx
```diff
@@ -0,0 +1,69 @@
...
+
+// ----- Render ----- //
+
+
+function SubscriptionSubmitButtons(props: PropTypes) {
+  return (
+    <div>
+      <div
+        id="component-paypal-button-checkout"
+        className={hiddenIf(props.paymentMethod !== 'PayPal', 'component-paypal-button-checkout')}
```
walaura: can this be a ternary instead of using css?

```jsx
<div>
  {props.paymentMethod === 'PayPal' ? <PayPalExpressButton  ... /> :  < Button ... />}
</div>
```

rupertbates: No because the PayPal loading is expensive and involves state changes so we only want to do it once. I've added a comment to explain it.

support-frontend/assets/pages/digital-subscription-checkout/components/checkoutForm.jsx
rupertbates: Actually there are so many props in `PropTypes` that are not in `SubscriptionSubmitButtons` I don't think this will work. Unless you have a cunning way of doing it?

PR #1602, opened 2019-03-05
Hide contribution type tabs if only one-off is available

tomrf1:
## Why are you doing this?
We sometimes only want to allow one-off contributions. In this case, there is no need to show the contribution type tabs.

## Screenshots
Only one-off:
picture 6

Others:
picture 7

### DIFF
--- a/support-frontend/assets/pages/new-contributions-landing/components/ContributionTypeTabs.jsx
+++ b/support-frontend/assets/pages/new-contributions-landing/components/ContributionTypeTabs.jsx
@@ -55,11 +55,17 @@ const mapDispatchToProps = (dispatch: Function) => ({
 // ----- Render ----- //
 
 function ContributionTypeTabs(props: PropTypes) {
+  const contributionTypes = getValidContributionTypes(props.dropMonthlyVariant);
+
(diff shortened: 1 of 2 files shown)


support-frontend/assets/pages/new-contributions-landing/components/ContributionTypeTabs.jsx
```diff
@@ -55,11 +55,17 @@ const mapDispatchToProps = (dispatch: Function) => ({
 // ----- Render ----- //
 
 function ContributionTypeTabs(props: PropTypes) {
+  const contributionTypes = getValidContributionTypes(props.dropMonthlyVariant);
+
+  if (contributionTypes.length === 1 && contributionTypes[0] === 'ONE_OFF') {
```
jranks123: Would `if (contributionTypes === ['ONE_OFF'])` work?

tomrf1: unfortunately no, you'd have to [implement equality check](https://stackoverflow.com/a/14853974)

PR #1605, opened 2019-03-06
Allow optional form message

tomrf1:
## Why are you doing this?
This change makes it possible to add a message at the top of the form component.
This will be used for the upcoming AU campaign, but for now nothing uses this feature

## Screenshots
**With contribution type tabs:**
picture 9

**Without contribution type tabs:**
picture 10

### DIFF
--- a/support-frontend/assets/pages/new-contributions-landing/components/ContributionFormContainer.jsx
+++ b/support-frontend/assets/pages/new-contributions-landing/components/ContributionFormContainer.jsx
@@ -120,6 +120,7 @@ const defaultCountryGroupSpecificDetails = {
   AUDCountries: {
     ...defaultHeaderCopyAndContributeCopy,
     headerCopy: australiaHeadline,
+    formMessage: 'Make a contribution',
   },
   International: defaultHeaderCopyAndContributeCopy,
   NZDCountries: defaultHeaderCopyAndContributeCopy,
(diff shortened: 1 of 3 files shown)


support-frontend/assets/pages/new-contributions-landing/components/ContributionFormContainer.jsx
```diff
@@ -145,6 +145,8 @@ export type CountryMetaData = {
   headerClasses?: string,
   // URL to fetch ticker data from. null/undefined implies no ticker
   tickerJsonUrl?: string,
+  // Optional message to display at the top of the form
+  formMessage?: React$Element<string>,
```
johnduffell: does this come from a file in s3 or something?

tomrf1: Good question. The `CountryMetaData` type is actually used in this file - I've just pushed a commit to clarify where it's used

tomrf1: So for the AU campaign the `formMessage` will be set in this file

johnduffell:
ah I see, hard coded tests basically, looks good thanks, 👍

tomrf1:
@johnduffell yes it's a terrible way to test, but I'm about to create a PR for the campaign which will make use of this feature!

PR #1612, opened 2019-03-06
inline Javascript into the page

johnduffell:
## Why are you doing this?

We are trying to make perf improvements to help people access the page quicker.  This means they are less likely to leave the page before it loads so they are more likely to contribute.

I noticed that with a slow network connection "Fast 3G in chrome", the JS takes 1.9 seconds to download.  Then there is a constant amount of time to get the form on the screen.  See the diagram:
image

However if we inline all the JS the connection just gets a clear run from when the main body starts downloading until all the JS is loaded, meaning that there is little contention.
image

As you can see the js is fully downloaded 1 second earlier when it's inlined.

Lighthouse difference is small

|before|after|
|---|---|
|image|image|

This is not an AB test, it would be similar difficulty to doing the SSR test, not sure if it's better to just do it, test it, or not do it at all?

The downside could be that people have to download the JS regardless of whether it's been cached.  Although that is flushed out at every deploy anyway, and most people presumably only hit the page once.

[**Trello Card**](https://trello.com/c/8PLMDwNB/358-try-inlining-main-js-bundle-into-html)

### DIFF
--- a/support-frontend/app/controllers/Application.scala
+++ b/support-frontend/app/controllers/Application.scala
@@ -114,7 +114,10 @@ class Application(
   private def contributionsHtml(countryCode: String, idUser: Option[IdUser])
     (implicit request: RequestHeader, settings: AllSettings) = {
 
-    val css = CSSElementForStage(assets.getFileContentsAsHtml, stage, RefPath("newContributionsLandingPageStyles.css"))
+    val elementForStage = CSSElementForStage(assets.getFileContentsAsHtml, Stages.CODE)_
+    val css = elementForStage(RefPath("newContributionsLandingPageStyles.css"))
+
+    val js = elementForStage(RefPath("newContributionsLandingPage.js"))
 
     val preload = List("GuardianTextSans-Regular", "GuardianTextSans-Medium", "GHGuardianHeadline-Bold").map { name =>
       Preload(s"//pasteup.guim.co.uk/webfonts/1.0.0/noalts-not-hinted/$name.woff2", "font", "font/woff2")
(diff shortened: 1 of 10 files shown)


support-frontend/app/controllers/Application.scala
```diff
@@ -114,7 +114,10 @@ class Application(
   private def contributionsHtml(countryCode: String, idUser: Option[IdUser])
     (implicit request: RequestHeader, settings: AllSettings) = {
 
-    val css = CSSElementForStage(assets.getFileContentsAsHtml, stage, RefPath("newContributionsLandingPageStyles.css"))
+    val elementForStage = CSSElementForStage(assets.getFileContentsAsHtml, stage)_
```
johnduffell: if preferred, I could have kept this as 1 parameter group and done something more like this:
```suggestion
    val elementForStage = CSSElementForStage(assets.getFileContentsAsHtml, stage, _)
```

support-frontend/app/views/main.scala.html
```diff
@@ -82,9 +82,6 @@
 		}
 	</script>
 
-    <script async type="text/javascript" src="@assets("googleTagManagerScript.js")"></script>
```
johnduffell: tag manager is moved later so it doesn't download too earlier ⬇️

support-frontend/app/views/main.scala.html
```diff
@@ -110,7 +107,6 @@
     @mainElement.fold(emptyDiv, withContent)
 
     <script src="https://polyfill.io/v2/polyfill.min.js?features=default,Array.prototype.find,fetch,Array.prototype.includes,Number.isInteger,Object.values,Object.entries"></script>
-    <script defer id="stripe-js" src="https://js.stripe.com/v3/"></script>
```
johnduffell: moved down ⬇️

johnduffell:
looks good in prod
image

PR #1615, opened 2019-03-07
Query params to omit PayPal and Stripe checkout JS

joelochlann:
Looks like this has a pretty big impact on Time To Interactive, while leaving FCP and FMP unchanged (as we'd expect)

# before, loading PayPal & Stripe checkout JS as normal
picture 455

# after, without loading PayPal & Stripe checkout JS
## (also has PayPal iframe removed from serverside-rendered template)
picture 453

### DIFF
--- a/support-frontend/app/views/newContributions.scala.html
+++ b/support-frontend/app/views/newContributions.scala.html
@@ -98,83 +98,36 @@
                   </label></span>
               </li>
             </ul>
-          </fieldset><div class="stripe-payment-request-button stripe-payment-request-button--hidden"></div><div class="form form--content"><div class="form-fields"><div class="form__field form__field--contribution-email"><label class="form__label" for="contributionEmail">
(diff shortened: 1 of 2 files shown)


support-frontend/app/views/newContributions.scala.html
```diff
@@ -98,83 +98,36 @@ <h1 class="header ">Help us deliver the independent journalism the world needs</
                   </label></span>
               </li>
             </ul>
-          </fieldset><div class="stripe-payment-request-button stripe-payment-request-button--hidden"></div><div class="form form--content"><div class="form-fields"><div class="form__field form__field--contribution-email"><label class="form__label" for="contributionEmail">
-            Email address</label><span class="form__input-with-icon"><input id="contributionEmail" class="form__input form__input--contributionEmail" type="email" autocapitalize="none" autocomplete="email" required="" placeholder="example@@domain.com" pattern="^[a-zA-Z0-9\.!#$%&amp;'*+/=?^_`{|}~-]+@@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)+$" style="background-image: url(&quot;data: image/png;
+          </fieldset>
+          <div class="stripe-payment-request-button stripe-payment-request-button--hidden"></div>
```
joelochlann: sorry for the noise here - all I really did is remove the PayPal iframe, which we shouldn't have in the serverside template anyway

but I messed around with the indentation to make sure I was deleting the right bit

PR #1617, opened 2019-03-07
correct icon beside direct debit payment option to be actual Direct Debit logo

twrichards:
## Why are you doing this?

An initial change as part of https://trello.com/c/1T6txwB4 I noticed that the symbol beside the Direct Debit payment method option was the same as card... confusing.

## Changes

- Change the react components to render the direct debit symbol (coloured accordingly) beside the Direct Debit payment method option, rather than the credit card symbol as before (which is confusing).
- PLUS matching changes to `newContributions.scala.html` (the server-side placeholder)

## Screenshots
| Before | image |
| --- | --- |
| **After** | image |


support-frontend/assets/components/svgs/directDebitSymbol.jsx
```diff
@@ -2,18 +2,33 @@
...
         d="M18.254 9.655v13.73s-7.444-1.436-7.444-6.893c0-4.353 5.51-6.272 7.444-6.837zm-14.618 6.55c0 3.2 5.371 7.878 14.119 8 .345 0 10.282-.41 10.129-11.817-.129-9.582-6.834-11.26-8.978-11.552V.252c8.548.58 15.27 6.078 15.27 12.75 0 7.06-7.52 12.804-16.766 12.804C8.163 25.806.68 20.06.68 13.003.681 5.945 8.163.2 17.41.2c.283 0 .565.005.844.016V9.4c-4.474-.346-14.618 1.897-14.618 6.805z"
-        fill="#212121"
+        fill={props.fill}
       />
     </svg>
   );
 }
+
+SvgDirectDebitSymbol.defaultProps = {
+  fill: '#212121',
```
walaura: could we do this using css instead?

walaura:
😻

support-frontend/assets/components/svgs/directDebitSymbol.jsx
twrichards: great shout - doing that now

twrichards: done

PR #1618, opened 2019-03-07
Address form fixes

rupertbates:
## Why are you doing this?

There were a number of bugs with the address section of the Digital Pack checkout
* Postcode should be 'Zip code' in the US
* Postcode should be optional in countries other than US, UK, AU & CA
* AU should have a mandatory state list

Other changes:
* We have decided that we don't need the county field (it was optional and not actually being saved to Zuora) so I have removed it.
* I have moved all the validation code out of `/digitalSubscriptionCheckoutReducer.js` into its own file.

[**Trello Card**](https://trello.com/c/m8N722T3/2282-bugs-in-the-us-digital-pack-checkout)

### DIFF
--- a/support-frontend/assets/pages/digital-subscription-checkout/helpers/validation.js
+++ b/support-frontend/assets/pages/digital-subscription-checkout/helpers/validation.js
@@ -2,7 +2,7 @@
 
 import type { Option } from 'helpers/types/option';
 import type { IsoCountry } from 'helpers/internationalisation/country';
-import { Dispatch } from 'redux';
+import type { Dispatch } from 'redux';
 import type { Action } from 'pages/digital-subscription-checkout/digitalSubscriptionCheckoutActions';
 import { getFormFields, setFormErrors } from 'pages/digital-subscription-checkout/digitalSubscriptionCheckoutActions';
 import type {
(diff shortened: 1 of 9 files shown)


lmath:
If AU states are mandatory, it's probably good to update the server side validation too: https://github.com/guardian/support-frontend/blob/master/support-frontend/app/utils/CheckoutValidationRules.scala#L40

support-frontend/assets/pages/digital-subscription-checkout/helpers/validation.js
```diff
@@ -0,0 +1,71 @@
...
+}
+
+const formIsValid = (state: State): boolean => {
+  const errors = getErrors(getFormFields(state));
+  return errors.length === 0;
+};
+
+function validateForm(dispatch: Dispatch<Action>, state: State) {
+  const errors = getErrors(getFormFields(state));
+  const valid = errors.length === 0;
```
lmath: I know this has just been moved here, but why not use `formIsValid`?

rupertbates: Because we need the list of errors to pass to `setFormErrors`

PR #1620, opened 2019-03-07
Load fonts via dotcom endpoint

jranks123:
## Why are you doing this?
Previously, for those visiting our site for the first time (i.e a large percentage of users), we were downloading 4 entire font files when they came to the site. This has a large impact on the time to interactive: by stripping out the fonts, the time to interactive was cut by 25%.

We realised that most users actually already have the fonts in their browser, if they've come from dotcom (i.e most users), but it is stored in the `gu.com` local storage, which isn't accessible from `support.theguardian.com`. 

To get around this, the dotcom team have build an endpoint on dotcom, `/font-loader`, that serves an iFrame which extracts the fonts from local storage and sends them to the parent window where it is hosted using postmessage. This PR inserts this iFrame into the page and waits for the fonts to be sent through, and then appends them directly into the head. This means that, for users who have come from `gu.com`, there is no network request and no downloading required to get the fonts! Excellent! (If they haven't come from `gu.com`, the iFrame will make a network request and return the fonts anyway, just a bit slower). 

One compromise we've made for this approach is to not use Titlepiece on our page. This is because `gu.com` doesn't save that font to the users local storage currently, as it's not used frequently enough. But I've left it in so that if anyone else wants to use it, they can. For the other fonts, they will now be loaded in the new way.

Note: this is a temporary solution. When dotcom becomes Dot Components, it will load the font files directly from pastup, like `support.theguardian.com` currently does, and this will mean that this hackery won't be necessary as the fonts will be in the users browser cache already. 

There are also a few design implementations, as dotcom doesn't actually download all the weights of fonts that we were using before. We have lost the medium weight on Guardian Text Sans Web, so we have had to go either regular or bold in places where we were using medium.

We now embolden the selected amount, as @ionamckendrick thought it was too weak in contrast with the now-bolder "Contrbute" button.

We've also stopped using titlepiece on the contributions site (although I've left it in as `/support` still use it)

|Before|
|---|
|image|

|After|
|---|
|image|

Due to this change, @jesseyuen and I had to make a slight change to one of the font presets that `/support` uses (look at the paragraph copy, it is now lighter in weight):

|Before|
|---|
image

|After|
|---|
image

# TODO
- ~~I feel like it might be nice to put that script in main.html.scala somewhere else and have it inserted, maybe with webpack? I'm unsure of exactly how I'd do that but suggestions welcome~~
- ~~Once that is done, add the script to the other pages (I think it's just the unsupported browser page that needs it, there might be a couple of others)~~
- Add the fonts to local storage on load and check for them before inserting the iFrame - probably in the next PR though
- ~~waiting on [this PR](https://github.com/guardian/frontend/pull/21215) to be merged so that I can replace the `ngrok` links~~


support-frontend/assets/stylesheets/gu-sass/typography.scss
```diff
@@ -8,7 +8,7 @@
 // ----- Font Families ----- //
 
 $gu-titlepiece: 'GT Guardian Titlepiece', Georgia, serif !default;
-$gu-headline: 'Guardian Headline', Georgia, serif !default;
+$gu-headline: 'Guardian Egyptian Web', Georgia, serif !default;
```
jranks123: this change is because frontend calls what we call Guardian Headline, Guardian Egyptian Web

support-frontend/conf/DEV.public.conf
```diff
@@ -16,7 +16,8 @@ support.url="https://support.thegulocal.com"
...
+#TODO use prod link to font-loader
+play.filters.headers.contentSecurityPolicy = "default-src 'self' www.paypalobjects.com www.paypal.com www.sandbox.paypal.com js.stripe.com pasteup.guim.co.uk ophan.theguardian.com j.ophan.co.uk media.guim.co.uk www.google-analytics.com www.googletagmanager.com tagmanager.google.com polyfill.io www.googleadservices.com googleads.g.doubleclick.net www.google.com www.google.co.uk optimize.google.com static.ads-twitter.com bat.bing.com bid.g.doubleclick.net t.co analytics.twitter.com stats.g.doubleclick.net www.youtube-nocookie.com connect.facebook.net www.facebook.com consumer.krxd.net cdn.krxd.net beacon.krxd.net secure.adnxs.com checkout.stripe.com fonts.googleapis.com ssl.gstatic.com www.gstatic.com fonts.gstatic.com sentry.io www.dwin1.com data: wss: 'unsafe-inline' 'unsafe-eval'  q.stripe.com payment.code.dev-guardianapis.com https://interactive.guim.co.uk/ https://176c9a2c.ngrok.io"
```
mchv: Not to change now, but the CSP is almost useless here due to number of domains allowed combined with the unsafe allowed

mchv: Quickly looking at domains `krux` `twitter analytics` probably don’t require to be added here

support-frontend/app/views/main.scala.html
```diff
@@ -83,11 +83,47 @@
 	</script>
 
     <script async type="text/javascript" src="@assets("googleTagManagerScript.js")"></script>
-
     <link rel="preload"  as="script" href="@assets(mainJsBundle)">
 
     </head>
   <body class="header-tweaks">
+    <script>
```
walaura: let's put this in dotcom so we dont have all these magic variables here

support-frontend/app/views/main.scala.html
```diff
@@ -83,11 +83,47 @@
...
+      (function(window, document) {
+      var head = document.getElementsByTagName('head')[0];
+      var hideIframeStyle = document.createElement('style');
+      hideIframeStyle.innerHTML= ".guardianFontLoader {display: none;}";
```
walaura: we could instead do 

```
iframe.style.display = 'none';
```

a bit below and avoid these four lines

support-frontend/assets/stylesheets/skeleton/webfonts.scss
```diff
@@ -2,7 +2,7 @@ $gu-gt-fonts-url: "//pasteup.guim.co.uk/webfonts/1.0.0";
 $gu-gt-font-variant: "noalts-not-hinted";
```
walaura: can we rename this file to `non-critical-webdonts.scss` or something and add a comment to clarify most webfonts aren't here?

support-frontend/assets/stylesheets/gu-sass/typography.scss
walaura: can we get frontend to rename their font? that's the wrong name 😭

support-frontend/conf/DEV.public.conf
rupertbates: We need Krux and Twitter for remarketing tags

walaura: is there any way to analise unused csp items?

mchv: @rupertbates let's catchup offline about this.

support-frontend/app/controllers/Application.scala
```diff
@@ -138,7 +134,7 @@ class Application(
       paymentApiStripeEndpoint = paymentAPIService.stripeExecutePaymentEndpoint,
       paymentApiPayPalEndpoint = paymentAPIService.payPalCreatePaymentEndpoint,
       idUser = idUser,
-      preload = preload
+      preload = Nil
```
johnduffell: if this isn't used, then remove it completely (it was only added for this call)
also remove the Preload case class and anywhere else it's used.

support-frontend/assets/stylesheets/skeleton/non-critical-webfonts.scss
```diff
@@ -0,0 +1,33 @@
+// Our core fonts (Guardian Headline, Guardian Text Egyptian Web and Guardian Text Sans Egyptian Web)
+// are loaded via an iFrame from dotcom. If you use any fonts in this file, it will result in
+// the font files being downloaded from pasteup
```
johnduffell: this isn't really the problem that they will be downloaded from pasteup, the problem is that will actually slow everything else down due to bandwidth contention with the critical stuff.
Being downloaded from pasteup would be okay if that was triggered by the JS after the critical stuff was in memory.

jranks123: I'm not saying it's a problem, I'm just saying what will literally happen

jranks123: I guess I could say "unless they are already in the browser cache"

jranks123: although IMO thats an broswer implentation detail

support-frontend/app/controllers/Application.scala
jranks123: 👍

PR #1621, opened 2019-03-07
load raven library on demand rather than bundling it

johnduffell:
We noticed that the raven library was in the bundle and taking up 39K (13K gzipped).  The rest of the bundle is 280kb (80kb gzipped).  This is about 15% which is something that is non critical to painting the page.

It seems to be working OK locally, just letting the promise run and discarding the result.  I'm not sure whether that is good practice though?  In case the bundle doesn't download, we might not realise.

[**Trello Card**](https://trello.com/c/E5XA4fqR/356-js-raven-load-on-demand)

|PROD|CODE|
|---|---|
|image|image|
|image|image|
|image|image|

### DIFF
--- a/support-frontend/assets/helpers/logger.js
+++ b/support-frontend/assets/helpers/logger.js
@@ -1,39 +1,45 @@
 // @flow
-import Raven from 'raven-js';
+// import Raven from 'raven-js';
+
+const EventualRaven = import('raven-js');
 
 // ----- Functions ----- //
 
-const init = () => {
+const init = () => EventualRaven.then((Raven) => {
   const dsn: string = 'https://65f7514888b6407881f34a6cf1320d06@sentry.io/1213654';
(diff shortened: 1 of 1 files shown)


walaura:
they do this in identity-frontend! its brill <3 

is raven called elsewhere or just there?

johnduffell:
@walaura 
>is raven called elsewhere or just there?

Not sure,  I just removed the import and the bundle got smaller, I assume it would have stayed in te bundle if it was used elsewhere? and/or the build woudl have failed?  Please tell me if I did something careless!

After:
image

walaura:
What about the rest of bundles? 😅

johnduffell:
@joelochlann was looking at the code coverage on initial page load and it wasn't totally clear what was what, but if we can make it only load the covered code for the first paint that would be fab!  Watch for another PR

PR #1636, opened 2019-03-11
Queue Paper thank you emails via support-workers

jacobwinch:
## Why are you doing this?

This allows us to send thank you emails via support-workers for the new paper checkout flows.

[**Trello Card**](https://trello.com/c/7RoY2JWS/2266-thank-you-email-card-and-direct-debit)

## Changes

* Add Paper email fields 
* Wire up SendThankYouEmail lambda for Paper
* Add simple integration test

## Screenshots

These should look (almost) the same as the existing emails, but in case you're interested:

**Collection/Voucher (card)**

image

**Collection/Voucher (direct debit)**

image

**Home Delivery (card)**

image

**Home Delivery (direct debit)**

image


support-workers/src/main/scala/com/gu/emailservices/PaperEmailFields.scala
```diff
@@ -0,0 +1,69 @@
...
+    List(
+      "delivery_address_line_1" -> address.lineOne.getOrElse(""),
+      "delivery_address_line_2" -> address.lineTwo.getOrElse(""),
+      "delivery_address_town" -> address.city.getOrElse(""),
+      "delivery_postcode" -> address.postCode.getOrElse(""),
+      "delivery_country" -> address.country.name
+    )
+  }
+
+  override val fields = List(
```
jacobwinch: Based on: https://github.com/guardian/subscriptions-frontend/blob/9ae6ee3d9f8d5f21bce81b2c76225d7fc07db6fd/app/model/Email.scala#L143-L207

support-workers/src/main/scala/com/gu/emailservices/PaperEmailFields.scala
```diff
@@ -0,0 +1,69 @@
...
+    "subscriber_id" -> subscriptionNumber,
+    "first_name" -> user.firstName,
+    "last_name" -> user.lastName,
+    "date_of_first_paper" -> formatDate(firstDeliveryDate.getOrElse(firstPaymentDate)),
+    "date_of_first_payment" -> formatDate(firstPaymentDate),
+    "package" -> productOptions.toString,
+    "subscription_rate" -> SubscriptionEmailFieldHelpers.describe(paymentSchedule, billingPeriod, currency)
+  ) ++ paymentFields ++ deliveryAddressFields.getOrElse(List())
+
+  //TODO - do we care about the billing and promotion fields - these are included in the emails queued via subscriptions-frontend?
```
jacobwinch: I've omitted the billing address fields from the email for now - we can wire them through if necessary, but I'm not convinced that including delivery and billing address in the email is valuable / looks appealing?

support-workers/src/main/scala/com/gu/emailservices/PaperEmailFields.scala
jacobwinch: However there are a few omissions:

- "IncludesDigipack" - I wrote a little helper function for this initially but haven't included it because this value doesn't seem to be used anyway!
- "delivery_county" - we don't collect this anymore
- "delivery_instructions" - we don't collect these as part of the new flow. Not sure if that's intentional @SarahLaughton?
- "title" - this can be wired through if necessary / if we want the emails to look more formal

support-workers/src/main/scala/com/gu/emailservices/PaperEmailFields.scala
jacobwinch: I'm not too sure what the promo fields are... there's also a todo related to them for the Digital Pack email: https://github.com/guardian/support-frontend/blob/26eadf20f80d733c051c51b893f4c0384addf311/support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala#L78

Any discounted payments are already mentioned in both emails due to: `SubscriptionEmailFieldHelpers.describe(paymentSchedule, billingPeriod, currency)`

Any idea whether we really need the promotionFields @rupertbates?

support-models/src/main/scala/com/gu/support/workers/states/SendThankYouEmailState.scala
```diff
@@ -3,12 +3,15 @@ package com.gu.support.workers.states
...
 import com.gu.support.workers.{PaymentMethod, SalesforceContactRecord, User, _}
+import com.gu.support.encoding.CustomCodecs.{decodeLocalTime, encodeLocalTime}
+import org.joda.time.LocalDate
 
 case class SendThankYouEmailState(
   requestId: UUID,
   user: User,
   product: ProductType,
   paymentMethod: PaymentMethod,
+  firstDeliveryDate: Option[LocalDate],
```
rupertbates: Will we not always have this field?

support-workers/src/main/scala/com/gu/emailservices/PaperEmailFields.scala
rupertbates: No it sounds like we can get rid of those comments

SarahLaughton:
@jacobwinch I'm not sure we've removed the delivery instructions on purpose, but we may not need it if delivery gets pulled from sale for a while. I'll keep you posted.

support-models/src/main/scala/com/gu/support/workers/states/SendThankYouEmailState.scala
jacobwinch: This will always be None for Recurring Contributions and Digital Pack.

PR #1650, opened 2019-03-12
Paper checkout links

rupertbates:
## Why are you doing this?
To be able to go from the paper landing page to the checkout page we need to pass the selected product and fulfilment options in the url.
This has required quite a large refactor of the paper landing page to make it work with the new `ProductPrices` objects from the catalog but has allowed me to delete quite a bit of the old pricing code from `subscriptions.js`

## Changes
* Load prices from the product catalog on the paper landing page rather than using the old hard coded values
* Update components to use the catalog types
* Generate links to the new paper checkout page when the `displayCheckout=true` query param is present

[**Trello Card**](https://trello.com/c/3FHVlJbc/2297-pass-paper-product-choices-from-landing-page-to-checkout-page)

### DIFF
--- a/support-frontend/.eslintrc
+++ b/support-frontend/.eslintrc
@@ -35,6 +35,7 @@
     "import/no-duplicates": "off",
     "import/no-extraneous-dependencies": "off",
     "padded-blocks": "off",
+    "import/first": "off",
     "import/prefer-default-export": "off",
     "jsx-a11y/label-has-for": [ 2, {
       "components": [ "Label" ],
--- a/support-frontend/assets/pages/paper-subscription-landing/paperSubscriptionLandingPageActions.js
+++ b/support-frontend/assets/pages/paper-subscription-landing/paperSubscriptionLandingPageActions.js
@@ -4,7 +4,7 @@
 
 import { ProductPagePlanFormActionsFor } from 'components/productPage/productPagePlanForm/productPagePlanFormActions';
 import { paperCheckoutUrl, paperSubsUrl } from 'helpers/routes';
-import { getPaperCheckout } from 'helpers/externalLinks';
+import { getLegacyPaperCheckout } from 'helpers/externalLinks';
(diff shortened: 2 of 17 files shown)


support-frontend/.eslintrc
```diff
@@ -35,6 +35,7 @@
     "import/no-duplicates": "off",
     "import/no-extraneous-dependencies": "off",
     "padded-blocks": "off",
+    "import/first": "off",
```
rupertbates: Turn off those annoying `Absolute imports should come before relative imports` errors

jacobwinch: 💯

support-frontend/assets/pages/paper-subscription-landing/paperSubscriptionLandingPageActions.js
```diff
@@ -2,39 +2,52 @@
...
 };
 
+const getCheckoutUrl = (state: State) => {
+  if (getQueryParameter('displayCheckout') === 'true') {
+    const product = state.page.plan.plan ? state.page.plan.plan : Everyday;
+    return paperCheckoutUrl(state.page.tab, product);
+  }
+
+  const { referrerAcquisitionData, abParticipations, optimizeExperiments } = state.common;
+  return state.page.plan.plan ? getPaperCheckout(
```
jacobwinch: Minor: now that we have two functions for getting a paper checkout url it might be a bit clearer to call this `getLegacyPaperCheckout` / `getSubsFrontendPaperCheckout` (or similar)

rupertbates: Yes, because it's in the externalLinks.js file I thought that would probably make it clear but I think you're probably right.

jacobwinch: That's a fair point, if I was looking at this code in an IDE (instead of reviewing in GitHub) I would just click through so it's no big deal.

PR #1652, opened 2019-03-12
Hide ticker due to issue

joelochlann:
The ticker isn't updating. Raw events in `acquisition_events_prod` don't have any campaignCodes against them, which means nothing will match the conditions for this campaign.

We'll fix tomorrow but right now we need to switch it off


tomrf1:
closing as the issue has been fixed

PR #1653, opened 2019-03-13
Do not load unnecessary payment JS

joelochlann:
Just acting on a thought I had, fairly late into the evening!

We're not checking the available contribution types before we initialise third-party libs. Most wastefully, we're loading PayPal recurring JS even on pages which only have one-off available (e.g. the current Australia campaign page). This will definitely have an impact on performance.

We're also running through some unnecessary Stripe logic on pages where particular payment methods are disabled (such as non-UK/US right now). We still have to actually load the main Stripe libs, since Stripe is used on all contribution types, but we shouldn't try and configure the Stripe libs for contribution types which aren't available (though admittedly this probably won't improve performance).

### DIFF
--- a/support-frontend/assets/pages/new-contributions-landing/contributionsLandingInit.js
+++ b/support-frontend/assets/pages/new-contributions-landing/contributionsLandingInit.js
@@ -89,7 +89,7 @@ function initialiseStripeCheckout(
 
 
 function initialisePaymentMethods(state: State, dispatch: Function) {
-  const { countryId, currencyId } = state.common.internationalisation;
+  const { countryId, currencyId, countryGroupId } = state.common.internationalisation;
   const { switches } = state.common.settings;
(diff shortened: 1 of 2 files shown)


support-frontend/assets/pages/new-contributions-landing/contributionsLandingInit.js
```diff
@@ -115,7 +117,10 @@ function initialisePaymentMethods(state: State, dispatch: Function) {
     });
   }
 
-  if (getQueryParameter('paypal-js') !== 'no') {
+  const hasRecurring = contributionTypes.includes('MONTHLY')
```
jranks123: I wonder if this could be more clearly named, to me it's not immediately clear what exactly has recurring. Maybe `recurringContributionsAvailable` or `recurringContributionsDisplayed`?

PR #1654, opened 2019-03-13
Rationalise payment methods

rupertbates:
## Why are you doing this?
We currently have 3 almost identical `PaymentMethod` types, one in contributions, one for digital subscriptions and one for paper subscriptions. This PR extracts one type from those and uses it everywhere. I've also introduced constants for the payment methods themselves and used those throughout the codebase.


lmath:
Good idea 👍

joelochlann:
this is great

PR #1656, opened 2019-03-13
Ticker post goal

joelochlann:
## before
Picture 467

## after
Picture 466


jranks123:
I.e:

image

joelochlann:
> Could we not say how much it has raised over the amount? Like you see on kickstarter, i. e $123,002 raised against $100,000 goal

This would be very easy to implement. I'll leave @ionamckendrick et al to make the call on this

ionamckendrick:
@jranks123 To reply to your comment above. Until today's last min decision, we were legally not allowed to accept any more money and that's why I redesigned the ticker this way (to be "full") Any subsequent donations were going into our main pot and could not be visually represented here. But, as you'll have now seen, we've been given the green light to accept more for the campaign. So we're going back to the original plan (kickstarted style)!

PR #1663, opened 2019-03-14
Make featured product hero always Guardian Weekly.

lmath:
## Why are you doing this?
As we near the end of the quarter, we would prefer to stop promoting Digital Pack

💰 

[**Trello Card**](https://trello.com/c/o2YsnqAl/2283-subs-page-march-april-updates-1st-15th-march-31st-march)

## Changes

* Get the featured product hero to always feature Guardian Weekly, unless there is a digital pack flash sale on 

## Screenshots
for  /<uk|us|au|eu|int|nz|ca>/subscribe 

image

### DIFF
--- a/support-frontend/assets/pages/subscriptions-landing/components/featuredProducts.jsx
+++ b/support-frontend/assets/pages/subscriptions-landing/components/featuredProducts.jsx
@@ -160,6 +160,7 @@ const getProducts = (
 
 const getProduct = (subsLinks: SubsUrls, countryGroupId: CountryGroupId): ?Product => {
   const products = getProducts(subsLinks, countryGroupId);
+  const promoteGuardianWeekly = true;
   switch (getQueryParameter('ab_product')) {
     case 'DigitalPack':
       return products.DigitalPack;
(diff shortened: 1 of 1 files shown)


support-frontend/assets/pages/subscriptions-landing/components/featuredProducts.jsx
```diff
@@ -168,6 +169,9 @@ const getProduct = (subsLinks: SubsUrls, countryGroupId: CountryGroupId): ?Produ
     case 'GuardianWeekly':
       return products.GuardianWeekly;
     default:
+      if (promoteGuardianWeekly) {
```
rupertbates: This means we will no longer be promoting the paper in the UK, is that what we want?

support-frontend/assets/pages/subscriptions-landing/components/featuredProducts.jsx
```diff
@@ -168,6 +169,9 @@ const getProduct = (subsLinks: SubsUrls, countryGroupId: CountryGroupId): ?Produ
...
+        return products.GuardianWeekly;
+      }
```
rupertbates: I know this isn't your change but there are 2 functionally identical if clauses here:
` if (countryGroupId === GBPCountries && flashSaleIsActive('DigitalPack', GBPCountries))`
and 
`if (countryGroupId !== GBPCountries && flashSaleIsActive('DigitalPack', countryGroupId))`
Could you remove one in this PR please?

lmath: Sure thing

support-frontend/assets/pages/subscriptions-landing/components/featuredProducts.jsx
lmath: I also wondered this. I've asked Chris about it.

support-frontend/assets/pages/subscriptions-landing/components/featuredProducts.jsx
```diff
@@ -160,6 +160,7 @@ const getProducts = (
 
 const getProduct = (subsLinks: SubsUrls, countryGroupId: CountryGroupId): ?Product => {
   const products = getProducts(subsLinks, countryGroupId);
+  const promoteGuardianWeekly = true;
```
rupertbates: I think it might be simpler to remove this variable and the if clause that relates to it, then also remove 
```
if (countryGroupId === GBPCountries) {
        return products.Paper;	
  }
```	      
that would give us the behaviour we want with a lot fewer ifs to read through?

lmath: Based on the card, I thought what we wanted was for two weeks to push GW and then have another PR to remove that and return to existing behaviour. Based on standup, I think we might just want 
```
default: 
    return products.GuardianWeekly;
```

But I shall find that out

rupertbates: I think that's right except that we want to keep digipack as long as the flash sale is on, so if you left that if clause in it would mean you could merge now and it would do the right thing

lmath: If we ordinarily need to maintain digital pack behaviour during a flash sale, then that's why we need the `const promoteGuardianWeekly` because we are currently running a digital pack flash sale but want to promote GW instead this time.

lmath: OK I misunderstood the requirements and you're totally right @rupertbates 

Thanks for your patience 😄

lmath:
Any chance of a +1 (or more comments) @rupertbates ?

PR #1673, opened 2019-03-15
'existing payment options' in Contributions checkout

twrichards:
### Why are you doing this?
Initially the motivation behind this change is to facilitate 'resubscribes' and other 'upsell' opportunities from within the manage my account area (`manage-frontend`  `manage.theguardian.com`), however if proven successful can easily be rolled out to all users (who are signed in and already have _recent_ payment methods listed with us) to reduce checkout friction (by avoiding entering their payment details again).

The choice to offer 'existing payment methods' in the existing **recurring** contributions checkout (will be replicated in the new Digital Pack and Print checkouts in due course) was to avoid rebuilding yet _another_ checkout flow somewhere, while efforts are ongoing to consolidate all checkouts in `support-frontend`.

Although this PR touches quite a few files (and has a number or pre-requisite PRs outside `support-frontend') the lines changed is relatively small given the significance of this change. The changes slotted in nicely with the existing UI and model, and will allow this 'existing payment options' feature to evolve nicely with wider changes to the checkout without having to maintain another system. 

### Changes

- Read new 'switches' for toggling `existingCard` and `existingDirectDebit` globally (switches introduced in https://github.com/guardian/support-admin-console/pull/14) 
- Added `members-data-api` integration client-side (with the URL passed through on `window` from public server-side config) - PLUS added to `contentSecurityPolicy`. 
- Call to the new `members-data-api` `/user-attributes/me/existing-payment-options` endpoint, passing the selected currency ISO as a required query parameter (see https://github.com/guardian/members-data-api/pull/376) shortly after page load, ONLY IF...
  - User is signed-in (otherwise there would be no point making the API call)
  - `displayExistingPaymentOptions` query param is `=true`
  - either `existingCard` OR `existingDirectDebit` switch is ON
- New REDUX actions (and accompanying state)...
  - Action: `UPDATE_SELECTED_EXISTING_PAYMENT_METHOD` State: `page.existingPaymentMethod` this tracks the specific selected payment method, and is `undefined` when one of the normal payment methods is selected. This new piece of state is in addition to the `paymentMethod` that was already there, which tracks the type of `PaymentMethod` selected (which in these new cases will be `ExistingCard` or `ExistingDirectDebit`.
  - Action: `SET_EXISTING_PAYMENT_METHODS` State: `page.form.existingPaymentMethod` this stores the list of existing payment method objects which comes back from `members-data-api` (but first filters based on `existingCard` and `existingDirectDebit` switches before storing). 
    - If the list is empty nothing existing payment related is rendered
    - If undefined then loading GIF is shown
    - If list contains just basic options (`paymentType` only, not full detail like `billingAccountId`) then some text is shown to invite the user to 're-enter their password'
    - If list contains full detail options then those options are rendered in the list under 'Payment Method' for the user to select.
- UI changes in `PaymentMethodSelector.jsx` which renders accordingly based on the different state _(see screenshots)_

### Screenshots
_all assuming new query param is present and `true`_ image

#### Providing the user is signed-in within the last hour...
image

#### Whilst loading (space reserved for one existing payment method - most likely scenario)...
image

#### If the user is signed-in but not in the last hour (but user does have existing payment methods available)...
image
_NOTE: since all the initial traffic will come from `manage-frontend` (where users already must be signed in within the last hour) it's very unlikely anyone will see this - although in future if we roll out more widely we want to review the copy_

#### If the user is not signed in - NO CHANGE (and notably NO AJAX call to `members-data-api`)
image


support-frontend/app/admin/settings/Settings.scala
```diff
@@ -101,7 +101,13 @@ object SettingsSource extends LazyLogging {
 
 }
 
-case class PaymentMethodsSwitch(stripe: SwitchState, payPal: SwitchState, directDebit: Option[SwitchState])
+case class PaymentMethodsSwitch(
+  stripe: SwitchState,
+  payPal: SwitchState,
+  directDebit: Option[SwitchState],
+  existingCard: Option[SwitchState],
```
jacobwinch: What benefit do we get from having two separate switches here? 

If I recall correctly the main purpose of these switches is that it allows us to disable a payment method if the third-party is down (e.g. we can hide Stripe completely if we know that Stripe Checkout will fail to load). For the reuse of existing payment methods we don't need to talk to third-parties client side, so I think a single switch could be sufficient?

support-frontend/app/admin/settings/Settings.scala
```diff
@@ -112,15 +118,15 @@ case class ExperimentSwitch(name: String, description: String, state: SwitchStat
...
 object PaymentMethodsSwitch {
   def fromConfig(config: Config): PaymentMethodsSwitch =
     PaymentMethodsSwitch(
       SwitchState.fromConfig(config, "stripe"),
       SwitchState.fromConfig(config, "payPal"),
-      if (config.hasPath("directDebit"))
-        Some(SwitchState.fromConfig(config, "directDebit"))
-      else
-        None
+      SwitchState.optionFromConfig(config, "direcDebit"),
```
jacobwinch:
```suggestion
      SwitchState.optionFromConfig(config, "directDebit"),
```

support-frontend/app/admin/settings/Settings.scala
```diff
@@ -112,15 +118,15 @@ case class ExperimentSwitch(name: String, description: String, state: SwitchStat
...
+      SwitchState.optionFromConfig(config, "existingCard"),
+      SwitchState.optionFromConfig(config, "existingDirecDebit")
```
jacobwinch:
```suggestion
      SwitchState.optionFromConfig(config, "existingDirectDebit")
```

support-frontend/assets/pages/new-contributions-landing/components/PaymentMethodSelector.jsx
```diff
@@ -75,14 +104,62 @@ function PaymentMethodSelector(props: PropTypes) {
...
 
-  const noPaymentMethodsErrorMessage = <GeneralErrorMessage classModifiers={['no-valid-payments']} errorHeading="Payment methods are unavailable" errorReason="all_payment_methods_unavailable" />;
+  const noPaymentMethodsErrorMessage =
+    (<GeneralErrorMessage
+      classModifiers={['no-valid-payments']}
+      errorHeading="Payment methods are unavailable"
+      errorReason="all_payment_methods_unavailable"
+    />);
+
+  // having to do this nasty cast because Flow sucks and type guards don't work through .filter
```
walaura:
```suggestion
  // having to do this nasty casty because Flow sucks and type guards don't work through .filter
```

support-frontend/assets/pages/new-contributions-landing/contributionsLanding.scss
```diff
@@ -742,6 +742,29 @@ form {
   margin-left: 5px;
 }
 
+.has-existing-payment-option-explainer {
```
walaura: the radios in `/components/forms/customFields/radioInput.jsx`support images now so perhaps this'd be a good time to migrate to them? Might be a tall order though

support-frontend/assets/helpers/paymentIntegrations/readerRevenueApis.js
```diff
@@ -49,10 +49,13 @@ type RegularDirectDebitPaymentFields = {|
   accountNumber: string,
 |};
 
+type RegualarExistingPaymentFields = {| billingAccountId: string |};
```
jacobwinch:
```suggestion
type RegularExistingPaymentFields = {| billingAccountId: string |};
```

support-frontend/assets/helpers/page/page.js
```diff
@@ -156,6 +167,29 @@ function init<S, A>(
       store.dispatch(setExperimentVariant(exp));
     });
 
+    // initiate fetch of existing payment methods
```
jacobwinch: I know that the query params will prevent us from actually making this call when we don't intend to, but this feels very checkout specific so I'm not sure that it should live in this function or file

support-frontend/app/admin/settings/Settings.scala
twrichards: Given out change of approach in `support-workers` to clone payment methods (due to concerns around auto-pay, see https://github.com/guardian/members-data-api/pull/379#issuecomment-476304459) we will actually have to call 3rd parties for existing due to limitations in Zuora, so being able to switch them separately is definitely necessary.

The original rationale was more to do with being able to push people in one direction and in-case we discovered issues in production say with re-using Direct Debit being able to switch it off without disabling the whole feature... which is still valid.

jacobwinch: Yep agreed, thanks for explaining 👍

support-frontend/app/admin/settings/Settings.scala
twrichards: oooo that was nasty - fixed now

support-frontend/assets/pages/new-contributions-landing/contributionsLanding.scss
twrichards: good shout - although I'd prefer to refactor the new and existing payment methods all at the same time, I'll make a card to do a follow up PR

support-frontend/assets/helpers/page/page.js
twrichards: absolutely! DONE in https://github.com/guardian/support-frontend/pull/1673/commits/56e51278be3003019ab65aa1ca9f990d73fb0dcc

jranks123:
Come talk to me before you merge this, I'm about to make the page cachable as part of [this PR](https://github.com/guardian/support-frontend/pull/1720), and as part of that we are removing the window.user object from the server response, and instead making a call here in `contributionsLandingInit` to identity to get the user info, which we then update the form state with

support-frontend/assets/helpers/existingPaymentMethods/existingPaymentMethods.js
```diff
@@ -0,0 +1,100 @@
...
+function isFullDetailExistingPaymentMethod(existingPaymentMethod: ExistingPaymentMethod): boolean %checks {
+  return !!existingPaymentMethod.billingAccountId;
+}
+
+function sendGetExistingPaymentMethodsRequest(
+  isoCurrency: IsoCurrency,
+  storeResponse: ExistingPaymentMethod[] => void,
+): void {
+  fetchJson(
+    `${window.guardian.existingPaymentOptionsEndpoint}?currencyFilter=${isoCurrency}`,
```
walaura: could we abstract this into `helpers/routes`? I'm not super happy with using `window.guardian` liberally as it's a magic global and it makes doing SSR and other optimizations very hard

support-frontend/assets/pages/new-contributions-landing/components/PaymentMethodSelector.jsx
```diff
@@ -75,14 +104,62 @@ function PaymentMethodSelector(props: PropTypes) {
...
+                <span className="radio-ui" />
+                <span className="radio-ui__label">{getExistingPaymentMethodLabel(existingPaymentMethod)}</span>
+                {getPaymentMethodLogo(mapExistingPaymentMethodToPaymentMethod(existingPaymentMethod))}
+              </label>
+              <div className="existing-payment-option-explainer">
+                  Used for your{' '}
+                {subscriptionsToExplainerList(existingPaymentMethod.subscriptions.map(subscriptionToExplainerPart))}
+              </div>
+            </li>
+            ))}
```
walaura: this entire bit is pretty nested and hard to parse, i'm  wonderin if it would be worth it to extract into its own component?

support-frontend/assets/pages/new-contributions-landing/contributionsLandingInit.js
```diff
@@ -117,6 +125,31 @@ function initialisePaymentMethods(state: State, dispatch: Function) {
...
+
+    // initiate fetch of existing payment methods
+    const userAppearsLoggedIn = doesUserAppearToBeSignedIn();
+    const existingDirectDebitON = switchIsOn(switches.recurringPaymentMethods, 'existingDirectDebit');
+    const existingCardON = switchIsOn(switches.recurringPaymentMethods, 'existingCard');
+    const existingPaymentsEnabledViaUrlParam = getQueryParameter('displayExistingPaymentOptions') === 'true';
+    if (userAppearsLoggedIn && (existingCardON || existingDirectDebitON) && existingPaymentsEnabledViaUrlParam) {
+      sendGetExistingPaymentMethodsRequest(
+        currencyId,
+        (allExistingPaymentMethods: ExistingPaymentMethod[]) => {
```
walaura: this function would work best as a promise imo

support-frontend/assets/helpers/checkouts.js
```diff
@@ -120,7 +123,8 @@ function getPaymentMethodToSelect(
 
 function getPaymentMethodFromSession(): ?PaymentMethod {
   const pm: ?string = storage.getSession('selectedPaymentMethod');
-  if (pm === 'DirectDebit' || pm === 'Stripe' || pm === 'PayPal') {
+  // can't use Flow types for these comparisons for some strange reason
```
joelochlann: There is a Flow pattern for this sort of thing. But you'd need to change `assets/helpers/paymentMethods.js` and everywhere that uses it so probably not for this PR. https://flow.org/en/docs/types/utilities/#toc-keys

```javascript
const paymentMethods = {
  Stripe: 'Stripe',
  PayPal: 'PayPal',
  ...
}

type PaymentMethod = $Keys<typeof paymentMethods>;
```

so then you get the union type, and a runtime value so you can do:

```javascript
if (Object.keys(paymentMethods).includes(pm)) {
  ...
}
```

support-frontend/assets/helpers/existingPaymentMethods/existingPaymentMethods.js
```diff
@@ -0,0 +1,100 @@
...
+  paymentType: ExistingPaymentType;
+};
+
+export type FullDetailExistingPaymentMethod = {
+  billingAccountId: string,
+  subscriptions: ExistingPaymentMethodSubscription[],
+  card?: string,
+  mandate?: string,
+  // can't use Intersection types to avoid repeating this as it breaks isFullDetailExistingPaymentMethod because; Flow!
+  paymentType: ExistingPaymentType
```
joelochlann: you can use the spread operator with Flow types as well - that might work for this use case?

support-frontend/assets/pages/new-contributions-landing/contributionsLandingInit.js
```diff
@@ -117,6 +125,31 @@ function initialisePaymentMethods(state: State, dispatch: Function) {
...
+          const filteredExistingPaymentMethods = allExistingPaymentMethods.filter(existingPaymentMethod => (
```
joelochlann: could this be called `switchedOnExistingPaymentMethods`?

support-frontend/assets/pages/new-contributions-landing/components/PaymentMethodSelector.jsx
```diff
@@ -101,6 +181,16 @@ function PaymentMethodSelector(props: PropTypes) {
               </label>
             </li>
           ))}
+          {
+            props.contributionType !== 'ONE_OFF' &&
```
joelochlann: could be worth pulling this predicate out into its own function

support-frontend/assets/helpers/existingPaymentMethods/existingPaymentMethods.js
```diff
@@ -0,0 +1,100 @@
...
+  name: string
+}
+
+type ExistingPaymentType = 'Card' | 'DirectDebit';
+
+export type PartialExistingPaymentMethod = {
+  paymentType: ExistingPaymentType;
+};
+
+export type FullDetailExistingPaymentMethod = {
```
joelochlann: under what conditions are full details returned from the endpoint as opposed to partial? and could we put this information in a comment above the type/encode it in the naming?

twrichards: depending on the freshness of the sign-in, see in PR description...
> If the user is signed-in but not in the last hour (but user does have existing payment methods available)...
... the name could be better though, along with a comment... will do

twrichards:
@walaura and @joelochlann thanks for all the comments - I will make all the suggested improvements in a follow up PR

support-frontend/assets/helpers/existingPaymentMethods/existingPaymentMethods.js
twrichards: tried this but Flow still complains elsewhere, for example... 
image

PR #1676, opened 2019-03-15
Campaign-specific landing page post-goal

tomrf1:
## Why are you doing this?
We need to display a custom form when the goal is reached for the 'Frontline' campaign

[**Trello Card**](https://trello.com/c/Q2nbxhnZ/471-landing-page-update-once-goal-is-reached)

### Before we reach the goal (unchanged)
Screenshot 2019-03-15 at 15 56 43

### After we reach the goal
Screenshot 2019-03-15 at 16 34 19


support-frontend/assets/pages/new-contributions-landing/components/ContributionTicker/ContributionTicker.jsx
```diff
@@ -13,6 +13,7 @@ type StateTypes = {|
 type PropTypes = {|
   // URL to fetch ticker JSON from
   tickerJsonUrl: string,
+  onGoalReached: () => void,
```
tomrf1: rather than wiring this component up to the redux store, it gets a callback to use if the goal has been reached

support-frontend/assets/pages/new-contributions-landing/components/ContributionFormContainer.jsx
```diff
@@ -167,6 +173,48 @@ function campaignSpecificDetails() {
   return {};
 }
 
+function goalReachedTemplate() {
+  if (isFrontlineCampaign()) {
```
joelochlann: I think this condition should be an `&&` with `props.tickerGoalReached` - otherwise in principle we could show a blank page if the ticker goal was reached but it wasn't the campaign page (though I don't think this would do anything in practice right now)

joelochlann: actually - it's probably fine the way it is

support-frontend/assets/pages/new-contributions-landing/components/ContributionFormContainer.jsx
```diff
@@ -172,6 +178,48 @@ function campaignSpecificDetails() {
...
+    return (
+      <div className="goal-reached">
+        <div className="goal-reached__message">
+          Thank you to everyone who supported ‘The Frontline’.
+          We’re no longer accepting contributions for the series, but you can still support
+          The Guardian’s journalism with a single or recurring contribution
+        </div>
```
jlieb10: Text should live in `p`s or `span`s in a perfect world but this won't break anything

PR #1684, opened 2019-03-18
minor QoL improvements to startup scripts

walaura:
## Why are you doing this?
I have recently been bestowed a new laptop and the very first thing I did right after getting the long end of my old charger into the new shorter charger with the intention of returning a shorter charger with the old laptop and keeping a long one myself, was getting `support-frontend` up and running. In the process I ran into a few quirks, mostly missing comments, I thought I could iron them out for everybody but also for me should I need to do this again.

## Changes
I commented the pr!

### DIFF
--- a/support-frontend/fetch-config.sh
+++ b/support-frontend/fetch-config.sh
@@ -8,5 +8,6 @@ if [[ -w ${CONFIG_DIR} ]]; then
     --profile membership
 else
   echo "ERROR! Cannot write to ${CONFIG_DIR}. Check it exists and you have write permissions to it.";
+  echo "  'mkdir ${CONFIG_DIR}' will create ${CONFIG_DIR} if it does not exist";
   echo "  'sudo chown -R $(whoami):admin ${CONFIG_DIR}' will make you the owner of ${CONFIG_DIR}";
 fi
--- a/support-frontend/nginx/support.conf
+++ b/support-frontend/nginx/support.conf
@@ -1,8 +1,7 @@
 server {
-    listen 443;
+    listen 443 ssl;
     server_name support.thegulocal.com;
 
-    ssl on;
     ssl_certificate STAR_thegulocal_com_exp2020-01-09.crt; ## Supplied by https://github.com/guardian/identity-platform#setup-nginx-for-local-development
     ssl_certificate_key STAR_thegulocal_com_exp2020-01-09.key; ## ditto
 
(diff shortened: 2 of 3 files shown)


support-frontend/nginx/support.conf
```diff
@@ -1,8 +1,7 @@
 server {
```
walaura: nginx was updated and this is the new syntax, using the old one throws deprecation warnings on open

support-frontend/fetch-config.sh
```diff
@@ -8,5 +8,6 @@ if [[ -w ${CONFIG_DIR} ]]; then
     --profile membership
 else
   echo "ERROR! Cannot write to ${CONFIG_DIR}. Check it exists and you have write permissions to it.";
+  echo "  'mkdir ${CONFIG_DIR}' will create ${CONFIG_DIR} if it does not exist";
```
walaura: i'll more often run into the dir not existing than not having perms to it

support-frontend/setup.sh
```diff
@@ -102,6 +102,8 @@ install_nginx() {
   if ! installed nginx; then
     brew install nginx
     EXTRA_STEPS+=("nginx has been installed. Ensure you have 'include sites-enabled/*' in your nginx configuration ${NGINX_ROOT}/nginx.conf and add '127.0.0.1 support.thegulocal.com' and '127.0.0.1 support-ui.thegulocal.com' to /etc/hosts")
+  else
+    EXTRA_STEPS+=("nginx is installed. Ensure you have 'include sites-enabled/*' in your nginx configuration ${NGINX_ROOT}/nginx.conf and add '127.0.0.1 support.thegulocal.com' and '127.0.0.1 support-ui.thegulocal.com' to /etc/hosts")
```
walaura: I keep forgetting to do these and then freak out

support-frontend/fetch-config.sh
```diff
@@ -8,5 +8,6 @@ if [[ -w ${CONFIG_DIR} ]]; then
...
   echo "  'sudo chown -R $(whoami):admin ${CONFIG_DIR}' will make you the owner of ${CONFIG_DIR}";
```
akash1810: This can be scripted, for example, [here's how Grid](https://github.com/guardian/grid/blob/master/fetch-config.sh#L26-L30) handles something similar.

walaura: i'm not confident enough in my shell chops to do this but happy to pair tomorrow!!

PR #1677, opened 2019-03-15
Consent banner

rupertbates:
## Why are you doing this?
Add a consent banner for users who haven't opted in or out of third party tracking

We will now check for the existence of the `GU_TK` cookie which stores a user's third party tracking settings (for this browser) and store these settings in the `common` state in Redux. 

If this cookie is not present then we show this consent banner asking them to opt in or out:

Screenshot 2019-03-15 at 17 22 20

If they choose to accept then we write the `GU_TK` cookie to record this choice and hide the banner.

[**Trello Card**](https://trello.com/c/Q6fM1gvZ/1-supporttheguardiancom)

### TODO:
- [x] Add to all relevant pages
- [ ] Add analytics (send to Ophan and GA as custom dimension)
- [ ] GTM triggers to respect tracking settings

PR #1685, opened 2019-03-19
Don't show the consent banner during post deploy tests

rupertbates:
## Why are you doing this?

The new consent banner introduced in #1677 is causing problems with the post deploy tests, this PR prevents it from showing during them.


johnduffell:
nice ! 👍

PR #1686, opened 2019-03-19
Hide Home Delivery behind a switch

walaura:
## Why are you doing this?

This hides the HD options in landing & new check out if the `hidePaperHomeDelivery` switch is flipped on. The switch is not on in prod so no changes will be visible as a result of this PR

[**Trello Card**](https://trello.com/c/8IpnkzhA/2307-stop-selling-print-home-delivery-subscriptions)

## Landing Changes
* Block changing options in reducer
* Force default option to be `Collection`
* Remove link at the bottom of the form

## Checkout changes
* Force default option to be `Collection`

## Screenshots
Screenshot 2019-03-19 at 11 52 02


support-frontend/assets/helpers/subscriptions.js
```diff
@@ -271,6 +271,18 @@ const getNewsstandSaving = (subscriptionMonthlyCost: number, newsstandWeeklyCost
 const getNewsstandPrice = (productOption: PaperProductOptions) =>
   newsstandPrices[productOption];
 
+
+// Paper product
+
+const paperHasDeliveryEnabled = (): boolean => {
+  try {
+    return window.guardian.settings.switches.experiments.hidePaperHomeDelivery.state === 'Off';
```
rupertbates: Might be good to change this switch to `paperHomeDeliveryEnabled` so that it matches how it is used in the app rather than being the opposite?

walaura: you are quite right

rupertbates: Also you can get this from `state.common.settings` rather than directly from the window object

support-frontend/assets/helpers/subscriptions.js
```diff
@@ -271,6 +271,18 @@ const getNewsstandSaving = (subscriptionMonthlyCost: number, newsstandWeeklyCost
 const getNewsstandPrice = (productOption: PaperProductOptions) =>
   newsstandPrices[productOption];
 
+
+// Paper product
+
+const paperHasDeliveryEnabled = (): boolean => {
+  try {
+    return window.guardian.settings.switches.experiments.paperHomeDeliveryEnabled.state === 'On';
```
rupertbates: You can get this from `state.common.settings` rather than directly from the window object

walaura: it's a bit of a hassle to pass around the state in all use cases and it locks this into redux. Given this is unchangeable through the app lifecycle I'm not sure it's worth it

rupertbates: Fair enough 👍

PR #1687, opened 2019-03-19
Consolidate campaign-specific settings

tomrf1:
## Why are you doing this?
The recent 'Frontline' campaign introduced some custom settings on the landing page.
This is going to be common for future campaigns, so this PR defines a `CampaignSettings` object, which can be accessed by components.

The campaign is determined by the url querystring param `campaignName=<name>`
The campaign name is stored in the redux store


support-frontend/assets/helpers/campaigns.jsx
```diff
@@ -0,0 +1,84 @@
...
+};
+
+export type Campaigns = {
+  [string]: CampaignSettings,
+};
+
+export const campaigns: Campaigns = {
+  thefrontline: {
+    formMessage: (
+      <div>
```
walaura: could this be a bit more abstracted? as in, more key:value objects? I'm not sure if i like putting css classnames, html, and react in helpers (in that order)

tomrf1: I'm not sure what you're suggesting - move the jsx into separate files?

walaura: sth like

```es6

   type FormMessage = {
     headline: ?string,
     body: ?string,
   }
```

```es6
   thefrontline: {
     formMessage: {
       headline: 'Make a contribution',
       body: 'to our dedicated series ‘The Frontline’'
     },
     ....
   }
```

tomrf1: ah I see, that would be nice. I suspect we'll need more flexibility than that, but I'll find out

joelochlann: it's a tricky one, since there always hard-to-predict changes with campaigns. maybe this could work for now and we could have conversations with design/UX about locking down the possibilities so that we don't need custom markup

walaura: if we keep the custom markup in here, would it be possible to move this helper inside the contributions flow route?

joelochlann: yeah that's a good idea, it's definitely only applicable to contributions

tomrf1: Ok, for now I've moved `campaigns.jsx` into `/pages/new-contributions-landing/` (rather than helpers/).
I'm going to create a card to follow up with UX about whether we can lock down the layout and move the jsx out of here

PR #1692, opened 2019-03-20
Make it explicit that clicking back from thankyou page does not take you back to landing page

tomrf1:
## Why are you doing this?
Clicking back from the /thankyou page takes you back to the page before the /contribute page.
This is actually preferable, because otherwise client-side routing would kick in and, based on the redux state, decide to go back to /thankyou again. Returning the user to /contribute could also get messy given the state they are now in.

This change makes this intention clearer but does not change the behaviour.

[**Trello Card**](https://trello.com/c/pvbHBgU2/121-npf-follow-up-bug-back-button-broken-by-clientside-routing-need-to-do-a-historypushstate)

### DIFF
--- a/support-frontend/assets/pages/new-contributions-landing/components/ContributionFormContainer.jsx
+++ b/support-frontend/assets/pages/new-contributions-landing/components/ContributionFormContainer.jsx
@@ -243,7 +243,11 @@ function ContributionFormContainer(props: PropTypes) {
   };
 
   return props.paymentComplete ?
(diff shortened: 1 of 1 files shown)


tomrf1:
@jlieb10 yep I wanted to make it really clear for future reference.
Unfortunately only `push` is available on this type:
```
declare export class Redirect extends React$Component<{|
    to: string | LocationShape,
    push?: boolean,
    from?: string,
    exact?: boolean,
    strict?: boolean
  |}> {}
```

PR #1694, opened 2019-03-20
Inline font-loader script separately into main.html.scala

jranks123:
## Why are you doing this?

Previously, we were calling the fontsloader script from the main JS bundle, in `page.js`. This proposed change inlines the script above the main js bundle.

this means it will get called as early as possible, and also as we are putting it in the main.scala.html, which all our pages use, it means we only have to put the call to the script into the codebase once.


support-frontend/app/lib/CustomHttpErrorHandler.scala
```diff
@@ -36,7 +36,8 @@ class CustomHttpErrorHandler(
         "Error 404",
         EmptyDiv("error-404-page"),
         Left(RefPath("error404Page.js")),
-        Left(RefPath("error404Page.css"))
+        Left(RefPath("error404Page.css")),
+        None
```
jranks123: Preferably, we wouldn't pass in `None` here, and then `fontLoaderBundle` wouldn't have to be an option.

support-frontend/app/views/main.scala.html
```diff
@@ -58,6 +59,17 @@
       </style>
     }
 
+    @linkedJs(path: RefPath) = {
+      <script type="text/javascript" src="@assets(path)"></script>
+    }
+
+      @inlineJs(inlineJs: StyleContent) = {
+        <script type="text/javascript">
```
joelochlann: one thing I'd be wary of here. there's a

```javascript
document.body.appendChild(iframe);
```

call in the JS. but as far as I can tell from looking at this, that JS would be executed before we even have a body - we're still in the `<head>` at this point.

Does this not cause any issues?

Even if not - could we not put the `<iframe>` directly into the HTML? Since we're embedding stuff serverside. If you give it an id you should still be able to bind up the event handler in the JS

jranks123: There is a `if (document.body) {` check in the script, and when George and I were pairing on this we did think about that, but we checked and in the context we are using it, there is always body at the point when this gets called. 

We could add the iFrame directly, I believe the point of doing it like this is that you could add the script to any template and it would work, without you having to manually add the iFrame in. I guess if we forget about others using this script then that would make a little more sense, slightly less javascript to run

jranks123: So we don't call `loadFonts()` until the `DOMContentLoaded` event fires in the DOM (see line 44 in the script

PR #1696, opened 2019-03-20
Read/write fonts to and from local storage

jranks123:
## Why are you doing this?
Previously, we were getting the fonts from the `gu.com/font-loader` endpoint, inserting them into the template and that was that. This PR also adds the fonts to local storage, and reads them from local storage too. The upshot is that on second page load, you don't get a FOUT (flash of unstyled text). 

Currently, I've set it up to read from localStorage then hit the `gu.com/font-loader` endpoint anyway, regardless of whether it found any fonts in there, as I didn't want to have to encode which fonts we were looking for (as this could potentially change in the future). I do, however, make sure that we don't insert them twice into the template.

### DIFF
--- a/support-frontend/assets/helpers/fontLoader.js
+++ b/support-frontend/assets/helpers/fontLoader.js
@@ -1,17 +1,50 @@
 // @flow
+import * as storage from 'helpers/storage';
+
 
 const fetchFonts = (window: Object, document: Document): void => {
   const head: null | HTMLHeadElement = document.querySelector('head');
 
-  const useFont = (font: { css: string }): void => {
-    if (head) {
+  const useFont = (font: { fontName: string, css: string }): void => {
+    const fontClassName = `gu_font__${font.fontName}`;
+    if (document.getElementsByClassName(fontClassName).length === 0 && head && font.fontName && font.css) {
       const style: HTMLStyleElement = document.createElement('style');
+      style.classList.add(`gu_font__${font.fontName}`);
       style.innerHTML = font.css;
       head.appendChild(style);
     }
   };
 
+  const saveFontToLocalStorage = (font: { fontName: string, css: string }) => {
+    const { fontName } = font;
+    if (fontName) {
+      const currentFontsString = storage.getLocal('guFonts');
+      const currentFonts = currentFontsString ? JSON.parse(currentFontsString) : {};
+      storage.setLocal('guFonts', JSON.stringify({
+        ...currentFonts,
+        [fontName]: font,
+      }));
+    }
+  };
+
+  const loadFontsFromLocalStorage = () => {
+    const fonts = storage.getLocal('guFonts');
+    if (fonts) {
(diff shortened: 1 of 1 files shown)


support-frontend/assets/helpers/fontLoader.js
```diff
@@ -1,17 +1,50 @@
...
       head.appendChild(style);
     }
   };
 
+  const saveFontToLocalStorage = (font: { fontName: string, css: string }) => {
+    const { fontName } = font;
+    if (fontName) {
+      const currentFontsString = storage.getLocal('guFonts');
+      const currentFonts = currentFontsString ? JSON.parse(currentFontsString) : {};
+      storage.setLocal('guFonts', JSON.stringify({
```
jlieb10: @jranks123 should we add a check around here to see if the typeface is already in localStorage.guFonts? Not sure if writing a second typeface with the same name will replace the previous typeface or if it will create a new one. If it creates a new typeface, we'll be incrementing the time to run loadFontsFromLocalStorage every time a user navigates around the site, I think?

support-frontend/assets/helpers/fontLoader.js
```diff
@@ -1,17 +1,50 @@
...
+
 
 const fetchFonts = (window: Object, document: Document): void => {
   const head: null | HTMLHeadElement = document.querySelector('head');
 
-  const useFont = (font: { css: string }): void => {
-    if (head) {
+  const useFont = (font: { fontName: string, css: string }): void => {
+    const fontClassName = `gu_font__${font.fontName}`;
+    if (document.getElementsByClassName(fontClassName).length === 0 && head && font.fontName && font.css) {
```
jlieb10: Good check here to make sure it's not written on the page multiple times 👍

support-frontend/assets/helpers/fontLoader.js
jranks123: This will overwrite what is there. 

In case you haven't come across it before, this bit of code uses [spread syntax ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#Spread_in_object_literals)(the `...`) :"

```
storage.setLocal('guFonts', JSON.stringify({
        ...currentFonts,
        [fontName]: font,
      }));
```
So the result of 

```{
        ...currentFonts,
        [fontName]: font,
      }
```
will be whatever currentFonts is, with the value for [fontName] either added if it isn't already in the object, or with the current value replaced with the value of `font`. 

And then `storage.setLocal` will overwrite whatever is there. 


And I think that is preferable, as it is possible that the fonts from dotcom may change, so we do want to overwrite it each time, to be safe, I think

jlieb10: Ok cool, just wanted to make sure there wasn't duplicating. Nice work.

walaura:
looking good!

PR #1698, opened 2019-03-20
Do not send full payment details to steps after CreateSubscription

pvighi:
## Why are you doing this?
Currently the "sendAcquisitionEvent" and "SendThankYouEmail" states of  support workers expect full payment details even though they don't use most of them.
In most cases only the type of payment method is needed, the exception being direct debit payments where some extra fields are needed for the confirmation email.

We are working on extending the state machine to allow adding subscriptions to existing accounts and in that scenario we won't have the full payment details available.

this PR adds a reduce version of the payment method object to be used in the states where the full one is not needed.


support-models/src/main/scala/com/gu/support/workers/states/SendThankYouEmailState.scala
```diff
@@ -2,15 +2,16 @@ package com.gu.support.workers.states
...
 import com.gu.support.encoding.CustomCodecs.{decodeLocalTime, encodeLocalTime}
 import org.joda.time.LocalDate
 
+
 case class SendThankYouEmailState(
   requestId: UUID,
   user: User,
   product: ProductType,
-  paymentMethod: PaymentMethod,
+  paymentMethod: PaymentMethodDisplayFields,
```
pvighi: before merging I'll rename the paymentMethod field here and in the acquisitionsEventState to displayPaymentMethod to make it different from the full payment method object that is in createSubscriptionState.
Unless anyone has of any reason not to do that

support-models/src/main/scala/com/gu/support/workers/states/PaymentMethodDisplayFields.scala
```diff
@@ -0,0 +1,98 @@
...
+                                          bankAccountNumberMask: String,
+                                          bankSortCode: String,
+                                          bankAccountName: String,
+                                          mandateId: Option[String],
+                                        ) extends PaymentMethodDisplayFields {
+  val `type`: String = "BankTransfer"
+
+}
+
+object WireFields {
```
pvighi: I used the intermediate wire fields to get the json serialisation to work the way I wanted. Maybe there is a more elegant  way to do this in circe without them.

rupertbates: Why do you need the intermediate classes? They look identical

support-models/src/main/scala/com/gu/support/workers/states/PaymentMethodDisplayFields.scala
```diff
@@ -0,0 +1,98 @@
...
+sealed trait JustTypePaymentMethodDisplayFields extends PaymentMethodDisplayFields
+
+case object CreditCardDisplayFields extends JustTypePaymentMethodDisplayFields{
+  override val `type` = "CreditCardReferenceTransaction"
+}
+case object PaypalDisplayFields extends JustTypePaymentMethodDisplayFields{
+  override val `type` = "Paypal"
+}
+case class DirectDebitDisplayFields(
+                                          bankAccountNumberMask: String,
```
jacobwinch: Minor: the indentation looks a bit strange here

support-models/src/main/scala/com/gu/support/workers/states/PaymentMethodDisplayFields.scala
pvighi: for non direct debit payment methods we just have an object that tells us what type of payment it is, but we need this to serialise to a json object with a type field.  
The "wire" case classes resemble the desired json structure more directly, So there is a direct debit one and a justType one that contains a type field.
We could remove some duplication if we unified the two direct debit classes (making the same class extend both traits) but I though it was cleaner this way

Having said all of that, there is probably a way to make circe generate this for us without intermediate classes, I just don't know if it would be easier to understand than this

support-models/src/main/scala/com/gu/support/workers/states/PaymentMethodDisplayFields.scala
pvighi: yeah Intellij insists in formatting it like that. I'll fix it

support-workers/src/main/scala/com/gu/support/workers/lambdas/SendAcquisitionEvent.scala
```diff
@@ -79,11 +79,11 @@ object SendAcquisitionEvent {
...
         }
 
-      def paymentProviderFromPaymentMethod(paymentMethod: PaymentMethod): thrift.PaymentProvider =
+      def paymentProviderFromPaymentMethod(paymentMethod: PaymentMethodDisplayFields): thrift.PaymentProvider =
         paymentMethod match {
-          case _: CreditCardReferenceTransaction => thrift.PaymentProvider.Stripe
-          case _: PayPalReferenceTransaction => thrift.PaymentProvider.Paypal
-          case _: DirectDebitPaymentMethod => thrift.PaymentProvider.Gocardless
+          case CreditCardDisplayFields => thrift.PaymentProvider.Stripe
+          case PaypalDisplayFields => thrift.PaymentProvider.Paypal
```
pvighi: This is a very simple change but I don't know how to test this. 
Is there a way to make sure this lambda still works as expected in CODE ?

pvighi:
We decided to do this in a different way so this is probably not needed, at least in its current form

PR #1699, opened 2019-03-20
Add ticker to storybook

joelochlann:
Also

- moves it into a general components directory (`components/ticker`)
- includes CSS directly with ES6 import

In sum: makes the ticker more reusable

### DIFF
--- a/support-frontend/assets/components/ticker/contributionTicker.jsx
+++ b/support-frontend/assets/components/ticker/contributionTicker.jsx
@@ -4,6 +4,8 @@
 import React, { Component } from 'react';
 import { classNameWithModifiers } from 'helpers/utilities';
 
+import './contributionTicker.scss';
+
 // ---- Types ----- //
 type StateTypes = {|
   totalSoFar: number,
@@ -36,7 +38,7 @@ const percentageTotalAsNegative = (total: number, goal: number) => {
 
 
 // ----- Component ----- //
-export class ContributionTicker extends Component<PropTypes, StateTypes> {
+export default class ContributionTicker extends Component<PropTypes, StateTypes> {
 
   constructor(props: PropTypes) {
     super(props);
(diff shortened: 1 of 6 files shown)


support-frontend/assets/components/ticker/contributionTicker.scss
```diff
@@ -1,4 +1,4 @@
-// -----  gu-sass ----- //
+@import '~stylesheets/gu-sass/gu-sass';
```
jlieb10: Maybe add a note to say this is bc the component is in Storybook

support-frontend/assets/components/ticker/contributionTicker.jsx
```diff
@@ -36,7 +38,7 @@ const percentageTotalAsNegative = (total: number, goal: number) => {
 
 
 // ----- Component ----- //
-export class ContributionTicker extends Component<PropTypes, StateTypes> {
+export default class ContributionTicker extends Component<PropTypes, StateTypes> {
```
jlieb10: Maybe we should document work TBD in the class somewhere. Like: (1) we'd like to edit the component to get its values either from JSON or with props and (2) add some documentation about the structure of the JSON the component expects.

support-frontend/assets/components/ticker/contributionTicker.scss
walaura: Nope! this import is also picked up by `support-frontend` [as normal](https://github.com/guardian/support-frontend/wiki/CSS) we started this migration a bit ago

support-frontend/assets/components/ticker/contributionTicker.jsx
joelochlann: I've actually documented those in this Trello card instead: https://trello.com/c/qXxukzxY/395-ticker-in-storybook-improvements-for-future-campaigns

PR #1705, opened 2019-03-21
Fix Prout seen messaging and add comment

jacobwinch:
## Why are you doing this?

To fix the Prout 'seen' message (see https://github.com/guardian/support-frontend/blob/master/.prout.json#L6-L8 and https://github.com/guardian/prout/pull/54).

Also adds a comment to explain the remote driver setup.


jorgeazevedo:
I get it! \o/

PR #1708, opened 2019-03-22
Add CreatedRequestId__c to subscription object in Zuora

jacobwinch:
## Why are you doing this?

This is groundwork that will help us to modify [this](https://github.com/guardian/support-frontend/blob/master/support-workers/src/main/scala/com/gu/support/workers/GetRecurringSubscription.scala#L16-L37) check to cope with:

- safely adding subscriptions products 
- safely adding a second contribution or subscription to an existing billing account (a new feature currently being worked on by Supporter Experience) 

We now add the request id to all subscriptions created by `support-workers`. This will allow us to add a safety check with prevents us from creating multiple subscriptions for a single `support-workers` execution (important because this lambda has automatic retries). I've found having the request id on the billing account useful for debugging in the past (and I don't think there's any cost associated with this), so we still write it at that level too.

[**Trello Card**](https://trello.com/c/hagyNZHy/2301-fix-existing-subscription-check-in-support-workers)

## Changes

* Add CreatedRequestId__c to `Subscription` (and fix related tests)


twrichards:
Having the CreatedRequestId at account level (existing functionality based on the description) does it get overwritten if subsequent requests modify that same billing account?

jacobwinch:
> Having the CreatedRequestId at account level (existing functionality based on the description) does it get overwritten if subsequent requests modify that same billing account?

No - and I don't think we'd ever want to do this given the name of the field.

PR #1711, opened 2019-03-22
Personal details form

rupertbates:
## Why are you doing this?
To reduce code duplication. We had the same fields defined twice in the paper and DP checkout forms, this PR extracts them to a new component.


support-frontend/assets/pages/digital-subscription-checkout/components/checkoutForm.jsx
```diff
@@ -163,58 +157,16 @@ function CheckoutForm(props: PropTypes) {
...
-                error={firstError('telephone', props.formErrors)}
+              <PersonalDetails
+                firstName={props.firstName}
+                setFirstName={props.setFirstName}
+                lastName={props.lastName}
+                setLastName={props.setLastName}
+                email={props.email}
+                telephone={props.telephone}
+                setTelephone={props.setTelephone}
+                formErrors={((props.formErrors: any): FormError<PersonalDetailsFormField>[])}
```
rupertbates: I couldn't find any other way to convince Flow to let me pass the errors array in even though we know that structurally it is ok

support-frontend/assets/components/subscriptionCheckouts/personalDetails.jsx
```diff
@@ -0,0 +1,92 @@
+// @flow
+import React from 'react';
+import { firstError } from '../../helpers/subscriptionsForms/validation';
```
walaura: I'm not sure I like this import style, could we stick to absolute imports when they start at the root?

rupertbates: Yup Intellij added this automatically but I agree that absolute imports are nicer

PR #1714, opened 2019-03-22
Add a little info to logging when a support-workers execution is started for a subscription

lmath:
## Why are you doing this?
Just wanted to make it a little easier to look at the logs and see subscription signups, and for which product. 👀 

We are currently just logging billing period.

[**Trello Card**](https://trello.com) - none


rupertbates:
It would be good to add this information into the request info object as well

PR #1717, opened 2019-03-25
Fix patrons code

walaura:
## Why are you doing this?
It was broken and not applying itself (ah types), now only the wording is broken (it's free for 1 year)

Screenshot 2019-03-25 at 13 21 54


jacobwinch:
@walaura - do we need a separate card to look into the problem with the wording?

PR #1718, opened 2019-03-25
Release support models 62

rupertbates:
## Why are you doing this?

To make the library projects releasable, they have not worked since the move to the mono repo but it turns out we do still need to sometimes release them separately because the payment api depends on support-models.


ademuk:
👍 LGTM. I don't have permission to approve.

PR #1719, opened 2019-03-25
Use t3.small instances

jacobwinch:
## Why are you doing this?

Preparing for 2019-2020 reservations.

https://aws.amazon.com/ec2/instance-types/t3/

[**Trello Card**](https://trello.com/c/ihXGzPyy/2313-work-to-support-aws-instance-reservations)

## Changes

* Use t3.small instead of t2.small


jacobwinch:
TeamCity build failed due to an unrelated problem, I'm re-running it now.

PR #1720, opened 2019-03-26
Make contributions landing page cacheable

jranks123:
## Why are you doing this?
This PR makes the contributions landing page a cached page. Previously, the page was not cached for 2 main reasons.

- If the user was signed in, we were putting the user's details in the response from the server (for the autofilling of the fields). We have moved this logic to the clientside

- We were including a unique CSRF token with each response from the server. Having looked at all the endpoints that the clientside calls that currently have CSRF protection, we decided that we no longer needed CSRF protection on them and thus we are able to remove this level of protection from these endpoints


johnduffell:
great!

support-frontend/assets/helpers/paymentIntegrations/readerRevenueApis.js
```diff
@@ -225,11 +222,10 @@ function postRegularPaymentRequest(
 function setPasswordGuest(
   password: string,
   guestAccountRegistrationToken: string,
-  csrf: CsrfState,
 ): Promise<boolean> {
 
   const data = { password, guestAccountRegistrationToken };
-  return logPromise(fetch(`${routes.contributionsSetPasswordGuest}`, requestOptions(data, 'same-origin', 'PUT', csrf)))
+  return logPromise(fetch(`${routes.contributionsSetPasswordGuest}`, requestOptions(data, 'same-origin', 'PUT', null)))
```
johnduffell: null???!!!

johnduffell: 😨

jranks123: Yup!

johnduffell: oh wait that's javascript, that's alright then 😆

johnduffell:
could have done this in 3 PRs to be honest.  Move the logic to the client side would have been one, then remove the CSRF as a second.  Then the final one would have been self contained and easy to revert in case of issues.

support-frontend/app/controllers/IdentityController.scala
```diff
@@ -81,6 +83,27 @@ class IdentityController(
...
+    type Attempt[A] = EitherT[Future, String, A]
+    request.user.traverse[Attempt, IdUser](identityService.getUser(_))
+      .fold(
+        _ => {
+          SafeLogger.info(s"Failed to retrieve IdUser as no authenticated user found in request")
+          InternalServerError
+        },
+        user => {
+          user.fold({
+            SafeLogger.info(s"User not found")
```
johnduffell:
```suggestion
            SafeLogger.info(s"User not found: ${request.user}")
```

support-frontend/app/controllers/IdentityController.scala
```diff
@@ -81,6 +83,27 @@ class IdentityController(
         )
     }
   }
+
+  def getUser(): Action[AnyContent] = maybeAuthenticatedAction().async { implicit request => {
+    type Attempt[A] = EitherT[Future, String, A]
+    request.user.traverse[Attempt, IdUser](identityService.getUser(_))
+      .fold(
+        _ => {
```
johnduffell: what's the underscore? should it be logged? and should it be a warn rather than an info?

PR #1723, opened 2019-03-26
Revert "Use t3.small instances"

jacobwinch:
Reverts guardian/support-frontend#1719

I've just seen this issue when deploying `members-data-api`:

`Launching a new EC2 instance. Status Reason: You have requested more instances (21) than your current instance limit of 20 allows for the specified instance type. Please visit http://aws.amazon.com/contact-us/ec2-request to request an adjustment to this limit. Launching EC2 instance failed.`

This may cause issues with other deploys so I'd like to back this out until we get the limit increased.

### DIFF
--- a/support-frontend/cloud-formation/cfn.yaml
+++ b/support-frontend/cloud-formation/cfn.yaml
@@ -31,7 +31,7 @@ Mappings:
     CODE:
       MaxInstances: 2
       MinInstances: 1
-      InstanceType: t3.small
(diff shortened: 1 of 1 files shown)


jacobwinch:
Wow, AWS support are fast:

`We have approved and processed your limit increase request(s).  It can sometimes take up to 30 minutes for this to propagate and become available for use.`

Closing this.

PR #1730, opened 2019-03-27
Support existing payment methods in support workers ( only cc for now)

pvighi:
## Why are you doing this?
This will enable customers to reuse their existing payment methods instead of having to enter the same details again. 
In these cases support frontend will execute the state machine providing the id of the Zuora account that contains the payment method we want to use. In this version this will only work for card reference payment methods, the state machine will fail if it finds any other type of payment method.

To make this work we will add a fork in the support workers state machine and instead of creating a new payment method it will get the payment details from Zuora. After this the state machine will resume normal execution. 

Before:
Screenshot 2019-03-27 at 14 01 22

After:
Screenshot 2019-03-27 at 14 00 25


support-workers/src/main/scala/com/gu/support/workers/lambdas/GetPaymentMethod.scala
```diff
@@ -0,0 +1,80 @@
...
+
+    val zuoraService = services.zuoraService
+    for {
+      account <- zuoraService.getObjectAccount(state.paymentFields.billingAccountId)
+      paymentId <- getOrFailWithMessage(account.DefaultPaymentMethodId, "Account has no default payment method")
+      getPaymentMethodResponse <- zuoraService.getPaymentMethod(paymentId)
+      _ <- ifFalseReturnError(getPaymentMethodResponse.paymentMethodStatus == "Active", "Account has a non active default payment method")
+      sfContactId <- getOrFailWithMessage(account.sfContactId__c, "account has no sfContact!")
+      paymentMethod <- asFuture(toPaymentMethod(getPaymentMethodResponse))
+      sfContact = SalesforceContactRecord(sfContactId, account.CrmId)
```
pvighi: Is it always safe to use account.crmId as the salesforce account id here ?

support-models/src/main/scala/com/gu/support/zuora/api/response/Responses.scala
```diff
@@ -76,6 +78,73 @@ object GetAccountResponse {
 
 case class GetAccountResponse(success: Boolean, basicInfo: BasicInfo) extends ZuoraResponse
 
+//this response cannot extend ZuoraResponse because this endpoint doesn't return a 'success' boolean field
+case class GetObjectAccountResponse(
+   IdentityId__c: Option[String],
+   sfContactId__c: Option[String],
+   CrmId: String,
```
pvighi: @paulbrown1982  Is it safe to assume every account should have crmId defined and fail to parse the response if they don't?
I checked in Zuora prod and we seem to just have 7 or so accounts with no crmId (1 with null the rest with empty strings)

support-workers/src/main/scala/com/gu/support/workers/lambdas/GetPaymentMethod.scala
```diff
@@ -0,0 +1,81 @@
...
+class GetPaymentMethod(servicesProvider: ServiceProvider = ServiceProvider)
+    extends ServicesHandler[GetPaymentMethodState, CreateZuoraSubscriptionState](servicesProvider) {
+
+  def this() = this(ServiceProvider)
+
+  override protected def servicesHandler(state: GetPaymentMethodState, requestInfo: RequestInfo, context: Context, services: Services) = {
+
+    val zuoraService = services.zuoraService
+    for {
+      account <- zuoraService.getObjectAccount(state.paymentFields.billingAccountId)
```
pvighi: Is there any check we should do to verify this account belongs to the user in state.user ?

support-models/src/main/scala/com/gu/support/workers/states/GetPaymentMethodState.scala
```diff
@@ -0,0 +1,19 @@
+package com.gu.support.workers.states
+
+import java.util.UUID
+import com.gu.support.encoding.Codec.deriveCodec
+import com.gu.support.workers.{User, _}
+
+case class GetPaymentMethodState(
```
jacobwinch: Does it make sense to define a new type here? 

If I understand it correctly, the JSON from support-frontend could be passed to this lambda _or_ CreatePaymentMethod, which means that anyone who wants to add new input state needs to remember to update (and test) both scenarios.

It might be less error prone to define a single type `StartExecutionState` (or similar), which is passed to both lambdas?

support-workers/src/main/scala/com/gu/support/workers/lambdas/GetPaymentMethod.scala
```diff
@@ -0,0 +1,83 @@
...
+          creditCardExpirationYear = cardResponse.creditCardExpirationYear,
+          creditCardType = cardResponse.creditCardType
+        )
+      )
+    }
+    case other => Left(s"unsupported payment method: '$other'")
+
+  }
+
+  def errorResponse(msg:String): Future[Nothing] = Future.failed(new Exception(msg))
```
jacobwinch: At the moment an unhandled exception will cause the Step Function execution to fail and result in an alert. If any of these errors are expected we should handle them gracefully here: https://github.com/guardian/support-frontend/blob/master/support-workers/src/main/scala/com/gu/support/workers/lambdas/FailureHandler.scala#L49-L68

pvighi: @twrichards  might know more about this, but as far as I understand frontend will only offer to reuse payment methods in cases where all these validations pass.

twrichards: Correct and so we would want alerts as it would indicate something is mis-implemented or being used maliciously

pvighi: Also, I just realised that a lot of this validations are not needed. I was initially replicating the csr acquisitions code where we need to return a nice error message to the user. But in this case i can make the fields non optional in the case class and let the json parsing fail if something required is missing

pvighi: .. Unless we want to leave a more generic "get zuora account" function in support workers for future reuse
🤔

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreatePaymentMethod.scala
```diff
@@ -26,7 +26,7 @@ class CreatePaymentMethod(servicesProvider: ServiceProvider = ServiceProvider)
       .map(paymentMethod =>
         HandlerResult(
           getCreateSalesforceContactState(state, paymentMethod),
-          requestInfo.appendMessage(s"Payment method is ${paymentMethod.toFriendlyString}")
```
jacobwinch: Did we intend to remove this here?

pvighi: No, this is a mistake. I should not have changed this lambda

pvighi: In fact maybe the new lambda that retrieves an existing payment method should also put this message in requestInfo to keep it consistent

twrichards: rebased out this line 😄

PR #1731, opened 2019-03-27
label existing payment reuse in acquisition event

twrichards:
## Why are you doing this?
Before rolling out 'existing payment options' feature (see https://github.com/guardian/support-frontend/pull/1673 and https://github.com/guardian/support-frontend/pull/1730) we must be able to attribute acquisitions accordingly (for use in the 'Acquisitions Slicer' for example).

## Changes
Wrap the `SendAcquisitionEventState` (previously what was passed through to build the `thrift.Acquisition`) along with the `RequestInfo` (which contains `accountExists` flag) in a new case class, which can be passed as the single paramater to the `AcquisitionSubmissionBuilder`... so that it can be used to add `"REUSED_EXISTING_PAYMENT_METHOD"` to the `labels` field of the acquisition event if applicable (i.e. if `accountExists` is true and hence the payment method was reused/cloned).

### DIFF
--- a/support-workers/src/main/scala/com/gu/support/workers/lambdas/SendAcquisitionEvent.scala
+++ b/support-workers/src/main/scala/com/gu/support/workers/lambdas/SendAcquisitionEvent.scala
@@ -34,9 +34,18 @@ class SendAcquisitionEvent(serviceProvider: ServiceProvider = ServiceProvider)
   ): FutureHandlerResult = {
     SafeLogger.info(s"Sending acquisition event to ophan: ${state.toString}")
     // Throw any error in the EitherT monad so that in can be processed by ErrorHandler.handleException
-    services.acquisitionService.submit(state).fold(errors => throw AnalyticsServiceErrorList(errors), _ => HandlerResult(Unit, requestInfo))
+    services.acquisitionService.submit(
+      SendAcquisitionEventStateAndRequestInfo(state, requestInfo)
+    ).fold(
+      errors => throw AnalyticsServiceErrorList(errors), _ => HandlerResult(Unit, requestInfo)
+    )
   }
 }
+
+case class SendAcquisitionEventStateAndRequestInfo(
+  state: SendAcquisitionEventState,
+  requestInfo: RequestInfo
+)
 
 object SendAcquisitionEvent {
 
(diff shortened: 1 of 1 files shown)


support-workers/src/main/scala/com/gu/support/workers/lambdas/SendAcquisitionEvent.scala
```diff
@@ -109,36 +118,37 @@ object SendAcquisitionEvent {
         }
       }
 
-      override def buildAcquisition(state: SendAcquisitionEventState): Either[String, thrift.Acquisition] = {
-        val (productType, productAmount) = state.product match {
+      override def buildAcquisition(a: SendAcquisitionEventStateAndRequestInfo): Either[String, thrift.Acquisition] = {
```
pvighi: maybe this parameter could be named stateAndInfo or something more descriptive than 'a'

twrichards: done!

support-workers/src/main/scala/com/gu/support/workers/lambdas/SendAcquisitionEvent.scala
```diff
@@ -34,29 +34,38 @@ class SendAcquisitionEvent(serviceProvider: ServiceProvider = ServiceProvider)
...
-    services.acquisitionService.submit(state).fold(errors => throw AnalyticsServiceErrorList(errors), _ => HandlerResult(Unit, requestInfo))
+    services.acquisitionService.submit(
+      SendAcquisitionEventStateAndRequestInfo(state, requestInfo)
+    ).fold(
+      errors => throw AnalyticsServiceErrorList(errors), _ => HandlerResult(Unit, requestInfo)
+    )
   }
 }
 
+case class SendAcquisitionEventStateAndRequestInfo(
```
pvighi: minor: I was going to complain about the name of this class but then I have to propose something better..
maybe AcquisitionBuildData, even though it's kinda meaningless at least it will still make sense if we decide to add or remove something and also it's shorter.
Just a suggestion, feel free to ignore it 100%
also, sorry this comment should have come before my other one

PR #1752, opened 2019-04-08
Catalog warnings

rupertbates:
## Why are you doing this?

If the Zuora catalog fails to load on app start this is a big problem for subs because pricing information will be unavailable, this PR adds a CloudWatch alarm so that we will get notified if that happens.

[**Trello Card**](https://trello.com/c/RLwKIy8d/2218-handle-catalog-load-failures-gracefully)

## Changes

* Send an event to CloudWatch to record catalog load failures in PROD
* Add a CloudWatch alarm to fire on failure

### DIFF
--- a/support-frontend/build.sbt
+++ b/support-frontend/build.sbt
@@ -20,7 +20,6 @@ libraryDependencies ++= Seq(
   "com.amazonaws" % "aws-java-sdk-stepfunctions" % awsClientVersion,
   "com.amazonaws" % "aws-java-sdk-sts" % awsClientVersion,
   "com.amazonaws" % "aws-java-sdk-s3" % awsClientVersion,
-  "com.amazonaws" % "aws-java-sdk-cloudwatch" % awsClientVersion,
   "org.typelevel" %% "cats-core" % "1.0.1",
   "com.dripower" %% "play-circe" % "2609.1",
   "com.gu" %% "fezziwig" % "0.8",
--- a/support-frontend/cloud-formation/cfn.yaml
+++ b/support-frontend/cloud-formation/cfn.yaml
@@ -217,8 +217,6 @@ Resources:
       Name: !Sub ${Stack}-${Stage}-${App}
       Port: 9000
       Protocol: HTTP
-      VpcId:
-        Ref: VpcId
       HealthCheckIntervalSeconds: 10
       HealthCheckPath: /healthcheck
       HealthCheckPort: 9000
@@ -318,6 +316,24 @@ Resources:
     DependsOn:
     - TargetGroup
     - ElasticLoadBalancer
+
+  CatalogLoadingFailureAlarm:
+    Type: AWS::CloudWatch::Alarm
+    Condition: CreateProdResources
+    Properties:
+      AlarmActions:
+        - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:reader-revenue-dev
+      AlarmName: Support Frontend Catalog Load Failure
+      MetricName: CatalogLoadingFailure
+      Namespace: support-frontend
+      Dimensions:
+        - Name: Environment
+          Value: PROD
+      ComparisonOperator: GreaterThanOrEqualToThreshold
+      Threshold: 1
+      Period: 60
+      EvaluationPeriods: 1
+      Statistic: Average
 
   StateMachineUnavailableMetric:
     Type: "AWS::Logs::MetricFilter"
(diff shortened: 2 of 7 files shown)


support-frontend/cloud-formation/cfn.yaml
```diff
@@ -217,8 +217,6 @@ Resources:
       Name: !Sub ${Stack}-${Stage}-${App}
       Port: 9000
       Protocol: HTTP
-      VpcId:
```
jacobwinch: Was about to ask about this and then I noticed that we specify the VPC twice 🤦‍♂️

support-frontend/build.sbt
```diff
@@ -20,7 +20,6 @@ libraryDependencies ++= Seq(
   "com.amazonaws" % "aws-java-sdk-stepfunctions" % awsClientVersion,
   "com.amazonaws" % "aws-java-sdk-sts" % awsClientVersion,
   "com.amazonaws" % "aws-java-sdk-s3" % awsClientVersion,
-  "com.amazonaws" % "aws-java-sdk-cloudwatch" % awsClientVersion,
```
jacobwinch: Does the functionality we need get pulled in by another sdk then?

support-frontend/cloud-formation/cfn.yaml
```diff
@@ -319,6 +317,24 @@ Resources:
     - TargetGroup
     - ElasticLoadBalancer
 
+  CatalogLoadingFailureAlarm:
+    Type: AWS::CloudWatch::Alarm
+    Condition: CreateProdResources
+    Properties:
+      AlarmActions:
```
jacobwinch: Might be worth adding an [AlarmDescription](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cw-alarm.html#cfn-cloudwatch-alarms-alarmdescription) for new engineers who may not understand the impact of this?

support-frontend/cloud-formation/cfn.yaml
```diff
@@ -319,6 +317,24 @@ Resources:
...
+      Namespace: support-frontend
+      Dimensions:
+        - Name: Environment
+          Value: PROD
+      ComparisonOperator: GreaterThanOrEqualToThreshold
+      Threshold: 1
+      Period: 60
+      EvaluationPeriods: 1
+      Statistic: Average
+
```
jacobwinch:
```suggestion
TreatMissingData: notBreaching

```

I think we may need this because it [defaults to missing](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cw-alarm.html#cfn-cloudwatch-alarms-treatmissingdata), which is probably not what we want (see: https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/AlarmThatSendsEmail.html#alarms-and-missing-data)

support-services/src/main/scala/com/gu/aws/AwsCloudWatchMetricPut.scala
```diff
@@ -27,7 +26,7 @@ object AwsCloudwatchMetricPut {
       dimensions: Map[MetricDimensionName, MetricDimensionValue]
   )
 
-  def apply(s3Client: AmazonCloudWatch)(request: MetricRequest): Try[Unit] = {
+  def apply(client: AmazonCloudWatch)(request: MetricRequest): Try[Unit] = {
```
jacobwinch: thanks for fixing this name 👍

support-frontend/build.sbt
jacobwinch: I can now see that this has just moved into another `build.sbt` file 🙈

support-frontend/cloud-formation/cfn.yaml
rupertbates: Have discussed offline and we are going to leave this

support-services/src/main/scala/com/gu/aws/AwsCloudWatchMetricPut.scala
```diff
@@ -46,17 +45,26 @@ object AwsCloudwatchMetricPut {
     metricDatum1.setValue(1.00)
     metricDatum1.setUnit(StandardUnit.Count)
     putMetricDataRequest.getMetricData.add(metricDatum1)
-    Try(s3Client.putMetricData(putMetricDataRequest)).map(_ => ())
+    Try(client.putMetricData(putMetricDataRequest)).map(_ => Unit)
```
johnduffell: Old PR I know, but I don't think this change is correct.
`Unit` is the type, `()` is the value.  `Unit` is a companion object of type `Unit.type` which is not in itself of type `Unit`.
This does happen to pass the type checker, because it will discard a non Unit value in Unit context, and it does work at run time because the value is discarded, but it's technically wrong 😎

PR #1755, opened 2019-04-11
Add guardian weekly

rupertbates:
## Why are you doing this?

This is the first step towards building the Guardian Weekly checkout - support-workers will now be able to create GW subscriptions

Still TODO:
- [ ] 6 for 6
- [ ] Emails


support-models/src/main/scala/com/gu/support/workers/Products.scala
```diff
@@ -42,22 +42,33 @@ case class Paper(
   override def describe: String = s"Paper-$fulfilmentOptions-$productOptions"
 }
 
+case class GuardianWeekly(
+  currency: Currency,
+  billingPeriod: BillingPeriod,
+  fulfilmentOptions: FulfilmentOptions,
```
lmath: Why do we need fulfilmentOptions here?

lmath: Ohhh I missed that "Domestic" and "RestOfWorld" are FulfilmentOptions

PR #1758, opened 2019-04-17
Make contribution types configurable

tomrf1:
## Why are you doing this?
Contribution types (aka frequencies) are now configurable from the Support Admin Console.
This lets us configure which contributionTypes are available in each region, and which is the default.

This change gets the contributionTypes settings from s3 (currently they're in the javascript) and sends them to the client.

## Changes

* Server gets the settings from `contributionTypes.json` (along with switches and amounts settings)
* Client uses the settings in window.guardian.settings

### DIFF
--- a/support-frontend/app/admin/settings/ContributionTypes.scala
+++ b/support-frontend/app/admin/settings/ContributionTypes.scala
@@ -1,0 +1,39 @@
+package admin.settings
+
+import com.gu.support.encoding.JsonHelpers
+import io.circe.{Decoder, Encoder}
+
+sealed trait ContributionType
+case object ONE_OFF extends ContributionType
+case object MONTHLY extends ContributionType
+case object ANNUAL extends ContributionType
+
+case object ContributionType {
+  implicit val contributionTypeEncoder: Encoder[ContributionType] = Encoder.encodeString.contramap[ContributionType](_.toString)
+
+  implicit val contributionTypeDecoder: Decoder[ContributionType] = JsonHelpers.decodeStringAndCollect {
+    case "ONE_OFF" => ONE_OFF
+    case "MONTHLY" => MONTHLY
+    case "ANNUAL" => ANNUAL
+  }
+}
+
+case class ContributionTypeSetting(contributionType: ContributionType, isDefault: Option[Boolean])
+
+case class ContributionTypes(
+  GBPCountries: Seq[ContributionTypeSetting],
+  UnitedStates: Seq[ContributionTypeSetting],
+  EURCountries: Seq[ContributionTypeSetting],
+  International: Seq[ContributionTypeSetting],
+  Canada: Seq[ContributionTypeSetting],
+  AUDCountries: Seq[ContributionTypeSetting],
+  NZDCountries: Seq[ContributionTypeSetting]
+)
+
+object ContributionTypes {
+  import io.circe.generic.auto._
+  import ContributionType._
+
+  implicit val contributionTypesDecoder = Decoder[ContributionTypes]
+  implicit val contributionTypesEncoder = Encoder[ContributionTypes]
+}
--- a/support-frontend/assets/helpers/globals.js
+++ b/support-frontend/assets/helpers/globals.js
@@ -57,6 +57,15 @@ const getSettings = (): Settings => getGlobal('settings') || {
       ANNUAL: [],
     },
   },
+  contributionTypes: {
+    GBPCountries: [],
+    UnitedStates: [],
+    EURCountries: [],
+    AUDCountries: [],
+    International: [],
+    NZDCountries: [],
+    Canada: [],
+  },
 };
 
 const getProductPrices = (): ?ProductPrices => getGlobal('productPrices');
--- a/support-frontend/assets/helpers/globals.js
+++ b/support-frontend/assets/helpers/globals.js
@@ -16,46 +16,24 @@ function getGlobal<T>(path: string = ''): ?T {
   return null;
 }
 
+const emptyAmountsSettings = {
+  ONE_OFF: [],
+  MONTHLY: [],
+  ANNUAL: [],
+};
+
 const getSettings = (): Settings => getGlobal('settings') || {
   switches: {
     experiments: {},
   },
   amounts: {
-    GBPCountries: {
-      ONE_OFF: [],
-      MONTHLY: [],
-      ANNUAL: [],
-    },
-    UnitedStates: {
-      ONE_OFF: [],
-      MONTHLY: [],
-      ANNUAL: [],
-    },
-    EURCountries: {
-      ONE_OFF: [],
-      MONTHLY: [],
-      ANNUAL: [],
-    },
-    AUDCountries: {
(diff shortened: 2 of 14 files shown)


support-frontend/assets/helpers/globals.js
```diff
@@ -16,46 +16,33 @@ function getGlobal<T>(path: string = ''): ?T {
...
+    GBPCountries: emptyAmountsSettings,
+    UnitedStates: emptyAmountsSettings,
+    EURCountries: emptyAmountsSettings,
+    AUDCountries: emptyAmountsSettings,
+    International: emptyAmountsSettings,
+    NZDCountries: emptyAmountsSettings,
+    Canada: emptyAmountsSettings,
+  },
+  contributionTypes: {
+    GBPCountries: [],
```
tomrf1: I wonder if it's worth combining the amounts + contributionTypes settings into one model, e.g.

```
type Amount = {
  value: string,
  isDefault?: boolean,
}

type ContributionTypeSetting = {
  isDefault?: boolean,
  amounts: Amounts[],
}

type ContributionTypesSettings = {
  [CountryGroupId]: ContributionTypeSetting[],
}
```

support-frontend/app/admin/settings/ContributionTypes.scala
```diff
@@ -0,0 +1,39 @@
...
+
+sealed trait ContributionType
+case object ONE_OFF extends ContributionType
+case object MONTHLY extends ContributionType
+case object ANNUAL extends ContributionType
+
+case object ContributionType {
+  implicit val contributionTypeEncoder: Encoder[ContributionType] = Encoder.encodeString.contramap[ContributionType](_.toString)
+
+  implicit val contributionTypeDecoder: Decoder[ContributionType] = JsonHelpers.decodeStringAndCollect {
```
joelochlann: is there a circe macro that could avoid this boilerplate? or maybe using enumeratum? just curious!

tomrf1: good suggestion - looks like enumeratum solves this https://github.com/lloydmeta/enumeratum#circe
([explanation of why it doesn't do this out of the box](https://stackoverflow.com/a/37015184))

support-models/src/main/scala/com/gu/support/encoding/JsonHelpers.scala
```diff
@@ -98,4 +98,11 @@ object JsonHelpers {
     def getField(key: String) = json.hcursor.downField(key).focus
   }
 
+  //Decodes an expected json string to T if the given partial function is defined for the string value
+  def decodeStringAndCollect[T](pf: PartialFunction[String,T]): Decoder[T] = Decoder.decodeString.emap { s =>
```
joelochlann: is this what works in conjunction with the stuff in `ContributionTypes.scala` to encode/decode a sum type to JSON? (asking for my own Scala curiosity!)

support-frontend/assets/pages/new-contributions-landing/contributionsLandingInit.js
```diff
@@ -63,13 +64,25 @@ function getInitialPaymentMethod(
...
+      contributionTypes[countryGroupId][0].contributionType;
+  };
+
+  const contributionType =
+    getContributionTypeFromUrlOrElse(getContributionTypeFromSessionOrElse(getDefaultContributionType()));
+
   return (
     // make sure we don't select a contribution type which isn't on the page
-    getValidContributionTypes(countryGroupId).includes(contributionType)
+    contributionTypes[countryGroupId].find(ct => ct.contributionType === contributionType)
```
joelochlann: the overall logic of this function is a bit hard to follow for me now. especially the appearance of the `contributionTypes[countryGroupId][0].contributionType` fallback in two places - is there a way to clarify?

support-frontend/assets/helpers/globals.js
joelochlann: Yes! We could put everything that's specific to region + contribution type in there.

joelochlann: (e.g. possibly the "editorialise amounts" copy, `Contributing £50 works out as 14p a day`)

support-models/src/main/scala/com/gu/support/encoding/JsonHelpers.scala
tomrf1: yep, it's just a convenient way to match on a json string, for case object decoding

support-frontend/assets/pages/new-contributions-landing/components/ContributionTypeTabs.jsx
```diff
@@ -18,6 +17,7 @@ import type { Switches } from 'helpers/settings';
 import type { CountryGroupId } from 'helpers/internationalisation/countryGroup';
 import { type State } from '../contributionsLandingReducer';
 import { updateContributionTypeAndPaymentMethod } from '../contributionsLandingActions';
+import type { ContributionTypes, ContributionTypeSetting } from '../../../helpers/contributions';
```
joelochlann: I think I'd rather not have a relative path to `helpers` in case the file moves... I know that there are other relative paths, but they're to very local contribution-specific files rather than the global helpers

support-frontend/assets/pages/new-contributions-landing/contributionsLandingInit.js
```diff
@@ -45,6 +45,7 @@ import type { ExistingPaymentMethod } from '../../helpers/existingPaymentMethods
 import { setExistingPaymentMethods } from '../../helpers/page/commonActions';
 import { doesUserAppearToBeSignedIn } from '../../helpers/user/user';
 import { isSwitchOn } from 'helpers/globals';
+import type { ContributionTypes } from '../../helpers/contributions';
```
joelochlann: ahh - I see lots more already existing relative paths to `helpers` here!

support-frontend/test/codecs/CirceDecodersTest.scala
```diff
@@ -192,7 +192,51 @@ class CirceDecodersTest extends WordSpec with MustMatchers {
           |            { "value": "25", "isDefault": true }
           |        ]
           |    }
-          |  }
+          |  },
+          |  "contributionTypes": {
+          |"GBPCountries": [
```
joelochlann: I know it's a bit of a pain but it would be nice to have correct indentation here to help skim-reading the file

support-frontend/assets/helpers/globals.js
tomrf1: the migration could be quite complicated so I'll create a trello card to address this

support-frontend/assets/pages/new-contributions-landing/contributionsLandingInit.js
tomrf1: I've just pushed https://github.com/guardian/support-frontend/pull/1758/commits/3c3495042def3dfdc794905cc47b12427e4d7f5b to attempt to address this.
It checks the url/session (which may or may not be valid) before using the settings.

PR #1760, opened 2019-04-23
Add vertical space to text under sub start radio chonkies

lucymonie:
## Why are you doing this?
This is work detailed in this [**Trello Card**](https://trello.com/c/h7RzWj8t/2348-vertical-spacing-below-form-ui)

## Changes

* Adds space (hopefully as required - @jesseyuen, please confirm is this looks okay)

### Before
Screen Shot 2019-04-24 at 10 22 16

### After
Screen Shot 2019-04-23 at 12 31 53

### DIFF
--- a/support-frontend/assets/components/text/text.scss
+++ b/support-frontend/assets/components/text/text.scss
@@ -42,3 +42,7 @@
 .component-text__sans {
   @include gu-fontset-body-sans;
 }
+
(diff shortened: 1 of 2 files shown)


walaura:
hiya! the `<Rows>` component that wraps these has a `size` prop that controls the spacing between elements, would be a cleaner way to support the same behaviour!

PR #1766, opened 2019-04-25
Social share and button update on thank you pages

jlieb10:
## Why are you doing this?

We released the new Thank You page designs knowing we had a few things left to be done. The most major thing to be done was releasing the social share buttons, which this PR addresses. This PR also addresses one minor concern which was that we were using a non-standard button. 

[**Trello Card**](https://trello.com/c/oo2MNCqE/959-thank-you-page-redesign-ft-social-share-v2)

## Changes

* Releasing Social Share buttons on "thank you" page and "password set" page
* Using standard `button` component across thank you pages

## Screenshots
Thank you page with share buttons:
Thank You With Social

Password set page with share buttons:
Password Set With Social

Facebook share window:
sharefb

Twitter share window:
sharetwitter

Linkedin share window
sharelinkedin

email share window (w default email set to gmail):
shareemail

### DIFF
--- a/support-frontend/assets/components/button/button.scss
+++ b/support-frontend/assets/components/button/button.scss
@@ -17,7 +17,6 @@
   transition: $gu-transition;
   justify-content: space-between;
   position: relative;
-  width: fit-content;
 
   &:hover, &:focus {
     outline: 0;
--- a/support-frontend/assets/components/button/button.scss
+++ b/support-frontend/assets/components/button/button.scss
@@ -17,6 +17,8 @@
   transition: $gu-transition;
   justify-content: space-between;
   position: relative;
+  width: fit-content;
+
   &:hover, &:focus {
     outline: 0;
   }
--- a/support-frontend/assets/components/socialShare/socialShare.jsx
+++ b/support-frontend/assets/components/socialShare/socialShare.jsx
@@ -14,42 +14,47 @@ import { trackComponentClick } from 'helpers/tracking/ophanComponentEventTrackin
 
 // ---- Types ----- //
 
-type Platform = 'facebook' | 'twitter' | 'linkedin' | 'email';
+type SharePlatform = 'facebook' | 'twitter' | 'linkedin' | 'email';
 
-type PropTypes = {| name: Platform |};
+type PropTypes = {| name: SharePlatform |};
 
 type SocialMedia = {
   link: string,
   svg: Node,
   a11yHint: string,
+  windowFeatures: string
 };
 
-type WindowFeatures = '' | 'menubar=no, toolbar=no, resizable=yes, scrollbars=yes, width=500, height=400';
+const SocialWindowFeatures = 'menubar=no, toolbar=no, resizable=yes, scrollbars=yes, width=500, height=400';
 
 // ----- Setup ----- //
 
 const socialMedia: {
-  [Platform]: SocialMedia,
+  [SharePlatform]: SocialMedia,
 } = {
   facebook: {
-    link: 'https://www.facebook.com/sharer/sharer.php?u=https%3A%2F%2Fsupport.theguardian.com%2Fuk%2Fcontribute',
+    link: 'https://www.facebook.com/sharer/sharer.php?u=https%3A%2F%2Fsupport.theguardian.com%2Fcontribute',
     svg: <SvgFacebook />,
     a11yHint: 'Share on facebook',
+    windowFeatures: SocialWindowFeatures,
   },
   twitter: {
-    link: 'https://twitter.com/intent/tweet?url=https%3A%2F%2Fsupport.theguardian.com%2Fuk%2Fcontribute&text=Join%20me%20and%20over%20one%20million%20others%20in%20supporting%20a%20different%20model%20for%20open%2C%20independent%20journalism.%20Together%20we%20can%20help%20safeguard%20The%20Guardian%E2%80%99s%20future%20%E2%80%93%20so%20more%20people%2C%20across%20the%20world%2C%20can%20keep%20accessing%20factual%20information%20for%20free',
+    link: 'https://twitter.com/intent/tweet?url=https%3A%2F%2Fsupport.theguardian.com%2Fcontribute&text=Join%20me%20and%20over%20one%20million%20others%20in%20supporting%20a%20different%20model%20for%20open%2C%20independent%20journalism.%20Together%20we%20can%20help%20safeguard%20The%20Guardian%E2%80%99s%20future%20%E2%80%93%20so%20more%20people%2C%20across%20the%20world%2C%20can%20keep%20accessing%20factual%20information%20for%20free',
     svg: <SvgTwitter />,
     a11yHint: 'Share on twitter',
+    windowFeatures: SocialWindowFeatures,
   },
   linkedin: {
-    link: 'http://www.linkedin.com/shareArticle?mini=true&url=https%3A%2F%2Fsupport.theguardian.com%2Fuk%2Fcontribute&title=Join%20me%20and%20over%20one%20million%20others%20in%20supporting%20a%20different%20model%20for%20open%2C%20independent%20journalism.%20Together%20we%20can%20help%20safeguard%20The%20Guardian%E2%80%99s%20future%20%E2%80%93%20so%20more%20people%2C%20across%20the%20world%2C%20can%20keep%20accessing%20factual%20information%20for%20free',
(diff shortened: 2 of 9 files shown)


support-frontend/assets/components/socialShare/socialShare.jsx
```diff
@@ -24,29 +24,30 @@ type SocialMedia = {
   a11yHint: string,
 };
 
+type WindowFeatures = '' | 'menubar=no, toolbar=no, resizable=yes, scrollbars=yes, width=500, height=400';
```
tomrf1: rather than writing out this string twice, we could make this a const and add a `windowFeatures: string` field to the `SocialMedia` type?

support-frontend/assets/components/button/button.scss
```diff
@@ -17,6 +17,8 @@
   transition: $gu-transition;
   justify-content: space-between;
   position: relative;
+  width: fit-content;
```
joelochlann: worth checking there's a graceful fallback in unsupported browser since it doesn't look like support is that great: https://caniuse.com/#search=fit-content

jlieb10: Thanks @joelochlann. From what I found, unless an absolute `width` or `max-width` is specified, it seems a _button used inside a form_ goes full width. As this is a reused component, I don't want to pass in any absolute widths but let's loop in @walaura to see if she has more expertise

support-frontend/assets/components/socialShare/socialShare.jsx
joelochlann: then you could just pass `socialMedia[props.name]` to `onShare` and get rid of the `if` condition

support-frontend/assets/components/socialShare/socialShare.jsx
```diff
@@ -24,29 +24,30 @@ type SocialMedia = {
...
 
 // ----- Setup ----- //
 
 const socialMedia: {
   [Platform]: SocialMedia,
```
joelochlann: I feel that `SharePlatform` could be a better name since it includes email?

support-frontend/assets/components/socialShare/socialShare.jsx
```diff
@@ -24,29 +24,30 @@ type SocialMedia = {
...
 } = {
   facebook: {
-    link: 'https://www.facebook.com/sharer/sharer.php?u=https%3A%2F%2Fsupport.theguardian.com%2F%3FINTCMP%3Dsocial&t=',
+    link: 'https://www.facebook.com/sharer/sharer.php?u=https%3A%2F%2Fsupport.theguardian.com%2Fuk%2Fcontribute',
```
joelochlann: these links are all hardcoded to the `uk` region. you should link to the geo-redirect URL https://support.theguardian.com/contribute

support-frontend/assets/pages/new-contributions-landing/components/SetPasswordForm.jsx
```diff
@@ -131,26 +131,24 @@ function SetPasswordForm(props: PropTypes) {
           errorMessage="Please enter a password between 6 and 20 characters long"
           required
         />
-        <ButtonWithRightArrow
-          componentClassName={classNameWithModifiers('form__submit', ['create-account'])}
-          buttonClassName={classNameWithModifiers('form__submit-button', ['create-account'])}
-          accessibilityHintId="accessibility-hint-create-account"
+        <Button
```
joelochlann: 👍 for switching to storybook components!

support-frontend/assets/components/button/button.scss
jlieb10: or maybe @SiAdcock because I think Laura is on hols

support-frontend/assets/components/socialShare/socialShare.jsx
```diff
@@ -59,17 +60,35 @@ export default function SocialShare(props: PropTypes) {
...
+        '',
+        features,
+      );
+    };
+  }
+
+
   return (
-    <a
+    <button
```
joelochlann: I just noticed the cursor doesn't change when hovering over these buttons, which might discourage people from clicking. You'll either need to set that hover state in your CSS - or would it be possible to use storybook buttons for these too?

support-frontend/assets/components/button/button.scss
joelochlann: Ah of course, it's a storybook component. Well, I can see that the reason it's going full width is because the [form is a flex container](https://github.com/guardian/support-frontend/blob/master/support-frontend/assets/pages/new-contributions-landing/contributionsLanding.scss#L452-L456). The default value of `align-items` is `stretch`, so this is why the children are full-width (try adding `align-items: flex-start` to the form style and see the difference!).

That `form--contribution` style is from our page-specific CSS though, so we definitely shouldn't change the generic storybook styling to affect it. If we really want these buttons to be left-aligned, we could put another class on those buttons and make them `align-self: flex-start`. **But** I'm not actually convinced we want them left-aligned & content-width.

Check with @ionamckendrick but I think this:

Screenshot 2019-04-26 at 10 58 28

is preferable to this:

Screenshot 2019-04-26 at 10 59 41

so no extra CSS needed to make the buttons content-width.

support-frontend/assets/components/socialShare/socialShare.jsx
jlieb10: I think it would be better to add social buttons to storybook instead of using the default storybook buttons for this, as we want the social icons to be bigger than the icons used next to the text

support-frontend/assets/components/button/button.scss
jlieb10: I think having a fitted width is part of using the storybook components- we should try to eliminate the stylistic variance of having full-width some places and auto-width others in my understanding

ionamckendrick: Really good one to discuss. I visually prefer full width when buttons are stacked together (the flex width looks weird) BUT I appreciate that for single CTAs the flex width is the pattern we're generally using. I'd vote for fixed width in this instance, but we should definitely pick up with Laura and a group of us to decide for the future.

support-frontend/assets/components/socialShare/socialShare.jsx
jlieb10: Fixed

support-frontend/assets/components/socialShare/socialShare.jsx
jlieb10: Fixed

joelochlann: actually or `PlatformParams` to correspond to the `Platform` key

support-frontend/assets/components/button/button.scss
joelochlann: (@ionamckendrick careful with the term flex width since flex is a very loaded technical term in CSS-land! so it might not mean what you think. content-width vs parent-width is probably the least ambiguous for this discussion)

jlieb10: @ionamckendrick thanks for the input and resolving this for now - will just leave them full-width for the time being

support-frontend/assets/components/socialShare/socialShare.jsx
```diff
@@ -10,45 +10,51 @@ import SvgFacebook from 'components/svgs/facebook';
...
 
 // ----- Setup ----- //
 
 const socialMedia: {
-  [Platform]: SocialMedia,
+  [SharePlatform]: SocialMedia,
 } = {
   facebook: {
-    link: 'https://www.facebook.com/sharer/sharer.php?u=https%3A%2F%2Fsupport.theguardian.com%2F%3FINTCMP%3Dsocial&t=',
+    link: 'https://www.facebook.com/sharer/sharer.php?u=https%3A%2F%2Fsupport.theguardian.com%2Fcontribute',
```
joelochlann: oh, one more thing on the tracking for these links. I would use this to generate them: http://reader-revenue-lynx.s3-website-eu-west-1.amazonaws.com/

in the second dropdown (source) you can choose EMAIL or SOCIAL and put `contributions-share-facebook`, `contributions-share-twitter`, etc in the Component ID field

jlieb10: it seems that facebook does support the urls that this tool generates however twitter strips the entire link out of the tweet when it has this parameter, linkedin strips out the value of the query parameter, and gmail unencodes the json (which probably isn't a technical problem but doesn't look very trustworthy to click)

joelochlann: Yep you're right! Those URL-encoded JSON blobs are not good for this use case. Let's leave out the tracking for the purposes of this PR and address in a future PR.

What we can do is create some friendly-URL redirects for these, e.g. https://support.theguardian.com/contribute/twitter-share, which redirect to a page with all the tracking info. We did something similar for the Australian campaign: https://support.theguardian.com/au/contribute/thefrontline

Unfortunately it's not the easiest thing to do right now - for Australia we did it in Fastly which was a bit of a pain (as @tomrf1 will tell you). We should probably make a little tool on the admin console that can set up redirects.

Of course we could always use https://bitly.com/ or similar to shorten the links but people might be less inclined to click on links which don't have the guardian domain.

joelochlann: and as per our chat conversation - INTCMP is fine for now

PR #1772, opened 2019-04-30
existing-payment-options: implement Direct Debit cloning

twrichards:
## Why are you doing this?
In the initial release of 'existing payment options' (re-use of existing payment details), `support-workers` was only able to handle cards (since Stripe tokens can be re-used on multiple billing accounts in Zuora). Direct Debit is a bit more complicated, and this PR makes it work. 

We need to fetch the existing Mandate ID from Zuora, then use it to first lookup the 'customer bank account' ID in GoCardless before then creating a **NEW mandate** in GoCardless under that same 'customer bank account', before then including the new mandateID and reference in the `PaymentMethod` passed to `CreateZuoraSubscription` (like any other Payment Method).

This approach means that customers will have multiple entries in 'manage direct debits' section of their online banking etc, one for each product, rather than having a single entry for all products, which would cause the frequency and amount to fluctuate in confusing way, and make it hard to cancel.

This should hopefully increase the overall quantity of acquisitions via existing payment options. 

## Changes

- There was already a `GoCardlessService` in `support-frontend` used for validation of bank details client-side, so this was migrated to `support-services` (and config to `support-config`)
- With this `GoCardlessService` migrated to a shared location, two additional functions were added
  - one to fetch the 'customer bank account' ID based on the mandateID of the existing one
  - then the second to created a new mandate under that 'customer bank account' and return an ADT containing the new mandate ID and the 'reference'
- These two new functions are now used in a `for` comprehension in the `GetPaymentMethod` lambda, for the case when the requested payment method type for re-use is Direct Debit.

**NOTE** that this function required new GoCardless API tokens for `support-workers` with read & write ability - `support-frontend` just needs read-only


support-services/build.sbt
```diff
@@ -13,5 +13,6 @@ libraryDependencies ++= Seq(
   "io.circe" %% "circe-generic" % circeVersion,
   "io.circe" %% "circe-generic-extras" % circeVersion,
   "io.circe" %% "circe-parser" % circeVersion,
-  "com.squareup.okhttp3" % "okhttp" % okhttpVersion
+  "com.squareup.okhttp3" % "okhttp" % okhttpVersion,
+  "com.gocardless" % "gocardless-pro" % "2.8.0"
```
johnduffell: support-services sounds a bit like a "common" type module to me.  Every dependency that is added to this, will be on the classpath of everything that depends on this, meaning more risk of snyk issues, binary incompatibilities, increase in artifact size and difficult to reuse. The gravity of a "common" appears to be so strong, that over time it ends up as a monolithic everything library. (I learned all this from my membership-common experience)
I am sure that opinion varies on this, but I think having a "common" or "services" is an anti pattern for the above reason.
In terms of here, I would recommend creating a new gocardless sbt module, and put only the requrements/dependencies for that there.  Then it would be self contained and reusable independently.
@rupertbates probably has a useful viewpoint on this matter!

twrichards: fair points (including those mentioned in person) - since the new functions added to `GoCardlessService` aren't used in support-frontend and the old function isn't used by `support-workers` I'm going to create a sealed trait (or something) for `GoCardlessService` which resides in `support-services` but have the actual implementation**s** in the respective projects where the functions are used and so the GoCardless library dependency will be on only the necessary projects.

twrichards: done!

rupertbates: I don't really agree with your concerns in this case John - there are 2 client apps for support-services and they both need the GoCardless library. I can't see any benefit from splitting that functionality between two separate projects, it's just duplication and runs the risk of things like the lib version getting updated in one project but not the other.

support-workers/src/main/scala/com/gu/emailservices/ContributionEmailFields.scala
```diff
@@ -27,6 +27,13 @@ case class ContributionEmailFields(
       "first payment date" -> formatDate(created.plusDays(10).toLocalDate),
       "payment method" -> "Direct Debit"
     )
+    case dd: ClonedDirectDebitPaymentMethod => List(
+      "account name" -> dd.bankTransferAccountName,
+      "sort code" -> hyphenate(dd.bankCode),
+      "Mandate ID" -> directDebitMandateId.getOrElse(""),
+      "first payment date" -> formatDate(created.plusDays(10).toLocalDate),
```
rupertbates: Maybe extract this 10 to a named val? Or create a function which takes `created` and returns the first payment date

support-services/build.sbt
twrichards: sigh - i have now refactored - shall i refactor it back to shared, or shall i make a `val` for the version in `LibraryVersions.scala` (like `jacksonVersion` for example)

support-workers/src/main/scala/com/gu/support/workers/lambdas/GetPaymentMethod.scala
```diff
@@ -65,10 +69,29 @@ class GetPaymentMethod(servicesProvider: ServiceProvider = ServiceProvider)
...
 
+  def cloneMandate(
+    existingDirectDebit: GetPaymentMethodDirectDebitResponse,
+    goCardlessService: GoCardlessWorkersService
+  ) : Future[ClonedDirectDebitPaymentMethod] = for {
+    customerBankAccountId <- goCardlessService.getCustomerAccountIdFromMandateId(existingDirectDebit.tokenId)
+    clonedMandateRefs <- goCardlessService.createNewMandateOnExistingCustomerAccount(customerBankAccountId)
+  } yield ClonedDirectDebitPaymentMethod(
+    tokenId = clonedMandateRefs.mandateId, // yes Zuora put the mandateId into tokenId
+    mandateId = clonedMandateRefs.reference, // and yes Zuora like to use the 'reference' in the 'mandateId' field, sigh
```
rupertbates: 😃

support-services/build.sbt
rupertbates: Personally I'd go back to shared but it's up to you if you've got the patience!

johnduffell: since there are only two clients it won't make any difference in practice at this stage, agreed.

Regarding the version going out of step, they are both in the same repo so we would have a common dependency version in Dependencies.scala.

As long as there will only ever be two clients, or if there were it would be obvious and easy to split the non-needed stuff out in some way, then that would be fine.  Perhaps in the case of membership-common there were other considerations (e.g. 2 complete changes of team members) that exacerbated it.  But the same happened on dotcom with the "common" module and I've seen it in many other projects.  I suppose it's a bit like microservices - split it down just enough, but not too much!  It's good that we're having the debate but maybe the specific line doesn't matter as much.

PR #1774, opened 2019-04-30
Bump webpack-dev-server to 3.1.11

jorgeazevedo:
## Why are you doing this?

According to Github, version 3.1.5 of webpack-dev-server is affected by CVE-2018-14732. And that warning at the top of the repo is freaking me out.

As far as I can tell this is only used for local development so I think we're good?


jlieb10:
https://github.com/guardian/support-frontend/pull/1795

PR #1777, opened 2019-05-03
Send tracking events when a user selects a payment method

rupertbates:
## Why are you doing this?
We want to track what payment methods users are selecting on the Digital Pack checkout page so that we can analyse drop off rates for each method.

### DIFF
--- a/support-frontend/assets/pages/digital-subscription-checkout/digitalSubscriptionCheckoutActions.js
+++ b/support-frontend/assets/pages/digital-subscription-checkout/digitalSubscriptionCheckoutActions.js
@@ -15,6 +15,8 @@ import type { ErrorReason } from '../../helpers/errorReasons';
 import type { PaymentMethod } from 'helpers/paymentMethods';
 import { PayPal } from 'helpers/paymentMethods';
 import type { Action as AddressAction } from 'components/subscriptionCheckouts/address/addressFieldsStore';
+import * as storage from 'helpers/storage';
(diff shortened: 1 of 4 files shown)


support-frontend/assets/pages/digital-subscription-checkout/digitalSubscriptionCheckoutActions.js
```diff
@@ -45,6 +47,8 @@ const formActionCreators = {
   setBillingPeriod: (billingPeriod: DigitalBillingPeriod): Action => ({ type: 'SET_BILLING_PERIOD', billingPeriod }),
   setPaymentMethod: (paymentMethod: PaymentMethod) => (dispatch: Dispatch<Action>, getState: () => State) => {
     const state = getState();
+    storage.setSession('selectedPaymentMethod', paymentMethod);
```
johnduffell: why do we have to put in storage? does all that get transferred into GA when the purchase happens?

rupertbates: I've noticed that we send the payment method as a custom dimension on any events we send and that it comes from the session. We may actually not need to as it is a session level dimension in GA but contributions set it and it doesn't hurt so I think it's safer to.

johnduffell: I thought that redux was the best place for state, is this needed in storage because it needs persist over page loads?

PR #1779, opened 2019-05-03
SaleHeader updates

jesseyuen:
## Why are you doing this?

Each product page has two banner options Default and Campaign. The campaign banner was confusingingly called SaleHeader even though it was not related to a _sales_ as such. 

The Campaign Banner also had not been updated to include flash sale text.

## Changes

* Rename SaleHeader to CampaignHeader because it makes more sense
* Introduce flash sale text to CampaignHeader

### DIFF
--- a/support-frontend/assets/pages/digital-subscription-landing/components/digitalSubscriptionLandingHeader.jsx
+++ b/support-frontend/assets/pages/digital-subscription-landing/components/digitalSubscriptionLandingHeader.jsx
@@ -171,7 +171,7 @@ function getCopy(product: SubscriptionProduct, country: CountryGroupId) {
     };
   }
   return {
-    heading: 'Digital Pack',
+    heading: 'Digital Pack subscriptions',
     subHeading: 'The premium Guardian experience, ad-free on all your devices',
   };
 }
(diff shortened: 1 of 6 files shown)


support-frontend/assets/pages/digital-subscription-landing/components/digitalSubscriptionLandingHeader.jsx
```diff
@@ -176,13 +176,14 @@ function getCopy(product: SubscriptionProduct, country: CountryGroupId) {
...
 }
 
-function SaleHeader(props: PropTypes) {
+function CampaignHeader(props: PropTypes) {
   const product: SubscriptionProduct = 'DigitalPack';
+  const copy = getCopy(product, props.countryGroupId);
   return (
     <ProductPagehero
       appearance="campaign"
-      overheading="Digital Pack subscriptions"
```
jfsoul: Looks like this might now change to just `Digital Pack`.  Is that intentional?

jesseyuen: Good pick up Jon! Cheers.

PR #1782, opened 2019-05-07
removed survey links, added documentation for implementing in the future

jlieb10:
## Why are you doing this?
The current survey has finished running, so I'm taking those links out and adding in documentation to help re-implement a configurable survey button again in the future

[**Trello Card**](https://trello.com/c/pB4N050G/1053-remove-user-survey-post-contribution)

## Changes

* Remove buttons from 'thank you' and 'thank you password set' pages
* Added documentation for adding back configurable survey links in the future, particularly on the 'thank you password set' page, which isn't connected to Redux by default

### DIFF
--- a/support-frontend/assets/pages/new-contributions-landing/components/ContributionThankYou/ContributionThankYou.jsx
+++ b/support-frontend/assets/pages/new-contributions-landing/components/ContributionThankYou/ContributionThankYou.jsx
@@ -14,7 +14,6 @@ import AnchorButton from 'components/button/anchorButton';
 import SvgArrowLeft from 'components/svgs/arrowLeftStraight';
 import { DirectDebit } from 'helpers/paymentMethods';
 import SpreadTheWord from 'components/spreadTheWord/spreadTheWord';
-import ContributionSurvey from '../ContributionSurvey/ContributionsSurvey';
 
 // ----- Types ----- //
 
@@ -64,7 +63,6 @@ function ContributionThankYou(props: PropTypes) {
           </section>
         ) : null}
         <MarketingConsent />
-        <ContributionSurvey contributionType={props.contributionType} />
         <SpreadTheWord />
         <div className="gu-content__return-link">
           <AnchorButton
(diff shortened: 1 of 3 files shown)


support-frontend/assets/pages/new-contributions-landing/components/ContributionThankYou/ContributionThankYou.jsx
```diff
@@ -64,7 +63,6 @@ function ContributionThankYou(props: PropTypes) {
           </section>
         ) : null}
         <MarketingConsent />
-        <ContributionSurvey contributionType={props.contributionType} />
```
tomrf1: A thought: rather than removing any code in this PR, could we just allow `ContributionSurvey` to decide whether or not to render?
E.g. in its `render` function: `if (surveyLink) return ( <div>...</div>)`
This seems to be something we keep switching on/off

jlieb10: @tomrf1 I started to implement this approach, but I think it's a little bit confusing to show the `<ContributionSurvey>` component on the thank you pages with the expectation that it won't show up. I'd rather take it off the pages and reintroduce it when we expect it to show up (and have the process of reintroducing it well documented). But I do take your point and I'm not sure how often this happens, so maybe it would be better to leave in place if we are doing a survey more than a couple times a year.

tomrf1: I'd personally prefer configurable code to commented out code - I'd be good to get thoughts of @jranks123 and @joelochlann here

jranks123: I'm leaning towards keeping it in and having it optionally on. It's something that we are always taking out and putting in and I think it's happened enough now to warrant a place in the code.

jlieb10: 👍

jlieb10: @tomrf1 @jranks123 thanks for calling it. I created a different branch and PR with the implementation here: https://github.com/guardian/support-frontend/pull/1785

PR #1783, opened 2019-05-07
Adds a version of navbar for checkout pages

lucymonie:
## Why are you doing this?
The aim here is to add the word 'Checkout' to the navbar for digital and paper checkouts, in a range of different formats for different screens.

The work is details in this [**Trello Card**](https://trello.com/c/rkxkI6rA/2341-checkout-header)

## Changes
* Mobile version of navbar in checkout
* Tablet version of navbar in checkout
* Desktop version of navbar in checkout

### Mobile
Screen Shot 2019-05-08 at 12 55 45

### Tablet
Screen Shot 2019-05-08 at 12 57 36

### Desktop
Screen Shot 2019-05-08 at 12 56 07

### DIFF
--- a/support-frontend/assets/components/headers/header/header.jsx
+++ b/support-frontend/assets/components/headers/header/header.jsx
@@ -20,8 +20,7 @@ import './header.scss';
 export type PropTypes = {|
   utility: Option<Node>,
   countryGroupId: ?CountryGroupId,
-  displayNavigation: boolean,
-  displayCheckout?: boolean | void,
+  display?: 'navigation' | 'checkout' | void,
 |};
 export type State = {|
   fitsLinksInOneRow: boolean,
@@ -52,13 +51,13 @@ const getMenuStateMetrics = ({ menuRef, logoRef, containerRef }): State => {
 type TopNavPropTypes = {|
   utility: Node,
   getLogoRef: (?Element) => void,
-  displayCheckout: boolean | void,
+  display: 'navigation' | 'checkout' | void,
 |};
 
-const TopNav = ({ displayCheckout, getLogoRef, utility }: TopNavPropTypes) => (
+const TopNav = ({ display, getLogoRef, utility }: TopNavPropTypes) => (
   <div className="component-header-topnav">
     <div className="component-header-topnav__utility">{utility}</div>
-    {displayCheckout && (
+    {display === 'checkout' && (
       <div className="component-header-topnav__checkout">
         <div />
         <div className="component-header-topnav--checkout-text">
@@ -80,8 +79,7 @@ export default class Header extends Component<PropTypes, State> {
   static defaultProps = {
     utility: null,
     countryGroupId: null,
-    displayNavigation: true,
-    displayCheckout: false,
+    display: 'navigation',
   };
 
   state = {
@@ -90,7 +88,7 @@ export default class Header extends Component<PropTypes, State> {
   };
 
   componentDidMount() {
-    if (this.props.displayNavigation && this.menuRef && this.logoRef && this.containerRef) {
+    if (this.props.display === 'navigation' && this.menuRef && this.logoRef && this.containerRef) {
       this.observer = onElementResize(
         [this.logoRef, this.menuRef, this.containerRef],
         ([logoRef, menuRef, containerRef]) => {
@@ -112,9 +110,7 @@ export default class Header extends Component<PropTypes, State> {
   observer: ElementResizer;
 
   render() {
-    const {
-      utility, displayNavigation, countryGroupId, displayCheckout,
-    } = this.props;
+    const { utility, display, countryGroupId } = this.props;
     const { fitsLinksInOneRow, fitsLinksAtAll } = this.state;
 
     return (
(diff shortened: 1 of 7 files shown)


support-frontend/assets/components/headers/header/header.jsx
```diff
@@ -45,16 +46,23 @@ const getMenuStateMetrics = ({ menuRef, logoRef, containerRef }): State => {
...
+  displayCheckout: boolean | void,
 |};
 
-const TopNav = ({ getLogoRef, utility }: TopNavPropTypes) => (
+const TopNav = ({ displayCheckout, getLogoRef, utility }: TopNavPropTypes) => (
   <div className="component-header-topnav">
     <div className="component-header-topnav__utility">{utility}</div>
+    {displayCheckout && (
+      <div className="component-header-topnav__checkout">
+        <div className="component-header-topnav--empty" />
```
jesseyuen: You could delete --empty if you added this css:

```
.component-header-topnav__checkout {
    //the existing css
    display: flex;
}

.component-header-topnav--checkout-text {
    //the existing css
    width: 100%;
    align-self: flex-end;
}
```

support-frontend/assets/components/headers/header/header.scss
```diff
@@ -213,3 +212,100 @@ $header-links-min-width: gu-span(12);
     }
   }
 }
+
+.component-header--display-checkout {
+  .component-header__wrapper {
+    max-width: 50rem;
+
+    @include mq($from: wide) {
```
jesseyuen: I think this needs to be @include mq($from: leftCol) or what ever the variable name is for one down from wide

support-frontend/assets/components/headers/header/header.jsx
```diff
@@ -20,6 +21,7 @@ export type PropTypes = {|
   utility: Option<Node>,
   countryGroupId: ?CountryGroupId,
   displayNavigation: boolean,
+  displayCheckout?: boolean | void,
```
walaura:
```suggestion
  display: 'navigation' | 'checkout' | null
```

i'd do something like this instead of two boolean parameters, since they are mutually exclusive (you can't display navigation AND display checkout)

support-frontend/assets/components/headers/header/header.scss
```diff
@@ -213,3 +219,103 @@ $header-links-min-width: gu-span(12);
...
+        justify-content: flex-end;
+      }
+
+      .component-header-topnav__utility {
+        display: none;
+      }
+
+      .component-header-topnav__checkout {
+        display: flex;
+        flex: 0 0 490px;
```
walaura: try to use `gu-span(*)` here rather than a fixed pixel size

support-frontend/assets/components/headers/header/header.scss
```diff
@@ -213,3 +219,103 @@ $header-links-min-width: gu-span(12);
...
+    max-width: 50rem;
+
+    @include mq($from: tablet, $until: desktop) {
+      max-width: 100%;
+      padding: 0;
+      margin: 0;
+    }
+
+    @include mq($from: leftCol) {
+      max-width: 60rem;
```
walaura: try to use `gu-span(*)` here rather than a fixed pixel size

support-frontend/assets/components/headers/header/header.scss
```diff
@@ -213,3 +219,103 @@ $header-links-min-width: gu-span(12);
     }
   }
 }
+
+.component-header--display-checkout {
+  .component-header__wrapper {
+    max-width: 50rem;
```
walaura: try to use `gu-span(*)` here rather than a fixed pixel size

support-frontend/assets/components/headers/header/header.scss
```diff
@@ -213,3 +219,103 @@ $header-links-min-width: gu-span(12);
...
+            padding-left: 2px;
+          }
+        }
+      
+        @include mq($from: tablet, $until: desktop) {
+          display: none;
+        }
+  
+        @include mq($from: leftCol) {
+          flex: 0 0 650px;
```
walaura: try to use `gu-span(*)` here rather than a fixed pixel size

support-frontend/assets/components/headers/header/header.scss
```diff
@@ -213,3 +219,103 @@ $header-links-min-width: gu-span(12);
...
+          :nth-child(2) {
+            display: inline-block;
+            vertical-align: top;
+          }
+  
+          @include mq($until: tablet) {
+            border-left: none;
+            border-top: none;
+            margin-top: -16px;
+            padding-left: 2px;
```
walaura: these two numbers look a bit magic, what are they doing? can we achieve this positioning in a cleaner way?

support-frontend/assets/components/headers/header/header.jsx
lucymonie: Thanks @walaura - shall do

support-frontend/assets/components/headers/header/header.scss
lucymonie: The issue here is that the lines in the nav need to line up with the structure of the main body and I don't seem to be able to get it right with whole columns. Am I going about this the wrong way?

support-frontend/assets/components/headers/header/header.scss
lucymonie: Again here - I'm not able to get the lines in the right place with columns. Any suggestions as to how to make this work?

support-frontend/assets/components/headers/header/header.scss
lucymonie: Here as well.

support-frontend/assets/components/headers/header/header.scss
lucymonie: You're right! The margin-top wasn't doing anything and the other one I have adjusted to be gu

support-frontend/assets/components/headers/header/header.scss
lucymonie: Here as well...

PR #1786, opened 2019-05-08
Separate public/private subnets

joelochlann:
As part of reconfiguring the Membership VPC to have public & private subnets, I'm going to placing load balancers in the public subnet and instances in the private subnet, as per best security practices. (The original motivation for this was to give [lambdas access to the internet as well as entities within the VPC](https://aws.amazon.com/premiumsupport/knowledge-center/internet-access-lambda-function/), but having a subnet which is unreachable from the internet has other security benefits too, and is in fact how the Subscriptions VPC is set up.)

As a first step for this, I need to separate out these params in CloudFormation for all of our apps. I'll need to manually upload the new CloudFormation in the console because I need to supply values for these params. When I do this, I'll leave the instances in the current public subnets, but move the load balancer into the new public subnets I've created. Once I've done this for all our ASGs, the final step is to attach new routing tables to the current public subnets, making them private (i.e. the default route goes to a NAT gateway rather than an internet gateway).

- [x] Stack updated for CODE
- [x] Stack updated for PROD

### DIFF
--- a/support-frontend/cloud-formation/cfn.yaml
+++ b/support-frontend/cloud-formation/cfn.yaml
@@ -5,10 +5,12 @@ Parameters:
     Type: String
     Description: VpcId of your existing Virtual Private Cloud (VPC)
     Default: vpc-e6e00183
-  Subnets:
-    Type: CommaDelimitedList
-    Description: The list of SubnetIds in your Virtual Private Cloud (VPC)
-    Default: subnet-cb91ae8d, subnet-a7b74ac2, subnet-179e8063
+  PrivateVpcSubnets:
+    Description: Private subnets to use for EC2 instances
+    Type: List<AWS::EC2::Subnet::Id>
+  PublicVpcSubnets:
+    Description: Public subnets to use for the ELB
+    Type: List<AWS::EC2::Subnet::Id>
   Stack:
     Description: Applied directly as a tag
     Type: String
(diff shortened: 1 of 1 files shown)


support-frontend/cloud-formation/cfn.yaml
```diff
@@ -50,8 +52,7 @@ Resources:
   AutoScalingGroup:
     Type: AWS::AutoScaling::AutoScalingGroup
     Properties:
-      AvailabilityZones: !GetAZs
-      VPCZoneIdentifier: !Ref Subnets
+      VPCZoneIdentifier: !Ref PrivateVpcSubnets
```
rupertbates: Do we not need to set availability zones any more?

joelochlann: Yep good point. `AvailabilityZones` isn't needed because we supply `VPCZoneIdentifier` (https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-as-group.html#cfn-as-group-availabilityzones). So as long as we supply a subnet for each AZ, it has the same effect. At one point while experimenting with moving subnets around, having it there caused issues because it got out of sync with `VPCZoneIdentifier` (`!GetAZs` always returns all the AZs for the region). So I decided to remove it across the board to avoid making the migration process harder.

Maybe there's an argument that having it there enforces a subnet in each AZ. But I feel like since the subnet parameters will be the same across every single app the chance of error is quite low.

PR #1793, opened 2019-05-10
existing-payment-options: tweaks to Direct Debit cloning

twrichards:
There was a long delay before I merged https://github.com/guardian/support-frontend/pull/1772 and another PR snuck in with a pattern match on `PaymentMethod`, which then became non-exhaustive when I finally merged https://github.com/guardian/support-frontend/pull/1772.

_NOTE, this only affects re-use of Direct Debit payment methods for contributions (for signed in users with existing products) but this was **already disabled** because of the support switches. Once this PR is merged and tested, hopefully we can turn this on._

Unfortunately it's not possible to make SBT error on **only** non-exhaustive pattern matches (it's either make all warnings into errors or none - which is a real shame) - this would've caught this in the merge build (i.e. failed and therefore wouldn't be deployable)
https://stackoverflow.com/questions/28327383/make-compile-fail-on-non-exhaustive-match-in-sbt  
... so **instead** I've made `MatchError` map to `RetryNone` (to prevent something that will never succeed from being retried - especially important if that thing has side-effects, like the mutations in GoCardless added in https://github.com/guardian/support-frontend/pull/1772).

Also **renamed** `GetPaymentMethod` lambda to `PreparePaymentMethodForReuse` (now that it does some mutation in GoCardless the name was misleading).

Also corrected the fields passed to the email to ensure parity with a brand new direct debit.

### DIFF
--- a/support-workers/src/main/scala/com/gu/support/workers/lambdas/PaymentMethodExtensions.scala
+++ b/support-workers/src/main/scala/com/gu/support/workers/lambdas/PaymentMethodExtensions.scala
@@ -1,13 +1,19 @@
 package com.gu.support.workers.lambdas
 
-import com.gu.support.workers.{CreditCardReferenceTransaction, DirectDebitPaymentMethod, PayPalReferenceTransaction, PaymentMethod}
+import com.gu.support.workers.{
+  ClonedDirectDebitPaymentMethod,
+  CreditCardReferenceTransaction,
+  DirectDebitPaymentMethod,
+  PayPalReferenceTransaction,
+  PaymentMethod
+}
 
 object PaymentMethodExtensions {
   implicit class PaymentMethodExtension[T <: PaymentMethod](self: T) {
     def toFriendlyString: String = self match {
       case _: CreditCardReferenceTransaction => "Stripe"
       case _: PayPalReferenceTransaction => "PayPal"
-      case _: DirectDebitPaymentMethod => "DirectDebit"
+      case _: DirectDebitPaymentMethod | _: ClonedDirectDebitPaymentMethod => "DirectDebit"
     }
   }
 }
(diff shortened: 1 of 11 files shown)


twrichards:
Unfortunately it's not possible to make SBT error on **only** non-exhaustive pattern matches (it's either make all warnings into errors or none - which is a real shame) - this would've caught this in the merge build (i.e. failed and therefore wouldn't be deployable)
https://stackoverflow.com/questions/28327383/make-compile-fail-on-non-exhaustive-match-in-sbt

johnduffell:
yes the error on warning is really annoying, it basically works really well on a new project, but as soon as you need to update play or scala or any other major library, you have to turn it off until you can fix all the deprecation warnings.

support-workers/src/main/scala/com/gu/support/workers/lambdas/PaymentMethodExtensions.scala
```diff
@@ -1,13 +1,19 @@
 package com.gu.support.workers.lambdas
 
-import com.gu.support.workers.{CreditCardReferenceTransaction, DirectDebitPaymentMethod, PayPalReferenceTransaction, PaymentMethod}
+import com.gu.support.workers.{
```
johnduffell: interesting way to write imports, is that normal?

twrichards: just it was complaining about the line length, but wasn't sure whether to a an `_` import. Thoughts?

twrichards:
TESTED in CODE ✅

support-workers/src/main/scala/com/gu/support/workers/lambdas/PaymentMethodExtensions.scala
johnduffell: fine by me! just looked funny.  I'm not a big fan of _ imports anyway but intellij used to revert to them pretty quickly

PR #1794, opened 2019-05-13
Default payment methods

rupertbates:
## Why are you doing this?
This PR fixes a number of issues with the checkouts:
* When we stopped preselecting a payment method on the checkout forms we made it impossible to checkout if there was only one payment method available because there was no way in the form to set it. 
* We need to show a validation error if the user tries to submit the form without selecting a payment method.
* When a user corrects a form error by filling in a mandatory field, we should remove the error message for that field (this was being done by the address component but not by the rest of the forms)

[**Trello Card**](https://trello.com/c/mrRfSU1v/2388-digipack-checkout-doesnt-select-the-payment-method-when-theres-only-1-option-eg-from-netherlands)


support-frontend/assets/helpers/subscriptionsForms/validation.js
```diff
@@ -33,6 +33,10 @@ function firstError<FormField>(field: FormField, errors: FormError<FormField>[])
 
 }
 
+function removeError<FormField>(field: FormField, formErrors: FormError<FormField>[]): FormError<FormField>[] {
```
lucymonie: I love this - I much prefer forms that adapt to inputs post error (which is what I **believe** this is doing)

lucymonie: And then I realised it was already in there - just somewhere else...!

rupertbates: It is 😃

johnduffell:
just linking back to one of the original PRs as I was trying to gather together all the info in one place: https://github.com/guardian/support-frontend/pull/1765

PR #1795, opened 2019-05-13
Bugfix for webpack-dev-server 3.1.11

jlieb10:
Upgrading webpack-dev-server to 3.1.11 causes constant console logging due to new security features. The current best supported work around is to add the `--disable-host-check` flag to the 
script in `package.json`

Please see webpack-dev-server's documentation of this issue [**here**](https://github.com/webpack/webpack-dev-server/issues/1604)

**TBD**: check to see if this flag is still necessary next time we upgrade webpack-dev-server

### DIFF
--- a/support-frontend/package.json
+++ b/support-frontend/package.json
@@ -11,7 +11,7 @@
     "build-ssr": "webpack --config webpack.ssr.js && node scripts/ssrScript.js",
     "build-prod": "npm-run-all clean validate test storybook-static 'webpack --config webpack.prod.js --env.prod'",
     "webpack": "webpack",
-    "webpack-dev-server": "webpack-dev-server --config webpack.dev.js --port 9211 --public support.thegulocal.com",
+    "webpack-dev-server": "webpack-dev-server --config webpack.dev.js --port 9211 --public support.thegulocal.com --disable-host-check",
     "devrun": "npm-run-all clean webpack-dev-server",
     "test": "echo 'Running JS tests' && jest",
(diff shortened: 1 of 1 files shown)


support-frontend/package.json
```diff
@@ -11,7 +11,7 @@
     "build-ssr": "webpack --config webpack.ssr.js && node scripts/ssrScript.js",
     "build-prod": "npm-run-all clean validate test storybook-static 'webpack --config webpack.prod.js --env.prod'",
     "webpack": "webpack",
-    "webpack-dev-server": "webpack-dev-server --config webpack.dev.js --port 9211 --public support.thegulocal.com",
+    "webpack-dev-server": "webpack-dev-server --config webpack.dev.js --port 9211 --public support.thegulocal.com --disable-host-check",
```
johnduffell: is it possible to add a comment with a link to the issue jonny was looking at?

johnduffell:
good, but see my comment. although it can be linked back to the PR, it might be nice to comment inline so it is more obvious.
Also, Rupert is looking at the master build failure.
👍 other than that thanks!

support-frontend/package.json
jlieb10: @johnduffell comments are not supported in json files so we settled for a PR with the documentation 😢

jlieb10:
@johnduffell your PR was definitely not the source of the issue - you just brought the `yarn.lock` file up to date with the packages that we are currently using, which should've happened when the last package was updated

Also, unfortunately inline comments are not supported in `json` files

jlieb10:
Related PR: https://github.com/guardian/support-frontend/pull/1774

PR #1788, opened 2019-05-09
new route for weekly checkout and split out bundle

johnduffell:
## Why are you doing this?

This creates the new route and skeleton bundle for the weekly checkout.
This was mostly mechanical however I couldn't bring myself to copy and paste all the bundle without moving the payment related stuff somewhere shared.  However I went on a learning journey on the way round so it prob needs a bit more cleanup after that.
It would be good to get a look to make sure I'm on the right track though!

 I didn't do a URL parameter in the end because getting it to work without logging in would be a bit tricky

[**Trello Card**](https://trello.com/c/DlpF9cS6/2332-guardian-weekly-checkout-url-and-page-skeleton)

## Changes

* add route
* add controller
* add page template
* copy bundle and make sure it only duplicates the non common stuff

## Screenshots

image

PR #1796, opened 2019-05-13
Fix Flow errors in master from failed merge

rupertbates:
## Why are you doing this?
#1788 caused some Flow errors in master when it was merged


johnduffell:
thanks for fixing! 👍

PR #1797, opened 2019-05-13
FIXED PreviewSubscribeRequest from creating duplicate customers & mandates in GoCardless

twrichards:
## Why are you doing this?
Due to what appears to be a really nasty Zuora bug, despite being in **preview mode** Zuora was creating a customer and mandate in GoCardless...
image

Fortunately not double payments, but likely [TBC] creating duplicate direct debit entries in customers online banking etc (one for the real subscription, and another which is 'never used').

## Changes

_(after lots of trial and error)_ removed `PaymentMethod` and forced `AutoPay` to `false` in the `PreviewSubscribeRequest` call

### DIFF
--- a/support-models/src/main/scala/com/gu/support/zuora/api/PreviewSubscribeRequest.scala
+++ b/support-models/src/main/scala/com/gu/support/zuora/api/PreviewSubscribeRequest.scala
@@ -2,7 +2,6 @@ package com.gu.support.zuora.api
 
 import com.gu.support.encoding.Codec
 import com.gu.support.encoding.Codec._
-import com.gu.support.workers.PaymentMethod
 
 object PreviewSubscribeItem {
   implicit val codec: Codec[PreviewSubscribeItem] = capitalizingCodec
@@ -11,7 +10,6 @@ object PreviewSubscribeItem {
 case class PreviewSubscribeItem(
   account: Account,
   billToContact: ContactDetails,
-  paymentMethod: PaymentMethod,
   subscriptionData: SubscriptionData,
   subscribeOptions: SubscribeOptions,
   previewOptions: PreviewOptions
@@ -25,9 +23,8 @@ object PreviewSubscribeRequest {
     PreviewSubscribeRequest(
       List(
         PreviewSubscribeItem(
-          subscribeItem.account,
+          subscribeItem.account.copy(autoPay = false), //this is to work-around bug in Zuora where duplicate mandate would be created in GoCardless
           subscribeItem.billToContact,
-          subscribeItem.paymentMethod,
           // This hack allows us to preview invoices further into the future (required to get a helpful payment schedule)
           subscribeItem.subscriptionData.copy(subscription = subscribeItem.subscriptionData.subscription.copy(initialTerm = 24)),
           subscribeItem.subscribeOptions,


twrichards:
The call in question is https://www.zuora.com/developer/api-reference/#operation/Action_POSTsubscribe (and to clarify this IMO is a bug in Zuora, not GoCardless... in preview mode Zuora should not be making any external calls to payment gateway)

jfsoul:
👏 nice work

twrichards:
TESTED in CODE ✅

mchv:
I am asking a lot, and assuming a lot of capabilities from our providers, but is there a way for us to create an automated test that would create a subscribe request in Zuora API Sandbox and check in go cardless number of customer and mandates?

johnduffell:
such a small change in the end! Nice work tracking that down in the zuora jungle 👍 

regarding automated test it's a good point and this would be possible in the integration tests. However we should also consider post deployment monitoring as an alternative to pre-deployment tests.

thinking about the cost and benefit, these IT tests are relatively expensive, and I think the chance of a regression are surely small, the consequences are relatively small to allow into production for a short time.  I think personally that we should save the IT tests for more serious such as actual overcharging, and ideally in such a way it would catch a broad range of issues rather than very specific issues.
Happy to join any discussion about moving things forward though.

twrichards:
The entries have now come through in my online banking, and I can confirm **as expected** that the illegitimate mandate created by the preview call has come through 😦 

@mchv that idea is worth some consideration, but I'm inclined to agree with @johnduffell that regression for this very specific issue is unlikely and we're best having a more general post deployment test(s) that follows a few acquisition flows and checks everything is all good (but such a test could included confirmation that only **one** customer, mandate and payment is made) - @jorgeazevedo is that something you might be able to help us set up?

jorgeazevedo:
yeah, I'm interested in this!

mario-galic:
Consider and alternative to functional test where we execute side-effectful assertion in a separate thread that simply logs an error on invalid mutation. For example,

```
mutate()
getNumberOfMandatesF map { num => assert(num == 0, "There should be a single DD mandate") }
  .andThen { case Failure(e) => logger.error("Duplicate DD mandates created! Fix ASAP!", e)  }  
}
```  


Sentry/Kibana should pick it up, and we can respond in matter of hours as opposed to months.

Note crashing a future like so has no impact on production as it happens off main thread.

johnduffell:
@twrichards I was being even more general than that, eg we could have a daily reconciliation that makes sure the number of new mandates and the number of new subscriptions with DD on are similar.
But anyway, I'd still prefer personally us to focus on things like money taken being correct before we introduce testing for additional mandates.  Things like mini payments and double payments sound more scary to me, and I don't know how quickly we would notice that. Admittedly our customers would notice quicker.  Just because we have had a problem before, doesn't mean it's the worst thing that might happen in future.
Contributions team has some checking of Data Lake Vs Stripe payments for the one off contributions, and that runs daily.
Having said that I think @mario-galic has a simple solution/easy enough to maintain that it might be worth doing anyway.

jfsoul:
I'm also more in favour of a scheduled reconciliation or sanity check type of solution in general.  If the existing GoCardless integration supports `getNumberOfMandatesF` pretty simply then it wouldn't hurt to check in another thread.  Having said that GoCardless API might only be eventually consistent after a mutation (a bit like AWS autoscaling) and that might be a bit fiddly and not worth the trouble.

mario-galic:
Say we use [retry](https://github.com/softwaremill/retry), then call site would be not much more than 

```
assertFireAndForget(
  getNumberOfMandatesF,
  (v: Int) => Try(assert(v == 1)),
  "Duplicate DD mandates created! Fix ASAP!"
)
```

where

```
  def assertFireAndForget[T](
      f: => Future[T],
      p: T => Try[/* Assertion */ Unit],
      msg: String,
      max: Int = 3,
      delay: FiniteDuration = 10.seconds
  )(implicit assertionEC: ExecutionContext): Future[Unit] = {

      retry.Pause(max, delay)(odelay.Timer.default) {
        f map p
      } map {
        case Failure(e) => println(msg, e)
        case Success(_) => println("success")
      }

  }
```

and a dedicated execution context

```
implicit val assertionExcecutionContext = 
  ExecutionContext.fromExecutor(Executors.newFixedThreadPool(2))
```

jacobwinch:
Instead of writing a separate test, or running a daily reconciliation job, we could also consider adding a new lambda function to run the type of check(s) that @mario-galic described. This would enable each Step Function execution to QA itself (low effort, low maintenance cost, high coverage), whilst ensuring that there is a clear separation of concerns. 

This lambda could run a number of checks, for example:
1. Did we charge the user the right amount of money?
2. Did we charge the user exactly once?
3. Did we end up creating exactly one GoCardless / Stripe / PayPal customer?
4. Did the user end up with the right product?

By throwing an exception if a failed check is encountered, the [existing alerting](https://github.com/guardian/support-frontend/blob/master/support-workers/cloud-formation/src/templates/cfn-template.yaml#L167-L185) already would pick up failures, notifying us almost immediately after a user has completed their journey.

Such a lambda could run in the parallel block (alongside ‘SendThankYouEmail’ and ‘SendAcqusitionEvent’) which would allow the [API](https://github.com/guardian/support-frontend/blob/master/support-frontend/app/services/stepfunctions/SupportWorkersClient.scala#L169-L174) which sits over this Step Function to return success to the client _before_ the sanity checks are complete. Including the sanity checks in the Step Function also allows us to deal with eventual consistency, by making use of Step Function retries.

image

mario-galic:
https://github.com/guardian/tip/pull/61 provides the following API:

```
chargeUser andThen { case _ =>
  TipAssert(
    getNumberOfCharges,
    (num: Int) => num == 1,
    "User should be charged only once. Fix ASAP!"
  )
}
```

PR #1799, opened 2019-05-14
Do not use test stripe token if logged in

tomrf1:
## Why are you doing this?
If you make a stripe payment while logged in but also have the `_test_username` cookie set then it causes an alarm. This is because the client uses the stripe test token, but the backend (support-workers) uses it in live mode if the identity cookie is present.

This change prevents the client from using the stripe test token if you're logged in.

[**Trello Card**](https://trello.com/c/n6sD0Xv6/1054-fix-intermittent-stripe-test-live-mode-crossover-errors-in-step-functions)

### DIFF
--- a/support-frontend/assets/helpers/user/user.js
+++ b/support-frontend/assets/helpers/user/user.js
@@ -82,7 +82,8 @@ const init = (dispatch: Function, actions: UserSetStateActions = defaultUserActi
 
   const emailFromBrowser = getEmailFromBrowser();
 
-  if (isTestUser()) {
+  // Prevent use of test stripe token if logged in, as support-workers will try it in live mode if the identity cookie present
+  if (isTestUser() && !userAppearsLoggedIn) {
     dispatch(setTestUser(true));
   }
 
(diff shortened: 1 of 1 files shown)


support-frontend/assets/helpers/user/user.js
```diff
@@ -82,7 +82,11 @@ const init = (dispatch: Function, actions: UserSetStateActions = defaultUserActi
 
   const emailFromBrowser = getEmailFromBrowser();
 
-  if (isTestUser()) {
+  /*
+   * Prevent use of the test stripe token if user us logged in, as support-workers will try it in live mode
+   * if the identity cookie is present
+   */
+  if (isTestUser() && !userAppearsLoggedIn) {
```
johnduffell: great, but is it easy to check if the user appears logged in as a test user? if so might be worth making that additional check to avoid the opposite problem.

tomrf1: It would be simple to also check if the user's email matches the test username, i.e.:
`window.guardian.email.startsWith(cookie.get('_test_username'))`
Would this help?

tomrf1: I've pushed a commit which I think solves this

johnduffell: nice work, that's a good idea. Now we would only see a problem if it was a valid test user but had expired (48H), or someone was using an incognito window to sign up as a test user, which all seems unlikely!

PR #1803, opened 2019-05-14
Configure campaigns to use URL params instead of query params

jlieb10:
## Why are you doing this?

So we can provide clean, easy to remember URLs for fundraising campaigns (without query parameters, etc)

Previously a campaign link would have looked like this: `https://support.thegulocal.com/us/contribute?campaign-name=toxicamerica` but now it will look like this: `https://support.thegulocal.com/us/contribute/toxicamerica`

[**Trello Card**](https://trello.com/c/pQuXtcoY/1071-configure-campaigns-to-use-url-params-instead-of-query-params)

## Changes

* Added a route to Play router for catching the campaign URL
* Updated the router controller to handle the new URLs, while still preserving functionality for all other types of contribute URLs
* Updated campaign recognition functionality in the frontend app, so it reads from the URL path instead of the query parameters


support-frontend/app/controllers/Application.scala
```diff
@@ -19,6 +20,7 @@ import services.{IdentityService, MembersDataService, PaymentAPIService}
 import utils.BrowserCheck
 import utils.RequestCountry._
 import views.{EmptyDiv, Preload}
+
```
jranks123: remove this line

support-frontend/app/controllers/Application.scala
```diff
@@ -102,20 +94,28 @@ class Application(
...
+                          ): Action[AnyContent] = maybeAuthenticatedAction().async { implicit request =>
+
     type Attempt[A] = EitherT[Future, String, A]
-    implicit val settings: AllSettings = settingsProvider.getAllSettings()
-    request.user.traverse[Attempt, IdUser](identityService.getUser(_)).fold(
-      _ => Ok(contributionsHtml(countryCode, None)),
-      user => Ok(contributionsHtml(countryCode, user))
-    ).map(_.withSettingsSurrogateKey)
+
+    val campaignCodeExistsButIsInvalid = campaignCode != "" && campaignCode != "toxicamerica"
```
tomrf1: what happens if in future we add a new campaign in the JS but forget to update this line?
Is it necessary to do this server-side check?

support-frontend/app/controllers/Application.scala
```diff
@@ -136,15 +136,15 @@ class Application(
       paymentApiStripeEndpoint = paymentAPIService.stripeExecutePaymentEndpoint,
       paymentApiPayPalEndpoint = paymentAPIService.payPalCreatePaymentEndpoint,
       existingPaymentOptionsEndpoint = membersDataService.existingPaymentOptionsEndpoint,
-      idUser = idUser
+      idUser = idUser,
```
jranks123: remove this comma

support-frontend/app/controllers/Application.scala
```diff
@@ -162,6 +162,20 @@ class Application(
     request =>
       Redirect("/" + path, request.queryString, MOVED_PERMANENTLY)
   }
+
+
+  private def getRedirectUrl(fastlyCountry: Option[CountryGroup]): String = {
+    fastlyCountry match {
```
jranks123: seeing as we're only calling this once we should probably put it back

support-frontend/app/views/newContributions.scala.html
```diff
@@ -180,6 +180,7 @@ <h1 class="gu-content__blurb-header">Help us deliver the independent journalism
         }
       };
     };
+
```
jranks123: remove this

support-frontend/assets/helpers/url.js
```diff
@@ -104,6 +106,10 @@ function isFrontlineCampaign(): boolean {
   return getQueryParameter('frontline-campaign') === 'true';
 }
 
+function isToxicAmericaCampaign(): boolean {
```
jranks123: maybe this belongs in `campaigns.jsx`

support-frontend/app/controllers/Application.scala
jlieb10: @tomrf1 we refactored to move the 'arbitrary' bit to the routes file, so it should be more obvious how it's working

support-frontend/assets/helpers/url.js
jlieb10: don't think we need it at all

support-frontend/conf/routes
```diff
@@ -45,23 +45,23 @@ GET /event/*eventId                                 controllers.Application.redi
...
 # ----- Contributions ----- #
 
-GET  /contribute                                        controllers.Application.contributeGeoRedirect()
-GET  /$country<(uk|us|au|eu|int|nz|ca)>/contribute      controllers.Application.contributionsLanding(country: String)
-GET  /$country<(uk|us|au|eu|int|nz|ca)>/thankyou        controllers.Application.contributionsLanding(country: String)
+GET  /contribute                                    controllers.Application.contributeGeoRedirect(campaignCode = "")
+GET  /$country<(uk|us|au|eu|int|nz|ca)>/contribute  controllers.Application.contributionsLanding(country: String)
+
+# ----- Toxic America Campaign ---- #
+GET  /contribute/toxicamerica                       controllers.Application.contributeGeoRedirect(campaignCode = "toxicamerica")
```
joelochlann: looking at this, I wonder if it would be simpler with just `GET    /contribute/:campaignCode` and a pattern match in the controller action, falling back to the georedirect. Ditto for the country-specific one. You can then get rid of the wildcard redirect routes, and it's clearer that those wildcard ones are really intended to cover deprecated campaigns

jlieb10: We had that setup before but the controller with the if statement got pretty messy because it needed to return a future (from what I remember). I think this approach is pretty declarative and lets us change out the endpoint when there's a new campaign in a more obvious place.

jranks123: @joelochlann We did have that at one point. It's basically a toss up between logic in the controller or logic in the routes file. I preferred this in the end but there's not much between them. I don't think either is necessarily simpler, I think both solutions basically implement the same logic, it's about which is clearer. 

`it's clearer that those wildcard ones are really intended to cover deprecated campaigns`

This would be an argument for what you are suggesting. The way we implemented it before was by pointing `GET /contribute/:campaignCode` to the contributions Landing controller, and made that take a campaign code parameter. This felt messy as there was suddenly quite  a lot of logic in that controller, as it had to check to see if there was a campaign code and if there was, whether it matched the whitelisted campaign code. But maybe adding a seperate controller for  `GET /contribute/:campaignCode` would solve this problem. When Joshua comes in I think we'll give `GET /contribute/:campaignCode` another whack

jranks123: Just saw that Joshua isn't going to be in today so I'll have a go myself

tomrf1: What's the advantage of having the server whitelist campaign codes?
If the path component after `/contribute/` is always used for campaign codes then the client can check it, and ignore it if it's not a genuine campaign. Or is that crazy?

jranks123: Ok @tomrf1 and @joelochlann I've simplified things

support-frontend/assets/pages/new-contributions-landing/components/ContributionFormContainer.jsx
```diff
@@ -172,8 +172,7 @@ function goalReachedTemplate() {
 }
 
 const campaignName = getCampaignName();
-const campaign = campaignName ? campaigns[campaignName] : null;
-
+const campaign = campaignName && campaigns[campaignName] ? campaigns[campaignName] : null;
```
tomrf1: not really important, but I think the `campaigns[campaignName]` check is unnecessary because of how `CampaignName` is defined:
`export type CampaignName = $Keys<typeof campaigns>`

support-frontend/conf/routes
tomrf1: nice!

support-frontend/assets/pages/new-contributions-landing/components/ContributionFormContainer.jsx
jranks123: Good point

PR #1805, opened 2019-05-16
Make contributionTypes configurable for campaigns

tomrf1:
## Why are you doing this?
In the past we used the `contributionTypes` query parameter to limit a campaign to one-offs only.
This change adds an optional `contributionTypes` field to the `Campaign` model, which overrides whatever is in the settings provided by the server.
It's still possible to do this from query parameters.

We already store the contributionTypes settings for all regions in `state.common.settings.contributionTypes`. This is initialised to the value from `window.guardian.settings`, but it is now overridden by `contributionsLandingInit.js` if there is a campaign.

[**Trello Card**](https://trello.com/c/SOyNM6vg/1074-make-contribution-types-configurable-from-campaign-settings)

## Screenshots
Screenshot 2019-05-16 at 13 41 31

### DIFF
--- a/support-frontend/assets/helpers/checkouts.js
+++ b/support-frontend/assets/helpers/checkouts.js
@@ -15,7 +15,6 @@ import type { IsoCountry } from 'helpers/internationalisation/country';
 import type { Currency, IsoCurrency, SpokenCurrency } from 'helpers/internationalisation/currency';
 import { currencies, spokenCurrencies } from 'helpers/internationalisation/currency';
 import type { Amount, SelectedAmounts } from 'helpers/contributions';
-import type { CountryGroupId } from 'helpers/internationalisation/countryGroup';
 import type { PaymentMethod } from 'helpers/paymentMethods';
 import { DirectDebit, PayPal, Stripe } from 'helpers/paymentMethods';
 import { ExistingCard, ExistingDirectDebit } from './paymentMethods';
@@ -47,13 +46,11 @@ function toPaymentMethodSwitchNaming(paymentMethod: PaymentMethod): PaymentMetho
 function getValidContributionTypesFromUrlOrElse(fallback: ContributionTypes): ContributionTypes {
   const contributionTypesFromUrl = getQueryParameter('contributionTypes');
   if (contributionTypesFromUrl) {
-    return generateContributionTypes(
-      contributionTypesFromUrl
-        .split(',')
-        .map(toContributionType)
-        .filter(Boolean)
-        .map(contributionType => { return {contributionType} })
-    );
+    return generateContributionTypes(contributionTypesFromUrl
+      .split(',')
+      .map(toContributionType)
+      .filter(Boolean)
+      .map(contributionType => ({ contributionType })));
   }
 
   return fallback;
--- a/support-frontend/assets/helpers/checkouts.js
+++ b/support-frontend/assets/helpers/checkouts.js
@@ -4,8 +4,10 @@
 
 import { getQueryParameter } from 'helpers/url';
 import {
-  type ContributionType, getFrequency,
+  type ContributionType, type ContributionTypes,
+  getFrequency,
   toContributionType,
+  generateContributionTypes,
 } from 'helpers/contributions';
 import * as storage from 'helpers/storage';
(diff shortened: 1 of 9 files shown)


support-frontend/assets/pages/new-contributions-landing/contributionsLandingInit.js
```diff
@@ -118,16 +119,15 @@ function initialisePaymentMethods(state: State, dispatch: Function) {
     dispatch(onThirdPartyPaymentAuthorised(paymentAuthorisation));
   };
 
-  const contributionTypes = getValidContributionTypes(countryGroupId);
```
tomrf1: this was incorrectly using a hardcoded set of contributionTypes before

support-frontend/assets/pages/new-contributions-landing/contributionsLandingInit.js
```diff
@@ -197,10 +212,13 @@ const init = (store: Store<State, Action, Function>) => {
 
   const state = store.getState();
 
-  initialisePaymentMethods(state, dispatch);
+  const contributionTypes = getContributionTypes(state);
+  dispatch(setContributionTypes(contributionTypes));
+
+  initialisePaymentMethods(state, dispatch, contributionTypes);
```
tomrf1: `state.common.settings.contributionTypes` will not yet have been updated in the redux store, so the local `contributionTypes` value needs to be passed into `initialisePaymentMethods` and `selectInitialContributionTypeAndPaymentMethod`
This feels like an anti-pattern...

support-frontend/assets/helpers/checkouts.js
```diff
@@ -42,34 +43,19 @@ function toPaymentMethodSwitchNaming(paymentMethod: PaymentMethod): PaymentMetho
...
       .map(toContributionType)
-      .filter(Boolean);
+      .filter(Boolean)
+      .map(contributionType => ({ contributionType })));
   }
 
   return fallback;
 }
 
-function getValidContributionTypes(countryGroupId: CountryGroupId): ContributionType[] {
```
tomrf1: this should have been removed when contributionTypes were made configurable

support-frontend/assets/pages/new-contributions-landing/contributionsLandingInit.js
joelochlann: haven't totally understood the issue yet - but wondering if keeping settings outside of the redux store would help? since they're immutable globals I've always thought maybe they shouldn't be in redux

tomrf1: I was wondering this also! Ideally they'd be determined during initialisation and then passed in as props. I'll see what I can do

support-frontend/assets/pages/new-contributions-landing/components/ContributionTypeTabs.jsx
```diff
@@ -56,7 +56,7 @@ const mapDispatchToProps = (dispatch: Function) => ({
 function ContributionTypeTabs(props: PropTypes) {
   const contributionTypes = props.contributionTypes[props.countryGroupId];
```
jlieb10: If we want to avoid making this part of Redux, we could check to see if the campaign contribution types exist here. This would be consistent with the other campaign configuration checks, which are all done inside the component using them. I don't know if that's the best idea, but it's working to keep the campaign configuration outside the state

tomrf1: the `contributionTypes` settings are also needed during initialisation in `contributionsLandingInit.js`, and it would be better to only calculate them once

jlieb10: ah right, i forgot they're needed earlier than at the point of rendering

support-frontend/assets/pages/new-contributions-landing/contributionsLandingInit.js
tomrf1: I'm going to slap a `TODO` in here and try this in another PR as it's quite a big change, and we need the functionality in this PR asap

PR #1807, opened 2019-05-17
Subs reducers refactor

rupertbates:
## Why are you doing this?

Up until now each of the subscriptions checkouts has its own Redux actions and reducer which have been copy and pasted from the previous checkout developed. 
This is a big problem for maintainability as it means that any changes have to be made in multiple places and also makes it a lot harder to add new checkouts as there is so much code needed to set them up.

This PR gets rid of `digitalCheckoutReducer.js`, `digitalCheckoutActions.js`, `paperCheckoutReducer.js` and `weeklyCheckoutReducer.js` and replaces them with `subscriptionsCheckoutReducer.js` and subscriptionsCheckoutActions.js` which work for all the products and should cover any future products as well.

There is still a lot of duplicate code in the validation and form submission code which I'd like to get rid off but this PR is already pretty big so I'll do that separately.

[**Trello Card**](https://trello.com/c/YzISKcSR/2392-refactor-the-common-code-across-the-checkouts)


support-frontend/assets/helpers/paymentIntegrations/readerRevenueApis.js
```diff
@@ -11,11 +11,11 @@ import { logPromise, pollUntilPromise } from 'helpers/promise';
 import { logException } from 'helpers/logger';
 import { fetchJson, getRequestOptions, requestOptions } from 'helpers/fetch';
 import trackConversion from 'helpers/tracking/conversions';
-import { type PaperFulfilmentOptions } from 'helpers/productPrice/fulfilmentOptions';
-import { type PaperProductOptions } from 'helpers/productPrice/productOptions';
+import type { FulfilmentOptions } from 'helpers/productPrice/fulfilmentOptions';
+import type { ProductOptions } from 'helpers/productPrice/productOptions';
```
johnduffell: what's the difference? if any

rupertbates: Outside the braces applies to everything within, inside only applies to the single type that follows the keyword

support-frontend/assets/helpers/subscriptionsForms/formReducer.js
```diff
@@ -0,0 +1,117 @@
...
+
+function createFormReducer(
+  initialCountry: IsoCountry,
+  product: SubscriptionProduct,
+  initialBillingPeriod: BillingPeriod,
+  startDate: Option<string>,
+  productOption: Option<ProductOptions>,
+  fulfilmentOption: Option<FulfilmentOptions>,
+) {
+  const user = getUser(); // TODO: Is this unnecessary? It could use the user reducer
```
lucymonie: I just looked at the initial state of the page and these details don't seem to be in the user object, but they are in the checkout object.

support-frontend/assets/pages/digital-subscription-checkout/helpers/paymentProviders.js
```diff
@@ -100,8 +108,8 @@ function showStripe(
 }
 
 function showPaymentMethod(
-  dispatch: Dispatch<Action>,
-  state: State,
+  dispatch: Dispatch<Action | Action>,
```
lucymonie: What goes on here?

support-frontend/assets/helpers/subscriptionsForms/formReducer.js
rupertbates: Yeah I think it isn't but I'm going to leave changing it for a future PR

lucymonie: 👍

PR #1810, opened 2019-05-17
Temporarily remove direct debit payment method

jfsoul:
## Why are you doing this?
There is going to be 2 GoCardless maintenance periods over the weekend so we are temporarily removing the direct debit payment method.

[**Trello Card**](https://trello.com/c/q1EZ6694/2402-investigate-gocardless-downtime-implications)


jfsoul:
Not yet, I'll put it on CODE in a bit.

jfsoul:
Looks good on CODE

PR #1819, opened 2019-05-21
Jtl ticker fix

jlieb10:
## Why are you doing this?

The ticker wasn't displaying when the amount was zero (falsey in JS). This hasn't been a problem previously because we've always had an initial amount. (Also apologies to reviewers for shoe-horning in a small copy change on this branch)

## Screenshots
zeroed ticker

### DIFF
--- a/support-frontend/assets/components/ticker/contributionTicker.jsx
+++ b/support-frontend/assets/components/ticker/contributionTicker.jsx
@@ -179,7 +179,7 @@ export default class ContributionTicker extends Component<PropTypes, StateTypes>
       https://davidwalsh.name/translate3d
       *************************************************************** */
     const progressBarAnimation = `translate3d(${percentageToTranslate(this.dataFromServer.totalContributed || 0, this.dataFromServer.goal || 0, this.tickerType)}%, 0, 0)`;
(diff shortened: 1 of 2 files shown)


support-frontend/assets/components/ticker/contributionTicker.jsx
```diff
@@ -179,7 +179,7 @@ export default class ContributionTicker extends Component<PropTypes, StateTypes>
       https://davidwalsh.name/translate3d
       *************************************************************** */
     const progressBarAnimation = `translate3d(${percentageToTranslate(this.dataFromServer.totalContributed || 0, this.dataFromServer.goal || 0, this.tickerType)}%, 0, 0)`;
-    const readyToRender = (this.state && this.state.count && this.state.count !== 0);
+    const readyToRender = (this.state && !Number.isNaN(this.state.count) && this.state.count > -1);
```
joelochlann: could be nicer for this check to happen when fetching the data from the server. feels a bit like sanity checks that are really about that data are spreading into other bits of logic. maybe if we do the `isNaN` checks in `getInitialTickerValues` and throw an error if we don't have real numbers? then maybe `readyToRender` could just depend on `this.dataFromServer`?

(that may not actually simplify things in practice but just a thought....)

PR #1822, opened 2019-05-22
Configure campaign currency

jlieb10:
## Why are you doing this?

All campaign-specific configuration should be abstracted to `campaigns.jsx` file. We identified that the local currency of the campaign is something that should be changed on a per-campaign basis and is currently being manually configured in the component. This PR moves the configuration to the configuration file, where it should be.


jlieb10:
@joelochlann would that mean putting like "GB" or "US" or "AU" in the `campaigns.jsx` file and then using the internationalisation helper in the component to get the currency?

PR #1823, opened 2019-05-22
add a new route to render the prod 500 error page in dev

johnduffell:
## Why are you doing this?

I want to make sentry flag every time a real person hits a 500 error.  The first problem was testing it, so I have added a new route so we can see the 500 error page in dev easily.

Normally it would show a stack trace, and running in runProd mode doesn't have hot reloading or the ability to quit play without killing sbt.

[**Trello Card**](https://trello.com/c/ecHp74BG/2356-spike-investigate-what-we-can-do-with-putting-error-page-alerts-into-sentry-to-make-it-obvious-that-something-is-a-p1-issue-incl)

## Screenshots

image

### DIFF
--- a/support-frontend/app/wiring/AppComponents.scala
+++ b/support-frontend/app/wiring/AppComponents.scala
@@ -2,7 +2,7 @@ package wiring
 
 import controllers.AssetsComponents
 import filters.{CacheHeadersCheck, SetCookiesCheck}
-import lib.CustomHttpErrorHandler
+import lib.{CustomHttpErrorHandler, ErrorController}
 import monitoring.{SentryLogging, StateMachineMonitor}
 import play.api.BuiltInComponentsFromContext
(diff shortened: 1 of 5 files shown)


support-frontend/app/wiring/AppComponents.scala
```diff
@@ -24,7 +24,7 @@ trait AppComponents extends PlayComponents
   with Monitoring {
   self: BuiltInComponentsFromContext =>
 
-  override lazy val httpErrorHandler = new CustomHttpErrorHandler(
+  private val customHandler: CustomHttpErrorHandler = new CustomHttpErrorHandler(
```
johnduffell: this should have been a lazy val, caused 500 errors on all 404s etc

johnduffell: fixed by https://github.com/guardian/support-frontend/pull/1828

PR #1824, opened 2019-05-22
Cloudwatch alarms for contributions

tomrf1:
## Why are you doing this?
Currently we have alarms to warn us if we have had no one-off contributions for an hour (in payment-api), but not for recurring.

This branch changes the `SendAcquisitionEvent` step to write cloudwatch metrics with name `PaymentSuccess`, and with the `PaymentProvider` and `ProductType` values as dimensions. 

The new cloudwatch alarms are for contributions only, but it would be easy to add more.

[**Trello Card**](https://trello.com/c/JEM3mVdt/1086-add-alarms-to-repeat-contributions)

### DIFF
--- a/support-workers/cloud-formation/src/templates/cfn-template.yaml
+++ b/support-workers/cloud-formation/src/templates/cfn-template.yaml
@@ -209,9 +209,8 @@ Resources:
     Condition: CreateProdResources
     Properties:
       AlarmActions:
-        - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:reader-revenue-dev
-      AlarmName: !Sub support-workers ${Stage} Recurring contributions with PayPal might be down
-      AlarmDescription: No successful recurring paypal contributions for an hour
+        - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:contributions-dev
+      AlarmName: !Sub support-workers ${Stage} No successful recurring paypal contributions for an hour
       MetricName: PaymentSuccess
       Namespace: support-frontend
       Dimensions:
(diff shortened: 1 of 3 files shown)


support-workers/cloud-formation/src/templates/cfn-template.yaml
```diff
@@ -203,3 +203,75 @@ Resources:
...
+        - Name: ProductType
+          Value: Contribution
+        - Name: Stage
+          Value: ${Stage}
+      ComparisonOperator: LessThanOrEqualToThreshold
+      Threshold: 0
+      Period: 300
+      EvaluationPeriods: 12
+      Statistic: Sum
+      TreatMissingData: breaching
```
johnduffell: this can be very confusing setting up an alarm.  if you have 12 evaluation periods, I have a feeling it will try hard to look back and find valid/nonmissing periods for longer than the 12 periods, and only if it can't find any data back for long enough it will go for breaching.  I think.  The documentation is very confusing though.

However if, you have copied this from the alarm that's working for single, then it must be right!

support-workers/src/main/scala/com/gu/support/workers/lambdas/SendAcquisitionEvent.scala
```diff
@@ -38,7 +41,13 @@ class SendAcquisitionEvent(serviceProvider: ServiceProvider = ServiceProvider)
     services.acquisitionService.submit(
       SendAcquisitionEventStateAndRequestInfo(state, requestInfo)
     ).fold(
-      errors => throw AnalyticsServiceErrorList(errors), _ => HandlerResult(Unit, requestInfo)
+      errors => throw AnalyticsServiceErrorList(errors),
+      _ => {
+        val cloudwatchEvent = paymentSuccessRequest(Configuration.stage, paymentProviderFromPaymentMethod(state.paymentMethod), state.product)
+        AwsCloudWatchMetricPut(client)(cloudwatchEvent)
+
+        HandlerResult(Unit, requestInfo)
```
johnduffell: `Unit` is a type! I realise it's not your change, but the value of `Unit` is `()`

support-workers/src/main/scala/com/gu/support/workers/lambdas/SendAcquisitionEvent.scala
```diff
@@ -38,7 +41,13 @@ class SendAcquisitionEvent(serviceProvider: ServiceProvider = ServiceProvider)
     services.acquisitionService.submit(
       SendAcquisitionEventStateAndRequestInfo(state, requestInfo)
     ).fold(
-      errors => throw AnalyticsServiceErrorList(errors), _ => HandlerResult(Unit, requestInfo)
+      errors => throw AnalyticsServiceErrorList(errors),
+      _ => {
+        val cloudwatchEvent = paymentSuccessRequest(Configuration.stage, paymentProviderFromPaymentMethod(state.paymentMethod), state.product)
+        AwsCloudWatchMetricPut(client)(cloudwatchEvent)
```
johnduffell: no contributions is a P1 issue. however no analytics is not a P1 issue.  So if we want to use this for detecting p1 issues we should probably have a separate alarm for it.  If not, we should make it more clear in the alarm description that it could just be analytics that's failing.

support-workers/src/main/scala/com/gu/support/workers/lambdas/SendAcquisitionEvent.scala
```diff
@@ -54,6 +63,13 @@ object SendAcquisitionEvent {
     override def getMessage: String = errors.map(_.getMessage).mkString(". ")
   }
 
+  def paymentProviderFromPaymentMethod(paymentMethod: PaymentMethod): thrift.PaymentProvider =
```
johnduffell: interesting that we are using thrift defintions for cloudwatch alarms. However at least it would set off the alarm if someone changed the thrift definition.

support-workers/src/main/scala/com/gu/support/workers/lambdas/SendAcquisitionEvent.scala
```diff
@@ -54,6 +63,13 @@ object SendAcquisitionEvent {
...
+    paymentMethod match {
+      case _: CreditCardReferenceTransaction => thrift.PaymentProvider.Stripe
+      case _: PayPalReferenceTransaction => thrift.PaymentProvider.Paypal
+      case _: DirectDebitPaymentMethod | _: ClonedDirectDebitPaymentMethod => thrift.PaymentProvider.Gocardless
```
johnduffell: what would happen if we got a new payment method?  Probably would want to add an alarm even if it were no payment for 24 hours.  I'm not sure if the code would throw an error if we missed one from the pattern match.

support-workers/cloud-formation/src/templates/cfn-template.yaml
tomrf1: I have indeed copied it from the payment-api alarm! It is a bit confusing though

support-workers/src/main/scala/com/gu/support/workers/lambdas/SendAcquisitionEvent.scala
tomrf1: good spot

support-workers/src/main/scala/com/gu/support/workers/lambdas/SendAcquisitionEvent.scala
tomrf1: yes `PaymentMethod` is a sealed trait so the compiler would warn.
Also, I haven't actually changed this code, just moved it out of a function

support-workers/src/main/scala/com/gu/support/workers/lambdas/SendAcquisitionEvent.scala
tomrf1: This was a good point - I have moved the cloudwatch request outside of the `future.fold` because we want it to happen regardless of whether the analytics request succeeds

PR #1826, opened 2019-05-24
Fix ticker counter for low amounts

tomrf1:
## Why are you doing this?
The ticker gets stuck if its initial value is below 100 (it's initialised by `totalContributed * 0.8`).
This change makes it jump straight to `totalContributed` if that's the case.

## Screenshots
Screenshot 2019-05-24 at 11 10 36

### DIFF
--- a/support-frontend/assets/components/ticker/contributionTicker.jsx
+++ b/support-frontend/assets/components/ticker/contributionTicker.jsx
@@ -128,14 +128,17 @@ export default class ContributionTicker extends Component<PropTypes, StateTypes>
   increaseTextCounter = () => {
     const nextCount = this.state.count + Math.floor(this.state.count / 100);
 
-    if (this.dataFromServer.totalContributed &&
-      (nextCount >= this.dataFromServer.totalContributed || nextCount <= this.state.count)) {
+    const finishedCounting = this.dataFromServer.totalContributed &&
+      (nextCount <= this.state.count || //count isn't going up because total is too small
+      nextCount >= this.dataFromServer.totalContributed);
+
+    if (finishedCounting) {
       this.setState({ count: this.dataFromServer.totalContributed });
     } else {
       this.setState({ count: nextCount });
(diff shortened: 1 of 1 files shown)


support-frontend/assets/components/ticker/contributionTicker.jsx
```diff
@@ -128,7 +128,8 @@ export default class ContributionTicker extends Component<PropTypes, StateTypes>
   increaseTextCounter = () => {
     const nextCount = this.state.count + Math.floor(this.state.count / 100);
 
-    if (this.dataFromServer.totalContributed && nextCount >= this.dataFromServer.totalContributed) {
+    if (this.dataFromServer.totalContributed &&
```
joelochlann: it's hard to tell just from the code what that `nextCount <= this.state.count` is actually doing - is there any way that variable names/comments could make it clear?

tomrf1: I've just pushed a commit that attempts to make it clearer

joelochlann: thanks!

support-frontend/assets/components/ticker/contributionTicker.jsx
```diff
@@ -128,13 +128,17 @@ export default class ContributionTicker extends Component<PropTypes, StateTypes>
   increaseTextCounter = () => {
     const nextCount = this.state.count + Math.floor(this.state.count / 100);
 
-    if (this.dataFromServer.totalContributed && nextCount >= this.dataFromServer.totalContributed) {
+    const finishedCounting = this.dataFromServer.totalContributed &&
+      (nextCount <= this.state.count || //count isn't going up because total is too small
```
jlieb10: Nice!

jlieb10:
Really great work figuring out this bug and with the naming and inline documentation @tomrf1 👍

PR #1829, opened 2019-05-24
Fix ticker behaviour when it exceeds goal

tomrf1:
## Why are you doing this?
The position of the progress bar relative to the marker is currently incorrect when we exceed the goal.
This PR fixes it, following the original logic in frontend introduced here https://github.com/guardian/frontend/pull/20874/files
I've had to make the marker a `div` rather than an `after` so that it can be transformed.

I've also fixed it so that it displays the current amount instead of the goal.

Before fix, for a very small amount above the goal it would exceed the marker by too much:
Screenshot 2019-05-24 at 17 26 59

## Screenshots
Above goal:
Screenshot 2019-05-24 at 17 40 29

Below goal:
Screenshot 2019-05-24 at 16 57 09

### DIFF
--- a/support-frontend/assets/components/ticker/contributionTicker.jsx
+++ b/support-frontend/assets/components/ticker/contributionTicker.jsx
@@ -205,11 +205,11 @@ export default class ContributionTicker extends Component<PropTypes, StateTypes>
               className="contributions-landing-ticker__filled-progress"
               style={{ transform: progressBarAnimation }}
             />
-            <div
-              className="contributions-landing-ticker__marker"
-              style={{ transform: markerAnimation }}
-            />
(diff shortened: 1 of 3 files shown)


jlieb10:
@joelochlann @tomrf1 - I built it this way intentionally because we didn't need it to be smart"
 about how much to show after the goal was reached, we just needed to demonstrate that it had been exceeded. I believe the designs for the ticker showed that we were always going to show the "goal" number on the right and never show the amount contributed after the goal was reached.

If there's a consensus to do it this way instead, that's fine- Just wanted to explain that I built it this way intentionally to reduce unneeded complexity and potential visual confusion.

PR #1832, opened 2019-05-28
Don't pass a promocode into the pricing lookup 

rupertbates:
Passing a promo code into the pricing lookup on the Guardian Weekly landing page is causing it to respond very slowly so we are going to stop passing them in while we investigate.


johnduffell:
MembershipSub-Promotions-PROD throttle events vs response time ELB
image

PR #1833, opened 2019-05-28
Improve price summary performance

rupertbates:
## Why are you doing this?
There was an issue with the `PriceSummaryService` which was causing price lookups with promo code to take a very long time. 

This was because of the way the code was structured which meant that a promo code lookup was  taking place deep inside a for loop resulting in many more table scans in Dynamo than were necessary.

This PR moves the promo code lookup to the start of the process and then passes the result through to subsequent functions to ensure there is only one table scan per price lookup.

Before this PR the Guardian Weekly landing page was taking up to 14 seconds to load, after this it is around 350 milliseconds.


support-services/src/main/scala/com/gu/support/pricing/PriceSummaryService.scala
```diff
@@ -12,20 +12,30 @@ import scala.math.BigDecimal.RoundingMode
 
 class PriceSummaryService(promotionService: PromotionService, catalogService: CatalogService) extends TouchpointService {
   private type GroupedPriceList = Map[(FulfilmentOptions, ProductOptions, BillingPeriod), Map[Currency, PriceSummary]]
+  case class PromotionWithCode(promoCode: PromoCode, promotion: Promotion)
+
+  private def findPromotion(maybePromoCode: Option[PromoCode]) =
+    for {
+      promoCode <- maybePromoCode
+      promotion <- promotionService.findPromotion(promoCode)
```
johnduffell: this will still be scanning the whole table (although only once in this case), I was under the impression the View table was there to allow lookup by promo code in the checkout.  I can see that based on the amount of data we have it's not too bad, but in that case do we need the View table at all?
This is all a bit fuzzy from recollection so I may have missed something else.

rupertbates: Unfortunately the view table doesn't have required fields in it for our purposes. We could look at adding them and switching to using the view table in a future PR if it seems useful.

PR #1839, opened 2019-05-30
feat: added a new productOption card to story book

damien-shaw-guardian:
feat: added a new productOption card to story book

## Why are you doing this?
This new card has been added to test out variants of the product option cards

[**Trello Card**](https://trello.com/c/7rn0MCD5/2324-product-option-card-design-iteration)

## Changes
New product option cards added

## Screenshots
Screen Shot 2019-05-30 at 15 16 49

Screen Shot 2019-05-30 at 15 17 00

Screen Shot 2019-05-30 at 15 17 12

### DIFF
--- a/support-frontend/assets/components/productOption/productOption.jsx
+++ b/support-frontend/assets/components/productOption/productOption.jsx
@@ -1,0 +1,68 @@
+// @flow
+import React from 'react';
+import type { Node } from 'react';
+import cx from 'classnames';
+
+// components
+import AnchorButton, { type PropTypes } from 'components/button/anchorButton';
+import { defaultProps } from 'components/button/_sharedButton';
+
+// styles
+import './productOption.scss';
+
+type Props = {
+  children: Node,
+};
+
+// hocs
+const withProductOptionsStyle = WrappedComponent => (props: PropTypes) => (
+  <div className="product-option__button">
+    <WrappedComponent {...props} />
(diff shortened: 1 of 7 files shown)


support-frontend/assets/components/productOption/productOption.jsx
```diff
@@ -0,0 +1,68 @@
+// @flow
+import React from 'react';
+import type { Node } from 'react';
```
lucymonie: I think you can roll these together if you want to, as
`import React, { type Node } from 'react';`

support-frontend/assets/components/productOption/productOption.jsx
```diff
@@ -0,0 +1,68 @@
...
+
+// hocs
+const withProductOptionsStyle = WrappedComponent => (props: PropTypes) => (
+  <div className="product-option__button">
+    <WrappedComponent {...props} />
+  </div>
+);
+
+// presentation components
+export const Line = ({ showLine, style }: { showLine: boolean, style?: { [string]: string } }) =>
```
rupertbates: I'm a bit worried that these names are very generic whereas their application is very specific (they are only used with the ProductOptions component), for instance we already have a Content element in layout.jsx which represents the content for the entire page.
Could you add some prefix onto them to make it clearer that they are subcomponents of the ProductOptions component?

PR #1840, opened 2019-05-31
Introductory promotions

rupertbates:
## Why are you doing this?

We want to use promotions to implement 6 for 6 rather than a lot of custom code because it allows a lot more flexibility for adding new prices and periods (n for n promotions) and would also mean that we could use them for other products. 

This is a first step towards that, it adds a new `IntroductoryPrice` type of promotion which describes n for n type promotions and hard codes an instance of this type (6 for 6) which applies to Guardian Weekly Quarterly pricing. The reason for hard coding this at this point is to avoid having to build the tooling to handle the new type in the membership promo codes tool.

The next step is to update the various React components to work with this new promotion type.

### DIFF
--- a/support-models/src/main/scala/com/gu/support/promotions/Benefit.scala
+++ b/support-models/src/main/scala/com/gu/support/promotions/Benefit.scala
@@ -3,10 +3,10 @@ package com.gu.support.promotions
 import cats.syntax.functor._
 import com.gu.support.encoding.Codec
 import com.gu.support.encoding.Codec.deriveCodec
+import com.gu.support.encoding.CustomCodecs._
 import io.circe.syntax._
 import io.circe.{Decoder, Encoder}
 import org.joda.time.{Days, Months}
-import com.gu.support.encoding.CustomCodecs._
 
 sealed trait Benefit
 
@@ -32,6 +32,29 @@ object IncentiveBenefit {
   val jsonName = "incentive"
 
   implicit val incentiveCodec: Codec[IncentiveBenefit] = deriveCodec
+}
+
+sealed trait IntroductoryPeriodType
+
+case object Issue extends IntroductoryPeriodType
+
+object IntroductoryPeriodType {
+  implicit val decodePeriod: Decoder[IntroductoryPeriodType] = Decoder.decodeString.map(code => fromString(code))
+  implicit val encodePeriod: Encoder[IntroductoryPeriodType] = Encoder.encodeString.contramap[IntroductoryPeriodType](_.toString)
+
+  private def fromString(s: String) = {
+    s.toLowerCase match {
+      case "issue" => Issue
+    }
+  }
+}
+
+case class IntroductoryPriceBenefit(price: Double, periodLength: Int, periodType: IntroductoryPeriodType) extends Benefit
+
+object IntroductoryPriceBenefit {
+  val jsonName = "introductory_price"
+
+  implicit val incentiveCodec: Codec[IntroductoryPriceBenefit] = deriveCodec
 }
 
 object Benefit {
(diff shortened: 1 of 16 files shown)


support-models/src/main/scala/com/gu/support/promotions/Benefit.scala
```diff
@@ -34,6 +34,29 @@ object IncentiveBenefit {
...
+
+object IntroductoryPriceBenefit {
+  val jsonName = "introductory_price"
+
+  implicit val incentiveCodec: Codec[IntroductoryPriceBenefit] = deriveCodec
+}
+
 object Benefit {
   implicit val encoder: Encoder[Benefit] = Encoder.instance {
     case d: DiscountBenefit => d.asJson
```
jfsoul: Do we need to add `IntroductoryPriceBenefit` here yet?

support-models/src/test/scala/com/gu/support/promotions/SerialisationSpec.scala
```diff
@@ -69,6 +69,17 @@ class SerialisationSpec extends FlatSpec with SerialisationTestHelpers with Lazy
     )
   }
 
+  it should "be able to decode an introductory price Promotion" in {
+    testDecoding[Promotion](Fixtures.introductoryPricePromotion,
+      promotion => {
+        promotion.introductoryPrice.isDefined shouldBe true
```
jfsoul: nitpick: `should be (defined)` might give a better fail message on this assertion.

support-services/src/main/scala/com/gu/support/pricing/PriceSummaryService.scala
```diff
@@ -61,16 +61,18 @@ class PriceSummaryService(promotionService: PromotionService, catalogService: Ca
...
-      promotion.promotion.discount.map(getDiscountedPrice(price, _, billingPeriod).value),
-      promotion.promotion.discount.flatMap(_.durationMonths).map(getNumberOfDiscountedPeriods(_, billingPeriod)),
-      promotion.promotion.discount,
-      promotion.promotion.freeTrial,
-      promotion.promotion.incentive
+      name = promotion.name,
+      description = promotion.description,
+      promoCode = promoCode,
+      discountedPrice = promotion.discount.map(getDiscountedPrice(price, _, billingPeriod).value),
+      numberOfDiscountedPeriods = promotion.discount.flatMap(_.durationMonths).map(getNumberOfDiscountedPeriods(_, billingPeriod)),
```
jfsoul: is this still valid for the new promotion type? do we need to update `getNumberOfDiscountedPeriods`?

support-services/src/main/scala/com/gu/support/promotions/PromotionService.scala
```diff
@@ -1,25 +1,49 @@
...
 import com.gu.support.touchpoint.TouchpointService
 import com.gu.support.zuora.api.SubscriptionData
 import com.typesafe.scalalogging.LazyLogging
+import org.joda.time.DateTime
 
 class PromotionService(config: PromotionsConfig, maybeCollection: Option[PromotionCollection] = None) extends TouchpointService with LazyLogging {
   val promotionCollection = maybeCollection.getOrElse(new DynamoPromotionCollection(config.tables))
 
+  val SixForSixPromotion = Promotion(
+    "Six For Six",
```
jfsoul: I'd find named parameters a little easier to follow, but understand if this is temporary.

support-models/src/main/scala/com/gu/support/promotions/Benefit.scala
rupertbates: Seems like we should doesn't it? Although it appears to work without, I'll investigate why...

support-services/src/main/scala/com/gu/support/pricing/PriceSummaryService.scala
rupertbates: No this field only applies to discounts as does `discountedPrice`. The reason these are separated out is because they are not part of the benefit itself, they need to be worked out in conjunction with the price.

PR #1841, opened 2019-05-31
Thank you 'set password' screen UX updates

jlieb10:
## Why are you doing this?
UX design changes requested by @ionamckendrick to indicate that the user isn't on the 'final' thank you page they will see, but still in part of the flow.

[**Trello Card**](https://trello.com/c/b7e7Lhue/1105-visual-design-and-implementation-for-set-password-sign-up-screens)

## Changes

* Removing page-level class for thank you set password screen
* Introducing component-level class granularity to make these screens more customizable in the future (as might be the case with forthcoming updates to the flow)

## Screenshots
set password page

### DIFF
--- a/support-frontend/assets/components/page/page.jsx
+++ b/support-frontend/assets/components/page/page.jsx
@@ -3,12 +3,10 @@
 // ----- Imports ----- //
 
 import React, { type Node } from 'react';
-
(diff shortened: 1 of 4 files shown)


tomrf1:
This change has broken the post-deployment tests, because it relies on a css class being present on the thank you page. Everything seems to be working fine in PROD through

PR #1845, opened 2019-06-03
Billing period component

rupertbates:
## Why are you doing this?
We have a lot of different places where pricing related copy is defined in the app, this leads to inconsistencies and makes it hard to make changes. This PR tries to centralise all pricing related copy in a new `priceDescription` module and makes that module aware of promotions, particularly introductory price type promotions which are used to model 6 for 6.

I have also created a new component `BillingPeriodSelector` which handles the display of available billing periods and their prices in the DP & GW checkout forms.

### DIFF
--- a/support-frontend/assets/components/subscriptionCheckouts/billingPeriodSelector.jsx
+++ b/support-frontend/assets/components/subscriptionCheckouts/billingPeriodSelector.jsx
@@ -10,6 +10,7 @@ import type { ProductPrices } from 'helpers/productPrice/productPrices';
 import { getProductPrice } from 'helpers/productPrice/productPrices';
 import type { IsoCountry } from 'helpers/internationalisation/country';
 import type { FulfilmentOptions } from 'helpers/productPrice/fulfilmentOptions';
+import type { Action } from 'helpers/subscriptionsForms/formActions';
 import {
   extendedGlyph,
   fromCountryGroupId,
@@ -25,9 +26,9 @@ type PropTypes = {|
   productPrices: ProductPrices,
   billingPeriods: BillingPeriod[],
   fulfilmentOption?: FulfilmentOptions,
-  country: IsoCountry,
+  billingCountry: IsoCountry,
   selected: BillingPeriod,
-  onChange: (BillingPeriod) => void,
+  onChange: (BillingPeriod) => Action,
 |}
 
 function getTitle(billingPeriod: BillingPeriod) {
@@ -47,14 +48,14 @@ function BillingPeriodSelector(props: PropTypes) {
         {props.billingPeriods.map((billingPeriod) => {
           const productPrice = getProductPrice(
             props.productPrices,
-            props.country,
+            props.billingCountry,
             billingPeriod,
             props.fulfilmentOption,
           );
           return (<RadioInputWithHelper
             text={getTitle(billingPeriod)}
             helper={getPriceDescription(
-              getGlyph(props.country),
+              getGlyph(props.billingCountry),
               productPrice,
               billingPeriod,
             )}
--- a/support-frontend/assets/components/subscriptionCheckouts/billingPeriodSelector.jsx
+++ b/support-frontend/assets/components/subscriptionCheckouts/billingPeriodSelector.jsx
@@ -1,0 +1,74 @@
+// @flow
+import React from 'react';
+
+import { Fieldset } from 'components/forms/fieldset';
+import { RadioInputWithHelper } from 'components/forms/customFields/radioInputWithHelper';
+import type { BillingPeriod } from 'helpers/billingPeriods';
+import { SixWeekly } from 'helpers/billingPeriods';
+import { FormSection } from 'components/checkoutForm/checkoutForm';
+import type { ProductPrices } from 'helpers/productPrice/productPrices';
+import { getProductPrice } from 'helpers/productPrice/productPrices';
+import type { IsoCountry } from 'helpers/internationalisation/country';
+import type { FulfilmentOptions } from 'helpers/productPrice/fulfilmentOptions';
+import {
+  extendedGlyph,
+  fromCountryGroupId,
+} from 'helpers/internationalisation/currency';
+import { getPriceDescription } from 'helpers/productPrice/priceDescriptions';
+import {
+  fromCountry,
+  GBPCountries,
+} from 'helpers/internationalisation/countryGroup';
+import { Domestic } from 'helpers/productPrice/fulfilmentOptions';
+
+type PropTypes = {|
+  productPrices: ProductPrices,
+  billingPeriods: BillingPeriod[],
+  fulfilmentOption?: FulfilmentOptions,
+  country: IsoCountry,
+  selected: BillingPeriod,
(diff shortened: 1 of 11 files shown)


support-frontend/assets/components/subscriptionCheckouts/billingPeriodSelector.jsx
```diff
@@ -0,0 +1,67 @@
...
+
+function BillingPeriodSelector(props: PropTypes) {
+  return (
+    <FormSection title="How often would you like to pay?">
+      <Fieldset legend="How often would you like to pay?">
+        {props.billingPeriods.map((billingPeriod) => {
+          const productPrice = getProductPrice(
+            props.productPrices,
+            props.billingCountry,
+            billingPeriod === SixWeekly ? Quarterly : billingPeriod, // for 6 for 6 we need the quarterly pricing
```
lucymonie: Is this because 6 for 6 is really a quarterly order with a discount applied?

support-frontend/assets/helpers/productPrice/priceDescriptions.js
```diff
@@ -0,0 +1,85 @@
...
+
+const billingPeriodQuantifier = (numberOfBillingPeriods: number, noun: string) =>
+  (numberOfBillingPeriods > 1 ?
+    `/${noun} for ${numberOfBillingPeriods} ${noun}s` :
+    ` for 1 ${noun}`);
+
+const billingPeriodDescription = (billingPeriod: BillingPeriod) =>
+  billingPeriodNoun(billingPeriod).toLowerCase();
+
+const standardRate = (glyph: string, price: number, billingPeriod: BillingPeriod) => `${displayPrice(glyph, price)} every ${billingPeriodDescription(billingPeriod)}`;
```
lucymonie:
```suggestion
const standardRate = (glyph: string, price: number, billingPeriod: BillingPeriod) => 
`${displayPrice(glyph, price)} every ${billingPeriodDescription(billingPeriod)}`;
```

support-frontend/assets/helpers/productPrice/priceDescriptions.js
```diff
@@ -0,0 +1,85 @@
...
+
+function getDiscountDescription(
+  glyph: string,
+  price: number,
+  discountedPrice: number,
+  numberOfDiscountedPeriods: ?number,
+  billingPeriod: BillingPeriod,
+) {
+  const noun = billingPeriodDescription(billingPeriod);
```
lucymonie: 'noun' seems a bit generic - is it possible to use a more descriptive term?

support-frontend/assets/helpers/productPrice/weeklyProductPrice.js
```diff
@@ -22,28 +33,38 @@ const getPromotionPrice = (country: IsoCountry, period: WeeklyBillingPeriod, pro
   ].join('');
 };
 
+function getPromotion(
```
lucymonie: Is there a reason to call `getPromotion` and then just calling `genericGetPromotion`? Would it be less code just to call `genericGetPromotion`?

support-frontend/assets/helpers/productPrice/weeklyProductPrice.js
```diff
@@ -22,28 +33,38 @@ const getPromotionPrice = (country: IsoCountry, period: WeeklyBillingPeriod, pro
...
+  return genericGetPromotion(
+    productPrices,
+    country,
+    billingPeriod,
+    fulfilmentOption,
+    NoProductOptions,
+  );
+}
+
+function regularPrice(
```
lucymonie: And here? Is this a pattern I'm missing?

support-frontend/assets/components/subscriptionCheckouts/billingPeriodSelector.jsx
rupertbates: Yes that's right, it's all a bit messy unfortunately

support-frontend/assets/helpers/productPrice/weeklyProductPrice.js
lucymonie: Also, should it be something like `getRegularPrice` and `genericGetRegularPrice`?

support-frontend/assets/pages/weekly-subscription-checkout/components/weeklyCheckoutForm.jsx
```diff
@@ -69,13 +65,14 @@ import { getWeeklyDays } from 'pages/weekly-subscription-checkout/helpers/delive
...
 import { routes } from 'helpers/routes';
+import { BillingPeriodSelector } from 'components/subscriptionCheckouts/billingPeriodSelector';
 
 // ----- Types ----- //
 
 type PropTypes = {|
   ...FormFields,
-  country: IsoCountry,
-  countryGroupId: CountryGroupId,
+  billingCountry: IsoCountry,
```
lucymonie: Yay! Very happy to see this 🎉

support-frontend/assets/pages/weekly-subscription-checkout/components/weeklyCheckoutForm.jsx
```diff
@@ -117,43 +116,12 @@ const BillingAddress = withStore(countries, 'billing', getBillingAddress);
...
-  [BillingPeriod]: {
-    title: string,
-    copy: string,
-    offer: Option<string>,
-    priceObject: Price,
-  }
-}
 
 function WeeklyCheckoutForm(props: PropTypes) {
-  const plans: Plans = Object.keys(displayBillingPeriods).reduce((ps, billingPeriod) => ({
```
lucymonie: I'm pretty happy to see the back of this piece of code...

support-frontend/assets/helpers/productPrice/weeklyProductPrice.js
rupertbates: Yeah I went backwards and forwards with it, I think it was originally there to reduce the number of params that need to be passed but I agree that it is not really adding any value now. I'll remove it.

support-frontend/assets/helpers/productPrice/weeklyProductPrice.js
rupertbates: Same as above

support-frontend/assets/components/subscriptionCheckouts/billingPeriodSelector.jsx
lucymonie: 👍

support-frontend/assets/helpers/productPrice/priceDescriptions.js
rupertbates: I think noun is appropriate though because is the noun version of the billing period eg. year, month etc. I'll rename the function instead, that is really only a little helper to avoid having to call `toLowerCase` in multiple places.

support-frontend/assets/helpers/productPrice/weeklyProductPrice.js
rupertbates: Actually I'll do this in my next PR because I've already done it in that branch and don't want to redo the work.

PR #1846, opened 2019-06-03
Remove 'new' prefix in the codebase

jlieb10:
## Why are you doing this?

Health work - best practice to remove 'new' prefix on files, folders, components, etc. as soon as possible.

[**Trello Card**](https://trello.com/c/sfnmWrpg/1129-clean-up-all-the-new-stuff-in-support-frontend-new-contributions-flow-marketingconsentnew)

## Screenshots
new-contrib
newcontrib


tomrf1:
This is great, though I'm wondering if we should delete all of the old (unused) contributions page files first?

jlieb10:
@tomrf1 @joelochlann I used https://github.com/github/eslint-plugin-github to identify unused exports but it's very inaccurate -- it marks anything that's an `export default ...` as an unused export. My plan is to manually look through the components folder to make sure everything is being imported somewhere. If I find any exports that aren't imported somewhere, I will remove them and re-check anything they were importing to make sure it's imported in another location.

support-frontend/assets/pages/contributions-landing/components/ContributionFormFields.jsx
```diff
@@ -163,6 +163,4 @@ function FormFields(props: PropTypes) {
   );
 }
 
-const ContributionFormFields = connect(mapStateToProps, mapDispatchToProps)(FormFields);
-
-export { ContributionFormFields };
+export default connect(mapStateToProps, mapDispatchToProps)(ContributionFormFields);
```
joelochlann: 🥇 for bringing this in to line with the repo-wide convention that components have default exports

joelochlann:
> I used https://github.com/github/eslint-plugin-github to identify unused exports but it's very inaccurate -- it marks anything that's an export default ... as an unused export.

oh, have you tried `unused-module` instead of `unused-export`? https://github.com/github/eslint-plugin-github/blob/master/docs/rules/unused-module.md

Where there's a default export, it should be the only export - meaning unused module is equivalent to unused export for those files

support-frontend/assets/pages/contributions-landing/components/ContributionFormFields.jsx
jlieb10: 👍

jlieb10:
@joelochlann no I will check it out though

PR #1847, opened 2019-06-03
Clean up legacy routes

joelochlann:
Some long-awaited cleanup of routes no longer used in the new payment flow world.

We don't need a specific route for `cancelURL` even though PayPal does accept one as a parameter with `create-payment` requests, since we just send users back to the contributions page if they cancel.

We do need a specific route for errors, since we need to tell the user their payment has failed. Confusingly, we're the ones who redirect users to this page, not PayPal, because we find this out when we make the `execute-payment` request after the user has completed the PayPal flow.

### DIFF
--- a/support-frontend/app/controllers/PayPalOneOff.scala
+++ b/support-frontend/app/controllers/PayPalOneOff.scala
@@ -67,7 +67,7 @@ class PayPalOneOff(
 
   def resultFromPaypalSuccess(success: PayPalSuccess, country: String, isTestUser: Boolean)(implicit request: RequestHeader): Result = {
     SafeLogger.info(s"One-off contribution for Paypal payment is successful")
-    val redirect = Redirect("/contribute/one-off/thankyou")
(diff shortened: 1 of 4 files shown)


support-frontend/app/controllers/PayPalOneOff.scala
```diff
@@ -65,17 +65,11 @@ class PayPalOneOff(
...
+  def resultFromPaypalSuccess(success: PayPalSuccess, country: String, isTestUser: Boolean)(implicit request: RequestHeader): Result = {
     SafeLogger.info(s"One-off contribution for Paypal payment is successful")
-    val redirect = Redirect {
-      country match {
-        // TODO: more cleanup
-        case Some(c) => s"/$c/thankyou"
-        case None => "/contribute/one-off/thankyou"
-      }
-    }
+    val redirect = Redirect("/contribute/one-off/thankyou")
```
tomrf1: shouldn't this be `s"/$c/thankyou"`?

PR #1850, opened 2019-06-03
Always send email to payment api

joelochlann:
We're changing the Payment API so it never uses the email from PayPal, only ever the email in the form.

The optional `signedInUserEmail` (which overrides the PayPal email if supplied, so signed-in users don't get an extra account created against their PayPal account, thankfully) is therefore deprecated. And we need mandatory `email` field instead.

This makes the change on the support-frontend side. From Payment API's perspective, it will just be receiving an extra param so nothing should break. Then we'll make the Payment API change so that email is mandatory in the request.


support-frontend/app/services/PaymentAPIService.scala
```diff
@@ -35,7 +35,10 @@ object PaymentAPIResponseError {
 }
 
 case class ExecutePaymentBody(
-    signedInUserEmail: Option[String],
+    // TODO: question: should we put the email in the paymentData for consistency with Stripe?
```
tomrf1: I think it's fine here

support-frontend/app/services/PaymentAPIService.scala
```diff
@@ -100,12 +97,11 @@ class PaymentAPIService(wsClient: WSClient, paymentAPIUrl: String)(implicit ec:
   def executePaypalPayment(
     paymentJSON: JsObject,
     acquisitionData: JsValue,
-    queryStrings: Map[String, Seq[String]],
-    email: Option[String],
+    email: String,
```
joelochlann: The existing code passes through arbitrary `queryStrings` to the Payment API. I can't see any use for this (the only query string it accepts is `mode=test`) so I've removed

support-frontend/app/services/PaymentAPIService.scala
```diff
@@ -100,12 +97,11 @@ class PaymentAPIService(wsClient: WSClient, paymentAPIUrl: String)(implicit ec:
...
     isTestUser: Boolean,
     userAgent: Option[String]
   )(implicit ec: ExecutionContext): EitherT[Future, PaymentAPIResponseError[PayPalError], PayPalSuccess] = {
-    val data = ExecutePaymentBody(email, acquisitionData, paymentJSON)
-    postPaypalData(data, queryStrings, isTestUser, userAgent).subflatMap(decodePaymentAPIResponse[PayPalError, PayPalSuccess])
+    val data = ExecutePaymentBody(Some(email), email, acquisitionData, paymentJSON)
```
jlieb10: Do we want this to accept `Some(email)` and `email` as args? My knowledge of Scala isn't extensive but I thought you could only call `Some` on an option, and email is now a string

support-frontend/assets/helpers/routes.js
```diff
@@ -54,8 +54,8 @@ function payPalCancelUrl(cgId: CountryGroupId): string {
   return `${getOrigin()}/${countryGroups[cgId].supportInternationalisationId}/contribute`;
 }
 
-function payPalReturnUrl(cgId: CountryGroupId): string {
-  return `${getOrigin()}/${countryGroups[cgId].supportInternationalisationId}/paypal/rest/return`;
+function payPalReturnUrl(cgId: CountryGroupId, email: string): string {
+  return `${getOrigin()}/${countryGroups[cgId].supportInternationalisationId}/paypal/rest/return?email=${email}`;
```
jlieb10: 👍

support-frontend/app/services/PaymentAPIService.scala
jlieb10: 👍

support-frontend/app/services/PaymentAPIService.scala
```diff
@@ -35,9 +35,14 @@ object PaymentAPIResponseError {
 }
 
 case class ExecutePaymentBody(
-    signedInUserEmail: Option[String],
-    acquisitionData: JsValue,
-    paymentData: JsObject
+  // TODO: remove this field once the Payment API switches over to use mandatory email field
```
jlieb10: Great documentation

support-frontend/app/services/PaymentAPIService.scala
jlieb10: Ah I think I understand - Some(email) _returns_ an option here and this will also get removed after the Payment API switches over

joelochlann: yep, a type of `Option[String]` has one of two forms, either `Some("some string")`, or `None`. To construct an `Option[T]` from a `T` you do `Some(thingOfTypeT)`

jlieb10:
Fix for this is here: https://github.com/guardian/support-frontend/pull/1853

PR #1851, opened 2019-06-04
Health: Reusable Button Component

jlieb10:
## Why are you doing this?

Small health upgrade: I deleted the last instance of the "button with right arrow" component in favor of the more generic button component shared across the support site. I also added the ability for the shared button to be disabled.

## Changes

* Removed the last instance of the 'button with right arrow' and replaced with our generic, reusable button component
* Removed the files for the button
* Added disabled prop to the generic button component


jlieb10:
Prout silently failed for this one but this is the related fix https://github.com/guardian/support-frontend/pull/1853

PR #1852, opened 2019-06-04
[WIP] Add gifting to weekly checkout

lucymonie:
## Why are you doing this?
The intention is that users can give Guardian Weekly as a gift, and specify this in the checkout.

There are a number of assumptions underpinning the changes, which are listed in the [**Trello Card**](https://trello.com/c/JwbDPHwv/2412-gifting-in-gw-checkout-frontend)

## Changes
* There's a checkbox that users can select that indicates the order is a gift
* Users who select 6 for 6 won't see the gifting option
* Users who select the gifting option won't see 6 for 6 as an alternative option further down in the checkout
* When the user checks the gifting box, additional fields appear asking for the gift recipient's details and a form for the recipient's address (aka the delivery address). There's still a section asking if billing is the same as delivery address, but the wording is slightly different to reflect that delivery is to a gift recipient.

## Work yet to be completed
There's additional business logic to be applied to the checkout around the billing period selected by the user and which other options should be visible in the checkout

## Not a gift (screenshots)

Screen Shot 2019-06-04 at 14 36 57
...
...
...
Screen Shot 2019-06-04 at 14 43 05

## Gift (screenshots)
Screen Shot 2019-06-04 at 14 41 02
Screen Shot 2019-06-04 at 14 36 09
...
...
...
Screen Shot 2019-06-04 at 14 42 21


support-frontend/assets/pages/weekly-subscription-checkout/components/weeklyCheckoutForm.jsx
```diff
@@ -213,9 +215,48 @@ function WeeklyCheckoutForm(props: PropTypes) {
...
+                lastName={props.lastNameGiftRecipient}
+                setLastName={props.setLastNameGift}
+                email={props.emailGiftRecipient}
+                setEmailGift={props.setEmailGift}
+                formErrors={((props.formErrors: any): FormError<PersonalDetailsFormField>[])}
+                isGiftRecipient
+              />
+            </FormSection>)
+          : null}
+          {props.orderIsAGift ? (
```
rupertbates: I don't think you need another conditional check here do you? you can just include this in the one above.

lucymonie: Very true. Thanks Rupert!

PR #1853, opened 2019-06-04
Post-deployment test false positive fix

jlieb10:
## Why are you doing this?

The post-deployment test was expecting a button with a certain selector to be present in order to confirm the page loaded. The selector changed on an earlier PR (https://github.com/guardian/support-frontend/pull/1851) so this test needs to be updated correspondingly. It is running and passing locally.

### DIFF
--- a/support-frontend/assets/components/button/_sharedButton.jsx
+++ b/support-frontend/assets/components/button/_sharedButton.jsx
@@ -32,6 +32,7 @@ type SharedButtonPropTypes = {|
   iconSide: IconSide,
   getRef?: (?Element) => void,
   modifierClasses: string[],
+  postDeploymentTestID?: string,
 |};
 
 type PropTypes = {
@@ -42,7 +43,7 @@ type PropTypes = {
 // ----- Render ----- //
 
 const SharedButton = ({
-  element, appearance, iconSide, modifierClasses, children, icon, getRef, ...otherProps
+  element, appearance, iconSide, modifierClasses, children, icon, getRef, postDeploymentTestID, ...otherProps
 }: PropTypes) => {
 
   const className = classNameWithModifiers('component-button', [
@@ -59,6 +60,7 @@ const SharedButton = ({
   return createElement(element, {
     className,
     ref: getRef,
+    id: postDeploymentTestID,
     ...otherProps,
   }, contents);
 };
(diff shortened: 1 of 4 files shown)


johnduffell:
what a great idea joe, this is similar to dotcom's way of doing "js-" selectors and I wholeheartely support that idea.

support-frontend/test/selenium/contributions/pages/ContributionsLanding.scala
```diff
@@ -8,7 +8,7 @@ case class ContributionsLanding(region: String, testUser: TestUser)(implicit val
 
   val url = s"${Config.supportFrontendUrl}/$region/contribute"
 
-  private val contributeButton = className("form__submit-button")
+  private val contributeButton = cssSelector("#contributions-landing-submit-contribution-button")
```
johnduffell: by class name with an additional class name was fine to be honest, the main thing was the clue in the name that it realtes to the test.  Even now, the actual name itself doesn't make it clear it's for the tests (making it harder to review the test for compliance with the new "rule".
I realise the fashion is against overloading names with extra type information, but this area is pretty weakly typed.

jlieb10: I think we should use ID (or another attr) to keep our separation of concerns clear. I think it's best if we keep class to indicate the way that something will be referenced on the DOM to apply styles. By keeping it separate, we are indicating this is an attribute which can't be arbitrarily changed and only checked against styles

johnduffell: fair enough originally I just thought class might be simpler :+1:   I am a bit stronger on the idea of putting a prefix to make it more obvious that it's an ID designed for this purpose rather than any other.
But either way better to get this shipped and then address the niceties in a followup to avoid having broken tests for longer than necessary

jlieb10: Cool i've added the qa prefix - now will merge when team city gives a passing build 👍

support-frontend/assets/components/button/_sharedButton.jsx
```diff
@@ -59,6 +66,7 @@ const SharedButton = ({
   return createElement(element, {
     className,
     ref: getRef,
+    id: postDeploymentTestID ? `qa-${postDeploymentTestID}` : null,
```
johnduffell: nice neat idea doing it at this point (when the type is lost, add the type information into the value)

PR #1854, opened 2019-06-04
add latency/5xx alarms and improve descriptions

johnduffell:
## Why are you doing this?

Two reasons:

1. Following the outage on fri 24th may, we didn't actually get an alarm when the instances were sick.  This is because latency skyrocketed and the ELB was throwing 5xx errors, however our alarm is only on instance 5xx errors.  (in any case, 5xx errors were being thrown around that time due to another issue that hadn't been rolled back)
This PR adds both alarms above, and defines them as high severity in the [alerts doc.](https://docs.google.com/document/d/1_3El3cly9d7u_jPgTcRjLxmdG2e919zCLvmcFCLOYAk/edit)

2. I have also taken the opportunity to move away from a technical notation of the alerts and more towards and action and impact based notation.
The essence is that every alarm name is of the form
`HIGH severity - <service> <what is happening>`
and every description is of the form
`Impact - <what is degraded or broken for the end user>. Follow the process in https://docs.google.com/document/d/1_3El3cly9d7u_jPgTcRjLxmdG2e919zCLvmcFCLOYAk/edit`

[**Trello Card**](https://trello.com/c/jYtcYsmQ/2414-did-we-and-should-we-get-alarms-for-the-initial-and-later-incidents-on-25th-may#)

### DIFF
--- a/support-frontend/cloud-formation/cfn.yaml
+++ b/support-frontend/cloud-formation/cfn.yaml
@@ -278,7 +278,8 @@ Resources:
     Properties:
       AlarmActions:
       - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:reader-revenue-dev
-      AlarmName: !Sub No healthy instances for support-frontend in ${Stage}
+      AlarmName: !Sub HIGH severity - No healthy instances for support-frontend in ${Stage}
+      AlarmDescription: Impact - Cannot sell any subscriptions or contributions products. https://docs.google.com/document/d/1_3El3cly9d7u_jPgTcRjLxmdG2e919zCLvmcFCLOYAk/edit
       MetricName: HealthyHostCount
       Namespace: AWS/ApplicationELB
       Dimensions:
@@ -301,7 +302,8 @@ Resources:
     Properties:
       AlarmActions:
       - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:reader-revenue-dev
-      AlarmName: !Sub High 5XX rate for support-frontend in ${Stage}
+      AlarmName: !Sub HIGH severity - support-frontend instances are returning 5XX errors in ${Stage}
+      AlarmDescription: Impact - Some or all actions on support website are failing. https://docs.google.com/document/d/1_3El3cly9d7u_jPgTcRjLxmdG2e919zCLvmcFCLOYAk/edit
       MetricName: HTTPCode_Target_5XX_Count
       Namespace: AWS/ApplicationELB
       Dimensions:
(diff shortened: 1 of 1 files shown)


rupertbates:
I like the general idea but we seem to only have high severity alerts which makes the severity a bit redundant. Are you planning on reclassifying some of the existing ones? The state machine execution failure is probably less high severity than most.

johnduffell:
thanks @rupertbates actually the severity isn't just over cloudwatch alarms, it should cover any alert we get within the team.  If you check the doc there are some low severity too. 
I added a "warnings" alarm a while ago, for post checkout failures e.g. setting password or signing up for email, but I can't see it any more.  Those would be dealt with in the next sprint.

johnduffell:
ah apparently I remove it in https://github.com/guardian/support-frontend/pull/1372 as we decided email should only be used for high severity issues!  I'm thinking having a clear framework for dealing with situations can be especially useful for people joining the team.

support-frontend/cloud-formation/cfn.yaml
```diff
@@ -278,7 +278,8 @@ Resources:
     Properties:
       AlarmActions:
       - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:reader-revenue-dev
-      AlarmName: !Sub No healthy instances for support-frontend in ${Stage}
+      AlarmName: !Sub HIGH severity - No healthy instances for support-frontend in ${Stage}
```
jfsoul: What do you think about alarming on <3 rather than 0 healthy instances?  I think there's an argument for an alarm if we are down to 1 healthy - or cycling instances.

support-frontend/cloud-formation/cfn.yaml
```diff
@@ -301,7 +302,8 @@ Resources:
     Properties:
       AlarmActions:
       - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:reader-revenue-dev
-      AlarmName: !Sub High 5XX rate for support-frontend in ${Stage}
+      AlarmName: !Sub HIGH severity - support-frontend instances are returning 5XX errors in ${Stage}
+      AlarmDescription: Impact - Some or all actions on support website are failing. Follow the process in https://docs.google.com/document/d/1_3El3cly9d7u_jPgTcRjLxmdG2e919zCLvmcFCLOYAk/edit
```
jfsoul: Is this alarm strictly necessary? I don't remember seeing high 5XX from target and that not also be reflected by the ELB having high 5XX

johnduffell: yes it is needed.  The reports are Either/OR, rather than one including the other.

support-frontend/cloud-formation/cfn.yaml
johnduffell: yeah good idea on membership and subs we have a less than the correct number for 10 minutes, as well as a no healthy instances for 1 minute.  This is because of being burnt in the past not finding out instances couldn't start until it was too late.
I think I will add a card for that.

johnduffell: the card is https://trello.com/c/mIRQe53c/2416-add-alert-for-less-than-required-instances-of-support-frontend-for-10-min-period

PR #1855, opened 2019-06-04
Add more useful error message when page fails to render

jranks123:
## Why are you doing this?

Adds the exception message to the reported error message in the instance where a page fails to render.

### DIFF
--- a/support-frontend/assets/helpers/render.js
+++ b/support-frontend/assets/helpers/render.js
@@ -19,10 +19,10 @@ const getElementOrBody = (id: ?string): HTMLElement => {
   return element;
 };
 
-const renderError = (e: any, id: ?string) => {
+const renderError = (e: Error, id: ?string) => {
   const element = getElementOrBody(id);
 
-  logException(`Fatal error rendering a page. ${id ? id : ""} Error string: ${e}`);
+  logException(`Fatal error rendering page: ${id ? id : ""}. Error message: ${e.message}. Stack trace: ${e.stack ? e.stack : "none"}`);
   import('pages/error/components/errorPage').then(({ default: ErrorPage }) => {
     if (element) {
       ReactDOM.render(ErrorPage({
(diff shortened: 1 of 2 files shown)


support-frontend/assets/helpers/render.js
```diff
@@ -19,10 +19,10 @@ const getElementOrBody = (id: ?string): HTMLElement => {
   return element;
 };
 
-const renderError = (id: ?string) => {
+const renderError = (e: any, id: ?string) => {
   const element = getElementOrBody(id);
 
-  logException(['Fatal error rendering a page', id ? `id:${id}` : null].join(' '));
+  logException(['Fatal error rendering a page.', id ? `id:${id}.` : null, 'Error string:', e].join(' '));
```
joelochlann: not your changes, but can we just use string interpolation here?

jranks123: sure thing

support-frontend/assets/helpers/render.js
```diff
@@ -19,10 +19,10 @@ const getElementOrBody = (id: ?string): HTMLElement => {
   return element;
 };
 
-const renderError = (id: ?string) => {
+const renderError = (e: any, id: ?string) => {
   const element = getElementOrBody(id);
 
-  logException(['Fatal error rendering a page', id ? `id:${id}` : null].join(' '));
+  logException(`Fatal error rendering a page. ${id ? id : ""} Error string: ${e}`);
```
joelochlann: needs to be `e.message`

support-frontend/assets/helpers/render.js
```diff
@@ -19,10 +19,10 @@ const getElementOrBody = (id: ?string): HTMLElement => {
   return element;
 };
 
-const renderError = (id: ?string) => {
+const renderError = (e: any, id: ?string) => {
```
joelochlann: I'd give this a type of `Error`

support-frontend/assets/helpers/render.js
joelochlann: and for good measure you could log `e.stack` too

PR #1857, opened 2019-06-04
Price summary updates

rupertbates:
## Why are you doing this?

We have previously had two types which describe subs product prices; `Price` which had the price and the currency and `ProductPrice` which had the price and any promotions which were applied. 
This led to a lot of code mapping between the two and made things complicated when trying to unify the pricing code across components. I have now merged the two types into `ProductPrice` which contains all the fields and used this throughout the subs components.

This has also improved the summary component which didn't really cope with 6 for 6 properly before.


support-frontend/assets/helpers/productPrice/productPrices.js
```diff
@@ -81,7 +78,12 @@ const hasDiscount = (promotion: ?Promotion): boolean %checks =>
   promotion !== undefined &&
   isNumeric(promotion.discountedPrice);
 
-function applyDiscount(price: Price, promotion: ?Promotion) {
+const hasIntroductoryPrice = (promotion: ?Promotion): boolean %checks =>
```
lucymonie: What does the `%` do in this line? I haven't seen this before.

rupertbates: It is to tell Flow that this is a check, some details (though not many) here: https://github.com/facebook/flow/issues/4723

PR #1858, opened 2019-06-05
Add gifting to Weekly checkout

lucymonie:
## Why are you doing this?
The intention is that users can give Guardian Weekly as a gift, and specify this in the checkout.

There are a number of assumptions underpinning the changes, which are listed in the [**Trello Card**](https://trello.com/c/JwbDPHwv/2412-gifting-in-gw-checkout-frontend)

## Changes
* There's a checkbox that users can select that indicates the order is a gift
* Users who select 6 for 6 won't see the gifting option
* Users who select the gifting option won't see 6 for 6 as an alternative option further down in the checkout
* When the user checks the gifting box, additional fields appear asking for the gift recipient's details and a form for the recipient's address (aka the delivery address). There's still a section asking if billing is the same as delivery address, but the wording is slightly different to reflect that delivery is to a gift recipient.

It's worth noting that as a result of Rupert's work, the delivery cost is determined by the delivery country, but the currency is determined by the billing country.

## Work that's out of scope
There's additional business logic to be applied to the checkout around the billing period selected by the user and which other options should be visible in the checkout (ie if the user has selected Quarterly in the landing page, we don't show 6 for 6 as an option in the checkout).

## Not a gift (screenshots)

Screen Shot 2019-06-04 at 14 36 57
...
...
...
Screen Shot 2019-06-04 at 14 43 05

## Gift (screenshots)
Screen Shot 2019-06-04 at 14 41 02
Screen Shot 2019-06-04 at 14 36 09
...
...
...
Screen Shot 2019-06-05 at 10 42 46

### DIFF
--- a/support-frontend/assets/components/subscriptionCheckouts/billingPeriodSelector.jsx
+++ b/support-frontend/assets/components/subscriptionCheckouts/billingPeriodSelector.jsx
@@ -44,8 +44,8 @@ function BillingPeriodSelector(props: PropTypes) {
             billingPeriod === SixWeekly ? Quarterly : billingPeriod, // for 6 for 6 we need the quarterly pricing
             props.fulfilmentOption,
           );
-          return !props.orderIsAGift || billingPeriod !== SixWeekly ?
-            <RadioInputWithHelper
+          return props.orderIsAGift && billingPeriod === SixWeekly ? null
+            : <RadioInputWithHelper
               text={billingPeriodTitle(billingPeriod)}
               helper={getPriceDescription(
                 extendedGlyphForCountry(props.billingCountry),
@@ -56,8 +56,7 @@ function BillingPeriodSelector(props: PropTypes) {
               name="billingPeriod"
               checked={billingPeriod === props.selected}
               onChange={() => props.onChange(billingPeriod)}
-            />
-          : null;
+            />;
         })}
       </Fieldset>
     </FormSection>);
--- a/support-frontend/assets/components/subscriptionCheckouts/billingPeriodSelector.jsx
+++ b/support-frontend/assets/components/subscriptionCheckouts/billingPeriodSelector.jsx
@@ -29,6 +29,7 @@ type PropTypes = {|
   billingCountry: IsoCountry,
   selected: BillingPeriod,
   onChange: (BillingPeriod) => Action,
+  orderIsAGift?: boolean | null,
 |}
 
 function BillingPeriodSelector(props: PropTypes) {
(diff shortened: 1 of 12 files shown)


support-frontend/assets/components/subscriptionCheckouts/billingPeriodSelector.jsx
```diff
@@ -29,6 +29,7 @@ type PropTypes = {|
   billingCountry: IsoCountry,
   selected: BillingPeriod,
   onChange: (BillingPeriod) => Action,
+  orderIsAGift?: boolean | null,
```
rupertbates: Use `Option<boolean>` here for consistency

support-frontend/assets/components/subscriptionCheckouts/billingPeriodSelector.jsx
```diff
@@ -42,26 +43,28 @@ function BillingPeriodSelector(props: PropTypes) {
...
-              extendedGlyphForCountry(props.billingCountry),
-              productPrice,
-              billingPeriod,
-            )}
-            offer={getAppliedPromoDescription(billingPeriod, productPrice)}
-            name="billingPeriod"
-            checked={billingPeriod === props.selected}
-            onChange={() => props.onChange(billingPeriod)}
-          />);
+          return !props.orderIsAGift || billingPeriod !== SixWeekly ?
```
rupertbates: Would `props.orderIsAGift && billingPeriod === SixWeekly` be easier to read here?

support-frontend/assets/components/subscriptionCheckouts/billingPeriodSelector.jsx
lucymonie: Thanks, shall do

support-frontend/assets/components/subscriptionCheckouts/billingPeriodSelector.jsx
lucymonie: Yes it would. I just don't like having null as the first option! But you're right. I'll change it.

support-frontend/assets/pages/weekly-subscription-checkout/components/weeklyCheckoutForm.jsx
```diff
@@ -212,7 +252,7 @@ function WeeklyCheckoutForm(props: PropTypes) {
               </FormSection>
             : null
           }
-          <FormSection title="When would you like your subscription to start?">
+          <FormSection title={`When would you like ${props.orderIsAGift ? 'the' : 'your'} subscription to start?`}>
             <Rows>
               <FieldsetWithError id="startDate" error={firstError('startDate', props.formErrors)} legend="When would you like your subscription to start?">
```
rupertbates: Should we use the same adjusted copy as above here? In which case maybe extract it to a const first

lucymonie: 👍

support-frontend/assets/helpers/subscriptionsForms/formReducer.js
```diff
@@ -79,13 +79,13 @@ function createFormReducer(
...
-        return { ...state, firstNameGiftRecipient: action.firstNameGiftRecipient, formErrors: removeError('firstName', state.formErrors) };
+        return { ...state, firstNameGiftRecipient: action.firstNameGiftRecipient, formErrors: removeError('firstNameGiftRecipient', state.formErrors) };
 
       case 'SET_LAST_NAME_GIFT':
-        return { ...state, lastNameGiftRecipient: action.lastNameGiftRecipient, formErrors: removeError('lastName', state.formErrors) };
+        return { ...state, lastNameGiftRecipient: action.lastNameGiftRecipient, formErrors: removeError('lastNameGiftRecipient', state.formErrors) };
 
       case 'SET_EMAIL_GIFT':
-        return { ...state, emailGiftRecipient: action.emailGiftRecipient };
+        return { ...state, emailGiftRecipient: action.emailGiftRecipient, formErrors: removeError('emailGiftRecipient', state.formErrors) };
```
rupertbates: This is optional in the existing checkout, was a specific decision made to make it mandatory because optional feels more appropriate to me?

lucymonie: I'm kind of relieved that you said this - it made me vaguely uncomfortable for this to be mandatory. I checked with @jesseyuen and he agrees it should be optional.

PR #1860, opened 2019-06-05
support latency metric should be on ApplicationELB not ELB

johnduffell:
## Why are you doing this?

I shipped https://github.com/guardian/support-frontend/pull/1854 but the latency alarm wasn't showing data, because I copied it from an ELB one and forgot to change the type.  This PR fixes that!

correct one is this metric:
image

and this is the one in the alarm
image


johnduffell:
this is all good now! 
image

PR #1862, opened 2019-06-05
Get token correctly from flash session

joelochlann:
... and rename `guestAccountToken` response from Payment API to `guestAccountCreationToken` for consistency with support-frontend


tomrf1:
The final issue was that we were initialising `formData.email` wrongly.
It's working now!

PR #1867, opened 2019-06-07
use dev-nginx for setting up nginx in DEV

akash1810:
## Why are you doing this?
dev-nginx contains a number of generic useful scripts to:
- issue certs locally (and automatically trust them)
- edit hosts file
- symlink nginx configs
- restart nginx

Using dev-nginx means:
- we're creating certificates locally, valid only to a single machine
- we do not need to update certs every year as they expire (https://github.com/guardian/support-frontend/pull/1355), we just need to re-run the script when certs expire
- we do not need an S3 bucket holding certs for local dev

Depends on:
- https://github.com/guardian/dev-nginx/pull/5
- https://github.com/guardian/homebrew-devtools/pull/35
- https://github.com/guardian/dev-nginx/pull/7

We probably could move to using [`setup-app`](https://github.com/guardian/dev-nginx/blob/master/README.md#setup-app) too, removing the `support.conf` file. However it's probably best to make small changes.

[**Trello Card**](https://trello.com/c/EhggXPfo/2148-investigate-using-mkcert-for-local-certificates)

## Changes

* Change 1 - erm, see above?
* Change 2 - see change 1

## Screenshots
n/a

### DIFF
--- a/support-frontend/nginx/Brewfile
+++ b/support-frontend/nginx/Brewfile
@@ -1,0 +1,2 @@
+tap "guardian/devtools"
+brew "guardian/devtools/dev-nginx"
--- a/support-frontend/nginx/README.md
+++ b/support-frontend/nginx/README.md
@@ -1,32 +1,20 @@
 # Support™ NGINX
 
-## Setup Nginx for `Identity-Platform`
-
-Support™ depends on Identity, so **you'll need to perform the**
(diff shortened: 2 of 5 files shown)


akash1810:
> Looks good, thanks for doing this. Is there a way to use these tools on Linux?

Yep, so long as `dev-nginx/bin` is [on your PATH](https://github.com/guardian/dev-nginx#manually). Maybe we should include explicit instructions in the dev-nginx for how to install mkcert on Linux?

PR #1868, opened 2019-06-07
add reduced healthy instances alarm

johnduffell:
## Why are you doing this?

@jfsoul suggested it, and I thought it was a nice idea https://github.com/guardian/support-frontend/pull/1854#pullrequestreview-245907984

The real reason is there could be an issue preventing instances coming into service.  Instances could also be silently cycling in and out of service.  It would be good to realise this because the site goes down completely, if it's in working hours.  This alarm is unlikely to have many false alarms so it should be valuable.  It is not indicating an actual issue affecting users immediately, so it doesn't have to be as urgent as the other urgent alarms?

[**Trello Card**](https://trello.com/c/mIRQe53c/2416-add-alert-for-less-than-required-instances-of-support-frontend-for-10-min-period)

Tested in CODE: yes

### DIFF
--- a/support-frontend/cloud-formation/cfn.yaml
+++ b/support-frontend/cloud-formation/cfn.yaml
@@ -42,6 +42,7 @@ Mappings:
     PROD:
       MaxInstances: 6
       MinInstances: 3
+      InsufficientInstances: 2
       InstanceType: t3.small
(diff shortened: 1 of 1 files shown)


johnduffell:
this was automatically merged as part of https://github.com/guardian/support-frontend/pull/1871

PR #1871, opened 2019-06-07
Clarify urgent/high severity alarm names

johnduffell:
note this PR branches off https://github.com/guardian/support-frontend/pull/1868 so the first commit is from that change. feel free to review separately.

## Why are you doing this?

After last night's alarm that @david-pepper looked into, that looked more serious than it should have been taken at that time, I have updated the alarm name to be more explitit

Instead of being "HIGH severity" it is now "URGENT 9-5" which should be better.

## Screenshots

before:
image

after:
image


support-frontend/cloud-formation/cfn.yaml
```diff
@@ -278,8 +281,15 @@ Resources:
...
-      AlarmDescription: Impact - Cannot sell any subscriptions or contributions products. Follow the process in https://docs.google.com/document/d/1_3El3cly9d7u_jPgTcRjLxmdG2e919zCLvmcFCLOYAk/edit
+      AlarmName: !Join
+        - ' '
+        - - !FindInMap [ StageVariables, !Ref Stage, AlarmUrgent ]
+          - !Ref 'Stage'
+          - 'no healthy instances for support-frontend'
+      AlarmDescription: !Join
+        - ' '
+        - - 'Impact - Cannot sell any subscriptions or contributions products.'
+          - !FindInMap [ StageVariables, !Ref Stage, AlarmProcess ]
```
johnduffell: I was hoping storing the process link and header in one place would make it easier to read the cloudformation, but make your own mind up!

support-frontend/cloud-formation/cfn.yaml
```diff
@@ -296,14 +306,52 @@ Resources:
...
+        - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:reader-revenue-dev
+      AlarmName: !Join
+        - ' '
+        - - !FindInMap [ StageVariables, !Ref Stage, AlarmUrgent ]
+          - !Ref 'Stage'
+          - 'Reduced number of healthy instances for support-frontend'
+      AlarmDescription: !Join
+        - ' '
+        - - 'Impact - Imminent issue cannot sell any subscriptions or contributions products.'
+          - !FindInMap [ StageVariables, !Ref Stage, AlarmProcess ]
```
joelochlann: given that all of these alarms are conditional on `CreateProdResources` (except `StateMachineUnavailableAlarm` for some reason, should it be?), I wonder if all the substitutions of `Stage` variables could be simplified somehow? Just write "prod" where appropriate? Is it a general rule that we never have alarms on CODE resources?

johnduffell: see what you mean that's a good point.  In that case I should really put them in a separate Map that isn't realted to stages.  I actually just wanted a simple constant, and the only way to do that in cloudformation seems to be to use a Mappings 😥

johnduffell: regarding alarms on CODE, I was thinking about that too, I was wondering if we should have exactly the same alarms but just going to a dummy mailbox.  This would make it a lot easier to test my changes and also minimise the differences between CODE and PROD.  In the end I decided not to address it at this point.

johnduffell: @joelochlann I have made a Constants section, and made that other alarm PROD only as per your suggestions.

johnduffell:
All good now! Hopefully.

image

PR #1872, opened 2019-06-07
Try polyfill v3

joelochlann:
We're seeing about 6 cases per half an hour where, even after the polyfill script loads, `Object.values` and `Object.entries` are not available. Extrapolating to a day, this accounts for the number of sentry errors we were seeing for `Object.values is not a function` (approx 250 - 350 per day)

I noticed v3 is actually the latest version. So I thought I'd try that and see if the problematic cases go away.

I've renamed the events so we can clearly distinguish ones from browser using the v3 polyfill

### DIFF
--- a/support-frontend/app/views/main.scala.html
+++ b/support-frontend/app/views/main.scala.html
@@ -148,7 +148,7 @@
     </script>
     <script
       id="polyfill-script"
-      src="https://polyfill.io/v2/polyfill.min.js?features=default,Array.prototype.find,fetch,Array.prototype.includes,Number.isInteger,Object.values,Object.entries"
+      src="https://polyfill.io/v3/polyfill.min.js?features=default,Array.prototype.find,fetch,Array.prototype.includes,Number.isInteger,Object.values,Object.entries"
       onload="polyfillOnLoad()"
       onerror="polyfillOnError()"
       onabort="polyfillOnAbort()"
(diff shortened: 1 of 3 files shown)


joelochlann:
> Possible next thing to try: put another script tag with a link to some arbitrary script next to the polyfill (with a similar response size) and see if we get the same error rate: is it polyfill being flakey or is the problem our side?

the error rate from the `onerror` handler is actually very low, lower than the number of cases where the polyfill's just don't work afterwards. so I think whatever the problem is might be a bug in polyfill's JS, at least for certain browsers.

are you thinking of defining some global in an external script, and then checking whether that global is available later? to check that the script did genuinely load but also execute?

support-frontend/app/views/main.scala.html
```diff
@@ -148,7 +148,7 @@
     </script>
     <script
       id="polyfill-script"
-      src="https://polyfill.io/v2/polyfill.min.js?features=default,Array.prototype.find,fetch,Array.prototype.includes,Number.isInteger,Object.values,Object.entries"
+      src="https://polyfill.io/v3/polyfill.min.js?features=default,Array.prototype.find,fetch,Array.prototype.includes,Number.isInteger,Object.values,Object.entries"
       onload="polyfillOnLoad()"
```
gtrufitt: I wonder if, instead of polyfillOnLoad you could use the Polyfill Callback functionality `https://polyfill.io/v3/polyfill.min.js?callback=polyfillOnLoad`? A bit of a shot in the dark but might be more reliable...

joelochlann: that's a great suggestion - didn't know about that callback param!

PR #1877, opened 2019-06-10
Revert "Price summary updates"

rupertbates:
Reverts guardian/support-frontend#1857


johnduffell:
would be useful to have an explanation as to why the revert was necessary (doesn't have to be in the original desc, but afterwards edit or comment would be useful.

PR #1881, opened 2019-06-10
Revert "Revert "Price summary updates""

rupertbates:
## Why are you doing this?

Reapply #1857 with a bug fix


johnduffell:
> bug fix

what/where is it?

rupertbates:
> > bug fix
> 
> what/where is it?

@rupertbates
The description was in the commit message 'Hard coded prices are still needed on the subs landing page'. Basically I removed some static price information because I thought that it was all loaded from Zuora now but I didn't notice that the subs landing page still uses it and Eslint/Flow didn't pick it up because it was held and accessed via a map

johnduffell:
great looks good, hope this will be working. go for it! 👍 
I think you mentioned somewhere adding a test, is that done elsewhere or just decided not to/at this stage?

rupertbates:
> great looks good, hope this will be working. go for it! 👍
> I think you mentioned somewhere adding a test, is that done elsewhere or just decided not to/at this stage?

That was in #1880 merged this morning

PR #1882, opened 2019-06-11
Revert "Revert "Cache list of promotions in memory""

rupertbates:
Reverts guardian/support-frontend#1876 which I reverted while trying to fix an issue yesterday but it wasn't the problem.


johnduffell:
yes looks good, we could put it into a Map rather than Iterator just to avoid scanning through, but at least it's only in memory.  I think the Iterator might be a limitation of the subclassing/overriding anyway.

👍 go for it!

PR #1883, opened 2019-06-12
bugfix: added defensiveness around object checking in `checkAmountOrOtherAmount`

jlieb10:
## Why are you doing this?

Landing page was breaking on local and code after one-off contributions because object was undefined. I added checks to make sure the object exists before we try to pull values from it. Larger analysis required to understand what changes were made that were caused this problem and why it only seems to affect local and code.

## Changes

* Added checks for object existing
* Linted code base

### DIFF
--- a/support-frontend/assets/components/subscriptionsByCountryGroup/subscriptionsByCountryGroup.jsx
+++ b/support-frontend/assets/components/subscriptionsByCountryGroup/subscriptionsByCountryGroup.jsx
@@ -76,7 +76,7 @@ function SubscriptionsByCountryGroup(props: PropTypes) {
 
   if (countryGroupId === GBPCountries) {
     return (
-      <div id='qa-component-subscriptions-by-country-group' className={className} {...otherProps}>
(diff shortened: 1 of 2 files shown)


tomrf1:
just digging into this a bit further, the call originates here https://github.com/guardian/support-frontend/blob/master/support-frontend/assets/helpers/user/user.js#L116
The `selectedAmounts` object in the state is empty at this point.
So it occurs when the user is signed in, and the amount hasn't yet been set

tomrf1:
...and it's because of the ordering here https://github.com/guardian/support-frontend/blob/master/support-frontend/assets/pages/contributions-landing/contributionsLanding.jsx#L53
the call to `formInit` will set the `selectedAmounts`, but it happens too late

johnduffell:
this PR won't be deployed in PROD any more as we are rolling back due to a voucher issue temporarily

jlieb10:
@johnduffell no problem, will it be redeployed as part of the post fix process?

PR #1884, opened 2019-06-12
Fix handling of redirects with special chars in the queryString

tomrf1:
## Why are you doing this?
We've had a steady flow of requests where the url querystring contains some of the text from the article body.
These appear to come from web crawlers. Investigating why that's happening (something wrong on dotcom?) is a separate piece of work.

This PR addresses the fact that these requests result in a 500 response, which is causing our alarms to go off.
The reason is that support-frontend tries to return a 302 redirect when the client hits e.g. `/contribute?acquisitionData=%7B&quo.%E2%80%99`, but when Play constructs the `Location` header of the redirect response it only encodes the value parts of the querystring, not the keys.
This results in an exception when akka-http processes the headers:
`java.lang.IllegalArgumentException: Character 8217 cannot match AsciiSet because it is out of range`

This PR fixes 2 issues:
1. Always encode the keys in the querystring on redirect responses.
2. Fixes the geo redirect where there is no campaign code. This was broken because if the campaignCode is an empty string then the url has an additional slash, e.g. `/uk/contribute/?INTCMP...`, which is not valid.

## Screenshots
With an offending URL (and my bookmarks toolbar):
`https://support.thegulocal.com/contribute?INTCMP=side_menu_support_contribute&acquisitionData=%7B&quo.%E2%80%99`
Screenshot 2019-06-12 at 14 47 45

### DIFF
--- a/support-frontend/app/controllers/Application.scala
+++ b/support-frontend/app/controllers/Application.scala
@@ -49,6 +49,13 @@ class Application(
     Ok(views.html.contributionsRedirect())
   }
 
+  //Encode the querystring parameter keys, as Play only encodes the values
+  def redirectWithEncodedQueryString(url: String, queryString: Map[String, Seq[String]] = Map.empty, status: Int = SEE_OTHER): Result = Redirect(
+    url = url,
+    queryString = queryString.map { case (k,v) => java.net.URLEncoder.encode(k, "utf-8") -> v },
+    status = status
+  )
+
   def geoRedirect: Action[AnyContent] = GeoTargetedCachedAction() { implicit request =>
     val redirectUrl = request.fastlyCountry match {
       case Some(UK) => buildCanonicalShowcaseLink("uk")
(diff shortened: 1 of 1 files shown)


support-frontend/app/controllers/Application.scala
```diff
@@ -49,6 +49,13 @@ class Application(
     Ok(views.html.contributionsRedirect())
   }
 
+  //Encode the querystring parameter keys, as Play only encodes the values
+  def redirectWithEncodedQueryString(url: String, queryString: Map[String, Seq[String]] = Map.empty, status: Int = SEE_OTHER): Result = Redirect(
+    url = url,
+    queryString = queryString.map { case (k,v) => java.net.URLEncoder.encode(k, "utf-8") -> v },
```
johnduffell: have you tested that an actually encoded input e.g. %7f in the incoming url, is decoded in the "key" and then gets reencoded suitably on the way out.  And that it doesn't get double encoded.
This seems to document the "incoming" side, which suggests the decoding is done as per RFC3986:
https://doc.akka.io/docs/akka-http/current/common/uri-model.html
From reading those, I'd guess yes, but I haven't got that far yet.

tomrf1: Yes, the screenshot does kind of show this, but we can confirm with:
`curl "https://support.thegulocal.com/contribute?INTCMP=side_menu_support_contribute&acquisitionData=%7B&quo.%E2%80%99" -v`
which returns:
```
< HTTP/1.1 302 Found
...
< location: /uk/contribute?INTCMP=side_menu_support_contribute&acquisitionData=%7B&quo.%E2%80%99=
...
```
The additional `=` at the end of the `location` header is because it's interpreted as a querystring parameter with an empty value

johnduffell: I see, so the akka/play query string parser assumes the query string is encoded using https://url.spec.whatwg.org/#urlencoded-serializing
This means it interprets the & as the start of a new key and the lack of an `=` as an empty value.  I'm not sure there's much we can do about it, assuming we're going to build the query string using the same mechanism.  But hopefully we wouldn't rely on that not happening.

johnduffell:
yes this is spot on, thanks and it looks like java's over generally named "URLEncoder" happens to be exactly what we need for query key/value encoding. 👍

johnduffell:
it was out at 11:23 - https://riffraff.gutools.co.uk/deployment/view/913055b4-1573-4a4f-a970-4fe2b5e11565 - but  prout didn't spot it until 12:51

PR #1886, opened 2019-06-12
add product option into the guardian delivery price lookup

johnduffell:
## Why are you doing this?

DD voucher checkout is breaking with an error 
```
Uncaught TypeError: Cannot read property 'Monthly' of undefined
    at m (productPrices.js:110)
    at E (submit.js:132)
    at formActions.js:98
```
This is because this param doesnt go thrugh to getProductPrices, it should be the last param
image

so when it arrives it is undefined at the moment, causing an empty results
image

The structure of the productPrices object is like this:
image

### DIFF
--- a/support-frontend/assets/helpers/subscriptionsForms/submit.js
+++ b/support-frontend/assets/helpers/subscriptionsForms/submit.js
@@ -134,6 +134,7 @@ function buildRegularPaymentRequest(
     state.page.billingAddress.fields.country,
     billingPeriod,
     fulfilmentOption,
+    productOption,
   );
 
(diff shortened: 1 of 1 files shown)


johnduffell:
we have done an assessment and can't find any reason this would break from a logical point of view by looking at the code.
Since @joelochlann has a fix to GA that's possibly costing us a lot of money that is already on master, that encourages us to want to get a clean master.
This has been tested in CODE as working.
We will test in PROD briefly and then revisit tomorrow morning.
This would be caught quicker by a "no DD for 24 hours" monitoring test

PR #1888, opened 2019-06-13
update sentry client side lib to the new one

johnduffell:
## Why are you doing this?

The [old official Raven library](https://docs.sentry.io/clients/javascript/) is deprecated, the new one is called [Sentry](https://docs.sentry.io/error-reporting/quickstart/?platform=browser).

[**Trello Card**](https://trello.com/c/ckxSy7oV/2427-update-raven-to-sentry-library)

## Screenshots

I got it working locally
image

### DIFF
--- a/support-frontend/assets/helpers/logger.js
+++ b/support-frontend/assets/helpers/logger.js
@@ -1,24 +1,25 @@
 // @flow
 // import Raven from 'raven-js';
 
-const EventualRaven = import('raven-js');
+const EventualSentry = import('@sentry/browser');
 
 // ----- Functions ----- //
 
-const init = () => EventualRaven.then((Raven) => {
+const init = () => EventualSentry.then((Sentry) => {
   const dsn: string = 'https://65f7514888b6407881f34a6cf1320d06@sentry.io/1213654';
   const { gitCommitId } = window.guardian;
 
-  Raven.default.config(dsn, {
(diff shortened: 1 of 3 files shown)


support-frontend/assets/helpers/logger.js
```diff
@@ -1,39 +1,40 @@
 // @flow
 // import Raven from 'raven-js';
```
jlieb10:
```suggestion
// import Sentry
```

joelochlann: can we delete this commented import?

support-frontend/assets/helpers/logger.js
```diff
@@ -1,39 +1,40 @@
...
+    Sentry.captureException(
       new Error(ex),
       {
         extra: context,
       },
     );
 
     if (window.console && console.error) {
-      console.error(ex);
+      console.error('sentry', ex);
```
jlieb10:
```suggestion
      console.error('sentry exception: ', ex);
```

support-frontend/assets/helpers/logger.js
johnduffell: thanks yes maybe that was left over from when I made sentry load async.
Although having thought about that particular change a bit more, I'm not sure whether loading it async would collect errors that happen early in the page load!

PR #1890, opened 2019-06-13
More lenient geolocation detection

joelochlann:
Viewing the contributions landing page through Google Translate doesn't work because the element that it renders the react app into contains a country code, and this country code detection doesn't work because the URL is in different (it's embedded as a parameter inside a Google Translate URL).

So
1. Don't put the country code in the id. This id isn't used by anything, there's no reason for it to be there
2. Look at the whole URL for the country code string anywhere, rather than the stricter check we currently have (it has to be the entire path, or the start of the path)
3. Fix the unhelpful error message `Fatal error trying to render a page`

### DIFF
--- a/support-frontend/app/controllers/Application.scala
+++ b/support-frontend/app/controllers/Application.scala
@@ -131,7 +131,7 @@ class Application(
 
     views.html.contributions(
       title = "Support the Guardian | Make a Contribution",
-      id = s"contributions-landing-page-$countryCode",
+      id = s"contributions-landing-page",
       js = js,
       css = css,
       fontLoaderBundle = fontLoaderBundle,
--- a/support-frontend/assets/helpers/internationalisation/countryGroup.js
+++ b/support-frontend/assets/helpers/internationalisation/countryGroup.js
@@ -103,20 +103,21 @@ const countryGroups: CountryGroups = {
 
 // ----- Functions ----- //
 
-function fromPath(path: string = window.location.pathname): ?CountryGroupId {
-  if (path === '/uk' || path.startsWith('/uk/')) {
+function fromUrl(): ?CountryGroupId {
+  const url = window.location.href;
+  if (url.includes('/uk')) {
     return GBPCountries;
-  } else if (path === '/us' || path.startsWith('/us/')) {
+  } else if (url.includes('/us')) {
     return UnitedStates;
-  } else if (path === '/au' || path.startsWith('/au/')) {
+  } else if (url.includes('/au')) {
     return AUDCountries;
-  } else if (path === '/eu' || path.startsWith('/eu/')) {
+  } else if (url.includes('/eu')) {
     return EURCountries;
-  } else if (path === '/int' || path.startsWith('/int/')) {
+  } else if (url.includes('/int')) {
     return International;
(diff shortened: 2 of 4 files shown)


support-frontend/assets/helpers/internationalisation/countryGroup.js
```diff
@@ -103,20 +103,21 @@ const countryGroups: CountryGroups = {
 
 // ----- Functions ----- //
 
-function fromPath(path: string = window.location.pathname): ?CountryGroupId {
-  if (path === '/uk' || path.startsWith('/uk/')) {
+function fromUrl(): ?CountryGroupId {
```
jlieb10: Not a major suggestion but we could make it more reusable by turning the the url a default parameter

```suggestion:
function fromUrl(url: string = window.location.href): ?CountryGroupId {
```

support-frontend/assets/helpers/internationalisation/countryGroup.js
```diff
@@ -103,20 +103,21 @@ const countryGroups: CountryGroups = {
...
+  const url = window.location.href;
```
jlieb10:
```suggestion
```

support-frontend/assets/helpers/render.js
```diff
@@ -46,7 +46,7 @@ const renderPage = (content: Object, id: string, callBack?: () => void) => {
       renderError(e, id);
     }
   } else {
-    logException(`Fatal error trying to render a page. id:${id}`);
+    logException(`Could not find element with id ${id} to render react app into`);
```
jlieb10: 👍 MUCH better error

support-frontend/app/controllers/Application.scala
```diff
@@ -131,7 +131,7 @@ class Application(
 
     views.html.contributions(
       title = "Support the Guardian | Make a Contribution",
-      id = s"contributions-landing-page-$countryCode",
```
jlieb10: I assume this isn't being used anywhere in the scss because it's an ID, but wanted to raise it as something to double-check

joelochlann: I've searched everywhere for `contributions-landing-page`, the only place it's used is to set the id here and in the JS. Maybe we should just have a `react-root` id everywhere and not bother parameterising it

support-frontend/assets/helpers/internationalisation/countryGroup.js
joelochlann: Yep that's how it was before, but the only place it's use is in `detect` and all the other functions called in `detect` accept no arguments and depend just on browser state. So for me this one accepting a parameter is confusing. I can't think of a use case where you would want to pass in a URL other than the page the browser is on to figure out the geolocation

joelochlann:
I've realised this change doesn't actually make it work inside the Google Translate iframe though. it renders into the root element, but just empties it out leaving a blank page. No errors in the console to help debug this time.

I see the same behaviour if I put airbnb.com into Google Translate.

I suspect trying to get a react app (especially one with third-party payment libs, iframes etc) trying to work inside the google translate iframe might be a rabbit hole so parking this for now

PR #1891, opened 2019-06-14
6 for 6 Guardian Weekly promotions

rupertbates:
## Why are you doing this?
We want to be able to create 6 for 6 subscriptions to Guardian Weekly

I have implemented 6 for 6 slightly differently to the way that it was done in the old site, it is now a promotion of the type `IntroductoryPrice` which has a price and a duration (£6 & 6 weeks). This will allow us to do 'n for n' promotions in the future but there will still be a bit of work needed to support that.

The website should now pass in `Quarterly` as the billing period, which is the subscription that the user will move onto after the introductory period, rather than `SixWeekly` as in the old site.

I've also had to add `productRatePlanChargeIds` for the GW `SixWeekly` billing period as this is used in setting up the introductory pricing period.

I have tested this on code against the old site and the subscriptions it creates are identical as far as I can see.

[**Trello Card**](https://trello.com/c/NQV3JqAa/2338-guardian-weekly-6-for-6-backend)

### DIFF
--- a/support-models/src/main/scala/com/gu/support/zuora/api/SubscriptionData.scala
+++ b/support-models/src/main/scala/com/gu/support/zuora/api/SubscriptionData.scala
@@ -2,15 +2,14 @@ package com.gu.support.zuora.api
 
 import cats.syntax.functor._
 import com.gu.support.catalog.{ProductRatePlanChargeId, ProductRatePlanId}
-import com.gu.support.encoding.CustomCodecs._
-import com.gu.support.encoding.Codec._
 import com.gu.support.encoding.Codec
+import com.gu.support.encoding.Codec._
+import com.gu.support.encoding.CustomCodecs.{decodeDateTime, encodeDateTime, monthDecoder, _}
+import com.gu.support.encoding.JsonHelpers._
+import com.gu.support.promotions.PromoCode
 import io.circe.syntax._
 import io.circe.{Decoder, Encoder, Json}
 import org.joda.time.{LocalDate, Months}
-import com.gu.support.encoding.CustomCodecs.monthDecoder
-import com.gu.support.encoding.JsonHelpers._
-import com.gu.support.promotions.PromoCode
 
 object RatePlanCharge {
   val fixedPeriod = "FixedPeriod"
@@ -37,15 +36,20 @@ object RatePlanCharge {
       .add(endDateCondition, Json.fromString(subscriptionEnd)))
   implicit val contributionDecoder: Decoder[ContributionRatePlanCharge] = decapitalizingDecoder
 
+  implicit val introductoryPriceEncoder: Encoder[IntroductoryPriceRatePlanCharge] = capitalizingEncoder
+  implicit val introductoryPriceDecoder: Decoder[IntroductoryPriceRatePlanCharge] = decapitalizingDecoder
+
   implicit val encodeRatePlanCharge: Encoder[RatePlanCharge] = Encoder.instance {
     case f: DiscountRatePlanCharge => f.asJson
     case s: ContributionRatePlanCharge => s.asJson
+    case i: IntroductoryPriceRatePlanCharge => i.asJson
   }
 
   implicit val decodeRatePlanCharge: Decoder[RatePlanCharge] =
     List[Decoder[RatePlanCharge]](
       Decoder[DiscountRatePlanCharge].widen,
-      Decoder[ContributionRatePlanCharge].widen
+      Decoder[ContributionRatePlanCharge].widen,
+      Decoder[IntroductoryPriceRatePlanCharge].widen,
     ).reduceLeft(_ or _)
 }
 
@@ -62,6 +66,12 @@ case class DiscountRatePlanCharge(
 case class ContributionRatePlanCharge(
   productRatePlanChargeId: ProductRatePlanChargeId,
   price: BigDecimal
+) extends RatePlanCharge
+
+case class IntroductoryPriceRatePlanCharge(
+  productRatePlanChargeId: ProductRatePlanChargeId,
+  price: BigDecimal,
+  triggerDate: LocalDate,
 ) extends RatePlanCharge
 
 sealed trait PeriodType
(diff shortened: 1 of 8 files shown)


support-models/src/main/scala/com/gu/support/zuora/api/SubscriptionData.scala
```diff
@@ -37,15 +38,22 @@ object RatePlanCharge {
       .add(endDateCondition, Json.fromString(subscriptionEnd)))
   implicit val contributionDecoder: Decoder[ContributionRatePlanCharge] = decapitalizingDecoder
 
+  implicit val introductoryPriceEncoder: Encoder[IntroductoryPriceRatePlanCharge] = capitalizingEncoder[IntroductoryPriceRatePlanCharge]
+    .mapJsonObject(_
```
johnduffell: would it be clearer if we change this to `{ .. }` instead of `( ... )` and give the param a name e.g. 
```suggestion
    .mapJsonObject { somethingName =>
      somethingName.add(triggerEvent, Json.fromString(specificEvent))
    }
```

support-services/src/main/scala/com/gu/support/promotions/PromotionApplicator.scala
```diff
@@ -55,6 +58,56 @@ class DiscountApplicator(discount: DiscountBenefit, config: PromotionsDiscountCo
...
+    // To genericise this for N for N or even to work with other products, we would need to have a separate product rate plan
+    // for the introductory period which can apply to all products (in the way that discounts work currently) and then change
+    // the fulfilment lambda to pick this up.
+
+    val introductoryPriceStartDate = subscriptionData.subscription.contractAcceptanceDate
+
+    for {
+      recurringRatePlan <- subscriptionData.ratePlanData.headOption
+      recurringProductRatePlanId = recurringRatePlan.ratePlan.productRatePlanId
+      productRatePlansForEnvironment <- GuardianWeekly.ratePlans.find(_._2.exists(_.id == recurringProductRatePlanId)).map(_._2)
```
johnduffell: can we map to _2 first?  even better if we can name it rather than refer to it by position
```suggestion
      somethingSomething = GuardianWeekly.ratePlans.map(_._2)
      productRatePlansForEnvironment <- somethingSomething.find(_.exists(_.id == recurringProductRatePlanId))
```

support-services/src/main/scala/com/gu/support/promotions/PromotionApplicator.scala
```diff
@@ -55,6 +58,56 @@ class DiscountApplicator(discount: DiscountBenefit, config: PromotionsDiscountCo
...
 
+class IntroductoryPriceApplicator(introductoryPriceBenefit: IntroductoryPriceBenefit) extends BenefitApplicator {
+  def applyTo(subscriptionData: SubscriptionData): SubscriptionData = {
+    val subscription = subscriptionData.subscription
+    val result = introductoryPriceRatePlanData(subscriptionData).map(
+      ratePlanData =>
+        subscriptionData.copy(
+          ratePlanData = subscriptionData.ratePlanData.::(ratePlanData),
+          subscription = subscription.copy(
+            contractAcceptanceDate = subscription.contractAcceptanceDate.plusWeeks(introductoryPriceBenefit.periodLength)
```
johnduffell: so if `introductoryPrice` exists on the promo code,we try to apply the applicator.  We add the period length to the start date (which would hopefully always be 6 at the moment, and we override the price in the charge from the promo.
If there is any error during the process the promotion would be silently not applied.  Do we have any safeguard to just prevent the subscription being created if they have an introductary period but it came back as None?

johnduffell:
good, I mentioned a few small clarity things, the only concern is the fact that it would probably silently not apply the promotion if there is any issue with the find or headOption command.  I think it's an error case if the promo has a introduction and we couldn't apply it properly, and we should refuse to subscribe, rather than just creating a default quarterly.  But perhaps this is only a very short term mechanism?
Other than that I think it's good work.  Just make a decision on the Option thing first 👍

support-services/src/main/scala/com/gu/support/promotions/PromotionApplicator.scala
rupertbates: This is not the same because we are searching a structure `Map[TouchPointEnvironment, List[ProductRatePlan]]` and want to find a `TouchPointEnvironment -> List[ProductRatePlan]` so that we can search only the product rate plans for the current environment. 
If we were to map first this is the same as just doing `ratePlans.values` which will flatten all the environments into one list.

support-models/src/main/scala/com/gu/support/zuora/api/SubscriptionData.scala
rupertbates: could do, it would just be `mapJsonObject(jsonObject => jsonObject.add...` which doesn't really seem worth it to me as the method name already tells you what the type is.

support-services/src/main/scala/com/gu/support/promotions/PromotionApplicator.scala
rupertbates: No we don't, this is a good point, we should put some thought into what we do in the general case here although it _should_ of course never happen!
I think this should be a different PR though

support-services/src/main/scala/com/gu/support/promotions/PromotionApplicator.scala
johnduffell: I forgot map on a Map is non law abiding in terms of restructuring a Map into a List of pairs.
However that's shocking that it would then notice if the value type is a List and then flatten them all together!

support-services/src/main/scala/com/gu/support/promotions/PromotionApplicator.scala
johnduffell: yeah separate PR should be fine.   My question is really "how will we know if it's broken?".  Not from the type checker and not from the error logs, maybe  from the tests, and eventually from the users.

PR #1892, opened 2019-06-14
Post deploy tests for subscriptions checkouts

rupertbates:
## Why are you doing this?

We want to catch any issues which prevent subscribers from purchasing our products as quickly as possible, one way of doing this is to add [Selenium](https://www.seleniumhq.org/) post deployment tests which run though checkout flows to ensure they can be completed successfully.

This PR adds tests for credit card checkout on Paper and Digital Pack

[**Trello Card**](https://trello.com/c/tOADTOUP/2425-add-post-deploy-tests-for-the-subs-checkouts)

[**Post deploy test docs**](https://github.com/guardian/support-frontend/wiki/Post-Deployment-Testing)


support-frontend/test/selenium/subscriptions/CheckoutsSpec.scala
```diff
@@ -0,0 +1,86 @@
...
+    register.next()
+
+    Then("they should be taken to the second register page")
+    assert(register.secondPageHasLoaded)
+
+    Given("that the user fills in the rest of their details")
+    register.fillInPersonalDetails()
+
+    When("they click the create account button")
+    register.createAccount()
```
jfsoul: Do we want to bypass this - I'm just wondering whether this is actually something we want to test or just that there's no way around it.

support-frontend/test/selenium/subscriptions/pages/PaperCheckout.scala
```diff
@@ -0,0 +1,30 @@
...
+  private val billingAddressIsDifferent = id("qa-billing-address-different")
+
+  def fillForm {
+    setValue(deliveryLineOne, "Kings Place")
+    setValue(deliveryCity, "London")
+    setValue(deliveryPostcode, "N19GU")
+    clickOn(billingAddressIsDifferent)
+    setValue(billingLineOne, "My house")
+    setValue(billingCity, "Bristol")
+    setValue(billingPostcode, "BS68QT")
```
jfsoul: Just checking this is not a real postcode!

rupertbates: Nope! 😄

support-frontend/test/selenium/subscriptions/CheckoutsSpec.scala
rupertbates: There's no way around it really because we need an identity user for the checkout

support-frontend/test/selenium/subscriptions/pages/PaperCheckout.scala
lucymonie: Isn't it Rupert's postcode?

PR #1893, opened 2019-06-14
Unused components purge

jranks123:
## Why are you doing this?

This PR deletes a whole lotta components 

YAY

@guardian/contributions 

@jlieb10 To find out which ones weren't being used, I simply went through each component and searched for it in in the codebase to see if it was being used. 

I did try using https://github.com/CVarisco/delete-react-zombies, which claims to be able to identify unused components for you, but I couldn't get it to work


jlieb10:
This is awesome @jranks123 ! Could you document here how you found the unused components?

PR #1894, opened 2019-06-17
Add contribution type to set password tracking

jranks123:
## Why are you doing this?
So we can see the split between one-off and recurring.


support-frontend/assets/pages/contributions-landing/components/SetPasswordForm.jsx
```diff
@@ -78,7 +81,7 @@ function onSubmit(props: PropTypes): Event => void {
   return (event) => {
     props.setPasswordHasBeenSubmitted();
     event.preventDefault();
-    trackComponentClick('set-password');
+    trackComponentClick(`set-password-${props.contributionType}`);
```
jlieb10: Is it helpful to separate monthly and annual? I guess this can all be joined up again as 'recurring' during data analysis

PR #1897, opened 2019-06-17
Use Parameter Store for config

tomrf1:
## Why are you doing this?
This app currently uses S3 for its secret configuration. It is now recommended ([according to a doc I wrote](https://github.com/guardian/recommendations/blob/master/config.md)) that we use Parameter Store instead.

#### Migrating to Parameter Store - general steps:
1. Use the [parameter-store-migration](https://github.com/guardian/parameter-store-migration) tool to migrate from the S3 file to PS, for each environment.
2. Update the app to use [simple-configuration](https://github.com/guardian/simple-configuration) to load config from Parameter Store. This is done in the Play app's `AppLoader.scala`.
3. Update the cloudformation file to stop downloading the private config file from S3, and grant necessary SSM + ASG permissions.

#### Current config situation:
- `/support-frontend/conf/application.conf` holds public, non-stage-specific config.
- `/support-frontend/conf/[STAGE].public.conf` holds public, stage-specific config.
- `/etc/gu/support-frontend.private.conf` holds private, stage-specific config, and is downloaded from s3 on startup (not in the repo).

#### New config situation
- continue to use `/support-frontend/conf/application.conf`
- continue to use `/support-frontend/conf/[STAGE].public.conf`
- private config is now stored in Parameter Store and fetched by the Play app on startup. It is combined with the public config that is loaded from file

The public config in `[STAGE].public.conf` could also be moved into Parameter Store - I don't have strong feelings one way or the other so I've left it as it is.

I have migrated the DEV private config to PS, but in DEV mode it will also look for `/etc/gu/support-frontend.private.conf`, to allow us to override values when running locally.


rupertbates:
A couple of questions: 
* Is the `private.conf.template` file actually used for anything or is it purely documentation? And if purely doc is there anything stopping it going out of date? If there isn't then we should probably get rid of it.
* This is just for support-frontend right? Support-workers is unaffected?

tomrf1:
@rupertbates I just chatted to @jfsoul about the private.conf.template file and we also concluded it's a bad idea, I'll remove it.

This is just for support-frontend, correct

support-frontend/app/wiring/AppLoader.scala
```diff
@@ -1,19 +1,46 @@
...
+    val identity = AppIdentity.whoAmI(defaultAppName = "support-frontend")
+    val loadedConfig = ConfigurationLoader.load(identity) {
+      case AwsIdentity(app, stack, stage, _) =>
+        val privateConfig = SSMConfigurationLocation(s"/$stack/$app/$stage")
+        val publicConfig = ResourceConfigurationLocation(s"$stage.public.conf")
+
+        ComposedConfigurationLocation(List(privateConfig, publicConfig))
+
+      case DevIdentity(_) =>
+        //If a local private config file exists then override any DEV Parameter Store config
```
rupertbates: 🥇

PR #1899, opened 2019-06-18
Add a post deployment test for Direct Debit

rupertbates:
## Why are you doing this?

We want to have a post deployment test which checks that Direct Debit is still working correctly so that we can tell immediately if we break it.

This PR adds Direct Debit as a payment method to the subs checkouts Selenium tests and uses it on the paper checkout. This means that we are now testing
* UK Digital Pack checkout with Stripe
* Paper checkout with Direct Debit

[**Trello Card**](https://trello.com/c/tOADTOUP/2425-add-post-deploy-tests-for-the-subs-checkouts)

### DIFF
--- a/support-frontend/assets/pages/paper-subscription-checkout/components/checkoutForm.jsx
+++ b/support-frontend/assets/pages/paper-subscription-checkout/components/checkoutForm.jsx
@@ -25,9 +25,6 @@ import { asControlled } from 'hocs/asControlled';
 import Form, { FormSection } from 'components/checkoutForm/checkoutForm';
 import Layout, { Content } from 'components/subscriptionCheckouts/layout';
 import Summary from 'components/subscriptionCheckouts/summary';
-import DirectDebitPopUpForm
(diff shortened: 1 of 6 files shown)


support-frontend/assets/pages/paper-subscription-checkout/components/checkoutForm.jsx
```diff
@@ -258,12 +255,6 @@ function CheckoutForm(props: PropTypes) {
             >
               Continue to payment
             </Button>
-            <DirectDebitPopUpForm
```
tomrf1: is this being used currently?

rupertbates: No it was a duplicate of this:
https://github.com/guardian/support-frontend/blob/909e45760fee9c6a30104763e24b88330d95d752/support-frontend/assets/components/subscriptionCheckouts/paymentMethodSelector.jsx#L84

PR #1900, opened 2019-06-18
Reset password endpoint & cohort 4 client side work

jlieb10:
## Why are you doing this?

We need to send users through Identity's 'reset password' flow (which recognizes whether or not a user has set a password already and changes its email copy accordingly. This PR also includes the necessary clientside work to integrate this endpoint which will be ready for when we start to recognize cohort4 users and send them to the 'set password' template.

[**Trello Card**](https://trello.com/c/LBoYGOds/1178-implement-password-reset-endpoint-on-support-frontend)

## Changes

* Added endpoint to support-frontend (paired with @jranks123 )
* Integrated the endpoint on the client

## Screenshots
'Thank you set password':
reset password screen
'Thank you password set': 
reset password final screen

### DIFF
--- a/support-frontend/assets/pages/contributions-landing/components/SetPasswordForm.jsx
+++ b/support-frontend/assets/pages/contributions-landing/components/SetPasswordForm.jsx
@@ -220,8 +220,8 @@ function SetPasswordForm(props: PropTypes) {
   return (
     <div className="set-password__form">
       {props.guestAccountCreationToken ?
-      renderNewPasswordForm(props) :
-      renderPasswordResetForm(props)}
+       renderNewPasswordForm(props) :
+       renderPasswordResetForm(props)}
     </div>
   );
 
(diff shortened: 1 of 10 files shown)


support-frontend/assets/pages/contributions-landing/components/SetPasswordForm.jsx
```diff
@@ -160,6 +185,43 @@ function SetPasswordForm(props: PropTypes) {
...
+          No thank you
+        </Button>
+      </form>
+    </div>);
+}
+
+function SetPasswordForm(props: PropTypes) {
+  return (
+    <div className="set-password__form">
+      {props.guestAccountCreationToken ?
```
joelochlann: ah OK, so this gives everyone who is **not** brand new and not signed in (i.e. people with full accounts with passwords, full accounts with social sign-in, full accounts which are validated, full accounts unvalidated, existing guest accounts) the password reset form?

jlieb10: This would rely on a backend implementation where we could send all "guests" to the "set password" page and the just-created guests would have a token, and the not-just-created guests would get the reset button. We wouldn't want to send all our traffic to this page, though. Just accurately identified "guests."

PR #1903, opened 2019-06-18
Pass CredentialsProvider to simple configuration

joelochlann:
Otherwise it can't find credentials when running locally


joelochlann:
@tomrf1 do you have a `[default]` section in your `~/.aws/credentials`?

johnduffell:
there seems to be a problem with prout, this overdue message only appeared at the moment when I merged #1888 and mine is defineitly out in prod but prout hasn't noticed the updated git commit id, (or marked it as overdue)

PR #1905, opened 2019-06-19
Add dev-nginx installation to support-frontned/setup.sh

jranks123:
## Why are you doing this?
Adds the installation of `[dev-ngnix](https://github.com/guardian/dev-nginx)` to `./setup.sh`. 

This is required by `nginx/setup.sh`, which is now called from the main setup script

### DIFF
--- a/support-frontend/setup.sh
+++ b/support-frontend/setup.sh
@@ -106,6 +106,8 @@ install_nginx() {
     EXTRA_STEPS+=("nginx is installed. Ensure you have 'include sites-enabled/*' in your nginx configuration ${NGINX_ROOT}/nginx.conf and add '127.0.0.1 support.thegulocal.com' and '127.0.0.1 support-ui.thegulocal.com' to /etc/hosts")
   fi
 
+  install_dev_nginx 
+
   ./nginx/setup.sh
 }
 
@@ -114,6 +116,14 @@ install_awscli() {
     brew install awscli
   fi
 }
+
+install_dev_nginx() {
+  if ! installed dev-nginx; then
+    brew tap "guardian/devtools"
+    brew install "guardian/devtools/dev-nginx"
+  fi
+}
+
 
 install_sbt() {
   if ! installed sbt; then


support-frontend/setup.sh
```diff
@@ -115,6 +117,14 @@ install_awscli() {
   fi
 }
 
+install_dev_nginx() {
+  if ! installed dev-nginx; then
+    brew tap "guardian/devtools"
```
akash1810: I think a `Brewfile` could make the setup simpler. A `Brewfile` is essentially a package.json for brew.

For example, we added it to one of our Tools recently - https://github.com/guardian/login.gutools/pull/44/files.

The simpler setup then reads something like `brew bundle` (to install deps) `yarn install && yarn build` `sbt run`.

akash1810: This change is probably too much for this PR though.

support-frontend/setup.sh
```diff
@@ -106,6 +106,8 @@ install_nginx() {
     EXTRA_STEPS+=("nginx is installed. Ensure you have 'include sites-enabled/*' in your nginx configuration ${NGINX_ROOT}/nginx.conf and add '127.0.0.1 support.thegulocal.com' and '127.0.0.1 support-ui.thegulocal.com' to /etc/hosts")
   fi
 
+  install_dev_nginx
```
akash1810: Should this go [here](https://github.com/guardian/support-frontend/pull/1905/files#diff-102794ef626ea07496b9fed8cf96887aL103) instead?

jranks123: It isn't needed until nginx/setup.sh is ran

jlieb10:
Just ran into the problem of trying to run `./setup.sh` without `dev-nginx` installed. Fortunately the error returned is helpful, but this solution would be even more helpful. I think moving to a brewfile would be a cleaner solution but doesn't need to happen urgently, and this change is in line with what we currently have.

joelochlann:
> Just ran into the problem of trying to run ./setup.sh without dev-nginx installed. Fortunately the error returned is helpful, but this solution would be even more helpful. I think moving to a brewfile would be a cleaner solution but doesn't need to happen urgently, and this change is in line with what we currently have.

PR here: https://github.com/guardian/support-frontend/pull/1916

joelochlann:
superseded by #1916

PR #1907, opened 2019-06-20
Increase fulfilment delay for voucher subs

jfsoul:
## Why are you doing this?
There are a couple of issues with our fulfilment delay calculation for vouchers:
- We are only applying a delay of 2-3 weeks rather than 3-4 weeks, which puts us out of step with subscriptions-frontend and unable to fulfil some subscriptions.
- We are currently calculating lead time based on whether the current day is after a Weds (correct) *AND* whether the current day is after the delivery day of the product (incorrect).  For example, an Everyday sub taken out on a Monday or Tuesday should have the same first delivery day - but at the moment the Tuesday one is pushed a week because Tuesday falls after the delivery day (Monday) of the product.

The latter issue above is resolved by now using a different rule:
- Find the date of the delivery day of the product in the _current_ week
- Add 3 or 4 weeks depending on whether the current time is after Weds 6am (voucher fulfilment cutoff).

This rule appears to also work for Guardian Weekly (except the cutoff is different) - however it will not work for home delivery if we were to re-enable it.  That should be done in a later PR once we have fixed the live issue affecting fulfilment.  In the meantime I have skipped HD tests.

Follow up work needed:
* Document fulfilment delay rules heavily because they are a minefield
* Implement a single rule that works for voucher, weekly and home delivery

[**Trello Card**](https://trello.com/c/eWa5ZctU/2370-increase-wait-time-calculation-in-print-checkout)

## Changes

* Increases delay of voucher fulfilment to 3-4 weeks from 2-3 weeks (roughly)
* Makes fulfilment delay only dependent on current day relative to cutoff (specifically not on the current day relative to delivery day)
* Adds a few more tests for good measure

### DIFF
--- a/support-frontend/assets/helpers/subscriptionsForms/deliveryDays.js
+++ b/support-frontend/assets/helpers/subscriptionsForms/deliveryDays.js
@@ -5,13 +5,13 @@ const milsInADay = 1000 * 60 * 60 * 24;
 
 export const numberOfWeeksWeDeliverTo = 4;
 
-const getNextDayOfTheWeek = (today: number, day: Day): Date => {
-  const diff = (7 + (day - new Date(today).getDay())) % 7;
+const getDateOfDeliveryDayInCurrentWeek = (today: number, day: Day): Date => {
+  const diff = (day - new Date(today).getDay()) % 7;
   return new Date(today + (diff * milsInADay));
 };
 
-const getNextDaysOfTheWeek = (today: number, day: Day, length: number = numberOfWeeksWeDeliverTo): Date[] => {
-  const initial = getNextDayOfTheWeek(today, day);
+const getDeliveryDays = (today: number, day: Day, length: number = numberOfWeeksWeDeliverTo): Date[] => {
+  const initial = getDateOfDeliveryDayInCurrentWeek(today, day);
   const rt = [initial];
   for (let i = 1; i <= length; i += 1) {
     rt.push(new Date(rt[i - 1].getTime() + (7 * milsInADay)));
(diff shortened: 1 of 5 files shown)


jfsoul:
@rupertbates Spotted a problem with Sunday paper: made a small change to adjust and added another test case to cover.

johnduffell:
Patricio has defined these rules in the endpoint that salesforce uses when CSRs are adding voucher subectiptions for people who call or email in!  Ideally support-frontend would share that rules.
https://github.com/guardian/support-service-lambdas/blob/ebc4425774d9f864fe8b8ebbb521f403a8e8fd4a/handlers/new-product-api/src/main/scala/com/gu/newproduct/api/productcatalog/NewProductApi.scala#L34

johnduffell:
This test has the full JSON example.  Also covers home delivery
https://github.com/guardian/support-service-lambdas/blob/master/handlers/new-product-api/src/test/scala/com/gu/newproduct/api/productcatalog/CatalogWireTest.scala

johnduffell:
the server side valudation uses the same model, the code is here: https://github.com/guardian/support-service-lambdas/blob/ebc4425774d9f864fe8b8ebbb521f403a8e8fd4a/handlers/new-product-api/src/main/scala/com/gu/newproduct/api/addsubscription/validation/DateValidation.scala

jfsoul:
@johnduffell I couldn't immediately see where those rules are interpreted so I am kind of guessing based on naming i.e. `startDaysAfterCutoff` whether the end result is definitely the same for us.  Otherwise I think sharing is the way forward on something like this and the modelling is neater that we have at the moment.

PR #1908, opened 2019-06-20
Generic sign in on thank you page

jlieb10:
## Why are you doing this?

We want as many people to sign in as possible. This is a tracked link with copy to go to profile.theguardian.com with email prefilled that renders for people who are not signed in. Hopefully this will up our number of people we can recognize on site after they've contributed.

[**Trello Card**](https://trello.com/c/dxrQCGzW/1186-thank-you-page-for-cohorts-34-sign-in)

## Screenshots
generic sign in cta

### DIFF
--- a/support-frontend/assets/pages/contributions-landing/components/ContributionThankYou/ContributionThankYou.jsx
+++ b/support-frontend/assets/pages/contributions-landing/components/ContributionThankYou/ContributionThankYou.jsx
@@ -11,10 +11,12 @@ import { type Action, setHasSeenDirectDebitThankYouCopy } from '../../contributi
 import type { PaymentMethod } from 'helpers/paymentMethods';
 import { ContributionThankYouBlurb } from './ContributionThankYouBlurb';
 import AnchorButton from 'components/button/anchorButton';
+import Button from 'components/button/button';
 import SvgArrowLeft from 'components/svgs/arrowLeftStraight';
 import { DirectDebit } from 'helpers/paymentMethods';
 import SpreadTheWord from 'components/spreadTheWord/spreadTheWord';
 import ContributionSurvey from '../ContributionSurvey/ContributionsSurvey';
+import { trackComponentClick } from 'helpers/tracking/ophan';
 
 // ----- Types ----- //
 
@@ -61,8 +63,8 @@ function ContributionThankYou(props: PropTypes) {
     <div className="thank-you__container">
       <div className="gu-content__form gu-content__form--thank-you">
         {props.contributionType !== 'ONE_OFF' ? (
-          <section className="confirmation">
-            <h3 className="confirmation__title">
+          <section className="contribution-thank-you-block">
+            <h3 className="contribution-thank-you-block__title">
               {`${directDebitHeaderSuffix}Look out for an email within three business days confirming your ${getSpokenType(props.contributionType)} recurring payment${directDebitMessageSuffix}`}
             </h3>
           </section>
@@ -70,15 +72,24 @@ function ContributionThankYou(props: PropTypes) {
         {!props.isSignedIn ?
           <section className="contribution-thank-you-block">
             <h3 className="contribution-thank-you-block__title">
-              Sign into The Guardian
+              Stay signed in to The Guardian
             </h3>
-            <p className="contribution-thank-you-block__message">If you stay signed into a validated account on each of your devices, you’ll notice far fewer messages asking you for financial support.</p>
-            <AnchorButton
-              href={`https://profile.theguardian.com/signin?email=${props.email}`}
+            <p className="contribution-thank-you-block__message">
+              As a valued contributor, we want to ensure you are having the best experience on our site.
+              To stop seeing requests for support at the bottom of articles, or in pop-up banners, please sign in
+              on each of the devices you use to access The Guardian – mobile, tablet, laptop or desktop.
+            </p>
+            <Button
               aria-label="Sign into The Guardian"
+              appearance="secondary"
+              onClick={
+                () => {
+                  trackComponentClick(`sign-into-the-guardian-link-${props.contributionType}`);
+                  window.location.href = `https://profile.theguardian.com/signin?email=${props.email}`;
+                }}
             >
               Sign in now
-            </AnchorButton>
+            </Button>
           </section> : null }
         <MarketingConsent />
         <ContributionSurvey isRunning={false} contributionType={props.contributionType} />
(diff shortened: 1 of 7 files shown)


support-frontend/assets/pages/contributions-landing/components/ContributionThankYou/ContributionThankYou.jsx
```diff
@@ -57,12 +63,34 @@ function ContributionThankYou(props: PropTypes) {
...
+              To stop seeing requests for support at the bottom of articles, or in pop-up banners, please sign in
+              on each of the devices you use to access The Guardian – mobile, tablet, laptop or desktop.
+            </p>
+            <Button
+              aria-label="Sign into The Guardian"
+              appearance="secondary"
+              onClick={
+                () => {
+                  trackComponentClick(`sign-into-the-guardian-link-${props.contributionType}`);
+                  window.location.href = `https://profile.theguardian.com/signin?email=${props.email}`;
```
jfsoul: Unfortunately this email query parameter is a feature that Identity have been looking to remove for a while - best to avoid using it.  The problem is that the email address could end up leaking to browser extensions and in GA, datalake etc.  Are the team using it anywhere else currently?

Trello card here: https://trello.com/c/epTU4D57/1218-remove-option-to-add-email-address-to-query-strings-sign-in-first-second-step

jlieb10: yeah we are using it when we 'force' sign in for recurring contributors whose emails we recognize as belonging to 'current' accounts

joelochlann: ah that's a shame. yes as Joshua points out we are using it currently.

maybe another solution on identity's end would be a redirect endpoint on profile.theguardian.com that accepts the email param, stores it in a cookie and redirects to the sign-in flow. that way there's no clientside code executed on the endpoint that has the email URL param, but the sign-in flow would have access to it via a cookie.

I think that should give the same functionality but avoid the security/privacy concerns?

support-frontend/assets/pages/contributions-landing/components/ContributionThankYou/ContributionThankYou.scss
```diff
@@ -89,32 +89,26 @@
     line-height: 22px;
   }
 
-  .confirmation {
+  .contribution-thank-you-block {
```
joelochlann: 👍 for the rename

support-frontend/assets/pages/contributions-landing/components/ContributionThankYou/ContributionThankYou.jsx
```diff
@@ -57,12 +61,34 @@ function ContributionThankYou(props: PropTypes) {
...
+              As a valued contributor, we want to ensure you are having the best experience on our site.
+              To stop seeing requests for support at the bottom of articles, or in pop-up banners, please sign in
+              on each of the devices you use to access The Guardian – mobile, tablet, laptop or desktop.
+            </p>
+            <Button
+              aria-label="Sign into The Guardian"
+              appearance="secondary"
+              onClick={
+                () => {
+                  trackComponentClick(`sign-into-the-guardian-link-${props.contributionType}`);
```
joelochlann: hopefully we could also get this info from pageview data on profile.theguardian.com/signin, where the referrer is the contributions thank you page. though that wouldn't distinguish one-off/recurring

just flagging for the future - this extra tracking is great for now as we're rolling this out but I'm always keen for links to be links if they can be

support-frontend/assets/pages/contributions-landing/components/ContributionThankYou/ContributionThankYou.jsx
jranks123: @jfsoul Could we do some kind of hashing of the email?

jranks123: @jlieb10 I wonder if we need to make it super clear which email address to sign in with, i.e the one they made the contribution with, if we are no longer going to send it to profile.thegu

jfsoul: Yes, there’s definitely another solution along those lines. Grab me and/or Guy tomorrow and we can decide the best way!

support-frontend/assets/pages/contributions-landing/components/ContributionThankYou/ContributionThankYou.jsx
jlieb10: yeah that's a good point. I think tracking could also be achieved with query params but i'm not that familiar with that method- in particular where that data ends up. Something for me to look into.

support-frontend/assets/pages/contributions-landing/components/ContributionThankYou/ContributionThankYou.jsx
jlieb10: @jranks123 good shout. the necessary consistency of email address was the reason i hardcoded it into the url - i'll go see if i can get some info from Guy. We already have a fairly big block of copy so I'm hesitant to suggest adding more.

support-frontend/assets/pages/contributions-landing/components/ContributionThankYou/ContributionThankYou.scss
jlieb10: i'll put a card in the backlog for combing through the thank you pages more thoroughly. Each of these blocks on the thank you page has its own class but the same styles.

support-frontend/assets/pages/contributions-landing/components/ContributionThankYou/ContributionThankYou.jsx
Mullefa: In [#486](https://github.com/guardian/identity-frontend/pull/486), profile.theguardian.com was updated to use the `email` query parameter to auto-fill email etc. This was a quick fix when there was no additional functionality for passing email. In general it's something we want to avoid, for the reasons noted by @jfsoul. 

Since that PR, additional functionality has been added to profile.theguardian.com that could provide an alternate way of passing email through. Specifically, there is [an endpoint on identity API](https://github.com/guardian/identity/blob/master/identity-api/src/main/scala/com/gu/identity/api/servlet/SigninTokenServlet.scala#L39) which can be used to encrypt a user's an email, which will then be [decrypted](https://github.com/guardian/identity-frontend/blob/master/app/com/gu/identity/frontend/controllers/Application.scala#L268) if passed as a query parameter for the /signin endpoint. If you used this solution, then you would benefit from a customised sign-in page based on the user's current status.

joelochlann: ah, nice!

PR #1909, opened 2019-06-20
Guardian Weekly Gifting

rupertbates:
## Why are you doing this?

This PR adds gifting to the Guardian Weekly checkout

This is a high level overview of the changes:
* Send gift recipient fields from the checkout form to the `/create` endpoint in the subscribe request
* Add a new `Option[GiftRecipient]` field to several of the support workers states
* When a gift recipient is present add an extra 'Delivery / Recipient Contact' record to Salesforce with the recipient's contact details - in this case do not update the delivery address on the buyer record to avoid overwriting the correct address (if this user already has a Salesforce record) with the gift recipient's address 
* Send the recipient's Salesforce account id to Zuora in the CRM Id - this is either the gift recipient or the buyer.
* Set the correct `ReaderType` on the Zuora account - either Direct or Gift
* Send the buyer's Salesforce id to the email service
* We are also now passing title through to Salesforce for both the buyer and gift recipient

[**Trello Card**](https://trello.com/c/Y8sgB8Gi/2374-gifting-in-gw-checkout-backend)

### DIFF
--- a/support-frontend/app/services/stepfunctions/SupportWorkersClient.scala
+++ b/support-frontend/app/services/stepfunctions/SupportWorkersClient.scala
@@ -26,27 +26,37 @@ import scala.concurrent.Future
 import scala.util.{Failure, Success, Try}
 
 object CreateSupportWorkersRequest {
+
   import codecs.CirceDecoders._
+
   implicit val codec: Codec[CreateSupportWorkersRequest] = deriveCodec
 }
+
 case class CreateSupportWorkersRequest(
-    firstName: String,
-    lastName: String,
-    billingAddress: Address,
-    deliveryAddress: Option[Address],
-    product: ProductType,
-    firstDeliveryDate: Option[LocalDate],
-    paymentFields: PaymentFields,
-    promoCode: Option[PromoCode],
-    ophanIds: OphanIds,
-    referrerAcquisitionData: ReferrerAcquisitionData,
-    supportAbTests: Set[AbTest],
-    email: String,
-    telephoneNumber: Option[String]
+  title: Option[String],
+  firstName: String,
+  lastName: String,
+  billingAddress: Address,
+  deliveryAddress: Option[Address],
+  titleGiftRecipient: Option[String],
+  firstNameGiftRecipient: String,
+  lastNameGiftRecipient: String,
+  emailGiftRecipient: Option[String],
+  product: ProductType,
+  firstDeliveryDate: Option[LocalDate],
+  paymentFields: PaymentFields,
+  promoCode: Option[PromoCode],
+  ophanIds: OphanIds,
+  referrerAcquisitionData: ReferrerAcquisitionData,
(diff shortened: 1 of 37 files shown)


support-frontend/app/services/stepfunctions/SupportWorkersClient.scala
```diff
@@ -26,27 +26,37 @@ import scala.concurrent.Future
...
-    product: ProductType,
-    firstDeliveryDate: Option[LocalDate],
-    paymentFields: PaymentFields,
-    promoCode: Option[PromoCode],
-    ophanIds: OphanIds,
-    referrerAcquisitionData: ReferrerAcquisitionData,
-    supportAbTests: Set[AbTest],
-    email: String,
-    telephoneNumber: Option[String]
+  title: Option[Title],
```
jfsoul: Curious - does this make https://trello.com/c/qmya13nc/2404-checkout-forms-do-not-store-title-field any easier? Or even resolve it?

support-models/src/main/scala/com/gu/salesforce/Salesforce.scala
```diff
@@ -1,73 +1,54 @@
...
+    implicit val encoder: Encoder[DeliveryContact] = deriveEncoder[DeliveryContact]
+      .mapJsonObject(
+        _.add("RecordTypeId", Json.fromString("01220000000VB50AAG"))
+          .wrapObject("newContact")
       )
-    // scalastyle:on parameter.number
   }
 
   //The odd field names on these class are to match with the Salesforce api and allow us to serialise and deserialise
   //without a lot of custom mapping code
```
jfsoul: I feel like this comment might have been intended for `NewContact`?

support-frontend/app/services/stepfunctions/SupportWorkersClient.scala
rupertbates: It resolves it, I thought I might as well do it while I'm here!

support-models/src/main/scala/com/gu/support/workers/states/CreateZuoraSubscriptionState.scala
```diff
@@ -10,11 +11,13 @@ import com.gu.support.encoding.CustomCodecs.{decodeLocalTime, encodeLocalTime}
 case class CreateZuoraSubscriptionState(
   requestId: UUID,
   user: User,
+  giftRecipient: Option[GiftRecipient],
   product: ProductType,
   paymentMethod: PaymentMethod,
   firstDeliveryDate: Option[LocalDate],
   promoCode: Option[PromoCode],
-  salesForceContact: SalesforceContactRecord,
+  salesForceContact: SalesforceContactRecord, //TODO: Remove this it is redundant now we have gifting
```
johnduffell: is this hard or not urgent or just not got around to it yet? (I noticed due to fixing up it:test failures)

PR #1910, opened 2019-06-20
Put sign-in details in user object

joelochlann:
We want the user's validation status (that is, whether they've validated their account by clicking on a link in an email) to be available in the client so that we can direct signed-in users who aren't validated to validate their address.

#### User object (which is, as before, only populated if the user is logged in) now looks like this:

Screenshot 2019-06-20 at 23 30 03

### why should they validate?
The members-data-api does not take account of one-off contributions history for unvalidated accounts, because we can't be sure that the owner of the email also made the contribution. We're encouraging people to sign-in so they don't see supporter asks anymore, but this won't happen for unvalidated one-off contributors.

### what about recurring?
The members-data-api exposes information on recurring contributors even if they're not validated. This is because we have the protection of forcing sign-in for people with passwords before a recurring contribution is made. However, if we make it clear to users that they need to validate and provide an easy way to do so, we might be able to add the same validation restriction on recurring contribution info from the members-data-api, and thereby remove the sign-in requirement to contribute. This would reduce friction and boost conversion (we have data to support this hypothesis)

### DIFF
--- a/support-frontend/app/services/HttpIdentityService.scala
+++ b/support-frontend/app/services/HttpIdentityService.scala
@@ -11,9 +11,10 @@ import com.gu.monitoring.SafeLogger._
 import config.Identity
 import io.circe.Encoder
 import io.circe.generic.semiauto.deriveEncoder
-import models.identity.UserIdWithGuestAccountToken
+import models.identity.{UserIdWithGuestAccountToken, UserWithSignInDetails}
 import models.identity.requests.CreateGuestAccountRequestBody
-import models.identity.responses.{GuestRegistrationResponse, SetGuestPasswordResponseCookies, UserResponse}
+import models.identity.responses.UserSignInDetailsResponse.UserSignInDetails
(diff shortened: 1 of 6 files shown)


support-frontend/app/services/HttpIdentityService.scala
```diff
@@ -194,6 +218,8 @@ class HttpIdentityService(apiUrl: String, apiClientToken: String)(implicit wsCli
 
 trait IdentityService {
   def getUser(user: IdMinimalUser)(implicit req: RequestHeader, ec: ExecutionContext): EitherT[Future, String, IdUser]
+  def getUserSignInDetails(email: String)(implicit ec: ExecutionContext): EitherT[Future, String, UserSignInDetails]
```
tomrf1: this method doesn't need to be on the public interface

joelochlann:
closing in preference to https://github.com/guardian/support-frontend/pull/1915

PR #544, opened 2018-03-02
Upgrade autoprefixer to version 8 and update supported browser list

svillafe:
## Why are you doing this?

Upgrading `autoprefixer` to version 8. Since version 8 autoprefixier uses browserslist, a tool already used by preset-env. The [suggested way of specifying](https://github.com/ai/browserslist#queries) which browser a project supports for browserslits is stating that in `package.json`.  Our current coverage is [here](https://goo.gl/KnaeBn)

[**Trello Card**](https://trello.com/c/HDwP2FhE/1350-setup-correctly-autoprefixier)

## Changes

* Upgrade autoprefixier
* Introduced browserslist's browsers
* Added a second category of supported browsers

### Chrome v38

| Page                     | Screen capture |
|--------------------------|----------------|
| Bundles Landing          |   picture 582  |
| US Contributions Landing |        picture 583          |
| Circles Landing          |        picture 585        |
| Checkout Page            |          picture 584       |

### IE 11
| Page                     | Screen capture |
|--------------------------|----------------|
| Bundles Landing          |  picture 590|
| US Contributions Landing |   picture 591|
| Circles Landing          |        picture 589       |
| Checkout Page            |  picture 592|
### Safari 8
| Page                     | Screen capture |
|--------------------------|----------------|
| Bundles Landing          |       picture 593         |
| US Contributions Landing |  picture 594|
| Circles Landing          |    picture 596|
| Checkout Page            |     picture 595           |

### Edge
| Page                     | Screen capture |
|--------------------------|----------------|
| Bundles Landing          |    picture 599       |
| US Contributions Landing |          picture 600|
| Circles Landing          |    picture 597            |
| Checkout Page            |       picture 598|

### Firefox 44
| Page                     | Screen capture |
|--------------------------|----------------|
| Bundles Landing          | picture 604 |
| US Contributions Landing | picture 601|
| Circles Landing          |  picture 605              |
| Checkout Page            |    picture 602|

### Samsung 4
| Page                     | Screen capture |
|--------------------------|----------------|
| Bundles Landing          |  picture 609|
| US Contributions Landing | #picture 610   |
| Circles Landing          |      picture 606|
| Checkout Page            |    picture 607    |

### Opera 25
| Page                     | Screen capture |
|--------------------------|----------------|
| Bundles Landing          |        picture 613        |
| US Contributions Landing |  picture 612|
| Circles Landing   |         picture 615       |
| Checkout Page            |    picture 614|

PR #1914, opened 2019-06-21
Create sign in token & return sign in url to client

jlieb10:
## Why are you doing this?
The best practice for redirecting contributors to the sign in flow is to do it using the encrypted token pattern. This PR uses identity's encrypted email token service to construct a custom sign in url and return it to the client on request, followed by a client-side redirect to the custom url (link behavior) 

[**Trello Card**](https://trello.com/c/dgx6xVxj/1187-create-encrypted-sign-in-token-return-sign-in-url-to-client)

## Changes

* Creates an endpoint with identity's encrypted email token service that returns a custom sign in url
* Consumes this endpoint on the client using link-like behavior


support-frontend/app/controllers/IdentityController.scala
```diff
@@ -81,14 +82,35 @@ class IdentityController(
         )
     }
   }
+
+  def createSignInURL(): Action[CreateSignInTokenRequest] = PrivateAction.async(circe.json[CreateSignInTokenRequest]) { implicit request =>
```
Mullefa: Note that (by proxy) this is giving public access to an endpoint which currently requires authentication. If we ok with the endpoint that creates encrypted emails being made public, then we could remove the authentication on it and also this proxy. @guardian/identity - any thoughts?

support-frontend/assets/pages/contributions-landing/components/ContributionThankYou/ContributionThankYou.jsx
```diff
@@ -45,6 +50,21 @@ function mapDispatchToProps(dispatch: Dispatch<Action>) {
...
+  fetch(routes.createSignInUrl, {
+    method: 'post',
+    headers: {
+      'Csrf-Token': csrf,
+      'Content-Type': 'application/json',
+    },
+    body: JSON.stringify(payload),
+  })
+    .then(response => response.json())
+    .then((data) => { window.location.href = data.signInLink; });
```
Mullefa: Do we need to handle API errors in this function?

support-frontend/app/controllers/IdentityController.scala
jlieb10: It's still CSRF protected, if that's any consolation

support-frontend/assets/pages/contributions-landing/components/ContributionThankYou/ContributionThankYou.jsx
jlieb10: Yes, we should 👍 thanks

support-frontend/app/controllers/IdentityController.scala
Mullefa: I've given this some more thought and think that my advice in #1908 was probably overkill (apologies). 

To give you some context, we initially created a way of encrypting a user's email address for scenarios where there was no other way to forward this data than as a query parameter; for example, a link in an email. 

The other advantage of using the encrypted email token was that the sign-in page could be personalised based on the user's state. However, this is not applicable in this scenario since someone (via contributing) could obtain an encrypted email for an account that isn't theirs and then navigate to the sign-in page to infer information about the account associated with the respective email address. If this PR was merged as is, then it would be necessary to merge [#544](https://github.com/guardian/identity-frontend/pull/544) first to mitigate this issue.

Since you are already using `fetch()`, I think the simplest solution (and something I should have suggested previously) would be to include the email address in a custom header; we can do the corresponding work on identity-frontend to utilise this header to autofill the email field. This solution is in accordance with the [advice in the OWASP docs](https://owasp-aasvs.readthedocs.io/en/latest/requirement-9.3.html).

Assuming other people agree, I can catch up with you later to pair on this.

jlieb10: @Mullefa - Good thinking around inferring sign in details, but what level of risk do you think this has? From our perspective, it is pretty important to be sending users to profile.theguardian.com using the email address they've just used to make a contribution so our copy will definitely be accurate for them.

In short: I'm wondering if you think this is a 'blocking' level of risk or if you think we can move forward with this and switch over to sending a header with their email when your side is ready to process that header.

lmath: Isn't the work needed in `identity-frontend` to support an email header to prefill the email mostly figuring out where add some tests + https://github.com/guardian/identity-frontend/compare/lm-email-param  ? I'm still not super familiar with identity-frontend so that could definitely be wrong.

jlieb10: @Mullefa @lmath I don't think you can actually route to a new page using fetch and I don't think one can send headers with a link - I could send a get request to profile.theguardian.com/signin with the header set (using fetch) but it would retrieve the document - not route the user there

jranks123: "The other advantage of using the encrypted email token was that the sign-in page could be personalised based on the user's state."

@Mullefa out of interest, what use case is this? The sign in page is unauthenticated, so I'm curious as to what user state info you would consider exposing to someone who isn't logged in (as in I'm genuinely curious, not saying I don't think there is anything that you could show),.

Mullefa: Hey Jonny - yeah, good question. These have previously been used as query parameters in email links. So a user would only be able to see the customised page if they owned the respective email account.

Mullefa: Given the points @jlieb10 made regarding headers and having spoken to @jlieb10 and @jfsoul, the current plan is to roll out implemented as is. As prerequisites to the merge, I'm going to update identity API so that the secret used to encrypt a user's email is not used to encrypt any other data and merge [#544](https://github.com/guardian/identity-frontend/pull/544).

jlieb10:
To make this merge possible, Identity did the following work: (A) they took off the custom styles on the sign in page that would indicate whether a user has social login only or password only (they AB tested this feature and it didn't improve sign in rates so it's unnecessarily leaky) and (B) they are now using a unique encryption key to create the email hash so even if someone cracked it by collecting and analyzing a ton of hashes, they wouldn't gain access to anything

support-frontend/app/services/HttpIdentityService.scala
```diff
@@ -58,12 +59,23 @@ object IdentityService {
...
-
 object GetUserTypeResponse {
   implicit val readsGetUserTypeResponse: Reads[GetUserTypeResponse] = Json.reads[GetUserTypeResponse]
   implicit val getUserTypeEncoder: Encoder[GetUserTypeResponse] = deriveEncoder
 }
 
+case class CreateSignInTokenResponse(encryptedEmail: String)
+object CreateSignInTokenResponse {
+  implicit val readCreateSignInTokenResponse: Reads[CreateSignInTokenResponse] = Json.reads[CreateSignInTokenResponse]
+  implicit val createSignInTokenResponseEncoder: Encoder[CreateSignInTokenResponse] = deriveEncoder
```
tomrf1: I _think_ this `createSignInTokenResponseEncoder` is unnecessary.
The `readCreateSignInTokenResponse` is used to deserialize the response in `createSignInToken`

Confusingly we're mixing two json libraries here, circe and play-json (that's not your fault)

support-frontend/app/services/HttpIdentityService.scala
```diff
@@ -58,12 +59,23 @@ object IdentityService {
...
+}
+
+case class CreateSignInLinkResponse(signInLink: String)
```
tomrf1: It looks like this belongs in `IdentityController`, since that's where it's used (as the response type).
Also, you don't need `readCreateSignInLinkResponse` in this case, because we use circe to serialize the response

support-frontend/app/services/HttpIdentityService.scala
jlieb10: Done ✅

support-frontend/app/services/HttpIdentityService.scala
jlieb10: Done ✅ thanks

support-frontend/app/controllers/IdentityController.scala
jlieb10: Marking this as resolved- thanks- please don't forget to review/approve @Mullefa @lmath

Mullefa:
Pre-requisite PRs merged on identity's side ([#1559](https://github.com/guardian/identity/pull/1559) and [#544](https://github.com/guardian/identity-frontend/pull/544)).

LGTM from an identity view point 👍

PR #1915, opened 2019-06-24
Get emailValidated from cookie

tomrf1:
## Why are you doing this?
If the user is logged in then the `GU_U` cookie has the `emailValidated` field. This means we know whether to tell the user to validate (aka verify) their email before getting the benefits of being signed in.

This also includes CTA for signed in, unverified users to click. The CTA loading is being tracked (loading) `sign-into-the-guardian-link-loaded-{payment-type}` (clicked) `sign-into-the-guardian-link-{payment-type}`

## Screenshots
verification CTA

### DIFF
--- a/support-frontend/assets/helpers/user/user.js
+++ b/support-frontend/assets/helpers/user/user.js
@@ -58,6 +58,14 @@ const signOut = () => { window.location.href = getSignoutUrl(); };
 
 const doesUserAppearToBeSignedIn = () => !!cookie.get('GU_U');
 
+// JTL: The user cookie is built to have particular values at
+// particular indices by design. Index 7 in the cookie object represents
+// whether a signed in user is validated or not. Though it's not ideal
+// to grab values at unnamed indexes, this is a decision made a long
+// time ago, on which a lot of other code relies, so it's unlikely
+// there will be a breaking change affecting our base without some advance
+// communication to a broader segment of engineering that also uses
+// the user cookie.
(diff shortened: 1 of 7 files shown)


support-frontend/assets/helpers/user/user.js
```diff
@@ -58,6 +58,20 @@ const signOut = () => { window.location.href = getSignoutUrl(); };
 
 const doesUserAppearToBeSignedIn = () => !!cookie.get('GU_U');
 
+const getEmailValidatedFromUserCookie = (guuCookie: ?string) => {
+  if (guuCookie) {
+    const tokens = guuCookie.split('.');
+    try {
+      const parsed = JSON.parse(atob(tokens[0]));
+      return !!parsed[7]
```
jlieb10: What are the chances of this index changing?

tomrf1: It would break a lot of things, this stuff is hardcoded in identity

jlieb10: got it

jlieb10: I added inline documentation

tomrf1: 👍

jlieb10:
Note: testing in production fails because of google optimize changes on the Digital Pack landing page -- the check to see that the page has loaded is only valid for one of the variants

PR #1916, opened 2019-06-24
Unify setup script

joelochlann:
There didn't seem much point in having a separate nginx setup script, and this way lets us use a single `Brewfile` to specify all Brew dependencies for the project.

This works for me locally, thought the real test is running it on a clean laptop. I tried on a UXer's machine but there were various other problems (nvm not installed, brew not installed) - it's probably not been tested for a while in an environment like that

If anyone has any bright ideas for testing let me know. Maybe I should try on a UXer's laptop after running the dev boostrap script at https://gu-strap.herokuapp.com/?


support-frontend/nginx/README.md
```diff
@@ -1,19 +1,15 @@
...
-brew bundle
 ```
-- Run [setup.sh](setup.sh) script from the root of the `support-frontend` project:
+./setup.sh
+```
 
+#### Running nginx
 ```
-./nginx/setup.sh
+sudo nginx
```
jlieb10: Ii don't run this with sudo - not sure if it depends on the way one installs it

jlieb10: could flag it as optional?

support-frontend/setup.sh
```diff
@@ -98,43 +98,40 @@ install_node() {
...
   fi
 }
 
 install_js_deps() {
   yarn install
 }
 
+install_brew_deps() {
+  # Installs dependencies defined in Brewfile
+  brew bundle
```
jlieb10: great refactor to use the brewfile 👍

support-frontend/nginx/README.md
joelochlann: I thought it had to be run with `sudo` in order to bind to port 80?

joelochlann: actually I've just tried it (I had to `sudo chown joseph_smith:admin /usr/local/var/log/nginx/access.log` and `sudo chown joseph_smith:admin /usr/local/var/log/nginx/error.log`), and it can indeed run without `sudo`!

joelochlann: interesting - I always thought you had to be running as `root` to bind to port below 1024. https://unix.stackexchange.com/questions/10735/allowing-a-user-to-let-listen-to-a-port-below-1024

maybe it applies only to Linux to Mac OS X? Maybe @akash1810 can shed some light since he's our current nginxpert...

support-frontend/setup.sh
joelochlann: props to @akash1810 for suggestion

support-frontend/Brewfile
```diff
@@ -1,2 +1,6 @@
 tap "guardian/devtools"
 brew "guardian/devtools/dev-nginx"
+brew "nginx"
```
akash1810: nginx will be [pulled in by `dev-nginx`](https://github.com/guardian/homebrew-devtools/blob/master/Formula/dev-nginx.rb#L8)

support-frontend/Brewfile
```diff
@@ -1,2 +1,6 @@
...
+brew "awscli"
+brew "sbt@1"
```
akash1810: minor: we could use the [`gu-scala` cask](https://github.com/guardian/homebrew-devtools/blob/master/Casks/gu-scala.rb) here instead

support-frontend/nginx/README.md
akash1810: my understanding was similar to yours, binding to a low port required `sudo`.

support-frontend/setup.sh
```diff
@@ -4,7 +4,7 @@ set -e
 readonly SYSTEM=$(uname -s)
 EXTRA_STEPS=()
 
-NGINX_ROOT=/usr/local/etc/nginx/
+NGINX_ROOT=/usr/local/etc/nginx
```
akash1810: this can be removed

support-frontend/setup.sh
```diff
@@ -98,43 +98,40 @@ install_node() {
...
 
-install_nginx() {
-  if ! installed nginx; then
-    brew install nginx
-    EXTRA_STEPS+=("nginx has been installed. Ensure you have 'include sites-enabled/*' in your nginx configuration ${NGINX_ROOT}/nginx.conf and add '127.0.0.1 support.thegulocal.com' and '127.0.0.1 support-ui.thegulocal.com' to /etc/hosts")
-  else
-    EXTRA_STEPS+=("nginx is installed. Ensure you have 'include sites-enabled/*' in your nginx configuration ${NGINX_ROOT}/nginx.conf and add '127.0.0.1 support.thegulocal.com' and '127.0.0.1 support-ui.thegulocal.com' to /etc/hosts")
-  fi
+setup_nginx() {
+  DOMAIN=support.thegulocal.com
```
akash1810: this variable can be removed

support-frontend/setup.sh
```diff
@@ -148,12 +145,13 @@ main () {
...
-  install_awscli
+  install_brew_deps
+  setup_nginx
   install_jdk
-  install_sbt
-  install_yarn
+  install_yarn_if_linux
   install_js_deps
+
+  EXTRA_STEPS+=("nginx is installed. Ensure you have 'include sites-enabled/*' in your nginx configuration ${NGINX_ROOT}/nginx.conf and add '127.0.0.1 support.thegulocal.com' and '127.0.0.1 support-ui.thegulocal.com' to /etc/hosts")
```
akash1810: this is no longer needed as:
- dev-nginx uses the [`servers` directory](https://github.com/guardian/dev-nginx/blob/master/script/link-config#L25)
- later versions of nginx (certainly any version installed on our new machines) will include the `include servers/*` directive in its default config
- `*.thegulocal.com` exists in DNS to resolve to `127.0.0.1` so we don't need to edits hosts file now

PR #1918, opened 2019-06-24
Update analytics for cohorts 3, 4, 5

jlieb10:
## Why are you doing this?
1. We need to get numbers for how many users are seeing the "sign in" button to get a sense of the percentage that are 'converting' to signed in
2. It makes sense to track the 'decline to set password' across the same dimension (contribution type) that we are now tracking those who do set a password

[**Trello Card**](https://trello.com/b/QVWhizic/contributions)

## Changes

* Added a tracking event to `ophan.js` for components loading
* Made trackable versions of the reusable button component using the `componentDidMount` lifecycle hook
* Integrated new tracking event and trackable button on the thank you page
* Updated the tracking on 'decline to set password' link to include contribution type

### DIFF
--- a/support-frontend/assets/components/button/trackableAnchorButton.jsx
+++ b/support-frontend/assets/components/button/trackableAnchorButton.jsx
@@ -1,0 +1,40 @@
+// @flow
+
+// ----- Imports ----- //
+
+import React, { Component } from 'react';
+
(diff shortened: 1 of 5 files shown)


jlieb10:
Regarding setting 'static' inside the class, this is flow's preferred syntax

PR #1919, opened 2019-06-24
Clean up unneeded props

joelochlann:
All of these props are actually supplied directly to the child components that need them via Redux wiring. They don't need to be supplied to the parent.


jlieb10:
@joelochlann TeamCity was erroring all of the builds on this repo yesterday evening but it seems to be fixed now so I re-ran this branch to pass

PR #1920, opened 2019-06-25
Update Brewfile

akash1810:
Full namespace for gu-scala cask

## Why are you doing this?

## Changes

* Change 1
* Change 2


akash1810:
This doesn't solve the issue @jesseyuen was experiencing; that fix will be raised in a different PR. This one can be closed.

PR #1922, opened 2019-06-25
Make first name and last name optional in gifting model

rupertbates:
## Why are you doing this?

Previously the giftee first name and last name defaulted to empty strings, this made it slightly more difficult to map to the `Option[String]` values used to model these fields on the server. This PR defaults them to null to make their status more explicit.


support-frontend/assets/helpers/paymentIntegrations/readerRevenueApis.js
```diff
@@ -75,8 +89,8 @@ export type RegularPaymentRequest = {|
   deliveryAddress: Option<RegularPaymentRequestAddress>,
   email: string,
   titleGiftRecipient?: Option<Title>,
-  firstNameGiftRecipient?: string,
-  lastNameGiftRecipient?: string,
+  firstNameGiftRecipient?: Option<string>,
+  lastNameGiftRecipient?: Option<string>,
```
johnduffell: what does it mean if it's both optional `?` and an option `Option`?

johnduffell: if this is a "payment request", surely the gift status is not relevant? payment would just be for billing info/frequency/amount and payment card number?

rupertbates: It can be missing entirely and if it is present it can be null

support-frontend/assets/helpers/subscriptionsForms/formReducer.js
```diff
@@ -43,9 +43,9 @@ function createFormReducer(
     billingAddressIsSame: null,
     billingPeriod: initialBillingPeriod,
     titleGiftRecipient: null,
-    firstNameGiftRecipient: '',
-    lastNameGiftRecipient: '',
-    emailGiftRecipient: '',
+    firstNameGiftRecipient: null,
+    lastNameGiftRecipient: null,
+    emailGiftRecipient: null,
```
johnduffell: shoudl all the "gift" stuff be bound together more tightly to avoid so many codependent nullable vaiables?  Like a `GiftInfo` data type that is a sum type containing either Gift(title, firstname, lastname, email) or NonGift

rupertbates: Yes I think it should and it is in the Scala model, but it had already been implemented like this when I got to it and it would be quite a big change to refactor it so I left it as is.

support-frontend/assets/helpers/paymentIntegrations/readerRevenueApis.js
johnduffell: sorry I wasn't clear, I meant what does it mean in real life!  I know what it means in terms of the code.
I guess missing means it's not a gift, and null means it wasn't entered by the gifter?

PR #1928, opened 2019-06-27
 Pre-render an 'empty' contributions landing page 

tomrf1:
## Why are you doing this?
Currently we do server-side rendering by putting an empty version of the landing page into the `contributions.scala.html` template, which then gets replaced when client-side rendering kicks in.

This PR is an initial proposal for switching to using pre-rendering: building the 'empty' version of the react components at _build_ time (on teamcity). The generated html is then loaded from file at runtime and embedded in the twirl template.
This has the advantage of keeping the 'empty' version of each component in the same file as the real thing, rather than split between `.scala.html` and `.jsx` files.

The empty version of each component must not depend on props or the redux store. The approach I have used is to replace e.g.
```
function ContributionForm(props: PropTypes) {...}

export default connect(mapStateToProps, mapDispatchToProps)(ContributionForm);
```
with 2 exports:
```
function withProps(props: PropTypes) {...}
function withoutProps() {...}

export const ContributionForm = connect(mapStateToProps, mapDispatchToProps)(withProps);
export const EmptyContributionForm = withoutProps;
```

## Screenshots
Screenshot 2019-06-27 at 13 50 10


support-frontend/assets/pages/contributions-landing/components/ContributionTypeTabs.jsx
```diff
@@ -94,4 +94,30 @@ function ContributionTypeTabs(props: PropTypes) {
...
 }
 
-export default connect(mapStateToProps, mapDispatchToProps)(ContributionTypeTabs);
+function withoutProps() {
+  return (
+    <fieldset className={classNameWithModifiers('form__radio-group', ['tabs', 'contribution-type'])}>
+      <legend className={classNameWithModifiers('form__legend', ['radio-group'])}>Recurrence</legend>
+      <ul className="form__radio-group-list form__radio-group-list--border">
+        {
+          ['a','b','c'].map(id => {
```
johnduffell: in fact, if we happen to know there are no AB tests running we could just use the actual one.  I think the tabs ordering test was running, which is why these were greyed out

johnduffell:
this is great and couldn't possibly be more hacky than the original mechanism! 😂 
I would say go for it if it works! 🥇 and there will be a lot more opportunities to improve the structure later.

support-frontend/app/controllers/Application.scala
```diff
@@ -129,9 +129,12 @@ class Application(
 
     val js = elementForStage(RefPath("contributionsLandingPage.js"))
 
+    val mainElement = assets.getSsrCacheContentsAsHtml(s"contributions-landing-page-$countryCode","contributions-landing.html")
```
joelochlann: unrelated to your PR but we should probably ditch the country code in the main element id, it's unnecessary and a potential source of bugs. in fact even better would be to have a universal root element id across all pages

tomrf1:
Lighthouse performance is consistent with master (in CODE)

PR #1929, opened 2019-06-27
Guardian Weekly thank you email

rupertbates:
## Why are you doing this?

When a user purchases a Guardian Weekly subscription we want to send them an email confirming what they have purchased.

Changes:
* Pass the promo code through to the `SendThankYouEmail` lambda so that we can tell when a subscription is 6 for 6.
* Add a `GuardianWeeklyEmailFields` case class to generate the necessary data for the email
* Update the `SubscriptionEmailFieldHelpers.describe` method to cope with 6 for 6

[**Trello Card**](https://trello.com/c/uTIjJqm5/2336-guardian-weekly-thank-you-email)

### DIFF
--- a/support-workers/src/main/scala/com/gu/emailservices/PaperFieldsGenerator.scala
+++ b/support-workers/src/main/scala/com/gu/emailservices/PaperFieldsGenerator.scala
@@ -18,8 +18,8 @@ object PaperFieldsGenerator {
     currency: Currency,
     paymentMethod: PaymentMethod,
     sfContactId: SfContactId,
-    directDebitMandateId: Option[String] = None,
-    promotion: Option[Promotion] = None
+    directDebitMandateId: Option[String],
+    promotion: Option[Promotion]
   ): List[(String, String)] = {
 
     val firstPaymentDate = SubscriptionEmailFieldHelpers.firstPayment(paymentSchedule).date
--- a/support-workers/src/main/scala/com/gu/emailservices/PaperFieldsGenerator.scala
+++ b/support-workers/src/main/scala/com/gu/emailservices/PaperFieldsGenerator.scala
@@ -2,6 +2,7 @@ package com.gu.emailservices
 
 import com.gu.i18n.Currency
 import com.gu.salesforce.Salesforce.SfContactId
+import com.gu.support.promotions.Promotion
 import com.gu.support.workers._
 import org.joda.time.LocalDate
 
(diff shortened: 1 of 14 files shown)


support-workers/src/main/scala/com/gu/emailservices/PaperFieldsGenerator.scala
```diff
@@ -0,0 +1,77 @@
...
+
+    val firstPaymentDate = SubscriptionEmailFieldHelpers.firstPayment(paymentSchedule).date
+
+    val paymentFields = getPaymentFields(paymentMethod, directDebitMandateId)
+
+    val deliveryAddressFields = getAddressFields(user)
+
+    val fields = List(
+      "ZuoraSubscriberId" -> subscriptionNumber,
+      "SubscriberKey" -> user.primaryEmailAddress,
```
paulbrown1982: I’d like to explore if we can remove the SubscriberKey field now that we’re no longer using ET. I’ll have a scan of the Workflow Service to see if it still uses it in any places rather than the EmailAddress.

support-workers/src/main/scala/com/gu/emailservices/PaperFieldsGenerator.scala
```diff
@@ -0,0 +1,77 @@
...
+      "EmailAddress" -> user.primaryEmailAddress,
```
paulbrown1982: @rupertbates I've just spotted in the Membership Workflow service that this field might be called 'Address' rather than EmailAddress - https://github.com/guardian/membership-workflow/blob/a5239afe520dc3df22c27a6c9cd24169424d314d/app/model/EmailPayload.scala#L15

support-workers/src/main/scala/com/gu/emailservices/PaperFieldsGenerator.scala
paulbrown1982: It doesn't seem to be used so I'll raise a PR to remove it.

support-workers/src/main/scala/com/gu/emailservices/PaperFieldsGenerator.scala
rupertbates: That's at the next level up, here: https://github.com/guardian/support-frontend/blob/35e1d03b276188509e101993f77720a8dff732c4/support-workers/src/main/scala/com/gu/emailservices/EmailFields.scala#L27

The fields are in the payload

support-workers/src/main/scala/com/gu/emailservices/PaperFieldsGenerator.scala
paulbrown1982: See: https://github.com/guardian/membership-workflow/pull/194

PR #1930, opened 2019-06-28
Make user authentication asynchronous

Mullefa:
## Why are you doing this?

Identity are wanting to centralise user authentication i.e. an application should authenticate a user by making a call to the authentication endpoint in identity API  (please talk to us in person for more context).

This changes the type signature of user authentication from:
```scala
RequestHeader => Option[U]
```
to:
```scala
RequestHeader => Future[Option[U]]
```
To lay the ground-work for this change, define and use the class `AsyncAuthenticatedBuilder` to support this change in type signature.  

__Notes:__
- this PR does not change the authentication mechanism used (we are still verifying the cookie's signature using a secret that is imported via the config); authenticating via the identity API will be done in a subsequent PR
- it is possible that some of the classes in this PR (`AsyncAuthenticatedBuilder`, `AsyncAuthenticationService`) could get moved to [identity-play-auth](https://github.com/guardian/identity-play-auth). This will only happen once we are confident of their general usefulness across Guardian applications that require user authentication.


lmath:
👍 lgtm from an identity POV but I suppose you'll want a RR dev to weigh in

support-frontend/app/actions/AsyncAuthenticatedBuilder.scala
```diff
@@ -0,0 +1,41 @@
...
+// In the context of Guardian applications, asynchronous authentication equals making a call to identity API,
+// to authenticate the user data.
+// TODO: considering porting this to identity-play-auth.
+// scalastyle:on
+class AsyncAuthenticatedBuilder[U](
+    userinfo: RequestHeader => Future[Option[U]],
+    defaultParser: BodyParser[AnyContent],
+    onUnauthorized: RequestHeader => Result
+)(implicit val executionContext: ExecutionContext)
+  extends ActionBuilder[({ type R[A] = AuthenticatedRequest[A, U] })#R, AnyContent] with LazyLogging { // scalastyle:ignore
```
johnduffell: good old type lambdas! - consider pull this out as a named type to make it less WTF (I have experience of this situation!)

johnduffell: I can't remember exactly and I'm now wondering if it's possbile in all cases but something like 
```scala
class A[U] {
  type R[A] = AuthenticatedRequest[A, U]
}
```
then
```scala
... ActionBuilder[A[U]#R, AnyContent] ...
````

File: support-frontend/app/actions/AsyncAuthenticatedBuilder.scala [hunk 2]
```diff
@@ -0,0 +1,41 @@
+package actions
+
+import com.typesafe.scalalogging.LazyLogging
+import play.api.mvc.Security.AuthenticatedRequest
+import play.api.mvc._
+
+import scala.concurrent.{ExecutionContext, Future}
+
+// scalastyle:off
+// This is based on AuthenticationBuilder in https://github.com/playframework/playframework/blob/1ca9d0af237ac45a011671b2e036d726cd05e4e7/core/play/src/main/scala/play/api/mvc/Security.scala
```
johnduffell:
```suggestion
// This is based on AuthenticatedBuilder in https://github.com/playframework/playframework/blob/1ca9d0af237ac45a011671b2e036d726cd05e4e7/core/play/src/main/scala/play/api/mvc/Security.scala
```

support-frontend/app/actions/AsyncAuthenticatedBuilder.scala
```diff
@@ -0,0 +1,41 @@
...
+  def invokeBlock[A](request: Request[A], block: AuthenticatedRequest[A, U] => Future[Result]): Future[Result] =
+    authenticate(request, block)
+
+  /**
+    * Authenticate the given block.
+    */
+  def authenticate[A](request: Request[A], block: AuthenticatedRequest[A, U] => Future[Result]): Future[Result] = {
+    userinfo(request).flatMap {
+      case Some(user) => block(new AuthenticatedRequest(user, request))
+      case None => Future.successful(onUnauthorized(request))
```
johnduffell: Not recommending a change, but here's an interesting friday topic:
the interesting thing is that this copy of the class is needed because the monad to use (option) is not a type parameter, so the logic isn't reusable.  Not that I'm suggestion the tagless final approach!
However an interesting point is that in some way map/getOrElse (or fold could be used) is more abstract than pattern match, as pattern match only works on ADTs wheras the high level methods work on OptionT as well https://typelevel.org/cats/datatypes/optiont.html

support-frontend/app/actions/AsyncAuthenticatedBuilder.scala
```diff
@@ -0,0 +1,41 @@
...
+// TODO: considering porting this to identity-play-auth.
+// scalastyle:on
+class AsyncAuthenticatedBuilder[U](
+    userinfo: RequestHeader => Future[Option[U]],
+    defaultParser: BodyParser[AnyContent],
+    onUnauthorized: RequestHeader => Result
+)(implicit val executionContext: ExecutionContext)
+  extends ActionBuilder[({ type R[A] = AuthenticatedRequest[A, U] })#R, AnyContent] with LazyLogging { // scalastyle:ignore
+
+  lazy val parser = defaultParser
```
johnduffell: it's risky in scala to use lazy vals and LazyLogging together with Futures.  lazy would get a (java) lock on the containing object, which could deadlock if there is a future that is mapped or flatmapped between the two.
(see https://alvinalexander.com/scala/look-how-scala-lazy-val-converted-to-java-code-jvm-bytecode )
In this case it will be fine, as the value of parser is a static value, but if the defaultParser was a code block or a method call, things get more risky.
We have seen this issue in membership-frontend.
There was a SIP to change that But I'm not sure if it got anywhere: https://docs.scala-lang.org/sips/improved-lazy-val-initialization.html#version-6---no-synchronization-on-this-and-concurrent-initialization-of-fields

support-frontend/app/actions/CustomActionBuilders.scala
```diff
@@ -53,14 +51,18 @@ class CustomActionBuilders(
   }
 
   private def maybeAuthenticated(onUnauthenticated: RequestHeader => Result): ActionBuilder[OptionalAuthRequest, AnyContent] =
-    new AuthenticatedBuilder(authenticatedIdUserProvider.andThen(Some.apply), cc.parsers.defaultBodyParser, onUnauthenticated)
+    new AsyncAuthenticatedBuilder(
+      userinfo = requestHeader => asyncAuthenticationService.authenticateUser(requestHeader).map(Some.apply),
+      cc.parsers.defaultBodyParser,
+      onUnauthenticated
```
johnduffell: funny to name the first parameter but not the rest, either all or none would be worth considering

johnduffell:
yeah I think it's good, and excellent work doing it in stages like this. I think if authenticated will need a further REST api call in future this would need considering in term of performance, and whether that puts more obligation on idapi to be up throughout the whole checkout process.
I have made a few comments but nothing really substantial!
Thanks for doing this 👍

support-frontend/app/actions/AsyncAuthenticatedBuilder.scala
Mullefa: I'm a but confused on how you'd pull it out since it's reliant on `U` which is a type parameter for the class. I've used [typeleve/kind-projector](https://github.com/typelevel/kind-projector) in other applications, which I'd be happy to use here to simplify the syntax (?)

support-frontend/app/actions/AsyncAuthenticatedBuilder.scala
Mullefa: So something like:
```scala
lazy val parser = {
  logger.info("initialising parser")
  defaultParser
}
```
would not be cool?

support-frontend/app/actions/AsyncAuthenticatedBuilder.scala
johnduffell: maybe it's not possible, as I was typing it all out I was starting to have doubts about scoping! sorry for the red herring 🐠

support-frontend/app/actions/AsyncAuthenticatedBuilder.scala
johnduffell: not that simple, the problem would actually be
```scala
lazy val parser = Future.successful(()).map{ _ =>
  logger.info("initialising parser")
  defaultParser
}
```
so basically the lazy val gets the lock, the `map` on the future uses a different thread, and the `logger.` would try to initialise another lazy val and get the same lock

johnduffell:
looks great to me! 👍

support-frontend/app/actions/AsyncAuthenticatedBuilder.scala
Mullefa: ah I see - so because lazy vals lock on the surrounding class instance, the synchronous version wouldn't be a problem (since the thread that has a lock on the class instance is the same thread that's evaluating the logging statement) (?)

johnduffell: yes exactly, that would be fine.
I'm not sure why they don't just create one object for each lazy val, maybe just perf reasons.  Even in my java days I used to just create an object for synchronising if I needed to do it rather than relying on something I didn't control.

PR #1932, opened 2019-06-28
push a GET to cloudwatch if a client side error happens

johnduffell:
## Why are you doing this?

The main dscription is https://github.com/guardian/support-service-lambdas/pull/331
We want to know when client side errors happen, in a way that is actionable.  At present they go into sentry and get lost, so this should wake us up to it.

[**Trello Card**](https://trello.com/c/80GozO7M/2429-beacon-error-pages-into-cloudwatch-and-alarm-on-them)

Tested locally, writing the metrics into CODE successfully.


support-frontend/app/admin/settings/SettingsProvider.scala
```diff
@@ -30,10 +31,17 @@ abstract class SettingsProvider[T] {
...
+  metricUrl: MetricUrl
+) {
 
   def getAllSettings(): AllSettings = {
-    AllSettings(switchesProvider.settings(), amountsProvider.settings(), contributionTypesProvider.settings())
+    AllSettings(
+      switchesProvider.settings(),
+      amountsProvider.settings(),
+      contributionTypesProvider.settings(),
+      metricUrl
```
johnduffell: I did this as a static value as I don't think we will change it without a deploy, as it's an infra thing rather than a user facing setting.  The additional wrapper looked to have a fair bit of ceremony over no special treatment.  We can always wrap it in the appropriate abstraction later.

support-frontend/assets/helpers/render.js
```diff
@@ -20,6 +20,8 @@ const getElementOrBody = (id: ?string): HTMLElement => {
 };
 
 const renderError = (e: Error, id: ?string) => {
+  fetch(window.guardian.settings.metricUrl, {'mode': 'no-cors'}); // ignore result, fire and forget
```
johnduffell: this uses window.guardian directly, as it is run when something has gone badly wrong, we don't want to rely on too much working at that point.  The other option is to put the getSettings outside the try block, but that is something that could easily break when adding settings (I think there was some issue recently with pricing missing)

support-frontend/conf/CODE.public.conf
```diff
@@ -15,8 +15,9 @@ support.url="https://support.code.dev-theguardian.com"
...
+play.filters.headers.contentSecurityPolicy = "default-src 'self' www.paypalobjects.com members-data-api.code.dev-theguardian.com metric-push-api-code.support.guardianapis.com www.paypal.com www.sandbox.paypal.com js.stripe.com pasteup.guim.co.uk ophan.theguardian.com j.ophan.co.uk media.guim.co.uk uploads.guim.co.uk www.google-analytics.com www.googletagmanager.com tagmanager.google.com polyfill.io www.googleadservices.com googleads.g.doubleclick.net www.google.com www.google.co.uk optimize.google.com static.ads-twitter.com bat.bing.com bid.g.doubleclick.net t.co analytics.twitter.com stats.g.doubleclick.net www.youtube-nocookie.com connect.facebook.net www.facebook.com consumer.krxd.net cdn.krxd.net beacon.krxd.net secure.adnxs.com checkout.stripe.com fonts.googleapis.com ssl.gstatic.com www.gstatic.com fonts.gstatic.com sentry.io www.dwin1.com data: wss: 'unsafe-inline' q.stripe.com payment.code.dev-guardianapis.com https://interactive.guim.co.uk/ https://www.theguardian.com/ https://theguardian.com/"
```
johnduffell: github isn't clear but the change on this line was add `metric-push-api-code.support.guardianapis.com`

PR #1933, opened 2019-06-28
Remove stub identity service

Mullefa:
## Why are you doing this?

@lmath said the stub service was motivated by identity API CODE having regular downtime; in which case, when developing locally you could update the config to use the stub.

However, the uptime of identity API CODE in its current state doesn't merit a need for this stub service, therefore, let's remove it.

Along with the general benefits of removing redundant code, this change reduces the surface area of [identity-play-auth](https://github.com/guardian/identity-play-auth) used by this project. If analogous changes can be made to all applications using identity-play-auth, we can simplify / reduce the scope of this library.


lmath:
gosh who introduced this in the first place? 🙈

johnduffell:
sounds good to me! it's always rarer to delete than add code, so it's good that it could be added when it was useful and removed when it isn't any more :+1:

PR #1935, opened 2019-06-28
Redirect with encoded query string on application and subscription controllers

jlieb10:
## Why are you doing this?
Extending @tomrf1 's work here: https://github.com/guardian/support-frontend/pull/1884

Which will eliminate 500 errors when users come to the site from a URL which includes a dash in the query parameters

Tested the offending urls: 
https://support.thegulocal.com/contribute?INTCMP=side_menu_support_contribute&acquisitionData=%7B&quo.%E2%80%99

https://support.thegulocal.com/subscribe?INTCMP=side_menu_support_contribute&acquisitionData=%7B&quo.%E2%80%99

### DIFF
--- a/support-frontend/app/lib/RedirectWithEncodedQueryString.scala
+++ b/support-frontend/app/lib/RedirectWithEncodedQueryString.scala
@@ -1,0 +1,14 @@
+package lib
+
+import play.api.http.Status
+import play.api.mvc.{Result, Results}
+
+object RedirectWithEncodedQueryString extends Results with Status {
+  //Encode the querystring parameter keys, as Play only encodes the values
+  def apply(url: String, queryString: Map[String, Seq[String]] = Map.empty, status: Int = SEE_OTHER): Result = Redirect(
+    url = url,
+    queryString = queryString.map { case (k,v) => java.net.URLEncoder.encode(k, "utf-8") -> v },
+    status = status
+  )
+}
+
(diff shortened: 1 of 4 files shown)


support-frontend/app/lib/RedirectWithEncodedQueryString.scala
```diff
@@ -0,0 +1,14 @@
+package lib
+
+import play.api.http.Status
+import play.api.mvc.{Result, Results}
+
+object RedirectWithEncodedQueryString extends Results with Status {
+  //Encode the querystring parameter keys, as Play only encodes the values
+  def apply(url: String, queryString: Map[String, Seq[String]] = Map.empty, status: Int = SEE_OTHER): Result = Redirect(
+    url = url,
```
jlieb10: @tomrf1 what do you think of this change for safety/defensiveness? Is there any risk?
```suggestion
    url = java.net.URLEncoder.encode(url, "utf-8"),
```

tomrf1: good question.
Based on [this answer](https://stackoverflow.com/a/7654605), you are correct that it should be encoded in the `Redirect` header.
Please test that redirects still work with the change

jlieb10: This change seems to break the redirect for contributions so I'm going to bin it

PR #1937, opened 2019-06-28
Include 6 for 6 and gifting

lucymonie:
## Why are you doing this?
The idea is to make the GW checkout as straightforward as possible, so no matter which billing period the user selects, we show gifting in the checkout, and we show 6 for 6 even if they selected quarterly or annual on the landing page. Based on this [**Trello Card**](https://trello.com/b/TjUElW0B/simple-and-coherent-product-supporting-the-guardian)

## Changes
* Removes the code that hides gifting if the user selects 6 for 6 
* Removes the code that hides 6 for 6 if the user selects gifting

## Screenshots
Screen Shot 2019-06-28 at 16 47 06
Screen Shot 2019-06-28 at 16 44 16


lucymonie:
Weirdly enough, no @jfsoul - that was never actually a thing (other than in a PR I opened 2 weeks but have closed). 

The thing that's more problematic is that we don't currently track which type of billing period the user has selected on the landing page and display that as the selected on in the checkout.

PR #1938, opened 2019-06-28
Use smaller image size for premium app

jfsoul:
## Why are you doing this?
We're loading a full size crop of one the images on the digipack product page (variant only).  Reducing that down to a more reasonable 1000px width, which reduces the weight significantly.

To be honest I'm not entirely clear what `sizes="(min-width: 480px) 100%, (max-width: 660px) 100%, 100%"` is trying to achieve.  To me that is saying at any width of viewport set the width of the image to 100% 😕 

## Changes

* Change 1
* Change 2

## Screenshots
Before (1.2MB):
image

After (161KB):
image


lucymonie:
Could you please add a screenshot @jfsoul?

jfsoul:
Good point, done now @lucymonie

PR #1941, opened 2019-07-01
Add alerts for DigiPack and Paper checkout flows

jorgeazevedo:
## Why are you doing this?

We've recently had an outage where paper checkout was down for 11am one day to 17h the next day. We've been wanting to add a cloud watch alarm that goes off when acquisition flatlines for a while now, but after that outage this was bumped in priority.

[**Trello Card**](https://trello.com/c/5vAIlA4j/2419-add-alarms-for-no-conversions-within-x-hours-days-to-all-variants)

## Changes

* Adds two alarms NoPaperAcquisitionInOneDayAlarm and NoDigipackAcquisitionInEightHoursAlarm. So that's 24h and 8h for paper and digipack respectively.
* The general idea is to use a calculate metric that sums all the individual payment type metrics. This allows us to have a global alert, on all acquisition, rather than independent alerts per payment type. The implementation is slightly obscure, if anyone knows of a clearer way to define them I'd be happy to refactor
* The emails have subjects
  *  ` URGENT 9-5 - ${Stage} support-workers No successful digipack checkouts in 8h`
  * `URGENT 9-5 - ${Stage} support-workers No successful paper checkouts in 24h`

### DIFF
--- a/support-workers/cloud-formation/src/templates/cfn-template.yaml
+++ b/support-workers/cloud-formation/src/templates/cfn-template.yaml
@@ -286,3 +286,138 @@ Resources:
       Statistic: Sum
       TreatMissingData: breaching
     DependsOn: SupportWorkersPROD
+
+
+  # This alarm is for the PaymentSuccess metric where
+  # ProductType == DigitalPack AND PaymentProvider in [Stripe, Gocardless, Paypal]
+  #
+  # The general idea here is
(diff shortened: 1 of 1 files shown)


jorgeazevedo:
Yeah, absolutely : -\ Thanks! I'll merge and see if it comes through alright. It's actually easier to interpret the CF when you see the final result, with the SUMs and FILLs plotted in a chart

PR #1942, opened 2019-07-01
Weekly landing page

rupertbates:
## Why are you doing this?
To launch GW checkout we need to update the landing page so that it goes to the new checkout. We will then run a test in Optimize to send a proportion of our traffic to the new flow for testing.

## Changes

* Update url generation code to switch between the old an new checkout based on experiment variant
* Update the copy around gifting on the GW landing page
* Remove Google auth from the new checkout page
* There are quite a few small changes which have been made automatically by Eslint

[**Trello Card**](https://trello.com/c/Wjqqlnrp/2436-phased-rollout-gw) & [also]( https://trello.com/c/t9d6wXyW/2428-remove-google-auth-from-gw-checkout-just-before-go-live)

### DIFF
--- a/support-frontend/app/controllers/WeeklySubscription.scala
+++ b/support-frontend/app/controllers/WeeklySubscription.scala
@@ -43,9 +43,7 @@ class WeeklySubscription(
   implicit val a: AssetsResolver = assets
 
 
-  //TODO: remove Google Auth from this endpoint before we go live
-  def displayForm(): Action[AnyContent] =
-    (PrivateAction andThen authAction andThen authenticatedAction(subscriptionsClientId)).async { implicit request =>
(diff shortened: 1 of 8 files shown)


support-frontend/app/controllers/WeeklySubscription.scala
```diff
@@ -43,9 +43,7 @@ class WeeklySubscription(
   implicit val a: AssetsResolver = assets
 
 
-  //TODO: remove Google Auth from this endpoint before we go live
-  def displayForm(): Action[AnyContent] =
-    (PrivateAction andThen authAction andThen authenticatedAction(subscriptionsClientId)).async { implicit request =>
+  def displayForm(): Action[AnyContent] = authenticatedAction(subscriptionsClientId).async { implicit request =>
```
rupertbates: Remove Google Auth

PR #1945, opened 2019-07-03
Prefill the US/CA state code in the Contributions checkout form

paulbrown1982:
Pre-fills the US State and Canada Province value on the Contributions flow, defaulting to the user's billing account or GEO-IP value.

## Why are you doing this?
There are thousands of recurring contributions assigned to the Alaska state due to people just picking the first item of the list, and a historical bug which set the state to Alaska back when the field was not required.

https://trello.com/c/ZexwHuvs

## Screenshots
Screen Shot 2019-07-08 at 17 29 57

## Considerations
- Check caching settings, if we're varying the checkout by country, we may cache state codes
- The new header needs adding to Fastly (tested on CODE as the header passes through from the browser)

### DIFF
--- a/support-frontend/app/controllers/Application.scala
+++ b/support-frontend/app/controllers/Application.scala
@@ -16,6 +16,7 @@ import cookies.ServersideAbTestCookie
 import com.gu.monitoring.SafeLogger
 import com.gu.monitoring.SafeLogger._
 import lib.RedirectWithEncodedQueryString
+import models.GeoData
 import play.api.mvc._
 import services.{IdentityService, MembersDataService, PaymentAPIService}
 import utils.BrowserCheck
@@ -51,7 +52,7 @@ class Application(
   }
 
   def geoRedirect: Action[AnyContent] = GeoTargetedCachedAction() { implicit request =>
-    val redirectUrl = request.fastlyCountry match {
+    val redirectUrl = request.geoData.countryGroup match {
       case Some(UK) => buildCanonicalShowcaseLink("uk")
       case Some(US) => "/us/contribute"
       case Some(Australia) => "/au/contribute"
@@ -66,7 +67,7 @@ class Application(
   }
 
   def contributeGeoRedirect(campaignCode: String): Action[AnyContent] = GeoTargetedCachedAction() { implicit request =>
-    val url = List(getRedirectUrl(request.fastlyCountry), campaignCode)
+    val url = List(getRedirectUrl(request.geoData.countryGroup), campaignCode)
       .filter(_.nonEmpty)
       .mkString("/")
 
@@ -103,7 +104,8 @@ class Application(
   ): Action[AnyContent] = maybeAuthenticatedAction().async { implicit request =>
     type Attempt[A] = EitherT[Future, String, A]
 
-    val stateCode = request.fastlyRegion
+    val geoData = request.geoData
+
     val campaignCodeOption = if (campaignCode != "") Some(campaignCode) else None
 
     // This will be present if the token has been flashed into the session by the PayPal redirect endpoint
@@ -111,12 +113,12 @@ class Application(
 
     implicit val settings: AllSettings = settingsProvider.getAllSettings()
     request.user.traverse[Attempt, IdUser](identityService.getUser(_)).fold(
-      _ => Ok(contributionsHtml(countryCode, stateCode, None, campaignCodeOption, guestAccountCreationToken)),
-      user => Ok(contributionsHtml(countryCode, stateCode, user, campaignCodeOption, guestAccountCreationToken))
+      _ => Ok(contributionsHtml(countryCode, geoData, None, campaignCodeOption, guestAccountCreationToken)),
+      user => Ok(contributionsHtml(countryCode, geoData, user, campaignCodeOption, guestAccountCreationToken))
     ).map(_.withSettingsSurrogateKey)
   }
 
-  private def contributionsHtml(countryCode: String, stateCode: Option[String], idUser: Option[IdUser],
+  private def contributionsHtml(countryCode: String, geoData: GeoData, idUser: Option[IdUser],
                                 campaignCode: Option[String], guestAccountCreationToken: Option[String])
                                (implicit request: RequestHeader, settings: AllSettings) = {
 
@@ -153,8 +155,7 @@ class Application(
       idUser = idUser,
       guestAccountCreationToken = guestAccountCreationToken,
       fontLoaderBundle = fontLoaderBundle,
-      countryCode = countryCode,
-      stateCode = stateCode
+      geoData = geoData
     )
   }
 
--- a/support-frontend/app/controllers/Application.scala
+++ b/support-frontend/app/controllers/Application.scala
@@ -19,7 +19,7 @@ import lib.RedirectWithEncodedQueryString
 import play.api.mvc._
 import services.{IdentityService, MembersDataService, PaymentAPIService}
 import utils.BrowserCheck
-import utils.RequestCountry._
+import utils.FastlyGEOIP._
 import views.{EmptyDiv, Preload}
 import scala.concurrent.{ExecutionContext, Future}
 
@@ -103,6 +103,7 @@ class Application(
   ): Action[AnyContent] = maybeAuthenticatedAction().async { implicit request =>
     type Attempt[A] = EitherT[Future, String, A]
 
+    val stateCode = request.fastlyRegion
     val campaignCodeOption = if (campaignCode != "") Some(campaignCode) else None
 
     // This will be present if the token has been flashed into the session by the PayPal redirect endpoint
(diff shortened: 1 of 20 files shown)


support-frontend/assets/helpers/user/userReducer.js
```diff
@@ -48,28 +45,13 @@ const initialState: User = {
 
 // ----- Functions ----- //
 
-function defaultStateOrProvince(countryGroup: CountryGroupId): string {
```
paulbrown1982: There are no circumstances where we would want to default a state to the first in the list, either the user must select their correct state (helped by the pre-selection from Geo-IP or their existing account), or it isn't required and it's sent through blank or null.

support-frontend/assets/pages/contributions-landing/components/ContributionFormFields.jsx
```diff
@@ -61,7 +61,7 @@ const mapStateToProps = (state: State) => ({
   lastName: getCheckoutFormValue(state.page.form.formData.lastName, state.page.user.lastName),
   email: getCheckoutFormValue(state.page.form.formData.email, state.page.user.email),
   checkoutFormHasBeenSubmitted: state.page.form.formData.checkoutFormHasBeenSubmitted,
-  state: state.page.form.formData.state,
+  state: getCheckoutFormValue(state.page.form.formData.state, state.page.user.stateField),
```
paulbrown1982: Not sure if this is needed or not, but it makes sense to be in parity with the other user-based fields.

support-frontend/assets/helpers/user/user.js
```diff
@@ -163,6 +166,9 @@ const init = (dispatch: Function, actions: UserSetStateActions = defaultUserActi
     });
   } else if (emailFromBrowser) {
     dispatch(setEmail(emailFromBrowser));
+    dispatch(setStateField(window.guardian.geoip.stateCode));
```
paulbrown1982: Note, we are not persisting the state in the browser session.

support-frontend/app/controllers/Application.scala
```diff
@@ -150,7 +153,9 @@ class Application(
       paymentApiPayPalEndpoint = paymentAPIService.payPalCreatePaymentEndpoint,
       existingPaymentOptionsEndpoint = membersDataService.existingPaymentOptionsEndpoint,
       idUser = idUser,
-      guestAccountCreationToken = guestAccountCreationToken
+      guestAccountCreationToken = guestAccountCreationToken,
+      fontLoaderBundle = fontLoaderBundle,
```
jlieb10: was this reordering intentional?

paulbrown1982: Yes, the order now matches the class's definition.

support-frontend/assets/helpers/user/user.js
```diff
@@ -139,9 +140,11 @@ const init = (dispatch: Function, actions: UserSetStateActions = defaultUserActi
     dispatch(setFirstName(window.guardian.user.firstName));
     dispatch(setLastName(window.guardian.user.lastName));
     dispatch(setFullName(`${window.guardian.user.firstName} ${window.guardian.user.lastName}`));
+    // default value from Identity Billing Address, or Fastly GEO-IP
+    dispatch(setStateField(window.guardian.user.address4 || window.guardian.geoip.stateCode));
     dispatch(setIsSignedIn(true));
     dispatch(setEmailValidated(getEmailValidatedFromUserCookie(cookie.get('GU_U'))));
-  } else if (userAppearsLoggedIn) {
+  } else if (userAppearsLoggedIn) { // TODO - remove in another PR as this condition is deprecated
```
jlieb10: Why is this condition deprecated? We are using it to surface a 'sign in' CTA on the thank you page

paulbrown1982: It's the condition (if statement) based on this value, not the value that's deprecated, because routes.oneOffContribAutofill does not exist.

jlieb10: Got it, thanks

support-frontend/app/views/contributions.scala.html
```diff
@@ -36,12 +39,17 @@
...
-    @for(fields <- user.privateFields; firstName <- fields.firstName; lastName <- fields.secondName) {
-        firstName: "@firstName",
-        lastName: "@lastName",
+      @for(fields <- user.privateFields) {
+        @for(firstName <- fields.firstName; lastName <- fields.secondName) {
+          firstName: "@firstName",
+          lastName: "@lastName",
+        }
+        @for(address4 <- fields.address4) {
+          address4: "@address4",
```
jlieb10: What is `address4`?

support-frontend/assets/helpers/user/user.js
jlieb10: Should this be `window.guardian.user.address4 || window.guardian.geoip.stateCode` in all of the places we are calling dispatch?

support-frontend/app/views/contributions.scala.html
paulbrown1982: It's the Identity API JSON model key name for where the State value is stored.

support-frontend/assets/helpers/user/user.js
paulbrown1982: emailFromBrowser does not guarantee that there's a `window.guardian.user` set of values. If that data's there then it'll be captured by `windowHasUser` condition above.

support-frontend/assets/helpers/user/user.js
```diff
@@ -163,6 +166,9 @@ const init = (dispatch: Function, actions: UserSetStateActions = defaultUserActi
     });
   } else if (emailFromBrowser) {
```
tjsilver: Could this if statement go inside the else so that `dispatch(setStateField(window.guardian.geoip.stateCode));` is not repeated?

tjsilver: Could this if statement nest inside the else below so that `dispatch(setStateField(window.guardian.geoip.stateCode));` is not repeated?

support-frontend/assets/helpers/user/user.js
jlieb10: Thanks

support-frontend/app/models/GeoData.scala
```diff
@@ -0,0 +1,19 @@
+package models
+
+import com.gu.i18n.{Country, CountryGroup}
+
+case class GeoData(countryCode: Option[String], stateCode: Option[String]) {
+  def countryGroup: Option[CountryGroup] = countryCode.flatMap(CountryGroup.byFastlyCountryCode)
+  def country: Option[Country] = countryCode.flatMap(CountryGroup.countryByCode)
+  def validatedStateNameForCountry: Option[String] = for {
```
tomrf1: `validatedStateNameForCountry` isn't being used, do we need it here?

support-frontend/app/models/GeoData.scala
```diff
@@ -0,0 +1,19 @@
...
+    c <- country
+    r <- stateCode
+    stateName <- c.statesByCode.get(r) orElse c.statesByCode.values.find(_ == r)
+  } yield stateName
+  def validatedStateCodeForCountry: Option[String] = for {
+    c <- country
+    r <- stateCode
+    stateName <- c.statesByCode.keySet.find(_ == r) orElse c.statesByCode.map(_.swap).get(r)
```
tomrf1: why do we need to support lookups against name as well?

paulbrown1982: Ah, good spot, I may have got confused with the Australia states that use the code as the name, but that's in the def now. I'll do some more AU testing to confirm I can remove this.

support-frontend/app/models/GeoData.scala
paulbrown1982: I added this to make the API more useful. I can remove it if you think that's better.

paulbrown1982: I've removed this.

support-frontend/app/models/GeoData.scala
paulbrown1982: I've removed this

support-frontend/assets/pages/contributions-landing/components/ContributionFormFields.jsx
paulbrown1982: Just tested some more in CODE, and this should be removed as the stateField will change depending on what options the user chooses, but none of the other user fields will.

support-frontend/app/views/contributions.scala.html
```diff
@@ -36,12 +39,20 @@
...
       }
-    @for(fields <- user.privateFields; firstName <- fields.firstName; lastName <- fields.secondName) {
-        firstName: "@firstName",
-        lastName: "@lastName",
+      @for(fields <- user.privateFields) {
+        @for(firstName <- fields.firstName; lastName <- fields.secondName) {
+          firstName: "@firstName",
+          lastName: "@lastName",
+        }
+        @for(address4 <- fields.billingAddress4 orElse fields.address4) {
```
paulbrown1982: We need to add extra privileges to our API key to retrieve this value properly (certainly in DEV/CODE).

support-frontend/assets/helpers/user/userActions.js
```diff
@@ -24,7 +24,7 @@ export type UserSetStateActions = {|
   setLastName: string => Action,
   setFullName: string => Action,
   setEmail: string => Action,
-  setStateField: string => Action,
+  setStateField: string => (Action | (Function => void)),
```
tomrf1: for clarity this line should now be moved to the end, below the comment about page state

support-frontend/assets/pages/contributions-landing/setUserStateActions.js
```diff
@@ -14,10 +18,26 @@ const setIsRecurringContributor = (): ((Function) => void) =>
     dispatch(setFormSubmissionDependentValue(() => ({ type: 'SET_IS_RECURRING_CONTRIBUTOR' })));
   };
 
-const setUserStateActions = {
-  ...defaultUserActionFunctions,
-  setIsSignedIn,
-  setIsRecurringContributor,
+const setStateFieldSafely = (pageCountryGroupId: CountryGroupId) =>
```
tomrf1: 👍

PR #1954, opened 2019-07-05
Redirect if campaign code is invalid

tomrf1:
## Why are you doing this?
During campaigns we link readers to e.g. `/us/contribute/toxicamerica`.
Validation of the campaign code at the end of the url path is done client-side.

This change ensures that if a campaign code is invalid or discontinued then we redirect to `/us/contribute`. Note that it's a client-side routing redirect rather than a proper redirect, but the url in the browser will be updated.

### DIFF
--- a/support-frontend/assets/pages/contributions-landing/components/ContributionFormContainer.jsx
+++ b/support-frontend/assets/pages/contributions-landing/components/ContributionFormContainer.jsx
@@ -5,7 +5,7 @@
 import React from 'react';
 import { connect } from 'react-redux';
 import { Redirect } from 'react-router-dom';
-import { type CountryGroupId } from 'helpers/internationalisation/countryGroup';
+import { type CountryGroupId, countryGroups } from 'helpers/internationalisation/countryGroup';
 import { type PaymentAuthorisation } from 'helpers/paymentIntegrations/readerRevenueApis';
 import DirectDebitPopUpForm from 'components/directDebit/directDebitPopUpForm/directDebitPopUpForm';
 import ContributionTicker from 'components/ticker/contributionTicker';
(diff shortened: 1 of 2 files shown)


support-frontend/assets/pages/contributions-landing/components/ContributionFormContainer.jsx
```diff
@@ -100,49 +101,61 @@ function withProps(props: PropTypes) {
     ...campaign || {},
   };
 
-  return props.paymentComplete ?
```
tomrf1: this diff is quite confusing but essentially I've just replaced the ternary operator and added a new condition that checks `campaignCodeParameter`

jlieb10: definitely prefer this for readability!

support-frontend/assets/pages/contributions-landing/components/ContributionFormContainer.jsx
```diff
@@ -100,49 +101,61 @@ function withProps(props: PropTypes) {
...
-            <h1 className="gu-content__blurb-header">{countryGroupDetails.headerCopy}</h1>
-          </div>
-          { countryGroupDetails.contributeCopy ?
-            <p className="gu-content__blurb-blurb">{countryGroupDetails.contributeCopy}</p> : null
-          }
+    return (<Redirect to={props.thankYouRoute} push={false} />);
+  }
+
+  if (props.campaignCodeParameter && !campaign) {
+    // A campaign code was supplied in the url path, but it's not a valid campaign
```
jlieb10: Very helpful

jlieb10:
Well done doing this client-side with react router - I initially thought this had to be accomplished with a play redirect 👍

PR #1955, opened 2019-07-05
Switch to using the new GW checkout for all traffic

rupertbates:
## Why are you doing this?
We want to switch over to using the new Guardian Weekly checkout for all traffic.

[**Trello Card**](https://trello.com/c/Gw0EzkE0/2451-launch-gw)

### DIFF
--- a/support-frontend/assets/pages/weekly-subscription-landing/components/weeklyForm.jsx
+++ b/support-frontend/assets/pages/weekly-subscription-landing/components/weeklyForm.jsx
@@ -6,10 +6,7 @@ import {
   billingPeriodTitle,
   weeklyBillingPeriods,
 } from 'helpers/billingPeriods';
-import { type CommonState } from 'helpers/page/commonReducer';
-import { getWeeklyCheckout } from 'helpers/externalLinks';
 import { sendTrackingEventsOnClick } from 'helpers/subscriptions';
-import { getPromoCode } from 'helpers/flashSale';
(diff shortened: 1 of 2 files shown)


support-frontend/assets/pages/weekly-subscription-landing/components/weeklyForm.jsx
```diff
@@ -23,28 +20,8 @@ import { getOrigin } from 'helpers/url';
...
-    referrerAcquisitionData,
-    billingPeriod,
-    countryGroupId,
-    abParticipations,
-    optimizeExperiments,
-    (billingPeriod === 'Annual' ? getPromoCode('GuardianWeekly', countryGroupId, '10ANNUAL') : null),
-  );
-
-};
+const getCheckoutUrl = (billingPeriod: WeeklyBillingPeriod): string =>
```
jfsoul: do we also need to pass through existing parameters?  Related to this current card where we haven't done that for print https://trello.com/c/MaLGg6Xq/2443-utm-parameters-not-passed-through-to-print-checkout-breaking-channel-reporting

rupertbates: I've put a note on that card to say that whatever the solution is it needs to be applied to GW checkout as well. I'd rather do that than make a change here without investigating fully.

jfsoul: Sounds good.

PR #1959, opened 2019-07-09
Persist params through to checkout from print landing page

lucymonie:
## Why are you doing this?
There seems to be an issue with utm params not getting passed through from the print landing page to the checkout as detailed in this [**Trello Card**](https://trello.com/c/MaLGg6Xq/2443-utm-parameters-not-passed-through-to-print-checkout-breaking-channel-reporting).

The theory is that the session may be disrupted as people pass through identity on their way from the landing page to checkout, which means we need to persist params from the landing page over to checkout.

## Changes
The print checkout url builder now includes all params rather than just fulfilment and product details.


rupertbates:
We will need to do this for GW as well. Did you talk to Joel about whether this is necessary?

PR #1571, opened 2019-02-27
Remove ssr test

johnduffell:
## Why are you doing this?

We ran the SSR (server side rendering) test and got +9%.  It has been on 100% but it was still there in a hacky way with a duplicate template.

In the process of combining it I realised that I was adding more specialised parameters to an already 12 specialised parameters "main" template.  Therefore I went through and removed all the redundant/over specific parameters and implemented it at a better level of abstraction.  This should make things easier in future.

I went through the routes file and tested all the main ones, the only one that looked weird was a reduandant one.

@jranks123

PR #1960, opened 2019-07-09
Authenticate users using new version of identity-play-auth

Mullefa:
This PR builds on #1930 to use the (c.f.) published version of [#1571](https://github.com/guardian/identity/pull/1571) in the identity repo.

To summarise: the new version of identity-auth-play is used to authenticate a user via the identity API instead of statically by 'just' verifying the signature of the cookie / crypto access token. This means that when a user's session is invalidated, it is invalidated for all Guardian sub domains (c.f. work on other Guardian client applications).

# Testing locally

#### Step 1
not-signed-in-subscription

#### Step 2
not-signed-in-subscription-2

#### Step 3
not-signed-in-subscription-3

#### Step 1
not-signed-in-subscription

#### Step 2
not-signed-in-subscription-3


support-frontend/app/actions/AsyncAuthenticatedBuilder.scala
```diff
@@ -15,7 +15,7 @@ import scala.concurrent.{ExecutionContext, Future}
 // TODO: considering porting this to identity-play-auth.
 // scalastyle:on
 class AsyncAuthenticatedBuilder[U](
-    userinfo: RequestHeader => Future[Option[U]],
+    userinfo: RequestHeader => Future[U],
```
Mullefa: No need to have a future of an option; failed future can be used to model no authenticated user.

support-frontend/app/controllers/Application.scala
```diff
@@ -7,7 +7,7 @@ import cats.data.EitherT
 import cats.implicits._
 import com.gu.i18n.CountryGroup
 import com.gu.i18n.CountryGroup._
-import com.gu.identity.play.IdUser
```
Mullefa: previously identity-play-auth was shipping it's own models for identity data, but if these are getting imported then identity-model should be used!

support-frontend/app/controllers/Application.scala
```diff
@@ -112,7 +112,7 @@ class Application(
     val guestAccountCreationToken = request.flash.get("guestAccountCreationToken")
 
     implicit val settings: AllSettings = settingsProvider.getAllSettings()
-    request.user.traverse[Attempt, IdUser](identityService.getUser(_)).fold(
+    request.user.traverse[Attempt, IdUser](user => identityService.getUser(user.user)).fold(
```
Mullefa: syntax is now `user.user` since I've removed an implicit def converting one type of a user to another

build.sbt
```diff
@@ -47,6 +47,8 @@ lazy val commonSettings = Seq(
     url("https://github.com/guardian/support-frontend"),
     "scm:git:git@github.com:guardian/support-frontend.git"
   )),
+  // https://www.scala-sbt.org/1.x/docs/Cached-Resolution.html
+  updateOptions := updateOptions.value.withCachedResolution(true)
```
johnduffell: thanks for the link, this is described as experimental so how do you find the benefits? does it look like this will be made a full feature soon?

support-frontend/app/models/identity/requests/CreateGuestAccountRequestBody.scala
```diff
@@ -22,7 +22,10 @@ object CreateGuestAccountRequestBody {
 
   def fromEmail(email: String): CreateGuestAccountRequestBody = CreateGuestAccountRequestBody(email, PublicFields(Some(guestDisplayName(email))))
 
-  implicit val writesCreateGuestAccountRequestBody: Writes[CreateGuestAccountRequestBody] = Json.writes[CreateGuestAccountRequestBody]
+  implicit val writesCreateGuestAccountRequestBody: Writes[CreateGuestAccountRequestBody] = {
+    import com.gu.identity.model.play.WritesInstances.publicFieldsWrites
```
Mullefa: identity-model-play has been added to provide type classes to serialise / deserialise case classes in identity-model using Play JSON

support-frontend/app/services/AuthenticationService.scala
```diff
@@ -1,38 +1,69 @@
...
+  case class Token(tokenText: String) extends AccessCredentials
 }
+case class IdMinimalUser(id: String, displayName: Option[String])
+case class AuthenticatedIdUser(credentials: AccessCredentials, user: IdMinimalUser)
 
 // TODO: consider porting this to identity-play-auth.
 class AsyncAuthenticationService(
-  authenticationService: AuthenticationService,
-  testUserService: TestUserService
+    identityPlayAuthService: IdentityPlayAuthService,
```
Mullefa: This is the major change in this PR: the `IdentityPlayAuthService` now authenticates users by calling identity API.

support-frontend/app/views/contributions.scala.html
```diff
@@ -39,17 +39,15 @@
...
-          firstName: "@firstName",
-          lastName: "@lastName",
-        }
-        @for(address4 <- fields.billingAddress4 orElse fields.address4) {
-          address4: "@address4",
-        }
-        @for(country <- fields.billingCountry orElse fields.country) {
-          country: "@country"
-        }
+      @for(firstName <- user.privateFields.firstName; lastName <- user.privateFields.secondName) {
```
Mullefa: `User` from identity-model is now being used (instead of `User` from identity-play-auth); in particular, the `privateFields` field is no longer modelled as an option (which is why a nested for comprehension is no longer necessary).

support-frontend/app/controllers/CreateSubscription.scala
```diff
@@ -48,7 +48,7 @@ class CreateSubscription(
     implicit request: AuthRequest[CreateSupportWorkersRequest],
     validator: CreateSupportWorkersRequest => Boolean
   ): Future[Result] = {
-    SafeLogger.info(s"[${request.uuid}] User ${request.user.id} is attempting to create a new ${request.body.product} subscription")
+    SafeLogger.info(s"[${request.uuid}] User ${request.user.user.id} is attempting to create a new ${request.body.product} subscription")
```
johnduffell: A nit pick that would probably be more work than it's worth to do anything with:
I find these ambiguous situations a bit confusing, they are both users, but is there a way of showing their distinguishing features in the name?
Not sure if it applies in this case, but a technique I use in value classes sometimes is to call the underlying value `value` and the actual class represents the actual concept
`case class User(value: String)`
rather than
`case class User(user: String)`

It does mean you end up with loads of `something.value` which ends up losing a bit of type safety, so there would be other solutions like `user.userString`

Mullefa: The reason why we now have `user.user` littered over the code base is because I've stopped using the [implicit def](https://github.com/guardian/identity-auth/blob/cb625a5f28ca3a07512a1cf721a0e1c2d95b5314/src/main/scala/com/gu/identity/play/AuthenticatedIdUser.scala#L28) which converts a `AuthenticatedUser` to an `IdMinimalUser`. I could do a rename e.g. instances of `AuthenticatedUser` could be renamed to `authenticatedUser` (?)

support-frontend/app/controllers/CreateSubscription.scala
```diff
@@ -85,8 +85,15 @@ class CreateSubscription(
       deliveryAddress = request.deliveryAddress,
       telephoneNumber = request.telephoneNumber,
       allowMembershipMail = false,
-      allowThirdPartyMail = user.statusFields.flatMap(_.receive3rdPartyMarketing).getOrElse(false),
-      allowGURelatedMail = user.statusFields.flatMap(_.receiveGnmMarketing).getOrElse(false),
+      // Previously the values for the fields allowThirdPartyMail and allowGURelatedMail
+      // were derived by looking for the fields: statusFields.receive3rdPartyMarketing and
+      // statusFields.receiveGnmMarketing in the JSON object that models a user.
+      // However, a query of the identity database indicates that these fields aren't defined for any users
```
johnduffell: nice work working all that out, so we were always writing false into Salesforce, regardless of people's actual preference?  And I think there is a job in membership-workflow that writes changes to these fields back into identitty and possibly vice versa.  I'm not sure if that job would be making things worse or better!  I'm not sure if those fields are used by the call centre in salesforce, and whether they affect braze!
Not sure if anyone else knows more about this off hand?

build.sbt
Mullefa: I _think_ it's decreased the build times, but that could just be psychological 😅

Not sure when it will be made a full-feature; seems to be fairly prominent across guardian sbt projects. But also happy to remove.

support-frontend/app/controllers/CreateSubscription.scala
Mullefa: Yeah, for `receiveGnmMarketing` I ran the following query against the PROD database:
```sql
select *
from users
where jdoc #>> '{statusFields,receiveGnmMarketing}' is not null
```
and it returned no results; same for `receive3rdPartyMarketing`. So I'd be surprised if it wasn't false for all cases. 

This is the one of the reasons why I wanted to remove the user models from identity-play-auth; its just another place where we have to keep the models up to date. The previously used [`StatusFields`](https://github.com/guardian/identity-auth/blob/1c6869141955b3da31675a3200c607f6ebeee2eb/src/main/scala/com/gu/identity/play/IdUser.scala#L5) were last updated 4 years ago.

support-frontend/app/services/AuthenticationService.scala
```diff
@@ -1,38 +1,69 @@
...
+  import AsyncAuthenticationService._
+
+  def authenticateUser(requestHeader: RequestHeader): Future[AuthenticatedIdUser] =
+    identityPlayAuthService.getUserFromRequest(requestHeader)
+      .map { case (credentials, user) => buildAuthenticatedUser(credentials, user) }
+      .unsafeToFuture()
+
+  def tryAuthenticateUser(requestHeader: RequestHeader): Future[Option[AuthenticatedIdUser]] =
+    authenticateUser(requestHeader)
+      .map(user => Option(user))
```
johnduffell: Option.apply handles `null` into `None`, is that something were expecting or is this better as 
```suggestion
      .map(user => Some(user))
```
or even
```suggestion
      .map(Some.apply)
```

support-frontend/app/services/AuthenticationService.scala
```diff
@@ -1,38 +1,69 @@
...
+      .handleError(_ => None) // TODO: log error?
```
johnduffell: not sure, for refernce this is eventually used to decide whether to prefill the email address on contributions landing page here https://github.com/guardian/support-frontend/pull/1960/files#diff-04600d093ee81c82983e9c9f0a0ef971R115
Contributions will then immediately do another identity request at that point (getUser) (and also discards any error without specific logging)
We should probably be concerned if that ratio dropped significantly even if it's an expected situation,  My feeling is it maybe wouldn't hurt to log this at info level?  But it does raise the question, how would we know if this was broken.

johnduffell: we should also avoid the two identity api requests in the long term, perhaps they can be consolidated in due course.  cc @tomrf1

support-frontend/app/services/AuthenticationService.scala
Mullefa: I think it's reasonable to treat `null` values as `None`, although I wouldn't expect a `null` to ever be returned.

build.sbt
johnduffell: fair enough, not seen it before, keep it in if it's well used here! I just wanted to know more about the decision process.

support-frontend/app/controllers/CreateSubscription.scala
johnduffell: right got it, getting rid of implicit conversions is better than not having clutter so it's surely the right decision you've made!
In that case yes either rename as you suggest and/or rename the second `.user` to be `minimalUser` whatever makes more sense.

support-frontend/app/services/AuthenticationService.scala
johnduffell: I agree with the first part, just when I see Option(blah) I always assume we're calling a java api that can return null and we're defending against that.
In second reading you're actually directly mapping both sides of a future to their respective sides of an option and removing the possibliity of a failed future.
Which might be better represented like replace the whole thing with this?
```
authenticateUser(requestHeader)
  .recover(failableUser => Success(failableUser.toOption))
```

support-frontend/app/controllers/CreateSubscription.scala
Mullefa: Ok, renaming to `minimalUser` as a stop-gap. Hopefully in the future, these case classes can get edited.

support-frontend/app/services/AuthenticationService.scala
Mullefa: I don't think that would compile since `authenticateUser()` doesn't return an option (?) What we want is method takes two functions: `Throwable => B` and `A => B` but I was unable to find such a method 🤷‍♂️

support-frontend/app/services/AuthenticationService.scala
Mullefa: It's a good point re 'how would we know if it this was broken'. I've added an info level log and also a TODO  noting that we might want to change this in future.

Mullefa:
@johnduffell I don't think rate-limiting will be an issue, but I'll double-check to be safe. Before deploying, I'll also (1) test on CODE; and (2) check that the scopes of the support-frontend access token give it permission to access the user fields used by the support-frontend application.

support-frontend/app/services/AuthenticationService.scala
johnduffell: sorry I have no idea why I wrote recover, I meant `transform` which does exactly that as far as I remember

Mullefa: Hmm not quite... `transform` either takes `Try[A] => Try[B]` or 2 functions: `Throwable => Throwable` and `A => B`, so I could do:
```scala
authenticateUser(requestHeader)
      .transform(user => Success(user.toOption))
```
but I'd probably have to introduce pattern matching to log the error so might end up being more verbose / less clear.

johnduffell: oh yeah I forgot about the adding logging thing!  I think at this point we're probably discussing whether it's quicker to take the motorway around the town or take the back route through the side streets! 👍 interesting discussion though, thanks

Mullefa: Ha, true! I do think it's interesting though that there isn't a method which does what we want; and it's not the only time I've had a need for something like this... Might raise with cats community and see what they say.

PR #1962, opened 2019-07-10
Adds tracking for support page CTAs

lucymonie:
## Why are you doing this?
The aim is to improve visibility of clicks on support page call-to-action buttons.

[**Trello Card**](https://trello.com/c/AZktsdvl/2454-dev-add-ga-tracking-to-subscribe-and-contribute-cta-buttons-in-why-support-page)

### DIFF
--- a/support-frontend/assets/pages/showcase/components/ctaContribute.jsx
+++ b/support-frontend/assets/pages/showcase/components/ctaContribute.jsx
@@ -9,6 +9,7 @@ import ArrowRightStraight from 'components/svgs/arrowRightStraight';
 
 import WithSupport from 'components/svgs/withSupport';
 import OneMillionCircles from 'components/svgs/oneMillionCircles';
+import { sendClickedEvent } from 'helpers/tracking/clickTracking';
 
 export default function CtaContribute() {
   return (
@@ -43,6 +44,7 @@ export default function CtaContribute() {
               aria-label={null}
               appearance="secondary"
               href="/contribute"
+              onClick={() => sendClickedEvent('support_page_cta_contribute')()}
             >
             Make a Contribution
             </AnchorButton>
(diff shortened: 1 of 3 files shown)


support-frontend/assets/pages/showcase/components/ctaSubscribe.jsx
```diff
@@ -22,7 +23,14 @@ export default function CtaSubscribe() {
         </p>
       </Text>
       <NarrowContent>
-        <AnchorButton aria-label={null} icon={<ArrowRightStraight />} href="/subscribe">Choose a Subscription</AnchorButton>
+        <AnchorButton
+          aria-label={null}
+          icon={<ArrowRightStraight />}
+          href="/subscribe"
+          onClick={() => sendClickedEvent('support_page_cta_subscribe')()}
```
rupertbates: There is also a function called `sendTrackingEventsOnClick` in subscriptions.js which does a very similar thing but also sends data to Ophan. Not sure which is the best bet in this case, probably the one you've picked?

lucymonie: Yes - I saw that function when I was reviewing your PR! I think it's less suitable as it wants you to specify a product as opposed to clicking on 'subscribe' or 'contribute'

support-frontend/assets/pages/showcase/components/ctaContribute.jsx
```diff
@@ -42,6 +43,7 @@ export default function CtaContribute() {
               icon={<ArrowRightStraight />}
               appearance="secondary"
               href="/contribute"
+              onClick={() => sendClickedEvent('support_page_cta_contribute')()}
```
rupertbates: Are you going to put a similar event on the subscribe button?

lucymonie: Yes, I have done :-)

lucymonie: See below :down

rupertbates: 🤦‍♂ Oh yeah!

PR #1965, opened 2019-07-11
Remove the SubscriberKey from email fields

rupertbates:
## Why are you doing this?
We no longer need to send the SubscriberKey field in the email JSON as this was an ExactTarget thing and now we use Braze, it is not needed.

This PR: https://github.com/guardian/membership-workflow/pull/194 removed it from membership-workflow


johnduffell:
for some reason the post deploy tests for this pr appeared on the one merged 4 minutes later https://github.com/guardian/support-frontend/pull/1967#issuecomment-510831882

Also they failed, which could have been because of the bug @damien-shaw-guardian and I are looking into with the deploy causing blank pages temporarily.

PR #1967, opened 2019-07-11
Fix the terms and conditions link for print checkout

lucymonie:
## Why are you doing this?
Because the current print checkout has terms and conditions that apply to GW. This updates the link to the print T&C.

[**Trello Card**](https://trello.com/c/3wn616jG/2444-tc-link-on-print-only-checkout-pointing-to-gw-terms)


johnduffell:
good spot there! 👍

PR #1969, opened 2019-07-12
Add ophan tracking to why support page and header nav

lucymonie:
## Why are you doing this?
To make sure that we have GA tracking and Ophan data collection set up for the why support page CTAs and also for the support header navbar.

[**Trello Card**](https://trello.com/c/FjJI3VaI/2455-tracking-audit-implement-tags-to-allow-reporting-from-datalake)

### DIFF
--- a/support-frontend/assets/pages/showcase/components/ctaSubscribe.jsx
+++ b/support-frontend/assets/pages/showcase/components/ctaSubscribe.jsx
@@ -7,8 +7,9 @@ import Text from 'components/text/text';
 import GridImage from 'components/gridImage/gridImage';
 import AnchorButton from 'components/button/anchorButton';
 import ArrowRightStraight from 'components/svgs/arrowRightStraight';
+import { trackComponentClick } from 'helpers/tracking/ophan';
 
-export default function CtaSubscribe(props: {clickHandler: Function}) {
+export default function CtaSubscribe() {
   return (
     <Content
       appearance="feature"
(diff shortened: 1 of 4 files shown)


support-frontend/assets/pages/showcase/showcase.jsx
```diff
@@ -26,6 +26,15 @@ import ConsentBanner from 'components/consentBanner/consentBanner';
...
 import './showcase.scss';
 import { Provider } from 'react-redux';
+import { sendClickedEvent } from 'helpers/tracking/clickTracking';
+import { trackComponentClick } from 'helpers/tracking/ophan';
+
+// ----- Tracking ----- //
+
+const clickHandler = (ctaType) => {
+  sendClickedEvent(`support_page_cta_${ctaType}`)();
+  trackComponentClick(`support-page-cta-${ctaType}`);
```
rupertbates: I think you can just call `trackComponentClick` because this calls GA as well as Ophan. Which means that you don't need this function

lucymonie: Great, thanks Rupert!

support-frontend/assets/pages/showcase/components/ctaSubscribe.jsx
```diff
@@ -26,7 +26,7 @@ export default function CtaSubscribe() {
         <AnchorButton
           icon={<ArrowRightStraight />}
           href="/subscribe"
-          onClick={() => sendClickedEvent('support_page_cta_subscribe')()}
```
johnduffell: as discussed it does seem a bit confusing that the old one had the extra set of `()` wheras the new one doesn't.  Although the new function does look like things are more sensible!  As discussed it might always be worth checking quickly.

PR #1972, opened 2019-07-16
add logging for which stripe account we're using

johnduffell:
## Why are you doing this?

We have had a few issues with the checkout using the wrong stripe account.  It seemed to make sense to add logging on the step functions to say which account it decided to use.

## Changes

* add logging to stripe for getting the right account from the currency
* add logger to support config  project
* update test names
* make logger work when running the tests locally

## Screenshots

image


support-config/src/main/scala/com/gu/support/config/SafeLogger.scala
```diff
@@ -0,0 +1,46 @@
+package com.gu.support.config
+
+import com.typesafe.scalalogging.StrictLogging
+import org.slf4j.{Marker, MarkerFactory}
+
+object SafeLogger extends StrictLogging {
```
rupertbates: This class already exists in `support-services`. Since you are adding it here and support-services depends on this library I think you can remove it from there.

support-workers/src/test/resources/logback.xml
```diff
@@ -1,6 +1,6 @@
 <configuration>
 
-    <appender name="STDOUT" class="io.symphonia.lambda.logging.DefaultConsoleAppender">
```
rupertbates: If we are no longer using the symphonia library we can remove it from the dependencies. Have you tested that this works ok with writing to CloudWatch?

support-config/src/main/scala/com/gu/support/config/SafeLogger.scala
johnduffell: good idea

support-workers/src/test/resources/logback.xml
johnduffell: this is the "src/test/" version, the normal one is still the same.

support-config/src/main/scala/com/gu/support/config/SafeLogger.scala
johnduffell: I have also moved the new one to the equivalent old package to avoid messing up the old code's important - that's pushed now

johnduffell:
this is working
```
[2019-07-17 10:40:48.655] 22be8e27-0126-4613-8c2b-0d9ef15e5700 INFO c.g.m.SafeLogger$ - StripeConfig: getting AU stripe account for Some(AUD)
```
and
```
[2019-07-17 10:41:44.780] 55f37680-f42e-455b-bb2b-7fd160d54ea6 INFO c.g.m.SafeLogger$ - StripeConfig: getting default stripe account for Some(EUR)
```

PR #1976, opened 2019-07-17
Stop two separate Ophan pageviews being fired per page load

rupertbates:
## Why are you doing this?
I have spotted that we are firing two Ophan page views per page with different viewIds. 
This is down to the way that Ophan is triggered just by importing it into a file: Because the googleTagManagerScript.js file imported the `doNotTrack` function from page.js (which fires the first page view) it meant that it was being fired twice.

[**Trello Card**](https://trello.com/c/pGaBYCf5/2483-investigate-double-counting-in-ophan-page-views)

### DIFF
--- a/support-frontend/assets/helpers/page/page.js
+++ b/support-frontend/assets/helpers/page/page.js
@@ -4,26 +4,47 @@
 
 import 'ophan';
 import type { Store } from 'redux';
-import { applyMiddleware, combineReducers, compose, createStore, type Reducer } from 'redux';
+import {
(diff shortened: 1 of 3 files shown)


support-frontend/assets/helpers/page/page.js
```diff
@@ -4,33 +4,55 @@
 
 import 'ophan';
```
johnduffell: worth adding a comment to this line, if it's having unexpected side effects?

PR #1977, opened 2019-07-18
Reading single contributor cookie and putting in state

jlieb10:
## Why are you doing this?
We want to know if a user has given a single contribution before so we can customize messaging on the landing and thank you pages. This is the first step to enable this work - we will probably also want to request information contribution history for returning contributors.

[**Trello Card**](https://trello.com/c/gnO74GFW/1270-recognize-single-contributors-on-landing-page-using-cookie)

## Changes

* Testing presence of cookie and adding to state using Redux
* Expanding user reducer and actions tests

### DIFF
--- a/support-frontend/assets/helpers/user/userReducer.js
+++ b/support-frontend/assets/helpers/user/userReducer.js
@@ -21,6 +21,7 @@ export type User = {
   isSignedIn: boolean,
   isRecurringContributor: boolean,
   emailValidated: boolean,
+  isReturningContributor: boolean,
 };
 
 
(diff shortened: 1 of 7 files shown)


support-frontend/assets/helpers/user/userReducer.js
```diff
@@ -21,6 +21,7 @@ export type User = {
   isSignedIn: boolean,
   isRecurringContributor: boolean,
   emailValidated: boolean,
+  isReturningContributor: boolean,
```
tomrf1: is the date of the contribution of any use for this upcoming test?

jlieb10: I don't think we should reveal the date or amount unless they are signed in

PR #1980, opened 2019-07-22
Tracking improvements

rupertbates:
## Why are you doing this?

Our tracking has got a bit messy as it has grown, for instance we respect consent settings and do not track headers in most places but not all, we will need to fix this as part of the upcoming consent work, and this PR makes that easier.

Also we have quite a lot of 'behavioural' tracking (tracking related to a user's behaviour on the site such as button clicks) and are about to add more to cover checkouts. This was all in the Ophan module even though it tracks through GA as well as Ophan. I've moved it into its own module 'behaviour' to make it cleaner. 

[**Trello Card**](https://trello.com/c/3gQHLqpZ/2467-consent-phase-2-dev-work)

## Changes
* Respect do not track header in all tracking
* Move all behavioural tracking to a separate module.

### DIFF
--- a/support-frontend/assets/components/headers/links/links.jsx
+++ b/support-frontend/assets/components/headers/links/links.jsx
@@ -5,7 +5,7 @@ import { routes } from 'helpers/routes';
 import { getPatronsLink } from 'helpers/externalLinks';
 import { type Option } from 'helpers/types/option';
 import { classNameWithModifiers } from 'helpers/utilities';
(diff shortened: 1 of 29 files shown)


paulbrown1982:
Just for the record, there is a PR to stop applying the DNT header for Ophan, in order for the logic to be consistent with DotCom - https://github.com/guardian/support-frontend/pull/2981

PR #1984, opened 2019-07-23
Second attempt - Authenticate users using new version of identity-play-auth

Mullefa:
Second attempt to authenticate users using new version of identity-play-auth; previous attempt was #1960. This PR only differs from #1960 in 2 places, which I'll annotate below.

The issue was that we were failing to create a guest account for a user wanting to make a recurring contribution. This was due to using the `PublicFields` model from identity-model rather than the (now removed) `PublicFields` model that was shipped with the old version of identity-play-auth. 

Specifically the old model was:
```scala
case class PublicFields(displayName: Option[String])
```
Whereas the new model is:
```scala
case class PublicFields(username: Option[String] = None, displayName: Option[String] = None, ...)
```
Therefore, whilst:
```scala
CreateGuestAccountRequestBody(email, PublicFields(Some(guestDisplayName(email))))
```
was compiling (when using the new model), the `username` field was getting set, not the `displayName` field (as this is now the second argument to the case class constructor).

Therefore, a create guest account request was failing, due to insufficient data being provided in the respective request body.


support-frontend/app/models/identity/requests/CreateGuestAccountRequestBody.scala
```diff
@@ -20,9 +20,13 @@ object CreateGuestAccountRequestBody {
 
   def guestDisplayName(email: String): String = email.split("@").headOption.getOrElse("Guest User")
 
-  def fromEmail(email: String): CreateGuestAccountRequestBody = CreateGuestAccountRequestBody(email, PublicFields(Some(guestDisplayName(email))))
+  def fromEmail(email: String): CreateGuestAccountRequestBody =
+    CreateGuestAccountRequestBody(email, PublicFields(displayName = Some(guestDisplayName(email))))
```
Mullefa: This is the first change from #1960; it fixes the issue documented in this PR.

support-frontend/test/models/identity/requests/CreateGuestAccountRequestBodyTest.scala
```diff
@@ -0,0 +1,18 @@
+package models.identity.requests
+
+import org.scalatest.{Matchers, WordSpec}
+import play.api.libs.json.Json
+
+class CreateGuestAccountRequestBodyTest extends WordSpec with Matchers {
```
Mullefa: This is the second change from #1960; it tests the fix introduced in this PR.

johnduffell:
thanks for fixing this! 👍 tough error there.

As a side note, this is partly why I like to use AnyVal case classes as much as possible to avoid having several identical parameters with the same type but different meanings.  I'm not sure if that would work well with the serialisers though.  On support-service-lambdas projects I tended to have a low level "wire" model and a higher level application domain model, where only the high level model had proper types, and the low level model was free of any defaults etc.

Mullefa:
@johnduffell - yeah I was thinking about ways to circumvent against this. AnyVal case classes are one approach and I'm sure we could handle the serialisation/deserialisation of them relatively easily (especially now there are identity-model-json and identity-model-circe libraries so developers wouldn't have to roll out their own serialisers/deserialisers per application). My only concern is the verbosity they add to the code base, but experiencing errors like this suggests that it's worth it. I've wanted to make some breaking changes to identity-model anyway to simplify some of the types used ([for example](https://github.com/guardian/identity/blob/master/identity-model/src/main/scala/com/gu/identity/model/Model.scala#L59)), so I could consider introducing AnyVals in conjunction with this.

Another option is to always used named parameters, but this is probably more verbose / less chance of sticking than AnyVals.

PR #1985, opened 2019-07-23
Checkout tracking

rupertbates:
## Why are you doing this?

We want to track the payment method that a user chooses to checkout with so that we can evaluate the drop out rate of each method.

This PR fires the `trackCheckoutSubmitAttempt` function when a user clicks the 'Proceed to payment' button tracking the product type and payment method. 

We are also now setting the paymentMethod custom dimension on the `trackCheckoutSubmitAttempt` and `trackPaymentMethodSelected` events as this makes some analyses easier. 

[**Trello Card**](https://trello.com/c/nCpd0NnR/2447-dev-add-tracking-to-digital-pack-checkout)

### DIFF
--- a/support-frontend/assets/helpers/subscriptionsForms/submit.js
+++ b/support-frontend/assets/helpers/subscriptionsForms/submit.js
@@ -241,7 +241,7 @@ function showPaymentMethod(
 
 function trackSubmitAttempt(paymentMethod: ?PaymentMethod, productType: SubscriptionProduct) {
   const componentId = `subs-checkout-submit-${productType}-${paymentMethod || ''}`;
-  trackCheckoutSubmitAttempt(componentId, componentId, paymentMethod);
+  trackCheckoutSubmitAttempt(componentId, productType, paymentMethod);
 }
 
 function submitForm(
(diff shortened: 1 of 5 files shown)


support-frontend/assets/helpers/tracking/googleTagManager.js
```diff
@@ -202,14 +202,15 @@ function successfulConversion(participations: Participations) {
...
-function gaEvent(gaEventData: GaEventData) {
+function gaEvent(gaEventData: GaEventData, additionalFields: ?Object) {
   maybeTrack(() => {
     if (window.googleTagManagerDataLayer) {
       window.googleTagManagerDataLayer.push({
         event: 'GAEvent',
         eventCategory: gaEventData.category,
         eventAction: gaEventData.action,
         eventLabel: gaEventData.label,
+        ...additionalFields,
```
jfsoul: What would this object look like in practice out of curiosity?

rupertbates:
```javascript
{
    event: "GAEvent", 
    eventCategory: "click", 
    eventAction: "payment-method-selected", 
    eventLabel: "Stripe", 
    paymentMethod: "Stripe"
}
```

support-frontend/assets/helpers/subscriptionsForms/submit.js
```diff
@@ -237,11 +239,20 @@ function showPaymentMethod(
   }
 }
 
+function trackSubmitAttempt(paymentMethod: ?PaymentMethod, productType: SubscriptionProduct) {
+  const componentId = `subs-checkout-submit-${productType}-${paymentMethod || ''}`;
+  trackCheckoutSubmitAttempt(componentId, componentId, paymentMethod);
```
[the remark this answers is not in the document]
rupertbates: Yes that's a good point, I'm going to put product type in there as that's most likely to be useful I would think.

PR #1987, opened 2019-07-24
fix: fixed getGlobal for the switches

damien-shaw-guardian:
## Why are you doing this?
This PR is all about turning Home Delivery back on, which has been switched off due to off line fulfilment issues, these issues have been resolved so Home Delivery can now be turned back on

[**Trello Card**](https://trello.com/c/XlcAd6cH/2434-turn-home-delivery-back-on-nb-fulfillment-delays-need-fixing)

## Changes

* fixed getGlobal so the switches could be obtained from the window object

## Screenshots
Screen Shot 2019-07-22 at 14 52 18


johnduffell:
good spot, missing dot 👍

PR #1995, opened 2019-07-26
Send ophan view event on landing page when cookie present

jlieb10:
## Why are you doing this?

We want to start tracking our RSC landing page views and conversions.

### DIFF
--- a/support-frontend/assets/helpers/user/defaultUserActionFunctions.js
+++ b/support-frontend/assets/helpers/user/defaultUserActionFunctions.js
@@ -2,6 +2,7 @@
 
 import { setSession } from 'helpers/storage';
 import { type Action } from './userActions';
+import { trackComponentLoad } from 'helpers/tracking/behaviour';
 
 // ----- Actions Creators ----- //
 
@@ -62,6 +63,14 @@ function setEmailValidated(emailValidated: boolean): Action {
 }
 
 function setIsReturningContributor(isReturningContributor: boolean): Action {
+  // JTL: We want to send an Ophan event when we recognize a user is a returning contributor and on the landing page
+  const isReturningContributorOnLandingPage = isReturningContributor &&
+    !!document.location.pathname.match(/^https:\/\/support\.\w+\.com\/\w\w\/contribute/);
+
+  if (isReturningContributorOnLandingPage) {
+    trackComponentLoad('returning-single-contributor-landing-page-view')
+  }
+
   return { type: 'SET_IS_RETURNING_CONTRIBUTOR', isReturningContributor };
 }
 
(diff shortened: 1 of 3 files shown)


support-frontend/assets/helpers/user/defaultUserActionFunctions.js
```diff
@@ -62,6 +63,14 @@ function setEmailValidated(emailValidated: boolean): Action {
 }
 
 function setIsReturningContributor(isReturningContributor: boolean): Action {
+  // JTL: We want to send an Ophan event when we recognize a user is a returning contributor and on the landing page
+  const isReturningContributorOnLandingPage = isReturningContributor &&
+    !!document.location.pathname.match(/^https:\/\/support\.\w+\.com\/\w\w\/contribute/);
```
tomrf1: Is it guaranteed to end with a slash?
Also, what about campaign pages?

support-frontend/assets/helpers/user/user.js
```diff
@@ -130,11 +130,11 @@ const init = (dispatch: Function, actions: UserSetStateActions = defaultUserActi
     dispatch(setPostDeploymentTestUser(true));
   }
 
-  if (getCookie('gu_recurring_contributor') === 'true') {
+  if (getCookie('gu_recurring_contributor')) {
```
tomrf1: is this a change in behaviour? What if it's `'false'`?

jlieb10: If it's false, then the code inside the block won't execute. I thought `===  true` was unnecessary - it's not a change in the behaviour

tomrf1: the comparison was against a string, not a boolean. So if the cookie value is `'false'` then I'd expect the expression to be true?

support-frontend/assets/helpers/user/defaultUserActionFunctions.js
jlieb10: I didn't include the 'end' bit of the regex (`$`) so it will match with a campaign code but it's true that it shouldn't include a slash at the end

tomrf1: ah yes 👍

support-frontend/assets/helpers/user/user.js
jlieb10: ohhh okay- got it

jlieb10: is there another possible value for this cookie? the string and bool thing is confusing

tomrf1: the value of a cookie is always a string, so we have to resolve it to a boolean.
It's set based on the response from members-data-api [here](https://github.com/guardian/frontend/blob/3337a586d533cd981f2023773c137bbbb23d1e35/static/src/javascripts/projects/common/modules/commercial/user-features.js#L64)

support-frontend/assets/helpers/user/defaultUserActionFunctions.js
jlieb10: i just looked at the code - the slash at the end here just indicates it's the end of the regex (JS regexes start and end with `/` -- the character "/" is indicated as `\/`)

jlieb10: (so this can be marked as resolved, I believe)

support-frontend/assets/helpers/user/user.js
jlieb10: got it - reverting now

PR #1999, opened 2019-07-31
QA Test for New Marketing Component

jlieb10:
## Why are you doing this?

We have a design for a new marketing component and this needs to be 'QA tested' alongside our current marketing component. We aren't looking for anything scientific in this test, just want to watch the two alongside each other for a couple days to make sure we are getting opt-ins through both. Then we will be able to switch over to the 'newer' component (with checkbox).

Doing this as an A/B test was unexpectedly complicated and involved touching a lot of files that I didn't expect to need to touch. I have added inline documentation to help with removing this as a test when that time comes.

## Changes

* Created new marketing option as `MarketingComponentWithCheckbox`
* Implemented it across contributions thank you pages as an AB test against the current `MarketingComponent`

## Screenshots
Control:
marketing abtest control

Variant ("new"):
marketing abtest variant

### DIFF
--- a/support-frontend/assets/components/marketingConsent/marketingConsent.jsx
+++ b/support-frontend/assets/components/marketingConsent/marketingConsent.jsx
@@ -13,6 +13,7 @@ import { logException } from 'helpers/logger';
 import Button from 'components/button/button';
 import NonInteractiveButton from 'components/button/nonInteractiveButton';
 import 'components/marketingConsent/marketingConsent.scss';
+import type { ThankYouPageMarketingComponentTestVariants } from 'helpers/abTests/abtestDefinitions';
 
 // ----- Types ----- //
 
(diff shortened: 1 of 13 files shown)


support-frontend/assets/components/marketingConsent/marketingConsent.jsx
```diff
@@ -13,19 +13,26 @@ import { logException } from 'helpers/logger';
...
   email: string,
   csrf: CsrfState,
-  onClick: (?string, CsrfState) => void,
+  onClick: (?string, CsrfState, ?ThankYouPageMarketingComponentTestVariants) => void,
   requestPending: boolean,
+  marketingComponentVariant?: ThankYouPageMarketingComponentTestVariants,
 |};
 
 type PropTypes = {|
-  ...ButtonPropTypes,
```
jlieb10: For some reason, doing the object spread here was causing eslint errors - I don't think it likes it when it comes to optional props

tomrf1:
I'm not sure I understand the reason for doing this as an a/b test. Surely we can tell if the new component is working or not regardless?

support-frontend/assets/components/marketingConsent/marketingConsent.jsx
tomrf1: that's annoying. At least we can revert this part later

jlieb10: 👍

PR #2005, opened 2019-08-01
Add alarm on failed user authentication

Mullefa:
## Why are you doing this?

There were some issues with user authentication, spotted by at @jfsoul; see #2003 for more context.

Motivated by this, introduce an alarm to automate reporting on this issue.

### DIFF
--- a/support-frontend/app/services/AuthenticationService.scala
+++ b/support-frontend/app/services/AuthenticationService.scala
@@ -81,6 +81,7 @@ object AsyncAuthenticationService {
     error match {
       // Utilises the custom error introduced in: https://github.com/guardian/identity/pull/1578
       case _: UserCredentialsMissingError => SafeLogger.info(s"unable to authorize user - $error")
+      // Cloud watch alarm based on this log statement.
       case _ => SafeLogger.error(scrub"unable to authorize user", error)
     }
 }
--- a/support-frontend/cloud-formation/cfn.yaml
+++ b/support-frontend/cloud-formation/cfn.yaml
@@ -506,6 +506,43 @@ Resources:
     DependsOn:
     - StateMachineUnavailableMetric
 
+  UnableToAuthorizeUserMetric:
+    Type: "AWS::Logs::MetricFilter"
+      Properties:
+        LogGroupName: !Sub support-frontend-${Stage}
+        FilterPattern: "ERROR unable to authorize user"
+        MetricTransformations:
+        - MetricNamespace: !Sub support-frontend-${Stage}
+          MetricName: unable-to-authorize-user
+          MetricValue: 1
+          DefaultValue: 0
+
+  UnableToAuthorizeUserAlarm:
+    Type: AWS::CloudWatch::Alarm
+      Condition: CreateProdResources
+      Properties:
+        AlarmActions:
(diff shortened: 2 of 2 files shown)


jfsoul:
An alternative to this would be to use sentry notifications for the support-server-side project.  Haven't looked/thought in detail but worth considering whether that does a _good enough_ job in this case.

Mullefa:
Thanks for the suggestion - is there an example you can point me to? Also, having a look at the logs over the last 12 hours, I think it'd probably be too noisy implemented as is - there are some access denied errors but I think these are completely valid e.g. the user's session could have been invalidated

support-frontend/app/services/AuthenticationService.scala
```diff
@@ -81,6 +81,7 @@ object AsyncAuthenticationService {
     error match {
       // Utilises the custom error introduced in: https://github.com/guardian/identity/pull/1578
       case _: UserCredentialsMissingError => SafeLogger.info(s"unable to authorize user - $error")
+      // Cloud watch alarm based on this log statement.
       case _ => SafeLogger.error(scrub"unable to authorize user", error)
```
johnduffell: if you want to have an alarm on a log line, it has to be quite clear, I'd recommend an error code 
```suggestion
      case _ => SafeLogger.error(scrub"AUTH_FAIL: unable to authorize user", error)
```
however @mario-galic had trouble with alarms being lumpy when using this technique of processing logs.  We think this might be because the AWS logs agent didn't always keep up with the logs being produced.
We do have a metric library in support-frontend which would mean it could be explicit and go straight into cloudwatch which I would prefer to do.

johnduffell: e.g. https://github.com/guardian/support-frontend/search?q=catalogFailureRequest&unscoped_q=catalogFailureRequest

johnduffell:
given how much luck/knowledge we have with sentry so far, I don't have faith we would keep on top of these that way.  An alarm makes more sense as we have a defined process, as long as it won't be noisy.  However I would prefer it's implemented using the metric API similar to the other places, for consistency and reliability.

Mullefa:
Thanks for the advice @johnduffell - I'll have a think to see if we can have an alarm around this that isn't too noisy, in which case I'll use the metric API.

johnduffell:
hi Guy, any more thoughts about this PR? I think it's a nice idea, we could perhaps pair on it some time.  I'm not sure why I suggested metric push api as that's for client side stuff.  WE could just use this directly https://github.com/guardian/support-frontend/blob/master/support-services/src/main/scala/com/gu/aws/AwsCloudWatchMetricPut.scala

alexflav23:
@Mullefa Is this still necessary or should we close it? It's been a while and it's been forgotten.

jfsoul:
@alexflav23 Guy is on leave for a while, but I still think there is some value in this idea - if you wanted to resurrect on his behalf? I agree with @johnduffell comment and would probably just directly put into a custom metric in cloudwatch.

rupertbates:
@Mullefa @jfsoul @johnduffell I'm going to close this PR unless someone has time to take this on this week?

Mullefa:
I defer to current team. If you don't require such an alarm, feel free to close 👍

PR #2006, opened 2019-08-02
Fix 2 fulfilment delay issues

jfsoul:
## Why are you doing this?
Fixes 2 issues with fulfilment delays for vouchers and Guardian Weekly, namely:

- Sunday was being considered as the first day of the week (day 0) when calculating the additional delay based on cutoff (e.g. extra week if it's after the fulfilment file is generated on a Wednesday for vouchers).
- The condition for the extra delay for vouchers was incorrect, the hour of day should only be considered when it's a Wednesday.

## Changes

* Change 1
* Change 2

### DIFF
--- a/support-frontend/assets/pages/weekly-subscription-checkout/helpers/deliveryDays.js
+++ b/support-frontend/assets/pages/weekly-subscription-checkout/helpers/deliveryDays.js
@@ -14,7 +14,8 @@ const getWeeklyDays = (today: ?number): Date[] => {
   const now = new Date(today || new Date().getTime());
   const currentWeekday = now.getDay();
   const weeksToAdd =
-      currentWeekday > extraDelayCutoffWeekday
+      currentWeekday > extraDelayCutoffWeekday || 
+      currentWeekday === 0 // Sunday is considered the last day of the week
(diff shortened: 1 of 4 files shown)


jfsoul:
> the hour of day should only be considered when it's a Wednesday

After making this change, I am now thinking of removing this 5am rule instead and seeing if anyone even notices @rupertbates

support-frontend/assets/pages/weekly-subscription-checkout/helpers/deliveryDays.js
```diff
@@ -14,7 +14,8 @@ const getWeeklyDays = (today: ?number): Date[] => {
   const now = new Date(today || new Date().getTime());
   const currentWeekday = now.getDay();
   const weeksToAdd =
-      currentWeekday > extraDelayCutoffWeekday
+      currentWeekday > extraDelayCutoffWeekday || 
+      currentWeekday === 0 // Sunday is considered the last day of the week
```
johnduffell: I know we discussed this a while ago, but  I still think we should align whatever the fulfilment cutoff is with the delivery cutoff week commencing, and then allow all valid days after that week.

PR #2009, opened 2019-08-02
Fix pre-rendering at build time

tomrf1:
## Why are you doing this?
Pre-rendering has been silently breaking since [this PR](https://github.com/guardian/support-frontend/pull/1932/files#diff-7c13c67470551a3cc3256ad71218daf5R23)

This is because node does not have `fetch` available.
I've checked that this functionality still works in the browser by confirming the request was sent

### DIFF
--- a/support-frontend/assets/helpers/render.js
+++ b/support-frontend/assets/helpers/render.js
@@ -3,6 +3,13 @@
 import ReactDOM from 'react-dom';
 import { logException } from 'helpers/logger';
 
+// Without this the build-time pre-rendering breaks, because fetch is undefined when running with node
+const safeFetch = (url: string, opts) => {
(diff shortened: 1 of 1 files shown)


support-frontend/assets/helpers/render.js
```diff
@@ -3,6 +3,13 @@
 import ReactDOM from 'react-dom';
 import { logException } from 'helpers/logger';
 
+// Without this the build-time pre-rendering breaks, because fetch is undefined when running with node
+const safeFetch = (url: string, opts) => {
+  if (typeof fetch !== 'undefined') {
+    fetch(url, opts)
+  }
```
johnduffell: can we have another alert that happens if fetch is not defined and we're on the client side? 😆 📈

PR #2010, opened 2019-08-02
feat: optimize test for product cards

damien-shaw-guardian:
## Why are you doing this?
We are adding a new experiment, which is to test if a change to the copy on the product option cards will improve conversion rates

[**Trello Card**](https://trello.com/c/Tm2rzQ30/2513-updates-to-cards-for-next-ab-test)

## Changes

* product option button card copy
* product option card size
* added optimize experiments to product option cards

### Old Product cards 
Screen Shot 2019-08-02 at 15 25 24

### New Product cards
Screen Shot 2019-08-02 at 15 28 46

PR #2012, opened 2019-08-05
Revert "feat: optimize test for product cards"

damien-shaw-guardian:
Reverts guardian/support-frontend#2010


tomrf1:
The post-deploy test failure is the `Check Stripe pop-up appears` test again, I need to fix this.
Have kicked off the tests again

PR #2013, opened 2019-08-05
Choice architecture test round 1 (labels test)

jlieb10:
## Why are you doing this?
We want to test the performance of adding labels to the contributions form. We will measure the performance of this test by looking at 

[**Trello Card**](https://trello.com/c/oNMxdCjp/1242-choice-architecture-r1-add-field-labels)

## Changes

* Added the test (basically was able to complete using SCSS only)
* Updated the legend on the form - applicable for both variants in this test. This change should improve accessibility even when the legend is not visible.

## Screenshots
Control: 
landingpage labels control

Variant:
landingpage labels variant

### DIFF
--- a/support-frontend/assets/pages/contributions-landing/components/ContributionAmount.jsx
+++ b/support-frontend/assets/pages/contributions-landing/components/ContributionAmount.jsx
@@ -160,7 +160,7 @@ function withProps(props: PropTypes) {
 
   return (
     <fieldset className={classNameWithModifiers('form__radio-group', ['pills', 'contribution-amount'])}>
-      <legend className={classNameWithModifiers('form__legend', ['radio-group'])}>How much would you like to contribute?</legend>
+      <legend className={classNameWithModifiers('form__legend', ['radio-group'])}>How much would you like to give?</legend>
       <ul className="form__radio-group-list">
         {validAmounts.map(renderAmount(currencies[props.currency], spokenCurrencies[props.currency], props))}
         <li className="form__radio-group-item">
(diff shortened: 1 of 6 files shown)


support-frontend/assets/pages/contributions-landing/components/ContributionAmount.jsx
```diff
@@ -160,7 +160,7 @@ function withProps(props: PropTypes) {
 
   return (
     <fieldset className={classNameWithModifiers('form__radio-group', ['pills', 'contribution-amount'])}>
-      <legend className={classNameWithModifiers('form__legend', ['radio-group'])}>Amount</legend>
+      <legend className={classNameWithModifiers('form__legend', ['radio-group'])}>How much would you like to contribute?</legend>
```
tomrf1: 💯

jlieb10:
@jranks123 @tomrf1 we are going to measure this test by comparing the volume of contributions and the volume of recurring cancellations. Can you advise on adding an analytics call somewhere to see whether or not those in the variant group are less likely to cancel a subscription?

ionamckendrick:
One copy mistake. The first line should be "How much would you like to *contribute?*"

jlieb10:
@ionamckendrick there are some versions where the first verb is give and the second is contribute and vice-versa - but the reason i did "give" for both is because "contribute" causes a line break on tablet

tomrf1:
Regarding tracking cancellations, I expect we can work that out in the analysis by matching up the identity ID with the cancellation. This assumes that we do have the cancellation data.
@jranks123 do you think this is possible?

jlieb10:
@ionamckendrick updated in screenshots

jranks123:
@jlieb10 yes, the cancellation information is in the subscriptions table. 

At what point are you planning on looking at the cancellations by the way? I would imagine that this test will be over before there is any meaningful cancellation data to look at. Are you committing to looking at some date in the future? If so, we should define exactly when that will happen upfront, and put something in place to make sure we don't forget.

It is really good that we are thinking about churn, it's something that we have not really given enough time to in the past. Figuring out how to effectively measure churn in an acquisitions test is a useful skill for us as a team to try and learn,.

jlieb10:
@jranks123 thanks!

I planned to pass along the date of the test launch to Jesse here and Jakub in the US who can add it to their calendars and decide when to start analyzing churn. We could also plan to do the analysis ourselves and also think about what a meaningful-enough amount of time is for this context.

In the meantime, we will look at the volume of contributions and the spread of contribution types to get a sense of how this test is performing. While the churn rate is something we'd like to factor in, you are right that there's no plan to run this long enough to get an immediate sense of churn.

PR #2015, opened 2019-08-05
Post-deploy test: check all instances of Stripe element

tomrf1:
## Why are you doing this?
This test periodically fails. I can't be certain, but I think it might be because there is always more than one stripe checkout element. So when it checks if the element is visible, it needs to be check each one.

I also want to find out why there's more than one of these elements...

### DIFF
--- a/support-frontend/test/selenium/util/Browser.scala
+++ b/support-frontend/test/selenium/util/Browser.scala
@@ -1,14 +1,16 @@
 package selenium.util
 
+import com.typesafe.scalalogging.LazyLogging
 import org.openqa.selenium.WebDriver
 import org.openqa.selenium.support.ui.ExpectedConditions.numberOfWindowsToBe
 import org.openqa.selenium.support.ui.{ExpectedCondition, ExpectedConditions, WebDriverWait}
 import org.scalatest.selenium.WebBrowser
+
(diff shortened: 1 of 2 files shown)


support-frontend/test/selenium/util/Browser.scala
```diff
@@ -20,6 +21,12 @@ trait Browser extends WebBrowser {
   def pageHasElement(q: Query): Boolean =
     waitUntil(ExpectedConditions.visibilityOfElementLocated(q.by))
 
+  // If there is more than 1 element matching the query then only one of them needs to be visible
+  def pageHasAtLeastOneVisibleElement(q: Query): Boolean = {
+    val elements = webDriver.findElements(q.by)
+    elements.asScala.exists(element => element.isDisplayed)
```
johnduffell: how about adding some logging to confirm your theory?

johnduffell:
thanks Tom, selenium is a whole world of magic! 
👍

support-frontend/test/selenium/util/Browser.scala
tomrf1: 👍 it now logs:
`INFO selenium.util.Browser Visibility of element .stripe_checkout_app: List(true, false, false)`

PR #2016, opened 2019-08-05
copy assets to S3 bucket during the deploy

johnduffell:
## Why are you doing this?

This is the second PR of three (first pr: https://github.com/guardian/support-frontend/pull/1968 ), the overall purpose is to make sure the assets are available throughout a deploy.  At present, any client accessing a new instance during a deploy may get a page containing the new hashes, however there's a chance they will hit an old instance when requesting an updated asset URL, getting a 404.

This PR was more complicated because we cloudformed the bucket, but there was no way to specify in riffraff that the bucket name varied depending on the stage.  There is expected to be a single bucket with a /PROD/... or /CODE/... keys inside it.

To get around this either a change would be needed in riffraff, or the simple solution of cloudforming the bucket in a separate deploy (support-frontend-static) and then hard coding the bucket name for the asset copy.

[**Trello Card**](https://trello.com/c/i8yIdGcu/2470-push-the-production-build-assets-to-an-s3-bucket)


build.sbt
```diff
@@ -57,7 +57,18 @@ lazy val commonDependencies = Seq(
...
+    `support-frontend`,
+    `support-workers`,
+    `support-models`,
+    `support-config`,
+    `support-internationalisation`,
+    `support-services`,
+    `support-frontend-static`
+  )
+
+lazy val `support-frontend-static` = (project in file("support-frontend-static"))
```
johnduffell: this is the new project to cloudform the actual bucket only.  It would only be deployed to PROD as there will only be one bucket for all stages.  It would need to go to prod before we deploy the branch to CODE.
This may sound convoluted, but it's still better than the standard situation where we just manually create the bucket effectively in prod, and then assume that it's there.

project/plugins.sbt
```diff
@@ -29,3 +29,4 @@ addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.9.0")
 
 addSbtPlugin("com.typesafe.sbt" % "sbt-native-packager" % "1.1.1")
 
+libraryDependencies += "javax.xml.bind" % "jaxb-api" % "2.3.1"
```
johnduffell: this is a dependency override as riffraff artifact uses a java API (via AWS S3 client 1.11) that has been removed in JDK11.  Since most people use JDK8 this isn't strictly needed so I can remove or add a comment?

support-frontend/build.sbt
```diff
@@ -68,15 +68,19 @@ riffRaffPackageName := "frontend"
 riffRaffUploadArtifactBucket := Option("riffraff-artifact")
 riffRaffUploadManifestBucket := Option("riffraff-builds")
 riffRaffArtifactResources += (file("support-frontend/cloud-formation/cfn.yaml"), "cfn/cfn.yaml")
-
-def getFiles(f: File): Seq[(File, String)] = {
-  f match {
-    case file if file.isFile => Seq((file, file.toString.replace("support-frontend/", "")))
-    case dir if dir.isDirectory => dir.listFiles.toSeq.flatMap(getFiles)
+riffRaffArtifactResources ++= getFiles(file("support-frontend/public/compiled-assets"), "assets-static")
```
johnduffell: I haven't moved the assets so they will still appear in the normal location on support.theguardian.com.  This is because the 3rd PR (todo) is to get fastly to send the requests to S3.

support-frontend/cloud-formation/CODE-deploy.sh
```diff
@@ -5,4 +5,5 @@ aws cloudformation deploy \
     --stack-name support-CODE-frontend \
     --template-file cfn.yaml \
     --region eu-west-1 \
-    --capabilities CAPABILITY_IAM
\ No newline at end of file
+    --capabilities CAPABILITY_IAM \
+    --tags App=frontend Stage=CODE Stack=support
```
johnduffell: this line was missing and was causing the CLI to remove the tags, which meant that riffraff couldn't find the code stack any more!

support-frontend/conf/riff-raff.yaml
```diff
@@ -1,6 +1,15 @@
...
 regions: [eu-west-1]
 deployments:
+  assets-static:
+    type: aws-s3
+    parameters:
+      app: frontend
+      bucket: support-frontend-static
+      prefixPackage: false
+      prefixStack: false
+      prefixStage: true
```
johnduffell: this will put /CODE/ on the start of the s3 key (Etc)

support-frontend-static/PROD-deploy.sh
```diff
@@ -0,0 +1,8 @@
+# Run this file to deploy the bucket
```
rupertbates: Maybe describe what this bucket is used for as there are a number of other buckets used by the app.

support-frontend-static/README.md
```diff
@@ -0,0 +1,33 @@
...
+# Deploying to PROD (manual)
+
+1. make your changes to cfn
+1. do a PR and get it approved
+1. merge to master
+1. check out master locally
+1. get janus credentials
+1. run ./PROD-deploy.sh
+
+# More information
```
rupertbates: Most of this is out of date now isn't it?

johnduffell: I thought so, but actually upon reading it it was all still relevant

support-frontend-static/PROD-deploy.sh
johnduffell: good idea, I've done that

support-frontend/cloud-formation/CODE-deploy.sh
jfsoul: That explains a lot!

johnduffell:
I have tried this in CODE and it seems to work!
image

johnduffell:
this has copied the assets successfully in prod too 
image

PR #2020, opened 2019-08-06
tidy up for s3 bucket for assets

johnduffell:
## Why are you doing this?

This follows on from https://github.com/guardian/support-frontend/pull/2016
I just noticed a few tidyup things that need doing.

- we don't need directory listings enabled
- we aren't actually useing the S3 website functionality, we're just using the standard public HTTP interface
- riffraff does a warning if you specify something to be the default value

I have recloudformed the bucket and it seems to be creating the right bucket, I will deploy to CODE and check the warning has gone and it seems to work still

### DIFF
--- a/support-frontend-static/cfn.yaml
+++ b/support-frontend-static/cfn.yaml
@@ -4,9 +4,7 @@ Resources:
   S3Bucket:
     Type: AWS::S3::Bucket
     Properties:
-      AccessControl: PublicRead
+      AccessControl: Private
       BucketName: support-frontend-static
       MetricsConfigurations: 
         - Id: EntireBucket
-      WebsiteConfiguration:
-        IndexDocument: index.html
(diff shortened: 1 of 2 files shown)


support-frontend-static/cfn.yaml
```diff
@@ -4,9 +4,7 @@ Resources:
   S3Bucket:
     Type: AWS::S3::Bucket
     Properties:
-      AccessControl: PublicRead
+      AccessControl: Private
```
johnduffell: private actually means "no public list permission on the bucket, but individual get access depends on the object"

support-frontend/conf/riff-raff.yaml
```diff
@@ -7,7 +7,6 @@ deployments:
       bucket: support-frontend-static
       prefixPackage: false
       prefixStack: false
-      prefixStage: true
```
johnduffell: this is riffraff's default and we get warnings if we confirm it in the deploy file 😢

PR #2025, opened 2019-08-08
Choice architecture test round 2 (Amounts first test) 

jlieb10:
Note: PR taken over by @jranks123 

## Why are you doing this?
We are testing the idea that selecting amounts first might give us an uplift. 

Moving the choice of amounts first means that we can't have different suggested amounts for the different products, so we can only have one set of suggested amounts. 

In order to isolate the change of putting the choice of amounts first, we have also tested this change. i.e having only one set of suggeted amounts. There are two variants of these amounts, set One and set Two. You can see them [here ](https://docs.google.com/spreadsheets/d/1BuAdoLzrSIWCNz0lfymNUM2XKGpqF8VwoCx7AVscXGc/edit?usp=sharing)

[**Trello Card**](https://trello.com/c/XcLbdC4p/1295-choice-architecture-r2-flip-order)

## Screenshots (To be updated)
Control:
control

Product First, Set One:
productFirstSetOne

Product First, Set Two:
productFirstSetTwo

Amounts First, Set One:
amountFiestSetOne

Amounts First, Set Two:
amountfirstSetTwo


johnduffell:
nice idea, I think there are some git conflicts here now though.
Also, even if there is no uplift in AV, there may be an uplift in usability!

support-frontend/assets/helpers/abTests/abtestDefinitions.js
```diff
@@ -1,10 +1,17 @@
 // @flow
 import type { Tests } from './abtest';
 import { get as getCookie } from 'helpers/cookie';
+import { SetOne, SetTwo } from 'helpers/abTests/data/testAmountsData';
 
 // ----- Tests ----- //
 export type LandingPageCopyReturningSinglesTestVariants = 'control' | 'returningSingle' | 'notintest';
 export type LandingPageChoiceArchitectureLabelsTestVariants = 'control' | 'withLabels' | 'notintest';
```
tomrf1: can be deleted

PR #2028, opened 2019-08-09
Increase cookie TTL (again)

jlieb10:
## Why are you doing this?
We'd like to be able to see if our contributors give again within a 12 month period as part of our new asking strategy

[**Trello Card**](https://trello.com/c/tgid0Gqh/1337-extend-contribution-cookie-to-1-year)


jlieb10:
@paulbrown1982 would you cosign this PR when you know if we're happy to do a 30 month TTL

PR #2031, opened 2019-08-13
fix support-workers it tests and make more logging

johnduffell:
## Why are you doing this?

I tried to run the `it:test` in support workers and I was getting 8 failures within these classes
```
[error] 	com.gu.zuora.ZuoraSpec
[error] 	com.gu.support.workers.integration.PreparePaymentMethodForReuseSpec
[error] 	com.gu.support.workers.integration.CreateZuoraSubscriptionSpec
[error] 	com.gu.support.workers.EndToEndSpec
```
I managed to fix some with changes in zuora UAT and DEV, but some I needed to edit the code as well.  Judging by the history I think they have been broke for a good few months, but I don't know if we can automate them easily.

EndToEnd was the hardest one, and it was caused by a change I made nearly a year ago.  It checked first for existing contributions against a hard coded identity id by querying them one by one, then it creates a new contribution against the same id.  Every single time!  we've run the test over 1000 times and we can only query about 600 before the test times out...
The fix done, which affects the prod code, was to only deduplicate subs if they are within the last 28 days.  I considered doing a hack on the tests, but I preferred to think about a way that made sense in prod too.

### DIFF
--- a/support-workers/src/it/resources/logback-test.xml
+++ b/support-workers/src/it/resources/logback-test.xml
@@ -1,3 +1,20 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<!-- Turn off logging during tests -->
-<configuration />
+<configuration>
+
+    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+        <filter class="ch.qos.logback.core.filter.EvaluatorFilter">
+            <evaluator class="ch.qos.logback.classic.boolex.OnMarkerEvaluator">
+                <marker>SENTRY</marker>
+            </evaluator>
+            <onMatch>DENY</onMatch>
+        </filter>
+
+        <encoder>
+            <pattern>[%d{yyyy-MM-dd HH:mm:ss.SSS}] %.-6level %logger{5} - %msg%n</pattern>
+        </encoder>
+    </appender>
+
+    <root level="info">
+        <appender-ref ref="STDOUT" />
+    </root>
+
+</configuration>
--- a/support-workers/src/main/scala/com/gu/support/workers/GetSubscriptionWithCurrentRequestId.scala
+++ b/support-workers/src/main/scala/com/gu/support/workers/GetSubscriptionWithCurrentRequestId.scala
@@ -22,7 +22,7 @@ object GetSubscriptionWithCurrentRequestId {
   )(implicit ec: ExecutionContext): Future[Option[DomainSubscription]] = for {
     accountNumbers <- zuoraService.getAccountFields(identityId, now())
       .withLogging("getAccountFields")
-    subscriptions <- accountNumbers.map(_.accountNumber).map{ zuoraAccountNumber =>
+    subscriptions <- accountNumbers.map(_.accountNumber).map { zuoraAccountNumber =>
       zuoraService.getSubscriptions(zuoraAccountNumber).withLogging(s"getSubscriptions($zuoraAccountNumber)")
     }.combineAll.withLogging("combineAll")
   } yield subscriptions.find(subscription => CreatedBySameRequest(requestId, subscription.existingSubscriptionRequestId))
--- a/support-workers/src/main/scala/com/gu/support/workers/GetSubscriptionWithCurrentRequestId.scala
+++ b/support-workers/src/main/scala/com/gu/support/workers/GetSubscriptionWithCurrentRequestId.scala
@@ -1,12 +1,15 @@
 package com.gu.support.workers
 
+import java.time.OffsetDateTime
 import java.util.UUID
 
 import cats.implicits._
+import com.gu.FutureLogging.LogImplicitFuture
 import com.gu.support.zuora.domain.{CreatedRequestId, DomainSubscription}
 import com.gu.zuora.ZuoraService
 
 import scala.concurrent.{ExecutionContext, Future}
+
 
 object GetSubscriptionWithCurrentRequestId {
 
@@ -14,10 +17,14 @@ object GetSubscriptionWithCurrentRequestId {
     zuoraService: ZuoraService,
     requestId: UUID,
     identityId: IdentityId,
-    billingPeriod: BillingPeriod
+    billingPeriod: BillingPeriod,
+    now: () => OffsetDateTime
   )(implicit ec: ExecutionContext): Future[Option[DomainSubscription]] = for {
-    accountNumbers <- zuoraService.getAccountFields(identityId)
-    subscriptions <- accountNumbers.map(_.accountNumber).map(zuoraService.getSubscriptions).combineAll
+    accountNumbers <- zuoraService.getAccountFields(identityId, now())
+      .withLogging("getAccountFields")
+    subscriptions <- accountNumbers.map(_.accountNumber).map{ zuoraAccountNumber =>
+      zuoraService.getSubscriptions(zuoraAccountNumber).withLogging(s"getSubscriptions($zuoraAccountNumber)")
+    }.combineAll.withLogging("combineAll")
   } yield subscriptions.find(subscription => CreatedBySameRequest(requestId, subscription.existingSubscriptionRequestId))
 
 }
--- a/support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
+++ b/support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
@@ -1,10 +1,11 @@
 package com.gu.support.workers.lambdas
 
+import java.time.OffsetDateTime
+
 import com.amazonaws.services.lambda.runtime.Context
 import com.gu.config.Configuration.zuoraConfigProvider
 import com.gu.monitoring.SafeLogger
 import com.gu.services.{ServiceProvider, Services}
-import com.gu.support.encoding.CustomCodecs._
 import com.gu.support.promotions.PromotionService
 import com.gu.support.workers._
 import com.gu.support.workers.states.{CreateZuoraSubscriptionState, SendThankYouEmailState}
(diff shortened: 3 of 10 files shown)


support-workers/src/it/resources/logback-test.xml
```diff
@@ -1,3 +1,20 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<!-- Turn off logging during tests -->
```
johnduffell: not sure why the logging was off, as it was necessary to work out what was going on.  I suppose I can turn on locally and turn off before pushing, but given we only ever run locally the logging would be nice.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -28,14 +32,20 @@ class CreateZuoraSubscription(servicesProvider: ServiceProvider = ServiceProvide
     context: Context,
     services: Services
   ): FutureHandlerResult = {
+    val now = () => OffsetDateTime.now
```
johnduffell: as a rule I prefer to pass in the side effects, to make the tests safer and the code easier to reason about

support-workers/src/main/scala/com/gu/zuora/ZuoraService.scala
```diff
@@ -29,9 +32,11 @@ class ZuoraService(val config: ZuoraConfig, client: FutureHttpClient, baseUrl: O
   def getAccount(accountNumber: String): Future[GetAccountResponse] =
     get[GetAccountResponse](s"accounts/$accountNumber", authHeaders)
 
-  def getAccountFields(identityId: IdentityId): Future[List[DomainAccount]] = {
+  def getAccountFields(identityId: IdentityId, now: OffsetDateTime): Future[List[DomainAccount]] = {
+    val recentDays = 28
```
johnduffell: since the step functions retry for 24 hours, I thought 28 days should be plenty

support-workers/src/test/resources/logback.xml
```diff
@@ -9,7 +9,7 @@
         </filter>
 
         <encoder>
-            <pattern>[%d{yyyy-MM-dd HH:mm:ss.SSS}] %X{AWSRequestId:-NO-REQUEST-ID} %.-6level %logger{5} - %msg%n</pattern>
+            <pattern>[%d{yyyy-MM-dd HH:mm:ss.SSS}] %.-6level %logger{5} - %msg%n</pattern>
```
johnduffell: since this is the test one, it doesn't ever have a request id

support-workers/src/test/scala/com/gu/support/workers/JsonFixtures.scala
```diff
@@ -534,7 +534,8 @@ object JsonFixtures {
               "Id": "0036E00000VlOPDQA3",
               "AccountId": "0016E00000f17pYQAQ"
           },
-          "acquisitionData": $acquisitionData
+          "acquisitionData": $acquisitionData,
+          $salesforceContactsJson
```
johnduffell: this was because the deserialiser was expecting this field too since gifting was introduced, and this fixture was only used in IT tests. https://github.com/guardian/support-frontend/pull/1909/files#diff-2c58c36244db5235e18da3b61dd55736R20

support-workers/src/test/scala/com/gu/zuora/ZuoraSpec.scala
```diff
@@ -19,6 +20,8 @@ class ZuoraSpec extends AsyncFlatSpec with Matchers {
 
   def uatService: ZuoraService = new ZuoraService(zuoraConfigProvider.get(true), RequestRunners.configurableFutureRunner(30.seconds))
 
+  val earlyDate = OffsetDateTime.of(2010, 1, 1, 0, 0, 0, 0, ZoneOffset.UTC)
```
johnduffell: need a long time ago to make sure we pick up the static value with a hard coded date as we are using the real UAT zuora

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -28,14 +32,20 @@ class CreateZuoraSubscription(servicesProvider: ServiceProvider = ServiceProvide
...
+        .withLogging("identity id")
+      maybeDomainSubscription <- GetSubscriptionWithCurrentRequestId(services.zuoraService, state.requestId, identityId, state.product.billingPeriod, now)
+          .withLogging("GetSubscriptionWithCurrentRequestId")
       previewPaymentSchedule <- PreviewPaymentSchedule(subscribeItem, state.product.billingPeriod, services, checkSingleResponse)
+          .withLogging("PreviewPaymentSchedule")
       thankYouState <- maybeDomainSubscription match {
         case Some(domainSubscription) => skipSubscribe(state, requestInfo, previewPaymentSchedule, domainSubscription)
+            .withLogging("skipSubscribe")
         case None => subscribe(state, subscribeItem, services).map(response => toHandlerResult(state, response, previewPaymentSchedule, requestInfo))
+          .withLogging("subscribe")
```
tomrf1: this seems like a lot of extra logging, will we see it in PROD?

johnduffell: yes we will, I'm open to adjusting the logging, but my usual problem is not enough logging when looking at support workers, hence I was generous

support-workers/src/main/scala/com/gu/zuora/ZuoraService.scala
tomrf1: comment please!

support-workers/src/main/scala/com/gu/support/workers/GetSubscriptionWithCurrentRequestId.scala
```diff
@@ -1,23 +1,30 @@
...
     identityId: IdentityId,
-    billingPeriod: BillingPeriod
+    billingPeriod: BillingPeriod,
+    now: () => OffsetDateTime
   )(implicit ec: ExecutionContext): Future[Option[DomainSubscription]] = for {
-    accountNumbers <- zuoraService.getAccountFields(identityId)
-    subscriptions <- accountNumbers.map(_.accountNumber).map(zuoraService.getSubscriptions).combineAll
+    accountNumbers <- zuoraService.getAccountFields(identityId, now())
+      .withLogging("getAccountFields")
+    subscriptions <- accountNumbers.map(_.accountNumber).map{ zuoraAccountNumber =>
```
jfsoul: space between `map` and `{` or intentionally not?

support-workers/src/main/scala/com/gu/zuora/ZuoraService.scala
```diff
@@ -29,9 +32,11 @@ class ZuoraService(val config: ZuoraConfig, client: FutureHttpClient, baseUrl: O
...
+    val recently = now.minusDays(recentDays).format(DateTimeFormatter.ISO_ZONED_DATE_TIME)
     // WARNING constructing queries from strings is inherently dangerous.  Be very careful.
-    val queryData = QueryData(s"select AccountNumber, CreatedRequestId__c from account where IdentityId__c = '${identityId.value}'")
+    val queryData = QueryData(s"select AccountNumber, CreatedRequestId__c from account where IdentityId__c = '${identityId.value}' and UpdatedDate > '$recently'")
```
jfsoul: I assume we're pretty confident this query will be more efficient than the previous version? Having not worked much with Zuora much is it reasonably safe to make adjustments to queries without being paranoid performance will be affected significantly?

johnduffell: yes, this is a built in field, we would not worry too much about performance from that point of view.  I think it's indexed under the covers.
For custom fields we can choose whether they are indexed or not, and non indexed we wouldn't be able to query.  I believe we created CreatedRequestId__c as a non indexed field, hence we just iterate through.

support-workers/src/test/scala/com/gu/support/workers/EndToEndSpec.scala
```diff
@@ -22,8 +22,9 @@ class EndToEndSpec extends LambdaSpec {
   they should "work with other currencies" in runSignupWithCurrency(EUR)
 
   def runSignupWithCurrency(currency: Currency) {
-    SafeLogger.info(createPayPalPaymentMethodContributionJson(currency))
-    val output = wrapFixture(createPayPalPaymentMethodContributionJson(currency))
+    val json = createStripePaymentMethodContributionJson(currency = currency)
```
johnduffell: I changed to stripe as once it was actually trying the subscribe, it turned out the paypal test BAID is expired.  Since stripe has a permanent test token, I decided to use that instead.

support-workers/src/test/scala/com/gu/support/workers/JsonFixtures.scala
```diff
@@ -64,7 +65,7 @@ object JsonFixtures {
           "isTestUser": false
         }
     """
-  val requestIdJson = "\"requestId\": \"e18f6418-45f2-11e7-8bfa-8faac2182601\""
+  def requestIdJson: String = s""""requestId": "${UUID.randomUUID()}\""""
```
johnduffell: this is randomised so that the end to end test will follow the subscribe path rather than the skipSubscribe one.  randomness in tests is a mixed thing, generally to be avoided in predictable tests but I think it's ok in this case.  We're not randomising the test behaviour, just an opaque field.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
johnduffell: doesn't look bad so far https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#logEventViewer:group=/aws/lambda/support-CreateZuoraSubscriptionLambda-PROD;stream=2019/08/14/[$LATEST]d92a77593ac54659a5acca38c8b8c74d;start=2019-08-13T13:40:58Z

PR #2034, opened 2019-08-14
Payment selection tests

damien-shaw-guardian:
## Why are you doing this?
The tests are to catch any changes to data before they go to production 

[**Trello Card**](https://trello.com/c/mih6w6j0/2536-product-card-clean-up)

## Changes

* add Tests for paymentSelection 

## Screenshots
N/A

### DIFF
--- a/support-frontend/assets/pages/digital-subscription-landing/components/paymentSelection/__tests__/paymentSelectionTests.js
+++ b/support-frontend/assets/pages/digital-subscription-landing/components/paymentSelection/__tests__/paymentSelectionTests.js
@@ -1,0 +1,71 @@
+// flow
+
+import { getProductOptions } from '../helpers/paymentSelection';
+
+jest.mock('ophan', () => {});
+
+describe.only('PaymentSelection', () => {
+  let productPrices;
+
+  beforeEach(() => {
+    productPrices = {
+      'United Kingdom': {
+        NoFulfilmentOptions: {
+          NoProductOptions: {
+            Monthly: {
+              GBP: {
+                price: 17.99,
+                currency: {
+                  glyph: '£',
+                  extendedGlyph: '£',
+                },
+                promotions: [],
+              },
+            },
+            Annual: {
+              GBP: {
+                price: 17.99,
+                currency: {
+                  glyph: '£',
+                  extendedGlyph: '£',
+                },
+                promotions: [],
+              },
+            },
+          },
+        },
+      },
+    };
+
+  });
+
+  it('return the correct production options ', () => {
+
+    const countryGroup = 'GBPCountries';
+    const expected = {
+      Monthly: {
+        GBP: {
+          price: 17.99,
+          currency: {
+            glyph: '£',
+            extendedGlyph: '£',
+          },
+          promotions: [],
+        },
+      },
+      Annual: {
+        GBP: {
+          price: 17.99,
+          currency: {
+            glyph: '£',
+            extendedGlyph: '£',
+          },
+          promotions: [],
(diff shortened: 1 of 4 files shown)


support-frontend/assets/pages/digital-subscription-landing/components/paymentSelection/__tests__/paymentSelectionTests.js
```diff
@@ -0,0 +1,148 @@
...
+            singlePeriod: 'year',
+            title: 'Annual',
+          },
+        ],
+      };
+
+      const result = mapStateToProps(state);
+
+      // JSON.stringify is used to fix the problem of comparing deep equality in the object
+      expect(JSON.stringify(result.paymentOptions[0], functionReplacer))
```
jfsoul: This workaround is maybe a bit too clever.  Are we trying to test every single property is the same? Is there a type for `paymentOptions` that would at least test the structure is valid?

johnduffell:
the description is a bit short, it would be good to have a bit more background on what we're hoping to get out of them and other approaches considered

support-frontend/assets/pages/digital-subscription-landing/components/paymentSelection/__tests__/paymentSelectionTests.js
johnduffell: we need a test for the functionReplacer too! what happens if it replaces everything with "function" 😆

johnduffell: comparing data is one of the hardest parts of writing tests.  but it's orders of magnitude easier than reporting differences in a sensible way.

johnduffell:
more tests is good, Id like to see a test for the functionReplacer before it goes out, because I don't trust it not to replace everything with "function".  Also the description could do with more background/examples of what we're trying to avoid.
Other than that thanks for writing the tests 👍

support-frontend/assets/pages/digital-subscription-landing/components/paymentSelection/__tests__/paymentSelectionTests.js
```diff
@@ -0,0 +1,148 @@
...
+  getCurrencySymbol,
+  getDisplayPrice,
+  getProductPrice,
+  getSavingPercentage,
+  mapStateToProps,
+} from '../helpers/paymentSelection';
+
+jest.mock('ophan', () => {}); // required to get the test to run! why?
+
+const functionReplacer = (k, v) => {
```
johnduffell:
```suggestion
const functionToStringConstant = (k, v) => {
```

support-frontend/assets/pages/digital-subscription-landing/components/paymentSelection/__tests__/paymentSelectionTests.js
```diff
@@ -0,0 +1,148 @@
...
+import {
+  getProductOptions,
+  getCurrencySymbol,
+  getDisplayPrice,
+  getProductPrice,
+  getSavingPercentage,
+  mapStateToProps,
+} from '../helpers/paymentSelection';
+
+jest.mock('ophan', () => {}); // required to get the test to run! why?
```
johnduffell: we had trouble with ophan firing when it was imported, I'm sure a fix went in to prevent that, if it was @rupertbates who made the change, but I'm not sure if that was pulled down.
If it's still breaking tests that sounds like there might still be a side effect somewhere we don't expect.

PR #2035, opened 2019-08-14
handle 4xx errors specifically in calls from service workers

johnduffell:
## Why are you doing this?

A while ago I was investigatng a step function that timed out after 24 hours.  It turned out there was a 4xx error occurring in one of the backends, i.e. client error.  I think something was missing (I forget the exact details).
It should have aborted straight away, but the step functions were turning any error when doing an http request as a Helper error, which was turning into a RetryUnlimited (i.e. 24 hours).

This PR separates out client errors (4xx) from general explosions, so that 4xx are always aborted and any other failure is retried.

Tested locally via it testing and remotely in CODE (GW 6 for 6 and contribution)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -7,7 +7,8 @@ skip in publish := true
 lazy val integrationTestSettings: Seq[Def.Setting[_]] = Defaults.itSettings ++ Seq(
   scalaSource in IntegrationTest := baseDirectory.value / "src" / "test" / "scala",
   javaSource in IntegrationTest := baseDirectory.value / "src" / "test" / "java",
-  testOptions in Test := Seq(Tests.Argument(TestFrameworks.ScalaTest, "-l", "com.gu.test.tags.annotations.IntegrationTest"))
+  testOptions in Test := Seq(Tests.Argument(TestFrameworks.ScalaTest, "-l", "com.gu.test.tags.annotations.IntegrationTest")),
+  parallelExecution in IntegrationTest := false
 )
 
 lazy val testScalastyle = taskKey[Unit]("testScalastyle")
--- a/build.sbt
+++ b/build.sbt
@@ -7,8 +7,7 @@ skip in publish := true
 lazy val integrationTestSettings: Seq[Def.Setting[_]] = Defaults.itSettings ++ Seq(
   scalaSource in IntegrationTest := baseDirectory.value / "src" / "test" / "scala",
   javaSource in IntegrationTest := baseDirectory.value / "src" / "test" / "java",
-  testOptions in Test := Seq(Tests.Argument(TestFrameworks.ScalaTest, "-l", "com.gu.test.tags.annotations.IntegrationTest")),
-  parallelExecution in IntegrationTest := false
+  testOptions in Test := Seq(Tests.Argument(TestFrameworks.ScalaTest, "-l", "com.gu.test.tags.annotations.IntegrationTest"))
 )
 
 lazy val testScalastyle = taskKey[Unit]("testScalastyle")
--- a/support-services/src/main/scala/com/gu/helpers/WebServiceHelper.scala
+++ b/support-services/src/main/scala/com/gu/helpers/WebServiceHelper.scala
@@ -13,18 +13,19 @@ import scala.concurrent.Future
 import scala.reflect.{ClassTag, classTag}
 import scala.util.{Failure, Success, Try}
 
-sealed abstract class WebServiceHelperErrorBase(responseCode: String, responseBody: Option[String]) extends Throwable
+sealed abstract class WebServiceHelperErrorBase(responseCode: String, responseBody: String) extends Throwable
 
 case class WebServiceHelperError[T: ClassTag](
   responseCode: String,
-  responseBody: Option[String]
+  responseBody: String,
+  message: String
 ) extends WebServiceHelperErrorBase(responseCode, responseBody) {
-  override def getMessage: String = s"${classTag[T]} - $responseCode: $responseBody"
+  override def getMessage: String = s"${classTag[T]} - $message: $responseCode: $responseBody"
 }
 
 case class WebServiceClientError(
   responseCode: String,
-  responseBody: Option[String]
+  responseBody: String
 ) extends WebServiceHelperErrorBase(responseCode, responseBody) {
   override def getMessage: String = s"$responseCode: $responseBody"
 }
@@ -69,25 +70,22 @@ trait WebServiceHelper[Error <: Throwable] {
       response <- httpClient(req)
       decodedResponse <- {
         val code = response.code().toString
-        SafeLogger.info(s"response code: $code")
-        val contentType = Option(response.header("content-type"))
-        SafeLogger.info(s"content type: $contentType")
-        val maybeResponseBody = Option(response.body).flatMap(body => Try(body.string()).toOption)
-        SafeLogger.info(s"response body: ${maybeResponseBody.map(_.length)} bytes")
         val failableDecodedResponse = for {
-          _ <- code.splitAt(1) match {
-            case ("2", _) => Success(())
-            case ("4", _) => Failure(maybeResponseBody.flatMap(decodeError(_).right.toOption).getOrElse(
-              WebServiceClientError(code, maybeResponseBody))
+          body <- Option(response.body).toRight(WebServiceHelperError(code, "", "no body")).toTry
+          contentType <- Option(body.contentType()).toRight(WebServiceHelperError(code, "", "no content type")).toTry
+          bodyString <- Try(body.string())
+          _ = SafeLogger.info(s"response $code body: ${bodyString.length} bytes")
+          _ <- if (contentType.`type`() == "application" && contentType.subtype() == "json") Success(())
+          else Failure(WebServiceHelperError(code, bodyString, s"wrong content type"))
+          result <- code.splitAt(1) match {
+            case ("2", _) => decode[A](bodyString).left.map { err =>
+              WebServiceHelperError[A](code, bodyString, s"failed to parse response: $err")
+            }.toTry
+            case ("4", _) => Failure((decodeError(bodyString).right.toOption).getOrElse(
+              WebServiceClientError(code, bodyString))
             )
-            case _ => Failure(WebServiceHelperError[A](code, maybeResponseBody))
+            case _ => Failure(WebServiceHelperError[A](code, bodyString, "unrecognised response code"))
           }
-          responseBody <- maybeResponseBody.toRight(WebServiceHelperError(code, maybeResponseBody)).toTry
-          result <- decode[A](responseBody)/*.left.map { err =>
-            throw decodeError(responseBody).right.getOrElse(
-              WebServiceHelperError[A](code, maybeResponseBody)
-            )
-          }*/.toTry
         } yield result
         Future.fromTry(failableDecodedResponse)
       }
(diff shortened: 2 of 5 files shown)


build.sbt
```diff
@@ -7,7 +7,8 @@ skip in publish := true
 lazy val integrationTestSettings: Seq[Def.Setting[_]] = Defaults.itSettings ++ Seq(
   scalaSource in IntegrationTest := baseDirectory.value / "src" / "test" / "scala",
   javaSource in IntegrationTest := baseDirectory.value / "src" / "test" / "java",
-  testOptions in Test := Seq(Tests.Argument(TestFrameworks.ScalaTest, "-l", "com.gu.test.tags.annotations.IntegrationTest"))
+  testOptions in Test := Seq(Tests.Argument(TestFrameworks.ScalaTest, "-l", "com.gu.test.tags.annotations.IntegrationTest")),
+  parallelExecution in IntegrationTest := false
```
johnduffell: I did this because the it tests were having timeouts in futures, however I think giving sbt 2GB instead of 1.5GB sped things up.  I'll play with removing that again tomorrow.

support-services/src/main/scala/com/gu/helpers/WebServiceHelper.scala
```diff
@@ -11,9 +11,23 @@ import scala.collection.immutable.Map.empty
...
+import scala.util.{Failure, Success, Try}
 
-case class WebServiceHelperError[T: ClassTag](responseCode: Int, responseBody: String) extends Throwable {
-  override def getMessage: String = s"${classTag[T]} - $responseCode: $responseBody"
+sealed abstract class WebServiceHelperErrorBase(responseCode: String, responseBody: String) extends Throwable
+
+case class WebServiceHelperError[T: ClassTag](
+  responseCode: String,
+  responseBody: String,
+  message: String
```
johnduffell: I added a message so we can be more clear what went wrong in the logs.

support-services/src/main/scala/com/gu/helpers/WebServiceHelper.scala
```diff
@@ -11,9 +11,23 @@ import scala.collection.immutable.Map.empty
...
+) extends WebServiceHelperErrorBase(responseCode, responseBody) {
+  override def getMessage: String = s"${classTag[T]} - $message: $responseCode: $responseBody"
+}
+
+case class WebServiceClientError(
```
johnduffell: this covers all 4xx errors

support-services/src/main/scala/com/gu/helpers/WebServiceHelper.scala
```diff
@@ -11,9 +11,23 @@ import scala.collection.immutable.Map.empty
...
 import scala.concurrent.Future
 import scala.reflect.{ClassTag, classTag}
+import scala.util.{Failure, Success, Try}
 
-case class WebServiceHelperError[T: ClassTag](responseCode: Int, responseBody: String) extends Throwable {
-  override def getMessage: String = s"${classTag[T]} - $responseCode: $responseBody"
+sealed abstract class WebServiceHelperErrorBase(responseCode: String, responseBody: String) extends Throwable
+
+case class WebServiceHelperError[T: ClassTag](
+  responseCode: String,
```
johnduffell: in reality is isn't really a String or an Int, but I changed to string so I can index into it easily

support-services/src/main/scala/com/gu/helpers/WebServiceHelper.scala
```diff
@@ -51,19 +65,32 @@ trait WebServiceHelper[Error <: Throwable] {
    */
   private def request[A](rb: Request.Builder)(implicit decoder: Decoder[A], errorDecoder: Decoder[Error], ctag: ClassTag[A]): Future[A] = {
     val req = wsPreExecute(rb).build()
-    SafeLogger.debug(s"Issuing request ${req.method} ${req.url}")
+    SafeLogger.info(s"Issuing request ${req.method} ${req.url}")
     for {
       response <- httpClient(req)
-    } yield {
-      val responseBody = response.body.string()
```
johnduffell: body and string are both nullable according to okhttp3 docs

support-services/src/main/scala/com/gu/helpers/WebServiceHelper.scala
```diff
@@ -51,19 +65,32 @@ trait WebServiceHelper[Error <: Throwable] {
...
+        val code = response.code().toString
+        val failableDecodedResponse = for {
+          body <- Option(response.body).toRight(WebServiceHelperError(code, "", "no body")).toTry
+          contentType <- Option(body.contentType()).toRight(WebServiceHelperError(code, "", "no content type")).toTry
+          bodyString <- Try(body.string())
+          _ = SafeLogger.info(s"response $code body: ${bodyString.length} bytes")
+          _ <- if (contentType.`type`() == "application" && contentType.subtype() == "json") Success(())
+          else Failure(WebServiceHelperError(code, bodyString, s"wrong content type"))
+          result <- code.splitAt(1) match {
+            case ("2", _) => decode[A](bodyString).left.map { err =>
```
jfsoul: Could have left `code` as an `int` and then we don't have to split a string and match on that.  There's also an `isSuccessful` method by the looks of it, but no equivalent for client errors.

support-workers/src/main/scala/com/gu/support/workers/exceptions/RetryImplicits.scala
```diff
@@ -15,8 +15,8 @@ object RetryImplicits {
       //Timeouts/connection issues and 500s
       case e @ (_: SocketTimeoutException | _: SocketException | _: WebServiceHelperError[_]) => new RetryUnlimited(message = e.getMessage, cause = throwable)
 
-      //Invalid Json
-      case e @ (_: ParsingFailure | _: MatchError) => new RetryNone(message = e.getMessage, cause = throwable)
+      //Invalid Json or http client error (4xx)
+      case e @ (_: ParsingFailure | _: MatchError | _: WebServiceClientError) => new RetryNone(message = e.getMessage, cause = throwable)
```
jfsoul: We're sure that there aren't any retryable 4XX errors?

johnduffell: well nothing is for sure, especially with zuora.  Having said that I have a feeling zuora errors are returned as 2xx with success=false or Success=false in the body, so I will investigate (and add tests)

support-services/src/main/scala/com/gu/helpers/WebServiceHelper.scala
johnduffell: yeah isSuccessful seems to be the only one in okhttp, unfortunately no nice ADT of response codes.
I agree I could have left as a int, but I'd still have to do `code % 100` and check it against 2 and 4, but I don't entirely feel like arithmetic makes sense on a response code.  More concretely it opens us up to off by one errors.  In a string each character is opaque when splitting so if it works in one case it should work in them all.
that's my reasoning but happy to use mod instead if you feel strongly!

support-services/src/main/scala/com/gu/helpers/WebServiceHelper.scala
```diff
@@ -51,19 +65,32 @@ trait WebServiceHelper[Error <: Throwable] {
...
       response <- httpClient(req)
-    } yield {
-      val responseBody = response.body.string()
-      SafeLogger.debug(responseBody)
-      decode[A](responseBody) match {
-        case Left(err) => throw decodeError(responseBody).right.getOrElse(
-          WebServiceHelperError[A](response.code(), responseBody)
-        )
-        case Right(value) => value
+      decodedResponse <- {
```
tomrf1: can this block be moved out into a function? Nested for/yields can be hard to follow, especially in github

johnduffell: good idea hadn't spotted how long that had got

support-services/src/main/scala/com/gu/helpers/WebServiceHelper.scala
```diff
@@ -51,19 +65,32 @@ trait WebServiceHelper[Error <: Throwable] {
...
+        val code = response.code().toString
+        val failableDecodedResponse = for {
+          body <- Option(response.body).toRight(WebServiceHelperError(code, "", "no body")).toTry
+          contentType <- Option(body.contentType()).toRight(WebServiceHelperError(code, "", "no content type")).toTry
+          bodyString <- Try(body.string())
+          _ = SafeLogger.info(s"response $code body: ${bodyString.length} bytes")
+          _ <- if (contentType.`type`() == "application" && contentType.subtype() == "json") Success(())
+          else Failure(WebServiceHelperError(code, bodyString, s"wrong content type"))
+          result <- code.splitAt(1) match {
```
tomrf1: I think this match block could also be a separate function, there's a lot going on in this `for` block

support-workers/src/main/scala/com/gu/support/workers/exceptions/RetryImplicits.scala
johnduffell: TODO- I need to check that these still work (I think they don't in this PR)
https://github.com/guardian/support-frontend/blob/ef690cebb233c7ed593840937cb3b78d2d9cc800/support-models/src/main/scala/com/gu/support/zuora/api/response/Responses.scala#L48

support-services/src/main/scala/com/gu/helpers/WebServiceHelper.scala
jfsoul: Maybe it would be easier to read quickly as an `if...else` with `code >= 200 && code < 300`? Separating out as Tom suggested is probably more meaningful though.

johnduffell:
```scala
      result <-
        if (code >= 200 && code < 300) 
          decode[A](bodyString).left.map { err =>
            WebServiceHelperError[A](code, bodyString, s"failed to parse response: $err")
          }.toTry
        else if (code >= 400 && code < 500) 
  Failure((decodeError(bodyString).right.toOption).getOrElse(
            WebServiceClientError(code, bodyString))
          )
        else Failure(WebServiceHelperError[A](code, bodyString, "unrecognised response code"))
      }
```

johnduffell: see what you mean about the WTF factor, I have an idea to make it read better

johnduffell:
```scala
      result <- code(0) + "xx" match {
        case "2xx" => decode[A](bodyString).left.map { err =>
          WebServiceHelperError[A](code, bodyString, s"failed to parse response: $err")
        }.toTry
        case "4xx" => Failure((decodeError(bodyString).right.toOption).getOrElse(
          WebServiceClientError(code, bodyString))
        )
        case _ => Failure(WebServiceHelperError[A](code, bodyString, "unrecognised response code"))
      }
```

support-services/src/main/scala/com/gu/helpers/WebServiceHelper.scala
johnduffell: sorry, body is nullable, but string might throw an IOException.  In any case we probably just want to blow up and retry with a RetryUnlimited

support-workers/src/main/scala/com/gu/support/workers/exceptions/RetryImplicits.scala
johnduffell: I have added the tests to ZuoraErrorsSpec.scala  below and they are now passing

johnduffell:
I think this is ready, @jfsoul @tomrf1 if you wanted to take a look please 👀

johnduffell:
tested in CODE, digi pack and contributions both worked

johnduffell:
thanks Jon I'll merge that tomorrow morning

PR #2038, opened 2019-08-15
Home delivery - ensure postcode is deliverable (within M25)

damien-shaw-guardian:
## Why are you doing this?
We are adding postcode validation into the paper checkout form (only when home delivery is selected).  This validation matches the postcode prefix (everything but the final three characters) against a hardcoded list of prefixes which we know our delivery partner can deliver to.  These prefixes nearly map to postcodes within the M25, but there are some exceptions.

Postcode validation currently occurs along with other validation, when the form is submitted.

[**Trello Card**](https://trello.com/c/XlcAd6cH/2434-turn-home-delivery-back-on-nb-fulfillment-delays-need-fixing)

## Changes

* New delivery check + test

## Screenshots

image


support-frontend/assets/components/subscriptionCheckouts/address/deliveryCheck.js
```diff
@@ -0,0 +1,45 @@
+const M25_POSTCODE_PREFIXES = [
```
johnduffell: here is the list from subs frontend
https://github.com/guardian/subscriptions-frontend/blob/0f01397786041b2b225e73bf764c8732f41c0948/assets/javascripts/modules/checkout/deliveryDetails.js#L25

support-frontend/assets/components/subscriptionCheckouts/address/postcodeFinder.jsx
```diff
@@ -59,9 +60,21 @@ const InputWithButton = ({ onClick, isLoading, ...props }) => (
...
 );
 
-const ComposedInputWithButton = compose(withLabel, asControlled, withError)(InputWithButton);
-const ComposedSelect = compose(withLabel)(Select);
+const withDeliveryCheck = WappedComponent => ({ results, ...props }) => (
+  <div>
+    <WappedComponent {...props} />
+    { results.length > 0 && postcodeIsWithinDeliveryArea(props.value) && (
+      <p>
+        Sorry, we cannot deliver a paper to an address with this postcode.
```
johnduffell:
```suggestion
        Sorry, we cannot deliver a paper outside the M25.
```

support-frontend/assets/components/subscriptionCheckouts/address/postcodeFinder.jsx
```diff
@@ -59,9 +60,21 @@ const InputWithButton = ({ onClick, isLoading, ...props }) => (
...
+        Please call Customer Services on: 0330 333 6767 or press Back to purchase a voucher subscription.
```
johnduffell:
```suggestion
        You can have vouchers delivered anywhere in the UK to redeem at your newsagent, please go here to subscribe.
```
the here would have to be a link

support-frontend/assets/components/subscriptionCheckouts/address/deliveryCheck.js
```diff
@@ -0,0 +1,45 @@
...
+  'W1', 'W10', 'W11', 'W12', 'W13', 'W14', 'W1A', 'W1B', 'W1C', 'W1D', 'W1F', 'W1G', 'W1H', 'W1J', 'W1K', 'W1M', 'W1N', 'W1P', 'W1R', 'W1S', 'W1T', 'W1U', 'W1V', 'W1W', 'W1X', 'W1Y',
+  'W2', 'W3', 'W4', 'W5', 'W6', 'W7', 'W8', 'W9',
+  'WC1', 'WC1A', 'WC1B', 'WC1E', 'WC1H', 'WC1N', 'WC1R', 'WC1V', 'WC1X', 'WC2', 'WC2A', 'WC2B', 'WC2E', 'WC2H', 'WC2N', 'WC2R',
+  'WD1', 'WD17', 'WD18', 'WD19', 'WD23', 'WD24', 'WD25', 'WD3', 'WD4', 'WD5', 'WD6', 'WD7',
+  'KT10', 'KT11', 'KT12', 'KT13', 'KT14', 'KT15', 'KT16', 'KT17', 'KT18', 'KT19', 'KT2', 'KT20', 'KT21', 'KT3', 'KT4', 'KT5', 'KT6', 'KT7', 'KT8', 'KT9',
+  'TN16',
+];
+
+const isPostcodeInPrefixes = (postcode, prefix) => {
+  // check postcode length is greater than 6 characters
```
johnduffell: in is that a todo? where/why?

support-frontend/assets/components/subscriptionCheckouts/address/deliveryCheck.js
```diff
@@ -0,0 +1,45 @@
...
+  const formattedPostcode = postcode.replace(' ', '').toUpperCase();
```
johnduffell: it's not really formatted, more compacted

support-frontend/assets/components/subscriptionCheckouts/address/postcodeFinder.jsx
```diff
@@ -59,9 +60,21 @@ const InputWithButton = ({ onClick, isLoading, ...props }) => (
   </div>
 );
 
-const ComposedInputWithButton = compose(withLabel, asControlled, withError)(InputWithButton);
-const ComposedSelect = compose(withLabel)(Select);
+const withDeliveryCheck = WappedComponent => ({ results, ...props }) => (
+  <div>
+    <WappedComponent {...props} />
+    { results.length > 0 && postcodeIsWithinDeliveryArea(props.value) && (
```
johnduffell: does this run every key press, or wait until you leave the field?

support-frontend/assets/components/subscriptionCheckouts/address/deliveryCheck.js
```diff
@@ -0,0 +1,45 @@
...
+  'UB1', 'UB10', 'UB11', 'UB2', 'UB3', 'UB4', 'UB5', 'UB6', 'UB7', 'UB8', 'UB9',
+  'W1', 'W10', 'W11', 'W12', 'W13', 'W14', 'W1A', 'W1B', 'W1C', 'W1D', 'W1F', 'W1G', 'W1H', 'W1J', 'W1K', 'W1M', 'W1N', 'W1P', 'W1R', 'W1S', 'W1T', 'W1U', 'W1V', 'W1W', 'W1X', 'W1Y',
+  'W2', 'W3', 'W4', 'W5', 'W6', 'W7', 'W8', 'W9',
+  'WC1', 'WC1A', 'WC1B', 'WC1E', 'WC1H', 'WC1N', 'WC1R', 'WC1V', 'WC1X', 'WC2', 'WC2A', 'WC2B', 'WC2E', 'WC2H', 'WC2N', 'WC2R',
+  'WD1', 'WD17', 'WD18', 'WD19', 'WD23', 'WD24', 'WD25', 'WD3', 'WD4', 'WD5', 'WD6', 'WD7',
+  'KT10', 'KT11', 'KT12', 'KT13', 'KT14', 'KT15', 'KT16', 'KT17', 'KT18', 'KT19', 'KT2', 'KT20', 'KT21', 'KT3', 'KT4', 'KT5', 'KT6', 'KT7', 'KT8', 'KT9',
+  'TN16',
+];
+
+const isPostcodeInPrefixes = (postcode, prefix) => {
```
johnduffell:
```suggestion
const postcodeHasPrefix = (postcode, prefix) => {
```

support-frontend/assets/components/subscriptionCheckouts/address/deliveryCheck.js
```diff
@@ -0,0 +1,45 @@
...
+  // check postcode length is greater than 6 characters
+  const formattedPostcode = postcode.replace(' ', '').toUpperCase();
+  const lastThreeCharacters = formattedPostcode.split('').slice(-3).join('');
+  const newPostcode = formattedPostcode.replace(lastThreeCharacters, '');
+
+  return newPostcode === prefix;
+};
+
+const postcodeIsWithinDeliveryArea = postcode => M25_POSTCODE_PREFIXES
+  .filter(prefix => isPostcodeInPrefixes(postcode, prefix)).length > 0;
```
johnduffell: `isPostcodeInPrefixes` is basically case insensitive version of startsWith, so why not just toUpper the entered text once before the filter, and then use startsWith. Otherwise we are normalising the entered text on every iteration.  I realise startsWith and split+replace have different behaviour, but I think both will work reasonably ish in the circumstances

support-frontend/assets/components/subscriptionCheckouts/address/__tests__/deliveryCheckTests.js
```diff
@@ -0,0 +1,13 @@
+import { postcodeIsWithinDeliveryArea } from 'components/subscriptionCheckouts/address/deliveryCheck';
+
+describe('Delivery Check', () => {
+  it('should return true if the postcode is in the delivery area', () => {
+    expect(postcodeIsWithinDeliveryArea('SE2 2LB')).toBeTruthy();
+    expect(postcodeIsWithinDeliveryArea('SE20 2LB')).toBeTruthy();
+    expect(postcodeIsWithinDeliveryArea('SE8 2AD')).toBeTruthy();
+    expect(postcodeIsWithinDeliveryArea('SW17 2LB')).toBeTruthy();
```
johnduffell: the first 4 tests are pretty much the same logic, why not just reduce to one.
Also add some partial and empty string, if that's something the code might have to deal with? e.g.
"", "N", "N1", "N111111111"

support-frontend/assets/components/subscriptionCheckouts/address/deliveryCheck.js
```diff
@@ -0,0 +1,45 @@
...
+  'WC1', 'WC1A', 'WC1B', 'WC1E', 'WC1H', 'WC1N', 'WC1R', 'WC1V', 'WC1X', 'WC2', 'WC2A', 'WC2B', 'WC2E', 'WC2H', 'WC2N', 'WC2R',
+  'WD1', 'WD17', 'WD18', 'WD19', 'WD23', 'WD24', 'WD25', 'WD3', 'WD4', 'WD5', 'WD6', 'WD7',
+  'KT10', 'KT11', 'KT12', 'KT13', 'KT14', 'KT15', 'KT16', 'KT17', 'KT18', 'KT19', 'KT2', 'KT20', 'KT21', 'KT3', 'KT4', 'KT5', 'KT6', 'KT7', 'KT8', 'KT9',
+  'TN16',
+];
+
+const isPostcodeInPrefixes = (postcode, prefix) => {
+  // check postcode length is greater than 6 characters
+  const formattedPostcode = postcode.replace(' ', '').toUpperCase();
+  const lastThreeCharacters = formattedPostcode.split('').slice(-3).join('');
```
lucymonie:
```suggestion
  const lastThreeCharacters = formattedPostcode.slice(-3);
```

support-frontend/assets/components/subscriptionCheckouts/address/deliveryCheck.js
lucymonie: You could also possibly skip all the reformatting and trimming of the postcode by taking the raw postcode entered by the user, by uppercasing it and then filtering the prefixes by checking if the postcode entered includes the prefix.

```
const postcodeIsWithinDeliveryArea = postcode => M25_POSTCODE_PREFIXES
  .filter(prefix => postcode.toUppercase().includes(prefix)).length > 0;
```

support-frontend/assets/components/subscriptionCheckouts/address/__tests__/deliveryCheckTests.js
lucymonie: I tend to agree - it would be great to break out all the steps involved in manipulating the postcode and test each one rather than multiple tests of the overall outcome.

support-frontend/assets/components/subscriptionCheckouts/address/deliveryCheck.js
lucymonie: There are some postcodes missing in the list here that are in the list from subs:
```
[ 'CM14',
  'DA6',
  'EC1',
  'EC2',
  'EC3',
  'EC4',
  'KT1',
  'N9',
  'RM5',
  'RM9',
  'WD2' ]
```

lucymonie: And there are also loads more in your list that aren't included in the subs list.

damien-shaw-guardian: the list from subs frontend is the wrong list, I tried to attach the correct list here but it didn't work, I can share it separately

lucymonie: 👍

support-frontend/assets/components/subscriptionCheckouts/address/postcodeFinder.jsx
damien-shaw-guardian: actually I need to add an onblur check

support-frontend/assets/components/subscriptionCheckouts/address/deliveryCheck.js
damien-shaw-guardian: starts with allows postcodes that are outside the M25, or if the delivery company decides it does not delivery to a postcode it is harder to change, in terms of operation cost this is a 0.003s operation in a single browser, so i wouldn't worry about the extra operation, currently this happens only on click (and will also happen on blur)

damien-shaw-guardian: using `includes` would allow invalid postcodes to pass e.g. 'GE1 5JK' would pass

support-frontend/assets/components/subscriptionCheckouts/address/__tests__/deliveryCheckTests.js
damien-shaw-guardian: this only really has two tests 'is the prefix inside the M25' or 'is the prefix outside the M25' all the other validation happens when we check if this is a valid postcode, before this M25 check

support-frontend/assets/components/subscriptionCheckouts/address/deliveryCheck.js
damien-shaw-guardian: updated just to use slice

support-frontend/assets/components/subscriptionCheckouts/address/postcodeFinder.jsx
damien-shaw-guardian: this has been removed in favour of triggering the M25 check onSubmit

support-frontend/assets/components/subscriptionCheckouts/address/__tests__/addressFieldStoreTests.js
```diff
@@ -0,0 +1,44 @@
+import { isHomeDeliveryInM25 } from 'components/subscriptionCheckouts/address/addressFieldsStore';
+
+jest.mock('ophan', () => {});
+
+describe('addressFieldStore', () => {
+
+  describe('isHomeDeliveryInM25 - this is a form error check so when this function returns "false" when there is an error', () => {
+
+    it('should return true when the order is a home delivery and the postcode in the M25', () => {
```
jfsoul: Big thumbs-up for the tests, although I think we are essentially testing `postcodeIsWithinDeliveryArea` in two test suites now.  Not a big deal, but if you wanted to isolate the unique functionality of `isHomeDeliveryInM25` you could just check whether `postcodeIsWithinDeliveryArea` is called or true is returned. In fact, it might be easier to test if the predicate for checking the postcode were passed in, rather than internal to `isHomeDeliveryInM25`.  I wouldn't go back and change this for this PR, just a thought for next time.

support-frontend/assets/components/subscriptionCheckouts/address/__tests__/deliveryCheckTests.js
```diff
@@ -0,0 +1,33 @@
+import { postcodeIsWithinDeliveryArea, postcodeHasPrefix } from 'components/subscriptionCheckouts/address/deliveryCheck';
+
+describe('Delivery Check', () => {
+  it('should return true if the postcode prefix matches the prefix', () => {
```
jfsoul: I'd nest another `describe` and separate the prefix test from the others.

support-frontend/assets/components/subscriptionCheckouts/address/deliveryCheck.js
```diff
@@ -0,0 +1,48 @@
+// @flow
```
jfsoul: Regarding file location: do we have a convention of this kind of file being in a `helpers` directory or similar?

support-frontend/assets/components/subscriptionCheckouts/address/deliveryCheck.js
```diff
@@ -0,0 +1,48 @@
...
+  'W2', 'W3', 'W4', 'W5', 'W6', 'W7', 'W8', 'W9',
+  'WC1', 'WC1A', 'WC1B', 'WC1E', 'WC1H', 'WC1N', 'WC1R', 'WC1V', 'WC1X', 'WC2', 'WC2A', 'WC2B', 'WC2E', 'WC2H', 'WC2N', 'WC2R',
+  'WD1', 'WD17', 'WD18', 'WD19', 'WD23', 'WD24', 'WD25', 'WD3', 'WD4', 'WD5', 'WD6', 'WD7',
+  'KT10', 'KT11', 'KT12', 'KT13', 'KT14', 'KT15', 'KT16', 'KT17', 'KT18', 'KT19', 'KT2', 'KT20', 'KT21', 'KT3', 'KT4', 'KT5', 'KT6', 'KT7', 'KT8', 'KT9',
+  'TN16',
+];
+
+export const postcodeHasPrefix = (postcode: Option<string>, prefix: string): boolean => {
+
+  const formattedPostcode = postcode !== null ? postcode.replace(' ', '').toUpperCase() : '';
```
jfsoul: Minor: you could flip this ternary to avoid negation of the predicate. Or do `if(postcode === null) { return false }` and not worry about it here.

support-frontend/assets/components/subscriptionCheckouts/address/deliveryCheck.js
```diff
@@ -0,0 +1,48 @@
...
+  const lastThreeCharacters = formattedPostcode.slice(-3);
+  const postcodePrefix = formattedPostcode.replace(lastThreeCharacters, '');
```
jfsoul: I guess this is equivalent to formattedPostcode.splice(0, formattedPostcode.length - 3) or similar if you wanted a one-liner.  Beware of out-by-one errors if you like my suggestion, because I haven't worked it out!

support-frontend/assets/components/subscriptionCheckouts/address/deliveryCheck.js
```diff
@@ -0,0 +1,48 @@
...
+
+  return postcodePrefix === prefix;
+};
+
+const postcodeIsWithinDeliveryArea = (postcode: Option<string>): boolean =>
+  postcode !== null && M25_POSTCODE_PREFIXES.filter(prefix => postcodeHasPrefix(postcode, prefix)).length > 0;
```
jfsoul: ah I see so we have already checked `postcode !== null` here - does flow not let you then just use a string type for the `postcodeHasPrefix` parameter?

support-frontend/assets/components/subscriptionCheckouts/address/deliveryCheck.js
lucymonie: And don't use splice because it mutates the array.

support-frontend/assets/components/subscriptionCheckouts/address/deliveryCheck.js
johnduffell: I see, I wonder wrong list in terms of what? did the list come from cex team or did you just go by the "M25" defintion?

support-frontend/assets/components/subscriptionCheckouts/address/deliveryCheck.js
johnduffell: see what you mean in case E1 is in the m25 but E16 is not.  Makes sense 💯

support-frontend/assets/components/subscriptionCheckouts/address/__tests__/deliveryCheckTests.js
johnduffell: fair enough in which case the first 4 tests are all the same by the looks in terms of logic?  Although more tests is always stated as a goal, it's actually better to avoid redundant tests as they have a maintenance cost.

support-frontend/assets/components/subscriptionCheckouts/address/deliveryCheck.js
johnduffell: helpers is one of my words to be avoided at all costs 😆 , this is because it includes more stuff than it excludes i.e. it's generic.  We should really group things according to cohesive functionality where possible. It does make organising things take significantly more thought, but it's a better time to do that thought than when trying to refactor later.

support-frontend/assets/components/subscriptionCheckouts/address/deliveryCheck.js
johnduffell: to be honest replace makes me nervous because we're cracking out a regex on the whole lot just to truncate the last 3 characters.  What we want to do and what we wrote doesn't exactly match.

Obviously it's client side so less likely to cause any issue, but my preference would be
- work out the length of the string
- subtract 3
- if less than 2, fail, else take the substring from 0 - n-1

then all the rules and potential side effects are plain

support-frontend/assets/components/subscriptionCheckouts/address/deliveryCheck.js
damien-shaw-guardian: this is only in place because flow, needs postcode type to be the same as postcodeIsWithinDeliveryArea postcode type, which means it (flow) also needs null to be handled. 

This doesn't really make sense because I have to handle a state that doesn't happen

support-frontend/assets/components/subscriptionCheckouts/address/deliveryCheck.js
damien-shaw-guardian: that's correct

support-frontend/assets/components/subscriptionCheckouts/address/deliveryCheck.js
damien-shaw-guardian: the list i received from Matthew Sullivan that he got from Newsprinters (the delivery company)

johnduffell: great thanks in that case it will be perfect 👍

support-frontend/assets/helpers/deliveryCheck.js
```diff
@@ -0,0 +1,47 @@
...
+  'KT10', 'KT11', 'KT12', 'KT13', 'KT14', 'KT15', 'KT16', 'KT17', 'KT18', 'KT19', 'KT2', 'KT20', 'KT21', 'KT3', 'KT4', 'KT5', 'KT6', 'KT7', 'KT8', 'KT9',
+  'TN16',
+];
+
+export const postcodeHasPrefix = (postcode: Option<string>, prefix: string): boolean => {
+
+  const formattedPostcode = postcode !== null ? postcode.replace(' ', '').toUpperCase() : '';
+  const postcodePrefix = formattedPostcode.slice(0, -3);
+
+  return postcodePrefix === prefix;
```
johnduffell: yep the slice with -3 makes more sense now! I hadn't realised that was easy until I checked: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice

johnduffell: only as a minor suggestion. this line is not too clear on its own becuase the varilable names are basically nothing to distinguish them.  To be clear there would have to be something non arbitrary to tell them apart.  I would suggest renaming to show which is actual and which is expected.
```suggestion
  return actualPrefix === expectedPrefix;
```

PR #2040, opened 2019-08-16
Modifications to enable a third US-based Stripe account

paulbrown1982:
Modifications to enable a third US-based Stripe account. Main changes are that the public key is chosen based on the page's country not the transacting currency. Also that the public key needs to be transmitted along with the payment and analytics information over to the Payment API so that it can be completely sure which account the token is generated from, and doesn't duplicate the based-on-country algorithm.

Use of the new key is controlled by a new 'Payment Method' feature switch which I need to add to the [admin console](https://github.com/guardian/support-admin-console).

## Why are you doing this?
To enable Discover and Diners Club cards to be used by USA customers. Also to move less money around in the back-office between UK and US-based USD company bank accounts.

[**Trello Card**](https://trello.com/c/23rljvzx)

## Screenshots
No fundamental changes to UX other than new cards (Diners and Discover) will be accepted. 

Before / with the Feature Switch off:
Screen Shot 2019-09-02 at 13 21 26
Screen Shot 2019-09-02 at 13 20 33
Screen Shot 2019-09-02 at 13 21 12

After / with the Feature Switch on:
Screen Shot 2019-09-02 at 10 22 26
Screen Shot 2019-09-02 at 13 20 07
Screen Shot 2019-09-02 at 13 19 44

### DIFF
--- a/support-config/src/main/resources/touchpoint.PROD.conf
+++ b/support-config/src/main/resources/touchpoint.PROD.conf
@@ -1,8 +1,12 @@
 touchpoint.backend.environments {
   PROD {
     stripe {
-      AUD {
+      AU {
         api.key.public = "pk_live_57xV50eFjPIA990PEGWJwoHp"
+        api.key.secret = "dummy-key"
+      }
+      US {
+        api.key.public = "pk_live_2O6zPMHXNs2AGea4bAmq5R7Z"
         api.key.secret = "dummy-key"
       }
       default {
--- a/support-config/src/main/scala/com/gu/support/config/StripeConfig.scala
+++ b/support-config/src/main/scala/com/gu/support/config/StripeConfig.scala
@@ -17,11 +17,11 @@ case class StripeConfig(defaultAccount: StripeAccountConfig,
     maybeCurrency match {
       case Some(AUD) => {
         SafeLogger.info(s"StripeConfig: getting AU stripe account for AUD")
-        forCountry(Some(Country.Australia))
+        australiaAccount
       }
       case _ => {
         SafeLogger.info(s"StripeConfig: getting default stripe account for ${maybeCurrency.map(_.iso).mkString}")
-        forCountry(None)
+        defaultAccount
       }
     }
   }
--- a/support-config/src/main/scala/com/gu/support/config/StripeConfig.scala
+++ b/support-config/src/main/scala/com/gu/support/config/StripeConfig.scala
@@ -1,24 +1,45 @@
 package com.gu.support.config
 
-import com.gu.i18n.Currency
+import com.gu.i18n.{Country, Currency}
 import com.gu.i18n.Currency.AUD
 import com.gu.monitoring.SafeLogger
 import com.typesafe.config.Config
 
 
-case class StripeConfig(defaultAccount: StripeAccountConfig, australiaAccount: StripeAccountConfig, version: Option[String] = None)
+case class StripeConfig(defaultAccount: StripeAccountConfig,
+                        australiaAccount: StripeAccountConfig,
+                        unitedStatesAccount: StripeAccountConfig,
+                        version: Option[String] = None)
   extends TouchpointConfig {
-  def forCurrency(maybeCurrency: Option[Currency]): StripeAccountConfig =
+  def forCurrency(maybeCurrency: Option[Currency]): StripeAccountConfig = {
     maybeCurrency match {
       case Some(AUD) => {
-        SafeLogger.info(s"StripeConfig: getting AU stripe account for $maybeCurrency")
-        australiaAccount
+        SafeLogger.info(s"StripeConfig: getting AU stripe account for AUD")
+        forCountry(Some(Country.Australia))
       }
       case _ => {
-        SafeLogger.info(s"StripeConfig: getting default stripe account for $maybeCurrency")
+        SafeLogger.info(s"StripeConfig: getting default stripe account for ${maybeCurrency.map(_.iso).mkString}")
+        forCountry(None)
+      }
+    }
+  }
+
+  def forCountry(maybeCountry: Option[Country]): StripeAccountConfig = {
+    maybeCountry match {
+      case Some(Country.Australia) => {
+        SafeLogger.info(s"StripeConfig: getting AU stripe account for Australia")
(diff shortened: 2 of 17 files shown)


support-frontend/assets/helpers/paymentIntegrations/oneOffContributions.js
```diff
@@ -64,6 +64,7 @@ export type StripeChargeData = {|
     stripePaymentMethod: StripePaymentMethod,
   },
   acquisitionData: PaymentAPIAcquisitionData,
+  publicKey: string
```
paulbrown1982: This tells the Payment API which backend Stripe account it should use to process the card token, in order to remain consistent with the front-end decision making (via the setting switch) about which account to use.

support-config/src/main/scala/com/gu/support/config/StripeConfig.scala
```diff
@@ -1,24 +1,47 @@
...
 import com.gu.i18n.Currency.AUD
 import com.gu.monitoring.SafeLogger
 import com.typesafe.config.Config
 
 
-case class StripeConfig(defaultAccount: StripeAccountConfig, australiaAccount: StripeAccountConfig, version: Option[String] = None)
+case class StripeConfig(defaultAccount: StripeAccountConfig,
+                        australiaAccount: StripeAccountConfig,
+                        unitedStatesAccount: StripeAccountConfig,
+                        version: Option[String])
```
paulbrown1982: I removed the Option[String] = None here as I tripped up on it and accidentally deleted some code that passed a value into it.

jlieb10: Doesn't `= None` just provide an initial value in the case that a value isn't passed in?

support-config/src/main/scala/com/gu/support/config/StripeConfig.scala
```diff
@@ -1,24 +1,47 @@
...
   extends TouchpointConfig {
-  def forCurrency(maybeCurrency: Option[Currency]): StripeAccountConfig =
+
+  // Still needed for SupportWorkers (recurring products) which don't support a US Stripe account yet.
```
jlieb10: ✅

support-frontend/assets/pages/contributions-landing/contributionsLandingActions.js
```diff
@@ -47,6 +47,7 @@ import type { PaymentMethod } from 'helpers/paymentMethods';
 import { DirectDebit, Stripe } from 'helpers/paymentMethods';
 import type { RecentlySignedInExistingPaymentMethod } from 'helpers/existingPaymentMethods/existingPaymentMethods';
 import { ExistingCard, ExistingDirectDebit } from '../../helpers/paymentMethods';
+import { getStripeKey, stripeAccountForContributionType } from '../../helpers/paymentIntegrations/stripeCheckout';
```
jlieb10:
```suggestion
import { getStripeKey, stripeAccountForContributionType } from 'helpers/paymentIntegrations/stripeCheckout';
```

support-frontend/assets/pages/contributions-landing/contributionsLandingActions.js
```diff
@@ -47,6 +47,7 @@ import type { PaymentMethod } from 'helpers/paymentMethods';
 import { DirectDebit, Stripe } from 'helpers/paymentMethods';
 import type { RecentlySignedInExistingPaymentMethod } from 'helpers/existingPaymentMethods/existingPaymentMethods';
 import { ExistingCard, ExistingDirectDebit } from '../../helpers/paymentMethods';
```
jlieb10:
```suggestion
import { ExistingCard, ExistingDirectDebit } from 'helpers/paymentMethods';
```

support-config/src/main/scala/com/gu/support/config/StripeConfig.scala
jlieb10: Is it that it's easier to read without it?

support-frontend/app/views/subscriptionCheckout.scala.html
```diff
@@ -53,6 +53,7 @@
           uat: "@uatStripeConfig.forCurrency(Some(AUD)).publicKey"
         }
       };
+      window.guardian.stripeKeyUnitedStates = window.guardian.stripeKeyDefaultCurrencies,
```
jlieb10: What does this do @paulbrown1982

paulbrown1982: It forces the US public key to be the same as the default one, because the new account won't be used for Subscriptions - it's easier to do this than to carry the 'what-product-is-it?' logic into the simple 'what country is it?' decision system about what Stripe account to use.

support-frontend/assets/helpers/paymentIntegrations/oneOffContributions.js
jlieb10: Got it- I thought it must be something like this

support-config/src/main/scala/com/gu/support/config/StripeConfig.scala
paulbrown1982: Yup, but it also means you can make the mistake I did before this commit: https://github.com/guardian/support-frontend/pull/2040/commits/ef28ff44d181fc1a63dd90e9039d6549e6ba4de2

jlieb10: Ah so it's because `None` isn't actually something we ever want there, so it's a bad default parameter

paulbrown1982: Not quite, let's chat f2f tomorrow.

support-config/src/main/resources/touchpoint.PROD.conf
```diff
@@ -1,10 +1,14 @@
 touchpoint.backend.environments {
   PROD {
     stripe {
-      AUD {
```
tomrf1: Discovered post-deploy:
This change needs to be reflected in s3 in the support-workers.private.conf, to avoid breaking recurring in AU

PR #2045, opened 2019-08-21
Fix GW gifting

jfsoul:
## Why are you doing this?
We are not correctly setting contact details in Zuora when gifting GW.  It appears that:

- the Salesforce Contact ID should be that of the recipient (i.e. giftee orElse buyer) rather than the buyer. 
- the soldTo contact should have the giftee's name and email details (address seems correct already)

This requires verifying in CODE/DEV before merging.

## Changes

* Change 1
* Change 2


jfsoul:
Verified with @arausuy that this looks good in salesforce/zuora.

PR #2046, opened 2019-08-22
Add an error summary to Weekly and Paper checkouts

lucymonie:
## Why are you doing this?
We recently applied an error summary to the DP checkout. We are now adding the error summary to Guardian Weekly and Paper subscription checkouts.

[**Trello Card**](https://trello.com/c/zrI6Chih/2572-add-error-summary-to-gw-and-paper-checkouts)

## Changes
* Refactored error summary changes so they can be applied more easily to the other checkouts.
* Added error summary to weekly checkout.
* Added error summary to paper checkout.
* Applied address type to error messages so it's clear if it's the billing address city or the delivery address postcode that's missing.
* Fixed the state validation to check that state is not null and is not an empty string as the validation was allowing the US state not to be set without triggering an error.

## Weekly error summary
Screen Shot 2019-08-22 at 17 41 31

## Paper error summary
Screen Shot 2019-08-23 at 16 37 52


support-frontend/assets/components/subscriptionCheckouts/payPalSubmitButton.jsx
```diff
@@ -84,17 +65,12 @@ function SubscriptionSubmitButtons(props: PropTypes) {
             billingPeriod={props.billingPeriod}
           />
         </div>
-        <Button
-          id="qa-submit-button"
```
jfsoul: could be worth running the post-deploy tests locally with this change just to double check!

support-frontend/assets/helpers/subscriptionsForms/formValidation.js
```diff
@@ -59,12 +59,12 @@ function withDeliveryValidation(state: WithDeliveryCheckoutState): AnyErrorType[
...
     }: Error<FormField>),
     ({
-      errors: applyAddressRules(getDeliveryAddressFields(state)),
+      errors: applyAddressRules(getDeliveryAddressFields(state), 'delivery'),
       errorAction: setAddressFormErrorsFor('delivery'),
     }: Error<AddressFormField>),
     ...(shouldValidateBillingAddress(formFields) ? [
       ({
-        errors: applyAddressRules(getBillingAddressFields(state)),
+        errors: applyAddressRules(getBillingAddressFields(state), 'billing'),
```
jfsoul: ah this is an interesting change - it intersects with Damien's PR for HD because the postcode check only happens in `'delivery'` address rules.  It'll be his conflict to merge probably though 😄

johnduffell:
nice work splitting into separate PRs
as a nit pick, I am not keen on "There is a problem" as the title, as it sounds a bit harsh.  I'd prefer something like "Please check the information"

support-frontend/assets/components/subscriptionCheckouts/payPalSubmitButton.jsx
```diff
@@ -84,17 +65,12 @@ function SubscriptionSubmitButtons(props: PropTypes) {
...
-          type="submit"
-          modifierClasses={props.paymentMethod === PayPal ? ['hidden'] : []}
-        >
-          Continue to payment
-        </Button>
       </div>
-      <span>{props.allErrors.length > 0 && <ErrorSummary errors={props.allErrors} />}</span>
+      {props.paymentMethod === PayPal && (
```
johnduffell: it is a bit funny that we have to replicate the errors in both buttons.  it would be better if we can split the errors somewhere that's common to both, as I think the errors are unrelated to which kind of button paypal or other we want to use

support-frontend/assets/helpers/subscriptionsForms/validation.js
```diff
@@ -23,7 +23,6 @@ function notNull<A>(value: A): boolean {
   return value !== null;
 }
 
-
```
johnduffell: was this intentional?

support-frontend/assets/components/subscriptionCheckouts/payPalSubmitButton.jsx
lucymonie: The difficulty that we have here is that the errors belong to the fields, and in following the pattern of keeping the implementation in the component, we need to pass the errors through to the button that's being used. Happy to discuss if this doesn't make sense.

johnduffell: discussed in person, sounds ok
The reason is it's because the aggregation of errors is actually implemented as a validation error on the "submit field" so it has to be contained in the same div.  And anyway most of the implementation is shared in the ErrorSummary component.

PR #2053, opened 2019-08-27
SCA: support stripe PaymentMethod field

johnduffell:
## Why are you doing this?

We want to support the new SCA so that we can still take payments from customers after 14th Sep.
Stripe has a set of great instructions to show to be SCA ready here: https://stripe.com/docs/payments/cards/saving-cards
This PR implements step 5.

At the moment we just take a card token from stripe, and push it into the create customer step in order to be able to charge them multiple times.
The new world requires us to use a payment_method instead of a card token (source).
This change keeps the old functionality so that we don't break existing code, but also means that when someone submits the field from the client side with a paymentMethod instead, it will try to use the SCA method instead.

A small concern is how Zuora will deal with things, as we will have a payment method id and a customer token, rather than a card and customer token so although we can just send things to zuora in the normal way, we don't know exactly what field to put them in and whether it will be a different payment gateway name.  There will be another PR do do this.

See zuora thread: https://community.zuora.com/t5/Zuora-Announcements/PSD2-amp-SCA-how-Zuora-is-helping-you-prepare/ba-p/29434

We will need a second PR to clean up the "old" way once it's no longer used, but it makes sense for the support workers to be agnostic in the mean time.
If it's not clear otherwise, we might want some logging or metrics so that we know when that point is reached though?  I haven't specifically added any yet.

[**Trello Card**](https://trello.com/c/6EtZduO5/2534-add-parameter-to-support-workers-to-trigger-new-intent-confirmation)

@rupertbates @lucymonie @tomrf1 @jfsoul this PR might be worth a review please! Let me know what you think.

### DIFF
--- a/support-frontend/test/utils/CheckoutValidationRulesTest.scala
+++ b/support-frontend/test/utils/CheckoutValidationRulesTest.scala
@@ -99,8 +99,13 @@ class DigitalPackValidationTest extends FlatSpec with Matchers {
     DigitalPackValidation.passes(requestMissingPostcode) shouldBe true
   }
 
-  it should "fail if the payment field received is an empty string" in {
-    val requestMissingState = validDigitalPackRequest.copy(paymentFields = StripePaymentFields(""))
+  it should "fail if the payment method payment field received is an empty string" in {
+    val requestMissingState = validDigitalPackRequest.copy(paymentFields = StripePaymentMethodPaymentFields(""))
+    DigitalPackValidation.passes(requestMissingState) shouldBe false
+  }
+
+  it should "fail if the source payment field received is an empty string" in {
+    val requestMissingState = validDigitalPackRequest.copy(paymentFields = StripeSourcePaymentFields(""))
     DigitalPackValidation.passes(requestMissingState) shouldBe false
   }
 
@@ -182,7 +187,7 @@ object TestData {
     lastName = "hopper",
     product = DigitalPack(Currency.USD, Monthly),
     firstDeliveryDate = None,
-    paymentFields = StripePaymentFields("test-token"),
+    paymentFields = StripePaymentMethodPaymentFields("test-token"),
     ophanIds = OphanIds(None, None, None),
     referrerAcquisitionData = ReferrerAcquisitionData(None, None, None, None, None, None, None, None, None, None, None, None),
     supportAbTests = Set(),
@@ -219,7 +224,7 @@ object TestData {
     lastName = "hopper",
     product = Paper(Currency.GBP, Monthly, HomeDelivery, Everyday),
     firstDeliveryDate = Some(someDateNextMonth),
-    paymentFields = StripePaymentFields("test-token"),
+    paymentFields = StripePaymentMethodPaymentFields("test-token"),
     ophanIds = OphanIds(None, None, None),
     referrerAcquisitionData = ReferrerAcquisitionData(None, None, None, None, None, None, None, None, None, None, None, None),
     supportAbTests = Set(),
--- a/support-models/src/main/scala/com/gu/support/workers/PaymentFields.scala
+++ b/support-models/src/main/scala/com/gu/support/workers/PaymentFields.scala
@@ -10,7 +10,9 @@ sealed trait PaymentFields
 
 case class PayPalPaymentFields(baid: String) extends PaymentFields
 
-case class StripePaymentFields(stripeToken: String) extends PaymentFields
+sealed trait StripePaymentFields extends PaymentFields
+case class StripeSourcePaymentFields(stripeToken: String) extends StripePaymentFields // pre SCA compatibility
+case class StripePaymentMethodPaymentFields(paymentMethod: String) extends StripePaymentFields
 
 case class DirectDebitPaymentFields(
   accountHolderName: String,
@@ -23,13 +25,15 @@ case class ExistingPaymentFields(billingAccountId: String) extends PaymentFields
 object PaymentFields {
   //Payment fields are input from support-frontend
   implicit val payPalPaymentFieldsCodec: Codec[PayPalPaymentFields] = deriveCodec
-  implicit val stripePaymentFieldsCodec: Codec[StripePaymentFields] = deriveCodec
+  implicit val stripeSourcePaymentFieldsCodec: Codec[StripeSourcePaymentFields] = deriveCodec
+  implicit val stripePaymentMethodPaymentFieldsCodec: Codec[StripePaymentMethodPaymentFields] = deriveCodec
   implicit val directDebitPaymentFieldsCodec: Codec[DirectDebitPaymentFields] = deriveCodec
   implicit val existingPaymentFieldsCodec: Codec[ExistingPaymentFields] = deriveCodec
 
   implicit val encodePaymentFields: Encoder[PaymentFields] = Encoder.instance {
     case p: PayPalPaymentFields => p.asJson
-    case s: StripePaymentFields => s.asJson
+    case s: StripeSourcePaymentFields => s.asJson
+    case s: StripePaymentMethodPaymentFields => s.asJson
     case d: DirectDebitPaymentFields => d.asJson
     case e: ExistingPaymentFields => e.asJson
   }
@@ -37,7 +41,8 @@ object PaymentFields {
   implicit val decodePaymentFields: Decoder[PaymentFields] =
     List[Decoder[PaymentFields]](
       Decoder[PayPalPaymentFields].widen,
-      Decoder[StripePaymentFields].widen,
+      Decoder[StripeSourcePaymentFields].widen,
+      Decoder[StripePaymentMethodPaymentFields].widen,
       Decoder[DirectDebitPaymentFields].widen,
       Decoder[ExistingPaymentFields].widen
     ).reduceLeft(_ or _)
--- a/support-workers/src/main/scala/com/gu/stripe/StripeService.scala
+++ b/support-workers/src/main/scala/com/gu/stripe/StripeService.scala
@@ -5,8 +5,10 @@ import com.gu.i18n.Currency
 import com.gu.okhttp.RequestRunners._
 import com.gu.stripe.Stripe._
 import com.gu.support.config.StripeConfig
+import io.circe.Decoder
 
 import scala.concurrent.{ExecutionContext, Future}
+import scala.reflect.ClassTag
 
 class StripeService(config: StripeConfig, client: FutureHttpClient, baseUrl: String = "https://api.stripe.com/v1")(implicit ec: ExecutionContext)
     extends WebServiceHelper[Stripe.StripeError] {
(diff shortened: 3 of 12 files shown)


support-models/src/main/scala/com/gu/support/workers/PaymentFields.scala
```diff
@@ -10,7 +10,9 @@ sealed trait PaymentFields
 
 case class PayPalPaymentFields(baid: String) extends PaymentFields
 
-case class StripePaymentFields(stripeToken: String) extends PaymentFields
+sealed trait StripePaymentFields extends PaymentFields
+case class StripeSourcePaymentFields(stripeToken: String) extends StripePaymentFields // pre SCA compatibility
+case class StripePaymentMethodPaymentFields(paymentMethod: String) extends StripePaymentFields
```
johnduffell: once we are not using the old mechanism any more, we can remove the stripeToken side of the type, and flatten it back down.

support-workers/src/main/scala/com/gu/stripe/StripeService.scala
```diff
@@ -15,16 +17,37 @@ class StripeService(config: StripeConfig, client: FutureHttpClient, baseUrl: Str
   val wsUrl = baseUrl
   val httpClient: FutureHttpClient = client
 
-  def createCustomer(token: String, currency: Currency): Future[Customer] =
-    postForm[Customer]("customers", Map(
-      "card" -> Seq(token)
-    ), getHeaders(currency))
+  def withCurrency(currency: Currency): WithCurrency = new WithCurrency(currency)
```
johnduffell: I could have made this less clunky perhaps
having a class with a parameter is basically partially applying all the methods inside with whatever the parameter is.
I could have just kept the currency as a parameter to `post` and then partially applied it, rather than making a class and wrapping everything.

The ultimate aim was I just wanted to get away from passing currency and stripeService all the way through a load of methods higher up, only in order to combine them at the bottom anyway.

support-workers/src/main/scala/com/gu/stripe/StripeService.scala
```diff
@@ -15,16 +17,37 @@ class StripeService(config: StripeConfig, client: FutureHttpClient, baseUrl: Str
...
+        case (map, version) => map + ("Stripe-Version" -> version)
+      }
+
+    private def getAuthorizationHeader() =
+      Map("Authorization" -> s"Bearer ${config.forCurrency(Some(currency)).secretKey}")
+
+  }
+}
+
+object CreateCustomer {
```
johnduffell: I wanted to split out the abstract definitions of the operations from the mechanism to actually make the calls - otherwise before long we end up with a lot of methods all more coupled than they should be.
However I'm open to other ways of doing it syntactically, as this specific implementation might be a bit too "FP" for the existing codebase.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreatePaymentMethod.scala
```diff
@@ -60,14 +60,15 @@ class CreatePaymentMethod(servicesProvider: ServiceProvider = ServiceProvider)
...
-        CreditCardReferenceTransaction(card.id, stripeCustomer.id, card.last4,
-          CountryGroup.countryByCode(card.country), card.exp_month, card.exp_year, card.zuoraCardType)
-      }
+  def createStripePaymentMethod(stripe: StripePaymentFields, stripeService: StripeService#WithCurrency): Future[CreditCardReferenceTransaction] =
+    (stripe match {
+      case StripeSourcePaymentFields(source) => CreateCustomer.fromSource(source)
+      case StripePaymentMethodPaymentFields(paymentMethod) => CreateCustomer.fromPaymentMethod(paymentMethod)
+    })(stripeService).map { stripeCustomer =>
+      val card = stripeCustomer.source
+      CreditCardReferenceTransaction(card.id, stripeCustomer.id, card.last4,
```
johnduffell: here I'm assuming the existing card.id will be alright for Zuora's primary token field.
I think zuora only uses it when it does a POST on /v1/charges https://stripe.com/docs/api/charges/create
and zuora actually posts it as "card" instead of "source" because they are assuming a version of stripe API from 2015.  This may be ignored by stripe, but I have a feeling that stripe let you get away with it, as we are actually making the same mistake in support-workers!  Anyway, we only have a single (default) source so I don't think there will be any problem regardless.

support-workers/src/main/scala/com/gu/stripe/StripeService.scala
```diff
@@ -15,16 +17,37 @@ class StripeService(config: StripeConfig, client: FutureHttpClient, baseUrl: Str
...
+  // https://stripe.com/docs/api/customers/create
+
+  def fromSource(token: String): StripeService#WithCurrency => Future[Customer] =
+    _.post[Customer]("customers", Map(
+      "source" -> Seq(token)
```
johnduffell: this was "card" but according to stripe docs card was removed in 2015 version of the API and we should have been using "source" all along! Somehow we got away with it though.

support-workers/src/main/scala/com/gu/stripe/StripeService.scala
tomrf1: How about:
`case class StripeServiceWithCurrency(service: StripeService, currency: Currency)`

support-workers/src/main/scala/com/gu/stripe/StripeService.scala
rupertbates: I can see what you're trying to achieve, but this feels considerably less easy to read than the previous version. If the original class had a lot of methods then I could see the value more but it only has one, so I think introducing another class and an object is overkill.

rupertbates: You could leave the existing class as is pass the `StripePaymentFields` object into `createCustomer` and match on it there to work out whether to pass `source` or `payment_method` into the Stripe call

support-workers/src/main/scala/com/gu/stripe/StripeService.scala
rupertbates: I think this is less clear than what we had before, see comment below

support-workers/src/main/scala/com/gu/stripe/StripeService.scala
rupertbates: Backwards compatibility FTW!

support-workers/src/test/scala/com/gu/support/workers/lambdas/CreatePaymentMethodStateDecoderSpec.scala
```diff
@@ -57,13 +57,23 @@ class CreatePaymentMethodStateDecoderSpec extends FlatSpec with Matchers with Mo
...
     val fieldsToTest = maybeState.map(state =>
       (state.product, state.paymentFields))
     fieldsToTest should be(Right(
       Contribution(5, GBP, Monthly),
-      StripePaymentFields(stripeToken)
+      StripeSourcePaymentFields(stripeToken)
+    ))
+  }
+
+  it should "be able to decode a contribution with Stripe payment method payment fields" in {
```
rupertbates: 👍 Well done for adding this

support-workers/src/main/scala/com/gu/stripe/StripeService.scala
johnduffell: yeah I think I'm looking at it from a purist point of view (i.e. thing X should only know about one concept) wheras you are looking for a more fewer things point of view (don't have too many "things" and fit the concepts where it makes sense at the moment and just keep the structure fluid as the balance changes)
I will go through again and tweak it back and see what it looks like.

johnduffell:
ok I have thought about all the comments from me tom and rupert and I have tried to make something that is the best of all the suggestions.  It is slightly weird because of the extra class to wrap the currency, but the additional object is gone, and the class isn't nested any more.  I haven't updated the tests yet but have a look and comment again please.

tomrf1:
Thanks @johnduffell , it looks good to me - I like the extra class!

johnduffell:
thanks tom I'm fixing up the tests now

support-workers/src/main/scala/com/gu/stripe/StripeService.scala
```diff
@@ -8,23 +8,36 @@ import com.gu.support.config.StripeConfig
...
-class StripeService(config: StripeConfig, client: FutureHttpClient, baseUrl: String = "https://api.stripe.com/v1")(implicit ec: ExecutionContext)
+class StripeService(val config: StripeConfig, client: FutureHttpClient, baseUrl: String = "https://api.stripe.com/v1")(implicit ec: ExecutionContext)
     extends WebServiceHelper[Stripe.StripeError] {
 
   // Stripe URL is the same in all environments
   val wsUrl = baseUrl
   val httpClient: FutureHttpClient = client
 
-  def createCustomer(token: String, currency: Currency): Future[Customer] =
+  def withCurrency(currency: Currency): StripeServiceForCurrency = new StripeServiceForCurrency(this, currency)
```
johnduffell: I had to add this back because we have to pass in the mock StripeServiceForCurrency, but as it's an end to end test so we have to pass it in from the top level.  The simplest way of doing so is to attach it to Services via this method and it all happens by magic.

support-workers/src/test/scala/com/gu/support/workers/AddressLineTransformerTest.scala
```diff
@@ -20,8 +20,6 @@ class AddressLineTransformerTest extends FlatSpec with Matchers with MockitoSuga
     accountNumber = "55779911"
   )
 
-  val stripePaymentFields = StripePaymentFields("test-token")
```
johnduffell: this wasn't actually used anwhere

johnduffell:
this is ready for reviewing/merging - however in other news although this PR might help our journey to SCA compliance, we will find out in the next couple of days from zuora exactly what we should do with the fields.  We know for sure it won't use a card ID, we probably have to send the Customer ID and the Payment Method ID to zuora.  I will ship that change in a subsequent PR as it doesn't affect this one.

support-frontend/test/utils/CheckoutValidationRulesTest.scala
```diff
@@ -182,7 +187,7 @@ object TestData {
     lastName = "hopper",
     product = DigitalPack(Currency.USD, Monthly),
     firstDeliveryDate = None,
-    paymentFields = StripePaymentFields("test-token"),
+    paymentFields = StripePaymentMethodPaymentFields("test-token"),
```
jfsoul: I haven't looked in detail at the tests that use these, but are we missing anything by not having a `StripeSourcePaymentFields` request now?

jfsoul: I suppose it is temporary anyway

johnduffell: We do have one, that's on line 108 actually. Due to the way these tests work it start with the default and overwrites it

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreatePaymentMethod.scala
johnduffell: just to update this comment - this assumption was incorrect, we need to pass the payment_method id instead, and use a stripe v2 gateway, but it does seem to work in sandbox.

johnduffell:
I have run this through with a test user and it worked for GW 6 for 6

PR #2055, opened 2019-08-28
SCA-ready credit card form in all Subscriptions checkouts

lucymonie:
## Why are you doing this?
The aim is to add a custom credit card form to all three subscription checkouts.

The initial work is to implement stripe elements with the existing stripe payment process, so we have a credit/debit card checkout that matches the designs in the [Trello Card](https://trello.com/c/O9EvNpNw/2559-build-sca-credit-card-stripe-elements-wireframes) and can (hopefully) quite easily add the steps required for SCA in a further piece of work.

It's worth noting that the implementation doesn't exactly match the design - the expiry and CVC elements aren't side by side. This is because the error messages make a mess of the layout when the elements are side by side. The variance with the design has been agreed with Gus and Jesse.

## Changes
* Implement stripe elements
* Add custom errors, and make them work with existing errors and error summary
* Make sure errors and submit work as expected
* Style stripe elements to look like part of the Gu checkout

## Not in scope but relevant questions
* Should the error summary update on change in the same way that the checkout fields do? To elaborate, the errors are deleted from next to the field on change in the field, but this behaviour isn't replicated in the error summary.
* ~~Is there a way of persevering the the paymentMethod when it's credit/debit card? This is because a change to the address country causes the page to refresh and reloads the payment selector. If the user has selected a payment method and possibly filled in some fields, this will be lost, and this could be annoying.~~  
* The thank you checkout page state currently appears scrolled down to the bottom of the page, which is not an optimal view of the page from a UX point of view. Is this deliberate?
* The opening and closing of the credit card form is a little jumpy and could benefit from a small animation to make it smoother.
* In the case of a card error, the message currently appears in the payment method section (see last screenshot, below), which may still be the best place for it. I have flagged this for UXD consideration.
* We have commented out one of the post-deployment tests because there are issues getting the card details into the iFrames and then getting the form to submit. We'll create a new card for this work and then the test can be reinstated.

## Screenshots
Screen Shot 2019-09-10 at 15 37 19

Screen Shot 2019-09-10 at 15 37 29

Screen Shot 2019-09-10 at 15 37 43

Screen Shot 2019-09-10 at 15 43 53

Screen Shot 2019-09-10 at 15 38 09

Screen Shot 2019-09-12 at 12 53 33


support-frontend/app/views/main.scala.html
```diff
@@ -30,6 +30,8 @@
       </script>
     }
 
+    <script defer id="stripe-js" src="https://js.stripe.com/v3/"></script>
```
rupertbates: This script tag already exists on line 148 of this file

lucymonie: Ah yes... I think Tom Forbes and I may have been making changes at the same time

support-frontend/assets/components/subscriptionCheckouts/stripeForm/stripeForm.jsx
```diff
@@ -0,0 +1,207 @@
...
+const invalidStyles = {
+  color: '#c70000',
+};
+
+// Main component
+
+class StripeForm extends Component<PropTypes, StateTypes> {
+  constructor(props) {
+    super(props);
+    this.state = {
```
rupertbates: We should handle all of this state in Redux. That that might also help with persisting the values when the address changes.

support-frontend/assets/components/subscriptionCheckouts/subscriptionSubmitButton.jsx
```diff
@@ -20,28 +19,34 @@ import './subscriptionSubmitButton.scss';
...
 
 
 function SubscriptionSubmitButton(props: PropTypes) {
   return (
-    <span>
-      {props.paymentMethod !== PayPal && (
+    <span className={props.paymentMethod === props.className ? 'show' : 'hide'}>
       <div className="component-submit-button">
-        <div className="component-submit-button--margin">
+        <div className={props.paymentMethod === DirectDebit ? 'component-submit-button--margin' : ''}>
```
rupertbates: Would be good to find a nicer way to do this, and the same for the PayPal submit button above.

support-frontend/assets/components/subscriptionCheckouts/stripeForm/stripeForm.jsx
```diff
@@ -0,0 +1,206 @@
...
+type StateTypes = {
+  cardNumber: Object,
+  cardExpiry: Object,
+  cardCvc: Object,
+  cardErrors: Array<Object>,
+}
+
+// Styles for stripe elements
+
+const baseStyles = {
```
rupertbates: See if we can move this to the CSS

support-frontend/assets/components/subscriptionCheckouts/stripeForm/stripeForm.jsx
```diff
@@ -0,0 +1,206 @@
...
+  requestStripeToken = (event) => {
+    event.preventDefault();
+    this.props.validateForm();
+    this.handleCardErrors();
+    if (this.props.stripe && this.props.allErrors.length === 0 && this.state.cardErrors.length === 0) {
+      const { stripe } = this.props;
+      stripe.createToken({ type: 'card', name: this.props.name })
+        .then(({ token, error }) => {
+          if (error) {
+            console.log(error);
```
rupertbates: Do we need to report this error to the user?

support-frontend/assets/components/subscriptionCheckouts/stripeForm/stripeForm.jsx
rupertbates: Decided to leave as is

support-frontend/assets/components/checkoutForm/checkoutForm.jsx
```diff
@@ -35,6 +35,35 @@ FormSection.defaultProps = {
   noBorder: false,
 };
 
+// Hidden version of form section
```
johnduffell: is "hidden" really the defining feature of this? if it really is hidden, then there must be some other reason for it to exist

support-frontend/assets/components/checkoutForm/checkoutForm.jsx
```diff
@@ -35,6 +35,35 @@ FormSection.defaultProps = {
...
+  headingSize: HeadingSize,
+  hidden?: boolean,
+  id?: Option<string>,
+|};
+
+const FormSectionHidden = ({
+  children, title, headingSize, hidden, id,
+}: FormSectionHiddenPropTypes) => (
+  <div id={id} className={`component-checkout-form-section ${hidden ? 'component-checkout-form-section--hidden' : ''}`}>
+    {!hidden && (
```
johnduffell: I guess it's like a hide by default thing, but to avoid a double negative it might be worth calling it something like `show` and `FormSectionShowHide`

support-frontend/assets/components/checkoutForm/checkoutForm.scss
```diff
@@ -33,7 +33,12 @@ div .component-checkout-form-section.component-checkout-form-section--no-border
   border-top: none;
 }
 
+div .component-checkout-form-section.component-checkout-form-section--hidden {
+  display: none;
+}
```
johnduffell: not sure if it's useful here, but in dotcom we used to have utility classes for things like hidden, meaning it was class `is-hidden` and it didn't depend on where it was.  I'm not sure if that's a proper BEM pattern but it did save having to reimplement hiding every time.
https://github.com/guardian/frontend/blob/840f5dbc0ebc03b606a75139d1b08bf949cb6fed/static/src/stylesheets/base/_helpers.scss#L87

support-frontend/assets/components/subscriptionCheckouts/subscriptionSubmitButton.scss
```diff
@@ -2,5 +2,16 @@
...
+    margin-bottom: $gu-v-spacing * 2;
+    margin-left: $gu-h-spacing / 2;
+}
+
+.show {
+    display: block;
+}
+
+.hide {
+    display: none;
```
johnduffell: ah in fact these are pretty much what i meant about utility classes, is it best practice to reuse these?

johnduffell:
looks good although I am not totally sure what I'm looking at! Great work digging through all that as the changes seem to be well spready about the codebase.
one thing I did think is how will this affect the post deployment tests?  I know we try to bypass the stripe stuff as much as possible, but it would be worth spinning them up to see how they handle it.
I think you can run them locally although it doesn't work for everyone.

PR #2056, opened 2019-08-28
fix: added missing postcodes

damien-shaw-guardian:
##  Why are you doing this?
There are some missing postcodes on the current home delivery M25_POSTCODE_PREFIXES list. This PR adds the missing postcodes

[**Trello Card**](https://trello.com/c/cXb40KLN/2583-update-postcodes-in-postcode-checker)

## Changes

* Updated the postcodes

## Screenshots
N/A


johnduffell:
thanks 👍  but given the list before was supposedly the official one from the supplier, what changed? (i.e. how do we know this one is actually right now)

PR #2057, opened 2019-08-29
Allow tests to be enabled from the url only

tomrf1:
## Why are you doing this?
Sometimes it's useful to try out a new test in PROD without opting users into the test.
It means we can integration test with e.g. our Stripe production account before putting it live to users.

Setting `urlParticipationOnly=true` on a test means you can only opt into the test if you set it in the url, e.g. `support.theguardian.com/uk/contribute#ab-stripeElements=stripeCardElement`

TODO - write some unit tests

### DIFF
--- a/support-frontend/assets/helpers/abTests/abtest.js
+++ b/support-frontend/assets/helpers/abTests/abtest.js
@@ -206,9 +206,8 @@ function getParticipations(
       const participationsFromUrl = getParticipationsFromUrl();
       if (participationsFromUrl && participationsFromUrl[testId]) {
         participations[testId] = participationsFromUrl[testId];
-      } else {
-        return;
       }
+      return;
     }
 
     if (testId in currentParticipation) {
--- a/support-frontend/assets/helpers/abTests/abtest.js
+++ b/support-frontend/assets/helpers/abTests/abtest.js
@@ -66,6 +66,7 @@ export type Test = {|
   canRun?: () => boolean,
   independent: boolean,
   seed: number,
+  urlParticipationOnly: boolean,
 |};
 
 export type Tests = { [testId: string]: Test }
@@ -199,6 +200,15 @@ function getParticipations(
 
     if (test.canRun && !test.canRun()) {
       return;
+    }
+
+    if (test.urlParticipationOnly) {
+      const participationsFromUrl = getParticipationsFromUrl();
+      if (participationsFromUrl && participationsFromUrl[testId]) {
+        participations[testId] = participationsFromUrl[testId];
+      } else {
+        return;
+      }
     }
 
     if (testId in currentParticipation) {


support-frontend/assets/helpers/abTests/abtest.js
```diff
@@ -201,6 +202,15 @@ function getParticipations(
       return;
     }
 
+    if (test.urlParticipationOnly) {
+      const participationsFromUrl = getParticipationsFromUrl();
+      if (participationsFromUrl && participationsFromUrl[testId]) {
+        participations[testId] = participationsFromUrl[testId];
```
johnduffell: would this be overwritten on line 215? and is that intentional?

johnduffell:
something you do have to be careful of is whether people would accidentally put links in e/g. an article or email, which would direct people to a specific bucket.
having the boolean is a good idea to prevent that issue from both directions, although I'm not totally certain of the implementation and whether that was the intention?

support-frontend/assets/helpers/abTests/abtest.js
tomrf1: good point, it should return here

support-frontend/assets/helpers/abTests/abtest.js
```diff
@@ -201,6 +202,15 @@ function getParticipations(
...
+      } else {
+        return;
```
johnduffell: so if any test is url participation and its not in the url, we would return undefined? (not in any test?)

tomrf1: we're inside a `forEach` here so it just abandons this particular test.

tomrf1: A nicer and more functional approach would be to use `reduce`, though it would still be mutating an object as it goes

johnduffell: did not know return returns from a foreach 🔢

support-frontend/assets/helpers/abTests/abtest.js
```diff
@@ -201,6 +202,14 @@ function getParticipations(
       return;
     }
 
+    if (test.urlParticipationOnly) {
+      const participationsFromUrl = getParticipationsFromUrl();
+      if (participationsFromUrl && participationsFromUrl[testId]) {
+        participations[testId] = participationsFromUrl[testId];
+      }
+      return;
```
johnduffell: can we just use an `else` instead of returning half way through?  then the structure seems a bit more level

support-frontend/assets/helpers/abTests/abtest.js
```diff
@@ -201,6 +202,14 @@ function getParticipations(
       return;
     }
 
+    if (test.urlParticipationOnly) {
```
johnduffell: I feel like "only" somehow makes it ambiguous.
Can we just say `participationMethod` and it's either `url` or `standard` (or something) then it's clear it's either/or

tomrf1:
Closing in favour of https://github.com/guardian/support-frontend/pull/2062, which is much simpler

PR #2058, opened 2019-08-29
Use Stripe Elements for single contributions

tomrf1:
## Why are you doing this?
For SCA we need to migrate from the Stripe Charges API to the Payment Intents API for one-off contributions.
This also means using Stripe Elements on the client-side.
[Backend work](https://github.com/guardian/payment-api/pull/173) on payment-api has been deployed. See that PR for an explanation of the flow.

This is implemented as an A/B test so we can see if it affects conversion.
Users in the variant will enter their card details through the new form, and if their bank wants to do 3DS authentication then this will popup before payment can complete.
An ophan component view event is sent when 3DS appears, so we can track acquisitions with/without 3DS.

Apologies for the size of this PR. I don't think it would have been helpful to split it up.

## Screenshots
Card details form displays when Stripe is selected on the one-off view, 'Contribute' button is disabled until form is filled out:
Screenshot 2019-08-29 at 14 08 50

No card details form if Paypal selected:
Screenshot 2019-08-29 at 14 09 10

Card details errors from Stripe Elements are displayed below as you go (you don't have to click 'Contribute' first):
Screenshot 2019-08-29 at 14 09 21

'Contribute' button only becomes enabled once all fields are filled out correctly:
Screenshot 2019-08-29 at 14 09 37

If card provider requires 3DS then it is displayed in a popup:
Screenshot 2019-08-29 at 14 09 48

A special error message is displayed if 3DS authentication fails:
Screenshot 2019-08-29 at 14 10 02


support-frontend/app/controllers/Application.scala
```diff
@@ -149,7 +149,7 @@ class Application(
       regularUatStripeConfig = regularStripeConfigProvider.get(true),
       regularDefaultPayPalConfig = payPalConfigProvider.get(false),
       regularUatPayPalConfig = payPalConfigProvider.get(true),
-      paymentApiStripeEndpoint = paymentAPIService.stripeExecutePaymentEndpoint,
+      paymentApiStripeUrl = paymentAPIService.stripeUrl,
```
tomrf1: Just return the base url, as we now hit several endpoints

support-frontend/app/views/main.scala.html
```diff
@@ -30,6 +30,8 @@
       </script>
     }
 
+    <script defer id="stripe-js" src="https://js.stripe.com/v3/"></script>
```
tomrf1: This was previously at the bottom of the body.
It prevented stripe from loading (`window.Stripe`) because the `<script>` element was not available at the point that react creates the components. Having it in the header solves this problem, and it's still marked with `defer` so I don't think it affects performance?

support-frontend/assets/pages/contributions-landing/components/ContributionForm.jsx
```diff
@@ -126,7 +129,7 @@ const mapDispatchToProps = (dispatch: Function) => ({
   openDirectDebitPopUp: () => { dispatch(openDirectDebitPopUp()); },
   setCheckoutFormHasBeenSubmitted: () => { dispatch(setCheckoutFormHasBeenSubmitted()); },
   createOneOffPayPalPayment: (data: CreatePaypalPaymentData) => { dispatch(createOneOffPayPalPayment(data)); },
-  setStripeV3HasLoaded: () => { dispatch(setStripeV3HasLoaded); },
+  setStripeV3HasLoaded: () => { dispatch(setStripeV3HasLoaded()); },
```
tomrf1: this was broken previously

support-frontend/assets/pages/contributions-landing/components/PaymentMethodSelector.jsx
```diff
@@ -55,11 +52,8 @@ type PropTypes = {|
   existingPaymentMethods: ExistingPaymentMethod[] | typeof undefined,
   paymentMethod: PaymentMethod,
   existingPaymentMethod: RecentlySignedInExistingPaymentMethod,
-  onPaymentAuthorisation: PaymentAuthorisation => void,
```
tomrf1: just removing some unused props in this file

support-frontend/assets/pages/contributions-landing/components/StripePaymentRequestButton/StripePaymentRequestButtonContainer.jsx
```diff
@@ -30,49 +31,35 @@ type PropTypes = {|
...
-      'load',
-      setStripeHasLoaded,
-    );
-  }
-};
-
 // ----- Component ----- //
 
-function StripePaymentRequestButtonContainer(props: PropTypes) {
+class StripePaymentRequestButtonContainer extends React.Component<PropTypes, void> {
```
tomrf1: This needs to be a class because we need to call `setStripeHasLoaded` on `componentDidMount`

support-frontend/assets/pages/contributions-landing/contributionsLandingActions.js
```diff
@@ -89,13 +93,14 @@ const updateContributionType = (contributionType: ContributionType): ((Function)
...
-  storage.setSession('selectedPaymentMethod', paymentMethod);
-  return ({ type: 'UPDATE_PAYMENT_METHOD', paymentMethod });
-};
+const updatePaymentMethod = (paymentMethod: PaymentMethod): Action =>
+  (dispatch: Function): void => {
+    // PayPal one-off redirects away from the site before hitting the thank you page
+    // so we need to store the payment method in the storage so that it is available on the
+    // thank you page in all scenarios.
+    storage.setSession('selectedPaymentMethod', paymentMethod);
+    dispatch(setFormSubmissionDependentValue(() => ({ type: 'UPDATE_PAYMENT_METHOD', paymentMethod })));
```
tomrf1: Changing the payment method is now a 'Form Submission Dependent Value' because the 'Contribute' button needs to be disabled only if the Stripe form is incomplete

support-frontend/assets/pages/contributions-landing/contributionsLandingReducer.js
```diff
@@ -130,11 +138,16 @@ function createFormReducer() {
       state: null,
       checkoutFormHasBeenSubmitted: false,
     },
+    stripeV3HasLoaded: false,
```
tomrf1: moved this out of `stripePaymentRequestButtonData` because it's now shared

paulbrown1982:
Have you tested what happens if you supply the card number that triggers SCA into the Stripe Checkout flow?

tomrf1:
@paulbrown1982 good question. I just tried it.
Payment API logs:
`Your card was declined. This transaction requires authentication`

and the client says:
```
Payment Attempt Failed
The transaction was temporarily declined. Please try entering your payment details again. Alternatively, try another payment method.
```

support-frontend/app/views/main.scala.html
jlieb10: still have no idea how it's possible that the entire html body hasn't been loaded and parsed by the time one is able to call 'document' within the parsed and executing javascript -- i wonder if this is something to do with the performance improvements that we did at the beginning of the year

Moving it to the header will cause it to start downloading earlier but the `defer` deprioritizes it executing. It may affect performance. Worth testing and debugging the above issue if it does.

support-frontend/assets/helpers/errorReasons.js
```diff
@@ -42,6 +42,9 @@ function appropriateErrorMessage(errorReason: ?ErrorReason): ?string {
       return 'We are unable to sign you up at this time';
     case 'internal_error':
       return 'Sorry, something has gone wrong. Please try again, or come back later.';
+    case 'card_authentication_error':
+      // TODO - is this copy ok?
+      return 'Your card issuer\'s authentication was unsuccessful and you have not been charged. Please try again or choose another payment method.';
```
jlieb10: I'm thinking we don't want to encourage them to try again w the same payment method if they've gotten this specific error

support-frontend/assets/helpers/abTests/abtestDefinitions.js
```diff
@@ -11,6 +11,7 @@ export type LandingPageChoiceArchitectureAmountsFirstTestVariants = 'control'
   | 'productFirstSetOne'
   | 'productFirstSetTwo'
   | 'notintest';
+export type StripeElementsTestVariants = 'control' | 'stripeCardElement' | 'notintest';
```
jlieb10: I'm confused why we are running this as a test -- isn't it a required change?

jlieb10: Or is it to measure the impact of the required change even if we don't have another option?

support-frontend/assets/pages/contributions-landing/components/ContributionForm.jsx
paulbrown1982: What would have been the impact of it being broken?

jlieb10:
Is it possible to test stripe elements in the same way we test the normal stripe checkout?

support-frontend/assets/pages/contributions-landing/components/ContributionForm.jsx
jlieb10: From what I can see, the only place this bit of state is being used is in the Stripe Payment Request Button (as `props.stripeHasLoaded`). In the case that we think it hasn't loaded, we attempt to set stripe up again but don't actually prevent anything from rendering, so the consequence is a fully functioning site that's sometimes making an unnecessary call to load Stripe again (background performance hit). I think I remember seeing two calls for Stripe when doing the performance improvements and not being able to figure out why it was happening so this is a great find.

https://github.com/guardian/support-frontend/blob/a6c3d8a0fda981d7c3ebabbe0184d2aeec0a33a9/support-frontend/assets/pages/contributions-landing/components/StripePaymentRequestButton/StripePaymentRequestButtonContainer.jsx

support-frontend/assets/pages/contributions-landing/contributionsLandingActions.js
jlieb10: I'm confused about why we need to disable the button completely instead of just error handling like we do in other cases of required fields not being filled in

support-frontend/assets/pages/contributions-landing/components/ContributionForm.jsx
tomrf1: In practice it probably never caused a problem for the Payment Request Button (through luck, the component always seems to get updated later on regardless), but the logic for setting the `stripeV3HasLoaded` state was broken.

support-frontend/assets/pages/contributions-landing/contributionsLandingActions.js
tomrf1: @jesseyuen thoughts?

support-frontend/assets/helpers/abTests/abtestDefinitions.js
tomrf1: It is required, but the deadline for SCA was put back by 18 months so it's a bit less urgent now!
We still want to get this live regardless, but this is a pretty major change to a core part of the payment flow so it's important to check that there aren't any issues. A/B testing will allow is to compare acquisitions

support-frontend/assets/helpers/paymentIntegrations/readerRevenueApis.js
```diff
@@ -105,11 +105,17 @@ export type RegularPaymentRequest = {|
 export type StripePaymentMethod = 'StripeCheckout' | 'StripeApplePay' | 'StripePaymentRequestButton';
 export type StripePaymentRequestButtonMethod = 'none' | StripePaymentMethod;
 
-export type StripeAuthorisation = {|
```
tomrf1: Split `StripeAuthorisation` into `StripeCheckoutAuthorisation` and `StripePaymentIntentAuthorisation` because one needs a `token` and one needs a `paymentMethodId`. After migration, `StripeCheckoutAuthorisation` can be removed

support-frontend/assets/helpers/abTests/abtestDefinitions.js
jlieb10: glad to know it's less urgent now

support-frontend/assets/pages/contributions-landing/components/StripeCardForm/StripeCardForm.jsx
```diff
@@ -0,0 +1,229 @@
...
+  {| name: 'Complete' |};
+
+type CardFieldName = 'CardNumber' | 'Expiry' | 'CVC';
+
+type StateTypes = {
+  [CardFieldName]: CardFieldState,
+  currentlySelected: CardFieldName | null,
+};
+
+const fieldStyle = {
```
jlieb10: I prefer styling react in this way but it breaks our convention - is it a requirement for stripe elements? do they not have constant ids or classes we can hook into?

tomrf1: Indeed, the docs say:
> No CSS styles should be applied directly to anything Elements inserts inside your container DOM element. Style the container DOM element instead.

support-frontend/app/views/main.scala.html
tomrf1: This was resolved by @jlieb10 [here](https://github.com/guardian/support-frontend/pull/2058/commits/c719765ee28e69232102108d96414beff32d16df)

PR #2059, opened 2019-08-29
feat: add contributions to the mobile menu

damien-shaw-guardian:
## Why are you doing this?
As part of the traffic drivers for subscriptions, the mobile header cta now links to subscriptions, this means that getting to contributions (where the previous mobile header cta linked to) will need to be accessed from subscriptions. This PR puts a link to contributions in the mobile menu

[**Trello Card**](https://trello.com/c/aXNyBKDt/2587-add-contributions-to-the-mobile-nav-on-support-frontend)

## Changes

* feat: add contributions to the mobile menu

## Screenshots
Screen Shot 2019-09-02 at 15 34 44

### DIFF
--- a/support-frontend/assets/components/headers/header/header.scss
+++ b/support-frontend/assets/components/headers/header/header.scss
@@ -319,10 +319,11 @@ $header-links-min-width: gu-span(12);
   }
 }
 
-.component-header-links__li.show-on-tablet {
-  display: none;
+.component-header-links__li.component-header-links__li--show-on-tablet {
+  display: block;
 
-  @include mq($until: desktop) {
-    display: block;
+  // custom breakpoint - we are using a 'custom breakpoint' because this is linked to the menu toggle which has a custom break point applied in js
+  @media (min-width: 884px) {
+    display: none;
   }
 }
--- a/support-frontend/assets/components/headers/header/header.scss
+++ b/support-frontend/assets/components/headers/header/header.scss
@@ -253,12 +253,12 @@ $header-links-min-width: gu-span(12);
         color: gu-colour(neutral-100);
         @include gu-fontset-body-sans;
         border-right: 1px solid gu-colour(brand-pastel);
-  
+
         @include mq($until: tablet) {
           flex: none;
           border-right: none;
         }
-      
+
         .component-header-topnav--checkout-text {
           border-top: 1px solid gu-colour(brand-pastel);
           border-left: 1px solid gu-colour(brand-pastel);
@@ -270,18 +270,18 @@ $header-links-min-width: gu-span(12);
             display: inline-block;
             vertical-align: top;
           }
-  
+
           @include mq($until: tablet) {
             border-left: none;
             border-top: none;
             padding-left: $gu-h-spacing /4;
           }
         }
-      
+
         @include mq($from: tablet, $until: desktop) {
           display: none;
         }
-  
+
         @include mq($from: leftCol) {
           flex: 0 0 650px;
         }
@@ -316,5 +316,13 @@ $header-links-min-width: gu-span(12);
 
   @include mq($until: tablet) {
     display: none;
+  }
+}
+
+.component-header-links__li.show-on-tablet {
+  display: none;
+
+  @include mq($until: desktop) {
+    display: block;
   }
 }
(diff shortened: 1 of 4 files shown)


jfsoul:
This should be the second link in the nav, rather than second-last.

Also it should be mobile only because I don't think there's room on desktop - is that the case? Screenshot would be good.

support-frontend/assets/helpers/deliveryCheck.js
```diff
@@ -5,7 +5,7 @@ const M25_POSTCODE_PREFIXES = [
   'CR0', 'CR2', 'CR3', 'CR4', 'CR5', 'CR6', 'CR7', 'CR8', 'CR9',
   'DA1', 'DA14', 'DA15', 'DA16', 'DA18', 'DA5', 'DA7', 'DA8', 'DA17',
   'E1', 'E10', 'E11', 'E12', 'E13', 'E14', 'E15', 'E16', 'E17', 'E18', 'E20', 'E1W', 'E2', 'E3', 'E4', 'E5', 'E6', 'E7', 'E8', 'E9', 'E98',
-  'EC1A', 'EC1M', 'EC1N', 'EC1R', 'EC1V', 'EC1Y', 'EC2A', 'EC2M', 'EC2N', 'EC2R', 'EC2V', 'EC2Y', 'EC3A', 'EC3M', 'EC3N', 'EC3P', 'EC3R', 'EC3V', 'EC4A', 'EC4M', 'EC4N', 'EC4P', 'EC4R', 'EC4V','EC4Y',
+  'EC1A', 'EC1M', 'EC1N', 'EC1R', 'EC1V', 'EC1Y', 'EC2A', 'EC2M', 'EC2N', 'EC2R', 'EC2V', 'EC2Y', 'EC3A', 'EC3M', 'EC3N', 'EC3P', 'EC3R', 'EC3V', 'EC4A', 'EC4M', 'EC4N', 'EC4P', 'EC4R', 'EC4V', 'EC4Y',
```
jfsoul: 😄

support-frontend/assets/components/headers/links/links.jsx
```diff
@@ -73,14 +80,15 @@ const Links = ({ location, getRef, countryGroupId }: PropTypes) => (
...
         }) => (
           <li
-            className={
+            className={[
               classNameWithModifiers(
                 'component-header-links__li',
                 [window.location.href.endsWith(href) ? 'active' : null],
-              )
+              ), (additionalClasses || null),
+            ].join(' ')
```
jfsoul: I think you could use the classnames utility we import to just do something like:

```
import 'cx' from 'classnames'

...

className={cx(
  classNameWithModifiers(
    ...
  ), additionalClasses)
}
```

support-frontend/assets/components/headers/header/header.scss
```diff
@@ -318,3 +318,11 @@ $header-links-min-width: gu-span(12);
     display: none;
   }
 }
+
+.component-header-links__li.show-on-tablet {
```
jfsoul: maybe this should be a modifier looking for an additional class

`.component-header-links__li--show-on-tablet`

support-frontend/assets/components/headers/header/header.scss
```diff
@@ -318,3 +318,12 @@ $header-links-min-width: gu-span(12);
     display: none;
   }
 }
+
+.component-header-links__li.component-header-links__li--show-on-tablet {
```
johnduffell: I'm not totally up on BEM, but I think "show on tablet" is a utility class rather than a modifier?

damien-shaw-guardian: The idea is to have this locally scoped to the header links, otherwise I would agree

support-frontend/assets/components/headers/header/header.scss
```diff
@@ -253,12 +253,12 @@ $header-links-min-width: gu-span(12);
...
         @include gu-fontset-body-sans;
         border-right: 1px solid gu-colour(brand-pastel);
-  
+
         @include mq($until: tablet) {
           flex: none;
           border-right: none;
         }
-      
+
```
johnduffell: which editor setting removes these spaces? I haven't seen much of this kind of change, but might be worth us all making this setting on to avoid this noise in future. @lucymonie any idea?
( from https://github.com/guardian/support-frontend/pull/1783/files#diff-a03618d9ea75a35d75993c4b81e6049cR261 )

PR #2060, opened 2019-08-30
Redirect /support to regional versions

jfsoul:
## Why are you doing this?
We recently added regional versions of the "Why Support" page at `/au/support` and `/uk/support`, but we hadn't yet changed the redirection from defaulting to `/uk/support`.  This does that so that when you navigate using the nav from Aus/US you should end up on the region-specific page.

[**Trello Card**](https://trello.com/c/DEsu2DM8/2586-why-support-redirect-support-always-goes-to-uk-version)

## Screenshots

image

### DIFF
--- a/support-frontend/app/controllers/Application.scala
+++ b/support-frontend/app/controllers/Application.scala
@@ -67,13 +67,13 @@ class Application(
   }
 
   def supportGeoRedirect: Action[AnyContent] = GeoTargetedCachedAction() { implicit request =>
-    val redirectUrl = request.geoData.countryGroup match {
-      case Some(US) => buildCanonicalShowcaseLink("us")
-      case Some(Australia) => buildCanonicalShowcaseLink("au")
-      case _ => buildCanonicalShowcaseLink("uk")
+    val supportPageVariant = request.geoData.countryGroup match {
+      case Some(US) => "us"
+      case Some(Australia) => "au"
+      case _ => "uk"
     }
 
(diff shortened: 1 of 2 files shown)


jfsoul:
This needs testing on CODE because it relies on fastly geo-location

support-frontend/app/controllers/Application.scala
```diff
@@ -66,8 +66,18 @@ class Application(
     RedirectWithEncodedQueryString(redirectUrl, request.queryString, status = FOUND)
   }
 
+  def supportGeoRedirect: Action[AnyContent] = GeoTargetedCachedAction() { implicit request =>
+    val redirectUrl = request.geoData.countryGroup match {
+      case Some(US) => buildCanonicalShowcaseLink("us")
+      case Some(Australia) => buildCanonicalShowcaseLink("au")
+      case _ => buildCanonicalShowcaseLink("uk")
```
johnduffell: no need to repeat `buildCanonicalShowcaseLink` three times, just do it outside

johnduffell: we are changing from a relative to an absolute url in the redirect too.  That is fine assuming it's correct, and I think it will be, but it is a subtle additional change.  https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Location#Directives

jfsoul: good point!

PR #2061, opened 2019-09-02
remove static asset S3 buckets from support frontend

johnduffell:
## Why are you doing this?

We planned to serve the assets from an S3 bucket so that there couldn't be a mistmatch between the deployed instances on separate requests as this would cause a blank or unstyled page.

Originally we planned to have an S3 bucket per env and upload to that, but riffraff doesn't support a per env bucket when it uploads the files as part of the deploy.

In the end we went for a single bucket in a separate cfn, so this original background work is no longer needed.

previous PR with explanation: https://github.com/guardian/support-frontend/pull/2016


johnduffell:
tested in code and worked fine via riffraff

johnduffell:
I have respun travis, it was complaining it lost a connection

PR #2062, opened 2019-09-02
Support viewing of an A/B test using the url, even if the test is off

tomrf1:
## Why are you doing this?
Sometimes it's useful to try out a new test in PROD without opting users into the test.
It means we can integration test with e.g. our Stripe production account before putting it live to users.

Previously `urlParticipations` was only being used to set a local storage value, so if the ab test was switch off then it had no effect.
This change means that even if an ab test is switched off, if you set `#ab-<test-name>=<variant-name>` in the url than you'll be in the ab test.

### DIFF
--- a/support-frontend/assets/helpers/abTests/abtest.js
+++ b/support-frontend/assets/helpers/abTests/abtest.js
@@ -222,9 +222,10 @@ const init = (
   const mvt: number = getMvtId();
   const participations: Participations = getParticipations(abTests, mvt, country, countryGroupId);
   const urlParticipations: ?Participations = getParticipationsFromUrl();
-  setLocalStorageParticipations({ ...participations, ...urlParticipations });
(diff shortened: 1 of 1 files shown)


tomrf1:
Yes I think the existing behaviour was strange. It would set the test/variant in local storage but not return it in the resulting `participations` object, meaning you had to reload the page again to actually be put in the test.

Now, if the test is switched off then the only way to opt in is to include it in the url. Anything in local storage is ignored if the test is off

johnduffell:
sounds good.  I think "opt in" is not quite true as it's not sticky (I can understand why though)  it's more like a "show for one page view"

However I still find this clearer than the other PR!  Thanks for exploring the options
👍

tomrf1:
@johnduffell good point regarding "opt-in", have changed the title

PR #2063, opened 2019-09-02
SCA: support workers: decide zuora payment gateway when deciding zuora payment method

johnduffell:
## Why are you doing this?

_Note: "payment method" is overloaded in this description, so be careful when reading._

This is part of the work to make SCA work with recurring payments.
We have made the support workers able to accept the old stripeToken and the new [stripe paymentMethod](https://stripe.com/docs/payments/payment-methods) and deal with them as appropriate: https://github.com/guardian/support-frontend/pull/2053 .  Once the customer has been POSTed into stripe, we have a customer id and either a card token or a stripe payment method which need to be written into zuora.  Finally, zuora needs to call either the charges or the payment intent api on stripe when it charges the customer.
Zuora has implemented the payment intent version as a [new payment gateway](https://knowledgecenter.zuora.com/CB_Billing/M_Payment_Gateways/Supported_Payment_Gateways/Stripe_Payment_Gateway).  This means we need to choose which payment gateway to be used based not just on the currency and stripe payment method, but also based on whether it's a card or stripe payment method token.

This PR is mostly a refactoring, at present the decision on which gateway is made by pattern matching on the type of the [support workers PaymentMethod](https://github.com/guardian/support-frontend/blob/master/support-models/src/main/scala/com/gu/support/workers/PaymentMethods.scala), which a later tweak to look at currency for the stripe (AUD).  This feels like an optimisation that doesn't always apply, so I have made it so the gateway is determined at the same time as the support workers PaymentMethod.

A following PR will make it choose the new gateway in the case where it's a stripe payment method.

[**Trello Card**](https://trello.com/c/3EXrocav/2589-make-support-workers-write-to-zuora-in-sca-mode)

## TO do
- test in code
- add units tests


build.sbt
```diff
@@ -7,6 +7,7 @@ skip in publish := true
 lazy val integrationTestSettings: Seq[Def.Setting[_]] = Defaults.itSettings ++ Seq(
   scalaSource in IntegrationTest := baseDirectory.value / "src" / "test" / "scala",
   javaSource in IntegrationTest := baseDirectory.value / "src" / "test" / "java",
+  resourceDirectory in IntegrationTest := baseDirectory.value / "src" / "test" / "resources",
```
johnduffell: the it tests didn't actually run for all the projects, as some of them had classpath resources that were not built into it-classes (as it was looking for it resources rather than normal ones).  This fixes that.

build.sbt
```diff
@@ -29,7 +30,7 @@ lazy val release = Seq[ReleaseStep](
 
 lazy val commonSettings = Seq(
   organization := "com.gu",
-  scalaVersion := "2.12.7",
+  scalaVersion := "2.12.8",
```
johnduffell: just to be up to date

build.sbt
```diff
@@ -81,6 +82,7 @@ lazy val `support-frontend` = (project in file("support-frontend"))
     (testOnly in Test) := ((testOnly in Test) dependsOn testScalastyle).evaluated,
     (testQuick in Test) := ((testQuick in Test) dependsOn testScalastyle).evaluated,
   ).dependsOn(`support-services`, `support-models`, `support-config`, `support-internationalisation`)
+  .aggregate(`support-services`, `support-models`, `support-config`, `support-internationalisation`)
```
johnduffell: previously the support-frontend build used to run test in the root, which would run all the tests in all projects.  As part of the snyk monitor changes we removed that extra step as we decided it was slowing things down, and we made sure the support-workers build was also reported to the PR status.
I just realised that in that case the tests for the depended on projects were not being run, as they were only aggregated at the root. https://www.scala-sbt.org/1.x/docs/Multi-Project.html#Aggregation

This is overall a change from before.  Originally if the tests were broken anywhere, support frontend wouldn't build, but support-workers would build unless its own direct tests were broken.
Now, support-workers and support-frontend will go out if their own test and dependent tests are all passing, but if their own tests only are broken, the other would still build successfully.

I don't feel like that's an issue, but there may be other effects I don't know of?

project/plugins.sbt
```diff
@@ -27,7 +27,7 @@ addSbtPlugin("net.vonbuchholtz" % "sbt-dependency-check" % "0.2.0")
 // dependency tracker plugin - needed for snyk cli integration
 addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.9.0")
 
-addSbtPlugin("com.typesafe.sbt" % "sbt-native-packager" % "1.1.1")
+addSbtPlugin("com.typesafe.sbt" % "sbt-native-packager" % "1.4.1")
```
johnduffell: just keeping up to date

support-models/src/main/scala/com/gu/support/workers/PaymentMethods.scala
```diff
@@ -4,11 +4,13 @@ import cats.syntax.functor._
 import com.gu.i18n.Country
 import com.gu.support.encoding.Codec
 import com.gu.support.encoding.Codec.capitalizingCodec
+import com.gu.support.zuora.api.{DirectDebitGateway, PayPalGateway, PaymentGateway}
 import io.circe.syntax._
 import io.circe.{Decoder, Encoder}
 
 sealed trait PaymentMethod {
   def `type`: String
+  def paymentGateway: PaymentGateway
```
johnduffell: now the actual Zuora payment gateway name is defined when the zuora payment method is determined.  Before it was determined at the point of writing the subscription to zuora based on this object type and the currency.

support-models/src/main/scala/com/gu/support/zuora/api/PaymentGateway.scala
```diff
@@ -14,12 +14,6 @@ sealed trait PaymentGateway {
 }
 
 object PaymentGateway {
-  def forPaymentMethod(paymentMethod: PaymentMethod, currency: Currency): PaymentGateway = paymentMethod match {
-    case _: PayPalReferenceTransaction => PayPalGateway
-    case _: CreditCardReferenceTransaction if currency == AUD => StripeGatewayAUD
-    case _: CreditCardReferenceTransaction => StripeGatewayDefault
-    case _: DirectDebitPaymentMethod | _: ClonedDirectDebitPaymentMethod => DirectDebitGateway
-  }
```
johnduffell: this is redundant as we store it on the object using inheritance instead of pattern matching

support-models/src/main/scala/com/gu/support/zuora/api/response/Responses.scala
```diff
@@ -80,14 +81,15 @@ case class GetAccountResponse(success: Boolean, basicInfo: BasicInfo) extends Zu
...
-   Currency: String
-  )
+  IdentityId__c: Option[String],
+  sfContactId__c: Option[String],
+  CrmId: Option[String],
+  DefaultPaymentMethodId: Option[String],
+  AutoPay: Boolean,
+  Balance: Double,
+  Currency: String,
+  PaymentGateway: PaymentGateway
```
johnduffell: when duplicating a payment method, we now read the gateway from the zuora repsponse so that we maintain the old one in the new account regardless of whether it's New SCA or Old token.

PR #2065, opened 2019-09-03
Modifications to enable a third US-based Stripe account

paulbrown1982:
Reverts guardian/support-frontend#2064, re-enabling, now that support-workers configs have been properly set:

Modifications to enable a third US-based Stripe account. Main changes are that the public key is chosen based on the page's country not the transacting currency. Also that the public key needs to be transmitted along with the payment and analytics information over to the Payment API so that it can be completely sure which account the token is generated from, and doesn't duplicate the based-on-country algorithm.

Use of the new key is controlled by a new 'Payment Method' feature switch which I need to add to the [admin console](https://github.com/guardian/support-admin-console).

## Why are you doing this?
To enable Discover and Diners Club cards to be used by USA customers. Also to move less money around in the back-office between UK and US-based USD company bank accounts.

[**Trello Card**](https://trello.com/c/23rljvzx)

## Screenshots
No fundamental changes to UX other than new cards (Diners and Discover) will be accepted. 

Before / with the Feature Switch off:
Screen Shot 2019-09-02 at 13 21 26
Screen Shot 2019-09-02 at 13 20 33
Screen Shot 2019-09-02 at 13 21 12

After / with the Feature Switch on:
Screen Shot 2019-09-02 at 10 22 26
Screen Shot 2019-09-02 at 13 20 07
Screen Shot 2019-09-02 at 13 19 44

### DIFF
--- a/support-config/src/main/resources/touchpoint.PROD.conf
+++ b/support-config/src/main/resources/touchpoint.PROD.conf
@@ -1,8 +1,12 @@
 touchpoint.backend.environments {
   PROD {
     stripe {
-      AUD {
+      AU {
         api.key.public = "pk_live_57xV50eFjPIA990PEGWJwoHp"
+        api.key.secret = "dummy-key"
+      }
+      US {
+        api.key.public = "pk_live_2O6zPMHXNs2AGea4bAmq5R7Z"
         api.key.secret = "dummy-key"
       }
       default {
--- a/support-config/src/main/scala/com/gu/support/config/StripeConfig.scala
+++ b/support-config/src/main/scala/com/gu/support/config/StripeConfig.scala
@@ -1,24 +1,47 @@
 package com.gu.support.config
 
-import com.gu.i18n.Currency
+import com.gu.i18n.{Country, Currency}
 import com.gu.i18n.Currency.AUD
 import com.gu.monitoring.SafeLogger
 import com.typesafe.config.Config
 
 
-case class StripeConfig(defaultAccount: StripeAccountConfig, australiaAccount: StripeAccountConfig, version: Option[String] = None)
+case class StripeConfig(defaultAccount: StripeAccountConfig,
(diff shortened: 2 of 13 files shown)


support-config/src/main/resources/touchpoint.PROD.conf
```diff
@@ -1,10 +1,14 @@
...
   PROD {
     stripe {
-      AUD {
+      AU {
         api.key.public = "pk_live_57xV50eFjPIA990PEGWJwoHp"
         api.key.secret = "dummy-key"
       }
+      US {
+        api.key.public = "pk_live_2O6zPMHXNs2AGea4bAmq5R7Z"
+        api.key.secret = "dummy-key"
```
johnduffell: I'm not sure what the dummy-key is for (although it was a useful error id when it went wrong)
Are we having a retro as I do think we could have some it:test for the config which could catch this if they are run before merging.  Same idea as here:
https://github.com/guardian/support-service-lambdas/blob/master/lib/s3ConfigValidator/src/test/scala/com/gu/test/S3ConfigFilesEffectsTest.scala

support-config/src/main/scala/com/gu/support/config/StripeConfig.scala
```diff
@@ -1,24 +1,47 @@
...
 import com.gu.i18n.Currency.AUD
 import com.gu.monitoring.SafeLogger
 import com.typesafe.config.Config
 
 
-case class StripeConfig(defaultAccount: StripeAccountConfig, australiaAccount: StripeAccountConfig, version: Option[String] = None)
+case class StripeConfig(defaultAccount: StripeAccountConfig,
+                        australiaAccount: StripeAccountConfig,
+                        unitedStatesAccount: StripeAccountConfig,
+                        version: Option[String])
```
johnduffell: to stop intellij doing this change these settings 
image

PR #2069, opened 2019-09-05
make tests that involve futures wait for the future single threadedly

johnduffell:
TODO
- flesh out description more with some links
- tidy up the code (this is the concept to get feedback, but there are probably syntactic and other low level simplifications even if we stick with the overall approach)
- test in CODE

## Why are you doing this?

This PR is a rough sketch (that works in testing) in order to get feedback on the approach.

----

Some test runs for support-workers were failing with a concurrency error.  This seems to be a scalatest bug, but triggered by us using a non serial execution context.  Someone else has hit it, and exploring the source code, scalatest doesn't seem to wait for the result properly (and may even not notice test failures) https://github.com/typelevel/cats-effect/issues/49

There's no reason we shouldn't use the built in serial one, so I decided to take the easy option, and make it work serially (it was suffering thread starvation in the Await call)  As otherwise we will end in concurrency whack a mole.  The concurrency issue was discovered during the migration to mono repo, but I'm not sure of any more background https://github.com/guardian/support-frontend/commit/d9f89a63274aad81457f9c355aa361e698cb49ca

See discussion on RR dev channel.

[**Trello Card**](https://trello.com/c/pBeYMyJn/2603-fix-parallel-execution-bug-in-tests)

### DIFF
--- a/support-workers/src/main/scala/com/gu/support/workers/lambdas/Handler.scala
+++ b/support-workers/src/main/scala/com/gu/support/workers/lambdas/Handler.scala
@@ -2,6 +2,7 @@ package com.gu.support.workers.lambdas
 
 import java.io.{InputStream, OutputStream}
 
+import cats.{Applicative, Functor, Id}
 import com.amazonaws.services.lambda.runtime.{Context, RequestStreamHandler}
 import com.gu.support.workers.exceptions.ErrorHandler
 import com.gu.support.workers.{ExecutionError, RequestInfo}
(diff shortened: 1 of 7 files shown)


support-workers/src/main/scala/com/gu/salesforce/SalesforceService.scala
```diff
@@ -177,9 +177,7 @@ class AuthService(config: SalesforceConfig)(implicit ec: ExecutionContext)
         "grant_type" -> Seq("password")
       ))
 
-    Retry(3) {
-      postAuthRequest
-    }
+    postAuthRequest
```
johnduffell: I think the whole lambda would retry anyway if there was an auth failure?  If not, it would still be more sensible to retry at that level.  I noticed this because when the test was deadlocking, with a 30 second timeout, it was actually taking >90 seconds.  I then noticed the retrying in the test logs.

support-workers/src/main/scala/com/gu/support/workers/lambdas/Handler.scala
```diff
@@ -10,43 +10,42 @@ import io.circe.{Decoder, Encoder}
 import scala.concurrent.duration._
 import scala.concurrent.{Await, ExecutionContext, Future}
 
-abstract class Handler[T, R](implicit decoder: Decoder[T], encoder: Encoder[R]) extends RequestStreamHandler {
+abstract class Handler[T, R](
```
johnduffell: I have folded Handler and FutureHandler together as the Handler was never used on its own.

support-workers/src/main/scala/com/gu/support/workers/lambdas/Handler.scala
```diff
@@ -10,43 +10,42 @@ import io.circe.{Decoder, Encoder}
...
-    ec: ExecutionContext
-) extends Handler[T, R] {
+  def FutureHandlerResult(r: R, ri: RequestInfo): FutureHandlerResult = Future.successful((r, ri)) // scalastyle:ignore method.name
 
-  type FutureHandlerResult = Future[(R, RequestInfo)]
+  override def handleRequest(is: InputStream, os: OutputStream, context: Context): Unit =
+    Await.result(
+      handleRequestFuture(is, os, context),
+      Duration(context.getRemainingTimeInMillis.toLong, MILLISECONDS)
+    )
```
johnduffell: now the Awaiting only happens at the very outside of the lambda.  There shouldn't be any Awaits anywhere further in.

support-workers/src/main/scala/com/gu/support/workers/lambdas/Handler.scala
```diff
@@ -10,43 +10,42 @@ import io.circe.{Decoder, Encoder}
...
-  def FutureHandlerResult(r: R, ri: RequestInfo): FutureHandlerResult = Future.successful((r, ri)) // scalastyle:ignore method.name
+  def handleRequestFuture(is: InputStream, os: OutputStream, context: Context): Future[Unit] = {
+    val eventualUnit: Future[Unit] = for {
+      inputData <- Future.fromTry(in(is))
+      (input, error, requestInfo) = inputData
+      result <- handlerFuture(input, error, requestInfo, context)
+      _ <- Future.fromTry(out(result._1, result._2, os))
+    } yield ()
+    eventualUnit.recover {
+      case t => ErrorHandler.handleException(t)
```
johnduffell: because this now uses Future rather than try..catch it won't catch the internal errors e.g. outofmemory https://www.scala-lang.org/api/current/scala/util/control/NonFatal$.html
I'm not sure if we would hope to do anything with those though?

support-workers/src/main/scala/com/gu/support/workers/lambdas/ServicesHandler.scala
```diff
@@ -9,12 +9,12 @@ import io.circe.{Decoder, Encoder}
 import scala.concurrent.duration._
 import scala.concurrent.{ExecutionContext, Future}
 
-abstract class ServicesHandler[T <: StepFunctionUserState, R](servicesProvider: ServiceProvider, d: Option[Duration] = None)(
+abstract class ServicesHandler[T <: StepFunctionUserState, R](servicesProvider: ServiceProvider)(
```
johnduffell: duration wasn't used anywhere so I remove it in a few places.

support-workers/src/test/scala/com/gu/support/workers/LambdaSpec.scala
```diff
@@ -7,9 +7,7 @@ import org.scalatest.{AsyncFlatSpec, FlatSpec, Matchers}
 
 abstract class LambdaSpec extends FlatSpec with Matchers with MockContext
 
-abstract class AsyncLambdaSpec extends AsyncFlatSpec with Matchers {
-  implicit override def executionContext = scala.concurrent.ExecutionContext.Implicits.global
```
johnduffell: this is the crucial change - now we use the serial execution context built into the scala async test framework.  This doesn't seem to hit the bug where it's running more than one test at the same time for some reason, possibly due to us Awaiting during the tests and it starting the next test before unblocking the Await and completing the previous one.

support-workers/src/main/scala/com/gu/support/workers/lambdas/Handler.scala
tomrf1: 💯

support-workers/src/main/scala/com/gu/support/workers/lambdas/Handler.scala
```diff
@@ -10,43 +10,42 @@ import io.circe.{Decoder, Encoder}
...
-  def handleRequest(is: InputStream, os: OutputStream, context: Context): Unit =
-    try {
-      in(is).flatMap {
-        case (i, error, requestInfo) =>
-          val result = handler(i, error, requestInfo, context)
-          out(result._1, result._2, os)
-      }.get
-    } catch ErrorHandler.handleException
-}
+  def HandlerResult(r: R, ri: RequestInfo): HandlerResult = (r, ri) // scalastyle:ignore method.name
```
tomrf1: Separate from your change, but this should really be a case class

johnduffell: yes I almost did that, then decided it might be too cluttered and resolved to do it next time!  Same for the tuple from `in` as well.  I might go ahead anyway now that you've spotted it too 😁

support-workers/src/main/scala/com/gu/support/workers/lambdas/Handler.scala
rupertbates: Currently they would be handled as a RetryException so they would make the lambda retry. I'm not quite sure how this change would affect that but I think we'd want to retain the existing behaviour.

support-workers/src/test/scala/com/gu/support/workers/LambdaSpec.scala
rupertbates: Looks good, I assume there was a reason why this is overridden here but I can't remember why it was and if everything still functions, then 👍

support-workers/src/main/scala/com/gu/support/workers/lambdas/Handler.scala
johnduffell: I have made this chnge now

PR #2070, opened 2019-09-05
survey implemented

jlieb10:
## Why are you doing this?
We are implementing a new survey to be seen by the US audience only. 

[**Trello Card**](https://trello.com/c/Blum7pOP/1393-survey-on-preferred-payment-methods-in-the-us)

## Changes

* Implement the survey with geo-restriction
* Bump survey component above marketing component

## Screenshots
US thank you page:
US thank you

Everywhere else thank you page:
UK thank you


jranks123:
Looks good to me! 

The only question I have is how long will it run for? There is likely a cost in the number of people who go on to verify their accounts/sign up for marketing emails, so it would be good to state upfront exactly how many responses we need to get before we take it down.

jlieb10:
@jranks123 good shout! Should've included that info in the PR: I'm not sure the exact target of responses but Paul estimates less than a week-long run, and Amanda has been consulted about it being higher on the page than the marketing component

PR #2075, opened 2019-09-09
Fix one-off contributions deployment tests

tomrf1:
## Why are you doing this?
We're currently running an A/B test on the one-off contributions page, with Stripe Elements.
This breaks these post-deploy tests.
This is a temporary fix to force the browser into the control variant (with stripe checkout). I've also reinstated the "Check Stripe pop-up appears" test.

TODO - rewrite tests for stripe elements

### DIFF
--- a/support-frontend/test/selenium/contributions/OneOffContributionsSpec.scala
+++ b/support-frontend/test/selenium/contributions/OneOffContributionsSpec.scala
@@ -1,9 +1,10 @@
 package selenium.contributions
 
+import org.openqa.selenium.JavascriptExecutor
 import org.scalatest.concurrent.Eventually
 import org.scalatest.time.{Minute, Seconds, Span}
 import org.scalatest.{BeforeAndAfter, BeforeAndAfterAll, FeatureSpec, GivenWhenThen}
-import selenium.contributions.pages.{ContributionsLanding, ContributionThankYou}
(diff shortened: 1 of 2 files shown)


support-frontend/test/selenium/contributions/OneOffContributionsSpec.scala
```diff
@@ -13,7 +14,14 @@ class OneOffContributionsSpec extends FeatureSpec with GivenWhenThen with Before
 
   override implicit val patienceConfig = PatienceConfig(Span(1, Minute), Span(5, Seconds))
 
-  before { driverConfig.reset() }
+  before {
+    driverConfig.reset()
+    // TODO - change tests to support new Stripe Elements UI
+    webDriver.asInstanceOf[JavascriptExecutor].executeScript(
+      """window.localStorage.setItem('gu.support.abTests',""" +
```
rupertbates: Is this the only way to set local storage? Is there nothing equivalent to `driverConfig.addCookie()`?

tomrf1: there is a way to access local storage, but I cannot get it to work properly. Not sure it's worth it
https://medium.com/@jonashavers/accessing-sessionstorage-and-localstorage-with-selenium-remotewebdriver-f3935d8d7d9b

PR #2082, opened 2019-09-12
use sbt 1.3 and scala 2.12.10

johnduffell:
## Why are you doing this?

two new updates are out
Sbt 1.3 https://github.com/sbt/sbt/releases/tag/v1.3.0
scala 2.12.10 https://github.com/scala/scala/releases/tag/v2.12.10

SBT has a new turbo mode that keeps the libraryDependencies warm in a classloader so that if you keep running `test` or `run` in the same sbt session it will start much quicker and the usual JIT benefits over time.  It is experiemental though, and seems to cause the support-workers build to hang, so I have removed it again.  It does make things a lot quicker if you keep running `test` locally.

[**Trello Card**](https://trello.com/c/dMbnejd3/2596-try-sbt-130-poss-big-compile-test-performance-improvement)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: scala
 scala:
-- 2.12.8
+- 2.12.10
 jdk:
 - openjdk8
 sudo: false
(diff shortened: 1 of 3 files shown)


johnduffell:
the support-workers build was hanging after running the tests so I have removed turbo mode again.

PR #2084, opened 2019-09-16
add new paymentintents gateway to allow reuse

johnduffell:
## Why are you doing this?

Now people have stripe paymentintents card details in zuora, after @twrichards updates last week
However it is possible for people to reuse card details, which means that the step functions must pick up the correct gateway together with the card details when fetching the existing account.

Some of this work is already done, to pass through the gatway correctly rather than just assuming the standard stripe gateway (otherwise it would have written incorrect data silently!), however the actual work to add the new gatway names was not present.  This means that when it tried to read the account from zuora, the gatway name was not understood and the lambda threw an error.

This change adds the new payment gateway names so that the step functions can proxy the gateway through for reuse.


twrichards:
✅ TEST IN CODE
🚀 time to ship 🚢

PR #2088, opened 2019-09-17
Add promo code for the Sept 2019 DP sale

rupertbates:
## Why are you doing this?
In order for the upcoming Digital Pack sale to work correctly we need to ensure that the landing page is using the correct promo code to fetch prices and also that it passes that promo code on to the checkout page.

[**Trello Card**](https://trello.com/c/1Acz8BH2/2628-set-up-digital-pack-pages-to-use-the-correct-promo-codes-for-the-september-sale)

### DIFF
--- a/support-frontend/assets/helpers/flashSale.js
+++ b/support-frontend/assets/helpers/flashSale.js
@@ -34,7 +34,7 @@ type SaleDetails = {
     price: number,
     annualPrice?: number,
     discountPercentage?: number,
-    saleCopy: SaleCopy,
+    saleCopy?: SaleCopy,
   },
 };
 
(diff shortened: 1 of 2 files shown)


support-frontend/assets/helpers/flashSale.js
```diff
@@ -34,7 +34,7 @@ type SaleDetails = {
     price: number,
     annualPrice?: number,
     discountPercentage?: number,
-    saleCopy: SaleCopy,
+    saleCopy?: SaleCopy,
```
rupertbates: The DP page no longer uses copy from the `Sale` object so I'm making this optional

PR #2091, opened 2019-09-20
support workers read details from stripe in SCA mode if requested

johnduffell:
## Why are you doing this?

This PR is part of a series to make our stuff SCA compatible using setupintents/paymentintents. Previous PR https://github.com/guardian/support-frontend/pull/2063

This means we can still take payments after banks start using SCA.  This is split up into smaller tasks.  This one follows on from the one that creates a customer from a payment method id.

Previously we created the customer from the token, which returned the card token/id and the card details e.g. brand, last 4, etc.  We then write those to Zuora.
With creating a customer from a payment method id, we write that id to zuora instead, but we need to call get payment method separately to get the card details.

[**Trello Card**](https://trello.com/c/3EXrocav/2589-make-support-workers-write-to-zuora-in-sca-mode)

This is tested in CODE with digital pack both stripe and gocardless still working in the old mode.
I have also tested in code with a hand crafted `"paymentFields":{"paymentMethod":"pm_card_visa"}` POSTed to the support-frontend, and it worked, the Account in zuora looked ok and I could apply a charge.

### DIFF
--- a/support-models/src/main/scala/com/gu/support/workers/PaymentFields.scala
+++ b/support-models/src/main/scala/com/gu/support/workers/PaymentFields.scala
@@ -18,13 +18,14 @@ case class StripePaymentMethodPaymentFields(paymentMethod: PaymentMethodId) exte
 
 object PaymentMethodId {
 
-  def apply(value: String): Option[PaymentMethodId] ={ println(s"value: $value")
-    if (value.forall { char =>
-      (char >= 'a' && char <= 'z') ||
-        (char >= 'A' && char <= 'Z') ||
-        (char >= '0' && char <= '9') ||
-        char == '_'
-    }) Some(new PaymentMethodId(value)) else None}
+  def apply(value: String): Option[PaymentMethodId] =
+    if (value.nonEmpty &&
+      value.forall { char =>
+        (char >= 'a' && char <= 'z') ||
+          (char >= 'A' && char <= 'Z') ||
+          (char >= '0' && char <= '9') ||
+          char == '_'
+      }) Some(new PaymentMethodId(value)) else None
 
   implicit val decoder: Decoder[PaymentMethodId] = Decoder.decodeString.emap { str =>
     apply(str).toRight(s"PaymentMethodId $str had invalid characters")
--- a/support-models/src/main/scala/com/gu/support/workers/PaymentFields.scala
+++ b/support-models/src/main/scala/com/gu/support/workers/PaymentFields.scala
@@ -1,18 +1,39 @@
 package com.gu.support.workers
 
+import cats.syntax.functor._
 import com.gu.support.encoding.Codec
 import com.gu.support.encoding.Codec.deriveCodec
-import io.circe.{Decoder, Encoder}
+import io.circe.{Encoder, _}
 import io.circe.syntax._
-import cats.syntax.functor._
 
 sealed trait PaymentFields
 
+
 case class PayPalPaymentFields(baid: String) extends PaymentFields
+
 
 sealed trait StripePaymentFields extends PaymentFields
 case class StripeSourcePaymentFields(stripeToken: String) extends StripePaymentFields // pre SCA compatibility
-case class StripePaymentMethodPaymentFields(paymentMethod: String) extends StripePaymentFields
+case class StripePaymentMethodPaymentFields(paymentMethod: PaymentMethodId) extends StripePaymentFields
+
+object PaymentMethodId {
+
+  def apply(value: String): Option[PaymentMethodId] ={ println(s"value: $value")
+    if (value.forall { char =>
+      (char >= 'a' && char <= 'z') ||
+        (char >= 'A' && char <= 'Z') ||
+        (char >= '0' && char <= '9') ||
+        char == '_'
+    }) Some(new PaymentMethodId(value)) else None}
+
+  implicit val decoder: Decoder[PaymentMethodId] = Decoder.decodeString.emap { str =>
+    apply(str).toRight(s"PaymentMethodId $str had invalid characters")
+  }
+  implicit val encoder: Encoder[PaymentMethodId] = Encoder.encodeString.contramap[PaymentMethodId](_.value)
+
+}
+case class PaymentMethodId private (value: String) extends AnyVal
+
 
 case class DirectDebitPaymentFields(
   accountHolderName: String,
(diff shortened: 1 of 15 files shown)


support-models/src/main/scala/com/gu/support/workers/PaymentFields.scala
```diff
@@ -1,18 +1,40 @@
...
+object PaymentMethodId {
+
+  def apply(value: String): Option[PaymentMethodId] =
+    if (value.nonEmpty &&
+      value.forall { char =>
+        (char >= 'a' && char <= 'z') ||
+          (char >= 'A' && char <= 'Z') ||
+          (char >= '0' && char <= '9') ||
+          char == '_'
+      }) Some(new PaymentMethodId(value)) else None
```
johnduffell: the main things we want to avoid are slashes and dots, as we concatenate this into a url and it's easier to validate than try to use a url builder to escape things.  We know that payment method ids are supposed to be fairly inoffensive, so this makes things easier to reason about.

support-models/src/main/scala/com/gu/support/workers/PaymentFields.scala
```diff
@@ -45,6 +67,15 @@ object PaymentFields {
       Decoder[StripePaymentMethodPaymentFields].widen,
       Decoder[DirectDebitPaymentFields].widen,
       Decoder[ExistingPaymentFields].widen
-    ).reduceLeft(_ or _)
+    ).reduceLeft(or(_,_))
```
johnduffell: the alternaitve OR was needed becuase the built in one discards all but the last error.  In our case that's pointless as most of the time we don't expect the last one to succeed.
The implementation I've done is not great, but this is the annoying thing about non deterministic code.
Given more motivation, I would get it to serialising the subtype as a separate data point, as per withDiscriminator or similar https://circe.github.io/circe/codecs/adt.html#the-future and then it takes the non determinism out of it

support-models/src/main/scala/com/gu/support/workers/PaymentFields.scala
```diff
@@ -45,6 +67,15 @@ object PaymentFields {
...
+
+  final def or[A, AA >: A](a: Decoder[A], d: => Decoder[AA]): Decoder[AA] = new Decoder[AA] {
+    final def apply(c: HCursor): Decoder.Result[AA] = a(c) match {
+      case r @ Right(_) => r
+      case Left(err)      => d(c).left.map { (decodingFailure: DecodingFailure) =>
+        DecodingFailure(err.message + " OR " + decodingFailure.message, err.history ++ decodingFailure.history)
```
johnduffell: this is not great, all the information is there if you squint, but it's all mashed together a bit.  It got me where I needed to be though.

support-workers/src/main/scala/com/gu/stripe/Stripe.scala
```diff
@@ -9,8 +9,6 @@ import io.circe.{Decoder, Encoder, Json}
 
 object Stripe {
 
-  sealed trait StripeObject
```
johnduffell: didn't look like this was used for anything

support-workers/src/main/scala/com/gu/stripe/Stripe.scala
```diff
@@ -39,40 +37,16 @@ object Stripe {
   }
 
   object StripeList {
-    implicit def codec[T](implicit encoder: Encoder[T], decoder: Decoder[T]): Codec[StripeList[T]] = deriveCodec[StripeList[T]]
-  }
-
-  case class StripeList[T](total_count: Int, data: Seq[T]) extends StripeObject
-
-  object Source {
```
johnduffell: source and customer are now different depending whether it's from payment method or token.

support-workers/src/main/scala/com/gu/stripe/StripeService.scala
```diff
@@ -48,4 +42,10 @@ class StripeServiceForCurrency(stripeService: StripeService, currency: Currency)
   private def getAuthorizationHeader() =
     Map("Authorization" -> s"Bearer ${config.forCurrency(Some(currency)).secretKey}")
 
-}
\ No newline at end of file
+  val createCustomerFromPaymentMethod = com.gu.stripe.createCustomerFromPaymentMethod.apply(this)_
```
johnduffell: not totally happy with this but it does keep the `stripeService.something(asdf)` syntax without making stripeService become non cohesive.  Hopefully someone can suggest a nicer way to avoid the more explicit but less auto complete friendly `Something(stripeService, asdf)`?  Or just say it's fine to go that way!

johnduffell: another option would be implicit class (extension method), but then you have to get the imports and implicits are usually a last resort

support-workers/src/main/scala/com/gu/stripe/StripeBrand.scala
```diff
@@ -0,0 +1,25 @@
+package com.gu.stripe
+
+import io.circe.Decoder
+
+// https://stripe.com/docs/api/payment_methods/object#payment_method_object-card-brand
+// See CreditCardType at https://www.zuora.com/developer/api-reference/#operation/Object_POSTPaymentMethod
+sealed abstract class StripeBrand(val value: String, val zuoraCreditCardType: Option[String])
```
johnduffell: I'm not sure why i put zuora stuff in the stripe card type, pattern match in the zuora code would have worked

support-workers/src/main/scala/com/gu/stripe/StripeBrand.scala
```diff
@@ -0,0 +1,25 @@
...
+  case object Amex extends StripeBrand("amex", Some("AmericanExpress"))
+  case object Diners extends StripeBrand("diners", Some("Diners"))
+  case object Discover extends StripeBrand("discover", Some("Discover"))
+  case object Jcb extends StripeBrand("jcb", Some("JCB"))
+  case object Mastercard extends StripeBrand("mastercard", Some("MasterCard"))
+  case object Unionpay extends StripeBrand("unionpay",  None)
+  case object Visa extends StripeBrand("visa", Some("Visa"))
+  case object Unknown extends StripeBrand("unknown", None)
+
+  val all = Seq(Amex, Diners, Discover, Jcb,Mastercard,Unionpay,Visa,Unknown)
```
johnduffell: once day we can switch to scala 3 enums and the risk of not adding to `all` will go away! https://dotty.epfl.ch/docs/reference/enums/enums.html

support-workers/src/main/scala/com/gu/stripe/StripeService.scala
rupertbates: I think this is a good approach

johnduffell:
@tomrf1 feel free to add comments in retrospect and I'll take them into account of course

support-models/src/main/scala/com/gu/support/workers/PaymentFields.scala
```diff
@@ -45,6 +67,15 @@ object PaymentFields {
       Decoder[StripePaymentMethodPaymentFields].widen,
       Decoder[DirectDebitPaymentFields].widen,
       Decoder[ExistingPaymentFields].widen
-    ).reduceLeft(_ or _)
+    ).reduceLeft(or(_,_))
+
+  final def or[A, AA >: A](a: Decoder[A], d: => Decoder[AA]): Decoder[AA] = new Decoder[AA] {
```
tomrf1: why could you not use circe's built-in behaviour for this?

tomrf1: oh sorry just saw your comment above!!

PR #2097, opened 2019-09-24
Move payment failure error message out of PaymentMethodSelector.jsx

rupertbates:
## Why are you doing this?

Up until now the error message component which is used when a payment failure occurs was inside the `PaymentMethodSelector` component. This was a problem when there is only one available payment method as it means that the error message is not visible.

This PR moves the error message out of `PaymentMethodSelector` and directly into the form.

[**Trello Card**](https://trello.com/c/BHpPtXww/2639-investigate-gw-australia-failures)

## Screenshots
Screenshot 2019-09-24 at 13 36 30

### DIFF
--- a/support-frontend/assets/pages/digital-subscription-checkout/components/digitalCheckoutForm.jsx
+++ b/support-frontend/assets/pages/digital-subscription-checkout/components/digitalCheckoutForm.jsx
@@ -68,6 +68,8 @@ import {
 } from 'helpers/productPrice/priceDescriptions';
 import { StripeProvider, Elements } from 'react-stripe-elements';
 import { getStripeKey } from 'helpers/paymentIntegrations/stripeCheckout';
+import GeneralErrorMessage
+  from 'components/generalErrorMessage/generalErrorMessage';
 
 
 // ----- Types ----- //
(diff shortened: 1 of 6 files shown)


lucymonie:
It looks good - though I think the error message isn't meant to be flush to the container. Is it missing some margin / padding?

support-frontend/assets/pages/digital-subscription-checkout/components/digitalCheckoutForm.jsx
```diff
@@ -182,6 +184,8 @@ class DigitalCheckoutForm extends Component<PropTypes, StateTypes> {
     const offerOnSelected = getAppliedPromoDescription(props.billingPeriod, productPrice);
     const helperSelected = getPriceDescription(productPrice, props.billingPeriod);
     const priceSummary = `${offerOnSelected || 'Enjoy your digital pack free for 14 days, then'} ${helperSelected}.`;
+    const submissionErrorHeading = props.submissionError === 'personal_details_incorrect' ? 'Failed to Create Subscription' :
+      'Payment Attempt Failed';
```
johnduffell: I'm not sure if the word "failed" is best fractice.  Perhaps more like "Sorry, we could not create your subscription" and "Sorry, we could not take your payment"

johnduffell:
looks good, my comment is about something that probably hasn't changed, but is always worth getting right
👍

support-frontend/assets/pages/digital-subscription-checkout/components/digitalCheckoutForm.jsx
rupertbates: I agree actually, will use something more user friendly

PR #2100, opened 2019-09-26
Support multiple stripe accounts

rupertbates:
## Why are you doing this?

To fix the following problem with the DP and GW checkouts:
* A user arrives on the checkout and we identify them as coming from UK or ROW and so use the default Stripe account
* They change their billing address to Australia, meaning we should now change the Stripe account to the Australian one, this was not happening
* The account creation fails in the support-workers step function because that (correctly) tries to use the Australian Stripe account but the token we have does not exist for that account.

[**Trello Card**](https://trello.com/c/BHpPtXww/2639-investigate-gw-australia-failures)

## Changes

* Create a new component `StripeProviderForCountry` which wraps the Stripe form and ensures that we  are using the correct account
* Use this component in all the subs checkouts. This has meant that we can change the checkout components back from classes to pure functions which is good from a simplicity point of view as this bug is actually a result of having to use class components


johnduffell:
would we ever want to use classes with their own props? I had previously assumed that not using redux for state was a bit like using a `var` in scala.  Obviously you are removing it because it is easier to solve the bug that way, but I just wondered whether we avoid state in classes as a best practice or only where it causes bugs?

rupertbates:
@johnduffell yes we strongly prefer functions over classes wherever possible. The only times we would really use local component state is if there is an expensive operation involved (for instance initialising PayPal) or if the state is purely a display implementation which would be time consuming and pointless to put into Redux such as the open or closed state of an accordion control. And I would say the second reason is still debatable.

PR #2101, opened 2019-09-27
Add Venmo as a payment option [DO NOT MERGE]

jranks123:
Things left to do: 

- Currently, Venmo is hardcoded in as a payment option for single contributions. We need to add it as a payment option to the config in the Support Admin Console, and make sure it is only on in the US

- Put behind a feature switch and make sure that the braintree library is only downloaded if the switch is on


jranks123:
Closing pending legal blockers

PR #2102, opened 2019-09-30
send redux event history to support-frontend and log it

johnduffell:
## Why are you doing this?

We have had a few more checkout failures for GW, where people are getting a ROW stripe token and submitting with AUD currency.
We don't really understand why this happens, but pushing the whole redux state to the server might help us to correlate together the steps to reproduce.

This PR just gathers all the action history and pushes it on one big string.  Then we can look at it later and try to reproduce.

We did try putting it into the support workers state, but the encryption only handles 4096KB so we could end up with failures due to the debugging, which would be ironically bad.

this is the post request from client to server
image

this is the support-frontend log
```
2019-10-01 17:17:15,567 [application-akka.actor.default-dispatcher-3] INFO  com.gu.monitoring.SafeLogger$ - f64cf20a-9a67-de3e-0000-000000000002: debug info Some({"type":"@@INIT"}
{"scope":"delivery","type":"SET_ADDRESS_LINE_1","lineOne":"sdjkhfgsd"}
{"scope":"delivery","type":"SET_TOWN_CITY","city":"sdjkhfgsd"}
{"type":"SET_POSTCODE","postCode":"sdjkhfgsd","scope":"delivery"}
{"type":"SET_BILLING_ADDRESS_IS_SAME","isSame":true}
{"type":"SET_COUNTRY_INTERNATIONALISATION","country":"GB"}
{"type":"SET_COUNTRY_CHANGED","country":"GB","scope":"billing"}
{"type":"SET_BILLING_PERIOD","billingPeriod":"SixWeekly"}
{"type":"SET_PAYMENT_METHOD","paymentMethod":"DirectDebit","country":"GB"}
{"type":"SET_PAYMENT_METHOD","paymentMethod":"Stripe","country":"GB"}
{"type":"SET_STRIPE_TOKEN","stripeToken":"tok_FudI4YFmYPwgjA"}
)
```

### DIFF
--- a/support-frontend/app/services/stepfunctions/Client.scala
+++ b/support-frontend/app/services/stepfunctions/Client.scala
@@ -11,6 +11,7 @@ import com.amazonaws.services.stepfunctions.model._
 import com.amazonaws.services.stepfunctions.{AWSStepFunctionsAsync, AWSStepFunctionsAsyncClientBuilder}
 import io.circe.Encoder
 import cats.implicits._
+import com.gu.monitoring.SafeLogger
 import services.stepfunctions.StateMachineErrors.Fail
 
 import scala.collection.JavaConversions._
@@ -35,6 +36,10 @@ class Client(client: AWSStepFunctionsAsync, arn: StateMachineArn) {
 
   private def startExecution(arn: String, input: String)(implicit ec: ExecutionContext): Response[StartExecutionResult] = convertErrors {
     AwsAsync(client.startExecutionAsync, new StartExecutionRequest().withStateMachineArn(arn).withInput(input))
+      .transform { theTry =>
+        SafeLogger.info(s"state machine result: $theTry")
+        theTry
+      }
   }
 
   def triggerExecution[T](input: T, isTestUser: Boolean, isExistingAccount: Boolean = false)(
--- a/support-frontend/app/services/stepfunctions/SupportWorkersClient.scala
+++ b/support-frontend/app/services/stepfunctions/SupportWorkersClient.scala
@@ -52,7 +52,7 @@ case class CreateSupportWorkersRequest(
   email: String,
   telephoneNumber: Option[String],
   deliveryInstructions: Option[String],
-  debugInfo: Option[String]
+  debugInfo: Option[String] = None
 )
 
 object SupportWorkersClient {
--- a/support-frontend/app/services/stepfunctions/SupportWorkersClient.scala
+++ b/support-frontend/app/services/stepfunctions/SupportWorkersClient.scala
@@ -120,7 +120,6 @@ class SupportWorkersClient(
     requestId: UUID,
     promoCode: Option[PromoCode] = None
   ): EitherT[Future, SupportWorkersError, StatusResponse] = {
-    SafeLogger.info(s"$requestId: debug info ${request.body.debugInfo}")
 
     val createPaymentMethodState = CreatePaymentMethodState(
       requestId = requestId,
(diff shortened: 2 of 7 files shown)


support-frontend/app/services/stepfunctions/Client.scala
```diff
@@ -35,6 +36,10 @@ class Client(client: AWSStepFunctionsAsync, arn: StateMachineArn) {
 
   private def startExecution(arn: String, input: String)(implicit ec: ExecutionContext): Response[StartExecutionResult] = convertErrors {
     AwsAsync(client.startExecutionAsync, new StartExecutionRequest().withStateMachineArn(arn).withInput(input))
+      .transform { theTry =>
+        SafeLogger.info(s"state machine result: $theTry")
```
johnduffell: the problem I had was the ARN was wrong in my local support-private config.  It didn't actually log that.  I don't think we need this logging, I think we should actually make `convertErrors` store the original error in the converted error so we can debug it properly.

support-frontend/app/services/stepfunctions/SupportWorkersClient.scala
```diff
@@ -119,6 +120,7 @@ class SupportWorkersClient(
     requestId: UUID,
     promoCode: Option[PromoCode] = None
   ): EitherT[Future, SupportWorkersError, StatusResponse] = {
+    SafeLogger.info(s"$requestId: debug info ${request.body.debugInfo}")
```
rupertbates: I think we should probably remove this logging. It will be pretty noisy and this is not where we need that information really.

johnduffell: Yes agree the whole point of thee exercise is to collect all the info on the support worker. feel free to remove that as I'm on the bus now!

johnduffell:
I guess the tests don't compile due to the extra field in the case class, i didn't get that far before lunch.

johnduffell:
Interesting, we might have to truncate/abbreviate it in some way, and log the full info on support frontend and just match up based on the request id.

johnduffell:
Looks like it might be just because we are encrypting too much data the fact it's a kms thing rather than a step function thing.

rupertbates:
Yes that is the reason, there is a comment describing this limitation here: https://github.com/guardian/support-frontend/blob/f6fa790df0cb836a417c6c4e3e7abf50d2d83f67/support-workers/src/main/scala/com/gu/support/workers/encoding/Encryption.scala#L45

johnduffell:
I have updated this to only send to support-frontend backend and log it there.  We will have to correlate them a bit more manually.

I have edited the title and description of this PR.

This does add to the case for just having a secure transient central store for debugging information per session or user for recent requests.

Incidentally, in the process of testing this, we worked out both a/the bug causing stripe token errors, and the EUR is not valid for DD bug.  So it may be that this extra debugging won't be needed in the short term.  But it's still worth having in some form.

johnduffell:
tested in code and looks ok, here is a cloudwatch log sample
image

support-frontend/app/services/stepfunctions/SupportWorkersClient.scala
```diff
@@ -51,7 +51,8 @@ case class CreateSupportWorkersRequest(
   supportAbTests: Set[AbTest],
   email: String,
   telephoneNumber: Option[String],
-  deliveryInstructions: Option[String]
+  deliveryInstructions: Option[String],
+  debugInfo: Option[String] = None
```
johnduffell:
```suggestion
  multiline: String,
  testing: String,
  ignore: String
```

support-frontend/app/services/stepfunctions/SupportWorkersClient.scala
johnduffell: this logging is back as sending to support workers causes a length limit to be exceeded

PR #2110, opened 2019-10-03
Change the selenium css selector used by the subscriptions landing page

rupertbates:
## Why are you doing this?

#2109 broke the post deployment test for the subscriptions showcase page because it removed an element from the page which Selenium was looking for to check that the page has loaded correctly. This PR changes the css selector to an element which does exist.

### DIFF
--- a/support-frontend/test/selenium/subscriptions/pages/SubsLandingPage.scala
+++ b/support-frontend/test/selenium/subscriptions/pages/SubsLandingPage.scala
@@ -8,7 +8,7 @@ class SubsLandingPage(implicit val webDriver: WebDriver) extends Page with Brows
 
   val url = s"${Config.supportFrontendUrl}/uk/subscribe"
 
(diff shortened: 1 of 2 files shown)


support-frontend/test/selenium/subscriptions/pages/SubsLandingPage.scala
```diff
@@ -8,7 +8,7 @@ class SubsLandingPage(implicit val webDriver: WebDriver) extends Page with Brows
 
   val url = s"${Config.supportFrontendUrl}/uk/subscribe"
 
-  private val header = cssSelector("#qa-component-subscriptions-by-country-group")
+  private val header = cssSelector(".subscriptions-landing-page")
```
johnduffell: the "qa" prefix is so we know it's there for testing, although it didn't work in this case, it's a good technique to maintain in general I think

PR #2111, opened 2019-10-04
new lambda to get stripe setup intent ids for the client side

johnduffell:
This PR is open but I need to self review it before it really makes sense.  Feel free to review now or wait until it's tidied up and I've added my own comments.

## Why are you doing this?

This is part of the programme to get stripe SCA support up and running.
This is step 1 from https://stripe.com/docs/payments/cards/saving-cards-after-payment

This is implemented as a lambda as serverless is a good thing for us to move towards where sensible, and this functionality has nothing to do with support frontend play app, and doesn't need access to most of the secrets and libraries that support-frontend has.
This is created as an independent lambda from the support-workers jar, as support workers already has a lot of dependencies and the startup time is 7+ seconds on a 51MB jar.  This jar comes out at 40+M which is mostly down to the aws s3 sdk library, but there is a lot of  huge DLLs for gitapi for some bizarre reason https://github.com/luben/zstd-jni
Keeping things in a big mono-jar is not really a sustainable way to create lambdas for other uses so I have teased apart those dependencies that we really need and tried to structure things so that the dependencies are more reasonable.  Hopefully this would be a more sustainable way if we want to make more use of lambdas in future.
This has required a bit of restructuring but hopefully I've got the essential without going overboard.

TODO 
- test on code - DONE
- add unit tests
- upload stripe live key mappings - DONE
- consider adding a health check or similar
- alarms needed - DONE

[**Trello Card**](https://trello.com/c/ut7hiijJ/2577-put-setupintent-client-secret-in-checkout-page-html)


build.sbt
```diff
@@ -49,7 +49,9 @@ lazy val commonSettings = Seq(
     "scm:git:git@github.com:guardian/support-frontend.git"
   )),
   // https://www.scala-sbt.org/1.x/docs/Cached-Resolution.html
-  updateOptions := updateOptions.value.withCachedResolution(true)
+  updateOptions := updateOptions.value.withCachedResolution(true),
+  dependencyStats / aggregate := false,
```
johnduffell: this command is useful to see what the jar size will be.

project/LibraryVersions.scala
```diff
@@ -1,7 +1,7 @@
 
 object LibraryVersions {
   val circeVersion = "0.11.1"
-  val awsClientVersion = "1.11.226"
+  val awsClientVersion = "1.11.642"
```
johnduffell: just bumping an old version

project/plugins.sbt
```diff
@@ -25,7 +25,7 @@ addSbtPlugin("com.timushev.sbt" % "sbt-updates" % "0.3.3")
 addSbtPlugin("net.vonbuchholtz" % "sbt-dependency-check" % "0.2.0")
 
 // dependency tracker plugin - needed for snyk cli integration
-addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.9.0")
+addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.10.0-RC1")
```
johnduffell: we are on sbt1.3.2 now but there are bugs with its interaction with this plugin.  this version of sbt dependency graph will appear to work rather than crash https://github.com/jrudolph/sbt-dependency-graph/issues/178#issuecomment-514529823
however it doesn't always produce the right result https://github.com/coursier/coursier/issues/1375
so we will need another version bump at some point, but this is just a small helper tool rather than something from the build itself.
Another option is just to remove it, and people can add it to their local ~/.sbt,  plugins.sbt file if they like to use it.

support-lambdas/stripe-intent/README.md
```diff
@@ -0,0 +1,14 @@
...
+stripe account.
+
+It is a lambda because it doesn't need a whole auto scaling group,
+the latency is not critical, and doing it this way provides better
+separation between codebases.
+
+It is a separate jar from the support-workers jar because the unused
+code would add to be loaded on every single step of the functions.
+Hopefully we can be a bit more cohesive and minimise dependencies if
+we separate things better.
```
johnduffell: this turned out to be harder than expected, for example because pulling in support-models pulls in acquisition-event-producer which pulls in various thrift stuff including a 3.6mb binaries of compression algorithm used by capi.  I think sometimes pulling in a "com.gu" library can get more than we bargained for as we are not always disciplined at not adding dependencies that not everyone would need.  Client side devs have some nice tooling so every import shows the size in KB of the file in the IDE, but server side we have no idea.

support-lambdas/stripe-intent/cfn.yaml
```diff
@@ -0,0 +1,95 @@
...
+      - CODE
+    Default: CODE
+
+Mappings:
+  StageMap:
+    CODE:
+      DomainName: stripe-intent-code.support.guardianapis.com
+      StripeConfig: test
+    PROD:
+      DomainName: stripe-intent.support.guardianapis.com
```
johnduffell: we (@tomrf1 @JoemG ) had a very brief discussion about appropriate host names for prod/code as they are making a similar style lambda. We have the existing code.support.dev-guardianapis.com convention but no certificate so this one is a bit easier.   But now would be a good time for objections.   These conventions are also present in support-service-lambdas repo.

support-lambdas/stripe-intent/cfn.yaml
```diff
@@ -0,0 +1,95 @@
...
+        Types:
+          - REGIONAL
+
+  BasePathMapping:
+    Type: "AWS::ApiGateway::BasePathMapping"
+    Properties:
+      RestApiId: !Ref ServerlessRestApi # auto generated by the Transform
+      DomainName: !Ref DomainName
+      Stage: !Sub Prod
+    DependsOn: ServerlessRestApiProdStage # auto generated by the Transform
```
johnduffell: this is a hack based on the predicted names of the low level resources.  AWS are working on proper support for custom domain names, at which point we can feel free to switch over to that https://github.com/awslabs/serverless-application-model/issues/783

support-lambdas/stripe-intent/src/main/scala/com/gu/handler/impure/S3Loader.scala
```diff
@@ -0,0 +1,44 @@
...
+      .build()
+
+    val s3Object: S3Object = s3Client.getObject(uri.getBucket, uri.getKey)
+
+    val source: BufferedSource = Source.fromInputStream(s3Object.getObjectContent)
+    try {
+      val conf = source.mkString
+      ConfigFactory.parseString(conf).withFallback(ConfigFactory.load())
+    } finally {
+      source.close()
```
tomrf1: you should also call `s3Object.close` here

support-lambdas/stripe-intent/src/main/scala/com/gu/stripeIntent/Handler.scala
```diff
@@ -0,0 +1,117 @@
...
+
+
+
+
+
+
+
+
+
+object ManualTest {
```
tomrf1: should this be somewhere else?

johnduffell: probably shouldn't exist at all, I think maybe an it:test would make more sense, as then we can check assertions.

support-lambdas/stripe-intent/src/main/scala/com/gu/stripeIntent/StripeService.scala
```diff
@@ -0,0 +1,61 @@
...
+import scala.reflect.ClassTag
+
+class StripeService(val config: StripePrivateKey, client: FutureHttpClient, baseUrl: String = "https://api.stripe.com/v1")(implicit ec: ExecutionContext)
+  extends WebServiceHelper[StripeError] {
+
+  // Stripe URL is the same in all environments
+  val wsUrl = baseUrl
+  val httpClient: FutureHttpClient = client
+
+  def get[A](endpoint: String)(implicit decoder: Decoder[A], errorDecoder: Decoder[StripeError], ctag: ClassTag[A]): Future[A] =
```
tomrf1: is this used?

support-lambdas/stripe-intent/src/main/scala/com/gu/handler/impure/S3Loader.scala
johnduffell: you mean just for completeness? as looking at the source code closing the Source would close the stream, and calling close on the s3 object does the same.

support-lambdas/stripe-intent/src/main/scala/com/gu/stripeIntent/StripeService.scala
johnduffell: no, good point, I'll remove

support-lambdas/stripe-intent/src/main/scala/com/gu/handler/impure/S3Loader.scala
tomrf1: oh good point, I thought it did more than that!

johnduffell: even the docs are not so clear. https://docs.aws.amazon.com/AmazonS3/latest/dev/RetrievingObjectUsingJava.html

support-workers/src/main/scala/com/gu/stripe/StripeService.scala
```diff
@@ -59,3 +66,30 @@ class StripeServiceForCurrency(val stripeService: StripeService, currency: Curre
   val getPaymentMethod = com.gu.stripe.getPaymentMethod.apply(this)_
 
 }
+
+object StripeError {
```
tomrf1: this is duplicated across 2 files

support-lambdas/stripe-intent/src/main/scala/com/gu/handler/impure/S3Loader.scala
tomrf1: indeed, but the actual code is clear!

support-workers/src/main/scala/com/gu/stripe/StripeService.scala
johnduffell: thanks yes, the whole "Stripe service" is.  In one side of my mind I wanted to go full cohesive refactor but at the moment things are more grouped by type than function.
I don't think the stripe service has any expensive dependencies of its own, but I briefly tried to move it to the new support-rest project, which they both already depend on.  However I think I got tripped up with some dependency on Stripe config.  I will have another go, as I don't think stripe client stuff has any fundamentally expensive dependencies that we don't want in support-rest.

johnduffell: ok this is done, it wasn't too bad.  Stripe service is still duplicated in name, but in reality it's configured in a different way and still shares the web service helper.

support-frontend/assets/helpers/tracking/acquisitions.js
```diff
@@ -44,7 +45,7 @@ export type ReferrerAcquisitionData = {|
   abTests: ?AcquisitionABTest[],
   // these aren't in the referrer acquisition data model on frontend, but they're convenient to include
   // as we want to include query parameters in the acquisition event to e.g. facilitate off-platform tracking
-  queryParameters: ?AcquisitionQueryParameters,
+  queryParameters: AcquisitionQueryParameters,
```
tomrf1: not sure how this change relates to this PR?

PR #2112, opened 2019-10-04
Fix Stripe account bug

rupertbates:
## Why are you doing this?
I discovered 2 more bugs with our Stripe implementation related to switching between different Stripe accounts.

1. The billing address country for users whose billing address and delivery address are the same is set when the 'Is the billing address the same as the delivery address?' checkbox is checked and then never updated. This country was then being used directly to determine the currency in the `submitForm()` function of submit.js. I have updated this function to now use the `getAddresses()` function which takes account of the `billingAddressIsSame` field when determining the billing address to use.

2. The `StripeProviderForCountry` component was not updating its state correctly so continued to use the first Stripe account loaded. Adding a `key` prop with the country value fixes this by forcing a state reset when the country changes. [More info here](https://medium.com/@albertogasparin/forcing-state-reset-on-a-react-component-by-using-the-key-prop-14b36cd7448e)

[**Trello Card**](https://trello.com/c/BHpPtXww/2639-fix-gw-australia-related-failures)


support-frontend/assets/helpers/subscriptionsForms/submit.js
```diff
@@ -119,9 +119,11 @@ function buildRegularPaymentRequest(
     deliveryInstructions,
   } = state.page.checkout;
 
+  const addresses = getAddresses(state);
+
   const price = getProductPrice(
     productPrices,
-    state.page.billingAddress.fields.country,
```
rupertbates: This value should not be used directly because it may be wrong if `billingAddressIsSame` is true

johnduffell: for long term safety, would it be worth making the "update address" action update both in the state?

rupertbates: I was thinking about that, it's quite tricky to do because each address is managed by a separate reducer. It probably is worth investigating though.

rupertbates: I was thinking about that, it's quite tricky to do because each address is managed by a separate reducer. It probably is worth investigating though.

support-frontend/assets/pages/weekly-subscription-checkout/components/weeklyCheckoutForm.jsx
```diff
@@ -334,6 +334,7 @@ function WeeklyCheckoutForm(props: PropTypes) {
             title="Your card details"
           >
             <StripeProviderForCountry
+              key={props.country} // this is necessary to force the component to use the correct Stripe account, see https://reactjs.org/docs/lists-and-keys.html#keys
```
johnduffell: rather than trying to remember it everywhere, could we just put it in one place here(see below link), and base it on the actual stripe public key rather than the country?
https://github.com/guardian/support-frontend/blob/950743bb04089deec52051ac233484d5d6a8209a/support-frontend/assets/components/subscriptionCheckouts/stripeForm/stripeProviderForCountry.jsx#L21

support-frontend/assets/helpers/subscriptionsForms/submit.js
johnduffell: yes definitely doesn't have to be in this pr, more of a general consideration in case people use auto complete in future

support-frontend/assets/pages/weekly-subscription-checkout/components/weeklyCheckoutForm.jsx
rupertbates: I thought I'd tried it at that level without success, but trying it again it appears to work so 👍

PR #2114, opened 2019-10-07
Remove Stripe Checkout for single contributions

tomrf1:
## Why are you doing this?
In the past we have used Stripe Checkout for contributions.
We have been running an A/B test to compare Stripe Checkout vs Stripe Elements + Payment Intents API.
This is for SCA compliance. Currently it is for single contributions only.
From the test we have found that the change does not impact revenue, and so are switching to 100% Stripe Elements + Payment Intents API.

This PR removes Stripe Checkout for single contributions. Recurring will be migrated later (it requires some separate backend work)

### DIFF
--- a/support-frontend/assets/helpers/paymentIntegrations/oneOffContributions.js
+++ b/support-frontend/assets/helpers/paymentIntegrations/oneOffContributions.js
@@ -59,7 +59,7 @@ export type StripeChargeData = {|
   paymentData: {
     currency: IsoCurrency,
     amount: number,
-    token: string,
+    token: string, // TODO - remove this field once it has been removed from payment-api
     email: string,
     stripePaymentMethod: StripePaymentMethod,
   },
@@ -134,19 +134,6 @@ function paymentResultFromObject(
   return Promise.resolve(PaymentSuccess);
 }
 
-// Sends a one-off payment request to the payment API and standardises the result
-// https://github.com/guardian/payment-api/blob/master/src/main/resources/routes#L17
-function postOneOffStripeExecutePaymentRequest(
-  data: StripeChargeData,
-  setGuestAccountCreationToken: (string) => void,
-  setThankYouPageStage: (ThankYouPageStage) => void,
-): Promise<PaymentResult> {
-  return logPromise(fetchJson(
-    paymentApiEndpointWithMode(`${window.guardian.paymentApiStripeUrl}/contribute/one-off/stripe/execute-payment`),
-    requestOptions(data, 'omit', 'POST', null),
-  ).then(result => paymentResultFromObject(result, setGuestAccountCreationToken, setThankYouPageStage)));
-}
-
 function postStripeCreatePaymentIntentRequest(data: CreateStripePaymentIntentRequest): Promise<Object> {
   return fetchJson(
     paymentApiEndpointWithMode(`${window.guardian.paymentApiStripeUrl}/contribute/one-off/stripe/create-payment`),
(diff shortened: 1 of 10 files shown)


support-frontend/test/selenium/contributions/OneOffContributionsSpec.scala
```diff
@@ -59,50 +59,19 @@ class OneOffContributionsSpec extends FeatureSpec with GivenWhenThen with Before
       When("they press the Stripe payment button")
       landingPage.selectStripePayment()
 
+      And("enter card details")
```
tomrf1: Previously this test got through to the /thankyou page by skipping the stripe checkout stage and sending a test token to the backend.
Now it actually populates the card details fields

support-frontend/assets/helpers/paymentIntegrations/oneOffContributions.js
```diff
@@ -59,7 +59,7 @@ export type StripeChargeData = {|
   paymentData: {
     currency: IsoCurrency,
     amount: number,
-    token: string,
+    token: string, // TODO - remove this field once it has been removed from payment-api
```
jlieb10: 👍

support-frontend/test/selenium/contributions/OneOffContributionsSpec.scala
jlieb10: 👍

support-frontend/assets/pages/contributions-landing/components/ContributionForm.jsx
```diff
@@ -174,9 +171,7 @@ const formHandlersForRecurring = {
 const formHandlers: PaymentMatrix<PropTypes => void> = {
   ONE_OFF: {
     Stripe: (props: PropTypes) => {
-      if (props.stripeElementsTestVariant !== 'stripeCardElement') {
-        openStripePopup(props);
```
jlieb10: Do we need to remove this function from the codebase or is it in use elsewhere?

tomrf1: good question. It's actually still used by the recurring contributions handler, which still uses stripe checkout. I'm working on removing this in a separate branch

support-frontend/assets/pages/contributions-landing/contributionsLandingActions.js
```diff
@@ -464,14 +445,6 @@ const paymentAuthorisationHandlers: PaymentMatrix<(
...
       if (paymentAuthorisation.paymentMethod === Stripe) {
-        if (paymentAuthorisation.token) {
-          return dispatch(executeStripeOneOffPayment(
-            stripeChargeDataFromCheckoutAuthorisation(paymentAuthorisation, state),
-            (token: string) => dispatch(setGuestAccountCreationToken(token)),
-            (thankYouPageStage: ThankYouPageStage) => dispatch(setThankYouPageStage(thankYouPageStage)),
-          ));
-        }
-
         if (paymentAuthorisation.paymentMethodId) {
```
jlieb10: as there's no longer two conditional branches inside the first `if` we could consider a rewrite as `(paymentAuthorisation.paymentMethod === Stripe && paymentAuthorisation.paymentMethodId)`

tomrf1: Good spot, I've just added an extra level of error logging for this as the `paymentMethodId` should always be present at this point.

PR #2119, opened 2019-10-08
"Reminder" proof of concept frontend component

jlieb10:
## Why are you doing this?
This is the frontend component for the reminder proof of concept

[**Trello Card**](https://trello.com/c/os7AbR3n/1502-poc-remind-me-later-feature)

## Screenshots
set password screen

### DIFF
--- a/support-frontend/assets/components/contributionsReminder/contributionsReminder.jsx
+++ b/support-frontend/assets/components/contributionsReminder/contributionsReminder.jsx
@@ -1,0 +1,162 @@
+// @flow
+
+// ----- Imports ----- //
+
+import React, { Component } from 'react';
+import SvgSubscribe from 'components/svgs/subscribe';
+import SvgSubscribed from 'components/svgs/subscribed';
+import TrackableButton from 'components/button/trackableButton';
+import { trackComponentClick, trackComponentLoad } from 'helpers/tracking/behaviour';
+import { connect } from 'react-redux';
+import type { Csrf as CsrfState } from 'helpers/csrf/csrfReducer';
+import { routes } from 'helpers/routes';
+import { logException } from 'helpers/logger';
+
+// ----- Types ----- //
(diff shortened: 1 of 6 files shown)


support-frontend/assets/components/contributionsReminder/contributionsReminder.jsx
```diff
@@ -0,0 +1,162 @@
...
+      successful: false,
+    });
+  }
+
+  renderButtonCopy = (): string => {
+    const { clicked, pending, successful } = this.state;
+    const defaultState = !clicked && !pending && !successful;
+    const pendingState = clicked && pending && !successful;
+    const failedState = clicked && !pending && !successful;
+    const successState = clicked && !pending && successful;
```
jlieb10: I realize these are overly verbose-- eg default could be just `!clicked` -- but I think this approach makes them more readable/debuggable

tomrf1: I think the simplest way to model this would be:
```
type State = 'Initial' | 'Pending' | 'Failed' | 'Successful'
...

state = 'Initial'
...

switch (state) {
  case 'Initial':
    return 'Remind me in December';
  case 'Pending':
    ...
}

```

jlieb10: I like demonstrating that the three client events add up to a state but i can refactor if you think your way is more maintainable

tomrf1: I think each client event would still map to a distinct state, e.g.
```
onButtonClicked = (): void => {
  this.setState('Pending');
}
```

jlieb10: i'm not 100% sure about this but i think state has to be an object

PR #2122, opened 2019-10-09
Postcode length restriction on delivery and billing

alexflav23:
## Why are you doing this?

Support workers experience issues submitting subscriptions to Salesforce where the postcode length is over 20 characters. This is also likely to be accidental user input, and we need to protect against it with a character limit in the postcode field.

[**Trello Card**](https://trello.com/c/ONmjZVen/2565-clientside-validate-checkout-fields-to-avoid-salesforce-errors)

## Changes

* Added postcode maximum length rule for billing and delivery postcode.

## Screenshots

UNADJUSTEDNONRAW_thumb_2

### DIFF
--- a/support-frontend/assets/components/subscriptionCheckouts/address/addressFieldsStore.js
+++ b/support-frontend/assets/components/subscriptionCheckouts/address/addressFieldsStore.js
@@ -76,7 +76,7 @@ const getFormFields = (state: State): FormFields => ({
 const isPostcodeOptional = (country: Option<IsoCountry>): boolean =>
   country !== 'GB' && country !== 'AU' && country !== 'US' && country !== 'CA';
 
-const checkpostCodeLength = (input: string | null): boolean => (input != null && input.length <= 20);
+const checkpostCodeLength = (input: string | null): boolean => ((input == null) || (input.length <= 20));
 
 const isStateNullable = (country: Option<IsoCountry>): boolean =>
   country !== 'AU' && country !== 'US' && country !== 'CA';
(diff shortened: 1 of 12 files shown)


support-frontend/assets/components/subscriptionCheckouts/address/addressFieldsStore.js
```diff
@@ -110,6 +112,10 @@ const applyBillingAddressRules = (fields: FormFields, addressType: AddressType):
       fields.country === 'CA' ? `Please select a ${addressType} province/territory.` : `Please select a ${addressType} state.`,
     ),
   },
+  {
+    rule: checkpostCodeLength(fields.postCode),
+    error: formError('postCode', `Postcode should not be longer than 20 characters.`),
```
johnduffell: looks like on the other address validation errors we also include `${addressType}` in the message, so the summary points out which it is, so best to add that.

alexflav23: Done now.

support-frontend/assets/components/subscriptionCheckouts/address/addressFieldsStore.js
```diff
@@ -110,6 +112,10 @@ const applyBillingAddressRules = (fields: FormFields, addressType: AddressType):
       fields.country === 'CA' ? `Please select a ${addressType} province/territory.` : `Please select a ${addressType} state.`,
     ),
   },
+  {
+    rule: checkpostCodeLength(fields.postCode),
+    error: formError('postCode', `Postcode for ${addressType} should not be longer than 20 characters.`),
```
johnduffell:
```suggestion
    error: formError('postCode', `Please enter a ${addressType} postcode no longer than 20 characters.`),
```
I would also consider moving the two postcode rules next to each other.  As on line 100-103 is another rule, for non empty postcodes.

PR #2124, opened 2019-10-10
Make the Payment Request button work for recurring contributions

paulbrown1982:
Enable the Payment Request button work for recurring contributions checkout page, and also confirm the payment frequency in its title.

The payment method in Zuora looks identical to if the card was tokenised via Stripe Checkout so there should be no downstream issue for the customer, other than when the button does eventually be used for Apple Pay, that won't be available in the manage-my-account portal.

I have made 3 trade-off decisions as part of this change:

1) I have added the requirement that we request the contributor's name via the Payment Request button. This means if they don't have a name set up in their system then the button won't show. I'm 99% sure this should have a negligible impact on the Apple Pay button showing up when single contribution is the default.
2) As browsers just store the name as 1 field, I have assumed the contributor's first name is the first word in that field (when split by `' '`) and the last name is all the other words.
3) This change does **not** release the button on the recurring tabs as an A/B test.

## Why are you doing this?
Apple Pay leverages the Payment Request button, so this is a stepping stone to enable Apple Pay for recurring too.

[**Trello Card**](https://trello.com/c/0f3yqsoz)

## To do
- [ ] Decide how State will work in the US/CA

## Screenshots

Screen Shot 2019-10-10 at 15 21 39

Screen Shot 2019-10-11 at 17 17 47

Screen Shot 2019-10-11 at 17 15 21

Screen Shot 2019-10-11 at 17 17 58

### DIFF
--- a/support-config/src/main/scala/com/gu/support/config/StripeConfig.scala
+++ b/support-config/src/main/scala/com/gu/support/config/StripeConfig.scala
@@ -16,11 +16,11 @@ case class StripeConfig(defaultAccount: StripeAccountConfig,
   def forCurrency(maybeCurrency: Option[Currency]): StripeAccountConfig = {
     maybeCurrency match {
       case Some(AUD) => {
-        SafeLogger.info(s"StripeConfig: getting AU stripe account for AUD")
+        SafeLogger.debug(s"StripeConfig: getting AU stripe account for AUD")
         australiaAccount
       }
       case _ => {
-        SafeLogger.info(s"StripeConfig: getting default stripe account for ${maybeCurrency.map(_.iso).mkString}")
+        SafeLogger.debug(s"StripeConfig: getting default stripe account for ${maybeCurrency.map(_.iso).mkString}")
         defaultAccount
       }
     }
@@ -29,15 +29,15 @@ case class StripeConfig(defaultAccount: StripeAccountConfig,
   def forCountry(maybeCountry: Option[Country]): StripeAccountConfig = {
     maybeCountry match {
       case Some(Country.Australia) => {
-        SafeLogger.info(s"StripeConfig: getting AU stripe account for Australia")
+        SafeLogger.debug(s"StripeConfig: getting AU stripe account for Australia")
         australiaAccount
       }
       case Some(Country.US) => {
-        SafeLogger.info(s"StripeConfig: getting US stripe account for United States")
+        SafeLogger.debug(s"StripeConfig: getting US stripe account for United States")
         unitedStatesAccount
       }
       case _ => {
-        SafeLogger.info(s"StripeConfig: getting default stripe account for ${maybeCountry.map(_.name).mkString}")
+        SafeLogger.debug(s"StripeConfig: getting default stripe account for ${maybeCountry.map(_.name).mkString}")
         defaultAccount
       }
     }
--- a/support-frontend/app/views/contributions.scala.html
+++ b/support-frontend/app/views/contributions.scala.html
@@ -103,5 +103,6 @@ window.guardian.paymentApiPayPalEndpoint = "@paymentApiPayPalEndpoint";
 
   window.guardian.stripeElements = @settings.switches.experiments.get("stripeElements").exists(_.isOn)
   window.guardian.forceCampaign = @settings.switches.experiments.get("forceCampaign").exists(_.isOn)
+  window.guardian.recurringStripePaymentRequestButton = @settings.switches.experiments.get("recurringStripePaymentRequestButton").exists(_.isOn)
   </script>
 }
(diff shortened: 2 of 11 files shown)


support-frontend/package.json
```diff
@@ -71,7 +71,7 @@
     "react-redux": "^5.1.1",
     "react-router": "^4.3.1",
     "react-router-dom": "^4.3.1",
-    "react-stripe-elements": "^4.0.1",
+    "react-stripe-elements": "^5.0.1",
```
paulbrown1982: I updated this for IE compatibility.

support-config/src/main/scala/com/gu/support/config/StripeConfig.scala
```diff
@@ -16,11 +16,11 @@ case class StripeConfig(defaultAccount: StripeAccountConfig,
   def forCurrency(maybeCurrency: Option[Currency]): StripeAccountConfig = {
     maybeCurrency match {
       case Some(AUD) => {
-        SafeLogger.info(s"StripeConfig: getting AU stripe account for AUD")
+        SafeLogger.debug(s"StripeConfig: getting AU stripe account for AUD")
```
paulbrown1982: Increased logging level of this stuff as it horribly spams the CloudWatch logs.

support-frontend/assets/pages/contributions-landing/components/StripePaymentRequestButton/StripePaymentRequestButton.jsx
```diff
@@ -177,7 +177,7 @@ function initialisePaymentRequest(props: PropTypes) {
     currency: props.currency.toLowerCase(),
     total: {
       label: 'The Guardian',
-      amount: props.amount,
+      amount: props.amount * 100,
```
paulbrown1982: This bug didn't manifest because the 'update' function was ran whenever any amount or payment request button was clicked anyway.

support-frontend/assets/pages/contributions-landing/components/StripePaymentRequestButton/StripePaymentRequestButton.jsx
```diff
@@ -80,6 +82,7 @@ const mapDispatchToProps = (dispatch: Function) => ({
     (paymentRequest: Object) => { dispatch(setStripePaymentRequestObject(paymentRequest)); },
   updateEmail: (email: string) => { dispatch(updateEmail(email)); },
   setStripePaymentRequestButtonClicked: () => { dispatch(setStripePaymentRequestButtonClicked()); },
+  setAssociatedPaymentMethod: () => { dispatch(updatePaymentMethod(Stripe)); },
```
paulbrown1982: I think I prefer this being a separate method than an additional `dispatch(updatePaymentMethod(Stripe))` line being put inside `setStripePaymentRequestButtonClicked`, what do you think @tomrf1 ?

tomrf1:
@paulbrown1982 this seems to be working locally now.
I've made 2 changes:
1. add a `key` attribute to the `StripeProvider` to tell react to create separate instances for ONE_OFF and REGULAR,
2. Maintain separate `StripePaymentRequestButtonData` instances in redux for ONE_OFF and REGULAR

support-frontend/assets/pages/contributions-landing/components/StripePaymentRequestButton/StripePaymentRequestButton.jsx
```diff
@@ -127,17 +152,18 @@ function updateAmount(amount: number, paymentRequest: Object | null) {
 function onClick(event, props: PropTypes) {
   event.preventDefault();
   trackComponentClick('apple-pay-clicked');
-  updateAmount(props.amount, props.stripePaymentRequestObject);
-  props.setStripePaymentRequestButtonClicked();
+  updateTotal(props);
+  props.setAssociatedPaymentMethod();
```
tomrf1: why was this call not necessary before?

paulbrown1982: Because there is a special thank you page for direct debit.

support-frontend/app/views/contributions.scala.html
```diff
@@ -103,5 +103,6 @@
 
   window.guardian.stripeElements = @settings.switches.experiments.get("stripeElements").exists(_.isOn)
   window.guardian.forceCampaign = @settings.switches.experiments.get("forceCampaign").exists(_.isOn)
+  window.guardian.recurringStripePaymentRequestButton = @settings.switches.experiments.get("recurringStripePaymentRequestButton").exists(_.isOn)
```
tomrf1: this can be enabled from the admin console later on

tomrf1:
Tested single + recurring.
I've put recurring behind a feature switch because we still need to review the UX and handling of US state. For now, let's just get it merged

tomrf1:
also tested in CODE. @paulbrown1982 are you happy for me to merge this?

paulbrown1982:
@tomrf1 Yup if you want to now's your last chance today I reckon, or you can wait till Monday?

tomrf1:
@paulbrown1982 thanks, definitely one for Monday morning!

PR #2125, opened 2019-10-10
Jy current dp product page

jesseyuen:
## Why are you doing this?

Merging my page into your branch!


jesseyuen:
> I'm not clear in some cases why you've used `<span>` for some titles, but otherwise this looks good!

Oh yeah! They were <h1> tags for some reason?? I was part way through making them <strong> tags. I'll pick this up in the next pass @lucymonie

PR #2126, opened 2019-10-10
2664 infinite resubmission bug

alexflav23:
## Why are you doing this?

[**Trello Card**](https://trello.com/c/sBBvEFdc/2664-fix-backend-infinite-loop-re-resubmission-of-checkout-requests-via-support-workers-when-we-encounter-salesforce-validation-error)

## Changes

* Propagating error cause as part of `WebServiceHelperError` stack.
* Preventing retries for `DecodingFailures`. If lambda steps cannot parse expected response, we abort retries. We still retry in the same fashion if service calls fail for different reasons, however this should cover all salesforce errors.


support-workers/src/main/scala/com/gu/emailservices/EmailService.scala
```diff
@@ -18,6 +18,7 @@ class EmailService(implicit val executionContext: ExecutionContext) {
     .withRegion(Regions.EU_WEST_1)
     .build()
 
+  Console.println(Configuration.contributionThanksQueueName)
```
rupertbates: Probably want to remove this

alexflav23:
> Looks good, remove a `Console.println` and then 🚀

Done, forgot about my advanced debugging technique.

PR #2128, opened 2019-10-10
Promotion copy on Guardian Weekly landing page

rupertbates:
## Why are you doing this?

Marketing still use the subscribe page to create promotion specific pages because this functionality is not currently available on support.

This PR updates the Guardian Weekly landing page so that if a promotional code is present on the query string `?promoCode=XXXXXX` then it will use the copy provided in the promotions tool in place of the standard copy for the page.

### Copy is set in the promotions tool:
Screenshot 2019-10-10 at 15 47 13

### It then appears in the page:
Screenshot 2019-10-10 at 15 48 08

## Changes
There are a number of changes to facilitate this:
* Added copy fields on to the `Promotion` scala type to pull the relevant copy from Dynamo
* Added a class `ProductPromotionCopy` which validates that a given promo code is valid for the current product and country and is not expired
* Pass the promotion copy through to the Javascript and get it into the redux store
* Use the copy in the page

There are a lot of files changed in this PR but most of this is because I split js code related to promotions out of the `productPrices.js` file into its own file which has meant lots of imports changing.

[**Trello Card**](https://trello.com/c/QjXywSpn/2648-enable-promo-code-functionality-for-gw-on-support-platform-mvp)


support-frontend/assets/helpers/productPrice/productPrices.js
```diff
@@ -21,35 +21,11 @@ import {
...
 } from 'helpers/internationalisation/currency';
 import { fixDecimals } from 'helpers/subscriptions';
-import type { Option } from 'helpers/types/option';
-import { getQueryParameter } from 'helpers/url';
+import type { Promotion } from 'helpers/productPrice/promotions';
+import { applyDiscount, getPromotion } from 'helpers/productPrice/promotions';
 
 // ----- Types ----- //
 
-export type DiscountBenefit = {
```
rupertbates: All of the deleted sections in this file have been moved to `promotions.js`

support-services/src/main/scala/com/gu/support/promotions/PromotionValidator.scala
```diff
@@ -44,7 +45,22 @@ object PromotionValidator {
         Some(ExpiredPromotion)
       else
         None
+
+    def validForAnyProductRatePlan(
```
rupertbates: Checks that a product has at least one rate plan which is valid for this promotion

support-frontend/assets/helpers/productPrice/promotions.js
```diff
@@ -0,0 +1,105 @@
...
+      ...price,
+      // $FlowIgnore - we have checked this above
+      price: promotion.introductoryPrice.price,
+    };
+  }
+  return price;
+}
+
+const matchesQueryParam = promotion =>
+  getQueryParameter('promoCode') === promotion.promoCode;
```
paulbrown1982: Should this be a case-insensitive comparison?

rupertbates: We haven't generally used a case-insensitive comparison although it might be a good idea. I think we wouldn't expect people to be typing it in

PR #2130, opened 2019-10-11
Amounts test round 4

tomrf1:
## Why are you doing this?
The next amounts test, to be run during the moment.

The control values are set from the admin console


tomrf1:
@paulbrown1982 yes ideally the admin console would allow use to configure tests as well as the control amounts, but it's easy to hack it like this for now

PR #2135, opened 2019-10-14
Add Optimize test into native AB test participations

rupertbates:
## Why are you doing this?

To test using the native AB test framework to allocate users to variants in Optimize tests.

**What I've done:**
- Created a test in Optimize
- Added it into the native ab test list in `abTestDefinitions.js` so the user will get automatically allocated to a variant by the framework
- Added some code to the digi pack landing page to fire a GA event with the `experimentId` and `experimentVariant` datalayer variables set so that the user is assigned to the correct cohort in GA/Optimize

There is quite a bit of scope to improve on this implementation but I want to test that it works before refining the approach any further

[**Trello Card**](https://trello.com/c/PsimymUQ/2674-optmize-rework)

### DIFF
--- a/support-frontend/assets/helpers/campaigns.jsx
+++ b/support-frontend/assets/helpers/campaigns.jsx
@@ -65,7 +65,7 @@ export const campaigns: Campaigns = {
           our <a href="https://www.theguardian.com/help/privacy-policy">Privacy Policy</a>.
         </p>
         {isUK ? (
-          <span className={`component-terms-privacy__divider`}>
+          <span className="component-terms-privacy__divider">
             <p>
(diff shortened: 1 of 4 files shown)


support-frontend/assets/helpers/campaigns.jsx
```diff
@@ -65,15 +65,15 @@ export const campaigns: Campaigns = {
           our <a href="https://www.theguardian.com/help/privacy-policy">Privacy Policy</a>.
         </p>
         {isUK ? (
-          <span className={`component-terms-privacy__divider`}>
+          <span className="component-terms-privacy__divider">
```
rupertbates: Changes to this file were from running `yarn lint --fix`

PR #2136, opened 2019-10-14
Ide settings

alexflav23:
## Why are you doing this?

[**Trello Card**](https://trello.com/c/SUwE8N6a/2665-intelij-add-sbt-plugin-to-mark-assets-as-sources-root-automatically)

## Changes

* Added another SBT sources directory not directly managed by SBT under the sub-project `build.sbt` definition.
* Add a name and module name for the root project so it's easier to know what's what in InteliJ.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -58,6 +58,10 @@ lazy val commonDependencies = Seq(
 )
 
 lazy val root = (project in file("."))
+  .settings(
+    name := "support-frontend",
+    moduleName := "support-frontend"
+  )
   .aggregate(
     `support-frontend`,
     `support-workers`,
--- a/build.sbt
+++ b/build.sbt
@@ -59,8 +59,8 @@ lazy val commonDependencies = Seq(
 
 lazy val root = (project in file("."))
   .settings(
-    name := "support-frontend",
-    moduleName := "support-frontend"
+    name := "support-frontend-root",
+    moduleName := "support-frontend-root"
   )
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -58,6 +58,10 @@ lazy val commonDependencies = Seq(
 )
 
 lazy val root = (project in file("."))
+  .settings(
```
rupertbates: Will this cause problems because there is already a support-frontend project beneath the root? Doesn't this make the sbt command `project support-frontend` ambiguous?

alexflav23: You're right, fixed.

rupertbates: Can we just leave it as root? Less typing 😃

support-frontend/build.sbt
```diff
@@ -95,4 +95,6 @@ javaOptions in Universal ++= Seq(
 
 javaOptions in Test += "-Dconfig.file=test/selenium/conf/selenium-test.conf"
 
+unmanagedSourceDirectories in Compile += (baseDirectory.value / "assets")
```
rupertbates: This is 🥇

build.sbt
alexflav23: @rupertbates Would be nice it if's not confusing in InteliJ, every other project is properly named while this one is called root, so it's not consistent and can clash, since if you don't put a module name SBT will use variable name.

PR #2140, opened 2019-10-15
AB test improvements

rupertbates:
## Why are you doing this?
This PR adds two new capabilities to the AB test framework
### Target page property
It is now possible to set an optional `targetPage` field on AB tests which specifies which page they should load on. This is to stop users who never visit a particular page from being included in the cohort for tests running on that page. This field should contain a regex containing a path which will be matched against the current page path when evaluating whether to run the test, eg. `'/(uk|us|eu|au|ca|nz|int)/subscribe/digital$'`. 

### Optimize Id 
It is now possible to add an `optimizeId` field to AB tests which allows the linking of a test set up through the native test framework to an experiment running in Optimize. When a test with an optimizeId is initialised we send a GA event to set the appropriate experiment and variant data for the user.

This PR also adds a test for the upcoming Digital Pack product page test

[**Trello Card**](https://trello.com/c/PsimymUQ/2674-optimize-rework)

### DIFF
--- a/support-frontend/assets/helpers/abTests/abtest.js
+++ b/support-frontend/assets/helpers/abTests/abtest.js
@@ -13,6 +13,7 @@ import { type CountryGroupId } from 'helpers/internationalisation/countryGroup';
 import { type AmountsRegions } from 'helpers/contributions';
 
 import { tests } from './abtestDefinitions';
+import { gaEvent } from 'helpers/tracking/googleTagManager';
 
 // ----- Types ----- //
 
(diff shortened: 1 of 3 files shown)


support-frontend/assets/helpers/abTests/abtest.js
```diff
@@ -171,12 +175,33 @@ function randomNumber(mvtId: number, independent: boolean, seed: number): number
...
 
-  const variantIndex = randomNumber(mvtId, independent, seed) % test.variants.length;
+  return randomNumber(mvtId, independent, seed) % test.variants.length;
+}
+
+function targetPageMatches(targetPage: ?string) {
+  if (!targetPage) { return true; }
 
-  return test.variants[variantIndex].id;
+  return new URL(window.location).pathname.match(targetPage) != null;
```
tomrf1: out of interest, why not just `window.location.pathname`?

rupertbates: Not sure actually! I've just changed it

PR #2145, opened 2019-10-16
update dependencies to remediate snyk vulnerabilities

tjsilver:
## Why are you doing this?
Snyk highlighted a number of vulnerabilities in the npm packages. This fix addresses many of them by upgrading the packages involved. The rest of the issues are not currently fixable as the vulnerable packages are sub-dependencies. 

[**Trello Card**](https://trello.com/c/lsnyaJ7b)

## Changes
Updated package.json


alexflav23:
@tjsilver I think we don't need to commit the npmtree here, otherwise LGTM.

alexflav23:
@tjsilver Having an issue with building the frontend as a result of this PR, might be worth having a look at the TeamCity log, I'm not sure what is going wrong.

tjsilver:
> @tjsilver Having an issue with building the frontend as a result of this PR, might be worth having a look at the TeamCity log, I'm not sure what is going wrong.

Yes thanks - just looking at that now - this PR is still draft and will make it ready to review when issues are sorted.

alexflav23:
@tjsilver Here's the excerpt below. I think a downstream library upgrade is causing this, though I am not quite sure which one and how. It seems to be breaking test compilation for `paymentSelectionTests.js`

```
Summary of all failing tests
[10:13:02]FAIL assets/pages/digital-subscription-landing/components/paymentSelection/__tests__/paymentSelectionTests.js
[10:13:02]  ● Test suite failed to run
[10:13:02]
[10:13:02]    TypeError: Invalid value used as weak map key
[10:13:02]        at WeakMap.set (native)
[10:13:02]
[10:13:02]      28 | };
[10:13:02]      29 | 
[10:13:02]    > 30 | export type AcquisitionQueryParameters = Array<QueryParameter>
[10:13:02]         |                                                               ^
[10:13:02]      31 | 
[10:13:02]      32 | export type OphanIds = {|
[10:13:02]      33 |   pageviewId: string,
[10:13:02]
[10:13:02]      at _interopRequireWildcard (assets/helpers/tracking/acquisitions.js:30:626)
[10:13:02]      at Object.<anonymous> (assets/helpers/tracking/acquisitions.js:14:13)
[10:13:02]      at Object.<anonymous> (assets/helpers/externalLinks.js:14:21)
[10:13:02]      at Object.<anonymous> (assets/pages/digital-subscription-landing/components/paymentSelection/helpers/paymentSelection.jsx:10:22)
[10:13:02]      at Object.<anonymous> (assets/pages/digital-subscription-landing/components/paymentSelection/__tests__/paymentSelectionTests.js:6:25)
[10:13:02]
```

tomrf1:
I've just pushed a commit which fixes the jest problem. Some things to note:
1. The line numbers given in the stack trace are completely misleading. I had to experiment with commenting out code until I found the actual cause.
2. The actual problem was the syntax of the ophan import at the top of acquisitions.js - I don't know why, but it's a simple fix.
3. I updated the `babel-jest` dependency but I don't think it helped fix this particular issue

jlieb10:
@tomrf1 could you share some insights or history as to how you figured this out! Great work 👍

tomrf1:
@jlieb10 
I'm afraid it's not very scientific:
1. observed the line numbers in the stack trace did not make sense (we should be able to trace function calls/imports)
2. started commenting things out to confirm that the code around the specified line number was unrelated to the issue
3. commented out all imports in `acquisitions.js` on a hunch. It predictably caused a bunch of new errors, but the `WeakMap` error had gone.
4. narrowed it down by uncommenting all but the ophan import, and I'd found the culprit

jlieb10:
@tomrf1 were you able to reproduce the TC build fail locally?

alexflav23:
@tomrf1  That's the weirdest thing I've seen, a confused stacktrace where line numbers make no sense. Why does that happen, is that because of the Flow pre-compilation before JS VM interpretation?

PR #2148, opened 2019-10-17
Remove old Optimize code

rupertbates:
## Why are you doing this?
We have changed the way we work with Optimize in the code base, this has meant we can delete a large amount of code that supported the old way of working.

### For reference the code needed to support the old way of working was:
- An Optimize script in the page which loaded a Javascript api. This also required code to take account of the users consent settings. 
- Custom code which queried the api to check if any experiments were running on the page and if so which variant the user is in. This was complex and error prone because of the asynchronous nature of the Optimize script and the fact that it could fail due to ad-blockers or network conditions. The most undesirable consequence of this was that we had to delay rendering the page until Optimize loaded or until we hit a 4 second time out, which was a poor user experience.
- Custom code to include Optimize experiments in tracking code which is sent to the data lake so that the Data & Insight team could do detailed analysis. Because we only got an experiment ID and variant number from Optimize not an experiment/variant name this was difficult to work with in DL.

### The new way of working:
- An extra optional field `optimizeId` on AB tests defined through support-frontend's native AB test framework.
- Some code which when a user is allocated to test with an `optimizeId` sends an event to GA to set the experiment name and ID for that user.

### Advantages of the approach:
- No delay before render. As we are allocating users into test variants ourselves using the native test framework and then telling Optimize about that allocation, we no longer need any Optimize script on the site, we don't need a lot of complicated code to make sure that script has loaded successfully and we don't need to delay rendering until it has succeeded or timed out.
- All data lake tracking now works in the same way as any other test so there is no custom code needed for Optimize and all tests have descriptive test and variant names.

[**Trello Card**](https://trello.com/c/PsimymUQ/2674-optimize-rework)


jfsoul:
This is in my top 5 PRs of all time, thank you.

PR #2149, opened 2019-10-17
Give SBT more memory

rupertbates:
## Why are you doing this?

SBT keeps running out of memory, this should fix it.

### DIFF
--- a/support-frontend/devrun.sh
+++ b/support-frontend/devrun.sh
@@ -2,4 +2,4 @@
 
 yarn devrun &
 yarn storybook &
-cd ..; sbt "project support-frontend" devrun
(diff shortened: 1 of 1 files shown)


alexflav23:
@rupertbates I missed this PR yesterday, there's a different way to do this. Remember test and compile don't necessarily execute in the same VM as SBT, I'm not sure what this setting does specifically, but we can always target a specific phase accurately with scoped options, e.g:

```scala
fork := true, // otherwise Java options are ignored
javaOptions in Test ++= Seq(
 "-XX:MaxMetaspaceSize=512m",
 "-Xmx=2g",
...
)
```

That way we can target what gets more memory, if we ever needed to.

PR #2159, opened 2019-10-22
Stripe Elements for recurring contributions (SCA)

tomrf1:
## Why are you doing this?
Single contributions already uses Stripe Elements and the Payment Intents API. This means it is SCA-compliant.

This PR does the same for recurring contributions. There are 2 aspects to this change:
1. Replace Stripe Checkout with Stripe Elements, with the card details form displayed in the landing page rather than as a popup.
2. Use the Setup Intents API for SCA-compliance.

It is implemented as an A/B test to ensure acquistions do not go down.

Roughly, the new recurring contributions flow is as follows:
1. client requests a new Setup Intent from a lambda (behind api gateway), which returns a `client_secret` - the identifier for the new Setup Intent. This request is made as soon as the stripe component is created to avoid slowing down the payment.
2. When user clicks 'Contribute', client requests a new Payment Method from Stripe (using the stripe.js `handleCardSetup` call) for the above `client_secret`.
3. client sends the resulting `paymentMethodId` to support-workers, our backend.
4. support-workers sends a request to Stripe to create a new Customer for the given `paymentMethodId`.
5. support-workers sends the Customer details to zuora to create the subscription.

This PR covers steps 1-3.

Here's a diagram. TODO - draw some formal diagrams for Stripe single and recurring.
stripe_recurring

Note - as @twrichards discovered when [implementing this for manage frontend](https://github.com/guardian/manage-frontend/pull/241), you have to use test card [`4000 0025 0000 3155`](https://stripe.com/docs/payments/cards/testing#regulatory-cards) for this to work.


support-frontend/app/controllers/Application.scala
```diff
@@ -156,6 +156,10 @@ class Application(
       classes = Some(classes)
     )
 
+    val stripeSetupIntentEndpoint =
+      if (stage == Stages.PROD) "https://stripe-intent.support.guardianapis.com/stripe-intent"
```
tomrf1: no point having this in config

support-frontend/app/views/contributions.scala.html
```diff
@@ -93,15 +94,16 @@
...
   window.guardian.existingPaymentOptionsEndpoint = "@existingPaymentOptionsEndpoint";
+  window.guardian.stripeSetupIntentEndpoint = "@stripeSetupIntentEndpoint";
   window.guardian.csrf = { token: "@CSRF.getToken.value" };
 
   @guestAccountCreationToken.map { guestAccountCreationToken =>
     window.guardian.guestAccountCreationToken = "@guestAccountCreationToken";
   }
 
-  window.guardian.stripeElements = @settings.switches.experiments.get("stripeElements").exists(_.isOn)
+  window.guardian.stripeElementsRecurring = @settings.switches.experiments.get("stripeElementsRecurring").exists(_.isOn)
```
tomrf1: this means we can switch it on/off from the admin console

support-frontend/assets/helpers/abTests/abtestDefinitions.js
```diff
@@ -7,6 +7,9 @@ import { V1 } from 'helpers/abTests/data/testAmountsData';
 // ----- Tests ----- //
 export type LandingPageCopyReturningSinglesTestVariants = 'control' | 'returningSingle' | 'notintest';
 export type LandingPageMomentBackgroundColourTestVariants = 'control' | 'yellow' | 'notintest';
+export type LandingPageStripeElementsRecurringTestVariants = 'control' | 'stripeElements' | 'notintest';
+
+const contributionsLandingPageMatch = '/(uk|us|eu|au|ca|nz|int)/contribute(/.*)?$';
```
tomrf1: The `targetPage` field in the `Test` object is a new feature for restricting tests to specific pages on support.thegaurdian.com. I will apply this matcher to our other landing page tests, but not in this PR

support-frontend/assets/helpers/stripe.js
```diff
@@ -23,6 +23,8 @@ export const stripeCardFormIsIncomplete = (
   contributionType: ContributionType,
   paymentMethod: PaymentMethod,
   stripeCardFormComplete: boolean,
-): boolean => contributionType === 'ONE_OFF' &&
-    paymentMethod === Stripe &&
-    !(stripeCardFormComplete);
+  stripeElementsRecurringTestVariant: string,
+): boolean =>
+  (contributionType === 'ONE_OFF' || stripeElementsRecurringTestVariant === 'stripeElements') &&
```
alexflav23: Would be good to have an enum for contribution types to avoid using literal values and comparisons: https://flow.org/en/docs/types/utilities/

support-frontend/assets/pages/contributions-landing/components/StripeCardForm/StripeCardForm.jsx
```diff
@@ -26,35 +27,45 @@ import { trackComponentLoad } from 'helpers/tracking/behaviour';
...
   checkoutFormHasBeenSubmitted: state.page.form.formData.checkoutFormHasBeenSubmitted,
+  setupIntentClientSecret: state.page.form.stripeCardFormData.setupIntentClientSecret,
+  paymentWaiting: state.page.form.isWaiting,
 });
 
 const mapDispatchToProps = (dispatch: Function) => ({
-  onPaymentAuthorised:
-    (paymentAuthorisation: PaymentAuthorisation) =>
-      dispatch(onThirdPartyPaymentAuthorised(paymentAuthorisation)),
+  onPaymentAuthorised: (paymentMethodId: string) =>
```
alexflav23: Could be beneficial to enforce something more concise than `string`, e.g using an enum type or similar.

support-frontend/assets/pages/contributions-landing/components/StripeCardForm/StripeCardForm.jsx
```diff
@@ -97,37 +108,11 @@ class CardForm extends Component<PropTypes, StateTypes> {
...
-          });
-        }
-      });
-    });
-
-    this.props.setHandleStripe3DS((clientSecret: string) => {
-      trackComponentLoad('stripe-3ds');
-      return this.props.stripe.handleCardAction(clientSecret);
-    });
+    if (this.props.contributionType === 'ONE_OFF') {
```
alexflav23: Same as above, may be worth while to use an enum instead of plain strings.

support-frontend/assets/helpers/stripe.js
tomrf1: Ideally I'd be able to use something like `contributionType === ContributionType.ONE_OFF` here, but it doesn't work despite having a [ContributionType](https://github.com/guardian/support-frontend/blob/master/support-frontend/assets/helpers/contributions.js#L28) type. I could also add an enum with the same values, though it's a bit rubbish

support-frontend/assets/pages/contributions-landing/components/StripeCardForm/StripeCardForm.jsx
tomrf1: the payment method here is a Stripe api concept - https://stripe.com/docs/payments/payment-methods
The `paymentMethodId` here is effectively a string token given to us by Stripe

alexflav23: @tomrf1 Makes sense.

support-frontend/assets/helpers/stripe.js
tomrf1: actually I'm not sure there's anything more we can do here with flow, unless I'm missing something

support-frontend/app/controllers/Application.scala
```diff
@@ -156,6 +156,10 @@ class Application(
...
+      else "https://stripe-intent-code.support.guardianapis.com/stripe-intent"
```
paulbrown1982: Not related to this PR, but ideally CODE endpoints would be on a separate subdomain or TLD so that there's less chance of a PROD cookie going to a CODE server. Might be worth an audit of our CODE TLDs. I've created a health [task](https://trello.com/c/292FoN2N) on our backlog.

tomrf1: We discussed this recently - [see this comment](https://github.com/guardian/support-frontend/pull/2111/files#r331962950)
Ideally it would use `code.support.dev-guardianapis.com`, though this could be updated later?

tomrf1: Discussed IRL - we can solve this after it goes live

PR #2167, opened 2019-10-24
Add stripe payment intent token to Subscription checkout pages

alexflav23:
## Why are you doing this?

[**Trello Card**](https://trello.com/c/e5IoEaKC/2489-spike-stripe-understand-setupintent-steps-needed-on-support-frontend)

## Changes

* Add a stage parameter to checkout templates to determine stripe intent Lambda URL for specific stage/environment.
* Passthrough stripe intent URL to client side HTML rendering via window object to allow later usage.


rupertbates:
Can you put a bit more detail into the PR about why we're doing this? Fine to copy it from Trello, it's just good to have it in the PR if we want to look back further down the line to see why a change was made.

support-frontend/app/controllers/DigitalSubscription.scala
```diff
@@ -5,46 +5,51 @@ import admin.settings.{AllSettings, AllSettingsProvider, SettingsSurrogateKeySyn
...
+)(
+  implicit val ec: ExecutionContext
+) extends AbstractController(components) with GeoRedirect with CanonicalLinks with Circe with SettingsSurrogateKeySyntax {
 
   import actionRefiners._
 
   implicit val a: AssetsResolver = assets
 
+  val stripeSetupIntentEndpoint: String =
+    if (stage == Stages.PROD) "https://stripe-intent.support.guardianapis.com/stripe-intent"
```
rupertbates: This should go into the appropriate public config file eg. `PROD.public.conf` or `UAT.public.conf`

support-frontend/app/controllers/PaperSubscription.scala
```diff
@@ -39,13 +39,17 @@ class PaperSubscription(
...
-  fontLoaderBundle: Either[RefPath, StyleContent]
+  fontLoaderBundle: Either[RefPath, StyleContent],
+  stage: Stage
 )(implicit val ec: ExecutionContext) extends AbstractController(components) with GeoRedirect with Circe with CanonicalLinks with SettingsSurrogateKeySyntax {
 
   import actionRefiners._
 
   implicit val a: AssetsResolver = assets
 
+  val stripeSetupIntentEndpoint: String =
```
rupertbates: Same comment as above

support-frontend/app/controllers/WeeklySubscription.scala
```diff
@@ -35,13 +35,19 @@ class WeeklySubscription(
...
-  fontLoaderBundle: Either[RefPath, StyleContent]
+  fontLoaderBundle: Either[RefPath, StyleContent],
+  stage: Stage
 )(implicit val ec: ExecutionContext) extends AbstractController(components) with GeoRedirect with Circe with CanonicalLinks with SettingsSurrogateKeySyntax {
 
   import actionRefiners._
 
   implicit val a: AssetsResolver = assets
 
+  val stripeSetupIntentEndpoint: String =
```
rupertbates: Same comment as above

support-frontend/app/wiring/Controllers.scala
```diff
@@ -64,7 +64,8 @@ trait Controllers {
     stringsConfig,
     allSettingsProvider,
     appConfig.supportUrl,
-    fontLoader
+    fontLoader,
+    appConfig.stage
```
rupertbates: As mentioned above, pass the endpoint url in here rather than the stage

support-frontend/app/wiring/Controllers.scala
```diff
@@ -79,7 +80,8 @@ trait Controllers {
...
```
rupertbates: Same as above

support-frontend/app/wiring/Controllers.scala
```diff
@@ -95,7 +97,8 @@ trait Controllers {
...
```
rupertbates: Same as above

rupertbates:
I do think it's good to fix stuff like formatting etc as you go along, but maybe a little at a time would be better? This PR is massive for the actual number of new code and it makes it hard to see what's important.

alexflav23:
@rupertbates About formatting, we have a build failure because of scalastyle, but it wasn't telling me which file was problematic, that's why I went round fixing random things, until the build went green.

I take your point about the config URL, I copied the approach from contributions without thinking it through, I'll adjust now.

tomrf1:
👍

support-frontend/app/wiring/Controllers.scala
```diff
@@ -33,6 +33,7 @@ trait Controllers {
     allSettingsProvider,
     appConfig.guardianDomain,
     appConfig.stage,
+    appConfig,
```
rupertbates: Can you just pass the url in here please as you have for the other 3 controllers below

alexflav23: that would make sense yes :)

PR #2166, opened 2019-10-24
Stripe Elements for recurring contributions (SCA)

tomrf1:
Original PR here https://github.com/guardian/support-frontend/pull/2159

PR #2168, opened 2019-10-24
Revert "Stripe Elements for recurring contributions (SCA)"

jlieb10:
Reverts guardian/support-frontend#2166


alexflav23:
@jlieb10 @tomrf1 and I paired today to test this on CODE and PROD and it worked well. We also deployed the Lambda function to prod on AWS and it seemed to work quite well during manual testing. Not only that but it's not even enabled.

PR #2177, opened 2019-10-30
Fix svgo npm dependency

tomrf1:
we're getting build failures because of an issue with svgo, which is picked up via one of our dependencies.
The error is:
`TypeError: Cannot set property 'multipassCount' of undefined`

This PR forces it to use `1.3.0` instead of the bad `1.3.1` version, which isn't a great solution...
see https://yarnpkg.com/lang/en/docs/selective-version-resolutions/


tomrf1:
this has now been fixed in svgo 1.3.2

PR #2179, opened 2019-10-31
6 for 6 refactor

rupertbates:
## Why are you doing this?
The way we have dealt with 6 for 6 up until now has been quite problematic, we have a `SixWeekly` billing period which we have used to represent it in Javascript, but this didn't have the 6 for 6 promotion attached to it or the Quarterly price, both of which are what give us the pricing terms to display to users. This meant that we had quite a bit of code which tried to swap out `SixWeekly` for `Quarterly` at points where we needed pricing or access to the promotion, and this in turn meant that you were never quite sure which billing period you were using.

This PR changes this set up so that the `SixWeekly` billing period now has the `Quarterly` price in the `ProductPrices` object and the 6 for 6 promotion has moved from the `Quarterly` billing period to `SixWeekly`. This is a much better representation of how prices are actually made up and means that we no longer need to count 6 for 6 as a special case in Javascript.

The final change is that the Javascript code is no longer switching to `SixWeekly` to `Quarterly` when it submits the create subscription request, I have moved the code to handle this into Scala as that is where we have the rest of the code to abstract the peculiarities of our Zuora catalog

[**Trello Card**](https://trello.com/c/ZxFfpHXx/2701-refactor-6-for-6-promotion)


support-frontend/assets/helpers/productPrice/__tests__/weeklyProductPrice.js
```diff
@@ -17,9 +17,10 @@ const productPrices = {
   'United Kingdom': {
     RestOfWorld: {
       NoProductOptions: {
-        Quarterly: {
+        Quarterly: { GBP: { price: 60, currency: 'GBP', promotions: [] } },
```
rupertbates: You can ignore this file it's all just test data

support-models/src/main/scala/com/gu/support/promotions/Promotion.scala
```diff
@@ -48,12 +48,12 @@ object Promotions {
     description = "Introductory offer",
     appliesTo = AppliesTo(
       Set(
-        "2c92a0086619bf8901661ab02752722f",
-        "2c92a0fe6619b4b301661aa494392ee2",
-        "2c92c0f9660fc4d70166109c01465f10",
-        "2c92c0f8660fb5d601661081ea010391",
-        "2c92c0f965f2122101660fb81b745a06",
-        "2c92c0f965dc30640165f150c0956859"
+        "2c92c0f965f2122101660fbc75a16c38",
```
rupertbates: This changes the product rate plan that the 6 for 6 offer applies to to `SixWeekly` rather than `Quarterly`

support-services/src/main/scala/com/gu/support/catalog/CatalogService.scala
```diff
@@ -14,6 +14,34 @@ object CatalogService {
...
+  private def getRatePlanId(billingPeriod: BillingPeriod, fulfilmentOptions: FulfilmentOptions)
+    = getProductRatePlan(environment, billingPeriod,fulfilmentOptions, NoProductOptions).map(_.id).getOrElse("")
+
+  private def fetchQuarterlyPrice(quarterlyId: ProductRatePlanId, sixWeeklyPriceList: Pricelist, catalogPrices: List[Pricelist]) = {
+    Pricelist(sixWeeklyPriceList.productRatePlanId,
+      catalogPrices.find(_.productRatePlanId == quarterlyId).map(_.prices).getOrElse(sixWeeklyPriceList.prices)
+    )
+  }
+
+  def adjustSixWeeklyPriceList(c: Catalog) = {
```
rupertbates: All of this is just to map between Zuora's representation of a 6 for 6 subs which is made up of 2 separate rate plans and the representation we use on the site which is a quarterly subscription with an introductory promotion

support-workers/src/main/scala/com/gu/zuora/ProductSubscriptionBuilders.scala
```diff
@@ -128,17 +118,21 @@ object ProductSubscriptionBuilders {
         case Failure(e) => throw new BadRequestException(s"First delivery date was not provided. It is required for a Guardian Weekly subscription.", e)
       }
 
-      val productRatePlanId = getProductRatePlanId(catalog.GuardianWeekly, guardianWeekly, isTestUser)
+      val recurringProductRatePlanId = validateRatePlan(guardianWeekly.productRatePlan(getTouchPointEnvironment(isTestUser)), guardianWeekly.describe)
```
rupertbates: Because a 6 for 6 sub in Zuora is made up of two separate rate plans we need to get the id for both, one for the standard recurring part and one for the promotional part

support-models/src/main/scala/com/gu/support/config/TouchPointEnvironments.scala
```diff
@@ -16,12 +16,13 @@ object TouchPointEnvironments {
...
-  def fromStage(stage: Stage): TouchPointEnvironment =
-    stage match {
-      case Stages.DEV => SANDBOX
-      case Stages.CODE => SANDBOX
-      case Stages.PROD => PROD
-    }
+  def fromStage(stage: Stage, isTestUser: Boolean = false): TouchPointEnvironment =
+    if (isTestUser) UAT else
+      stage match {
+        case Stages.DEV => SANDBOX
```
alexflav23: You can use `|` syntax to flatten this. `case Stages.DEV | Stages.CODE => SANDBOX`

support-models/src/main/scala/com/gu/support/workers/ProductTypeExtensions.scala
```diff
@@ -0,0 +1,63 @@
...
+    ratePlans: Map[TouchPointEnvironment, scala.List[ProductRatePlan[Product]]],
+    predicate: ProductRatePlan[Product] => Boolean): Option[ProductRatePlan[Product]] =
+    ratePlans.getOrElse(environment, Nil).find(predicate)
+
+  implicit class ContributionExtensions(contribution: Contribution) {
+    def productRatePlan(environment: TouchPointEnvironment): Option[ProductRatePlan[Product]] =
+      None
+  }
+
+  implicit class DigitalPackExtensions(digitalPack: DigitalPack) {
```
alexflav23: Should use a value class and use `AnyVal`.

support-models/src/main/scala/com/gu/support/workers/ProductTypeExtensions.scala
```diff
@@ -0,0 +1,63 @@
...
+    def productRatePlan(environment: TouchPointEnvironment): Option[ProductRatePlan[Product]] =
+      productRatePlanFromPlans(environment, digitalPack.catalogType.ratePlans, productRatePlanPredicate)
+
+    def productRatePlanPredicate(productRatePlan: ProductRatePlan[Product]): Boolean =
+      productRatePlan.billingPeriod == digitalPack.billingPeriod
+  }
+
+  implicit class PaperExtensions(paper: Paper) {
```
alexflav23: Again you cna use `AnyVal` here too for more performance.

support-models/src/main/scala/com/gu/support/workers/ProductTypeExtensions.scala
```diff
@@ -0,0 +1,63 @@
...
+    def productRatePlan(environment: TouchPointEnvironment): Option[ProductRatePlan[Product]] =
+      productRatePlanFromPlans(environment, paper.catalogType.ratePlans, productRatePlanPredicate)
+
+    def productRatePlanPredicate(productRatePlan: ProductRatePlan[Product]): Boolean =
+      productRatePlan.fulfilmentOptions == paper.fulfilmentOptions &&
+        productRatePlan.productOptions == paper.productOptions
+  }
+
+  implicit class GuardianWeeklyExtensions(guardianWeekly: GuardianWeekly) {
```
alexflav23: Could be a value class.

support-services/src/main/scala/com/gu/support/catalog/CatalogService.scala
```diff
@@ -14,6 +14,34 @@ object CatalogService {
 }
 
 class CatalogService(val environment: TouchPointEnvironment, jsonProvider: CatalogJsonProvider) extends LazyLogging {
+
+  private def getRatePlanId(billingPeriod: BillingPeriod, fulfilmentOptions: FulfilmentOptions)
+    = getProductRatePlan(environment, billingPeriod,fulfilmentOptions, NoProductOptions).map(_.id).getOrElse("")
+
+  private def fetchQuarterlyPrice(quarterlyId: ProductRatePlanId, sixWeeklyPriceList: Pricelist, catalogPrices: List[Pricelist]) = {
```
alexflav23: Line too long, we should do each argument on a line. Also same feedback as above, we may only need `private[this]`.

support-workers/src/main/scala/com/gu/support/workers/lambdas/SendThankYouEmail.scala
```diff
@@ -41,8 +44,18 @@ class SendThankYouEmail(thankYouEmailService: EmailService, servicesProvider: Se
     case _ => Future.successful(None)
   }
 
+  def getProductRatePlanId(product: ProductType, isTestUser: Boolean) = {
+    val touchpointEnvironment = TouchPointEnvironments.fromStage(Configuration.stage, isTestUser)
+    (product match {
+      case d: DigitalPack => d.productRatePlan(touchpointEnvironment)
+      case p: Paper => p.productRatePlan(touchpointEnvironment)
```
alexflav23: Do we need this if `product` is already the correct ADT, or is it because we added `produxtRatePlan`.

support-services/src/test/scala/com/gu/support/pricing/PriceSummaryServiceSpec.scala
```diff
@@ -55,6 +53,11 @@ class PriceSummaryServiceSpec extends FlatSpec with Matchers {
     guardianWeekly(UK)(Domestic)(NoProductOptions)(Annual)(GBP).promotions
       .find(_.promoCode == GuardianWeekly.AnnualPromoCode).get.discountedPrice shouldBe Some(135.00)
 
+    // SixWeekly should have the 6 for 6 promotion and the discount
+    guardianWeekly(UK)(Domestic)(NoProductOptions)(SixWeekly)(GBP).promotions.size shouldBe 2
+    guardianWeekly(UK)(Domestic)(NoProductOptions)(SixWeekly)(GBP).promotions
+      .find(_.promoCode == GuardianWeekly.SixForSixPromoCode).get.introductoryPrice.get.price shouldBe 6
```
alexflav23: There's a trait called `OptionValues` in ScalaTest which we should prefer here, which will give you a `.value` method to use on testing options. It will wrap exceptions  thrown by`.get` in the right way.

support-services/src/main/scala/com/gu/support/catalog/CatalogService.scala
```diff
@@ -14,6 +14,34 @@ object CatalogService {
...
+    // price, ie. the quarterly price as the £6 is available through the introductory promotion object
+    val ratePlanIdsToSwap = Map(
+      getRatePlanId(SixWeekly, Domestic) -> getRatePlanId(Quarterly, Domestic),
+      getRatePlanId(SixWeekly, RestOfWorld) -> getRatePlanId(Quarterly, RestOfWorld)
+    )
+
+    Catalog(
+      c.prices.map(priceList =>
+        ratePlanIdsToSwap.get(priceList.productRatePlanId)
+          .map(quarterlyId => fetchQuarterlyPrice(quarterlyId, priceList, c.prices))
```
alexflav23: You can flatten the lambda if you want: `.map(fetchQuarterlyPrice(_, priceList, c.prices))`

support-services/src/main/scala/com/gu/support/catalog/CatalogService.scala
```diff
@@ -14,6 +14,34 @@ object CatalogService {
 }
 
 class CatalogService(val environment: TouchPointEnvironment, jsonProvider: CatalogJsonProvider) extends LazyLogging {
+
+  private def getRatePlanId(billingPeriod: BillingPeriod, fulfilmentOptions: FulfilmentOptions)
```
alexflav23: `private` allows cross instance method access, meaning all instances of CatalogueService can access each other's private methods. I think what you want here is `private[this]`, which is a cheaper inline accessor.

support-workers/src/main/scala/com/gu/zuora/ProductSubscriptionBuilders.scala
```diff
@@ -5,37 +5,26 @@ import java.util.UUID
...
-      case dp: DigitalPack => ratePlans.find(rp => rp.billingPeriod == dp.billingPeriod).map(_.id)
-      case p: Paper => ratePlans.find(rp => rp.fulfilmentOptions == p.fulfilmentOptions && rp.productOptions == p.productOptions).map(_.id)
-      case gw: GuardianWeekly => ratePlans.find(rp => rp.billingPeriod == gw.billingPeriod && rp.fulfilmentOptions == gw.fulfilmentOptions).map(_.id)
-      case _ => None
-    }
+  private def getTouchPointEnvironment(isTestUser: Boolean) = TouchPointEnvironments.fromStage(Configuration.stage, isTestUser)
 
-    Try(maybeProductRatePlanId.get) match {
+  def validateRatePlan(maybeProductRatePlan: Option[ProductRatePlan[catalog.Product]], productDescription: String): ProductRatePlanId =
+    Try(maybeProductRatePlan.map(_.id).get) match {
```
alexflav23: Why `.get` instead of a direct match on the `Option`? We're introducing a `Try` largely for the sake of it.

support-models/src/main/scala/com/gu/support/workers/ProductTypeExtensions.scala
```diff
@@ -0,0 +1,63 @@
...
+
+object ProductTypeExtensions {
+
+  protected def productRatePlanFromPlans(
+    environment: TouchPointEnvironment,
+    ratePlans: Map[TouchPointEnvironment, scala.List[ProductRatePlan[Product]]],
+    predicate: ProductRatePlan[Product] => Boolean): Option[ProductRatePlan[Product]] =
+    ratePlans.getOrElse(environment, Nil).find(predicate)
+
+  implicit class ContributionExtensions(contribution: Contribution) {
```
alexflav23: Could be a `value class` for automated inlining.

PR #2181, opened 2019-11-01
Stripe Payment Method creation errors: stringify error object in log

tomrf1:
I've noticed a few error logs in sentry caused by failure to request a new Payment Method from Stripe. The error logging is currently not very helpful though:
`Error creating Payment Method: [object Object]`


tomrf1:
Just a follow-up:
The error from Stripe is `card_declined`. I didn't realise it can decline cards at this point in the flow.
The user will be shown a generic payment failure message, so all is fine.

PR #2182, opened 2019-11-01
Payment request button: get user's state

tomrf1:
## Why are you doing this?
We will soon support the Payment Request button for recurring contributions.
This PR is to get the user's state/province if they are in the US/Canada - this is required by support-workers.
Note - it gets the state from the user's billing address associated with the chosen payment method.
If for some reason the state is not available then it fails with a general error ([it's not customisable](https://stripe.com/docs/stripe-js/reference#payment-response-object)). I tested this by using a UK address on the US landing page:
Screenshot 2019-11-01 at 09 07 52


tomrf1:
@jlieb10 I agree it's very confusing having 2 meanings of the word 'state' - this is already the case across the codebase but it would be nice to change it.

tomrf1:
@jlieb10 I agree it's very confusing having 2 meanings of the word 'state' - this is already the case across the codebase but it would be nice to change it.

PR #2183, opened 2019-11-01
Work in progress: Transaction fees, client-side including analytics

jlieb10:
## Why are you doing this?
We are introducing the ability for contributors to cover transaction fees, pending analysis from finance. This adds the fee to the contribution and sends the fee amount as part of the acquisition data object.

## Changes

* Change 1
* Change 2

### DIFF
--- a/support-frontend/assets/helpers/tracking/acquisitions.js
+++ b/support-frontend/assets/helpers/tracking/acquisitions.js
@@ -11,6 +11,7 @@ import type { Participations } from 'helpers/abTests/abtest';
 import * as storage from 'helpers/storage';
 import { getAllQueryParamsWithExclusions } from 'helpers/url';
 import { getCampaignName } from 'helpers/campaigns';
+import { getTransactionFee } from 'helpers/contributions';
 
 // ----- Types ----- //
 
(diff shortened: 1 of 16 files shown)


support-frontend/assets/helpers/tracking/acquisitions.js
```diff
@@ -223,6 +225,8 @@ const getAbTests = (
 function derivePaymentApiAcquisitionData(
   referrerAcquisitionData: ReferrerAcquisitionData,
   nativeAbParticipations: Participations,
+  baseContributionAmount?: number,
```
tomrf1: does `baseContributionAmount` need to be optional? Shouldn't we always have it at this point?

support-frontend/assets/helpers/tracking/acquisitions.js
```diff
@@ -59,6 +60,7 @@ export type PaymentAPIAcquisitionData = {|
   source: ?string,
   abTests: ?AcquisitionABTest[],
   gaId: ?string,
+  transactionFee?: number,
```
tomrf1: 👍

tomrf1: we'll need to update the model in payment-api before putting this live, but this is the right approach.

support-frontend/assets/helpers/tracking/acquisitions.js
jlieb10: nope!

PR #2186, opened 2019-11-04
Tweak contribution type query parameter

jlieb10:
## Why are you doing this?
Best practice is to use `kebab-case` instead of `camelCase` for URLs. Also, allowing for the new `single` terminology to be used instead of `one_off`

[**Trello Card**](https://trello.com/c/yd5RqBQQ/1594-landing-page-ab-test-one-product-only-single)

### DIFF
--- a/support-frontend/assets/helpers/contributions.js
+++ b/support-frontend/assets/helpers/contributions.js
@@ -270,6 +270,8 @@ function toContributionType(s: ?string): ?ContributionType {
         return 'MONTHLY';
       case 'ONE_OFF':
         return 'ONE_OFF';
+      case 'SINGLE':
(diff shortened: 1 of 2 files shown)


support-frontend/assets/helpers/contributions.js
```diff
@@ -270,6 +270,8 @@ function toContributionType(s: ?string): ?ContributionType {
         return 'MONTHLY';
       case 'ONE_OFF':
```
tjsilver: Do we need a TODO to remember to remove ONE_OFF in the future?

jlieb10: I think we can leave it as it might be variably referred to as either

PR #2187, opened 2019-11-05
[WIP] Temporarily disable postcode lookup

jfsoul:
## Why are you doing this?

We appear to have hit our account limit of requests for the address finder service getAddressIO.  This temporarily removes the lookup functionality from checkouts altogether while we update our account.

Next steps:
- Improve error message for when service returns this kind of error (i.e. 5XX)
- Put this removal behind a switch to make this easier if it happens again

## Screenshots

image

### DIFF
--- a/support-frontend/assets/components/subscriptionCheckouts/address/postcodeFinder.jsx
+++ b/support-frontend/assets/components/subscriptionCheckouts/address/postcodeFinder.jsx
@@ -76,39 +76,40 @@ class PostcodeFinder extends Component<PropTypes> {
     const {
       id, postcode, results, isLoading, setPostcode, fetchResults, error, onPostcodeUpdate, onAddressUpdate,
     } = this.props;
-    return (
-      <div>
-        <ComposedInputWithButton
-          error={error}
-          label="Postcode"
-          onClick={() => { fetchResults(postcode); }}
-          id={id}
-          setValue={(val) => {
-            setPostcode(val);
-            onPostcodeUpdate(val);
-          }}
-          isLoading={isLoading}
-          value={postcode}
-        />
-        {(results.length > 0) &&
-          <ComposedSelect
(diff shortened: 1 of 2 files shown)


support-frontend/assets/pages/contributions-landing/components/StripePaymentRequestButton/StripePaymentRequestButton.jsx
```diff
@@ -213,7 +213,7 @@ function setUpPaymentListener(props: PropTypes, paymentRequest: Object, paymentM
 
     const stateUpdateOk =
       props.stripeAccount !== 'ONE_OFF' && (props.countryGroupId === UnitedStates || props.countryGroupId === Canada) ?
-      updatePayerState(token, props.updateState) : true;
+        updatePayerState(token, props.updateState) : true;
```
lucymonie: Are these just linting changes? John Duffell would definitely ask you to remove these 😉

support-frontend/assets/components/subscriptionCheckouts/address/postcodeFinder.jsx
```diff
@@ -76,21 +76,22 @@ class PostcodeFinder extends Component<PropTypes> {
...
     } = this.props;
-    return (
-      <div>
-        <ComposedInputWithButton
-          error={error}
-          label="Postcode"
-          onClick={() => { fetchResults(postcode); }}
-          id={id}
-          setValue={(val) => {
+    return true ? '' :
```
lucymonie: I probably wouldn't put a switch inside the component itself, but in the place where the component sits. I would make the page responsible for determining whether to render it. But then I guess you have to implement it three times, which is more tedious to manage, so: okay.

Is there any way to change this so that it not a raw boolean that might be confusing to future devs?

PR #2189, opened 2019-11-05
Better handling of payment request button errors

tomrf1:
## Why are you doing this?
Currently, the Payment Request button is only enabled for single contributions.
Error handling is currently done within the browser popup, e.g.:
Screenshot 2019-11-05 at 11 51 10

We will soon be enabling the Payment Request button for recurring as well.
There are a couple of issues with error handling within the popup:
1. We don't have control over what the error says
2. The Payment Request button has a 30 second timeout if we do not call the `complete` callback.

The 2nd point here is particularly problematic, because our backend for recurring (support-workers) can in extreme cases take 30 seconds to complete. This means the popup displays an error even though the payment eventually completes successfully and the thankyou page is displayed:
Screenshot_2019-10-31_at_13 19 37

The solution is to stop using the popup for displaying errors.
Instead, we can dismiss the popup immediately after the user clicks submit and use our page to display progress/error information.
This means when the user clicks submit, our standard 'spinner' is displayed:
Screenshot 2019-11-05 at 11 57 34

And failures are displayed below the Payment Request button:
Screenshot 2019-11-05 at 11 58 08

The payment error message for the Payment Request button is now stored separately from the existing error message. This means that if the user attempts payment with both the Payment Request button and the form below, and both attempts fail, then 2 separate errors are stored and displayed:
Screenshot 2019-11-05 at 11 30 11

There's also a new error message for when the browser's saved details are incomplete:
Screenshot 2019-11-05 at 14 14 33


support-frontend/assets/helpers/errorReasons.js
```diff
@@ -45,6 +46,8 @@ function appropriateErrorMessage(errorReason: ?ErrorReason): ?string {
       return 'Sorry, something has gone wrong. Please try again, or come back later.';
     case 'card_authentication_error':
       return 'You have not been charged. Please check your payment details and try again, or choose another payment method.';
+    case 'incomplete_payment_request_details':
+      return 'TODO - message about missing payment request details.';
```
tomrf1: awaiting final copy

support-frontend/assets/pages/contributions-landing/components/StripePaymentRequestButton/StripePaymentRequestButton.jsx
```diff
@@ -144,19 +154,15 @@ function updatePayerState(token: Object, setState: (UsState | CaState | null) =>
...
+
 }
 
-// Calling the complete function will close the pop up payment window
-const onComplete = (complete: Function) => (res: PaymentResult) => {
+const onComplete = () => (res: PaymentResult) => {
   if (res.paymentStatus === 'success') {
     trackComponentClick('apple-pay-payment-complete');
-    complete('success');
-  } else if (res.paymentStatus === 'failure') {
```
tomrf1: no longer any need to check for failure here, as we're not using the browser popup

support-frontend/assets/pages/contributions-landing/components/StripePaymentRequestButton/StripePaymentRequestButton.jsx
```diff
@@ -208,11 +214,18 @@ const availablePaymentRequestButtonPaymentMethod = (result: Object): StripePayme
 
 function setUpPaymentListener(props: PropTypes, paymentRequest: Object, paymentMethod: StripePaymentMethod) {
   paymentRequest.on('token', ({ complete, token, ...data }) => {
+
+    // Always dismiss the payment popup immediately - any pending/success/failure will be displayed on our own page.
+    // This is because `complete` must be called within 30 seconds or the user will see an error.
+    // Our backend (support-workers) can in extreme cases take longer than this, so we must call complete now.
+    // This means that the browser's payment popup will be dismissed, and our own 'spinner' will be displayed until
+    // the backend job finishes.
```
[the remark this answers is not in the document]
tjsilver: 👍

PR #2196, opened 2019-11-06
Feature/remove old checkout

alexflav23:
## Why are you doing this?

[**Trello Card**](https://trello.com/c/96kKFLLT/2719-remove-old-checkout-code-implementation-from-support-frontend)

## Changes

* Removed `stripeToken` references from everywhere.
* Removed feature flag `window.guardian.stripeElementsSubscriptions`. Stripe Payment method API now the only available Stripe Checkout mechanism.
* Removed if condition in `isPostDeployUser()`, Selenium tests use payment intent API for checkout just like normal users, with the fake `4242 4242 4242 4242`credit card.

### DIFF
--- a/support-frontend/app/controllers/Application.scala
+++ b/support-frontend/app/controllers/Application.scala
@@ -65,8 +65,8 @@ class Application(
   def geoRedirect: Action[AnyContent] = GeoTargetedCachedAction() { implicit request =>
     val redirectUrl = request.geoData.countryGroup match {
       case Some(UK) => buildCanonicalShowcaseLink("uk")
-      case Some(US) => "/us/support"
-      case Some(Australia) => "/au/support"
+      case Some(US) => "/us/contribute"
+      case Some(Australia) => "/au/contribute"
       case Some(Europe) => "/eu/contribute"
       case Some(Canada) => "/ca/contribute"
       case Some(NewZealand) => "/nz/contribute"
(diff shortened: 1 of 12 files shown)


support-frontend/app/controllers/Application.scala
```diff
@@ -65,8 +65,8 @@ class Application(
   def geoRedirect: Action[AnyContent] = GeoTargetedCachedAction() { implicit request =>
     val redirectUrl = request.geoData.countryGroup match {
       case Some(UK) => buildCanonicalShowcaseLink("uk")
-      case Some(US) => "/us/contribute"
-      case Some(Australia) => "/au/contribute"
+      case Some(US) => "/us/support"
```
rupertbates: This seems unrelated to this change and will conflict with the PR to do this work, I'd revert it

support-frontend/assets/helpers/subscriptionsForms/submit.js
```diff
@@ -195,36 +194,16 @@ function checkStripeUserType(
   isTestUser: boolean,
   price: number,
   currency: IsoCurrency,
-  stripeToken: Option<string>,
-  stripePaymentMethodId: Option<string>,
+  stripePaymentMethodId: ?string,
```
rupertbates: Probably want `Option<string> ` here? Otherwise it will accept undefined

support-frontend/assets/helpers/subscriptionsForms/submit.js
```diff
@@ -236,12 +215,11 @@ function showPaymentMethod(
   currency: IsoCurrency,
   country: IsoCountry,
   paymentMethod: Option<PaymentMethod>,
-  stripeToken: Option<string>,
-  stripePaymentMethod: Option<string>,
+  stripePaymentMethod: ?string,
```
rupertbates: Probably want Option<string> here too

PR #2197, opened 2019-11-07
Failure emails

alexflav23:
## Why are you doing this?

This is a 2 in 1 special PR

- Print failure email Trello Card: [**Trello Card**](https://trello.com/c/Jrau1rZr/2271-print-failure-email)
- Guardian weekly failed email: https://trello.com/c/dLmrZO7r/2337-guardian-weekly-failure-email

## Changes

* Added classes for `GuardianWeeklyFailed` and `PrintFailed` emails.
* Added manual tests to check against live Braze deployment/membership-workflow that they work.

### DIFF
--- a/support-workers/src/test/scala/com/gu/support/workers/lambdas/FailureHandlerSpec.scala
+++ b/support-workers/src/test/scala/com/gu/support/workers/lambdas/FailureHandlerSpec.scala
@@ -3,7 +3,7 @@ package com.gu.support.workers.lambdas
 import java.io.ByteArrayOutputStream
 
 import com.amazonaws.services.sqs.model.SendMessageResult
-import com.gu.emailservices.{EmailService, FailedContributionEmailFields, FailedDigitalPackEmailFields, IdentityUserId}
+import com.gu.emailservices.{EmailService, FailedContributionEmailFields, FailedDigitalPackEmailFields, FailedGuardianWeeklyEmailFields, FailedPaperEmailFields, IdentityUserId}
 import com.gu.monitoring.SafeLogger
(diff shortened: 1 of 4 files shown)


support-workers/src/test/scala/com/gu/support/workers/lambdas/FailureHandlerSpec.scala
```diff
@@ -181,6 +181,49 @@ object DigiPackFailureHandlerManualTest extends Matchers {
         sys.exit(0)
       }
   }
+}
+
+object GuardianWeeklyFailureHandlerTest extends Matchers {
```
jfsoul: Do you think there’s any unit tests to add as well, or do the existing ones still cover?

alexflav23: @jfsoul Unit tests are not possible here unfortunately, these are manual checks only. How would you unit that that a specific email box has received an email without writing a mail server? I might be wrong but I don't think that can work.

jfsoul: Sure, probably not - I was just thinking was there anything to add in `FailureHandlerSpec` class for example, but probably new product failure emails don't change failure handling as such, just which template to trigger.

alexflav23: @jfsoul Exactly, the way we handle the failure doesn't change, so we have specific objects to trigger sending a specific failure email to an email address. We then look at the sent email and assess if it gets sent and the format is correct. It's fairly manual.

PR #2200, opened 2019-11-07
Add postcode lookup flag

alexflav23:
## Why are you doing this?

[**Trello Card**](https://trello.com/c/OQhOWhfc/2711-put-the-address-finder-aka-postcode-lookup-behind-a-switch)
[**Address lookup error**](https://trello.com/c/wvgdTwTm/2712-add-distinct-error-message-on-the-address-finder-for-an-internal-server-error)

## Changes

* Added a switch called `checkoutPostcodeLookup` to enable/disable using the external address service during postcode lookup.
* Change logic to return `Promise.resolve([])` when postcode lookup is disabled, by-passing the API Call. Designed to be used only when external service is unavailable.
* Not sure how to use `getGlobal` convention on this one. @rupertbates please advise.
* Added switches to `DEV`, `CODE`, and `PROD` on Amazon S3. Can be checked here: https://support.gutools.co.uk/switches
* Added a failure handler to return specific error messages based on `GetAddress` backend route response status codes.

### DIFF
--- a/support-frontend/assets/components/subscriptionCheckouts/address/postcodeLookup.js
+++ b/support-frontend/assets/components/subscriptionCheckouts/address/postcodeLookup.js
@@ -7,14 +7,30 @@ export type PostcodeFinderResult = {|
   city?: string,
 |};
 
+
+function handleErrors(response: Response) {
+  if (response.ok) {
+    return response
+  } else {
+    if (response.status === 500) {
+      throw new Error("External address service temporarily unavailable. Please proceed with checkout. We apologise for the inconvenience")
+    } else if (response.status === 400) {
+      throw new Error('We couldn\'t find this postcode, please check and try again or enter your address below.',)
+    } else {
+      throw new Error("Error while contacting address API: ", response.statusText)
+    }
+  }
+}
+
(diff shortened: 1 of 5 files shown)


support-frontend/assets/components/subscriptionCheckouts/address/postcodeLookup.js
```diff
@@ -7,8 +7,14 @@ export type PostcodeFinderResult = {|
   city?: string,
 |};
 
-const getAddressesForPostcode = (postcode: string): Promise<PostcodeFinderResult[]> =>
-  fetch(postcodeLookupUrl(postcode))
-    .then(res => res.json());
+const getAddressesForPostcode = (postcode: string): Promise<PostcodeFinderResult[]> => {
+
+  if (window.guardian.checkoutPostcodeLookup) {
```
alexflav23: @rupertbates How would I use `getGlobals` convention here? I'm not cool enough in react/redux terms to work it out.

rupertbates: Just `getGlobal('checkoutPostcodeLookup')` doesn't need any redux involvement

PR #2215, opened 2019-11-12
Make the support header logo accessible

jlieb10:
## Why are you doing this?
I saw the opportunity for an accessibility quick win by adding alternative text to our linked logo in the header -- I added it and used our accessibility hint mixin to make it SR-only


SiAdcock:
This is great, thanks for doing this. I have a few meta-questions that should not impede merging

- can this also be implemented on the Subscribe landing page? The alt text currently reads "The Guardian logo", which is not as helpful as "Return to The Guardian"
- I know the mixin wasn't implemented here, but the mixin's name `accessibility-hint` is not as strong as it could be. We have a similar mixin in the design system called `visuallyHidden` which I think more strongly conveys the intention
- This mixin has a bit more going on than [the design system's equivalent](https://github.com/guardian/source-components/commit/1da3eb29ca6a244243d0dd1401ec3d4708394ccb#diff-bfbf76d70befa2c7a06917b6b8b41ba4R3-R10). We might steal some bits! Do you happen to know what the purpose of the negative margin is?

jlieb10:
hey @SiAdcock thanks!

in response to your questions:

1. I'm going to tag @lucymonie and @damien-shaw-guardian in case either of them want to update it for subs

2. okay! will update the name

3. i'm not sure at all -- this code smells of 'i saw it on stack overflow' (which is fine! but would be nice to understand it completely)

jlieb10:
actually nevermind @damien-shaw-guardian and @lucymonie - i took care of the link updates requested by @SiAdcock

PR #2217, opened 2019-11-14
Send Stripe payment type to backend for recurring/subscriptions

tomrf1:
## Why are you doing this?
There are several types of Stripe payments in the [ophan model](https://github.com/guardian/ophan/blob/master/event-model/src/main/thrift/acquisition.thrift#L35).
But only one is currently supported by support-workers for recurring/subscriptions.
We need it to also support `STRIPE_APPLE_PAY` and `STRIPE_PAYMENT_REQUEST_BUTTON` for our analytics.

This change makes the client send a new field, `stripePaymentType`, to support-workers for Stripe payments.
In a separate PR I will change support-workers to receive this field and to set the paymentProvider field appropriately in the acquisition event.

This is the payload being sent to the backend, now with the new field:
Screenshot 2019-11-14 at 12 59 58

### DIFF
--- a/support-frontend/assets/helpers/paymentIntegrations/readerRevenueApis.js
+++ b/support-frontend/assets/helpers/paymentIntegrations/readerRevenueApis.js
@@ -56,8 +56,14 @@ type ProductFields = RegularContribution | DigitalSubscription | PaperSubscripti
 
 type RegularPayPalPaymentFields = {| baid: string |};
 
-type RegularStripePaymentIntentFields = {| paymentMethod: string |};
-type RegularStripeCheckoutPaymentFields = {| stripeToken: string |};
+type RegularStripePaymentIntentFields = {|
+  paymentMethod: string,                  // The ID of the Stripe Payment Method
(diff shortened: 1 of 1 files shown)


support-frontend/assets/helpers/paymentIntegrations/readerRevenueApis.js
```diff
@@ -34,6 +34,9 @@ import type { Title } from 'helpers/user/details';
 
 // ----- Types ----- //
 
+export type StripePaymentMethod = 'StripeCheckout' | 'StripeApplePay' | 'StripePaymentRequestButton' | 'StripeElements';
```
tomrf1: just moved these 2 types to the top to avoid use-before-definition

paulbrown1982: We should check later whether there's a way to see if people are using Google Pay as part of the payment request button. But don't let that hold up this going out.

tomrf1: good point, will create a card

PR #2219, opened 2019-11-15
Send Stripe payment type to ophan for recurring contributions

tomrf1:
## Why are you doing this?
There are several types of Stripe payments in the ophan model.
But only one is currently supported by support-workers for recurring/subscriptions.
We need it to also support `STRIPE_APPLE_PAY` and `STRIPE_PAYMENT_REQUEST_BUTTON` for our analytics.

1. receive the new `stripePaymentType` field from the client
2. add it to the `PaymentFields` object
3. send the new types to ophan in the acquisition event

The client already sends a `stripePaymentType` field - https://github.com/guardian/support-frontend/pull/2217

### DIFF
--- a/support-workers/src/main/scala/com/gu/support/workers/lambdas/PreparePaymentMethodForReuse.scala
+++ b/support-workers/src/main/scala/com/gu/support/workers/lambdas/PreparePaymentMethodForReuse.scala
@@ -72,7 +72,8 @@ class PreparePaymentMethodForReuse(servicesProvider: ServiceProvider = ServicePr
           creditCardExpirationMonth = cardResponse.creditCardExpirationMonth,
           creditCardExpirationYear = cardResponse.creditCardExpirationYear,
           creditCardType = cardResponse.creditCardType,
-          paymentGateway = paymentGateway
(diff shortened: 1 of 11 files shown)


support-workers/src/main/scala/com/gu/support/workers/lambdas/PreparePaymentMethodForReuse.scala
```diff
@@ -72,7 +72,8 @@ class PreparePaymentMethodForReuse(servicesProvider: ServiceProvider = ServicePr
           creditCardExpirationMonth = cardResponse.creditCardExpirationMonth,
           creditCardExpirationYear = cardResponse.creditCardExpirationYear,
           creditCardType = cardResponse.creditCardType,
-          paymentGateway = paymentGateway
+          paymentGateway = paymentGateway,
+          stripePaymentType = None
```
tomrf1: this info has presumably come from zuora, so we do not have the `stripePaymentType`

tomrf1:
@paulbrown1982 thanks. These extra stripe payment types are already sent by payment-api and correctly handled by ophan, so it should be fine

PR #2221, opened 2019-11-18
Remove test to recognize people who gave single contributions in the past

jlieb10:
## Why are you doing this?
This test, which has been running since August, conflicts with the moment landing page copy changes.

Because of a bunch of fake contributions that made their way into our actual contributions table on 11 September 2019, the test needs to be manually analyzed excluding this date. The big picture when doing this analysis is a global AV uplift of +5% from a very small number of contributors.


tomrf1:
👍 thanks!

PR #2226, opened 2019-11-19
Remove Stripe Checkout for recurring contributions

tomrf1:
## Why are you doing this?
We have been running an A/B test to compare Stripe Elements for recurring contributions vs the old Stripe Checkout (popup) implementation.
Stripe Elements has performed better, and so Stripe Checkout can now be removed.

Note - Stripe Checkout is still used by the Payment Request button (though it doesn't use the popup). This is the last thing to be migrated to the SCA-compliant Stripe integration. Then we can clean up any remaining Stripe Checkout code.

I've tested card and Payment Request button payments.

Screenshot 2019-11-19 at 15 44 51

### DIFF
--- a/support-frontend/assets/pages/contributions-landing/components/ContributionForm.jsx
+++ b/support-frontend/assets/pages/contributions-landing/components/ContributionForm.jsx
@@ -15,7 +15,6 @@ import {
 } from 'helpers/contributions';
 import { type ErrorReason } from 'helpers/errorReasons';
 import type { IsoCountry } from 'helpers/internationalisation/country';
-import { openDialogBox } from 'helpers/paymentIntegrations/stripeCheckout';
 import { type PaymentAuthorisation } from 'helpers/paymentIntegrations/readerRevenueApis';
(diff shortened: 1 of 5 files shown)


support-frontend/assets/pages/contributions-landing/components/ContributionForm.jsx
```diff
@@ -247,9 +220,9 @@ function withProps(props: PropTypes) {
   const baseClass = 'form';
 
   const classModifiers = ['contribution', 'with-labels'];
-
+  
   const showSecureStripeContainer: boolean = props.paymentSecuritySecureTransactionGreyNonUKVariant !== 'control' || props.countryGroupId === 'GBPCountries';
-  const showSecureButtonBg: boolean = showSecureStripeContainer && props.paymentMethod === Stripe && (props.stripeElementsRecurringTestVariant === 'stripeElements' || props.contributionType === 'ONE_OFF');
+  const showSecureButtonBg: boolean = showSecureStripeContainer && props.paymentMethod === Stripe;
```
tjsilver: 🥇

PR #2229, opened 2019-11-21
Bump acquisition event producer for new component type

jfsoul:
## Why are you doing this?
Using latest version of acquisition event producer library so that we can use the new component type for the subscriptions banner.

There was a transitive dependency bump for circe, so to make things easier I have bumped here too.

## Changes

* Change 1
* Change 2

### DIFF
--- a/support-frontend/app/views/ViewHelpers.scala
+++ b/support-frontend/app/views/ViewHelpers.scala
@@ -9,5 +9,5 @@ object ViewHelpers {
   def outputJson[T : Encoder](value: T): String =
     value
       .asJson
-      .pretty(Printer.noSpaces.copy(dropNullValues = true))
+      .printWith(Printer.noSpaces.copy(dropNullValues = true))
 }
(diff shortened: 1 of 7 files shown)


support-lambdas/stripe-intent/build.sbt
```diff
@@ -13,6 +13,12 @@ riffRaffManifestProjectName := "support:lambdas:Stripe Intent"
 riffRaffArtifactResources += (file("support-lambdas/stripe-intent/cfn.yaml"), "cfn/cfn.yaml")
 riffRaffManifestBranch := Option(System.getenv("BRANCH_NAME")).getOrElse("unknown_branch")
 riffRaffBuildIdentifier := Option(System.getenv("BUILD_NUMBER")).getOrElse("DEV")
+assemblyMergeStrategy in assembly := {
+  case x if x.endsWith("module-info.class") => MergeStrategy.discard
+  case y =>
+    val oldStrategy = (assemblyMergeStrategy in assembly).value
+    oldStrategy(y)
+}
```
jfsoul: Needed to add this because of a duplicate error coming from jackson. We have done the same on the support-workers `build.sbt`.

support-frontend/app/views/ViewHelpers.scala
```diff
@@ -9,5 +9,5 @@ object ViewHelpers {
   def outputJson[T : Encoder](value: T): String =
     value
       .asJson
-      .pretty(Printer.noSpaces.copy(dropNullValues = true))
+      .printWith(Printer.noSpaces.copy(dropNullValues = true))
```
jfsoul: `.printWith` has replaced the deprecated `.pretty`

PR #2232, opened 2019-11-22
Fix input empty field placeholder bug

tomrf1:
This has only just become an issue, I think it's related to the preact upgrade which has exposed some faulty logic in this component

### DIFF
--- a/support-frontend/assets/pages/contributions-landing/components/ContributionTextInput.jsx
+++ b/support-frontend/assets/pages/contributions-landing/components/ContributionTextInput.jsx
@@ -12,7 +12,7 @@ type PropTypes = {|
   id: string,
   name: string,
   label: string,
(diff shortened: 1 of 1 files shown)


support-frontend/assets/pages/contributions-landing/components/ContributionTextInput.jsx
```diff
@@ -12,7 +12,7 @@ type PropTypes = {|
   id: string,
   name: string,
   label: string,
-  placeholder: boolean | string,
+  placeholder: boolean | string | null,
```
rupertbates: Is boolean still a valid value for this?

tomrf1: no - just removed it!

PR #2242, opened 2019-11-25
Revert "Revert "New Guardian Weekly gifting v 1.0""

rupertbates:
Reverts guardian/support-frontend#2238

The bug was caused by a change in the original PR which changed the number of preview invoices which we request from Zuora to 1 for gifts so that we didn't send a second one through to the thank you email which would then include it and cause confusion because gifts should be a one time payment.

However it turns out that this is not necessary and actually causes problems when the first delivery date (contract acceptance date) is further in the future than the end of the first billing period. This is because in this case Zuora will not return any invoice data at all and our code will fail to parse the response because it expects that data to be there.

I have changed this behaviour so that firstly we request the standard number of invoices when previewing gifts and secondly the preview response parsing will cope with missing invoice data.

### DIFF
--- a/support-rest/src/main/scala/com/gu/rest/WebServiceHelper.scala
+++ b/support-rest/src/main/scala/com/gu/rest/WebServiceHelper.scala
@@ -104,11 +104,11 @@ trait WebServiceHelper[Error <: Throwable] {
       case "2xx" =>
         decode[A](responseBody).left.map { err =>
           decodeError(responseBody).right.getOrElse(
-            WebServiceHelperError[A](codeBody, s"failed to parse response: $err", err)
+            WebServiceHelperError[A](codeBody, s"failed to parse response. Error was: $err, Response was: $responseBody", err)
           )
         }.toTry
       case "4xx" =>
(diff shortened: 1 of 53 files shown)


support-rest/src/main/scala/com/gu/rest/WebServiceHelper.scala
```diff
@@ -104,7 +104,7 @@ trait WebServiceHelper[Error <: Throwable] {
       case "2xx" =>
         decode[A](responseBody).left.map { err =>
           decodeError(responseBody).right.getOrElse(
-            WebServiceHelperError[A](codeBody, s"failed to parse response: $err", err)
+            WebServiceHelperError[A](codeBody, s"failed to parse response. Error was: $err, Response was: $responseBody", err)
```
jfsoul: Do you want to leave this in? Not sure what uses this helper, so we might be logging something secret (for example).

rupertbates: It goes through the SafeLogger so will be fine, but I will take it out once this has deployed successfully. I'd like to leave it in for now though in case of further errors.

PR #2247, opened 2019-11-26
Refactoring the paypal button

jlieb10:
## Why are you doing this?
We shouldn't have 'createElement' inside the render loop


paulbrown1982:
Is this change still needed @jlieb10 ?

jlieb10:
@paulbrown1982 This work was related to the upgrade to Preact X that we had to rollback. This is a good refactor IMO but does not solve the problem that Preact X introduced. I will close this for now but leave the branch on Github.

PR #2251, opened 2019-11-27
Log lambda execution results to Elasticsearch

rupertbates:
## Why are you doing this?
As part of our health OKR for this quarter we want to start logging all step functions executions to Elasticsearch so that we can do analysis of them more easily.

The way that this works is 
* We log the result of each lambda execution to CloudWatch as Json
* A CloudWatch subscription sends all log messages matching a particular filter to Elasticsearch (via the LogsToElasticsearch_support-elk-domain lambda) 
* Elasticsearch has a pipeline set up which extracts the Json into separate fields

[**Trello Card**](https://trello.com/c/99hEG7T9/2637-failure-reporting-tool)

## Changes
* Create a `LambdaExecutionResult` class to capture the data we are interested in
* Log results from the `FailureHandler` and `SendAcquisitionEvent` lambdas - these are the two exit points for the step function

### DIFF
--- a/support-workers/elasticsearch/cloudWatchToElasticSearchLambda.js
+++ b/support-workers/elasticsearch/cloudWatchToElasticSearchLambda.js
@@ -1,0 +1,261 @@
+// v1.1.2
+var https = require('https');
+var zlib = require('zlib');
+var crypto = require('crypto');
+
+var endpoint = 'vpc-support-elk-domain-vcesheatvi6xspmp4sd6dnfkq4.eu-west-1.es.amazonaws.com';
+
+// Set this to true if you want to debug why data isn't making it to
+// your Elasticsearch cluster. This will enable logging of failed items
+// to CloudWatch Logs.
+var logFailedResponses = false;
+
+exports.handler = function(input, context) {
+  // decode input from base64
+  var zippedInput = new Buffer.from(input.awslogs.data, 'base64');
+
+  // decompress the input
+  zlib.gunzip(zippedInput, function(error, buffer) {
+    if (error) { context.fail(error); return; }
+
+    // parse the input from JSON
+    var awslogsData = JSON.parse(buffer.toString('utf8'));
+
+    // transform the input to Elasticsearch documents
+    var elasticsearchBulkData = transform(awslogsData);
+
+    // skip control messages
+    if (!elasticsearchBulkData) {
+      console.log('Received a control message');
+      context.succeed('Control message handled successfully');
+      return;
+    }
+
+    // post documents to the Amazon Elasticsearch Service
+    post(elasticsearchBulkData, function(error, success, statusCode, failedItems) {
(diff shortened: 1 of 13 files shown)


support-workers/elasticsearch/cloudWatchToElasticSearchLambda.js
```diff
@@ -0,0 +1,261 @@
+// v1.1.2
+var https = require('https');
+var zlib = require('zlib');
+var crypto = require('crypto');
+
+var endpoint = 'vpc-support-elk-domain-vcesheatvi6xspmp4sd6dnfkq4.eu-west-1.es.amazonaws.com';
```
jfsoul: worth putting this as an env variable?

support-workers/elasticsearch/cloudWatchToElasticSearchLambda.js
```diff
@@ -0,0 +1,261 @@
...
+exports.handler = function(input, context) {
+  // decode input from base64
+  var zippedInput = new Buffer.from(input.awslogs.data, 'base64');
+
+  // decompress the input
+  zlib.gunzip(zippedInput, function(error, buffer) {
+    if (error) { context.fail(error); return; }
+
+    // parse the input from JSON
+    var awslogsData = JSON.parse(buffer.toString('utf8'));
```
jfsoul: Would this ever fail and would we care?

support-workers/elasticsearch/cloudWatchToElasticSearchLambda.js
```diff
@@ -0,0 +1,261 @@
...
+    ].join('\n') + '\n';
+  });
+  return bulkRequestBody;
+}
+
+function buildSource(message, extractedFields) {
+  if (extractedFields) {
+    var source = {};
+
+    for (var key in extractedFields) {
```
jfsoul: Could we use `Object.keys` or something and map over the keys?

support-workers/elasticsearch/cloudWatchToElasticSearchLambda.js
```diff
@@ -0,0 +1,261 @@
...
+
+  var canonicalHeaders = Object.keys(request.headers)
+  .sort(function(a, b) { return a.toLowerCase() < b.toLowerCase() ? -1 : 1; })
+  .map(function(k) { return k.toLowerCase() + ':' + request.headers[k]; })
+  .join('\n');
+
+  var signedHeaders = Object.keys(request.headers)
+  .map(function(k) { return k.toLowerCase(); })
+  .sort()
+  .join(';');
```
jfsoul: is it just me or is the indentation a bit off on these vars?

jfsoul: also can we make any/all of the `var`s `const`s? Presumably we're running on a later version of node.

PR #2255, opened 2019-11-28
Add weekly gifting button to showcase page

lucymonie:
## Why are you doing this?
The aim is to draw people's attention to the option to give the Guardian Weekly as a gift on the showcase page, and give them a link directly to the gifting landing page.

[**Trello Card**](https://trello.com/c/GXgcYSBM/2764-gw-gifting-add-a-way-to-get-to-the-gifting-landing-page-from-show-case-page)

## Screenshots
Screen Shot 2019-11-28 at 18 06 45

Screen Shot 2019-11-28 at 18 06 21

Screen Shot 2019-11-28 at 17 52 34

Screen Shot 2019-11-28 at 17 52 20


support-frontend/assets/components/button/_sharedButton.jsx
```diff
@@ -14,6 +14,8 @@ import './button.scss';
 export const Appearances = {
   primary: 'primary',
   secondary: 'secondary',
+  tertiary: 'tertiary',
```
rupertbates: What are 'appearances'?

support-frontend/assets/pages/subscriptions-landing/copy/subscriptionCopy.jsx
```diff
@@ -116,11 +117,20 @@ const guardianWeekly: ProductCopy = {
...
+      ctaButtonText: 'Find out more',
+      link: subsLinks.GuardianWeekly,
+      analyticsTracking: sendTrackingEventsOnClick('weekly_cta', 'GuardianWeekly', abTest),
+    },
+    {
+      ctaButtonText: 'See gift options',
+      link: subsLinks.GuardianWeeklyGift,
+      analyticsTracking: sendTrackingEventsOnClick('weekly_cta_gift', 'GuardianWeekly', abTest),
+      modifierClasses: '',
+      hierarchy: 'second',
```
jfsoul: What's the difference between hierarchy and index? I'm not sure I see how hierarchy is used at the moment as we're only checking for `hierarchy='first'` which isn't ever the case yet?

support-frontend/assets/components/button/button.scss
```diff
@@ -47,6 +48,26 @@
...
+  background-color: transparent;
+  border: solid 1px gu-colour(neutral-20);
+
+  &:hover, &:focus {
+    background-color: gu-colour(neutral-97);
+  }
+}
+
+.component-button--tertiaryFeature {
+  color: #FFE500;
```
damien-shaw-guardian: can we use variables for this,  I think there is one called highlight-main which is a yellow

support-frontend/assets/components/button/button.scss
```diff
@@ -47,6 +48,26 @@
...
+  background-color: transparent;
+  border: solid 1.5px #FFE500;
```
damien-shaw-guardian: can we use variables for this too

support-frontend/assets/components/subscriptionsProductDescription/subscriptionsProductDescription.jsx
```diff
@@ -17,6 +17,17 @@ type PropTypes = {
   buttons: ProductButton[],
 };
 
+const getAppearance = (isFeature, index, hierarchy) => {
```
damien-shaw-guardian: would getButtonAppearance be a better name? as this seems specific to the buttons

support-frontend/assets/components/button/_sharedButton.jsx
lucymonie: This is about the type of look the button has ('primary' is yellow, 'secondary' is solid back with white text, etc).

support-frontend/assets/pages/subscriptions-landing/copy/subscriptionCopy.jsx
lucymonie: Sorry - this is probably a bit opaque from the screen shots I've provided! This is used when there are two different app stores that we link to (on mobile). In this case, the 'equal first' hierarchy overrides the index, which would otherwise make the second button a tertiary rather than secondary button.
Screen Shot 2019-12-02 at 11 43 54

lucymonie: And we only check for 'first' so I will remove the one that's 'second'

support-frontend/assets/components/button/button.scss
lucymonie: Yes, good pickup.

support-frontend/assets/components/button/button.scss
lucymonie: Thanks 👍

support-frontend/assets/components/subscriptionsProductDescription/subscriptionsProductDescription.jsx
lucymonie: Sure

PR #2263, opened 2019-12-02
Update postcodes

alexflav23:
## Why are you doing this?

[**Trello Card**](https://trello.com/c/YhpjgY7L/2713-update-home-delivery-postcodes-with-latest-list)

## Changes

* Added latest postcode list to `deliveryCheck.js`.


support-frontend/assets/helpers/deliveryCheck.js
```diff
@@ -1,6 +1,6 @@
 // @flow
 
-const M25_POSTCODE_PREFIXES = [
+const M25_POSTCODE_OLD_PREFIXES = [
```
rupertbates: Is there a good reason for keeping the separation between old and new?

PR #2268, opened 2019-12-04
Landing page font fix

jlieb10:
## Why are you doing this?
There was an unnecessary stylesheet being imported into the stylesheet for the new landing page. This was causing a CSS reset to happen after much intentional CSS had been applied. This PR removes that import and fixes the bug.

## Screenshots
Fixed form:
fixed form


ionamckendrick:
Thanks so much for doing this so quickly Joshua!

PR #2273, opened 2019-12-09
Don't offer 27th of Dec as a delivery date for GW

rupertbates:
## Why are you doing this?

There is no Guardian Weekly issue on December the 27th so we shouldn't offer this as a potential delivery date.

[**Trello Card**](https://trello.com/c/pxFZq1ZM/2798-exclude-the-27th-of-dec-from-list-of-delivery-dates-for-guardian-weekly)


lucymonie:
Will this need to be removed after Christmas?

rupertbates:
> Will this need to be removed after Christmas?

No it will be fine to leave. SX are doing some work on start dates which we are hoping to integrate into the site in the new year so this will be able to go then

PR #2220, opened 2019-11-15
New landing page template

tjsilver:
## Why are you doing this?

We are testing whether a new design of landing page will provide any uplift.

[**Trello Card**](https://trello.com/c/bv0IWb3W/1562-implement-a-new-template-for-the-landing-page)

## Changes

* Control - original landing page
* new_template - new landing page

### Mobile
image

### Tablet (740px +)
image

### Desktop (980px +)
image

### Wide (1440px +)
image

PR #2279, opened 2019-12-10
Landing page template test, round 2

jlieb10:
## Why are you doing this?
The landing page template test isn't going very well so we are releasing a second round with less variance from the first.

[**Trello Card**](https://trello.com/c/rKzsjt5X/1704-new-round-of-testing-on-the-lp-including-header-and-font-changes)

## Changes

* Changed header to the control version and adapted accordingly to the new layout
* Changed typography to match control version
* Changed colour scheme to match control version
* Adapted footer to fit the new layout

## Screenshots
For comprehensive screenshots, see the previous PR: https://github.com/guardian/support-frontend/pull/2220

Updates reflected below:
new landing page

### DIFF
--- a/support-frontend/assets/helpers/abTests/abtestDefinitions.js
+++ b/support-frontend/assets/helpers/abTests/abtestDefinitions.js
@@ -109,7 +109,7 @@ export const tests: Tests = {
     optimizeId: 'sj4_I5OAT3SJpqgnxtJ6Xg',
   },
 
-  newLandingPageTemplateTest: {
+  newLandingPageTemplateTestR2: {
     type: 'OTHER',
     variants: [
       {
--- a/support-frontend/assets/pages/contributions-landing/components/ContributionFormContainer.jsx
+++ b/support-frontend/assets/pages/contributions-landing/components/ContributionFormContainer.jsx
@@ -43,7 +43,7 @@ const mapStateToProps = (state: State) => ({
   tickerGoalReached: state.page.form.tickerGoalReached,
   isReturningContributor: state.page.user.isReturningContributor,
   countryId: state.common.internationalisation.countryId,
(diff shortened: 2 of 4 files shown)


ionamckendrick:
One thing that I think needs work is for the footer (blue box) to be further down. It's so close to the form. Would that be ok @jlieb10 ?

jlieb10:
@ionamckendrick i agree but i think let's focus on adding some padding there after we know we're going to go with this template-- it's more complicated than it looks because of those lines inside the footer and i don't think we should let it stop us from releasing the test today

jlieb10:
@tomrf1 the magic numbers can definitely be revisited in the future - i think it's acceptable to have a couple in the codebase if they are clearly labelled and no one is looking for them to represent anything except design specificity.

ionamckendrick:
@jlieb10 Sorry I didn't reply on the footer: your logic makes sense! I wasn't sure how much work it was / how fiddly. Let's revisit later.

jesseyuen:
I wonder why the landing page isnt doing great. Any hypotheses @jlieb10 @ionamckendrick? Merry Xmas!

jlieb10:
Heya @jesseyuen ! When are you coming back to iterate on it with us?!

Some hypotheses are that:
- The "G" header isn't as compelling/recognizable/reassuring as seeing the full "The Guardian" on the page (changed)
- The type was too large, distracting from the form that we ultimately needs to become the user's center of attention (changed)
- Having the form on the far left makes it secondary to the "content area" of the page on larger widths - so possibly one can feel "done" with the page when they've read what is in the content area without interacting with the form (not changed)

I think @ionamckendrick had some other ideas as well

merry xmas!

jesseyuen:
Sounds like it could be time to talk to some users @jlieb10!

PR #2280, opened 2019-12-10
Use scalatest version 3.2.0-M1 in all projects

rupertbates:
## Why are you doing this?

#2137 upgraded scalatest in support-frontend to 3.2.0-M1 but all the other projects were still using 3.0.8. 

This was causing `NoClassDefFoundError` when using the `SerialisationTestHelpers` class from `support-models` in other projects so I have upgraded all the projects to use version 3.2.0-M1

[**Trello Card**](https://trello.com/c/AAgmcmce/2809-upgrade-scalatest-in-support-frontend)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -56,7 +56,7 @@ lazy val commonSettings = Seq(
 
 lazy val commonDependencies = Seq(
   "com.typesafe" % "config" % "1.3.2",
-  "org.scalatest" %% "scalatest" % "3.0.8" % "it, test"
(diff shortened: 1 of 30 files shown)


johnduffell:
it's not clear from this or the original PR why we need to use a Milestone version of scalatest in the first place, the latest stable is 3.1.2 at the moment.  Any ideas?

johnduffell:
I only ask because the latest docs refer to 3.1 and they have moved a lot of things into different packages and/or renamed them, so it's harder to get the examples running.

PR #2285, opened 2019-12-12
Add support-logs.gutools.co.uk

jorgeazevedo:
## Why are you doing this?

Following on from Rupert's work to [get logs onto kibana](https://github.com/guardian/support-frontend/pull/2251), this PR sets up the infrastructure required to make it available at https://support-logs.gutools.co.uk. I've also moved the project to a root level dir `support-logs`.

So. Since Kibana is in the private subnet of our VPC, the traffic needs to jump through a few hoops to from our office to it.

The general idea is we have an ec2 instance running NGINX(_KibanaProxy_) proxying traffic between Kibana and Cognito, and an ALB (_ProxyLoadBalancer_) serving traffic over HTTPS using a certificate from ACM.

Additionally, we've configured NS1 to point support-logs.gutools.co.uk. to support-logs.membership-aws.gutools.co.uk. In the cloudformation here, we add a record to point support-logs.membership-aws.gutools.co.uk to the load balancer DNS name.

Finally, I've also set up [a TeamCity job](https://teamcity.gutools.co.uk/viewType.html?buildTypeId=memsub_Support_SupportLogs) so the NGINX instance can use amigo.

The NGINX instance was configured as [per aws documentation](https://aws.amazon.com/premiumsupport/knowledge-center/kibana-outside-vpc-nginx-elasticsearch).

## Screenshots

The cloudformation is so difficult to follow, a diagram would probably be useful no? Is that a common way to document it?

Cloudera_EDH_-_VP_Online


jorgeazevedo:
Thanks! Will do : -)

PR #2287, opened 2019-12-13
Elasticsearch retention policy lambda

jorgeazevedo:
## Why are you doing this?

Now that we have [serious logs](https://github.com/guardian/support-frontend/pull/2251) being streamed to kibana, it's important to have a retention policy applied to the data.

Kibana has a "turn key" solution for this called [Index Lifecycle Policies](https://www.elastic.co/guide/en/kibana/current/index-lifecycle-policies.html), which is unfortunately not available in AWS's elasticsearch service. Reason being that aws uses the open-source version of elasticsearch, and ILP is an x-pack feature available [on the Basic plan and above](https://www.elastic.co/subscriptions).

Luckily,  [amazon's documentation](https://docs.aws.amazon.com/elasticsearch-service/latest/developerguide/curator.html) sketches out a solution to this using lambda and the elasticsearch python library (specifically a package called _curator_). This PR is an implementation of that solution exactly as described in the docs. The idea is for the lambda to delete indices older than 30 days and for it to run every day.

On the cloudformation front, I've opted to use the AWS Serverless Application Model (AWS SAM) to specify the lambda which greatly simplifies the code. Still, the configuration might be a bit hard to follow. Some key ideas:

For the lambda to be able to make API calls to Kibana (in the VPC), it needs to both have a networking path to get there and appropriate "API credentials". To make this work:
- The lambda is in the VPC in the same subnets as the ProxyInstance. Its security group is locked down to only allows outbound HTTPS calls to the kibana security group.
- A policy to give lambda permissions to use the elastic search API

The serverless template has built-in functionality to set a cron rule to trigger the execution. In the background, aws is actually setting up a cloudwatch event.

It's also worth noting that when using the serverless template, we're actually creating a _lambda application_ rather than just a function. To it shows up under the Applications section. I'll add a screenshot.

On the build and deploy side, this PR includes a configure script to install dependencies, a makefile to zip up the code, a riff-raff.yaml with the deployment definition and a corresponding [TeamCity job](https://teamcity.gutools.co.uk/viewType.html?buildTypeId=memsub_Support_SupportLogsRetention).

## Screenshots

How the retention fits in with [the proxying infrastructure](https://github.com/guardian/support-frontend/pull/2285).

Cloudera_EDH_-_VP_Online

How it shows up in aws

Lambda_Management_Console

### DIFF
--- a/support-logs/retention-lambda/lambda_function.py
+++ b/support-logs/retention-lambda/lambda_function.py
@@ -24,7 +24,7 @@ def lambda_handler(event, context):
     index_list = curator.IndexList(es)
 
     # Filters by age, anything with a time stamp older than 30 days in the index name.
-    index_list.filter_by_age(source='name', direction='older', timestring='%Y.%m.%d', unit='days', unit_count=24)
+    index_list.filter_by_age(source='name', direction='older', timestring='%Y.%m.%d', unit='days', unit_count=90)
 
     print("Found %s indices to delete" % len(index_list.indices))
 
--- a/support-logs/retention-lambda/lambda_function.py
+++ b/support-logs/retention-lambda/lambda_function.py
@@ -1,0 +1,33 @@
+import boto3
+from requests_aws4auth import AWS4Auth
+from elasticsearch import Elasticsearch, RequestsHttpConnection
+import curator
+
+host = 'vpc-support-elk-domain-vcesheatvi6xspmp4sd6dnfkq4.eu-west-1.es.amazonaws.com'
+region = 'eu-west-1'
+service = 'es'
+credentials = boto3.Session().get_credentials()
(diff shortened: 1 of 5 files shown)


support-logs/retention-lambda/lambda_function.py
```diff
@@ -0,0 +1,33 @@
...
+        hosts = [{'host': host, 'port': 443}],
+        http_auth = awsauth,
+        use_ssl = True,
+        verify_certs = True,
+        connection_class = RequestsHttpConnection
+    )
+
+    index_list = curator.IndexList(es)
+
+    # Filters by age, anything with a time stamp older than 30 days in the index name.
```
rupertbates: Can we bump this to 3 months to start off with?

jorgeazevedo: I think that's OK for the logs that are currently being streamed. I set it to something short for testing purposes!

PR #2295, opened 2019-12-17
Send gift promo codes to the gift version of the GW product page

rupertbates:
## Why are you doing this?
We have a feature on the site where people can go to the url `/p/[any promo code]` and then be redirected to the product page for the promoted product. We want this endpoint to be aware of gift promotions so that if the promo code is a Guardian Weekly gift offer then the user will be redirected to the gift version of the product page.

I have also updated the promotion terms page so that for gift promotions the 'Get this offer' button goes directly to the gift version of the product page
Screen Shot 2019-12-17 at 17 03 10

[**Trello Card**](https://trello.com/c/uo6uhzY2/2770-redirect-gift-promotions-to-the-gifting-page)


alexflav23:
LGTM

PR #2297, opened 2019-12-18
Remove Krux from CSP

jfsoul:
## Why are you doing this?
Our contract with Krux is coming to an end at the end of the month and we have got the green light to remove it completely.  It has already been disabled through tag manager, this removes CSP exceptions for Krux, which will no longer be needed.

## Changes

* Remove krxd domains from CSP


alexflav23:
I've always "loved' how play implements CSP, it's really great.

PR #2310, opened 2019-12-20
Remove the reminder endpoint from this repo and add it to api-gateway

jlieb10:
## Why are you doing this?
We need to access the reminder api from www.theguardian.com & support.theguardian.com -- and decided the best place for the endpoint to be was not support.theguardian.com - it should be a more general location, so API gateway made sense

[**Trello Card**](https://trello.com/c/tv5Q7zXa/1731-remove-reminders-endpoint-from-support-frontend-and-implement-with-api-gateway)

## Changes

* Use the new api endpoint in this repo
* remove the old, now unused route, service, controller, etc.


support-frontend/assets/helpers/routes.js
```diff
@@ -38,9 +38,12 @@ const routes: {
   guardianWeeklySubscriptionLandingGift: '/subscribe/weekly/gift',
   postcodeLookup: '/postcode-lookup',
   createSignInUrl: '/identity/signin-url',
-  createReminder: '/contribute/remind-me',
+  createReminder: 'https://contribution-reminders-code.support.guardianapis.com/remind-me',
```
paulbrown1982: Will this still work @jlieb10?

jlieb10: No that needs to be cleaned up - thank you

jlieb10: using `createReminderEndpoint` (below) instead

jlieb10: (which should probably be `getReminderEndpoint` actually)

jlieb10:
```suggestion
```

PR #2317, opened 2020-01-03
Accessibility: remove 'required' fields from contribution form

tjsilver:
## Why are you doing this?

According to the new [Source Design System](https://zeroheight.com/2a1e5182b/p/35ba96) it is preferable to remove the 'required' label from form fields and display an 'optional' label where applicable instead. 'Required' fields add unnecessary noise for screen readers.

[**Trello Card**](https://trello.com/c/t9D8oECp)

## Changes

* Remove `showRequiredLabel` prop from ContributionTextInput and ContributionState components
* Change function to show 'optional' label instead of 'required' in ContributionTextInput component (it seemed pointless to do this for the ContributionState component as it's only used in one place and is required, but can implement if better practice to do so).

### Before (US example with state field)

image

### After 

image


jlieb10:
nice work @tjsilver !!

PR #2318, opened 2020-01-03
Expanded options on the post contribution reminder

jlieb10:
## Why are you doing this?
In support of our KR on reminders, we are expanding the post-contribution reminder feature.

[**Trello Card**](https://trello.com/c/2B3RQXRA/1634-update-the-post-contribution-reminder-to-offer-different-frequencies)

## Changes
Frontend: Adding form elements from support-ui, wiring them to update the request we are sending to the backend

Backend: modifying the event being sent to the lambda to be backwards compatible -- to support the two different lambdas, one of which is in code and one of which is in prod ( cc: @JoemG )

## Screenshots
set up a reminder

### DIFF
--- a/support-frontend/assets/components/contributionsReminder/contributionsReminder.jsx
+++ b/support-frontend/assets/components/contributionsReminder/contributionsReminder.jsx
@@ -3,14 +3,15 @@
 // ----- Imports ----- //
 
 import React, { Component } from 'react';
-import SvgSubscribe from 'components/svgs/subscribe';
-import SvgSubscribed from 'components/svgs/subscribed';
 import TrackableButton from 'components/button/trackableButton';
 import { trackComponentClick, trackComponentLoad } from 'helpers/tracking/behaviour';
 import { connect } from 'react-redux';
 import type { Csrf as CsrfState } from 'helpers/csrf/csrfReducer';
 import { routes } from 'helpers/routes';
 import { logException } from 'helpers/logger';
+import { Fieldset } from 'components/forms/fieldset';
+import { RadioInput } from 'components/forms/customFields/radioInput';
+import { Label as FormLabel } from 'components/forms/label';
 
 // ----- Types ----- //
 type PropTypes = {
@@ -22,6 +23,8 @@ type ButtonState = 'initial' | 'pending' | 'success' | 'fail';
 
 type StateTypes = {
   buttonState: ButtonState;
+  selectedDate: string | null;
+  selectedDateAsWord: string | null,
 }
 
 const mapStateToProps = state => ({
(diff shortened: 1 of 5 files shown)


support-frontend/assets/components/contributionsReminder/contributionsReminder.jsx
```diff
@@ -60,50 +76,59 @@ class ContributionsReminder extends Component<PropTypes, StateTypes> {
...
+            Set up a reminder to contribute again
           </h3>
           <p className="contribution-thank-you-block__message">
             {'Lots of readers choose to make single contributions at various points in the year. Opt in to receive a reminder in case you would like to support our journalism again. This will be a single email, with no obligation.'}
           </p>
 
+          <FormLabel label="Remind me in:" htmlFor={null}>
+            <Fieldset legend="Choose one of the following dates to receive your reminder:">
+              <RadioInput id="march2020" text="March 2020" name="reminder" onChange={evt => this.setDateState(evt, '2020-03-19 00:00:00', 'march2020')} />
+              <RadioInput id="june2020" text="June 2020" name="reminder" onChange={evt => this.setDateState(evt, '2020-06-01 00:00:00', 'june2020')} />
```
tjsilver: Just wondering if it would be more efficient/DRYer to use a loop or map to create the radio inputs from an array of objects?

support-frontend/assets/components/contributionsReminder/contributionsReminder.jsx
```diff
@@ -60,50 +76,58 @@ class ContributionsReminder extends Component<PropTypes, StateTypes> {
...
       default:
-        return 'Remind me in March';
+        return 'Set my reminder';
     }
   }
 
   render() {
     const { email, csrf } = this.props;
+
     const requestParams = {
```
paulbrown1982: I think you should just move this whole object down to line 142.

jlieb10: I've just inlined the object

jlieb10:
@paulbrown1982 @tjsilver done

support-frontend/assets/components/contributionsReminder/contributionsReminder.jsx
jlieb10: done

jlieb10:
@paulbrown1982 can you mark your requested change as done?

@alexflav23 is investigating why the build is failing -- will merge when the investigation is done

jlieb10:
this PR need to be modified when we put the API Gateway endpoint intro production

JoemG:
> this PR need to be modified when we put the API Gateway endpoint intro production

We can make the change in the PR that removes the endpoint

PR #2319, opened 2020-01-07
Apple Pay/PRB: fall back on state/province in the form 

tomrf1:
If Apple Pay/Payment Request button does not provide a state/province, use the one in the form (taken from user's geolocation).

Some US users are unable to pay with Apple Pay because the state is missing.

Also renamed `state` to `stateOrProvince` because state is a very overloaded term!

## Screenshots
This error message will now only be displayed if the state is also not in the form:
Screenshot 2020-01-07 at 10 21 25

### DIFF
--- a/support-frontend/assets/pages/contributions-landing/components/StripePaymentRequestButton/StripePaymentRequestButton.jsx
+++ b/support-frontend/assets/pages/contributions-landing/components/StripePaymentRequestButton/StripePaymentRequestButton.jsx
@@ -72,6 +72,7 @@ type PropTypes = {|
   otherAmounts: OtherAmounts,
   contributionType: ContributionType,
   countryGroupId: CountryGroupId,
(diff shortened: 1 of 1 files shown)


support-frontend/assets/pages/contributions-landing/components/StripePaymentRequestButton/StripePaymentRequestButton.jsx
```diff
@@ -83,7 +84,7 @@ type PropTypes = {|
   updateEmail: string => void,
   updateFirstName: string => void,
   updateLastName: string => void,
-  updateState: (state: UsState | CaState | null) => void,
+  updateStateOrProvince: (state: UsState | CaState | null) => void,
```
tjsilver: 💯 really helpful update to name!

PR #2320, opened 2020-01-07
Added random pre-selection to radio button feature

jlieb10:
## Why are you doing this?
UX came up with an additional solution for pre-selected radio buttons following the release of this feature. This PR will update the feature with the new pre-selected options.

[**Trello Card**](https://trello.com/c/Lnan3eec/1759-randomising-pre-selected-option-on-radio-buttons-on-post-contribution-reminder)

## Screenshots
random preselected radio button

### DIFF
--- a/support-frontend/assets/components/contributionsReminder/contributionsReminder.jsx
+++ b/support-frontend/assets/components/contributionsReminder/contributionsReminder.jsx
@@ -34,30 +34,26 @@ const mapStateToProps = state => ({
   email: state.page.form.formData.email,
 });
 
+const reminderDates: Array<ReminderDate> = [
+  {
+    dateName: 'March 2020',
+    timeStamp: '2020-03-19 00:00:00',
+  },
+  {
+    dateName: 'June 2020',
+    timeStamp: '2020-03-01 00:00:00',
+  },
+  {
+    dateName: 'December 2020',
+    timeStamp: '2020-12-01 00:00:00',
+  },
+];
+
+const randomIndex: number = Math.floor(Math.random() * (reminderDates.length));
+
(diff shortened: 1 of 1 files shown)


support-frontend/assets/components/contributionsReminder/contributionsReminder.jsx
```diff
@@ -31,6 +36,28 @@ const mapStateToProps = state => ({
 
 // ----- Render ----- //
 class ContributionsReminder extends Component<PropTypes, StateTypes> {
+  reminderDates: Array<ReminderDate>;
```
tomrf1: looks like this can just be defined as a const outside of the class?

support-frontend/assets/components/contributionsReminder/contributionsReminder.jsx
```diff
@@ -31,6 +36,28 @@ const mapStateToProps = state => ({
...
+      {
+        dateName: 'June 2020',
+        timeStamp: '2020-03-01 00:00:00',
+      },
+      {
+        dateName: 'December 2020',
+        timeStamp: '2020-12-01 00:00:00',
+      },
+    ];
+    this.randomIndex = Math.floor(Math.random() * (this.reminderDates.length));
```
tomrf1: does this need to happen in a constructor?

jlieb10: no, i'll refactor it to be outside the class 👍

support-frontend/assets/components/contributionsReminder/contributionsReminder.jsx
jlieb10: agreed 👍

PR #2327, opened 2020-01-09
Fix deployment of KibanaProxy

jorgeazevedo:
## Why are you doing this?

I've been investigating why KibanaProxy was running an out of date ami, and realised the riff raff yaml was misconfigured. It had `amiEncrypted` set to true, which is not the case for the `ubuntu-xenial-with-nginx`

In the process of debugging this I actually ended up simplifying the set up slightly, so I kept those changes as well.

I should also add that I added a scheduled deployment to the project so the ami doesn't get stale. [I've set it to weekly](https://riffraff.gutools.co.uk/deployment/schedule/fd7ed3b5-d93a-4c95-95ca-8f9057d0a140/edit)

### DIFF
--- a/support-logs/riff-raff.yaml
+++ b/support-logs/riff-raff.yaml
@@ -6,9 +6,7 @@ deployments:
     type: cloud-formation
     parameters:
       templatePath: cfn.yaml
-      amiParametersToTags:
-        NginxAMI:
-          BuiltBy: amigo
-          AmigoStage: PROD
-          Recipe: ubuntu-xenial-with-nginx
+      amiTags:
+        AmigoStage: PROD
+        Recipe: ubuntu-xenial-with-nginx
       amiEncrypted: true
--- a/support-logs/riff-raff.yaml
+++ b/support-logs/riff-raff.yaml
@@ -9,4 +9,3 @@ deployments:
       amiTags:
         AmigoStage: PROD
         Recipe: ubuntu-xenial-with-nginx
-      amiEncrypted: true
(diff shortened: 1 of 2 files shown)


support-logs/riff-raff.yaml
```diff
@@ -6,9 +6,6 @@ deployments:
     type: cloud-formation
     parameters:
       templatePath: cfn.yaml
-      amiParametersToTags:
-        NginxAMI:
-          BuiltBy: amigo
-          AmigoStage: PROD
-          Recipe: ubuntu-xenial-with-nginx
-      amiEncrypted: true
```
jacobwinch: I'm not sure what data is being logged to / stored on this instance, but you can request an encrypted copy via https://amigo.gutools.co.uk/recipes/ubuntu-xenial-with-nginx/edit if you'd prefer to use an encrypted volume (generally I think that's preferred if there is any personal data involved)?

jorgeazevedo: oohhh interesting! It does look quite easy to make it available. I'll give it a try

jorgeazevedo: Using encryption increases the complexity of this a bit, as it needs to be set up to use KMS. 

I'd rather go the other way and not store any data to begin with. The only thing that stores data on this box is nginx's access and error logs, which I've just pushed a commit to disable!

Fun fact: there's a lot of misleading advice on the internet that to disable nginx logs you can do `access_log off;`, when it reality, that stores the logs in a file called `off`!

PR #2329, opened 2020-01-14
Make PRB/Apple Pay SCA-compliant (A/B test)

tomrf1:
## Why are you doing this?
We have removed Stripe checkout (the old, non-SCA compliant flow) for card payments. But the Payment Request button and Apple Pay still use Stripe checkout.
This change migrates it to the new flows for single and recurring. I have implemented it as an A/B test to ensure it doesn't hurt acquisitions.

The key change is that we switch the handler from:
`paymentRequest.on('token', ...`
to:
`paymentRequest.on('paymentmethod', ...`
Tokens were part of the Stripe Checkout API, whereas PaymentMethods are part of the Payment Intent (for single) and Setup Intent (for recurring) APIs. A PaymentMethod is attached to a Payment/Setup Intent.

### Single contributions
This largely uses the existing Payment Intents logic, which sends a PaymentMethodId to our backend and handles 3D Secure if necessary.
Full flow described here: https://github.com/guardian/payment-api/pull/173

### Recurring contributions
When the user clicks 'pay':
- Request a new Setup Intent (from an endpoint we own). This returns a `clientSecret` associated with the Setup Intent.
- Use the stripe library's confirmCardSetup handler to link the Setup Intent to the PaymentMethod. This call will trigger the 3D Secure popup if required.
- The rest is handled by the [existing recurring contributions logic](https://github.com/guardian/support-frontend/pull/2159).

Note - the request for the `clientSecret` will slightly increase the time it takes to complete a recurring contribution setup. The A/B test will help determine if this matters.
I did try to make it request the `clientSecret` immediately after page load (which is what the Stripe card form does for recurring), but it's very complicated with the PRB and I don't want to add this complexity unless it's really necessary.


support-frontend/assets/helpers/abTests/abtestDefinitions.js
```diff
@@ -40,14 +41,14 @@ export const tests: Tests = {
     targetPage: usContributionsLandingPageMatch,
   },
 
-  recurringStripePaymentRequestButton: {
```
tomrf1: This old test was no longer running

support-frontend/assets/pages/contributions-landing/components/StripePaymentRequestButton/StripePaymentRequestButton.jsx
```diff
@@ -27,21 +27,21 @@ import {
...
 import type {
-  State,
+  State, Stripe3DSResult,
   StripePaymentRequestButtonData,
 } from 'pages/contributions-landing/contributionsLandingReducer';
 import type { Action } from 'pages/contributions-landing/contributionsLandingActions';
 import {
   onThirdPartyPaymentAuthorised,
-  paymentWaiting as setPaymentWaiting,
+  paymentWaiting as setPaymentWaiting, setHandleStripe3DS,
```
tjsilver: Should `setHandleStripe3DS` be on a separate line?

support-frontend/assets/helpers/abTests/abtestDefinitions.js
```diff
@@ -40,14 +41,14 @@ export const tests: Tests = {
...
+  stripePaymentRequestButtonSca: {
```
paulbrown1982: Note the PRB is disabled on PROD recurring flow. Apple Pay is still enabled.

PR #2330, opened 2020-01-14
Add kibana link to description

jorgeazevedo:
## Why are you doing this?

Kibana is now live, so we can add this here for convenience!


jorgeazevedo:
Test message, please ignore

Test_-_Chat

PR #2331, opened 2020-01-14
401 retry

rupertbates:
## Why are you doing this?
We have been getting regular 401 authentication errors from Zuora which have been causing step function executions to fail. We think that it is likely that these are temporary glitches and that a retry would succeed so this PR changes the error mapping between http errors and retry types to make a 401 response map to `RetryLimited`.

I have also increased the time between retry attempts for `RetryLimited` errors to give the systems which are experiencing problems a bit longer to recover.

[**Trello Card**](https://trello.com/c/XtArUSWE/2795-alter-retry-behaviour-of-zuora-authentication-errors)

### DIFF
--- a/support-workers/src/main/scala/com/gu/support/workers/exceptions/RetryImplicits.scala
+++ b/support-workers/src/main/scala/com/gu/support/workers/exceptions/RetryImplicits.scala
@@ -24,14 +24,15 @@ object RetryImplicits {
         new RetryUnlimited(message = e.getMessage, cause = throwable)
 
       //WebServiceClientError
-      case e @ (clientError: WebServiceClientError) =>
-        if (clientError.codeBody.code == "401"){ // Authentication clientError (401)
-          new RetryLimited(message = e.getMessage, cause = throwable)
-        } else {
-          new RetryNone(message = e.getMessage, cause = throwable)
-        }
+      case e @ (_: WebServiceClientError) if e.codeBody.code == "401" =>
+        // We are retrying on 401s now because we have been receiving this
+        // response from Zuora during maintenance windows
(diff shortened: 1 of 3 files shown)


support-workers/src/main/scala/com/gu/support/workers/exceptions/RetryImplicits.scala
```diff
@@ -23,8 +23,16 @@ object RetryImplicits {
       case e @ (_: SocketTimeoutException | _: SocketException | _: WebServiceHelperError[_]) =>
         new RetryUnlimited(message = e.getMessage, cause = throwable)
 
+      //WebServiceClientError
+      case e @ (clientError: WebServiceClientError) =>
+        if (clientError.codeBody.code == "401"){ // Authentication clientError (401)
```
jfsoul: Could you put this as a guard in the case matcher? ` case e @ ... if e.codeBody.code == "401" =>`

jfsoul: Also worth expanding briefly on why we're retying 401 now in a comment? Maybe people would find the PR anyway if they were really curious.

support-workers/src/main/scala/com/gu/support/workers/exceptions/RetryImplicits.scala
```diff
@@ -23,8 +23,16 @@ object RetryImplicits {
...
+          new RetryLimited(message = e.getMessage, cause = throwable)
+        } else {
+          new RetryNone(message = e.getMessage, cause = throwable)
+        }
+
       //Invalid Json or http client error (4xx)
```
jfsoul: Update this comment

PR #2332, opened 2020-01-14
handle Salesforce maintenance errors with RetryUnlimited behaviour

twrichards:
In light of an upcoming Salesforce maintenance window (where SF will go into ReadOnly mode), we anticipate errors when trying to create contact (during `support-workers` step function) and hence we want to treat the error code **`INSERT_UPDATE_DELETE_NOT_ALLOWED_DURING_MAINTENANCE`** as a **`RetryUnlimited`** scenario (i.e. keep retrying with existing back-off strategy until the request succeeds - at which point customer will get success confirmation email).

_Note `RetryUnlimited` will timeout after 24 hours - at which point customer will get failure email and we will be alerted (via alarm) - however we don't expect this to happen for this upcoming maintenance window._

[**Trello Card**](https://trello.com/c/ivg3G0JE)

### DIFF
--- a/support-models/src/main/scala/com/gu/salesforce/Salesforce.scala
+++ b/support-models/src/main/scala/com/gu/salesforce/Salesforce.scala
@@ -2,7 +2,6 @@ package com.gu.salesforce
 
 import cats.syntax.functor._
 import com.gu.i18n.Title
-import com.gu.salesforce.Salesforce.SalesforceErrorResponse.{expiredAuthenticationCode, rateLimitExceeded}
 import com.gu.support.encoding.Codec
 import com.gu.support.encoding.Codec.deriveCodec
 import com.gu.support.encoding.CustomCodecs._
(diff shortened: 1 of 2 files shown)


twrichards:
FYI @MahatHussein @david-pepper

support-models/src/main/scala/com/gu/salesforce/Salesforce.scala
```diff
@@ -174,13 +173,21 @@ object Salesforce {
...
   ) extends Throwable {
-    def asRetryException: RetryException = if (errorCode == expiredAuthenticationCode || errorCode == rateLimitExceeded)
+
+    val errorsToRetryUnlimited = List(
+      SalesforceErrorResponse.expiredAuthenticationCode,
+      SalesforceErrorResponse.rateLimitExceeded,
+      SalesforceErrorResponse.readOnlyMaintenance
+    )
+
+    def asRetryException: RetryException = if (errorsToRetryUnlimited.contains(errorCode))
```
jfsoul: What's the downside to (temporarily over the weekend) using retry unlimited for all errors? Increased costs running lambdas repeatedly? Anything else?

twrichards: i think this is something for subs/contributions to decide (@rupertbates @tomrf1)

twrichards:
We kinda need to merge this morning, so we have the day to observe any adverse effects before we leave it for the weekend (to do its job), so some 👍 would be appreciated...

twrichards:
✅ tested in `CODE`

PR #2333, opened 2020-01-16
Upgrade deps

alexflav23:
## Why are you doing this?

## Changes

* Bumping all dependencies to work with Node 12.14.1.
* Updating NVMRC file to work only with Node 12.14.1


jfsoul:
With regards to the non-compatibility of pasteup, I suspect the best route will be to replace that with source foundations (our design system).  We only seem to be using the colour palette so hopefully this is not too difficult a migration.

damien-shaw-guardian:
No longer needed the upgrade was completed in this PR https://github.com/guardian/support-frontend/pull/2352

PR #2335, opened 2020-01-17
Remove state encryption

rupertbates:
## Why are you doing this?

Since AWS Step functions are now encrypted at rest, we no longer need the extra encryption in the payload body.

I have already switched off encryption using the configuration option in support-frontend, this PR removes it completely. The next step will be to remove the base 64 encoding of the state as this was only needed because we were encrypting the payload (to encode invalid characters to valid json)

[**Trello Card**](https://trello.com/c/ZPylvrLf/2791-remove-encryption-from-support-workers)


support-frontend/app/services/stepfunctions/Encryption.scala
```diff
@@ -1,57 +0,0 @@
...
-  private val kms = AWSKMSClientBuilder
-    .standard()
-    .withCredentials(CredentialsProvider)
-    .withRegion(Regions.EU_WEST_1)
-    .build()
-
-  override def encrypt(data: String): Array[Byte] = {
-    val plainText = ByteBuffer.wrap(data.getBytes(utf8))
-    val req = new EncryptRequest()
-      .withKeyId(encryptionKeyId)
```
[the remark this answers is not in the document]
jfsoul: As discussed already added checklist item for this.

support-frontend/app/services/stepfunctions/StateWrapper.scala
```diff
@@ -31,11 +32,11 @@ class StateWrapper(encryption: EncryptionProvider, useEncryption: Boolean) {
       decoded <- decodeState(unwrapped.state)(decoder)
     } yield decoded
 
-  private def encodeState[T](state: T)(implicit encoder: Encoder[T]): String = encodeToBase64String(encryption.encrypt(state.asJson.noSpaces))
+  private def encodeState[T](state: T)(implicit encoder: Encoder[T]): String = encodeToBase64String(state.asJson.noSpaces.getBytes(utf8))
 
   private def decodeState[T](state: String)(implicit decoder: Decoder[T]): Try[T] = for {
     state <- Try(Base64.getDecoder.decode(state))
-    decrypted <- Try(encryption.decrypt(state))
+    decrypted <- Try(new String(state, utf8))
```
alexflav23: You can use `java.nio.charset.StandardCharsets.UTF_8` so you don't have to have literal values.

support-workers/src/main/scala/com/gu/support/workers/encoding/Conversions.scala
```diff
@@ -21,9 +21,9 @@ object Conversions {
   }
 
   implicit class FromOutputStream(val self: ByteArrayOutputStream) {
-    def toClass[T](encrypted: Boolean)(implicit decoder: Decoder[T]): T = {
+    def toClass[T](implicit decoder: Decoder[T]): T = {
       val is = self.toInputStream
-      val str = Encryption.decrypt(Streamable.bytes(is), encrypted)
+      val str = new String(Streamable.bytes(is), utf8)
```
alexflav23: Same feedback as above, we can pass in `java.nio.charset.StandardCharsets.UTF_8`.

support-workers/src/main/scala/com/gu/support/workers/encoding/Conversions.scala
```diff
@@ -21,9 +21,9 @@ object Conversions {
   }
 
   implicit class FromOutputStream(val self: ByteArrayOutputStream) {
```
alexflav23: Could be a value class, `extends AnyVal`.

support-workers/src/main/scala/com/gu/support/workers/encoding/Encoding.scala
```diff
@@ -22,7 +21,7 @@ object Encoding {
     for {
       wrapper <- unWrap(is)
       state <- Try(Base64.getDecoder.decode(wrapper.state))
-      decrypted <- Try(decrypt(state, wrapper.requestInfo.encrypted))
+      decrypted <- Try(new String(state, utf8))
```
alexflav23: Same thing with charset.

support-workers/src/main/scala/com/gu/support/workers/encoding/Wrapper.scala
```diff
@@ -31,7 +31,7 @@ object Wrapper {
     wrapString(handlerResult.value.asJson.noSpaces, handlerResult.requestInfo)
 
   def wrapString(string: String, requestInfo: RequestInfo): JsonWrapper =
-    JsonWrapper(encodeToBase64String(encrypt(string, requestInfo.encrypted)), None, requestInfo)
+    JsonWrapper(encodeToBase64String(string.getBytes(utf8)), None, requestInfo)
```
alexflav23: More charset fun.

support-frontend/app/services/stepfunctions/StateWrapper.scala
```diff
@@ -21,8 +20,10 @@ class StateWrapper(encryption: EncryptionProvider, useEncryption: Boolean) {
   implicit private val wrapperEncoder = deriveEncoder[JsonWrapper]
   implicit private val wrapperDecoder = deriveDecoder[JsonWrapper]
 
+  val utf8 = "UTF-8"
```
alexflav23: I think the easiest way to address the rest of my feedback is to replace this with: 

```scala
val utf8 = java.nio.charset.StandardCharsets.UTF_8
```

PR #2341, opened 2020-01-22
Remove Base 64 encoding and tidy up after encryption

rupertbates:
## Why are you doing this?
As a follow up to #2338 this PR removes all code related to base 64 encoding the state field in support-workers. Base 64 encoding is not longer needed because this was only in place to enable us to pass encrypted text in Json. I have also removed some left over encryption related code.

This leaves us with a slightly strange situation where the `state` field in the json passed to support-workers is stringified json rather than just a json object, but to change this is a much larger bit of work so can be done (if we decide it's worth doing) in a separate PR.

[**Trello Card**](https://trello.com/c/ZPylvrLf/2791-remove-encryption-from-support-workers)


rupertbates:
> Might be obvious but why is passing json rather than stringified json much more work?

Because of the need to support both initially (so that support-frontend and support-workers don't get out of sync during deploy) and the refactoring that would be required to go from a
plain `JsonWrapper` class containing a string to a `JsonWrapper[T]` where T is the type of contained state. 

I did start to make the changes in this PR but they got pretty extensive at which point I decided to limit the scope for simplicities sake

PR #2343, opened 2020-01-29
Updating country sent to support-workers 

JoemG:
The payment request button has a bug when used with recurring. It currently uses the country from the page(uri) instead of from the card address. This causes an error in Zuora as we can end up with invalid combinations of province/country mixing Canada and US.

This change is to send the country from the billing address of the card in the authentication response from stripe. As we send the state from the same place, in theory they will always match up in Zuora and prevent invalid combinations.

## Why are you doing this?

[**Trello Card**](https://trello.com/c/L4G6cyzc/1572-payment-requests-button-send-state-to-backend-for-us-canada)

## Screenshots

image

### DIFF
--- a/support-frontend/assets/pages/contributions-landing/components/StripePaymentRequestButton/StripePaymentRequestButton.jsx
+++ b/support-frontend/assets/pages/contributions-landing/components/StripePaymentRequestButton/StripePaymentRequestButton.jsx
@@ -63,7 +63,7 @@ import GeneralErrorMessage
   from 'components/generalErrorMessage/generalErrorMessage';
 import { getAvailablePaymentRequestButtonPaymentMethod } from 'helpers/checkouts';
 import type { StripePaymentRequestButtonScaTestVariants } from 'helpers/abTests/abtestDefinitions';
-import { fromString as countryFromString} from "../../../../helpers/internationalisation/country";
+import { fromString as countryFromString } from '../../../../helpers/internationalisation/country';
 
 // ----- Types -----//
 
@@ -198,19 +198,21 @@ function updatePayerCountry(
 ): boolean {
 
 
-  console.log("updating payer country",  countryFromCard,
+  console.log(
+    'updating payer country', countryFromCard,
     countryFromForm,
-    setCountry,);
-  if (!!countryFromCard){
-    console.log("setting country to iso")
-    const isoCountry= countryFromString(countryFromCard);
-    if (!!isoCountry) {
-      console.log("setting country")
+    setCountry,
+  );
+  if (countryFromCard) {
+    console.log('setting country to iso');
+    const isoCountry = countryFromString(countryFromCard);
+    if (isoCountry) {
+      console.log('setting country');
       setCountry(isoCountry);
       return true;
-    } else {
-      return false;
     }
+    return false;
+
   } else if (countryFromForm) {
     return true;
   }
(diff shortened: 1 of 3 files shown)


support-frontend/assets/pages/contributions-landing/contributionsLandingActions.js
```diff
@@ -351,7 +358,8 @@ const regularPaymentRequestFromAuthorisation = (
     city: null, // required go cardless field
     state: state.page.form.formData.state,
     postCode: null, // required go cardless field
-    country: state.common.internationalisation.countryId,
+    country:
+      state.page.form.formData.country ? state.page.form.formData.country : state.common.internationalisation.countryId,
```
paulbrown1982: When would these two ever be different?

paulbrown1982: In all other instances the same processing/source-selection happens for state as does country, except for here...

support-frontend/assets/pages/contributions-landing/components/StripePaymentRequestButton/StripePaymentRequestButton.jsx
```diff
@@ -184,7 +188,28 @@ function updatePayerStateOrProvince(
...
+
+  if (countryFromCard) {
+    const isoCountry = countryFromString(countryFromCard);
+    if (isoCountry) {
+      setCountry(isoCountry);
+      return true;
+    }
+    return false;
 
+  } else if (countryFromForm) {
```
tomrf1: `countryFromForm` is not optional so no need for this check

support-frontend/assets/pages/contributions-landing/components/StripePaymentRequestButton/StripePaymentRequestButton.jsx
```diff
@@ -184,7 +188,28 @@ function updatePayerStateOrProvince(
...
+  setCountry: IsoCountry => void,
+): boolean {
+
+  if (countryFromCard) {
+    const isoCountry = countryFromString(countryFromCard);
+    if (isoCountry) {
+      setCountry(isoCountry);
+      return true;
+    }
+    return false;
```
tomrf1: if `countryFromCard` is invalid then we can still use the `countryFromForm`, so I don't think it's necessary to fail.
In fact, since we always have a fallback this function can be:
```
function updatePayerCountry(
  countryFromCard?: string,
  setCountry: IsoCountry => void,
): void
```

JoemG: I am going to get rid of the function and do the update logic inline

support-frontend/assets/pages/contributions-landing/contributionsLandingActions.js
JoemG: If I am on `/ca/contribute` and use the single pay button with a card that has a US based address

`state.common.internationalisation.countryId` will be CA
`state.page.form.formData.country` will be US (taken from the Stripe auth res)

tomrf1: Currently:
`state.common.internationalisation` has country but not state
`state.page.form` has state but not country
When we POST we use both.

This branch adds a `country` field to `state.page.form`.
It may be preferable to instead update the `country` in `state.common.internationalisation`. But I'm not sure it's safe to go mutating that (there it is tied up with currency as well).
Let's discuss with @paulbrown1982 when we're all around

support-frontend/assets/pages/contributions-landing/components/StripePaymentRequestButton/StripePaymentRequestButton.jsx
tomrf1: 👍 thanks

support-frontend/assets/pages/contributions-landing/contributionsLandingActions.js
JoemG: agreed `state.page.form.formData.country` to be renamed `state.page.form.formData.billingCountry`

support-frontend/assets/pages/contributions-landing/contributionsLandingActions.js
```diff
@@ -165,6 +167,11 @@ const updateState = (state: UsState | CaState | null): ((Function) => void) =>
     dispatch(setFormSubmissionDependentValue(() => ({ type: 'UPDATE_STATE', state })));
   };
 
+const updateBillingCountry = (billingCountry: IsoCountry): ((Function) => void) =>
+  (dispatch: Function): void => {
+    dispatch(setFormSubmissionDependentValue(() => ({ type: 'UPDATE_BILLING_COUNTRY', billingCountry })));
```
tomrf1: `setFormSubmissionDependentValue` not necessary here

PR #2346, opened 2020-01-31
Paper savings

rupertbates:
## Why are you doing this?
The paper product page has various pieces of copy which show the savings a subscriber would make vs the retail price of the paper. These are currently set in code unlike the prices themselves which are stored in Zuora. This means that if we change the prices we also need to deploy a code change to update this copy.

It would be much easier if we specified the saving vs retail in Zuora as well so that we can change all price information in one place and have the site pick it up without a PR/deploy.

support thegulocal com_uk_subscribe_paper

[**Trello Card**](https://trello.com/c/cGKePEXi/2858-implement-print-price-rises-for-new-users-on-print)

### DIFF
--- a/support-frontend/assets/pages/paper-subscription-landing/components/hero/discountCopy.jsx
+++ b/support-frontend/assets/pages/paper-subscription-landing/components/hero/discountCopy.jsx
@@ -5,20 +5,20 @@ import { type Node } from 'react';
 import { flashSaleIsActive, getSaleCopy } from 'helpers/flashSale';
 import { GBPCountries } from 'helpers/internationalisation/countryGroup';
 
-type DiscountCopy = {
+export type DiscountCopy = {
   roundel: string[],
   heading: string | Node,
 };
 
-const discountCopy: DiscountCopy = {
-  roundel: ['Save up to', '37%', 'a year'],
-  heading: 'Save up to 37% a year with a subscription',
-};
+const discountCopy = (discountPercentage: number): DiscountCopy => ({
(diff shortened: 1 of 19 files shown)


support-frontend/assets/pages/paper-subscription-landing/components/hero/discountCopy.jsx
```diff
@@ -5,20 +5,27 @@ import { type Node } from 'react';
...
-  roundel: ['Save up to', '37%', 'a year'],
-  heading: 'Save up to 37% a year with a subscription',
-};
+const discountCopy = (discountPercentage: number): DiscountCopy => (
+  discountPercentage > 0 ?
+    {
+      roundel: ['Save up to', `${discountPercentage}%`, 'a year'],
+      heading: `Save up to ${discountPercentage}% a year with a subscription`,
+    } :
+    {
```
rupertbates: This default is for when the page is being 'server side rendered' during the build process because at that point the pricing information is not available.

support-frontend/assets/pages/paper-subscription-landing/components/hero/discountCopy.jsx
```diff
@@ -5,20 +5,27 @@ import { type Node } from 'react';
...
+      roundel: [],
+      heading: `Save money with a subscription`,
+    }
```
rupertbates: This default is for when the page is being 'server side rendered' during the build process because at that point the pricing information is not available.

PR #2347, opened 2020-02-04
feat: updated to node 12

damien-shaw-guardian:
feat: updated to node 12

## Why are you doing this?
AWS has deprecated node v8 and moved to node v12 as a result all the lamdas on AWS have been upgraded, we have also decide to do the same for support frontend to allow us upgrade the tool we use such webpack and preact 

## Changes

* bumped to node 12
* move all npm packages to support-frontend/support-frontend

## Screenshots
N/A


damien-shaw-guardian:
No longer need upgrade was done in this PR https://github.com/guardian/support-frontend/pull/2352

PR #2348, opened 2020-02-05
Fix react linting

damien-shaw-guardian:
## Why are you doing this?
There are a number of react linting errors which I have experienced lately, that will be present if add the new react linting rules to the project in the future. In this PR I have fixed a number of lint rules to help prepare for the future

[**Trello Card**](https://trello.com/b/TjUElW0B/subscriptions-team)

## Changes

* Fix eslint react linting rules

## Screenshots
N/A

### DIFF
--- a/support-frontend/assets/components/subscriptionCheckouts/summary.jsx
+++ b/support-frontend/assets/components/subscriptionCheckouts/summary.jsx
@@ -196,7 +196,7 @@ const ShowDropDown = (props: {
 );
 
 
-const Mobile = props => (
+const Mobile = (props: { showDropDown: boolean }) => (
   <span className={styles.mobileOnly}>
     {!props.showDropDown && <HideDropDown {...props} />}
     {props.showDropDown && <ShowDropDown {...props} />}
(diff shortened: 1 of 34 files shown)


support-frontend/assets/components/subscriptionCheckouts/summary.jsx
```diff
@@ -21,9 +22,11 @@ export type DataListItem = {
 
 type PropTypes = {
   billingPeriod: BillingPeriod,
-  changeSubscription?: string | null,
+  changeSubscription?: Option<string>,
+  // eslint-disable-next-line react/no-unused-prop-types
```
rupertbates: Are these props actually used or not?

damien-shaw-guardian: They are used, I'm sure why flow has a problem with them

rupertbates: It's lint that has the problem rather than Flow, I've just tried this and it works fine without disabling those validations for me

PR #2350, opened 2020-02-07
upgrade: dev dependencies

damien-shaw-guardian:
## Why are you doing this?

I have upgraded the seed random dev dependency because it caused the build to but this error `seedrandom.js:32 Uncaught EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content` and thus brought support-frontend down

## Changes

* upgrade seedrandom dependency

## Screenshots
N/A


damien-shaw-guardian:
No longer needed!

PR #2352, opened 2020-02-10
Upgrade Node to v12

rupertbates:
## Why are you doing this?
Node 8's End of Life is December 31, 2019 so this PR bumps our Node version to 12 and updates some dependencies to versions which work with v12


package.json
```diff
@@ -1,5 +0,0 @@
-{
-  "dependencies": {
-    "webpack-bundle-analyzer": "^3.6.0"
```
jfsoul: I think I saw contributions team using this the other day, but it feels like something we can install globally or in the project itself if we want it?

cc: @tjsilver @tomrf1

tjsilver: I discovered that it's already installed in the parent package.json: https://github.com/guardian/support-frontend/blob/master/package.json

tomrf1: we did briefly use it for comparing react vs preact. It's likely to be useful again, but can be addressed separately from this PR

rupertbates: It is already in the package.json within `/support-frontend` I think it's just a mistake that it is here as well

PR #2354, opened 2020-02-11
state/province form field not rendered for one off contributions

JoemG:
## Why are you doing this?

We currently do not send the state/province to payment API, instead we use the geo-IP header to determine the location.

As this appears to be sufficient for our reporting purposes, we should remove the field from the form as it should give increased conversion.

[**Trello Card**](https://trello.com/c/LIej1lCY/1680-remove-the-state-field-from-the-form-for-single-contributions)

## Screenshots

image


tomrf1:
Re-running the post-deploy tests, travis was failing to download stuff

PR #2357, opened 2020-02-11
removing amazon pay test

JoemG:
## Why are you doing this?

[**Trello Card**](https://trello.com/c/NU6AYP71/1733-release-amazon-pay-to-100-to-the-audience)

## Changes

* Change 1
* Change 2

### DIFF
--- a/support-frontend/assets/helpers/abTests/abtestDefinitions.js
+++ b/support-frontend/assets/helpers/abTests/abtestDefinitions.js
@@ -2,35 +2,12 @@
 import type { Tests } from './abtest';
 
 // ----- Tests ----- //
-export type AmazonPaySingleUSTestVariants = 'control' | 'amazonPay' | 'notintest';
 export type StripePaymentRequestButtonScaTestVariants = 'control' | 'sca' | 'notintest';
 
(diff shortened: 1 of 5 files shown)


support-frontend/assets/helpers/abTests/abtestDefinitions.js
```diff
@@ -2,35 +2,12 @@
...
-    audiences: {
-      ALL: {
-        offset: 0,
-        size: 1,
-      },
-    },
-    isActive: true,
-    referrerControlled: false,
-    seed: 13,
-    targetPage: usContributionsLandingPageMatch,
```
JoemG: Does removing this mean that it will now appear on all pages - I think we still only want this on the US page. Would it be more appropriate to increase the audience size?

paulbrown1982: You might need to add a country filter criteria to checkouts.js

tomrf1: We do not load amazon pay for non-US users: https://github.com/guardian/support-frontend/blob/master/support-frontend/assets/helpers/paymentIntegrations/amazonPay.js#L74

PR #2369, opened 2020-02-28
DisplayName changes

QuarpT:
## Why are you doing this?

we are deprecating identity displayName. clients will now derive display name from username and firstname

To maintain the current test user flow, the first name of a test user is propagated to the username. Then username can be used to identify if a user is a test user.

https://github.com/guardian/identity/pull/1695

## Changes

* use username to determine if test user
* derive the displayName field from identity username falling back to the first name

### DIFF
--- a/support-frontend/app/models/identity/requests/CreateGuestAccountRequestBody.scala
+++ b/support-frontend/app/models/identity/requests/CreateGuestAccountRequestBody.scala
@@ -1,7 +1,6 @@
 package models.identity.requests
 
 import akka.util.ByteString
-import com.gu.identity.model.PublicFields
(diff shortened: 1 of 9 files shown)


support-frontend/app/models/identity/requests/CreateGuestAccountRequestBody.scala
```diff
@@ -10,18 +9,13 @@ import play.api.libs.ws.{BodyWritable, InMemoryBody}
 // Example request:
 // =================
 // {
-//   "email": "a@b.com",
-//   "publicFields": {
-//     "displayName": "a"
```
QuarpT: we no longer need to send a displayName on guest user creation

PR #2370, opened 2020-02-28
Remove 'circles' variant from choice cards test

tomrf1:
## Why are you doing this?
This variant has not been doing well


jlieb10:
Just to note that if someone has already been assigned to the circle variant, and comes back to the site, they will still see circles. This is an edge case.

PR #2371, opened 2020-02-28
Add social images for subscription landing pages

jfsoul:
## Why are you doing this?
When subscription landing pages are shared on facebook/twitter etc we'd like the shared item to include a relevant image to market ourselves a bit more effectively.

While I was there, I noticed that we don't even provide twitter with a card type to use, so I've added that which should be a start for meaningful twitter shares.

I also did a little code reshuffle - moving GW routes into the GW controller which is more consistent with paper and digital.

[**Trello Card**](https://trello.com/c/nYyzs11c/2384-open-graph-tag-optimisation)

## Changes

* Add specific open graph images to product pages
* Refactor controller methods (move for consistency)

## Screenshots
image
image


jfsoul:
@rupertbates fair point about config - I'll add this to a card that I need to create anyway for improving image choices

PR #2372, opened 2020-03-02
Only send valid state codes down to Step Function

paulbrown1982:
## Why are you doing this?
We receive at around one AWS alarm per day where the value provided in the State field is incorrect, either in terms of what Zuora can understand (2-letter code or full name), and in the set that matches the provided country (e.g. providing Ontario when country is United States).

This change includes a better lookup function in `country.js` called `stateProvinceFromString` that will search for a state based on a provided string of 2-characters, the full state name, or a 3-letter code (which isn't a validly structured code in USA and CA but is in Australia). The function also ensures that it looks up in the correct set of states, respective to the provided country. If there is no match, it returns null.

I also updated the logic which determines whether a state code is required by Zuora. 
For all other instances, including Single, we will not pass on the state value unless it matches the country's set of states - this is a change from the existing logic that would pass-through whatever value was entered. We see so many unparsable or mismatched values in the DB or Zuora, that the state value not worth collecting if it does not map based on the `stateProvinceFromString` search function.

I also renamed the `updateState` function and `'UPDATE_STATE'` reducer matcher to make it clear it's about the country's state / province rather than anything React related. 
 

[**Trello Card**](https://trello.com/c/pK5SzRGy)

## Screenshots
No change to the UI other than an updated error message.

Before:
Screen Shot 2020-03-02 at 13 54 06

After:
Screen Shot 2020-03-02 at 13 53 57

### DIFF
--- a/support-frontend/assets/helpers/internationalisation/country.js
+++ b/support-frontend/assets/helpers/internationalisation/country.js
@@ -427,7 +427,8 @@ export const isInStripePaymentRequestAllowedCountries = (country: IsoCountry) =>
 // ----- Functions ----- /
 function sateFromMap(l: string, states: {[p: string]: string}): ?StateProvince {
   const s = l.toUpperCase();
-  return states[s] ? s : Object.keys(states).find(key => s.startsWith(key) || states[key] === s);
+  return states[s] ? s :
+    Object.keys(states).find(key => states[key].toUpperCase() === s || (s.length === 3 && s.startsWith(key)));
 }
 
 function usStateFromString(s: string): Option<UsState> {
--- a/support-frontend/assets/helpers/internationalisation/country.js
+++ b/support-frontend/assets/helpers/internationalisation/country.js
@@ -17,7 +17,6 @@ import {
   UnitedStates,
 } from './countryGroup';
 import type { CountryGroupId } from './countryGroup';
-import type { ContributionType } from '../contributions';
 
 // ----- Setup ----- //
 
@@ -115,12 +114,16 @@ const auStates: {
   JBT: 'Jervis Bay Territory',
 };
 
-const newspaperCountries = {
+const newspaperCountries: {
+  [string]: string,
+} = {
   GB: 'United Kingdom',
   IM: 'Isle of Man',
 };
 
-const countries = {
+const countries: {
+  [string]: string,
+} = {
   GB: 'United Kingdom',
   US: 'United States',
   AU: 'Australia',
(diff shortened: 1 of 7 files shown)


support-frontend/assets/helpers/internationalisation/country.js
```diff
@@ -421,20 +425,26 @@ export const isInStripePaymentRequestAllowedCountries = (country: IsoCountry) =>
   stripePaymentRequestAllowedCountries.includes(country);
 
 // ----- Functions ----- /
+function sateFromMap(l: string, states: {[p: string]: string}): ?StateProvince {
```
rupertbates: should be `stateFromMap`?

paulbrown1982: Whoops, yup will rename to `stateProvinceFromMap`.

support-frontend/assets/helpers/internationalisation/country.js
```diff
@@ -421,20 +425,26 @@ export const isInStripePaymentRequestAllowedCountries = (country: IsoCountry) =>
...
+  const s = l.toUpperCase();
+  return states[s] ? s :
+    Object.keys(states).find(key => states[key].toUpperCase() === s || (s.length === 3 && s.startsWith(key)));
```
jlieb10: Should we call `toUpperCase()` on `s` as well, so we know we are comparing like for like?

support-frontend/assets/pages/contributions-landing/components/StripePaymentRequestButton/StripePaymentRequestButton.jsx
```diff
@@ -263,23 +246,28 @@ function onPayment(
...
+      countryAndStateUpdateOk = true;
+    } else if (!props.stateOrProvince) {
+      logException('Missing address_state in payment request token and no state/province selected in the form');
     }
   }
 
-  const nameUpdateOk: boolean = props.stripeAccount !== 'ONE_OFF' ?
+  // If recurring, Salesforce needs a valid last name.
+  const nameUpdateOk: boolean = props.contributionType !== 'ONE_OFF' ?
     updatePayerName(paymentRequestData, props.updateFirstName, props.updateLastName) : true;
```
rupertbates:
```suggestion
  const nameUpdateOk: boolean = props.contributionType === 'ONE_OFF' || updatePayerName(paymentRequestData, props.updateFirstName, props.updateLastName);
```

paulbrown1982: I did think of doing it that way, however it then inverts the logic to the comment above. If you feel it's better I'll do it this way round for both the `...Ok` variables?

rupertbates: I think it works better with the name of the variable and is simpler, maybe tweak the comment slightly if you think it's misleading?

support-frontend/assets/pages/contributions-landing/components/StripePaymentRequestButton/StripePaymentRequestButton.jsx
```diff
@@ -260,26 +243,29 @@ function onPayment(
...
+  let countryAndStateValueOk = props.contributionType === 'ONE_OFF' || !['US', 'CA'].includes(validatedCountryFromCard);
+  if (validatedCountryFromCard) {
+    props.updateBillingCountry(validatedCountryFromCard);
+    const validatedBillingStateOrProvinceFromCard: Option<StateProvince> =
+      stateProvinceFromString(validatedCountryFromCard, billingStateOrProvinceFromCard);
+    if (validatedBillingStateOrProvinceFromCard) {
+      props.updateStateOrProvince(validatedBillingStateOrProvinceFromCard);
+      countryAndStateValueOk = true;
+    } else if (!props.stateOrProvince) {
+      logException('Missing address_state in payment request token and no state/province selected in the form');
```
tomrf1: this error will be logged even if a state/province is not required, e.g. a UK user

paulbrown1982: Ah yes, before it only logged if US and Canada. I'll raise a fix.

paulbrown1982: Fix is here: https://github.com/guardian/support-frontend/pull/2378

PR #2373, opened 2020-03-02
make sure the right version of node is running in devrun.sh

jlieb10:
## Why are you doing this?
I keep accidentally not switching the version of node before running this script. This will prevent that from happening, assuming everyone uses nvm to manage their node versions.


jlieb10:
This was reverted as it wasn't properly pulling in the source script it needs

PR #2376, opened 2020-03-02
Choice cards round 2 test

tjsilver:
## Why are you doing this?

This is round 2 of the choice cards test, where we implement only the rectangular variant and the updated design with hover state. The circular amount variant was been removed as it didn't perform as well. 

[**Trello Card**](https://trello.com/c/NaAsLE0h/1867-choice-cards-implement-round-2)

## Changes

* Update test definitions
* Update choice cards to latest design with hover state and new background colour

### Control
image

### Variant - rectangles (no hover state shown)
image

### Variant - rectangles - with hover state (£50) - desktop
image

### Variant - rectangles - with hover state (£50) - mobile
image


ionamckendrick:
This looks great now! Thanks for clarifying and adding the pics 👏

PR #2382, opened 2020-03-05
Add paypal to paper acquisition alarm

jfsoul:
## Why are you doing this?
The `no paper acqusitions` alarm went off yesterday and while investigating why that might've been we noticed that we had neglected to add paypal to the metrics behind the alarm since that was added to the paper checkout.

Note: I feel like we also need an alarm like this for GW acquisitions - I'll make a card.

## Changes

* Include paypal metric in the summed metrics for this alarm

N/A

## Screenshots
N/A


johnduffell:
nice work! 👍

jorgeazevedo:
Brilliant, thanks Jon!

PR #2387, opened 2020-03-06
New A/B test for Payment Request Button

paulbrown1982:
## Why are you doing this?
We want to see if the Payment Request Button really is worth using or not. It may be that its presence is a net detractor. Apple Pay remains fully rolled out if it's available on the browser.

## Changes

* This change sets up a new A/B test for the Payment Request Button across both checkouts. To simplify things, it also fully enables the SCA variant of the Payment Request Button as that variant is performing equally as well as the non-SCA version.

## Testing
 - [x] Locally
 - [x] CODE


paulbrown1982:
All tested on CODE, merging now.

PR #2388, opened 2020-03-09
[support-workers] fix intermittent issues with PreparePaymentMethodForReuse step

twrichards:
There were two reasons for the occasional failures with `PreparePaymentMethodForReuse` step of `support-workers`...

- Some users had missing `creditCardType` on their Zuora `PaymentMethod`s and so we were seeing Decoding failures as this was `String` rather than `Option[String]` (this field type also had to be updated on `CreditCardReferenceTransaction` as posting empty string produces a different error)
_I discovered some of these were creeping in on Amex cards updated via MMA, since the SCA changes - https://github.com/guardian/manage-frontend/pull/371 and https://github.com/guardian/membership-common/pull/606 will fix/prevent the creation of anymore of those_

- For some users their DirectDebit has Zuora payment gateway `GoCardless - Zuora Instance` rather than the main `GoCardless` Zuora payment gateway (there is little knowledge of why there are two different ones, but they point to the same GoCardless account). Fix was simply to add a `PaymentGateway` `case object` to represent it.

**I replicated both these failure scenarios and have tested the fix in `CODE`.**

### DIFF
--- a/support-models/src/main/scala/com/gu/support/zuora/api/response/Responses.scala
+++ b/support-models/src/main/scala/com/gu/support/zuora/api/response/Responses.scala
@@ -115,7 +115,7 @@ case class GetPaymentMethodDirectDebitResponse(
 case class GetPaymentMethodCardReferenceResponse(
   `type`: String,
   paymentMethodStatus: String,
-  creditCardType: String,
+  creditCardType: Option[String],
   tokenId: String,
   secondTokenId: String,
   creditCardCountry: Option[String] = None,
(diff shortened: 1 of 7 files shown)


support-models/src/main/scala/com/gu/support/zuora/api/response/Responses.scala
```diff
@@ -115,7 +115,7 @@ case class GetPaymentMethodDirectDebitResponse(
 case class GetPaymentMethodCardReferenceResponse(
   `type`: String,
   paymentMethodStatus: String,
-  creditCardType: String,
+  creditCardType: Option[String],
```
jfsoul: Payment methods can be reused fine when this is `None`?

twrichards: Correct. It's not needed in Zuora to process the payment, we just have it there for good measure.

PR #2398, opened 2020-03-13
Make step function retry on 429 rate limit responses

rupertbates:
## Why are you doing this?
Recently we had some support-workers failures caused by Zuora rate limiting [retro here](
https://docs.google.com/document/d/18SNixCPb13Ja0RDNpVfGNsZQwZVI8xGoxEHPIVlT1TA/edit)

One action from this was to change our error handling to retry (with the appropriate backoff) in the event of a rate limiting error.

[**Trello Card**](https://trello.com/c/mmvhzSkb/2899-retry-on-zuora-rate-limiting-errors)

### DIFF
--- a/support-workers/src/main/scala/com/gu/support/workers/exceptions/RetryImplicits.scala
+++ b/support-workers/src/main/scala/com/gu/support/workers/exceptions/RetryImplicits.scala
@@ -28,6 +28,12 @@ object RetryImplicits {
         // response from Zuora during maintenance windows
         new RetryLimited(message = e.getMessage, cause = throwable)
 
+      case e @ (_: WebServiceClientError) if e.codeBody.code == "429" =>
+        // We are retrying on 429 (rate limit) errors now because we have been hitting Zuora's rate limit at
+        // times when their responses are particularly slow. Retrying with a back off should ensure that these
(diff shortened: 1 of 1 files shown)


support-workers/src/main/scala/com/gu/support/workers/exceptions/RetryImplicits.scala
```diff
@@ -28,6 +28,12 @@ object RetryImplicits {
         // response from Zuora during maintenance windows
         new RetryLimited(message = e.getMessage, cause = throwable)
 
+      case e @ (_: WebServiceClientError) if e.codeBody.code == "429" =>
+        // We are retrying on 429 (rate limit) errors now because we have been hitting Zuora's rate limit at
```
johnduffell: is this code zuora specific? I notice also 401 retry seems to be for zuora errors in particular, which is probably ok as it will fail after a short delay just wanted to make sure that's intended.

rupertbates: It's not Zuora specific no because the error we get is an http error however I think it makes sense to retry any rate limit error.

PR #2396, opened 2020-03-13
Makes the AU State field mandatory on the AUD checkout

paulbrown1982:
This change makes the AU State field mandatory on the AU checkout, except for when GEO-IP determines the person to be in a country that uses AUD but is not Australia (e.g. Tuvalu, Norfolk Island, Nauru and Kir).

Another way to say this is:  If the geo-ip is anything except Tuvalu, Norfolk Island, Nauru or Kir, then the state dropdown will show on the AUD checkout will be present and will be a required field.

## Why are you doing this?
It will improve our data collection, and help for a while whilst we do the DPIA for: https://github.com/guardian/support-frontend/pull/2392

 

[**Trello Card**](https://trello.com/c/hgjWg0Yr)

## Accessibility test checklist
 - [ ] [Tested with screen reader](https://accessibility.gutools.co.uk/testing/web/screen-readers/)
 - [ ] [Navigable with keyboard](https://accessibility.gutools.co.uk/testing/web/keyboard-navigation/)
 - [ ] [Colour contrast passed](https://accessibility.gutools.co.uk/testing/web/colour-contrast/)

## Screenshots

Screen Shot 2020-03-13 at 12 41 38
---
Screen Shot 2020-03-13 at 14 06 08

PR #2402, opened 2020-03-16
Prefill AU State / Province with GEO-IP value on Recurring Contribution checkout

paulbrown1982:
This change fixes how the new compulsory AU State Field (introduced in https://github.com/guardian/support-frontend/pull/2396) gets pre-populated by GEO-IP on page load. The new solution is much simpler, not requiring the `fromCountryGroup` function that should not return AU country for AUDCountries group, because multiple countries in our model use AUD.

## Why are you doing this?
I'm making this change because there has been a small reduction in recurring contribution AV from Australia since the introduction of the field on Friday. It also makes everything consistent with the 3 checkout 'country groups' that require a state.

## Screenshots

SSB
SSA


paulbrown1982:
Tested in CODE.

PR #2404, opened 2020-03-17
Use firstname rather than displayname for test users

johnduffell:
## Why are you doing this?

Recently Identity made some changes to make it less likely people would reveal personal information as part of their profile.  The significant part of that was to make it so if we create a guest user with a displayName, the displayName is ignored and returned as "user".  This means we don't detect the user as a test user, and instead process the subscription using prod databases (which fails in support-workers).
Now we are supposed to create with the firstName set to a testUser string, and then read from the username field, which will match if it's a test user.

This PR updates the code to write both first and second/last name to identity on guest checkout.  Then the returned username is used to decide if they are a test user.  The email address is ignored, so people can feel free to use their own email address to test emails.

I have also updated the test users blurb to make it  abit more accurate based on what I know (although a bit more vague)

Further steps might be needed. I'm now not surprised we have so much trouble with support workers failure because the client and server side make separate decisions based on separate information whether it's a test user.  In the old world it was all decided server side based on the logged in user.
There may also need some checks on subs @jfsoul I think you mentioned earlier? I forgot to make a note of exactly what.

### DIFF
--- a/support-frontend/app/controllers/RegularContributions.scala
+++ b/support-frontend/app/controllers/RegularContributions.scala
@@ -130,7 +130,7 @@ class RegularContributions(
       // TODO: in a subsequent PR set these values based on the respective user.
       allowThirdPartyMail = false,
       allowGURelatedMail = false,
-      isTestUser = testUsers.isTestUser(user.privateFields.firstName)
+      isTestUser = testUsers.isTestUser(user.publicFields.username)
     )
   }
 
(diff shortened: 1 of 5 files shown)


jfsoul:
Re: testing on subs - my point was that I wondered whether this was definitely the only place where we used displayname for test user verification.

johnduffell:
@jfsoul interestingly, I have just noticed that both displayName and username is returned as the test code, where identity has detected it's a test user, even if it doesn't appear in identity. So that wouldn't end up causing an issue.  This is presumably reflected in the fact the post deploy tests for those didn't fail.
image

PR #2408, opened 2020-03-20
BUGFIX: Remove whitespace in and around a name for payment request button

jlieb10:
#1104  

## Why are you doing this?
We had an error where a first name was not getting set due to insufficient whitespace handling on the data we get when a user uses the payment request or apple pay button. This fixes that problem.

Before:
"     jean   claude   van     damme     " => `["","","","", "jean","",...etc ]`
After:
"     jean   claude   van      damme     " => `["jean","claude","van","damme"]`

### DIFF
--- a/support-frontend/assets/pages/contributions-landing/components/StripePaymentRequestButton/StripePaymentRequestButton.jsx
+++ b/support-frontend/assets/pages/contributions-landing/components/StripePaymentRequestButton/StripePaymentRequestButton.jsx
@@ -151,7 +151,8 @@ function updatePayerEmail(data: Object, setEmail: string => void) {
 }
 
 function updatePayerName(data: Object, setFirstName: string => void, setLastName: string => void): boolean {
-  const nameParts = data.payerName.split(' ');
+  // NB: This turns "    jean    claude    van    damme     " into ["jean", "claude", "van", "damme"]
+  const nameParts = data.payerName.trim().replace(/\s+/g, ' ').split(' ');
   if (nameParts.length > 2) {
     setFirstName(nameParts[0]);
     setLastName(nameParts.slice(1).join(' '));


support-frontend/assets/pages/contributions-landing/components/StripePaymentRequestButton/StripePaymentRequestButton.jsx
```diff
@@ -151,7 +151,8 @@ function updatePayerEmail(data: Object, setEmail: string => void) {
 }
 
 function updatePayerName(data: Object, setFirstName: string => void, setLastName: string => void): boolean {
-  const nameParts = data.payerName.split(' ');
+  // NB: This turns "    jean    claude    van    damme     " into ["jean", "claude", "van", "damme"]
+  const nameParts = data.payerName.trim().replace(/\s+/g, ' ').split(' ');
```
paulbrown1982: The replace() regex will catch everything from the trim, but I guess there's no harm!

jlieb10: so my thinking is that the trim will totally *remove* any whitespace before and after the name and the replace will go inside and replace multiple instance of whitespace with a single space

jlieb10: just calling the replace could leave a single space before and after the name

paulbrown1982: Ah yes sorry I missed the single space in the regex replace value, I thought it was empty-string. More tea required this morning.

PR #2392, opened 2020-03-11
Attempt to stop card testing activity by using the Google reCAPTCHA SDK

JoemG:
This change implements the Google reCAPTCHA v2 and v3 SDKs behind 2 independent feature switches.

v2 error

**v3 - stripe disabled**
image
**v2 - robot checkbox**
image

PR #2412, opened 2020-03-23
Ensure the captcha only applies to Card payment methods

paulbrown1982:
## Why are you doing this?
This change ensure the captcha (introduced in https://github.com/guardian/support-frontend/pull/2392) only applies when using the Stripe payment method.

### DIFF
--- a/support-frontend/assets/pages/contributions-landing/checkoutFormIsSubmittableActions.js
+++ b/support-frontend/assets/pages/contributions-landing/checkoutFormIsSubmittableActions.js
@@ -25,6 +25,7 @@ import {
   type Action as ContributionsLandingAction,
   setFormIsValid,
 } from './contributionsLandingActions';
+import { ExistingCard, Stripe } from 'helpers/paymentMethods';
 
 
 // ----- Types ----- //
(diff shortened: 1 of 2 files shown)


support-frontend/assets/pages/contributions-landing/checkoutFormIsSubmittableActions.js
```diff
@@ -119,6 +120,7 @@ function enableOrDisableForm() {
 
     const v2RecaptchaCheck =
       window.guardian.recaptchaV2
+      && [Stripe, ExistingCard].includes(state.page.form.paymentMethod)
```
paulbrown1982: @tomrf1 @JoemG @jlieb10 Is it better to do:

`&& state.page.form.stripeCardFormData.formComplete` ?

tomrf1: I would still check paymentMethod is `Stripe`. Not sure the `ExistingCard` check is necessary here, we don't do it elsewhere.

tomrf1:
We also need to update the rendering here https://github.com/guardian/support-frontend/commit/aa930e830f0b127af889f1772fae4aa9a0cae1b3#diff-4d3161a2ba6c0e92ab30a377cec1f42eR265
The error should not display unless `props.paymentMethod === 'Stripe'`

paulbrown1982:
Testing in CODE now.

paulbrown1982:
Tested in CODE. All payment methods work fine.

tomrf1:
great work!

PR #2413, opened 2020-03-24
Add imovo

rupertbates:
## Why are you doing this?
We want to move to using i-movo to provide a digital voucher service for our papers replacing the current physical voucher book which we post out.

To do this we have created a new product in Zuora called 'Newspaper Digital Voucher', however since we will not need to sell both the old voucher product and the new one at the same time it is possible to just update the product rate plan ids of the current `Collection` paper product and all the existing subscription creation code will switch over to using the new product.

This PR adds the new product rate plan ids for the digital voucher product alongside the old ids and uses a boolean flag to decide which to use (in this PR we continue to use the old ones).
This means that it is easy to switch between products as necessary although only one will ever be live at any time.

[**Trello Card**](https://trello.com/c/OrhmJz7N/2919-support-workers-support-for-new-product)


support-models/src/main/scala/com/gu/support/catalog/Product.scala
```diff
@@ -67,76 +69,137 @@ case object Contribution extends Product {
...
 case object Paper extends Product {
-  private lazy val ukOnly = List(CountryGroup.UK)
+
+  private def collection(productRatePlanId: ProductRatePlanId, productOptions: ProductOptions, description: String): ProductRatePlan[Paper.type] =
+    ProductRatePlan(productRatePlanId, Monthly, Collection, productOptions, description, List(CountryGroup.UK))
+
+  private def homeDelivery(productRatePlanId: ProductRatePlanId, productOptions: ProductOptions, description: String): ProductRatePlan[Paper.type] =
+    ProductRatePlan(productRatePlanId, Monthly, HomeDelivery, productOptions, description, List(CountryGroup.UK))
+
+  private lazy val useDigitalVoucher = false
```
rupertbates: This flag determines whether we use the new digital voucher product or the old paper one.

support-services/src/test/scala/com/gu/support/catalog/CatalogServiceSpec.scala
```diff
@@ -26,15 +26,15 @@ class CatalogServiceSpec extends AsyncFlatSpec with Matchers {
       Monthly,
       HomeDelivery,
       Everyday
-    ) shouldBe Some(Price(62.79, GBP))
+    ) shouldBe Some(Price(67.99, GBP))
```
jfsoul: Are these just because we haven't updated the test resources in a while?

rupertbates: Yes exactly

PR #2416, opened 2020-03-25
Add 'Support The Guardian' to the digitalSubscriptionLandingHeader

rupertbates:
## Why are you doing this?

We currently have a big First impression take over on the front that drives people directly to the digital subs landing page in UK and ROW (yay).

The issue is that there is a big, abrupt jump from the message on the banner, which is all about trust and support, and the landing page where it's very product-oriented.

As a quick fix we are going to adjust the language on the page header (the yellow thing below) to be "The digital subscription. Support The Guardian and enjoy two innovative apps and ad-free"
### Old Version
image (7)

### New Version
Screen Shot 2020-03-25 at 10 04 08


jesseyuen:
MY EYES! 😵 @lucymonie @rupertbates

PR #2417, opened 2020-03-25
Order summary: digital subscriptions

lucymonie:
## Why are you doing this?
This is the first step towards the new checkout rollout, using the design system 🎉 

Details and a link to the design are in the [Trello Card](https://trello.com/c/06UjK9kY/2935-checkouts-order-summary). The card assumes that this feature will be rolled out to all checkouts. As I've had to build a new component (because of using the design system) it was straightforward to leave the existing one in the other checkouts and just add this to digital. It should be uncomplicated to roll this out to the other checkouts from here if this is the preferred outcome.

## Changes
New order summary, per the design. 

    The text in the list that says 'So you'll pay £5.99/month for 3 months, then £11.99/month' on the desktop version doesn't make a lot of sense when read out by the screen reader. 
   This isn't highly relevant as it doesn't have input fields or buttons
    Since this uses design system components, we can assume it meets an AA standard for colour contrast.

## Screenshots: mobile

Screen Shot 2020-03-30 at 19 02 59

## Screenshots: tablet

Screen Shot 2020-03-30 at 19 03 24

## Screenshots: desktop

Screen Shot 2020-03-30 at 19 04 01


support-frontend/assets/components/subscriptionCheckouts/endSummary/endSummary.jsx
```diff
@@ -0,0 +1,26 @@
...
+import * as styles from './endSummaryStyles';
+
+const Dot = () => <div css={styles.dot} />;
+
+function EndSummary(props: {description: string}) {
+  return (
+    <ul css={styles.list}>
+      <li css={styles.listMain}><Dot />14-day free trial</li>
+      <li css={styles.listMain}>
+        <span><Dot />First 3 months half price</span>
```
rupertbates: This should copy come from the promotion not be hard coded

johnduffell:
> The text in the list that says 'So you'll pay £5.99/month for 3 months, then £11.99/month' on the desktop version doesn't make a lot of sense when read out by the screen reader.

Good work spotting that. - Is that something we can/should fix or just an observation?

support-frontend/assets/components/subscriptionCheckouts/endSummary/endSummary.jsx
lucymonie: True! Thanks Rupert

lucymonie:
> > The text in the list that says 'So you'll pay £5.99/month for 3 months, then £11.99/month' on the desktop version doesn't make a lot of sense when read out by the screen reader.
> 
> Good work spotting that. - Is that something we can/should fix or just an observation?

Yes, I think we should address this. I have mentioned it to Gus, and hope that we can have a think about our wording more broadly. Probably not something we'll fix at this stage as this wording is right across subscriptions at the moment.

PR #2427, opened 2020-04-01
Move Stripe Intent creation behind the Play backend

JoemG:
Currently Stripe Setup Intents are created by a lambda behind an api gateway.
This PR begins the process of moving the creation of Setup Intents behind endpoints in our Play backend.
This is so that we can use Recaptcha verification.

1. Removes Recaptcha V3 (we're not using it), and much of the old Recaptcha V2 logic.
2. Removes the Recaptcha A/B test.
3. Moves the Recaptcha V2 component inside the StripeCardForm component.
4. Adds the `/stripe/create-setup-intent/recaptcha` route for creating Setup Intents with a Recaptcha token. The controller behind this invokes the lambda directly and returns the Setup Intent's client_secret to the client.
5. Where recaptcha is not yet being used for recurring contributions we continue to request the Setup Intent when the user clicks 'Contribute' (from the api-gateway endpoint).

### Clicking Contribute before completing the recaptcha:
Screen Shot 2020-04-03 at 09 22 02

### Clicking Contribute after completing the recaptcha:
Screen Shot 2020-04-03 at 09 22 08


support-frontend/assets/pages/contributions-landing/checkoutFormIsSubmittableActions.js
```diff
@@ -116,25 +117,19 @@ function enableOrDisableForm() {
...
-        state.common.abParticipations.recaptchaPresenceTest === 'recaptchaPresent'
-      )
-      && !state.page.user.isPostDeploymentTestUser
-        ? state.page.form.v2IsLowRisk
-        : true;
-
     const formIsValid = getFormIsValid(formIsValidParameters(state));
     dispatch(setFormIsValid(formIsValid));
 
+    const auStripeRecurringNotVerified =
```
paulbrown1982, line 123: I wouldn't prefix this with `au` because you're already passing the countryGroup into the `recaptchaEnabled` decision function so the boolean is country agnostic.

support-frontend/assets/pages/contributions-landing/components/StripeCardForm/StripeCardForm.jsx
```diff
@@ -151,23 +192,66 @@ class CardForm extends Component<PropTypes, StateTypes> {
...
-      ).then((result) => {
-        if (result.client_secret) {
-          this.handleCardSetupForRecurring(result.client_secret);
-        } else {
-          throw new Error(`Missing client_secret field in response from ${window.guardian.stripeSetupIntentEndpoint}`);
+      if (recaptchaEnabled(this.props.countryGroupId, this.props.contributionType)) {
+        /* Recaptcha verification is required for setupIntent creation.
+        If setupIntentClientSecret is ready then complete the payment now.
+        If setupIntentClientSecret is not ready then componentDidUpdate will complete the payment when it arrives. */
+        if (this.props.setupIntentClientSecret) {
```
paulbrown1982, line 236: Why did the race condition come back into play?

tomrf1: We request the client_secret when the user completes the captcha. Recaptcha immediately displays a green tick, but it may be some time before the client_secret is returned by our backend.

We then have 2 choices:
1. disable the Contribute button until the client_secret is ready (not great UX).
2. allow the user to click Contribute, bring up the usual spinner, and when the client_secret arrives complete the payment.

We choice the 2nd option because the UX is much better.

support-frontend/assets/pages/contributions-landing/checkoutFormIsSubmittableActions.js
tomrf1: good point

PR #2437, opened 2020-04-03
DO NOT MERGE: remove unneeded paypal call from support-workers

johnduffell:
## Why are you doing this?

Recently we had a [paypal failure in support workers](https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#logsV2:logs-insights$3FqueryDetail$3D~(end~'2020-04-03T14*3a35*3a00.000Z~start~'2020-04-03T14*3a34*3a00.000Z~timeType~'ABSOLUTE~tz~'Local~editorString~'fields*20*40timestamp*2c*20*40message*0a*7c*20sort*20*40timestamp*20asc*0a~isLiveTail~false~queryId~'9cbdd4a7-18d9-4e60-9f57-de1901c193e5~source~(~'*2faws*2flambda*2fsupport-CreatePaymentMethodLambda-PROD))) because the returned paypal response didn't have an EMAIL in the response.  According to [their docs](https://developer.paypal.com/docs/archive/nvp-soap-api/merchant/ba-update-nvp/#payer-information-fields) (for the API that's deprecated 3 years+ ago) the email field is returned so I've no idea why it wasn't there.

However having had a look through, the email field has been requested since the [beginning of time](https://github.com/guardian/support-frontend/commit/d235819f6d2de121b81aac4c97979e294d30d417) and nothing actually uses it.  The only side effect would be to verify the BAID that came from the client was valid, assuming we need to?

So I have gone through and tried to remove it, and made everything compile.  I haven't tested in code or otherwise apart from the automated tests locally.

Anyone know any other reason we might want to make the call or should we just test in CODE and ship this? It would make the whole support checkout process slightly quicker for paypal, and more reliable.  This was just a 5 pounds monthly contribution, but even so complexity is bad.

@rupertbates @tomrf1 @paulbrown1982 any knowledge about this area that might help?


johnduffell:
so this is actually serialised into zuora for the subscribe call and the email is a required field https://knowledgecenter.zuora.com/BB_Introducing_Z_Business/How_Do_I_._._._%3F/How_do_I_use_the_Paypal_BAID_payment_method_in_Zuora%3F
image

PR #2440, opened 2020-04-06
Removing API Gateway from Setup Intent Lambda

JoemG:
## Why are you doing this?

We no longer want to open up the intent lambda externally, removing the API traffic and any future invocations need to be done directly using the AWS SDK.

This should prevent card testers from using our stripe api access to generate intents.

[**Trello Card**](https://trello.com/c/1nR0l7gx/1947-improve-security-of-stripe-setup-intent-creation-1)

## Changes

* Remove Gateway
* Add alarm for lambda failures


tomrf1:
Can you add a lambda alarm please, e.g. like this one https://github.com/guardian/contributions-ticker-calculator/blob/master/cloudformation.yaml#L244

support-lambdas/stripe-intent/cfn.yaml
```diff
@@ -11,31 +11,6 @@ Parameters:
       - CODE
     Default: CODE
 
-Mappings:
-  StageMap:
-    CODE:
-      DomainName: stripe-intent-code.support.guardianapis.com
-      StripeConfig: test
```
johnduffell, line 18: it would probably still need to know which stripe config to use

JoemG: good spot!

support-lambdas/stripe-intent/cfn.yaml
```diff
@@ -17,6 +17,10 @@ Mappings:
       StripeConfig: test
     PROD:
       StripeConfig: live
+  Constants:
+    Alarm:
+      Process: Follow the process in https://docs.google.com/document/d/1_3El3cly9d7u_jPgTcRjLxmdG2e919zCLvmcFCLOYAk/edit
+      Urgent: URGENT 9-5 -
```
johnduffell, line 23: it's not essential to do the join and the constants for the alarm name, this was just because we changed it a couple of times in one that had loads of alarms.  Feel free to just inline it to make it more intelligable

support-lambdas/stripe-intent/cfn.yaml
```diff
@@ -61,8 +65,15 @@ Resources:
...
+            - ' '
+            - - !FindInMap [ Constants, Alarm, Urgent ]
+              - 'PROD'
+              - 'Stripe create setup intent Lambda failing'
+        AlarmDescription: !Join
+            - ' '
+            - - 'Impact - Credit card checkout will not work for subscriptions or recurring contributions.'
+              - !FindInMap [ Constants, Alarm, Process ]
         Namespace: AWS/Lambda
         MetricName: Errors
```
johnduffell, line 78: if you are concerned about missing some types of failure e.g. stripe is down, as mentioned in chat, feel free to add an alarm on a custom metric in support-frontend.  There is an AwsCloudwatchMetricPut file I think which should let you put arbitrary metrics.
But I suspect we would return support-frontend 5xx errors if anything went wrong anyway.

support-lambdas/stripe-intent/cfn.yaml
johnduffell: oh just realised I was looking at the last commit only somehow, you didn't actually add this for the one alarm, ignore it!

support-lambdas/stripe-intent/cfn.yaml
```diff
@@ -70,84 +59,29 @@ Resources:
...
+        MetricName: Errors
+        Dimensions:
+          - Name: FunctionName
+            Value: !Ref LambdaFunctionOverHttps
+        Statistic: Sum
+        ComparisonOperator: GreaterThanThreshold
+        Threshold: 0
+        EvaluationPeriods: 1
+        Period: 60
+        TreatMissingData: "breaching"
```
johnduffell, line 87: missing data wouldn't be breaching, it would just be noone has caused an error/called the lambda I think

support-lambdas/stripe-intent/cfn.yaml
JoemG: good point, I will remove it and it will default to missing

PR #2443, opened 2020-04-06
Add iMovo tab for print landing page

lucymonie:
## Why are you doing this?
We are adding an iMovo subscription card to our print subscriptions offering. These changes make that available on the frontend. Here is the [Trello Card](https://trello.com/c/iB7TGjA6/2917-imovo-frontend-changes).

Initially when this is merged, we will continue with voucher booklet and home delivery tabs. We will switch over to iMovo when it launches in June 2020 (exact date TBC). There is some further work to the packshot for the iMovo tab that needs doing before it can be switched on.

## Changes

* Added a new tab for iMovo
* Updated the home delivery tab to new style
* Updated the voucher booklet tab to new style
* Updated checkout to handle for iMovo when the useDigitalVoucher switch is flipped to true

## Home delivery tab
support thegulocal com_uk_subscribe_paper_delivery(wide) (1)

support thegulocal com_uk_subscribe_paper_delivery(mobileMedium) (1)

## Voucher booklet tab
support thegulocal com_uk_subscribe_paper_delivery(wide)

support thegulocal com_uk_subscribe_paper_delivery(mobileMedium)

## Subscription card tab (packshot yet to be updated)
support thegulocal com_uk_subscribe_paper_delivery(wide) (2)

support thegulocal com_uk_subscribe_paper_delivery(mobileMedium) (2)


johnduffell:
looking that this, it feels like you have to make a choice up front.  I guess that is not the case, that it's flexible later? 
image

johnduffell:
home delivery tab screenshot (the first one) is inconsistent
image

support-frontend/assets/pages/digital-subscription-landing/components/digitalSubscriptionLandingHeader.jsx
```diff
@@ -214,16 +211,6 @@ function DigitalSubscriptionLandingHeader(props: PropTypes) {
         }
         <div className="digital-subscription-landing-header__cta" />
       </LeftMarginSection>
-      <HeroHanger>
-        <AnchorButton
-          aria-label="See Subscription options for Digital Subscription"
-          onClick={sendTrackingEventsOnClick('options_cta_click', 'DigitalPack', null)}
```
rupertbates, line 220: Are we losing this tracking completely?

support-frontend/assets/pages/paper-subscription-landing/components/content/deliveryTab.jsx
```diff
@@ -20,6 +18,21 @@ import {
...
 import { Collection } from 'helpers/productPrice/fulfilmentOptions';
 import { paperHasDeliveryEnabled } from 'helpers/subscriptions';
+import { Accordion, AccordionRow } from '@guardian/src-accordion';
+import { css } from '@emotion/core';
+import { neutral } from '@guardian/src-foundations/palette';
+import { textSans } from '@guardian/src-foundations/typography';
+import { space } from '@guardian/src-foundations';
+
+
+const accordionContainer = css`
```
rupertbates, line 28: Is this in the component file because it didn't seem worth moving to a separate styles file?

support-frontend/assets/pages/paper-subscription-landing/components/content/helpers.jsx
```diff
@@ -26,11 +26,13 @@ import {
...
   useDotcomContactPage,
 } from 'helpers/dotcomContactPage';
+import InfoSvg from './info.svg';
 
 // Types
 export type ContentPropTypes = {|
   selectedTab: ActiveTabState,
-  setTabAction: typeof setTab
+  setTabAction: typeof setTab,
+  useDigitalVoucher?: Option<boolean>,
```
rupertbates, line 35: Does this need to be an optional prop **and** `Option<boolean>`?

rupertbates: Also is it actually used? I can't see anywhere where it is.

support-frontend/assets/pages/paper-subscription-landing/components/content/subsCardTab.jsx
```diff
@@ -0,0 +1,95 @@
...
+);
+
+const SubscriptionCardTab = ({ getRef, setTabAction, selectedTab }: ContentTabPropTypes) => (
+  <div className="paper-subscription-landing-content__focusable" tabIndex={-1} ref={(r) => { getRef(r); }}>
+    <SubsCardFaqBlock />
+    <ContentForm selectedTab={selectedTab} setTabAction={setTabAction} title="Pick your subscription package below" />
+    <ContentHelpBlock
+      faqLink={
+        <a
+          href="https://www.theguardian.com/subscriber-direct/subscription-frequently-asked-questions"
```
rupertbates, line 79: Would be nice to move this url to routes.js as it's used all over the place but fine to leave for this pr

support-frontend/assets/pages/paper-subscription-landing/components/content/subsCardTab.jsx
```diff
@@ -0,0 +1,95 @@
...
+    </Text>
+    <Text>
+      <div css={accordionContainer}>
+        <Accordion>
+          <AccordionRow label="Collecting from multiple newsagents">
+            <p>
+              Present your card to a newsagent each time you collect the paper. The newsagent will scan your
+              card and will bee reimbursed for each transaction automatically.
+            </p>
+            <p><a href="https://imovo.org/guardianstorefinder">Find your nearest participating retailer</a></p>
```
rupertbates, line 56: maybe put the url in a const above and use in both accordion rows?

support-frontend/package.json
```diff
@@ -68,7 +68,9 @@
     "@emotion/babel-preset-css-prop": "^10.0.27",
     "@emotion/cache": "^10.0.29",
     "@emotion/core": "^10.0.27",
+    "@emotion/serialize": "^0.11.16",
```
rupertbates, line 71: Do we still need this now the SSR has been removed?

support-frontend/assets/pages/paper-subscription-landing/paperSubscriptionLandingPage.jsx
```diff
@@ -48,10 +49,12 @@ const content = (
     >
       <CampaignHeader />
       {paperHasDeliveryEnabled() &&
-        <Content needsHigherZindex>
+        <Content needsHigherZindex innerBackground="grey">
           <Text>
             <LargeParagraph>
-              We offer two different subscription types: voucher booklets and home delivery.
+              {useDigitalVoucher === null || useDigitalVoucher === false || useDigitalVoucher === undefined
```
rupertbates, line 55: can `useDigitalVoucher` ever be undefined? Doesn't it default to null?

support-frontend/assets/helpers/globals.js
```diff
@@ -25,29 +25,37 @@ const emptyAmountsSettings = {
...
-    International: [],
-    NZDCountries: [],
-    Canada: [],
-  },
-  metricUrl: '',
+const getSettings = (): Settings => {
+  const globalSettings = getGlobal('settings');
+  if (globalSettings) {
+    const useDigitalVoucher = getGlobal('useDigitalVoucher');
+    return { ...globalSettings, useDigitalVoucher };
```
johnduffell, line 32: why not just put `useDigitalVoucher` in `window.guardian.settings.useDigitalVoucher` rather than separately then having the logic to copy it over.
I'm not clear of the purpose of the default/null values below either - may be just a client side thing I don't understand though!

support-frontend/assets/pages/paper-subscription-landing/components/content/subsCardTab.jsx
```diff
@@ -0,0 +1,95 @@
...
+    <Text>
+      You can collect the newspaper from your local store or have your copies delivered by your newsagent.
+    </Text>
+    <Text>
+      <div css={accordionContainer}>
+        <Accordion>
+          <AccordionRow label="Collecting from multiple newsagents">
+            <p>
+              Present your card to a newsagent each time you collect the paper. The newsagent will scan your
+              card and will bee reimbursed for each transaction automatically.
```
johnduffell, line 54:
```suggestion
              card and will be reimbursed for each transaction automatically.
```

johnduffell: 🐝

support-frontend/assets/pages/paper-subscription-landing/components/tabs.jsx
```diff
@@ -38,27 +40,38 @@ type DispatchPropTypes = {|
 type PropTypes = {|
   ...StatePropTypes,
   ...DispatchPropTypes,
+  useDigitalVoucher: Option<boolean>,
 |};
 
+const getTabTitle = (useDigitalVoucher, k) => {
+  if (k === 'HomeDelivery' || useDigitalVoucher) {
```
johnduffell, line 47: what is `k`?

johnduffell: I see that it's "key" now but it must mean tabName or something?
I assume the exception is Voucher, so I'd prefer to see that as the exception, as it looks like Home delivery or use digital voucher is the exception
```
if (k === 'Voucher' && useDigitalVoucher) {
 return 'Voucher Booklet';
} 
```

support-frontend/assets/pages/paper-subscription-landing/paperSubscriptionLandingPage.jsx
johnduffell, line 55: can we tighten up the type of useDigitalVoucher so it is guaranteed to be false or true to avoid this kind of belt and braces?

support-models/src/main/scala/com/gu/support/catalog/Product.scala
```diff
@@ -76,7 +76,7 @@ case object Paper extends Product {
   private def homeDelivery(productRatePlanId: ProductRatePlanId, productOptions: ProductOptions, description: String): ProductRatePlan[Paper.type] =
     ProductRatePlan(productRatePlanId, Monthly, HomeDelivery, productOptions, description, List(CountryGroup.UK))
 
-  private lazy val useDigitalVoucher = false
+  lazy val useDigitalVoucher = false
```
johnduffell, line 79: doesn't need to be lazy given it's a boolean. although I realise that wasn't your change here!

support-frontend/assets/pages/paper-subscription-landing/components/content/info.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="19" height="19" viewBox="0 0 19 19" fill="none" xmlns="http://www.w3.org/2000/svg">
```
johnduffell, line 1: just wanted to check this has been through the svgo tool to make sure it's compacted as much as possible

support-frontend/assets/pages/digital-subscription-landing/components/digitalSubscriptionLandingHeader.jsx
lucymonie: The design excludes the big yellow button that used to be in this panel, so yes we won't have this tracking anymore. However, there is still tracking attached to the link clicks in the product cards, so we'll still get information about whether the use clicks through etc.

support-frontend/assets/pages/paper-subscription-landing/components/content/deliveryTab.jsx
lucymonie: TLDR; Yes
---------
If you want the long answer, it's this: if we were to separate this styling into separate files, we would have to have three separate files, one for each tab. There's a very strong principle with this sort of css that it must not be shared. In order to resist the temptation to share it, and because it was small, I have left it in the files. I know this is not your preference, and I'm happy to discuss if that would be helpful.

support-frontend/assets/helpers/globals.js
lucymonie: @rupertbates - are you able to explain this process? I don't feel confident that I would explain it correctly!

support-frontend/assets/pages/paper-subscription-landing/components/content/helpers.jsx
lucymonie: No, I don't think we need it to be an optional prop, but we will need it as the text of the switch link needs to be conditional on which collection method it is. It does need to be optional as it can be null or true.

support-frontend/assets/pages/paper-subscription-landing/components/content/info.svg
lucymonie: No it hasn't, and I didn't know about that. Tell me more!

support-frontend/assets/pages/paper-subscription-landing/components/content/subsCardTab.jsx
lucymonie: I knew my sticky 'eeeee' key would eventually get into my code... 😆

support-frontend/assets/pages/paper-subscription-landing/components/content/subsCardTab.jsx
lucymonie: Thanks - shall do

support-frontend/assets/pages/paper-subscription-landing/components/content/subsCardTab.jsx
lucymonie: Yes, good point

support-frontend/assets/pages/paper-subscription-landing/components/tabs.jsx
lucymonie: For the moment, the voucher booklet actually is the default. But this will change, so I'm torn! We will probably refactor this code in a bit once we're sure we don't need to switch vouchers back on. I guess that will be a good opportunity to get rid of some of this complexity.

You're so right about k. That has to go!

support-frontend/assets/pages/paper-subscription-landing/paperSubscriptionLandingPage.jsx
lucymonie: Yes it does (default to null). At one point it was undefined but I think that was due to another issue I was having so most likely overkill! I will remove.

lucymonie: I think we can get rid of undefined, but I think we need null and false, unfortunately. 👎

support-frontend/package.json
lucymonie: I believe we do need it, but I will check.

support-models/src/main/scala/com/gu/support/catalog/Product.scala
lucymonie: So should it be `val useDigitalVoucher = false`?

support-frontend/assets/pages/paper-subscription-landing/components/content/subsCardTab.jsx
lucymonie: I actually tried to do this, and found that it didn't work as it tried to plunk the endpoint onto support, and it needs to go to gu.co.uk. 😞

support-frontend/assets/pages/paper-subscription-landing/paperSubscriptionLandingPage.jsx
lucymonie: Actually, I got rid of it all with falsy 👍

lucymonie: `!useDigitalVoucher` seemed to cover it all off!

support-frontend/package.json
lucymonie: Just to confirm, yes we do need it.

support-frontend/assets/pages/paper-subscription-landing/components/content/subsCardTab.jsx
johnduffell: maybe just put it in a const somewhere, I guess the point is duplication although this kind of thing tends to be pretty search-and-replaceable

support-frontend/assets/helpers/globals.js
johnduffell: if there's a good reason then don't worry, I just wanted to flag it as looking a bit at odds with the rest of the code

support-frontend/assets/pages/paper-subscription-landing/components/content/info.svg
johnduffell: I think you did it before I saw this, I mean this https://jakearchibald.github.io/svgomg/

johnduffell: this could be a place for tooling for the quality okr, atomatically check in a PR to make sure images are optimised

support-frontend/assets/pages/paper-subscription-landing/components/tabs.jsx
```diff
@@ -38,27 +40,42 @@ type DispatchPropTypes = {|
...
+  useDigitalVoucher: Option<boolean>,
 |};
 
+// This is a temporary workaround while we have both iMovo and vouchers
+// We can get rid of this when we drop vouchers
+const getTabTitle = (useDigitalVoucher, fulfilmentMethod) => {
+  if (fulfilmentMethod === 'HomeDelivery' || useDigitalVoucher) {
+    return tabs[fulfilmentMethod].name;
+  }
+  return 'Voucher Booklet';
```
johnduffell, line 52: with this I guess I didn't explain my logic quite right, I just feel like if I was explaining to people 
I would say "if it's digital voucher but we're not using digital voucher, call it `Voucher Booklet` instead (other wise use the default name)"
I wouldn't say "if it's home delivery or we're using digital voucher use the default name, otherwise use `Voucher Booklet`" 
As home delivery and useDigitalVoucher feel like they're in different dimensions/unrelated to me so it's not clear why they're in the same if statement.
However, if that's not how you'd think of it, then keep it as is!  It's clearly correct logic even if not how I'd write it.

PR #2446, opened 2020-04-07
Fix broken error message in postcode lookup

lucymonie:
## Why are you doing this?
This is a fix for an error introduced by a previous PR: https://github.com/guardian/support-frontend/pull/2434

[**Trello Card**](https://trello.com/c/u9mWQrXG/2967-postcode-lookup-error-message-broken-arising-from-the-change-to-the-message-for-the-error-summary)

## Problem
image (2)

## Fix
Screen Shot 2020-04-07 at 14 45 17


johnduffell:
yeah looks good, I don't get the whole temporary covid message thing though, why not just put the real link in? Is that something to do with links not working in error messages?

PR #2448, opened 2020-04-08
Adding a CloudWatch metric to track the number of requests to the setup intent endpoints

paulbrown1982:
Adding a CloudWatch metric to track the number of requests to the setup intent endpoints. Also added a tiny bit of validation to the recaptcha version to ensure the client actually passes a token value.

## Why are you doing this?
So we can see the volume of traffic using these endpoints.

[**Trello Card**](https://trello.com/c/1nR0l7gx)

## Testing
- [x] CODE
Screen Shot 2020-04-08 at 11 37 58


paulbrown1982:
All looks good in CODE.

PR #2450, opened 2020-04-08
[DO NOT MERGE] Use recaptcha for contributions in all regions

tomrf1:
## Why are you doing this?
Currently only AU has recaptcha. We are applying this to all regions.
This applies to recurring and single contributions.

- Deletes the `/stripe/create-setup-intent` endpoint
- Does not require recaptcha for the post-deploy test user
- On the backend for recurring, if the request is for the Stripe test account then the recaptcha token is not verified

post deploy will fail until corresponding payment api branch has been deployed (we still want to deploy the clientside code first though)
https://github.com/guardian/payment-api/pull/194/files


support-frontend/app/views/contributions.scala.html
```diff
@@ -112,8 +112,6 @@
 
   window.guardian.forceCampaign = @settings.switches.experiments.get("forceCampaign").exists(_.isOn)
 
-  window.guardian.recaptchaV2 = @settings.switches.experiments.get("recaptchaV2").exists(_.isOn)
```
tomrf1, line 115: it no longer makes sense to have a switch, as recaptcha is mandatory

JoemG:
```tomrf1 requested review from JoemG and paulbrown1982 and removed request for JoemG 18 hours ago```
ah so I am not the only one to suffer from that weird UI glitch...

File: support-frontend/app/controllers/StripeController.scala [hunk 2]
```diff
@@ -72,6 +72,20 @@ class StripeController(
     }
   }
 
+  def createSetupIntentPRB: Action[SetupIntentRequest] = PrivateAction.async(circe.json[SetupIntentRequest]) { implicit request =>
+
+    val cloudwatchEvent = createSetupIntentRequest(stage, "CSRF-only")
```
paulbrown1982, line 77: `val cloudwatchEvent = createSetupIntentRequest(stage, "PRB-CSRF-only")`

support-frontend/app/controllers/StripeController.scala
```diff
@@ -72,6 +72,20 @@ class StripeController(
...
+    AwsCloudWatchMetricPut(client)(cloudwatchEvent)
+
+    stripeService(request.body.stripePublicKey).fold(
+      error => {
+        logger.error(s"Returning status InternalServerError for Create Stripe Intent request because: $error")
```
paulbrown1982: `logger.error(s"Returning status InternalServerError for Create Stripe Intent PRB request because: $error")`

support-frontend/app/controllers/StripeController.scala
```diff
@@ -45,41 +45,47 @@ class StripeController(
...
-        },
-        identity
-      )
-    } else {
-      logger.warn(s"Returning status BadRequest for Create Stripe Intent Recaptcha request because user provided no one-time-token value")
-      Future.successful(BadRequest("reCAPTCHA one-time-token required"))
-    }
+    val testPublicKeys = Set(testStripeConfig.australiaAccount.publicKey,
+      testStripeConfig.defaultAccount.publicKey,
+      testStripeConfig.unitedStatesAccount.publicKey)
```
paulbrown1982, line 53: There's no recurring United States Stripe account, so this set may end up having an empty string, so maybe you should have a `filter(_.nonEmpty)`?

tomrf1: Is that how it's currently configured? With an empty string?

JoemG: maybe we should remove `testStripeConfig.unitedStatesAccount.publicKey` from the set? 
If we aren't going to receive it from frontend then we shouldn't be checking against it in the backend?

There's no key in ssm - if it doesn't exist I think removing it from the code would be the clearer thing to do

JoemG: https://github.com/guardian/support-frontend/pull/2450/commits/9e5d6b4c201a1e459be8d485aa78c609f5189e26

PR #2451, opened 2020-04-09
Add patron info after T&Cs

tjsilver:
## Why are you doing this?

[**Trello Card**](https://trello.com/c/FGxNlXZd/1967-add-patrons-to-the-tcs-on-recurring-only)

## Changes

* Add patron info to T&Cs on recurring

### Before
image

### After (full page)
image

### UK
image

### US
image

###  AU
image

### EU
image

### INT
image

### NZ
image

### CA
image


hyunaleecasual:
Thank you so much for sharing before/ after screen! If possible, could you please share a full screen instead of the cropped versions? Any country's screen is good, I just wanted to have a glance within the entire screen. Thank you!

hyunaleecasual:
Thank you so much!! This is looking great! Thank you for your hard work!!

PR #2452, opened 2020-04-09
Revert "cope with missing stripe.elements()"

jfsoul:
Reverts guardian/support-frontend#2430


jfsoul:
> good idea assuming we got a decent answer from stripe

@johnduffell decent enough 😄 - I forwarded to subs dev the day after this happened if you're interested.

PR #2455, opened 2020-04-14
 Recaptcha for one off contributions (AU)

JoemG:
## Why are you doing this?

We want to enable recaptcha for one off contributions to prevent card testing.

Depends on https://github.com/guardian/payment-api/pull/191

[**Trello Card**](https://trello.com/c/fe75gfT8/1976-recaptcha-for-one-off-contributions)

## Changes

* Adding recaptcha to singles for AU

## Screenshots
image


tomrf1:
I've just tested what happens if the token (which we store in redux) expires after 2 mins.
The recaptcha widget automatically displays an error telling the user to try again, and it works fine after a retry.

PR #2462, opened 2020-04-17
Update scala-uri dependency

tomrf1:
Snyk is reporting a vulnerability in spray-json, a dependency of scala-uri:
`com.netaporter:scala-uri_2.12@0.4.16 › io.spray:spray-json_2.12@1.3.2`

Newer versions of scala-uri do not depend on spray-json.

### DIFF
--- a/support-frontend/app/actions/CustomActionBuilders.scala
+++ b/support-frontend/app/actions/CustomActionBuilders.scala
@@ -1,6 +1,6 @@
 package actions
 
-import com.netaporter.uri.dsl._
+import io.lemonlabs.uri.typesafe.dsl._
 import config.Configuration.IdentityUrl
 import play.api.mvc.Results._
 import play.api.mvc.Security.AuthenticatedRequest
@@ -39,11 +39,11 @@ class CustomActionBuilders(
   private val idSkipValidationReturn: (String, String) = "skipValidationReturn" -> "true"
 
   private def idWebAppRegisterUrl(path: String, clientId: String, idWebAppRegisterPath: String): String =
-    idWebAppUrl.value / idWebAppRegisterPath ?
+    (idWebAppUrl.value / idWebAppRegisterPath ?
       ("returnUrl" -> s"$supportUrl$path") &
       idSkipConfirmation &
       idSkipValidationReturn &
-      "clientId" -> clientId
+      "clientId" -> clientId).toString
 
   def onUnauthenticated(identityClientId: String): RequestHeader => Result = request => {
     SeeOther(idWebAppRegisterUrl(request.uri, identityClientId, "signin"))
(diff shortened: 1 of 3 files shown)


support-frontend/app/actions/CustomActionBuilders.scala
```diff
@@ -39,11 +39,11 @@ class CustomActionBuilders(
   private val idSkipValidationReturn: (String, String) = "skipValidationReturn" -> "true"
 
   private def idWebAppRegisterUrl(path: String, clientId: String, idWebAppRegisterPath: String): String =
-    idWebAppUrl.value / idWebAppRegisterPath ?
+    (idWebAppUrl.value / idWebAppRegisterPath ?
       ("returnUrl" -> s"$supportUrl$path") &
       idSkipConfirmation &
       idSkipValidationReturn &
-      "clientId" -> clientId
+      "clientId" -> clientId).toString
```
johnduffell, line 46: toString makes me nervous (due to no type checking as it works on anything)
I assume that's the only way though rather than a toUrl or something

support-frontend/app/actions/CustomActionBuilders.scala
```diff
@@ -39,11 +39,11 @@ class CustomActionBuilders(
   private val idSkipValidationReturn: (String, String) = "skipValidationReturn" -> "true"
 
   private def idWebAppRegisterUrl(path: String, clientId: String, idWebAppRegisterPath: String): String =
-    idWebAppUrl.value / idWebAppRegisterPath ?
+    (idWebAppUrl.value / idWebAppRegisterPath ?
```
johnduffell, line 42: check that there are no `/` in the `idWebAppRegisterPath`, as I discovered these are now escaped automatically (which is probably correct, but it's a change in behaviour)
https://github.com/guardian/membership-common/pull/608/files#diff-4ab30be5dbdbc76f24180331be757b19R8

support-frontend/app/actions/CustomActionBuilders.scala
tomrf1: I know what you mean, but that appears to be all we've got:
https://github.com/lemonlabsuk/scala-uri/blob/6b5bb3357b7d60e78514ae5664c9d4f1b46210a4/shared/src/main/scala/io/lemonlabs/uri/Uri.scala#L86

support-frontend/app/actions/CustomActionBuilders.scala
tomrf1: Oh that's nasty! Good spot.
In this case, it's only ever called with the string `"signin"`:
https://github.com/guardian/support-frontend/blob/f00be2f4a2a495fb87025e34476125357e71984e/support-frontend/app/actions/CustomActionBuilders.scala#L49
Do you think it's safe to leave it as it is?

johnduffell: in that case fine, if it works it works.  Could be worth just inlining the "signin" as no point having a parameter if it's never changed

tomrf1: good point!

support-frontend/app/actions/CustomActionBuilders.scala
johnduffell: no problem not much we can do then!

PR #2471, opened 2020-04-23
Remove Direct Debit from payment methods

rupertbates:
## Why are you doing this?

Direct Debit is currently broken


tomrf1:
we should get these into the switchboard

PR #2480, opened 2020-04-29
Introduce artificial ContributionsService test

nicl:
**WIP**

## Why are you doing this?

The aim is to indicate in Ophan/Data Lake whether a request was served locally in Frontend or by a remote epic (from the new Contributions Service).

This is a temporary solution while we validate and migrate the Contributions Service, at which point it will no longer be needed.

https://trello.com/c/VLxYd8go/115-full-variants-test-only-on-frontend

## Changes

To do this, we pass an `isRemote` flag in the query param and then convert that into a fake `ContributionsService` test.

### DIFF
--- a/support-frontend/assets/helpers/abTests/abtest.js
+++ b/support-frontend/assets/helpers/abTests/abtest.js
@@ -287,6 +287,9 @@ function getParticipations(
   const acquisitionDataTest: ?AcquisitionABTest = getTestFromAcquisitionData();
   const isRemote = getIsRemoteFromAcquisitionData();
 
+  // This is a temporary addition to help us compare remote and locally rendered
+  // epics on Frontend (as part of testing out the new Contributions Service).
+  // It will be removed once the new service is shown to be working correctly.
   if (isRemote) {
     participations['ContributionsService'] = 'remote';
   }
--- a/support-frontend/assets/helpers/abTests/abtest.js
+++ b/support-frontend/assets/helpers/abTests/abtest.js
@@ -137,6 +137,22 @@ function getParticipationsFromUrl(): ?Participations {
   return null;
 }
 
+function getIsRemoteFromAcquisitionData(): boolean {
+  const queryString = getQueryParameter('acquisitionData');
+
+  if (!queryString) {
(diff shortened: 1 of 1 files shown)


support-frontend/assets/helpers/abTests/abtest.js
```diff
@@ -137,6 +137,22 @@ function getParticipationsFromUrl(): ?Participations {
...
+function getIsRemoteFromAcquisitionData(): boolean {
+  const queryString = getQueryParameter('acquisitionData');
+
+  if (!queryString) {
+    return false;
+  }
+
+  try {
+    const data = JSON.parse(queryString);
+    return data ? data.isRemote === 'true' : false;
```
jfsoul, line 149: I guess I don't have much context on the other side of this (frontend?), but can we not model it as an A/B test in the acquisition data parameter?

nicl: @jfsoul yeah, so really this is like a meta-test that wraps a bunch of others. Ideally, we'd store `isRemote` directly in the data lake, but that requires a fair bit of work, so we are using the fact that multiple ab tests are allowed/ab test is a bit like a key-value store in the data lake instead.

An alternative would be to amend the abTest parameter in acquisitionsData to support multiple values, but that's more work and riskier so decided against given this is meant to be temporary.

(But if useful, might be easier to discuss via video!)

jfsoul: Ok, that makes sense - I think as a temporary solution this is more acceptable.

support-frontend/assets/helpers/abTests/abtest.js
```diff
@@ -269,6 +285,11 @@ function getParticipations(
   const participations: Participations = {};
 
   const acquisitionDataTest: ?AcquisitionABTest = getTestFromAcquisitionData();
+  const isRemote = getIsRemoteFromAcquisitionData();
+
+  if (isRemote) {
```
jfsoul, line 290: Maybe add a comment to say this is temporary with some explanation.

PR #2484, opened 2020-05-04
Corporate subs work - Part 1

rupertbates:
## Why are you doing this?
This is part one of the work to add corporate Digital Subscriptions to the site. It includes the back end work required to support the new type of digital sub.

[**Design Doc**](https://docs.google.com/document/d/1fd75C_getkqoz9_JErU_zl92M9lcKKWerT-zo9vP0_g/edit#heading=h.9y9a4whav5t5)
[**Trello Card**](https://trello.com/c/3wIOuGuk/3027-corporate-subs-integration-work-part-1)

## Changes

* `paymentFields` and `paymentMethod` become `Option[T]` on all state models
* New `Option[RedemptionData]` on all state models to capture the `redemptionCode` and `corporateAccountId` - these values are passed to Zuora on sub creation in custom fields
* Add a new type of `ProductOption` (`Corporate`) for use by digital subs 
* Add new product rate plans for corporate digisubs in all environments - these are differentiated from rate plans by the `Corporate` `ProductOption`
* Add a new `ReaderType` (`Corporate`) - this is passed to Zuora on sub creation in a custom field to enable us to identify this sub type in downstream systems

I have used an integration test to run a sign up for a corporate sub and checked that it makes its way into the members-data-api
Screen Shot 2020-05-04 at 16 40 42

I've also run the post deployment tests against this in CODE and they passed


support-frontend/app/services/stepfunctions/SupportWorkersClient.scala
```diff
@@ -126,6 +126,7 @@ class SupportWorkersClient(
       requestId = requestId,
       user = user,
       giftRecipient = getGiftRecipient(request.body),
+      None, // TODO
```
rupertbates, line 129: This will get picked up in part 2 of this work

support-frontend/app/controllers/RegularContributions.scala
```diff
@@ -88,9 +88,9 @@ class RegularContributions(
         InternalServerError
       },
       { statusResponse =>
-        if (!testUsers.isTestUser(request)) {
+        if (!testUsers.isTestUser(request) && body.paymentFields.isDefined) {
```
johnduffell, line 91: why not just pattern match and use the `aaa` below?

```suggestion
        body.paymentFields match { case Some(aaa) if !testUsers.isTestUser(request) =>
```

support-frontend/app/services/stepfunctions/SupportWorkersClient.scala
```diff
@@ -44,7 +44,7 @@ case class CreateSupportWorkersRequest(
   emailGiftRecipient: Option[String],
   product: ProductType,
   firstDeliveryDate: Option[LocalDate],
-  paymentFields: PaymentFields,
+  paymentFields: Option[PaymentFields],
```
johnduffell, line 47: making things optional is quite low level and doesn't necessarily explain why things are optional and in which combination - if we want more expressiveness consider rolling your own sum/product types to lock it down better.
However I realise this is a low level model as it's used for going between the lambdas, so it would complicate the serialisers or require a separate low level model.
Also theres a risk of going full membership-common although I think that was mostly confusing down to the implicit chains!
```scala
// example, this would replace `product: ProductType and paymentFields`
sealed trait TheProduct
case class Corporate extends TheProduct
sealed abstract class PaidProduct(paymentFields: PaymentFields) extends TheProduct
case class ....(paymentFields: PaymentFields) extends PaidProduct
```
See what you think - I am probably slightly in favour of keeping it low level for ease of serialisation but it depends what kind of bugs we are expecting.

rupertbates: I did consider that actually because it would have made the changes quite a bit less extensive but I felt that using an option was as explicit and a bit more standard - it's obvious to any Scala dev at a single glance that it might not be there whereas a custom model takes a little bit more investigation to understand. Also we have `PaymentFields` in the first two lambdas and `PaymentMethod` after that (when the fields have been converted into a payment method) so it would have meant doing the same thing for both classes which is starting to get a bit too elaborate for the benefit I think.

johnduffell: > obvious to any Scala dev at a single glance that it might not be there

yes I agree there, certainly a step above nulls,, but I was thinking both telling and enforcing a why/when it might not be there rather than making it easy to see that it may or may not be there.  Just to make Direct subs with no payment details a compiler error rather than a mess in zuora (in the true spirit of ADTs)
In the last week I've become slightly more in favour of elaborate types over elaborate repairs to zuora 🙈 
But happy to leave it as is, we have to balance added understanding time for everyone, over the smaller risk of something bad happening. 👍

support-frontend/app/controllers/RegularContributions.scala
```diff
@@ -88,12 +88,15 @@ class RegularContributions(
         InternalServerError
       },
       { statusResponse =>
-        if (!testUsers.isTestUser(request) && body.paymentFields.isDefined) {
-          monitoredRegion(body.billingAddress.country).map { region =>
-            val tipPath = TipPath(region, RecurringContribution, monitoredPaymentMethod(body.paymentFields.get), guestCheckout)
-            verify(tipPath, tipMonitoring.verify)
+        body.paymentFields.map(
```
johnduffell, line 91: good improvement on my previous suggestion, although since we're side effecting foreach would be clearer
```suggestion
        body.paymentFields.foreach(
```

support-models/src/main/scala/com/gu/support/catalog/ProductOptions.scala
```diff
@@ -1,44 +1,71 @@
...
+  implicit val keyDecoder: KeyDecoder[ProductOptions] = (key: String) => fromString(key, allProductOptions)
+}
+
+object DigitalProductOptions {
+  val productOptions: List[catalog.DigitalProductOptions] = List(Corporate, NoProductOptions)
+
+  implicit val decoder: Decoder[DigitalProductOptions] =
+    Decoder.decodeString.emap(code => fromString(code, productOptions).toRight(s"unrecognised product options '$code'"))
+
+  implicit val encode: Encoder[DigitalProductOptions] = Encoder.encodeString.contramap[DigitalProductOptions](_.toString)
```
johnduffell, line 58: using toString we would have to be careful about refactoring later - if we made into a case class for example, this would silently serialise to the wrong thing.  I know not all the serialisers are tested fully.
An alternative would be to give all the case objects a field for `serialisedValue` which I would usually do, but I can see why it could add boilerplate.

support-models/src/main/scala/com/gu/support/workers/ProductTypeRatePlans.scala
```diff
@@ -39,7 +39,8 @@ object ProductTypeRatePlans {
 
   implicit val digitalRatePlan: ProductTypeRatePlans[DigitalPack] = (product: DigitalPack, environment: TouchPointEnvironment, fixedTerm) =>
     product.catalogType.ratePlans.getOrElse(environment, Nil).find(productRatePlan =>
-      productRatePlan.billingPeriod == product.billingPeriod
+      (productRatePlan.billingPeriod == product.billingPeriod && productRatePlan.productOptions == product.productOptions) ||
+        (productRatePlan.productOptions == Corporate && product.productOptions == Corporate) // We don't care about the billing period for corporates
```
johnduffell, line 43: I can see the logic here if I squint far enough, but I feel it would make more sense to branch based on what we are looking for.  I also think the naming could be improved.
rename `product` to `productToFind` and `productRatePlan` to `candidateProduct`
make it
```scala
if (productToFind.productOptions == Corporate) {
  candidateProduct.productOptions == Corporate
} else {
  (productRatePlan.billingPeriod == product.billingPeriod && productRatePlan.productOptions == product.productOptions)
}
```

support-models/src/main/scala/com/gu/support/workers/redemption/RedemptionData.scala
```diff
@@ -0,0 +1,35 @@
...
+import com.gu.support.encoding.Codec
+import com.gu.support.zuora.api.{ReaderType, Subscription}
+import io.circe.syntax._
+import io.circe.{Decoder, Encoder}
+
+abstract class RedemptionData(val redemptionCode: RedemptionCode) {
+  def redeem(subscription: Subscription) = subscription.copy(redemptionCode = Some(redemptionCode))
+}
+
+case class Corporate(override val redemptionCode: RedemptionCode, corporateAccountId: CorporateAccountId) extends RedemptionData(redemptionCode) {
```
johnduffell, line 13: we also have `com.gu.support.catalog.Corporate`, consider being a bit more verbose in the class name rather than just reling on package name?

support-models/src/main/scala/com/gu/support/workers/redemption/RedemptionData.scala
```diff
@@ -0,0 +1,35 @@
...
+  override def redeem(subscription: Subscription) =
+    super.redeem(subscription)
+      .copy(
+        corporateAccountId = Some(corporateAccountId),
```
johnduffell, line 17: this and redemptionCode are new (non/indexed?) fields in zuora Subscription object? or just internal model

support-models/src/main/scala/com/gu/support/zuora/api/SubscriptionData.scala
```diff
@@ -118,6 +122,7 @@ object ReaderType {
     s match {
       case Gift.value => Gift
       case Agent.value => Agent
+      case Corporate.value => Corporate
       case _ => Direct
```
johnduffell, line 126: interesting that we assume Direct rather than failing, this would reduce the number of "can't subscribe" incidents but could cause some incorrect data in the long term

support-models/src/main/scala/com/gu/support/workers/redemption/RedemptionData.scala
johnduffell: (I can see below that they are zuora custom fields)

support-models/src/test/scala/com/gu/support/zuora/api/Fixtures.scala
```diff
@@ -269,15 +269,15 @@ object Fixtures {
...
+    SubscribeItem(account(), contactDetails, None, Some(creditCardPaymentMethod), invalidMonthlySubsData, SubscribeOptions())
   ))
 
   val incorrectPaymentMethod = SubscribeRequest(
     List(
       SubscribeItem(account(),
       contactDetails,
       None,
-      payPalPaymentMethod,
+        Some(payPalPaymentMethod),
```
johnduffell, line 280: indentation

support-services/src/main/scala/com/gu/aws/AwsCloudWatchMetricPut.scala
```diff
@@ -63,11 +63,11 @@ object AwsCloudWatchMetricPut {
         MetricDimensionName("Environment") -> MetricDimensionValue(environment.toString)
       ))
 
-  def paymentSuccessRequest(stage: Stage, paymentProvider: PaymentProvider, productType: ProductType): MetricRequest =
+  def paymentSuccessRequest(stage: Stage, paymentProvider: Option[PaymentProvider], productType: ProductType): MetricRequest =
```
johnduffell, line 66: doesn't really make sense for a paymentSuccessRequest to happen when there is no payment provider.  So it should only be called if there is a payment provider.

support-services/src/test/resources/catalog-uat.json
```diff
@@ -2821,6 +2821,83 @@
...
+          "Saving__c": null,
+          "DefaultTerm__c": "12",
+          "RatePlanType__c": "Base",
+          "PromotionCode__c": null,
+          "TrialPeriodDays__c": null,
+          "productRatePlanCharges": [
+            {
+              "id": "2c92c0f871c64cea0171dfacb1696aad",
+              "name": "Corporate Digital Subscription",
+              "type": "OneTime",
```
johnduffell, line 2842: I thought subscribers had to be covered by a charge at all times in our model?  I guess you discussed this and all the systems will handle it?  I know we don't have a fulfilment process for this so it may be different - uses members data api instead.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreatePaymentMethod.scala
```diff
@@ -20,38 +20,51 @@ class CreatePaymentMethod(servicesProvider: ServiceProvider = ServiceProvider)
...
+            requestInfo
+              .appendMessage(s"Payment method is ${paymentMethod.toFriendlyString}")
+              .appendMessage(s"Product is ${state.product.describe}")
+          ))
+    ).getOrElse(
+      Future.successful(HandlerResult(
+        getCreateSalesforceContactState(state, None),
+        requestInfo
+          .appendMessage(s"No Payment method present")
+          .appendMessage(s"Product is a corporate Digital Subscription")
```
johnduffell, line 40: > Product is a corporate Digital Subscription

is it? looks like all we checked that it's not got payment fields (back to my comments about modelling with ADTs)

johnduffell: if we start accidentally putting through other things without payment fields, due to a bug, this could confuse people looking into it.

support-workers/src/main/scala/com/gu/support/workers/lambdas/SendAcquisitionEvent.scala
```diff
@@ -66,7 +66,7 @@ class SendAcquisitionEvent(serviceProvider: ServiceProvider = ServiceProvider)
       _ => HandlerResult((), requestInfo)
     )
 
-    val cloudwatchEvent = paymentSuccessRequest(Configuration.stage, paymentProviderFromPaymentMethod(state.paymentMethod), state.product)
+    val cloudwatchEvent = paymentSuccessRequest(Configuration.stage, state.paymentMethod.map(paymentProviderFromPaymentMethod), state.product)
```
johnduffell, line 69: as per earlier, looks like it does make sense for no payment method, but it should be more like subscriptionSuccessRequest.  Ideally the metric should be renamed also, but there would be a bit more work checking the alarms as well.

johnduffell:
looks great, thanks.
my main concerns are firstly have we got enough deserialiser tests for the corporate stuff, I can see lots of test changes but not sure how many new ones.
Secondly a bigger concern is how much hangs on that innocuous looking Option, - since our "promo code becoming None" issue, I'm wondering how much it would take for subs to come through with None by mistake.  I can imagine how long it would take to deal with all the faulty subs, we'd have to put the payment data back in and turn on auto pay, and then deal with any payment failure.  if we hadn't taken payment for more than a month or so, we would then have to decide whether to write off some of the debt, and we'd need comms too.

johnduffell:
About the "option is more explicit and standard" discussion - having slept on it I feel slightly more strongly about the danger of Option, because it looks like one of the root causes of the promo code issue was us going in and making the types line up in a way that didn't make sense due to None meaning something wasn't there, but having a different reason for being missing.  Although against my point, is the fact that the Option here is in the data model, rather than in some logic.

take a look at the offending for comprehension that involves 3 similar looking optional things
https://github.com/guardian/support-frontend/blob/e2743a49d79f668c10304366754f02c67875cca7/support-workers/src/main/scala/com/gu/zuora/ProductSubscriptionBuilders.scala#L201-L204

- maybePromoCode is an `Option` and optional based on whether the user input it or not, so we want to ignore if not present.
- findPromotionCode is an `Option` and optional based on where it was found or not - this is actually an error, it should be enforced to be "one or more" ideally NonEmptyList through the types. (this didn't actually cause the issue)
- applyPromotion returns the subscription either changed or unchanged, it's optional based on whether the promotion was valid for the subscription, but it should be enforced to be error or the updated subscription. (this was actually the issue in question)

So the first option means one thing and the other two mean something different, but they have all been squashed into the first meaning because the types line up.

support-services/src/main/scala/com/gu/aws/AwsCloudWatchMetricPut.scala
rupertbates: Are you sure about this? It is called by the SendAcquisitionEvent lambda which suggests that it relates to a successful sign up rather than a payment event but I don't really know what it's used for so it's unclear whether we want to register anything here or not

support-services/src/test/resources/catalog-uat.json
rupertbates: Yes this appears to work, subs are created and picked up by members-data-api. I went with a one time charge rather than recurring as it seemed wasteful to bill £0 on a recurring basis

johnduffell: check the data lake too, I think the raw->clean subscriptions from zuora job might make that assumption.  I know I had a fun time rewriting that job when GDPR came in (retention issues)

support-services/src/main/scala/com/gu/aws/AwsCloudWatchMetricPut.scala
johnduffell: no I realised later and made another comment and forgot to come back and edit this one

> as per earlier, looks like it does make sense for no payment method, but it should be more like subscriptionSuccessRequest. Ideally the metric should be renamed also, but there would be a bit more work checking the alarms as well.

so it's more like a rename needed to make it reflect what it actually does.  I know refactoring inherently has a risk, but it also has benefits for future clarity.

support-services/src/test/resources/catalog-uat.json
johnduffell: it looks like as long as the effective end date of the charge is populated, it will work OK.  The exception is Friend, which was a one day charge, which ends when the next one starts.
https://github.com/guardian/ophan-data-lake/blob/418ad335c5dce925a1632ca07237b79b3483f485/etl/subscriptions/src/main/scala/ophan/etl/subscriptions/JoinedToClean.scala#L135
So it may be a new exception if we want to model it like friend, or we may want to make it a charge that renews with the sub, even if it's a one time charge.

The job I linked above is due to be rewritted to use the clean zuora incremental tables, and the friend special case may be removed as I don't know if friends still exist.

johnduffell: you can see the rate plan name and product name in the DL are by the following logic
https://github.com/guardian/ophan-data-lake/blob/418ad335c5dce925a1632ca07237b79b3483f485/etl/subscriptions/src/main/scala/ophan/etl/subscriptions/JoinedToClean.scala#L117-L122
First find one within the effective dates including special case for friend
failing that find the one started most recently (this was a last attempt because there were lots of GW migrated in with charges that were messed up - however this should find the corporate one time charges)
once that process has finished, find the most expensive of the ones in the list (there would only be one for corporate and it would be free therefore the most expensive)

support-services/src/main/scala/com/gu/aws/AwsCloudWatchMetricPut.scala
rupertbates: I'm going to leave it out of this PR as the knock on effects could be quite extensive

johnduffell: good thinking better as a separate PR 👍

support-models/src/main/scala/com/gu/support/catalog/ProductOptions.scala
rupertbates: This is the prevailing pattern in this code base although I've noticed that `ReaderType` does provide a serialisation value so I'm going to leave it like this for now. If we think there is sufficient value to changing this across the board then we should do that in a separate PR

support-models/src/main/scala/com/gu/support/workers/states/PaymentDetails.scala
```diff
@@ -0,0 +1,63 @@
...
+
+  override def isEmpty = true
+}
+
+object PaymentDetails {
+  implicit def paidProductEncoder[T: Encoder]: Encoder[PaidProduct[T]] = deriveEncoder[PaidProduct[T]]
+    .mapJson { json =>
+      val jsonObject = json.asObject.getOrElse(JsonObject.empty)
+      val value = jsonObject("value")
+      value.getOrElse(Json.Null)
```
johnduffell, line 41: minor but I'd prob pattern match first rather than encoding to json the default way and looking inside, because the contents of the string "value" here is not checked by the compiler

support-frontend/app/services/stepfunctions/SupportWorkersClient.scala
```diff
@@ -128,7 +128,7 @@ class SupportWorkersClient(
       giftRecipient = getGiftRecipient(request.body),
       None, // TODO
       product = request.body.product,
-      paymentFields = request.body.paymentFields,
+      paymentFields = request.body.paymentFields.map(PaidProduct(_)).getOrElse(FreeProduct),
```
johnduffell, line 131: I think we might have already discussed this, but since the types can't ensure that the sub is a corporate digital pack at this stage, we need a clear validation at this point.  i could see some existing separate validation but I think it might need a new validator.
The existing validator has actually suffered the "option" issue again in that we only validate that the individual fields are present if the whole are present, but not the conditions that the whole should be present https://github.com/guardian/support-frontend/pull/2484/files#diff-a7bab7639dad7567439a537a17cfc4a8R22

support-models/src/main/scala/com/gu/support/workers/states/FailureHandlerState.scala
```diff
@@ -13,8 +13,8 @@ case class FailureHandlerState(
   giftRecipient: Option[GiftRecipient],
   redemptionData: Option[RedemptionData],
   product: ProductType,
-  paymentFields: Option[PaymentFields], //Will be present if CreatePaymentMethod failed
-  paymentMethod: Option[PaymentMethod], //Will be present if anything after CreatePaymentMethod failed
+  paymentFields: PaymentDetails[PaymentFields], //Will be present if CreatePaymentMethod failed
+  paymentMethod: PaymentDetails[PaymentMethod], //Will be present if anything after CreatePaymentMethod failed
```
johnduffell, line 17: I thihnk these are technically `Option[PaymentDetails[PaymentFields]]` and `Option[PaymentDetails[PaymentMethod]]` as the two optionalities mean different things

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreatePaymentMethod.scala
```diff
@@ -22,24 +22,23 @@ class CreatePaymentMethod(servicesProvider: ServiceProvider = ServiceProvider)
...
     SafeLogger.debug(s"CreatePaymentMethod state: $state")
-    state.paymentFields.map(
-      paymentFields =>
-        createPaymentMethod(paymentFields, state.user, state.product.currency, services)
+
+    state.paymentFields match {
+      case FreeProduct => Future.successful(HandlerResult(
+        getCreateSalesforceContactState(state, FreeProduct),
+        requestInfo
+          .appendMessage(s"No Payment method present")
```
johnduffell, line 30:
```suggestion
          .appendMessage(s"Free product has no payment method")
```

support-models/src/main/scala/com/gu/support/workers/states/FailureHandlerState.scala
rupertbates: Yes! Well spotted.

support-models/src/main/scala/com/gu/support/workers/ProductTypeRatePlans.scala
rupertbates: I'm going to leave this for now because the changes would be quite extensive - there are similar extension methods for the other product types as well.

johnduffell: ok didn't realise it's a bigger pattern, makes sense to pick it up when the time is better 👍

support-frontend/app/services/stepfunctions/SupportWorkersClient.scala
rupertbates: Have refactored the validators to take account of the new type. There is still some work to do here in phase 2 but not for this PR

support-models/src/main/scala/com/gu/support/zuora/api/SubscriptionData.scala
rupertbates: We actually never deserialise this in the app, we only serialise to send to Zuora but I've added a new `Unknown` value to be doubly safe.

johnduffell: consider just deleting the deserialiser if it's unused at present

support-frontend/app/utils/CheckoutValidationRules.scala
```diff
@@ -79,11 +94,18 @@ object DigitalPackValidation {
     import createSupportWorkersRequest.product._
     import createSupportWorkersRequest.billingAddress._
 
+    def isCorporateSub(request: CreateSupportWorkersRequest) =
+      request.paymentFields == FreeProduct //TODO: also need a valid code, this is coming in part 2
```
johnduffell, line 98: got it, we could also check product is digitalpack (or read from dynamo and match it up if we want to be more flexible again in part 2)

johnduffell: oh sorry, did't notice this is specifically the digital pack validator! ignore

PR #2486, opened 2020-05-05
Remove the unused `allowFixedTermSubs` flag

rupertbates:
## Why are you doing this?
This is a hangover from when we were releasing GW gifting, it is no longer needed


support-workers/src/test/scala/com/gu/zuora/ProductSubscriptionBuildersSpec.scala
```diff
@@ -94,10 +94,4 @@ class ProductSubscriptionBuildersSpec extends AnyFlatSpec with Matchers with Pro
     isTestUser = false,
     contractEffectiveDate = saleDate)
 
-  def checkAllowed(assertion: => Assertion) = {
```
johnduffell, line 97: I was confused when I saw it above - it's gone now but probably should have called this `onlyIfFixedTermAllowed`

PR #2490, opened 2020-05-05
Digital subscriptions mobile checkout tweaks

lucymonie:
## Why are you doing this?
These are some updates to the mobile digital checkout order summary and end summary.

[**Trello Card**](https://trello.com/c/J8FaPxdm/3028-mobile-checkouts-tweaks)

## Changes
* Update font sizes in order summary
* Remove total line
* Reverse order of first two bullet points in end summary
* Remove stray empty box element from checkout (desktop view)
* This is a bit sneaky but I fixed the out-of-alignment FAQs in the footer for this page

## Screenshots
support thegulocal com_subscribe_digital_checkout_promoCode=DK0NT24WG(mobileMedium) (1)

support thegulocal com_subscribe_digital_checkout_promoCode=DK0NT24WG(wide) (5)


support-frontend/assets/pages/digital-subscription-checkout/components/digitalPaymentTerms.jsx
```diff
@@ -4,19 +4,18 @@ import { FormSection } from 'components/checkoutForm/checkoutForm';
...
   paymentMethod: Option<PaymentMethod>,
 }) {
   return (
-    <FormSection>
-      <div className={styles.endSummay}>
-        <EndSummary />
-      </div>
-      {(props.paymentMethod === DirectDebit) && <DirectDebitTerms />}
-    </FormSection>
+    <span>
```
rupertbates, line 13: Do we want to return an empty span here for non-DD payment methods?

lucymonie: You do have to return jsx or null. I guess I could do a ternary and return null otherwise?

rupertbates: Would be cleaner I think

GHaberis: Could you move the conditional out of this component into the parent:

```jsx
{props.paymentMethod === DirectDebit && <DigitalPaymentTerms />}
```

You could then simplify this `DigitalPaymentTerms` component by removing the `paymentMethod` prop, and you could also remove the wrapping `<span>` ?

lucymonie: That would be nicer all other things being equal, but the parent is currently more of a composer of he pieces rather than a manager of presentation logic, so I'd slightly prefer to leave it in the sub-component.

support-frontend/assets/pages/digital-subscription-checkout/components/endSummary/endSummary.jsx
```diff
@@ -10,21 +10,25 @@ const Dot = () => <div css={styles.dot} />;
...
-      </li>
       <li>
         <Dot /><div css={styles.listMain}>You can cancel any time</div>
       </li>
     </ul>
   );
 }
 
+EndSummary.defaultProps = {
+  endSummaryMobile: false,
```
GHaberis, line 31: Is this `endSummaryMobile` prop necessary?

lucymonie: No! Great pickup

support-frontend/assets/pages/digital-subscription-checkout/components/endSummary/endSummarySelector.js
```diff
@@ -10,6 +10,7 @@ import { getBillingDescription } from 'helpers/productPrice/priceDescriptionsDig
 export type EndSummaryProps = {
   priceDescription: string,
   promotion: Option<?string>,
+  endSummaryMobile?: boolean,
```
GHaberis, line 13: I couldn't see this new prop `endSummaryMobile` in use, is it required?

lucymonie: No - I added it for the previous solution I had to this problem. I'll remove it. 👍

PR #2495, opened 2020-05-12
Fix Weekly product cards

lucymonie:
## Why are you doing this?
The Guardian Weekly product cards were broken by the changes made to the print product page. This work was to reinstate the GW styles on that page.

[**Trello Card**](https://trello.com/c/wpZqQyIo/3054-fix-gw-product-cards)

## Changes
* Apply a theme to the product cards
* Make a new version of the productPlanPageFormLabel component so there can be two distinct versions of this card.

## Screenshots (broken)
Screen Shot 2020-05-12 at 11 41 07

## Screenshots (fixed)
support thegulocal com_uk_subscribe_weekly(wide)

support thegulocal com_uk_subscribe_paper(wide) (4)


support-frontend/assets/components/productPage/productPagePlanForm/productPagePlanForm.jsx
```diff
@@ -39,24 +41,39 @@ export default function ProductPagePlanForm<P:string>({
...
-                  footer={((price || saving) ?
-                    <ProductPagePlanFormPrice title={price} copy={saving} />
-                    : null
-                  )}
-                >
-                  {copy}
-                </ProductPagePlanFormLabel>
+              <div className={`component-product-page-plan-form__item--${theme}`}>
+                {theme === 'light' && (
+                  <ProductPagePlanFormLabel
```
GHaberis, line 52: `ProductPagePlanFormLabel` (light theme) and `ProductPagePlanFormLabelDark` (dark theme) seem very similar, they take the same `props` and `children`, also their `jsx` markup and styles do not look different - other than the use of a darker colour palette for in `ProductPagePlanFormLabelDark` styles. It feels like there's an opportunity to remove the duplication here and have a single component where you pass the theme as a prop and just apply a modifier `--dark` or `--light` modifier to the component's elements' classNames depending on the value of the theme?

lucymonie: Yes, on the one hand you are totally correct about the duplication. I guess my thinking was that by keeping them separate, it's easier to get rid of the old (dark) theme when it is phased out, which I believe it will be, rather than having a snaggled mess of css.

GHaberis: Ah I see! I didn't realise removing the dark theme was on the horizon, ok in that case it's a sensible decision 👍

lucymonie: However, I am checking this assumption with Cloe before merging based on this assumption!

lucymonie: Cloe says that is correct 👍

PR #2504, opened 2020-05-15
Make payment and redemption details into an `Either`

rupertbates:
## Why are you doing this?

This is a refactor of #2484 to change the way that redemptions and payment data is modelled.

In the first model payment fields were modelled as a `PaymentDetails` type which could be either a `PaidProduct` or a `FreeProduct` which was used for redemptions.
There was then a separate Option[RedemptionData] field for the redemption details

This is much better modelled as an Either[PaymentFields, RedemptionData] because you must always have exactly one of these types present. The previous model allowed for invalid states where you could have a `FreeProduct` with `None` for the redemption data or a `PaidProduct` which also has redemption data.

Also `FreeProduct` doesn't really describe redemptions accurately because they are not free they're just paid for by someone other than the redeemer.


support-frontend/app/utils/CheckoutValidationRules.scala
```diff
@@ -34,13 +34,13 @@ object PaidProductValidation {
...
 
-  def hasValidPaymentDetailsForPaidProduct(paymentDetails: PaymentDetails[PaymentFields]): Boolean =
-    paymentDetails match {
-      case PaidProduct(paymentFields) => noEmptyPaymentFields(paymentFields)
-      case FreeProduct => false
-    }
+  def hasValidPaymentDetailsForPaidProduct(paymentDetails: Either[PaymentFields, RedemptionData]): Boolean =
+    paymentDetails fold (
+      paymentFields => noEmptyPaymentFields(paymentFields),
+      _ => false
```
johnduffell, line 40: now this line is less clear, from the source in Github.  would a pattern match help? or even just name the `_` assuming scalacheck allows that.

johnduffell: or even a `// comment`, although again not checked by the compiler

support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
```diff
@@ -43,30 +43,30 @@ case class DigitalPackEmailFields(
...
       "Account Name" -> dd.bankTransferAccountName,
       "Default payment method" -> "Direct Debit",
       "MandateID" -> dd.mandateId
     )
-    case PaidProduct(_: CreditCardReferenceTransaction) => List("Default payment method" -> "Credit/Debit Card")
-    case PaidProduct(_: PayPalReferenceTransaction) => Seq("Default payment method" -> "PayPal")
-    case FreeProduct => Nil
+    case Left(_: CreditCardReferenceTransaction) => List("Default payment method" -> "Credit/Debit Card")
+    case Left(_: PayPalReferenceTransaction) => Seq("Default payment method" -> "PayPal")
+    case Right(_) => Nil
```
johnduffell, line 69: another less clear, could benefit from naming the underscore?

support-frontend/app/utils/CheckoutValidationRules.scala
johnduffell: or add the `: RedemptionData` which would be type checked I think
```suggestion
      _: RedemptionData => false
```

rupertbates: Yes I did think about that but was discouraged by Intellij because it's an 'unused' parameter. I think you're right though and the 'use' is as documentation.

PR #2506, opened 2020-05-20
Redemption code db tables

johnduffell:
## Why are you doing this?

For various forms of redemption e.g. gifting/corporate we want a db to store the codes so we can keep track of which codes exist (and possibly other data such as which product and how many licences in total.
The tables have the code as the primary key, and can have further data as needed, probably total available redemptions, and possibly a cache of active code usage from zuora (so we can still check whether any codes are available independently of zuora being up)

[**Trello Card**](https://trello.com/c/lZcb6e4F/3049-create-data-store-for-corporate-codes)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -105,7 +105,7 @@ lazy val `support-workers` = (project in file("support-workers"))
     integrationTestSettings,
     libraryDependencies ++= commonDependencies
   ).dependsOn(`support-services`, `support-models` % "test->test;it->test;compile->compile", `support-config`, `support-internationalisation`)
-  .aggregate(`support-services`, `support-models`, `support-config`, `support-internationalisation`, `stripe-intent`)
+  .aggregate(`support-services`, `support-models`, `support-config`, `support-internationalisation`, `stripe-intent`, `support-redemptiondb`)
 
 
 lazy val `support-models` = (project in file("support-models"))
--- a/build.sbt
+++ b/build.sbt
@@ -70,7 +70,8 @@ lazy val root = (project in file("."))
     `support-config`,
     `support-internationalisation`,
     `support-services`,
-    `stripe-intent`
+    `stripe-intent`,
+    `support-redemptiondb`
   )
 
 lazy val testScalastyle = taskKey[Unit]("testScalastyle")
@@ -158,3 +159,6 @@ lazy val `stripe-intent` = (project in file("support-lambdas/stripe-intent"))
     integrationTestSettings,
     libraryDependencies ++= commonDependencies,
   ).dependsOn(`support-rest`, `support-config`)
+
+lazy val `support-redemptiondb` = (project in file("support-redemptiondb"))
+  .enablePlugins(RiffRaffArtifact)
--- a/support-redemptiondb/build.sbt
+++ b/support-redemptiondb/build.sbt
@@ -1,0 +1,12 @@
+
+name := "support-redemptiondb"
+description:= "packages the database cfn for the redemption code db"
+
+assemblyJarName := "dummy.jar"
+riffRaffPackageType := assembly.value
+riffRaffUploadArtifactBucket := Option("riffraff-artifact")
+riffRaffUploadManifestBucket := Option("riffraff-builds")
+riffRaffManifestProjectName := "support:db:redemptiondb"
+riffRaffArtifactResources += (file("support-redemptiondb/cfn.yaml"), "cfn/cfn.yaml")
+riffRaffManifestBranch := Option(System.getenv("BRANCH_NAME")).getOrElse("unknown_branch")
+riffRaffBuildIdentifier := Option(System.getenv("BUILD_NUMBER")).getOrElse("DEV")
--- a/support-redemptiondb/cfn.yaml
+++ b/support-redemptiondb/cfn.yaml
@@ -4,11 +4,10 @@ Parameters:
     Type: String
     Description: Stage for the tables
     AllowedValues:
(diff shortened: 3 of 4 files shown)


build.sbt
```diff
@@ -104,7 +105,7 @@ lazy val `support-workers` = (project in file("support-workers"))
     integrationTestSettings,
     libraryDependencies ++= commonDependencies
   ).dependsOn(`support-services`, `support-models` % "test->test;it->test;compile->compile", `support-config`, `support-internationalisation`)
-  .aggregate(`support-services`, `support-models`, `support-config`, `support-internationalisation`, `stripe-intent`)
+  .aggregate(`support-services`, `support-models`, `support-config`, `support-internationalisation`, `stripe-intent`, `support-redemptiondb`)
```
johnduffell, line 108: I considered adding a new build a bit like support-logs, but it seemed excessive to have a build agent queued up for a 10 second build, together with all the associated configuration that isn't version controlled etc.
Another option would be to have a build that just calls a shell script that copies the files, then at least as little as possible is configured in TC.

support-redemptiondb/cfn.yaml
```diff
@@ -0,0 +1,58 @@
+AWSTemplateFormatVersion: '2010-09-09'
+Parameters:
+  Stage:
+    Type: String
+    Description: Stage for the tables
+    AllowedValues:
+      - CODE
+      - PROD
```
johnduffell, line 8: I decided in the end to have CODE and PROD (like the promo code tool) and then within PROD we have the touchpoint stages of DEV/UAT/PROD

rupertbates: I think this is the best approach 👍

support-redemptiondb/cfn.yaml
```diff
@@ -0,0 +1,58 @@
...
+Resources:
+  RedemptionCodeDynamoTableDEV:
```
rupertbates, line 10: I guess schema is all TBC at this point

support-redemptiondb/build.sbt
```diff
@@ -0,0 +1,12 @@
+
+name := "support-redemptiondb"
+description:= "packages the database cfn for the redemption code db"
+
+assemblyJarName := "dummy.jar"
```
rupertbates, line 5: `dummy.jar` because there is no jar?

build.sbt
rupertbates: What is it that tells SBT to copy the artifacts, the `RiffRaffArchitect` plugin?

johnduffell: yes, it will do so from the contents of the `riffRaffArtifactResources` sbt key
https://github.com/guardian/support-frontend/pull/2506/files#diff-90a52a2822abc32a0ffc37008d56294aR10

support-redemptiondb/build.sbt
johnduffell: yes, normally the jar is the key part (after all, it's an SBT build) and the riffRaffArtifactResources are extras on the side.  but this time it's backwards.
I should mentioned that an alternative approach is the approach I used in https://github.com/guardian/support-frontend/tree/master/support-frontend-static which was just get people to run a script locally to deploy.  But generally I dislike people deploying or building PROD stuff from their laptops as it's less controlled.

support-redemptiondb/cfn.yaml
johnduffell: yes
 in fact we only need to specify the primary keys and other search/ordering keys in the cfn.  the rest of the data can be arbitrary.  However we will indeed probably need more PRs on the cfn.

support-redemptiondb/cfn.yaml
```diff
@@ -0,0 +1,58 @@
...
+    Type: AWS::DynamoDB::Table
+    Properties:
+      AttributeDefinitions:
+        - AttributeName: "redemptionCode"
+          AttributeType: "S"
+      BillingMode: "PAY_PER_REQUEST"
+      KeySchema:
+        - AttributeName: "redemptionCode"
+          KeyType: "HASH"
+      TableName: !Sub redemption-codes-DEV-${Stage}
```
johnduffell, line 20: I thought the first touch point stage is DEV but looking at https://github.com/guardian/support-frontend/blob/968f37325dc68e7b51349a9797633618024ed4a8/support-models/src/main/scala/com/gu/support/config/TouchPointEnvironments.scala#L13
perhaps it should be SANDBOX?

johnduffell: looking at membership-common, members-data-api it is always referred to as DEV/UAT/PROD, so I think support is the outlier there
https://github.com/search?q=org%3Aguardian+sandbox+uat&type=Code

rupertbates: Yes I'd like to change it to DEV/UAT/PROD at some point

johnduffell:
I have deployed this to PROD as it's a manual deploy, not CI
https://riffraff.gutools.co.uk/deployment/view/25ae0ac5-5acc-47cf-8bae-bf289b35252c

image

https://eu-west-1.console.aws.amazon.com/dynamodb/home?region=eu-west-1#tables:selected=redemption-codes-DEV-PROD
https://eu-west-1.console.aws.amazon.com/dynamodb/home?region=eu-west-1#tables:selected=redemption-codes-UAT-PROD
https://eu-west-1.console.aws.amazon.com/dynamodb/home?region=eu-west-1#tables:selected=redemption-codes-PROD-PROD

PR #2507, opened 2020-05-20
Add survey link to thank you pages in subscriptions checkout

GHaberis:
## Why are you doing this?

We want a mechanism to receive regular/constant user feedback. We'd therefore like to add a link through to a survey on the thank you pages in subscriptions checkout.

[**Trello Card**](https://trello.com/c/Xy7HqSFa/3052-add-digital-subscription-survey-link-to-thank-you-page-post-purchase)

## Changes

- Create new `SubscriptionsSurvey.jsx` component. This component takes a prop `product`, using this prop it selects the appropriate survey link. If no survey link is available it returns `null`.
- Include this `SubscriptionsSurvey` component in the layouts for `thankYouContent.jsx` and `thankYouPendingContent.jsx`
- I had to update `thankYouPendingContent.jsx` so it had access to `state` using  `connect` from `react-redux`, I did this so I could then access `props.product`.

**note:** I initially wanted to style `SubscriptionsSurvey` using the design system, however I found the existing styles on the page didn't match those available in the design system. EG. I'd hoped to use the `@guardian/src-foundations/typography` lib to get font styles, unfortunately non of the pre-existing `<h3>` tag styles on these pages match the available styles exported by source components. EG. They all have `font-size: 22px` whilst `@guardian/src-foundations/typography` exposes `20px` and  `24px` as available sizes, but not `22px`. I decided consistency with the existing styles was the favourable path - I looked at updating the pre-existing elements styles but it's not insignificant - ideally we'll do this as a separate change at some point.

## Screenshots

**Live action on mobile:**

subs-survey

**Desktop:**

Screenshot 2020-05-20 at 16 57 35

**Tablet:**

Screenshot 2020-05-20 at 16 59 53

**Mobile:**

Screenshot 2020-05-20 at 16 58 58


support-frontend/assets/components/subscriptionCheckouts/subscriptionsSurvey/SubscriptionsSurvey.jsx
```diff
@@ -0,0 +1,43 @@
...
+const surveyLinks: {[key: SubscriptionProduct]: string} = {
+  DigitalPack: 'https://www.surveymonkey.co.uk/r/QF9ZGQR',
+  GuardianWeekly: 'https://www.surveymonkey.co.uk/r/QFNYV5G',
+  Paper: 'https://www.surveymonkey.co.uk/r/Q37XNTV',
+};
+
+export const SubscriptionsSurvey = ({ product }: PropTypes) => {
+  const surveyLink = surveyLinks[product];
+  const title = 'Tell us about your subscription';
+  const message = 'Please fill out this short form to help us learn a little more about why you purchased a subscription';
```
johnduffell, line 27: this does seema bit verbose, I'm sure some of the words can be deleted without changing the meaning (although maybe it came from someone who knows more than me!)

support-frontend/assets/components/subscriptionCheckouts/subscriptionsSurvey/SubscriptionsSurvey.jsx
```diff
@@ -0,0 +1,43 @@
...
+
+const surveyLinks: {[key: SubscriptionProduct]: string} = {
+  DigitalPack: 'https://www.surveymonkey.co.uk/r/QF9ZGQR',
+  GuardianWeekly: 'https://www.surveymonkey.co.uk/r/QFNYV5G',
+  Paper: 'https://www.surveymonkey.co.uk/r/Q37XNTV',
+};
+
+export const SubscriptionsSurvey = ({ product }: PropTypes) => {
+  const surveyLink = surveyLinks[product];
+  const title = 'Tell us about your subscription';
```
johnduffell, line 26: is this tell us more about your purchase or experience? as they only just started the subscription.

support-frontend/assets/components/subscriptionCheckouts/subscriptionsSurvey/SubscriptionsSurvey.jsx
johnduffell: also short form is not clear, perhaps you could say something quantified eg "2 minute" or "4 question"

PR #2508, opened 2020-05-21
use separated event model/event client

johnduffell:
## Why are you doing this?

See https://github.com/guardian/acquisition-event-producer/pull/62

[**Trello Card**](https://trello.com/c/PXslLgrv/3060-split-up-acquisition-event-producer-to-reduce-jar-sizes)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -55,7 +55,8 @@ lazy val commonSettings = Seq(
 
 lazy val commonDependencies = Seq(
   "com.typesafe" % "config" % "1.3.2",
-  "org.scalatest" %% "scalatest" % "3.2.0-M1" % "it, test"
+  "org.scalatest" %% "scalatest" % "3.2.0-M1" % "it, test",
+  "com.typesafe.scala-logging" %% "scala-logging" % "3.9.2"
 )
 
 lazy val root = (project in file("."))
--- a/project/LibraryVersions.scala
+++ b/project/LibraryVersions.scala
@@ -5,4 +5,5 @@ object LibraryVersions {
   val catsVersion = "1.4.0"
   val jacksonVersion = "2.10.1"
   val okhttpVersion = "3.10.0"
+  val scalaUriVersion = "2.2.2"
 }
--- a/support-frontend/build.sbt
+++ b/support-frontend/build.sbt
@@ -20,6 +20,7 @@ libraryDependencies ++= Seq(
   "com.amazonaws" % "aws-java-sdk-stepfunctions" % awsClientVersion,
   "com.amazonaws" % "aws-java-sdk-sts" % awsClientVersion,
   "com.amazonaws" % "aws-java-sdk-s3" % awsClientVersion,
+  "com.amazonaws" % "aws-java-sdk-lambda" % awsClientVersion,
   "org.typelevel" %% "cats-core" % "1.0.1",
   "com.dripower" %% "play-circe" % "2609.1",
   "com.gu" %% "fezziwig" % "1.3",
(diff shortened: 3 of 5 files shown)


build.sbt
```diff
@@ -55,7 +55,8 @@ lazy val commonSettings = Seq(
 
 lazy val commonDependencies = Seq(
   "com.typesafe" % "config" % "1.3.2",
-  "org.scalatest" %% "scalatest" % "3.2.0-M1" % "it, test"
+  "org.scalatest" %% "scalatest" % "3.2.0-M1" % "it, test",
+  "com.typesafe.scala-logging" %% "scala-logging" % "3.9.2"
```
johnduffell, line 59: we were incorrectly relying on a transitive dependency pulling this in.  I decided to put in common, after some thought.  But anything we put in common is really hard to escape so be careful!

support-models/build.sbt
```diff
@@ -5,8 +5,8 @@ name := "support-models"
 description := "Scala library to provide shared step-function models to Guardian Support projects."
 
 libraryDependencies ++= Seq(
-  "com.gu" %% "acquisition-event-producer-play26" % "4.0.27", //this should really be split into models and producer
-                                                              // so we don't have to pull in thrift binary compression libs etc
+  "com.gu" %% "acquisition-event-models-play26" % "4.0.29-SNAPSHOT",
+  "joda-time" % "joda-time" % "2.10.1",
```
johnduffell, line 9: we were incorrectly relying on this appearing as a transitive dependency

project/LibraryVersions.scala
```diff
@@ -5,4 +5,5 @@ object LibraryVersions {
   val catsVersion = "1.4.0"
   val jacksonVersion = "2.10.1"
   val okhttpVersion = "3.10.0"
+  val scalaUriVersion = "2.2.2"
```
johnduffell, line 8: we were using different versions between workers and frontend, neither quite up to date (workers was pulling in an old domain of simulacrum which caused undetected binary incompatibilities - clashing class names)

support-workers/build.sbt
```diff
@@ -38,7 +38,8 @@ libraryDependencies ++= Seq(
   "org.scala-stm" %% "scala-stm" % "0.8",
   "io.sentry" % "sentry-logback" % "1.7.4",
   "com.google.code.findbugs" % "jsr305" % "3.0.2",
-  "com.gocardless" % "gocardless-pro" % "2.8.0"
+  "com.gocardless" % "gocardless-pro" % "2.8.0",
+  "com.gu" %% "acquisition-event-client-play26" % "4.0.29"
```
johnduffell, line 42: next step (if we decide to take it) would be to move this dependency from the support-workers jar to a specific jar just for the send acquisition event.

support-frontend/build.sbt
```diff
@@ -21,6 +20,7 @@ libraryDependencies ++= Seq(
   "com.amazonaws" % "aws-java-sdk-stepfunctions" % awsClientVersion,
   "com.amazonaws" % "aws-java-sdk-sts" % awsClientVersion,
   "com.amazonaws" % "aws-java-sdk-s3" % awsClientVersion,
+  "com.amazonaws" % "aws-java-sdk-lambda" % awsClientVersion,
```
johnduffell, line 23: this was previously getting pulled in via support-models -> acquisition event producer -> acquisition value calculator client

johnduffell:
tested in CODE and working with DP and contribution OK - just need a review now

PR #2512, opened 2020-05-29
Fix isRemote handling - is a boolean

nicl:
## Why are you doing this?

This is a fix for https://github.com/guardian/support-frontend/pull/2480.

The actual query param (once JSONified) looks like:

`{..."isRemote":true}`

### DIFF
--- a/support-frontend/assets/helpers/abTests/abtest.js
+++ b/support-frontend/assets/helpers/abTests/abtest.js
@@ -146,7 +146,7 @@ function getIsRemoteFromAcquisitionData(): boolean {
 
   try {
     const data = JSON.parse(queryString);
-    return data ? data.isRemote === 'true' : false;
+    return data && !!data.isRemote;
   } catch {
     console.error('Cannot parse acquisition data from query string');
     return false;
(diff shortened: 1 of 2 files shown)


support-frontend/assets/helpers/abTests/abtest.js
```diff
@@ -146,7 +146,7 @@ function getIsRemoteFromAcquisitionData(): boolean {
 
   try {
     const data = JSON.parse(queryString);
-    return data ? data.isRemote === 'true' : false;
+    return data ? data.isRemote === true : false;
```
tjmw, line 149: Could this just be:

```js
return data ? data.isRemote : false;
```

nicl: Yes, although I might prefer `return data ? !!data.isRemote : false` . I hesitated to do this as prefer being explicit, but this is safer, particularly given unclear how to test locally.

nicl:
@johnduffell thanks. I've added some basic tests. This code is intended to be short-lived, so I'll remove once we've validated the remote epic service.

PR #2514, opened 2020-05-29
Redemption code api

johnduffell:
## Why are you doing this?

This PR is part of the corporate subs work.
We are storing the codes in a set of dynamo tables, we need an endpoint for the client side to check the codes against dynamo.

This PR only covers the check status endpoint, this can be merged and will validate the overall approach, then based on the result there, I can implement the support-workers call later.

[**Trello Card**](https://trello.com/c/tF4SGp5U/3051-add-simple-corporate-code-get-update-call)

The API is provisional and has no rate limiting etc, but at present: 
GET https://support.thegulocal.com/redemption/ITTEST-AVAILABLE/status
gives a 200 with empty body
GET https://support.thegulocal.com/redemption/ITTEST-USED/status
gives a 404 with body "CODE_ALREADY_USED" and a Content-Type: text/plain; charset=UTF-8
GET https://support.thegulocal.com/redemption/ITTEST-MISSING/status
gives a 404 with body "NO_SUCH_CODE"
everything else is a 500 error

you can also pass in ?isTestUser=true to use UAT.  If we're expecting people to be logged in anyway, we could get it from the authenticated user, at the cost of an extra call to identity.


support-frontend/app/admin/settings/Settings.scala
```diff
@@ -46,7 +46,7 @@ object Settings {
       buf <- Either.catchNonFatal {
         val inputStream = s3.getObject(source.bucket, source.key).getObjectContent
         Source.fromInputStream(inputStream)
-      }
+      }.leftMap(ex => new RuntimeException(s"couldn't getObject content for source: $source", ex))
```
johnduffell, line 49: add useful logging

support-frontend/build.sbt
```diff
@@ -93,4 +93,6 @@ javaOptions in Universal ++= Seq(
 
 javaOptions in Test += "-Dconfig.file=test/selenium/conf/selenium-test.conf"
 
+scalacOptions += "-Ypartial-unification"
```
johnduffell, line 96: needed so we can sequence an Either or any other higher order operation on something with more than one type parameter. actually I refactored the code that needed it out, but this is good to have on anyway
explanation of what it is here:
https://gist.github.com/djspiewak/7a81a395c461fd3a09a6941d4cd040f2

support-services/src/main/scala/com/gu/support/promotions/dynamo/DynamoService.scala
```diff
@@ -33,7 +31,7 @@ object DynamoService {
     new InstanceProfileCredentialsProvider(false)
   )
 
-  def forTable[T](table: String)(implicit e: ExecutionContext, decoder: Decoder[T]): DynamoService[T] = {
+  def forTable[T](table: String)(implicit decoder: Decoder[T]): DynamoService[T] = {
```
johnduffell, line 34: I just did some tidy up on this file anyway, I was going to put the new stuff here it but I realised none of it was actually related other than it also hitting dynamo

support-services/src/main/scala/com/gu/support/redemption/DynamoTableAsync.scala
```diff
@@ -0,0 +1,82 @@
...
+    new DynamoTableAsync(dynamoDBClient, table, key)
+  }
+}
+trait DynamoLookup {
+  def lookup(value: String): Future[Option[Map[String, AttributeValue]]]
+}
+
+trait DynamoUpdate {
+  def update(primaryKeyValue: String, updateName: String, updateValue: Boolean): Future[Unit]
+}
```
johnduffell, line 31: the two traits are to prevent the operation having to depend on the whole service, which is good in itself.  The practical benefit of that is easier reuse in the tests, the test mocks can just be a simple inline single abstract method with a pattern match, which makes the tests a hundred times clearer and banishes mockito 🎉

support-services/src/test/scala/com/gu/support/redemption/RedemptionSpec.scala
```diff
@@ -0,0 +1,79 @@
...
+class RedemptionSpec extends AsyncFlatSpec with Matchers {
+
+  "getCodeStatus" should "handle an available code" in {
+    val getCodeStatus = GetCodeStatus.withDynamoLookup {
+      case "CODE" => Future.successful(Some(Map("available" -> new AttributeValue().withBOOL(true))))
+    }
+    getCodeStatus(RedemptionCode("CODE")).map {
+      _ should be(Right(()))
+    }
+  }
```
johnduffell, line 20: if we wanted to make the test really neat, we could get parameterise `GetCodeStatus` on the `Future` and pass in `Id`.  This would get rid of the Future.successful and the .map, but I think the overall effect on the code is negative, and I'm happy enough with the clarity of the tests without.

support-services/src/main/scala/com/gu/support/redemption/DynamoTableAsync.scala
johnduffell: perhaps a better option would be to separate the below `DynamoTableAsync` into two classes.  I put them together to reduce the amount of stuff we have and because they both depend on the same parameters, but in hindsight that seems a bit unnecessary to have them grouped

support-frontend/build.sbt
rupertbates: I'd leave it out if it's not needed for this PR just to keep things simple

johnduffell: sounds sensible, done that now.

support-services/build.sbt
```diff
@@ -5,6 +5,8 @@ name := "support-services"
 description := "Scala library to provide shared services to Guardian Support projects."
 
 libraryDependencies ++= Seq(
+  "software.amazon.awssdk" % "dynamodb" % "2.13.26",
```
johnduffell, line 8: the 1.x dynamo async client used a legacy way to contact dynamo which didn't allow the specification of a ConditionExpression together with attributes.  I needed to apply a condition to the update call to stop it from creating a redemption code if the provided one was not already in the DB(!!!)
there was a non async client that might have supported it, but I wouldn't want the whole server to hang if dynamo is being slow/timing out.
Aside from that, the 2.x client is supposed to be faster startup, more lightweight deps, and supports a nicer async (java 8) rather than the custom one.
On that basis we should endeavour to move everything across over time - it's fairly mind numbing though.

support-services/build.sbt
```diff
@@ -5,6 +5,8 @@ name := "support-services"
...
+  "org.scala-lang.modules" %% "scala-java8-compat" % "0.9.1",
```
johnduffell, line 9: new dep but it doesn't have any onward deps https://mvnrepository.com/artifact/org.scala-lang.modules/scala-java8-compat_2.12/0.9.1

support-services/build.sbt
johnduffell: https://mvnrepository.com/artifact/software.amazon.awssdk/dynamodb/2.13.26
```
$ cs fetch software.amazon.awssdk:dynamodb:2.13.26|xargs -I{} ls -sk {}|sed 's# .*/#kB #'|sort -n
kB annotations-2.13.26.jar
kB apache-client-2.13.26.jar
kB aws-core-2.13.26.jar
kB commons-logging-1.2.jar
kB http-client-spi-2.13.26.jar
kB httpclient-4.5.9.jar
kB jackson-annotations-2.10.4.jar
kB netty-codec-http-4.1.46.Final.jar
kB netty-common-4.1.46.Final.jar
kB sdk-core-2.13.26.jar
12kB reactive-streams-1.0.3.jar
24kB netty-reactive-streams-2.0.4.jar
28kB profiles-2.13.26.jar
32kB eventstream-1.0.1.jar
32kB protocol-core-2.13.26.jar
36kB netty-reactive-streams-http-2.0.4.jar
36kB netty-resolver-4.1.46.Final.jar
36kB netty-transport-native-unix-common-4.1.46.Final.jar
44kB slf4j-api-1.7.28.jar
104kB utils-2.13.26.jar
108kB aws-json-protocol-2.13.26.jar
128kB auth-2.13.26.jar
152kB netty-transport-native-epoll-4.1.46.Final-linux-x86_64.jar
236kB netty-nio-client-2.13.26.jar
272kB netty-buffer-4.1.46.Final.jar
316kB netty-codec-4.1.46.Final.jar
320kB httpcore-4.4.11.jar
328kB commons-codec-1.11.jar
344kB jackson-core-2.10.4.jar
408kB regions-2.13.26.jar
448kB netty-codec-http2-4.1.46.Final.jar
464kB netty-transport-4.1.46.Final.jar
492kB netty-handler-4.1.46.Final.jar
2096kB dynamodb-2.13.26.jar
2272kB jackson-databind-2.10.4.jar
```

support-services/build.sbt
```diff
@@ -5,6 +5,8 @@ name := "support-services"
...
   "com.amazonaws" % "aws-java-sdk-dynamodb" % awsClientVersion,
```
johnduffell, line 10:
```
$ cs fetch com.amazonaws:aws-java-sdk-dynamodb:1.11.642|xargs -I{} ls -sk {}|sed 's# .*/#kB #'|sort -n
kB aws-java-sdk-core-1.11.642.jar
kB aws-java-sdk-kms-1.11.642.jar
kB aws-java-sdk-s3-1.11.642.jar
kB commons-logging-1.2.jar
kB httpclient-4.5.9.jar
kB ion-java-1.0.2.jar
kB joda-time-2.8.1.jar
28kB jmespath-java-1.11.642.jar
48kB jackson-annotations-2.6.0.jar
48kB jackson-dataformat-cbor-2.6.7.jar
256kB jackson-core-2.6.7.jar
320kB httpcore-4.4.11.jar
328kB commons-codec-1.11.jar
1148kB jackson-databind-2.6.7.2.jar
1148kB jackson-databind-2.6.7.2.jar
1644kB aws-java-sdk-dynamodb-1.11.642.jar
```

support-services/src/main/scala/com/gu/support/redemption/DynamoTableAsync.scala
```diff
@@ -0,0 +1,86 @@
...
+  def apply(table: String, key: String)(implicit e: ExecutionContext): DynamoTableAsync = {
+    val dynamoDBClient = DynamoDbAsyncClient.builder
+      .credentialsProvider(CredentialsProvider)
+      .region(Region.EU_WEST_1)
+      .build
+    new DynamoTableAsync(dynamoDBClient, table, key)
+  }
+}
+trait DynamoLookup {
+  def lookup(value: String): Future[Option[Map[String, Boolean]]]
```
johnduffell, line 33: the implementation will discard all non-boolean attrributes. We can always extend it in future, but it didn't seem worth a full scala model, and the aws model is so verbose I wanted to keep it out of the API.
Could rename this to DynamoLookupBoolean in the meantime though?
I don't want to move too much code that should be unit tested into the client though.  As I don't plan to unit test the AWS api if at all possible!

support-services/src/test/scala/com/gu/support/redemption/GetCodeStatusSpec.scala
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+//  it should "handle an code with invalid attrbibute type " in {
+//    val getCodeStatus = GetCodeStatus.withDynamoLookup {
+//      case "CODE" => Future.successful(Some(Map("available" -> new AttributeValue().withS("true"))))
+//    }
+//    recoverToSucceededIf[RuntimeException] {
+//      getCodeStatus(RedemptionCode("CODE"))
+//    }
+//  }
```
johnduffell, line 57: this test would have to happen against DynamoTableAsync further up in this PR, as the code has been pushed down to simplify the API.  It could be worth extracting the non-side effect parts of that file, and testing them? the tests would be messy but wouldn't be mixed with any business logic tests such as the current one.
https://github.com/guardian/support-frontend/blob/79d715962389a72a0c7bdb10ee6efa055297544a/support-services/src/main/scala/com/gu/support/redemption/DynamoTableAsync.scala#L63

support-services/src/main/scala/com/gu/support/redemption/DynamoTableAsync.scala
rupertbates: We are going to have to be able to return a `corporateAccountId` to store in Zuora as this is how we will link an account to a particular corporate customer.

My mocked implementation of this which I used to build the redemption form is here: https://github.com/guardian/support-frontend/pull/2518/files#diff-4d6b86fc69b601966db937093ee55092R105

It returns a case class with the redemption code and the account id (it actually has a corporate name as well but we are not actually using this currently)
https://github.com/guardian/support-frontend/pull/2518/files#diff-1f241c7b97cb23a815da6c5e5221eceeR5

support-services/src/main/scala/com/gu/support/redemption/DynamoTableAsync.scala
rupertbates: Fine to leave them grouped I think

support-services/src/main/scala/com/gu/support/redemption/WithLogging.scala
```diff
@@ -0,0 +1,31 @@
...
+import com.gu.monitoring.SafeLogger
+
+import scala.concurrent.{ExecutionContext, Future}
+
+trait WithLogging {
+
+  implicit class LogImplicit2[A](value: Future[A]) {
+
+    // this is just a handy method to add logging to the end of any value
+    def withLoggingAsync(message: String)(implicit executionContext: ExecutionContext): Future[A] = {
```
rupertbates, line 12: This already exists here: https://github.com/guardian/support-frontend/blob/e6b51f248825528661f5160b5dc86498284a4598/support-lambdas/stripe-intent/src/main/scala/com/gu/handler/LogImplicit.scala#L12

support-frontend/app/controllers/RedemptionsController.scala
```diff
@@ -0,0 +1,35 @@
+package controllers
+
+import actions.CustomActionBuilders
+import com.gu.support.config.{Stage, TouchPointEnvironments}
+import com.gu.support.redemption.{GetCodeStatus, RedemptionCode, RedemptionTable}
+import play.api.mvc.{AbstractController, Action, AnyContent, ControllerComponents}
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class RedemptionsController(
```
rupertbates, line 10: I have a version of this in my corporate subs part 2 PR so there'll be a bit of a merge to do. We could keep this in a separate controller maybe but would need to get the naming right

support-services/src/main/scala/com/gu/support/promotions/dynamo/DynamoService.scala
rupertbates: Do we not need an execution context here?

support-services/src/main/scala/com/gu/support/redemption/DynamoTableAsync.scala
```diff
@@ -0,0 +1,86 @@
...
+
+import scala.collection.JavaConverters._
+import scala.compat.java8.FutureConverters._
+import scala.concurrent.{ExecutionContext, Future}
+
+object DynamoTableAsync {
+
+  val ProfileName = "membership"
+
+  lazy val CredentialsProvider =  AwsCredentialsProviderChain.builder.credentialsProviders(
```
rupertbates, line 19: I think you should be able to use the CredentialsProvider here:
https://github.com/guardian/support-frontend/blob/5a66eeae1ce6bb0b0bd1ef4bef64fde72492572b/support-services/src/main/scala/com/gu/aws/aws.scala#L9

support-services/src/main/scala/com/gu/support/redemption/DynamoTableAsync.scala
```diff
@@ -0,0 +1,86 @@
...
+import software.amazon.awssdk.auth.credentials.{AwsCredentialsProviderChain, InstanceProfileCredentialsProvider, ProfileCredentialsProvider}
+import software.amazon.awssdk.regions.Region
+import software.amazon.awssdk.services.dynamodb.DynamoDbAsyncClient
+import software.amazon.awssdk.services.dynamodb.model.{AttributeValue, GetItemRequest, UpdateItemRequest}
+
+import scala.collection.JavaConverters._
+import scala.compat.java8.FutureConverters._
+import scala.concurrent.{ExecutionContext, Future}
+
+object DynamoTableAsync {
```
rupertbates, line 15: Maybe call this `RedemptionTableAsync` and the classes below `RedemptionLookup` or `RedemptionDynamoLookup` etc.? 

I'm not entirely sure but Dynamo feels a bit too generic. We already have a `DynamoTables` class in the promotions namespace for instance.

support-services/src/main/scala/com/gu/support/redemption/DynamoTableAsync.scala
```diff
@@ -0,0 +1,86 @@
...
+  def update(primaryKeyValue: String, updateName: String, updateValue: Boolean): Future[Unit]
+}
+
+class DynamoTableAsync(
+  dynamoDBAsyncClient: DynamoDbAsyncClient,
+  table: String,
+  primaryKeyName: String
+)(implicit e: ExecutionContext) extends LazyLogging with DynamoLookup with DynamoUpdate {
+
+  override def lookup(primaryKeyValue: String): Future[Option[Map[String, Boolean]]] = {
```
rupertbates, line 46: is `primaryKeyValue` the redemptionCode? If so it might be be good to call this parameter `redemptionCode`? Less mental translation to do...

support-services/src/main/scala/com/gu/support/redemption/DynamoTableAsync.scala
```diff
@@ -0,0 +1,86 @@
...
+      maybeAttributes.map(attributes =>
+        attributes.flatMap {
+          // only need to handle boolean values at the moment
+          case (key, value) => Option(value.bool).map(Boolean2boolean).map(bool => key -> bool)
+        }
+      )
+    }
+  }
+
+  override def update(primaryKeyValue: String, updateName: String, updateValue: Boolean): Future[Unit] = {
```
rupertbates, line 69: What is `updateName`?

support-services/src/main/scala/com/gu/support/redemption/GetCodeStatus.scala
```diff
@@ -0,0 +1,47 @@
+package com.gu.support.redemption
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.{Failure, Success}
+
+object GetCodeStatus {
+
+  sealed abstract class RedemptonInvalid(val clientCode: String)
```
rupertbates, line 8: typo, should be `RedemptionInvalid`

support-services/src/main/scala/com/gu/support/promotions/dynamo/DynamoService.scala
johnduffell: no, it compiles without, and all of the promo code stuff is using the blocking client anyway (probably shouldn't be but it hasn't caused an incident so far)

support-services/src/main/scala/com/gu/support/redemption/DynamoTableAsync.scala
johnduffell: ok I will update suitably and make it support a string too

support-services/src/main/scala/com/gu/support/redemption/WithLogging.scala
johnduffell: yes I didn't think it was worth adding a dependency just for a bit of syntactic sugar so I just copied and pasted it.  There are a couple of copies already in support-frontend repo.  Happy to consider alternative approaches though.

support-frontend/app/controllers/RedemptionsController.scala
johnduffell: one way to do it might be to just merge both, and then do a separate pr to connect them together

support-services/src/main/scala/com/gu/support/redemption/DynamoTableAsync.scala
johnduffell: there isn't anything redemption specific in the code (and intentionally so for testing and separation reasons), it's only exists to put a simplified and scala-friendly facade on dynamo async table api to make the business logic nicer.
Yes in reality the scala api for it is only used by redemptions, but that is not relevant to the class itself (and that's why it's in the `com.gu.support.redemption` package)
If the need happened, I'd be happy to pull it out of the package and use it in other places, perhaps the promo codes stuff will use this same code if we update it to aws-sdk v2.
Adding the word redemption encourages redemption specific code to end up here.
Makes sense or still prefer a rename?

support-services/src/main/scala/com/gu/support/redemption/DynamoTableAsync.scala
johnduffell: I tried originally but that's the AWS SDK v1 version, so it's in a different package although has pretty much the same name and functionality.

support-services/src/main/scala/com/gu/support/redemption/DynamoTableAsync.scala
johnduffell: it depends what we passed in as primaryKeyName on line 43.
I think the decision here is hinged on whether this is a redemptions client or a dynamo-to-scala layer as per the above discussion.

support-services/src/main/scala/com/gu/support/redemption/DynamoTableAsync.scala
johnduffell: the name of the field to update.  but good point, I think the param names of this function could do with a bit of attention

support-services/src/main/scala/com/gu/support/redemption/DynamoTableAsync.scala
johnduffell: will do, but interesting comment, I find it easier to use code with fewer interdependencies, which is actually not the same as fewer classes, as things often have to be finer grained to avoid a dependency that's not strictly necessary.  I wonder if there's some element of tradeoff there.

support-workers/build.sbt
```diff
@@ -55,6 +55,7 @@ assemblyMergeStrategy in assembly := {
   case PathList("models", xs@_*) => MergeStrategy.discard
   case x if x.endsWith("io.netty.versions.properties") => MergeStrategy.first
   case x if x.endsWith("module-info.class") => MergeStrategy.discard
+  case "mime.types" => MergeStrategy.first
```
johnduffell, line 58: this occurs in both aws sdks, both are slightly different but similar enough. It's just a mapping from extension to mime type so probably not used by anything we do anyway.

support-services/src/main/scala/com/gu/support/redemption/DynamoTableAsync.scala
johnduffell: couldn't get the naming I was happy with, decided grouping them into a case class makes it clearer at the expense of more code.  diff here:
https://github.com/guardian/support-frontend/pull/2514/files/8d5afa03d4654d2fc35f2807e4de07f5bbb0351a..cfb89a238ed7d9d95810eb88c805a8992b4588d9

support-services/src/main/scala/com/gu/support/redemption/DynamoTableAsync.scala
johnduffell: so as far as I can tell for the client side we don't need any information other than is it valid, on the support-workers side we will need to know the `corporateAccountId` and whether it's valid.  If I'm reading the code and your comment correctly.
I am updating to return those from the underlying call, and use the appropriately.
If we do need extra information I'm happy to add, but starting with the simplest possible implementation, even serialising to json seems unnecessary on first sight?

support-services/src/main/scala/com/gu/support/redemption/DynamoTableAsync.scala
rupertbates: Yes that's fine

support-services/src/main/scala/com/gu/support/redemption/DynamoTableAsync.scala
rupertbates: We will need the corporateAccountId in support-frontend but I think this interface should be fine.

PR #2515, opened 2020-06-01
Landing page design system components test

tjsilver:
## Why are you doing this?

We want to test the landing page implemented using design system (DS) components against the existing page. The DS components are designed with accessibility in mind.

[**Trello Card**](https://trello.com/c/wKOxKTvQ/1883-design-system-form-integration)

## Changes
* Control: no change
* Variant: landing page implementing design system components
   * DS components used in text input fields and radio buttons only (in addition to existing choice cards)
   * Stripe elements fields (credit card) are styled to look like DS components as far as possible but there are limitations as Stripe provides limited styling opportunities
   * State/province picker is styled to look like a DS component (not available currently in DS).

## Control
image
image

### Desktop
image
image

### Mobile
image
image
image

### Mobile - iOS
image
image
image

### Existing payment methods
image

### Errors 
image

image
image
image


support-frontend/assets/pages/contributions-landing/components/ContributionStateDs.jsx
```diff
@@ -0,0 +1,183 @@
+// @flow
```
tomrf1, line 1: Does this need to be a separate file? Having all this duplicated code feels a bit risky as we have to remember to maintain both

tjsilver: Good point, but hopefully we will only have one file after the test (hopefully the DS one)!?

support-frontend/assets/pages/contributions-landing/components/ContributionFormFields.jsx
```diff
@@ -88,9 +94,8 @@ function withProps(props: PropTypes) {
     checkoutFormHasBeenSubmitted,
   } = props;
 
-  return (
-    <div className="form-fields">
-      <h3 className="hidden-heading">Your details</h3>
+  const renderControl = () => (
+    <>
```
JoemG, line 98: ? empty <>

support-frontend/assets/pages/contributions-landing/components/ContributionFormFields.jsx
```diff
@@ -155,6 +160,79 @@ function withProps(props: PropTypes) {
         isValid={checkBillingState(billingState)}
         formHasBeenSubmitted={checkoutFormHasBeenSubmitted}
       />
+    </>
```
JoemG, line 163:
```suggestion
```

SiAdcock:
Looks good to me! I have two comments really, not blockers

- I notice you're using `<InlineError>`, which is currently undocumented, not explicitly listed as a component available for consumption and is used exclusively within other Source components. This is fine, the component rarely changes and I don't plan on removing it. It would be interesting to know whether you found this useful as a component for external consumption?
- There are some SVGs you've inlined... are these icons that are not available in Source right now, or did you need to customise them to serve your use case? Is there anything we can do in Source to make our icons more useful?

Excellent work on this @tjsilver and @tomrf1

tomrf1:
@SiAdcock `<InlineError>` is very useful where we need to implement our own component, perhaps in practice there will always be cases where we need it?

PR #2518, opened 2020-06-02
Corporate Subscription Redemption Form

rupertbates:
## Why are you doing this?

This PR adds a corporate subscription redemption form which will allow users with a valid corporate subscription code to create a Digital Subscription.

The flow is as follows:
1. Redemption form
Screen Shot 2020-06-04 at 15 14 26
2. When user clicks 'Activate' the user is redirected to Identity (unless they are already logged in) then return to:
Screen Shot 2020-06-04 at 15 16 57
3. If subscription creation is successful they are redirected to the thank you page
Screen Shot 2020-06-04 at 15 17 30

## TODO
- [ ] Styling to match [designs](https://www.figma.com/file/81Ku9XImBCx1jQXIP8EoF2/Corporate-Subs-2020?node-id=603%3A285) 
- [ ] Handle users with existing Digital Subscriptions
- [ ] Tracking
- [x] Handle submission errors
- [ ] Handle test users
- [ ] Accessibility testing
 
[**Trello Card**](https://trello.com/c/xfNSDWLO/3029-corporate-subs-integration-work-part-2-redemption-form)


support-frontend/app/services/stepfunctions/SupportWorkersClient.scala
```diff
@@ -120,8 +120,7 @@ class SupportWorkersClient(
   def createSubscription(
     request: AnyAuthRequest[CreateSupportWorkersRequest],
     user: User,
-    requestId: UUID,
-    promoCode: Option[PromoCode] = None
```
rupertbates, line 124: This was unused

support-frontend/app/utils/CheckoutValidationRules.scala
```diff
@@ -61,7 +61,7 @@ object AddressAndCurrencyValidationRules {
   }
 
 
-  def hasStateIfRequired(countryFromRequest: Country, stateFromRequest: Option[String], currencyFromRequest: Currency): Boolean =
```
rupertbates, line 64: Also unused

support-frontend/app/controllers/RedemptionController.scala
```diff
@@ -0,0 +1,147 @@
...
+
+  implicit val settings: AllSettings = settingsProvider.getAllSettings()
+  val title = "Support the Guardian | Redeem your code"
+  val id = EmptyDiv("subscriptions-redemption-page")
+  val js = "subscriptionsRedemptionPage.js"
+  val css = "digitalSubscriptionCheckoutPage.css" //TODO: Don't need this?
+
+  def displayForm(redemptionCode: String): Action[AnyContent] = NoCacheAction().async {
+    implicit request =>
+      getCorporateCustomer(redemptionCode).value.map(
```
johnduffell, line 51: if the future is failed, we will get a 500 error.  Failed future would be if the lookup broke, rather than just being a code missing/used.
Not sure if it's any use to recover the future to a success of left, at least with a 5xx we will be notified.

support-frontend/app/controllers/RedemptionController.scala
```diff
@@ -0,0 +1,147 @@
...
+            user = None,
+            submitted = false
+          ))
+      )
+
+  }
+
+  def displayThankYou(stage: String): Action[AnyContent] = authenticatedAction(subscriptionsClientId) {
+    implicit request =>
+          Ok(views.html.main(
```
johnduffell, line 73: indentation

support-frontend/app/controllers/RedemptionController.scala
```diff
@@ -0,0 +1,147 @@
...
+    authenticatedAction(subscriptionsClientId).async {
+      implicit request: AuthRequest[Any] =>
+        val processingPage = for {
+          user <- identityService.getUser(request.user.minimalUser)
+          corporateCustomer <- getCorporateCustomer(redemptionCode)
+        } yield showProcessing(redemptionCode, corporateCustomer, user)
+
+        processingPage.value.map(
+          _.fold(
+            BadRequest(_),
```
johnduffell, line 95: this will potentially send the identity url or any json parse error of the identity response back to the client which is not ideal for security.
The identityService should probably use failed future for any unexpected problem with the call, and Left should be reserved for signalling something specific e.g. user not found.

support-frontend/app/controllers/RedemptionController.scala
```diff
@@ -0,0 +1,147 @@
...
+            result => result
+          )
+        )
+    }
+
+  def showProcessing(
+    redemptionCode: RedemptionCode,
+    corporateCustomer: CorporateCustomer,
+    user: IdUser)(implicit request: AuthRequest[Any]): Result = {
```
johnduffell, line 104:
```suggestion
    user: IdUser
  )(implicit request: AuthRequest[Any]): Result = {
```

support-frontend/app/controllers/RedemptionValidationResult.scala
```diff
@@ -0,0 +1,10 @@
+package controllers
+
+case class RedemptionValidationResult(valid: Boolean, errorMessage: Option[String])
+
+import com.gu.support.encoding.Codec
+import com.gu.support.encoding.Codec._
+
+object RedemptionValidationResult {
+  implicit val codec: Codec[RedemptionValidationResult] = deriveCodec
```
johnduffell, line 9:
```suggestion
  implicit val encoder: Encoder[RedemptionValidationResult] = deriveEncoder
```

support-frontend/app/utils/CheckoutValidationRules.scala
```diff
@@ -94,21 +94,25 @@ object DigitalPackValidation {
...
+    val allowCorporateSubs = true
 
-    def hasValidPaymentDetails(request: CreateSupportWorkersRequest) = paymentFields.fold(
-      paymentFields => PaidProductValidation.noEmptyPaymentFields(paymentFields),
-      redemptionData => isCorporateSub(redemptionData)
-    )
+    def isValidCorporateSub(redemptionData: RedemptionData) =
+      SimpleCheckoutFormValidation.passes(createSupportWorkersRequest) &&
+        hasStateIfRequired(country, state) &&
+        allowCorporateSubs //TODO: validate redemption data here? This would mean returning a Future
```
johnduffell, line 102: This is similar to promo codes and we don't validate them until CreateZuoraSubscription (in fact we didn't even validate in CreateZuoraSubscription until the one for one incident!).  This validation is to avoid support workers failures caused by invalid submissions/hacker http scanners/etc I think.

Reminds me though, someone on chat the other day was saying "parse, don't validate" as a principle (in the most general sense), and it made me think about this code.
Might be a bit much for a refactor, but food for thought.

support-frontend/app/views/subscriptionRedemptionForm.scala.html
```diff
@@ -0,0 +1,43 @@
...
+@import com.gu.support.encoding.CustomCodecs._
+@import com.gu.support.redemptions.redemptions.RedemptionCode
+@import helper.CSRF
+@import views.ReactDiv
+@import com.gu.support.redemptions.CorporateCustomer
+@import io.circe.syntax._
+
+@import io.circe.Json
+@(
+        title: String,
```
johnduffell, line 13: indentation?

support-frontend/app/views/subscriptionRedemptionForm.scala.html
```diff
@@ -0,0 +1,43 @@
...
+)(implicit assets: AssetsResolver, requestHeader: RequestHeader, settings: AllSettings)
+
+@main(title = title, mainJsBundle = Left(RefPath(js)), mainElement = mainElement, mainStyleBundle = Left(RefPath(css)), fontLoaderBundle = fontLoaderBundle, csrf = csrf) {
+  <script type="text/javascript">
+      window.guardian = window.guardian || {};
+      window.guardian.stage = "@stage";
+      window.guardian.userCode = "@redemptionCode";
+      window.guardian.error = "@customerOrError.left.getOrElse("")";
+      window.guardian.submitted = @submitted;
+      window.guardian.corporateCustomer = @Html(customerOrError.right.map(_.asJson.noSpaces).getOrElse(Json.Null.noSpaces));
```
johnduffell, line 34: why the `@Html` and not just @?  adding `Html`  will prevent it being escaped correctly.

support-frontend/app/controllers/RedemptionController.scala
```diff
@@ -0,0 +1,147 @@
...
+  def displayThankYou(stage: String): Action[AnyContent] = authenticatedAction(subscriptionsClientId) {
+    implicit request =>
+          Ok(views.html.main(
+            title = title,
+            mainElement = id,
+            mainJsBundle = Left(RefPath(js)),
+            mainStyleBundle = Left(RefPath(css)),
+            fontLoaderBundle = fontLoaderBundle,
+          ){
+            Html(s"""<script type="text/javascript">window.guardian.stage = "${stage}";</script>""")
```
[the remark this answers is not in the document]
rupertbates: I'm not sure I understand what you're saying, there is no user supplied input being used here

support-frontend/app/views/subscriptionRedemptionForm.scala.html
johnduffell: correction - it will prevent it being escaped INcorrectly, I just tested it! 😆

support-frontend/assets/components/forms/customFields/validated.jsx
```diff
@@ -0,0 +1,72 @@
...
+      <Element
+        aria-hidden={!error}
+        aria-atomic="true"
+        aria-live="polite"
+        htmlFor={htmlFor}
+        className="component-form-error__error"
+      >
+        {(error === 'Temporary COVID message') && (
+          <li className="component-form-error__summary-error">
+            The address and postcode you entered is outside of our delivery area. You may want to
```
johnduffell, line 51: not sure if this is relevant to digi?

support-frontend/assets/pages/subscriptions-redemption/thankYouContainer.js
```diff
@@ -0,0 +1,16 @@
...
+
+import { connect } from 'react-redux';
+import ThankYouContent from 'pages/digital-subscription-checkout/thankYouContent';
+import type { RedemptionPageState } from 'pages/subscriptions-redemption/subscriptionsRedemptionReducer';
+import { Stripe } from 'helpers/paymentMethods';
+
+function mapStateToProps(state: RedemptionPageState) {
+  return {
+    countryGroupId: state.common.internationalisation.countryGroupId,
+    paymentMethod: Stripe,
```
johnduffell, line 11: if it's redemption it won't be Stripe, there won't be one, (in the scala we did an Either, but I'm not sure how it goes in flow)

support-frontend/test/utils/CheckoutValidationRulesTest.scala
```diff
@@ -134,6 +137,15 @@ class DigitalPackValidationTest extends AnyFlatSpec with Matchers {
     DigitalPackValidation.passes(requestMissingAddressLineAndCity) shouldBe false
   }
 
+  it should "succeed when there is a valid corporate sub" in {
+    val corporateSub = validDigitalPackRequest.copy(
+      product = DigitalPack(GBP, Monthly, Corporate),
+      paymentFields = Right(CorporateRedemption("test-code", "1"))
+    )
+
+    DigitalPackValidation.passes(corporateSub) shouldBe true
```
johnduffell, line 146: consider testing the other cases, e.g. that corporate does NOT work for GW/paper/etc.

I know we decided against it due to sheer amount of work, but if we could have got it into the types, then we wouldn't need tests as the type system would prove it.

support-frontend/assets/components/forms/customFields/validated.jsx
rupertbates: It's not but this component is a generic one that can be used elsewhere

johnduffell: I don't fully understand, if it's got paper HD stuff in it it's not really generic?  If it was, any product specific stuff should be passed in as a parameter.  Happy to defer though as I'm not totally up on structuring client side stuff.

rupertbates: Yes it's come directly from the error component:
https://github.com/guardian/support-frontend/blob/6de367509e0813e5aaae7852693e10d1f828b5e3/support-frontend/assets/components/forms/customFields/error.jsx#L35
It is a hack that was put in place to handle the Covid message without having to do a major refactor, with any luck we can remove it from both places in the not too distant future.

support-frontend/app/controllers/RedemptionController.scala
rupertbates: 👍  we can work this out when we integrate the 2 PRs

support-frontend/assets/pages/subscriptions-redemption/subscriptionsRedemption.jsx
```diff
@@ -0,0 +1,77 @@
...
+            paperFulfilmentOptions={null}
+          />
+          <SubscriptionFaq subscriptionProduct="DigitalPack" />
+        </Footer>}
+    >
+      <CheckoutStage
+        checkoutForm={<RedemptionForm />}
+        thankYouContentPending={
+          <ThankYouPendingContent
+            countryGroupId={countryGroupId}
```
GHaberis, line 53: similar to my  comment elsewhere There `<ThankYouPendingContent>` and `<ThankYouContent>` components have the same injected props, to remove some repetition here you could move these to an Object as so:

```js
const thankYouProps = {
  countryGroupId,
  marketingConsent: (
    <MarketingConsent
      render={({ title, message }) => <Text title={title}>{message}</Text>}
    />
  ),
};
```

and spread this object's properties in the Components:

```jsx
<ThankYouPendingContent {...thankYouProps} />
<ThankYouContent {...thankYouProps} />
```

support-frontend/assets/pages/digital-subscription-checkout/thankYouContent.jsx
```diff
@@ -2,33 +2,44 @@
...
 
-import React from 'react';
-import { connect } from 'react-redux';
+import * as React from 'react';
 
 import { type CountryGroupId } from 'helpers/internationalisation/countryGroup';
 
 import Content from 'components/content/content';
 import Text, { LargeParagraph } from 'components/text/text';
-import MarketingConsent from 'components/subscriptionCheckouts/thankYou/marketingConsentContainer';
```
GHaberis, line 12: I'm having trouble following why we're now passing `marketingConsent` as prop into `thankYouContent.jsx` and `thankYouPendingContent.jsx` now rather than importing as before?

support-frontend/app/controllers/RedemptionController.scala
```diff
@@ -80,6 +82,24 @@ class RedemptionController(
       })
   }
 
+  def displayError(redemptionCode: RedemptionCode, error: String)(implicit request: AuthRequest[Any]): Result = {
+    SafeLogger.error(scrub"An error occurred while trying to process redemption code - ${redemptionCode}. Error was - ${error}")
```
johnduffell, line 86: we should also do a cloudwatch putMetric at this point so we know it's broken.  We have one already for warnings (non critical failures on server side) so it might be worth piggy backing on that

support-frontend/assets/pages/digital-subscription-checkout/thankYouContent.jsx
rupertbates: Because we have to map from a different redux state to the marketing consent props depending on where the control is used.

support-frontend/app/controllers/RedemptionController.scala
rupertbates: Have changed the way this works

support-frontend/assets/pages/subscriptions-redemption/api.js
```diff
@@ -0,0 +1,138 @@
...
+}
+
+function validate(userCode: string) {
+  const validationUrl = `${getOrigin()}/subscribe/redeem/validate/${userCode}`;
+  return fetchJson(validationUrl, {});
+}
+
+function doValidation(userCode: string, dispatch: Dispatch<Action>) {
+  validate(userCode).then((result: ValidationResult) => {
+    dispatch({ type: 'SET_ERROR', error: result.errorMessage });
```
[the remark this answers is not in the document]
rupertbates: Yes because that's how any previous error is cleared from state

support-frontend/assets/pages/subscriptions-redemption/api.js
```diff
@@ -0,0 +1,138 @@
...
+    routes.subscriptionCreate,
+    data,
+    participations,
+    csrf,
+    () => {},
+    () => {},
+  ).then(handleSubscribeResult);
+}
+
+export { doValidation, validateUserCode, submitCode, createSubscription };
```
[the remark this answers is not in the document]
rupertbates: Have removed the export

support-frontend/assets/components/forms/customFields/validated.jsx
```diff
@@ -0,0 +1,72 @@
...
+    <div className={getClassName(error, valid)}>
+      {children && children}
+      <Element
+        aria-hidden={!error}
+        aria-atomic="true"
+        aria-live="polite"
+        htmlFor={htmlFor}
+        className="component-form-error__error"
+      >
+        {(error === 'Temporary COVID message') && (
```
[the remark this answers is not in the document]
rupertbates: Interesting I hadn't noticed that, I copied it over from error.jsx but you're right it's not needed

support-frontend/app/controllers/RedemptionController.scala
```diff
@@ -0,0 +1,181 @@
...
+            submitted = false
+          ))
+      )
+
+  }
+
+  def displayThankYou(stage: String): Action[AnyContent] = authenticatedAction(subscriptionsClientId).async {
+    implicit request: AuthRequest[Any] =>
+      identityService.getUser(request.user.minimalUser).fold(
+        error => thankYouPage(stage, None),
```
johnduffell, line 75: re changed the way this works, looks good
```suggestion
        error => {
          SafeLogger.error(scrub"failed to show user on thank you page. Error was - ${error}")
          thankYouPage(stage, None)
        },
```
don't thint it's critical enough to send it to the urgent 9-5 cloudwatch alarm, but make your own mind up based on what the effect woudl be.

support-frontend/test/utils/CheckoutValidationRulesTest.scala
rupertbates: I've added a test to check that only digisubs can use corporate redemptions

support-frontend/app/controllers/RedemptionController.scala
rupertbates: Discussed offline, we're going to revisit tracking and alarms in a separate PR

support-frontend/assets/components/forms/customFields/validated.jsx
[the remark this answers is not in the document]
johnduffell: @GHaberis the postcode message was discussed in my comment just below - result was "It is a hack that was put in place to handle the Covid message without having to do a major refactor, with any luck we can remove it from both places in the not too distant future."

PR #2526, opened 2020-06-08
Check redemption code validity in step functions

johnduffell:
## Why are you doing this?

Following on in the corporate codes work, we need to check the codes in support workers to make sure they are stil valid.
We also need to mark them as redeemed after the sub is created in zuora.

[**Trello Card**](https://trello.com/c/bth1aQ7O/3100-mark-corporate-codes-as-redeemed-when-checkout-done)


support-config/src/main/scala/com/gu/support/config/ZuoraConfigProvider.scala
```diff
@@ -2,6 +2,7 @@ package com.gu.support.config
 
 
 import com.gu.support.catalog.{ProductRatePlanChargeId, ProductRatePlanId}
+import com.gu.support.workers.BillingPeriod.Annual
```
johnduffell, line 5: I moved the objects inside BillingPeriod as `com.gu.support.workers.Monthly` didn't make so much sense, noticed  when I was writing the tests

support-models/src/main/scala/com/gu/support/workers/redemption/RedemptionData.scala
```diff
@@ -6,18 +6,9 @@ import com.gu.support.zuora.api.{ReaderType, Subscription}
 import io.circe.syntax._
 import io.circe.{Decoder, Encoder}
 
-abstract class RedemptionData(val redemptionCode: RedemptionCode){
-  def redeem(subscription: Subscription) = subscription.copy(redemptionCode = Some(redemptionCode))
```
johnduffell, line 10: the redeem functionality has been moved out to keep the data model clean

support-services/src/main/scala/com/gu/support/redemption/RedemptionCode.scala
```diff
@@ -0,0 +1,13 @@
+package com.gu.support.redemption
+
+object RedemptionCode {
+  def apply(codeAsString: String): Either[String, RedemptionCode] = {
+    // make sure no one can inject anything bad
```
johnduffell, line 5: actually, the IT test proves that dynamo client can handle all the characters, but even so, it makes sense to parse the redemption code string into the types as early as possible, so we can assume it's correct from the type.

support-services/src/test/scala/com/gu/support/redemption/DynamoTableAsyncITSpec.scala
```diff
@@ -0,0 +1,37 @@
...
+import org.scalatest.flatspec.AsyncFlatSpec
+import org.scalatest.matchers.should.Matchers
+
+@IntegrationTest
+class DynamoTableAsyncITSpec extends AsyncFlatSpec with Matchers {
+
+  val table: DynamoTableAsync = RedemptionTable.forEnvAsync(TouchPointEnvironments.SANDBOX)
+
+  "lookup" should "handle all chars" in {
+    val crazyCode = """ITTEST- !"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\]^_`abcdefghijklmnopqrstuvwxyz{|}~"""
```
johnduffell, line 16: didn't think about it at the time, but maybe we should also add some codepoints outside of standard ASCII e.g. emojis and non latin charset

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -114,61 +121,72 @@ class CreateZuoraSubscription(servicesProvider: ServiceProvider = ServiceProvide
...
         state.requestId,
         state.user.billingAddress.country,
         state.promoCode,
         state.firstDeliveryDate,
         promotionService,
-        readerType,
         stage,
         isTestUser
-      )
+      ).leftMap(_.msg))
```
johnduffell, line 169: we used to use `toString` but now we do `msg`.  The toString would have something like "NotApplicable" wheras the message is something like "This promotion is not applicable"  I hope this won't break something and hopefully it would be more friendly.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -38,8 +43,10 @@ class CreateZuoraSubscription(servicesProvider: ServiceProvider = ServiceProvide
     services: Services
   ): FutureHandlerResult = {
     val now = () => OffsetDateTime.now
+    val now2 = () => LocalDate.now(DateTimeZone.UTC)
```
johnduffell, line 46: unfortunately we have an old style joda time, and a new style java time.  We should probably switch away from joda now that the built in one does it all.

support-workers/src/main/scala/com/gu/zuora/ProductSubscriptionBuilders.scala
```diff
@@ -26,38 +32,38 @@ object ProductSubscriptionBuilders {
...
-
+      isTestUser: Boolean,
+      getCodeStatus: GetCodeStatus,
+      now: () => LocalDate
+    )(implicit ec: ExecutionContext): EitherT[Future, String, SubscriptionData] = {
+
+      val contractEffectiveDate = now()
+      val delay = paymentMethod match {
+        case Left((config, _)) => config.defaultFreeTrialPeriod + config.paymentGracePeriod
+        case Right(_) => 0
```
johnduffell, line 64: this looks a bit messed up because I used an Either and Tuple2 rather than going full custom ADT. (Right is info needed for corporate, Left is the info needed for standard subs)
I realised the types were not specific enough when I was trying to unit test it, as I would have had to make up loads of irrelevant data.  Happy to go for a custom ADT or other solution to make things more clear in the pattern matching.

support-workers/src/main/scala/com/gu/zuora/ProductSubscriptionBuilders.scala
```diff
@@ -68,53 +74,74 @@ object ProductSubscriptionBuilders {
         contractEffectiveDate = contractEffectiveDate
       )
 
-      val redeemedSub = paymentMethod.fold(
-        _ => subscriptionData.subscription,
-        redemptionData => redemptionData.redeem(subscriptionData.subscription)
+      paymentMethod.fold({ case (_, maybePromoCode) =>
```
johnduffell, line 77: although fold is more high level, I think a pattern match might be clearer here.

support-workers/src/main/scala/com/gu/zuora/ProductSubscriptionBuilders.scala
```diff
@@ -68,53 +74,74 @@ object ProductSubscriptionBuilders {
...
-      val redeemedSub = maybeRedemptionData.map(_.redeem(subscriptionData.subscription)).getOrElse(subscriptionData.subscription)
-      subscriptionData.copy(subscription = redeemedSub)
+    def withRedemption(
+      subscription: Subscription,
+      redemptionData: RedemptionData,
+      getCodeStatus: GetCodeStatus
+    )(implicit ec: ExecutionContext): EitherT[Future, GetCodeStatus.RedemptionInvalid, Subscription] = {
+      val withCode = subscription.copy(redemptionCode = Some(redemptionData.redemptionCode))
+      redemptionData match {
+        case CorporateRedemption(redemptionCode, _) =>
```
johnduffell, line 97: here we ignore the corporate id in the payload in favour of the one that comes back from getCodeStatus.  I am not sure if the one in the payload is trusted or whether it came from the client side?

support-workers/src/test/scala/com/gu/zuora/BuildDigitalPackSubscriptionSpec.scala
```diff
@@ -0,0 +1,102 @@
...
+          corporateAccountId = None
+        )
+      )
+    }
+
+  lazy val promotionService = mock[PromotionService]
+  lazy val saleDate = new LocalDate(2020, 6, 5)
+
+  lazy val corporate = buildDigitalPackSubscription(
+    DigitalPack(GBP, null /* FIXME should be Option-al for a corp sub */ , Corporate),
```
johnduffell, line 74: we specifically don't check the billing period when we look up the rate plan for corporate, so presumably the field should be removed rather than set arbitrarily?  but that's probably one for a separate PR

support-workers/src/test/scala/com/gu/zuora/BuildDigitalPackSubscriptionSpec.scala
```diff
@@ -0,0 +1,102 @@
...
+
+  lazy val monthly = buildDigitalPackSubscription(
+    DigitalPack(GBP, Monthly),
+    UUID.fromString("f7651338-5d94-4f57-85fd-262030de9ad5"),
+    Country.UK,
+    Left((ZuoraDigitalPackConfig(14, 2), None)),
+    promotionService,
+    DEV,
+    isTestUser = false,
+    new GetCodeStatus(_ => Future.failed(new Throwable)),
```
johnduffell, line 98: technically the GetCodeStatus is only relevant to corporate subs, so would ideally only be included somehow in the `Right` rather than having to be also specified for `Left` standard subs.
But I'd rather keep the side effects separate, and I'd also rather not do anything too fancy with the types to keep them separate but checked together.  Would certainly end up involving implicits.
Other ideas welcome though.

support-config/src/main/scala/com/gu/support/config/ZuoraConfigProvider.scala
rupertbates: I do think there is value in refactoring but at the same time this PR has 50 files changed for what is actually a very small piece of functionality. Also quite often refactoring can be more about changing code to fit our own personal preferences for code structure rather than any actual tangible benefit. 
I would rather assume that the person who originally wrote the code had a reason for doing it that way and work with the existing structure, hopefully resulting in a more consistent codebase.

support-services/src/main/scala/com/gu/support/redemption/RedemptionCode.scala
```diff
@@ -0,0 +1,13 @@
...
+    val validChars = (('A' to 'Z') ++ ('0' to '9') ++ Seq('-')).toSet
+    if (codeAsString.forall(validChars.contains))
+      Right(new RedemptionCode(codeAsString))
+    else
+      Left(s"redemption code must only include ${validChars.mkString("")}")
+  }
+}
+case class RedemptionCode private (codeAsString: String) extends AnyVal
```
rupertbates, line 13: I would say `value` is more consistent with other classes in this codebase than `codeAsString`

support-workers/src/main/scala/com/gu/support/workers/exceptions/ErrorHandler.scala
```diff
@@ -1,39 +1,86 @@
...
+
+      //Invalid Json
+      case e@(_: ParsingFailure | _: MatchError) =>
+        new RetryNone(message = e.getMessage, cause = throwable)
+
+      //Any Exception that we haven't specifically handled
+      case e: Throwable => new RetryLimited(message = e.getMessage, cause = throwable)
+    }
+
+  def fromStripeError(throwable: StripeError): RetryException = throwable.`type` match {
```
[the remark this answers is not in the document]
rupertbates: I think this is another example of where we would be better leaving alone, the change doesn't have anything to do with the purpose of this PR. If the change is worth doing then make it in a separate PR.

support-workers/src/main/scala/com/gu/services/Services.scala
```diff
@@ -30,5 +32,6 @@ class Services(isTestUser: Boolean) {
   lazy val acquisitionService = AcquisitionServiceBuilder.build(isTestUser)
   lazy val promotionService = new PromotionService(promotionsConfigProvider.get(isTestUser))
   lazy val goCardlessService = new GoCardlessWorkersService(goCardlessConfigProvider.get(isTestUser))
+  lazy val redemptionService = RedemptionTable.forEnvAsync(TouchPointEnvironments.fromStage(stage, isTestUser))
```
rupertbates, line 35: Can you use a config provider approach here similar to the 2 lines above for consistency?

support-config/src/main/scala/com/gu/support/config/ZuoraConfigProvider.scala
johnduffell: true, that certainly fits in with chestertons fence https://fs.blog/2020/03/chestertons-fence/
i.e. don't change something until you understand why it was done that way.

I did just assume that it was an oversight to have multiple "Monthly" classes in support-models with no positive distinction - here actually the second one is monthly Billing Period (this screenshot is master)
image

However I will remove this as I see that it makes it hard to review this code even if we do want to change it.

support-services/src/main/scala/com/gu/support/redemption/RedemptionCode.scala
johnduffell: I used to do the same until recently, but then I realised if we call them all `value`, then it's going the same way as the ubiquitous `toString` that will compile regardless, but won't always do what you expect.  I don't feel as strongly as for toString though so happy to change it.

support-workers/src/main/scala/com/gu/services/Services.scala
johnduffell: sure, I only didn't because it seemed unneccesary to have a class for one line of code, but will do

johnduffell: it looks like the config provider system is designed to get the configuration from AWS store, but RedemptionTable is more like the AcquisitionServiceBuilder above in that it doesn't actually need any config, just needs to know the env.  So I propose to add the config provider as/when we need config.

support-workers/src/main/scala/com/gu/zuora/ProductSubscriptionBuilders.scala
```diff
@@ -26,40 +31,51 @@ object ProductSubscriptionBuilders {
...
+      getCodeStatus: GetCodeStatus
+    ) extends SubscriptionPaymentType
 
+    def apply(
+      digitalPack: DigitalPack,
+      requestId: UUID,
+      subscriptionPaymentType: SubscriptionPaymentType,
+      environment: TouchPointEnvironment,
+      now: () => LocalDate
+    )(implicit ec: ExecutionContext): EitherT[Future, Either[PromoError, GetCodeStatus.RedemptionInvalid], SubscriptionData] = {
```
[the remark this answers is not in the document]
rupertbates: I think this is fine for now, it obviously isn't extensible any further than these two types but we can cross that bridge when necessary. This is when Flow's `PromoError | RedemptionInvalid` would be really nice.

support-workers/src/main/scala/com/gu/zuora/ProductSubscriptionBuilders.scala
rupertbates: I believe it comes from the client so best to use the one from getCodeStatus

support-workers/src/test/scala/com/gu/zuora/BuildDigitalPackSubscriptionSpec.scala
rupertbates: Yes this would be better but would require a lot of knock on changes

support-workers/src/main/scala/com/gu/zuora/ProductSubscriptionBuilders.scala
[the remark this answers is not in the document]
johnduffell: practically speaking agree - maybe scala 3 will be out by then so we can use union types.

mathematically it's extensible as far as you like - you just use sum types recursively, but that's like implementing your own linked like, i.e. 
```scala
Either[PromoError, Either[GetCodeStatus.RedemptionInvalid, AnotherError]]
```

PR #2532, opened 2020-06-09
Error tidy up

rupertbates:
## Why are you doing this?
This PR change the `error` component used on the subs checkouts to use the `InlineError` component from [Source](https://github.com/guardian/source) which provides us with the correct icon and styling.

It also moves the error message above the form field with the error in line with design system guidelines.

Finally I have refactored to remove the 'Temporary Covid message' hack which we had in place.

### Previous error component
Screen Shot 2020-06-09 at 17 57 52

### New error component
Screen Shot 2020-06-09 at 17 55 35

### DIFF
--- a/support-frontend/assets/components/forms/customFields/error.jsx
+++ b/support-frontend/assets/components/forms/customFields/error.jsx
@@ -5,41 +5,35 @@
 import React, { type Node } from 'react';
 
 import { type Option } from 'helpers/types/option';
+import { InlineError } from '@guardian/src-inline-error';
 
 import './error.scss';
 
 // ----- Types ----- //
 export type PropsForHoc = {
-  error: Option<string>,
+  error: Option<string> | Option<Node>,
 };
 
 type Props = PropsForHoc & {
-  htmlFor: Option<string>,
   children?: Option<Node>,
 };
 
 // ----- Component ----- //
 
-function Error({ error, htmlFor, children }: Props) {
-  const Element = htmlFor ? 'label' : 'div';
+function Error({ error, children }: Props) {
+  const errorNode = error === 'Temporary COVID message' ? (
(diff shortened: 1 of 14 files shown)


support-frontend/assets/components/subscriptionCheckouts/address/addressFields.jsx
```diff
@@ -53,9 +53,7 @@ type PropTypes<GlobalState> = {|
   ...StatePropTypes<GlobalState>,
 |}
 
-const StaticInputWithLabel = withLabel(Input);
-const InputWithLabel = asControlled(StaticInputWithLabel);
-const InputWithError = withError(InputWithLabel);
+const InputWithError = compose(asControlled, withLabel, withError)(Input);
```
lucymonie, line 56: This seems much cleaner! Nice

support-frontend/assets/components/forms/customFields/error.jsx
```diff
@@ -7,39 +7,24 @@ import React, { type Node } from 'react';
...
 // ----- Component ----- //
 
-function Error({ error, htmlFor, children }: Props) {
-  const Element = htmlFor ? 'label' : 'div';
+function Error({ error, children }: Props) {
   return (
     <div className={error ? 'component-form-error' : null}>
-      {children && children}
-      <Element
-        aria-hidden={!error}
```
lucymonie, line 29: I'm assuming that the removal of the aria tags here is okay since these messages aren't on that page at all unless there's an error, therefore aren't going to confuse a screen-reader user.

rupertbates: I've discussed this with Simon A and we believe it is fine, he is doing some more investigation and I'll update if needed.

PR #2533, opened 2020-06-10
Fix getaddress.io error handling

Jpsi:
# Why?

- **Current situation**: We use the service getAddressIO to look up addresses from a postcode on all of our checkouts.

- **Problem**: This service has changed their response so we are misinterpreting invalid postcode errors as server errors, which trigger 5XX alarms if too many of them happen in quick succession.

- **Proposed change**: Fix getAddressIO error handling

- **Expected benefit**: Show correct error message to user and stop the alarm.

# How?

- Interpret the error type rather than message body - it looks like the service returns a different message now.
- [**Trello Card**](https://trello.com/c/c64aWiPH)

# Screenshots

## Before
Screenshot 2020-06-10 at 01 02 02

## After
Screenshot 2020-06-10 at 00 58 37

### DIFF
--- a/support-frontend/app/controllers/GetAddress.scala
+++ b/support-frontend/app/controllers/GetAddress.scala
@@ -7,6 +7,7 @@ import com.gu.support.getaddressio.GetAddressIOService
 import io.circe.syntax._
 import play.api.libs.circe.Circe
 import play.api.mvc.{AbstractController, Action, AnyContent, ControllerComponents}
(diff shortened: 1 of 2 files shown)


johnduffell:
> Interpret the error type rather than message body - it looks like the service returns a different message now.

That was down to changes I made to the Web service helper a while ago and I didn't realise anything was relying on this.  If we weren't moving away from getaddress, I'd suggest adding a test.

PR #2535, opened 2020-06-11
Script for identifying "orphan" .js/.jsx/.scss modules

GHaberis:
## Why are you doing this?

Whilst working on the repo I've come across some modules which are present in the code, but aren't dependencies of any other modules (and they also aren't entry points for the generated bundles). They were probably depended upon once but are no longer in use now. 

It's not great developer experience to keep these in the codebase as these modules create additional noise when searching through the code, and developers then have to spend time working out if and how they're used. It also leaves us maintaining redundant code.

I've therefore written a little script using the follow NPM module https://www.npmjs.com/package/madge

`madge` is a developer tool that traverses a dependency tree and can do neat things like produce a dependency map, identify circular dependencies and [identify orphan modules](https://www.npmjs.com/package/madge#orphans)! It works for JavaScript (AMD, CommonJS, and ES6 modules) and for CSS preprocessors (Sass, Stylus, and Less).

## Changes

- I've introduced a new `js` script in `scripts/orphans.js`. This script queries the `assets` directory and identifies any `.js`/`.jsx`/`.scss` files which aren't imported elsewhere. The script exclues "entry points" from the orphans list. The entry points are  used by `Webpack` as starting poits when building `.js`/`.css` bundles, and I'd therefore not expect them to be dependencies. The entry points are imported from a new module `support-frontend/webpack.entryPoints.js`, which is now also used by our webpack config when building bundles. 
- I've included a new script `orphans` in the `package.json`. That can be run directly with the command line command `yarn orphans`.
- I've also included this as part of the `yarn validate` step.
- When we run the script it logs a list of orphans to the command line, it doesn't prevent the build from running or fail validation.

## Screenshots

orphans

## Next steps

This script should help us maintain the codebase, when we want to remove orphan modules (redundant code). It doesn't do anything other than identify orphans right now. I expect next steps will be to look at the results and start tidying the code and removing orphans manually. 

I've cross checked a number of the results with the codebase and the orphans it's identified are indeed not referenced anywhere else in the code!

### DIFF
--- a/support-frontend/package.json
+++ b/support-frontend/package.json
@@ -4,7 +4,7 @@
   "description": "Frontend for the supporter platform.",
   "scripts": {
     "clean": "echo 'Cleaning Public Folder' && rimraf public/compiled-assets",
-    "validate": "echo 'Validating JS' && npm-run-all 'lint --fix' flow",
+    "validate": "echo 'Validating JS' && npm-run-all 'lint --fix' flow && yarn orphans",
     "flow": "flow",
     "lint": "eslint --ext .js --ext .jsx .",
     "build-dev": "webpack --config webpack.dev.js",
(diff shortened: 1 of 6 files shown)


GHaberis:
@rupertbates thanks for looking, I think this just needs an approval before I can merge now.

PR #2538, opened 2020-06-11
Prevent users with an existing Digital Subscription signing up for a corporate subscription

rupertbates:
## Why are you doing this?
We want to prevent users who already have a live Digital Subscription from also signing up for a corporate subscription. 

This PR takes the same approach as a regular digisub which is to redirect the user to this page:

Screen Shot 2020-06-11 at 17 35 19

We may want to do something more tailored to redemptions at some point but this is what we have agreed on for the proof of concept.

[**Trello Card**](https://trello.com/c/DH2PVZ9U/3097-handle-users-with-existing-digital-subscriptions)

### DIFF
--- a/support-frontend/app/controllers/RedemptionController.scala
+++ b/support-frontend/app/controllers/RedemptionController.scala
@@ -122,7 +122,7 @@ class RedemptionController(
   def displayProcessing(redemptionCode: String): Action[AnyContent] =
     authenticatedAction(subscriptionsClientId).async {
       implicit request: AuthRequest[Any] =>
-        val processingPage: EitherT[Future, String, Result] = for {
+        val processingPage = for {
           user <- identityService.getUser(request.user.minimalUser)
           corporateCustomer <- getCorporateCustomer(redemptionCode)
           hasDigipack <- EitherT.right(userHasDigitalSubscription(membersDataService, request.user))
--- a/support-frontend/app/controllers/RedemptionController.scala
+++ b/support-frontend/app/controllers/RedemptionController.scala
@@ -13,7 +13,7 @@ import controllers.RedemptionController._
 import io.circe.syntax._
 import play.api.libs.circe.Circe
 import play.api.mvc.{AbstractController, Action, AnyContent, ControllerComponents, Result}
-import services.{IdentityService, TestUserService}
+import services.{AccessCredentials, AuthenticatedIdUser, IdentityService, MembersDataService, TestUserService}
 import views.EmptyDiv
 import views.html.helper.CSRF
 import views.html.subscriptionRedemptionForm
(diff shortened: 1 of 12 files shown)


support-frontend/app/controllers/UserDigitalSubscription.scala
```diff
@@ -0,0 +1,37 @@
...
+import actions.CustomActionBuilders.AuthRequest
+import com.gu.monitoring.SafeLogger
+import play.api.mvc.Result
+import play.api.mvc.Results._
+import play.mvc.Http.Status.FOUND
+import services.{AccessCredentials, AuthenticatedIdUser, MembersDataService}
+import SafeLogger._
+import scala.concurrent.{ExecutionContext, Future}
+
+trait UserDigitalSubscription {
```
johnduffell, line 12: why not just make this a standalone function rather than using traits?

support-frontend/app/controllers/RedemptionController.scala
```diff
@@ -121,10 +122,14 @@ class RedemptionController(
   def displayProcessing(redemptionCode: String): Action[AnyContent] =
     authenticatedAction(subscriptionsClientId).async {
       implicit request: AuthRequest[Any] =>
-        val processingPage = for {
+        val processingPage: EitherT[Future, String, Result] = for {
           user <- identityService.getUser(request.user.minimalUser)
           corporateCustomer <- getCorporateCustomer(redemptionCode)
-        } yield showProcessing(redemptionCode, corporateCustomer, user)
+          hasDigipack <- EitherT.right(userHasDigitalSubscription(membersDataService, request.user))
```
johnduffell, line 128: will be ok, but we could do the check first, then we don't depend on the calls to identity and dynamo

support-frontend/app/controllers/UserDigitalSubscription.scala
```diff
@@ -0,0 +1,37 @@
...
+        membersDataService.userAttributes(cookies).value map {
+          case Left(err: MembersDataService.MembersDataServiceError) => {
+            SafeLogger.error(scrub"Got an error from members-data-api ${err}")
+            false
+          }
+          case Right(response: MembersDataService.UserAttributes) => response.contentAccess.digitalPack
+        }
+      case _ => {
+        SafeLogger.warn(
+          s"Couldn't call members-data-api to check subscription status for user ${user.minimalUser.id} because their access cookies were missing"
```
johnduffell, line 28: good idea to add the logging for the false cases (although I still doubt we'd notice if it started happening)

support-frontend/app/controllers/RedemptionController.scala
rupertbates: Good suggestion 👍

PR #2539, opened 2020-06-11
consolidate step function error/retry decision logic into one place

johnduffell:
## Why are you doing this?

The retry logic trips me up every time I change it, and the several similar asRetryException methods, including one on any Throwable, mean that breakages aren't always apparent unless there are good tests.
There are two big match statements where one chains onto the other, and various exceptions have been added to each, with no particular distinction.

This PR just moves all the match statement together, and puts normal functions for the various conversions, rather than implicit classes.

This is a precursor to adding another exception to the list (for a failure to look up redemption codes.)


support-workers/src/main/scala/com/gu/support/workers/exceptions/RetryImplicits.scala
```diff
@@ -1,83 +0,0 @@
...
-
-  implicit class StripeConversions(val throwable: StripeError) extends AnyVal {
-    def asRetryException: RetryException = throwable.`type` match {
-      case "api_connection_error" | "api_error" | "rate_limit_error" => new RetryUnlimited(throwable.asJson.noSpaces, cause = throwable)
-      case "authentication_error" => new RetryLimited(throwable.asJson.noSpaces, cause = throwable)
-      case "card_error" | "invalid_request_error" | "validation_error" => new RetryNone(throwable.asJson.noSpaces, cause = throwable)
-    }
-  }
-
-  implicit class AwsSQSConversions(val throwable: AmazonSQSException) extends AnyVal {
```
johnduffell, line 57: this and the two implicits below were not actually used.  I am wondering if that is because someone assumed they would automatically get picked up by having the right named method.
I could add them to the match statement, but since it hasn't cropped up since, leaving the behaviour as it is seems more sensible.  Happy to go either way though.

PR #2540, opened 2020-06-12
make product subscription builders into normal functions

johnduffell:
## Why are you doing this?

The subscription builders are using implicit classes and traits which adds boilerplate and difficulty in navigating outside of an IDE.  This PR turns them into normal functions, and injects the date rather than depending on it directly, to make the code more testable.

precursor to https://github.com/guardian/support-frontend/pull/2526

### DIFF
--- a/support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
+++ b/support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
@@ -37,14 +37,14 @@ class CreateZuoraSubscription(servicesProvider: ServiceProvider = ServiceProvide
     context: Context,
     services: Services
   ): FutureHandlerResult = {
-    val now = () => OffsetDateTime.now
-    val now2 = () => LocalDate.now(DateTimeZone.UTC)
+    val nowJavaTime = () => OffsetDateTime.now
+    val todayJodaDate = () => LocalDate.now(DateTimeZone.UTC)
     for {
-      subscriptionData <- buildSubscriptionData(state, services.promotionService, now2)
+      subscriptionData <- buildSubscriptionData(state, services.promotionService, todayJodaDate)
       subscribeItem = buildSubscribeItem(state, subscriptionData)
       identityId <- Future.fromTry(IdentityId(state.user.id))
         .withLogging("identity id")
-      maybeDomainSubscription <- GetSubscriptionWithCurrentRequestId(services.zuoraService, state.requestId, identityId, state.product.billingPeriod, now)
+      maybeDomainSubscription <- GetSubscriptionWithCurrentRequestId(services.zuoraService, state.requestId, identityId, state.product.billingPeriod, nowJavaTime)
           .withLogging("GetSubscriptionWithCurrentRequestId")
       previewPaymentSchedule <- PreviewPaymentSchedule(subscribeItem, state.product.billingPeriod, services, checkSingleResponse)
           .withLogging("PreviewPaymentSchedule")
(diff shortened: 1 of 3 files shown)


support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -111,61 +116,64 @@ class CreateZuoraSubscription(servicesProvider: ServiceProvider = ServiceProvide
...
-      case _ => ReaderType.Direct
-    }
-    val stage = Configuration.stage
+    val environment = TouchPointEnvironments.fromStage(Configuration.stage, isTestUser)
 
-    state.product match {
-      case c: Contribution => Right(c.build(state.requestId, config))
-      case d: DigitalPack => d.build(
+    val eventualErrorOrSubscriptionData = state.product match {
+      case c: Contribution => EitherT.pure[Future, Throwable](buildContributionSubscription(c, state.requestId, config))
```
johnduffell, line 142: it's slightly premature to make it into EitherT rather than just Either, as we don't use Futures yet, but I thought it would simplify the next PR to remove all that types noise.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -35,8 +38,10 @@ class CreateZuoraSubscription(servicesProvider: ServiceProvider = ServiceProvide
     services: Services
   ): FutureHandlerResult = {
     val now = () => OffsetDateTime.now
+    val now2 = () => LocalDate.now(DateTimeZone.UTC)
```
tomrf1, line 41: please explain!

johnduffell: sorry forgot to add the explanation here - it's because the existing `now` is a java one, wheras the one we use is a joda one.  Technically joda is not needed any more as the functionality is in the java library now, but I wasn't sure whether I should change the new one to use the java time, or the old one to use a joda time, or neither.  Could be one for a health task one way or the other.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
tomrf1: perhaps it's time for zio :)

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -34,12 +37,14 @@ class CreateZuoraSubscription(servicesProvider: ServiceProvider = ServiceProvide
     context: Context,
     services: Services
   ): FutureHandlerResult = {
-    val now = () => OffsetDateTime.now
+    val nowJavaTime = () => OffsetDateTime.now
```
rupertbates, line 40: How much work would it be to change the uses of java time over to Joda (or the other way but I think that would be a lot more work)?

johnduffell: just to get an idea -
java to joda would be 7 files
getting rid of joda would be 52 files

It's fairly mechanical in either direction (the lead on the java time project was the joda-time person)
joda has no dependencies so it would Not be a massive artifact size win to remove it
users are officially "asked" to migrate to the Java 8 time apis as it's no longer in development
> Joda-time is no longer in active development except to keep timezone data up to date. From Java SE 8 onwards, users are asked to migrate to java.time (JSR-310) - a core part of the JDK which replaces this project.

my personal opinion if we're going to do either, just get onto the new api in one go as then it's cleaner, but either way not in this PR.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
johnduffell: I'd certainly be happy to consider it, or any other non "wtf" type of abstraction of effects!
image
Would be great to have a stronger guarantee of not accidentally pushing data to prod during IT tests, and simpler unit testability.
Not sure how we'd justify it as a quality objective though!

PR #2541, opened 2020-06-12
use joda time most places for consistency

johnduffell:
## Why are you doing this?

We noticed that joda and java time were mixed in the same codebase.
It makes sense to be consistent.
It would be slightly better to choose java time, as joda time has been superseded from java 8, but that would be a health task for another day.

There are two other places we use java time `ofDays`, but they are to call the test users library
https://github.com/guardian/support-frontend/blob/9cf8941e823860e17f409a1d12577ce7aa770fb4/support-frontend/test/selenium/util/PostDeployTestUser.scala#L15
https://github.com/guardian/support-frontend/blob/f00be2f4a2a495fb87025e34476125357e71984e/support-frontend/app/services/TestUserService.scala#L13
calling this function
https://github.com/guardian/identity-test-users/blob/461286656b7443b8b0e43e10b89426b68138f756/src/main/scala/com/gu/identity/testing/usernames/TestUsernames.scala#L27
which is probably owned by us but hasn't been touched in over 2 years.  Those can stay as they are.

### DIFF
--- a/support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
+++ b/support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
@@ -1,7 +1,5 @@
 package com.gu.support.workers.lambdas
 
-import java.time.OffsetDateTime
-
 import cats.data.EitherT
 import cats.implicits._
 import com.amazonaws.services.lambda.runtime.Context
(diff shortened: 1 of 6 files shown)


support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -44,14 +41,14 @@ class CreateZuoraSubscription(servicesProvider: ServiceProvider = ServiceProvide
     context: Context,
     services: Services
   ): FutureHandlerResult = {
-    val nowJavaTime = () => OffsetDateTime.now
-    val todayJodaDate = () => LocalDate.now(DateTimeZone.UTC)
+    val now = () => DateTime.now(DateTimeZone.UTC)
+    val today = () => LocalDate.now(DateTimeZone.UTC)
```
johnduffell, line 45: we still need a "today" and a "right now" in the end (unless we want to change the behaviour of the zuora sub deduplication)

support-workers/src/test/scala/com/gu/zuora/ZuoraITSpec.scala
```diff
@@ -0,0 +1,147 @@
+package com.gu.zuora
```
johnduffell, line 1: I even used the `git mv` function to try to keep the history but it didn't work!  This is a rename of ZuoraSpec, as I wanted an actual ZuoraSpec that wasn't an IT one.

support-workers/src/test/scala/com/gu/zuora/ZuoraITSpec.scala
```diff
@@ -0,0 +1,147 @@
...
+import org.scalatest.matchers.should.Matchers
+
+import scala.concurrent.duration._
+
+@IntegrationTest
+class ZuoraITSpec extends AsyncFlatSpec with Matchers {
+
+  def uatService: ZuoraService = new ZuoraService(zuoraConfigProvider.get(true), RequestRunners.configurableFutureRunner(30.seconds))
+
+  // actual sub "CreatedDate": "2017-12-07T15:47:21.000+00:00",
```
johnduffell, line 24: I added this one line to confirm the created date of the subscription it is targeting

JoemG:
💯  good work splitting the IT test out

johnduffell:
thanks for approval @JoemG 
Just for the record, I didn't split out the itegration test, I just renamed it to add `IT` to the name, and used the original name for a new test.
If `git mv` had done what I expected, it would have been clearer!

PR #2545, opened 2020-06-16
Mj aus moment social sharing referral codes

michaelbjacobson:
- For all Aus Moment contributions, a unique referral code is generated and logged to a specific user via Identity. The code is generated from a unix timestamp and a random number salt, which are then converted to Base36 and concatenated.

- To get around social media conditions on the number of query params allowed in a shared url, we use a single param for Aus Moment shares. The param is called ausAcquisitionsData and contains an underscore delimited list of data in a predefined order: `ausAcquisitionsData=<pagename>_<sharing platform>_<unique referral code>`

- If ausAcquisitionsData exists as a request param, then it is de-serialised by an Aus Moment specific function, and the data are mapped to Ophan keys like so:
```
{
    componentId: `${page}_${socialPlatform}`,
    source: 'SOCIAL',
    queryParameters: [
        {
            name: 'referralCode',
            value: referralCode,
        },
    ],
};
```

- If, however, this param does not exist for the request, we fall back on the standard deserialiser function.


support-frontend/assets/helpers/tracking/acquisitions.js
```diff
@@ -261,11 +262,22 @@ function deriveSubsAcquisitionData(
 
 }
 
+function deserialiseAusMomentAcquisitionData(serialised: string): Object {
+  const [page, socialPlatform, referralCode] = serialised.split('_');
+  return {
+    componentId: `${page}_${socialPlatform}`,
+    source: 'SOCIAL',
+    campaignCode: referralCode,
```
tomrf1, line 270: I've just realised an issue here. Elsewhere it overrides the `campaignCode` if the user has arrived on a campaign page. In this case that would be the /au/contribute contributions page.

tomrf1: I think we have 2 options for the referralCode here:
1. Include it in `componentId`, which would then just be the full `<page>_<platform>_<code>`
2. Add it to the `queryParameters` field - https://github.com/guardian/support-frontend/blob/master/support-frontend/assets/helpers/tracking/acquisitions.js#L46

@paulbrown1982 any thoughts on what's better from an analysis perspective?

support-frontend/assets/helpers/tracking/acquisitions.js
```diff
@@ -261,11 +262,22 @@ function deriveSubsAcquisitionData(
 
 }
 
+function deserialiseAusMomentAcquisitionData(serialised: string): Object {
```
tomrf1, line 265: return type can be `ReferrerAcquisitionData`

support-frontend/assets/components/socialShare/socialShare.jsx
```diff
@@ -40,25 +40,29 @@ const socialMedia: {
...
+    link: referralCodeQueryParameter =>
+      `https://twitter.com/intent/tweet?url=https%3A%2F%2Fsupport.theguardian.com%2Fcontribute%3FausAcquisitionData%3Dthankyou_twitter_${referralCodeQueryParameter || ''}&text=Join%20me%20and%20over%20one%20million%20others%20in%20supporting%20a%20different%20model%20for%20open%2C%20independent%20journalism.%20Together%20we%20can%20help%20safeguard%20The%20Guardian%E2%80%99s%20future%20%E2%80%93%20so%20more%20people%2C%20across%20the%20world%2C%20can%20keep%20accessing%20factual%20information%20for%20free`,
     svg: <SvgTwitter />,
     a11yHint: 'Share on twitter',
     windowFeatures: SocialWindowFeatures,
   },
   linkedin: {
-    link: 'http://www.linkedin.com/shareArticle?mini=true&url=https%3A%2F%2Fsupport.theguardian.com%2Fcontribute',
+    link: referralCodeQueryParameter =>
+      `http://www.linkedin.com/shareArticle?mini=true&url=https%3A%2F%2Fsupport.theguardian.com%2Fcontribute%3FausAcquisitionData%3Dthankyou_linkedin_${referralCodeQueryParameter || ''}`,
```
paulbrown1982, line 58: should we change this to https ?

michaelbjacobson: Good spot @paulbrown1982, I've made that change now 🙂

support-frontend/assets/helpers/tracking/acquisitions.js
michaelbjacobson: @tomrf1 @paulbrown1982 has a decision been reached about how to solve this problem?

tomrf1: The referralCode will need to be query-able, for the emailing.
To me having it in `queryParameters` seems more robust than sticking it on the end of the `componentId` string.
@paulbrown1982 what do you think?

support-frontend/assets/helpers/tracking/acquisitions.js
michaelbjacobson: @tomrf1 the result of calling this function or `deserialiseJsonObject` is stored as `paramData`. In turn, this is passed into `buildReferrerAcquisitionData`, which expects a generic object as its argument. As the new `deserialiseAusMomentAcquisitionData` function should be able to replace the default `deserialiseJsonObject` function, I wanted to make their signatures as similar as possible. 

Are you happy with all this or would you like me to change the return type still? It's totally doable, it just means a few changes in different places.

support-frontend/assets/helpers/tracking/acquisitions.js
paulbrown1982: The referralCode / shareCode should be inside `queryParameters`, and not go anywhere near `campaignCode`.

paulbrown1982: The `componentId` would be the call to action that the user clicked on to arrive at this pageview. In this case, it would be the Twitter link, so that should probably be `<page>_<platform>`.

michaelbjacobson: Ok @tomrf1 @paulbrown1982 , I've moved the referralCode into queryParams, see the updated PR description for more.

support-frontend/assets/helpers/tracking/acquisitions.js
```diff
@@ -261,11 +262,24 @@ function deriveSubsAcquisitionData(
...
+  const [page, socialPlatform, referralCode] = serialised.split('_');
+  return {
+    componentId: `${page}_${socialPlatform}`,
+    source: 'SOCIAL',
+    queryParameters: {
+      referralCode,
```
tomrf1, line 271: I think this needs to be structured differently?
See the `AcquisitionQueryParameters` type

support-frontend/assets/helpers/tracking/acquisitions.js
tomrf1: 👍

support-frontend/assets/helpers/tracking/acquisitions.js
michaelbjacobson: Thanks @tomrf1, I'm cross with myself for missing that. I've sorted it now and updated the PR description.

support-frontend/assets/helpers/tracking/acquisitions.js
```diff
@@ -174,9 +175,12 @@ function buildReferrerAcquisitionData(acquisitionData: Object = {}): ReferrerAcq
   const referrerPageviewId = acquisitionData.referrerPageviewId ||
     getQueryParameter('REFPVID');
 
-  // This was how referrer pageview id used to be passed.
-  const campaignCode =
-    getCampaignName() || acquisitionData.campaignCode || getQueryParameter('INTCMP');
+  // Override default campaign code assignment for Aus Moment 2020
```
michaelbjacobson, line 178: @tomrf1 are you happy with this?

support-frontend/assets/components/socialShare/socialShare.jsx
```diff
@@ -98,5 +101,10 @@ export default function SocialShare(props: PropTypes) {
       </p>
     </button>
   );
-
 }
+
+SocialShare.defaultProps = {
+  referralCode: '',
```
tomrf1, line 107: since the field is optional, should the default be `undefined`?

tomrf1: or can we just not have defaultProps here?

michaelbjacobson: ✅

PR #2546, opened 2020-06-16
add test for CreateZuoraSubscription digipack

johnduffell:
## Why are you doing this?

This PR adds a more comprehensive test for the create zuora subscription logic, this is in preparation for adding a test that checks the redemption code is correctly marked as used, which would not be possible otherwise without being an IT test.

### DIFF
--- a/support-workers/src/main/scala/com/gu/support/workers/PreviewPaymentSchedule.scala
+++ b/support-workers/src/main/scala/com/gu/support/workers/PreviewPaymentSchedule.scala
@@ -3,6 +3,7 @@ package com.gu.support.workers
 import com.gu.services.Services
 import com.gu.support.zuora.api.response.{Charge, PreviewSubscribeResponse}
 import com.gu.support.zuora.api.{PreviewSubscribeRequest, SubscribeItem}
+import com.gu.zuora.{ZuoraService, ZuoraSubscribeService}
 import org.joda.time.LocalDate
 
 import scala.concurrent.ExecutionContext.Implicits.global
@@ -13,7 +14,7 @@ object PreviewPaymentSchedule {
   def apply(
     subscribeItem: SubscribeItem,
     billingPeriod: BillingPeriod,
-    services: Services,
+    zuoraService: ZuoraSubscribeService,
     singleResponseCheck: Future[List[PreviewSubscribeResponse]] => Future[PreviewSubscribeResponse]
   ): Future[PaymentSchedule] = {
     val numberOfInvoicesToPreview: Int = billingPeriod match {
@@ -23,7 +24,7 @@ object PreviewPaymentSchedule {
       case SixWeekly => 2
     }
     singleResponseCheck(
-      services.zuoraService.previewSubscribe(PreviewSubscribeRequest.fromSubscribe(subscribeItem, numberOfInvoicesToPreview))
+      zuoraService.previewSubscribe(PreviewSubscribeRequest.fromSubscribe(subscribeItem, numberOfInvoicesToPreview))
     ).map(response => paymentSchedule(response.invoiceData.flatMap(_.invoiceItem)))
   }
 
--- a/support-workers/src/main/scala/com/gu/support/workers/PreviewPaymentSchedule.scala
+++ b/support-workers/src/main/scala/com/gu/support/workers/PreviewPaymentSchedule.scala
@@ -1,9 +1,8 @@
 package com.gu.support.workers
 
-import com.gu.services.Services
 import com.gu.support.zuora.api.response.{Charge, PreviewSubscribeResponse}
 import com.gu.support.zuora.api.{PreviewSubscribeRequest, SubscribeItem}
-import com.gu.zuora.{ZuoraService, ZuoraSubscribeService}
+import com.gu.zuora.ZuoraSubscribeService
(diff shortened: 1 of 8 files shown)


support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -43,19 +44,43 @@ class CreateZuoraSubscription(servicesProvider: ServiceProvider = ServiceProvide
...
+    val zuoraService = services.zuoraService
+    val isTestUser = state.user.isTestUser
+    val config: ZuoraConfig = zuoraConfigProvider.get(isTestUser)
+    doStepsFromDeps(state, requestInfo, now, today, promotionService, redemptionService, zuoraService, config)
+  }
+}
+object CreateZuoraSubscription {
+  import com.gu.FutureLogging._
+
+  def doStepsFromDeps(
```
johnduffell, line 58: this needs a better name, there's already servicesHandler (which is an arbitrary name), I needed another arbitrary name that somehow says that this does the same things, but has more limited parameters specific to this lambda

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -43,19 +44,43 @@ class CreateZuoraSubscription(servicesProvider: ServiceProvider = ServiceProvide
...
         .withLogging("identity id")
-      maybeDomainSubscription <- GetSubscriptionWithCurrentRequestId(services.zuoraService, state.requestId, identityId, state.product.billingPeriod, now)
-          .withLogging("GetSubscriptionWithCurrentRequestId")
-      previewPaymentSchedule <- PreviewPaymentSchedule(subscribeItem, state.product.billingPeriod, services, checkSingleResponse)
-          .withLogging("PreviewPaymentSchedule")
+      maybeDomainSubscription <- GetSubscriptionWithCurrentRequestId(zuoraService, state.requestId, identityId, state.product.billingPeriod, now)
+        .withLogging("GetSubscriptionWithCurrentRequestId")
+      paymentMethodWithPaymentSchedule <-
+        state.paymentMethod.leftMap(pm =>
+          PreviewPaymentSchedule(subscribeItem, state.product.billingPeriod, zuoraService, checkSingleResponse)
```
johnduffell, line 77: we only need to get the payment schedule for paid subs, this should make corporate subs execute quicker.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -137,10 +163,10 @@ class CreateZuoraSubscription(servicesProvider: ServiceProvider = ServiceProvide
     state: CreateZuoraSubscriptionState,
     promotionService: => PromotionService,
     getCodeStatus: => GetCodeStatus,
-    today: () => LocalDate
+    today: () => LocalDate,
+    config: ZuoraConfig
   ): Future[SubscriptionData] = {
     val isTestUser = state.user.isTestUser
-    val config = zuoraConfigProvider.get(isTestUser)
```
johnduffell, line 143: calling this went off to a global val and hit S3, hence the parameter

support-workers/src/main/scala/com/gu/zuora/ZuoraService.scala
```diff
@@ -18,8 +18,19 @@ import org.joda.time.format.ISODateTimeFormat
 
 import scala.concurrent.{ExecutionContext, Future}
 
+trait ZuoraSubscribeService {
```
johnduffell, line 21: only 4 methods seem to be used, so I pulled them out to simplify the tests

support-workers/src/test/scala/com/gu/support/workers/CreateZuoraSubscriptionStepsSpec.scala
```diff
@@ -0,0 +1,80 @@
...
+import com.gu.zuora.ZuoraSubscribeService
+import org.joda.time.{DateTime, LocalDate}
+import org.scalatest.flatspec.AsyncFlatSpec
+import org.scalatest.matchers.should.Matchers
+
+import scala.concurrent.Future
+
+class CreateZuoraSubscriptionStepsSpec extends AsyncFlatSpec with Matchers {
+
+  it should "create a Digital Pack standard (paid) subscription" in {
```
johnduffell, line 27: there will be a corporate subs test too in the next pr

support-workers/src/main/scala/com/gu/support/workers/PreviewPaymentSchedule.scala
```diff
@@ -13,7 +13,7 @@ object PreviewPaymentSchedule {
   def apply(
     subscribeItem: SubscribeItem,
     billingPeriod: BillingPeriod,
-    services: Services,
+    zuoraService: ZuoraSubscribeService,
```
johnduffell, line 16: this only uses "previewSubscribe" but it's not unit tested anyway. if we want to add unit test later we could use a single abstract method trait to make the tests clear.

PR #2548, opened 2020-06-16
mark redemption code as redeemed after sub created

johnduffell:
## Why are you doing this?

Part of corporate subs, we want to make sure once the subscription is created, we mark the code as successfully redeemed.
If there is a failure marking the sub as redeemed, the step will fail and retry, the repeat execution will check the code again and find it's not redeemed, but it should find the existing sub, skipSubscribe, and then successfully mark as redeemed on the retry hopefully.

[**Trello Card**](https://trello.com/c/YHcNJ2yF/3105-mark-corporate-codes-as-redeemed-when-checkout-done)

### DIFF
--- a/support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
+++ b/support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
@@ -81,7 +81,7 @@ object CreateZuoraSubscription {
         case None => subscribe(subscribeItem, zuoraService).map(response => toHandlerResult(state, response, paymentMethodWithPaymentSchedule, requestInfo))
           .withLogging("subscribe")
       }
-      _ <- updateRedemptionCode(state.paymentMethod, subscriptionData, SetCodeStatus.withDynamoLookup(redemptionService))
+      _ <- updateRedemptionCodeIfApplicable(state.paymentMethod, subscriptionData, SetCodeStatus.withDynamoLookup(redemptionService))
     } yield thankYouState
   }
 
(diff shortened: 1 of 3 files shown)


support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -81,6 +81,7 @@ object CreateZuoraSubscription {
         case None => subscribe(subscribeItem, zuoraService).map(response => toHandlerResult(state, response, paymentMethodWithPaymentSchedule, requestInfo))
           .withLogging("subscribe")
       }
+      _ <- updateRedemptionCode(state.paymentMethod, subscriptionData, SetCodeStatus.withDynamoLookup(redemptionService))
```
rupertbates, line 84: I realise this wasn't introduced in this PR but maybe move the match against `maybeDomainSubscription` above to a separate method? This for comprehension is getting pretty big

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -230,4 +231,20 @@ object CreateZuoraSubscription {
...
+    subscriptionData: SubscriptionData,
+    setCodeStatus: SetCodeStatus
+  ): Future[Unit] =
+    paymentMethod match {
+      case Right(rd: RedemptionData) =>
+        setCodeStatus(
+          RedemptionCode(rd.redemptionCode).right.get,
+          RedemptionTable.AvailableField.CodeIsUsed
+        )
+      case Left(_: PaymentMethod) if (subscriptionData.subscription.redemptionCode.isEmpty) => Future.successful(())
```
rupertbates, line 246: Explain this if statement to me 😄

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
johnduffell: good idea.  the whole skip/don't skip thing seems to involve several methods that all call each other dotted around, but for simplicity I had decided that was out of scope for this PR or a preceding refactor.  Happy to add a health card to revisit in a separate PR though?

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
johnduffell: For a paid sub, we should not be planning to put anything in the redemption code field of zuora.   That logic is hidden away in the builders.
If that all matches, then we can continue without updating the DB.
If there's a mismatch, then we throw an error (like a runtime assertion)

However if it needs explaining then it's not clear enough.  Having looked again, I wonder if it makes sense to call the "updateRedemptionCode" at all, if there is no redemption code.  Assuming we think the check adds value, we should have it as a separate step.

Having said that, I raised the separate refactor PR so that I could add the test for this (which is at the bottom of this PR) so maybe the value of the check is low now compared with the cost.

johnduffell:
thanks for approval @rupertbates
this PR can only be merged if this previous pr is merged: https://github.com/guardian/support-frontend/pull/2546
Otherwise the test I have added will not be possible (at least without a lot more complexity) so we would need to revisit.

PR #2549, opened 2020-06-16
removing geo check for og image

JoemG:
## Why are you doing this?
We don't think the crawler geo-ip will resolve correctly. 
We decided to use the moment OG image for all users for the duration of the moment.

[**Trello Card**](https://trello.com/c/AilQGrjm/2114-update-the-ogimage-on-the-aus-thank-youpage)


JoemG:
> Great!

I will merge tomorrow 👍

PR #2550, opened 2020-06-17
Use correct campaignCode during Aus campaign, use switch

tomrf1:
We can configure campaign-specific landing pages.
This is done by matching on the path of the url, and in the upcoming moment this will be /au/contribute.
In the past, the `campaignCode` has been set based on the path. Instead, we want to configure the `campaignCode` - in this case it is `Aus_moment_2020`.

Also, the campaign is now enabled based on the `ausMomentEnabled` switch, so no need for commented out code!

Redux store now has correct campaignCode:
Screen Shot 2020-06-17 at 13 19 00

### DIFF
--- a/support-frontend/app/views/contributions.scala.html
+++ b/support-frontend/app/views/contributions.scala.html
@@ -110,7 +110,7 @@
   }
 
   window.guardian.forceCampaign = @settings.switches.experiments.get("forceCampaign").exists(_.isOn)
-  window.guardian.ausMomentEnabled = false
+  window.guardian.ausMomentEnabled = @settings.switches.experiments.get("ausMomentEnabled").exists(_.isOn)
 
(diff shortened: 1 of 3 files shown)


support-frontend/app/views/contributions.scala.html
```diff
@@ -110,7 +110,7 @@
   }
 
   window.guardian.forceCampaign = @settings.switches.experiments.get("forceCampaign").exists(_.isOn)
-  window.guardian.ausMomentEnabled = @settings.switches.experiments.get("ausMomentEnabled").exists(_.isOn)
+  window.guardian.ausMomentEnabled = false
```
JoemG, line 113:
```suggestion
@settings.switches.experiments.get("ausMomentEnabled").exists(_.isOn)
```

PR #2552, opened 2020-06-17
Remove unused AB test variants of ad-free copy

rupertbates:
## Why are you doing this?
Part 1 of a number of PRs to remove unused code from the codebase.

This was a number of variants of the ad-free copy which appears on the digisub landing page which we presumably used in a test at some point.

It seems like the winning variant was `AdFreeSectionC` so I've renamed that to `AdFreeSection` and deleted all the others.


support-frontend/assets/components/adFreeSection/adFreeSection.scss
```diff
@@ -1,61 +1,114 @@
...
+    padding-bottom: 80px;
   }
 }
 
-.component-ad-free-section__col {
-  width: 100%;
-  padding-right: 40px;
-  @include mq($from: leftCol) {
-    width: 440px;
+.ad-free-section-c__content {
```
GHaberis, line 31: Shall we rename any reference to `.ad-free-section-c__*` to `.ad-free-section__*` in here and the component?

There's a few references to `.ad-free-section-b__*` in the `.scss` that can be removed now too.

rupertbates: Yes good spot

PR #2547, opened 2020-06-16
Corporate Subs Redemption Form Styling

rupertbates:
## Why are you doing this?
This PR updates the corporate subs redemption form to match the designs.
[**Figma file**](https://www.figma.com/file/81Ku9XImBCx1jQXIP8EoF2/Corporate-Subs-2020?node-id=565%3A1079)
[**Trello Card**](https://trello.com/c/Z7dcEaJO/3096-redemption-form-styling-accessibility)

## Changes
- Upgrade Source to v 1.4 to make use of the `user-feedback` module
- Put redemption form behind Google Auth.
- Add a `footerAside` prop to the `CheckoutLayout` component, this is an aside that wraps underneath rather than above the main content in a mobile view.
- Match form styling to the Figma designs

## Accessibility test checklist
 - [x] [Tested with screen reader](https://accessibility.gutools.co.uk/testing/web/screen-readers/)
 - [x] [Navigable with keyboard](https://accessibility.gutools.co.uk/testing/web/keyboard-navigation/)
 - [x] [Colour contrast passed](https://accessibility.gutools.co.uk/testing/web/colour-contrast/)

### Desktop
Screen Shot 2020-06-16 at 14 06 12
### Mobile
support thegulocal com_subscribe_redeem_welcome_test-code(iPhone X) (1)

### Desktop
Screen Shot 2020-06-16 at 12 05 01
### Mobile
support thegulocal com_subscribe_redeem_welcome_test-code(iPhone X)

PR #2558, opened 2020-06-18
Fix layout issue in redemption form

rupertbates:
## Why are you doing this?
I noticed that #2547 introduced a bug where the aside would show on the wrong side of the form in desktop layouts:
Screen Shot 2020-06-18 at 14 01 28

This PR fixes it:
Screen Shot 2020-06-18 at 15 00 49

I also noticed that the `CheckoutLayout` control is never used without an `aside` so I've made it non-optional as it simplifies the code.


rupertbates:
I agree about using emotion but I'm going to do that in a separate PR because I need to get this out for Cloe and Gus to work with

PR #2560, opened 2020-06-19
Ds remove address test

rupertbates:
## Why are you doing this?
We want to test removing the address fields from Digital Subscriptions to see if it results in an increase in conversion.

Note that this PR removes server side validation of address fields for all digisub checkouts. I think this is acceptable as we are happy to accept subs without an address for those in the 'no address' variant, so it is not a disaster if somehow those in the other variant manage to submit a purchase without a valid address.

[**Trello Card**](https://trello.com/c/x0iBur2v/3090-ab-test-does-removing-address-fields-from-ds-checkout-increase-conversion-rate)

### DIFF
--- a/support-frontend/app/utils/CheckoutValidationRules.scala
+++ b/support-frontend/app/utils/CheckoutValidationRules.scala
@@ -102,9 +102,9 @@ object DigitalPackValidation {
     def isValidPaidSub(paymentFields: PaymentFields) =
       SimpleCheckoutFormValidation.passes(createSupportWorkersRequest) &&
         hasStateIfRequired(country, state) &&
-        hasPostcodeIfRequired(country, postCode) &&
+        //hasAddressLine1AndCity(billingAddress) &&
+        //hasPostcodeIfRequired(country, postCode) &&
         currencyIsSupportedForCountry(country, currency) &&
-        hasAddressLine1AndCity(billingAddress) &&
         PaidProductValidation.noEmptyPaymentFields(paymentFields)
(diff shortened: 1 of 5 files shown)


support-frontend/app/utils/CheckoutValidationRules.scala
```diff
@@ -102,9 +102,9 @@ object DigitalPackValidation {
     def isValidPaidSub(paymentFields: PaymentFields) =
       SimpleCheckoutFormValidation.passes(createSupportWorkersRequest) &&
         hasStateIfRequired(country, state) &&
-        hasPostcodeIfRequired(country, postCode) &&
+        //hasAddressLine1AndCity(billingAddress) &&
+        //hasPostcodeIfRequired(country, postCode) &&
```
rupertbates, line 106: I'm going to leave these lines here commented out for the duration of the test

support-frontend/assets/helpers/abTests/abtestDefinitions.js
```diff
@@ -123,4 +124,28 @@ export const tests: Tests = {
...
+    ],
+    audiences: {
+      ALL: {
+        offset: 0,
+        size: 1,
+      },
+    },
+    isActive: true,
+    referrerControlled: false,
+    seed: 7,
```
johnduffell, line 146: unrelated, but `landingPageDesignSystemTest` and `fancyAddressTest` both have the same seed, I presume they weren't running at the same time

PR #2561, opened 2020-06-19
remove corporate details from client side

johnduffell:
## Why are you doing this?

This PR is part of the coporate subs work.
We have implemented the back end and front end separately and now need to integrate them together.
It became clear that some of the assumptions about which fields were needed were differeent between the two sides.
This PR removes the corporate name/id from the client side as they are not needed in the end.
The types are more lined up between frontend and backend too.
But the code itself isn't connected yet.
There are a few changes to permissions etc at the same time.

Tested and working in CODE for corporate, it checks the code and marks it as available=false

[**Trello Card**](https://trello.com/c/8ipoyypI/3130-integrate-ui-and-step-functions-for-redemption-stuff)


support-frontend/app/controllers/Promotions.scala
```diff
@@ -40,6 +40,7 @@ class Promotions(
         case GuardianWeekly => routes.WeeklySubscription.weeklyGeoRedirect(promotionTerms.isGift).url
         case DigitalPack => routes.DigitalSubscriptionController.digitalGeoRedirect().url
         case Paper => routes.PaperSubscription.paper(false).url
+        case Contribution => routes.Application.contributeGeoRedirect("").url
```
johnduffell, line 43: this was just a scalac warning that kept popping up due to bad types so I fixed it the easy way

support-frontend/app/controllers/RedemptionController.scala
```diff
@@ -51,10 +50,14 @@ class RedemptionController(
   val js = "subscriptionsRedemptionPage.js"
   val css = "digitalSubscriptionCheckoutPage.css" //TODO: Don't need this?
 
-  def displayForm(redemptionCode: String): Action[AnyContent] = googleAuthAction.async {
+  def displayForm(redemptionCode: RawRedemptionCode): Action[AnyContent] = googleAuthAction.async {
```
johnduffell, line 53: RawRedemptionCode is a String, it needs to be uppercased before use, probably in the next PR

support-frontend/app/controllers/RedemptionController.scala
```diff
@@ -51,10 +50,14 @@ class RedemptionController(
...
     implicit request =>
-      getCorporateCustomer(redemptionCode).value.map(
-        customerOrError =>
+      val eventualErrorOrUnit = for {
+        redemptionCode <- EitherT.fromEither[Future](RedemptionCode(redemptionCode))
```
johnduffell, line 56: need to leftMap this and replace with `"Please check the code and try again"`

support-frontend/app/views/subscriptionRedemptionForm.scala.html
```diff
@@ -28,10 +27,9 @@
     <script type="text/javascript">
       window.guardian = window.guardian || {};
       window.guardian.stage = "@stage";
-      window.guardian.userCode = "@redemptionCode";
-      window.guardian.error = "@customerOrError.left.getOrElse("")";
+      window.guardian.userCode = "@Html(redemptionCode.replaceAll(""""""", """\\""""))";
+      window.guardian.error = "@Html(maybeRedemptionError.getOrElse("").replaceAll(""""""", """\\""""))";
```
johnduffell, line 31: this should escape the strings properly into the JS without doing html escaping (i.e. ending up with &quot; and &amp; etc.

support-frontend/assets/pages/subscriptions-redemption/subscriptionsRedemptionReducer.js
```diff
@@ -40,7 +34,6 @@ export type Action =
   | { type: 'SET_STAGE', stage: Stage }
 
 const initialState: RedemptionFormState = {
-  corporateCustomer: getGlobal('corporateCustomer'),
   stage: getGlobal('stage') || 'form',
   userCode: getGlobal('userCode'),
```
johnduffell, line 38: I wasn't quite sure the difference in purpose between corporateCustomer.redemptionCode and userCode.
Also, it didn't look like corporateCustomer ever got updated (by what was returned from the ajax endpoint when you edited the code)
In the end I decided to just use userCode everywhere as I think that's enough.

support-models/src/main/scala/com/gu/support/redemptions/RedemptionCode.scala
```diff
@@ -0,0 +1,31 @@
...
+    if (value.forall(validCharsSet.contains))
+      Right(new RedemptionCode(value))
+    else {
+      val groupsDesc = validChars.map {
+        case (from, to) if from == to => s""""$from""""
+        case (from, to) => s"$from-$to"
+      }
+      val withSep = groupsDesc.foldRight(List[String]()) { case (next, soFar) => if (soFar.isEmpty) List(" and ", next) else ", " :: next :: soFar }
+      val description = withSep.tail.mkString
+      Left(s"redemption code must only include $description")
```
johnduffell, line 20: I was thinking about hard coding the error message to the one we want on the client side, rather than generating this and then leftMapping it later, however I prefer to keep the client side "friendly" messages in the view/controller, also this is the dev friendly message that the step functions will report if there is any json parse error.
Happy to go either way on it though.
Other simple option is just hard code here to `A-Z, 0-9 and "-"` rather than being so DRY!

support-models/src/main/scala/com/gu/support/redemptions/RedemptionData.scala
```diff
@@ -2,24 +2,23 @@ package com.gu.support.redemptions
...
 import cats.syntax.functor._
 import com.gu.support.encoding.Codec
-import com.gu.support.redemptions.redemptions.{CorporateAccountId, RedemptionCode}
 import io.circe.syntax._
 import io.circe.{Decoder, Encoder}
 
 sealed abstract class RedemptionData(val redemptionCode: RedemptionCode)
 
-case class CorporateRedemption(override val redemptionCode: RedemptionCode, corporateAccountId: CorporateAccountId) extends RedemptionData(redemptionCode)
+case class CorporateRedemption(override val redemptionCode: RedemptionCode) extends RedemptionData(redemptionCode)
```
johnduffell, line 10: I'm wondering whether we need this ADT or whether all of them can be modelled with a single value class.  Depends if we're going to use the same dynamo table really.

support-models/src/main/scala/com/gu/support/redemptions/RedemptionData.scala
```diff
@@ -2,24 +2,23 @@ package com.gu.support.redemptions
...
+    case c: CorporateRedemption => c.asJson
+  }
 
-    implicit val decoder: Decoder[RedemptionData] =
-      List[Decoder[RedemptionData]](
-        Decoder[CorporateRedemption].widen
-      ).reduceLeft(_ or _)
+  implicit val decoder: Decoder[RedemptionData] =
+    List[Decoder[RedemptionData]](
+      Decoder[CorporateRedemption].widen
```
johnduffell, line 22: will this be able to distinguish between CorporateRedemption and GiftRedemption if they both have the same field or one is a subset of the other?  A bit hypothetical since this code is just-in-case, but we will need to be aware.

support-services/src/main/scala/com/gu/support/redemption/DynamoTableAsync.scala
```diff
@@ -19,8 +19,9 @@ object DynamoTableAsync {
   val ProfileName = "membership"
 
   lazy val CredentialsProvider =  AwsCredentialsProviderChain.builder.credentialsProviders(
-     ProfileCredentialsProvider.builder.profileName(ProfileName).build,
-     InstanceProfileCredentialsProvider.builder.asyncCredentialUpdateEnabled(false).build
+    ProfileCredentialsProvider.builder.profileName(ProfileName).build,
+    InstanceProfileCredentialsProvider.builder.asyncCredentialUpdateEnabled(false).build,
+    EnvironmentVariableCredentialsProvider.create()
```
johnduffell, line 24: this was needed when running in AWS, the lambda gets the credentials from the environment, but somehow AWS SDK v1 was picking them up regardless.

support-frontend/app/views/subscriptionRedemptionForm.scala.html
rupertbates: Yes I would extract this to a helper function

support-frontend/assets/pages/subscriptions-redemption/subscriptionsRedemptionReducer.js
rupertbates: `userCode` was the code input by the user whereas `redemptionCode` was the one returned from the server. I think what you've done is fine though

support-models/src/main/scala/com/gu/support/redemptions/RedemptionCode.scala
```diff
@@ -0,0 +1,31 @@
...
+
+object RedemptionCode {
+  def apply(value: String): Either[String, RedemptionCode] = {
+    // make sure no one can inject anything bad
+    val validChars = List('A' -> 'Z', '0' -> '9', '-' -> '-')
+    val validCharsSet = validChars.flatMap { case (from, to) => (from to to) }.toSet
+    if (value.forall(validCharsSet.contains))
+      Right(new RedemptionCode(value))
+    else {
+      val groupsDesc = validChars.map {
```
rupertbates, line 14: Is all of this just generating an error message? If so could you move it into a separate method for to keep this one simple

support-models/src/main/scala/com/gu/support/redemptions/RedemptionCode.scala
rupertbates: Yeah I think I'd just hard code the restrictions here rather than generating it. It's a lot of code for not a massive amount of benefit

support-models/src/main/scala/com/gu/support/redemptions/RedemptionData.scala
rupertbates: If they have exactly the same field then we would need a type descriminator to aid the serialisation, but as you mentioned above, at that point we may not need the subclasses any more

support-models/src/main/scala/com/gu/support/redemptions/RedemptionData.scala
rupertbates: Maybe not but let's leave it for now until we're sure.

support-frontend/app/views/JsStringLiteral.scala
```diff
@@ -0,0 +1,25 @@
+package views
+
+import play.twirl.api.Html
+
+object JsStringLiteral {
+
+  def apply(rawString: String): Html = {
```
johnduffell, line 7: after moving the escaper over, I realised it wasn't going to be as safe as I'd hoped.
This implementation should handle special characters both in Html and Javascript.
The mappings I got from https://stackoverflow.com/questions/8749001/escaping-html-entities-in-javascript-string-literals-within-the-script-block
I have tested it and it seems to work
string `TEST-CODE < > " ' \ & `
ended up in the source as:
image
and in the page showed up as 
image

PR #2566, opened 2020-06-23
connect frontend and backend redemption code logic

johnduffell:
## Why are you doing this?

This PR is the critical backend/frontend connection PR, means we can use redemption codes in UAT/test user mode against the real DB.

I was struggling with test user mode a lot, so I added a widget to the new page, that will add a flash to show we're in test user mode, and why.

image

[**Trello Card**](https://trello.com/c/8ipoyypI/3130-integrate-ui-and-step-functions-for-redemption-stuff)


support-frontend/app/controllers/RedemptionController.scala
```diff
@@ -35,7 +37,8 @@ class RedemptionController(
   testUsers: TestUserService,
   components: ControllerComponents,
   fontLoaderBundle: Either[RefPath, StyleContent],
-  googleAuthAction: AuthAction[AnyContent]
+  googleAuthAction: AuthAction[AnyContent],
+  dynamoLookup: Boolean => DynamoLookup
```
johnduffell, line 41: need to change to a named trait

support-frontend/app/controllers/RedemptionController.scala
```diff
@@ -26,6 +28,10 @@ import views.html.subscriptionRedemptionForm
 
 import scala.concurrent.{ExecutionContext, Future}
 
+trait DynamoTableAsyncForUser {
```
johnduffell, line 31: wasn't quite sure what to call this and couldn't find a precedent - I looked at the payment api service and it seems to just have it as a parameter on the actual method, Promotion service seems to use the TouchpointServiceProvider which handles test users and accessing the config together.
In the end I just called it what it is, get the DynamoTableAsync for user.

support-frontend/app/controllers/RedemptionController.scala
```diff
@@ -50,26 +57,46 @@ class RedemptionController(
...
+          case Some(eventualErrorOrUser) =>
+            eventualErrorOrUser.fold(
+              message => {
+                SafeLogger.error(scrub"could not fetch user - assuming normal backend: $message")
+                None
+              },
+              testUserFromRequest.fromIdUser
+            )
+        }
+        normalisedCode = redemptionCode.toUpperCase(Locale.UK)
```
johnduffell, line 79: could take this out, I just thought it's nice that the code is displayed in caps on first load

support-frontend/assets/pages/subscriptions-redemption/api.js
```diff
@@ -27,7 +28,9 @@ function validate(userCode: string) {
   if (userCode === '') {
     return Promise.resolve({ valid: false, errorMessage: 'Please enter your code' });
   }
-  const validationUrl = `${getOrigin()}/subscribe/redeem/validate/${userCode}`;
+  const whyIsTestUser = getGlobal<string>('whyIsTestUser');
```
johnduffell, line 31: wasn't sure if we should just grab the global constant from this part of the code or we should do it once centrally

support-frontend/assets/components/headers/header/header.jsx
```diff
@@ -6,6 +6,7 @@ import React, { Component, type Node } from 'react';
```
johnduffell, line 6: the changes in this file are to add the banner, as it was hard to debug, but it would be nice to have this feature back as we had it in subs and membership frontend.  Easy to remove if we don't like it though.

support-frontend/app/controllers/RedemptionController.scala
```diff
@@ -50,26 +57,46 @@ class RedemptionController(
...
-            stage = "checkout",
-            redemptionCode = redemptionCode,
-            maybeRedemptionError = maybeError,
-            user = None,
-            submitted = false
-          )).withHeaders(CacheControl.noCache)
-      )
-    }
+      for {
+        whyIsTestUser <- request.user.map { authenticatedIdUser =>
```
rupertbates, line 66: Could you break out the code involved in this assignment to a new method please? Also shouldn't it just be `isTestUser` rather than `whyIsTestUser`?

johnduffell: I wasn't sure about naming, actually this also gives the reason they are a test user, e.g. whether it came from the cookie or the identity display name.

johnduffell: how about `testUserStatus`?

support-frontend/app/controllers/RedemptionController.scala
```diff
@@ -127,21 +142,27 @@ class RedemptionController(
           userHasSub =>
             if (userHasSub)
               Future.successful(redirectToExistingThankYouPage)
-            else
-              tryToShowProcessingPage(redemptionCode)
+            else {
+              identityService.getUser(request.user.minimalUser).value.flatMap{
+                case Left(message) =>
+                  SafeLogger.error(scrub"could not fetch user: $message")
+                  Future.successful(InternalServerError("could not fetch user"))
```
johnduffell, line 149: they should be correctly logged in by now.  However we could just assume they are not test user at this point, so they can still check out

support-frontend/app/controllers/RedemptionController.scala
```diff
@@ -50,26 +57,34 @@ class RedemptionController(
...
-            stage = "checkout",
-            redemptionCode = redemptionCode,
-            maybeRedemptionError = maybeError,
-            user = None,
-            submitted = false
-          )).withHeaders(CacheControl.noCache)
-      )
-    }
+      for {
+        whyIsTestUser <- testUserFromRequest.getTestUserStatus(request)
```
johnduffell, line 66: the comment has gone now @rupertbates but we could call this `testUserStatus` perhaps.  either/or we could consider a named ADT rather than leaning on Option in this unexpected way.

support-frontend/app/controllers/RedemptionController.scala
rupertbates: I'd go for `isTestUser` it is pretty standard in the code base both in the scala and js code

support-frontend/assets/components/headers/header/header.jsx
johnduffell: also, if we remove this splash banner, we would be able to go back to boolean for isTestUser/uatMode.

The reason this is in one PR is that whether we do this or not affects whether we use boolean or option<string> for test user mode for this new code.
If we did this in a later PR, we would have to rewrite all the test user code I just added to use new types, which would add a lot more work.

support-frontend/app/controllers/RedemptionController.scala
rupertbates: Do we need it to be an `Option[String]` or can we map it to a boolean?

johnduffell: ok will do, originally I specifically wanted to avoid that to make it clear this wasn't the usual true/false, but the IDE should keep people right via the type checkers anyway.

johnduffell: to answer "need", the string on the true side explains whether you were marked as a test user due to the cookie or due to being logged in.  If that information is not useful then a boolean would be enough
image

rupertbates: If the string is only so we can display the username in the flash I'd just turn this into a boolean and reword the flash so it doesn't use the displayname

johnduffell: ok something like this: 
image

support-frontend/app/controllers/RedemptionController.scala
rupertbates: Yes worth leaving in I think

support-frontend/app/controllers/RedemptionController.scala
```diff
@@ -127,21 +142,27 @@ class RedemptionController(
           userHasSub =>
             if (userHasSub)
               Future.successful(redirectToExistingThankYouPage)
-            else
-              tryToShowProcessingPage(redemptionCode)
+            else {
+              identityService.getUser(request.user.minimalUser).value.flatMap{
```
rupertbates, line 146: I don't think we need this call to identity here, we are doing it in `tryToShowProcessingPage` below

johnduffell: ooh good spot there! 👍

support-frontend/app/controllers/RedemptionController.scala
```diff
@@ -50,26 +57,34 @@ class RedemptionController(
...
-            css = css,
-            fontLoaderBundle = fontLoaderBundle,
-            csrf = None,
-            uatMode = false,
-            stage = "checkout",
-            redemptionCode = redemptionCode,
-            maybeRedemptionError = maybeError,
-            user = None,
-            submitted = false
-          )).withHeaders(CacheControl.noCache)
```
johnduffell, line 70: just realised I lost this noCache directive, I assumed authenticated pages are always nocache but I will check.

johnduffell: it has a `cache-control: no-cache, private` in the response locally, so OK

Jpsi:
Very useful to have the red test user widget, thank you!

PR #2567, opened 2020-06-24
Corporate subs thank you email

johnduffell:
## Why are you doing this?

This send the right sqs message for the corp subs thank you email.

Depends on the refactoring in https://github.com/guardian/support-frontend/pull/2572

[**Trello Card**](https://trello.com/c/BN05KB70/3070-jess-king-thank-you-email-api-trigger)


support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
```diff
@@ -38,55 +38,61 @@ import com.gu.support.workers._
...
+        "Trial period" -> "14", //TODO: depends on Promo code or zuora config
+        "Subscription details" -> SubscriptionEmailFieldHelpers.describe(paymentSchedule, billingPeriod, currency, promotion)
+      )
+    case Right(rd) => List(
+      "Subscription details" -> "Group subscription"
+    )
+  })
 
-  override def payload: String = super.payload(user.primaryEmailAddress, "digipack")
+  override def payload: String = super.payload(user.primaryEmailAddress, if (paymentMethod.isLeft) "digipack" else "digipack-corp")
```
johnduffell, line 96: https://github.com/guardian/membership-workflow/pull/244

support-workers/src/main/scala/com/gu/support/workers/lambdas/SendThankYouEmail.scala
```diff
@@ -89,10 +88,10 @@ class SendThankYouEmail(thankYouEmailService: EmailService, servicesProvider: Se
           productOptions = p.productOptions,
           billingPeriod = p.billingPeriod,
           user = state.user,
-          paymentSchedule = state.paymentSchedule,
+          paymentSchedule = state.paymentMethod.left.toOption.map(_._2).getOrElse(PaymentSchedule(List())),
```
johnduffell, line 91: not sure if I'm a massive fan of the _1 and _2 everywhere as it doesn't read well, but we get the built in logic for free.  Could create a custom case class plus codec, then also the json state would be more clear.

support-workers/src/test/scala/com/gu/support/workers/JsonFixtures.scala
```diff
@@ -318,29 +318,31 @@ object JsonFixtures {
...
-       |        "date": "2019-02-14",
-       |        "amount": 11.99
-       |      },
-       |      {
-       |        "date": "2019-03-14",
-       |        "amount": 11.99
-       |      }
-       |    ]
-       |  }
+       |  "paymentMethod": [
```
johnduffell, line 327: default codec makes a json list for tuples

support-models/src/test/scala/com/gu/support/workers/SerialisationSpec.scala
```diff
@@ -51,6 +52,25 @@ class SerialisationSpec extends AsyncFlatSpec with SerialisationTestHelpers with
     testDecoding[SendThankYouEmailState](thankYouEmailJson())
   }
 
+  it should "roundtrip successfully" in {
```
johnduffell, line 55: not sure if this test is in the right place or even needed - the others are all using hard coded json

support-models/src/main/scala/com/gu/support/workers/states/SendThankYouEmailState.scala
```diff
@@ -13,13 +13,12 @@ case class SendThankYouEmailState(
   user: User,
   giftRecipient: Option[GiftRecipient],
   product: ProductType,
-  paymentMethod: Either[PaymentMethod, RedemptionData],
+  paymentMethod: Either[(PaymentMethod, PaymentSchedule), RedemptionData],
```
rupertbates, line 16: Can you explain this change please?

johnduffell: yes, the payment schedule only exists where there are payments, if it's a free sub there's no concept of a payment schedule

johnduffell: I think I commented elsewhere that a tuple isn't so clear and maybe a case class would be more descriptive, if you agree?

rupertbates: I'm not too sure about this new model, it seems as though it would be clearer to retain the existing `paymentMethod` type and make `paymentSchedule` an Option. 
I realise that that is slightly less tight from a data validity point of view though. Maybe we could introduce a wrapper type `PaymentMethodWithSchedule` in place of this tuple?

rupertbates: We overlapped with our above comments, case class seems like a good solution to me

support-workers/src/main/scala/com/gu/emailservices/EmailFields.scala
```diff
@@ -1,16 +1,17 @@
 package com.gu.emailservices
 
+import com.gu.i18n.Currency
 import com.gu.salesforce.Salesforce.SfContactId
+import com.gu.support.promotions.Promotion
+import com.gu.support.workers.{BillingPeriod, User}
 import io.circe.generic.auto._
 import io.circe.syntax._
 
-// scalastyle:off
```
johnduffell, line 7: scalastyle doesn't seem to object to what's there now, maybe this was an old rule

support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
```diff
@@ -57,44 +54,53 @@ case class DigitalPackEmailFields(
...
-    "Address 2" -> "", //TODO: Remove this from Braze template
-    "City" -> "", //TODO: Remove this from Braze template
-    "Post Code" -> "", //TODO: Remove this from Braze template
-    "Country" -> user.billingAddress.country.name,
-    "Date of first payment" -> formatDate(SubscriptionEmailFieldHelpers.firstPayment(paymentSchedule).date),
-    "Currency" -> currency.glyph,
-    "Trial period" -> "14", //TODO: depends on Promo code
-    "Subscription details" -> SubscriptionEmailFieldHelpers.describe(paymentSchedule, billingPeriod, currency, promotion)
-  ) ++ paymentFields
+      override def payload: String = super.payload(user.primaryEmailAddress, if (paidSubPaymentData.isDefined) "digipack" else "digipack-corp")
```
johnduffell, line 102: https://github.com/guardian/membership-workflow/pull/244

PR #2569, opened 2020-06-24
Show CCPA CMP to US users

GHaberis:
## Why are you doing this?

From July 1st the main Guardian domain will be collecting US users consent state in relation to CCPA. We will also be required to do this across on `support.theguardian.com`.

@guardian/commercial-dev have been working on a library https://www.npmjs.com/package/@guardian/consent-management-platform that can be used to surface a CCPA compliant CMP to our US audience. If a user is in the US, we should initialise the CCPA CMP, otherwise we should continue using the current CMP.

The CCPA CMP will show on the user's first page view, at which point they have the opportunity to opt out of tracking, if they don't opt out at this point they will be considered to have opted in all subsequent page views and the CMP will not show again.

[**Trello Card**](https://trello.com/c/bs3xSiSJ/3101-implement-ccpa-cmp-for-us-audience-on-supporttheguardiancom)

## Changes

- `page.js` now imports `import { init as initCMP } from '@guardian/consent-management-platform'`. If a user is in The 'US' we call `initCMP` with `useCCPA: true` when we initialise the page, this will show the CCPA CMP to that user if required.
- `getTrackingConsent` in `thirdPartyTrackingConsentTest.js` is now asynchronous and returns a `Promise<ThirdPartyTrackingConsent> ` instead of `ThirdPartyTrackingConsent`. This is necessary because we have to use `onIabConsentNotification` to obtain CCPA `consentState` which is an asynchronous action.
- I've deleted the `trackingConsent` prop in the Redux store and removed all associated actions/dispatchers. I've chosen to do this as initialising the store would have to become an asynchronous action following the updates to `getTrackingConsent`, doing this may have performance implications and would involve a significant refactor. The `trackingConsent` prop was only used by the `consentBanner.jsx` to determine whether it should show so I've moved the logic to call `getTrackingConsent` into the component itself.
- Updated `getData` in `googleTagManager.js` to be asynchronous as it relies upon  the result of `getTrackingConsent`.
- Added the following domains to the `content-security-policy`: https://ccpa-service.sp-prod.net, https://ccpa-notice.sp-prod.net, https://consent.theguardian.com, https://ccpa.sp-prod.net. This was necessary to get the CCPA CMP to load within an iframe.

## Todo

- Put CCPA behind a feature switch so it can easily be enabled/disabled.

## Screenshots

**CCPA CMP on Mobile...**

support-us-cmp

**CCPA CMP on Desktop...**

support-us-cmp-dt


GHaberis:
I'm looking at why the command `yarn build-ssr` now fails, it's doing this because we're importing `import { init as initCMP } from '@guardian/consent-management-platform'` in page.js, and it fails with the following error: `ReferenceError: __uspapi is not defined`. `__uspapi`  must be referenced within the @guardian/consent-management-platform or it's dependencies.

GHaberis:
> I'm looking at why the command `yarn build-ssr` now fails, it's failing because we're importing `import { init as initCMP } from '@guardian/consent-management-platform'` in `page.js`, and it fails with the following error: `ReferenceError: __uspapi is not defined`. `__uspapi` must be referenced within the `@guardian/consent-management-platform` or it's dependencies.

This fails  because at build time we're trying to generate static HTML versions of the `contributions-landing.html` and `showcase.html` pages. We're doing this by running the `'pages/showcase/showcase'` and `pages/contributions-landing/EmptyContributionsLandingPage` entry points on the server at build time, these entry points contain code that expects to run in a browser, with references to the browser API within, to get around this we're using `JSDOM` to mock various aspects  of the browser API. Some of the dependencies pulled in by `@guardian/consent-management-platform` are failing because they're not designed to be run on the server this way.

I'd be keen to discuss options here. A short-term work around is to turn off server side rendering (SSR) for these two pages. Another work around is using webpack we could create an alias (stub) of `@guardian/consent-management-platform` that's used by webpack instead of the real `@guardian/consent-management-platform`  at build time, only for these SSR pages. Neither are ideal. A longer term solution would be to revisit how we do SSR.

GHaberis:
I've fixed the server side rendering by stubbing `window.__uspapi` on the server.

lucymonie:
This is more of a question about the text content, but I was wondering about the decision to show a California-specific message to the whole of the US. This could seem a little tone-deaf in some parts of the US where California is regarded as ...well, California.

GHaberis:
> This is more of a question about the text content, but I was wondering about the decision to show a California-specific message to the whole of the US. This could seem a little tone-deaf in some parts of the US where California is regarded as ...well, California.

It's a good question, honestly I'm not sure of the detail but I believe CCPA is being adopted as a blueprint by many other states, and there many organisations are also applying it across all states as best practise.

support-frontend/assets/helpers/tracking/ccpa.js
```diff
@@ -0,0 +1,17 @@
...
+
+import {
+  detect as detectCountry,
+  type IsoCountry,
+} from 'helpers/internationalisation/country';
+
+// ----- Functions ----- //
+
+export const ccpaEnabled = (): boolean => {
+  const useCCPA = true; // set false to switch CCPA off
```
tomrf1, line 13: Is it worth having this in the switchboard?

ripecosta:
> This is more of a question about the text content, but I was wondering about the decision to show a California-specific message to the whole of the US. This could seem a little tone-deaf in some parts of the US where California is regarded as ...well, California.

I can provided some context regarding decisions. The label for this link was decided by the DPO (IIRC) so that the connection between the options and CCPA (California Consumer Privacy Act) is clear. Because geolocation is sometimes not precise enough to differentiate between US state, it's been decided to apply it to the US as a whole to be on the safe side.

support-frontend/assets/helpers/page/page.js
```diff
@@ -136,11 +134,22 @@ function init<S, A>(
   pageReducer?: CommonState => Reducer<S, A> | null,
   thunk?: boolean = false,
 ): Store<*, *, *> {
-
   try {
+    /**
+     * If ccpaEnabled initialise CCPA CMP as early
+     * as possible so subsequent call so
```
ripecosta, line 140:
```suggestion
     * as possible so subsequent call to
```

support-frontend/assets/helpers/tracking/ccpa.js
GHaberis: Yes definitely, this was my intention which I listed as a todo, I'm not sure if/how we uses switches on support-frontend, any guidance/examples would be appreciated.

support-frontend/assets/helpers/tracking/thirdPartyTrackingConsent.js
```diff
@@ -13,16 +16,48 @@ export type ThirdPartyTrackingConsent = typeof OptedIn
...
+         * */
+        if (typeof consentState !== 'boolean') {
+          throw new Error('consentState not a boolean');
+        } else {
+          // consentState true means the user has OptedOut
+          resolve(consentState ? OptedOut : OptedIn);
+        }
+      });
+    }).catch((err) => {
+      console.log('***** ERR *****', err);
```
ripecosta, line 36: Was leaving this in deliberate or just for testing purposes?

GHaberis: thanks for pointing out, that should be removed

lucymonie:
> > This is more of a question about the text content, but I was wondering about the decision to show a California-specific message to the whole of the US. This could seem a little tone-deaf in some parts of the US where California is regarded as ...well, California.
> 
> I can provided some context regarding decisions. The label for this link was decided by the DPO (IIRC) so that the connection between the options and CCPA (California Consumer Privacy Act) is clear. Because geolocation is sometimes not precise enough to differentiate between US state, it's been decided to apply it to the US as a whole to be on the safe side.

Thanks for the background @ripecosta! Sounds like it has been through 

> > This is more of a question about the text content, but I was wondering about the decision to show a California-specific message to the whole of the US. This could seem a little tone-deaf in some parts of the US where California is regarded as ...well, California.
> 
> I can provided some context regarding decisions. The label for this link was decided by the DPO (IIRC) so that the connection between the options and CCPA (California Consumer Privacy Act) is clear. Because geolocation is sometimes not precise enough to differentiate between US state, it's been decided to apply it to the US as a whole to be on the safe side.

Thanks for the background @ripecosta. I totally agree that showing it to the whole of the US is great. I just think the wording could annoy some users who might not be fans of California.

mxdvl:
Would you like to move to the new CMP (v4)?

GHaberis:
> Would you like to move to the new CMP (v4)?

What does this entail? I can't find much information about v4 and how it differs to v3, it looks like `3.4.10` is the latest published version at https://www.npmjs.com/package/@guardian/consent-management-platform.

GHaberis:
@mxdvl from what I can see `v4` includes from https://github.com/guardian/frontend CCPA and TCFv2, this PR is for CCPA only, we will upgrade to TCFv2 in a following PR, which is when I think it'll be required to move to `v4` of `@guardian/consent-management-platform`.

support-frontend/assets/helpers/page/page.js
```diff
@@ -134,11 +132,22 @@ function init<S, A>(
...
+     * If ccpaEnabled initialise CCPA CMP as early
+     * as possible so subsequent call to
+     * onIabConsentNotification returns the correct
+     * consentState.
+    * */
+    if (ccpaEnabled()) {
+      initCMP({
+        useCcpa: true,
+      });
+    }
```
mxdvl, line 146: To update to V4:

```suggestion
    if (ccpaEnabled()) {
      init({
        inInUsa: ccpaEnabled(),
      });
    }
```

ripecosta: CMP v4 is still in beta so I would advise against using it until a stable and thoroughly tested version is released.

GHaberis: I've opted to release this with v3 for now and upgrade to v4 when we are ready to release TCFv2.

support-frontend/assets/components/consentBanner/consentBanner.jsx
```diff
@@ -1,79 +1,83 @@
 // @flow
-import React from 'react';
-import type { Dispatch } from 'redux';
-import { connect } from 'react-redux';
-import type { ThirdPartyTrackingConsent } from '../../helpers/tracking/thirdPartyTrackingConsent';
+import React, { Component } from 'react';
+import { getTrackingConsent, writeTrackingConsentCookie, type ThirdPartyTrackingConsent } from '../../helpers/tracking/thirdPartyTrackingConsent';
```
rupertbates, line 3: could just be `from 'helpers/tracking...` I think, I realise this was already here though so up to you if you want to tidy it up

support-frontend/assets/components/consentBanner/consentBanner.jsx
```diff
@@ -1,79 +1,83 @@
...
+type StateTypes = {
+  showBanner: boolean,
 }
 
-function mapStateToProps(state) {
-  return {
-    trackingConsent: state.common.trackingConsent,
-  };
-}
+class ConsentBanner extends Component<{}, StateTypes> {
```
rupertbates, line 20: You could use hooks to manage the state rather than a class as Tom F has done in [this PR](https://github.com/guardian/support-frontend/pull/2625)

support-frontend/assets/helpers/page/commonActions.js
```diff
@@ -18,7 +13,6 @@ export type SetCountryAction = { type: 'SET_COUNTRY_INTERNATIONALISATION', count
 export type Action =
   | SetCountryAction
   | { type: 'SET_EXISTING_PAYMENT_METHODS', existingPaymentMethods: ExistingPaymentMethod[] }
-  | { type: 'SET_TRACKING_CONSENT', trackingConsent: ThirdPartyTrackingConsent }
```
rupertbates, line 21: It's quite nice to have this just within the consent form rather than the global state I think

support-frontend/assets/helpers/page/page.js
```diff
@@ -134,11 +132,23 @@ function init<S, A>(
...
   thunk?: boolean = false,
 ): Store<*, *, *> {
-
   try {
+    /**
+     * Dynamically load @guardian/consent-management-platform
+     * on condition we're not server side rendering (ssr) the page.
+     * @guardian/consent-management-platform breaks ssr otherwise.
+     */
+    if (!getGlobal('ssr') && ccpaEnabled()) {
```
rupertbates, line 141: Could we check the user's country here to avoid loading the cmp when we don't need to?

GHaberis: That's what's happening, we only import `@guardian/consent-management-platform` when `ccpaEnabled()` is `true` (if the user is in the US).

support-frontend/assets/helpers/tracking/ccpa.js
```diff
@@ -0,0 +1,18 @@
...
+import {
+  detect as detectCountry,
+  type IsoCountry,
+} from 'helpers/internationalisation/country';
+import { isTestSwitchedOn } from 'helpers/globals';
+
+// ----- Functions ----- //
+
+export const ccpaEnabled = (): boolean => {
+  const useCCPA = isTestSwitchedOn('ccpaEnabled');
```
rupertbates, line 14: Why are we managing this as a test rather than a standard switch?

support-frontend/assets/helpers/tracking/thirdPartyTrackingConsent.js
```diff
@@ -1,6 +1,10 @@
 // @flow
 
+// import { onIabConsentNotification } from '@guardian/consent-management-platform';
```
rupertbates, line 3: Remove commented line

support-frontend/assets/helpers/page/page.js
rupertbates: Ah I've just seen that `ccpaEnabled` does check that 👍

support-frontend/scripts/ssrScript.js
```diff
@@ -12,6 +12,7 @@ global.URLSearchParams = require('url').URLSearchParams;
 global.window = window;
 global.window.guardian = {
   settings: {},
+  ssr: true,
```
rupertbates, line 15: This is a nice solution

support-frontend/assets/helpers/tracking/ccpa.js
GHaberis: It was suggested elsewhere I use `isTestSwitchedOn` but I agree a non-test switch seems more appropriate. Will set this up  now.

support-frontend/assets/components/consentBanner/consentBanner.jsx
GHaberis: As discussed this `consentBanner.jsx` will be deprecated shortly so I'm going to avoid further updates  in this PR.

PR #2571, opened 2020-06-25
Use emotion in layout

rupertbates:
## Why are you doing this?
A follow up to #2547 to replace the scss styling in the `CheckoutLayout` component with emotion.

There is still one bit of scss used by the related `Content` component which is a bit more complicated to remove because of the way that it applies a multiline border to all of it's children but this gets us most of the way there.


support-frontend/assets/components/subscriptionCheckouts/layoutStyles.js
```diff
@@ -0,0 +1,49 @@
+// @flow
+
+import { css } from '@emotion/core';
+import { from } from '@guardian/src-foundations/mq';
+import { space } from '@guardian/src-foundations';
+import { neutral } from '@guardian/src-foundations/palette';
+
+export const mainCss = css`
+  max-width: ${space[280]}px;
```
lucymonie, line 9: Does this work? This isn't how I understand `space`, but I may have misunderstood!

rupertbates: No it didn't! Have fixed it.

PR #2572, opened 2020-06-25
refactor email state to prepare for corp subs email

johnduffell:
## Why are you doing this?

This email is a precursor to adding the corporate subs email. PR: https://github.com/guardian/support-frontend/pull/2567
This makes the state more accurately represent the possible states, meaning fewer invalid states and fewer "this should never happen" situations as they will be compiler errors.  This reduces the "cry wolf" effect so any Option or Either are more likely to be valid states that we should deal with properly.

I also tidied up the email fields classes as many of the parameters are the same for all emails, but we were passing them in separately for each email.

### Changes
1. move payment schedule to the "paid"/non redemption side of the Either
1. add case class to wrap them both with serliasers and test updates
1. tidy up *EmailFields classes to not pass the same parameter in duplicated across all classes
1. extract method on `subscribe`/`skipSubscribe` pattern match in CreateZuoraSubscription lambda
1. and inline some trivial methods into the new method.


support-workers/src/main/scala/com/gu/emailservices/ContributionEmailFields.scala
```diff
@@ -3,56 +3,56 @@ package com.gu.emailservices
...
-      "account name" -> dd.bankTransferAccountName,
-      "account number" -> mask(dd.bankTransferAccountNumber),
-      "sort code" -> hyphenate(dd.bankCode),
-      "Mandate ID" -> dd.mandateId,
-      "first payment date" -> formatDate(created.plusDays(10).toLocalDate),
-      "payment method" -> "Direct Debit"
-    )
-    case Left(_: PayPalReferenceTransaction) => List("payment method" -> "PayPal")
-    case Left(_: CreditCardReferenceTransaction) => List("payment method" -> "credit / debit card")
-    case Right(_: RedemptionData) => Nil
```
johnduffell, line 42: this was an invalid state as there's no such thing as a redeemable contribution, so this will fail the lambda earlier and cause a support worker failure now

support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
```diff
@@ -38,35 +39,41 @@ import com.gu.support.workers._
...
+    case dd: ClonedDirectDebitPaymentMethod => List(
       "Sort Code" -> hyphenate(dd.bankCode),
       "Account number" -> mask(dd.bankTransferAccountNumber),
       "Account Name" -> dd.bankTransferAccountName,
       "Default payment method" -> "Direct Debit",
       "MandateID" -> dd.mandateId
     )
-    case Left(_: CreditCardReferenceTransaction) => List("Default payment method" -> "Credit/Debit Card")
-    case Left(_: PayPalReferenceTransaction) => Seq("Default payment method" -> "PayPal")
-    case Right(redemptionData) => Nil
```
johnduffell, line 69: this will come back in the next PR...I just cleaned them all up in a consistent way in this PR!

support-workers/src/main/scala/com/gu/emailservices/EmailFields.scala
```diff
@@ -1,16 +1,17 @@
 package com.gu.emailservices
 
+import com.gu.i18n.Currency
 import com.gu.salesforce.Salesforce.SfContactId
+import com.gu.support.promotions.Promotion
+import com.gu.support.workers.{BillingPeriod, User}
 import io.circe.generic.auto._
 import io.circe.syntax._
 
-// scalastyle:off
```
johnduffell, line 7: scalastyle doesn't seem to complain or edit this block when I compile, so I removed it

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -65,45 +65,39 @@ object CreateZuoraSubscription {
...
           PreviewPaymentSchedule(subscribeItem, state.product.billingPeriod, zuoraService, checkSingleResponse)
-            .withLogging("PreviewPaymentSchedule").map(ps => (pm, ps))
+            .withLogging("PreviewPaymentSchedule").map(ps => PaymentMethodWithSchedule(pm, ps))
         ).leftSequence
-      thankYouState <- maybeDomainSubscription match {
-        case Some(domainSubscription) => skipSubscribe(state, requestInfo, paymentMethodWithPaymentSchedule, domainSubscription)
-          .withLogging("skipSubscribe")
-        case None => subscribe(subscribeItem, zuoraService).map(response => toHandlerResult(state, response, paymentMethodWithPaymentSchedule, requestInfo))
-          .withLogging("subscribe")
-      }
```
johnduffell, line 83: I'm think there was a comment to extract this to a method in a previous PR, that PR was already big so I did it here.  It's only in this file so easy to revert if desired though.

support-workers/src/main/scala/com/gu/support/workers/lambdas/SendThankYouEmail.scala
```diff
@@ -38,8 +38,8 @@ class SendThankYouEmail(thankYouEmailService: EmailService, servicesProvider: Se
     } yield HandlerResult(emailResult, requestInfo)
   }
 
-  def fetchDirectDebitMandateId(state: SendThankYouEmailState, zuoraService: ZuoraService): Future[Option[String]] = state.paymentMethod match {
-    case Left(_: DirectDebitPaymentMethod) =>
+  def fetchDirectDebitMandateId(state: SendThankYouEmailState, zuoraService: ZuoraService): Future[Option[String]] = state.paymentOrRedemptionData match {
+    case Left(_: PaymentMethodWithSchedule) =>
```
johnduffell, line 42: todo

support-workers/src/main/scala/com/gu/support/workers/lambdas/SendThankYouEmail.scala
```diff
@@ -50,71 +50,57 @@ class SendThankYouEmail(thankYouEmailService: EmailService, servicesProvider: Se
...
+    val productRatePlanId = getProductRatePlanId(state.product, state.user.isTestUser, state.giftRecipient.isDefined)
     val maybePromotion = getAppliedPromotion(
       servicesProvider.forUser(state.user.isTestUser).promotionService,
       state.promoCode,
       state.user.billingAddress.country,
       productRatePlanId
     )
 
-    thankYouEmailService.send(
+    val subscriptionEmailFields: Either[String, SubscriptionEmailFields] = state.paymentOrRedemptionData.left.toOption.toRight("can't send a corporate/gift email yet").map(paymentMethodWithSchedule =>
```
johnduffell, line 61: this will change in the next pr

support-models/src/main/scala/com/gu/support/workers/states/SendThankYouEmailState.scala
```diff
@@ -13,13 +13,12 @@ case class SendThankYouEmailState(
   user: User,
   giftRecipient: Option[GiftRecipient],
   product: ProductType,
-  paymentMethod: Either[PaymentMethod, RedemptionData],
+  paymentOrRedemptionData: Either[PaymentMethodWithSchedule, RedemptionData],
```
rupertbates, line 16: Definitely agree with renaming this field, however I think we'll need to support the old schema in the deserialiser because we have 2 separate deploys for support-frontend and support-workers

johnduffell: this is only between two states of support workers though so I think it would be ok.
I did think about that if there are checkouts in flight at that moment, but I decided to leave it.  Worse case we will get a failed email if we are really unlucky, we can trigger the lambda manually.
Does that make sense?

rupertbates: Oh yes you're right 👍

johnduffell: so in the end you were right, but not in the way you thought!
Two things
1. SendThankyouEmailState and SendAcquisitionEventState both rely on the same fields, so everything seemed to work but the lambda was actually looping away in the background.  I managed to fix that beforing finding out that....
1. ... FailureHandler relies on paymentFields OR paymentMethod (although all it really needs is the PaymentProvider).  I think the solution would be to add that to the state and pass it through, rather than adding a third special case.

The fix should be independent of this PR so I will do a third PR (tomorrow), to go before this one, with the fixes, to keep it clear and easy to review!

johnduffell: for the record the third PR was here: https://github.com/guardian/support-frontend/pull/2577

PR #2574, opened 2020-06-25
don't show the auth screen on an internal server error

johnduffell:
## Why are you doing this?

When we use the maybe identity auth function, and there is a failure in the controller (our normal logic) this causes the identity register/signin screen to show.

This PR fixes it so it's only identity failures that cause the login screen.

This will also affect failures in mandatory authed screens, so a 5xx error in the checkout for e.g. GW, will correctly show a 5xx error page rather than redirecting to login. Hopefully we are not relying on that anywhere, but at the moment we could be having 100% 5xx error on any page behind identity , and the 5xx alarm woudln't go off.

optional auth:
image

### DIFF
--- a/support-frontend/app/actions/AsyncAuthenticatedBuilder.scala
+++ b/support-frontend/app/actions/AsyncAuthenticatedBuilder.scala
@@ -1,10 +1,13 @@
 package actions
 
+import cats.data.EitherT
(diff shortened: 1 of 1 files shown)


johnduffell:
technically you spotted it first Rupert 😂

PR #2577, opened 2020-06-26
Add tests and type checking for analytics and failure states de/serialisation

johnduffell:
## Why are you doing this?

I did some changes to make the payment schedule only fetched for paid subs and not corporate subs, to pass on to the send-email lambda.
When testing in CODE everything seemed to work and I got the normal thank you page, but I luckily noticed before merging there were a load of state machine executions still in "running" state.  The Analytics lambda was looping.
On investigation it turned out the SendEmail and SendAnalytics lambdas use the same input, but the analytics one uses a separate deserialiser, and there are no unit or integration tests to pick it up.

This PR adds type checking and unit tests for this.
I also noticed the FailureHandler is supposed to be able to deserialise any of the States successfully, so I added similar type checks and unit tests for that too.

This means people should realise at compile time that there is a link, and even people write a custom de/serialiser, the tests should pick it up.  The only risk is that a new kind of state might be added in future, and not added to the tests.

There is also a simplification to the FailureHandler in advance of the new PaymentMethodWithSchedule type coming in the email state in https://github.com/guardian/support-frontend/pull/2572/files#r445743286 - we would have needed a third Option in the FailureState to continue the current approach, so I've just unified it into a single simple non-optional field.

NOTE: For this PR, we need to ship the support-frontend change FIRST and then the support-workers.  This is because the extra field needs to be in place before support workers looks for it.


support-models/src/main/scala/com/gu/support/workers/states/FailureHandlerState.scala
```diff
@@ -12,11 +12,19 @@ case class FailureHandlerState(
   user: User,
   giftRecipient: Option[GiftRecipient],
   product: ProductType,
-  paymentFields: Option[Either[PaymentFields, RedemptionData]], //Will be present if CreatePaymentMethod failed
-  paymentMethod: Option[Either[PaymentMethod, RedemptionData]], //Will be present if anything after CreatePaymentMethod failed
```
johnduffell, line 16: FailureHandler needs the PaymentProvider to log to the ELK stack.
It extracted it from the paymentFields/paymentMethod, and passed it on (as an option) on to the ELK stack. 
 paymentprovider is not actually optional, it's just due to how the types worked out and the way it grabbed it.  

With the new PaymentMethodWithSchedule type coming in the email state in https://github.com/guardian/support-frontend/pull/2572/files#r445743286 - I could have would have added a third Option in the FailureState to continue the current approach.

Instead I've just unified it into a single simple non-optional field.

support-models/src/main/scala/com/gu/support/workers/states/FailureHandlerState.scala
```diff
@@ -12,11 +12,19 @@ case class FailureHandlerState(
...
-) extends StepFunctionUserState
+) extends FailableState
+
+trait FailableState extends StepFunctionUserState {
+  // only required fields needed here
+  def requestId: UUID
+  def user: User
+  def giftRecipient: Option[GiftRecipient]
+  def product: ProductType
+  def paymentProvider: PaymentProvider
```
johnduffell, line 26: this will force all input states that could fail, to provide the necessary fields

support-models/src/main/scala/com/gu/support/workers/PaymentProvider.scala
```diff
@@ -0,0 +1,63 @@
...
+
+object PaymentProvider {
+
+  val all = List(
+    Stripe,
+    StripeApplePay,
+    PayPal,
+    DirectDebit,
+    Existing,
+    RedemptionNoProvider
```
johnduffell, line 28: moved from support-workers project - only this variable was updated as it used to be 
```scala
List(Stripe, PayPal, DirectDebit, Existing)
```

support-models/src/main/scala/com/gu/support/workers/states/SendAcquisitionEventState.scala
```diff
@@ -6,22 +6,37 @@ import com.gu.support.encoding.CustomCodecs._
...
+
+trait SendAcquisitionEventState extends FailableState {
+  def requestId: UUID
+  def giftRecipient: Option[GiftRecipient]
+  def product: ProductType
+  def paymentMethod: Either[PaymentMethod, RedemptionData]
+  def firstDeliveryDate: Option[LocalDate]
+  def promoCode: Option[PromoCode]
+  def acquisitionData: Option[AcquisitionData]
+}
```
johnduffell, line 22: this trait forces the output of the previous, createZuora, step (namely the SendEmail state) to provide the correct fields for the acquisition state.
Unfortunately you can't inherit a case class and you can't deserialise a trait with the semiauto, so the trait is a duplicate of the Impl below.  However at least they are right beside each other.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreatePaymentMethod.scala
```diff
@@ -65,6 +65,7 @@ class CreatePaymentMethod(servicesProvider: ServiceProvider = ServiceProvider)
       state.user,
       state.giftRecipient,
       state.product,
+      PaymentProvider.fromPaymentMethod(paymentMethod.left.toOption),
```
johnduffell, line 68: when serialising the state, it just looks up the correct PaymentProvider from the payment method.

Other approaches could be to pass it through like we do for everything else `state.paymentProvider` or we could even `map` the encoder to add it to the JSON if we want to be clever (the latter would lose the compile safety but we could still unit test it)

PR #2581, opened 2020-06-30
don't use PROD in the name of the Redemption tables

johnduffell:
## Why are you doing this?

We called the redemption tables redemption-codes-((tpstage))-((gustage))
This got a bit silly with redemption-codes-PROD-PROD
So this PR takes out the stage if it's gu:PROD

[**Trello Card**](https://trello.com/c/02NsmWw6/3081-update-redemption-tables-to-not-have-a-code-prod-stage)

When we ship this, we need to manually deploy the tables with riffraff.
Support frontend/workers will be temporarily not working for corporate subs in between the deploys.
The IT tests are broken in this branch until the tables are updated in gu:PROD (as we use the gu:PROD tp:DEV table for the tests)


johnduffell:
this is deployed to PROD with riffraff and looks good
image

PR #2582, opened 2020-07-01
Tp support landing page overflow bug fix

TomPretty:
## Why are you doing this?

Fixes an issue with the design system choice cards that led to an overflow at certain screen sizes.

We will make a fix to the design system when we have time and then should be able to remove this.

Screenshot 2020-06-29 at 15 39 32

## Screenshots

Screenshot 2020-07-01 at 10 04 52

Screenshot 2020-07-01 at 10 04 40

Screenshot 2020-07-01 at 10 04 25

Screenshot 2020-07-01 at 10 04 15

Screenshot 2020-07-01 at 10 03 47

### DIFF
--- a/support-frontend/build-tc
+++ b/support-frontend/build-tc
@@ -3,6 +3,12 @@
 # Exit if any of these commands fail
 set -e
 
+export NVM_DIR="$HOME/.nvm"
+[[ -s "$NVM_DIR/nvm.sh" ]] && . "$NVM_DIR/nvm.sh"  # This loads nvm
+
+nvm install
+nvm use
+
 # Installing yarn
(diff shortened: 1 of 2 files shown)


support-frontend/build-tc
```diff
@@ -3,6 +3,12 @@
 # Exit if any of these commands fail
 set -e
 
+export NVM_DIR="$HOME/.nvm"
```
tomrf1, line 6: This is unrelated to the change. Teamcity's nvm step has stopped working, so we're doing it in the build script instead

tomrf1:
@SiAdcock FYI

PR #2585, opened 2020-07-02
remove survey from corp subs thank you pages

johnduffell:
## Why are you doing this?

The digi sub thank you page survey is designed for people who have made a purchase decision, not for corporate recipients.

This PR removes the digi sub survey from anyone who hasn't paid for their subscription at checkout time.

[**Trello Card**](https://trello.com/c/3JbmYsrN/3157-remove-survey-link-from-corporate-thank-you-page)

Tested in DEV on the thankyou, thankyou-pending (not present) and the normal paid thankyou (present)

### DIFF
--- a/support-frontend/assets/pages/digital-subscription-checkout/digitalSubscriptionCheckout.jsx
+++ b/support-frontend/assets/pages/digital-subscription-checkout/digitalSubscriptionCheckout.jsx
@@ -72,7 +72,7 @@ const content = (
     >
       <CheckoutStage
         checkoutForm={<CheckoutForm />}
-        thankYouContentPending={<ThankYouPendingContent {...thankyouProps} />}
+        thankYouContentPending={<ThankYouPendingContent includePaymentCopy={true} {...thankyouProps} />}
         thankYouContent={<ThankYouContent {...thankyouProps} />}
         subscriptionProduct="DigitalPack"
       />
--- a/support-frontend/assets/pages/subscriptions-redemption/subscriptionsRedemption.jsx
+++ b/support-frontend/assets/pages/subscriptions-redemption/subscriptionsRedemption.jsx
@@ -52,7 +52,7 @@ const content = (
     >
       <CheckoutStage
         checkoutForm={<RedemptionForm />}
-        thankYouContentPending={<ThankYouPendingContent {...thankyouProps} />}
+        thankYouContentPending={<ThankYouPendingContent includePaymentCopy={false} {...thankyouProps} />}
         thankYouContent={<ThankYouContent {...thankyouProps} />}
       />
       <ConsentBanner />
--- a/support-frontend/assets/pages/digital-subscription-checkout/thankYouContent.jsx
+++ b/support-frontend/assets/pages/digital-subscription-checkout/thankYouContent.jsx
@@ -66,7 +66,10 @@ function ThankYouContent(props: PropTypes) {
         </Text>
         <AppsSection countryGroupId={props.countryGroupId} />
       </Content>
-      <SubscriptionsSurvey product={DigitalPack} />
+      {props.includePaymentCopy ?
+        <SubscriptionsSurvey product={DigitalPack}/>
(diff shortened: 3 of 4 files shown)


support-frontend/assets/pages/subscriptions-redemption/subscriptionsRedemption.jsx
```diff
@@ -52,7 +52,7 @@ const content = (
     >
       <CheckoutStage
         checkoutForm={<RedemptionForm />}
-        thankYouContentPending={<ThankYouPendingContent {...thankyouProps} />}
+        thankYouContentPending={<ThankYouPendingContent includePaymentCopy={false} {...thankyouProps} />}
```
GHaberis, line 55: You could add `includePaymentCopy: false` to the `thankyouProps` object and save having to put an inline prop here.

support-frontend/assets/pages/subscriptions-redemption/subscriptionsRedemption.jsx
```diff
@@ -52,7 +52,7 @@ const content = (
...
         thankYouContent={<ThankYouContent {...thankyouProps} />}
```
GHaberis, line 56: Do we want to pass this `includePaymentCopy` prop to `< ThankYouContent />` too? If so you could add `includePaymentCopy: false` to the `thankyouProps` object and save having to put an inline prop here.

support-frontend/assets/pages/digital-subscription-checkout/digitalSubscriptionCheckout.jsx
```diff
@@ -72,7 +72,7 @@ const content = (
     >
       <CheckoutStage
         checkoutForm={<CheckoutForm />}
-        thankYouContentPending={<ThankYouPendingContent {...thankyouProps} />}
+        thankYouContentPending={<ThankYouPendingContent includePaymentCopy={true} {...thankyouProps} />}
         thankYouContent={<ThankYouContent {...thankyouProps} />}
```
GHaberis, line 76: Do we want to pass this includePaymentCopy prop to < ThankYouContent /> too? If so you could add includePaymentCopy: false to the thankyouProps object and save having to put an inline prop here.

support-frontend/assets/pages/subscriptions-redemption/subscriptionsRedemption.jsx
johnduffell: no that one seems to actually be an indirect thankYouContainer.jsx that wraps the actual ThankYouContent and adds a bunch of properties to the map including this one.  The "pending" one seems to directly call the view.  I thought about using the same technique, but it seemed overkill for just a single property, hence why I inlined it.
I'm not sure whether one of them should be renamed to make it clear one is direct and other is a container, or whether I should have made both a container?

support-frontend/assets/pages/digital-subscription-checkout/digitalSubscriptionCheckout.jsx
johnduffell: it would be `true` rather than false in this case, but the answer is the same as I mentioned in the other case.

support-frontend/assets/pages/subscriptions-redemption/subscriptionsRedemption.jsx
GHaberis: OK I see what you're saying `<ThankYouContent />` is a wrapper to `thankYouContainer.jsx` and this wrapper already passes the `includePaymentCopy` property so no need to include it here. It seems overkill to replicate it. In which case this is fine as is.

support-frontend/assets/pages/digital-subscription-checkout/digitalSubscriptionCheckout.jsx
GHaberis: Based on comments elsewhere this is fine as is.

PR #2587, opened 2020-07-03
allow ssm tunnel for ssh into elk kibana logs proxy server

johnduffell:
## Why are you doing this?

So we can SSH into elk proxy without using the bastion, which is more secure and works off VPN.
https://github.com/guardian/ssm-scala/blob/master/README.md#in-aws

### DIFF
--- a/support-logs/cfn.yaml
+++ b/support-logs/cfn.yaml
@@ -51,6 +51,10 @@ Resources:
           - ssm:ListInstanceAssociations
           - ssm:DescribeInstanceProperties
           - ssm:DescribeDocumentParameters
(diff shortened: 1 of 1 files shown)


support-logs/readme.md
```diff
@@ -36,3 +38,12 @@ PUT _ingest/pipeline/extractJsonPipeline
     ]
 }
 ```
+
+## Setup
```
rupertbates, line 42: Maybe add a link to information about SSM in here?

johnduffell: good idea, will do

PR #2610, opened 2020-07-13
Remove live app price

rupertbates:
## Problem 

1) Premium app prices in non-uk countries are out of date (due to app stores changing prices when the exchange rate changes)

2) In EU, Canada, and NZ, the price is not shown, instead a '7 day free trial' message is shown. 

## Solution

1) remove price from all regions
2) only show trial offer in all regions . ie: '7 day free trial' 

## Definition of Success

No prices, and only 7 day free trial offer

[**Trello Card**](https://trello.com/c/9959HmKl/3116-update-showcase-page-remove-prices-only-show-trial-offer-consistent-in-all-regions)

## Screenshots
support thegulocal com_uk_subscribe(iPhone X)

### DIFF
--- a/support-frontend/assets/helpers/subscriptions.js
+++ b/support-frontend/assets/helpers/subscriptions.js
@@ -66,35 +66,8 @@ export const subscriptionPricesForDefaultBillingPeriod: {
     [CountryGroupId]: number,
   }
 } = {
-  PremiumTier: {
-    GBPCountries: 5.99,
-    UnitedStates: 6.99,
-    AUDCountries: 7.99,
-    International: 5.99,
-  },
-  DigitalPack: {
-    GBPCountries: 11.99,
-    UnitedStates: 19.99,
-    AUDCountries: 21.50,
-    International: 19.99,
-  },
-  GuardianWeekly: {
-    GBPCountries: 37.50,
-    EURCountries: 61.30,
-    UnitedStates: 75,
-    Canada: 80,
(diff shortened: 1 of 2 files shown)


support-frontend/assets/helpers/subscriptions.js
```diff
@@ -66,36 +66,9 @@ export const subscriptionPricesForDefaultBillingPeriod: {
     [CountryGroupId]: number,
   }
 } = {
-  PremiumTier: {
```
rupertbates, line 69: None of these prices are used anymore

support-services/src/test/scala/com/gu/support/redemption/DynamoTableAsyncITSpec.scala
```diff
@@ -18,6 +18,7 @@ class DynamoTableAsyncITSpec extends AsyncFlatSpec with Matchers {
       lookedup should be(Some(Map(
         "redemptionCode" -> DynamoString(allCharsCode),
         "available" -> DynamoBoolean(true),
+        "type" -> DynamoString("Corporate"),
```
rupertbates, line 21: This is unrelated to this PR but this IT test was failing so I fixed it

support-frontend/assets/helpers/subscriptions.js
lucymonie: Good to know, thanks

PR #2612, opened 2020-07-14
Change copy when Aus moment goal reached

tomrf1:
Before goal:
Screen Shot 2020-07-14 at 15 41 01

After goal:
Screen Shot 2020-07-14 at 15 40 40

### DIFF
--- a/support-frontend/assets/components/legal/termsPrivacy/termsPrivacy.jsx
+++ b/support-frontend/assets/components/legal/termsPrivacy/termsPrivacy.jsx
@@ -8,8 +8,8 @@ import { privacyLink, contributionsTermsLinks, philanthropyContactEmail } from '
 import type { CountryGroupId } from 'helpers/internationalisation/countryGroup';
 import { type IsoCurrency, fromCountryGroupId, currencies } from 'helpers/internationalisation/currency';
 import type { ContributionType } from 'helpers/contributions';
-import { type CampaignName, campaigns } from 'helpers/campaigns';
 import './termsPrivacy.scss';
+import type { CampaignSettings } from 'helpers/campaigns';
 
 
 // ---- Types ----- //
@@ -17,7 +17,7 @@ import './termsPrivacy.scss';
 type PropTypes = {|
   countryGroupId: CountryGroupId,
   contributionType: ContributionType,
-  campaignName: ?CampaignName,
+  campaignSettings: ?CampaignSettings,
 |};
 
 // ----- Component ----- //
@@ -72,8 +72,8 @@ function TermsPrivacy(props: PropTypes) {
     </div>
   );
 
-  if (props.campaignName && campaigns[props.campaignName] && campaigns[props.campaignName].termsAndConditions) {
-    return campaigns[props.campaignName].termsAndConditions(
+  if (props.campaignSettings && props.campaignSettings.termsAndConditions) {
+    return props.campaignSettings.termsAndConditions(
       contributionsTermsLinks[props.countryGroupId],
       philanthropyContactEmail[props.countryGroupId],
     );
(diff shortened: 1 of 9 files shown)


support-frontend/assets/pages/digital-subscription-checkout/digitalSubscriptionCheckout.jsx
```diff
@@ -72,7 +72,7 @@ const content = (
     >
       <CheckoutStage
         checkoutForm={<CheckoutForm />}
-        thankYouContentPending={<ThankYouPendingContent includePaymentCopy={true} {...thankyouProps} />}
+        thankYouContentPending={<ThankYouPendingContent includePaymentCopy {...thankyouProps} />}
```
tomrf1, line 75: just letting the linter do its thing here

support-frontend/assets/helpers/campaigns.jsx
```diff
@@ -19,17 +19,18 @@ export type CampaignSettings = {
...
-  [string]: CampaignSettings,
+  [string]: (goalReached: boolean) => CampaignSettings,
 };
 
 const currentCampaignName = ausMomentEnabled('AU') ? 'au/contribute' : null;
 
 export const campaigns: Campaigns = currentCampaignName ? {
-  // TODO - the rest of the campaign settings
-  [currentCampaignName]: {
+  [currentCampaignName]: (goalReached: boolean) => ({
```
paulbrown1982, line 28: The type of this value not the same between the two values of `currentCampaignName`. If it's true, the return is a function, as in the definition, if false, it's an (empty) object like it used to be.

tomrf1: good spot! You are clearly a better type checker than flow is

tomrf1: ah wait, it is correct actually.
If `currentCampaignName` is `null` then we set `campaigns` to be an empty object.

But - this is all a bit needlessly complicated, I will simplify it

tomrf1: I've just pushed a refactor of the campaigns logic, hopefully it's easier to reason about now!

support-frontend/assets/helpers/campaigns.jsx
```diff
@@ -18,46 +22,46 @@ export type CampaignSettings = {
...
 };
 
-export type Campaigns = {
-  [string]: CampaignSettings,
-};
-
-const currentCampaignName = ausMomentEnabled('AU') ? 'au/contribute' : null;
+const currentCampaignPath: string | null = ausMomentEnabled('AU') ? 'au/contribute' : null;
 
-export const campaigns: Campaigns = currentCampaignName ? {
```
tomrf1, line 27: The `campaigns` object was creating unnecessary compexity. I've replaced it with a single `campaign` const, which is returned by `getCampaignSettings()` if the user is on the right url and the campaign is switched on

support-frontend/assets/components/legal/termsPrivacy/termsPrivacy.jsx
```diff
@@ -8,16 +8,16 @@ import { privacyLink, contributionsTermsLinks, philanthropyContactEmail } from '
...
+import type { CampaignSettings } from 'helpers/campaigns';
 
 
 // ---- Types ----- //
 
 type PropTypes = {|
   countryGroupId: CountryGroupId,
   contributionType: ContributionType,
-  campaignName: ?CampaignName,
+  campaignSettings: ?CampaignSettings,
```
paulbrown1982, line 20: As discussed, this would be better as `campaignSettings: CampaignSettings | null`

tomrf1: 👍

PR #2613, opened 2020-07-15
Stripe country problem

rupertbates:
## Why are you doing this?
Since we moved to using Stripe elements inline we have been initialising the country with the standard internationalisation country from `state.common.internationalisation.countryId`. 

For Guardian Weekly however, we should be using the billing country as this can be different from the internationalisation country and is needed to work out the correct Stripe account to use.

[**Trello Card**](https://trello.com/c/Z5cL7KSf/3177-gw-gifts-sometimes-use-wrong-stripe-account-client-side)

### DIFF
--- a/support-frontend/assets/helpers/subscriptionsForms/formActions.js
+++ b/support-frontend/assets/helpers/subscriptionsForms/formActions.js
@@ -95,7 +95,6 @@ const formActionCreators = {
     return dispatch({
       type: 'SET_PAYMENT_METHOD',
       paymentMethod,
-      country: state.common.internationalisation.countryId,
     });
   },
   setBillingAddressIsSame: (isSame: boolean): Action => ({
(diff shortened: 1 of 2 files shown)


support-frontend/assets/helpers/subscriptionsForms/formActions.js
```diff
@@ -95,7 +95,6 @@ const formActionCreators = {
     return dispatch({
       type: 'SET_PAYMENT_METHOD',
       paymentMethod,
-      country: state.common.internationalisation.countryId,
```
rupertbates, line 98: This is never actually used and is not defined in the type

support-frontend/assets/pages/weekly-subscription-checkout/components/weeklyCheckoutForm.jsx
```diff
@@ -308,7 +306,7 @@ function WeeklyCheckoutForm(props: PropTypes) {
             title="Your card details"
           >
             <StripeProviderForCountry
-              country={props.country}
+              country={props.billingCountry}
```
lucymonie, line 309: This seems to be the critical point!

PR #2614, opened 2020-07-15
don't access S3 during class loading/initialisation

johnduffell:
## Why are you doing this?

When running the tests, the DEV config is loaded from S3.

At present this happens from the class loading process (various vals in objects) which is a static initialiser once it gets to bytecode.

This means that if it fails there is an [ExceptionInInitializerError](https://docs.oracle.com/javase/8/docs/api/java/lang/ExceptionInInitializerError.html) which is a [LinkageError](https://docs.oracle.com/javase/8/docs/api/java/lang/LinkageError.html) and therefore not caught by [NonFatal](https://www.scala-lang.org/api/2.12.0/scala/util/control/NonFatal$.html)
Futures only catch non fatal errors when flatmapping or otherwise, if there is a fatal error, the exception is propagated to the thread's exception handler, and the future is never completed.  This means that the tests hang as things are not cleaned up in the future chain.

The reason vals are involved at all, are because config loading is an expensive process, and having a static val means the result is cached between warm lambda executions.

This is something useful to fix anyway, but even more so if we intend to automate the IT tests, as hanging is always a tricky one to deal with.

## Changes
This PR moves the vals to be in a class initialiser instead and replaces them with a single lazy val in order to cache the result.
The advantage over just making all of them lazy, is mainly that it provides a clear direction of dependency and reduces the risk of later addition of vals undoing the work.  If everything is a lazy val, things can get complicated as they can depend on each other, and get [deadlocks](https://github.com/guardian/membership-common/pull/421) or null pointer exceptions depending whether futures are involved or not.  Therefore it is a good idea to [minimise the number of lazy vals](https://docs.scala-lang.org/sips/improved-lazy-val-initialization.html) in the code.

### Before:
SBT
image
Intellij
image

### After:
SBT
image
Intellij
image


support-workers/src/main/scala/com/gu/config/Configuration.scala
```diff
@@ -4,11 +4,14 @@ import com.gu.config.loaders.PrivateConfigLoader
...
 import com.gu.salesforce.SalesforceConfigProvider
 import com.gu.support.config._
-import com.typesafe.config.ConfigFactory
+import com.typesafe.config.{Config, ConfigFactory}
 
 import scala.util.Try
 
 object Configuration {
+
+  def load(): Configuration = new Configuration(loadConfig)
```
johnduffell, line 13: it only actually loads the first time, but I wasn't sure of the best name.  It should only be called once "at the edge" anyway.

johnduffell:
tested in CODE and a small fix made!

support-workers/src/main/scala/com/gu/config/Configuration.scala
```diff
@@ -19,9 +22,16 @@ object Configuration {
 
   SafeLogger.info(s"Load from S3: $loadFromS3, Stage: $stage")
 
-  val config = PrivateConfigLoader
-    .forEnvironment(loadFromS3)
-    .load(stage, ConfigFactory.load())
+  // this is static so it persists between lambda executions, but lazy so it doesn't cause a fatal error at class loading time
+  private lazy val loadConfig = PrivateConfigLoader
+    .forEnvironment(Configuration.loadFromS3)
+    .load(Configuration.stage, ConfigFactory.load())
```
johnduffell, line 28: configfactory.load uses the current thread context classloader.  However this is not recommended when running code within SBT (test or run) as in a Future pool `the context classloader of the thread that backs the future is the scala library classloader which is not able to load project classes`
https://www.scala-sbt.org/1.x/docs/In-Process-Classloaders.html#Troubleshooting

So I need to work out what that really means, it seems bizarre that typesafe config and sbt tests wouldn't work in a Future, so there must be a proper solution.

rupertbates: 😱

johnduffell: I have fixed it to specifiy the classloader used to load the current class.  I discovered someone else hit the same issue with sbt 1.3 (sbt does fancy classloader stuff for perf reasons)
https://www.jan0sch.de/post/typesafe-config-sbt-classloader-issues/

johnduffell: the current thread classloader seems to be designed so that parent application servers can load things from the child classloader, so for example play loading application config.  That may explain why typesafe config defaults to loading from the thread loader instead of the current one.
More reading if anyone is interested:
https://www.infoworld.com/article/2077344/find-a-way-out-of-the-classloader-maze.html

johnduffell:
tested in CODE (DS and contribution) and also all IT tests passing locally.

PR #2616, opened 2020-07-17
UK amounts test

tomrf1:


tomrf1:
@michaelbjacobson yes that's a good point, it worries me too. Really the minor and major parts should be stored properly.

michaelbjacobson:
> @michaelbjacobson yes that's a good point, it worries me too. Really the minor and major parts should be stored properly.

The simplest thing would probably be to simply store all currency values as integers, representing the total number of cents/pennies, and then format correctly when displaying.

tomrf1:
@michaelbjacobson Yep. I've just had a look through the code at how this value gets passed around and I'm not feeling especially confident, despite successfully testing it. I'm going to remove that variant from the test for now, and we can consider proper representation of currencies in the client another time

PR #2618, opened 2020-07-20
use new email identifier for imovo thank you email

johnduffell:
## Why are you doing this?

This PR makes paper voucher use the new subscription card email identifier once the switch has been turned on.

Depends on https://github.com/guardian/membership-workflow/pull/266

[**Trello Card**](https://trello.com/c/pEUPHrCF/3171-print-thank-you-email-for-imovo-subs-card)

This means that once the switch is on we will send the new email, meaning that new customers will get information about the subscription card rather than about the old paper voucher.

I will test by shipping the workflow change and then run this PR in DEV with the switch turned on, or at least an if statement


johnduffell:
I was going to test in CODE, but since we are doing an end to end test soon and the change is pretty trivial I decided to merge anyway.

PR #2619, opened 2020-07-21
Fix image overlapping text in subscription CTA

i-hardy:
## Why are you doing this?

From the tablet breakpoint to screens about 1700px wide, the image in the 'Subscribe' CTA overlaps the text and cuts it off to varying degrees

[**Trello Card**](https://trello.com/c/Pay20ZIg)

## Changes

* Added a max-width rule for the text in the CTA. I experimented with a couple of other approaches but they all required much more significant overriding of the Content component CSS or creating new components; this was the simplest fix.

## Screenshots

**Tablet - Vertical**

Screenshot 2020-07-21 at 11 17 35

**Tablet - Horizontal**

Screenshot 2020-07-21 at 11 17 51

**Mobile (large) - Horizontal**

Screenshot 2020-07-21 at 11 20 14

**Desktop**

Screenshot 2020-07-21 at 11 56 58

### DIFF
--- a/support-frontend/assets/pages/showcase/showcase.scss
+++ b/support-frontend/assets/pages/showcase/showcase.scss
@@ -115,7 +115,7 @@
 
       @include mq($from: tablet, $until: leftCol) {
         .component-text {
-          max-width: 80%;
+          max-width: 75%;
         }
 
         .component-content__image {
--- a/support-frontend/assets/pages/showcase/showcase.scss
+++ b/support-frontend/assets/pages/showcase/showcase.scss
@@ -113,19 +113,23 @@
         }
       }
 
-      @include mq($from: tablet, $until: desktop) {
+      @include mq($from: tablet, $until: leftCol) {
+        .component-text {
+          max-width: 80%;
+        }
+
         .component-content__image {
           right: -200px;
         }
-
+        
         .component-left-margin-section__content .component-grid-image {
           height: 120%;
         }
       }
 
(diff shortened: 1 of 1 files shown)


support-frontend/assets/pages/showcase/showcase.scss
```diff
@@ -113,19 +113,23 @@
...
+      @include mq($from: tablet, $until: leftCol) {
+        .component-text {
+          max-width: 75%;
+        }
+
         .component-content__image {
           right: -200px;
         }
-
+
```
johnduffell, line 124: not sure which editor you're using but watch out for stray white space changes as they cause noise in the diffs.  I think IntelliJ automatically strips out trailing whitespace as soon as you edit a file, not sure if that's something you have to enable in other editors.

GHaberis: We use `eslint` which can be used to auto-fix code when it's run with the `--fix` flag. We have a validation script in our `package.json` that can be run via the command `yarn validate` on the command line, this calls `eslint` with the `--fix` flag, this should hopefully tidy up whitespace.

i-hardy: Looks like Eslint only checks/validates the .js and .jsx files and not the .scss files.

GHaberis: completely missed this was a `.scss` file 😆

PR #2622, opened 2020-07-22
Adding new copy for digital voucher thank you page

i-hardy:
## Why are you doing this?

When the user is buying a digital voucher we have different copy for the 'What's next' bullet points and will no longer show the date the vouchers are valid from.

[**Trello Card**](https://trello.com/c/Bp69nPoG)

## Changes

* Added the new copy behind the existing useDigitalVoucher switch
* Removed the date if the fulfilment type is collection and useDigitalVoucher is true

## Screenshots

**Paper voucher thank you page**
Screenshot 2020-07-22 at 17 03 35

**Digital voucher thank you page**
Screenshot 2020-07-22 at 17 04 36

### DIFF
--- a/support-frontend/assets/helpers/productPrice/fulfilmentOptions.js
+++ b/support-frontend/assets/helpers/productPrice/fulfilmentOptions.js
@@ -6,7 +6,6 @@ import { countryGroups } from 'helpers/internationalisation/countryGroup';
 
 const HomeDelivery: 'HomeDelivery' = 'HomeDelivery';
 const Collection: 'Collection' = 'Collection';
-const DigitalVoucher: 'DigitalVoucher' = 'DigitalVoucher';
 const Domestic: 'Domestic' = 'Domestic';
 const RestOfWorld: 'RestOfWorld' = 'RestOfWorld';
 const NoFulfilmentOptions: 'NoFulfilmentOptions' = 'NoFulfilmentOptions';
@@ -23,9 +22,8 @@ export type FulfilmentOptions =
   | typeof Domestic
   | typeof RestOfWorld
   | typeof NoFulfilmentOptions
-  | typeof DigitalVoucher
 
-export { HomeDelivery, Collection, DigitalVoucher, Domestic, RestOfWorld, NoFulfilmentOptions };
+export { HomeDelivery, Collection, Domestic, RestOfWorld, NoFulfilmentOptions };
 
 const getWeeklyFulfilmentOption = (country: IsoCountry) =>
   (countryGroups.International.countries.includes(country)
--- a/support-frontend/assets/helpers/productPrice/fulfilmentOptions.js
+++ b/support-frontend/assets/helpers/productPrice/fulfilmentOptions.js
@@ -6,6 +6,7 @@ import { countryGroups } from 'helpers/internationalisation/countryGroup';
 
 const HomeDelivery: 'HomeDelivery' = 'HomeDelivery';
 const Collection: 'Collection' = 'Collection';
+const DigitalVoucher: 'DigitalVoucher' = 'DigitalVoucher';
 const Domestic: 'Domestic' = 'Domestic';
 const RestOfWorld: 'RestOfWorld' = 'RestOfWorld';
 const NoFulfilmentOptions: 'NoFulfilmentOptions' = 'NoFulfilmentOptions';
@@ -22,8 +23,9 @@ export type FulfilmentOptions =
   | typeof Domestic
   | typeof RestOfWorld
   | typeof NoFulfilmentOptions
+  | typeof DigitalVoucher
 
-export { HomeDelivery, Collection, Domestic, RestOfWorld, NoFulfilmentOptions };
+export { HomeDelivery, Collection, DigitalVoucher, Domestic, RestOfWorld, NoFulfilmentOptions };
 
 const getWeeklyFulfilmentOption = (country: IsoCountry) =>
   (countryGroups.International.countries.includes(country)
--- a/support-frontend/assets/pages/paper-subscription-checkout/components/thankYou.jsx
+++ b/support-frontend/assets/pages/paper-subscription-checkout/components/thankYou.jsx
@@ -8,9 +8,10 @@ import { connect } from 'react-redux';
 
 import { HeroPicture } from 'pages/paper-subscription-landing/components/hero/heroPicture';
 import type { FulfilmentOptions } from 'helpers/productPrice/fulfilmentOptions';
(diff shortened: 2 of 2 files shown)


support-frontend/assets/pages/paper-subscription-checkout/components/thankYou.jsx
```diff
@@ -73,12 +83,42 @@ const whatNext: {[FulfilmentOptions]: Element<*>} = {
...
+            Keep an eye on your inbox. You should receive an email confirming the details of your subscription,
+            and another email shortly afterwards that contains details of how you can pick up your papers from today!
+          </span>,
+          <span>
+            You will receive your Subscription Card in your subscriber pack in the post, along with your home
+            delivery letter.
+          </span>,
+          <span>
+            Visit your chosen participating newsagent to pick up your paper using your Subscription Card, or
+            arrange a delivery using home your delivery letter.
```
kelvin-chappell, line 100:
```suggestion
            arrange a home delivery using your delivery letter.
```

support-frontend/assets/helpers/productPrice/fulfilmentOptions.js
```diff
@@ -6,6 +6,7 @@ import { countryGroups } from 'helpers/internationalisation/countryGroup';
 
 const HomeDelivery: 'HomeDelivery' = 'HomeDelivery';
 const Collection: 'Collection' = 'Collection';
+const DigitalVoucher: 'DigitalVoucher' = 'DigitalVoucher';
```
lucymonie, line 9: After thinking about this a bit more, I think it's better if we stick with HomeDelivery and Collection as the two fulfilment options for papers. I think it might be better if the subscription card text and the paper vouchers text are subsets of Collection.

i-hardy: I think you're right; I've removed the DigitalVoucher type and reworked the way we're getting the 'What next' text so the specific digital voucher copy sits as an option under Collection.

support-frontend/assets/pages/paper-subscription-checkout/components/thankYou.jsx
```diff
@@ -32,53 +33,68 @@ import { type FormFields, getFormFields } from 'helpers/subscriptionsForms/formF
...
+        <OrderedList items={textItems.map(item => <span>{item}</span>)} />
       </p>
     </Text>
-  ),
-};
+  );
+}
+
+function WhatNext(fulfilmentOption, useDigitalVoucher = false) {
+  if (fulfilmentOption === Collection && useDigitalVoucher) {
```
GHaberis, line 87: This condition `fulfilmentOption === Collection && useDigitalVoucher` is also used when assigning a value to `hideStartDate` in the `ThankYouContent` component. Could you pass `hideStartDate` to `WhatNext` so you didn't need to make this conditional check twice? If you do this then `hideStartDate` may not be the most clear name about what this `fulfilmentOption === Collection && useDigitalVoucher` condition signifies then?

support-frontend/assets/pages/paper-subscription-checkout/components/thankYou.jsx
```diff
@@ -32,53 +33,68 @@ import { type FormFields, getFormFields } from 'helpers/subscriptionsForms/formF
...
+  [Collection]: {
+    default: [
+      `Look out for an email from us confirming your subscription.
+        It has everything you need to know about how to manage it in the future.`,
+      'You will receive your personalised book of vouchers.',
+      'Exchange your voucher for a newspaper at your newsagent or wherever you buy your paper',
+    ],
+    digitalVoucher: [
+      `Keep an eye on your inbox. You should receive an email confirming the details of your subscription,
+        and another email shortly afterwards that contains details of how you can pick up your papers from today!`,
```
johnduffell, line 67: is it always from today, or can they pick a start date still?

support-frontend/assets/pages/paper-subscription-checkout/components/thankYou.jsx
```diff
@@ -32,53 +33,68 @@ import { type FormFields, getFormFields } from 'helpers/subscriptionsForms/formF
...
     isPending: boolean,
+    useDigitalVoucher: boolean,
 };
 
+// ----- Map State/Props ----- //
+
+function mapStateToProps(state: WithDeliveryCheckoutState) {
+  return {
+    ...getFormFields(state),
+    useDigitalVoucher: state.common.settings.useDigitalVoucher,
```
johnduffell, line 44: presumably this is existing, but I didn't think useDigitalVoucher was state, more a universal constant only changed with a deploy

support-frontend/assets/pages/paper-subscription-checkout/components/thankYou.jsx
```diff
@@ -32,53 +33,68 @@ import { type FormFields, getFormFields } from 'helpers/subscriptionsForms/formF
...
-  ),
-};
+  );
+}
+
+function WhatNext(fulfilmentOption, useDigitalVoucher = false) {
+  if (fulfilmentOption === Collection && useDigitalVoucher) {
+    return whatNextElement(whatNextText[Collection].digitalVoucher);
+  }
+  return whatNextElement(whatNextText[fulfilmentOption].default);
```
johnduffell, line 90: minor but the if and else part of this both repeat the call to `whatNextElement(` so it would be a good habit to put only the bits that change in the if statement.

support-frontend/assets/pages/paper-subscription-checkout/components/thankYou.jsx
```diff
@@ -104,12 +120,12 @@ function ThankYouContent({
             </Text>
           )
         }
-        {startDate &&
+        {(startDate && !hideStartDate) &&
           <Text title={fulfilmentOption === HomeDelivery ? 'You will receive your newspapers from' : 'You can start using your vouchers from'}>
```
rupertbates, line 124: Is the 'You can start using your vouchers from' copy ok for Subscription Card?

support-frontend/assets/pages/paper-subscription-checkout/components/thankYou.jsx
i-hardy: It is a constant, but it's incorporated into the Redux state despite not changing, I think just to make accessing it more convenient.

support-frontend/assets/pages/paper-subscription-checkout/components/thankYou.jsx
i-hardy: That entire section with the start date is hidden when using the card- `hideStartDate` will be true if `useDigitalVoucher` is true and the fulfillment option is Collection.

PR #2623, opened 2020-07-22
Automate the IT tests

johnduffell:
Post merge TODO:
- remove support-workers TC sbt build step (after merge)
- make support:it-test-runner auto deploy in riffraff (after merge)

Long term documentation on the wiki: https://github.com/guardian/support-frontend/wiki/Automated-IT-Tests

## Why are you doing this?

The IT tests are very important to know that the system is working correctly.
Like all tests, it is convenient to run them locally to develop them.

However, automation of everything helps with team consistency and makes it more likely that the person who broke them will notice first and fix them before it slows down othe team members.

With the IT tests there have been a few occasions where one developer finds the tests are broken, but doesn't think it was their changes.  Then there is a process of working out whether they are broken in github or it's an environment issue etc.

With them automated, there is a guarantee that the tests are working, and if there are any issues, it must be something locally.  This reduces the cost and hassle factor of the tests and makes them more maintained.

The tricky part of automating them in the past has been the need for credentials.  The tests need to acces the test zuora/stripe etc.

## How to get around the credential issue

The process I have taken is the simplest one using existing tooling e.g. riffraff etc. as per https://docs.google.com/document/d/1mjYNVdrc8v4KWOiD7B6dtKfQ2F5XvR7fq12oDPBSOZg/edit#heading=h.8escnh1mq4fg

The support-workers build spits out two new artifacts, the support-workers-it.jar with all of the tests and code in (it's 80+MB, even bigger than the non test support workers which is "only" 60MB) Then riffraff will upload this to S3 when support-workers is deployed.
It also builds a new it-test-runner project which is includes cloudformation etc for a lambda which is invoked every hour.  This lambda grabs the above IT jar and runs all the tests ending in *Spec.scala with the required permissions available for DEV.
As the lambda is running it spits the test results into cloudwatch and the alarm will go off for the "prod" version of the tests, if there are failures.

The logs will be in PROD but currently are available in CODE and don't look perfect, but usableish [/aws/lambda/support-it-tests-CODE](https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#logsV2:log-groups/log-group/$252Faws$252Flambda$252Fsupport-it-tests-CODE)
The metrics are [in cloudwatch](https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#metricsV2:graph=~(region~'eu-west-1~view~'timeSeries~stacked~false~start~'-P1D~end~'P0D~period~60~annotations~(horizontal~(~(label~'it-test-failed*20*3e*200*20for*201*20datapoints*20within*201*20minute~value~0)))~title~'non*20urgent*20-*20IT*20Tests*20are*20failing~metrics~(~(~'support-frontend~'it-test-succeeded~'Stage~'CODE)~(~'.~'it-test-failed~'.~'.))~yAxis~(left~(min~0)));query=~'*7bsupport-frontend*2cStage*7d) 

Possible future work:
- run all IT tests, not just the ones depended on by support-workers that end in *Spec
- abort if the tests in the support-workers-it.jar are not recent enough

[**Trello Card**](https://trello.com/c/GS3PiNN8/3163-automate-the-it-tests)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -177,4 +177,4 @@ lazy val `it-test-runner` = (project in file("support-lambdas/it-test-runner"))
 //    commonSettings,
 //    integrationTestSettings,
 //    libraryDependencies ++= commonDependencies
-  )//.dependsOn(`support-workers`)
+  ).dependsOn(`support-services`)// todo very excessive
--- a/build.sbt
+++ b/build.sbt
@@ -10,7 +10,8 @@ lazy val integrationTestSettings: Seq[Def.Setting[_]] = Defaults.itSettings ++ S
   javaSource in IntegrationTest := baseDirectory.value / "src" / "test" / "java",
   resourceDirectory in IntegrationTest := baseDirectory.value / "src" / "test" / "resources",
   testOptions in Test += Tests.Argument(TestFrameworks.ScalaTest, "-l", "com.gu.test.tags.annotations.IntegrationTest"),
-  testOptions in Test += Tests.Argument(TestFrameworks.ScalaTest, "-eU")
+  testOptions in Test += Tests.Argument(TestFrameworks.ScalaTest, "-eU"),
+  testOptions in IntegrationTest += Tests.Argument(TestFrameworks.ScalaTest, "-eU")
 )
 
 lazy val release = Seq[ReleaseStep](
--- a/build.sbt
+++ b/build.sbt
@@ -5,13 +5,15 @@ import scala.sys.process._
 
 skip in publish := true
 
-lazy val integrationTestSettings: Seq[Def.Setting[_]] = Defaults.itSettings ++ Seq(
+lazy val integrationTestSettings: Seq[Def.Setting[_]] = Project.inConfig(IntegrationTest)(baseAssemblySettings) ++ Defaults.itSettings ++ Seq(
   scalaSource in IntegrationTest := baseDirectory.value / "src" / "test" / "scala",
   javaSource in IntegrationTest := baseDirectory.value / "src" / "test" / "java",
   resourceDirectory in IntegrationTest := baseDirectory.value / "src" / "test" / "resources",
   testOptions in Test += Tests.Argument(TestFrameworks.ScalaTest, "-l", "com.gu.test.tags.annotations.IntegrationTest"),
   testOptions in Test += Tests.Argument(TestFrameworks.ScalaTest, "-eU"),
-  testOptions in IntegrationTest += Tests.Argument(TestFrameworks.ScalaTest, "-eU")
+  testOptions in IntegrationTest += Tests.Argument(TestFrameworks.ScalaTest, "-eU"),
+  test in (IntegrationTest, assembly) := {},
+  aggregate in (IntegrationTest, assembly) := false
 )
 
 lazy val release = Seq[ReleaseStep](
@@ -74,7 +76,8 @@ lazy val root = (project in file("."))
     `support-internationalisation`,
     `support-services`,
     `stripe-intent`,
-    `support-redemptiondb`
+    `support-redemptiondb`,
+    `it-test-runner`
   )
 
 lazy val testScalastyle = taskKey[Unit]("testScalastyle")
@@ -108,7 +111,7 @@ lazy val `support-workers` = (project in file("support-workers"))
     integrationTestSettings,
     libraryDependencies ++= commonDependencies
   ).dependsOn(`support-services`, `support-models` % "test->test;it->test;compile->compile", `support-config`, `support-internationalisation`)
-  .aggregate(`support-services`, `support-models`, `support-config`, `support-internationalisation`, `stripe-intent`, `support-redemptiondb`)
+  .aggregate(`support-services`, `support-models`, `support-config`, `support-internationalisation`, `stripe-intent`)
 
 
 lazy val `support-models` = (project in file("support-models"))
@@ -166,3 +169,12 @@ lazy val `stripe-intent` = (project in file("support-lambdas/stripe-intent"))
 
 lazy val `support-redemptiondb` = (project in file("support-redemptiondb"))
   .enablePlugins(RiffRaffArtifact)
+
+lazy val `it-test-runner` = (project in file("support-lambdas/it-test-runner"))
+  .enablePlugins(JavaAppPackaging, RiffRaffArtifact)
+//  .configs(IntegrationTest)
+  .settings(
+//    commonSettings,
+//    integrationTestSettings,
+//    libraryDependencies ++= commonDependencies
+  )//.dependsOn(`support-workers`)
--- a/build.sbt
+++ b/build.sbt
@@ -5,15 +5,12 @@ import scala.sys.process._
 
 skip in publish := true
 
-lazy val integrationTestSettings: Seq[Def.Setting[_]] = Project.inConfig(IntegrationTest)(baseAssemblySettings) ++ Defaults.itSettings ++ Seq(
+lazy val integrationTestSettings: Seq[Def.Setting[_]] = Defaults.itSettings ++ Seq(
   scalaSource in IntegrationTest := baseDirectory.value / "src" / "test" / "scala",
   javaSource in IntegrationTest := baseDirectory.value / "src" / "test" / "java",
   resourceDirectory in IntegrationTest := baseDirectory.value / "src" / "test" / "resources",
   testOptions in Test += Tests.Argument(TestFrameworks.ScalaTest, "-l", "com.gu.test.tags.annotations.IntegrationTest"),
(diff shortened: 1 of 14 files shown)


build.sbt
```diff
@@ -5,13 +5,15 @@ import scala.sys.process._
 
 skip in publish := true
 
-lazy val integrationTestSettings: Seq[Def.Setting[_]] = Defaults.itSettings ++ Seq(
+lazy val integrationTestSettings: Seq[Def.Setting[_]] = Project.inConfig(IntegrationTest)(baseAssemblySettings) ++ Defaults.itSettings ++ Seq(
```
johnduffell, line 8: we need to enable the sbt assembly plugin in the "it" context in order to product the jar (this only actually applies to the support-workers project as we are not aggregating separately)

build.sbt
```diff
@@ -5,13 +5,15 @@ import scala.sys.process._
...
   scalaSource in IntegrationTest := baseDirectory.value / "src" / "test" / "scala",
   javaSource in IntegrationTest := baseDirectory.value / "src" / "test" / "java",
   resourceDirectory in IntegrationTest := baseDirectory.value / "src" / "test" / "resources",
   testOptions in Test += Tests.Argument(TestFrameworks.ScalaTest, "-l", "com.gu.test.tags.annotations.IntegrationTest"),
   testOptions in Test += Tests.Argument(TestFrameworks.ScalaTest, "-eU"),
-  testOptions in IntegrationTest += Tests.Argument(TestFrameworks.ScalaTest, "-eU")
+  testOptions in IntegrationTest += Tests.Argument(TestFrameworks.ScalaTest, "-eU"),
+  test in (IntegrationTest, assembly) := {},
+  aggregate in (IntegrationTest, assembly) := false
```
johnduffell, line 16: - we don't want to run the tests when wer'e building them to run in AWS.
- we don't want to assembly a jar for every single project (yet) just for the main one.

build.sbt
```diff
@@ -108,7 +111,7 @@ lazy val `support-workers` = (project in file("support-workers"))
     integrationTestSettings,
     libraryDependencies ++= commonDependencies
   ).dependsOn(`support-services`, `support-models` % "test->test;it->test;compile->compile", `support-config`, `support-internationalisation`)
-  .aggregate(`support-services`, `support-models`, `support-config`, `support-internationalisation`, `stripe-intent`, `support-redemptiondb`)
+  .aggregate(`support-services`, `support-models`, `support-config`, `support-internationalisation`, `stripe-intent`)
```
johnduffell, line 114: the DB was never related to support workers, so removed it from the build definition.  This now happens in the build-tc script of support-workers-mono build.
So effectively it happens in the same place, but only in Teamcity.

build.sbt
```diff
@@ -166,3 +169,12 @@ lazy val `stripe-intent` = (project in file("support-lambdas/stripe-intent"))
...
   .enablePlugins(RiffRaffArtifact)
+
+lazy val `it-test-runner` = (project in file("support-lambdas/it-test-runner"))
+  .enablePlugins(JavaAppPackaging, RiffRaffArtifact)
+//  .configs(IntegrationTest)
+  .settings(
+//    commonSettings,
+//    integrationTestSettings,
+//    libraryDependencies ++= commonDependencies
+  ).dependsOn(`support-services`)// todo very excessive
```
johnduffell, line 180: we only need the cloudwatch metric put code so this dependency really bloats the lambda jar.  this will be probably be a copy and paste job for simplicity.

support-lambdas/it-test-runner/cfn.yaml
```diff
@@ -0,0 +1,105 @@
...
+                - lambda:InvokeFunction
+              Resource: !Sub "arn:aws:logs:${AWS::Region}:${AWS::AccountId}:log-group:/aws/lambda/it-tests-${Stage}:log-stream:*"
+        - Statement:
+            - Effect: Allow
+              Action: s3:GetObject
+              Resource: !Sub arn:aws:s3:::support-workers-dist/support/${Stage}/it-tests/support-workers-it.jar
+        - Statement:
+            - Effect: Allow
+              Action: cloudwatch:PutMetricData
+              Resource: "*"
```
johnduffell, line 63: https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/iam-access-control-overview-cw.html
> CloudWatch does not have any resources for you to control using policies resources, so use the wildcard character (*) in IAM policies

support-lambdas/it-test-runner/src/main/scala/com/gu/aws/AwsCloudWatchMetricPut.scala
```diff
@@ -0,0 +1,51 @@
+package com.gu.aws
```
johnduffell, line 1: this is copied from support-services, as the jar was 56mb with the whole of support-services etc pulled in.
(It's still 25MB even without but we don't want to be too silly at this stage.)
The other option would be being DRY and pulling the aws stuff into a separate library rather than the current tiered setup.

support-workers/build.sbt
```diff
@@ -46,10 +46,11 @@ riffRaffPackageType := assembly.value
 riffRaffManifestProjectName := s"support:support-workers-mono"
 riffRaffManifestBranch := Option(System.getenv("BRANCH_NAME")).getOrElse("unknown_branch")
 riffRaffBuildIdentifier := Option(System.getenv("BUILD_NUMBER")).getOrElse("DEV")
-riffRaffManifestVcsUrl := "git@github.com/guardian/support-workers.git"
+riffRaffManifestVcsUrl := "git@github.com/guardian/support-frontend.git"
 riffRaffUploadArtifactBucket := Option("riffraff-artifact")
 riffRaffUploadManifestBucket := Option("riffraff-builds")
 riffRaffArtifactResources += (file("support-workers/cloud-formation/target/cfn.yaml"), "cfn/cfn.yaml")
+riffRaffArtifactResources += (file("support-workers/target/scala-2.12/support-workers-it.jar"), "it-tests/support-workers-it.jar")
```
johnduffell, line 53: the artifact is built with "it:assembly" and rather than riffraff artifact plugin find it automatically, we hard code the jar and pick it up as a side artifact during the main upload and deploy.
Another option would be to have a `IntegrationTest / riffraff.yaml` and all the rest, and upload and deploy the IT separately from the rest of the workers, effectively as a separate project.
However it didn't look like riffraff artifact plugin supports alternative names for the yaml file so I would have to add that parameter first.

support-lambdas/it-test-runner/src/main/scala/com/gu/RunITTests.scala
```diff
@@ -0,0 +1,101 @@
...
+      )
+      _ <- Try(Runner.main(a)) // unfortunately calls System.exit(1) if any tests fail...
+    } yield ()
+    result match {
+      case Failure(exception) => log(s"Failed with exception ${exception.toString}")
+        exception.printStackTrace(System.out)
+        System.exit(9)
+        "not possible"
+      case _ => log("RAN OK!")
+        "Finshed and RAN OK!"
```
rupertbates, line 60: typo 'Finished'

support-lambdas/it-test-runner/src/main/scala/com/gu/aws/AwsCloudWatchMetricPut.scala
rupertbates: Thanks for this comment, I was wondering about it. I think it might be worth extracting this stuff to a separate project as it's not great having copy and paste code around and I think this is not the only place where it has happened, but I'd be happy to leave that out of this PR & put in a health card for it.

support-lambdas/it-test-runner/src/main/scala/com/gu/RunITTests.scala
johnduffell:
```suggestion
        "Finished and RAN OK!" // this won't happen because Runner calls System.exit
```

support-lambdas/it-test-runner/src/main/scala/com/gu/aws/AwsCloudWatchMetricPut.scala
johnduffell: yes its definitely not the only place.  I think a separate PR.
more generally, if we go down the cohesive modules route rather than layered approach, we could remove a lot of unneeded stuff from the lambda jars and make things more reusable.  if we do decide jar size/startup is a key quality metric we should do more analysis first.

johnduffell: in the end, because i copied it into the same package, it's causing some runtime exceptions, as they are all packaged and loaded by the same class loader even though compiled separately.  So I will pull it out in my next PR.

PR #2626, opened 2020-07-24
extract single s3 loader and cloudwatch client

johnduffell:
## Why are you doing this?

In the previous PR I copied the cloudwatch and S3 clients.
https://github.com/guardian/support-frontend/pull/2623/files#r459784462

The suggestion was to pull them into a separate module so we can reuse elsewhere.  I also found once I deployed the test runner it was getting class loader errors sometimes because the super-jar with all dependencies in had the same code with slight differences.  This means the IT tests are not running correctly yet.

I noticed we were calling the s3 client in so many different ways across the code base, I simplified the interface as much as possible and made them all comply with the new one.

The cloudwatch interface was consistent already so I didn't touch that.

I also made a folder for the cohesive modules, to distinguish them from the main "layered" approach, and I moved the "rest" module already created in there too.  This is to stop support-* blindness with the amount of folders we have already.  Other organisations are possible though if we prefer.  I am possibly gravitating to the "handlers (support-lambdas)" and "libs (support-modules)" layout that I did in the support-service-lambdas repo unintentionally.

I also did some tidy up on the build definition to make what I'm adding less of a mess.

In the end, hopefully this will also fix the duplicate class definition issue in the IT tests that I shipped in the above PR!

Tested in CODE:
support-frontend settings (contr amounts/types working)
getprices (product information page has prices for DS, GW)
support-workers config (can check out GW)
stripe-intent config (can check out GW)
it-test-runner downloading tests jar from S3


build.sbt
```diff
@@ -29,10 +33,23 @@ lazy val release = Seq[ReleaseStep](
   pushChanges
 )
 
-lazy val commonSettings = Seq(
+inThisBuild(Seq(
```
johnduffell, line 36: ThisBuild is the default "Zero" project that it falls back to, so we don't have to add them separately to each project.

build.sbt
```diff
@@ -1,17 +1,19 @@
...
 
 lazy val integrationTestSettings: Seq[Def.Setting[_]] = Defaults.itSettings ++ Seq(
   scalaSource in IntegrationTest := baseDirectory.value / "src" / "test" / "scala",
   javaSource in IntegrationTest := baseDirectory.value / "src" / "test" / "java",
   resourceDirectory in IntegrationTest := baseDirectory.value / "src" / "test" / "resources",
   testOptions in Test += Tests.Argument(TestFrameworks.ScalaTest, "-l", "com.gu.test.tags.annotations.IntegrationTest"),
   testOptions in Test += Tests.Argument(TestFrameworks.ScalaTest, "-eU"),
-  testOptions in IntegrationTest += Tests.Argument(TestFrameworks.ScalaTest, "-eU")
+  testOptions in IntegrationTest += Tests.Argument(TestFrameworks.ScalaTest, "-eU"),
+  libraryDependencies += scalatest % "it",
```
johnduffell, line 16: we can't put scalatest on the "it" classpath unless we add the IntegrationTest config axis to the project, so it should add here instead of commonDeps.

build.sbt
```diff
@@ -29,10 +31,17 @@ lazy val release = Seq[ReleaseStep](
...
   scalaVersion := "2.12.10",
-  resolvers ++= Seq(Resolver.sonatypeRepo("releases"), Resolver.bintrayRepo("guardian", "ophan")),
+  dependencyStats / aggregate := false,
+  dependencyTree / aggregate := false,
+  // https://www.scala-sbt.org/1.x/docs/Cached-Resolution.html
+  updateOptions := updateOptions.value.withCachedResolution(true),
+  resolvers ++= Seq(Resolver.sonatypeRepo("releases")), // libraries that haven't yet synced to maven central
+))
+
+lazy val releaseSettings = Seq(
```
johnduffell, line 44: separated release settings out, because they only apply to a minority of projects.

build.sbt
```diff
@@ -49,22 +58,20 @@ lazy val commonSettings = Seq(
...
 
 lazy val commonDependencies = Seq(
   "com.typesafe" % "config" % "1.3.2",
-  "org.scalatest" %% "scalatest" % "3.2.0" % "it, test",// if the following PR is merged in v3.2.1, remove "-eU" parameter from options above - PR https://github.com/scalatest/scalatest/pull/1842
+  scalatest % "test",
   "com.typesafe.scala-logging" %% "scala-logging" % "3.9.2"
 )
 
 lazy val root = (project in file("."))
+  .disablePlugins(ReleasePlugin, SbtPgp, Sonatype, AssemblyPlugin)
```
johnduffell, line 70: disabling unnecessary plugins improves the sbt startup time (although there are still loads of settings to load)

build.sbt
```diff
@@ -49,22 +58,20 @@ lazy val commonSettings = Seq(
...
   .settings(
     name := "support-frontend-root",
-    moduleName := "support-frontend-root"
+    moduleName := "support-frontend-root",
+    publish / skip := true,
```
johnduffell, line 74: moved from above, it was "loose" in the sbt file. 
Ideally for clarity, this root sbt file would just define relationships between the modules, and the modules themselves would defined any details about themselves.  But that isn't desirable to be done in this PR.

build.sbt
```diff
@@ -75,19 +82,20 @@ lazy val root = (project in file("."))
     `support-services`,
     `stripe-intent`,
     `support-redemptiondb`,
-    `it-test-runner`
+    `it-test-runner`,
+    `module-aws`,
+    `module-rest`,
```
johnduffell, line 87: `rest` module was missing from the aggregate.  I tried removing the root project explicit definition, as SBT will auto aggregate all the projects for you, but it had a name clash with `support-frontend` subproject so it got an undesirable name.

build.sbt
```diff
@@ -133,40 +140,44 @@ lazy val `support-config` = (project in file("support-config"))
...
   .configs(IntegrationTest)
   .settings(
-    commonSettings,
+    releaseSettings,
     integrationTestSettings,
     libraryDependencies ++= commonDependencies
-  ).dependsOn(`support-internationalisation`, `support-models`, `support-config`, `support-rest`)
-  .aggregate(`support-internationalisation`, `support-models`, `support-config`, `support-rest`)
+  ).dependsOn(`support-internationalisation`, `support-models`, `support-config`, `module-rest`, `module-aws`)
+  .aggregate(`support-internationalisation`, `support-models`, `support-config`, `module-rest`, `module-aws`)
```
johnduffell, line 147: I went for "module" rather than "support" as the prefix, to separate them from the standard layered approach we used for many of the modules.  Happy for other conventions, as long as we avoid a wall of `support-*` with no obvious high level structure.

project/plugins.sbt
```diff
@@ -20,9 +20,9 @@ addSbtPlugin("org.scalastyle" %% "scalastyle-sbt-plugin" % "1.0.0" excludeAll Ex
 
 addSbtPlugin("com.gu" % "sbt-riffraff-artifact" % "1.1.3")
 
-addSbtPlugin("com.timushev.sbt" % "sbt-updates" % "0.3.3")
+//addSbtPlugin("com.timushev.sbt" % "sbt-updates" % "0.3.3")
 
-addSbtPlugin("net.vonbuchholtz" % "sbt-dependency-check" % "0.2.0")
+//addSbtPlugin("net.vonbuchholtz" % "sbt-dependency-check" % "0.2.0")
```
johnduffell, line 25: these didn't seem to be used so were just slowing things down.  They may have just been added as a convenience when someone was doing updates/dependencies?

support-frontend/app/admin/settings/Settings.scala
```diff
@@ -35,27 +36,22 @@ object AllSettings {
...
     } yield settings
 
   def fromLocalFile[T: Decoder](source: SettingsSource.LocalFile): Either[Throwable, T] =
     for {
       buf <- Either.catchNonFatal {
         Source.fromFile(source.path)
       }
-      settings <- fromBufferedSource(buf)
+      settings <- decode[T](buf.mkString)
+      _ <- Try(buf.close()).toEither
```
johnduffell, line 54: I have changed all uses of S3 to use the new abstraction I created, which doesn't require you to close the stream yourself.  It is not actually `Future` compatible, but generally we seem to rely on blocking threads rather than async anyway. (which is a risk to server stability)

support-frontend/app/controllers/StripeController.scala
```diff
@@ -52,7 +52,7 @@ class StripeController(
 
     val v2RecaptchaToken = request.body.token
 
-    val cloudwatchEvent = createSetupIntentRequest(stage, "v2Recaptcha")
+    val cloudwatchEvent = AwsCloudWatchMetricSetup.createSetupIntentRequest(stage, "v2Recaptcha")
     AwsCloudWatchMetricPut(client)(cloudwatchEvent)
```
johnduffell, line 56: the domain specific convenience methods are all moved to `AwsCloudWatchMetricSetup` so that the actual `Put`ter can be generic

support-frontend/app/wiring/Controllers.scala
```diff
@@ -201,7 +200,7 @@ trait Controllers {
     actionRefiners,
     appConfig.guardianDomain,
     appConfig.identity.webappUrl,
-    () => AwsCloudWatchMetricPut(client)(setupWarningRequest(appConfig.stage))
+    () => AwsCloudWatchMetricPut(AwsCloudWatchMetricPut.client)(setupWarningRequest(appConfig.stage))
```
johnduffell, line 203: there's something weird about how you have to call AwsCloudWatchMetricPut.  I didn't bother to work out what it should look like in this PR as it is consistent, so I didn't have to pick a "winning" API like I did with S3.

support-frontend/app/wiring/Services.scala
```diff
@@ -12,14 +13,14 @@ import controllers.DynamoTableAsyncForUser
 import play.api.BuiltInComponentsFromContext
 import play.api.libs.ws.ahc.AhcWSComponents
 import services._
-import services.aws.AwsS3Client.s3
```
johnduffell, line 15: I'm not sure why we have an implicit s3 client, but we do, so I just went with it for now!

support-models/build.sbt
```diff
@@ -17,3 +17,5 @@ libraryDependencies ++= Seq(
   "com.fasterxml.jackson.core" % "jackson-annotations" % jacksonVersion,
   "com.fasterxml.jackson.datatype" % "jackson-datatype-jsr310" % jacksonVersion
 )
+
+resolvers ++= Seq(Resolver.bintrayRepo("guardian", "ophan")) //event producer library
```
johnduffell, line 21: this was previously added to all projects, but actually only support-models seems to need it (not even support-workers)

support-redemptiondb/build.sbt
```diff
@@ -3,7 +3,7 @@ name := "support-redemptiondb"
 description:= "packages the database cfn for the redemption code db"
 
 assemblyJarName := "dummy.jar"
-riffRaffPackageType := assembly.value
+riffRaffPackageType := new File("dummy.jar")
```
johnduffell, line 6: I changed this so we didn't need the sbt-assembly pluigin enabled on this project.

support-services/src/main/scala/com/gu/support/catalog/AwsS3ClientJson.scala
```diff
@@ -0,0 +1,26 @@
...
+
+import com.amazonaws.services.s3.AmazonS3URI
+import com.gu.aws.AwsS3Client
+import com.typesafe.scalalogging.LazyLogging
+import io.circe._
+import io.circe.parser._
+
+import scala.util.{Failure, Success}
+
+object AwsS3ClientJson extends LazyLogging{
```
johnduffell, line 11: pulled out from the original S3 client, otherwise we would have needed circe as a dependency of the module-aws.

build.sbt
```diff
@@ -102,10 +110,9 @@ lazy val `support-frontend` = (project in file("support-frontend"))
   .aggregate(`support-services`, `support-models`, `support-config`, `support-internationalisation`)
 
 lazy val `support-workers` = (project in file("support-workers"))
-  .enablePlugins(JavaAppPackaging, RiffRaffArtifact)
+  .enablePlugins(RiffRaffArtifact).disablePlugins(ReleasePlugin, SbtPgp, Sonatype)
```
johnduffell, line 113: didn't actually need Java app packaging, as that is designed to add platform specific shell scripts, but all we need is a plain jar file.

project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.3.9
+sbt.version=1.3.13
```
johnduffell, line 1: updated this as the sbt-dependency-graph plugin has issues which are more fully fixed in this version, and snyk relies on that plugin

support-frontend/app/admin/settings/Settings.scala
tomrf1: It looks like in support-frontend this is only used for getting the settings, and this operation does have its own EC:
https://github.com/guardian/support-frontend/blob/master/support-frontend/app/admin/settings/SettingsProvider.scala#L134

An alternative would be to use `scala.concurrent.blocking`

build.sbt
rupertbates: module works for me

support-frontend/app/admin/settings/Settings.scala
rupertbates: Is the blocking here necessary or just the way it's been written?

johnduffell: not necessary, but InputStream is fundamentally a blocking abstraction and that's what AWS sdk exposes.
The only way around it if you want to stream data is to use some kind of reactive stream e.g. [fs2](https://github.com/functional-streams-for-scala/fs2) and an alternative AWS library e.g https://github.com/laserdisc-io/fs2-aws (from a quick google)
In our case, the files are small enough to fit in memory, so we would be happy with a Future[Array[Byte]] or similar, but under the covers this would still have to use some non blocking mechanism or a separate thread context.

Whether that's a priority depends whether it causes an incident!  But as we add more and more to support-frontend app it does add to the risk.  I think higher priority would be move promotions dynamo access over to the new non-blocking AWS SDK.

PR #2632, opened 2020-07-29
component event parameters for identity authentication tracking

QuarpT:
## Why are you doing this?

This will allow data analysts to track how many users are using different sign in mechanisms to authenticate using the Guardian website.

https://trello.com/c/37cmz78X/2206-improve-and-create-additional-ophan-events-sent-from-identity-pages-and-identity-api

## Changes

Pass component event parameters for all signin referrers. These parameters get propagated from /signin to IDAPI which sends the component events on behalf of the user. IDAPI sends the component event to Ophan using the /signin page view ID and the new SC_GU_U cookie. This means the /signin pageview entry has session information associated and the component ID used for signing in.


support-frontend/app/actions/CustomActionBuilders.scala
```diff
@@ -43,7 +43,8 @@ class CustomActionBuilders(
       ("returnUrl" -> s"$supportUrl$path") &
       idSkipConfirmation &
       idSkipValidationReturn &
-      "clientId" -> clientId).toString
+      "clientId" -> clientId &
+      createAuthenticationComponentEventTuple(SigninRedirectForSupporters)).toString
```
johnduffell, line 47: probably a bit overkill having a special mechanism just to create a fixed string, would this do the same thing?
```suggestion
      "componentEventParams" -> "componentType=identityauthentication&componentId=signin_redirect_for_supporters"
    ).toString
```

support-frontend/test/controllers/SubscriptionsTest.scala
```diff
@@ -142,7 +142,12 @@ class SubscriptionsTest extends AnyWordSpec with Matchers with TestCSRFComponent
 
       val result = fakeRequestAuthenticatedWith(actionRefiner = loggedOutActionRefiner)
       status(result) mustBe 303
-      header("Location", result).value must startWith("https://identity-url.local")
+      header("Location", result).value must be("https://identity-url.local/signin?" +
+        "returnUrl=/&" +
+        "skipConfirmation=true&" +
+        "skipValidationReturn=true&" +
+        "clientId=subscriptions&" +
+        "componentEventParams=componentType%3Didentityauthentication%26componentId%3Dsignin_redirect_for_supporters")
```
johnduffell, line 150: good addition of a test here especially with the `&` inside a parameter value 👍

support-frontend/app/actions/CustomActionBuilders.scala
QuarpT: I agree 😄 
I reused code from projects where we have multiple event IDs that need to be encoded in different places. But you're right, it's not necessary here.

PR #2633, opened 2020-07-29
Remove ds address test

rupertbates:
## Why are you doing this?
We are finishing the Digital Subscription address test. There is no significant difference in conversion rate between the control and the 'no address' version so we are reverting to the control.

[**Trello Card**](https://trello.com/c/LNJ6clbg/3160-turn-off-ab-test-for-no-address)

### DIFF
--- a/support-frontend/assets/helpers/abTests/abtestDefinitions.js
+++ b/support-frontend/assets/helpers/abTests/abtestDefinitions.js
@@ -35,53 +35,6 @@ export const tests: Tests = {
     seed: 5,
   },
 
-  fancyAddressTest: {
-    type: 'OTHER',
-    variants: [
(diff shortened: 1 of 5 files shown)


support-frontend/assets/helpers/abTests/abtestDefinitions.js
```diff
@@ -35,53 +33,6 @@ export const tests: Tests = {
     seed: 5,
   },
 
-  fancyAddressTest: {
```
johnduffell, line 38: is this just left over cleanup from the other test?

rupertbates: Yup

PR #2634, opened 2020-07-29
New, compliant footer - Storybook component

i-hardy:
## Why are you doing this?

We need a more consistent and legally compliant footer across multiple pages; this introduces the new footer as a Storybook component ready to be implemented on pages as required. Legal text shown in the top section is currently just arbitrary as that's where the component renders children, this will probably be refined as we actually implement the component.

I'd particularly appreciate feedback on using the `styled` utility in `footerCompliant/containers/contentStyles.js`. It seemed like the best approach to re-implement some of the various modifiers that we have on the existing Content component, but I'm not sure if there's a better approach for adapting styles based on props.

[**Trello Card**](https://trello.com/c/m1SSOQiq)

## Changes

* Bringing initial Footer component in Storybook in line with the designs

## Screenshots

**Mobile, with additional text**
Screenshot 2020-07-29 at 16 51 14

**Mobile, links only**
Screenshot 2020-07-29 at 16 55 27

**Tablet, with additional text**
Screenshot 2020-07-29 at 16 56 52

**Desktop, links only**
Screenshot 2020-07-29 at 16 57 09


support-frontend/assets/components/footerCompliant/BackToTop.jsx
```diff
@@ -0,0 +1,69 @@
...
+      text-decoration: none;
+      & .css-${iconContainer.name} {
+        background-color: ${brandAlt[400]};
+      }
+    }
+`;
+
+
+const textStyles = css`
+    ${textSans.small({ fontWeight: 'bold' })};
```
rupertbates, line 56: I've noticed issues with passing fontWeight into options in the past so generally I separate it out. If you've tested across a few browsers and it works though feel free to ignore.

SiAdcock: Oh, are you able to characterise those issues @rupertbates?

support-frontend/assets/components/footerCompliant/footer.jsx
```diff
@@ -22,22 +23,21 @@ type PropTypes = {|
...
 
 
 // ----- Component ----- //
 
 function Footer({
-  disclaimer, privacyPolicy, children, countryGroupId, appearance,
+  disclaimer, privacyPolicy, children, countryGroupId,
 }: PropTypes) {
 
   // It would probably be helpful to replace the Content and Rows components with emotion
```
rupertbates, line 35: Does this comment still apply or should it be changed/removed?

support-frontend/assets/components/footerCompliant/footer.jsx
```diff
@@ -50,11 +50,30 @@ function Footer({
...
+            <a href="https://www.theguardian.com/help/privacy-policy">Privacy Policy</a>
+          </li>
+          <li css={link}>
+            <a href="https://www.theguardian.com/help/contact-us">Contact us</a>
+          </li>
+          <li css={link}>
+            <a href="https://www.theguardian.com/subscriber-direct/subscription-frequently-asked-questions">FAQs</a>
+          </li>
+          <li css={link}>
+            <a href="https://www.theguardian.com/info/2014/aug/06/guardian-observer-digital-subscriptions-terms-conditions">Terms & Conditions</a>
```
rupertbates, line 67: This url is for the Digital subscription only, do we need to parameterise this for use with other products?

support-frontend/assets/components/footerCompliant/BackToTop.jsx
rupertbates: Yes the bold wasn't getting applied

i-hardy: I've checked in Firefox, Chrome and Safari locally and had a look in IE too via Browserstack and it all looked fine. Might be an issue with specific fonts? If the weight you specify isn't in the font file(s) downloaded then browsers will usually just default to the nearest one they have.

support-frontend/assets/components/footerCompliant/containers/contentStyles.js
```diff
@@ -0,0 +1,44 @@
...
+  after: boolean
+|}
+
+export const WithMargins = styled.div`
+  display: flex;
+  ${from.tablet} {
+    ${({ before }: WithMarginsProps) => (before ? `&:before { ${marginRules} }` : '')}
+    ${({ after }: WithMarginsProps) => (after ? `&:after { ${marginRules} }` : '')}
+  }
+`;
```
SiAdcock, line 24: In the past we have recommended not using `styled` to create or extend components. Every `styled` component rendered creates an additional `Context.Consumer` component, which can lead to [performance issues](https://calendar.perfplanet.com/2019/the-unseen-performance-costs-of-css-in-js-in-react-apps/). Things may have changed since the analysis in that article was performed. Also this will only become appararent if you use these components a significant number of times on your page (say, more than 50).

The alternative using `css` is more verbose:

```tsx
const withMarginsStyle = ({before, after}: WithMarginsProps) => css`
  display: flex;
  ${from.tablet} {
    ${before ? `&:before { ${marginRules} }` : '')}
    ${after ? `&:after { ${marginRules} }` : '')}
  }
`

export const WithMargins = ({children, ...props}) => (
  <div css={withMarginsStyle(props)}>
    {children}
  </div>
)
```

I do find this approach clearer, and more consistent with the rest of your components. The inline functions in the `styled` approach are a bit magical. That might just be a personal thing 😊

SiAdcock: Actually I just realised you can this more efficient and avoid the logic in the CSS altogether:

```tsx
const withMarginsStyle = css`
  display: flex;
`

const afterStyle = css`
  ${from.tablet} {
    &:after { 
      ${marginRules} 
    }
  }
`

const beforeStyle = css`
  ${from.tablet} {
    &:before { 
      ${marginRules} 
    }
  }
`

export const WithMargins = ({children, before, after}) => (
  <div css={[
     withMarginsStyle,
     before ? beforeStyle : '',
     after ? afterStyle : '',
   ]}>
    {children}
  </div>
)
```

SiAdcock: This is the most verbose, but avoiding dynamic styles is, in the long term, a worthwhile investment. If all styles can be made static, they can be identified and extracted at build time. This would avoid the user's device having to do the work of calculating which styles to apply, and may save us from needing a runtime like Emotion running on the client at all (I'm pushing [these docs for Treat JS](https://seek-oss.github.io/treat/background) on everyone I know right now!)

support-frontend/assets/components/footerCompliant/BackToTop.jsx
SiAdcock: Interesting! @i-hardy makes a good point. Here are the [available font wieghts for each font](https://www.theguardian.design/2a1e5182b/p/95d5d0-code/t/98bef6).

support-frontend/assets/components/footerCompliant/containers/contentStyles.js
i-hardy: That makes a lot of sense, thanks! And I agree on the benefits of minimising the work that has to get done on the device. I'll make some changes.

support-frontend/assets/components/footerCompliant/footerStyles.js
```diff
@@ -2,24 +2,81 @@
...
 import { css } from '@emotion/core';
 import { textSans } from '@guardian/src-foundations/typography';
-import { neutral } from '@guardian/src-foundations/palette';
+import { brand, neutral, brandText, brandAlt } from '@guardian/src-foundations/palette';
+import { from, until } from '@guardian/src-foundations/mq';
 import { space } from '@guardian/src-foundations';
 
 export const componentFooter = css`
-  padding-top: ${space[2]}px;
+  background-color: ${brand[400]};
```
SiAdcock, line 10: Maybe we can use [`brandBackground.primary`](https://www.theguardian.design/2a1e5182b/p/1377a6-tokens/t/63cc10) to be more explicit

support-frontend/assets/components/footerCompliant/footerStyles.js
```diff
@@ -2,24 +2,81 @@
...
   a {
-    color: ${neutral[100]};
-  text-decoration: underline;
-    :hover, :visited {
+    color: ${brandText.anchorPrimary};
+    :hover {
       text-decoration: underline;
-      color: ${neutral[100]};
+      color: ${brandAlt[400]};
     }
```
SiAdcock, line 19: I would try to avoid nesting styles where possible. It's better to attach the styles directly to the element to which they ought to be applied. It makes it easier to update / delete the styles later, as you have a better understanding on the impact.

support-frontend/assets/components/footerCompliant/footer.jsx
```diff
@@ -50,11 +50,30 @@ function Footer({
...
             </Rows>
           </div>
-        </Content>
+        </FooterContent>
       }
-      <Content border appearance={appearance}>
+      <FooterContent border>
+        <ul css={linksList}>
+          <li css={link}>
+            <a href="https://www.theguardian.com/help/privacy-policy">Privacy Policy</a>
```
SiAdcock, line 57: Would the [Link component](https://www.theguardian.design/2a1e5182b/p/43c26b-link/b/048fd1) be of use here?

support-frontend/assets/components/footerCompliant/footerStyles.js
i-hardy: I'm going to leave it in for now as it's applying styling to various links in the legal text passed in, which otherwise show up as browser defaults. Lucy and I were planning to pair on actually rolling this out onto pages next week so I've left a comment to come back to it, as it would definitely be preferable to reduce the specificity here.

support-frontend/package.json
```diff
@@ -76,6 +76,7 @@
     "@guardian/src-choice-card": "^1.6.0",
     "@guardian/src-foundations": "^1.6.0",
     "@guardian/src-icons": "^1.6.0",
+    "@guardian/src-link": "^2.0.0",
```
SiAdcock, line 79: This is probably fine as it is, but in general I'd recommend keeping the version numbers of all `src-*` components in sync because I [can't guarantee compatibility](https://www.theguardian.design/2a1e5182b/p/876251-getting-started/t/9634cf) across major versions 🙏

support-frontend/assets/components/footerCompliant/footerStyles.js
lucymonie: Yes, it seems like we should restructure this.

support-frontend/assets/components/footerCompliant/BackToTop.jsx
lucymonie: Ah, maybe that's the problem. I think I've mostly just been using 'bold' though...

support-frontend/assets/components/footerCompliant/BackToTop.jsx
```diff
@@ -0,0 +1,69 @@
...
+  brandText,
+  brandAlt,
+} from '@guardian/src-foundations/palette';
+import { textSans } from '@guardian/src-foundations/typography';
+
+const iconHeight = '42px';
+
+const iconContainer = css`
+    position: relative;
+    float: right;
```
lucymonie, line 14: Is float actually needed here? I think the position absolute below will handle all this for you.

i-hardy: Float enables the white circle behind the icon to actually have an intrinsic width and height, as it's an inline element; using `display: inline-block` works the same but requires more faffing about to position it relative to the text, whereas with float it just sits where it's supposed to.

lucymonie: Okay cool.

PR #2635, opened 2020-07-30
Add ReaderType to ProductRatePlans

rupertbates:
## Why are you doing this?
Up until now [Product Rate Plans](https://knowledgecenter.zuora.com/Central_Platform/API/G_SOAP_API/E1_SOAP_API_Object_Reference/ProductRatePlan) as modelled in the support stack have been made up of `billingPeriod`, `fulfilmentOptions` and `productOptions` where: 
- `fulfilmentOptions` relates to how the product is distributed to the customer (eg. `HomeDelivery` or `Collection` for paper) 
- `productOptions` relates to variations in the product offering (eg. `Weekend` or `EveryDay` for paper).

To model gift subscriptions for Guardian Weekly an additional boolean `fixedTerm` was then added as all the gift rate plans were fixed term.  

After that to support corporate Digital Subscriptions a new `ProductOption` was added - `Corporate` which applied to Digisubs only.

There are a couple of problems with these two additions:
- It is not certain that all gift rate plans will continue to be fixed term going forward, and even if they are the relationship between whether a rate plan is fixed term and whether it is a gift is non-obvious.
- By making `Corporate` a `ProductOption` it becomes difficult to model corporate subscriptions for paper should that become necessary as for paper the `productOption` is already used to determine the frequency of the subscription.
- Corporate subs and Gift subs are actually very similar concepts but are modelled in very different ways (an additional boolean vs a new `ProductOption`).

The solution to all of these issues is to remove the corporate product option and the fixedTerm boolean and replace them with the concept of a `ReaderType` which can be either `Direct` - standard user, `Corporate` or `Gift`. `ReaderType` already exists in the domain model for exactly this purpose so this change is just to introduce it as a field on the `ProductRatePlan` case class and to use it when creating subscriptions.

[**Trello Card**](https://trello.com/c/EVYAhPNK/3173-gifting-system-integration-zuora-create-new-limit-term-products)

### DIFF
--- a/support-models/src/main/scala/com/gu/support/catalog/Product.scala
+++ b/support-models/src/main/scala/com/gu/support/catalog/Product.scala
@@ -4,7 +4,8 @@ import com.gu.i18n.CountryGroup
 import com.gu.support.config.TouchPointEnvironment
 import com.gu.support.config.TouchPointEnvironments.{PROD, SANDBOX, UAT}
 import com.gu.support.workers._
-import com.gu.support.zuora.api.ReaderType.{Corporate, Gift}
+import com.gu.support.zuora.api.ReaderType
+import com.gu.support.zuora.api.ReaderType.{Corporate, Direct, Gift}
 import io.circe.{Decoder, Encoder}
 
 import scala.collection.immutable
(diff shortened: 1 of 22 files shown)


support-models/src/main/scala/com/gu/support/catalog/Product.scala
```diff
@@ -39,17 +40,23 @@ case object DigitalPack extends Product {
...
+        ProductRatePlan("2c92c0f971c65df50171dfabef87093d", Monthly, NoFulfilmentOptions, NoProductOptions, "Digital Subscription Redemption Code",
+          readerType = Corporate
+        ),
       ),
       SANDBOX -> List(
         ProductRatePlan("2c92c0f84bbfec8b014bc655f4852d9d", Monthly, NoFulfilmentOptions, NoProductOptions, "Digital Subscription Monthly"),
         ProductRatePlan("2c92c0f94bbffaaa014bc6a4212e205b", Annual, NoFulfilmentOptions, NoProductOptions, "Digital Subscription Annual"),
-        ProductRatePlan("2c92c0f971c65dfe0171c6c1f86e603c", Monthly, NoFulfilmentOptions, Corporate, "Digital Subscription Redemption Code"),
+        ProductRatePlan("2c92c0f971c65dfe0171c6c1f86e603c", Monthly, NoFulfilmentOptions, NoProductOptions, "Digital Subscription Redemption Code",
+          readerType = Corporate
```
johnduffell, line 58: something funny about the formatting here, all the parameters are for the same `ProductRatePlan` object I think? but some of them are on one line and not others. (and mostly positional but one named)

if so I'd prefer to be consistent, either all on one line if readable, if not all on separate lines.
If it gets that we can't see the wood for the trees we can factor out the repetition a bit better.

support-models/src/main/scala/com/gu/support/workers/ProductTypeRatePlans.scala
```diff
@@ -1,46 +1,48 @@
...
+      productRatePlan.readerType == readerType
     )
   }
 
   implicit val digitalRatePlan: ProductTypeRatePlans[DigitalPack] = (product: DigitalPack, environment: TouchPointEnvironment, fixedTerm) =>
     product.catalogType.ratePlans.getOrElse(environment, Nil).find(productRatePlan =>
-      (productRatePlan.billingPeriod == product.billingPeriod && productRatePlan.productOptions == product.productOptions) ||
-        (productRatePlan.productOptions == Corporate && product.productOptions == Corporate) // We don't care about the billing period for corporates
+      (productRatePlan.billingPeriod == product.billingPeriod && productRatePlan.readerType == product.readerType) ||
+        (productRatePlan.readerType == Corporate && product.readerType == Corporate) // We don't care about the billing period for corporates
```
johnduffell, line 45: interestinlgly we will care about the billing period for gifts.

I would like to get rid of all these implicits and just use normal functions, a bit like we did with the productSubscriptionBuilders.  Something to keep in mind if we make significant changes to the code that uses them.

support-models/src/main/scala/com/gu/support/workers/Products.scala
```diff
@@ -61,7 +63,7 @@ case class GuardianWeekly(
 object ProductType {
   import com.gu.support.encoding.CustomCodecs._
   implicit val decoderDigital: Decoder[DigitalPack] = deriveDecoder[DigitalPack]
-    .prepare(_.withFocus(_.mapObject(_.checkKeyExists("productOptions", Json.fromString("NoProductOptions")))))
+    .prepare(_.withFocus(_.mapObject(_.checkKeyExists("readerType", Json.fromString("Direct")))))
```
johnduffell, line 66: I realise it's existing but what is this for? it looks like it's saying that we will refuse to decide non-Direct digitalpack, is that just a safety check until corporate and gift are released?

rupertbates: It's to handle defaulting the `readerType` to `Direct` if it is not present in the Json. Circe doesn't respect default values in case classes so we need to do a bit more work to avoid having to add it in everywhere.

support-models/src/main/scala/com/gu/support/catalog/Product.scala
rupertbates: It's because the last one is the only one which supplies a `readerType` the others just accept the default of `Direct`. I'll rejig the formatting to make it a bit neater.

Jpsi:
I agree, very clear description! 👍

support-models/src/main/scala/com/gu/support/workers/Products.scala
johnduffell: ah I see so "check" is more like "ensure" rather than "assert", makes sense now

support-models/src/main/scala/com/gu/support/workers/ProductTypeRatePlans.scala
Jpsi: Noted, thank you!

PR #2637, opened 2020-07-30
gift code generator logic

johnduffell:
## Why are you doing this?

We want to generate codes for gifting.  Eventually they will be issued by email to people who are buying gifts and who are receiving gifts.

This PR adds the code to generate those codes, and the relevant tests, as a standalone unit.

It is important to get this right as if we start giving everyone the same code or they are too predictable, codes could be stolen or not possible to create. Hence the separate PR.

[**Trello Card**](https://trello.com/c/fZGSle0b/3222-gifting-backend-logic-to-create-the-actual-codes)

## Code format
codes are in the form gd03-asdfgh
g = gift as oppsed to corporate
d = digi sub (could have been ds for clarity if preferred for this and corporate - consult userhelp/cust ops perhaps?)
03 = duration in months (currently 03, 06, 12 and recurring is not yet supported)
asdfgh = random code for security to discourage cracking (TODO this should be longer now as to guess one would be 1 year but if there are lots issued the chance could increase)

lower case as it matches Url conventuons better and easier to type in.


rupertbates:
Perhaps add a description of why the codes are in the form that they are to the PR? Have we documented it anywhere?

support-services/src/main/scala/com/gu/support/redemption/generator/GiftGenerator.scala
```diff
@@ -0,0 +1,84 @@
...
+
+  def apply(code: IntsToTypableString.Code): GenerateGiftCode =
+    (duration: GiftDuration) => {
+      val init = prefix + duration.code + "-"
+      GiftCode(init + code.value).get
+    }
+
+}
+
+object IntsToTypableString {
```
rupertbates, line 57: I'm not that keen on `Typable` could we call this `IntsToCode` or `IntsToCodeSuffix` or I think my preference would be `CodeSuffixGenerator`?

support-services/src/main/scala/com/gu/support/redemption/generator/GiftGenerator.scala
```diff
@@ -0,0 +1,84 @@
...
+
+  case class Code private(value: String) extends AnyVal
```
rupertbates, line 59: And call this class `CodeSuffix` to avoid confusion with the `GiftCode` class?

support-services/src/main/scala/com/gu/support/redemption/generator/GiftGenerator.scala
```diff
@@ -0,0 +1,84 @@
...
+  object Code {
```
rupertbates, line 60: Then maybe this object `Suffix`?

support-services/src/main/scala/com/gu/support/redemption/generator/GiftGenerator.scala
```diff
@@ -0,0 +1,84 @@
+package com.gu.support.redemption.generator
+
+import java.security.SecureRandom
+
+import com.gu.support.redemption.generator.ConstructCode.GenerateGiftCode
+
+object GiftGenerator {
```
rupertbates, line 7: GiftCodeGenerator might be more exact

support-services/src/test/scala/com/gu/support/redemption/generator/GiftGeneratorSpec.scala
```diff
@@ -0,0 +1,52 @@
...
+    ConstructCode(IntsToTypableString.Code("000000").get)(GiftDuration.Gift3Month).value should be("gd03-000000")
+    ConstructCode(IntsToTypableString.Code("000000").get)(GiftDuration.Gift6Month).value should be("gd06-000000")
+    ConstructCode(IntsToTypableString.Code("000000").get)(GiftDuration.Gift12Month).value should be("gd12-000000")
+  }
+
+}
+
+class IntsToTypableStringSpec extends AnyFlatSpec with Matchers {
+
+  it should "extreme values" in {
```
rupertbates, line 37: This test name doesn't really make sense

PR #2638, opened 2020-07-31
make the IT tests run daily rather than hourly

johnduffell:
## Why are you doing this?

The IT tests are putting too many records into SF dev, until this is resolved we will make them run daily instead of hourly

[**Trello Card**](https://trello.com/c/LKwe8ApH/3241-auto-clean-it-test-data-from-sf-dev)

### DIFF
--- a/support-lambdas/it-test-runner/cfn.yaml
+++ b/support-lambdas/it-test-runner/cfn.yaml
@@ -40,7 +40,7 @@ Resources:
         RerunTests:
           Type: Schedule
           Properties:
-            Schedule: 'rate(1 hour)'
+            Schedule: 'rate(1 day)'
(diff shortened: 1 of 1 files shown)


support-lambdas/it-test-runner/cfn.yaml
```diff
@@ -40,7 +40,7 @@ Resources:
         RerunTests:
           Type: Schedule
           Properties:
-            Schedule: 'rate(1 hour)'
+            Schedule: 'rate(1 day)'
```
johnduffell, line 43: "day" is from https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/ScheduledEvents.html

PR #2639, opened 2020-07-31
A/B test disabling Stripe's fraud detection

paulbrown1982:
## Why are you doing this?
If Stripe's fraud detection is enabled then it drops cookies. For TCFv2 compliance we do not want this.
So we're A/B testing disabling fraud detection


paulbrown1982:
@tomrf1 If this stops any cookies from being set by the Stripe SDK (and thus excludes it from PECR), then I may need your help on Monday to make this an A/B test so we can measure if there's any impact of having no fraud detection.

PR #2642, opened 2020-08-03
Defer loading of paypal + amazon pay

tomrf1:
## Why are you doing this?
For TCFv2 compliance we must only load these SDKs once the user has shown intent.
So we will now only load them when the user selects the relevant payment method.
This change only affects:
1. PayPal Express, which is only used for recurring,
2. Amazon pay, which is only for single.

Unaffected by this change:
1. PayPal for single, because it does not use an SDK
2. Gocardless
3. Stripe - we are trying a separate solution for this

This may result in a short delay in seeing the button when the user selects the payment method, as we are no longer initialising the SDKs on page load.
I have added a progress 'spinner' for the paypal express button, e.g.:
Screen Shot 2020-08-03 at 11 54 52

And we are already using one for the amazon pay button:
Screen Shot 2020-08-03 at 12 00 13

### Testing
Does not affect the 'existing payment method' feature, for when you arrive via MMA:
Screen Shot 2020-08-03 at 13 16 08

### DIFF
--- a/support-frontend/assets/pages/contributions-landing/contributionsLandingActions.js
+++ b/support-frontend/assets/pages/contributions-landing/contributionsLandingActions.js
@@ -46,11 +46,15 @@ import type { Action as PayPalAction } from 'helpers/paymentIntegrations/payPalA
 import { setFormSubmissionDependentValue } from './checkoutFormIsSubmittableActions';
 import { type State, type ThankYouPageStage, type UserFormData, type Stripe3DSResult } from './contributionsLandingReducer';
 import type { PaymentMethod } from 'helpers/paymentMethods';
-import { AmazonPay, DirectDebit, Stripe } from 'helpers/paymentMethods';
+import { AmazonPay, DirectDebit, PayPal, Stripe } from 'helpers/paymentMethods';
 import type { RecentlySignedInExistingPaymentMethod } from 'helpers/existingPaymentMethods/existingPaymentMethods';
 import { ExistingCard, ExistingDirectDebit } from 'helpers/paymentMethods';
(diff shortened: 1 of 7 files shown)


support-frontend/assets/pages/contributions-landing/contributionsLandingActions.js
```diff
@@ -251,6 +261,31 @@ const setUserTypeFromIdentityResponse =
         ({ type: 'SET_USER_TYPE_FROM_IDENTITY_RESPONSE', userTypeFromIdentityResponse })));
     };
 
+// We defer loading 3rd party payment SDKs until the user selects one, or one is selected by default
+const loadPaymentSdkIfNecessary = (
```
paulbrown1982, line 265: Would it be cleaner to split this into 2 functions, and have the 2 callees listen to the relevant 'begun-loading' variable and select the right function name, rather than passing them through?

`const loadPayPalSdkIfNecessary = (contributionType) => (dispatch: Function): void`
`loadAmazonPaySdkIfNecessary = (countryGroupId, isTestUser), => (dispatch: Function): void`

tomrf1: yes, but then each callee has to `switch` on the `paymentMethod`, which seemed worse to me. Which do you prefer?

paulbrown1982: There are only 2 callees at the moment, so I vote for separate functions rather than argument expansion. Least power principle applies to arguments sets as well as argument types.

tomrf1: 👍

PR #2644, opened 2020-08-04
Fix subscriptions checkouts post deploy tests

Jpsi:
## Why are you doing this?

- **Background**: We use automated tests on the production site to check for breaking changes after a new deploy (aka post-deploy tests - see their [wikipage](https://github.com/guardian/support-frontend/wiki/Post-Deployment-Testing)).
- **Problem**: We have had to disable the subscriptions checkouts post deploy tests because [this Identity PR](https://github.com/guardian/identity-frontend/pull/593) stopped selenium from being able to create a test user via the Identity site.
- **Solution**: We will need to test create a test user via an api call and use this in the post deploy tests. More details are [here](https://docs.google.com/document/d/1R6aJ17x8iN0p7H2VnZE4hAIQbdT41eTciRUYnQ3YfE8/edit?ts=5ee8cadf)

[**Trello Card**](https://trello.com/c/VSBybe3U)

## Changes

* There are now two ways to create a PostDeployTestUser - one using the traditional test username approach (using `com.gu.identity.testing.usernames.TestUsernames`), and a new one using cookies generated by the new endpoint (introduced in https://github.com/guardian/identity/pull/1768). The former is used for the Contributions checkouts tests, whereas the latter is used for the Subscriptions checkouts tests.


support-frontend/test/selenium/util/PostDeployTestUser.scala
```diff
@@ -24,3 +28,31 @@ class PostDeployTestUser(driverConfig: DriverConfig) extends TestUser {
...
+
+  private val client = new OkHttpClient()
+  private val requestToIdapi = new Request.Builder()
+    .url(Config.idapiNewTestUserUrl)
+    .addHeader(Config.idapiClientAccessTokenName, Config.idapiClientAccessTokenSecret)
+    .post(RequestBody.create(MediaType.parse("application/json"), "{}"))
+    .build()
+  private val response = client.newCall(requestToIdapi).execute()
+  private val responseBodyAsJson = parse(response.body().string()).getOrElse(Json.Null)
+  private val cursor = responseBodyAsJson.hcursor
```
johnduffell, line 42: I would probably pull the identity call out into a separate class that returns the cookie values in a `List[(String, String)]` as a minimum.
You can then check the response code and the json structure in there, and either throw a suitable exception (or even put it into an `Either`) so that we can diagnose things when they go wrong.
As it is, we might just throw NoSuchElementException on line 50/51 and it wouldn't be clear what went wrong.
Ideally we can have enough information to decide is the a bug or a transient identity issue and enough to raise an issue with Identity team without having to run them locally.

support-frontend/test/selenium/util/PostDeployTestUser.scala
```diff
@@ -24,3 +28,31 @@ class PostDeployTestUser(driverConfig: DriverConfig) extends TestUser {
...
+
+  private def addTestUserCookies(testUsername: String) = {
+    driverConfig.addCookie(name = "pre-signin-test-user", value = testUsername)
```
johnduffell, line 45: do we need this one for subs? I think only contributions allows non signed in test users

support-frontend/test/selenium/util/PostDeployTestUser.scala
```diff
@@ -24,3 +28,31 @@ class PostDeployTestUser(driverConfig: DriverConfig) extends TestUser {
...
+    driverConfig.addCookie(name = "_post_deploy_user", value = "true") // This enables the tests to use the mocked payment services
+    driverConfig.addCookie(name = "GU_TK", value = "1.1") //To avoid consent banner, which messes with selenium
+    for (n <- 0 until 3) {
+      val cookie_name = cursor.downField("values").downN(n).downField("key").as[String].right.get
+      val cookie_value = cursor.downField("values").downN(n).downField("value").as[String].right.get
+      driverConfig.addCookie(name = cookie_name, value = cookie_value)
+    }
+  }
+
+  val username = "postDeployTestUser"
```
johnduffell, line 56: why can this just be hard coded, I assumed it would have to be a proper test user string?

support-frontend/test/selenium/conf/selenium-test.conf
```diff
@@ -9,6 +9,11 @@ identity {
   webapp.url = ${?IDENTITY_WEBAPP_URL}
   test.users.secret = ${?IDENTITY_TEST_USERS_SECRET}
 }
+idapi {
+  newTestUser.url = ${?IDAPI_NEW_TEST_USER_URL}
+  clientAccessToken.name = ${?IDAPI_CLIENT_ACCESS_TOKEN_NAME}
+  clientAccessToken.secret = ${?IDAPI_CLIENT_ACCESS_TOKEN_SECRET}
+}
```
johnduffell, line 16: not something you've added, but it's funny that this whole file seems to be just an indirection from lower case dotted format to caps with underscores.

I'd be tempted to hard code the test user url in the code as it's not a secret and it's known at compile time (at least per stage) so that would simplify the amount of steps to get to the real value.

support-frontend/test/selenium/util/PostDeployTestUser.scala
```diff
@@ -24,3 +28,31 @@ class PostDeployTestUser(driverConfig: DriverConfig) extends TestUser {
...
+class PostDeployTestUserSubs(driverConfig: DriverConfig) extends TestUser {
+
+  private val client = new OkHttpClient()
+  private val requestToIdapi = new Request.Builder()
+    .url(Config.idapiNewTestUserUrl)
+    .addHeader(Config.idapiClientAccessTokenName, Config.idapiClientAccessTokenSecret)
+    .post(RequestBody.create(MediaType.parse("application/json"), "{}"))
+    .build()
+  private val response = client.newCall(requestToIdapi).execute()
+  private val responseBodyAsJson = parse(response.body().string()).getOrElse(Json.Null)
```
rupertbates, line 41: It would be good to deserialise this response to a case class using [Circe](https://github.com/circe/circe). There are loads of examples of this in the code base, just ask me on chat if you'd like any pointers

support-frontend/test/selenium/util/PostDeployTestUser.scala
```diff
@@ -24,3 +28,31 @@ class PostDeployTestUser(driverConfig: DriverConfig) extends TestUser {
...
+  private val cursor = responseBodyAsJson.hcursor
+
+  private def addTestUserCookies(testUsername: String) = {
+    driverConfig.addCookie(name = "pre-signin-test-user", value = testUsername)
+    driverConfig.addCookie(name = "_test_username", value = testUsername, domain = Some(Config.guardianDomain))
+    driverConfig.addCookie(name = "_post_deploy_user", value = "true") // This enables the tests to use the mocked payment services
+    driverConfig.addCookie(name = "GU_TK", value = "1.1") //To avoid consent banner, which messes with selenium
+    for (n <- 0 until 3) {
```
rupertbates, line 49: Once you have an object structure for the response via Circe you can `.forEach` over the `values` array to add them. That will mean that if a cookie is added or removed this code won't need to change.

support-frontend/test/selenium/util/PostDeployTestUser.scala
```diff
@@ -24,3 +28,31 @@ class PostDeployTestUser(driverConfig: DriverConfig) extends TestUser {
   val username = testUsers.generate()
   addTestUserCookies(username)
 }
+
+class PostDeployTestUserSubs(driverConfig: DriverConfig) extends TestUser {
+
+  private val client = new OkHttpClient()
```
rupertbates, line 34: An OkHttpClient is expensive to create so you should create it as a val on an object and reuse it for all tests

support-frontend/test/selenium/util/PostDeployTestUser.scala
Jpsi: Got it, thanks! Addressed in the new commit

support-frontend/test/selenium/util/PostDeployTestUser.scala
Jpsi: 👍

support-frontend/test/selenium/util/PostDeployTestUser.scala
Jpsi: Thank you! Addressed this in the new commit, do let me know if it's better.

support-frontend/test/selenium/util/PostDeployTestUser.scala
Jpsi: I moved the OkHttpClient into an object, but also the request itself, meaning that all tests would now use the same test user cookies. Not sure if that makes more sense, or if each test should re-request a fresh set of cookies (using the same singleton OkHttpClient)

support-frontend/test/selenium/util/PostDeployTestUser.scala
```diff
@@ -31,28 +34,56 @@ class PostDeployTestUserContribs(driverConfig: DriverConfig) extends TestUser {
...
+  }
+
+  def decodeCookies: List[IdapiCookie] = {
+    val responseDecoded: Either[DecodingFailure, IdapiSuccessfulResponse] = getResult.as[IdapiSuccessfulResponse]
+    var cookies = List[IdapiCookie]()
+    responseDecoded match {
+      case Left(l) => logger.error(s"Failed to decode IDAPI response: ${l}")
+      case Right(r) => cookies = r.values
+    }
+    cookies
```
johnduffell, line 87: avoid using a var, and avoid hiding errors, if you can't handle them then just return a suitable failable type (either/try/option)
```suggestion
    responseDecoded match {
      case Left(error) => {
        logger.error(s"Failed to decode IDAPI response: $error")
        None
      }
      case Right(r) => Some(r.values)
    }
```

support-frontend/test/selenium/util/PostDeployTestUser.scala
```diff
@@ -31,28 +34,56 @@ class PostDeployTestUserContribs(driverConfig: DriverConfig) extends TestUser {
...
+
+  private val request: Request = new Request.Builder()
+    .url(Config.idapiNewTestUserUrl)
+    .addHeader(Config.idapiClientAccessTokenName, Config.idapiClientAccessTokenSecret)
+    .post(RequestBody.create(MediaType.parse("application/json"), "{}"))
+    .build()
+
+  private def getResult: Json = {
+    val response: Response = client.newCall(request).execute()
+    parse(response.body().string()).getOrElse(Json.Null)
```
johnduffell, line 77:
```suggestion
    val body = response.body().string()
    parse(body) match {
      case None => Left("could not JSON parse body $body")
      case Some(json) => Right(json)
    }
```

support-frontend/test/selenium/util/PostDeployTestUser.scala
```diff
@@ -31,28 +34,56 @@ class PostDeployTestUserContribs(driverConfig: DriverConfig) extends TestUser {
...
+  def decodeCookies: List[IdapiCookie] = {
+    val responseDecoded: Either[DecodingFailure, IdapiSuccessfulResponse] = getResult.as[IdapiSuccessfulResponse]
+    var cookies = List[IdapiCookie]()
+    responseDecoded match {
+      case Left(l) => logger.error(s"Failed to decode IDAPI response: ${l}")
+      case Right(r) => cookies = r.values
+    }
+    cookies
+  }
+}
```
johnduffell, line 89: The separation looks great, I would just try to standardise on a failure type e.g. `Either[String, A]` and use that throughout.
The purpose of failure types is to consolidate error handling where the error can be dealt with (similar to exceptions)
So where the errors are detected, push them into the `Left` straight away and make any error producing function return an Either.
Then in the code that uses the cookies, we can add the cookies if the call worked, but if it didn't, don't even try to run the test, just abort and fail the test with the error message stored in the Left.
As it is, the test will run and get stuck at the log in screen, which would cause a bit of head scratching to work out actually e.g. identity returned a 5xx error.

support-frontend/test/selenium/util/PostDeployTestUser.scala
Jpsi: That makes a lot of sense (I was trying to get it to do something similar but couldn't get the syntax right). Thank you!

support-frontend/test/selenium/util/PostDeployTestUser.scala
Jpsi: Makes sense - thank you!

Jpsi: @johnduffell - I just read up on Try/Success/Failure, which to my beginner eyes looks like a more readable structure than Either/Left/Right, at least at first sight. Is there a benefit in using the latter?

johnduffell: sorry only just saw this.

Either is a generic type on both sides, so you can put whatever you like in either the left or right.  Convention is that the left is failure, but people certainly use Either in other contexts and it's perfectly valid.

Try is specifically relating to exceptions so it's like an Either that the left is hard coded to Throwable.
The naming reflects that specificity, but the downside is you have to wrap any message you have in a Throwable which has a whole load more baggage.

The other option (which I don't recommend here) is to have a custom failure type and you can call it what you want and hard code the failure side to contain a String.

In this case I think Either with a String as the failure i.e. `Either[String,A]` is your best option.  Although the naming is a bit odd, it fits with convention, and you won't have to mess with Throwables.

Jpsi: I now see the logic (that Try/Success/Failure involves the use of a Throwable, which has more baggage than just a String). Thank you for explaining!

support-frontend/test/selenium/util/IdapiBypassRecaptchaRequest.scala
```diff
@@ -0,0 +1,54 @@
...
+  def getCookies(): Either[String, List[IdapiCookie]] = {
+    getResult() match {
+      case Left(error) => Left(error)
+      case Right(result) => {
+        result.as[IdapiSuccessfulResponse] match {
+          case Left(error) => Left(s"Could not deserialise IDAPI response: $error")
+          case Right(r) => Right(r.values)
+        }
+      }
+    }
```
johnduffell, line 40: when using a Either or other specific error type, you can use a for comprehension get rid of most of the boilerplate but maintain the error in the background
```suggestion
    for {
      result <- getResult()
      deserialised <- result.as[IdapiSuccessfulResponse].leftMap(error => s"Could not deserialise IDAPI response: $error")
    } yield deserialised.values
```

support-frontend/test/selenium/util/IdapiBypassRecaptchaRequest.scala
```diff
@@ -0,0 +1,54 @@
...
+    .post(RequestBody.create(MediaType.parse("application/json"), "{}"))
+    .build()
+
+  private def getResult(): Either[String, Json] = {
+    val response: Response = client.newCall(request).execute()
+    val body: String = response.body().string()
+    parse(body) match {
+      case Left(error) => Left(s"Could not parse $body as JSON from response with code ${response.code()}: $error")
+      case Right(json) => Right(json)
+    }
```
johnduffell, line 28: if you're only changing the left you can use leftMap
```suggestion
    parse(body).leftMap { error =>
      s"Could not parse $body as JSON from response with code ${response.code()}: $error"
    }
```

support-frontend/test/selenium/util/PostDeployTestUser.scala
```diff
@@ -4,86 +4,32 @@ import java.time.Duration.ofDays
...
     driverConfig.addCookie(name = "GU_TK", value = "1.1") //To avoid consent banner, which messes with selenium
-    IdapiRequest.decodeCookies.foreach { cookie =>
-      driverConfig.addCookie(name = cookie.key, value = cookie.value)
+    bypassRecaptchaCookies match {
+      case None => 
+      case Some(cookies) => {
+        cookies.foreach {cookie =>
+          driverConfig.addCookie(name = cookie.key, value = cookie.value)
+        }
+      }
```
johnduffell, line 29: you can avoid the pattern match by flattening the Option and List together
```suggestion
    bypassRecaptchaCookies.toList.flatten.foreach { cookie =>
      driverConfig.addCookie(name = cookie.key, value = cookie.value)
    }
```

support-frontend/test/selenium/util/IdapiBypassRecaptchaRequest.scala
Jpsi: Nice!

support-frontend/test/selenium/util/IdapiBypassRecaptchaRequest.scala
Jpsi: Will read up on for comprehension now

support-frontend/test/selenium/util/PostDeployTestUser.scala
Jpsi: I see!

Jpsi: Added

support-frontend/test/selenium/util/IdapiBypassRecaptchaRequest.scala
Jpsi: Added

support-frontend/test/selenium/util/PostDeployTestUser.scala
Jpsi: Reverted back to using a test user string, which is needed for the Contributions checkouts tests.

support-frontend/test/selenium/util/PostDeployTestUser.scala
Jpsi: I am not sure about the purpose of this cookie, but would be happy to investigate further. Removing it causes some of the Stripe tests to fail, so I'd leave it in to limit the scope of the current PR.

support-frontend/test/selenium/conf/selenium-test.conf
Jpsi: Could do! I'd leave it as it is for now as this way all the IDAPI endpoint config values are stored in the same place in Travis, rather than being scattered in different places.

support-frontend/test/selenium/util/IdapiBypassRecaptchaRequest.scala
Jpsi: I had a look at this one for a bit and I'm still not sure how this particular for comprehension gets resolved. Leaving it out for a future refactoring PR.

support-frontend/test/selenium/subscriptions/ProductPagesSpec.scala
```diff
@@ -58,7 +58,10 @@ class ProductPagesSpec extends AnyFeatureSpec
 
   Feature("Digital Pack product page") {
     Scenario("Basic loading") {
-      val testUser = new PostDeployTestUser(driverConfig)
+      val testUser = IdapiTestUserRequest.getCookies match {
```
rupertbates, line 61: I don't think this user needs cookies, we're just testing that the page loads and you don't need to be signed in for that.

support-frontend/test/selenium/util/PostDeployTestUser.scala
rupertbates: I think it might be better to get a new set for each test because for Digital Sub if the user already has a subscription and tries to checkout again then they are directed to a different page. This would mean that if we had several Digisub tests (eg. for different currencies) the second one wouldn't be able to access the checkout, it could be quite confusing for whoever was trying to write the test.

support-frontend/test/selenium/subscriptions/ProductPagesSpec.scala
Jpsi: Nice catch - fixed!

support-frontend/test/selenium/util/PostDeployTestUser.scala
Jpsi: Got it, definitely makes sense. Fixed.

rupertbates (approved):
Nice work 👍

PR #2645, opened 2020-08-04
server-side A/B test

tomrf1:
Replaces the existing (but currently unused) server-side testing logic. This is because it was setting a cookie, which we do not need.
I could have assigned variants based on the MVT cookie, but decided it's fine for now to do it randomly.
It also still just supports 'Control' and 'Variant', as this is still all we need for now.

I've also made it general enough that any page can use it. So the `main.scala.html` template now takes a `Map[String,Participation]` containing any server-side tests, and the client picks them up in `window.guardian.serversideTests` so that they can be recorded in the acquisition event along with all the client-side tests.


support-frontend/app/admin/ServersideAbTest.scala
```diff
@@ -1,23 +1,31 @@
...
-      case "Variant" => Some(Variant)
-      case _ => None
-    }).getOrElse(computeParticipation)
-  }
 
-  private def computeParticipation: Participation = if (Random.nextBoolean) Control else Variant
+  def generateParticipations(testNames: List[String]): Map[String, Participation] =
+    testNames.map(_ -> computeParticipation).toMap
+
+  def computeParticipation: Participation = if (Random.nextBoolean) Control else Variant
```
paulbrown1982, line 14: So this makes our server-side A/B tests non-sticky? (That's fine, and actually gives us more options and is better for GDPR) it'd just be good to socialise this amongst dev teams so everyone is aware.

tomrf1: Correct - and it's how we've done server-side tests previously.
If we want to use MVT then we have to worry about having a unique seed value as we do with client-side tests

PR #2649, opened 2020-08-04
Update logic for showing philanthropic ask

TomPretty:
## Why are you doing this?

Remove the philanthropic ask from the contributions LP when a user comes via Apple News or Google AMP as per [this card](https://trello.com).

### DIFF
--- a/support-frontend/assets/components/legal/termsPrivacy/termsPrivacy.jsx
+++ b/support-frontend/assets/components/legal/termsPrivacy/termsPrivacy.jsx
@@ -10,6 +10,7 @@ import { type IsoCurrency, fromCountryGroupId, currencies } from 'helpers/intern
 import type { ContributionType } from 'helpers/contributions';
 import './termsPrivacy.scss';
 import type { CampaignSettings } from 'helpers/campaigns';
+import { getReferrerAcquisitionData } from 'helpers/tracking/acquisitions';
 
 
 // ---- Types ----- //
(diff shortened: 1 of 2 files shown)


support-frontend/assets/components/legal/termsPrivacy/termsPrivacy.jsx
```diff
@@ -79,6 +80,19 @@ function TermsPrivacy(props: PropTypes) {
     );
   }
 
+  const isUSContributor = props.countryGroupId === 'UnitedStates';
+  const isNotOneOffContribution = props.contributionType !== 'ONE_OFF';
+  const referrerAcquisitionData = getReferrerAcquisitionData();
```
tomrf1, line 85: We can get the `referrerAcquisitionData` from the redux store. The `getReferrerAcquisitionData` function should really only be called once, at the start.

So this component's props should instead receive a new field:
`referrerAcquisitionData: ReferrerAcquisitionData`

And the parent component can get this from redux in its `mapStateToProps` with:
`referrerAcquisitionData: state.common.referrerAcquisitionData`

Happy to pair on this!

TomPretty: I figured this might be the case! I'll give you a shout if I get stuck, but hopefully can get this sorted tomorrow morning!

TomPretty: This should be fixed now!

support-frontend/assets/pages/contributions-landing/components/ContributionForm.jsx
```diff
@@ -116,6 +117,7 @@ const mapStateToProps = (state: State) => ({
   stripeV3HasLoaded: state.page.form.stripeV3HasLoaded,
   amazonPayOrderReferenceId: state.page.form.amazonPayData.orderReferenceId,
   checkoutFormHasBeenSubmitted: state.page.form.formData.checkoutFormHasBeenSubmitted,
+  referrerSource: state.common.referrerAcquisitionData.source,
```
tomrf1, line 120: great!

PR #2652, opened 2020-08-04
Digital Subscription gift plans

rupertbates:
## Why are you doing this?

We want to add gifting as an option to the Digital Subscription product, the initial offering will be for one year or 3 month fixed term plans and this PR adds new product rate plans for these.

[**Trello Card**](https://trello.com/c/EVYAhPNK/3173-gifting-system-integration-zuora-create-new-limit-term-products)

### DIFF
--- a/support-models/src/main/scala/com/gu/support/catalog/Product.scala
+++ b/support-models/src/main/scala/com/gu/support/catalog/Product.scala
@@ -218,224 +218,73 @@ case object Paper extends Product {
 }
 
 case object GuardianWeekly extends Product {
-  private def productRatePlan(
+  private def domestic(
     id: String,
     billingPeriod:
     BillingPeriod,
-    fulfilmentOptions: FulfilmentOptions,
     description: String,
     productRatePlanChargeId: Option[String] = None,
     readerType: ReaderType = Direct
   ) = ProductRatePlan(
-    id, billingPeriod, fulfilmentOptions, NoProductOptions, description,
+    id, billingPeriod, Domestic, NoProductOptions, description,
+    productRatePlanChargeId = productRatePlanChargeId, readerType = readerType
+  )
+
+  private def restOfWorld(
+    id: String,
+    billingPeriod:
+    BillingPeriod,
+    description: String,
+    productRatePlanChargeId: Option[String] = None,
+    readerType: ReaderType = Direct
+  ) = ProductRatePlan(
+    id, billingPeriod, RestOfWorld, NoProductOptions, description,
     productRatePlanChargeId = productRatePlanChargeId, readerType = readerType
   )
 
   lazy val ratePlans: Map[TouchPointEnvironment, List[ProductRatePlan[GuardianWeekly.type]]] =
     Map(
       PROD -> List(
-        productRatePlan(
-          "2c92a0086619bf8901661ab545f51b21",
-          SixWeekly,
-          RestOfWorld,
-          "Guardian Weekly 6 for 6, rest of world delivery",
-          productRatePlanChargeId = Some("2c92a0086619bf8901661ab546091b23")
-        ),
-        productRatePlan(
-          "2c92a0fe6619b4b601661ab300222651",
-          Annual,
-          RestOfWorld,
-          "Guardian Weekly annual, rest of world delivery"
-        ),
-        productRatePlan(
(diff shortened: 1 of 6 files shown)


support-models/src/main/scala/com/gu/support/catalog/Product.scala
```diff
@@ -212,224 +218,73 @@ case object Paper extends Product {
...
-          "Guardian Weekly quarterly, domestic delivery"
-        ),
-        productRatePlan(
-          "2c92a00e6dd988e2016df85387417498",
-          Quarterly,
-          Domestic,
-          "Guardian Weekly three months, domestic delivery",
-          readerType = Gift)
-
+        restOfWorld("2c92a0086619bf8901661ab545f51b21", SixWeekly, "Guardian Weekly 6 for 6, rest of world delivery",
```
rupertbates, line 248: I've moved these back to one per line because I was finding it very hard to work with in the previous format, the list was too long to scan easily. To reduce the number of params on each line I've added a new helper method above so that it is only Six for Six plans which break onto two lines

support-services/src/test/scala/com/gu/support/pricing/PriceSummaryServiceSpec.scala
```diff
@@ -30,8 +32,25 @@ class PriceSummaryServiceSpec extends AsyncFlatSpec with Matchers {
     val digitalPack = service.getPrices(DigitalPack, List(discountPromoCode))
     digitalPack(UK)(NoFulfilmentOptions)(NoProductOptions)(Monthly)(GBP).price shouldBe 11.99
     digitalPack(UK)(NoFulfilmentOptions)(NoProductOptions)(Monthly)(GBP).promotions.head.discountedPrice shouldBe Some(8.39)
-    digitalPack(UK)(NoFulfilmentOptions)(NoProductOptions)(Annual)(GBP).price shouldBe 99
-    digitalPack(UK)(NoFulfilmentOptions)(NoProductOptions)(Annual)(GBP).promotions.head.discountedPrice shouldBe Some(91.57)
+    digitalPack(UK)(NoFulfilmentOptions)(NoProductOptions)(Annual)(GBP).price shouldBe 119
```
rupertbates, line 35: The catalog-prod.json hadn't been updated since the price rise

johnduffell: oops, I wonder if we should make it auto update when you run tests with credentials?

support-models/src/main/scala/com/gu/support/catalog/Product.scala
johnduffell: it's a tricky balance to tabulate complicated data, makes sense to keep it under review

support-models/src/main/scala/com/gu/support/catalog/Product.scala
```diff
@@ -212,224 +218,73 @@ case object Paper extends Product {
...
   ) = ProductRatePlan(
-    id, billingPeriod, fulfilmentOptions, NoProductOptions, description,
+    id, billingPeriod, Domestic, NoProductOptions, description,
+    productRatePlanChargeId = productRatePlanChargeId, readerType = readerType
+  )
+
+  private def restOfWorld(
+    id: String,
+    billingPeriod:
+    BillingPeriod,
```
johnduffell, line 236: typo
```suggestion
    billingPeriod: BillingPeriod,
```

support-models/src/main/scala/com/gu/support/catalog/Product.scala
```diff
@@ -212,224 +218,73 @@ case object Paper extends Product {
 }
 
 case object GuardianWeekly extends Product {
-  private def productRatePlan(
+  private def domestic(
     id: String,
     billingPeriod:
     BillingPeriod,
```
johnduffell, line 224: actually typo already existed 
```suggestion
    billingPeriod: BillingPeriod,
```

support-services/src/test/scala/com/gu/support/pricing/PriceSummaryServiceSpec.scala
rupertbates: I don't think it really matters, we're just testing that we can read the catalog.json successfully not that the prices are up to date.

PR #2654, opened 2020-08-05
Tidy up send email logic to make adding gifting/giftee emails easler

johnduffell:
### background
There has been previous work to simplify the send email to reduce the number of parameters passed around and tighten up the types in this PR https://github.com/guardian/support-frontend/pull/2572

This introduced a bunch of doubly nested "apply" method and inline trait implementations.

Although this was more DRY, syntactically it was over complicated and against our proposed coding recommendations https://docs.google.com/document/d/1vRIYAU2OnRkHzf4dLCuzxiA-3HsKh1PDmkiLc-1GmfQ/edit#

Ideally the code should make sense how you would explain it to someone, and there should be minimal boilerplate and maximal business logic (or at least they shouldn't be mixed together)

### summary of changes
This PR removes those traits/apply methods and replaces with normal case classes as normal parameters.  It additionally removes the inheritance/overriding from the EmailFields class and makes it into a normal class.

### why is this needed

There should be no functionality changes, it should just make more sense to all developers than the previous implementation.  Also be more conducive to have the gift emails added.

This has enabled the DigitalPackEmailFields class to have various functions pulled out standalone, and likewise shortened the logic in the actual Email Send lambda.  Both of these were (correctly) showing warnings about excessively long methods.

## Changes

* turn SubscriptionEmailFields and AllProductsEmailFields into case classes instead of traits
* pull out sensibly named functions from the long Digital pack email fields and turn into an object rather than a class.
* pull all the (almost identical) Failed email fields classes into one place and make them functions rather then extends/overrides classes.

 tested in CODE - standard DS, GW and contributions emails all worked fine.


support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
```diff
@@ -2,106 +2,81 @@ package com.gu.emailservices
...
-//  },
-//  "DataExtensionName": "digipack"
-//}
+object DigitalPackEmailFields {
 
-case class DigitalPackEmailFields(
-  paidSubPaymentData: Option[PaymentMethodWithSchedule]
-) extends SubscriptionEmailFields {
+  def build(
+    subscriptionEmailFields: SubscriptionEmailFields,
```
johnduffell, line 12: all the `apply` repetition in the old file is collapsed into this one case class

support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
```diff
@@ -2,106 +2,81 @@ package com.gu.emailservices
 
 import com.gu.emailservices.SubscriptionEmailFieldHelpers._
 import com.gu.i18n.Currency
-import com.gu.salesforce.Salesforce.SfContactId
 import com.gu.support.promotions.Promotion
 import com.gu.support.workers._
 import com.gu.support.workers.states.PaymentMethodWithSchedule
 
-// Output Json should look like this:
```
johnduffell, line 10: I don't know if this JSON was still in date but better to have it in a test than a comment.

support-workers/src/main/scala/com/gu/emailservices/FailedContributionEmailFields.scala
```diff
@@ -1,8 +1,20 @@
 package com.gu.emailservices
 
-import com.gu.salesforce.Salesforce.SfContactId
+object FailedEmailFields {
+
+  def contribution(email: String, identityUserId: IdentityUserId): EmailFields =
+    failedEmailFields("contribution-failed", email, identityUserId)
```
johnduffell, line 6: I took email and identityUserId as parameters and passed directly on to failedEmailFields.  Although this could be simpler as a partial application, I don't think this code is complex enough to benefit from such an advanced technique.  Like this it's simpler and more compact than the old multi file class based approach.

support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
```diff
@@ -2,106 +2,81 @@ package com.gu.emailservices
...
+    promotion: Option[Promotion],
+    billingPeriod: BillingPeriod,
+    user: User,
+    currency: Currency,
+    paymentSchedule: PaymentSchedule
+  ): Seq[(String, String)] = {
+    List(
+      "Subscription term" -> billingPeriod.noun,
+      "Payment amount" -> SubscriptionEmailFieldHelpers.formatPrice(SubscriptionEmailFieldHelpers.firstPayment(paymentSchedule).amount),
+      "Address 1" -> "", //TODO: Remove this from Braze template
```
rupertbates, line 70: Do we actually need to send it at all? I think Braze is quite tolerant of missing fields.

support-workers/src/main/scala/com/gu/emailservices/GuardianWeeklyEmailFields.scala
```diff
@@ -1,56 +1,41 @@
...
-
-      override def payload: String = super.payload(user.primaryEmailAddress, "guardian-weekly")
-
-      override def userId: Either[SfContactId, IdentityUserId] = Left(sfContactId)
-    }
-
+object GuardianWeeklyEmailFields {
+  def build(
+    subscriptionEmailFields: SubscriptionEmailFields,
+    fulfilmentOptions: FulfilmentOptions,
```
rupertbates, line 14: looks like `fulfilmentOptions` is unused

support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
johnduffell: maybe not but that code is just moved rather than edited - changing the actual fields sent is outside the scope of this PR

support-workers/src/main/scala/com/gu/emailservices/GuardianWeeklyEmailFields.scala
johnduffell: yes very well spotted! I noticed that too as it was highlighted in grey in IntelliJ, but decided to keep the change out of this PR

johnduffell:
thanks for simplification suggestions, I decided to fix them up in this PR, otherwise I will never do it!  And I am going to test everything anyway so it makes sense to do so.

johnduffell:
tested in CODE - standard DS, GW and contributions emails all worked fine.

PR #2660, opened 2020-08-07
Digital Subscription gift checkout preparation work

rupertbates:
## Why are you doing this?
This is some preparatory work for the Digital Subscription checkout server side work

## Changes

* Add an integration test for creating a DS gift sub
* Extract product specific builders to separate objects because the ProductSubscriptionBuilders class was getting very big and hard to scan even before I add in the code related to DS gifting purchase and redemption

[**Trello Card**](https://trello.com/c/TcVj29Ji/3175-gifting-checkout-spike-server-side)


support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/GuardianWeeklySubscriptionBuilder.scala
```diff
@@ -0,0 +1,56 @@
...
+    country: Country,
+    maybePromoCode: Option[PromoCode],
+    firstDeliveryDate: Option[LocalDate],
+    promotionService: PromotionService,
+    readerType: ReaderType,
+    environment: TouchPointEnvironment,
+    contractEffectiveDate: LocalDate = LocalDate.now(DateTimeZone.UTC)
+  ): Either[PromoError, SubscriptionData] = {
+
+    val contractAcceptanceDate = Try(firstDeliveryDate.get) match {
```
johnduffell, line 31: not your doing, but `Try(option.get)` is an interesting one!

PR #2662, opened 2020-08-11
Digital Subscription gift checkout - server side work

rupertbates:
## Why are you doing this?

This is the server side work to support purchasing and redeeming Digital Subscription gifts.

Digital Subscription gifts are the first product we have which requires two separate checkouts, one to purchase the gift and one to redeem it. The table below shows the types of checkout which are required by the different reader types (Direct, Gift or Corporate)

### Mapping of ReaderTypes to checkout types for Digital Subscription
|          | Direct | Gift | Corporate |
|----------|--------|------|-----------|
| Purchase |    ✔    |   ✔  (gifter) |           |
| Redemption   |        |   ✔  (giftee) |     ✔      |

Most of this PR focuses on changes to the `DigitalSubscriptionBuilder` class to handle the two different checkout types, there are now two separate methods `buildPurchase` and `buildRedemption` to make the branching in the logic clearer. 

I have converted the ADT `RedemptionData` to a plain case class as the two usages we envisioned for it are actually identical and so don't require subclassing.

## Changes

* Remove separate subclass for CorporateRedemption - gifting can share the RedemptionData subclass
* Separate out build methods on DigitalSubscriptionBuilder into those which deal with redemption and those which deal with purchases
* Add code to DigitalSubscriptionBuilder to handle purchase of gift subscriptions
* Add placeholder to DigitalSubscriptionBuilder for handling redemption of gift subscriptions

[**Trello Card**](https://trello.com/c/TcVj29Ji/3175-gifting-checkout-spike-server-side)

### TODO
- [ ] Implement DS giftee redemption
- [ ] More tests

### DIFF
--- a/support-services/src/main/scala/com/gu/support/redemption/GetCodeStatus.scala
+++ b/support-services/src/main/scala/com/gu/support/redemption/GetCodeStatus.scala
@@ -13,6 +13,7 @@ object GetCodeStatus {
   sealed abstract class RedemptionInvalid(val clientCode: String)
   case object NoSuchCode extends RedemptionInvalid("NO_SUCH_CODE")
   case object CodeAlreadyUsed extends RedemptionInvalid("CODE_ALREADY_USED")
+  case object InvalidReaderType extends RedemptionInvalid("INVALID_READER_TYPE")
 
   private def statusFromDynamoAttr(attrs: Map[String, DynamoLookup.DynamoValue]): Either[String, RedemptionTable.AvailableField] =
     for {
--- a/support-services/src/main/scala/com/gu/support/redemption/generator/GiftCodeGenerator.scala
+++ b/support-services/src/main/scala/com/gu/support/redemption/generator/GiftCodeGenerator.scala
@@ -28,11 +28,6 @@ object GiftDuration {
 
   case object Gift12Month extends GiftDuration("12")
 
-  def fromBillingPeriod = (billingPeriod: BillingPeriod) => billingPeriod match {
-    case Quarterly => Gift3Month
-    case Annual => Gift12Month
-  }
-
 }
 
 object CodeBuilder {
--- a/support-services/src/main/scala/com/gu/support/redemption/generator/GiftCodeGenerator.scala
+++ b/support-services/src/main/scala/com/gu/support/redemption/generator/GiftCodeGenerator.scala
@@ -3,6 +3,7 @@ package com.gu.support.redemption.generator
 import java.security.SecureRandom
 
 import com.gu.support.redemption.generator.CodeBuilder.GenerateGiftCode
+import com.gu.support.workers.{Annual, BillingPeriod, Quarterly}
 
 object GiftCodeGenerator {
 
@@ -26,6 +27,11 @@ object GiftDuration {
   case object Gift6Month extends GiftDuration("06")
 
   case object Gift12Month extends GiftDuration("12")
+
+  def fromBillingPeriod = (billingPeriod: BillingPeriod) => billingPeriod match {
+    case Quarterly => Gift3Month
+    case Annual => Gift12Month
+  }
 
 }
 
--- a/support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionBuilder.scala
+++ b/support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionBuilder.scala
@@ -10,10 +10,11 @@ import com.gu.support.config.{TouchPointEnvironment, ZuoraDigitalPackConfig}
 import com.gu.support.promotions.{PromoCode, PromoError, PromotionService}
 import com.gu.support.redemption.GetCodeStatus
 import com.gu.support.redemption.GetCodeStatus.{InvalidReaderType, RedemptionInvalid}
+import com.gu.support.redemption.generator.GiftDuration.{Gift12Month, Gift3Month}
 import com.gu.support.redemption.generator.{GiftCodeGenerator, GiftDuration}
 import com.gu.support.redemptions.{RedemptionCode, RedemptionData}
 import com.gu.support.workers.ProductTypeRatePlans._
-import com.gu.support.workers.{BillingPeriod, DigitalPack}
+import com.gu.support.workers.{Annual, BillingPeriod, DigitalPack, Quarterly}
 import com.gu.support.zuora.api.ReaderType.{Corporate, Direct, Gift}
 import com.gu.support.zuora.api.{ReaderType, SubscriptionData}
 import com.gu.zuora.subscriptionBuilders.ProductSubscriptionBuilders.{applyPromoCode, buildProductSubscription, validateRatePlan}
@@ -88,9 +89,14 @@ object DigitalSubscriptionBuilder {
     EitherT.fromEither[Future](withPromoApplied.left.map(Left.apply))
   }
 
+  def fromBillingPeriod(billingPeriod: BillingPeriod): GiftDuration = billingPeriod match {
+    case Annual => Gift12Month
+    case _ => Gift3Month
+  }
+
   def generateRedemptionCode(readerType: ReaderType, billingPeriod: BillingPeriod, subscriptionData: SubscriptionData) = {
     if (readerType == Gift) {
-      val code = GiftCodeGenerator.randomGiftCodes.next().withDuration(GiftDuration.fromBillingPeriod(billingPeriod))
+      val code = GiftCodeGenerator.randomGiftCodes.next().withDuration(fromBillingPeriod(billingPeriod))
       subscriptionData.copy(
         subscription = subscriptionData.subscription.copy(redemptionCode = Some(code.value))
       )
--- a/support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionBuilder.scala
+++ b/support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionBuilder.scala
@@ -12,6 +12,7 @@ import com.gu.support.redemption.GetCodeStatus
 import com.gu.support.redemptions.{CorporateRedemption, RedemptionData}
 import com.gu.support.workers.DigitalPack
 import com.gu.support.workers.ProductTypeRatePlans._
+import com.gu.support.zuora.api.ReaderType.Direct
 import com.gu.support.zuora.api.{ReaderType, Subscription, SubscriptionData}
 import com.gu.zuora.subscriptionBuilders.ProductSubscriptionBuilders.{applyPromoCode, buildProductSubscription, validateRatePlan}
 import org.joda.time.LocalDate
@@ -41,7 +42,10 @@ object DigitalSubscriptionBuilder {
     readerType: ReaderType,
     purchase: SubscriptionPurchase
   )(implicit ec: ExecutionContext): EitherT[Future, Either[PromoError, GetCodeStatus.RedemptionInvalid], SubscriptionData] = {
-    val delay = purchase.config.defaultFreeTrialPeriod + purchase.config.paymentGracePeriod
+    val delay = if(readerType == Direct)
+      purchase.config.defaultFreeTrialPeriod + purchase.config.paymentGracePeriod
+    else 0 // Gift purchases don't have a free trial period
+
     val contractAcceptanceDate = contractEffectiveDate.plusDays(delay)
 
     val subscriptionData = buildProductSubscription(
--- a/support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionBuilder.scala
+++ b/support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionBuilder.scala
@@ -89,14 +89,14 @@ object DigitalSubscriptionBuilder {
     EitherT.fromEither[Future](withPromoApplied.left.map(Left.apply))
   }
 
-  def fromBillingPeriod(billingPeriod: BillingPeriod): GiftDuration = billingPeriod match {
+  def toDuration(billingPeriod: BillingPeriod): GiftDuration = billingPeriod match {
     case Annual => Gift12Month
     case _ => Gift3Month
   }
 
   def addRedemptionCodeIfGift(readerType: ReaderType, billingPeriod: BillingPeriod, subscriptionData: SubscriptionData) = {
     if (readerType == Gift) {
-      val code = GiftCodeGenerator.randomGiftCodes.next().withDuration(fromBillingPeriod(billingPeriod))
+      val code = GiftCodeGenerator.randomGiftCodes.next().withDuration(toDuration(billingPeriod))
       subscriptionData.copy(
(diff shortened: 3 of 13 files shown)


johnduffell:
great description, especially the table/naming convention

support-services/src/main/scala/com/gu/support/redemption/generator/GiftCodeGenerator.scala
```diff
@@ -27,6 +28,11 @@ object GiftDuration {
 
   case object Gift12Month extends GiftDuration("12")
 
+  def fromBillingPeriod = (billingPeriod: BillingPeriod) => billingPeriod match {
```
johnduffell, line 31: two points:
in terms of dependencies, at the moment this generator is completely standalone.  The code to map billing periods is more how we want to interface with the generator based on what the workers input is so is not really generation code.  I would prefer to see this function in the subscription builder, as that is the code responsible for making a generated code from whatever input we already have.

Secondly I think this could just be a normal method:
```suggestion
  def fromBillingPeriod(billingPeriod: BillingPeriod): GiftDuration = billingPeriod match {
```

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionBuilder.scala
```diff
@@ -5,90 +5,166 @@ import java.util.UUID
...
 
-case class SubscriptionPaymentCorporate(
+case class SubscriptionRedemption(
   redemptionData: RedemptionData,
   getCodeStatus: GetCodeStatus
 ) extends SubscriptionPaymentType
 
 object DigitalSubscriptionBuilder {
 
+  type BuildResult = EitherT[Future, Either[PromoError, RedemptionInvalid], SubscriptionData]
```
johnduffell, line 41: 👍 good idea

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionBuilder.scala
```diff
@@ -5,90 +5,166 @@ import java.util.UUID
...
+    productRatePlanId: ProductRatePlanId,
+    requestId: UUID,
+    readerType: ReaderType,
+    purchase: SubscriptionPurchase
+  )(implicit ec: ExecutionContext): BuildResult = {
+
+    val (contractAcceptanceDelay, autoRenew, initialTerm) = if (readerType == Gift)
+      (0, false, purchase.billingPeriod.monthsInPeriod)
+    else
+      (purchase.config.defaultFreeTrialPeriod + purchase.config.paymentGracePeriod, true, 12)
```
johnduffell, line 71: clever idea using a tuple in this way, this might make it easier to follow
```suggestion
    val (autoRenew, initialTerm, contractAcceptanceDelay) =
      if (readerType == Gift)
        (false, purchase.billingPeriod.monthsInPeriod, 0)
      else
        (true, 12, purchase.config.defaultFreeTrialPeriod + purchase.config.paymentGracePeriod)
```

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionBuilder.scala
```diff
@@ -5,90 +5,166 @@ import java.util.UUID
...
-      case SubscriptionPaymentDirect(_, maybePromoCode, country, promotionService) =>
-        EitherT.fromEither[Future](
-          applyPromoCode(promotionService, maybePromoCode, country, productRatePlanId, subscriptionData)
-            .left.map(Left.apply)
-        )
-      case SubscriptionPaymentCorporate(redemptionData, getCodeStatus) =>
-        withRedemption(subscriptionData.subscription, redemptionData, getCodeStatus)
-          .map(subscription => subscriptionData.copy(subscription = subscription))
-          .leftMap(Right.apply)
+    val withRedemptionCode = generateRedemptionCode(readerType, purchase.billingPeriod, subscriptionData)
```
johnduffell, line 85: How about `addRedemptionCodeIfGift` (and `applyPromoIfAvailable`)?
I'm already not totally happy with the existing naming, this and the applyPromoCode when I've done previous work on this code.
It would be ok if it was `maybePromoCode.map(applyPromoCode(...))` and if the `if (readerType == Gift)` was outside.  But both of these are actually conditional within the logic.
`generateRedemptionCode` is not just conditionally getting the code generated but it's actually adding it to the subscription data.

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionBuilder.scala
```diff
@@ -5,90 +5,166 @@ import java.util.UUID
...
+    today: LocalDate,
+    productRatePlanId: ProductRatePlanId,
+    requestId: UUID,
+    redemptionCode: RedemptionCode,
+  )(implicit ec: ExecutionContext): BuildResult = {
+    // TODO: RB implement this
+    val errorType: Either[PromoError, RedemptionInvalid] = Right(InvalidReaderType)
+    EitherT.leftT(errorType)
   }
 }
```
johnduffell, line 170: probably worth splitting this file/object further into redemption and purchase related stuff.  This would let us have a more clear grouping about which behaviours apply to each situation and which are common to both.

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionBuilder.scala
```diff
@@ -5,90 +5,166 @@ import java.util.UUID
...
+
+    readerType match {
+      case Corporate => buildCorporateRedemption(
+        today,
+        productRatePlanId,
+        requestId,
+        redemption.redemptionData.redemptionCode,
+        redemption.getCodeStatus
+      )
+      case Gift => buildGiftRedemption(
```
johnduffell, line 118: I have some vague nervousness that we are using Gift in the state to both mean buy a gift and redeem a gift, and it's also a field on the subscription itself in zuora.  Everywhere before, readertype just went directly from the step function into zuora, but now it is conditional on a separate field (payment details) as to what it really means.
I think this is complicated by the fact that corporate redemption actually creates a zuora sub, wheras gift redemption just modifies a previously created one. 

Perhaps this just mostly inherent complexity due to the new offerings and the way we store in zuora, but even if we don't think the state model needs changing, we could still e.g. rename `Gift` to `GifterOrGiftee` or similar to make gotchas less likely in future.  Any thoughts?

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/ProductSubscriptionBuilders.scala
```diff
@@ -42,12 +42,10 @@ object ProductSubscriptionBuilders {
...
     readerType: ReaderType,
-    initialTermMonths: Int = 12
+    autoRenew: Boolean = true,
+    initialTerm: Int = 12,
+    initialTermPeriodType: PeriodType = Month
   ): SubscriptionData = {
-    val (initialTerm, autoRenew, initialTermPeriodType) = if (readerType == ReaderType.Gift)
-      (initialTermInDays(contractEffectiveDate, contractAcceptanceDate, initialTermMonths), false, Day)
-    else
-      (12, true, Month)
```
johnduffell, line 50: oh now I realised where the tuple came from 😂

support-workers/src/test/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionBuilderSpec.scala
```diff
@@ -86,9 +102,19 @@ class DigitalSubscriptionBuilderSpec extends AsyncFlatSpec with Matchers {
...
+
+  lazy val threeMonthGiftPurchase = DigitalSubscriptionBuilder.build(
+    DigitalPack(GBP, Quarterly, Gift),
+    UUID.fromString("f7651338-5d94-4f57-85fd-262030de9ad5"),
+    SubscriptionPurchase(ZuoraDigitalPackConfig(14, 2), None, Quarterly, Country.UK, promotionService),
     SANDBOX,
     () => saleDate
   ).value.map(_.right.get)
 
+  //TODO: RB test all cases
```
johnduffell, line 118: GW gifting is a good one to check that we have the tests too, as I'd imagine that's not a super common code path.

johnduffell (approved):
great work, just added a few comments and questions but nothing that I consider essential 👍

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionBuilder.scala
rupertbates: Maybe we should replace `Gift` with `Gifter` and `Giftee`? That might be an issue for downstream issues (primarily reporting) because we've been using `Gift` for GW gifters but it would be more accurate.

rupertbates: Systems that could be affected:
- manage
- renewal functionality on subscribe
- DL reporting

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionBuilder.scala
johnduffell: I realised later that this bit of code was moved from lower down the PR! but feel free to make changes to make it clearer as you see fit

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionBuilder.scala
```diff
@@ -82,15 +83,20 @@ object DigitalSubscriptionBuilder {
...
-    val withRedemptionCode = generateRedemptionCode(readerType, purchase.billingPeriod, subscriptionData)
-    val withPromoApplied = applyPromoCode(purchase.promotionService, purchase.maybePromoCode, purchase.country, productRatePlanId, withRedemptionCode)
+    val withRedemptionCode = addRedemptionCodeIfGift(readerType, purchase.billingPeriod, subscriptionData)
+    val withPromoApplied = applyPromoCodeIfPresent(purchase.promotionService, purchase.maybePromoCode, purchase.country, productRatePlanId, withRedemptionCode)
 
     EitherT.fromEither[Future](withPromoApplied.left.map(Left.apply))
   }
 
-  def generateRedemptionCode(readerType: ReaderType, billingPeriod: BillingPeriod, subscriptionData: SubscriptionData) = {
+  def fromBillingPeriod(billingPeriod: BillingPeriod): GiftDuration = billingPeriod match {
```
johnduffell, line 92: flip the name now it's here
```suggestion
  def toGiftDuration(billingPeriod: BillingPeriod): GiftDuration = billingPeriod match {
```

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionBuilder.scala
```diff
@@ -82,15 +83,20 @@ object DigitalSubscriptionBuilder {
...
+    case Annual => Gift12Month
+    case _ => Gift3Month
```
johnduffell, line 94: in relality this is a partial function
annual->12month
quarter->3month
other->impossible presumably enforced by the validators in support-frontend backend

If this is changed or broken, we might silently create incorrect gifts which we may only realise when people who were expecting (6 month? recurring?) gifts find they are ending after only 3 months.  I wouldn't look forward to that repair operation.

Ideally we would only be able to represent valid gift durations in the state model ADT, meaning it wouldn't even compile if this assumption was broken or removed at the other end.

Failing that I think we should return an error e.g. Option at this point to at least fail before getting bad data into the zuora db.

rupertbates: The thing is that all this really has nothing to do with the sub we are creating, it is just to generate the 'gd12' or 'gd03' part of the code. This was the reason why I was a bit wary of having information in the codes in the first place, because it is increasing the code complexity for something which is at best cosmetic and at worst misleading.

johnduffell: oh yeah good point forgot about that - of course the original billing period object is the one that deals with zuora
In that case I'm less concerned as the user will still get what was paid for even if the gift-code isn't helpful to us.

johnduffell: we could have a conversation about the code format - the code wasn't in my mind cosmetic, actually it was to make it easier for userhelp etc to understand what the code is without having to go into the system to look things up.  But I agree the default assumption seemed to be that the code has meaning in it, which I guess wasn't questioned enough in your opinion.

rupertbates: I think it's marginal, I can see that there could be some value from a userhelp point of view but on the other hand it's another source of bugs so by default I would have gone for a YAGNI approach.

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionBuilder.scala
rupertbates: I'll address this more in the next PR

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionBuilder.scala
rupertbates: After discussion we are going to leave this for now

support-services/src/main/scala/com/gu/support/redemption/GetCodeStatus.scala
```diff
@@ -13,6 +13,7 @@ object GetCodeStatus {
   sealed abstract class RedemptionInvalid(val clientCode: String)
   case object NoSuchCode extends RedemptionInvalid("NO_SUCH_CODE")
   case object CodeAlreadyUsed extends RedemptionInvalid("CODE_ALREADY_USED")
+  case object InvalidReaderType extends RedemptionInvalid("INVALID_READER_TYPE")
```
johnduffell, line 16: @rupertbates 
this change is causing a match error
```
[warn] /Users/john_duffell/code/support-frontend/support-frontend/app/controllers/RedemptionController.scala:230:56: match may not be exhaustive.
[warn] It would fail on the following input: InvalidReaderType
[warn]       _ <- EitherT(getCodeStatus(codeToCheck)).leftMap {
[warn]                  
```
on this line
https://github.com/guardian/support-frontend/blob/69996239214edcc7a42f6c8fb8b14e3ada0725f2/support-frontend/app/controllers/RedemptionController.scala#L230

PR #2672, opened 2020-08-18
Switch on subscription cards for test

lucymonie:
## Why are you doing this?
In preparation for a production end-to-end test of the digital subscriptions card for print (aka iMovo), this is a PR to update the variable `useDigitalVoucher`, so it's true. This means that we can test the digital subscriptions card purchase process is working correctly. 

[**Trello Card**](https://trello.com/c/RwCVUP46/3194-preparation-work-for-imovo-test)

I have deployed this to `CODE` and it works correctly except that the thank you page isn't merged in yet (see: https://github.com/guardian/support-frontend/pull/2622) and the email (I think @johnduffell is working on this).


johnduffell:
the email should be working.  If not there must be a bug somewhere
It was shipped last month
https://github.com/guardian/support-frontend/pull/2618
and
https://github.com/guardian/membership-workflow/pull/266

PR #2293, opened 2019-12-17
Make support-models depend on released version of support-internation…

tomrf1:
Currently if you release support-models it tries to release support-internationalisation as well.
Instead, make support-models depend on the released version of support-internationalisation

PR #2673, opened 2020-08-19
Move payment-api to the mono-repo

tomrf1:
## Why are you doing this?
`payment-api` is the backend for single contributions.

The backend for recurring contributions (support-workers) already lives in the mono-repo.
It is convenient for `payment-api` to be moved because it depends on projects in the mono-repo, and currently we have to publish them to maven so that payment-api can use them.

I spent a lot of time checking for secrets in the payment-api repo and its history and established that there are none.

### Changes:
This PR contains the entire commit history of the repo.
For reviewing purposes, only start with the ["make it compile" commit](https://github.com/guardian/support-frontend/commit/6d2370307f0a72ae39027f52a186170e42d4f8c1). diff here: https://github.com/guardian/support-frontend/pull/2673/files/8cd03204a6913c9ff606ad48d15adabe1e4b9894..a1299ccd95ac97e5f6f76e7d61084d22161f940b

- directory `payment-api` becomes `support-payment-api`
- riffraff project name `support:payment-api` becomes `support:payment-api-mono`
- `support-models` and `support-internationalisation` are no longer remote dependencies
- updated scalatest + mockito dependencies to align with the mono-repo. I'm not a fan of using mockito here but that's for another day

### Follow up tasks:
- remove sbt-release for `support-models` and `support-internationalisation`?
- remove [this hack](https://github.com/guardian/support-frontend/pull/2293)
- leave the private payment-api repo in place, as a convenient record?


johnduffell:
have/will you  checked the auto scheduled deploys and also added a restriction to the old riffraff project?

tomrf1:
@johnduffell thank you for the reminder!
I've just done the following for support:payment-api:
- disabled continuous deployment 
- added a restriction
- disabled scheduled deploys

Once this is merged I'll set it up for the new project name

support-payment-api/build.sbt
```diff
@@ -61,25 +59,7 @@ libraryDependencies ++= Seq(
...
-    testOptions in TeamCityTest += Tests.Argument("-l", "tags.RequiresMembershipCredentials"),
-    buildInfoKeys := Seq[BuildInfoKey](
-      name,
-      BuildInfoKey.constant("gitCommitId", Option(System.getenv("BUILD_VCS_NUMBER"))
-        .getOrElse(git.gitHeadCommit.value.getOrElse("Unknown Head Commit")))
-    ),
-    buildInfoPackage := "app",
-    buildInfoOptions += BuildInfoOption.ToMap
-  )
+enablePlugins(SystemdPlugin, PlayService, RoutesCompiler, RiffRaffArtifact, JDebPackaging, BuildInfoPlugin)
```
johnduffell, line 62: I think in the other projects, the enable plugins happens in the root build.sbt.  I am not sure which is better, but I thought I'd flag it in case you have an opinion.

johnduffell (approved):
nice, mostly seems to be test changes due to the modulatisation of scalatest, thanks for doing this 👍

johnduffell:
you might also want to make the payment api a required status in the settings->braches-> restrictions in the github repo.

johnduffell:
also make sure snyk is checking it.  I think it should pick up automatically when it runs snyk in the support-frontend project, as you have added this as an aggregated project, but just check that.

support-payment-api/build.sbt
rupertbates: It looks like you are enabling/disabling some in the root and some here. It might be better to be consistent (do it all in the root?)

rupertbates (approved):
Looks good, just one suggestion about enabling/disabling plugins

👏 👏 👏   for taking this on!

support-payment-api/build.sbt
tomrf1: 👍 I've moved it to the root build.sbt

support-payment-api/.prout.json
```diff
@@ -0,0 +1,5 @@
+{
+  "checkpoints": {
+    "PROD": { "url": "https://payment.guardianapis.com/healthcheck", "overdue": "15M" }
```
rtyley, line 3: @Jpsi this will interest you - there was a small unanticipated issue caused by moving this file into the mono-repo - there are now two Prout checkpoints called `PROD` in the https://github.com/guardian/support-frontend monorepo - the one newly moved in above, and the one already present here:

https://github.com/guardian/support-frontend/blob/291a4caf73108bdc06442e5ef7bb02a48fd6c25e/.prout.json#L2-L4

...so Prout doesn't accept there being two checkpoints with the same name in the same GitHub repo - which kind of makes sense because it uses those checkpoint names in the GitHub labels it generates, and it's ambiguous if there are two different `PROD`s.

Unfortunately Prout just throws an exception for this particular error if you attempt to look at the diagnostic for this:

https://prout-bot.herokuapp.com/view/guardian/support-frontend

image

The exception -not visible to the user, but [visible in eg Sentry](https://sentry.io/organizations/the-guardian/issues/1872374324/?project=49913) - is the only helpful message saying what's gone wrong:

image

Prout could be improved to report this problem better - but in the meantime, if the checkpoint here is renamed from `PROD` to something like `PROD-payment-api` then the error will resolve itself.

Jpsi: @rtyley thank you, that makes sense! (Also, thank you for showing how to diagnose it, I wasn't aware of Sentry and it makes it very clear 👍)

PR #2674, opened 2020-08-19
Hide supporter count on Aus map

tomrf1:
## Why are you doing this?
The Aus campaign is over and we are no longer updating the supporter count value.
But we do still want to link to this map.

#### Before
Screen Shot 2020-08-19 at 11 43 41

#### After
Screen Shot 2020-08-19 at 11 43 53


michaelbjacobson (approved):
👍 👍 👍

PR #2675, opened 2020-08-19
Set options prop on Stripe Elements

tomrf1:
## Why are you doing this?
card https://trello.com/c/EyeJ64OO/2226-react-stripe-elements-console-warning
When the user first clicks on a Stripe Elements field, the following warning message appears in the console:

```breadcrumbs.js:58 Unsupported prop change on Elements: You cannot change the `options` prop after setting the `stripe` prop.```

It doesn't cause any problems.

The solution is to always pass the `options` prop to `Elements`, otherwise you hit [this warning](https://github.com/stripe/react-stripe-js/blob/e13fae556b1a667a5dfa1130c00d5ec4c2bc154e/src/components/Elements.tsx#L125).
All fields in the `options` type are optional: https://stripe.com/docs/js/elements_object/create


paulbrown1982 (approved):
👍

PR #2676, opened 2020-08-19
Fix support-internationalisation publishing

mario-galic:
## Why are you doing this?

support-service-lambdas requires Scala 2.13 build

## Changes

* It was syntactically broken and missing sbt plugins for publishing
* update to latest release process
* The release was successful https://repo.maven.apache.org/maven2/com/gu/support-internationalisation_2.13/

### DIFF
--- a/support-internationalisation/build.sbt
+++ b/support-internationalisation/build.sbt
@@ -1,6 +1,7 @@
 import sbtrelease.ReleaseStateTransformations._
 
 name := "support-internationalisation"
+organization := "com.gu"
 
 crossScalaVersions := Seq("2.11.8", "2.12.7", "2.13.3")
 
--- a/support-internationalisation/build.sbt
+++ b/support-internationalisation/build.sbt
@@ -2,10 +2,14 @@ import sbtrelease.ReleaseStateTransformations._
 
 name := "support-internationalisation"
 
-crossScalaVersions := Seq("2.11.8", "2.12.7")
+crossScalaVersions := Seq("2.11.8", "2.12.7", "2.13.3")
 
 description := "Scala library to provide internationalisation classes to Guardian Membership/Subscriptions/support projects."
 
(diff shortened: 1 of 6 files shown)


support-internationalisation/build.sbt
```diff
@@ -1,11 +1,16 @@
...
 
 name := "support-internationalisation"
+organization := "com.gu"
 
-crossScalaVersions := Seq("2.11.8", "2.12.7")
+crossScalaVersions := Seq("2.11.8", "2.12.7", "2.13.3")
 
 description := "Scala library to provide internationalisation classes to Guardian Membership/Subscriptions/support projects."
 
+libraryDependencies += "org.scalatest" %% "scalatest" % "3.1.1" % "test"
```
johnduffell, line 10: there is already 3.2.1 on the classpath so it shouldn't need 3.1.1 as well

support-internationalisation/build.sbt
```diff
@@ -1,11 +1,16 @@
 import sbtrelease.ReleaseStateTransformations._
 
 name := "support-internationalisation"
+organization := "com.gu"
```
johnduffell, line 4: this should have been set at the root level under ThisBuild so it should appear everywhere.

johnduffell:
I think you have tried to run sbt in the support-internationalisation directory, but actually that's just part of a multi module build.  You need to run sbt at the root and then type `project support-internationalisation ` to get to the right module.

mario-galic:
It does not seem to work for me via `project support-internationalisation`. I'll go ahead and close this. Note there is a successful 0.13 release https://repo.maven.apache.org/maven2/com/gu/support-internationalisation_2.13/0.13/ so we'll have to remember to bump `version.sbt`.

johnduffell:
Maybe it is broken for some reason but that is the way the project is structured now

> we'll have to remember to bump version.sbt.

if you have released 0.13 then go ahead and bump it, happy to +1 that.

PR #2681, opened 2020-08-25
Apply box-sizing reset

i-hardy:
## Why are you doing this?

Currently the CSS reset for support-frontend (in assets/stylesheets/skeleton/reset-src.scss) does not globally set the `box-sizing` rule to `border-box`.

Without this reset, the width and height of elements are calculated without taking into account the size of any padding or border applied, which can create a lot of knock-on visual bugs and difficult-to-reason-about CSS. This MDN page illustrates the different box-sizing properties well: https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing

This also makes it difficult to share components from Source with DCR and other products that do have the reset in place, as we may see odd visual differences.

Applying the reset resulted in some copy moving slightly due to reduced space, and a few visual issues documented on the Trello card. After a lot of wrangling I decided to put the old `box-sizing: content-box;` rule back in place for the header, as it currently has a pretty complex styling setup managed partially in Javascript which would need to be substantially overhauled, which I don't think is worth the time it would require.

[**Trello Card**](https://trello.com/c/0TFkyIGB)

## Changes

* Applied a global reset for the `box-sizing` rule to `border-box`
* Fixed a resulting visual bug on the 'Why support' showcase page

## Screenshots

**Fixes for documented issues**

Screenshot 2020-08-25 at 16 48 28

Screenshot 2020-08-25 at 16 48 44

Screenshot 2020-08-25 at 16 49 11


support-frontend/yarn.lock
```diff
@@ -3085,6 +3085,18 @@
     telejson "^3.2.0"
     util-deprecate "^1.0.2"
 
+"@stripe/react-stripe-js@^1.1.2":
```
rupertbates, line 3088: Is this yarn.lock update needed for this PR? It doesn't look like any dependencies have changed.

rupertbates (approved):
👍  Good work, just one question about the yarn.lock

support-frontend/yarn.lock
i-hardy: I had to `yarn install` when I switched back to main as the react-stripe dependency had been added by another PR. I'd assumed it was just some minor dependency bumps but it looks like this dependency wasn't in the lockfile at all before, which is weird as it seems to have been added months ago: https://github.com/guardian/support-frontend/blame/main/support-frontend/package.json#L87 🤔

rupertbates: That is odd, maybe the lock file hasn't been updated since it was added? 😕 
I'm happy to leave this change in though.

PR #2686, opened 2020-08-28
Fix for safari recaptcha bug (recurring contributions)

tomrf1:
- safari only
- recurring only

The bug:
1. select card
2. select paypal. It gets stuck on card.

It is caused by recaptcha being loaded twice.
It uses some state called `calledRecaptchaRender` to ensure this doesn't happen, and this works in other browsers but not safari for some reason.

This fix is not ideal, and we should try to understand it better.
But it does at least work, by ensuring the recaptcha div is empty before trying to load it.

### DIFF
--- a/support-frontend/assets/pages/contributions-landing/components/StripeCardForm/StripeCardForm.jsx
+++ b/support-frontend/assets/pages/contributions-landing/components/StripeCardForm/StripeCardForm.jsx
@@ -184,7 +184,7 @@ const CardForm = (props: PropTypes) => {
     }
   };
 
-  const recaptchaElementEmpty = (): boolean => {
+  const recaptchaElementNotEmpty = (): boolean => {
     const el = document.getElementById('robot_checkbox');
     if (el) {
       return el.children.length > 0;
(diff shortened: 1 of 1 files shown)


TomPretty (approved):
Would be good to get to the bottom of why this is happening! Good quick fix though! 👍

support-frontend/assets/pages/contributions-landing/components/StripeCardForm/StripeCardForm.jsx
```diff
@@ -184,9 +184,22 @@ const CardForm = (props: PropTypes) => {
...
+    if (el) {
+      return el.children.length > 0;
+    }
+    return true;
+  };
+
   // Creates a new setupIntent upon recaptcha verification
   const setupRecurringRecaptchaCallback = () => {
     setCalledRecaptchaRender(true);
+    // Fix for safari, where the calledRecaptchaRender state handling does not work. TODO - find a better solution
```
paulbrown1982, line 198: In the old days we'd inspect the element's offsetHeight to know if it was visible or not...

PR #2693, opened 2020-09-07
Digital Subscription Gift Redemption

rupertbates:
## Why are you doing this?

This PR introduces the mechanism for redeeming a Digital Subscription gift code.

A full description of how this works is [here](https://docs.google.com/document/d/1tey3Px090ZaBwwOOvlCK6lbPl7MW_3MCJjqYer3NSeY/edit?pli=1#) but to summarise: 
- We find the subscription which was created when the gifter purchased the gift
- We update that subscription to set the `gifteeIdentityId` custom field to the Identity id of the giftee, and to set the `currentTerm` length to the length of gift purchased + the time elapsed since purchase. We also have to change the `currentTermPeriodType` to `Day` from `Month` so that this term is accurate to the nearest day.

### TODO
- [ ] Code expiry
- [ ] Salesforce record

## Changes
* Gift redemption code
* Change gift code suffix length to 8 characters
* Add integration tests for code redemption
* Start using [mockito-scala](https://github.com/mockito/mockito-scala) instead of regular Mockito as it handles various Scala features better

 - [ ] [PR experience rated](https://forms.gle/N6FsTGG8JQFGV4Ha9)


support-models/src/main/scala/com/gu/support/zuora/api/response/QueryResponse.scala
```diff
@@ -19,10 +27,22 @@ object PaymentMethodRecord {
   implicit val codec: Codec[PaymentMethodRecord] = deriveCodec
 }
 
+object SubscriptionRedemptionFields {
+  implicit val decoder: Decoder[SubscriptionRedemptionFields] = decapitalizingDecoder[SubscriptionRedemptionFields]
+    .prepare(_.withFocus(_.mapObject(_
+      .renameField("RedemptionCode__c", "RedemptionCode")
+      .renameField("GifteeIdentityId__c", "GifteeIdentityId")
```
johnduffell, line 34: presume this does work ok with the upper case `GifteeIdentityId` rather than `gifteeIdentityId` as in the case class?  but should it match?

johnduffell: nm just realised it's a decapitalizingDecoder

support-models/src/main/scala/com/gu/support/zuora/api/response/Responses.scala
```diff
@@ -200,3 +200,9 @@ object PreviewSubscribeResponse {
 }
 
 case class PreviewSubscribeResponse(invoiceData: List[InvoiceDataItem], success: Boolean) extends ZuoraResponse
+
+object UpdateRedemptionDataResponse {
+  implicit val codec: Codec[UpdateRedemptionDataResponse] = deriveCodec
```
johnduffell, line 205: if its a response it would be a decoder not a codec I think

support-services/src/test/scala/com/gu/support/redemption/generator/GiftCodeGeneratorSpec.scala
```diff
@@ -7,13 +7,13 @@ class GiftCodeGeneratorSpec extends AnyFlatSpec with Matchers {
...
     val giftCode = GiftCodeGenerator.fromInts(Iterator.continually(0)).next.withDuration(GiftDuration.Gift3Month)
-    giftCode.value should be("gd03-000000")
+    giftCode.value should be("gd03-00000000")
   }
 
   it should "always produce different codes with the built in random" in {
     val generateGiftCode = GiftCodeGenerator.randomGiftCodes
     val giftCodes = generateGiftCode.map(_.withDuration(GiftDuration.Gift3Month))
-    val numberToCheck = 10000
+    val numberToCheck = 100000
```
johnduffell, line 16: is it a lot slower? I was thinking when I wrote it even 10000 was probably excessive, even 100 is probably plenty to catch the obvious errors

support-workers/build.sbt
```diff
@@ -27,7 +27,8 @@ libraryDependencies ++= Seq(
   "com.fasterxml.jackson.core" % "jackson-databind" % jacksonVersion,
   "com.fasterxml.jackson.core" % "jackson-annotations" % jacksonVersion,
   "com.fasterxml.jackson.dataformat" % "jackson-dataformat-cbor" % jacksonVersion,
-  "org.mockito" % "mockito-core" % "3.0.0" % "it,test",
+  "org.mockito" % "mockito-scala_2.12" % "1.15.0" % "it,test",
+  "org.mockito" % "mockito-scala-scalatest_2.12" % "1.15.0" % "it,test",
```
johnduffell, line 31:
```suggestion
  "org.mockito" %% "mockito-scala" % "1.15.0" % "it,test",
  "org.mockito" %% "mockito-scala-scalatest" % "1.15.0" % "it,test",
```

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -233,3 +250,81 @@ object CreateZuoraSubscription {
     }
 
 }
+
+object DigitalSubscriptionGiftRedemption {
+
+  def ifDigitalSubscriptionGiftRedemption(product: ProductType, paymentMethod: Either[PaymentMethod, RedemptionData]) = {
```
johnduffell, line 256:
```suggestion
  def maybeDigitalSubscriptionGiftRedemption(product: ProductType, paymentMethod: Either[PaymentMethod, RedemptionData]) = {
```

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -44,36 +51,44 @@ class CreateZuoraSubscription(servicesProvider: ServiceProvider = ServiceProvide
...
+    val giftCodeGenerator = services.giftCodeGenerator
     val isTestUser = state.user.isTestUser
     val config: ZuoraConfig = services.config.zuoraConfigProvider.get(isTestUser)
-    CreateZuoraSubscription(state, requestInfo, now, today, promotionService, redemptionService, zuoraService, config)
+
+    ifDigitalSubscriptionGiftRedemption(state.product, state.paymentMethod).map(redemptionData =>
+      redeemGift(redemptionData, state.user.id, requestInfo, state, zuoraService, services.catalogService)
+    ).getOrElse(
+      createSubscription(state, requestInfo, now, today, promotionService, redemptionService, zuoraService, giftCodeGenerator, config)
+    )
```
johnduffell, line 62: I know it's less high level, but pattern match can be clearer when dealing with both sides of option or either - consider:
```suggestion
    maybeDigitalSubscriptionGiftRedemption(state.product, state.paymentMethod) match {
      case Some(redemptionData) =>
        redeemGift(redemptionData, state.user.id, requestInfo, state, zuoraService, services.catalogService)
      case None =>
        createSubscription(state, requestInfo, now, today, promotionService, redemptionService, zuoraService, giftCodeGenerator, config)
    }
```

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -233,3 +250,81 @@ object CreateZuoraSubscription {
...
+  ): Future[HandlerResult[SendThankYouEmailState]] = {
+    for {
+      giftSubscriptionFields <- zuoraService.getSubscriptionFromRedemptionCode(redemptionData.redemptionCode)
+      validatedSubscription <- Future.fromTry(validateRedemptionData(giftSubscriptionFields))
+      fullGiftSubscription <- zuoraService.getSubscriptionById(validatedSubscription.id)
+      newTermLength <- Future.fromTry(calculateNewTermLength(fullGiftSubscription, catalogService))
+      updateDataResponse <- zuoraService.updateSubscriptionRedemptionData(validatedSubscription.id, gifteeUserId, newTermLength)
+      handlerResult <- Future.fromTry(buildHandlerResult(updateDataResponse, state, redemptionData, requestInfo))
+    } yield handlerResult
+  }
```
johnduffell, line 279: v. minor: we don't need the extra braces

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -233,3 +250,81 @@ object CreateZuoraSubscription {
...
+    state: CreateZuoraSubscriptionState,
+    zuoraService: ZuoraService,
+    catalogService: CatalogService
+  ): Future[HandlerResult[SendThankYouEmailState]] = {
+    for {
+      giftSubscriptionFields <- zuoraService.getSubscriptionFromRedemptionCode(redemptionData.redemptionCode)
+      validatedSubscription <- Future.fromTry(validateRedemptionData(giftSubscriptionFields))
+      fullGiftSubscription <- zuoraService.getSubscriptionById(validatedSubscription.id)
+      newTermLength <- Future.fromTry(calculateNewTermLength(fullGiftSubscription, catalogService))
+      updateDataResponse <- zuoraService.updateSubscriptionRedemptionData(validatedSubscription.id, gifteeUserId, newTermLength)
```
johnduffell, line 276: is there a better name for `updateDataResponse` or is it literally just in case something strange is returned we want to save it?

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -233,3 +250,81 @@ object CreateZuoraSubscription {
...
+  }
+
+  def validateRedemptionData(existingSub: SubscriptionRedemptionQueryResponse) = {
+    existingSub.records.headOption.map(
+      existingSubFields =>
+        if (existingSubFields.gifteeIdentityId.isEmpty)
+          Success(existingSubFields)
+        else
+          Failure(new RuntimeException(GetCodeStatus.CodeAlreadyUsed.clientCode))
+    ).getOrElse(Failure(new RuntimeException(GetCodeStatus.NoSuchCode.clientCode)))
```
johnduffell, line 288: minor but not sure we need to use the client code here, as it's not needed to be machine readable, this will just be in the support workers failure.  Just toStringing the status object should be okay and clear enough to humans.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -233,3 +250,81 @@ object CreateZuoraSubscription {
...
+    gifteeUserId: String,
+    requestInfo: RequestInfo,
+    state: CreateZuoraSubscriptionState,
+    zuoraService: ZuoraService,
+    catalogService: CatalogService
+  ): Future[HandlerResult[SendThankYouEmailState]] = {
+    for {
+      giftSubscriptionFields <- zuoraService.getSubscriptionFromRedemptionCode(redemptionData.redemptionCode)
+      validatedSubscription <- Future.fromTry(validateRedemptionData(giftSubscriptionFields))
+      fullGiftSubscription <- zuoraService.getSubscriptionById(validatedSubscription.id)
```
johnduffell, line 274: is this second call only needed to get the rate plan so that we can find out the billing period?

support-workers/src/main/scala/com/gu/zuora/ZuoraService.scala
```diff
@@ -88,6 +89,22 @@ class ZuoraService(val config: ZuoraConfig, client: FutureHttpClient, baseUrl: O
     } yield ddId).value
   }
 
+  def getSubscriptionFromRedemptionCode(redemptionCode: RedemptionCode): Future[SubscriptionRedemptionQueryResponse] = {
+    val queryData = QueryData(
+      //TODO RB expire codes
+      s"select id, RedemptionCode__c, GifteeIdentityId__c from subscription where RedemptionCode__c = '${redemptionCode.value}' and status = 'Active'"
```
johnduffell, line 95: generally it would be risky to insert some user input into a query directly, but since RedemptionCode is validated when it's wrapped I think we're ok as no quotes are possible.

johnduffell: do we need `RedemptionCode__c ` at all in the select?

support-workers/src/test/scala/com/gu/support/workers/CreateZuoraSubscriptionStepsSpec.scala
```diff
@@ -139,15 +143,16 @@ class CreateZuoraSubscriptionStepsSpec extends AsyncFlatSpec with Matchers {
       }
     }
 
-    val result = CreateZuoraSubscription(
+    val result = CreateZuoraSubscription.createSubscription(
```
johnduffell, line 146: definitely more consistent now to have the proper named function

support-workers/src/test/scala/com/gu/support/workers/integration/CreateZuoraSubscriptionSpec.scala
```diff
@@ -50,51 +52,59 @@ class CreateZuoraSubscriptionSpec extends AsyncLambdaSpec with MockServicesCreat
...
+  def testCreateSubscription(json: String) =
+    createSubscription(json)
+      .map(
+        _._1.subscriptionNumber.length should be > 1
+      )
+}
+
+trait CreateZuoraSubscriptionBaseSpec extends AsyncLambdaSpec with MockServicesCreator with MockContext {
+
+  def createSubscription(
```
johnduffell, line 98: I agree in modularising this better, but why put these in a trait rather than an object?  if in an object then it is self contained/can't depend on anything in the subclass unless it's passed in.  trait doesn't have that guarantee.

support-workers/src/test/scala/com/gu/support/workers/integration/GiftSubcriptionZuoraSpec.scala
```diff
@@ -0,0 +1,64 @@
...
+import com.gu.support.workers.lambdas.{DigitalSubscriptionGiftRedemption, HandlerResult}
+import com.gu.support.workers.states.{CreateZuoraSubscriptionState, SendThankYouEmailState}
+import com.gu.test.tags.annotations.IntegrationTest
+import org.mockito.ArgumentMatchers.any
+import io.circe.parser.decode
+
+import scala.concurrent.Future
+
+@IntegrationTest
+class GiftSubcriptionZuoraSpec extends CreateZuoraSubscriptionBaseSpec {
```
johnduffell, line 19:
```suggestion
class GiftSubscriptionZuoraSpec extends CreateZuoraSubscriptionBaseSpec {
```

support-workers/src/test/scala/com/gu/support/workers/integration/GiftSubcriptionZuoraSpec.scala
```diff
@@ -0,0 +1,64 @@
...
+    redeemSubscription(giftCode.value)
+      .map(_.value.user.id shouldBe idId)
+
+  it should "throw a CodeAlreadyUsed exception redeem a Digital Pack gift subscription" in
+    recoverToExceptionIf[RuntimeException](
+      redeemSubscription(giftCode.value)
+    ).map(_.getMessage shouldBe CodeAlreadyUsed.clientCode)
+
+
+  it should "throw a NoSuchCode exception redeem a Digital Pack gift subscription" in {
```
johnduffell, line 43: for even better checking, move this test to the top, and use the generated code.  Then we have proved the selected code does not exist before we create it.

support-workers/src/test/scala/com/gu/zuora/ZuoraITSpec.scala
```diff
@@ -39,6 +40,24 @@ class ZuoraITSpec extends AsyncFlatSpec with Matchers {
...
+  it should "handle invalid redemption codes" in {
+    val invalidRedemptionCode = "xxxx-0000"
+    uatService.getSubscriptionFromRedemptionCode(RedemptionCode(invalidRedemptionCode).right.get).map {
+      response =>
+        response.records.size shouldBe 0
+    }
+  }
+
   it should "be resistant to 'ZOQL injection'" in {
     // try https://github.com/guardian/support-service-lambdas/blob/master/lib/zuora/src/main/scala/com/gu/util/zuora/SafeQueryBuilder.scala
```
johnduffell, line 62: unrelated, but that branch has been renamed now
```suggestion
    // try https://github.com/guardian/support-service-lambdas/blob/main/lib/zuora/src/main/scala/com/gu/util/zuora/SafeQueryBuilder.scala
```

support-workers/src/test/scala/com/gu/zuora/ZuoraITSpec.scala
```diff
@@ -39,6 +40,24 @@ class ZuoraITSpec extends AsyncFlatSpec with Matchers {
...
+    uatService.getSubscriptionFromRedemptionCode(RedemptionCode(redemptionCode).right.get).map {
+      response =>
+        response.records.size shouldBe 1
+        response.records.head.gifteeIdentityId shouldBe None
+        response.records.head.redemptionCode shouldBe Some(redemptionCode)
+    }
+  }
+
+  it should "handle invalid redemption codes" in {
+    val invalidRedemptionCode = "xxxx-0000"
```
johnduffell, line 54: I'm surprised that redemption code passes the validation to get into the RedemptionCode type?  if so it shouldn't

support-workers/src/test/scala/com/gu/zuora/ZuoraITSpec.scala
```diff
@@ -39,6 +40,24 @@ class ZuoraITSpec extends AsyncFlatSpec with Matchers {
     }
   }
 
+  it should "retrieve subscription redemption information from a redemption code" in {
+    val redemptionCode = "gd12-fnc4ox"
```
johnduffell, line 44: might be worth calling it gd12-autotest so that people don't try to use (or make it obvious some other way)

support-workers/src/test/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionBuilderSpec.scala
```diff
@@ -96,6 +105,7 @@ class DigitalSubscriptionBuilderSpec extends AsyncFlatSpec with Matchers {
         )))
       })),
     SANDBOX,
+    giftCodeGeneratorService,
```
johnduffell, line 108: funny that we need a gift code generator for a corporate sub, would be better if the types were narrower

support-workers/src/test/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionBuilderSpec.scala
```diff
@@ -112,9 +123,23 @@ class DigitalSubscriptionBuilderSpec extends AsyncFlatSpec with Matchers {
...
+  lazy val threeMonthGiftRedemption: Future[Either[PromoError, GetCodeStatus.RedemptionInvalid]] = DigitalSubscriptionBuilder.build(
+    DigitalPack(GBP, Quarterly, Gift),
+    UUID.fromString("f7651338-5d94-4f57-85fd-262030de9ad5"),
+    SubscriptionRedemption(RedemptionData(RedemptionCode("any-code").right.get),
+      new GetCodeStatus({
+        case "CODE" => Future.successful(Some(Map(
+          "available" -> DynamoLookup.DynamoBoolean(true),
+          "corporateId" -> DynamoLookup.DynamoString("1")
+        )))
+      })),
```
johnduffell, line 139: I'm struggling to understand the types a bit here, is the indentation 2 too many here? I think we expect an error and we've got an EitherT, and the error itself is an Either, so we just want to get that.  And we only have Future because we might have a promo code involved.
I feel like the test itself is not clear as some of the expectation is here, can we just move this code into the test method?

johnduffell (commented):
looks great, I think the overall structure is easy to follow.  I have made some comments and some nit picks but in the context there is nothing substantial.  I have perhaps not been totally clear in all of them so feel free to ask for more if it's not clear!

support-models/src/main/scala/com/gu/support/zuora/api/response/Responses.scala
rupertbates: A codec has both an encoder and decoder in it. It's convenient to have both in case you want to use an encoder in a test for instance.

support-services/src/test/scala/com/gu/support/redemption/generator/GiftCodeGeneratorSpec.scala
rupertbates: Around 897ms - seems ok to me?

support-workers/src/test/scala/com/gu/support/workers/integration/GiftSubcriptionZuoraSpec.scala
rupertbates: That is a good idea. It will only work if we generate a new code every test run rather than using the same one every time but I think that's fine.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
rupertbates: I'm checking the message in an IT test so it's nice to have a val to compare it to

support-workers/src/test/scala/com/gu/support/workers/integration/CreateZuoraSubscriptionSpec.scala
rupertbates: I had them in an Object originally but it caused at least one test to hang indefinitely - I think it was to do with the way that Scalatest uses its own execution context for async tests. Converting to a trait solved this.

support-workers/src/main/scala/com/gu/zuora/ZuoraService.scala
rupertbates: Re the injection point, also worth noting that this is Zoql not SQL so it's a lot more constrained in what it can do.

On whether we need `RedemptionCode__c` in the select, we do check it in an integration test although we don't use it anywhere else. I think I'll leave it in unless you feel strongly that it should be removed.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
rupertbates: Yes

support-workers/src/test/scala/com/gu/zuora/ZuoraITSpec.scala
rupertbates: It does, are you thinking of the GiftCode validation? It wouldn't pass that.

rupertbates: It does pass, this is a `RedemptionCode` which has different validation to a `GiftCode`. This does make me wonder if this could be an issue though - that we have two types with different validation which both go into the same custom field in Zuora.

support-models/src/main/scala/com/gu/support/zuora/api/response/Responses.scala
johnduffell: sure I would generally say YAGNI and add it (in the test code) if you do later. but not a major point.

support-services/src/test/scala/com/gu/support/redemption/generator/GiftCodeGeneratorSpec.scala
johnduffell: 1 second for such an unlikely test to fail seems disproportionate, unit tests are generally very quick. I would recommend reducing it as it's actually just designed to find major breakages e.g. total loss of randomness.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
johnduffell: I think it's just success false/true and a possible error message, so happy to resolve this one

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
johnduffell: fair enough 👍

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
johnduffell: thought so, fair enough.
Is there anything else we need to do to ensure idempodency?  should we be updating the CreatedInRequestId custom field when we redeem, and then checking it before rejecting as already redeemed?

johnduffell: I'm thinking of the case when zuora is V slow but succeeding, so the step function times out and retries, but actually is succeding slowly.

support-workers/src/main/scala/com/gu/zuora/ZuoraService.scala
johnduffell: if it's not used then take it out as dependencies between us and zuora data are more expensive than compile checked ones within the code.  It's easier to remove it now when it's fresh than later when it's been in PROD a while.

support-workers/src/test/scala/com/gu/support/workers/integration/CreateZuoraSubscriptionSpec.scala
johnduffell: if it really has some state that breaks things then consider make it its own class rather than extending, that's still simpler/self contained than extending.

We can revisit, but we did have in the [coding standards](https://docs.google.com/document/d/1vRIYAU2OnRkHzf4dLCuzxiA-3HsKh1PDmkiLc-1GmfQ/edit#) to avoid inheritance and overriding, in favour of normal functions.  But I think if the hanging problem is impossible, creating a standalone class would be a better middle ground as there's not much unexpected a composed class can do.

support-workers/src/test/scala/com/gu/support/workers/integration/GiftSubcriptionZuoraSpec.scala
johnduffell: oh yeah didn't spot that, thanks for updating it 👍

support-workers/src/test/scala/com/gu/zuora/ZuoraITSpec.scala
johnduffell: good point, yes we have overloaded the code field based on the value of ReaderType (or the first digit being 'g' or 'c').  This was alreaddy the intention even when matt spreadsheet was proposing the code format to be very consistent, but now we have diverged them further.
I would expect any code reading the subscription to interpret this field in conjunction with ReaderType, and for analysis in the DL they could do the same, or use the first letter.

There can't be the case where a sub is both corporate and gift, so the field would not be used at the same time.  Think of it as an `Option[Either[GiftCode,CorporateCode]]`
I can definitely see that it adds complexity to have a different data model at rest (zuora) from in transit (step functions) and we generally try to keep them as similar as possible.  This is also related to the ReaderType discussion we had as now the step function execution and zuora sub don't have a 1:1 mapping any more.

support-workers/src/main/scala/com/gu/zuora/ZuoraService.scala
johnduffell: and yes Zoql is much more constrained so the risk is definitely much lower, and DS are not a big target, also it would be difficult to extract any PII via any issue with this query.
I only mentioned because this is the critical point where a vuln would be caused, when data is merged into a higher level structure, and we are relying on validation that happens elsewhere rather than proper escaping at the point of constructing the query.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
rupertbates: That is a good point, I'll investigate 👍

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -263,32 +276,48 @@ object DigitalSubscriptionGiftRedemption {
...
-      newTermLength <- Future.fromTry(calculateNewTermLength(fullGiftSubscription, catalogService))
-      updateDataResponse <- zuoraService.updateSubscriptionRedemptionData(validatedSubscription.id, gifteeUserId, newTermLength)
-      handlerResult <- Future.fromTry(buildHandlerResult(updateDataResponse, state, redemptionData, requestInfo))
-    } yield handlerResult
-
-  def validateRedemptionData(existingSub: SubscriptionRedemptionQueryResponse) = {
+    zuoraService.getSubscriptionFromRedemptionCode(redemptionData.redemptionCode).flatMap(
+      redemptionQueryResponse =>
+        getSubscriptionState(redemptionQueryResponse, state.requestId.toString) match {
+          case u: Unredeemed => doRedemption(u.subscriptionId, state, redemptionData, requestInfo, zuoraService, catalogService)
```
johnduffell, line 287:
```suggestion
          case Unredeemed(subscriptionId) => doRedemption(subscriptionId, state, redemptionData, requestInfo, zuoraService, catalogService)
```

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -263,32 +276,48 @@ object DigitalSubscriptionGiftRedemption {
...
+          case RedeemedInThisRequest => Future.fromTry(buildHandlerResult(UpdateRedemptionDataResponse(true), state, redemptionData, requestInfo))
+          case Redeemed => Future.fromTry(Failure(new RuntimeException(GetCodeStatus.CodeAlreadyUsed.clientCode)))
+          case NotFound => Future.fromTry(Failure(new RuntimeException(GetCodeStatus.NoSuchCode.clientCode)))
```
johnduffell, line 290:
```suggestion
          case Redeemed => Future.failed(new RuntimeException(GetCodeStatus.CodeAlreadyUsed.clientCode))
          case NotFound => Future.failed(new RuntimeException(GetCodeStatus.NoSuchCode.clientCode))
```

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -233,3 +251,109 @@ object CreateZuoraSubscription {
...
+
+  private def getSubscriptionState(existingSub: SubscriptionRedemptionQueryResponse, requestId: String) = {
+    existingSub.records.headOption.map(
+      existingSubFields =>
+        if (existingSubFields.gifteeIdentityId.isEmpty)
+          Unredeemed(existingSubFields.id)
+        else if (existingSubFields.createdRequestId == requestId)
+          RedeemedInThisRequest
+        else Redeemed
+    ).getOrElse(NotFound)
```
johnduffell, line 302: I'd slightly prefer a pattern match here, for a runtime assertion of one result but it also helps with clarity a bit 
```suggestion
    existingSub.records match {
      case existingSubFields :: Nil if (existingSubFields.gifteeIdentityId.isEmpty) =>
        Unredeemed(existingSubFields.id)
      case existingSubFields :: Nil if (existingSubFields.createdRequestId == requestId) =>
        RedeemedInThisRequest
      case existingSubFields :: Nil =>
        Redeemed
      case _ =>
        NotFound
    }
```

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -263,32 +276,48 @@ object DigitalSubscriptionGiftRedemption {
...
-    ).getOrElse(Failure(new RuntimeException(GetCodeStatus.NoSuchCode.clientCode)))
+          Unredeemed(existingSubFields.id)
+        else if (existingSubFields.createdRequestId == requestId)
+          RedeemedInThisRequest
+        else Redeemed
+    ).getOrElse(NotFound)
   }
 
-  def calculateNewTermLength(subscription: Subscription, catalogService: CatalogService) = {
+  private def doRedemption(
```
johnduffell, line 305: we have a redeemGift and a doRedemption, which don't have any meaningful distinction in the name.
The difference in the function is that the first one does checks first, wheras the second one just goes the actual steps.
Perhaps we could call them tryToRedeem and redeemInZuora

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -301,7 +330,7 @@ object DigitalSubscriptionGiftRedemption {
     } yield Success(newTermLength)).getOrElse(Failure(new RuntimeException(s"Unable to calculate new term length for subscription ${subscription}")))
```
johnduffell, line 330: Just something to bear in mind rather than a request to change - it's great to standardise on Try/Future to reduce boilerplate in the for comprehensions, but I can't help thinking all the "new RuntimeException" is just boilerplate elsewhere.  If we were implementing this function in isolation we would just use `Either[String, Int]`

support-workers/src/test/scala/com/gu/support/workers/integration/GiftSubscriptionZuoraSpec.scala
```diff
@@ -18,43 +20,48 @@ import scala.concurrent.Future
...
+    redeemSubscription(giftCode.value, requestId)
+      .map(_.value.user.id shouldBe idId)
+
+  it should "return a successful redemption result if a subscription has already been redeemed in the current request" in
+    redeemSubscription(giftCode.value, requestId)
       .map(_.value.user.id shouldBe idId)
 
-  it should "throw a CodeAlreadyUsed exception redeem a Digital Pack gift subscription" in
+  it should "throw a CodeAlreadyUsed exception if a subscription has been redeemed in a previous request" in {
+    val previousRequestId = UUID.randomUUID()
```
johnduffell, line 53: that would be `subsequentRequestId`

support-models/src/main/scala/com/gu/support/zuora/api/SubscribeRequest.scala
```diff
@@ -25,3 +28,14 @@ object SubscribeRequest {
...
+
+import com.gu.support.encoding.Codec._
+
+object UpdateRedemptionDataRequest {
+  implicit val encoder: Encoder[UpdateRedemptionDataRequest] = deriveEncoder[UpdateRedemptionDataRequest].mapJsonObject(
+    _.renameField("gifteeIdentityId", "GifteeIdentityId__c")
+  )
+}
+
+case class UpdateRedemptionDataRequest(gifteeIdentityId: String, currentTerm: Int, currentTermPeriodType: PeriodType)
```
johnduffell, line 40: we need to add the CreatedInRequestId__c field to this and update it.

if you have the patience, could be worth renaming the field to something like ProcessedInRequestId__c (create a new one in zuora, and delete the old one) as they are only needed while a step function exection is in progress AND zuora is having issues.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
rupertbates: Maybe, although it does have to be converted to an exception at some point so it might as well be here

support-models/src/main/scala/com/gu/support/zuora/api/SubscribeRequest.scala
rupertbates: 🤦  Yes! I've updated the tests to catch this properly now as well

support-workers/src/test/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionBuilderSpec.scala
rupertbates: This is the same for all the other tests in this spec, they are all built up in the same way. I think it might be clearer if we did change them to move these definitions into the test itself but I don't want to do them all in this PR so I'm going to leave this as is.

support-workers/src/test/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionBuilderSpec.scala
rupertbates: The crazy world of Digital Subs!

support-workers/src/test/scala/com/gu/support/workers/integration/CreateZuoraSubscriptionSpec.scala
rupertbates: 👍

support-workers/src/test/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionBuilderSpec.scala
johnduffell: fair enough, leave it for another day 👍

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
johnduffell: fair enough - this was the thing that prompted me to think about the balance keeping the modules self contained e.g. GiftCodeGenerator vs keeping them in the overall structure i.e. avoiding loads of conversions in the for comprehensions

johnduffell (approved):
thanks, looks like everything is addressed so if you're happy I am! 👏 
nice work on this, it's not a simple bit of work to get all that connected up and I think the result is very clear despite the additional complexity needed.
Probably worth final testing in CODE at least to make sure the existing is still working then 👍

rupertbates:
> thanks, looks like everything is addressed so if you're happy I am! 👏
> nice work on this, it's not a simple bit of work to get all that connected up and I think the result is very clear despite the additional complexity needed.
> Probably worth final testing in CODE at least to make sure the existing is still working then 👍

Tested in CODE and all working fine.

PR #2697, opened 2020-09-10
Log Json if there is a parsing failure in IdapiTestUserRequest

rupertbates:
## Why are you doing this?

A tiny PR to log the Json received from IDAPI if there is a parsing failure in the IdapiTestUserRequest class

### DIFF
--- a/support-frontend/test/selenium/util/IdapiTestUserRequest.scala
+++ b/support-frontend/test/selenium/util/IdapiTestUserRequest.scala
@@ -1,17 +1,14 @@
 package selenium.util
 
-import okhttp3.{OkHttpClient, Request, RequestBody, MediaType, Response}
-import io.circe.generic.semiauto.deriveDecoder
-import io.circe.{Json, Decoder, DecodingFailure}
-import io.circe.parser.parse
-
-import org.slf4j.LoggerFactory
-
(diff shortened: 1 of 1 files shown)


support-frontend/test/selenium/util/IdapiTestUserRequest.scala
```diff
@@ -1,17 +1,14 @@
...
-
-import org.slf4j.LoggerFactory
-
-import cats.syntax.either._
+import io.circe.{Decoder, Json}
+import okhttp3._
 
 class IdapiTestUserRequest {
 
-  private def getResult(): Either[String, Json] = {
```
Jpsi, line 14: Is this because of [this convention](https://www.oreilly.com/library/view/learning-scala/9781449368814/ch04.html#:~:text=Functions%20with%20Side%20Effects%20Should,modifies%20data%20outside%20its%20scope)? (I didn't know about it, it seems like a useful convention)

Jpsi (approved):
Quite useful, thank you!

support-frontend/test/selenium/util/IdapiTestUserRequest.scala
rupertbates: Yes

PR #2699, opened 2020-09-11
Remove default payment ab test

TomPretty:
## Why are you doing this?
Remove default payment test and roll variant out as the default. Also fixes a bug where the paymentMethod was being set to "None" after a one-off contribution with paypal (as it leaves then redirects back to the site).


tomrf1 (approved):
👍

PR #2703, opened 2020-09-14
Remove variant 1 from retention test

tomrf1:
It's performing poorly


TomPretty (approved):
💯 nice! Hopefully we'll stop loosing money now!

PR #2705, opened 2020-09-14
Digital Subscription gift code expiry

rupertbates:
## Why are you doing this?

We want Digital Subscription gift codes to expire if they haven't been redeemed within one year. This PR does that

## Changes
* Rename `SubscriptionState` to `SubscriptionRedemptionState`
* introduces a new `Expired` `SubscriptionRedemptionState`
* Update the length of the initial term for gifts to 13 months so that it is longer than the expiry period - this will get set to the correct length upon redemption

### DIFF
--- a/support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
+++ b/support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
@@ -254,6 +254,8 @@ object CreateZuoraSubscription {
 
 object DigitalSubscriptionGiftRedemption {
 
+  val expirationTimeInMonths = 12
+
   sealed abstract class SubscriptionRedemptionState(val clientCode: String)
 
   object Unredeemed {
@@ -298,7 +300,7 @@ object DigitalSubscriptionGiftRedemption {
 
   def getSubscriptionState(existingSub: SubscriptionRedemptionQueryResponse, requestId: String) =
     existingSub.records match {
-      case existingSubFields :: Nil if existingSubFields.contractEffectiveDate.plusYears(1).isBefore(LocalDate.now()) =>
+      case existingSubFields :: Nil if existingSubFields.contractEffectiveDate.plusMonths(expirationTimeInMonths).isBefore(LocalDate.now()) =>
         Expired
       case existingSubFields :: Nil if existingSubFields.gifteeIdentityId.isEmpty =>
         Unredeemed(existingSubFields.id)
--- a/support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
+++ b/support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
@@ -254,18 +254,24 @@ object CreateZuoraSubscription {
 
 object DigitalSubscriptionGiftRedemption {
 
-  sealed trait SubscriptionState
+  sealed abstract class SubscriptionState(val clientCode: String)
 
-  case class Unredeemed(subscriptionId: String) extends SubscriptionState
+  object Unredeemed {
+    val clientCode = "unredeemed"
+  }
+
+  case class Unredeemed(subscriptionId: String) extends SubscriptionState(Unredeemed.clientCode)
 
   // This can happen if Zuora is responding very slowly - a redemption request may succeed but not return a response
   // until after the CreateZuoraSubscription lambda has timed out meaning that the redemption will be retried with the
   // same requestId. In this case we want the lambda to succeed so that we progress to the next lambda
-  case object RedeemedInThisRequest extends SubscriptionState
+  case object RedeemedInThisRequest extends SubscriptionState("redeemed_in_this_request")
 
-  case object Redeemed extends SubscriptionState
+  case object Redeemed extends SubscriptionState("redeemed")
 
-  case object NotFound extends SubscriptionState
+  case object Expired extends SubscriptionState("expired")
+
+  case object NotFound extends SubscriptionState("not_found")
 
   def maybeDigitalSubscriptionGiftRedemption(product: ProductType, paymentMethod: Either[PaymentMethod, RedemptionData]) = {
     product match {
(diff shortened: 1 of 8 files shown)


support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -254,18 +254,24 @@ object CreateZuoraSubscription {
 
 object DigitalSubscriptionGiftRedemption {
 
-  sealed trait SubscriptionState
+  sealed abstract class SubscriptionRedemptionState(val clientCode: String)
 
-  case class Unredeemed(subscriptionId: String) extends SubscriptionState
+  object Unredeemed {
+    val clientCode = "unredeemed"
```
johnduffell, line 260: can this just be inlined like the others for simplicity?
If we change the value, presumably it would be relied on independently by the client, so we would want the test to check the actual value (or could just check the type is an `Unredeemed` if that is all we care about)

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -286,13 +292,14 @@ object DigitalSubscriptionGiftRedemption {
...
-          case Redeemed => Future.failed(new RuntimeException(GetCodeStatus.CodeAlreadyUsed.clientCode))
-          case NotFound => Future.failed(new RuntimeException(GetCodeStatus.NoSuchCode.clientCode))
+          case otherState: SubscriptionRedemptionState => Future.failed(new RuntimeException(otherState.clientCode))
         }
     )
 
-  private def getSubscriptionState(existingSub: SubscriptionRedemptionQueryResponse, requestId: String) =
+  def getSubscriptionState(existingSub: SubscriptionRedemptionQueryResponse, requestId: String) =
     existingSub.records match {
+      case existingSubFields :: Nil if existingSubFields.contractEffectiveDate.plusYears(1).isBefore(LocalDate.now()) =>
```
johnduffell, line 301: to aid testability, I would prefer to pass today's date in, keep the non ref transparent stuff at the outside.  I realise it would add to the function signatures so it depends if you think that's a valid use of boilerplate though?

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionBuilder.scala
```diff
@@ -70,7 +70,7 @@ object DigitalSubscriptionBuilder {
   )(implicit ec: ExecutionContext): BuildResult = {
 
     val (contractAcceptanceDelay, autoRenew, initialTerm) = if (readerType == Gift)
-      (0, false, purchase.billingPeriod.monthsInPeriod)
+      (0, false, 13) // Expiry for gift redemption codes is 12 months so set the initial term to 13 to allow for this
```
johnduffell, line 73: will we remember to change this, if later we decide to extend the expiry?
Also wondering if this will affect any reporting or revenue spreading that we do (preumsbaly it will already have to take gifts into account specially)

support-workers/src/test/scala/com/gu/support/workers/DigitalSubscriptionGiftRedemptionSpec.scala
```diff
@@ -0,0 +1,75 @@
...
+      ))
+    val state = DigitalSubscriptionGiftRedemption.getSubscriptionState(response, "123")
+    state.clientCode shouldBe RedeemedInThisRequest.clientCode
+  }
+
+  it should "identify an expired subscription code from a SubscriptionRedemptionQueryResponse" in {
+    val response = SubscriptionRedemptionQueryResponse(
+      List(SubscriptionRedemptionFields(
+        id = "1",
+        contractEffectiveDate = new LocalDate(1999, 12, 31),
```
johnduffell, line 68: we should probably do boundary testing here rather than the extreme - e.g. make sure -365 days works but 366 doesn't, or similar.

support-workers/src/test/scala/com/gu/support/workers/DigitalSubscriptionGiftRedemptionSpec.scala
```diff
@@ -0,0 +1,75 @@
...
+        "done": true,
+        "size": 1
+    }
+    """
+
+    testDecoding[SubscriptionRedemptionQueryResponse](
+      jsonResponse,
+      subResponse => subResponse.records.head.contractEffectiveDate.getDayOfMonth shouldBe 9
+    )
+  }
```
johnduffell, line 30: not a major thing but does this belong in a different test class? This one is mostly testing `getSubscriptionState` which is nice and clear seeing all the different variations on their own.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
rupertbates: I'd rather not unless we specifically need to for testing purposes as it just adds to the cognitive load when reading through the code

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
rupertbates: It couldn't be accessed through `Unredeemed.clientCode` then in the way that we can access `Redeemed.clientCode` because `Unredeemed` is a case class not a case object

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionBuilder.scala
rupertbates: I've updated this so that it is explicitly dependent on the expiry time so that if we change that then this will update automatically.
Yes reporting will have to be updated anyway.

support-workers/src/test/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionBuilderSpec.scala
```diff
@@ -78,7 +79,7 @@ class DigitalSubscriptionBuilderSpec extends AsyncFlatSpec with Matchers {
       readerType shouldBe Gift
       redemptionCode.isDefined shouldBe true
       redemptionCode.get.substring(0, 4) shouldBe "gd03"
-      initialTerm shouldBe 13
+      initialTerm shouldBe DigitalSubscriptionGiftRedemption.expirationTimeInMonths + 1
```
johnduffell, line 82: I never know how much we should "share" between the tests and code. as the tests are supposed to be independent and the next level of simplicity!  But in this case I'm sure it won't do any harm either way.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
johnduffell: fair enough, I know what you mean 👍 

> as it just adds to the cognitive load when reading through the code

although that is the referential transparency/FP simplicity debate in a nutshell!

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionBuilder.scala
johnduffell: 👍

johnduffell (approved):
looks great 👍 thanks

PR #2706, opened 2020-09-15
Add `enableDigitalSubGifting` into Switches

rupertbates:
## Why are you doing this?
To add the `enableDigitalSubGifting` switch into the Switches object

Follows on from the admin console PR https://github.com/guardian/support-admin-console/pull/157

### DIFF
--- a/support-frontend/app/admin/settings/Switches.scala
+++ b/support-frontend/app/admin/settings/Switches.scala
@@ -6,6 +6,7 @@ import com.gu.support.encoding.Codec.deriveCodec
 case class Switches(
   oneOffPaymentMethods: PaymentMethodsSwitch,
   recurringPaymentMethods: PaymentMethodsSwitch,
+  enableDigitalSubGifting: SwitchState,
   useDotcomContactPage: Option[SwitchState],
   enableRecaptchaBackend: SwitchState,
   enableRecaptchaFrontend: SwitchState,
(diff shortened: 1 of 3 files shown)


support-frontend/app/admin/settings/Switches.scala
```diff
@@ -6,6 +6,7 @@ import com.gu.support.encoding.Codec.deriveCodec
 case class Switches(
   oneOffPaymentMethods: PaymentMethodsSwitch,
   recurringPaymentMethods: PaymentMethodsSwitch,
+  enableDigitalSubGifting: SwitchState,
```
johnduffell, line 9: what happens if the value is not there? or is it guaranteed to be there as long as the SAC PR is running in PROD?

I am just thinking about default values, in my mind from a time when the dotcom switches file was corrupted, and many things went back to the "safe state" which wasn't so safe in all cases.

rupertbates: It is definitely there in the file as you can see by looking at the console: https://support.gutools.co.uk/switches

If it weren't however we'd get a decoding error from Circe when we tried to load the switches. It's debatable whether we should have some sort of fall back state here rather than just allowing it to crash.

lucymonie (approved):

support-frontend/app/admin/settings/Switches.scala
johnduffell: yep I can see the item in the PROD json https://s3.console.aws.amazon.com/s3/object/support-admin-console/PROD/switches.json?region=eu-west-1&tab=overview already
`  "enableDigitalSubGifting": "Off",`

Agree it's not something for this PR, but something to be aware of as we add more critical switches, whtehr we should have some kind of swimlaning between them.

johnduffell (approved):

PR #2710, opened 2020-09-16
run IT tests hourly instead of daily

johnduffell:
## Why are you doing this?

We reduced the frequency of running IT tests because SF DEV was filling up quickly and eventually tests were failing  and it was blocking SX team.  Now that we have a PR to clean it up, we can bring it back to higher frequency. PR here to be merged also: https://github.com/guardian/support-service-lambdas/pull/700

This PR is also reverted in this change set, as these are the tests which were failing: https://github.com/guardian/support-frontend/pull/2636

The changes mean we are likely to find out about breakages within the working day, rather than a subsequent day.  This means we can either fix the tests or find potential issues quickly.

[**Trello Card**](https://trello.com/c/LKwe8ApH/3241-auto-clean-it-test-data-from-sf-dev)

### DIFF
--- a/support-lambdas/it-test-runner/cfn.yaml
+++ b/support-lambdas/it-test-runner/cfn.yaml
@@ -40,7 +40,7 @@ Resources:
         RerunTests:
           Type: Schedule
           Properties:
-            Schedule: 'rate(1 day)'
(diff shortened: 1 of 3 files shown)


kelvin-chappell (approved):
👍

PR #2712, opened 2020-09-17
Digital Subscription redemption form part 1

rupertbates:
## Why are you doing this?

We need to update the corporate sub redemption form to cope with gift subscriptions. This PR does some preparatory refactoring:

* Move code for determining the state of a gift sub (redeemed, expired etc.) out of `CreateZuoraSubscription` into a separate class so that it can be used from support-frontend.
* Change the redemption form url from `/subscribe/redeem/welcome/[the code]` to `/subscribe/redeem?redemptionCode=[the code]` as this is more flexible, particularly when we want to display the form without a pre-filled code
* Move code which was in `com.gu.support.redemption` into either `com.gu.support.redemption.corporate` or `com.gu.support.redemption.gifting` to make clearer which type of redemption it relates to.


support-frontend/conf/routes
```diff
@@ -109,7 +109,7 @@ GET  /uk/subscribe/paper/delivery                                  controllers.P
 GET  /subscribe/paper/checkout                                     controllers.PaperSubscription.displayForm()
 
 # redemptions
-GET  /subscribe/redeem/welcome/:redemptionCode                     controllers.RedemptionController.displayForm(redemptionCode: String)
+GET  /subscribe/redeem                                             controllers.RedemptionController.displayForm(redemptionCode: String)
```
johnduffell, line 112: good idea, I hit this this morning when I was trying to demo a blank redemption page.  I assume the /r/<code> redirect will still work?

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -291,27 +272,13 @@ object DigitalSubscriptionGiftRedemption {
   ): Future[HandlerResult[SendThankYouEmailState]] =
     zuoraService.getSubscriptionFromRedemptionCode(redemptionData.redemptionCode).flatMap(
       redemptionQueryResponse =>
-        getSubscriptionState(redemptionQueryResponse, state.requestId.toString) match {
+       GiftRedemptionState.getSubscriptionState(redemptionQueryResponse, state.requestId.toString) match {
```
johnduffell, line 275:
```suggestion
        GiftRedemptionState.getSubscriptionState(redemptionQueryResponse, state.requestId.toString) match {
```

johnduffell: in case it's not clear I was adding the space back in above!

johnduffell: not from this PR, but getSubscriptionState could take a UUID instead of a String.  This means that we get better localised type checking, so if we later change state.requestId to another type, we will get a compile error, rather than it running toString on whatever the new type is.  I'm not sure if our tests would cover that adequately.

johnduffell (approved):
great making it a small PR, it encourages quicker reviews! 👏 
just a couple of question/minor comments but nothing show stopping

support-frontend/conf/routes
rupertbates: Yes it does, it converts the code to a querystring

Jpsi (approved):
Breaking large files into smaller pieces definitely makes it easier to read. Thank you!

johnduffell:
move zuora service to support services so that support-frontend has access to zuora.  I can see why, and it was part of our decision process about where to store the codes, but that is a singificant new dependency to have, as so far we have intentionally not relied on zuora.

PR #2713, opened 2020-09-17
normalise email key names for braze

johnduffell:
## Why are you doing this?

Membership workflow lowercases and replaces spaces with _. 
See https://github.com/guardian/membership-workflow/blob/2eea2601b284519e67dbcdb2226ce89b3c81c6cc/app/model/BrazeCampaignTriggerPayload.scala#L11

This is presumably because just the old ET system allowed those and we used them, but braze doesn't, and membership workflow is a convenient place where everything passes through, microservice style.

However it doesn't make sense to retain that intenitonally, as it's just confusing when disucssing things between braze users and devs.

I have only done this for digital pack, because that is the one I'm touching, but it really should be done to them all.  Should I do them at the same time?

I have tested using the manual email tests and it works okay still.

### DIFF
--- a/support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
+++ b/support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
@@ -21,15 +21,15 @@ object DigitalPackEmailFields {
           paymentFields(pm, directDebitMandateId) ++
             directReaderFields(promotion, billingPeriod, user, currency, paymentSchedule)
         case None /*Corporate*/ => List(
-          "Subscription details" -> "Group subscription"
+          "subscription_details" -> "Group subscription"
         )
       }
 
     val fields = List(
-      "ZuoraSubscriberId" -> subscriptionNumber,
-      "EmailAddress" -> user.primaryEmailAddress,
-      "First Name" -> user.firstName,
-      "Last Name" -> user.lastName,
+      "zuorasubscriberid" -> subscriptionNumber,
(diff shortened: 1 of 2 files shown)


support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
```diff
@@ -21,15 +21,15 @@ object DigitalPackEmailFields {
...
+          "subscription_details" -> "Group subscription"
         )
       }
 
     val fields = List(
-      "ZuoraSubscriberId" -> subscriptionNumber,
-      "EmailAddress" -> user.primaryEmailAddress,
-      "First Name" -> user.firstName,
-      "Last Name" -> user.lastName,
+      "zuorasubscriberid" -> subscriptionNumber,
```
Jpsi, line 29: I can see that this key used to be CamelCase and so results in no underscores after the transformation. Just curious if there is any particular reason the keys followed different conventions to start with (e.g. `ZuoraSubscriberId` vs `First Name`.

johnduffell: not sure really, it's possible some of them are unused now, but it's hard to tell without going through braze.
They would have been added at different times by different people though and as long as they match at our end and braze end, there's no additional requirement on them.

lucymonie (approved):
LGTM! I'm interested to see that Mickey Mouse is allowed into the office 🐭

PR #2715, opened 2020-09-18
Send ophan event for 'return to guardian' link in new thankyou page

tomrf1:
We're comparing the conversion rate for each action on the page.
The 'return to the guardian' link tracking was missed on the new page.

_(just to prove it's the same on both now)_
New thankyou page event:
Screen Shot 2020-09-18 at 13 28 39

Old thankyou page event:
Screen Shot 2020-09-18 at 13 29 08


paulbrown1982 (approved):
👍  good spot

PR #2716, opened 2020-09-24
Pass ZuoraServiceProvider to RedemptionController

rupertbates:
## Why are you doing this?

As part of the Digital Subscriptions gifting work we need to be able to check a gift code's validity in Zuora. This PR makes a `ZuoraService` available to the `RedemptionController` class for this purpose.

NB the dependency passed into `RedemptionController` is actually a `ZuoraServiceProvider` which is used to get a `ZuoraService` configured for the correct environment depending on whether the request is from a test user or a regular user.

### TODO:
- [ ] Update config in all environments 


support-services/src/main/scala/com/gu/zuora/ZuoraServiceProvider.scala
```diff
@@ -0,0 +1,15 @@
...
+
+import com.gu.okhttp.RequestRunners.configurableFutureRunner
+import com.gu.support.config.{ZuoraConfig, ZuoraConfigProvider}
+import com.gu.support.touchpoint.TouchpointServiceProvider
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.duration.DurationInt
+
+class ZuoraServiceProvider(configProvider: ZuoraConfigProvider)(implicit ec: ExecutionContext)
+  extends TouchpointServiceProvider[ZuoraService, ZuoraConfig](configProvider){
```
johnduffell, line 11: we use a ZuoraSubscribeService for the few operations that are needed in the backend, to make mocking easier/safer.  In this case, we would presumably only need one operation, so it makes sense to have a ZuoraLookupService  in a similar way if we want to share the actual ZuoraService class.

johnduffell (approved):
looks  good, pretty uncontroversial as we have had the discussion already 👏 
just one comment about whether we should expose the entire service or just the operation that we need.

support-services/src/main/scala/com/gu/zuora/ZuoraServiceProvider.scala
rupertbates: This is a good point. I've gone back and broken out the functionality related to gifting into a separate class which I'd been wanting to do anyway as `ZuoraService` was getting much too big. 
I've also created a trait `ZuoraGiftLookupService` which just has the lookup method on it and this is all that now gets exposed in support-frontend

johnduffell: great work breaking up zuoraservice, yes we have more classes now, but each one is more cohesive  👏

johnduffell (approved):

PR #2717, opened 2020-09-24
add gifting emails - with dummy data

johnduffell:
## Why are you doing this?

As part of the DS gifting work, we need to trigger a pair of emails at purchase time (thank you email and giftee notification email) and a separate one at redemption time for the giftee when they redeem.

## Changes

* make the email builder return a List rather than a single email
* model the attributes as an ADT rather than building a list of key/value pairs (for Digi sub only at the moment)
* add logic to send the three new emails (as placeholder data only)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -37,6 +37,7 @@ inThisBuild(Seq(
   // https://www.scala-sbt.org/1.x/docs/Cached-Resolution.html
   updateOptions := updateOptions.value.withCachedResolution(true),
   resolvers ++= Seq(Resolver.sonatypeRepo("releases")), // libraries that haven't yet synced to maven central
+  scalacOptions += "-Ypartial-unification",
 ))
 
 lazy val releaseSettings = Seq(
--- a/support-modules/aws/src/main/scala/com/gu/aws/AwsCloudWatchMetricPut.scala
+++ b/support-modules/aws/src/main/scala/com/gu/aws/AwsCloudWatchMetricPut.scala
@@ -45,7 +45,7 @@ object AwsCloudWatchMetricPut {
     metricDatum1.setValue(request.value)
     metricDatum1.setUnit(StandardUnit.Count)
     putMetricDataRequest.getMetricData.add(metricDatum1)
-    Try(client.putMetricData(putMetricDataRequest)).map(_ => Unit)
+    Try(client.putMetricData(putMetricDataRequest)).map(_ => ())
   }
 
 }
--- a/support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
+++ b/support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
@@ -149,7 +149,7 @@ class DigitalPackEmailFields(
   }
 
   private def wrap(dataExtensionName: String, fields: DigitalSubscriptionEmailAttributes) = for {
-    attributePairs <- asFlattenedPairs(fields.asJsonObject)
+    attributePairs <- JsonToAttributes.asFlattenedPairs(fields.asJsonObject)
   } yield EmailFields(attributePairs, Left(sfContactId), user.primaryEmailAddress, dataExtensionName)
 
   private def giftRecipientNotification =
@@ -201,31 +201,6 @@ class DigitalPackEmailFields(
 
 object DigitalPackEmailFields {
 
-  type Failable[A] = Either[String, A]
-  def asFlattenedPairs(value: JsonObject): Failable[List[(String, String)]] = {
-    def flattenToPairs(value: JsonObject, failablePairsSoFar: Failable[Map[String, String]]): Failable[Map[String, String]] =
-      value.toList.foldLeft(failablePairsSoFar) {
-          case (failablePairsSoFar, (fieldName, jValue)) =>
-            failablePairsSoFar.flatMap { pairsSoFar =>
-              jValue.asString match {
-                case None =>
-                  jValue.asObject match {
-                    case None => Left(s"all values should be string or object: ${fieldName} -> $value")
-                    case Some(obj) =>
-                      flattenToPairs(obj, Right(pairsSoFar))
-                  }
-                case Some(string) =>
-                  if (pairsSoFar.contains(fieldName))
-                    Left(s"found duplicate key ${fieldName} in case classes")
-                  else
-                    Right(pairsSoFar + (fieldName -> string))
-              }
-          }
-        }
-
-    flattenToPairs(value, Right(Map.empty)).map(_.toList)
-  }
-
   def paymentFields(paymentMethod: PaymentMethod, directDebitMandateId: Option[String]): PaymentFieldsAttributes =
     paymentMethod match {
       case dd: DirectDebitPaymentMethod => DDAttributes(
@@ -243,5 +218,37 @@ object DigitalPackEmailFields {
       case _: CreditCardReferenceTransaction => CCAttributes()
       case _: PayPalReferenceTransaction => PPAttributes()
     }
+
+}
+
+object JsonToAttributes {
+
+  type Failable[A] = Either[String, A]
+
+  // this turns a JsonObject into a List of key/value pairs for insertion into the existing EmailFields object.
+  // (We don't want to change the non DS emails at this point, so it makes sense to convert the case classes to the existing structure.)
+  def asFlattenedPairs(value: JsonObject): Failable[List[(String, String)]] = {
+    def flattenToPairs(value: JsonObject, failablePairsSoFar: Failable[Map[String, String]]): Failable[Map[String, String]] =
+      value.toList.foldLeft(failablePairsSoFar) {
+        case (failablePairsSoFar, (fieldName, jValue)) =>
+          failablePairsSoFar.flatMap { pairsSoFar =>
+            jValue.asString match {
+              case None =>
+                jValue.asObject match {
+                  case None => Left(s"all values should be string or object: ${fieldName} -> $value")
+                  case Some(obj) =>
+                    flattenToPairs(obj, Right(pairsSoFar))
+                }
+              case Some(string) =>
+                if (pairsSoFar.contains(fieldName))
+                  Left(s"found duplicate key ${fieldName} in case classes")
+                else
+                  Right(pairsSoFar + (fieldName -> string))
+            }
+          }
+      }
+
+    flattenToPairs(value, Right(Map.empty)).map(_.toList)
+  }
 
 }
--- a/support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
+++ b/support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
@@ -1,78 +1,242 @@
 package com.gu.emailservices
 
+import cats.implicits._
+import com.gu.emailservices.DigitalSubscriptionEmailAttributes.PaymentFieldsAttributes
+import com.gu.emailservices.DigitalSubscriptionEmailAttributes.PaymentFieldsAttributes.{CCAttributes, DDAttributes, PPAttributes}
 import com.gu.emailservices.SubscriptionEmailFieldHelpers._
-import com.gu.i18n.Currency
-import com.gu.support.promotions.Promotion
 import com.gu.support.workers._
 import com.gu.support.workers.states.PaymentMethodWithSchedule
+import com.gu.support.zuora.api.ReaderType
+import io.circe._
+import io.circe.generic.semiauto.deriveEncoder
+import io.circe.syntax._
+
+sealed trait DigitalSubscriptionEmailAttributes extends Product with Serializable
+object DigitalSubscriptionEmailAttributes {
+
+  implicit val e1: Encoder.AsObject[BasicDSAttributes] = deriveEncoder
+  implicit val e2: Encoder.AsObject[DirectDSAttributes] = deriveEncoder
+  implicit val e3: Encoder.AsObject[GifteeRedemptionAttributes] = deriveEncoder
+  implicit val e4: Encoder.AsObject[GifterPurchaseAttributes] = deriveEncoder
+  implicit val e5: Encoder.AsObject[GifteeNotificationAttributes] = deriveEncoder
+
+  implicit val encoder: Encoder.AsObject[DigitalSubscriptionEmailAttributes] = Encoder.AsObject.instance {
+    case v: BasicDSAttributes => v.asJsonObject
+    case v: DirectDSAttributes => v.asJsonObject
+    case v: GifteeRedemptionAttributes => v.asJsonObject
+    case v: GifterPurchaseAttributes => v.asJsonObject
+    case v: GifteeNotificationAttributes => v.asJsonObject
+  }
+
+
+  sealed trait PaymentFieldsAttributes extends Product with Serializable {
+    val default_payment_method: String
+  }
+  object PaymentFieldsAttributes {
+
+    implicit val e1: Encoder.AsObject[DDAttributes] = deriveEncoder
+    implicit val e2: Encoder.AsObject[CCAttributes] = deriveEncoder
+    implicit val e3: Encoder.AsObject[PPAttributes] = deriveEncoder
+
+    implicit val encoder: Encoder.AsObject[PaymentFieldsAttributes] = Encoder.AsObject.instance {
+      case v: DDAttributes => v.asJsonObject
+      case v: CCAttributes => v.asJsonObject
+      case v: PPAttributes => v.asJsonObject
+    }
+
+    case class DDAttributes(
+      account_number: String,
+      sort_code: String,
+      account_name: String,
+      mandateid: String,
+      default_payment_method: String = "Direct Debit",
+    ) extends PaymentFieldsAttributes
+
+    case class CCAttributes(
+      default_payment_method: String = "Credit/Debit Card",
+    ) extends PaymentFieldsAttributes
+
+    case class PPAttributes(
+      default_payment_method: String = "PayPal",
+    ) extends PaymentFieldsAttributes
+
+  }
+
+
+  case class BasicDSAttributes(
+    zuorasubscriberid: String,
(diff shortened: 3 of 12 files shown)


build.sbt
```diff
@@ -37,6 +37,7 @@ inThisBuild(Seq(
   // https://www.scala-sbt.org/1.x/docs/Cached-Resolution.html
   updateOptions := updateOptions.value.withCachedResolution(true),
   resolvers ++= Seq(Resolver.sonatypeRepo("releases")), // libraries that haven't yet synced to maven central
+  scalacOptions += "-Ypartial-unification",
```
johnduffell, line 40: this is moved to the top level.  It was already in stripe-intent and payment-api.  It is needed to do sequence on Either[String, A] without needing a type alias. 

> Cats relies on improved type inference via the fix for SI-2712, which is not enabled by default.

from https://typelevel.org/cats/

support-modules/aws/src/main/scala/com/gu/aws/AwsCloudWatchMetricPut.scala
```diff
@@ -45,7 +45,7 @@ object AwsCloudWatchMetricPut {
     metricDatum1.setValue(request.value)
     metricDatum1.setUnit(StandardUnit.Count)
     putMetricDataRequest.getMetricData.add(metricDatum1)
-    Try(client.putMetricData(putMetricDataRequest)).map(_ => Unit)
+    Try(client.putMetricData(putMetricDataRequest)).map(_ => ())
```
johnduffell, line 48: this was just a bug/nonsense and it doesn't even compile in scala 2.13.  Can put in another PR if necessary but it will need fixing at some point.

support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
```diff
@@ -1,78 +1,242 @@
 package com.gu.emailservices
```
johnduffell, line 1: this file was originally a single bulder with a couple of branch points to switch between corp/direct.
There is now a top level branching point between corp/direct/gift purchase/gift redemption which then calls out to a separate method for each one.
This is similar in some ways to the SendThankYouEmail buildEmail function, that has a top level branching point to choose between Contribution/DS/paper/weekly and calls out to each builder in a separate file.

I decided not to make a totally different file for each subcategory of DS, as there is some shared code, and putting it in a class rather than an object means we don't have to pass around so many parameters, as they are all in scope.  However I'm happy to keep things consistent in that way if people prefer that.

support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
```diff
@@ -1,78 +1,242 @@
...
+    subscription_details: String,
+    paymentAttributes: PaymentFieldsAttributes,
+    date_of_first_payment: String,
+  ) extends DigitalSubscriptionEmailAttributes
+
+  case class GifteeNotificationAttributes(
+    gifter_first_name: String,
+    gift_personal_message: String,
+    gift_code: String,
+  ) extends DigitalSubscriptionEmailAttributes
```
johnduffell, line 110: these lines define the interface between support-workers and braze emails in one place.  This should make it easier in future when working out the contract between us and braze.

support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
```diff
@@ -1,78 +1,242 @@
...
+                  if (pairsSoFar.contains(fieldName))
+                    Left(s"found duplicate key ${fieldName} in case classes")
+                  else
+                    Right(pairsSoFar + (fieldName -> string))
+              }
+          }
+        }
+
+    flattenToPairs(value, Right(Map.empty)).map(_.toList)
   }
```
johnduffell, line 222: this just turns a JsonObject into a List of key/value pairs for insertion into the existing EmailFields object.  We don't want to change the non DS emails at this point, so it makes sense to convert the case classes to the existing structure.

support-workers/src/main/scala/com/gu/emailservices/EmailFields.scala
```diff
@@ -4,13 +4,18 @@ import com.gu.i18n.Currency
...
 import com.gu.support.workers.{BillingPeriod, User}
-import io.circe.generic.auto._
+import io.circe.Encoder
+import io.circe.generic.semiauto._
 import io.circe.syntax._
 
 case class EmailPayloadContactAttributes(SubscriberAttributes: Map[String, String])
 case class EmailPayloadTo(Address: String, ContactAttributes: EmailPayloadContactAttributes)
-case class EmailPayload(To: EmailPayloadTo, DataExtensionName: String, SfContactId: Option[String], IdentityUserId: Option[String]) {
-  lazy val jsonString: String = this.asJson.toString
```
johnduffell, line 13: this file used to use the "auto" generated encoders, I assume it was because we were serialising based on `this` rather than a compile time known type.  Given that uses shapeless magic, and we don't use them anywhere else, and it wasn't necessary to apply it to `this` anyway, I just converted for consistency.

support-workers/src/main/scala/com/gu/support/workers/lambdas/SendThankYouEmail.scala
```diff
@@ -61,11 +61,11 @@ class SendThankYouEmail(servicesProvider: ServiceProvider = ServiceProvider)
...
           )
-        )
-      case _: DigitalPack => Right(DigitalPackEmailFields.build(
-        getSubscriptionEmailFields(state, directDebitMandateId),
+        ).map(List(_))
+      case d: DigitalPack => new DigitalPackEmailFields(getSubscriptionEmailFields(state, directDebitMandateId)).build(
         paidSubPaymentData = state.paymentOrRedemptionData.left.toOption,
-      ))
+        readerType = d.readerType
+      )
```
johnduffell, line 68: the builder may return a Left either if we try to send a redemption email for a non corp/gift sub, or if the ADT structure could not be serialised to a flat key/value pairs list.
Because the Either is sequenced this lambda should still be idempodent, so if we want to send multiple emails, and one fails, it will refuse to send either.  This means if it is retrying repeatedy, the gifter wouldn't receive multiple emails!

support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
rupertbates: I'm also not that sure that all of what we do send is used by the emails so it would be nice to check that before writing a test

support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
```diff
@@ -1,78 +1,242 @@
...
+object DigitalSubscriptionEmailAttributes {
+
+  implicit val e1: Encoder.AsObject[BasicDSAttributes] = deriveEncoder
+  implicit val e2: Encoder.AsObject[DirectDSAttributes] = deriveEncoder
+  implicit val e3: Encoder.AsObject[GifteeRedemptionAttributes] = deriveEncoder
+  implicit val e4: Encoder.AsObject[GifterPurchaseAttributes] = deriveEncoder
+  implicit val e5: Encoder.AsObject[GifteeNotificationAttributes] = deriveEncoder
+
+  implicit val encoder: Encoder.AsObject[DigitalSubscriptionEmailAttributes] = Encoder.AsObject.instance {
+    case v: BasicDSAttributes => v.asJsonObject
```
rupertbates, line 24: Why encode to JsonObject rather than Json?

support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
```diff
@@ -1,78 +1,242 @@
...
-    ) ++ fieldsForReaderType
-
-    val dataExtensionName = if (paidSubPaymentData.isDefined) "digipack" else "digipack-corp"
-
-    EmailFields(fields, Left(sfContactId), user.primaryEmailAddress, dataExtensionName)
+    paidSubPaymentData: Option[PaymentMethodWithSchedule],
+    readerType: ReaderType
+  ): Either[String, List[EmailFields]] =
+    (paidSubPaymentData, readerType) match {
+      case (Some(PaymentMethodWithSchedule(pm, paymentSchedule)), ReaderType.Gift) =>
```
rupertbates, line 135: Is there some way to label what each of these cases means? Maybe break out the result into a separate function?
```scala
case (Some(PaymentMethodWithSchedule(pm, paymentSchedule)), ReaderType.Gift) => giftPurchase(pm)
case (Some(PaymentMethodWithSchedule(pm, paymentSchedule)), _) => purchase(pm, paymentSchedule)
```
Not sure if that will just add clutter...

support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
rupertbates: This is useful information include it in a comment?

support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
johnduffell: agree it would be nice to reassess what is used.  a test would still provide value for regressions until then.

I wouldn't want to remove anything in this refactoring PR as if something breaks it adds uncertainty - in future either @GHaberis could check in braze, or we could do some eyeballing on the emails.  I'm not sure if some of the fields may be stored in braze for analysis purposes?  I think George was creating a doc for the braze->support workers interface but I'm not sure how far he got.

support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
johnduffell: good idea, will do.  I might even pull that function out to a separate file, as it's nothing to do with emails really, it's just a simple conversion function.

support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
johnduffell: more specific type, a Json could also be a JArray, JString, JNumber etc, so it saves checking that before we start traversing in the flattening function.

I'm not totally sold on it (as I prefer json stuff to be as totally standard as possible, as anything new always requires consuilting the docs) so happy to change back and have the extra bit of scala code and a runtime check if you prefer?

support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
johnduffell: yeah actually it's the unclear use of Some/None that's making it hard for me.

another option is to not use "Option" and roll a clearly named type (or even a type alias if that works?), so we would have 
```scala
case (Purchase(purchaseInfo), ReaderType.Gift) => giftPurchase(purchaseInfo.pm)
case (Purchase(purchaseInfo), _) => purchase(purchaseInfo)
case (Redemption, ReaderType.Corporate) => wrap("digipack-corp", buildCorpRedemptionAttributes).map(List(_))
case (Redemption, ReaderType.Gift) => wrap("digipack-gift-redemption", buildGiftRedemptionAttributes).map(List(_))
case (Redemption, _) => Left("can only redeem gift and corporate subs")
```

johnduffell: I'll first try moving the call to "wrap" into the functions that already exist, see how that helps.

johnduffell: turns out it's trivial to alias, seems to work locally
```scala
  def build(
    paidSubPaymentData: Option[PaymentMethodWithSchedule],
    readerType: ReaderType
  ): Either[String, List[EmailFields]] = {

    val Purchase = Some
    val Redemption = None

    (paidSubPaymentData, readerType) match {
      case (Purchase(paymentInfo), ReaderType.Gift) =>
        List(
          buildGiftPurchaseAttributes(paymentInfo.paymentMethod),
          buildGiftNotificationAttributes
        ).sequence
      case (Purchase(paymentInfo), _) => directAttributes(paymentInfo).map(List(_))
      case (Redemption, ReaderType.Corporate) => buildCorpRedemptionAttributes.map(List(_))
      case (Redemption, ReaderType.Gift) => buildGiftRedemptionAttributes.map(List(_))
      case (Redemption, _) => Left("can only redeem gift and corporate subs")
    }
  }
```

support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
rupertbates: No happy to go with this

support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
```diff
@@ -1,78 +1,254 @@
...
+    subscription_details = details
+  )
 
-    val dataExtensionName = if (paidSubPaymentData.isDefined) "digipack" else "digipack-corp"
+  def build(
+    paidSubPaymentData: Option[PaymentMethodWithSchedule],
+    readerType: ReaderType
+  ): Either[String, List[EmailFields]] = {
+
+    val Purchase = Some
```
rupertbates, line 135: This is really nice 👏

support-workers/src/main/scala/com/gu/emailservices/EmailFields.scala
```diff
@@ -31,7 +36,7 @@ case class EmailFields(
       DataExtensionName = dataExtensionName,
       SfContactId = userId.left.toOption.map(_.id),
       IdentityUserId = userId.right.toOption.map(_.id)
-    ).jsonString
+    ).asJson.toString
```
rupertbates, line 39:
```suggestion
    ).asJson.noSpaces
```
toString behaves slightly odd if the Json isn't a string I think and even though this isn't it's probably better for consistency not to use it

support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
johnduffell: while I was looking it up, I found scala 3 has a really middle ground between type alias and value class, which should reduce the boilerplate for impementing a new type just for a bit of validation.
http://dotty.epfl.ch/docs/reference/other-new-features/opaques.html

support-workers/src/test/scala/com/gu/support/workers/integration/SendThankYouEmailSpec.scala
```diff
@@ -122,176 +132,134 @@ object SendContributionEmail extends App {
...
-    ))),
-  )
-  val service = new EmailService(realConfig.contributionThanksQueueName)
-  service.send(ef)
+    )
+  }
 
-}
-object SendWeeklySubscriptionGiftEmail extends App {
+  //
```
rupertbates, line 233: Remove comment

support-workers/src/test/scala/com/gu/support/workers/integration/SendThankYouEmailSpec.scala
```diff
@@ -122,176 +132,134 @@ object SendContributionEmail extends App {
...
 
-  val mandateId = "65HK26E"
-  val billingAddressWithCountry = Address(
+  val countryOnlyAddress = Address(lineOne = None, lineTwo = None, city = None, state = None, postCode = None, country = UK)
+
+  val billingOnlyUser = User("1234", addressToSendTo, None, "Mickey", "Mouse", billingAddress = countryOnlyAddress)
+
+  //
```
rupertbates, line 239: Remove comment

support-workers/src/test/scala/com/gu/support/workers/integration/SendThankYouEmailSpec.scala
```diff
@@ -122,176 +132,134 @@ object SendContributionEmail extends App {
...
-    Some(new LocalDate(2019, 3, 26)),
-    PaymentMethodWithSchedule(directDebitPaymentMethod, PaymentSchedule(List(Payment(new LocalDate(2019, 3, 25), 37.50)))),
-    giftRecipient = Some(GiftRecipient(None, "Earl", "Palmer", None))
-  )
-
-  val service = new EmailService(realConfig.contributionThanksQueueName)
-  service.send(ef)
 
-}
+  //
```
rupertbates, line 260: Remove comment

support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
rupertbates: Happy to keep it as is

support-workers/src/main/scala/com/gu/emailservices/EmailFields.scala
johnduffell: good point, I'll change.
this was actually moved from line 13 above.   I  intended to change to spaces2, as that's what it calls under the covers, and it's more type safe as you mention.
Do you have a particular preference between nospaces and spaces2?

johnduffell: image

johnduffell: I've done spaces2 for now, happy to change before merging though

rupertbates (approved):
👍 great stuff!

support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
johnduffell: I have added tests for my own piece of mind - they have shown that there are no changes to the JSON output for corporate and direct digital subs.  From visual looking, there are 3 fields we may be able to remove from DS direct so I may play with that a bit.
image

PR #2718, opened 2020-09-24
remove unused DS email fields

johnduffell:
## Why are you doing this?

Rupert mentioned in https://github.com/guardian/support-frontend/pull/2717#discussion_r494334861 that some of the existing email fields are probably not used.
I have done a check of the DS emails and found 3 fields, which I removed, and the emails still seem to send fine in PROD and TEST.
This PR removes the relevant fields.

## sc reenshots
This is the PROD email after removing the fields
image


rupertbates (approved):
Brilliant, well done for following this up 👏

PR #2721, opened 2020-09-25
Update @guardian/consent-management-platform and related dependencies

GHaberis:
## Why are you doing this?

This PR updates `@guardian/consent-management-platform` to the latest version, as used on `frontend`. This latest version of the library contains the new `getConsentFor(vendor, consentState)` interface for obtaining consent for vendors as detailed here: https://www.npmjs.com/package/@guardian/consent-management-platform#getconsentforvendor-consentstate

To use the latest version of `@guardian/consent-management-platform` I needed to update the webpack config to put all `@guardian` scope npm packages through the babel pipeline, further details of why are listed in this PR https://github.com/guardian/frontend/pull/23011 which made  the equivalent change on `frontend`. This also required bringing all `babel` and `webpack` related dependencies up to date. I've also had to:

- Switch from using `.babelrc` to `babel.config.json`, this is babel's recommended format https://babeljs.io/docs/en/babel-plugin-transform-runtime
- Removed the `@babel/plugin-proposal-object-rest-spread` which is now handled by babel `7.11.0`.
- Updated our content security policy to include `https://sourcepoint.theguardian.com`.


sndrs:
looks good 👍

does `getConsentFor` cover the use case you had for [`onConsentChangeEvent`](https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/helpers/tracking/thirdPartyTrackingConsent.js)?

mxdvl (approved):
Great to see this.

Worth noting that `https://consent.theguardian.com` is no longer used by the CMP. It might be worth removing it from the CSP.

On a side note: would defining the CSP as an array and then concatenating into a string—maybe using [mkString][]—make diffing more readable?

[mkString]: https://www.scala-lang.org/api/2.7.5/scala/Iterable.html#mkString%28String%29

GHaberis:
> Worth noting that `https://consent.theguardian.com` is no longer used by the CMP. It might be worth removing it from the CSP.
> 
> On a side note: would defining the CSP as an array and then concatenating into a string—maybe using [mkString](https://www.scala-lang.org/api/2.7.5/scala/Iterable.html#mkString%28String%29)—make diffing more readable?

Thanks for feedback on CSP, I'll update the paths and test on `CODE`. Regarding readability, your suggestion would definitely help. We're due to revisit the CSP at some point so this is something I'll bear in mind when we do that, it's been in this current format for as long as I've seen it and making the changes suggested wasn't necessarily in the scope of this ticket.

GHaberis:
> looks good 👍
> 
> does `getConsentFor` cover the use case you had for [`onConsentChangeEvent`](https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/helpers/tracking/thirdPartyTrackingConsent.js)?

Yes it looks useful to us, I hope to update our code to use the new `getConsentFor` function in a separate PR 👍

sndrs (approved):
lgtm

PR #2724, opened 2020-09-28
Re-introduce social share referral codes

tomrf1:
So that we can track successful referrals from social media.
The query string parameter is now `referralData=${source}_${platform}_${referralCode}`, e.g. `referralData=thankyou_facebook_ABCDEFG`.

This feature is disabled by default, and enabled if a campaign is live and its `createReferralCodes` field is true.

This applies only to the new thank you page, as the old one will soon be removed.


michaelbjacobson (approved):
Awesome! 👍

PR #2725, opened 2020-09-28
Digital Subscription gift redemption form 2

rupertbates:
## Why are you doing this?

We want people who have been gifted a Digital Subscription code to be able to redeem that code, this PR adapts the existing corporate subscriptions redemption form to also work with gift subscriptions

## Changes

* Unify the types used by corporate and gift redemption - previously they were very similar but separate
* Make the validation in `RedemptionController` aggregate the validation of corporate and gift codes with some renaming of classes for consistency
* Return readerType to the redemption form on successful code validation because it's needed for the `create` request - previously this was hard coded to `Corporate` as that was all that the form was used for

### TODO

- [ ] Move corporate codes to lowercase for consistency with gifting
- [ ] Only validate when code is the correct length
- [ ] Use provider pattern for CorporateCodeValidator for consistency with other services


support-frontend/app/controllers/RedemptionController.scala
```diff
@@ -220,19 +237,40 @@ class TestUserFromRequest(identityService: IdentityService, testUsers: TestUserS
 
 }
 
-class GetCorporateCustomer(dynamoLookup: DynamoTableAsyncForUser) {
+class CodeValidator(zuoraLookupService: ZuoraGiftLookupService, dynamoLookup: DynamoTableAsyncForUser) {
+
+  def validate(inputCode: String, isTestUser: Boolean)(implicit ec: ExecutionContext): EitherT[Future, String, ReaderType] =
```
johnduffell, line 242:
```suggestion
  def validationErrorOrReaderType(code: String, isTestUser: Boolean)(implicit ec: ExecutionContext): EitherT[Future, String, ReaderType] =
```

support-services/src/main/scala/com/gu/support/redemption/CodeValidationResult.scala
```diff
@@ -0,0 +1,30 @@
...
+
+case object CodeAlreadyUsed extends InvalidCode("code_already_used")
+
+case object CodeExpired extends InvalidCode("code_expired")
+
+case object InvalidReaderType extends InvalidCode("invalid_reader_type")
+
+object ValidGiftCode { val clientCode = "valid_gift_code"}
+
+case class ValidGiftCode(subscriptionId: String) extends ValidCode(ValidGiftCode.clientCode)
```
johnduffell, line 23: regarding ValidGiftCode and ValidCode, assuming the distinction is necessary, would be nice to capture that distinction in the name

support-frontend/app/controllers/RedemptionController.scala
```diff
@@ -220,19 +237,40 @@ class TestUserFromRequest(identityService: IdentityService, testUsers: TestUserS
...
+    EitherT(getValidationResult(inputCode, isTestUser).map {
+      case ValidCorporateCode(_) => Right(Corporate)
+      case ValidGiftCode(_) => Right(Gift)
+      case invalidCode => Left(validationResultToErrorMessage(invalidCode))
```
johnduffell, line 246: add explicit type to make sure we get a warning if we add another Valid[....] and haven't added to the pattern match
also consider inlining this too
```suggestion
      case CodeAlreadyUsed => Left("This code has already been redeemed")
      case CodeExpired => Left("This code has expired")
      case _: InvalidCode => Left("Please check the code and try again")
```

support-frontend/app/controllers/RedemptionController.scala
```diff
@@ -220,19 +237,40 @@ class TestUserFromRequest(identityService: IdentityService, testUsers: TestUserS
...
+            giftValidationResult <- giftValidator.validate(redemptionCode, "")
+            mergedResult <- if (giftValidationResult == CodeNotFound)
+              corporateValidator.validate(redemptionCode)
+            else
+              Future.successful(giftValidationResult)
+          } yield mergedResult
+      }.fold(
+      malformedCode => Future.successful(malformedCode),
+      validationResult => validationResult
+    )
```
johnduffell, line 267: may be slightly clearer with pattern match?
```suggestion
      } match {
        case Left(malformedCode) => Future.successful(malformedCode),
        case Right(validationResult) => validationResult
    }
```

johnduffell: in fact just replace the fold with `.merge`, and replace line 254 with 
```scala
.leftMap(_ => Future.successful(CodeMalformed))
```

support-services/src/main/scala/com/gu/support/redemption/CodeValidationResult.scala
```diff
@@ -0,0 +1,30 @@
+package com.gu.support.redemption
+
+import com.gu.support.redemption.corporate.CorporateCodeValidator.CorporateId
+
+sealed abstract class CodeValidationResult(val clientCode: String)
```
johnduffell, line 5: can we call this CodeStatus as it's a bit more specific than Result and I was mixing it up with RedemptionValidationResult (or not sure if that should also/instead consider a rename?)

support-frontend/app/controllers/RedemptionController.scala
```diff
@@ -220,19 +237,40 @@ class TestUserFromRequest(identityService: IdentityService, testUsers: TestUserS
...
-        case GetCodeStatus.CodeAlreadyUsed => "Your code has already been redeemed"
-      }
-    } yield ()
+    RedemptionCode(inputCode)
+      .leftMap(_ => CodeMalformed)
+      .map {
+        redemptionCode =>
+          for {
+            giftValidationResult <- giftValidator.validate(redemptionCode, "")
+            mergedResult <- if (giftValidationResult == CodeNotFound)
```
johnduffell, line 259: I'm always slightly wary of using the "try one thing and if it fails try another" logic because it takes away the certainty about what used to happen, if it were to start failing, or anything about which code path it follows if something goes wrong.

Also it perhaps adds a special case where CodeNotFound is not an InvalidCode.  I was trying to think of an alternative name for InvalidCode to make that clearer, then I was unsure why it was like that.

The alternative would be to inspect the first character of the code, so the codes stop being opaque but we do gain the determinism.  (the codes are not actually opaque anyway, as the RedemptionCode type asserts the overall format).  What you you think?

support-frontend/app/views/subscriptionRedemptionForm.scala.html
```diff
@@ -29,6 +31,7 @@
       window.guardian = window.guardian || {};
       window.guardian.stage = @JsStringLiteral(stage);
       window.guardian.userCode = @JsStringLiteral(redemptionCode);
+      window.guardian.readerType = @JsStringLiteral.fromOption(maybeReaderType.map(_.toString));
```
johnduffell, line 34: if we can add a suitable method instead of using toString, this would not be affected if we were to add an extra field or fail to compile if we change the ReaderType object in some incompatible way.  I'm especially nervous because this is the "zuora.api ReaderType" object but it's getting all over the place now

support-frontend/assets/pages/subscriptions-redemption/api.js
```diff
@@ -114,14 +121,19 @@ function buildRegularPaymentRequest(
...
   user: User,
   currencyId: IsoCurrency,
   countryId: IsoCountry,
   participations: Participations,
   csrf: Csrf,
   dispatch: Dispatch<Action>,
 ) {
-  const data = buildRegularPaymentRequest(userCode, user, currencyId, countryId, participations);
+  if (readerType == null) {
+    dispatchError(dispatch, 'An error occurred while redeeming this code, please try again later');
```
johnduffell, line 133: how will we know if users are seeing this error? This applies more generally to failures of redemption - do we expect them to contact the call centre/userhelp/do we want an alarm/something else?

support-services/src/main/scala/com/gu/support/redemption/CodeValidationResult.scala
```diff
@@ -0,0 +1,30 @@
...
+
+sealed abstract class ValidCode(clientCode: String) extends CodeValidationResult(clientCode)
+
+sealed abstract class InvalidCode(clientCode: String) extends CodeValidationResult(clientCode)
+
+case object CodeNotFound extends CodeValidationResult("code_not_found")
```
johnduffell, line 11: why is "code not found" not invalid? (in terms of the ADT it isn't)

support-services/src/main/scala/com/gu/support/redemption/corporate/CorporateCodeValidator.scala
```diff
@@ -33,15 +29,15 @@ object GetCodeStatus {
...
 }
 
-class GetCodeStatus(dynamoLookup: DynamoLookup) extends WithLogging {
+class CorporateCodeValidator(dynamoLookup: DynamoLookup) extends WithLogging {
 
-  import GetCodeStatus._
+  import CorporateCodeValidator._
 
-  def apply(code: RedemptionCode)(implicit ec: ExecutionContext): Future[Either[RedemptionInvalid, CorporateId]] =
+  def validate(code: RedemptionCode)(implicit ec: ExecutionContext): Future[CodeValidationResult] =
```
johnduffell, line 40: good to stop this being called "apply" but I think could be more positively named e.g. getStatus, as validate doesn't tell you what happens depending whether it's valid or not

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -161,7 +160,7 @@ object CreateZuoraSubscription {
   private def buildSubscriptionData(
     state: CreateZuoraSubscriptionState,
     promotionService: => PromotionService,
-    getCodeStatus: => GetCodeStatus,
+    getCodeStatus: => CorporateCodeValidator,
```
johnduffell, line 163:
```suggestion
    corporateCodeValidator: => CorporateCodeValidator,
```

support-services/src/main/scala/com/gu/support/redemption/gifting/GiftCodeValidator.scala
```diff
@@ -0,0 +1,35 @@
...
+import com.gu.support.zuora.api.response.SubscriptionRedemptionQueryResponse
+import com.gu.zuora.{ZuoraGiftLookupService, ZuoraService}
+import org.joda.time.LocalDate
+
+import scala.concurrent.{ExecutionContext, Future}
+
+object GiftCodeValidator {
+  val expirationTimeInMonths = 12
+
+  def getSubscriptionState(existingSub: SubscriptionRedemptionQueryResponse, requestId: String): CodeValidationResult =
```
johnduffell, line 15:
```suggestion
  def getSubscriptionState(existingSub: SubscriptionRedemptionQueryResponse, requestId: Option[String]): CodeValidationResult =
```
then we don't need to pass in an empty string for nothing

johnduffell:
unfortunately my comments turned into a bit of a hodgepodge but hopefully you can make sense of them, we can chat through any that are not clear or I can clarify tomorrow!  Would be interested in any areas you had concerns with.  If we allowed the client and server side to look at the first character of the code, I think it would improve a lot of the code, but it depends whether we envisage that prefix changing.

support-services/src/main/scala/com/gu/support/redemption/CodeValidationResult.scala
rupertbates: `ValidCode` is an abstract class, it's extended by `ValidGiftCode` and `ValidCorporateCode` to represent the set of valid code types.

johnduffell: ok got it now! 👍

support-services/src/main/scala/com/gu/support/redemption/CodeValidationResult.scala
rupertbates: Because it only means that the code was not found in one of the two systems not that we have determined that there is a specific problem with it. I can see how you could decide to make it extend `InvalidCode` but I think this approach is more flexible

support-frontend/app/controllers/RedemptionController.scala
rupertbates: I really don't want to rely on parsing strings to work out what type of code we have. If we think that it is too much of a risk to try to validate with two separate systems then I would rather have a separate url for each redemption type. As you've noted elsewhere `CodeNotFound` is actually not an `InvalidCode` for this reason.

support-frontend/assets/pages/subscriptions-redemption/api.js
rupertbates: This _should_ never happen it's just that we need to go from an `Option<ReaderType>` to a `ReaderType` to keep Flow happy

rupertbates: The more general point is a good one, we should have a think about what could be worth alerting on but I can only really think of code validation failures which probably aren't worth an alarm. Other failures will be picked up by the standard step function failure alerts. 
Can you think of anything else?

support-frontend/app/views/subscriptionRedemptionForm.scala.html
rupertbates: I've realised there is already a `value` field on `ReaderType` so I will use that.

support-frontend/app/controllers/RedemptionController.scala
johnduffell: let's leave it as the try one then the other, I accept your reasons for not wanting to look inside the code and I think that having a separate url would be similar complexity on the backend, only more confusing for readers, so overall worse.

support-services/src/main/scala/com/gu/support/redemption/CodeValidationResult.scala
johnduffell: yeah makes sense, I agree with your point about not subclassing, based on that I am wondering in my head whether the naming could be better, or we could call it a CorporateCodeNotFound to contain the intermediate state, and have a separate CodeNotFound that covers eventual failure and extends InvalidCode.
However that may get messy in pursuit of trying to reduce the WTF factor?

support-frontend/app/controllers/RedemptionController.scala
rupertbates: I would rather keep the method name as the verb for what is happening 'validate' and allow the return type to specify what is returned. It feels redundant to include the return type in the method name.

support-frontend/assets/pages/subscriptions-redemption/api.js
johnduffell: yeah I think as long as it's clear for users to call up we are going to have to rely on that more generally.  If not clear we can add a contact us link so they can find the phone number.
It might be worth having an alarm if the user comes from a clicked link, but the code is not valid, as that should generally be the case.  If it turns out that people commonly click the link a second time, we can go back and exclude "code already used"?
In other areas we have a "no X for Y hours" alarm, but I don't know we can rely on lack of conversions at this stage.

johnduffell (approved):
great work, nothing unclear, although there are a few outstanding comments to consider left before merging.
👍

support-frontend/app/controllers/RedemptionController.scala
johnduffell: fair enough.  Sounds like I am thinking of the FP way of "only the result matters" whereas you are leaning more to the "ask it to do something"

support-services/src/main/scala/com/gu/support/redemption/CodeValidationResult.scala
rupertbates: I don't think it's right to say that there's not much shared between the two code types
If we look at the concrete types:
```
case object CodeNotFound extends CodeStatus("code_not_found")
case object CodeMalformed extends InvalidCode("code_malformed")
case object CodeAlreadyUsed extends InvalidCode("code_already_used")
case object CodeExpired extends InvalidCode("code_expired")
case object InvalidReaderType extends InvalidCode("invalid_reader_type")
```
They could all apply to both types of redemption. We don't currently check expiry for corporate codes but I actually think that's an oversight and we will probably need to add it. Similarly `InvalidReaderType` is a problem which could occur in either case.

support-services/src/main/scala/com/gu/support/redemption/corporate/CorporateCodeValidator.scala
rupertbates: It is for consistency with the `GiftCodeValidator` but happy to change both to `getStatus`.

support-services/src/main/scala/com/gu/support/redemption/CodeValidationResult.scala
johnduffell: fair point yes you are right there.  I still don't know why I have unease about this type so I don't think it should stop it going into prod!

johnduffell (approved):
thanks, great work on this one 👍

support-workers/src/test/scala/com/gu/support/workers/integration/DigitalSubscriptionGiftRedemptionIntegrationSpec.scala
```diff
@@ -25,6 +24,7 @@ class DigitalSubscriptionGiftRedemptionIntegrationSpec extends AsyncLambdaSpec w
   val createSubRequestId = UUID.randomUUID()
   val redeemSubRequestId = UUID.randomUUID()
   val giftCode = new GiftCodeGeneratorService().generateCode(Annual)
+  SafeLogger.info(s"Gift code = ${giftCode}")
```
johnduffell, line 27: didn't realise that anything in the test class will happen right at the startt  of the test run
image
might be better to log it in the first test instead @rupertbates ?

PR #2727, opened 2020-09-28
email lambda returns a list rather than a single item

johnduffell:
## Why are you doing this?

this is a followup fix to https://github.com/guardian/support-frontend/pull/2717/files#diff-681c59db0de6db865fe598c3f08c823eR29
Where it was changed to be a List, I should have changed the deserialiser to also use a list.  I didn't expect the result was used anywhere and I forgot to run the IT tests.

This suggests that the IT test runner is not working correctly, as it should have detected this issue.


Jpsi (approved):
Thank you!

PR #2702, opened 2020-09-14
Digital gift subscriptions: checkout and thank you pages

lucymonie:
## Why are you doing this?
This is part one of the frontend DS gifting work: the checkout.

[**Trello Card**](https://trello.com/c/5NHvddVH/3307-set-up-ds-gifting-routes-and-clientside-data-flow)
[**Designs in Figma**](https://www.figma.com/file/CMWbat4kilJheLOPfXb8ix/Gifting---Digital-Subscriptions?node-id=517%3A2591)

## Changes
This is part one of a three-step rollout of the new gifting checkout. The steps are as follows:
1. MVP DS gift checkout in current checkout style but with date picker and gift message.

Not included in this PR:
2. Progressive disclosure checkout with playbacks (per designs)
3. Design systems components (per designs).

## Did you enjoy your PR experience?

 - [ ] [PR experience rated](https://forms.gle/N6FsTGG8JQFGV4Ha9)

## Accessibility test checklist
 - [x] [Tested with screen reader](https://accessibility.gutools.co.uk/testing/web/screen-readers/)
 - [x] [Navigable with keyboard](https://accessibility.gutools.co.uk/testing/web/keyboard-navigation/)
 - [x] [Colour contrast passed](https://accessibility.gutools.co.uk/testing/web/colour-contrast/)

NB: The grey text in the order summary section of the thank you page is not AAA at that size.

## Screenshots
support thegulocal com_subscribe_digital_checkout_gift_promoCode=ANNUAL-INTRO-UK period=Annual(wide)

support thegulocal com_subscribe_digital_checkout_gift_promoCode=ANNUAL-INTRO-UK period=Annual(wide) (1)

support thegulocal com_subscribe_digital_checkout_gift_promoCode=ANNUAL-INTRO-UK period=Annual(wide) (2)

PR #2730, opened 2020-09-29
send gift message through in the emails

johnduffell:
#Why are you doing this?

This PR adds the gift message field, to get it through from the client side to the giftee email in the SQS queue.

## Changes

See individual PR comments inline 

TODO
- test in CODE/locally


support-frontend/app/services/stepfunctions/SupportWorkersClient.scala
```diff
@@ -105,64 +108,78 @@ class SupportWorkersClient(
...
-          )
-        } getOrElse {
-          SafeLogger.error(scrub"[$requestId] Failed to parse ${success.arn} to a jobId after triggering Step Function execution for ${user.id} $request")
-          StatusResponse(
-            status = Status.Failure,
-            trackingUri = "",
-            failureReason = Some(CheckoutFailureReasons.Unknown)
-          )
+    for {
+      giftRecipient <- EitherT.fromEither[Future](request.body.giftRecipient.map(getGiftRecipient(_, request.body.product)).sequence)
```
johnduffell, line 141: the key change here is that it's now an error to have inconsistent giftee details from the client side, rather than just silently continuing as a non-gift subscription

support-frontend/assets/helpers/paymentIntegrations/readerRevenueApis.js
```diff
@@ -95,6 +95,14 @@ export type RegularPaymentRequestAddress = {|
   city: Option<string>,
 |};
 
+type GiftRecipientType = {|
+  title?: Option<Title>,
+  firstName: string,
+  lastName: string,
+  email?: Option<string>,
+  // message: Option<string>,
```
johnduffell, line 103: this will conflict with https://github.com/guardian/support-frontend/pull/2702/files#r494923958
the format the JSON should arrive in for giftee details is changed in my proposal, so we will need to chat some time!
@lucymonie

support-models/src/main/scala/com/gu/support/workers/GiftRecipient.scala
```diff
@@ -1,12 +1,79 @@
...
 import com.gu.support.encoding.Codec.deriveCodec
 import com.gu.support.encoding.CustomCodecs._
+import com.gu.support.workers.GiftRecipient.{DigiSub, Weekly}
+import io.circe._
+import io.circe.syntax._
 
-case class GiftRecipient(title: Option[Title], firstName: String, lastName: String, email: Option[String])
-
+sealed trait GiftRecipient {
+  val giftRecipientType: String
```
johnduffell, line 13: this I was a bit up in the air, so this whole file is a bit of a FIXME.

We need to keep different info for weekly and DS giftees, so I introduced two subclasses here (rather than the everything optional approach as that just pushes the validation later in a "this can't happen" kind of way)

however the subclasses are compatible, which falls foul of the default "try one and then try the other" way.  i.e. if it has first+last+email, the deserialiser has no idea if it's a weekly or a DS giftee.

I decided to add a discrimintor to the JSON but I couldn't decide whether to use the standard semiauto and expose in the case classes, or whether to make it deterministic and select the deserialiser in a custom Codec. ( I have implemented the built in option, and roughly the deterministic option is roughly commented out below)

support-models/src/main/scala/com/gu/support/workers/GiftRecipient.scala
```diff
@@ -1,12 +1,79 @@
...
+    case c: Weekly => c.asJson
+    case c: DigiSub => c.asJson
+  }
+
+  def or[A, AA >: A](da: => Decoder[A], daa: => Decoder[AA]): Decoder[AA] = new Decoder[AA] {
+    final def apply(c: HCursor): Decoder.Result[AA] = da(c) match {
+      case r @ Right(_) => r
+      case Left(failure)      => daa(c).leftMap(secondFail => DecodingFailure(failure.message + " / " + secondFail.message, failure.history ++ secondFail.history))
+    }
+  }
```
johnduffell, line 34: the "or" method is not essential. It was needed for debugging, as the built in `or` tries them all, and throws away all but the last error.

support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
```diff
@@ -126,18 +126,22 @@ class DigitalPackEmailFields(
...
     val Redemption = None
 
     (paidSubPaymentData, readerType) match {
       case (Purchase(paymentInfo), ReaderType.Gift) =>
-        List(
-          giftPurchaserConfirmation(paymentInfo.paymentMethod),
-          giftRecipientNotification
-        ).sequence
+        for {
+          giftRecipient <- maybeGiftRecipient.toRight("Gift redemption must have a gift recipient")
```
johnduffell, line 139: this can't happen, but it's not actually enforced by the types so it would be a coding error, but at this stage it is a choice between blowing up loudly or sending the emails anyway with some dummy data

support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
```diff
@@ -180,7 +184,7 @@ class DigitalPackEmailFields(
     ))
 
   private def corpRedemption =
-    wrap("digipack-corp", directOrCorpFields("Group subscription"))
+    wrap("digipack-corporate-redemption", directOrCorpFields("Group subscription"))
```
johnduffell, line 187: I decided to change this at the same time to make it more consistent
https://github.com/guardian/membership-workflow/pull/270/files#diff-9145604e78c7f805b7bdb0c49e8785baR54

support-frontend/app/services/stepfunctions/SupportWorkersClient.scala
lucymonie: Thanks for flagging this JD - I'll change this to the new model on clientside 👍

support-frontend/assets/helpers/paymentIntegrations/readerRevenueApis.js
lucymonie: 👍

support-models/src/main/scala/com/gu/support/workers/GiftRecipient.scala
```diff
@@ -1,12 +1,79 @@
...
+
+  implicit val decoder: Decoder[GiftRecipient] =
+    or(Decoder[Weekly], Decoder[DigiSub].widen)
+
+  case class Weekly(
```
rupertbates, line 39: Still thinking about the serialisation approach but can we call these classes `WeeklyGiftRecipient` and `DigitalGiftRecipient` so they more specifically describe what they represent? I would actually probably go for `GuardianWeeklyGiftRecipient` and DigitalSusbcriptionGiftRecipient` but that's a bit of a mouthful!

support-frontend/app/services/stepfunctions/SupportWorkersClient.scala
lucymonie: Or actually, it's probably best if I leave it as you've handled the changes in submit.js and we just make sure it works end-to-end

support-models/src/main/scala/com/gu/support/workers/GiftRecipient.scala
johnduffell: thanks rupert will do on the longer names.

In the slightly longer term I'd like to add a convention to our coding standards about ADTs structure as we have a few different ways
e.g. 
- one side we have several Monthly in the codebase including `com.gu.support.workers.Monthly` (which is actually a BillingPeriod) on one hand and
- the other side we have things like `com.gu.support.workers.GiftRecipient.WeeklyGiftRecipient` which gives great auto complete from entering `GiftRecipient.` but has a bit of duplication if you are looking at the whole type name.

I actually think/agree the latter is best overall as it's both clear in github, and gives good auto complete, but if we write it down as a rule then it becomes easier to be consistent when adding and refactoring.

support-models/src/main/scala/com/gu/support/workers/GiftRecipient.scala
johnduffell: rupert and I have discussed and decided that firstly the default fields thing doesn't actually do what I expected, the default value is ignored so it would just deserialise to the wrong one anyway.  We decided that the deterministic approach is better and it avoids polluting the scala model with serialisation specific stuff.

support-frontend/app/utils/CheckoutValidationRules.scala
```diff
@@ -3,14 +3,16 @@ package utils
...
+import com.gu.support.zuora.api.ReaderType
 import com.typesafe.scalalogging.LazyLogging
 import org.joda.time.LocalDate
 import services.stepfunctions.CreateSupportWorkersRequest
+import services.stepfunctions.CreateSupportWorkersRequest.GiftRecipientRequest
 
 object CheckoutValidationRules {
   def validatorFor(productType: ProductType): CreateSupportWorkersRequest => Boolean = productType match {
-    case _: DigitalPack => DigitalPackValidation.passes
+    case d: DigitalPack => DigitalPackValidation.passes(d)
```
johnduffell, line 15: it would be great to revisit the validation stuff, and consolidate it with the code that turns the data into the State model.
We already know this logic is duplicated between client and server side, but even within the server side it happens either when we convert it into the CreatePaymentMethodState, or later in a leaky way if the types don't retain the information.
In terms of the backend I think we should aim to do this once in one go, (and ideally retain the information in the types, if appropriate to avoid dealing "this can't happen" situations later)  This also means we are less likely to forget to add this validation when we change the model.
background reading
 https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/

support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
```diff
@@ -126,18 +126,22 @@ class DigitalPackEmailFields(
...
 
     (paidSubPaymentData, readerType) match {
       case (Purchase(paymentInfo), ReaderType.Gift) =>
-        List(
-          giftPurchaserConfirmation(paymentInfo.paymentMethod),
-          giftRecipientNotification
-        ).sequence
+        for {
+          giftRecipient <- maybeGiftRecipient.toRight("Gift redemption must have a gift recipient")
+          emailFieldses <- List(
```
rupertbates, line 140: `emailFields`?

rupertbates (approved):
One tiny nitpick but this is a real improvement, thanks!

support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
johnduffell: 😂

johnduffell: oh actually that's what intellij suggests for a double plural! 
image
maybe `emails` would make more sense as fields is a bit of a useless addition in the context but it is important that there are more than one email.

johnduffell:
tested in CODE,
- Digi sub Direct
- GW gift
- GW standard

all emails arrived, subs look fine in zuora UAT, step function executions succeeded.

In the process I found that if support workers and frontend are deployed separately, any gift subscriptions will fail (GW gift being the only one available in PROD).
Having thought about it, I think that support-frontend needs to go out first, then if any gifts are purchased they will keep retrying.  Then deploy support-workers, at which point any outstanding gifts will be unblocked.

PR #2731, opened 2020-09-30
remove thank you page test

tomrf1:
The new thank you page is performing much better than the old, so this change removes the test.

TODO later - remove all code for the old thank you page.


TomPretty (approved):
🔥 it'll feel good to delete all that old code when we get round to it!

PR #2732, opened 2020-09-30
fix post deploy test for thankyou page

tomrf1:
it looks for this class when checking to see if the thankyou page has loaded, so I've added it back in (and removed the styling, as it was for the old page only)
(test was broken here https://github.com/guardian/support-frontend/pull/2731)

Now works locally:
Screen Shot 2020-09-30 at 10 33 31


TomPretty (approved):
👍

PR #2733, opened 2020-09-30
alarm when two IT test runs fail in a row

johnduffell:
## Why are you doing this?

The IT tests are noisy because there are random failures from network issues.  This updates the alarm to be quieter hopefully as it will need two failures in a row.

[**Trello Card**](https://trello.com/c/of5WhYtb/3357-change-ittestfailurealarm-to-trigger-on-failing-tests-if-there-were-also-failing-tests-in-the-previous-run-1u-33)


tomrf1 (approved):
thanks!

PR #2735, opened 2020-10-01
rename dynamo lookup types and add pre-zuora gift code validation

rupertbates:
## Why are you doing this?

Some small tidy up stuff following on from #2693 

## Changes

* Only validate redemption codes with the server when they are of the correct length - this is because we are now validating with Zuora which can be slow and has a rate limit so we don't want to do this any more than we have to. NB. This means that we will now need to ensure that corporate codes are the same length or change this code to take account of the difference.
* Rework the way that dynamo table lookups are managed for test/non-test users to be consistent with the 'ServiceProvider' pattern used for other services.

### DIFF
--- a/support-frontend/assets/pages/subscriptions-redemption/api.js
+++ b/support-frontend/assets/pages/subscriptions-redemption/api.js
@@ -42,7 +42,7 @@ function dispatchReaderType(dispatch: Dispatch<Action>, readerType: Option<Reade
   dispatch({ type: 'SET_READER_TYPE', readerType });
 }
 
-function doValidation(userCode: string, dispatch: Dispatch<Action>) {
+function validateWithServer(userCode: string, dispatch: Dispatch<Action>) {
   validate(userCode).then((result: ValidationResult) => {
     dispatchError(dispatch, result.errorMessage);
     dispatchReaderType(dispatch, result.readerType);
@@ -53,7 +53,11 @@ function doValidation(userCode: string, dispatch: Dispatch<Action>) {
 
 function validateUserCode(userCode: string, dispatch: Dispatch<Action>) {
   dispatch({ type: 'SET_USER_CODE', userCode });
-  doValidation(userCode, dispatch);
+  if (userCode.length === 13) {
+    validateWithServer(userCode, dispatch);
+  } else {
+    dispatchError(dispatch, 'Please check the code and try again');
+  }
 }
 
 function submitCode(userCode: string, dispatch: Dispatch<Action>) {
--- a/support-frontend/assets/pages/subscriptions-redemption/api.js
+++ b/support-frontend/assets/pages/subscriptions-redemption/api.js
@@ -53,7 +53,8 @@ function validateWithServer(userCode: string, dispatch: Dispatch<Action>) {
 
 function validateUserCode(userCode: string, dispatch: Dispatch<Action>) {
   dispatch({ type: 'SET_USER_CODE', userCode });
-  if (userCode.length === 13) {
+  const codeLength = getGlobal('codeLength') || 13;
+  if (userCode.length === codeLength) {
     validateWithServer(userCode, dispatch);
   } else {
     dispatchError(dispatch, 'Please check the code and try again');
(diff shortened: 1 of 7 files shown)


support-frontend/assets/pages/subscriptions-redemption/api.js
```diff
@@ -53,7 +53,11 @@ function doValidation(userCode: string, dispatch: Dispatch<Action>) {
 
 function validateUserCode(userCode: string, dispatch: Dispatch<Action>) {
   dispatch({ type: 'SET_USER_CODE', userCode });
-  doValidation(userCode, dispatch);
+  if (userCode.length === 13) {
```
johnduffell, line 56: just a thought - I know why you are wary about adding knowledge about the code structure to the codebase.
since it's only that we don't want to hit zuora itself, and we already have partial validation here: https://github.com/guardian/support-frontend/blob/c1f856d7b573077b4fb2ff340b7beb7dec0cf7d5/support-models/src/main/scala/com/gu/support/redemptions/RedemptionCode.scala#L9
so would it make sense add the new "13" limit in the same place and do the check on the backend?
Then at least all the assumptions about the code structure are centralised.

johnduffell (approved):
a bit random mixed those two things together!
👍 on the renaming, that looks a more consistent overall and no more complicated in isolation
one comment on the code checking

support-frontend/assets/pages/subscriptions-redemption/api.js
rupertbates: What do you think about having it in both places? I think there is value in not making unnecessary http requests so it's worth having it in the client but maybe we should have it in the server as well?

johnduffell: good idea, both places would make sense at runtime even if not in the codebase twice so something like this in the template
```js
window.guardian.codeLength = ${RedemptionCode.length}
```

johnduffell: this is just a general case of duplication of validation between client and server though

rupertbates: It serves different purposes though, on the client it is preventing unnecessary http requests and on the server it is preventing us hitting Zuora unnecessarily if the client validation has failed for some reason.

support-models/src/main/scala/com/gu/support/redemptions/RedemptionCode.scala
```diff
@@ -6,12 +6,14 @@ import com.gu.support.encoding.Codec
...
   def apply(value: String): Either[String, RedemptionCode] = {
     val upper = value.toUpperCase(Locale.UK)
     // make sure no one can inject anything bad
     val validChars = List('A' -> 'Z', '0' -> '9', '-' -> '-')
     val validCharsSet = validChars.flatMap { case (from, to) => (from to to) }.toSet
-    if (upper.forall(validCharsSet.contains))
+    if (upper.length == length && upper.forall(validCharsSet.contains))
       Right(new RedemptionCode(upper))
     else
       Left(s"redemption code must only include A-Z, 0-9 and '-'")
```
johnduffell, line 19:
```suggestion
      Left(s"redemption code must only include A-Z, 0-9 and '-' and be 13 characters long")
```

support-frontend/assets/pages/subscriptions-redemption/api.js
johnduffell: yeah that's what I realised in the end - I think we both agree it's sensible and appropraite to validate in both places, but we don't want to define the rules in two+ places in the source code due to maintainability and discoverabilty reasons.
Change looks good though 👍

support-frontend/assets/pages/subscriptions-redemption/api.js
```diff
@@ -53,7 +53,12 @@ function doValidation(userCode: string, dispatch: Dispatch<Action>) {
 
 function validateUserCode(userCode: string, dispatch: Dispatch<Action>) {
   dispatch({ type: 'SET_USER_CODE', userCode });
-  doValidation(userCode, dispatch);
+  const codeLength = getGlobal('codeLength') || 13;
```
johnduffell, line 56: how will we know if we are relying on the 13? we will have to also change here if we change the rule, but at least it will be more discoverable via the "find usages" and search for "codeLength" in the JS.
The alternative would be to blow up if it's not present, hopefully before we ship to prod.

johnduffell: which reminds me, we should not forget to add post deploy tests for gifting

support-models/src/main/scala/com/gu/support/redemptions/RedemptionCode.scala
rupertbates: I'm removing the server side validlation for now because it would mean changing test data in Dynamo. I have a follow up PR coming that will do that anyway so I'll roll it into that.

PR #2737, opened 2020-10-01
make thank you email it test run in the hourly test run

johnduffell:
## Why are you doing this?

The SendThankYouEmailIT was failing but somehow wasn't picked up by the automated IT test runs.

Turned out it was because the runner requires all the test names to end in "Spec" https://github.com/guardian/support-frontend/blob/596a3ed349ed085dcfd13ded6b44a768fd128f41/support-lambdas/it-test-runner/src/main/scala/com/gu/RunITTests.scala#L47

I have renamed the test to suit.

there are 204 that run in the automated run (now 205).  When I run support-workers it:test locally it runs 291 tests.  So we might need to check this a bit more generally.


kelvin-chappell (approved):
😖

johnduffell:
look on the bright side, it used to run zero automatically, let's just say this is the 80:20 rule in action!

PR #2739, opened 2020-10-01
allow new email ittest to access sqs queue

johnduffell:
## Why are you doing this?

Having added an IT test that wasn't actually running https://github.com/guardian/support-frontend/pull/2737 , it was failing because it relies on writing to an SQS queue.

This PR adds the relevant permission.

tested in "PROD" with the result: `after: succeeded 208, failed 0, canceled 0, ignored 4, pending 0`


paulbrown1982 (approved):
👍Good catch

PR #2746, opened 2020-10-02
Update copy

TomPretty:
## Why are you doing this?

environment -> climate copy update


paulbrown1982 (approved):
👍

project/.bloop/support-frontend-build.json
```diff
@@ -0,0 +1,2566 @@
+{
```
tomrf1, line 1: I don't think you want this!

TomPretty: ahhhh idk why that keeps happening atm, I need to investigate

PR #2751, opened 2020-10-05
Tp make header fit on two lines

TomPretty:
## Why are you doing this?

Prevent the header wrapping onto 3 lines. Added a leftCol break point, and made the header slightly wider between desktop and leftCol. 

## images

before 
Screenshot 2020-10-05 at 09 51 04

after
Screenshot 2020-10-05 at 09 53 19


tomrf1 (approved):
great

PR #2752, opened 2020-10-05
remove old thank you flow

michaelbjacobson:
### What does this PR do?

Removes now unused code for the old thank you flow, as per [this Trello card](https://trello.com/c/XtEzTh2d/2307-delete-old-thankyou-page-code).


TomPretty (commented):
Looking good. We could potentially remove the `new-flow` folder, and move all those files into the `ContributionThankYou` folder. Also, there might be some other components that are no longer used - e.g. the old `ContributionsSurvey` component?

michaelbjacobson:
> Looking good. We could potentially remove the `new-flow` folder, and move all those files into the `ContributionThankYou` folder. Also, there might be some other components that are no longer used - e.g. the old `ContributionsSurvey` component?

Good shout @TomPretty! I meant to remove the `new-flow` directory, but it totally slipped my mind! So is the `ContributionsSurvey` component unlikely to be reused in future?

TomPretty:
Good question! I would probably defer to @tomrf1 for a definitive answer there. It's a not a huge loss if we delete it because we can always dig it out of the history. I think the component was designed just for the old flow, so I'm guessing we wont use it again. 

For the record, I think these components were only being used by the old flow

- `ContributionsSurvey`
- `ContributionsReminder` (+ test file)
- `MarketingConsentContainer` (under `contributions-landing/components`, not the subs one)

Potentially the `SpreadTheWord` component too, but that is being used by the `payPalError` component. But that doesn't seem to be used anywhere else. 

I guess I don't know how long we want to spend cleaning up here - I suppose it's non essential, but leaving unused code does clutter the repo.

tomrf1:
As Tom says, the old `ContributionsSurvey`, `MarketingConsentContainer` and `ContributionsReminder` components have been replaced, so it's safe to delete them.

Regarding `SpreadTheWord` - let's delete it. I've never seen the `payPalError` component before, but it's not being used and so should also be deleted. It looks like a [legacy thing](https://github.com/guardian/support-frontend/commit/01d227e3702ffb954b2b3083381512b4e5ee6a54)

tomrf1:
actually, I was wrong about the payPalError - it is still used - https://github.com/guardian/support-frontend/blob/main/support-frontend/conf/routes#L151
So we need to leave it and `SpreadTheWord` for now

tomrf1:
_Actually_ - the paypal failure page looks pretty bad with the `SpreadTheWord` component!
See https://support.theguardian.com/paypal/rest/error
I vote we remove `SpreadTheWord` from it, and delete it. But leave `payPalError` in place

michaelbjacobson:
@TomPretty @tomrf1, all the aforementioned dead code has now been excised from the repo!

tomrf1 (approved):
what a satisfying PR!

PR #2756, opened 2020-10-06
tidy zuora subscription lambda 1 of 2

johnduffell:
## Why are you doing this?

This ia  nit of pre tidying of the zuora subscription lambda, so that the way is clearer to get the code and date into the email lambda.

Part 1 (this one) is to update the redemption code stuff to match the consistent naming convention.  Also a bit of tidy up.

Part 2 will be getting rid of the warnings of the CreateZuoraSubscription file especially this one: 
image
this is pretty similar to the issue that SendThankYouEmail had, with the varying logic for contribution/DS/GW/Paper, so hopefully I can apply a similar technique to remove the excessive method length here.  But that will be the next PR.

### DIFF
--- a/support-services/src/main/scala/com/gu/support/redemption/corporate/CorporateCodeStatusUpdater.scala
+++ b/support-services/src/main/scala/com/gu/support/redemption/corporate/CorporateCodeStatusUpdater.scala
@@ -5,15 +5,15 @@ import com.gu.support.redemptions.RedemptionCode
 
 import scala.concurrent.{ExecutionContext, Future}
 
-object SetCodeStatus {
+object CorporateCodeStatusUpdater {
 
-  def withDynamoLookup(dynamoLookup: DynamoUpdate): SetCodeStatus = new SetCodeStatus(dynamoLookup)
+  def withDynamoUpdate(dynamoLookup: DynamoUpdate): CorporateCodeStatusUpdater = new CorporateCodeStatusUpdater(dynamoLookup)
 
 }
 
-class SetCodeStatus(dynamoUpdate: DynamoUpdate) extends WithLogging {
+class CorporateCodeStatusUpdater(dynamoUpdate: DynamoUpdate) extends WithLogging {
 
-  def apply(code: RedemptionCode, codeStatus: RedemptionTable.AvailableField)(implicit e: ExecutionContext): Future[Unit] =
+  def setStatus(code: RedemptionCode, codeStatus: RedemptionTable.AvailableField)(implicit e: ExecutionContext): Future[Unit] =
     dynamoUpdate.update(code.value, DynamoFieldUpdate(RedemptionTable.AvailableField.name, codeStatus.encoded))
       .withLoggingAsync(s"marked redemption status of $code to $codeStatus")
 
--- a/support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
+++ b/support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
@@ -1,22 +1,25 @@
 package com.gu.support.workers.lambdas
 
+import java.util.UUID
+
 import cats.data.EitherT
 import cats.implicits._
 import com.amazonaws.services.lambda.runtime.Context
 import com.gu.config.Configuration
+import com.gu.i18n.Country
 import com.gu.monitoring.SafeLogger
 import com.gu.services.{ServiceProvider, Services}
 import com.gu.support.catalog
 import com.gu.support.catalog._
-import com.gu.support.config.{TouchPointEnvironments, ZuoraConfig}
-import com.gu.support.promotions.{PromoError, PromotionService}
+import com.gu.support.config.{TouchPointEnvironment, TouchPointEnvironments, ZuoraConfig}
+import com.gu.support.promotions.{PromoCode, PromoError, PromotionService}
 import com.gu.support.redemption._
 import com.gu.support.redemption.corporate._
 import com.gu.support.redemption.gifting.GiftCodeValidator
 import com.gu.support.redemption.gifting.generator.GiftCodeGeneratorService
 import com.gu.support.redemptions.RedemptionData
 import com.gu.support.workers._
-import com.gu.support.workers.lambdas.CreateZuoraSubscription.createSubscription
+import com.gu.support.workers.lambdas.CreateZuoraSubscription.{Impure, createSubscription}
 import com.gu.support.workers.lambdas.DigitalSubscriptionGiftRedemption.{maybeDigitalSubscriptionGiftRedemption, redeemGift}
 import com.gu.support.workers.states.{CreateZuoraSubscriptionState, PaymentMethodWithSchedule, SendThankYouEmailState}
 import com.gu.support.zuora.api.ReaderType.Gift
@@ -61,7 +64,7 @@ class CreateZuoraSubscription(servicesProvider: ServiceProvider = ServiceProvide
       case Some(redemptionData) =>
         redeemGift(redemptionData, requestInfo, state, zuoraGiftService, services.catalogService)
       case None =>
-        createSubscription(state, requestInfo, now, today, promotionService, redemptionService, zuoraService, giftCodeGenerator, config)
+        createSubscription(state, requestInfo, config, Impure(now, today, promotionService, redemptionService, zuoraService, giftCodeGenerator))
     }
   }
 }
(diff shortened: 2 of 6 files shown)


support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -70,19 +73,26 @@ object CreateZuoraSubscription {
 
   import com.gu.FutureLogging._
 
-  def createSubscription(
-    state: CreateZuoraSubscriptionState,
-    requestInfo: RequestInfo,
+  case class Impure(
```
johnduffell, line 76: I wasn't sure what to call this, but this function `createSubscription` is the one that I'm going to rethink a lot in part 2.  I added it initially because there was an error about there being more than 8 parameters (which is a fair point)
But I think the refactor proper deserves a separate PR to avoid confusion.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -261,12 +271,14 @@ object CreateZuoraSubscription {
 
 object DigitalSubscriptionGiftRedemption {
 
-  def maybeDigitalSubscriptionGiftRedemption(product: ProductType, paymentMethod: Either[PaymentMethod, RedemptionData]) = {
+  def maybeDigitalSubscriptionGiftRedemption(
+    product: ProductType,
+    paymentMethod: Either[PaymentMethod, RedemptionData]
+  ): Option[RedemptionData] =
```
johnduffell, line 277: scalastyle was complaining it's a public method with no return type, and after adding it was >160 width

support-services/src/main/scala/com/gu/support/redemption/corporate/CorporateCodeStatusUpdater.scala
```diff
@@ -5,15 +5,15 @@ import com.gu.support.redemptions.RedemptionCode
 
 import scala.concurrent.{ExecutionContext, Future}
 
-object SetCodeStatus {
+object CorporateCodeStatusUpdater {
 
-  def withDynamoLookup(dynamoLookup: DynamoUpdate): SetCodeStatus = new SetCodeStatus(dynamoLookup)
+  def withDynamoUpdate(dynamoLookup: DynamoUpdate): CorporateCodeStatusUpdater = new CorporateCodeStatusUpdater(dynamoLookup)
```
rupertbates, line 10:
```suggestion
  def withDynamoUpdate(dynamoUpdate: DynamoUpdate): CorporateCodeStatusUpdater = new CorporateCodeStatusUpdater(dynamoUpdate)
```

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
rupertbates: I'm not a fan of introducing concepts like pure and impure at a class level, it seems quite FP jargony to me and doesn't really say anything about what it contains from an application logic point of view.

An alternative approach could be to group all the services into a wrapper class `SubscriptionServices` or similar and I'd also love to get rid of one of `today` and `now` as it seems redundant to pass both.

johnduffell: agree, in the part 2 refactor I realised it makes more sense for code clarity to group things by their lifecycle than whether they are pure or impure.
So first the things that come from the environment/config that are the same for every execution, and then the things that come from the State/input that vary per "run".
Then you can concentrate on the business logic separately from the plumbing.
I think this is pretty much what you suggested with a new SubscriptionServices class.
Ideally any extra structure should add meaning to the code, rather than obscure it.

Regarding the today and now, I agree and there must be a way to convert them so I'll have a quick look.

johnduffell: I have managed to get rid of the today/now separation.  I have got rid of the Impure class completely and made it normal class parameters.
Have a look and see what you think.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -41,45 +40,39 @@ class CreateZuoraSubscription(servicesProvider: ServiceProvider = ServiceProvide
...
+      new ZuoraSubscriptionCreator(now, services.promotionService, services.redemptionService, services.zuoraService, services.giftCodeGenerator, config)
+    }
 
     maybeDigitalSubscriptionGiftRedemption(state.product, state.paymentMethod) match {
       case Some(redemptionData) =>
-        redeemGift(redemptionData, requestInfo, state, zuoraGiftService, services.catalogService)
+        redeemGift(redemptionData, requestInfo, state, services.zuoraGiftService, services.catalogService)
       case None =>
-        createSubscription(state, requestInfo, config, Impure(now, today, promotionService, redemptionService, zuoraService, giftCodeGenerator))
+        createSubscription.create(state, requestInfo)
```
johnduffell, line 55: it does seem much cleaner in terms of following the logic by getting rid of all the services code from this line.

build.sbt
```diff
@@ -4,7 +4,7 @@ import sbtrelease.ReleaseStateTransformations._
 
 import scala.sys.process._
 
-val scalatest = "org.scalatest" %% "scalatest" % "3.2.1"
+val scalatest = "org.scalatest" %% "scalatest" % "3.2.2"
```
johnduffell, line 7: the fix I made that went into 3.2.1 turned out to break something else which was fixed in 3.2.2.  I was having trouble with it not telling me which test failed, just which sbt project, so I thought bumping it would be a good idea.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -47,42 +46,43 @@ class CreateZuoraSubscription(servicesProvider: ServiceProvider = ServiceProvide
...
-    val today = () => LocalDate.now(DateTimeZone.UTC)
-    val promotionService = services.promotionService
-    val redemptionService = services.redemptionService
-    val zuoraGiftService = services.zuoraGiftService
-    val zuoraService = services.zuoraService
-    val giftCodeGenerator = services.giftCodeGenerator
-    val isTestUser = state.user.isTestUser
-    val config: ZuoraConfig = services.config.zuoraConfigProvider.get(isTestUser)
+
+    val createSubscription = {
```
rupertbates, line 50: What is the thinking behind declaring these vals within a block?

rupertbates (approved):
Just one question but otherwise looks good 👍

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
johnduffell: scoping - I realised there were a load of vals that were only used in the one place so this makes that clear.
It would defintely be more consistent to actually turn it into a separate private function, but that seemed overkill when I actually tried it out.
What you you reckon?

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -47,42 +46,43 @@ class CreateZuoraSubscription(servicesProvider: ServiceProvider = ServiceProvide
...
+      val now = () => DateTime.now(DateTimeZone.UTC)
+      val isTestUser = state.user.isTestUser
+      val config: ZuoraConfig = services.config.zuoraConfigProvider.get(isTestUser)
+      new ZuoraSubscriptionCreator(now, services.promotionService, services.redemptionService, services.zuoraService, services.giftCodeGenerator, config)
+    }
```
johnduffell, line 55: how about this?
```suggestion
    val createSubscription = new ZuoraSubscriptionCreator(
      () => DateTime.now(DateTimeZone.UTC),
      services.promotionService, 
      services.redemptionService, 
      services.zuoraService, 
      services.giftCodeGenerator, 
      services.config.zuoraConfigProvider.get(state.user.isTestUser)
    )
```

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
johnduffell: I had a brain wave - see below https://github.com/guardian/support-frontend/pull/2756/files/d4271139f3be9cddb0bff2d5bfdc2e52077c5051#r501782946

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
rupertbates: I think it is less surprising so 👍

PR #2757, opened 2020-10-07
Check name is not null in contributions thank you page

tomrf1:
The `firstName` field in the redux state is `string | null`.
The fact that flow didn't catch this is pretty shocking. Let's move to typescript!

Seen in sentry:
Screen Shot 2020-10-07 at 07 43 50


TomPretty (approved):
😱 oh gosh, yeah - we need to ditch flow already!

PR #2759, opened 2020-10-07
Open "find your nearest participating retailer" link in new tab

GHaberis:
## Why are you doing this?

The "find your nearest participating retailer" link on https://support.theguardian.com/uk/subscribe/paper takes you off the support site when clicked, it'd be better to open this in a new tab. This PR does this be adding, `target="_blank"` to open the new tab and `rel="noopener noreferrer"` to prevent unsafe linking to cross-origin destinations, as described here: https://web.dev/external-anchors-use-rel-noopener/


rupertbates (commented):
👍

rupertbates (approved):

PR #2760, opened 2020-10-07
Split DS builder into purchase/corp and rationalise parameters (part 2 of 2)

johnduffell:
## Why are you doing this?

I need to make some changes to get the generated redemption code and giftee start/end date through to the email step.  This requires changes to the relevant state and also the gift-purchase and gift-redemption code path of CreateZuoraSubscription lambda.

Before I try to thread things back through, I want to make sure there are no existing warnings or scalastyle errors in the relevant code and the files are not too long or complex so it's easier to get the fields back cleanly through the functions from where they are generated.
 
e.g.
I noticed the CreateZuoraSubscription lambda has a few red scalastyle errors which were fair points.
image

Tested in CODE and working ok for contribtuion & digital pack with correct emails sent.

## Changes

* convert ProductTypeRatePlans from implicit to normal functions
* split CreateZuoraSubscription.scala file, remove buildSubscribeItem etc, buildSubscriptionData
* put parameters that are not from the State (services/config etc) into class level values rather than passing through
* separate digi sub builder into Purchase and Redemption classes


support-models/src/main/scala/com/gu/support/workers/ProductTypeRatePlans.scala
```diff
@@ -1,70 +1,54 @@
 package com.gu.support.workers
 
+import com.gu.support.catalog
 import com.gu.support.catalog.{Product, ProductRatePlan}
 import com.gu.support.config.TouchPointEnvironment
 import com.gu.support.zuora.api.ReaderType
 import com.gu.support.zuora.api.ReaderType.Corporate
 
-
-trait ProductTypeRatePlans[P] {
```
johnduffell, line 9: there seemed to be a lot of boilerplate and implicits here, for something that was only used in 4 places, and each of those could just be a normal function call.

support-models/src/main/scala/com/gu/support/workers/Products.scala
```diff
@@ -20,15 +20,13 @@ sealed trait ProductType {
...
   def describe: String
-  val catalogType: com.gu.support.catalog.Product
 }
 
 case class Contribution(
   amount: BigDecimal,
   currency: Currency,
   billingPeriod: BillingPeriod
 ) extends ProductType {
-  override val catalogType = com.gu.support.catalog.Contribution
```
johnduffell, line 31: this doesn't seem to be used, even by the serialisers.  (I know that because if the amount is missing from a contribution, it gets put through as a digital pack)

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -89,12 +80,19 @@ object CreateZuoraSubscription {
     impure: Impure
   ): Future[HandlerResult[SendThankYouEmailState]] = {
     import impure._
-    val corporateCodeValidator = CorporateCodeValidator.withDynamoLookup(redemptionService)
+    val subscriptionDataBuilder = {
+      val corporateCodeValidator = CorporateCodeValidator.withDynamoLookup(redemptionService)
+      val dsPurchaseBuilder = new DigitalSubscriptionPurchaseBuilder(config.digitalPack, promotionService, giftCodeGenerator, today)
+      val dsRedemptionBuilder = new DigitalSubscriptionCorporateRedemptionBuilder(corporateCodeValidator, today)
+      new SubscriptionBuilder(dsPurchaseBuilder, dsRedemptionBuilder, promotionService, config.contributionConfig _)
+    }
```
johnduffell, line 88: I did try some arbitrary groupings in case classes like I did in the SendThankYouEmail equivalent, but none of them were really clear.
I think splitting between per-request and reusable parameters makes most sense, as then the business decision making logic is not obscured by promoServices and configs.
This also means we are more likely to stay within the 8 parameter limit and 50 lines limit.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -154,106 +152,6 @@ object CreateZuoraSubscription {
       state.acquisitionData
     )
 
-  private def buildSubscribeItem(state: CreateZuoraSubscriptionState, subscriptionData: SubscriptionData): SubscribeItem = {
```
johnduffell, line 157: all these non-product-specific builder functions are moved to a new file SubscriptionData.scala

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -154,106 +152,6 @@ object CreateZuoraSubscription {
...
-      account = buildAccount(state),
-      billToContact = buildContactDetails(state.user, None, state.user.billingAddress),
-      soldToContact = state.user.deliveryAddress map (buildContactDetails(state.user, state.giftRecipient, _, state.user.deliveryInstructions)),
-      paymentMethod = state.paymentMethod.left.toOption,
-      subscriptionData = subscriptionData,
-      subscribeOptions = SubscribeOptions(generateInvoice = billingEnabled, processPayments = billingEnabled)
-    )
-  }
-
-  private def buildSubscriptionData(
```
johnduffell, line 170: this is the function that was originally triggering scalastyle.  this function is moved to SubscriptionBuilder.scala and restructured slightly to use the purchase/corporate builders instead of passing in the SubscriptionPurchase/SubscriptionRedemption class.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -316,12 +214,13 @@ object DigitalSubscriptionGiftRedemption {
     (for {
       ratePlan <- subscription.ratePlans.headOption
       productRatePlan <- catalogService.getProductRatePlanFromId(catalog.DigitalPack, ratePlan.productRatePlanId)
-      newEndDate = LocalDate.now()
+      startDate = LocalDate.now() // TODO needed for thank you email
```
johnduffell, line 217: I have added TODOs where I have identified which values need to end up in the State in the next PR

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionBuilder.scala
```diff
@@ -4,161 +4,47 @@ import java.util.UUID
...
-  codeValidator: CorporateCodeValidator
-) extends SubscriptionPaymentType
+}
 
 object DigitalSubscriptionBuilder {
 
   type BuildResult = EitherT[Future, Either[PromoError, InvalidCode], SubscriptionData]
 
   def build(
+    subscriptionPaymentTypeBuilder: SubscriptionPaymentTypeBuilder,
```
johnduffell, line 34: we lose the pattern match that used to be on line 56-61 and the code is in what used to be the data structure (as it's no longer a data structure)

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionCorporateRedemptionBuilder.scala
```diff
@@ -0,0 +1,60 @@
...
+  codeValidator: CorporateCodeValidator,
+  today: () => LocalDate,
+) {
+
+  def build(
+    redemptionData: RedemptionData,
+    digitalPack: DigitalPack,
+    requestId: UUID,
+    environment: TouchPointEnvironment,
+  )(implicit ec: ExecutionContext): EitherT[Future, InvalidCode, SubscriptionData] = {
```
johnduffell, line 30: this and the below are split out from the old DigitalSubscriptionBuilder object.
The main benefit is they don't need to conform to a common type so we have got rid of the old
`type BuildResult = EitherT[Future, Either[PromoError, InvalidCode], SubscriptionData]`

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/SubscribeItemBuilder.scala
```diff
@@ -0,0 +1,53 @@
+package com.gu.zuora.subscriptionBuilders
```
johnduffell, line 1: 3 related functions, all moved here from the main CreateZuoraSubscription.scala file

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/SubscriptionBuilder.scala
```diff
@@ -0,0 +1,94 @@
...
+          state.promoCode,
+          state.firstDeliveryDate,
+          promotionService,
+          if (state.giftRecipient.isDefined) ReaderType.Gift else ReaderType.Direct,
+          environment
+        ).leftMap(BuildSubscribePromoError)
+        )
+    }
+
+  private def buildDigiSub(
```
johnduffell, line 62: this replaces the generic parts of the DigitalSubscriptionBuilder.scala

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/SubscriptionBuilder.scala
```diff
@@ -0,0 +1,94 @@
...
+case class BuildSubscribeRedemptionError(cause: InvalidCode) extends RuntimeException
+
+class SubscriptionBuilder(
+  digitalSubscriptionPurchaseBuilder: DigitalSubscriptionPurchaseBuilder,
+  digitalSubscriptionCorporateRedemptionBuilder: DigitalSubscriptionCorporateRedemptionBuilder,
+  promotionService: => PromotionService,
+  config: BillingPeriod => ZuoraContributionConfig,
+) {
+
+  def build(
```
johnduffell, line 29: this replaces both the original >50lines function that was in CreateZuoraSubscription.scala originally

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/SubscriptionBuilder.scala
```diff
@@ -0,0 +1,94 @@
...
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
+
+case class BuildSubscribePromoError(cause: PromoError) extends RuntimeException
+
+case class BuildSubscribeRedemptionError(cause: InvalidCode) extends RuntimeException
+
+class SubscriptionBuilder(
+  digitalSubscriptionPurchaseBuilder: DigitalSubscriptionPurchaseBuilder,
```
rupertbates, line 23: It's a bit strange that we now have instantiated classes being passed in for the digi sub builders while all the others are static objects. Why did you choose to move to the instantiated class approach for digi sub?

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionCorporateRedemptionBuilder.scala
rupertbates: 🎉  that was a Frankenstein type!

rupertbates: a Frankentype

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -47,12 +41,12 @@ class CreateZuoraSubscription(servicesProvider: ServiceProvider = ServiceProvide
     services: Services
   ): FutureHandlerResult = {
 
-    val createSubscription = new ZuoraSubscriptionCreator(
+    val createSubscription = ZuoraSubscriptionCreator.construct(
```
rupertbates, line 44: Could use apply here? I know we're avoiding apply but in my mind that's for anything other than constructors

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -105,6 +97,29 @@ class ZuoraSubscriptionCreator(
...
+      now,
+      zuoraService,
+      new SubscriptionBuilder(
+        new DigitalSubscriptionPurchaseBuilder(config.digitalPack, promotionService, giftCodeGenerator, () => now().toLocalDate),
+        new DigitalSubscriptionCorporateRedemptionBuilder(
+          CorporateCodeValidator.withDynamoLookup(redemptionService),
+          () => now().toLocalDate
+        ),
+        promotionService,
+        config.contributionConfig _
```
rupertbates, line 118: Anonymous function convertible to a method value - according to Intellij

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/SubscriptionBuilder.scala
```diff
@@ -0,0 +1,94 @@
...
+  digitalSubscriptionCorporateRedemptionBuilder: DigitalSubscriptionCorporateRedemptionBuilder,
+  promotionService: => PromotionService,
```
rupertbates, line 25:
```suggestion
  promotionService: PromotionService,
```

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/SubscriptionBuilder.scala
```diff
@@ -0,0 +1,94 @@
...
+import com.gu.zuora.subscriptionBuilders.ProductSubscriptionBuilders.buildContributionSubscription
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
+
+case class BuildSubscribePromoError(cause: PromoError) extends RuntimeException
+
+case class BuildSubscribeRedemptionError(cause: InvalidCode) extends RuntimeException
+
+class SubscriptionBuilder(
```
rupertbates, line 22: SubscriptionDataBuilder would be more precise

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/SubscriptionBuilder.scala
```diff
@@ -0,0 +1,94 @@
...
+          promotionService,
+          if (state.giftRecipient.isDefined) ReaderType.Gift else ReaderType.Direct,
+          environment
+        ).leftMap(BuildSubscribePromoError)
+        )
+    }
+
+  private def buildDigiSub(
+    state: CreateZuoraSubscriptionState,
+    d: DigitalPack,
```
rupertbates, line 64: I'd call this parameter `digitalPack`

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -67,26 +61,24 @@ class CreateZuoraSubscription(servicesProvider: ServiceProvider = ServiceProvide
 
 class ZuoraSubscriptionCreator(
   now: () => DateTime,
-  promotionService: PromotionService,
-  redemptionService: DynamoLookup with DynamoUpdate,
   zuoraService: ZuoraSubscribeService,
-  giftCodeGenerator: GiftCodeGeneratorService,
-  config: ZuoraConfig,
+  subscriptionDataBuilder: SubscriptionBuilder,
```
rupertbates, line 65:
```suggestion
  subscriptionDataBuilder: SubscriptionDataBuilder,
```

I'd already put this suggestion below but for consistency...

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/SubscriptionBuilder.scala
johnduffell: discussed on a call, summary is
I didn't want to concern the decision logic with non State specific dependencies e.g. config 
if we pass them in all the way in/out as parameters or as a product type we have to explicit about every single one which makes the signatures more complicated.
if we pass as a partially applied/class, it's only relevant what it can do, not what it contains - the things that went into it are encapsulated/private and can't be used directly.  This is likely to produce a simpler signature.  Although of course functions are not serialisable and there is no way to get back the original components by default.

rupertbates (approved):
Good work 👍

PR #2763, opened 2020-10-08
fix list of requestable image sizes on home delivery tab

i-hardy:
## Why are you doing this?

Fixes an issue where we are requesting an image at a size which we do not have a crop for.

## Changes

* Adjust the requested sizes of the image on the home delivery tab to match the sizes we actually have


rupertbates (approved):
👍

PR #2764, opened 2020-10-09
Fix off by one error in the support reminder dates

TomPretty:
## Why are you doing this?

Javascript dates run from 0-11, but we want dates that run from 1-12.


tomrf1 (approved):

paulbrown1982 (approved):
👍  🤦

PR #2769, opened 2020-10-13
Add new Digital Sub gift purchase fields into the checkout

rupertbates:
## Why are you doing this?

Following on from #2730 we need to pass two new fields through to the backend when completing a Digital Subscription gift purchase - `giftMessage` and `giftDeliveryDate`.

I have also changed the way that the gift delivery date is formatted in the Redux data store from `dd\MM\yyyy` to `yyyy-MM-dd`. This is consistent with how all other dates are represented in the checkouts and is the expected format for the deserialiser code in the backend.

[**Trello Card**](https://trello.com/c/7N8rZbzM/3396-pass-gift-message-gift-delivery-date-to-server-on-gift-purchase)

### DIFF
--- a/support-frontend/assets/helpers/formValidation.js
+++ b/support-frontend/assets/helpers/formValidation.js
@@ -42,8 +42,7 @@ export const checkEmail: (string | null) => boolean = input => isNotEmpty(input)
 export const checkOptionalEmail: (string | null) => boolean = input => isEmpty(input) || isValidEmail(input);
 
 export const checkGiftStartDate: (string | null) => boolean = (rawDate) => {
-  const dateArray = rawDate ? rawDate.split('/') : null;
-  const date = dateArray ? new Date(`${dateArray[1]} ${dateArray[0]} ${dateArray[2]}`) : null;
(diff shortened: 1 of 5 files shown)


lucymonie (approved):
I got nervous about these changes because of the browser issues we had with dates. So I went on your branch and ran the code in FireFox so I can affirm that it's all good!

~🍰 ~

Sadly I have to cross out the cake as something is now broken with the form-level errors. The `checkGiftStartDate` function uses the giftDeliveryDate to determine if there should be an error on submission of the form.

lucymonie (changes requested):
However, something is now broken with the form-level errors. The checkGiftStartDate function uses the giftDeliveryDate to determine if there should be an error on submission of the form.

rupertbates:
> However, something is now broken with the form-level errors. The checkGiftStartDate function uses the giftDeliveryDate to determine if there should be an error on submission of the form.

Thanks for spotting this, should be fixed now

johnduffell (approved):

lucymonie (approved):

PR #2770, opened 2020-10-13
pass through generated gift code into emails

johnduffell:
Reminder: support-frontend will need to be deployed before support-workers and all current executions completed, as the analyticsInfo is not cross compatible.

## Why are you doing this?

As part of the corporate sub gifting work, we generate a gift code as part of the purchase process, and we want to send that in emails to both the purchaser and recipient.

This PR passes the code from the CreateZuoraSubscription step to the email step, and uses it instead of a placeholder when generating the email fields.

TODO: Test in CODE!

## Changes

I have added inline PR comments to help understand each change


support-frontend/app/services/stepfunctions/SupportWorkersClient.scala
```diff
@@ -154,7 +154,7 @@ class SupportWorkersClient(
         user = user,
         giftRecipient = giftRecipient,
         product = request.body.product,
-        paymentProvider = PaymentProvider.fromPaymentFields(request.body.paymentFields.left.toOption),
+        analyticsInfo = AnalyticsInfo(giftRecipient.isDefined, PaymentProvider.fromPaymentFields(request.body.paymentFields.left.toOption)),
```
johnduffell, line 157: * add new isGiftPurchase field so that sendacquisition/failure lambda can still recognise gifts of any type
* wrap paymentProvider and isGiftPurchase fields that are only used by acquisition/failure lambda in dedicated AnalyticsInfo object for clarity and ease of use

I had to keep it all on one line as this method is already at the 50 line limit 😢

support-models/src/main/scala/com/gu/support/workers/GiftRecipient.scala
```diff
@@ -54,6 +55,52 @@ object GiftRecipient {
...
+
+  def apply(value: String): Option[GiftCode] =
+    Some(value)
+      .filter(_.matches(raw"""gd(03|06|12)-[a-km-z02-9]{8}"""))
+      .map(new GiftCode(_))
+
+  implicit val e1 = Encoder.encodeString.contramap[GiftCode](_.value)
+  implicit val d1 = Decoder.decodeString.emap(GiftCode(_).toRight("invalid gift code"))
+
+}
```
johnduffell, line 70: GiftCode object moved to models, so we can serialise it between steps

support-models/src/main/scala/com/gu/support/workers/GiftRecipient.scala
```diff
@@ -54,6 +55,53 @@ object GiftRecipient {
...
+  def asDigiSub: Option[DigitalSubGiftRecipientWithCode] = this match { case a: DigitalSubGiftRecipientWithCode => Some(a); case _ => None }
+  def giftRecipient: GiftRecipient
+}
+object GiftRecipientAndMaybeCode {
+
+  case class NonDigitalSubGiftRecipient(giftRecipient: WeeklyGiftRecipient) extends GiftRecipientAndMaybeCode
+  case class DigitalSubGiftRecipientWithCode(
+    giftRecipient: DigitalSubGiftRecipient,
+    giftCode: GiftCode
+  ) extends GiftRecipientAndMaybeCode
```
johnduffell, line 82: new ADT to hold the gift code - only if it's a DS gift.

I wasn't totally sure about the naming. non digital is basically weekly, but for the purposes of this code we only care that it's not digital.

support-workers/src/main/scala/com/gu/WithLoggingSugar.scala
```diff
@@ -5,12 +5,12 @@ import com.typesafe.scalalogging.LazyLogging
...
 import scala.concurrent.Future
 
-object FutureLogging extends LazyLogging {
+object WithLoggingSugar extends LazyLogging {
 
   implicit class LogImplicitFuture[A](op: Future[A]) {
 
     // this is just a handy method to add logging to the end of any for comprehension
-    def withLogging(message: String): Future[A] = {
+    def withEventualLogging(message: String): Future[A] = {
```
johnduffell, line 13: I wanted to add a withLogging, and it needed to be clear to the compiler and us which was being called, so I distinguished the naming.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -75,24 +78,36 @@ class ZuoraSubscriptionCreator(
 
     val environment = TouchPointEnvironments.fromStage(Configuration.stage, state.user.isTestUser)
 
+    val maybeGiftRecipWithCode = state.giftRecipient.map(maybeAddCode(state.product.billingPeriod, _))
+      .withLogging("gift recipient with code")
```
johnduffell, line 82: now we generate the code up front rather than in the builder.  This does complicate this top level a bit, but the alternative of changing the type returned by the builder so we could pick out the code, was not turning out to be nice.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -248,9 +265,9 @@ object DigitalSubscriptionGiftRedemption {
         HandlerResult(SendThankYouEmailState(
           state.requestId,
           state.user,
-          state.giftRecipient,
+          None,
```
johnduffell, line 268: there's never a gift recipient for a redemption actually.

I was having a think about approaches for the State.. some thoughts below that we can think about collectively:

The one size fits all model makes sense if most execution types are similar, but now that we have more disparate things to do, some of these are starting to make less sense in some situations than others.  I can see this on line 275 just below here:
image

which we could indeed add more options, but maybe they should not exist at all if it's a redemption (ie use a sum type rather than a product type)
I have slightly pushed the sum type approach with recent PRs, making the GiftRecipient type GWGift+DSGift, 

Downside: does complicate the serialisers a bit.  And makes the raw JSON diverge a bit from the scala representation.
Upside: But it does give us the type safety and lack of having a lot of Options or empty strings to fill in.  And also means we don't have to keep getting things out of Options if they "must" exist in reality.

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionPurchaseBuilder.scala
```diff
@@ -48,35 +40,29 @@ class DigitalSubscriptionPurchaseBuilder(
...
-    )
-
-    val withRedemptionCode = addRedemptionCodeIfGift(digitalPack.readerType, giftCodeGenerator.generateCode(billingPeriod), subscriptionData)
-    val withPromoApplied = applyPromoCodeIfPresent(promotionService, maybePromoCode, country, productRatePlanId, withRedemptionCode)
-
-    EitherT.fromEither[Future](withPromoApplied)
-  }
+    val maybeRedemptionCode = (digitalPack.readerType, maybeGiftCode) match {
+      case (Gift, Some(giftCode)) => Some(giftCode)
+      case (Gift, _) | (_, Some(_)) => None // FIXME coding error - possible unredeemable sub
```
johnduffell, line 45: I think we should probably explode at this point - the types we have defined are not good enough to make this impossible, but it's impossible

support-models/src/main/scala/com/gu/support/workers/GiftRecipient.scala
```diff
@@ -54,6 +55,53 @@ object GiftRecipient {
...
+    Some(value)
+      .filter(_.matches(raw"""gd(03|06|12)-[a-km-z02-9]{8}"""))
+      .map(new GiftCode(_))
+
+  implicit val e1 = Encoder.encodeString.contramap[GiftCode](_.value)
+  implicit val d1 = Decoder.decodeString.emap(GiftCode(_).toRight("invalid gift code"))
+
+}
+
+sealed trait GiftRecipientAndMaybeCode {
```
johnduffell, line 72: I am going to rename this, something like `SendThankYouEmailGiftPurchase` is a more meaningful name, so far it's called by the "what" rather than the "why"

johnduffell:
```
GiftPurchase
+- WeeklyGiftPurchase
+- DigitalSubGiftPurchase
```

support-workers/src/test/scala/com/gu/support/workers/errors/StripeErrorsSpec.scala
```diff
@@ -89,13 +89,13 @@ class StripeErrorsSpec extends AsyncLambdaSpec with MockWebServerCreator with Mo
...
-      errorJson <- decode[ErrorJson](executionError.Cause).toOption
-      stripeError <- decode[StripeError](errorJson.errorMessage).toOption
+      wrapper <- decode[JsonWrapper](JsonFixtures.cardDeclinedJsonStripe)
+      executionError <- wrapper.error.toRight("wrapper.error was empty")
+      errorJson <- decode[ErrorJson](executionError.Cause)
+      stripeError <- decode[StripeError](errorJson.errorMessage)
     } yield stripeError
 
-    stripeError.get.code should be(Some("card_declined"))
+    stripeError.flatMap(_.code.toRight("no code")) should be(Right("card_declined"))
```
johnduffell, line 98: this test was failing with None.get, I had to change to Either to get the real error, and it is worth keeping it like that

support-workers/src/test/scala/com/gu/support/workers/integration/DigitalSubscriptionGiftRedemptionIntegrationSpec.scala
```diff
@@ -19,47 +19,48 @@ import scala.concurrent.Future
...
 @IntegrationTest
 class DigitalSubscriptionGiftRedemptionIntegrationSpec extends AsyncLambdaSpec with MockContext {
-  val createZuoraHelper = new CreateZuoraSubscriptionHelper()
-
-  val createSubRequestId = UUID.randomUUID()
-  val redeemSubRequestId = UUID.randomUUID()
-  val giftCode = new GiftCodeGeneratorService().generateCode(Annual)
-  SafeLogger.info(s"Gift code = ${giftCode}")
-  val mockCodeGenerator = mock[GiftCodeGeneratorService]
-  when(mockCodeGenerator.generateCode(any[BillingPeriod])).thenReturn(giftCode)
```
johnduffell, line 29: all this was being run when the test run initialised so the log statements were appearing at the top of the logs.

support-models/src/main/scala/com/gu/support/workers/GiftRecipient.scala
```diff
@@ -54,6 +55,53 @@ object GiftRecipient {
...
+  def asDigiSub: Option[DigitalSubGiftPurchase] = this match { case a: DigitalSubGiftPurchase => Some(a); case _ => None }
+  def giftRecipient: GiftRecipient
+}
+object GiftPurchase {
+
+  case class WeeklyGiftPurchase(giftRecipient: WeeklyGiftRecipient) extends GiftPurchase
+  case class DigitalSubGiftPurchase(
+    giftRecipient: DigitalSubGiftRecipient,
+    giftCode: GiftCode
+  ) extends GiftPurchase
```
johnduffell, line 82: new ADT to represent a product specific purchased gift, for DS only, so far it also includes the gift code.
In the next PR it will also get the `last_redemption_date` field added.

support-models/src/main/scala/com/gu/support/workers/GiftRecipient.scala
```diff
@@ -54,6 +55,53 @@ object GiftRecipient {
...
+object GiftPurchase {
+
+  case class WeeklyGiftPurchase(giftRecipient: WeeklyGiftRecipient) extends GiftPurchase
+  case class DigitalSubGiftPurchase(
+    giftRecipient: DigitalSubGiftRecipient,
+    giftCode: GiftCode,
+  ) extends GiftPurchase
+
+  val circeDiscriminator = new CirceDiscriminator("giftPurchaseType")
+  private val discriminatorNonDigiSub = "NonDigiSub"
```
rupertbates, line 85: Can we make this `Weekly` and the other one `DigitalSub` or `DigitalSubscription`? 
I'd prefer to avoid abbreviations where possible or at least restrict them to one standard one abbreviation rather than mixing between DigitalSubscription/DigitalSub/DigiSub

johnduffell: good idea, it's good to be consistent, or drop words completely, rather than random contractions

johnduffell: I should also change the existing one too on line 39 of this same file
https://github.com/guardian/support-frontend/blob/d150942050cec1f665e022eb3dae994e337ac7fe/support-models/src/main/scala/com/gu/support/workers/GiftRecipient.scala#L38

support-models/src/main/scala/com/gu/support/workers/GiftRecipient.scala
rupertbates: I wonder whether we actually need a separate type for `GiftCode` we could just use `RedemptionCode` for everything.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -75,24 +78,36 @@ class ZuoraSubscriptionCreator(
 
     val environment = TouchPointEnvironments.fromStage(Configuration.stage, state.user.isTestUser)
 
+    val maybeGiftRecipWithCode = state.giftRecipient.map(maybeAddCode(state.product.billingPeriod, _))
```
rupertbates, line 81:
```suggestion
    val maybeGiftRecipientWithCode = state.giftRecipient.map(maybeAddCode(state.product.billingPeriod, _))
```

support-models/src/main/scala/com/gu/support/workers/GiftRecipient.scala
johnduffell: good point, I did vaguely wonder that at one point.  I will look and decide if to put that change in this PR.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
johnduffell: I need to avoid the habit of shortening things.  I will have a look through the whole PR for these kind of things.

johnduffell: `maybeGiftRecipientWithCode` is a bit low level anyway, the type is already `GiftRecipient` in reflection of that.
Also I will rename `maybeAddCode` to `addGiftPurchaseDetails` as follows
```scala
    val maybeGiftPurchase = state.giftRecipient.map(addGiftPurchaseDetails(state.product.billingPeriod, _))
```

support-models/src/main/scala/com/gu/support/workers/GiftRecipient.scala
johnduffell: I had a look, I think there is a case for sharing the data types in terms of reducing the number of classes, on the other side there is no real logical benefit of them being compatible, and some potential confusion as they are treated totally separately in the functions.
redemption code is for validating the untrusted input from the user, whereas the gift code is the generated one that will be written into zuora.
I propose rename `GiftCode` to `GeneratedGiftCode` in this PR, then it's clear the purpose.  

But overall it's not something I feel strongly about keeping separately so happy to change it again or in future.

support-models/src/main/scala/com/gu/support/workers/GiftRecipient.scala
```diff
@@ -35,22 +37,70 @@ object GiftRecipient {
...
+    this match { case a: DigitalSubscriptionGiftPurchase => Some(a); case _ => None }
+  def giftRecipient: GiftRecipient
+}
+object GiftPurchase {
+
+  case class WeeklyGiftPurchase(giftRecipient: WeeklyGiftRecipient) extends GiftPurchase
+  case class DigitalSubscriptionGiftPurchase(
+    giftRecipient: DigitalSubscriptionGiftRecipient,
+    giftCode: GeneratedGiftCode,
+  ) extends GiftPurchase
```
johnduffell, line 84: (re-made comment as it had become "outdated")

new ADT to represent a product specific purchased gift

for DS gifts, so far it also includes the gift code. In the next PR it will also get the last_redemption_date field added.
for GW gifts there is no additional information generated in the subscribe step over a basic GW

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -75,24 +78,37 @@ class ZuoraSubscriptionCreator(
 
     val environment = TouchPointEnvironments.fromStage(Configuration.stage, state.user.isTestUser)
 
+    val maybeGiftPurchase = state.giftRecipient.map(addGiftPurchaseDetails(state.product.billingPeriod, _))
+      .withLogging("gift recipient with code")
```
johnduffell, line 82: (comment re-added as it was "outdated")

now we generate the code up front rather than in the builder. This does complicate this top level a bit, but the alternative of changing the type returned by the builder so we could pick out the code, was not turning out to be nice.

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionPurchaseBuilder.scala
```diff
@@ -2,81 +2,63 @@ package com.gu.zuora.subscriptionBuilders
...
-    val (contractAcceptanceDelay, autoRenew, initialTerm) =
-      if (digitalPack.readerType == ReaderType.Gift)
-        (0, false, GiftCodeValidator.expirationTimeInMonths + 1)
-      else
-        (config.defaultFreeTrialPeriod + config.paymentGracePeriod, true, 12)
+
+    val (contractAcceptanceDelay, autoRenew, initialTerm, maybeRedemptionCode) =
+      (digitalPack.readerType, maybeGiftCode) match {
+        case (Gift, Some(giftCode)) => (0, false, GiftCodeValidator.expirationTimeInMonths + 1, Some(giftCode))
+        case (Gift, _) | (_, Some(_)) => throw new RuntimeException("coding error - possible unredeemable sub")
```
johnduffell, line 38: we need to explode at this point - the types we have defined are not good enough to make this impossible, but it's impossible

support-models/src/main/scala/com/gu/support/workers/GiftRecipient.scala
```diff
@@ -35,22 +37,70 @@ object GiftRecipient {
...
+  implicit val encoder: Encoder[GiftPurchase] = Encoder.instance {
+    case c: WeeklyGiftPurchase => c.asJson
+    case c: DigitalSubscriptionGiftPurchase => c.asJson
+  }
+
+  implicit val decoder =
+    circeDiscriminator.decode[GiftPurchase](Map(
+      discriminatorWeekly -> Decoder[WeeklyGiftPurchase],
+      discriminatorDigitalSubscription -> Decoder[DigitalSubscriptionGiftPurchase]
     ))
```
johnduffell, line 104: I don't really get why it should be so much harder to serialise a sum type than a product type.  I can see why, given how they are implemented in circe, but I feel like there's so much boilerplate.  It almost puts me off modelling things this way.
Should we consider the provided `withDiscriminator` ? https://circe.github.io/circe/codecs/adt.html#the-future

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
rupertbates: Happy to discuss this further offline

rupertbates (approved):
👍

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
johnduffell: conveniently, next in the list for post standup pairing time is you and me!

johnduffell:
deployed together with https://github.com/guardian/support-frontend/pull/2773 as I retargeted that one to main

PR #2773, opened 2020-10-14
Add last redemption date to both emails sent at DS gift purchase time

johnduffell:
## Why are you doing this?

This PR adds last redemption date field.  The field name is agree with @GHaberis .  This lets the recipient know they have a time limit on redemption which is an important part of T&C.

[**Trello Card**](https://trello.com/c/RvP8A1nw/3192-gifting-email-api-triggers-for-the-gifter-giftee-emails-generated-95)

## Changes

* see inline comments where necessary


support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -99,12 +99,13 @@ class ZuoraSubscriptionCreator(
...
   }
 
-  private def maybeAddCode(billingPeriod: BillingPeriod, giftRecipient: GiftRecipient) =
+  private def maybeAddCode(billingPeriod: BillingPeriod, giftRecipient: GiftRecipient, today: () => LocalDate) =
     giftRecipient match {
       case digitalSubGiftRecipient: DigitalSubGiftRecipient =>
         val giftCode = giftCodeGeneratorService.generateCode(billingPeriod)
           .withLogging("Generated code for Digital Subscription gift")
-        DigitalSubGiftPurchase(digitalSubGiftRecipient, giftCode)
+        val lastRedemptionDate = today().plusMonths(GiftCodeValidator.expirationTimeInMonths).minusDays(1)
```
johnduffell, line 107: I took off one day just because a zuora 12 month sub runs from e.g. 2020-10-14 00:00 to 2021-10-14 00:00 but if we add 12 months to today, it looks like we include today.  Also it's a nice fudge factor.

rupertbates (approved):
👍

johnduffell:
tested in CODE DS annual and monthly, plus recurreing 7gbp per month contribution. Including thank you emails OK.
seems to be working okay in prod.

support-models/src/main/scala/com/gu/support/workers/states/SendAcquisitionEventState.scala
```diff
@@ -12,7 +12,6 @@ import org.joda.time.LocalDate
 
 trait SendAcquisitionEventState extends FailureHandlerState {
   def requestId: UUID
-  def giftRecipient: Option[GiftRecipient]
```
johnduffell, line 15: I should have replaced this with `def analyticsInfo: AnalyticsInfo`

johnduffell: oh ignore that, it is actually in `MinimalFailureHandlerState` now

PR #2776, opened 2020-10-16
Enable stripe fraud detection only if we have required consents

TomPretty:
## Why are you doing this?
We want to only load stripe with fraud detection if we have the required user consents (fraud detections drops a couple of cookies) as per [this card](https://trello.com/c/Owz6zkeD/2278-stripefraud-ab-test-remove-the-test-and-make-the-fraud-tool-available-only-when-we-get-the-users-consent).

### DIFF
--- a/support-frontend/assets/helpers/stripe.js
+++ b/support-frontend/assets/helpers/stripe.js
@@ -3,27 +3,9 @@
 import { useEffect, useState } from 'react';
 import { loadStripe, type Stripe as StripeSDK } from '@stripe/stripe-js/pure';
 import { onConsentChange } from '@guardian/consent-management-platform';
-import { logException } from 'helpers/logger';
 import { type PaymentMethod, Stripe } from 'helpers/paymentMethods';
 import type { IsoCountry } from 'helpers/internationalisation/country';
 import type { ContributionType } from 'helpers/contributions';
-
-
-const setupStripe = (setStripeHasLoaded: () => void) => {
-  if (window.Stripe) {
-    setStripeHasLoaded();
-  } else {
-    const htmlElement = document.getElementById('stripe-js');
-    if (htmlElement !== null) {
-      htmlElement.addEventListener(
-        'load',
-        setStripeHasLoaded,
-      );
-    } else {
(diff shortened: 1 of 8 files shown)


support-frontend/assets/pages/contributions-landing/components/StripeCardForm/StripeCardFormContainer.jsx
```diff
@@ -31,31 +26,14 @@ type PropTypes = {|
...
-
   const stripeAccount = stripeAccountForContributionType[props.contributionType];
   const stripeKey = getStripeKey(
     stripeAccount,
     props.country,
     props.isTestUser,
   );
 
-  React.useEffect(() => {
-    if (!props.stripeHasLoaded) {
```
tomrf1, line 48: there's some more code cleanup to be done here
`stripeHasLoaded` and `setStripeHasLoaded` are no longer needed. These can also be removed from redux.
Also the `setupStripe` function in `stripe.js`

TomPretty: Yeah good point - I'll get on that now

TomPretty: Removed!

tomrf1 (approved):
great!

support-frontend/assets/helpers/stripe.js
```diff
@@ -51,9 +38,47 @@ function getStripeKey(stripeAccount: StripeAccount, country: IsoCountry, isTestU
...
+    },
+    [stripeAccount],
+  );
+
+  return stripeObjects;
+};
+
 
 export {
-  setupStripe,
```
paulbrown1982, line 56: Are we sure only the contributions checkout was dependent on this export?

TomPretty: Yeah it was just being used in the `StripeCardFormContainer` and `StripePaymentRequestButtonContainer` components, which now both use `useStripeObjects` instead

paulbrown1982 (approved):
👍  LGTM

PR #2781, opened 2020-10-19
Fix issue with image container overlapping the 'Choose a subscription' button on the support page

i-hardy:
## Why are you doing this?

This is a bug fix for an issue where the container for the image in the 'Subscribe' section on the support showcase page was still present in the DOM at mobile breakpoints, even when the image is not being shown, and was overlapping the button and breaking its interactivity.
image

## Changes

* Make the image container `display: none` below the tablet breakpoint

### DIFF
--- a/support-frontend/assets/pages/showcase/showcase.scss
+++ b/support-frontend/assets/pages/showcase/showcase.scss
@@ -90,7 +90,10 @@
         right: 0;
         height: 100%;
         float: right;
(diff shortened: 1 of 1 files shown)


tomrf1 (approved):
👍

johnduffell (approved):
so quick! thanks 🥇 
I have added the image to the description for clarity

PR #2783, opened 2020-10-19
Don't send an acquisition event for gift redemptions

rupertbates:
## Why are you doing this?

We don't want to send the standard acquisition event when a user redeems a gift subscription. This is because we have already sent this event when the gift was originally bought so to send it again would be a duplication.

[**Trello Card**](https://trello.com/c/FfFZ8IEk/3322-update-sendacquisitionevent-lambda-to-ignore-redemptions-1u-25)

### DIFF
--- a/support-workers/src/main/scala/com/gu/support/workers/lambdas/SendAcquisitionEvent.scala
+++ b/support-workers/src/main/scala/com/gu/support/workers/lambdas/SendAcquisitionEvent.scala
@@ -16,6 +16,7 @@ import com.gu.support.catalog.{Contribution => _, DigitalPack => _, Paper => _, 
 import com.gu.support.promotions.DefaultPromotions
 import com.gu.support.workers._
 import com.gu.support.workers.states.SendAcquisitionEventState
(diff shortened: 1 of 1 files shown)


support-workers/src/main/scala/com/gu/support/workers/lambdas/SendAcquisitionEvent.scala
```diff
@@ -56,7 +57,18 @@ class SendAcquisitionEvent(serviceProvider: ServiceProvider = ServiceProvider)
       )
     )
 
-    // Throw any error in the EitherT monad so that in can be processed by ErrorHandler.handleException
+    state.product match {
+      case d:DigitalPack if d.readerType == Gift && state.paymentOrRedemptionData.isRight =>
```
johnduffell, line 61:
```suggestion
      case d: DigitalPack if d.readerType == Gift && state.paymentOrRedemptionData.isRight =>
```

johnduffell (approved):
simple but effective

PR #2788, opened 2020-10-21
Two small improvements to the SendAcquisitionEvent lambda

rupertbates:
## Why are you doing this?

There are a couple of small improvements in this PR

- Add a `name` val to the `PaymentProvider` class rather than relying on `toString` as this is in line with our coding guidelines
- Use this `PaymentProvider` class when firing the `paymentSuccess` Cloudwatch metric rather than the identically named class from Ophan. This allows us to avoid having to deal with an Option and also makes it easier to remove the Thrift dependency in future.

### DIFF
--- a/support-workers/src/main/scala/com/gu/support/workers/lambdas/SendAcquisitionEvent.scala
+++ b/support-workers/src/main/scala/com/gu/support/workers/lambdas/SendAcquisitionEvent.scala
@@ -58,7 +58,7 @@ class SendAcquisitionEvent(serviceProvider: ServiceProvider = ServiceProvider)
     )
 
     state.product match {
-      case d:DigitalPack if d.readerType == Gift && state.paymentOrRedemptionData.isRight =>
+      case d: DigitalPack if d.readerType == Gift && state.paymentOrRedemptionData.isRight =>
         // We don't want to send an acquisition event for Digital subscription gift redemptions as we have already done so on purchase
         Future.successful(HandlerResult((), requestInfo))
       case _ =>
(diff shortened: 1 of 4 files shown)


johnduffell:
> Add a name val to the PaymentProvider class rather than relying on toString as this is in line with our coding guidelines

I agree it is a good idea, so I will add it to the coding standards doc https://docs.google.com/document/d/1vRIYAU2OnRkHzf4dLCuzxiA-3HsKh1PDmkiLc-1GmfQ/edit#

support-workers/src/main/scala/com/gu/support/workers/lambdas/SendAcquisitionEvent.scala
```diff
@@ -68,19 +68,21 @@ class SendAcquisitionEvent(serviceProvider: ServiceProvider = ServiceProvider)
...
     ).fold(
       errors => throw AnalyticsServiceErrorList(errors),
       _ => HandlerResult((), requestInfo)
     )
+  }
 
-    val maybePaymentProvider = state.paymentOrRedemptionData.left.toOption.map(_.paymentMethod).map(paymentProviderFromPaymentMethod)
-    val cloudwatchEvent = paymentSuccessRequest(Configuration.stage, maybePaymentProvider, state.product)
+  def sendPaymentSuccessMetric(state: SendAcquisitionEventState) = {
+    // Used for Cloudwatch alarms: https://github.com/guardian/support-frontend/blob/920e638c35430cc260acdb1878f37bffa1d12fae/support-workers/cloud-formation/src/templates/cfn-template.yaml#L210
```
johnduffell, line 83: good comment

johnduffell (approved):
great, I had previously noticed the thrift/ophan dependency but didn't actually change it so thanks for doing that 👏

PR #2792, opened 2020-10-21
Use sum types (sealed trait) in email/analytics state

johnduffell:
# Why are you doing this?

This is a preparatory PR for passing the start and end date through to the email, when someone has redeemed a gift subscription.

I have refactored so that the SendThankYouEmailState is now a sealed trait rather than a case class.  This means that there is one subclass for each type of checkout, with appropriate fields.  So one per product, and for DS, there are several further subclasses also covering gifting and corporate.

Unfortunately it is a big change, so far I only applied it to the specific zuora->email state.

There are also a few other loose ends that could be pulled at, I have tried to draw a line under it at this stage, but there is some low level resultant duplication and mis-structuring for example in the PaperFieldsGenerator the code to look up promotion codes that could be addressed in future.

The important thing is to get the idea and naming correct, and the concepts clear.  But of course all comments accepted regardless!

## Changes

* see inline comments

TESTING in CODE correct emails received (not a test user):
- GW gift DONE (cc payemnt)
- GW - DONE (dd)
- paper - DONE ( dd)
- DS direct DONE(dd)
- recurring contribution DONE (card payment)


support-models/src/main/scala/com/gu/support/workers/GiftRecipient.scala
```diff
@@ -39,10 +42,8 @@ object GiftRecipient {
   private val discriminatorWeekly = "Weekly"
   private val discriminatorDigitalSubscription = "DigitalSubscription"
 
-  implicit val e1 = circeDiscriminator.add(deriveEncoder[WeeklyGiftRecipient], discriminatorWeekly)
-  implicit val d1 = deriveDecoder[WeeklyGiftRecipient]
-  implicit val e2 = circeDiscriminator.add(deriveEncoder[DigitalSubscriptionGiftRecipient], discriminatorDigitalSubscription)
-  implicit val d2 = deriveDecoder[DigitalSubscriptionGiftRecipient]
```
johnduffell, line 45: small improvement to `CirceDiscriminator` to save a line of code for each subtype

support-models/src/main/scala/com/gu/support/workers/GiftRecipient.scala
```diff
@@ -70,41 +71,6 @@ object GeneratedGiftCode {
 
 }
 
-sealed trait GiftPurchase {
```
johnduffell, line 73: this trait has been made redundant as it's all included in the new ProductTypeCreated type

support-models/src/main/scala/com/gu/support/workers/ProductTypeRatePlans.scala
```diff
@@ -8,14 +8,6 @@ import com.gu.support.zuora.api.ReaderType.Corporate
 
 object ProductTypeRatePlans {
 
-  def productTypeRatePlan(product: ProductType, environment: TouchPointEnvironment, readerType: ReaderType): Option[ProductRatePlan[Product]] =
-    product match {
-      case d: DigitalPack => digitalRatePlan(d, environment)
-      case p: Paper => paperRatePlan(p, environment)
-      case w: GuardianWeekly => weeklyRatePlan(w, environment, readerType)
-      case _ => None
-    }
```
johnduffell, line 17: this is no longer needed, as we always already know the exact product

support-models/src/main/scala/com/gu/support/workers/states/SendAcquisitionEventState.scala
```diff
@@ -10,24 +10,18 @@ import io.circe.Decoder
 import io.circe.generic.semiauto.deriveDecoder
 import org.joda.time.LocalDate
 
-trait SendAcquisitionEventState extends FailureHandlerState {
+trait SendAcquisitionEventState extends StepFunctionUserState {
```
johnduffell, line 13: SendAcquisitionEvent was never a state that could go to the failure state, so this type was too limiting

support-models/src/main/scala/com/gu/support/workers/states/SendAcquisitionEventState.scala
```diff
@@ -10,24 +10,18 @@ import io.circe.Decoder
...
   def requestId: UUID
-  def product: ProductType
-  def paymentOrRedemptionData: Either[PaymentMethodWithSchedule, RedemptionData]
-  def firstDeliveryDate: Option[LocalDate]
-  def promoCode: Option[PromoCode]
```
johnduffell, line 18: these are all inside appropriate subclasses of `ProductTypeCreated`

support-models/src/main/scala/com/gu/support/workers/states/SendAcquisitionEventState.scala
```diff
@@ -10,24 +10,18 @@ import io.circe.Decoder
...
+  def productTypeCreated: ProductTypeCreated
   def acquisitionData: Option[AcquisitionData]
+  def analyticsInfo: AnalyticsInfo
```
johnduffell, line 17: this used to be pulled in from FailureHandlerState

support-models/src/main/scala/com/gu/support/workers/states/SendThankYouEmailState.scala
```diff
@@ -2,36 +2,139 @@ package com.gu.support.workers.states
...
-  giftPurchase: Option[GiftPurchase],
-  product: ProductType,
   analyticsInfo: AnalyticsInfo,
-  paymentOrRedemptionData: Either[PaymentMethodWithSchedule, RedemptionData],
-  firstDeliveryDate: Option[LocalDate],
-  promoCode: Option[PromoCode],
+  productTypeCreated: ProductTypeCreated,
   salesForceContact: SalesforceContactRecord,
-  accountNumber: String,
-  subscriptionNumber: String,
```
johnduffell, line 22: the removed fields are now under the right subclasses of `ProductTypeCreated`

support-models/src/main/scala/com/gu/support/workers/states/SendThankYouEmailState.scala
```diff
@@ -2,36 +2,139 @@ package com.gu.support.workers.states
...
   acquisitionData: Option[AcquisitionData]
 ) extends SendAcquisitionEventState
 
+case class PurchaseInfo(
+  paymentMethod: PaymentMethod,
+  paymentSchedule: PaymentSchedule,
+  promoCode: Option[PromoCode],
+  accountNumber: String,
+  subscriptionNumber: String,
+)
```
johnduffell, line 32: just grouping a bunch of things you always have in the state for a purchase (in the existing model).  Although it probably should be refined later as necessary e.g. contribution doesn't have a promo code, or care about the payment schedule.

support-models/src/main/scala/com/gu/support/workers/states/SendThankYouEmailState.scala
```diff
@@ -2,36 +2,139 @@ package com.gu.support.workers.states
...
+      lastRedemptionDate: LocalDate,
+      purchaseInfo: PurchaseInfo,
+    ) extends DigitalSubscriptionCreated with PurchaseCreated
+
+    case class DigitalSubscriptionCorporateRedemptionCreated(
+      product: DigitalPack,
+      subscriptionNumber: String,
+    ) extends DigitalSubscriptionCreated
+
+    case class DigitalSubscriptionGiftRedemptionCreated( //tbc
```
johnduffell, line 76: this is where the new fields will be added for start/end date of the gift.  (Start date would be today, and end date would be either 3 or 12 months forward)

support-models/src/main/scala/com/gu/support/workers/states/SendThankYouEmailState.scala
```diff
@@ -2,36 +2,139 @@ package com.gu.support.workers.states
...
+    circeDiscriminator.decode[ProductTypeCreated](Map(
+      discriminatorContribution -> Decoder[ContributionCreated],
+      discriminatorDigitalSubscriptionDirectPurchase -> Decoder[DigitalSubscriptionDirectPurchaseCreated],
+      discriminatorDigitalSubscriptionGiftPurchase -> Decoder[DigitalSubscriptionGiftPurchaseCreated],
+      discriminatorDigitalSubscriptionCorporateRedemption -> Decoder[DigitalSubscriptionCorporateRedemptionCreated],
+      discriminatorDigitalSubscriptionGiftRedemption -> Decoder[DigitalSubscriptionGiftRedemptionCreated],
+      discriminatorPaper -> Decoder[PaperCreated],
+      discriminatorGuardianWeekly -> Decoder[GuardianWeeklyCreated],
+    ))
+}
```
johnduffell, line 133: there is a bit much copy and paste here for my liking.  we could get the compiler to wire some of it up correctly for us, along these lines:
(but if we might switch to circe-extras (which uses macros and does more automatically) then it may not be worth playing too much.)
```scala
  val circeDiscriminator = new CirceDiscriminator("productTypeCreatedType")

  implicit val discriminatorContribution = Discriminator[ContributionCreated]("Contribution")
  implicit val discriminatorDigitalSubscriptionDirectPurchase = Discriminator[DigitalSubscriptionDirectPurchaseCreated]("DigitalSubscriptionDirectPurchase")
  implicit val discriminatorDigitalSubscriptionGiftPurchase = Discriminator[DigitalSubscriptionDirectPurchaseCreated]("DigitalSubscriptionGiftPurchase")
  implicit val discriminatorDigitalSubscriptionCorporateRedemption = Discriminator[DigitalSubscriptionCorporateRedemptionCreated]("DigitalSubscriptionCorporateRedemption")
  implicit val discriminatorDigitalSubscriptionGiftRedemption = Discriminator[DigitalSubscriptionGiftRedemptionCreated]("DigitalSubscriptionGiftRedemption")
  implicit val discriminatorPaper = Discriminator[PaperCreated]("Paper")
  implicit val discriminatorGuardianWeekly = Discriminator[GuardianWeeklyCreated]("GuardianWeekly")

  implicit val encoder: Encoder[ProductTypeCreated] = Encoder.instance {
    case c: ContributionCreated => c.asJson
    case c: DigitalSubscriptionDirectPurchaseCreated => c.asJson
    case c: DigitalSubscriptionGiftPurchaseCreated => c.asJson
    case c: DigitalSubscriptionCorporateRedemptionCreated => c.asJson
    case c: DigitalSubscriptionGiftRedemptionCreated => c.asJson
    case c: PaperCreated => c.asJson
    case c: GuardianWeeklyCreated => c.asJson
  }

  implicit val decoder =
    circeDiscriminator.decode[ProductTypeCreated](List(
      DiscriminatedDecoder[ContributionCreated],
      DiscriminatedDecoder[DigitalSubscriptionDirectPurchaseCreated],
      DiscriminatedDecoder[DigitalSubscriptionGiftPurchaseCreated],
      DiscriminatedDecoder[DigitalSubscriptionCorporateRedemptionCreated],
      DiscriminatedDecoder[DigitalSubscriptionGiftRedemptionCreated],
      DiscriminatedDecoder[PaperCreated],
      DiscriminatedDecoder[GuardianWeeklyCreated],
    ))
}
```

support-models/src/main/scala/com/gu/support/workers/states/SendThankYouEmailState.scala
```diff
@@ -2,36 +2,139 @@ package com.gu.support.workers.states
...
+
 import com.gu.support.encoding.Codec
 import com.gu.support.encoding.Codec._
 
 object SendThankYouEmailState {
-  import com.gu.support.encoding.CustomCodecs._
   implicit val codec: Codec[SendThankYouEmailState] = deriveCodec
 }
-
-case class PaymentMethodWithSchedule(paymentMethod: PaymentMethod, paymentSchedule: PaymentSchedule)
```
johnduffell, line 34: this is flattened into the paymentInfo class now

support-workers/src/main/scala/com/gu/emailservices/ContributionEmailFields.scala
```diff
@@ -1,52 +1,62 @@
...
 
-  def build(
-    allProductsEmailFields: AllProductsEmailFields,
-    created: DateTime,
-    amount: BigDecimal,
-    paymentMethod: PaymentMethod
-  ): EmailFields = {
-    import allProductsEmailFields._
+class ContributionEmailFields(
+  getMandate: String => Future[Option[String]],
```
johnduffell, line 12: before, the getMandate was done at the top level.  But now we have different products, some of which don't have payment methods.
The low intervention solution would be to still run it at the top level by making the pattern match more comprehensive.
Since by the time we actually need the value, we already know whether it's DD, I decided to push it down to the point of need.  Then we only concern ourselves about it in one place, however we do need to pass in the mandate service, and we do need to return a `Future` (or at least `F` where `F` is a `Functor`)

There is undoubtedly an opprtunity to make getMandate into a proper Service, and probably reduce the duplication a bit between cont/DS/paper+weekly.  I will see whether to do it in this PR or save for later.

support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
```diff
@@ -112,61 +115,64 @@ object DigitalSubscriptionEmailAttributes {
...
       last_redemption_date = formatDate(giftPurchase.lastRedemptionDate),
     ))
 
-  private def giftPurchaserConfirmation(paymentMethodWithSchedule: PaymentMethodWithSchedule, giftPurchase: DigitalSubscriptionGiftPurchase) = {
+  private def giftPurchaserConfirmation(giftPurchase: DigitalSubscriptionGiftPurchaseCreated)(implicit ec: ExecutionContext) = {
     import giftPurchase._
+    import purchaseInfo._
+
+    val promotion = paperFieldsGenerator.getAppliedPromotion(
+      promoCode,
```
johnduffell, line 171: I'm not sure if gift purchases can have a promo code actually?

support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
```diff
@@ -192,38 +198,53 @@ class DigitalPackEmailFields(
...
-    ))
+  private def corpRedemption(subscriptionNumber: String) =
+    wrap("digipack-corporate-redemption", directOrCorpFields("Group subscription", subscriptionNumber))
+
+  private def directThankYou(directPurchase: DigitalSubscriptionDirectPurchaseCreated)(implicit ec: ExecutionContext) = {
+
+    val promotion = paperFieldsGenerator.getAppliedPromotion(
+      directPurchase.purchaseInfo.promoCode,
+      user.billingAddress.country,
+      ProductTypeRatePlans.digitalRatePlan(directPurchase.product, touchPointEnvironment).map(_.id).getOrElse("")
```
johnduffell, line 209: this `""` was already there but it can probably be ironed out one day

support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
```diff
@@ -112,61 +115,64 @@ object DigitalSubscriptionEmailAttributes {
 
 }
 class DigitalPackEmailFields(
```
johnduffell, line 117: this file hasn't really changed, it's just shuffled a bit.  (In another PR) it should probably have a separate file for each type of digital subscription execution rather than such a big class.

support-workers/src/main/scala/com/gu/emailservices/EmailFields.scala
```diff
@@ -40,17 +40,3 @@ case class EmailFields(
...
-  currency: Currency,
-  sfContactId: SfContactId,
-  directDebitMandateId: Option[String],
-)
-
-case class SubscriptionEmailFields(
-  allProductsEmailFields: AllProductsEmailFields,
-  subscriptionNumber: String,
-  promotion: Option[Promotion] = None
-)
```
johnduffell, line 56: these are no longer needed as the new type is a far neater way of gathering the appropriate values together.

support-workers/src/main/scala/com/gu/emailservices/GuardianWeeklyEmailFields.scala
```diff
@@ -1,40 +1,43 @@
...
-      promotion,
-      giftRecipient
-    ) ++ additionalFields
+    val giftRecipientFields = gw.giftRecipient.toList.flatMap(
+      recipient =>
+        List(
+          "giftee_first_name" -> recipient.firstName,
+          "giftee_last_name" -> recipient.lastName,
+        )
+    )
```
johnduffell, line 31: for some reason gifts were being handled by the `PaperFieldsGenerator.fieldsFor` even though they only exist for GW.  I have moved it out here instead as the giftRecipient field no longer exists on paper subs.

support-workers/src/main/scala/com/gu/emailservices/PaperFieldsGenerator.scala
```diff
@@ -1,51 +1,70 @@
...
+  def getAppliedPromotion(
+    maybePromoCode: Option[PromoCode],
+    country: Country,
+    productRatePlanId: ProductRatePlanId
+  ): Option[Promotion] =
+    for {
+      promoCode <- maybePromoCode
+      promotionWithCode <- promotionService.findPromotion(promoCode).toOption
+      validPromotion <- promotionService.validatePromotion(promotionWithCode, country, productRatePlanId, isRenewal = false).toOption
+    } yield validPromotion.promotion
```
johnduffell, line 26: this probably shouldn't really be here in the long term as promos also apply to digital subs.

support-workers/src/main/scala/com/gu/emailservices/PaperFieldsGenerator.scala
```diff
@@ -1,51 +1,70 @@
...
-    firstDeliveryDate: Option[LocalDate],
-    currency: Currency,
-    paymentMethodWithSchedule: PaymentMethodWithSchedule,
-    directDebitMandateId: Option[String],
-    promotion: Option[Promotion],
-    giftRecipient: Option[GiftRecipient]
-  ): List[(String, String)] = {
+    productRatePlanId: Option[ProductRatePlanId],
+    fixedTerm: Boolean,
+    firstDeliveryDate: LocalDate
```
johnduffell, line 34: ideally the structure of the Fields Generators should generally reflect the structure of the new types.  Given this code is special to paper+GW there shouldn't really be still 6 parameters to this method, everything should have arrived in one class. since this is a bit lower level this maybe a todo for another PR, depending how the test changes look.

support-workers/src/main/scala/com/gu/emailservices/PaperFieldsGenerator.scala
```diff
@@ -1,51 +1,70 @@
...
 
     val deliveryAddressFields = getAddressFields(user)
 
-    val giftRecipientFields = giftRecipient.map(
-      recipient =>
-        List(
-          "giftee_first_name" -> recipient.firstName,
-          "giftee_last_name" -> recipient.lastName,
-        )
-    ).getOrElse(Nil)
```
johnduffell, line 35: gifting is GW and not paper so moved out.

support-workers/src/main/scala/com/gu/emailservices/SubscriptionEmailFieldHelpers.scala
```diff
@@ -30,14 +30,13 @@ object SubscriptionEmailFieldHelpers {
 
   def describe(
     paymentSchedule: PaymentSchedule,
-    billingPeriod: BillingPeriod,
-    currency: Currency,
+    product: ProductType,
     promotion: Option[Promotion],
-    isGift: Boolean = false
+    fixedTerm: Boolean = false
```
johnduffell, line 35: this should be `fixedTerm`, because it determines the wording of the message saying how much you'll pay and when.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -162,27 +170,51 @@ object ZuoraSubscriptionCreator {
     }
   }
 
+  // scalastyle:off cyclomatic.complexity
```
johnduffell, line 173: the pattern match was too big for scalastyle

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -162,27 +170,51 @@ object ZuoraSubscriptionCreator {
...
+            case product: GuardianWeekly =>
+              GuardianWeeklyCreated(product, state.giftRecipient.map(_.asWeekly.get), purchaseInfo, state.firstDeliveryDate.get)
+          }
+        case _: Redemption =>
+          state.product match {
+            case product: DigitalPack if product.readerType == ReaderType.Corporate =>
+              DigitalSubscriptionCorporateRedemptionCreated(product, subscriptionNumber.value)
+            case product: DigitalPack if product.readerType == ReaderType.Gift =>
+              DigitalSubscriptionGiftRedemptionCreated(product/*tbc*/)
+            case _ => throw new RuntimeException("could not create value state")
```
johnduffell, line 209: this would be redeeming other products than digi, or other readertypes than corp/gift

support-workers/src/main/scala/com/gu/support/workers/lambdas/SendAcquisitionEvent.scala
```diff
@@ -211,10 +224,18 @@ object SendAcquisitionEvent {
           if (stateAndInfo.requestInfo.accountExists) Some("REUSED_EXISTING_PAYMENT_METHOD") else None,
           if (isSixForSix(stateAndInfo)) Some("guardian-weekly-six-for-six") else None,
           if (stateAndInfo.state.analyticsInfo.isGiftPurchase) Some("gift-subscription") else None,
-          stateAndInfo.state.paymentOrRedemptionData.right.map(_ => "corporate-subscription").toOption
+          stateAndInfo.state.productTypeCreated match {
+            case _: DigitalSubscriptionCorporateRedemptionCreated => Some("corporate-subscription")
+            case _ => None
+          }
```
johnduffell, line 230: all of these 2 line pattern matches can be reduced in boilerplate by adding e.g `def asDigitalSubscriptionCorporateRedemptionCreated: Option[DigitalSubscriptionCorporateRedemptionCreated]` or `def is DigitalSubscriptionCorporateRedemptionCreated: Boolean` function on the type, but there's a tradeoff in terms of clutter.
Unless you know a simpler way?

support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
johnduffell: (previously was on line 123 of SendThankYouEmail.scala 
https://github.com/guardian/support-frontend/pull/2792/files#diff-9c81d24a8e7c52abfde88fe9288edc673a3fe9ddc9fb734ae7b31bac675a3a6eL123 )

support-workers/src/main/scala/com/gu/support/workers/lambdas/SendAcquisitionEvent.scala
johnduffell: I raised this on scala channel, and there are a few points thanks to Adam
- we are moving the product type to the type level wheras it was more at the value level before. This allows the compiler to check a lot more in terms of grouping of possible values (this was the aim of the PR).  This also means pattern match is the basic tool to get back to the value level, rather than if statements.
- in scala standard library they added a lot of built in functions to get back to the value level e.g. `.left.toOption`
- we could add similar functions here `.as[Blah]` but the question would be is that abstraction worth learning over and above the standard matching, for the benefit it gives.
- it would be possible to do something generic with ClassTag thanks to Mario (and even generic implicit class), but it may not be the best way.

support-workers/src/main/scala/com/gu/emailservices/PaperFieldsGenerator.scala
```diff
@@ -1,51 +1,71 @@
...
-    firstDeliveryDate: Option[LocalDate],
-    currency: Currency,
-    paymentMethodWithSchedule: PaymentMethodWithSchedule,
-    directDebitMandateId: Option[String],
-    promotion: Option[Promotion],
-    giftRecipient: Option[GiftRecipient]
-  ): List[(String, String)] = {
+    productRatePlanId: Option[ProductRatePlanId],
+    fixedTerm: Boolean,
+    firstDeliveryDate: LocalDate,
```
johnduffell, line 34: ideally the structure of the Fields Generators should generally reflect the structure of the new types. Given this code is special to paper+GW there shouldn't really be still 6 parameters to this method, everything should have arrived in one class. since this is a bit lower level this maybe a todo for another PR, depending how the test changes look.

support-workers/src/main/scala/com/gu/support/workers/lambdas/SendAcquisitionEvent.scala
johnduffell: this line could be slightly simplified to but I haven't thought whether it's a good idea yet.  Maybe one for a later PR.

```suggestion
          stateAndInfo.state.productTypeCreated.collect { case _: DigitalSubscriptionCorporateRedemptionCreated => "corporate-subscription" }
```
we would also add this collect function to the trait
```scala
sealed trait ProductTypeCreated {
  def product: ProductType
  def collect[T](f: PartialFunction[ProductTypeCreated, T]): Option[T] = f.lift(this)
}
```

support-models/src/main/scala/com/gu/support/workers/GiftRecipient.scala
```diff
@@ -34,26 +39,11 @@ object GiftRecipient {
...
-  implicit val d2 = deriveDecoder[DigitalSubscriptionGiftRecipient]
-
-  implicit val encoder: Encoder[GiftRecipient] = Encoder.instance {
-    case c: WeeklyGiftRecipient => c.asJson
-    case c: DigitalSubscriptionGiftRecipient => c.asJson
-  }
+  val discriminatedType = new DiscriminatedType[GiftRecipient]("giftRecipientType")
+  implicit val weeklyCodec = discriminatedType.variant[WeeklyGiftRecipient]("Weekly")
+  implicit val dsCodec = discriminatedType.variant[DigitalSubscriptionGiftRecipient]("DigitalSubscription")
+  implicit val codec = discriminatedType.codec(List(weeklyCodec, dsCodec))
```
johnduffell, line 45: if we standardise the discriminator field name (e.g. `"_type"`) then we could make this simpler
```suggestion
  implicit val weeklyCodec = deriveVariantCodec[WeeklyGiftRecipient]("Weekly")
  implicit val dsCodec = deriveVariantCodec[DigitalSubscriptionGiftRecipient]("DigitalSubscription")
  implicit val codec = deriveTopLevelCodec[GiftRecipient](List(weeklyCodec, dsCodec))
```

support-models/src/main/scala/com/gu/support/workers/GiftRecipient.scala
```diff
@@ -70,53 +60,48 @@ object GeneratedGiftCode {
...
-
-  implicit val e1 = circeDiscriminator.add(deriveEncoder[WeeklyGiftPurchase], discriminatorWeekly)
-  implicit val d1 = deriveDecoder[WeeklyGiftPurchase]
-  implicit val e2 = circeDiscriminator.add(deriveEncoder[DigitalSubscriptionGiftPurchase], discriminatorDigitalSubscription)
-  implicit val d2 = deriveDecoder[DigitalSubscriptionGiftPurchase]
-
-  implicit val encoder: Encoder[GiftPurchase] = Encoder.instance {
-    case c: WeeklyGiftPurchase => c.asJson
-    case c: DigitalSubscriptionGiftPurchase => c.asJson
+class DiscriminatedType[TOPLEVEL](discriminatorFieldName: String) {
```
rupertbates, line 63: Move this to it's own file maybe? I'd also put it in the `com.gu.support.encoding` package

support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
rupertbates: They can't currently

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -162,27 +175,51 @@ object ZuoraSubscriptionCreator {
...
     SendThankYouEmailState(
       requestId = state.requestId,
       user = state.user,
-      giftPurchase = maybeGiftPurchase,
-      product = state.product,
       analyticsInfo = state.analyticsInfo,
-      paymentOrRedemptionData = paymentOrRedemptionData,
-      firstDeliveryDate = state.firstDeliveryDate,
-      promoCode = state.promoCode,
+      productTypeCreated = paymentOrRedemptionData match {
```
rupertbates, line 192: This is quite nasty, I think you need to break the nested matches out into separate functions if we're going to go down this route

johnduffell: good point and even scalastyle was complaining about it

support-models/src/main/scala/com/gu/support/workers/GiftRecipient.scala
johnduffell: good spot 👍

johnduffell:
- renamed ProductTypeCreated to SendThankYouEmailProductSpecificState
- flattened PurchaseInfo out completely
- removed PurchaseCreated trait completely
- broke up create zuora subscription state construction logic
- move discriminatedType class into correct package

- `SendThankYouEmailState` contains a `SendThankYouEmailProductSpecificState`
- `SendThankYouEmailProductSpecificState` has subclasses as follows:
image
- each of the subclasses contains a `ProductType`.  This could be further flattened out if necessary.

johnduffell:
I am just seeing if i get flatten together SendThankYouEmailState and SendThankYouEmailProductSpecificState

support-models/src/main/scala/com/gu/support/workers/states/SendThankYouEmailState.scala
```diff
@@ -1,37 +1,100 @@
...
+    user: User,
+    salesForceContact: SalesforceContactRecord,
+    product: DigitalPack,
+    giftRecipient: DigitalSubscriptionGiftRecipient,
+    giftCode: GeneratedGiftCode,
+    lastRedemptionDate: LocalDate,
+    paymentMethod: PaymentMethod,
+    paymentSchedule: PaymentSchedule,
+    accountNumber: String,
+  ) extends SendThankYouEmailDigitalSubscriptionState
```
johnduffell, line 49: following the earlier chat, we may need `promoCode: Option[PromoCode],` in here too

support-models/src/main/scala/com/gu/support/workers/GiftRecipient.scala
rupertbates: We could default it to `_type` and leave the option to override if necessary?

support-frontend/app/services/stepfunctions/SupportWorkersClient.scala
```diff
@@ -154,7 +154,7 @@ class SupportWorkersClient(
         user = user,
         giftRecipient = giftRecipient,
         product = request.body.product,
-        analyticsInfo = AnalyticsInfo(giftRecipient.isDefined, PaymentProvider.fromPaymentFields(request.body.paymentFields.left.toOption)),
+        analyticsInfo = AnalyticsInfo(giftRecipient.isDefined, PaymentProvider.fromPaymentFields(request.body.paymentFields.left.toOption), requestId),
```
rupertbates, line 157: Why do we need to pass `requestId` into `AnalyticsInfo` now? Isn't this unnecessary duplication as we have it at the top level as well?

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -96,31 +100,39 @@ class ZuoraSubscriptionCreator(
...
+        Some(DigitalSubscriptionGiftCreationDetails(digitalSubscriptionGiftRecipient, giftCode, lastRedemptionDate))
+      case _ =>
+        None
     }
 
 }
 
 object ZuoraSubscriptionCreator {
 
+  case class DigitalSubscriptionGiftCreationDetails(
```
rupertbates, line 121: What do you think about calling this `DigitalSubscriptionGiftPurchaseDetails` I think it would be good to standardise on calling the two types of gift checkout 'Purchase' & 'Redemption'. Creation doesn't seem very specific to me.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -199,6 +189,106 @@ object ZuoraSubscriptionCreator {
...
+        paymentOrRedemptionData,
+        maybeDigitalSubscriptionGiftCreationDetails
+      ),
+      acquisitionData = state.acquisitionData
+    )
+
+  val Purchase = Left
+  type Redemption = Right[PaymentMethodWithSchedule, RedemptionData]
+
+  // scalastyle:off cyclomatic.complexity
```
rupertbates, line 214: Do we still need to disable scalastyle here?

support-frontend/app/services/stepfunctions/SupportWorkersClient.scala
johnduffell: just because I didn't want to go back and alter the other places that use it.  Although it may not be many places in reality - I will check.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
johnduffell: will do, yeah I agree standardisation is good here

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
johnduffell: yes otherwise it complains 
image
although I can't work out why 13 and not 8.  I can only assume it's how it generates the bytecode for the match statement.
There is a "countCases: Y/N" option in scalastyle, but it's only  in the dev branch http://www.scalastyle.org/rules-dev.html#parameters-13:~:text=countCases,-Count

johnduffell: re 13 vs 8, looks like scalastyle bugs really.

looks like it searches for IF, WHILE, DO, FOR, CASE and || / &&
https://github.com/scalastyle/scalastyle/blob/ec14399543d2d5ccf93c3713aa5df21793844791/src/main/scala/org/scalastyle/scalariform/CyclomaticComplexityChecker.scala#L43
So I guess the if guard on each case has confused it into thinking it's another code path.
Then it adds one for the default case (which doesn't exist as it's covered by the `case _`.

support-frontend/app/services/stepfunctions/SupportWorkersClient.scala
johnduffell: I think the primary reason for this might have gone away - I will have a look.

johnduffell: removed, it was totally unnecessary in the end

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
johnduffell: fixed this now.

support-models/src/main/scala/com/gu/support/workers/states/SendThankYouEmailState.scala
johnduffell: added this in

support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
johnduffell: they are now added to the code just in case we need them in future.

support-models/src/main/scala/com/gu/support/workers/states/SendThankYouEmailState.scala
johnduffell: actually now using DiscriminatedType is much simpler than even this example

support-models/src/main/scala/com/gu/support/encoding/DiscriminatedType.scala
```diff
@@ -0,0 +1,121 @@
...
+
+  def getSingle[A](list: List[A], message: String): A = list match {
+    case Nil => throw new RuntimeException(s"no codec for: $message")
+    case a :: Nil => a
+    case many => throw new RuntimeException(s"duplicate codec for: $message: $many")
+  }
+
+  def encoder[TOPLEVEL](allCodecs: List[VariantEncoder[_ <: TOPLEVEL]]): Encoder[TOPLEVEL] = {
+    if (allCodecs.groupBy(_.discriminatorFieldName).size != 1)
+      throw new RuntimeException("coding error - non matching discriminator names")
```
johnduffell, line 26: after the refactor, a few things that would have been compile errors are now runtime exceptions

support-models/src/main/scala/com/gu/support/workers/states/SendThankYouEmailState.scala
```diff
@@ -1,37 +1,100 @@
...
+
+  implicit val codec = DiscriminatedType.codec[SendThankYouEmailState](List(
+    Variant.codec[SendThankYouEmailContributionState]("Contribution", "productType"),
+    Variant.codec[SendThankYouEmailDigitalSubscriptionDirectPurchaseState]("DigitalSubscriptionDirectPurchase", "productType"),
+    Variant.codec[SendThankYouEmailDigitalSubscriptionGiftPurchaseState]("DigitalSubscriptionGiftPurchase", "productType"),
+    Variant.codec[SendThankYouEmailDigitalSubscriptionCorporateRedemptionState]("DigitalSubscriptionCorporateRedemption", "productType"),
+    Variant.codec[SendThankYouEmailDigitalSubscriptionGiftRedemptionState]("DigitalSubscriptionGiftRedemption", "productType"),
+    Variant.codec[SendThankYouEmailPaperState]("Paper", "productType"),
+    Variant.codec[SendThankYouEmailGuardianWeeklyState]("GuardianWeekly", "productType"),
+  ))
```
johnduffell, line 98: if we want a particular discriminator, we now have to specify it correctly on every line, to avoid a runtime exeception

support-models/src/main/scala/com/gu/support/workers/GiftRecipient.scala
johnduffell: If you look at this specific commit, you can see the changes resulting from this simplification
https://github.com/guardian/support-frontend/pull/2792/commits/4612be809e1c9e333383b60fddf4988801149384
not sure if it's better or not, it's moved some compiler errors to runtime, but equally, it's only using normal functions and classes rather than nested classes.
Let me know what you think @rupertbates  - since it's the last commit it's easy to decide whether or not to include it.

support-models/src/main/scala/com/gu/support/encoding/DiscriminatedType.scala
rupertbates: Could we not have just changed the constructor of the `DiscriminatedType` class to have a default? Eg.
```
class DiscriminatedType[TOPLEVEL](discriminatorFieldName: String = "__type")
```

johnduffell: we could, but I don't think we actually want or need a default, that would only be a tradeoff to get better syntax.

my original point was purely about the syntax looking a bit weird, but we could only improve it by standardising the discriminator field name.

rupertbates (approved):
👍 good stuff, we should continue trying to use this approach in the other lambdas

PR #2796, opened 2020-10-23
email the gift redeemer the dates

johnduffell:
## Why are you doing this?

When someone redeems a gift subscription, we want to email them the start and end date of the fixed term.

This PR passes the dates through to the email step and includes them in the email structure.

[**Trello Card**](https://trello.com/c/QLWyYjK8/3403-gifting-add-gift-start-end-date-to-the-giftee-redemption-email)

## Changes

* see inline comments


support-services/src/main/scala/com/gu/support/redemption/CodeStatus.scala
```diff
@@ -18,13 +18,18 @@ case object CodeExpired extends InvalidCode("code_expired")
 
 case object InvalidReaderType extends InvalidCode("invalid_reader_type")
 
-object ValidGiftCode { val clientCode = "valid_gift_code"}
+object UnredeemedGiftCode { val clientCode = "valid_gift_code"}
```
rupertbates, line 21: Can you explain the changes in these classes please?

johnduffell: for the non controversial changes, regardless of whether it has been redeemed in this request or is about to be redeemed, we need the subscription id so we can check the billing period so we know when the subscription should start and end.  This is so we can pass on the start and end date to the next step to send in the email.

Due to the above, I changed `CodeRedeemedInThisRequest` to be a case class containing the subscription id.

---

When rejigging the redeem gift functionality to support the above.   I needed to reuse most of the `redeemInZuora` code, so I decided to call that function even if it was already redeemed in the request.

That function will decide whether to write to zuora based on whether it's `UnredeemedGiftCode` or `CodeRedeemedInThisRequest`


The other option (which I'm starting to think would be better) would be to keep the ADT flatter as before, and flatten out the code to match, so basically 3 cases in redeemGift
- invalid: throw it out
- valid: get-dates-etc THEN update in zuora
- redeemed in this request: get-dates-etc

This would slightly complicate the code as we'd need a case class to hold the "get dates etc" return values, but at least the more widely used `CodeStatus` data structure would be simpler and better separation between the cases.

rupertbates (approved):
Nice! 👍

support-services/src/main/scala/com/gu/support/redemption/CodeStatus.scala
johnduffell: I changed to the "slightly complicate the code" option above, rather than complicating the model in commit "change gift redemption logic so it doesn't rely on a common supertype"

PR #2799, opened 2020-10-28
add scheduled emails facility for giftee notification emails

johnduffell:
Depends on https://github.com/guardian/membership-workflow/pull/273

## Why are you doing this?

As part of digi sub gifting we want to send the gift recipient notification email on a specific date.

To do this we will ask braze (via membership-workflow, via contribution-thanks SQS queue) to send the email on a future date.

This PR adds the field to the SQS payload.

https://docs.google.com/document/d/1dGh1dMWsAXCS-BdRY8Qp9mxlbM2b3Bjg7273_DjxzU0/edit?usp=sharing

[**Trello Card**](https://trello.com/c/Qhk9ihTL/3279-add-scheduled-email-functionality-to-membership-workflow)


support-models/src/main/scala/com/gu/salesforce/Salesforce.scala
```diff
@@ -202,7 +202,8 @@ object Salesforce {
   }
 
   object Authentication {
-    implicit val codec: Codec[Authentication] = deriveCodec
+    implicit val decodeDateTime: Decoder[DateTime] = Decoder.decodeLong.map(new DateTime(_))
+    implicit val codec: Decoder[Authentication] = deriveDecoder
```
johnduffell, line 206:
```
{
    "access_token": "asjhdgksajhdgaj",
    "instance_url": "https://aksjhdg--DEV.my.salesforce.com",
    "id": "https://test.salesforce.com/id/kjasghdkjashdjkasd/askjdhakjsdhkajshdka",
    "token_type": "Bearer",
    "issued_at": "1603976679073",
    "signature": "dskjfhlskjdfhlksdjfhlksdjfhlksdjhf"
}
```

support-models/src/main/scala/com/gu/support/promotions/Promotion.scala
```diff
@@ -27,7 +27,7 @@ case class Promotion(
 }
 
 object Promotion {
-  import com.gu.support.encoding.CustomCodecs.{decodeDateTime, encodeDateTime}
+  implicit val decodeDateTime: Decoder[DateTime] = Decoder.decodeString.map(DateTime.parse)
```
johnduffell, line 30: image

support-models/src/test/scala/com/gu/support/promotions/SerialisationSpec.scala
```diff
@@ -111,9 +115,11 @@ class SerialisationSpec extends AsyncFlatSpec with SerialisationTestHelpers with
       appliesTo,
       "",
       Map[Channel, Set[PromoCode]]("" -> Set()),
-      DateTime.now(),
+      DateTime.now(ISOChronology.getInstanceUTC),
```
johnduffell, line 118: DateTime.now() uses ISO chronology/London by default, but when you parse an ISO date you get the ISO chronology/UTC.  This means it fails the test despite seeming to be the same.

support-models/src/test/scala/com/gu/support/promotions/SerialisationSpec.scala
```diff
@@ -111,9 +115,11 @@ class SerialisationSpec extends AsyncFlatSpec with SerialisationTestHelpers with
...
       None, None, None
     )
+    implicit val encodeDateTime: Encoder[DateTime] = Encoder.encodeString.contramap(ISODateTimeFormat.dateTime().print)
```
johnduffell, line 121: before this change the test was incorrectly encoding as a Long because it had picked up the encoder we use for sending dates to the client side.  Other tests were covering the correct format, and I'm not even sure if this test is useful, but I've fixed it up anyway.

support-services/src/main/scala/com/gu/support/promotions/PromotionTerms.scala
```diff
@@ -18,9 +19,9 @@ case class PromotionTerms(
 
 object PromotionTerms {
 
-  import com.gu.support.encoding.CustomCodecs.{encodeDateTime, decodeDateTime}
+  implicit val encodeDateTime: Encoder[DateTime] = Encoder.encodeLong.contramap(_.getMillis)
```
johnduffell, line 22: client side needs a Long timestamp rather than ISO, because the most reliable way across all browsers is millis since 1970.

support-workers/src/main/scala/com/gu/emailservices/EmailFields.scala
```diff
@@ -4,15 +4,26 @@ import com.gu.i18n.Currency
...
+import com.gu.support.encoding.CustomCodecs._
+import org.joda.time.format.ISODateTimeFormat
 
 case class EmailPayloadContactAttributes(SubscriberAttributes: Map[String, String])
 case class EmailPayloadTo(Address: String, ContactAttributes: EmailPayloadContactAttributes)
-case class EmailPayload(To: EmailPayloadTo, DataExtensionName: String, SfContactId: Option[String], IdentityUserId: Option[String])
+case class EmailPayload(
+  To: EmailPayloadTo,
+  DataExtensionName: String,
+  SfContactId: Option[String], // TODO delete this and make IdentityId non Option
```
johnduffell, line 19: this was a request made by paul brown, that us using SF contact causes difficulties and that we should use identity id where possible (which is all cases for support-workers)

support-workers/src/main/scala/com/gu/emailservices/EmailFields.scala
```diff
@@ -4,15 +4,26 @@ import com.gu.i18n.Currency
...
+  IdentityUserId: Option[String],
+  ScheduledTime: Option[DateTime], // None means immediate
+)
 
 object EmailPayload {
+  implicit val encodeDateTime: Encoder[DateTime] = Encoder.encodeString.contramap(ISODateTimeFormat.dateTime().print)
```
johnduffell, line 25: we want to send the scheduled time to membership-workflow as a nicely formatted date, so that it's easier to human read.

support-models/src/main/scala/com/gu/salesforce/Salesforce.scala
rupertbates: You can import a `Decoder[DateTime]` from `CustomCodecs`

support-frontend/assets/pages/digital-subscription-checkout/components/datePicker/datePickerHelpers.test.js
```diff
@@ -8,53 +8,73 @@ import {
 } from './helpers';
 import { monthText } from 'pages/paper-subscription-checkout/helpers/subsCardDays';
 
+describe('datePickerHelpers', () => {
```
johnduffell, line 11: ignore the changes in this file, they are from merging main into this branch, no idea what happened there.  This PR is already merged: https://github.com/guardian/support-frontend/pull/2812/files

support-workers/src/main/scala/com/gu/emailservices/EmailFields.scala
```diff
@@ -24,19 +35,20 @@ case class EmailFields(
...
         Address = email,
         ContactAttributes = EmailPayloadContactAttributes(SubscriberAttributes = fields.toMap)
       ),
       DataExtensionName = dataExtensionName,
       SfContactId = userId.left.toOption.map(_.id),
-      IdentityUserId = userId.right.toOption.map(_.id)
-    ).asJson.spaces2
-  }
+      IdentityUserId = userId.right.toOption.map(_.id),
+      deliveryDate.map(_.toDateTime(new LocalTime(8, 0), DateTimeZone.UTC)),
```
johnduffell, line 51: this means it will go at 8am UTC on the relevant day.  This is TBC

support-workers/src/main/scala/com/gu/emailservices/EmailFields.scala
```diff
@@ -24,19 +35,20 @@ case class EmailFields(
...
+    ).asJson.printWith(Printer.spaces2.copy(dropNullValues = true))
```
johnduffell, line 52: dropNullValues isn't the default but we use it in some places to prevent None becoming null.  For the payload to membership-workflow it just reduces the clutter.

support-models/src/main/scala/com/gu/salesforce/Salesforce.scala
rupertbates: Could we leave the most commonly used one (probably the long version?) in there and move the other version out to where it's needed? I think it's good to have at least one version in there because it is the obvious place to look for it in the codebase

support-models/src/test/scala/com/gu/support/promotions/SerialisationSpec.scala
rupertbates: Dates - always a problem!

rupertbates (approved):
One question/suggestion about where to put codecs but otherwise 👍

support-models/src/main/scala/com/gu/salesforce/Salesforce.scala
johnduffell: I'm not sure I could work out a most commonly used one, it's always determined based on what you need for that purpose and we use each of the 4 once!  I'd prefer where possible to use ISO most commonly as it's human readable.
- the long encoder is only used once to serialise a promo date to the client side.
- the long decoder is used to decode the salesforce credential expiry time
- the ISO decoder is used to get promo start/end dates from dynamo
- the ISO encoder is used to push the delivery date to membership-workflow (and in a dynamo test)

Actually, I think based on your comments I will actually put both in there, but each in a suitably named nested object.  That means people have to choose explicitly but it's easy to find.   There was somewhere I realised was using the wrong one so it makes sense to be explicit.

The other option is to not serialise DateTime directly, but convert it to a String or Long in the case class.  But I think that will make the case classes less type safe and add boiler plate at the point of use.

rupertbates: I like this approach 👍

PR #2802, opened 2020-10-30
fix IT tests that access zuora UAT

johnduffell:
## Why are you doing this?

Since the recent UAT env refresh, the IT tests have been failing.  This PR fixes them to use some fresh subscriptions.

### DIFF
--- a/support-workers/src/test/scala/com/gu/zuora/Fixtures.scala
+++ b/support-workers/src/test/scala/com/gu/zuora/Fixtures.scala
@@ -13,7 +13,7 @@ import org.joda.time.LocalDate
 
 //noinspection TypeAnnotation
 object Fixtures {
-  val accountNumber = "A00071408"
+  val accountNumber = "A00084679"
 
   val salesforceAccountId = "0013E00001ASmI6QAL"
   val salesforceId = "0033E00001CpBZaQAN"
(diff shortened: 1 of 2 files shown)


david-pepper:
Why are we hardcoding the ids of test data records, rather than creating the records as part of the test itself? This approach would allow for the tests to be more dynamic, and remove the constraint on a static piece of test data being present for said tests. 

Furthermore, this data will be cleared down at some point in the future, and these tests will need to be reconfigured all over again if we continue to use this approach

johnduffell:
totally agree it's a cost - the same applies to the DEV ones too
for repeatbility always better to be self contained
we have to balance that against that
* it will fill up quicker if they create stuff every time
* the tests will be slower and more internal complexity if they are doing extra subscribe calls

johnduffell:
this PR is mainly to get us back where we were, so the alarm isn't shouting every hour!

david-pepper:
> fair points thanks for that, and it is great that this work has flagged up this issue. It would certainly be a useful task to do in the medium term now that we are aware. I would consider doing the changes under pressure of failing tests, but I don't think it's quite simple enough to slip it through at the same time.
> 
> For the immediate issue, let's get this PR reviewed in terms of getting back to where we were at the start of the week so that the tests are working still. Then we can plan something more long term in perhaps in the innovation/enginerring time.

Yep I agree. This PR will fix the failing tests and revert them to a state that is no worse than the state that they were in before they began to fail. Happy to approve on that basis.

david-pepper (approved):

PR #2806, opened 2020-11-03
make the gift date tests more determinstic

johnduffell:

This PR passes in a specific date to the date range code so that the tests don't compare two separately acquired dates.
This means they will not fail occasionally due to a millisecond having passed while the test was running.

```
17:09:00
  FAIL assets/pages/digital-subscription-checkout/components/datePicker/datePickerHelpers.test.js
17:09:00
    ● getRange › should be able to get the latest available date from now
17:09:00
  17:09:00
      expect(received).toEqual(expected) // deep equality
17:09:00
  17:09:00
      Expected: 2021-01-06T17:08:59.555Z
17:09:00
      Received: 2021-01-06T17:08:59.554Z
```

https://trello.com/c/4l2cBx52/3393-fix-non-deterministic-test

## Why are you doing this?

Unreliable tests can cause us to have to rebuild and can delay changes going out.


johnduffell:
closed in favour of https://github.com/guardian/support-frontend/pull/2812

PR #2803, opened 2020-10-30
clean up git stuff

johnduffell:
## Why are you doing this?

Just a few things to clean up  See inline comments.

PR #2808, opened 2020-11-03
roll back to sbt 1.3.13 due to intermittent dev assertion failure

johnduffell:

This reverts sbt back to 1.3.13.

Previous PR to update: https://github.com/guardian/support-frontend/pull/2803

## Why are you doing this?

Some people were getting  a duplicate compllations error.  It seems to be a common problem online https://github.com/sbt/sbt/issues/5911
It seems that there is a new assertion against more than one compilation start time for a subproject being in the same millisecond,  but this situation can occur (rightly or wrongly) during play compilation.

Reverting until there is a solution either from play or sbt.

### DIFF
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=1.4.1
+sbt.version=1.3.13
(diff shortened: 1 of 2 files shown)


rupertbates (approved):
👍

PR #2815, opened 2020-11-10
Fix broken selenium test

lucymonie:
## Why are you doing this?
After merging in the new digital gift subs landing page, there was a selenium error, which seems to be due to a missing id that the test is looking for. This was removed inadvertently when we replaced the footer in this page, and thus the FAQs for the page, which contained the id that was being looked for.

I have reinstated this to the new digital footer.

The PR that introduced this error is this one: https://github.com/guardian/support-frontend/pull/2775


johnduffell (approved):
thanks 👏

PR #2816, opened 2020-11-10
handle optional fields in emails

johnduffell:

This PR adds code to handle None Options in the email builder code.

It also adds a test to make sure it is working.

## Why are you doing this?

during gifting testing there were errors in the logs if somoene didn't fill in a gift message.
```
coding error: all values should be string or object: gift_personal_message -> object[gifter_first_name -> "n0ZoXIrKEPrtpPjgBvc",gift_personal_message -> null,gift_code -> "gd12-blahblah",last_redemption_date -> "Tuesday, 9 November 2021",duration -> "12 months"]
```
This is because an Option[String] that is None, is treated as null rather than String or Object.

This situation wasn't in any of the tests.


rupertbates (approved):
👍

PR #2817, opened 2020-11-10
send the gift notification to the giftee rather than gifter

johnduffell:

This PR passes the gift recipient SF contact id from the zuora step through to the email step, so that we can send the notification email to the correct person.

## Why are you doing this?

During testing we noticed the gift purchase TY email and notification email both went to the purchaser.

### DIFF
--- a/support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
+++ b/support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
@@ -4,6 +4,7 @@ import cats.implicits._
 import com.amazonaws.services.lambda.runtime.Context
 import com.gu.config.Configuration
 import com.gu.monitoring.SafeLogger
+import com.gu.salesforce.Salesforce.SfContactId
 import com.gu.services.{ServiceProvider, Services}
 import com.gu.support.catalog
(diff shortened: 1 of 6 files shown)


support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -233,13 +234,18 @@ class NextState(
...
       case (product: DigitalPack, _: Redemption) if product.readerType == ReaderType.Corporate =>
-        SendThankYouEmailDigitalSubscriptionCorporateRedemptionState(state.user, state.salesforceContacts.buyer, product, subscriptionNumber.value)
+        dsCorp(product)
       case (product: DigitalPack, _: Redemption) if product.readerType == ReaderType.Gift =>
         throw new RuntimeException("wrong code path for gift redemption")
       case _ => throw new RuntimeException("could not create value state")
     }
   // scalastyle:on cyclomatic.complexity
 
+  private def dsCorp(product: DigitalPack) =
```
rupertbates, line 244:
```suggestion
  private def dsCorporate(product: DigitalPack) =
```

rupertbates (approved):
With one suggestion 👍

PR #2820, opened 2020-11-11
Update ophan tracker js to 1.3.23

mchv:

Update ophan tracker dependency to latest version.
There should not be a breaking change ([see latest changes](https://github.com/guardian/ophan/blob/master/tracker-js/changes.md)) and [latest PR](https://github.com/guardian/ophan/pull/3964)

## Why are you doing this?

Keeping dependencies up-to-date.


johnduffell (approved):
not clear whether this is something we really need, but happy to keep things bumped up to date all the same 👍

PR #2822, opened 2020-11-11
use correct term months for gift recipient thank you email

johnduffell:

This PR passes the number of months term through from the create subscription step into the email step,  then this is used for the thank you recipient email.

## Why are you doing this?

Otherwise it gets the default value of 1 month
image

### DIFF
--- a/support-models/src/main/scala/com/gu/support/workers/states/SendThankYouEmailState.scala
+++ b/support-models/src/main/scala/com/gu/support/workers/states/SendThankYouEmailState.scala
@@ -1,6 +1,7 @@
 package com.gu.support.workers.states
 
 import com.gu.salesforce.Salesforce.SfContactId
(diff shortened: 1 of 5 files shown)


support-models/src/main/scala/com/gu/support/workers/states/SendThankYouEmailState.scala
```diff
@@ -63,8 +64,7 @@ object SendThankYouEmailState {
     user: User,
     sfContactId: SfContactId,
     product: DigitalPack,
```
johnduffell, line 66: `product: DigitalPack,` also has a billing period in it, but it's wrong as we didn't actually know when we populated it.  I will look to address this as part of https://trello.com/c/Mwb7aMda/3379-add-discriminator-to-producttype

rupertbates (approved):
👍

PR #2825, opened 2020-11-13
Remove Guardian auth from the redemption page

rupertbates:

We don't need to limit access to the redemption form to Guardian employees anymore

### DIFF
--- a/support-frontend/app/controllers/RedemptionController.scala
+++ b/support-frontend/app/controllers/RedemptionController.scala
@@ -61,7 +61,7 @@ class RedemptionController(
 
   val testUserFromRequest = new TestUserFromRequest(identityService, testUsers)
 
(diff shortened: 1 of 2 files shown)


support-frontend/app/controllers/RedemptionController.scala
```diff
@@ -61,7 +61,7 @@ class RedemptionController(
 
   val testUserFromRequest = new TestUserFromRequest(identityService, testUsers)
 
-  def displayForm(redemptionCode: RawRedemptionCode): Action[AnyContent] = (googleAuthAction andThen maybeAuthenticatedAction()).async {
```
johnduffell, line 64: there is unused code after removing this https://github.com/guardian/support-frontend/blob/32147ee138ad2cf9c731dbba7f92424c1e69f5e0/support-frontend/app/controllers/RedemptionController.scala#L45

johnduffell (approved):
thanks, this will work (although there is some cleanup as well) 👍

PR #2826, opened 2020-11-13
Migrate acquisition event producer to mono repo

rupertbates:

The acquisition-event-producer library is used to publish acquisition events to Ophan, Google Analytics and Athena from the `SendAcquisitionEvent` lambda.` 

Moving it into the support mono repo will mean that we can use it directly rather than having to release to Bintray on every change.


.gitignore
```diff
@@ -1,12 +1,48 @@
...
+target
+project/project
+storybook-static
+paparazzi
+
+bower_components
+public/stylesheets
+assets/javascripts/lib/bower-components/
+
+support-frontend/conf/keys.conf
```
johnduffell: weren't most of these in already?

johnduffell (approved):
it's very hard to review!  I think just do the minimum to get it to build and get it merged, and then we can take it from there in terms of tidyup.

PR #2827, opened 2020-11-16
Add `giftRedemptionDate` to the `UpdateRedemptionDataRequest` class

rupertbates:
We want to be able to record the date a gift subscription was redeemed in a custom field in Zuora so that it is available to analysts in the Data Lake. 

This PR updates the `GifteeRedemptionDate` custom field with the current date when a gift code is redeemed.

[**Trello Card**](https://trello.com/c/wSuf5u9V/3453-add-redemption-date-field)


johnduffell (approved):
looks good but I'm surprised we didn't have any tests to update, should we have?

rupertbates:
> looks good but I'm surprised we didn't have any tests to update, should we have?

I did think the same thing, but I can't really see anything to test!

PR #2829, opened 2020-11-17
Test contribution type and amount in the thankyou page header

tomrf1:
### One-off with card
Screen Shot 2020-11-17 at 15 13 09

### One-off paypal (not supported because we lose the required info in the redirect)
Screen Shot 2020-11-17 at 15 12 33

### 'other' amount
Screen Shot 2020-11-17 at 15 13 41

### Monthly
Screen Shot 2020-11-17 at 15 15 28

### Annual
Screen Shot 2020-11-17 at 15 16 54

### DIFF
--- a/support-frontend/assets/pages/contributions-landing/components/ContributionThankYou/ContributionThankYouHeader.jsx
+++ b/support-frontend/assets/pages/contributions-landing/components/ContributionThankYou/ContributionThankYouHeader.jsx
@@ -71,7 +71,7 @@ const ContributionThankYouHeader = ({
     if (thankyouPageHeadingTestVariant && !payPalOneOff && amount) {
       switch (contributionType) {
         case 'ONE_OFF':
-          return `Thank you for supporting us with ${currencies[currency].glyph}${amount} today`;
+          return `Thank you for supporting us today with ${currencies[currency].glyph}${amount}`;
         case 'MONTHLY':
           return `Thank you ${nameAndTrailingSpace}for choosing to contribute ${currencies[currency].glyph}${amount} each month`;
         case 'ANNUAL':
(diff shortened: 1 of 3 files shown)


support-frontend/assets/pages/contributions-landing/components/ContributionThankYou/ContributionThankYouHeader.jsx
```diff
@@ -39,22 +43,53 @@ const directDebitSetupText = css`
...
+  amount,
+  currency,
+  thankyouPageHeadingTestVariant,
 }: ContributionThankYouHeaderProps) => {
-  const shouldShowName = name && name.length < MAX_DISPLAY_NAME_LENGTH;
+
+  const title = (): string => {
+    const nameAndTrailingSpace: string = name && name.length < MAX_DISPLAY_NAME_LENGTH ? `${name} ` : '';
+    // Do not show special header to paypal/one-off as we don't have the relevant info after the redirect
+    const payPalOneOff = paymentMethod === 'PayPal' && contributionType === 'ONE_OFF';
```
TomPretty, line 69: could we remove these from the test by making the targetPage just the TY page and using `canRun` to exclude them? hmmm I can't tell if that has access to the redux state...

tomrf1: unfortunately it seems we cannot rely on access to that info there. Even the function that decides which tests to put the user in doesn't have it

tomrf1: I _think_ it's ok for these users to be in the test because they'll be evenly split between control/variant.
Also we can probably exclude them in the analysis if necessary

TomPretty: yeah makes sense 👍

TomPretty (approved):
Looks great! 🚀

PR #2830, opened 2020-11-18
Digital Subscription gift revenue recognition

rupertbates:

For Digital Subscription gifts we want a different revenue recognition model than for normal subscriptions. Whereas for normal subs we begin to recognise the revenue from the point of purchase, for gift subscriptions we want to recognise the revenue from the point at which the giftee redeems their gift code.

To do this we have set up a 'Billing Manual' revenue recognition rule which we use in the charge attached to the gift subscription rate plans.

When subscriptions with this rate plan are created a Revenue Schedule is created but all revenue remains ‘Undistributed’ until it is manually distributed either through the Zuora UI or via an api call.

There are more details in [this document](https://docs.google.com/document/d/1yNeaR2l1Ss_unXygntuntmRX-GicDelryuK25vmqToc/edit#heading=h.15hbk4mzmzt0)

### DIFF
--- a/support-models/src/main/scala/com/gu/support/zuora/api/response/SubscriptionsResponse.scala
+++ b/support-models/src/main/scala/com/gu/support/zuora/api/response/SubscriptionsResponse.scala
@@ -50,5 +50,9 @@ case class RevenueSchedulesResponse(revenueSchedules: List[RevenueSchedule])
 object RevenueSchedule {
   implicit val codec: Codec[RevenueSchedule] = deriveCodec
 }
-case class RevenueSchedule(number: String, recognitionRuleName: String)
+case class RevenueSchedule(
+  number: String,
+  amount: BigDecimal,
+  undistributedUnrecognizedRevenue: BigDecimal
+)
 
--- a/support-models/src/main/scala/com/gu/support/zuora/api/response/SubscriptionsResponse.scala
+++ b/support-models/src/main/scala/com/gu/support/zuora/api/response/SubscriptionsResponse.scala
@@ -9,17 +9,46 @@ object SubscriptionsResponse {
   implicit val codec: Codec[SubscriptionsResponse] = deriveCodec
 }
 
+case class SubscriptionsResponse(subscriptions: List[Subscription])
+
 object Subscription {
   implicit val codec: Codec[Subscription] = deriveCodec
 }
+
+case class Subscription(
+  customerAcceptanceDate: LocalDate,
+  accountNumber: String,
+  subscriptionNumber: String,
+  status: String,
+  CreatedRequestId__c: Option[String],
+  ratePlans: List[RatePlan]
+)
 
 object RatePlan {
   implicit val codec: Codec[RatePlan] = deriveCodec
 }
 
-case class SubscriptionsResponse(subscriptions: List[Subscription])
+case class RatePlan(
+  productId: String,
+  productName: String,
+  productRatePlanId: String,
+  ratePlanCharges: List[RatePlanCharge]
+)
 
-case class Subscription(customerAcceptanceDate: LocalDate, accountNumber: String, subscriptionNumber: String, status: String, CreatedRequestId__c: Option[String], ratePlans: List[RatePlan])
+object RatePlanCharge {
+  implicit val codec: Codec[RatePlanCharge] = deriveCodec
+}
 
-case class RatePlan(productId: String, productName: String, productRatePlanId: String)
+case class RatePlanCharge(id: String, name: String)
+
+object RevenueSchedulesResponse {
+  implicit val codec: Codec[RevenueSchedulesResponse] = deriveCodec
+}
+
+case class RevenueSchedulesResponse(revenueSchedules: List[RevenueSchedule])
+
+object RevenueSchedule {
+  implicit val codec: Codec[RevenueSchedule] = deriveCodec
+}
+case class RevenueSchedule(number: String, recognitionRuleName: String)
 
--- a/support-services/src/main/scala/com/gu/zuora/ZuoraGiftService.scala
+++ b/support-services/src/main/scala/com/gu/zuora/ZuoraGiftService.scala
@@ -2,6 +2,8 @@ package com.gu.zuora
 
 import cats.data.OptionT
 import cats.implicits.catsStdInstancesForFuture
+import com.gu.monitoring.SafeLogger
+import com.gu.monitoring.SafeLogger.Sanitizer
 import com.gu.okhttp.RequestRunners.FutureHttpClient
 import com.gu.rest.WebServiceHelper
 import com.gu.support.config.ZuoraConfig
@@ -14,6 +16,7 @@ import io.circe.syntax.EncoderOps
 import org.joda.time.LocalDate
 
 import scala.concurrent.{ExecutionContext, Future}
+import scala.util.{Failure, Success}
 
 trait ZuoraGiftLookupService extends TouchpointService {
   def getSubscriptionFromRedemptionCode(redemptionCode: RedemptionCode): Future[SubscriptionRedemptionQueryResponse]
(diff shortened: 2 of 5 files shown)


support-services/src/main/scala/com/gu/zuora/ZuoraGiftService.scala
```diff
@@ -44,8 +50,73 @@ class ZuoraGiftService(val config: ZuoraConfig, client: FutureHttpClient)(implic
...
+          Success(successfulResponse)
+        case Success(failedResponse) =>
+          // Treat a ZuoraResponse where success = false as a failure
+          val responseMessage = failedResponse.errorMessage.getOrElse("unknown")
+          val errorMessage = s"Failed to update gifteeIdentityId on Digital Subscription gift for user $gifteeIdentityId. Error was ${responseMessage}"
+          Failure(new RuntimeException(errorMessage))
+        case Failure(originalError) =>
+          val errorMessage = s"Failed to update gifteeIdentityId on Digital Subscription gift for user $gifteeIdentityId."
+          Failure(new RuntimeException(errorMessage, originalError))
+      }
```
johnduffell, line 68: it's very verbose but this will make it clear what happened when things go wrong 👏

support-services/src/main/scala/com/gu/zuora/ZuoraGiftService.scala
```diff
@@ -44,8 +50,73 @@ class ZuoraGiftService(val config: ZuoraConfig, client: FutureHttpClient)(implic
...
+        val errorMessage = scrub"Failed to set up revenue recognition for Digital Subscription gift $subscriptionNumber."
+        SafeLogger.error(errorMessage) //TODO: Add an alarm
+        Success(())
+    }
+  }
+
+  private def getRevenueSchedule(ratePlanChargeId: String): Future[RevenueSchedulesResponse] =
+    get[RevenueSchedulesResponse](s"revenue-schedules/subscription-charges/${ratePlanChargeId}", authHeaders)
+
+  private def distributeRevenueByStartAndEndDates(
```
johnduffell, line 111:
```suggestion
  // https://www.zuora.com/developer/api-reference/#operation/PUT_RevenueByRecognitionStartandEndDates
  private def distributeRevenueByStartAndEndDates(
```

support-models/src/main/scala/com/gu/support/zuora/api/response/SubscriptionsResponse.scala
```diff
@@ -9,17 +9,46 @@ object SubscriptionsResponse {
...
+  implicit val codec: Codec[RevenueSchedulesResponse] = deriveCodec
+}
 
-case class RatePlan(productId: String, productName: String, productRatePlanId: String)
+case class RevenueSchedulesResponse(revenueSchedules: List[RevenueSchedule])
+
+object RevenueSchedule {
+  implicit val codec: Codec[RevenueSchedule] = deriveCodec
+}
+case class RevenueSchedule(number: String, recognitionRuleName: String)
```
johnduffell, line 53: do we use `recognitionRuleName` ?

support-services/src/main/scala/com/gu/zuora/ZuoraGiftService.scala
```diff
@@ -44,8 +50,73 @@ class ZuoraGiftService(val config: ZuoraConfig, client: FutureHttpClient)(implic
...
+        SafeLogger.error(errorMessage) //TODO: Add an alarm
+        Success(())
+      case Failure(originalError) =>
+        val errorMessage = scrub"Failed to set up revenue recognition for Digital Subscription gift $subscriptionNumber."
+        SafeLogger.error(errorMessage, originalError) //TODO: Add an alarm
+        Success(())
+      case _ =>
+        val errorMessage = scrub"Failed to set up revenue recognition for Digital Subscription gift $subscriptionNumber."
+        SafeLogger.error(errorMessage) //TODO: Add an alarm
+        Success(())
```
johnduffell, line 104: I know it will be slightly less easy to read the message, but I feel like the error handling outweighs the actual logic.  What do you think?
```suggestion
      case other =>
        val errorMessage = scrub"Failed to set up revenue recognition for Digital Subscription gift $subscriptionNumber. Error was $other"
        SafeLogger.error(errorMessage) //TODO: Add an alarm
        Success(())
```

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -323,23 +323,42 @@ object DigitalSubscriptionGiftRedemption {
...
+          zuoraService.setupRevenueRecognition,
+          zuoraService,
+          catalogService
+        )
+        case CodeRedeemedInThisRequest(subscriptionId) => doZuoraUpdates(
+          subscriptionId,
+          (_: Int) => Future.successful(ZuoraSuccessOrFailureResponse(success = true, None)),
+          (_: Subscription, _:TermDates) => Future.successful(()),
+          zuoraService,
+          catalogService
```
johnduffell, line 339: maybe it's marginal here, but to help us focus on the business logic/reduce clutter, I'd be tempted to make `doZuoraUpdates` a method in  new `class ZuoraUpdater` and pass those services in as constructor parameters.  Then we do zuoraUpdater.doZuoraUpdates each time instead of those two lines.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -323,23 +323,42 @@ object DigitalSubscriptionGiftRedemption {
...
-      updateDataResponse <- updateIfNecessary(newTermLength)
-      handlerResult <- Future.fromTry(buildHandlerResult(updateDataResponse, state, requestInfo, dates))
+      handlerResult <- Future.fromTry(buildHandlerResult(state, requestInfo, termDates))
     } yield handlerResult
   }
 
+  private def doZuoraUpdates(
+    subscriptionId: String,
+    updateGiftIdentityIdCall: Int => Future[ZuoraSuccessOrFailureResponse],
+    setupRevenueRecognitionCall: (Subscription, TermDates) => Future[Unit],
```
johnduffell, line 350: I know I started it, but we're proliferating anonymous functions a bit here, consider a type alias or trait to gain nice naming or type checking respectively.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -323,23 +323,42 @@ object DigitalSubscriptionGiftRedemption {
...
-        case CodeRedeemedInThisRequest(subscriptionId) => Future.successful(subscriptionId, (_: Int) => Future.successful(UpdateRedemptionDataResponse(true)))
+          zuoraService.updateSubscriptionRedemptionData(subscriptionId, state.requestId.toString, state.user.id, LocalDate.now(), _),
+          zuoraService.setupRevenueRecognition,
+          zuoraService,
+          catalogService
+        )
+        case CodeRedeemedInThisRequest(subscriptionId) => doZuoraUpdates(
+          subscriptionId,
+          (_: Int) => Future.successful(ZuoraSuccessOrFailureResponse(success = true, None)),
+          (_: Subscription, _:TermDates) => Future.successful(()),
```
johnduffell, line 337: I'm not sure this really is idempodent now.  The lambda can fail at any point on the first iteration - e.g. zuora could time out when doing the upddate identity id call, yet succeed.  Then when the lambda runs again it will be redemmed in this request, but the revenue distribution would still need to happen.
We would need to redo the distribution really, to be sure, or check and do if necessary.

support-services/src/main/scala/com/gu/zuora/ZuoraGiftService.scala
rupertbates: Yes I think you're probably right

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
rupertbates: We're only passing one function in now so I'm going to leave as is

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
rupertbates: I've updated this so that all code paths now call `setupRevenueRecognition` and that function checks whether the distribute revenue call has already happened before calling it again

support-services/src/main/scala/com/gu/zuora/ZuoraGiftService.scala
```diff
@@ -77,37 +77,47 @@ class ZuoraGiftService(val config: ZuoraConfig, client: FutureHttpClient)(implic
...
+    else
+      for {
+        revenueScheduleNumber <- OptionT.fromOption(revenueSchedule.revenueSchedules.headOption.map(_.number))
+        successOrFailureResponse <- OptionT.liftF(distributeRevenueByStartAndEndDates(revenueScheduleNumber, termDates))
+      } yield successOrFailureResponse
+
+  private def revenueScheduleIsAlreadyDistributed(revenueSchedule: RevenueSchedulesResponse) =
+    revenueSchedule.revenueSchedules.exists(revenueSchedule =>
+      // revenueSchedule.amount and revenueSchedule.undistributedUnrecognizedRevenue will be
+      // the same until the distribute revenue call has been made
```
johnduffell, line 116: I presume zuora otherwise would automatically recognise the revenue over the charge dates.  Did you have to change the product charge so that the revenue is not distributed by default?

rupertbates: Yes you set a 'Billing manual' revenue recognition rule on the product charge and that tells Zuora to keep all revenue as 'Undistributed' until it is manually distributed, either through the UI or via and api call

johnduffell (approved):
great work on this, the PR desciption is important to fill in well for this one to help future understanding, as it's not a simple situation.  But the code looks great!

PR #2832, opened 2020-11-18
send custom attributes to workflow to be inserted into braze

johnduffell:

depends on https://github.com/guardian/membership-workflow/pull/275

This pushes the unmanaged gift start/end date attributes and term length into the membership workflow email queue.
This means they will arrive in braze as user custom attributes once the above is merged.

https://trello.com/c/EZlDR9wB/3429-gifting-braze-custom-data-attribute-5

## Why are you doing this?

The marketing and retention teams need to be able to target gift recipients in the days after redemption and before expiry, for onboarding and renewal comms.
Normally they get these dates from the DL subscriptions extract from zuora, but we don't extract those details from zuora into the DL/braze at present.  This is a quick bypass, as the data is write only (you can't aamend a gift redemption at present.)

### DIFF
--- a/support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
+++ b/support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
@@ -12,6 +12,7 @@ import io.circe._
 import io.circe.generic.semiauto.deriveEncoder
 import io.circe.syntax._
 import org.joda.time.LocalDate
+import org.joda.time.format.ISODateTimeFormat
 
 import scala.concurrent.{ExecutionContext, Future}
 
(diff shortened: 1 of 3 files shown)


support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
```diff
@@ -154,11 +155,22 @@ class DigitalPackEmailFields(
...
+    userAttributes: Option[JsonObject] = None,
   ): EmailFields = {
     val attributePairs = JsonToAttributes.asFlattenedPairs(fields.asJsonObject).left.map(
       error => throw new RuntimeException(s"coding error: $error")
     ).merge
-    EmailFields(attributePairs, Left(sfContactId), emailAddress, dataExtensionName, deliveryDate)
+    EmailFields(attributePairs, Left(sfContactId), emailAddress, dataExtensionName, deliveryDate, userAttributes)
+  }
+
+  private def wrap2[UserAttributes: Encoder.AsObject](
```
rupertbates, line 166: Do we really need the wrap2 method? Can't we just do `userAttributes.map(_.asJsonObject)` at the call site?

rupertbates (approved):
👍  with one question

support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
johnduffell: good spot, although I feel like both the wrap methods shouldn't really exist, they were kind of one of the strings I didn't pull at when I was doing the refactors
I will see if I can not add the second one now.

PR #2828, opened 2020-11-16
Gifting: digital subs landing page - fixes

lucymonie:
## What are you doing in this PR?
These are fixes from the [digital gift subscriptions landing page PR](https://github.com/guardian/support-frontend/pull/2775). This pull request needed to be merged because we wanted to begin end-to-end testing, but there were a couple of things that were outstanding from feedback on the PR.

* Remove orderIsAGift param from Promotions.scala
* Make no animations the default behaviour in gifting
* Update the text for the gifting hero (screenshot below)

The details of these fixes are in this [trello Card](https://trello.com/c/ElhMe2AO/3433-digital-subscriptions-gift-landing-page-follow-up-work-90)

In addition, this PR contains fixes for issues that were picked up in end-to-end testing:
* The change link in the gift checkout order summary was pointing to the non-gift landing page
* The copy on the gift thank you page reads: "Your Digital subscription order has been placed successfully" and it should read: "Your Digital subscription gift for [firstname] has been placed successfully" (screenshot below)
* The heading on the gift thank you page that reads: "Tell us about your subscription" should read: "Tell us about your experience" (screenshot below)
* There should be a sign-in for ad-free reading section on the regular DS thank you page (screenshot of what's been added, below)

### Updated hero text

Screen Shot 2020-11-18 at 11 43 19

### Giftee name included in first paragraph

Screen Shot 2020-11-17 at 12 44 00

### Tell us about your experience

Screen Shot 2020-11-17 at 12 44 18

### Add sign-in for ad-free to thank you page

Screen Shot 2020-11-17 at 12 53 14

PR #2834, opened 2020-11-19
Fix styles when user prefers reduced motion

lucymonie:
This is because I broke `prefers-reduced-motion` in this PR: https://github.com/guardian/support-frontend/pull/2828
😳 

### This is how it looked when it was broken
Screen Shot 2020-11-19 at 10 35 29

### And this is it fixed
Screen Shot 2020-11-19 at 10 37 54


GHaberis (approved):
Looks good 👍

PR #2835, opened 2020-11-19
Update CSP to allow CCPA iframe in USA

GHaberis:

@mxdvl raised an issue where US user's who click the Privacy Settings link see an iframe that fails to load rather than the Sourcepoint CMP, an example of this blocked iframe can be seen below. This occurs because our content security policy (CSP) blocks the iframe's domain "https://ccpa-pm.sp-prod.net". This PR adds this domain to the list of allowed domains in our CSP.

image (2)


mxdvl (approved):
LGTM.

Maybe we could use [multiline strings and `stripMargin`][1] to make it easier to spot differences?

[1]: https://alvinalexander.com/scala/how-to-create-multiline-strings-heredoc-in-scala-cookbook/

PR #2839, opened 2020-11-19
Honour consent state in personalisedAdvertising flag in Australia

GHaberis:
Now we're using the latest version of `@guardian/consent-management-platform` @ `v6.7.4` we need to check Australian consent state's `personalisedAdvertising` flag if available.


mxdvl (approved):
LGTM

PR #2840, opened 2020-11-19
Fix selenium selector for GW product page tests

i-hardy:

This adds a page ID and uses that as the selector for the Selenium tests verifying that the Guardian Weekly product pages load in prod, replacing the usage of a class name as a selector now we've mostly migrated the page to Emotion.


GHaberis (approved):
LGTM 👍

PR #2841, opened 2020-11-20
Add an alarm for failures in the Zuora distribute revenue api call

rupertbates:
Following from #2830 we want to add an alarm when the calls to Zuora to set up revenue distribution fail.

### DIFF
--- a/support-services/src/main/scala/com/gu/aws/AwsCloudWatchMetricSetup.scala
+++ b/support-services/src/main/scala/com/gu/aws/AwsCloudWatchMetricSetup.scala
@@ -18,8 +18,12 @@ object AwsCloudWatchMetricSetup {
         MetricDimensionName("Environment") -> MetricDimensionValue(environment.envValue)
       ))
 
-  def revenueDistributionFailureRequest: MetricRequest =
-    getMetricRequest(MetricName("RevenueDistributionFailure"), Map())
+  def revenueDistributionFailureRequest(stage: Stage): MetricRequest =
+    getMetricRequest(MetricName("RevenueDistributionFailure"),
+      Map(
+        MetricDimensionName("Stage") -> MetricDimensionValue(stage.toString)
+      )
+    )
 
   def paymentSuccessRequest(stage: Stage, paymentProvider: PaymentProvider, productType: ProductType): MetricRequest =
     getMetricRequest(
--- a/support-services/src/main/scala/com/gu/aws/AwsCloudWatchMetricSetup.scala
+++ b/support-services/src/main/scala/com/gu/aws/AwsCloudWatchMetricSetup.scala
@@ -15,8 +15,11 @@ object AwsCloudWatchMetricSetup {
   def catalogFailureRequest(environment: TouchPointEnvironment): MetricRequest =
     getMetricRequest(MetricName("CatalogLoadingFailure"),
       Map(
-        MetricDimensionName("Environment") -> MetricDimensionValue(environment.toString)
+        MetricDimensionName("Environment") -> MetricDimensionValue(environment.envValue)
       ))
+
+  def revenueDistributionFailureRequest: MetricRequest =
+    getMetricRequest(MetricName("RevenueDistributionFailure"), Map())
 
   def paymentSuccessRequest(stage: Stage, paymentProvider: PaymentProvider, productType: ProductType): MetricRequest =
     getMetricRequest(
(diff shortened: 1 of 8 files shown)


support-services/src/main/scala/com/gu/aws/AwsCloudWatchMetricSetup.scala
```diff
@@ -15,9 +15,12 @@ object AwsCloudWatchMetricSetup {
   def catalogFailureRequest(environment: TouchPointEnvironment): MetricRequest =
     getMetricRequest(MetricName("CatalogLoadingFailure"),
       Map(
-        MetricDimensionName("Environment") -> MetricDimensionValue(environment.toString)
+        MetricDimensionName("Environment") -> MetricDimensionValue(environment.envValue)
```
johnduffell, line 18: good spot!

support-services/src/main/scala/com/gu/aws/AwsCloudWatchMetricSetup.scala
```diff
@@ -15,9 +15,12 @@ object AwsCloudWatchMetricSetup {
...
       ))
 
+  def revenueDistributionFailureRequest: MetricRequest =
+    getMetricRequest(MetricName("RevenueDistributionFailure"), Map())
```
johnduffell, line 22: we would need a Stage -> PROD/CODE/DEV in the map as well, so it doesn't alarm when people are testing

support-workers/cloud-formation/src/templates/cfn-template.yaml
```diff
@@ -207,6 +207,23 @@ Resources:
       Statistic: Sum
     DependsOn: SupportWorkersPROD
 
+    RevenueRecognitionSetupFailureAlarm:
+      Type: AWS::CloudWatch::Alarm
+      Condition: CreateProdResources
+      Properties:
+        AlarmActions:
+          - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:reader-revenue-dev
+        AlarmName: !Sub Revenue distribution for Digital Subscription gift failed in ${Stage}
```
johnduffell, line 216: this is maybe not URGENT-9-5 but it definitely needs looking at as it is not self correcting when it goes back to OK like other alarms.

support-services/src/main/scala/com/gu/aws/AwsCloudWatchMetricSetup.scala
rupertbates: It will only run for PROD support-workers so it won't be a problem for DEV or CODE, we might get some test user failures but it would involve quite a lot of code change to avoid that so I'd like to leave it as is for now and only do that work if it becomes a problem

support-workers/cloud-formation/src/templates/cfn-template.yaml
rupertbates: What do you suggest? Including some wording to that effect?

support-services/src/main/scala/com/gu/aws/AwsCloudWatchMetricSetup.scala
rupertbates: You're right, I had thought that because we were only creating the alarm for the PROD stage we wouldn't need it but of course that's just the alarm, the metric needs to have the stage as well. I've added this now.

support-workers/cloud-formation/src/templates/cfn-template.yaml
johnduffell: good question, maybe `DATA CORRUPTED - Revenue distribution for Digital Subscription gift failed in ${Stage}`  And there's a link to a doc that we can put more details in for each alarm, that usually goes in the description.
https://github.com/guardian/support-frontend/blob/920e638c35430cc260acdb1878f37bffa1d12fae/support-frontend/cloud-formation/cfn.yaml#L62

this is basically a data issue that will have been crystallised by this point.  The idea is to be clear to someone who doesn't know about that alarm whether they need to panic or what to do.

Ideally we would have a daily/weekly report of all redeemed gifts that don't have the right revenue distribution so they won't go away until they're fixed.  Maybe one for the new Data lake.

support-services/src/main/scala/com/gu/aws/AwsCloudWatchMetricSetup.scala
johnduffell: yes, and from your comment I guess you edited it, I agree about test users ideally shouldn't trigger this, but we get alarms for test users already as illustrated by the gifting testing session, that's one for another debate I think.

johnduffell (approved):
just the alarm wording to be clear to someone not in the know. hopefully it will never go off anyway! 👍

PR #2849, opened 2020-11-24
Set httpOnly and secure flags on csrf cookie

tomrf1:
This was a recommendation from the security audit of our site.

"The GU_support_csrf (https://support.theguardian.com) was identified as being set without the Secure flag. The Secure flag set forces a browser to only send the cookie in requests to sites protected with SSL (HTTPS sites). The HttpOnly flag forces the cookie to not be accessible via any client-side script (example: JavaScript using document.cookie)."

Screen Shot 2020-11-24 at 13 52 05

### DIFF
--- a/support-frontend/conf/application.conf
+++ b/support-frontend/conf/application.conf
@@ -2,7 +2,10 @@ play.application.loader = "wiring.AppLoader"
 
 googleAuth.domain = "guardian.co.uk"
 
-play.filters.csrf.cookie.name = "GU_support_csrf"
+play.filters.csrf.cookie {
+  name = "GU_support_csrf"
+  httpOnly = true
+}
 
 # cache hashed assets for one year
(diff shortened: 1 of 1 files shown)


rupertbates:
Have you checked that we don't access that cookie from Javascript anywhere? I wouldn't think we do but it's possible.

tomrf1:
> Have you checked that we don't access that cookie from Javascript anywhere? I wouldn't think we do but it's possible.

We don't access the cookie directly: https://github.com/guardian/support-frontend/search?q=GU_support_csrf

It's passed in from the template like this: https://github.com/guardian/support-frontend/blob/main/support-frontend/app/views/subscriptionCheckout.scala.html#L66

I'm not sure if this is standard practice...

rupertbates (approved):
🌮  LGTM

PR #2850, opened 2020-11-24
Reinstate 'best deal' on gift price cards

lucymonie:
The 'best deal' yellow label had gone missing from the gift price cards. The work done here is to reinstate this feature.

## Why are you doing this?
Because it will help users to identify that they would pay less money per month for 12 months worth of a gift subscription than for 3 months.

## Screenshots
Screen Shot 2020-11-24 at 13 50 22

Screen Shot 2020-11-24 at 13 50 05


GHaberis (approved):
👍

PR #2851, opened 2020-11-25
Add an error message to redemptionForm.jsx if id user names are missing

rupertbates:
## What are you doing in this PR?
Some Identity accounts such as those created via the single contributions checkout do not have a first name and last name set on them. This is a problem for gift redemption because these are mandatory fields for creating a subscription.

The longer term solution for this is to add a form to collect these fields for users who don't have them, but for launch we are just going to put an error message in place informing users that they need to set these fields via MMA.

[**Trello Card**](https://trello.com/c/s8bUR3HP/3467-investigate-redemption-errors-for-users-created-through-single-contribution-checkout)

## Accessibility test checklist
 - [ ] [Tested with screen reader](https://accessibility.gutools.co.uk/testing/web/screen-readers/)
 - [x] [Navigable with keyboard](https://accessibility.gutools.co.uk/testing/web/keyboard-navigation/)
 - [x] [Colour contrast passed](https://accessibility.gutools.co.uk/testing/web/colour-contrast/)

## Screenshots

Screen Shot 2020-11-23 at 18 03 22

PR #2852, opened 2020-11-25
Set user details correctly in redemption form

rupertbates:

This PR is a follow up to #2851. I noticed that the user information was not being set correctly in `window.guardian.user` by the server so the message was showing for all users.


GHaberis (approved):
👍

PR #2854, opened 2020-11-25
Only write address details to Salesforce for GW gifts, not DS

rupertbates:

## Why are you doing this?

Currently for Digital Subscription gifts we are writing the billing address of the gifter into the giftee's Salesforce contact record. This is because for Guardian Weekly gifts we want to add the delivery address to the contact however for DS gifts this does not apply.

### DIFF
--- a/support-workers/src/main/scala/com/gu/salesforce/SalesforceService.scala
+++ b/support-workers/src/main/scala/com/gu/salesforce/SalesforceService.scala
@@ -17,7 +17,7 @@ import okhttp3.Request
 
 import scala.concurrent.duration._
 import scala.concurrent.stm._
-import scala.concurrent.{Await, ExecutionContext, Future}
+import scala.concurrent.{ExecutionContext, Future}
 
 class SalesforceService(config: SalesforceConfig, client: FutureHttpClient)(implicit ec: ExecutionContext)
   extends WebServiceHelper[SalesforceErrorResponse] {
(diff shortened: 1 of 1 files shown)


support-workers/src/main/scala/com/gu/salesforce/SalesforceService.scala
```diff
@@ -51,15 +51,24 @@ class SalesforceService(config: SalesforceConfig, client: FutureHttpClient)(impl
...
-        user.deliveryAddress.getOrElse(user.billingAddress)
+        user
       )
     } yield SalesforceContactRecordsResponse(contactRecord, recipientContactRecord)
   }
 
-  private def maybeAddGiftRecipient(contactRecord: SalesforceContactRecord, maybeGiftRecipient: Option[GiftRecipient], deliveryAddress: Address) =
+  private def deliveryAddressForGiftRecipientType(giftRecipient: GiftRecipient, user: User) =
+    giftRecipient match {
+      case _: GiftRecipient.WeeklyGiftRecipient => Some(user.deliveryAddress.getOrElse(user.billingAddress))
```
rupertbates, line 61: Should we just use deliveryAddress here rather than falling back the billingAddress if it's None?

johnduffell: agree we should, I suspect that it is "not possible" but someone added that to get rid of the Optionality in the types rather than doing a `.get` or throwing an exception.
When we turn all the State case classes into traits like we did for the email state, the delivery address would not be optional for GW and wouldn't exist for DS so this problem will not exist.

rupertbates: Agreed

paulbrown1982: Yeah if there's a bug caused by the [] Use billing address as delivery address feature (or vice-versa) then it should be reconciled at a higher level not just left to be rescued via this random fallback.

support-workers/src/main/scala/com/gu/salesforce/SalesforceService.scala
```diff
@@ -51,15 +51,24 @@ class SalesforceService(config: SalesforceConfig, client: FutureHttpClient)(impl
...
-  private def maybeAddGiftRecipient(contactRecord: SalesforceContactRecord, maybeGiftRecipient: Option[GiftRecipient], deliveryAddress: Address) =
+  private def deliveryAddressForGiftRecipientType(giftRecipient: GiftRecipient, user: User) =
+    giftRecipient match {
+      case _: GiftRecipient.WeeklyGiftRecipient => user.deliveryAddress
+      case _: GiftRecipient.DigitalSubscriptionGiftRecipient  => None
+    }
+
+  private def maybeAddGiftRecipient(contactRecord: SalesforceContactRecord, maybeGiftRecipient: Option[GiftRecipient], user: User) =
     maybeGiftRecipient
       .filter(recipient => recipient.firstName != "" && recipient.lastName != "")
```
johnduffell, line 67: existing code I know, but it seems a bit late to be checking for empty strings! I'm sure the validators check this also

rupertbates: Yeah I was a bit confused about this as well, but as you say it was existing code so I left it.

paulbrown1982 (approved):
👍  LGTM, I agree the empty recipient name check should not be necessary here.

johnduffell (commented):
looks good, as long as salesforce can accept those fields missing (which I am pretty sure is the case) 👍

PR #2856, opened 2020-11-26
Correct crops of packshots in Hero Component

GHaberis:

This PR fixes broken packshot images used by the `hero.jsx` component, specifically linked to `editionsPackshotShort` and `editionsPackshotAusShort`, both images support a limited set of crops in the Grid: 500 x 300, 1000 x 600, 1800 x 1080.

Because our srcSizes contained crop dimensions that weren't on offer we were attempting to render crops that didn't exist, leading to instances of a broken packshot images. This PR switches to using supported crops.

### Before

IMG_2175

### After

Screenshot 2020-11-26 at 16 21 39


i-hardy (approved):
LGTM 👍

PR #2860, opened 2020-11-30
Add gifting url info to sharing links

lucymonie:
This is an update to the url builder for sharing links. This function wasn't taking gifting into account, so if someone tried to share the GW or DS gifting pages (using the share button on the browser on mobile) they would share the non-gift version of the page.

## Why are you doing this?
So people can share the gifting pages.

Screen Shot 2020-11-30 at 11 42 06

Screen Shot 2020-11-30 at 11 41 36


rupertbates (approved):
Nice 👍

PR #2861, opened 2020-11-30
Use less permissive ssm permissions

TomPretty:
Update the ssm policy to use a less permissive one, as per [this card](https://trello.com/c/qNgYO8yT/2435-replace-amazonec2roleforssm-with-custom-policy-on-payment-api).


tomrf1 (approved):
🎉

PR #2863, opened 2020-12-01
write gift notification date to zuora, when DS is purchased

johnduffell:

This PR makes the step functions write the gift notification date to zuora when a DS gift is purchased.

This field will end up in the DL, so we can do a one off report on this in a couple of weeks.

https://trello.com/c/gRopJsmH/3498-write-gift-delivery-date-to-zuora

## Why are you doing this?

CEX team are concerned that if a lot of gifts are scheduled for e.g. christmas day, it might cause a risk to support, as the call centre is not open over christmas/boxing day.

## Screenshots

Tested in DEV:
https://apisandbox.zuora.com/apps/Subscription.do?method=view&id=2c92c0f975e5ba1e01761e9e1109498d
image

Tested in UAT:
https://apisandbox.zuora.com/apps/Subscription.do?method=view&id=2c92c0f975e5ba2601761e9f907d5bcb
image

I also tested a normal Direct DS in UAT.

PROD field name:
image

### DIFF
--- a/support-models/src/main/scala/com/gu/support/zuora/api/SubscriptionData.scala
+++ b/support-models/src/main/scala/com/gu/support/zuora/api/SubscriptionData.scala
@@ -152,10 +152,6 @@ object SubscriptionProductFeature {
 case class SubscriptionProductFeature(featureId: String)
 
 object Subscription {
-  implicit val decoder: Decoder[Subscription] = decapitalizingDecoder[Subscription].prepare(
-    _.withFocus(_.mapObject(_.renameField("PromotionCode__c", "promoCode")))
-      .withFocus(_.mapObject(_.renameField("ReaderType__c", "readerType")))
-  )
 
   implicit val encoder: Encoder[Subscription] = capitalizingEncoder[Subscription].mapJsonObject(_
     .copyField("PromoCode", "PromotionCode__c")
@@ -163,6 +159,8 @@ object Subscription {
     .renameField("ReaderType", "ReaderType__c")
     .renameField("RedemptionCode", "RedemptionCode__c")
     .renameField("CorporateAccountId", "CorporateAccountId__c")
+    .renameField("CreatedRequestId", "CreatedRequestId__c")
+    .renameField("GiftNotificationEmailDate", "GiftNotificationEmailDate__c")
   )
 }
 
@@ -170,7 +168,7 @@ case class Subscription(
   contractEffectiveDate: LocalDate,
   contractAcceptanceDate: LocalDate,
   termStartDate: LocalDate,
-  createdRequestId__c: String,
+  createdRequestId: String,
   autoRenew: Boolean = true,
   initialTermPeriodType: PeriodType = Month,
   initialTerm: Int = 12,
@@ -180,7 +178,7 @@ case class Subscription(
   promoCode: Option[PromoCode] = None,
   redemptionCode: Option[RawRedemptionCode] = None,
   corporateAccountId: Option[String] = None,
-  giftNotificationEmailDate__c: Option[LocalDate] = None,
+  giftNotificationEmailDate: Option[LocalDate] = None,
 )
 
 object RatePlanChargeData {
@@ -200,7 +198,7 @@ case class RatePlanData(
 )
 
 object SubscriptionData {
-  implicit val codec: Codec[SubscriptionData] = capitalizingCodec
+  implicit val encoder: Encoder[SubscriptionData] = capitalizingEncoder
 }
 
 case class SubscriptionData(ratePlanData: List[RatePlanData], subscription: Subscription)
--- a/support-models/src/main/scala/com/gu/support/zuora/api/SubscriptionData.scala
+++ b/support-models/src/main/scala/com/gu/support/zuora/api/SubscriptionData.scala
@@ -179,7 +179,8 @@ case class Subscription(
   readerType: ReaderType = ReaderType.Direct,
   promoCode: Option[PromoCode] = None,
   redemptionCode: Option[RawRedemptionCode] = None,
-  corporateAccountId: Option[String] = None
+  corporateAccountId: Option[String] = None,
+  giftNotificationEmailDate__c: Option[LocalDate] = None,
 )
 
 object RatePlanChargeData {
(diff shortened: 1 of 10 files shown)


support-models/src/main/scala/com/gu/support/zuora/api/SubscriptionData.scala
```diff
@@ -179,7 +179,8 @@ case class Subscription(
   readerType: ReaderType = ReaderType.Direct,
   promoCode: Option[PromoCode] = None,
   redemptionCode: Option[RawRedemptionCode] = None,
-  corporateAccountId: Option[String] = None
+  corporateAccountId: Option[String] = None,
+  giftNotificationEmailDate__c: Option[LocalDate] = None,
```
rupertbates, line 183: Can we remove the `__c` from the end of this field and map it as we do with the other custom fields on this object here: https://github.com/guardian/support-frontend/blob/8cd4badd66b080321f84f79147459e1351ed48c0/support-models/src/main/scala/com/gu/support/zuora/api/SubscriptionData.scala#L165-L164

support-workers/src/main/scala/com/gu/emailservices/EmailFields.scala
```diff
@@ -50,7 +50,7 @@ case class EmailFields(
       DataExtensionName = dataExtensionName,
       SfContactId = userId.left.toOption.map(_.id),
       IdentityUserId = userId.right.toOption.map(_.id),
-      deliveryDate.map(_.toDateTime(new LocalTime(8, 0), DateTimeZone.UTC)),
+      deliveryDate.map(_.toDateTime(LocalTime.now().plusSeconds(90))),//new LocalTime(8, 0), DateTimeZone.UTC)),
```
rupertbates, line 53: Is this change intentional?

johnduffell: no good spot, i thought I unticked that one

support-models/src/main/scala/com/gu/support/zuora/api/SubscriptionData.scala
johnduffell: I was wondering that, we should probably do the same for createdRequestId__c: String, on line 173 too.

johnduffell: I've removed it from both now so it should be consistent.

support-models/src/main/scala/com/gu/support/zuora/api/SubscriptionData.scala
```diff
@@ -152,25 +152,23 @@ object SubscriptionProductFeature {
 case class SubscriptionProductFeature(featureId: String)
 
 object Subscription {
-  implicit val decoder: Decoder[Subscription] = decapitalizingDecoder[Subscription].prepare(
-    _.withFocus(_.mapObject(_.renameField("PromotionCode__c", "promoCode")))
-      .withFocus(_.mapObject(_.renameField("ReaderType__c", "readerType")))
-  )
```
johnduffell, line 158: this was out of date and we should never need to deserialise this as it's part of the request not the response.
I removed it completely and turned all the codecs that depend on it into Encoders and the maain code still seemed to compile okay.

support-models/src/test/scala/com/gu/support/zuora/api/SerialisationSpec.scala
```diff
@@ -36,17 +36,71 @@ class SerialisationSpec extends AsyncFlatSpec with SerialisationTestHelpers with
...
+              |    "InitialTerm" : 3,
+              |    "RenewalTerm" : 12,
+              |    "TermType" : "TERMED",
+              |    "ReaderType__c" : "Gift",
+              |    "RedemptionCode__c" : "gd03-asdfghjq",
+              |    "CreatedRequestId__c" : "requestId",
+              |    "GiftNotificationEmailDate__c" : "2020-12-25"
+              |  }
+              |}""".stripMargin).toTry.get
+    json.mapObject(_.mapValues(_.dropNullValues)) should be(expected)
```
johnduffell, line 94: getting rid of the nulls was tricky - we drop them in the json printer
https://github.com/guardian/support-frontend/blob/547ca4ea72ba19598165c89abb443ec289b5d9de/support-modules/rest/src/main/scala/com/gu/rest/WebServiceHelper.scala#L142
 but if we only go as far as the json, that doesn't apply.

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionPurchaseBuilder.scala
```diff
@@ -28,15 +28,16 @@ class DigitalSubscriptionPurchaseBuilder(
...
-    val (contractAcceptanceDelay, autoRenew, initialTerm, maybeRedemptionCode) =
-      (digitalPack.readerType, maybeGiftCode) match {
-        case (Gift, Some(giftCode)) => (0, false, GiftCodeValidator.expirationTimeInMonths + 1, Some(giftCode))
-        case (Gift, _) | (_, Some(_)) => throw new RuntimeException("coding error - possible unredeemable sub")
-        case _ => (config.defaultFreeTrialPeriod + config.paymentGracePeriod, true, 12, None)
+    val (contractAcceptanceDelay, autoRenew, initialTerm, maybeRedemptionCode, deliveryDate) =
+      (digitalPack.readerType, maybeGiftCode, maybeDeliveryDate) match {
+        case (Gift, Some(giftCode), Some(deliveryDate)) => (0, false, GiftCodeValidator.expirationTimeInMonths + 1, Some(giftCode), Some(deliveryDate))
+        case (Gift, _, _) | (_, Some(_), _) | (Gift, _, None) => throw new RuntimeException("coding error - possible unredeemable sub")
+        case _ => (config.defaultFreeTrialPeriod + config.paymentGracePeriod, true, 12, None, None)
```
johnduffell, line 40: now it will blow up if we get a Gift but we don't have a delivery date to put into zuora, otherwise we might not notice.  But it does mean that if we break that, all gift checkouts will fail, rather than going through with missing data.
If we think that's too harsh but still want some checking, we could make it a metric.
Alternatively, once we finish refactoring the state types, this is the sort of thing that would be checked by the type system anyway.

rupertbates: I think that's fine

rupertbates (approved):
Good work 👍

PR #2868, opened 2020-12-03
update jackson dependency

tomrf1:
and also play-json (which depends on jackson-databind) to the version that this repo's version of play uses

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,6 +1,7 @@
 import SeleniumTestConfig.SeleniumTest
 import sbt.Keys.{publishTo, resolvers, scalaVersion, skip, updateOptions, organization}
 import sbtrelease.ReleaseStateTransformations._
+import LibraryVersions.jacksonVersion
 
 import scala.sys.process._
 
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -182,7 +183,7 @@ lazy val `acquisition-event-producer` = (project in file("acquisition-event-prod
     libraryDependencies ++= Seq(
       "com.gu" %% "ophan-event-model" % "0.0.17" excludeAll ExclusionRule(organization = "com.typesafe.play"),
       "com.gu" %% "fezziwig" % "1.3",
-      "com.typesafe.play" %% "play-json" % "2.6.14",
+      "com.typesafe.play" %% "play-json" % "2.7.4",
```
johnduffell, line 186: good, but is there a way to get the play version from the play plugin?  e.g. com.play.SbtPlugin.playVersion or something?

project/LibraryVersions.scala
```diff
@@ -3,7 +3,7 @@ object LibraryVersions {
   val circeVersion = "0.12.1"
   val awsClientVersion = "1.11.827"
   val catsVersion = "1.4.0"
-  val jacksonVersion = "2.10.5"
+  val jacksonVersion = "2.12.0"
```
johnduffell, line 6: when I updated jackson, I stayed on the same minor version but just bumped the patch version.  That seemed to be enough to keep snyk happy but with less risk.

johnduffell: see
https://github.com/guardian/members-data-api/blob/112029e534827aa476b450b7d8757ab0a9a2c9b9/project/Dependencies.scala#L47

johnduffell (approved):
great, thanks for doing that! :+

build.sbt
tomrf1: that would be nice! It might be possible by accessing `PlayScala.projectSettings`? But that would be quite brittle!

tomrf1: Also, the play-json version doesn't actually align with the play plugin version, I've just gone with what our play version depends on

project/LibraryVersions.scala
tomrf1: I see - jackson updates always seem pretty low risk to me! The tests pass and I've tested a contribution in CODE, so I think it's safe

PR #2866, opened 2020-12-03
send product discriminator when creating a subscription

johnduffell:
## What are you doing in this PR?

This PR gets the client side to send a discriminator field when creating a subscription.  This means that it is explicit about which product the user is signing up for, rather than just looking at the shape of the data.

The next PR will actually use the PR on the server side.

https://trello.com/c/Mwb7aMda/3379-add-discriminator-to-producttype

## Why are you doing this?

We had a contribution that came through a few months without an amount, so the system tried to put it through as a digital subscription.  Luckily it failed for other reasons (payment) but this is a flaky situation to be in, so we decided to make things more explicit in case it breaks as we refactor things or add a new product.

## Testing
part of the POST request from testing locally, per product
```
product: {productType: "Contribution", amount: 5, currency: "GBP", billingPeriod: "Monthly"}
product: {productType: "DigitalPack", currency: "GBP", billingPeriod: "Quarterly", readerType: "Gift"}
product: {productType: "DigitalPack", currency: "GBP", billingPeriod: "Monthly", readerType: "Direct"}
six day paper:
product: {
billingPeriod: "Monthly"
currency: "GBP"
fulfilmentOptions: "Collection"
productOptions: "Sixday"
productType: "Paper"
}
GW gift:
product: {
billingPeriod: "Quarterly"
currency: "GBP"
fulfilmentOptions: "Domestic"
productType: "GuardianWeekly"
}
```

PR #2869, opened 2020-12-03
use discriminator on the backend to deserialise product object

johnduffell:

Following on from https://github.com/guardian/support-frontend/pull/2866

This adds the code on the server side to pay attention to the incoming productType field in the client side payload.

https://trello.com/c/Mwb7aMda/3379-add-discriminator-to-producttype

## Why are you doing this?

Same reasons as the above PR.

## Testing

All tests and IT tests working from support-frontend and support-workers
Need final test in CODE once approved


support-models/src/test/scala/com/gu/support/workers/SerialisationSpec.scala
```diff
@@ -16,11 +16,6 @@ import org.scalatest.flatspec.AnyFlatSpec
 
 class SerialisationSpec extends AnyFlatSpec with SerialisationTestHelpers with LazyLogging with EitherValues {
 
-  "Contribution JSON with a billing period set" should "be decoded correctly" in {
-    val input = contribution(billingPeriod = Annual)
-    testDecoding[Contribution](input, _.billingPeriod shouldBe Annual)
-  }
```
johnduffell, line 22: this is covered in the above test class

support-models/src/main/scala/com/gu/support/workers/Products.scala
```diff
@@ -57,25 +57,14 @@ case class GuardianWeekly(
 
 object ProductType {
   import com.gu.support.encoding.CustomCodecs._
-  implicit val decoderDigital: Decoder[DigitalPack] = deriveDecoder[DigitalPack]
-    .prepare(_.withFocus(_.mapObject(_.checkKeyExists("readerType", Json.fromString("Direct")))))
```
johnduffell, line 61: we used to accept a missing readerType, but it looks like the frontend does send it.  So I have dropped that functionality as the standard `DiscriminatedType` doesn't have facility for special cases at the moment.
It wouldn't be too hard to add the extra complexity, but I don't know if it's too big of a risk to drop that?  Thoughts please.

johnduffell: this was added only in may for corporate subs, apparently just to get the tests working https://github.com/guardian/support-frontend/pull/2484/files#diff-6bc499707cdd7d347bfaa7ce1da6dcf4fc67d86af1c0ca30c801f4e3d1b704f1R64
and changed later to ReaderType 
https://github.com/guardian/support-frontend/pull/2635/files#r463085289

rupertbates: If it is always sent, which would make sense, then I think we should remove it

rupertbates (approved):
LGTM

PR #2872, opened 2020-12-03
make the test users page more useful

johnduffell:

This PR adds a nice link to the test users Wiki page.  Also makes it a bit clearer what to use for what.

Ideally it could just have a list of links with the details prefilled for each product, but that might be a job for another day!

https://trello.com/c/Og2MS2Ts/3450-add-link-from-test-users-page-to-test-users-wiki-page

## Why are you doing this?

When we were testing gift DS throughout the whole subs team, a lot of us were having trouble!  Devs and non devs alike.

## Screenshots

image


rupertbates (approved):
Nice improvements 👍

PR #2877, opened 2020-12-04
use identity id instead of sf contact id for most emails

johnduffell:

This PR removes SF contact where it is not needed, for sending emails.  It only remains for gift notification for DS, as we do not have an identity id for the giftee notification email address.

https://trello.com/c/IfG3s7tM/3505-user-identity-id-instead-of-email-address-for-sending-emails

## Why are you doing this?

This is because when the email gets to braze, it somehow messes things up if we send using an SF contact id instead of identity ID. @paulbrown1982 has requested we use identity id where possible.

## Testing

tested with the IT tests, also running the manual email send, all 9 emails arrived successfully.
ToDO, test in CODE manually once it's approved.

### DIFF
--- a/support-models/src/main/scala/com/gu/support/workers/states/SendThankYouEmailState.scala
+++ b/support-models/src/main/scala/com/gu/support/workers/states/SendThankYouEmailState.scala
@@ -22,7 +22,6 @@ object SendThankYouEmailState {
 
   case class SendThankYouEmailContributionState(
     user: User,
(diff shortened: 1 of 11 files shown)


paulbrown1982:
The only other use case I can think of where SfContactId will be needed is payment failure and cancellation emails of identity orphan subscribers.

rupertbates (approved):
LGTM

PR #2881, opened 2020-12-07
update to scala 2.13

johnduffell:

update to scala 213

Fix up any code changes needed to make it work.

Also some extra bumps e.g. play.
Bump any libraries that did not support scala 2.13 when we last bumped them.

## Why are you doing this?

Scala 2.13 is the latest stable version.  We want to be able to use the latest maintained libraries to keep on with security updates.  As time goes on other teams are not supporting scala 2.12 any more, also external dependencies e.g. play will drop support.  Scala 3 is out in Release candidate so we really should get up to date on this.

There are few significant changes that will affect us, mainly improvements in the compiler and performance, and also cross binary compatibility with scala 3 which will reduce a major headache.


build.sbt
```diff
@@ -32,12 +32,11 @@ lazy val release = Seq[ReleaseStep](
 
 inThisBuild(Seq(
   organization := "com.gu",
-  scalaVersion := "2.12.12",
+  scalaVersion := "2.13.5",
   dependencyTree / aggregate := false,
   // https://www.scala-sbt.org/1.x/docs/Cached-Resolution.html
   updateOptions := updateOptions.value.withCachedResolution(true),
   resolvers ++= Seq(Resolver.sonatypeRepo("releases")), // libraries that haven't yet synced to maven central
-  scalacOptions += "-Ypartial-unification",
```
johnduffell, line 40: this is the default in scala 2.13

johnduffell, line 40: this is the default in scala 2.13

build.sbt
```diff
@@ -196,7 +195,7 @@ lazy val `acquisition-event-producer` = (project in file("acquisition-event-prod
     bintrayOrganization := Some("guardian"),
     bintrayRepository := "ophan",
     publishMavenStyle := true,
-    addCompilerPlugin("org.scalamacros" % "paradise" % "2.1.1" cross CrossVersion.full), // for simulacrum
+    scalacOptions += "-Ymacro-annotations",
```
johnduffell, line 198: macro annotations are built in now, rather than an external plugin

project/LibraryVersions.scala
```diff
@@ -2,8 +2,9 @@
 object LibraryVersions {
   val circeVersion = "0.12.1"
   val awsClientVersion = "1.11.827"
-  val catsVersion = "1.4.0"
-  val jacksonVersion = "2.12.0"
+  val catsVersion = "2.1.1"
+  val jacksonVersion = "2.10.5"
+  val jacksonDatabindVersion = "2.10.5.1"
```
johnduffell, line 7: scala jackson module now throws an error if the jackson databind version is not the same minor version, so downgrading.
I do think that we need urgently to get onto play 2.8 and AWS client v2 asap, as this problem will be impossible to get out of soon, as jackson say this version is no longer supported and they will not release any more updates.

project/LibraryVersions.scala
```diff
@@ -2,8 +2,9 @@
 object LibraryVersions {
   val circeVersion = "0.12.1"
   val awsClientVersion = "1.11.827"
-  val catsVersion = "1.4.0"
-  val jacksonVersion = "2.12.0"
+  val catsVersion = "2.1.1"
```
johnduffell, line 5: if we go any further on the cats version, it pulls in a simulacrum library that has an assembly merge conflict, which I wasn't sure how to resolve, as one is a scalafix thing and the other is a real typeclass implementation.

project/plugins.sbt
```diff
@@ -7,7 +7,7 @@ addSbtPlugin("com.jsuereth" % "sbt-pgp" % "1.1.0")
 
 addSbtPlugin("org.xerial.sbt" % "sbt-sonatype" % "2.0")
 
-addSbtPlugin("com.typesafe.play" % "sbt-plugin" % "2.7.5") // when updating major version, also update play-circe version
+addSbtPlugin("com.typesafe.play" % "sbt-plugin" % "2.7.9") // when updating major version, also update play-circe version
```
johnduffell, line 10: this is only a patch update.  The comment only applies to the first two numbers e.g. 2.7->2.8

support-frontend/build.sbt
```diff
@@ -13,41 +13,42 @@ resolvers += "Guardian Platform Bintray" at "https://dl.bintray.com/guardian/pla
...
+  "com.gu" %% "tip" % "0.6.4",
   "com.googlecode.libphonenumber" % "libphonenumber" % "8.10.4",
   // This is required to force aws libraries to use the latest version of jackson
-  "com.fasterxml.jackson.core" % "jackson-databind" % jacksonVersion,
+  "com.fasterxml.jackson.core" % "jackson-databind" % jacksonDatabindVersion,
   "com.fasterxml.jackson.core" % "jackson-annotations" % jacksonVersion,
   filters,
   ws
 )
+dependencyOverrides += "com.fasterxml.jackson.core" % "jackson-databind" % jacksonDatabindVersion
```
johnduffell, line 51: this is needed to stop us getting onto a too late (and incompatible) version of databind, that is pulled in via identity.  We need to get onto play 2.8 to solve this.

support-payment-api/src/main/scala/services/ContributionsStoreService.scala
```diff
@@ -52,6 +53,9 @@ object ContributionsStoreQueueService {
     private implicit val messageEncoder = Encoder[Message] { message =>
       import io.circe.generic.auto._
 
+      // copied from earlier version of circe-core to prevent seconds appearing in the serialised form
+      implicit val encodeLocalDateTime: Encoder[LocalDateTime] = (a: LocalDateTime) => Json.fromString(a.toString)
```
johnduffell, line 57: one of the tests was failing due to the last part of the seriased form being `00:00:00` rather than the expected `00:00` so I had to add the old copy and pasted impelmentation.

support-workers/build.sbt
```diff
@@ -69,7 +69,7 @@ assemblyMergeStrategy in (IntegrationTest, assembly) := {
     oldStrategy(y)
 }
 IntegrationTest / assembly / test := {}
-IntegrationTest / assembly / aggregate := false
+assembly / aggregate := false
```
johnduffell, line 72: this is quite useful to apply to all assembly, this means you can run `support-workers / assembly` in sbt and not get all the sub-jars pulled in

support-frontend/app/services/MembersDataService.scala
```diff
@@ -79,6 +79,12 @@ class MembersDataService(apiUrl: String)(implicit val ec: ExecutionContext, wsCl
   }
 
   private def validateJson[T](json: JsValue)(implicit reader: Reads[T]) =
-    json.validate[T].asEither.leftMap(JsonValidationError.apply)
+    json.validate[T].asEither.leftMap(
+      (x: collection.Seq[(JsPath, collection.Seq[PlayJsonValidationError])]) => {
+        val apply1: Seq[(JsPath, Seq[PlayJsonValidationError])] => JsonValidationError = JsonValidationError.apply _
+        val x2 = x.toSeq.map({ case (k, v) => (k, v.toSeq) }) // needed until play json uses immutable Seq - see this page:
```
tomrf1, line 85: :(

support-models/src/main/scala/com/gu/support/encoding/DiscriminatedType.scala
```diff
@@ -6,7 +6,7 @@ import io.circe.generic.decoding.DerivedDecoder
 import io.circe.generic.encoding.DerivedAsObjectEncoder
 import io.circe.generic.semiauto.{deriveDecoder, deriveEncoder}
 import io.circe.syntax._
-import io.circe._
+import io.circe.{Codec => _, _}
```
tomrf1, line 9: this is odd!

tomrf1 (approved):
great work!

support-models/src/main/scala/com/gu/support/encoding/DiscriminatedType.scala
johnduffell: very odd! we have a "Codec" in the same package as this file.
Previously in scala 2.12 it picked that one.  Since scala 2.13 it now seems to prioritise the one in the import.
So I had to redirect the imported one to `_` so that only the required one is available.

support-frontend/build.sbt
johnduffell: at the time of this PR we were on play 2.7.  Now we are on play 2.8, but this problem apparently still exists, so maybe one for an even later version.

PR #2882, opened 2020-12-09
PRB A/B test

tomrf1:
We are bringing back the Payment Request button (the purple button in chrome).
We want to find out if it competes with Paypal, and if it impacts retention as well as AV.

Only some users are able to see the button because it depends on their browser and if they have payments set up. So for the test analysis we need the ability to filter both the control and the variant for only people who can see the button.
We do this by emitting a special `payment-request-button-loaded` event to ophan, to indicate that the user _could_ see it.
This cannot be done at the time that they're assigned to the test, because we do not know until later on if their browser supports the button.

### control
Screen Shot 2020-12-09 at 13 58 40

### variant
Screen Shot 2020-12-09 at 13 58 22

### DIFF
--- a/support-frontend/assets/helpers/abTests/abtestDefinitions.js
+++ b/support-frontend/assets/helpers/abTests/abtestDefinitions.js
@@ -34,6 +34,28 @@ export const tests: Tests = {
     seed: 1,
   },
 
+  stripePaymentRequestButtonDec2020: {
+    type: 'OTHER',
+    variants: [
+      {
+        id: 'control',
+      },
+      {
+        id: 'PRB',
+      },
+    ],
+    audiences: {
+      ALL: {
+        offset: 0,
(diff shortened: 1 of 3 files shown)


TomPretty (approved):
Looks good 👍 I was a little unsure why the old style amounts were there (i.e. not frequency buttons) but running locally it's all good, so I'm guessing that's just an old screenshot? (from the first time this was implemented?!)

tomrf1:
> Looks good 👍 I was a little unsure why the old style amounts were there (i.e. not frequency buttons) but running locally it's all good, so I'm guessing that's just an old screenshot? (from the first time this was implemented?!)

No these are new screenshots. It's the one-off contribution view

TomPretty:
Ahhhh ofc! 🙈

PR #2862, opened 2020-12-01
Hide option to start subscribing to Guardian Weekly on Christmas day

GHaberis:
Remove the 25 December date from the Guardian Weekly checkout. Customers who select this date will not receive a paper until the week after.

## Screenshots

Screenshot 2020-12-01 at 10 31 04

PR #2887, opened 2020-12-14
Prevent users from starting GW gift on Dec 25

GHaberis:

Remove the 25 December date from the Guardian Weekly **Gifting** checkout. Customers who select this date will not receive a paper until the week after.

Non-gifting PR previously merged: https://github.com/guardian/support-frontend/pull/2862

### DIFF
--- a/support-frontend/assets/pages/weekly-subscription-checkout/components/weeklyCheckoutFormGifting.jsx
+++ b/support-frontend/assets/pages/weekly-subscription-checkout/components/weeklyCheckoutFormGifting.jsx
@@ -230,6 +230,13 @@ function WeeklyCheckoutFormGifting(props: PropTypes) {
               <FieldsetWithError id="startDate" error={firstError('startDate', props.formErrors)} legend="Gift delivery date">
                 {days.map((day) => {
                   const [userDate, machineDate] = [formatUserDate(day), formatMachineDate(day)];
+                  const hideDate = new RegExp('-12-25$').test(machineDate);
+
+                  // Don't render input if Christmas day
+                  if (hideDate) {
+                    return null;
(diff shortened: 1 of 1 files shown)


support-frontend/assets/pages/weekly-subscription-checkout/components/weeklyCheckoutFormGifting.jsx
```diff
@@ -230,6 +230,13 @@ function WeeklyCheckoutFormGifting(props: PropTypes) {
               <FieldsetWithError id="startDate" error={firstError('startDate', props.formErrors)} legend="Gift delivery date">
                 {days.map((day) => {
                   const [userDate, machineDate] = [formatUserDate(day), formatMachineDate(day)];
+                  const hideDate = new RegExp('-12-25$').test(machineDate);
```
rupertbates, line 233: It would be neater to use filter on the collection here before mapping but I'll leave that up to you

rupertbates (approved):

support-frontend/assets/pages/weekly-subscription-checkout/components/weeklyCheckoutFormGifting.jsx
GHaberis: True, the fix was originally implemented as a semi-urgent request and I felt this was the quickest way to apply the fix and only call `formatMachineDate(day)` once. Now it looks like we should be replacing this logic to [utilise the fulfilment date calculator](https://trello.com/c/bovZtRnp/2788-utilise-fulfilment-date-calculator-to-centralise-logic-for-fulfilment-dates) so this is a temporary fix anyhow.

PR #2888, opened 2020-12-14
Fix for checkout selenium test: update inputId to be just 'id'

lucymonie:
## What are you doing in this PR?
This is a fix to reinstate a missing id for a selenium test.

## Why are you doing this?
Because of an error introduced in this PR: https://github.com/guardian/support-frontend/pull/2786

PR #2786, opened 2020-10-20
Roll out design system in subscriptions checkouts

lucymonie:
## Why are you doing this?
The design system is here, and it's time we had it in our checkouts.

[**Trello Card**](https://trello.com/c/vCUGarHe/3343-add-design-systems-components-to-subs-checkouts)

## Changes
* Input fields -> design system components
* Buttons -> design system components
* Radio buttons -> design system components
* Text area -> design system components
* Select -> design system components

I refactored the payment selector component so the PayPal button is now outside this component, which is consistent with other payment methods, and makes the keyboard navigation better. 

I have added an h1 to the pages to improve the screen reader navigation of the page, but it is not the perfect solution. There is a [trello card](https://trello.com/c/sVxQeQnZ/3504-add-h1s-to-checkout-pages) to improve this.

## What hasn't changed
* The [expandable link](https://theguardian.design/2a1e5182b/p/43c26b-link/b/048fd1/t/93efb9) isn't yet available in the DS, so I have updated the existing one to match the new look, and used DS fonts and buttons in the expanding sections.

## Accessibility test checklist
 - [x] [Tested with screen reader](https://accessibility.gutools.co.uk/testing/web/screen-readers/)
 - [x] [Navigable with keyboard](https://accessibility.gutools.co.uk/testing/web/keyboard-navigation/)
 - [x] [Colour contrast passed](https://accessibility.gutools.co.uk/testing/web/colour-contrast/)

## Screenshot
There are so many checkouts, I didn't want to include shots of all of them! Here is the paper subs card one, and a [link to live page, for comparison](https://support.theguardian.com/subscribe/paper/checkout?fulfilment=HomeDelivery&product=Sixday)

support thegulocal com_subscribe_paper_checkout_fulfilment=HomeDelivery product=Sixday(wide) (2)

Here is the DS gift one as well

support thegulocal com_subscribe_digital_checkout_gift_period=Quarterly(wide) (4)

PR #2889, opened 2020-12-14
Remove inputId and replace with id for selenium tests

lucymonie:
This is a second attempt to remove the inputIds from the code, so these can be passed in correctly and picked up by selenium. First attempt to fix in this PR: https://github.com/guardian/support-frontend/pull/2888

The issue was introduced in this PR: https://github.com/guardian/support-frontend/pull/2786


johnduffell (approved):
thanks! 👏

johnduffell:
this doesn't seem to be an actual PROD issue. 
 seems the issue is with this button on the paper DD checkout
image

there is a small span over the top of it, and selenium won't try to click the input as it noticed the click will go to the span instead of the input

image


also affects GW DD checkout

I'm not sure what the fix is as presumably that span is designed to be there for animation purposes or similar

but actually clicking in that area does work, so it's not an production issue

PR #2890, opened 2020-12-15
change zuora state to be a trait/sum type

johnduffell:

This PR follows on from https://github.com/guardian/support-frontend/pull/2792
where Rupert said
image
so after 5 months here is the next one.

### Data structures

This PR changes the zuora state to be a sum type rather than having a load of optional fields.  This provides cross validation that the correct fields are supplied together, and reduces the amount of "this should never happen" that we have to deal with in the business logic.

Specifically, there is a containing "Wrapper" state to hold the fields that are common across all products.  Under that is the product specific trait as mentioned above which only contains fields that do not apply to all products.

### Code structure

The structure has been changed to :
1. the main handler deals with the wrapper and then a main top level match statement that decides which product specific Handler class to call.
1. Each Handler class takes the product specific state, does the operations to subscribe/redeem as appropriate, and returns the product specific SendThankYouEmailState for the given product.
1. most products (not gift redemption) also have a Builder class which takes the product specific state, and returns a SubscribeItem suitable for passing to Zuora.  This was already existing in many cases, but it is now consistent and simplified.

There is a slight increase in the amount of model code and code to transform between the models.. However this is balanced by a reduction in complexity of the business logic itself.

## Why are you doing this?

The current code is hard to separate the different fields needed for different types of product.  This neatly segregates them in a type safe manner so that it becomes more maintainable.

## Future work
There are many other tidy ups possible, but the main ones relevant to this PR are
- removing the product specific fields from the `User` class
- removing the product specific fields from the Wrapper  (common) class that are only needed for the elasticsearch support-logs event.
- using circe's `withDiscriminator` instead of the DiscriminatedType class.

## Still to do
* All tests pass - DONE
* IT tests - DONE
* Testing in CODE:

- [x] contribution monthly DD
- [x] ds CC
- [x] ds gift
- [x] ds gift redemption
- [x] gw
- [x] gw gift
- [x] paper/voucher


support-models/src/main/scala/com/gu/support/workers/User.scala
```diff
@@ -11,13 +11,13 @@ case class User(
...
   billingAddress: Address,
-  deliveryAddress: Option[Address] = None,
+  deliveryAddress: Option[Address] = None,// TODO should later be moved out of User and into the relevant case class
   telephoneNumber: Option[String] = None,
   allowMembershipMail: Boolean = false,
   allowThirdPartyMail: Boolean = false,
   allowGURelatedMail: Boolean = false,
   isTestUser: Boolean = false,
-  deliveryInstructions: Option[String] = None
+  deliveryInstructions: Option[String] = None// TODO should later be moved out of User and into the relevant case class
```
johnduffell, line 20: this is home delivery paper specific so shouldn't be provided everywhere, but that's not for this PR.  Delivery address above is specific to physical products only.

support-models/src/main/scala/com/gu/support/workers/states/CreateZuoraSubscriptionState.scala
```diff
@@ -1,31 +1,100 @@
...
   user: User,
-  giftRecipient: Option[GiftRecipient],
   product: ProductType,
   analyticsInfo: AnalyticsInfo,
-  paymentMethod: Either[PaymentMethod, RedemptionData],
   firstDeliveryDate: Option[LocalDate],
   promoCode: Option[PromoCode],
-  salesforceContacts: SalesforceContactRecords,
   acquisitionData: Option[AcquisitionData]
 ) extends FailureHandlerState
```
johnduffell, line 25: (most) product specific state is removed and put into the productSpecificState field.

Some obvious ones are remaining and appear unused e.g. promo code and first devliery date.  This is because they are used by FailureHandlerState elasticsearch step as key-value pairs.  They are not needed as strongly typed data so can be moved into AnalyticsInfo with the other similar fields.  This can happen in another (reasonably small) PR rather than adding here.

support-models/src/main/scala/com/gu/support/workers/states/CreateZuoraSubscriptionState.scala
```diff
@@ -1,31 +1,100 @@
...
 import com.gu.support.workers.{PaymentMethod, SalesforceContactRecord, User, _}
 import org.joda.time.LocalDate
-import com.gu.support.encoding.CustomCodecs.{decodeLocalTime, encodeLocalTime}
+import com.gu.support.encoding.CustomCodecs.{decodeLocalTime, encodeLocalTime, decodeCountry, encodeCountryAsAlpha2}
+import com.gu.support.encoding.{Codec, DiscriminatedType}
 import com.gu.support.redemptions.RedemptionData
+import com.gu.support.workers.GiftRecipient.{DigitalSubscriptionGiftRecipient, WeeklyGiftRecipient}
 
-case class CreateZuoraSubscriptionState(
+case class CreateZuoraSubscriptionWrapperState(
```
johnduffell, line 16: the old state that covered everything optionally was called `CreateZuoraSubscriptionState`

Now that it's split into "state that you always have", and "product specific state", I have called them `CreateZuoraSubscriptionWrapperState` and `CreateZuoraSubscriptionState` respectively.  It is more of a minimal change.

However it would make more sense to keep the original name for the "wrapper" and call the product specific state `CreateZuoraSubscriptionProductSpecificState`

Thoughts about naming (or simplifying it conceptially) would be nice.  I couldn't really flatten it into a single level and retain the simplicity of the second level.

support-models/src/main/scala/com/gu/support/zuora/api/SubscriptionData.scala
```diff
@@ -176,7 +178,7 @@ case class Subscription(
   termType: String = "TERMED",
   readerType: ReaderType = ReaderType.Direct,
   promoCode: Option[PromoCode] = None,
-  redemptionCode: Option[RawRedemptionCode] = None,
+  redemptionCode: Option[Either[GeneratedGiftCode, RedemptionCode]] = None,
```
johnduffell, line 181: this still serialises the same.  This was so the "builder" is responsible for the codes and the thing that creates the next state can get hold of the value.
TODO I think on reflection this should be reverted as there is probably a neater way

support-workers/docs/error-handling.md
```diff
@@ -18,7 +18,7 @@ Most errors should be logged to Sentry so they can be investigated. However tran
 
 | Error type           | Description                                     | Retry?    |
 |----------------------|-------------------------------------------------|-----------|
-| Input parsing errors | Errors caused by invalid input JSON             | No        |
+| Input parsing errors | Errors caused by invalid input JSON             | Yes       |
```
johnduffell, line 21: I didn't know it was supposed to fail - in practice it actually retries, which is very useful if you are deploying a change to the format - as you can deploy support-fronetend first and then support-workers later.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateSalesforceContact.scala
```diff
@@ -32,20 +35,123 @@ class CreateSalesforceContact extends ServicesHandler[CreateSalesforceContactSta
...
-      state.product,
-      state.analyticsInfo,
-      state.paymentMethod,
-      state.firstDeliveryDate,
-      state.promoCode,
-      response,
-      state.acquisitionData
-    )
+}
+class NextState(state: CreateSalesforceContactState) {
```
johnduffell, line 39: this is all temporary code to turn the "old style" state into the product specific state.  Once CreateSalesforceContactState is converted, it will move upstream again.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -1,411 +1,149 @@
...
-import org.joda.time.{DateTime, DateTimeZone, Days, LocalDate}
+import org.joda.time.{DateTime, DateTimeZone}
 
 import scala.concurrent.ExecutionContext.Implicits.global
-import scala.concurrent.Future
-import scala.util.{Failure, Success, Try}
 
 class CreateZuoraSubscription(servicesProvider: ServiceProvider = ServiceProvider)
-  extends ServicesHandler[CreateZuoraSubscriptionState, SendAcquisitionEventState](servicesProvider) {
+  extends ServicesHandler[CreateZuoraSubscriptionWrapperState, SendAcquisitionEventState](servicesProvider) {
```
johnduffell, line 19: this file has changed (simplified!) so significantly that it's worth just viewing it alone rather than as a diff to ensure that it makes sense in its own right.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateSalesforceContact.scala
```diff
@@ -32,20 +35,123 @@ class CreateSalesforceContact extends ServicesHandler[CreateSalesforceContactSta
...
+    purchase: PaymentMethod
+  ): CreateZuoraSubscriptionWrapperState =
+    CreateZuoraSubscriptionWrapperState(CreateZuoraSubscriptionDigitalSubscriptionDirectPurchaseState(
+      user.billingAddress.country,
+      product,
+      purchase,
+      promoCode,
+      salesforceContactRecord,
+    ), requestId, user, product, analyticsInfo, firstDeliveryDate, promoCode, acquisitionData)
+
```
johnduffell, line 156: the file below (CreateZuoraSubscription.scala) is not rendered bydefault, but that is the major change in this PR! so be sure to look carefully.

support-workers/src/main/scala/com/gu/support/workers/lambdas/PreparePaymentMethodForReuse.scala
```diff
@@ -38,19 +37,25 @@ class PreparePaymentMethodForReuse(servicesProvider: ServiceProvider = ServicePr
       crmId <- getOrFailWithMessage(account.CrmId, s"Zuora account $accountId has not CrmId")
       paymentMethod <- toPaymentMethod(getPaymentMethodResponse, services.goCardlessService, account.PaymentGateway)
       sfContact = SalesforceContactRecord(sfContactId, crmId)
+      contribution <- Future.fromTry(state.product match {
+        case c: Contribution => Success(c)
+        case _ => Failure(new RuntimeException("not yet supported reuse payment method for other than contributions"))
```
johnduffell, line 42: reuse will fail by default for non contributions - before it would have appeared to have gone ahead (but not filled in various product specific fields eg delivery date/promo code/gift contact.)

support-workers/src/main/scala/com/gu/support/workers/lambdas/PreparePaymentMethodForReuse.scala
```diff
@@ -38,19 +37,25 @@ class PreparePaymentMethodForReuse(servicesProvider: ServiceProvider = ServicePr
...
+      })
     } yield HandlerResult(
-        CreateZuoraSubscriptionState(
-          requestId = UUID.randomUUID(),
```
johnduffell, line 43: this was replacing the request ID with a new one - I can't see why that was a good idea so have preserved the existing one

support-workers/src/main/scala/com/gu/zuora/ZuoraSubscriptionCreator.scala
```diff
@@ -0,0 +1,68 @@
+package com.gu.zuora
```
johnduffell, line 1: this file is various functionality removed and tidied up from CreateZuoraSubscription.scala main file.

support-workers/src/main/scala/com/gu/zuora/productHandlers/ZuoraContributionHandler.scala
```diff
@@ -0,0 +1,24 @@
+package com.gu.zuora.productHandlers
```
johnduffell, line 1: now each product type has a standard "...Handler" class that takes the product specific state and returns the next state.
Most of the Handler classes also refer to the existing ...Buidler classes that builds the zuora subcribe data.  However they have been tidied up for the new data types.

support-workers/src/main/scala/com/gu/zuora/ZuoraSubscriptionCreator.scala
```diff
@@ -0,0 +1,65 @@
...
+  now: () => DateTime,
+  userId: String,
+  requestId: UUID,
+) {
+
+  def preview(
+    subscribeItem: SubscribeItem,
+    billingPeriod: BillingPeriod,
+  ): Future[PaymentSchedule] =
+    PreviewPaymentSchedule.preview(subscribeItem, billingPeriod, zuoraService, checkSingleResponse)
```
johnduffell, line 27: we could always inline `PreviewPaymentSchedule.preview` one day as this is the only place it's used

johnduffell: same goes for the below `ensureSubscriptionCreated` function I'm sure it could get simplified some day

support-workers/src/main/scala/com/gu/zuora/productHandlers/ZuoraDigitalSubscriptionGiftRedemptionHandler.scala
```diff
@@ -0,0 +1,86 @@
+package com.gu.zuora.productHandlers
```
johnduffell, line 1: this is moved from CreateZuoraSubscription main file and minor tidying up but otherwise unchanged.

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/ContributionSubscriptionBuilder.scala
```diff
@@ -0,0 +1,29 @@
+package com.gu.zuora.subscriptionBuilders
```
johnduffell, line 1: these "...Builder" files already existed for some products and in some form, now they are flattened into a consistent set and given simpler and more consistent input/output parameters so that it's easy to jump to the right one and understand what happens in each case.

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionCorporateRedemptionBuilder.scala
```diff
@@ -55,6 +51,7 @@ class DigitalSubscriptionCorporateRedemptionBuilder(
 
     redeemedSubcription
       .map(subscription => subscriptionData.copy(subscription = subscription))
+      .map(subscriptionData => subscribeItemBuilder.build(subscriptionData, state.salesForceContact, None, None))
```
johnduffell, line 54: creating the SubscribeItem doesn't call any external services, so I moved it here rather than in the handler.

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/GuardianWeeklySubscriptionBuilder.scala
```diff
@@ -1,65 +1,71 @@
...
+      promotionProductRatePlanId,
+      subscriptionData
+    ).map { subscriptionData =>
+      val soldToContact = state.giftRecipient match {
+        case None =>
+          SubscribeItemBuilder.buildContactDetails(Some(state.user.primaryEmailAddress), state.user.firstName, state.user.lastName, state.user.deliveryAddress.get, None)
+        case Some(gR) =>
+          SubscribeItemBuilder.buildContactDetails(gR.email, gR.firstName, gR.lastName, state.user.deliveryAddress.get, None)
+      }
+      subscribeItemBuilder.build(subscriptionData, state.salesforceContacts.recipient, Some(state.paymentMethod), Some(soldToContact))
```
johnduffell, line 60: as with the others, creating the SubscribeItem doesn't call any external services, so I moved it here rather than in the handler.

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/ProductSubscriptionBuilders.scala
```diff
@@ -21,60 +21,13 @@ object ProductSubscriptionBuilders {
...
-      List(
-        RatePlanChargeData(
-          ContributionRatePlanCharge(contributionConfig.productRatePlanChargeId, price = contribution.amount) //Pass the amount the user selected into Zuora
-        )
-      ),
-      readerType = Direct
-    )
-  }
-
-  def buildProductSubscription(
```
johnduffell, line 38: this function moved to the SubscribeItemBuilder, as it needs the `createdRequestId` which is not product-specific state, so isn't available to the individual builders any more.

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/SubscribeItemBuilder.scala
```diff
@@ -1,53 +1,109 @@
...
+    subscriptionData: SubscriptionData,
+    salesForceContact: SalesforceContactRecord,
+    maybePaymentMethod: Option[PaymentMethod],
+    soldToContact: Option[ContactDetails]
+  ): SubscribeItem = {
+    val billingEnabled = maybePaymentMethod.isDefined
     SubscribeItem(
-      account = buildAccount(state),
-      billToContact = buildContactDetails(state.user, None, state.user.billingAddress),
-      soldToContact = state.user.deliveryAddress map (buildContactDetails(state.user, state.giftRecipient, _, state.user.deliveryInstructions)),
```
johnduffell, line 15: bill contact is not product specific, but sold to contact is, so these have been refactored as approriate

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/SubscribeItemBuilder.scala
```diff
@@ -1,53 +1,109 @@
...
+        redemptionCode = redemptionCode,
+        giftNotificationEmailDate = giftNotificationEmailDate,
+      )
+    )
+
+}
+
+object SubscribeItemBuilder {
+
+  def buildContactDetails(
```
johnduffell, line 90: this is now just a smart constructor for `ContactDetails` so we could call this `ContactDetails.apply`, the onyl issue there was scala can't accept them both having default arguments.  WE could call that another name e.g. `ContactDetails.create`, but I have stuck it here for now.

support-models/src/main/scala/com/gu/support/workers/states/CreateZuoraSubscriptionState.scala
```diff
@@ -1,31 +1,100 @@
...
+    firstDeliveryDate: LocalDate,
+    promoCode: Option[PromoCode],
+    salesforceContacts: SalesforceContactRecords,
+  ) extends CreateZuoraSubscriptionState
+
+  private val discriminatedType = new DiscriminatedType[CreateZuoraSubscriptionState]("productType")
+  implicit val codec: Codec[CreateZuoraSubscriptionState] = discriminatedType.codec(List(
+    discriminatedType.variant[CreateZuoraSubscriptionContributionState]("Contribution"),
+    discriminatedType.variant[CreateZuoraSubscriptionDigitalSubscriptionDirectPurchaseState]("DigitalSubscriptionDirectPurchase"),
+    discriminatedType.variant[CreateZuoraSubscriptionDigitalSubscriptionGiftPurchaseState]("DigitalSubscriptionGiftPurchase"),
```
rupertbates, line 93: These discriminator strings could be shared as vals between this class and `SendThankYouEmailState`

support-models/src/main/scala/com/gu/support/workers/states/CreateZuoraSubscriptionState.scala
rupertbates: I think I'd rather keep `CreateZuoraSubscriptionState` as the outer class as it maps to the lambda name and then call the inner class something like `CreateZuoraSubscriptionProductState`.

support-models/src/main/scala/com/gu/support/workers/states/CreateZuoraSubscriptionState.scala
```diff
@@ -1,31 +1,100 @@
...
+  implicit val codec = deriveCodec[CreateZuoraSubscriptionWrapperState]
+}
+
+sealed trait CreateZuoraSubscriptionState
 
 object CreateZuoraSubscriptionState {
-  import com.gu.support.encoding.CustomCodecs._
-  implicit val codec: Codec[CreateZuoraSubscriptionState] = deriveCodec
+
+  case class CreateZuoraSubscriptionContributionState(
```
rupertbates, line 35: How do you feel about dropping `CreateZuoraSubscription` from the start of all these class names? They are all inside the `CreateZuoraSubscriptionState` object so it may not really be needed and it does make the names very verbose.

rupertbates: It would be nice if we could genericise these classes in some way like 
`CreateZuoraSubscriptionState[Product, ReaderType, AcquisitionType]` but that's not going to give us the varying structures we want is it?

support-models/src/main/scala/com/gu/support/workers/states/CreateZuoraSubscriptionState.scala
johnduffell: yeah that does sound like a better way, will do that

support-models/src/main/scala/com/gu/support/workers/states/CreateZuoraSubscriptionState.scala
johnduffell: there's no technical reason why they need to be the same, but I can see we want consistency so I will do that.

support-models/src/main/scala/com/gu/support/workers/states/CreateZuoraSubscriptionState.scala
johnduffell: yep I think the verboseness is definitely out of control.  I will try that - as long as we don't end up clashing too much with the preceding state and next state.
Regarding the genericising it,  I agree that there is commonality across many of them and the we could represent those similarities in some way, e.g. more complex hierarchy with `trait DeliverableSubscription` or a deeper tree structure and all the rest, that would increase the reuse and communicate more to the compiler, but having gone down that route on the last refactor it ended up being over complicated.
I quite like the nice flat self contained set of products and I think it does get most of the benefit without being totally confusing for new people.  But have a think on it and there may well be a middle ground.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateSalesforceContact.scala
johnduffell: image

support-models/src/main/scala/com/gu/support/workers/states/CreateZuoraSubscriptionState.scala
rupertbates: I notice that the product specific state does have a `product` field in it which duplicates the top level product, is it not straightforward to switch code over to using this?

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -45,367 +27,121 @@ class CreateZuoraSubscription(servicesProvider: ServiceProvider = ServiceProvide
...
-      services.config.zuoraConfigProvider.get(state.user.isTestUser),
-      state.user.isTestUser,
-    )
-
-    maybeDigitalSubscriptionGiftRedemption(state.product, state.paymentMethod) match {
-      case Some(redemptionData) =>
-        redeemGift(redemptionData, requestInfo, state, services.zuoraGiftService, services.catalogService)
-      case None =>
-        createSubscription.create(state, requestInfo)
+    val zuoraProductHandlers = new ZuoraProductHandlers(services, state)
```
rupertbates, line 30: Seems a bit wasteful to construct 7 objects when we're only going to use one. Could you move the construction of the appropriate handler class into the match statement below?
Eg.
```scala
case redemptionState: DigitalSubscriptionGiftRedemptionState =>
        new ZuoraDigitalSubscriptionGiftRedemptionHandler(
          services.zuoraGiftService,
          services.catalogService,
          state.user,
          state.requestId,
        ).redeemGift(redemptionState)
```

support-models/src/main/scala/com/gu/support/workers/states/CreateZuoraSubscriptionState.scala
johnduffell: yes, everything is switched over within this lambda, the only reason it is still at the top level is because this state needs to deserialise directly as FailureHandlerState, as FailureHandler it needs to send the correct failure email to the user.
This won't be a problem to fix to use a suitable enum or even the productType value, but I think it makes more sense to do it separately.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -45,367 +27,121 @@ class CreateZuoraSubscription(servicesProvider: ServiceProvider = ServiceProvide
...
-        case otherState: CodeStatus => Future.failed(new RuntimeException(otherState.clientCode))
-      }
-      handlerResult <- Future.fromTry(buildHandlerResult(state, requestInfo, termDates))
-    } yield handlerResult
-  }
+  val zuoraDigitalSubscriptionGiftPurchaseHandler = new ZuoraDigitalSubscriptionGiftPurchaseHandler(
+    zuoraSubscriptionCreator, now,
+    new DigitalSubscriptionGiftPurchaseBuilder(
+      services.promotionService,
+      () => now().toLocalDate,
```
rupertbates, line 91: Why pass a function rather than a LocalDate here?

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -45,367 +27,121 @@ class CreateZuoraSubscription(servicesProvider: ServiceProvider = ServiceProvide
...
-        acquisitionData = state.acquisitionData
-      ), requestInfo)
-    )
-  }
+  val zuoraDigitalSubscriptionCorporateRedemptionHandler = new ZuoraDigitalSubscriptionCorporateRedemptionHandler(
+    zuoraSubscriptionCreator,
+    CorporateCodeStatusUpdater.withDynamoUpdate(services.redemptionService),
+    new DigitalSubscriptionCorporateRedemptionBuilder(
+      CorporateCodeValidator.withDynamoLookup(services.redemptionService),
+      () => now().toLocalDate,
```
rupertbates, line 104: Why a function?

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
johnduffell: agree and I would like to improve that - it just seemed to obscure the elegance of the match statement, I will have a try again.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
johnduffell: just to give us control over when the date is picked.  but I think you're right in practical terms it may not make a difference.

support-workers/src/main/scala/com/gu/zuora/ZuoraSubscriptionCreator.scala
rupertbates: Why move it here? Just for file size reasons?

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/ProductSubscriptionBuilders.scala
```diff
@@ -1,17 +1,17 @@
 package com.gu.zuora.subscriptionBuilders
 
 import java.util.UUID
```
rupertbates, line 3: Nitpick: There are quite a few unused imports here

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/SubscriptionDataBuilder.scala
```diff
@@ -1,97 +1,8 @@
...
-import scala.concurrent.ExecutionContext.Implicits.global
-import scala.concurrent.Future
+import com.gu.support.promotions.PromoError
+import com.gu.support.redemption.InvalidCode
 
 case class BuildSubscribePromoError(cause: PromoError) extends RuntimeException
 
 case class BuildSubscribeRedemptionError(cause: InvalidCode) extends RuntimeException
-
-class SubscriptionDataBuilder(
```
rupertbates, line 22: Now this class is no longer here the file name should change. Maybe to `BuildSubscribeErrors`?

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
rupertbates: I think the simple approach should win out here. Otherwise anyone going through this code will spend time wondering about it.

support-workers/src/main/scala/com/gu/zuora/ZuoraSubscriptionCreator.scala
johnduffell: that and it didn't seem to have any reason to be in there as it's more like detail, not really part of the view from 10000 feet enough

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
johnduffell: anonymous functions are indeed too confusing, but I want to make sure the date is defintely picked at the moment of execution, rather than when the services are created, which would cause subtle data bugs in zuora.  So I have moved to a normal class for clarity in the commit I just pushed.

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/ProductSubscriptionBuilders.scala
johnduffell: done

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/SubscriptionDataBuilder.scala
johnduffell: done and I also moved to the handlers package where it's used rather than builders

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
johnduffell: see above comment, it's a class now

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
johnduffell: this is still to do and should be reasonably self contained, all the other comments are resolved

support-workers/src/main/scala/com/gu/support/workers/lambdas/PreparePaymentMethodForReuse.scala
rupertbates, line 42:
```suggestion
        case _ => Failure(new RuntimeException("Reusing payment methods is not yet supported for products other than contributions"))
```

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
johnduffell: will change all to lazy val (after discussion)

rupertbates (approved):
👍 nice work!

johnduffell:
all testing completed in CODE -
the zuora account and subscription look right, and the confirmation emails all look right.  Let's see how it goes in PROD!

PR #2891, opened 2020-12-16
New acquistions data pipeline

rupertbates:

We currently send all our acquisition data to the data lake via Ophan using the Thrift and the acquisition-event-producer project. This has a number of disadvantages:
- Changes to the schema are very difficult to make as they need to be propagated through a number of systems including libraries which need to be deployed to Sonatype
- Thrift is quite a heavyweight dependency to load in a lambda
- It takes 24hrs for data to appear in the data lake 
- There are a quite ridiculous number of systems involved in actually getting what is quite simple data into its final destination in the data lake

To solve these issues we are going to switch to writing data directly to a new table in the datalake dataset in BigQuery. This means that events will be available instantaneously and schema updates only have to happen in one repo.

This is the first PR related to this work, the full list of work to complete is below:
- [x] Send subscriptions to BigQuery
- [ ] Send one off contributions to BigQuery
- [ ] Send Zuora subscription id and account id information to BigQuery

[**Trello Card**](https://trello.com/c/4N6xz0Nh/3367-acquisitions-data-pipeline-work-15)


support-modules/bigquery/src/main/scala/com/gu/support/acquisitions/BigQueryService.scala
```diff
@@ -0,0 +1,47 @@
...
+      ))
+      .build().getService
+
+  def tableInsertRow(datasetName: String, tableName: String, acquisitionDataRow: AcquisitionDataRow): Either[String, Unit] = {
+    try {
+      val tableId = TableId.of(datasetName, tableName)
+      val rowContent = AcquisitionDataRowMapper.mapToTableRow(acquisitionDataRow)
+      val insertRequest = InsertAllRequest.newBuilder(tableId).addRow(rowContent).build
+
+      val response = bigQuery.insertAll(insertRequest) // Seems there is currently no Async way to do this
```
johnduffell, line 31: if you want to get control back straight away so you can run something else concurrently you can do something like this
```suggestion
      Future(bigQuery.insertAll(insertRequest)).map { response =>
```

support-workers/src/main/scala/com/gu/acquisitions/AcquisitionDataRowBuilder.scala
```diff
@@ -0,0 +1,215 @@
...
+
+  private def getAcquisitionTypeDetails(s: SendThankYouEmailState): AcquisitionTypeDetails =
+    s match {
+      case s: SendThankYouEmailContributionState => AcquisitionTypeDetails(
+        Some(s.paymentMethod),
+        None,
+        Direct,
+        Purchase,
+        Some(s.accountNumber),
+        None //TODO: we should have a sub number for this
```
johnduffell, line 129: interesting - I stopped passing them through for products where they were not used, was that a regression?

support-workers/src/test/scala/com/gu/acquisitions/BigQuerySpec.scala
```diff
@@ -0,0 +1,80 @@
...
+        config.privateKeyId,
+        Nil.asJavaCollection
+      ))
+      .build().getService
+
+  "BigQuery" should "be able to run a query" in {
+    val query = s"""select * from ${BigQuerySchema.datasetName}.${BigQuerySchema.tableName} where amount = 9999 and event_timestamp > TIMESTAMP("2020-12-14 00:20:00");"""
+    val queryConfig = QueryJobConfiguration.newBuilder(query).build
+
+    val tableResult = bigQuery.query(queryConfig)
```
johnduffell, line 41: do we need query functionality to insert something?  I can't see where this is used other than the test

johnduffell: perhaps we could combine this with the below test, and query the record we just inserted?

support-workers/src/main/scala/com/gu/services/Services.scala
```diff
@@ -41,5 +42,6 @@ class Services(isTestUser: Boolean, val config: Configuration) {
   lazy val redemptionService = RedemptionTable.forEnvAsync(TouchPointEnvironments.fromStage(stage, isTestUser))
   lazy val catalogService = CatalogService(TouchPointEnvironments.fromStage(stage, isTestUser))
   lazy val giftCodeGenerator = new GiftCodeGeneratorService
+  lazy val bigQueryService = new BigQueryService(bigQueryConfigProvider.get(isTestUser))
```
johnduffell, line 45: we do have a lot in this Services class, it's reminding me strongly of the TouchPointBackend class now! 😢

johnduffell (approved):
looks great, there's a lot of code there but I think it's mostly detailed models and getting it into the right format.
Is the idea to retire the "Old" lambda once the new one is being relied on?  If so the name is fine, but if we want to keep it we need a better name.
Nice work getting through all this, will be nice to see the result in getting rid of the thrift binaries/etc from the JAR for all the lambdas, and breaking the chain via Ophan/AWS. 👏 

👍

support-workers/src/main/scala/com/gu/acquisitions/AcquisitionDataRowBuilder.scala
rupertbates: Well we didn't use them before but the Data Design team are keen to have them so they can join to the zuora export tables in the data lake

support-workers/src/test/scala/com/gu/acquisitions/BigQuerySpec.scala
rupertbates: I added that test while developing just to test the connection was working correctly, I'm happy to remove it if you think it would be better?

support-workers/src/main/scala/com/gu/services/Services.scala
rupertbates: Yeah it is quite big, could do with a refactor at some point.

rupertbates:
Yes the idea is to get rid of the old lambda, we are still a little way off that though

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -241,9 +241,21 @@ class NextState(
     }
   // scalastyle:on cyclomatic.complexity
 
+  private def contribution(product: Contribution, purchase: PaymentMethodWithSchedule) =
```
johnduffell, line 244:
```suggestion
  private def contribution(product: Contribution, paymentMethod: PaymentMethod) =
```

support-workers/src/test/scala/com/gu/acquisitions/BigQuerySpec.scala
johnduffell: probably not useful in itself, although I did wonder if the "insert" call returns success, how do we know it actually arrived in bigquery, unless we query back the record.
So my suggestion would be remove this test, and query back in the "insert" test to be sure it actually arrived.

support-workers/cloud-formation/src/templates/cfn-template.yaml
```diff
@@ -343,7 +343,7 @@ Resources:
                 - Name: PaymentProvider
                   Value: Stripe
                 - Name: ProductType
-                  Value: DigitalPack
+                  Value: DigitalSubscription
```
johnduffell, line 346: I think the product type is still coming through as DigitalPack, so the alarm is going off but the data is still coming through.

johnduffell: https://github.com/guardian/support-frontend/pull/2899

support-workers/src/main/scala/com/gu/acquisitions/AcquisitionDataRowBuilder.scala
```diff
@@ -0,0 +1,215 @@
...
+        None, //TODO: if we rework digital gift modelling in Zuora we should include the relevant ids here
+        None
+      )
+    }
+
+  private def buildLabels(state: SendAcquisitionEventState, accountExists: Boolean) =
+    Set(
+      if (accountExists) Some("REUSED_EXISTING_PAYMENT_METHOD") else None,
+      if (isSixForSix(state)) Some("GUARDIAN_WEEKLY_SIX_FOR_SIX") else None,
+      if (state.analyticsInfo.isGiftPurchase) Some("GIFT_SUBSCRIPTION") else None,
```
johnduffell, line 185: changing this from gift-subscription to GIFT_SUBSCRIPTION broke the flagging in GA of gift subs because it relied on this string https://github.com/guardian/acquisition-event-producer/pull/50/files#diff-eadf20916668eea4ad74100913282140802379c1d17a35411e24832732b7868bR89

PR #2892, opened 2020-12-16
use acquisition-event-producer

tomrf1:
ophan is now returning 400s since a play upgrade.
This change makes payment-api use the same dependency as support-workers, which is still working

I cannot test the actual ophan integration until it's in PROD, but have confirmed it still creates the correct payload in CODE


paulbrown1982 (approved):
👍  Benefit of making payment-api part of the mono-repo!

PR #2893, opened 2020-12-16
encode ophan query string

tomrf1:


kelvin-chappell (approved):
👍 
but I thought OKHTTP encodes query params?

paulbrown1982:
It doesn’t encode `{` or `}` which Ophan seems not to accept from 10th. We need to check that this won’t double-encode.

PR #2895, opened 2020-12-17
Quick fix for LP layout bug

TomPretty:

Quick fix for a visual bug on the LP. This was caused by an update to source. The css we're using is a little hacky - it assumes knowledge of the internal structure of the `Radio` component. This css broke when source swapped out a `div` element for a `span` element. 

We should come up with a proper fix for this, but as it stands I don't think we can quite replicate the current behaviour without this hacky css. The tricky part to replicate is having the whole area between the label text and the image being clickable. That might require a change to source. In the mean time I think this quick fix will work. This is also potentially an argument for having some sort of visual regression testing (chromatic?).

### current (broken)
Screenshot 2020-12-17 at 15 01 16

### fixed
Screenshot 2020-12-17 at 15 01 22


paulbrown1982 (approved):
👍

PR #2896, opened 2020-12-17
Temporary fix to stop the tests failing on direct debit subscription payments

i-hardy:

This stops click events aimed at the Source checkbox component used to confirm that the user can authorise a direct debit from getting intercepted by a `<span>` element that overlays the input to provide the nice ticking animation.

This is a **temporary** fix- this ought to be fixed in Source itself and I'll be raising a PR to that effect- but will stop the post-deploy tests from constantly failing.


rupertbates (approved):
👍

PR #2897, opened 2020-12-18
Fix the Selenium tests for the paper landing page

i-hardy:

This adds an ID to the main element on the paper landing page for the tests to select, as we're now styling with Emotion and the old class name is obsolete.


johnduffell (approved):
great, that will be a lot more reliable 👍 thanks

PR #2898, opened 2020-12-18
Prevent double submission of Stripe payment form

i-hardy:

On hitting the 'Pay Now' button when paying by credit card, the button will now be disabled while we retrieve the token from Stripe. The button will be re-enabled on any error, and at the point where the 'Processing' overlay is shown.

[**Trello Card**](https://trello.com/c/eMuYMAnH)

## Why are you doing this?

This fixes an issue where users were occasionally repeatedly clicking the subscribe button before the 'Processing' overlay appeared, causing errors in processing the payment.

### DIFF
--- a/support-frontend/assets/components/subscriptionCheckouts/stripeForm/stripeForm.jsx
+++ b/support-frontend/assets/components/subscriptionCheckouts/stripeForm/stripeForm.jsx
@@ -113,6 +113,7 @@ const StripeForm = (props: StripeFormPropTypes) => {
       errorIncomplete: 'Please enter a valid CVC number',
     },
   });
+  const [disableButton, setDisableButton] = useState<boolean>(false);
 
   const stripe = stripeJs.useStripe();
   const elements = stripeJs.useElements();
@@ -257,6 +258,13 @@ const StripeForm = (props: StripeFormPropTypes) => {
     }
   };
 
+  const handleCardSetupAndPay = () => handleCardSetup(setupIntentClientSecret)
+    .then(props.setStripePaymentMethod)
+    .then(() => {
+      setDisableButton(false);
+      props.submitForm();
+    });
+
   const requestSCAPaymentMethod = (event) => {
     event.preventDefault();
     props.validateForm();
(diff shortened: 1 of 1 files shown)


support-frontend/assets/components/subscriptionCheckouts/stripeForm/stripeForm.jsx
```diff
@@ -257,17 +258,26 @@ const StripeForm = (props: StripeFormPropTypes) => {
...
     checkRecaptcha();
 
     if (stripe && props.allErrors.length === 0 && cardErrors.length === 0 && !recaptchaError) {
+      setDisableButton(true);
       if (setupIntentClientSecret) {
-        handleCardSetup(setupIntentClientSecret)
-          .then(props.setStripePaymentMethod)
-          .then(() => props.submitForm());
+        handleCardSetupAndPay();
       } else if (recaptchaCompleted) {
```
GHaberis, line 281: From what I can see we only enable a disabled submit button (`setDisableButton(false)`) when the `handleCardSetupAndPay` call is made from within the `if` block of this `if...else if` statement. 

If the code were to go into the `else if` block here is there any other scenario that could follow where we'd need to enable a disabled submit button? Or at this point are we confident the only course of action is the form will submit successfully? For example, in this `else if` block we `setPaymentWaiting(true)`, however in the event the `handleStripeError` function is called we undo this (`setPaymentWaiting(false)`), would we also need to do the same for the submit button (`setDisableButton(false)`) in the event of a stripe error?

i-hardy: If the `else if` block is entered, it means we're still waiting for `setupIntentClientSecret` to be available. When it becomes available the second `useEffect` hook will run, and if this `else if block _has_ been entered and set `paymentWaiting` to `true`, that hook will call `handleCardSetupAndPay`.

`handleStripeError` is only ever called in `handleCardSetup`, which is itself only called in `handleCardSetupAndPay`, which will always re-enable the button in the case of either success or error.

GHaberis: OK thanks for clarifying this @i-hardy 👍

GHaberis (approved):
Looks good to me 👍

PR #2900, opened 2020-12-18
Fix some mapping bugs in AcquisitionDataRowMapper

rupertbates:
#2891 started sending subscription acquisition data to BigQuery, however there were a couple of mapping bugs which means that some data isn't being be saved to BigQuery correctly. This PR fixes this.

### DIFF
--- a/support-modules/bigquery/src/main/scala/com/gu/support/acquisitions/AcquisitionDataRowMapper.scala
+++ b/support-modules/bigquery/src/main/scala/com/gu/support/acquisitions/AcquisitionDataRowMapper.scala
@@ -14,9 +14,9 @@ object AcquisitionDataRowMapper {
       acquisition.promoCode.map("promo_code" -> _),
       acquisition.paymentProvider.map("payment_provider" -> paymentProviderName(_)),
       acquisition.printOptions.map(p => "print_options" -> Map(
-        "product" -> p.product,
+        "product" -> p.product.value,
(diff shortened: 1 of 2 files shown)


support-modules/bigquery/src/main/scala/com/gu/support/acquisitions/AcquisitionDataRowMapper.scala
```diff
@@ -14,9 +14,9 @@ object AcquisitionDataRowMapper {
       acquisition.promoCode.map("promo_code" -> _),
       acquisition.paymentProvider.map("payment_provider" -> paymentProviderName(_)),
       acquisition.printOptions.map(p => "print_options" -> Map(
-        "product" -> p.product,
+        "product" -> p.product.value,
         "delivery_country_code" -> p.deliveryCountry.alpha2
-      )),
+      ).asJava),
```
paulbrown1982, line 19: Why does this need to be converted to a Java map?

paulbrown1982 (approved):
👍  LGTM, just asking a question for interest.

support-modules/bigquery/src/main/scala/com/gu/support/acquisitions/AcquisitionDataRowMapper.scala
rupertbates: It's a Java sdk and that's what it requires unfortunately otherwise it fails silently 🤦

PR #2904, opened 2020-12-24
Add post-deployment tests for digital subscription gifting

i-hardy:

This adds post-deployment Selenium tests for the digital subscription gift landing page and checkout. It also updates the way we determine the billing period in the digital subs checkout in order to enable the tests to run without being dependent on a query string in the URL.

[**Trello Card**](https://trello.com/c/GbmnFfY0)

### DIFF
--- a/support-frontend/test/selenium/subscriptions/pages/DigitalPackGiftCheckout.scala
+++ b/support-frontend/test/selenium/subscriptions/pages/DigitalPackGiftCheckout.scala
@@ -4,7 +4,7 @@ import org.openqa.selenium.WebDriver
 import selenium.util.Config
 
 class DigitalPackGiftCheckout(implicit val webDriver: WebDriver) extends CheckoutPage {
-  val url = s"${Config.supportFrontendUrl}/subscribe/digital/checkout/gift?period=Quarterly"
+  val url = s"${Config.supportFrontendUrl}/subscribe/digital/checkout/gift"
 
   private val giftRecipientFirstName = id("firstNameGiftRecipient")
   private val giftRecipientLastName = id("lastNameGiftRecipient")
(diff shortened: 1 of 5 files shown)


support-frontend/test/selenium/subscriptions/pages/DigitalPackGiftCheckout.scala
```diff
@@ -0,0 +1,25 @@
+package selenium.subscriptions.pages
+
+import org.openqa.selenium.WebDriver
+import selenium.util.Config
+
+class DigitalPackGiftCheckout(implicit val webDriver: WebDriver) extends CheckoutPage {
+  val url = s"${Config.supportFrontendUrl}/subscribe/digital/checkout/gift"
+
+  private val giftRecipientFirstName = id("firstNameGiftRecipient")
```
johnduffell, line 9: not a blocker, just wondering, should we query these by class and put a "qa-something" class on them?

johnduffell (approved):
looks good, thanks for this! 👏 
is there another PR coming for the redemption form?

support-frontend/test/selenium/subscriptions/pages/DigitalPackGiftCheckout.scala
i-hardy: I'd generally favour using IDs over class names; we're gradually phasing out named classes in general as we migrate to Emotion. I did think about using data attributes as QA-specific selectors (eg. data-test="firstNameGiftRecipient"), but I couldn't figure out how to use non-ID or -class selectors with ScalaTest.

i-hardy:
> is there another PR coming for the redemption form?

...There's going to need to be as I completely forgot about the redemption form when working on this! 🤦

PR #2906, opened 2021-01-04
Fix payment-api switches permission

tomrf1:
We're getting a lot of ERROR logs for payment-api:
`Error udpating cache for recaptchaSwitch:. com.amazonaws.services.s3.model.AmazonS3Exception: Access Denied`
This particular error isn't actually causing any problems, but we shouldn't have ERRORs!

This is because the cloudformation grants permission to s3 objects under `/support-admin-console/PROD/*`, but PROD actually requires the CODE object as well.


paulbrown1982 (approved):
👍

PR #2908, opened 2021-01-07
Bigquery contributions

rupertbates:
#2891 Added acquisition tracking to BigQuery for subscriptions and recurring contributions, this PR does the same for single contributions via the payment-api

[**Trello Card**](https://trello.com/c/SrukASHo/3536-acquisitions-data-pipeline-work-contributions-0)


build.sbt
```diff
@@ -124,8 +124,8 @@ lazy val `support-payment-api` = (project in file("support-payment-api"))
     buildInfoPackage := "app",
     buildInfoOptions += BuildInfoOption.ToMap,
     libraryDependencies ++= commonDependencies
-  ).dependsOn(`support-models`, `support-internationalisation`, `acquisition-event-producer`)
```
rupertbates, line 127: acquisition-event-producer is brought in by support-models

support-config/src/main/scala/com/gu/support/config/AmazonPayConfig.scala
```diff
@@ -5,7 +5,7 @@ import com.typesafe.config.Config
 case class AmazonPayConfig(
   sellerId: String,
   clientId: String
-) extends TouchpointConfig
```
rupertbates, line 8: Removing the requirement for these config classes to extend `TouchpointConfig` means that payment-api doesn't need to depend on support-config which is not really relevant for it.

support-payment-api/src/main/scala/backend/AmazonPayBackend.scala
```diff
@@ -124,6 +129,14 @@ class AmazonPayBackend(cloudWatchService: CloudWatchService,
     ophanService.submitAcquisition(acquisition)
       .bimap(BackendError.fromOphanError, _ => ())
 
+  private def submitAcquisitionToBigQuery(
+    acquisition: AmazonPayAcquisition,
+    contributionData: ContributionData
+  ): EitherT[Future, BackendError, Unit] =
+    EitherT.fromEither[Future](bigQueryService.tableInsertRow(
```
tomrf1, line 136: Any idea if we need to worry about this being a blocking call?
Currently we return a response to the client after the `trackContribution` tasks have finished, even though their result does not affect the response. Which is perhaps not ideal

rupertbates: This is a good point, I wasn't too worried about it in a lambda but on an EC2 box we probably need to be a bit more careful. I've wrapped the call in a blocking future as suggested here:
https://docs.scala-lang.org/overviews/core/futures.html#blocking-inside-a-future

tomrf1: thanks!

tomrf1 (approved):
Looks good!
I just tested a single contribution in CODE and there were no errors

PR #2910, opened 2021-01-08
return list of interesting cookies for diagnostics purposes

johnduffell:

This PR adds a diagnostics endpoint to return several useful cookies in the response.  The user can then forward to us for debugging reasons.
We do have to be careful, as we are turning cookies from the server side into a response, so we don't want to make anything accsible via XSS or other JS that wouldn't otherwise be.  Therefore I have hard coded a list of allowed cookies which should not contain anything sensisive.  For example, it would be a very bad idea to return all cookies, because that would include the SC_GU_U identity log in cookie in the list allowing someone to impersonate the user if they got hold of it.

## Why are you doing this?

There seem to be a lot of strange issues with ad free and support asks.  Members data api generally seems to be ok, so there may be some client side issue or cookies issue.
This will give us more visibility of the cookies so we can help users a bit more.
They can send the output to userhelp and we can have a look from our end.

## Screenshots
```
* gu.contributions.recurring.contrib-timestamp.Monthly
 ( domain: None, httpOnly: true, path: /, maxAge: None, sameSite: None, secure: false )
 = 1607002648056
```

### DIFF
--- a/support-frontend/app/controllers/DiagnosticsController.scala
+++ b/support-frontend/app/controllers/DiagnosticsController.scala
@@ -1,0 +1,39 @@
+package controllers
+
+import actions.CustomActionBuilders
+import play.api.mvc._
+import Results.Ok
+
+class DiagnosticsController(
+  actionRefiners: CustomActionBuilders,
+) {
+
+  val relevantCookies = List(
+    "gu_user_features_expiry",
+    "gu_paying_member",
+    "GU_AF1",
+    "gu_action_required_for",
+    "gu_digital_subscriber",
+    "gu_hide_support_messaging",
+    "gu_recurring_contributor",
+    "gu_one_off_contribution_date",
+    "gu.contributions.recurring.contrib-timestamp.Monthly",
+    "gu.contributions.recurring.contrib-timestamp.Annual",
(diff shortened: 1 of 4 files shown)


support-frontend/app/controllers/DiagnosticsController.scala
```diff
@@ -0,0 +1,39 @@
...
+    "gu_action_required_for",
+    "gu_digital_subscriber",
+    "gu_hide_support_messaging",
+    "gu_recurring_contributor",
+    "gu_one_off_contribution_date",
+    "gu.contributions.recurring.contrib-timestamp.Monthly",
+    "gu.contributions.recurring.contrib-timestamp.Annual",
+    "gu.contributions.contrib-timestamp",
+    "gu_article_count_opt_out",
+    "gu_contributions_reminder_signed_up",
```
johnduffell, line 24: this list was lifted directly from 
https://github.com/guardian/frontend/blob/126dfcbc1aa961650b7f7ff41ee50a12782bb62e/static/src/javascripts/projects/common/modules/commercial/user-features.js#L9-L37

tomrf1 (approved):
makes sense!

rupertbates (approved):
👍

PR #2912, opened 2021-01-11
fix IT tests after bigquery work

johnduffell:

This fixes two IT test failues introduced after #2908 was merged.
Also improved the error message when DiscriminatedType can't find a decoder.
Also make sure the Failure Handler IT test is automated.

https://trello.com/c/Q5PC0AKM/3569-fix-it-tests

## Why are you doing this?

The IT tests need to be passing so that the noise level is low.  We need them to not be noisy as then we will trust them more.  They are our only source of confirmation that our intragrations with external systems are still in good health, other than what we notice when testing locally or via alarms.


support-workers/src/test/scala/com/gu/support/workers/JsonFixtures.scala
```diff
@@ -575,7 +575,7 @@ object JsonFixtures {
...
+      "state": {"requestId":"299f0204-8f82-f479-0000-00000000e33d","user":{"id":"30001643","primaryEmailAddress":"test@gu.com","firstName":"FygxilMxB5QMoAkmuIc","lastName":"FygxilMxB5QMoAkmuIc","billingAddress":{"country":"GB"},"country":"GB","state":null,"allowMembershipMail":false,"allowThirdPartyMail":false,"allowGURelatedMail":false,"isTestUser":false},"product":{"productType":"DigitalPack","currency":"GBP","billingPeriod":"Annual","readerType":"Direct"},"analyticsInfo":{"paymentProvider": "Stripe","isGiftPurchase":false},"paymentFields":{"userId":"30001643","stripeToken":"tok_chargeDeclined"},"acquisitionData":{"ophanIds":{"pageviewId":"jaignx8mlwazhhpaq9tk","visitId":null,"browserId":null},"referrerAcquisitionData":{"campaignCode":null,"referrerPageviewId":null,"referrerUrl":null,"componentId":null,"componentType":null,"source":null,"abTest":null},"supportAbTests":[{"name":"usRecurringCopyTest","variant":"notintest"},{"name":"ukRecurringAmountsTest","variant":"lower"},{"name":"usRecurringAmountsTest","variant":"notintest"}]}},
```
rupertbates, line 578: Is there an actual change here? It's hard to see

rupertbates (approved):
Thanks for fixing this

support-workers/src/test/scala/com/gu/acquisitions/BigQuerySpec.scala
```diff
@@ -77,7 +77,7 @@ class BigQuerySpec extends AsyncFlatSpec with Matchers with LazyLogging {
       None
     )
 
-    service.tableInsertRow(dataRow) shouldBe Right(())
+    service.tableInsertRow(dataRow).value.map(_ shouldBe Right(()))
```
johnduffell, line 80: sorry I should have pointed out, this test was failing because Right(()) was not the same as Future(Not completed yet)

support-workers/src/test/scala/com/gu/support/workers/JsonFixtures.scala
johnduffell: the change here is 
`{"productType":"DigitalSubscription",`
changed to 
`{"productType":"DigitalPack",`

I'm not sure how it worked before as the serialisers have always used "DigitalPack" (although we could change it one day)

johnduffell: in fact it looks like you changed it specifically in the below PR can you remember if it was intentional?
https://github.com/guardian/support-frontend/pull/2891/files#diff-7588de636d51f82be09e89499144f270ac9f4960265e47112b3fdeaa96d498dcR578

johnduffell: looks like it's only used in that one test so I think it should be safe to merge it

support-workers/src/test/scala/com/gu/support/workers/lambdas/FailureHandlerSpec.scala
```diff
@@ -22,7 +22,7 @@ import org.scalatest.matchers.should.Matchers
 import scala.concurrent.{ExecutionContext, Future}
 
 @IntegrationTest
-class FailureHandlerIT extends AsyncLambdaSpec with MockContext {
+class FailureHandlerITSpec extends AsyncLambdaSpec with MockContext {
```
johnduffell, line 25: the IT test autoimated runner only finds tests that end in "Spec" which is usually but not always the case

PR #2914, opened 2021-01-12
throw an exception if the IT test metrics don't succeed

johnduffell:

This will make the put metric request throw an exception if the metric isn't sent.

## Why are you doing this?

There seems to be an issue where metrics for the IT tests are not always getting sent.  This means we get alarms going off, but reading the logs, the tests all ran as they should.

I noticed in the code, if the put metric call throws an exception, this will be ignored.


michaelbjacobson (approved):
Looks good to me! 👍🏼

PR #2915, opened 2021-01-12
add a short delay before trying to pay with DD

johnduffell:

This adds a short sleep to DD checkouts to make sure the button really is available.

https://trello.com/c/GHM1Ib1W/3570-fix-selenium-tests

## Why are you doing this?

WE  are not sure why this is failing in the automated run but not locally.  It is thinking it clicked the button but it doesn't actually seem to click it.  Perhaps there's a race condition there.


i-hardy (approved):
😴

PR #2916, opened 2021-01-12
Use CC checkout instead of DD for DS gift post deploy test

johnduffell:

Change DS gift checkout post deploy test to use CC instead of DD

https://trello.com/c/GHM1Ib1W/3570-fix-selenium-tests

## Why are you doing this?

 there seems to be a weird bug that only affects gift+DD combination!
When it clicks on the confirm DD button, the Pay button is there, but clicking it doesn't seem to have much effect (at least the first click).

We tried a 1s delay in this PR: https://github.com/guardian/support-frontend/pull/2915 but this is a backup PR in case that doesn't work enough.  We prefer to add a delay and still have the testing, rather than not testing the DD at all with DS.


i-hardy (approved):
It sleeps like the sleep hasn't helped, so let's see if this one works instead!

PR #2917, opened 2021-01-13
Get amounts tests from support-admin-console

TomPretty:
Allows for amounts tests on the contributions LP to be configured from support-admin-console, as per [this card](https://trello.com/c/5KqrDNqc/2340-implement-amounts-testing-on-the-rrcp). The matching support-admin-console PR is [here](https://github.com/guardian/support-admin-console/pull/190).

I think there are quite a few unused functions in the `contributions.js` helper file. I'll raise a future PR to clean those up. Also, I think we could do some future work to simplify the logic around how we handle the 'other' amount field. Too many places in the app need to know that either there is a `selectedAmount` or/and `otherAmount`. I think only the part of the page rendering the form should care if the user has selected one of the options or has entered in a custom amount in the field. Everywhere else just cares about the amount the user is about to contribute. 

## TODO

- [ ] create code s3 json
- [ ] create PROD s3 json (including hardcoded tests?)


support-frontend/assets/helpers/__tests__/formValidation.js
```diff
@@ -129,21 +129,9 @@ describe('formValidation', () => {
...
-        isDefault: true,
-      },
-      ANNUAL: {
-        value: '100',
-        spoken: 'one hundred',
-        isDefault: true,
-      },
+      ONE_OFF: 50,
+      MONTHLY: 15,
+      ANNUAL: 100,
```
TomPretty, line 134: It might be because we have a weird setup for our tests, but flow wasn't catching this type error, which isn't great

support-frontend/assets/helpers/abTests/abtest.js
```diff
@@ -334,6 +329,23 @@ function getParticipations(
   return participations;
 }
 
+function getAmountsTestParticipations(countryGroupId: CountryGroupId, settings: Settings): ?Participations {
+  if (!settings.amounts) {
+    return null;
+  }
```
TomPretty, line 335: I think this is only ever `null` for server side rendering

support-frontend/assets/helpers/abTests/abtest.js
```diff
@@ -334,6 +329,23 @@ function getParticipations(
...
+  const { test } = settings.amounts[countryGroupId];
+
+  if (!test || !test.isLive) {
+    return null;
+  }
+
+  const variants = ['CONTROL', ...test.variants.map(variant => variant.name)];
```
TomPretty, line 342: I'll add a check to support-admin-console to prevent a user naming a variant 'CONTROL'

support-frontend/assets/helpers/abTests/helpers.js
```diff
@@ -1,50 +1,133 @@
...
 
-// Returns a new AmountsRegions by combining currentAmountsRegions with any test participation amounts
-export function overrideAmountsForParticipations(
-  abParticipations: Participations,
-  currentAmountsRegions: AmountsRegions,
-): AmountsRegions {
-
-  return Object.keys(abParticipations).reduce((amountsRegions, testName) => {
-    const test = allTests[testName];
+const FALLBACK_AMOUNTS: FallbackAmounts = {
```
TomPretty, line 12: I copied these over from the tool today

support-frontend/assets/helpers/abTests/helpers.js
```diff
@@ -1,50 +1,133 @@
...
-      return overrideAmountsForTest(variant, test, amountsRegions);
-    }
+export function getAmounts(
+  settings: Settings,
+  abParticipations: Participations,
+  countryGroupId: CountryGroupId,
+): ContributionAmounts {
+  if (!settings.amounts) {
+    return FALLBACK_AMOUNTS[countryGroupId];
+  }
```
TomPretty, line 120: Again, I think this will only happen during ssr, so we don't really need the fallbacks, but it feels a little safer

support-frontend/assets/helpers/contributions.js
```diff
@@ -390,7 +408,7 @@ function getCustomAmountA11yHint(
   }
 
   return `Enter an amount of ${config[countryGroupId][contributionType].minInWords}
-    ${spokenCurrency} or more for your 
+    ${spokenCurrency} or more for your
```
TomPretty, line 411: This function isn't actually used. There are quite a few in this file that aren't. I'll create a follow up PR to remove them all.

support-frontend/assets/helpers/contributions.js
```diff
@@ -433,15 +451,15 @@ const contributionTypeRadios = [
...
 
   return amounts.map(amount => ({
-    value: amount.value,
-    text: `${currencies[currencyId].glyph}${amount.value}`,
-    accessibilityHint: getAmountA11yHint(contributionType, currencyId, numbersInWords[amount.value]),
+    value: amount.toString(),
+    text: `${currencies[currencyId].glyph}${amount}`,
+    accessibilityHint: getAmountA11yHint(contributionType, currencyId, numbersInWords[amount]),
   }));
 }
```
TomPretty, line 464: Not actually used

support-frontend/assets/helpers/abTests/helpers.js
tomrf1: 👍

support-frontend/assets/helpers/abTests/helpers.js
```diff
@@ -1,50 +1,133 @@
...
 
-import { tests as allTests } from 'helpers/abTests/abtestDefinitions';
-import { type AmountsRegions } from 'helpers/contributions';
-import type { Test, Participations } from 'helpers/abTests/abtest';
+import type { Participations } from 'helpers/abTests/abtest';
+import type { Settings } from 'helpers/settings';
+import type { CountryGroupId } from 'helpers/internationalisation/countryGroup';
+import type { ContributionAmounts } from 'helpers/contributions';
 
-function overrideAmountsForTest(
```
tomrf1, line 7: 🎉

tomrf1:
> I think only the part of the page rendering the form should care if the user has selected one of the options or has entered in a custom amount in the field. Everywhere else just cares about the amount the user is about to contribute.

Totally agree - this is worth doing, and also review whether the validation of the 'other' amount is safe enough

support-frontend/assets/helpers/abTests/abtest.js
```diff
@@ -344,10 +352,14 @@ const init = (
...
   const urlParticipations: ?Participations = getParticipationsFromUrl();
   const serverSideParticipations: ?Participations = getServerSideParticipations();
+  const amountsTestParticipations: ?Participations = getAmountsTestParticipations(countryGroupId, settings);
+  const localStorageParticipations: Participations = getLocalStorageParticipations();
   const combinedParticipations: Participations = {
     ...participations,
     ...urlParticipations,
     ...serverSideParticipations,
+    ...amountsTestParticipations,
+    ...localStorageParticipations,
```
TomPretty, line 362: @tomrf1 I've tried this instead. Before, the `localStorageParticipations` were only overwritting the participations coming from `abTestDefinitions.js`. Now by putting it here, you can overwrite any participations

TomPretty: although maybe url participations should take come last? (highest priority)

tomrf1: Yes you'd want the url to override. How's this?
```
    ...participations,
    ...amountsTestParticipations,
    ...serverSideParticipations,
    ...localStorageParticipations,
    ...urlParticipations,
```

TomPretty: just pushed that reorder up!

tomrf1: thanks!

tomrf1 (approved):
great work!

PR #2919, opened 2021-01-15
payment-api error handling improvements

tomrf1:
#### Some terms used in payment-api:
"failed payment" - caused by an issue with a customer's card/account/details. Has a cloudwatch metric, `failed-payment`.
"payment error" - any other cause, i.e. software/infrastructure/integration problems. . Has a cloudwatch metric, `payment-error`.

#### The problem:
1. We currently send `ERROR` logs for failed payments, e.g. card declined payment failures. These are not errors.
2. We do not have alarms for some genuine errors, e.g. failure to complete post-contribution actions like analytics.

#### The solution:
The aim of this PR is to ensure all ERROR logs are software/infrastucture/integration errors that require immediate investigation. Additionally, these errors should always trigger an alarm.

#### Changes:
1. Lower `payment-error` alarms threshold to 1. These should indicate something is broken and needs investigating.
2. Stop using `logger.error()` in random places, and instead only use it inside `CloudWatchService`. This is the class which writes the cloudwatch metrics, and so it is the place that decides if something is a `payment-error` or just a `failed-payment`. So it is now responsible for deciding what is an `ERROR` log whenever a payment goes wrong.
3. For AmazonPay, handle `InvalidPaymentMethod` as a `failed-payment` (so no alarm).
4. Fix handling of Stripe failure codes. Some normal payment failures were being classed as 'unknown'.
5. Actually use the `post-payment-tasks-error` metric, via `CloudWatchService.recordPostPaymentTasksError()`. There was already an alarm on this metric, so this would have caught the ophan issue if it was actually being used!

#### For consideration:
Add an alarm for any time an `ERROR` log appears in cloudwatch logs?
If we stick to using `CloudWatchService` properly anytime something goes wrong then this shouldn't be necessary, as the existing cloudwatch metrics will trigger alarms.
Also, it might result in 'death by alarms' if it goes off at the same time as an existing alarm.

### DIFF
--- a/support-payment-api/src/main/resources/cloud-formation.yaml
+++ b/support-payment-api/src/main/resources/cloud-formation.yaml
@@ -507,43 +507,62 @@ Resources:
       Statistic: Sum
       TreatMissingData: breaching
 
-  ThreeOrMorePaypalPaymentErrorsIn15Mins:
+  PaypalPaymentError:
       Type: AWS::CloudWatch::Alarm
       Condition: CreateProdResources
       Properties:
         AlarmActions:
         - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:contributions-dev
-        AlarmName: !Sub ${App} ${Stage} Multiple paypal errors in the last fifteen minutes for one-off contributions via the payment-api
+        AlarmName: !Sub ${App} ${Stage} Paypal payment error in the last fifteen minutes for one-off contributions via the payment-api
         MetricName: payment-error
         Namespace: support-payment-api-PROD
         Dimensions:
                 - Name: payment-provider
                   Value: Paypal
         ComparisonOperator: GreaterThanThreshold
-        Threshold: 3
+        Threshold: 1
         Period: 900
         EvaluationPeriods: 1
         Statistic: Sum
         TreatMissingData: notBreaching
 
-  ThreeOrMoreStripePaymentErrorsIn15Mins:
+  StripePaymentError:
       Type: AWS::CloudWatch::Alarm
       Condition: CreateProdResources
       Properties:
         AlarmActions:
         - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:contributions-dev
-        AlarmName: !Sub ${App} ${Stage} Multiple stripe errors in the last fifteen minutes for one-off contributions via the payment-api
+        AlarmName: !Sub ${App} ${Stage} Stripe payment error in the last fifteen minutes for one-off contributions via the payment-api
         MetricName: payment-error
         Namespace: support-payment-api-PROD
         Dimensions:
                 - Name: payment-provider
                   Value: Stripe
         ComparisonOperator: GreaterThanThreshold
-        Threshold: 3
+        Threshold: 1
         Period: 900
         EvaluationPeriods: 1
         Statistic: Sum
         TreatMissingData: notBreaching
+
+  AmazonPayPaymentError:
+    Type: AWS::CloudWatch::Alarm
+    Condition: CreateProdResources
+    Properties:
+      AlarmActions:
+        - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:contributions-dev
+      AlarmName: !Sub ${App} ${Stage} Amazon Pay payment error in the last fifteen minutes for one-off contributions via the payment-api
+      MetricName: payment-error
+      Namespace: support-payment-api-PROD
+      Dimensions:
+        - Name: payment-provider
+          Value: AmazonPay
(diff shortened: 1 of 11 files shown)


support-payment-api/src/main/scala/backend/SubscribeWithGoogleBackend.scala
```diff
@@ -53,7 +53,7 @@ case class SubscribeWithGoogleBackend(databaseService: ContributionsStoreService
 
   private def handleAlreadyProcessedPayment(googleRecordPayment: GoogleRecordPayment): EitherT[Future, BackendError, Unit] = {
     logger.error(s"Received a duplicate payment id for payment : $googleRecordPayment")
-    cloudWatchService.recordPostPaymentTasksError(PaymentProvider.SubscribeWithGoogle)
+    cloudWatchService.recordPostPaymentTasksError(PaymentProvider.SubscribeWithGoogle,"")
```
tomrf1, line 56: we still need to delete all SWG code!

support-payment-api/src/main/scala/services/CloudWatchService.scala
```diff
@@ -36,10 +37,13 @@ class CloudWatchService(cloudWatchAsyncClient: AmazonCloudWatchAsync, environmen
...
+      logger.error(s"Payment error with $paymentProvider", error)
       put("payment-error", paymentProvider)
-    else
+    } else {
+      logger.info(s"Payment failure with $paymentProvider", error)
       put("failed-payment", paymentProvider)
+    }
   }
 
   def isPaymentError(error: Exception): Boolean = {
```
tomrf1, line 49: Rather than matching on strings like this perhaps a better solution would be to encode `payment-error` vs `failed-payment` in the type system. It would be quite a big refactor though

support-payment-api/src/main/scala/backend/AmazonPayBackend.scala
```diff
@@ -65,7 +65,7 @@ class AmazonPayBackend(
   private def handleResponse(response: Either[AmazonPayApiError, AuthorizationDetails] , amazonPayRequest: AmazonPayRequest, clientBrowserInfo: ClientBrowserInfo)  = {
     response.toEitherT[Future]
       .leftMap { error =>
-        logger.error(s"Something went wrong with ${amazonPayRequest.paymentData.orderReferenceId}: ${error.getMessage}")
+        logger.info(s"Something went wrong with AmazonPay orderReferenceId: ${amazonPayRequest.paymentData.orderReferenceId}")
         cloudWatchService.recordFailedPayment(error, PaymentProvider.AmazonPay)
```
rupertbates, line 69: Do we lose anything here by not having the `orderReferenceId` in the error?

support-payment-api/src/main/scala/backend/StripeBackend.scala
```diff
@@ -54,7 +54,7 @@ class StripeBackend(
   def createCharge(chargeData: LegacyStripeChargeRequest, clientBrowserInfo: ClientBrowserInfo): EitherT[Future, StripeApiError, StripeCreateChargeResponse] =
     stripeService.createCharge(chargeData)
       .leftMap(err => {
-        logger.error(s"unable to create Stripe charge ($chargeData)", err)
+        logger.info(s"unable to create Stripe charge ($chargeData)")
```
rupertbates, line 57: Same question as above

rupertbates (approved):
This looks like a big improvement, well done 👍 . I've just asked a couple of questions about potentially losing some context which could be useful for investigating errors

support-payment-api/src/main/scala/backend/AmazonPayBackend.scala
tomrf1: we will still have it in the preceding `info` log, but we're leaving it to `recordFailedPayment` to decide whether it should be logged as an `error`. I think it should be easy enough for anyone investigating an alarm to find this

support-payment-api/src/main/resources/cloud-formation.yaml
```diff
@@ -507,51 +507,70 @@ Resources:
...
+        AlarmName: !Sub ${App} ${Stage} Paypal payment error in the last fifteen minutes for one-off contributions via the payment-api
         MetricName: payment-error
         Namespace: support-payment-api-PROD
         Dimensions:
                 - Name: payment-provider
                   Value: Paypal
         ComparisonOperator: GreaterThanThreshold
-        Threshold: 3
+        Threshold: 1
         Period: 900
```
johnduffell, line 524: period could change to 60 here so we find out quciker

tomrf1: 👍

PR #2920, opened 2021-01-15
Fix validation issues with the date picker in timezones outside of the UK

i-hardy:

This fixes the issue where the date picker in the digital subs gift checkout is failing validation when left on today's date if the user is in a location where midnight UTC is the previous day in their timezone.

It removes all validation regarding whether the date is 'too early', and amends the way we confirm the date to the user so it will always show the correct date relative to their timezone.

[**Trello Card**](https://trello.com/c/TCZHMVV0)

## Why are you doing this?

This has been causing the Selenium tests for the gift checkout to fail as the CI server is located in the US, and may have been preventing gift checkouts for users outside of the UK.

The backend doesn't care if a date that the user selects is in the past- the gift will always be sent immediately if the delivery date is not in the future- so there's no practical requirement for client-side validation outside of making sure the user sees something sensible if they click 'Check date'

### DIFF
--- a/support-frontend/assets/helpers/formValidation.js
+++ b/support-frontend/assets/helpers/formValidation.js
@@ -19,8 +19,6 @@ export const isEmpty: (string | null) => boolean = input =>
 
 export const isNotEmpty: (string | null) => boolean = input => !isEmpty(input);
 
-export const isNotInThePast: (Date | null) => boolean = date => !DateUtils.isPastDay(date);
-
 export const isNotTooFarInTheFuture: (Date | null) => boolean = (date) => {
   const rangeDate = new Date();
   rangeDate.setDate(rangeDate.getDate() + daysFromNowForGift);
--- a/support-frontend/assets/helpers/formValidation.js
+++ b/support-frontend/assets/helpers/formValidation.js
@@ -43,7 +43,7 @@ export const checkOptionalEmail: (string | null) => boolean = input => isEmpty(i
 
 export const checkGiftStartDate: (string | null) => boolean = (rawDate) => {
   const date = rawDate ? new Date(rawDate) : null;
-  return isNotEmpty(rawDate) && isNotInThePast(date) && isNotTooFarInTheFuture(date);
+  return isNotEmpty(rawDate) && isNotTooFarInTheFuture(date);
 };
 
 export const checkAmount: (string, CountryGroupId, ContributionType) =>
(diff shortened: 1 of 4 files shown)


support-frontend/assets/helpers/formValidation.js
```diff
@@ -43,7 +43,7 @@ export const checkOptionalEmail: (string | null) => boolean = input => isEmpty(i
 
 export const checkGiftStartDate: (string | null) => boolean = (rawDate) => {
   const date = rawDate ? new Date(rawDate) : null;
-  return isNotEmpty(rawDate) && isNotInThePast(date) && isNotTooFarInTheFuture(date);
+  return isNotEmpty(rawDate) && isNotTooFarInTheFuture(date);
```
GHaberis, line 46: Can we remove the entire `isNotInThePast` function? Is this used anywhere else in the codebase?

GHaberis: It's still possible for a user to select a date in the past via the calendar component? Although unlikely. Do we know happens downstream if a user were to do this and submit the form?

johnduffell: so it would send the email immediately, and it would write the provided date in to zuora/SF/data lake, which could be confusing, but is unlikely to be looked at.
It would be possible to cap it in the back end to today, if we thought it was worth it.

johnduffell (approved):
thanks for fixing! 👏 let's get those tests passing

support-frontend/assets/helpers/formValidation.js
GHaberis: Thanks @johnduffell, as long as the email is sent I think we're good here 👍 I agree the chances of someone cross-checking the provided date in to zuora/SF/data lake in this edge case scenario are low!

GHaberis (approved):
Nice work @i-hardy! 👍

support-frontend/assets/helpers/formValidation.js
i-hardy: Yeah, we can actually delete a couple of functions, I've just taken this and the `dateIsPast` function out.

With regards to the user selecting a date in the past, it's still not possible to do via the date picker. You can manually enter a past date in the input fields if you want, but since this won't cause an issue for the backend I don't see the point in trying to prevent it client-side. However if you enter a date in the past and hit the 'Check date' button, it will always tell you that the gift will be delivered on today's date- it didn't seem sensible to confirm to the user that yes, we would deliver their gift in 1992.

GHaberis: That's great, thanks for explaining @i-hardy. I didn't know the date picker prevented historic date selection so the risk is even smaller than I'd imagined.  

> However if you enter a date in the past and hit the 'Check date' button, it will always tell you that the gift will be delivered on today's date

It sounds like from @johnduffell's comment that this is in fact the case so I don't see a problem with us telling the user that in the rare event they enter a historic date.

PR #2923, opened 2021-01-18
Storybook Konmari Part 2 - Tidy up and refresh of documented components

i-hardy:

This clears out old and unused components from Storybook and tidies up existing Storybook components so they're better presented. It also adds some existing components that were not documented in Storybook- our two image components, the country switcher, the checkout expander and the date picker.

[**Trello Card**](https://trello.com/c/mZwJAupm)

## Why are you doing this?

Making our Storybook tidier and more relevant to the codebase as it currently is makes it more useful as a tool for documenting/exploring our components and building new ones.


GHaberis (approved):
Nice work Imogen, looking forward to seeing the results of this!

PR #2924, opened 2021-01-18
Fix error summary

lucymonie:
This PR is to fix an issue with the error summary in the checkout. The problem was that the error summary was showing up with empty bullet points when there were no errors, rather than just not showing at all in this case. In addition, I noticed that the stripe card errors weren't appearing in the summary, so I have made sure they are getting passed in correctly.

[**Trello Card**](https://trello.com/c/AJ4rw7GK/3525-fix-issue-in-checkout-error-summary-where-bullet-points-still-show-after-errors-have-been-removed)

## Why are you doing this?
This arose from changes to the checkouts when implementing the design system in this PR: https://github.com/guardian/support-frontend/pull/2786


support-frontend/assets/components/subscriptionCheckouts/stripeForm/stripeForm.jsx
```diff
@@ -269,6 +269,7 @@ const StripeForm = (props: StripeFormPropTypes) => {
     });
 
   const requestSCAPaymentMethod = (event) => {
+    console.log('called request method');
```
i-hardy, line 272: Stray console.log!

i-hardy (approved):
That stray log needs removing, otherwise this looks good!

support-frontend/assets/components/subscriptionCheckouts/stripeForm/stripeForm.jsx
lucymonie: Whoops! Thanks 🦑

lucymonie:
Thanks @i-hardy! 🎉

PR #2925, opened 2021-01-19
Fix issues with price cards incorrectly displaying content from promo codes

i-hardy:

This fixes two issues relating to the price cards when promo codes were used:

- Promo codes applying a 100% discount will now show the £0 price correctly, instead of defaulting back to the base price
- Where more than one promo code is applied for a specific product, the Guardian Weekly landing page will no longer always show the first one regardless of the query string.

[**Trello Card**](https://trello.com/c/TpBHm3KU)


GHaberis (commented):
Looks good 👍

rupertbates (approved):
Nice 👍

PR #2926, opened 2021-01-21
add metric for server side validation failures

johnduffell:

Adding a metric for any kind of server side validation failure

## Why are you doing this?

At the moment we return a 4xx error for validation failures.  This is not monitored anywhere except a line in the logs.
```
2021-01-21T17:31:15,640+0000 [application-akka.actor.default-dispatcher-4] ERROR com.gu.monitoring.SafeLogger$ - [ff50d14a-2928-92ec-0000-000000000001] Failed to create new Monthly Paper-HomeDelivery-Everyday subscription, due to RequestValidationError(validation of the request body failed)
```

I am about to improve the validation, so it made sense to have a metric, so that we can easily compare before/after error rates.  It would be nice to have an alarm on this, which we can add once we know the baseline.

## Screenshots
https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#metricsV2:graph=~(metrics~(~(~'support-frontend~'ServerSideValidationFailure~'Product~'Paper~'Stage~'DEV))~view~'timeSeries~stacked~false~region~'eu-west-1~start~'-PT1H~end~'P0D~stat~'Sum~period~60);query=~'*7bsupport-frontend*2cProduct*2cStage*7d
image

### DIFF
--- a/support-frontend/app/controllers/CreateSubscription.scala
+++ b/support-frontend/app/controllers/CreateSubscription.scala
@@ -7,10 +7,12 @@ import akka.stream.scaladsl.Flow
 import akka.util.ByteString
 import cats.data.EitherT
 import cats.implicits._
+import com.gu.aws.{AwsCloudWatchMetricPut, AwsCloudWatchMetricSetup}
 import com.gu.identity.model.{User => IdUser}
 import com.gu.monitoring.SafeLogger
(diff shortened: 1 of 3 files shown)


tomrf1 (approved):
👍

PR #2927, opened 2021-01-21
validate HD postcodes on the server side

johnduffell:

This PR adds validation on the server side for Home Delivery being inside the M25.
It also adds validation to guardian weekly that seems to have been missed.

Tested on dev with contribution, HD in/out of the M25, digital sub.
Will be monitored in PROD for the rate of validation errors remaining stable.

## Why are you doing this?

Some Home Delivery are getting through outside the M25.  This should be prevented by the client side, but it's not watertight, and the right combination of exceptions and paypal buttons can allow a POST to the server side with a postcode not in area.
This causes us to take money for something we can't fulfil, which is not right.
We could fix up the client side also, but the server side should definitely keep out invalid data where possible as a last resort.

## Related PR
I will check success by looking at the rate of SS validation errors, to make sure nothing is broken
https://github.com/guardian/support-frontend/pull/2926


support-frontend/app/utils/CheckoutValidationRules.scala
```diff
@@ -123,29 +124,175 @@ object DigitalPackValidation {
...
+    formattedPostcode.slice(0, formattedPostcode.length - 3)
 
   }
+
+  def postcodeIsWithinDeliveryArea(postcode: String): Boolean =
+    M25_POSTCODE_PREFIXES.contains(getPrefix(postcode))
+
+  // NOTE also in deliveryCheck.js
+
+  val M25_POSTCODE_OLD_PREFIXES = List(
```
rupertbates, line 198: Is this list duplicated in the client side now? It would be good to pass it through from here to the client so we don't have to maintain two lists.

rupertbates (dismissed):
Really good work, I've just made one suggestion about sharing the list of postcodes between client and server

support-frontend/assets/helpers/deliveryCheck.js
```diff
@@ -1,57 +1,8 @@
...
-  'WD172', 'WD173', 'WD174', 'WD180', 'WD187', 'WD188', 'WD194', 'WD195', 'WD196', 'WD197', 'WD231', 'WD232', 'WD233', 'WD234', 'WD244', 'WD245', 'WD246', 'WD247', 'WD250', 'WD257', 'WD258', 'WD259', 'WD31', 'WD33', 'WD34', 'WD35', 'WD36', 'WD37', 'WD38', 'WD48', 'WD50', 'WD61', 'WD62', 'WD63', 'WD64', 'WD65', 'WD77', 'WD78', 'WD79'];
-
-const M25_POSTCODE_PREFIXES = M25_POSTCODE_OLD_PREFIXES.concat(M25_NEW_PREFIXES);
+const M25_POSTCODE_PREFIXES = getGlobal('homeDeliveryPostcodes') || [];
```
johnduffell, line 5: wasn't sure what to do if the value does not exist as the type checker will not notice any more - as it is it will block all purchases!
The post deploy tests seem to check subs card rather than home delivery at the moment. https://github.com/guardian/support-frontend/blob/2df3358d1e443d21c2c6b82db256631937268c0e/support-frontend/test/selenium/subscriptions/CheckoutsSpec.scala#L48-L52

support-frontend/app/utils/CheckoutValidationRules.scala
johnduffell: I have done that now

support-frontend/assets/helpers/deliveryCheck.js
rupertbates: I think this is low risk because it's rendered directly into the page. We could add another test for HD or change the existing one if you're worried?

rupertbates (approved):
👍 LGTM

support-frontend/assets/helpers/deliveryCheck.js
johnduffell: There's no test to ensure that it really is rendered into the page.  But I agree that part is unlikely to be broken inadvertently.  I will leave it as it is.

PR #2930, opened 2021-01-25
Fix issues with form validation causing issues with paying via Stripe

i-hardy:

This fixes a current bug where if a user who has selected to pay by debit/credit card hits the 'Pay' button without filling out some/all of the card fields, we still attempt to get the payment token from Stripe and submit the form.

It also fixes a bug where if there is a validation issue elsewhere on the form that prevents submission we get an error from Stripe on the second submission attempt- this PR stops the component trying to submit any details to Stripe until all validation checks have passed.

It also adds `@testing-library/react` and associated dependencies to the codebase, in order to facilitate unit testing of React components.

[**Trello Card**](https://trello.com/c/ebS5fdAt)

## Why are you doing this?

The bug could cause the checkout to get into a really confusing validation state and cause issues for users trying to pay.

I've introduced the unit testing partly based on a conversation in the last Subscriptions retro; we have some unit testing of client-side code, but there's a lot of quite complex logic that's not adequately tested, often inside the components themselves (as in this case). Being able to test the components in isolation will hopefully help both in tackling future bugs and in documenting component logic and expected behaviour.

### DIFF
--- a/support-frontend/assets/components/subscriptionCheckouts/stripeForm/stripeForm.jsx
+++ b/support-frontend/assets/components/subscriptionCheckouts/stripeForm/stripeForm.jsx
@@ -114,6 +114,7 @@ const StripeForm = (props: StripeFormPropTypes) => {
     },
   });
   const [disableButton, setDisableButton] = useState<boolean>(false);
+  const [submitReady, setSubmitReady] = useState<boolean>(false);
 
   const stripe = stripeJs.useStripe();
   const elements = stripeJs.useElements();
@@ -121,13 +122,6 @@ const StripeForm = (props: StripeFormPropTypes) => {
   /**
    * Handlers
    */
-
-  const getAllCardErrors = () => ['cardNumber', 'cardExpiry', 'cardCvc'].reduce((acc, field) => {
-    if (cardFieldsData[field].error.length > 0) {
-      acc.push({ field: [field], message: cardFieldsData[field].error });
-    }
-    return acc;
-  }, []);
 
   const handleStripeError = (errorData: any): void => {
     setPaymentWaiting(false);
@@ -201,27 +195,30 @@ const StripeForm = (props: StripeFormPropTypes) => {
   };
 
   const handleCardErrors = () => {
-    // eslint-disable-next-line array-callback-return
-    ['cardNumber', 'cardExpiry', 'cardCvc'].map((field) => {
-      if (cardFieldsData[field].empty === true) {
-        setCardFieldsData(prevData => ({
-          ...prevData,
-          [field]: {
-            ...prevData[field],
-            error: cardFieldsData[field].errorEmpty,
-          },
-        }));
-      } else if (!cardFieldsData[field].complete) {
-        setCardFieldsData(prevData => ({
-          ...prevData,
-          [field]: {
-            ...prevData[field],
-            error: prevData[field].errorIncomplete,
-          },
-        }));
+    const newCardFieldsErrors = [];
+    const newCardFieldsData = Object.entries(cardFieldsData).reduce((newData, [cardFieldName, existingFieldData]) => {
+      if (existingFieldData.empty) {
+        const error = existingFieldData.errorEmpty;
+        // eslint-disable-next-line no-param-reassign
+        newData[cardFieldName] = {
+          ...existingFieldData,
+          error,
+        };
+        newCardFieldsErrors.push({ field: [cardFieldName], message: error });
+      } else if (!existingFieldData.complete) {
+        const error = existingFieldData.errorIncomplete;
+        // eslint-disable-next-line no-param-reassign
+        newData[cardFieldName] = {
+          ...existingFieldData,
+          error,
+        };
+        newCardFieldsErrors.push({ field: [cardFieldName], message: error });
       }
-      setCardErrors(getAllCardErrors());
-    });
+      return newData;
+    }, { ...cardFieldsData });
+
+    setCardFieldsData(newCardFieldsData);
+    setCardErrors(newCardFieldsErrors);
   };
 
   const handleChange = (event) => {
@@ -271,18 +268,7 @@ const StripeForm = (props: StripeFormPropTypes) => {
     props.validateForm();
     handleCardErrors();
     checkRecaptcha();
-
-    if (stripe && props.allErrors.length === 0 && cardErrors.length === 0 && !recaptchaError) {
-      console.log('what up');
-      setDisableButton(true);
-      if (setupIntentClientSecret) {
-        handleCardSetupAndPay();
-      } else if (recaptchaCompleted) {
-        // User has completed the form and recaptcha, but we're still waiting for the setupIntentClientSecret to
-        // come back. A hook will complete subscription once the setupIntentClientSecret is available.
-        setPaymentWaiting(true);
-      }
-    }
+    setSubmitReady(true);
   };
 
   /**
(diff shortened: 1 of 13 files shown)


support-frontend/assets/helpers/abTests/__tests__/abtestTest.js
```diff
@@ -39,6 +39,10 @@ describe('basic behaviour of init', () => {
     window.matchMedia = window.matchMedia || jest.fn(() => ({ matches: false }));
   });
 
+  afterEach(() => {
+    window.localStorage.clear();
+  });
+
```
i-hardy, line 45: Changes to this file are because upgrading to JSDOM 15 broke these tests- the tests were assuming a non-operational localStorage whereas JSDOM 15 has a functional one.

support-frontend/assets/components/subscriptionCheckouts/stripeForm/stripeForm.jsx
```diff
@@ -192,9 +184,9 @@ const StripeForm = (props: StripeFormPropTypes) => {
     });
   };
 
-  const setupRecurringHandlers = (): void => {
+  const setupRecurringHandlers = async () => {
```
GHaberis, line 187: I'm interested to know whether we need to use `async` / `await` here? I see we're calling an async function `fetchPaymentIntent` but we're not waiting on its return result, so I think what will happen here is  `setupRecurringHandlers` will return `undefined` once `fetchPaymentIntent('dummy)` has resolved?

We're also not waiting on `setupRecurringHandlers` to resolve asynchronously when we call it [here](https://github.com/guardian/support-frontend/blob/a469c698620b3c1840b31cc5ddaea45b6825b664/support-frontend/assets/components/subscriptionCheckouts/stripeForm/stripeForm.jsx#L281).

support-frontend/assets/components/subscriptionCheckouts/stripeForm/stripeForm.jsx
```diff
@@ -203,27 +195,31 @@ const StripeForm = (props: StripeFormPropTypes) => {
...
+          const error = existingFieldData.errorIncomplete;
+          // eslint-disable-next-line no-param-reassign
+          newData[cardFieldName] = {
+            ...existingFieldData,
+            error,
+          };
+          newCardFieldsErrors.push({ field: [cardFieldName], message: error });
+        }
+        return newData;
+      }, { ...cardFieldsData });
```
johnduffell, line 219: it looks like this whole thing is replacing the "error" in the field data based on errorEmpty and errorIncomplete result, and also building up `newCardFieldsErrors` a mapping from field name to error.
However it looks like it's editing in place the newData within the reduce, but then are we relying on that side effect as we then do a setCardFieldsData below with the same result.
I might have my scala hat on but I feel like it might be clearer use `map` or similar rather than `reduce` and also build a fresh data structure with the "data" and "errors".  Then that can be `unzip`ped or similar to get the two separate results.

support-frontend/assets/components/subscriptionCheckouts/stripeForm/stripeForm.jsx
```diff
@@ -304,6 +290,24 @@ const StripeForm = (props: StripeFormPropTypes) => {
     }
   }, [setupIntentClientSecret]);
 
+  // Handle the payment once all state updates to the card data, errors list, etc
+  // have completed after hitting the submit button
+  useEffect(() => {
+    if (stripe && submitReady && props.allErrors.length === 0 && cardErrors.length === 0 && !recaptchaError) {
```
johnduffell, line 296: it's not clear what "ready" means here, perhaps `submitReady` could be `formSubmitted` or `submitPressed`?

support-frontend/assets/components/subscriptionCheckouts/stripeForm/stripeForm.jsx
```diff
@@ -203,27 +195,31 @@ const StripeForm = (props: StripeFormPropTypes) => {
...
-      }
-      setCardErrors(getAllCardErrors());
-    });
+    const newCardFieldsErrors = [];
+    const newCardFieldsData = Object.entries(cardFieldsData)
+      .reduce((newData, [cardFieldName, existingFieldData]) => {
+        if (existingFieldData.empty) {
+          const error = existingFieldData.errorEmpty;
+          // eslint-disable-next-line no-param-reassign
+          newData[cardFieldName] = {
```
GHaberis, line 204: Could we do this inside the `if` block to prevent us ignoring `no-param-reassign` rule?

```js
const error = existingFieldData.errorEmpty;

newCardFieldsErrors.push({ field: [cardFieldName], message: error });

return {
 ...newData,
 [cardFieldName]: {
   ...existingFieldData,
   error
 }
};
```

johnduffell (approved):
great work, I did make some comments especially about the reduce function, but thanks so much for fixing that as I am sure this is costing us money and it's certainly frustrating me! 👏

support-frontend/assets/components/subscriptionCheckouts/stripeForm/stripeForm.jsx
GHaberis: >  build a fresh data structure with the "data" and "errors"

I think this is a good suggestion, I think it could be clearer than modifying `newCardFieldsErrors` as a side effect of the reduce.

support-frontend/assets/components/subscriptionCheckouts/stripeForm/stripeForm.jsx
GHaberis: I think because we can set this `submitReady` value to `false` if the form has validation errors, `formSubmitted` or `submitPressed` may be misleading as setting those variable names to `false` implies the user hasn't clicked submit? 

How about `readyToSubmit`?

GHaberis:
@i-hardy it's excellent to see tests in place for a React component. The actions and assertions in `stripeForm.test.jsx` were easy to follow, hopefully this sets a precedence for future component development, great work in plugging in the boilerplate to get this all working. I've left a few comments, but otherwise this is really good work!

support-frontend/assets/components/subscriptionCheckouts/stripeForm/stripeForm.jsx
johnduffell: `submitPressedAndValidationPending`? I have nothing against long names if they're clear.  It's just the word ready in my mind that is unclear. But happy to go with the majority view as scala conventions aren't the same as client side ones.

support-frontend/assets/components/subscriptionCheckouts/stripeForm/stripeForm.jsx
i-hardy: I think this was just something I did while trying to get the tests to play nicely with all the background async stuff the component does; it doesn't seem to be necessary though so I'll revert.

support-frontend/assets/components/subscriptionCheckouts/stripeForm/stripeForm.jsx
i-hardy: I generally try to avoid this kind of thing because it massively increases the time complexity of the reduce- the spread operator performs an internal iteration so you're essentially iterating over your (increasingly large) object again on each iteration of the loop, giving you an O(n<sup>2</sup>) operation for no actual gains in terms of immutability, since you create a brand new object as your starting accumulator value. But in this case the object is so small it's not going to have any real impact, so happy to switch over.

support-frontend/assets/components/subscriptionCheckouts/stripeForm/stripeForm.jsx
i-hardy: I'm all for clearer variable names and I wasn't super happy with `submitReady`- switching it to `readyToSubmitPendingValidation`, which I think is about as clear as it can get.

johnduffell:
followup PR https://github.com/guardian/support-frontend/pull/2933

PR #2921, opened 2021-01-18
Remove redux debug information

rupertbates:
## What are you doing in this PR?
We have discovered that trying to stringify the state when it contains an error can cause the following exception:
```
Uncaught TypeError: Converting circular structure to JSON
    --> starting at object with constructor 'Object'
    --- property '__v' closes the circle
    at JSON.stringify (<anonymous>)
    at formReducer.js:70
    at redux.js:435
    at redux.js:435
    at p (<anonymous>:1:36402)
    at v (<anonymous>:1:36684)
    at <anonymous>:1:40069
    at Object.h [as dispatch] (redux.js:213)
    at e (<anonymous>:1:40553)
    at index.js:11
    at dispatch (redux.js:563)
    at formValidation.js:81
    at Array.forEach (<anonymous>)
    at b (formValidation.js:80)
    at h (formValidation.js:99)
    at paperCheckoutForm.jsx:148
    at Object.dispatch (index.js:8)
    at dispatch (<anonymous>:1:28545)
    at Object.validateForm (redux.js:449)
    at Object.onClick [as click] (stripeForm.jsx:273)
    at HTMLButtonElement.k (preact.module.js:1)
    at HTMLButtonElement.r (helpers.js:70)
```

Steps to reproduce are:
- Go to paper home delivery checkout
- Put in an address outside of the M25
- Submit the form

No error message will be shown but the form will not submit.

This PR removed the redux debugging information until we can investigate this further.

PR #2931, opened 2021-01-26
don't put a react component inside a redux action

johnduffell:

Use a plaain string in the redux Action instead of a react component with a link.

https://trello.com/c/O0LclNV2/3586-add-back-debug-info-for-checkouts

## Why are you doing this?

The original PR stated that the error system could only pass around strings (presumably due to it going into an Action): https://github.com/guardian/support-frontend/pull/2434
then later the covid message was removed and it was inlined by mistake, so the Action had a react component in it.
caused this https://github.com/guardian/support-frontend/pull/2921
and it lso breaks Redux devtools (it hangs!)

the docs suggest not really allowed: https://redux.js.org/tutorials/fundamentals/part-2-concepts-data-flow#state-is-read-only

> Since actions are plain JS objects, they can be logged, serialized, stored, and later replayed for debugging or testing purposes.

## Screenshots

TODO

### DIFF
--- a/support-frontend/assets/components/subscriptionCheckouts/address/addressFieldsStore.jsx
+++ b/support-frontend/assets/components/subscriptionCheckouts/address/addressFieldsStore.jsx
@@ -1,7 +1,6 @@
 // @flow
 
 // ----- Imports ----- //
-import React from 'react';
 import { combineReducers, type Dispatch } from 'redux';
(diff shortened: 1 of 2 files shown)


rupertbates (approved):
Looks good, well done for picking this up 👍

PR #2936, opened 2021-01-28
make a read timeout cause a retry unlimited

johnduffell:

Fix the step functions so they will retry unlimited on an SSL error

https://trello.com/c/RmyRUx0V/3589-fix-stripeerrorsspec-test

## Why are you doing this?

The stripe errors spec was failing intermittently.  This is because we were expecting a Socket timeout on connect, but sometimes the connection was quicker than 1ms (presumably already open).  This meant we had a read timeout instead, which is returned as an SSL exception.

There is a point to be made about whether some SSL exceptions should fail straight away.  Otherwise it might be 24 hours before we find out about an issue with the hostname (e.g. salesforce migrates to a new URL?)  We could check the string of the exception as well perhaps?

### DIFF
--- a/support-workers/src/main/scala/com/gu/support/workers/exceptions/ErrorHandler.scala
+++ b/support-workers/src/main/scala/com/gu/support/workers/exceptions/ErrorHandler.scala
@@ -1,7 +1,5 @@
 package com.gu.support.workers.exceptions
 
-import java.net.{SocketException, SocketTimeoutException}
-
 import com.gu.acquisition.model.errors.AnalyticsServiceError
 import com.gu.monitoring.SafeLogger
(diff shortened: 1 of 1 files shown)


rupertbates (approved):
👍 How did you notice this?

johnduffell:
> How did you notice this?

because the build was randomly failing, I recently subscribed to teamcity "main" build failures (instead of "master") and have seen a few, also on a branch sometimes.
I confirmed it by setting the connect timeout to 10000ms and the read timeout to 1ms and it failed with the same error every time.
There's a small chance this isn't actually the issue, but at least it will be better than before.
This is just the last month or so of "main" builds 
image
and if we include every build in the month, most of these should prbably have been passing (the higher points are the failures)
image

PR #2938, opened 2021-01-29
Track country dropdown clicks/interactions

GHaberis:
This PR fixes a couple small issues with the tracking on our country dropdown list. 

1. If `trackProduct` was undefined in `onCountryGroupSelect` we'd never track the interactions, this meant we would't track interactions from our `/support` and `/subscribe` routes, however we're now able to remove the `if { ... }` block around the `sendTrackingEventsOnClick` call as this takes an optional `product` argument.

2. Specify the correct value for `trackProduct` from the `/subscribe/digital` route: 'DigitalPack'

3. Remove the incorrect value for `trackProduct` from the `/subscribe` route: 'GuardianWeekly'

[**Trello Card**](https://trello.com/c/AZDrxIgG/3566-fix-ophan-click-tracking-on-the-country-dropdown-toggle)


i-hardy (approved):
Looks good!

PR #2939, opened 2021-01-29
New order summary component for subscriptions checkouts

i-hardy:

The designs for adding the new print & digital option in the print checkouts include an updated version of the order summary which needs to be able to handle the display of more than one product. Also currently all our checkouts have their own, slightly different order summaries.

This adds an initial version of a new, generic order summary component, developed in Storybook and decoupled from Redux state.

[**Trello Card**](https://trello.com/c/OSsIhbE0)

## Why are you doing this?

Separating the presentational aspect of the order summary from the product-specific logic that determines what we display in it means this component should be reusable across all our checkouts in the future.

This is also a part of the print & digital work that's easy to break out into a separate PR, rather than having one giant PR with everything in it once all the work is done.

## Screenshots

I'm showing the component in the checkout for context- it's not been added yet!

**Mobile**
Screenshot 2021-01-29 at 16 41 56

**Tablet**
Screenshot 2021-01-29 at 16 46 25

**Desktop**
Screenshot_2021-01-29 Support the Guardian Newspaper Subscription

### DIFF
--- a/support-frontend/assets/components/orderSummary/orderSummary.jsx
+++ b/support-frontend/assets/components/orderSummary/orderSummary.jsx
@@ -1,6 +1,7 @@
 // @flow
 
 import React, { type Node } from 'react';
+import { SvgInfo } from '@guardian/src-icons';
 import typeof GridImageType from 'components/gridImage/gridImage';
 import { type GridImg } from 'components/gridImage/gridImage';
 import * as styles from './orderSummaryStyles';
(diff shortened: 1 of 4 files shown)


support-frontend/assets/components/orderSummary/orderSummary.jsx
```diff
@@ -0,0 +1,61 @@
...
+  total: string,
+};
+
+function OrderSummary(props: PropTypes) {
+  return (
+    <aside css={styles.wrapper}>
+      <div css={styles.topLine}>
+        <div css={styles.topLineBorder}>
+          <h3 css={styles.title}>Order summary</h3>
+          <a href={props.changeSubscription}>Change</a>
```
GHaberis, line 28: Do we still want to render this link in the event `props.changeSubscription` is an empty string?

i-hardy: Good point, we probably don't.

GHaberis (approved):
Looks good! 👍

PR #2941, opened 2021-02-01
Supporter product data

rupertbates:
This PR adds a new module `supporter-product-data` which is a Scala step function responsible for polling Zuora for supporter subscription data and writing it to a DynamoDB table for use by `members-data-api`

A full description of what is involved is [here](https://docs.google.com/document/d/1qGwmZ6LxF5Lc51O8GI2IkTuaDuyDiHqSm_QmLhWBvUo/edit?ts=600ebee8#)

[**Trello Card**](https://trello.com/c/MkDloZxc/3576-improve-retrieval-of-supporter-information-from-zuora-members-data-api-etl-pipeline-80)

## Todo:
- [x] UAT config
- [ ] UAT full sync


support-modules/aws/src/main/scala/com/gu/aws/AWSAsyncHandler.scala
```diff
@@ -1,25 +1,24 @@
...
 import com.amazonaws.handlers.AsyncHandler
 import com.amazonaws.{AmazonServiceException, AmazonWebServiceRequest}
-import com.gu.monitoring.SafeLogger
+import com.typesafe.scalalogging.LazyLogging
 
 import scala.concurrent.{Future, Promise}
 
 class AwsAsyncHandler[Request <: AmazonWebServiceRequest, Response](f: (Request, AsyncHandler[Request, Response]) => JFuture[Response], request: Request)
-    extends AsyncHandler[Request, Response] {
+    extends AsyncHandler[Request, Response] with LazyLogging{
```
johnduffell, line 11: do we even use sentry for server side Error reporting nowadays?  I have never looked at it!

rupertbates: Me neither so I would guess not. We could probably remove it

support-modules/rest/src/main/scala/com/gu/rest/WebServiceHelper.scala
```diff
@@ -172,13 +176,13 @@ trait WebServiceHelper[Error <: Throwable] {
...
 
-  private def buildRequest(endpoint: String, headers: ParamMap, params: ParamMap): Request.Builder =
+  protected def buildRequest(endpoint: String, headers: ParamMap, params: ParamMap): Request.Builder =
     new Request.Builder()
       .url(endpointUrl(endpoint, params))
       .headers(buildHeaders(headers))
 
   private def endpointUrl(endpoint: String, params: ParamMap): HttpUrl = {
-    val withSegments = endpoint.split("/").foldLeft(urlBuilder) {
+    val withSegments = endpoint.split("/", -1).foldLeft(urlBuilder) {
```
johnduffell, line 185: what is the -1 for?
```
scala> "asdf/asdas/errwr".split("/",-1)
val res0: Array[String] = Array(asdf, asdas, errwr)

scala> "asdf/asdas/errwr".split("/")
val res1: Array[String] = Array(asdf, asdas, errwr)

```

johnduffell: ah I see gets rid of trailing slashes. Is that intended?
```
scala> "asdf/asdas/errwr/".split("/",-1)
val res2: Array[String] = Array(asdf, asdas, errwr, "")

scala> "asdf/asdas/errwr/".split("/")
val res3: Array[String] = Array(asdf, asdas, errwr)

```

supporter-product-data/build-tc
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -e
+
+# build scala
+cd ..
+./sbt "project supporter-product-data" riffRaffUpload
```
johnduffell, line 6: good idea making it a script.  FYI I recently found you can do `supporter-product-data / riffRaffUpload` to run a task in a project without having to switch to it

supporter-product-data/build.sbt
```diff
@@ -0,0 +1,75 @@
...
+    "-SupporterProductDataQueryZuora-",
+    "-SupporterProductDataFetchResults-",
+    "-SupporterProductDataUpdateDynamo-"
+  ).foreach(functionPartial =>
+    s"aws lambda update-function-code --function-name support${functionPartial}CODE --s3-bucket $s3Bucket --s3-key $s3Path --profile membership --region eu-west-1".!!
+  )
+
+}
+
+lazy val deployToProd = inputKey[Unit]("Directly update AWS lambda code to PROD instead of via RiffRaff for faster feedback loop")
```
johnduffell, line 60: ooh direct to PROD is an interesting one, I would definitely prefer to encourage people to go via riffraff for PROD.

supporter-product-data/cloudformation/cfn.yaml
```diff
@@ -0,0 +1,383 @@
...
+Conditions:
+  CreateProdResources: !Equals [!Ref "Stage", "PROD"]
+
+Mappings:
+  StageVariables:
+    CODE:
+      S3Bucket: supporter-product-data-export-code
+      # Every 10 mins between 7am and 7pm Mon-Fri
+      scheduleRate: "cron(*/10 7-19 ? * 1-5 *)"
+    UAT:
```
johnduffell, line 23: so we have a UAT stage in AWS - normally stages in AWS are code/prod and touchpoint stages are dev/uat/prod.  Why that way?

supporter-product-data/cloudformation/cfn.yaml
```diff
@@ -0,0 +1,383 @@
...
+      - CODE
+    Default: CODE
+
+Conditions:
+  CreateProdResources: !Equals [!Ref "Stage", "PROD"]
+
+Mappings:
+  StageVariables:
+    CODE:
+      S3Bucket: supporter-product-data-export-code
```
johnduffell, line 20: kate mentioned today that S3 bucket names shouldn't really be in public repos, although I'm not sure why exactly

supporter-product-data/cloudformation/cfn.yaml
```diff
@@ -0,0 +1,383 @@
...
+          - Sid: DenyIncorrectEncryptionHeader
+            Effect: Deny
+            Principal: "*"
+            Action: s3:PutObject
+            Resource: !Sub arn:aws:s3:::${SupporterProductExportBucket}/*
+            Condition:
+              StringNotEquals:
+                s3:x-amz-server-side-encryption:
+                  - AES256
+                  - aws:kms
```
johnduffell, line 69: why kms rather than s3 managed encryption?
I think s3 managed encryption is cheaper from a quick look

supporter-product-data/cloudformation/cfn.yaml
```diff
@@ -0,0 +1,383 @@
...
+                  - s3:ListBucket
+                Resource: !Sub arn:aws:s3:::${SupporterProductExportBucket}
+        - PolicyName: SSMConfigParams
+          PolicyDocument:
+            Version: 2012-10-17
+            Statement:
+              - Effect: Allow
+                Action:
+                  - ssm:GetParametersByPath
+                Resource: !Sub arn:aws:ssm:${AWS::Region}:${AWS::AccountId}:parameter/supporter-product-data/*
```
johnduffell, line 139: we have private s3 bucket on line 109 and we have ssm down here. Do we use both?

supporter-product-data/cloudformation/cfn.yaml
```diff
@@ -0,0 +1,383 @@
...
+  SupporterProductDataScheduledRule:
+    Type: "AWS::Events::Rule"
+    Properties:
+      Description: "Trigger regular data synchronisation from Zuora"
+      ScheduleExpression: !FindInMap [StageVariables, !Ref Stage, scheduleRate]
+      State: "ENABLED"
+      Targets:
+        -
+          Arn: !Ref SupporterProductData
+          Id: !Sub trigger_state_machine-${Stage}-1
```
johnduffell, line 329: why "-1" ?

supporter-product-data/cloudformation/cfn.yaml
```diff
@@ -0,0 +1,383 @@
...
+      AlarmActions:
+        - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:reader-revenue-dev
+      AlarmName: !Join
+        - ' '
+        - - 'URGENT 9-5 -'
+          - !Ref 'Stage'
+          - 'Failed to extract supporter product data from Zuora'
+      AlarmDescription: !Join
+        - ' '
+        - - 'Impact - members-data-api attribute data may be incorrect. Fix SupporterProductData ASAP!'
```
johnduffell, line 370:
```suggestion
        - - 'Impact - New subscribers will not get their digital benefits.'
```

supporter-product-data/cloudformation/cfn.yaml
```diff
@@ -0,0 +1,383 @@
...
+
+  SupporterProductDataAlarm:
+    Type: AWS::CloudWatch::Alarm
+    Condition: CreateProdResources
+    Properties:
+      AlarmActions:
+        - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:reader-revenue-dev
+      AlarmName: !Join
+        - ' '
+        - - 'URGENT 9-5 -'
```
johnduffell, line 365: I'm not totally set either way, but I would say it's not "drop everything - it's down" urgent

supporter-product-data/cloudformation/cfn.yaml
```diff
@@ -0,0 +1,383 @@
...
+          - !Ref 'Stage'
+          - 'Failed to extract supporter product data from Zuora'
+      AlarmDescription: !Join
+        - ' '
+        - - 'Impact - members-data-api attribute data may be incorrect. Fix SupporterProductData ASAP!'
+          - 'Follow the process in https://docs.google.com/document/d/1_3El3cly9d7u_jPgTcRjLxmdG2e919zCLvmcFCLOYAk/edit'
```
johnduffell, line 371: needs a section in this doc.  that doc is still useful but might use a bit of a spring clean in general.

supporter-product-data/cloudformation/dynamo-tables.yaml
```diff
@@ -0,0 +1,56 @@
...
+    AllowedValues:
+      - PROD
+      - UAT
+      - CODE
+    Default: CODE
+Mappings:
+  StageVariables:
+    CODE:
+      DynamoWriteCapacity: 1
+      DynamoReadCapacity: 5
```
johnduffell, line 16: why do we need 5? also did you consider on-demand instead for dev/uat databases?

supporter-product-data/cloudformation/dynamo-tables.yaml
```diff
@@ -0,0 +1,56 @@
...
+    UAT:
+      DynamoWriteCapacity: 1
+      DynamoReadCapacity: 5
+    PROD:
+      DynamoWriteCapacity: 10
+      DynamoReadCapacity: 200
```
johnduffell, line 22: we can keep this under review for PROD

supporter-product-data/cloudformation/validate-cloudformation.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+aws --region eu-west-1 cloudformation validate-template --template-body file://cfn.yaml --profile membership
```
johnduffell, line 3: is something missing from the file// url here?  or does it require you to CD to the right directory first? could add `dirname $0` to the url if that's the case

support-modules/rest/src/main/scala/com/gu/rest/WebServiceHelper.scala
rupertbates: It means that we retain trailing slashes on urls when they're specified. Previously they were being stripped which isn't normally a problem, but the Zuora `batch-query/` endpoint is picky about it, it will 404 if the trailing slash isn't present

supporter-product-data/build.sbt
rupertbates: Yeah I might remove this, but when you're starting out this is a lot quicker

supporter-product-data/cloudformation/cfn.yaml
rupertbates: Just for simplicity really, the 2 stages and 3 environments configuration gets really complicated and I think it's overkill for this

supporter-product-data/cloudformation/cfn.yaml
rupertbates: Ah yes good spot, I'd been meaning to remove the bucket, we use SSM in this app. This is a classic copy and paste CFN job!

supporter-product-data/cloudformation/cfn.yaml
rupertbates: Copied from here and didn't notice: https://github.com/guardian/fulfilment-lambdas/blob/ba9cca4a2b36f7f9592a30b2a62a358b79e06644/cloudformation/cloudformation.yaml#L414

rupertbates: Have updated

supporter-product-data/cloudformation/dynamo-tables.yaml
rupertbates: Yes I've been tweaking it as I go along. That read capacity comes from the attribute database in members-data-api which I think is a fairly good guide, might be able to drop the write capacity though

supporter-product-data/cloudformation/dynamo-tables.yaml
rupertbates: I didn't actually though that's not a bad idea, I'll look into it

support-modules/rest/src/main/scala/com/gu/rest/WebServiceHelper.scala
johnduffell: got it, thanks, that makes complete sense 👍

supporter-product-data/cloudformation/cfn.yaml
johnduffell: in that case what is CODE?  I could understand dev/uat/prod or dev/code/prod but code/uat/prod seems confusing

rupertbates: Let's discuss

supporter-product-data/src/main/scala/com/gu/conf/ZuoraQuerierConfig.scala
```diff
@@ -0,0 +1,60 @@
...
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.collection.JavaConverters._
+import scala.concurrent.Future
+
+case class ZuoraQuerierConfig(
+  url: String,
+  partnerId: String,
+  username: String,
+  password: String,
+  discountProductRatePlanIds: List[String]
```
johnduffell, line 18: since we have a closed set of stages, we can just put the non-private information e.g. the rate plans in a Map which would make the config simpler.  I think the more stuff that goes through the normal CI/CD process the better.  Also we get type checking for free and it's easier to follow.

supporter-product-data/src/main/scala/com/gu/conf/ZuoraQuerierConfig.scala
```diff
@@ -0,0 +1,60 @@
...
+        new RuntimeException(s"Missing config value for parameter $name in ZuoraQuerierConfig")
+      ).toTry.get
+
+  private def getParametersByPath(path: String) = {
+    val request: GetParametersByPathRequest = new GetParametersByPathRequest()
+      .withPath(path)
+      .withRecursive(false)
+      .withWithDecryption(true)
+
+    AwsAsync(client.getParametersByPathAsync, request)
```
johnduffell, line 57:
```suggestion
    AwsAsync(client.getParametersByPathAsync, request)map(_.getParameters.asScala.toList)
```

supporter-product-data/cloudformation/cfn.yaml
rupertbates: Again, copied from fulfilment-lambdas, I've changed it now.

supporter-product-data/src/main/scala/com/gu/conf/ZuoraQuerierConfig.scala
rupertbates: I did it that way initially but then you end up passing the stage value around as well as the config which is already stage specific so it's a bit more messy

supporter-product-data/cloudformation/dynamo-tables.yaml
rupertbates: Ok I've updated this to use on demand for code/uat

supporter-product-data/cloudformation/cfn.yaml
rupertbates: Updated as discussed

supporter-product-data/cloudformation/validate-cloudformation.sh
rupertbates: The script is in the same directory as the cfn file

supporter-product-data/src/main/scala/com/gu/conf/ZuoraQuerierConfig.scala
johnduffell: I just want to get rid of the AWS bit, it could be lookup from a Map in this piece of code if that makes more sense structurally, or even just put the stage inside the Config object if you prefer.  Either way it'll get all the benefits I mentioned, the only downside is a slight inconsistency of where the items come from.

supporter-product-data/cloudformation/cfn.yaml
```diff
@@ -0,0 +1,347 @@
...
+      S3Bucket: supporter-product-data-export-dev
+      # Every 10 mins between 7am and 7pm Mon-Fri
+      scheduleRate: "cron(*/10 7-19 ? * 1-5 *)"
+    UAT:
+      S3Bucket: supporter-product-data-export-uat
+      scheduleRate: "cron(*/10 7-19 ? * 1-5 *)"
+    PROD:
+      S3Bucket: supporter-product-data-export-prod
+      # Every 5 mins
+      scheduleRate: "cron(*/5 * ? * * *)"
```
johnduffell, line 29: we schedule every 5 minutes, but the state machine might retry for 20 minutes when fetching the results.  I'm just wondering what would happen if two are doing the incremental query at the same time?

supporter-product-data/src/main/scala/com/gu/lambdas/FetchResultsLambda.scala
```diff
@@ -0,0 +1,44 @@
...
+object FetchResultsLambda {
+  def fetchResults(stage: Stage, jobId: String) = {
+    SafeLogger.info(s"Attempting to fetch results for jobId $jobId")
+    for {
+      config <- ZuoraQuerierConfig.load(stage)
+      service = new ZuoraQuerierService(config, configurableFutureRunner(60.seconds))
+      result <- service.getResults(jobId)
+      _ = assert(result.status == Completed, s"Job with id $jobId is still in status ${result.status}")
+      batch = result.batches.headOption.toRight(s"No batches were returned in the batch query response for jobId $jobId").right.get
+      fileId = batch.fileId.toRight(s"Batch.fileId was missing in jobId $jobId").right.get
```
johnduffell, line 30: doing right.get on an either doesn't actually report the contents of the left in the exception - that only works for `Try`.
```scala
    def get: B = e match {
      case Right(b) => b
      case _        => throw new NoSuchElementException("Either.right.get on Left")
    }
```

rupertbates: That's annoying...

supporter-product-data/src/main/scala/com/gu/conf/ZuoraQuerierConfig.scala
rupertbates: Done

supporter-product-data/src/main/scala/com/gu/services/S3Service.scala
```diff
@@ -0,0 +1,49 @@
...
+
+    val upload = transferManager.upload(
+      bucketName(stage),
+      filename,
+      inputStream,
+      objectMetadata
+    )
+
+    Future(
+      blocking(upload.waitForUploadResult())
```
johnduffell, line 41: `blocking` doesn't always do anything useful depending on the execution context, it can hang if the number of threads can't expand. Why not addProgressListener and listen for a TRANSFER_COMPLETED_EVENT in a similar way to our AwsAsyncHandler class?

supporter-product-data/src/main/scala/com/gu/lambdas/UpdateDynamoLambda.scala
```diff
@@ -0,0 +1,113 @@
...
+class ContextTimeOutCheck(context: Context) extends TimeOutCheck {
+  override def timeRemainingMillis = context.getRemainingTimeInMillis
+}
+
+class UpdateDynamoLambda extends Handler[UpdateDynamoState, UpdateDynamoState] {
+  override protected def handlerFuture(input: UpdateDynamoState, context: Context) = {
+    SafeLogger.info(s"Starting write to dynamo task for ${input.recordCount} records from ${input.filename}")
+    writeToDynamo(Stage.fromEnvironment, input, new ContextTimeOutCheck(context))
+  }
+}
```
johnduffell, line 31: you can use single abstact method syntax to reduce the boilerplate while keeping the nice naming, if you think it's still clear enough
```suggestion
trait TimeOutCheck {
  def timeRemainingMillis(): Int
}

class UpdateDynamoLambda extends Handler[UpdateDynamoState, UpdateDynamoState] {
  override protected def handlerFuture(input: UpdateDynamoState, context: Context) = {
    SafeLogger.info(s"Starting write to dynamo task for ${input.recordCount} records from ${input.filename}")
    writeToDynamo(Stage.fromEnvironment, input, () => context.getRemainingTimeInMillis)
  }
}
```

supporter-product-data/src/main/scala/com/gu/lambdas/UpdateDynamoLambda.scala
```diff
@@ -0,0 +1,113 @@
...
+    val dynamoDBService = DynamoDBService(stage)
+
+    val unProcessed = getUnprocessedItems(csvReader, state.processedCount)
+
+    val validUnprocessed = unProcessed.filter { case (itemOrError, _) => itemOrError.isRight }
+    val invalidUnprocessed = unProcessed.filter { case (itemOrError, _) => itemOrError.isLeft }
+
+    if (invalidUnprocessed.nonEmpty)
+      SafeLogger.error(
+        scrub"There were ${invalidUnprocessed.length} CSV read failures from file ${state.filename} with line numbers ${invalidUnprocessed.map(_._2).mkString(",")}"
```
johnduffell, line 49: how would we know if this happens? perhaps we need to send a metric for alarming?

supporter-product-data/src/main/scala/com/gu/lambdas/UpdateDynamoLambda.scala
```diff
@@ -0,0 +1,113 @@
...
+
+  def writeToDynamo(stage: Stage, state: UpdateDynamoState, timeOutCheck: TimeOutCheck): Future[UpdateDynamoState] = {
+    val csvStream = S3Service.streamFromS3(stage, state.filename)
+    val csvReader = csvStream.asCsvReader[SupporterRatePlanItem](rfc.withHeader)
+    val dynamoDBService = DynamoDBService(stage)
+
+    val unProcessed = getUnprocessedItems(csvReader, state.processedCount)
+
+    val validUnprocessed = unProcessed.filter { case (itemOrError, _) => itemOrError.isRight }
+    val invalidUnprocessed = unProcessed.filter { case (itemOrError, _) => itemOrError.isLeft }
```
johnduffell, line 45: each item should appear in exactly one of the lists so it might be clearer to use partition e.g.
```suggestion
    val (validUnprocessed, invalidUnprocessed) = unProcessed.partition { case (itemOrError, _) => itemOrError.isRight }
```

supporter-product-data/src/main/scala/com/gu/lambdas/UpdateDynamoLambda.scala
```diff
@@ -0,0 +1,113 @@
...
+
+    if (invalidUnprocessed.nonEmpty)
```
johnduffell, line 47: only do this on the first pass?
```suggestion
    if (invalidUnprocessed.nonEmpty && state.processedCount == 0)
```

supporter-product-data/src/main/scala/com/gu/lambdas/UpdateDynamoLambda.scala
johnduffell: actually if we use `collect` then we can `filter` and `.right.get` (so delete line 53) at the same time.  I'm not sure there's such a thing as partitionCollect or something.
```suggestion
    val validUnprocessed = unProcessed.collect { case (Right(item), index) => (item, index) }
    val invalidUnprocessed = unProcessed.collect { case (Left(error), _) => (error, index) }
```

johnduffell, line 45: delete line 53 and line 49 delete the `.map(_._2)` and
```suggestion
    val validUnprocessed = unProcessed.collect { case (Left(item), index) => (item, index) }
    val invalidUnprocessed = unProcessed.collect { case (Left(_), index) => index }
```

supporter-product-data/src/main/scala/com/gu/lambdas/UpdateDynamoLambda.scala
```diff
@@ -0,0 +1,113 @@
...
+          SafeLogger.info(
+            s"Aborting processing - time remaining: ${timeOutCheck.timeRemainingMillis / 1000} seconds, buffer: ${timeoutBufferInMillis / 1000} seconds"
+          )
+          return processedCount
+        }
+        SafeLogger.info(
+          s"Continuing processing - time remaining: ${timeOutCheck.timeRemainingMillis / 1000} seconds, buffer: ${timeoutBufferInMillis / 1000} seconds"
+        )
+
+        Await.result(writeBatch(group, dynamoDBService), 120.seconds)
```
johnduffell, line 89: foldLeft on an iterator cant terminate early without the return trick which I guess is why you needed the Await?  awaiting will block an extra thread and also will throw any failed futures

I can't help thinking that we're trying to reimplement error recovery and replay ourselves.  What would the cloud native version look like?  Run the query and load the CSV into a sqs queue or kinesis stream, then process the stream/queue until it is empty?  Or would we just run the query from scratch and reprocess everything if something goes wrong?

supporter-product-data/src/main/scala/com/gu/model/Stage.scala
```diff
@@ -0,0 +1,27 @@
...
+  implicit val decoder: Decoder[Stage] = Decoder.decodeString.emap(fromString)
+
+  def fromString(str: String): Either[String, Stage] =
+    List(DEV, UAT, PROD)
+      .find(_.value == str)
+      .toRight(s"Unknown batch status $str")
+
+  def fromEnvironment = fromString(Option(System.getenv("Stage"))
+    .getOrElse("DEV"))
+    .getOrElse(Stage.DEV)
```
johnduffell, line 25:
```suggestion

  def fromEnvironment: Stage =
    Option(System.getenv("Stage"))
      .flatMap(fromString(_).toOption)
      .getOrElse(Stage.DEV)

```

supporter-product-data/src/main/scala/com/gu/model/dynamo/SupporterRatePlanItem.scala
```diff
@@ -0,0 +1,29 @@
...
+object SupporterRatePlanItem {
+  implicit val decoder: HeaderDecoder[SupporterRatePlanItem] =
+    HeaderDecoder.decoder(
+      identityId.zuoraName,
+      gifteeIdentityId.zuoraName,
+      ratePlanId.zuoraName,
+      productRatePlanId.zuoraName,
+      productRatePlanName.zuoraName,
+      termEndDate.zuoraName
+    )(SupporterRatePlanItem.apply)
```
johnduffell, line 27: this is great - most places that we use CSVs we rely heavily on field ordering, so this is a really nice feature to give us more certainty

supporter-product-data/src/main/scala/com/gu/services/S3Service.scala
rupertbates: Good suggestion, have implemented

supporter-product-data/src/main/scala/com/gu/model/zuora/request/BatchQueryRequest.scala
```diff
@@ -0,0 +1,32 @@
...
+
+object ZoqlExportQuery{
+  implicit val encoder: Encoder[ZoqlExportQuery] = deriveEncoder[ZoqlExportQuery].mapJsonObject(_
+    .add("type", Json.fromString("zoqlexport"))
+  )
+}
+
+object BatchQueryRequest{
+  implicit val encoder: Encoder[BatchQueryRequest] = deriveEncoder[BatchQueryRequest].mapJsonObject(_
+    .add("format", Json.fromString("csv"))
```
johnduffell, line 25: just a thought - the data compresses excellently (but not the query execution time) if you specify gzip here, do you think it's worth it?  There is also content-encoding header which would make more sense as it's more transparent but I'm not sure if zuora respects that.

supporter-product-data/src/main/scala/com/gu/lambdas/UpdateDynamoLambda.scala
rupertbates: The issue is the lambda timing out when we have a large number of records, that would be an problem whether we write the items to a queue or directly to dynamo. The only other way round it would be to use something other than a lambda/step function like aws glue but there are other disadvantages to that such as the modularity of the codebase and familiarity of the tech to people who will need to maintain this code. I think this is approach gives us a decent compromise.

supporter-product-data/src/main/scala/com/gu/lambdas/UpdateDynamoLambda.scala
rupertbates: I've added a metric/alarm for this and dynamo write failures

supporter-product-data/src/main/scala/com/gu/model/zuora/request/BatchQueryRequest.scala
rupertbates: I did think about it and decided against it. The size of the file is not really a problem even for a full query so it doesn't seem worth the extra complexity

supporter-product-data/cloudformation/cfn.yaml
```diff
@@ -345,3 +345,55 @@ Resources:
...
+      MetricName: CsvReadFailure
+      Namespace: supporter-product-data
+      Dimensions:
+        - Name: Environment
+          Value: PROD
+      ComparisonOperator: GreaterThanOrEqualToThreshold
+      Threshold: 1
+      Period: 60
+      EvaluationPeriods: 1
+      Statistic: Average
```
johnduffell, line 373: minor as the threshold is 1 anyway but
```suggestion
      Statistic: Sum
```

supporter-product-data/cloudformation/cfn.yaml
```diff
@@ -345,3 +345,55 @@ Resources:
...
+      MetricName: DynamoWriteFailure
+      Namespace: supporter-product-data
+      Dimensions:
+        - Name: Environment
+          Value: PROD
+      ComparisonOperator: GreaterThanOrEqualToThreshold
+      Threshold: 1
+      Period: 60
+      EvaluationPeriods: 1
+      Statistic: Average
```
johnduffell, line 399: as above 
```suggestion
      Statistic: Sum
```

supporter-product-data/src/main/scala/com/gu/services/DynamoDBService.scala
```diff
@@ -0,0 +1,72 @@
...
+import java.time.{LocalDate, ZoneId, ZoneOffset}
+import java.util.concurrent.CompletionException
+import scala.collection.JavaConverters._
+import scala.compat.java8.FutureConverters._
+import scala.concurrent.{ExecutionContext, Future}
+
+
+class DynamoDBService(client: DynamoDbAsyncClient, tableName: String) {
+
+  def writeItem(item: SupporterRatePlanItem)(implicit executionContext: ExecutionContext) = {
```
johnduffell, line 21: for me loads of the code gets red underlined with a warning, does that happen for you?  I think it did offer to use the scala linter rules when I first imported so that may be why
image

supporter-product-data/src/main/scala/com/gu/services/SelectActiveRatePlansQuery.scala
```diff
@@ -0,0 +1,39 @@
...
+
+import com.gu.model.FieldsToExport._
+import com.gu.model.Stage
+import com.gu.model.Stage.{DEV, PROD}
+
+import java.time.LocalDate
+
+object SelectActiveRatePlansQuery {
+
+  val name = "select-active-rate-plans"
```
johnduffell, line 11:
```suggestion
  val name = "active-rate-plans"
```

supporter-product-data/src/main/scala/com/gu/services/SelectActiveRatePlansQuery.scala
```diff
@@ -0,0 +1,39 @@
...
+
+  val isNotDSGift = "Subscription.RedemptionCode__c = ''"
+  val isRedeemedDSGift = s"(Subscription.RedemptionCode__c != '' AND ${gifteeIdentityId.zuoraName} != '')"
+
+  def excludeDiscountProductRatePlans(discountProductRatePlanIds: List[String]) =
+    discountProductRatePlanIds
+      .map(discountProductRatePlanId => s"${productRatePlanId.zuoraName} != '$discountProductRatePlanId' AND")
+      .mkString("\n")
```
johnduffell, line 19: just to make the query consistent below do this and add the AND to the end of line 34
```suggestion
      .map(discountProductRatePlanId => s"${productRatePlanId.zuoraName} != '$discountProductRatePlanId'")
      .mkString(" AND\n")
```

supporter-product-data/src/main/scala/com/gu/services/SelectActiveRatePlansQuery.scala
```diff
@@ -0,0 +1,39 @@
...
+          ${identityId.zuoraName},
+          ${gifteeIdentityId.zuoraName},
+          ${ratePlanId.zuoraName},
+          ${productRatePlanId.zuoraName},
+          ${productRatePlanName.zuoraName},
+          ${termEndDate.zuoraName}
+            FROM
+            rateplan
+            WHERE
+            ${termEndDate.zuoraName} >= '$date' AND
```
johnduffell, line 32: that's good, so even if we change the date, it still carries on from where we left off?

supporter-product-data/src/main/scala/com/gu/services/SelectActiveRatePlansQuery.scala
```diff
@@ -0,0 +1,39 @@
...
+      .mkString("\n")
+
+  def query(date: LocalDate, discountProductRatePlanIds: List[String]): String =
+    s"""SELECT
+          ${identityId.zuoraName},
+          ${gifteeIdentityId.zuoraName},
+          ${ratePlanId.zuoraName},
+          ${productRatePlanId.zuoraName},
+          ${productRatePlanName.zuoraName},
+          ${termEndDate.zuoraName}
```
johnduffell, line 28: traditionally the subscription validity is based on being covered by a charge rather than just being within term.  I think that was more of a concern when things were originally migrated in, and newly acquirable products have the charges set up to fill the subscription term.  Is this logic matching the old members-data-api logic?

supporter-product-data/src/main/scala/com/gu/services/SelectActiveRatePlansQuery.scala
johnduffell: oh ignore that, I realised it doesn't submit any query on subsequent increments

supporter-product-data/src/main/scala/com/gu/services/ZuoraQuerierService.scala
```diff
@@ -0,0 +1,59 @@
...
+  val authHeaders = Map(
+    "apiSecretAccessKey" -> config.password,
+    "apiAccessKeyId" -> config.username
+  )
+
+  def postQuery(queryType: QueryType): Future[BatchQueryResponse] = {
+    val queries = queryType match {
+      case Full => List(
+        ZoqlExportQuery(
+          s"${SelectActiveRatePlansQuery.name}-${UUID.randomUUID().toString}",
```
johnduffell, line 33:
```suggestion
          s"${SelectActiveRatePlansQuery.name}-${LocalDateTime.now(ZoneId.of("UTC")).toString}",
```
might be a little more helpful for later debugging
```
scala> LocalDateTime.now(ZoneId.of("UTC")).toString
res2: String = 2021-02-05T17:31:10.332
```

supporter-product-data/src/main/scala/com/gu/services/ZuoraQuerierService.scala
```diff
@@ -0,0 +1,59 @@
...
+    )
+    postJson[BatchQueryResponse](s"batch-query/", request.asJson, authHeaders)
+  }
+
+  def getResults(id: String): Future[BatchQueryResponse] =
+    get(s"batch-query/jobs/$id", authHeaders)(
+      BatchQueryResponse.decoder, // having to explicitly pass these implicits or the compiler gets confused by ambiguous implicit values
+      BatchQueryErrorResponse.decoder,
+      classTag[BatchQueryResponse]
+    )
```
johnduffell, line 52: seems ok to me
```suggestion
    get[BatchQueryResponse](s"batch-query/jobs/$id", authHeaders)
```

supporter-product-data/src/test/scala/com/gu/Fixtures.scala
```diff
@@ -0,0 +1,8 @@
+package com.gu
+
+import scala.io.Source
+
+//noinspection SourceNotClosed
+object Fixtures {
+  def loadQueryResults: String = Source.fromURL(getClass.getResource("/query-results.csv")).mkString
```
johnduffell, line 7: marginal, but it's one small file so the indirection may be unnecessary
```suggestion
  def loadQueryResults: String =
    """
Account.IdentityId__c,Subscription.GifteeIdentityId__c,RatePlan.Id,ProductRatePlan.Id,ProductRatePlan.Name,Subscription.TermEndDate,Subscription.RedemptionCode__c,Subscription.GifteeIdentityId__c
13380875,,2c92a00777427d35017744760ace4c77,2c92a0ff5345f9220153559d915d5c26,Percentage,2022-01-27,,
13380875,,2c92a00777427d35017744760aef4c7c,2c92a0fb4edd70c8014edeaa4e972204,Digital Pack Annual,2022-01-27,,
3204264,105320827,2c92a00b7742896c017744780f57375e,2c92a00773adc09d0173b99e4ded7f45,Digital Subscription One Year Fixed,2022-01-28,gd12-aw655rhh,105320827
""".trim
```

supporter-product-data/src/test/scala/com/gu/model/dynamo/SerialisationSpec.scala
```diff
@@ -0,0 +1,28 @@
+package com.gu.model.dynamo
+
+import com.gu.Fixtures
+import kantan.csv.ops.toCsvInputOps
+import kantan.csv.rfc
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+
+class SerialisationSpec extends AnyFlatSpec with Matchers {
+  SupporterRatePlanItem.getClass.getSimpleName should "deserialise correctly" in {
```
johnduffell, line 10: clever way to get the test name from the source code, nice bonus for refactoring later!

supporter-product-data/src/test/scala/com/gu/model/zuora/response/SerialisationSpec.scala
```diff
@@ -0,0 +1,41 @@
...
+    "startTime": "2021-01-18T17:13:48+0000",
+    "version": "1.0",
+    "format": "CSV",
+    "name": "Fulfilment-Queries",
+    "id": "2c92c0f977091c5c0177167d05f77d7c",
+    "offset": 0
+}
+    """
+    val result = decode[BatchQueryResponse](json)
+    result.isRight shouldBe true
```
johnduffell, line 38: you can use inside or pattern here to check the fields you're interested in
https://www.scalatest.org/user_guide/using_matchers#matchingAPattern

```suggestion
    result should matchPattern { case Right(BatchQueryResponse("", Submitted, List(BatchQueryItem("", "", 0)))) => }
```

supporter-product-data/src/test/scala/com/gu/lambdas/UpdateDynamoSpec.scala
```diff
@@ -0,0 +1,38 @@
...
+      UpdateDynamoState(
+        filename = "New Subscriptions-2c92c0867728a2e801772ade71bb56a5",
+        recordCount = 50,
+        processedCount = 0
+      ),
+      new TimeOutCheck {
+        override def timeRemainingMillis = 9999999
+      }
+    ).map(_ =>
+      succeed
```
johnduffell, line 27: how do we know the item got to dynamo?

supporter-product-data/src/test/scala/com/gu/services/DynamoDBServiceSpec.scala
```diff
@@ -0,0 +1,24 @@
...
+class DynamoDBServiceSpec  extends AsyncFlatSpec with Matchers {
+
+  "DynamoDBService" should "stream be able to insert an item" in {
+    val service = DynamoDBService(DEV)
+    val item = SupporterRatePlanItem(
+      "999999",
+      None,
+      "1", "1", "Digital Subscription", LocalDate.parse("2021-01-01")
+    )
+    service.writeItem(item).map(_ => succeed)
```
johnduffell, line 21: this will pass as long as an exception wasn't thrown. worth to check that, if not rename to 
`"DynamoDBService" should "not throw an exception on insert"`

supporter-product-data/src/test/scala/com/gu/services/S3ServiceSpec.scala
```diff
@@ -0,0 +1,23 @@
...
+@IntegrationTest
+class S3ServiceSpec extends AsyncFlatSpec with Matchers {
+
+  "S3Service" should "stream to S3 successfully" in {
+    val initialString = "text".getBytes()
+    val stream = new ByteArrayInputStream(initialString)
+    S3Service
+      .streamToS3(DEV, s"test-file-${UUID.randomUUID().toString}", stream, initialString.length)
+      .map(_ =>
+        succeed
```
johnduffell, line 20: would be worth reading back S3 at this point

supporter-product-data/src/test/scala/com/gu/services/ZuoraQuerierServiceSpec.scala
```diff
@@ -0,0 +1,43 @@
...
+    } yield result
+
+    futureResult.map {
+      response =>
+        response.id shouldNot be("")
+        response.status shouldBe Submitted
+    }
+  }
+
+  it should "upload a file to S3 successfully" in {
```
johnduffell, line 31:
```suggestion
  it should "stream a file from zuora to S3 successfully" in {
```

supporter-product-data/src/test/scala/com/gu/services/ZuoraQuerierServiceSpec.scala
```diff
@@ -0,0 +1,43 @@
...
+    val fileId = "2c92c085771fa1a70177202f9ae63130"
+    val futureResult = for {
+      config <- ZuoraQuerierConfig.load(DEV)
+      service = new ZuoraQuerierService(config, configurableFutureRunner(60.seconds))
+      response <- service.getResultFileResponse(fileId)
+      uploadResult <- S3Service.streamToS3(DEV, s"upload-test-file-$fileId", response.body.byteStream(), response.body.contentLength)
+    } yield uploadResult
+    futureResult.map(_ =>
+      succeed
```
johnduffell, line 40: check if it really got there

supporter-product-data/src/test/scala/com/gu/services/ZuoraQuerierServiceSpec.scala
```diff
@@ -0,0 +1,43 @@
...
+
+    futureResult.map {
+      response =>
+        response.id shouldNot be("")
+        response.status shouldBe Submitted
+    }
+  }
+
+  it should "upload a file to S3 successfully" in {
+    val fileId = "2c92c085771fa1a70177202f9ae63130"
```
johnduffell, line 32: this file will only be there for 180 days

supporter-product-data/src/test/scala/conf/ZuoraQuerierConfigSpec.scala
```diff
@@ -0,0 +1,20 @@
+package conf
+
+import com.gu.conf.ZuoraQuerierConfig
+import com.gu.model.Stage.DEV
+import com.gu.test.tags.annotations.IntegrationTest
+import org.scalatest.flatspec.AsyncFlatSpec
+import org.scalatest.matchers.should.Matchers
+
+@IntegrationTest
```
johnduffell, line 9: these new IT tests won't automatically run by the IT test runner, we would need to add them to the relevant config so that it will pick them up.

supporter-product-data/src/test/scala/conf/ZuoraQuerierConfigSpec.scala
```diff
@@ -0,0 +1,20 @@
...
+class ZuoraQuerierConfigSpec extends AsyncFlatSpec with Matchers{
+  "ZuoraQuerierConfig" should "load from SSM" in {
+    val futureConfig = ZuoraQuerierConfig.load(DEV)
+    futureConfig.map {
```
johnduffell, line 13:
```suggestion
    ZuoraQuerierConfig.load(DEV).map {
```

johnduffell:
> A full description of what is involved is here

That info is really helpful - we need to make sure information is still available in future, it's probably fine as it seems to be in the shared drive, but do move to the wiki if necessary.

supporter-product-data/src/main/scala/com/gu/conf/ZuoraQuerierConfig.scala
```diff
@@ -55,21 +52,15 @@ object ZuoraQuerierConfig {
       )
     }
 
-  private def getParameterValue(configPath: String, name: String, params: List[Parameter]) =
+  private def findParameterValue(name: String, params: List[Parameter]) =
     params
-      .find(_.getName == configPath + name)
+      .find(_.getName.endsWith(name))
```
johnduffell, line 57: I had to check there that none are a suffix of another - if we're sure we're only fetching non-recursively, how about `split` it on a dot and check the `last`?

supporter-product-data/src/main/scala/com/gu/conf/ZuoraQuerierConfig.scala
```diff
@@ -55,21 +52,15 @@ object ZuoraQuerierConfig {
...
 
-  private def getParametersByPath(path: String) = {
-    val request: GetParametersByPathRequest = new GetParametersByPathRequest()
-      .withPath(path)
-      .withRecursive(false)
-      .withWithDecryption(true)
+  private def findParameterOrThrow(name: String, params: List[Parameter]) =
+    findParameterValue(name, params).toRight(
+      new RuntimeException(s"Missing config value for parameter $name in ZuoraQuerierConfig")
+    ).toTry.get
```
johnduffell, line 63:
```suggestion
    findParameterValue(name, params).getOrElse(
      throw new RuntimeException(s"Missing config value for parameter $name in ZuoraQuerierConfig")
    )
```

supporter-product-data/src/main/scala/com/gu/lambdas/QueryZuoraLambda.scala
```diff
@@ -19,17 +20,18 @@ class QueryZuoraLambda extends Handler[QueryZuoraState, FetchResultsState] {
...
-object QueryZuoraLambda {
+object QueryZuoraLambda extends StrictLogging {
   def queryZuora(stage: Stage, queryType: QueryType) = {
-    SafeLogger.info(s"Attempting to submit ${queryType.value} query to Zuora")
+    logger.info(s"Attempting to submit ${queryType.value} query to Zuora")
+    val attemptedQueryTime = LocalDateTime.now.minusMinutes(1)
     for {
       config <- ZuoraQuerierConfig.load(stage)
       service = new ZuoraQuerierService(config, configurableFutureRunner(60.seconds))
       result <- service.postQuery(queryType)
```
johnduffell, line 30: presumaably todo pass the attemptedQueryTime into the query

supporter-product-data/src/test/scala/com/gu/services/ParameterStoreServiceSpec.scala
```diff
@@ -0,0 +1,22 @@
...
+
+@IntegrationTest
+class ParameterStoreServiceSpec extends AsyncFlatSpec with Matchers {
+  ParameterStoreService.getClass.getSimpleName should "be able to put and get a parameter" in {
+    val paramName = "integrationTestParam"
+    val paramVal = LocalDateTime.now.format(DateTimeFormatter.ISO_DATE_TIME)
+    val service = ParameterStoreService(DEV)
+    for {
+      _ <- service.putParameter(paramName, paramVal)
+      result <- service.getParameter(paramName)
```
johnduffell, line 19: good to check the value was really written 👏

supporter-product-data/src/main/scala/com/gu/lambdas/QueryZuoraLambda.scala
johnduffell: ok I hve got it now - we pass in the previous time from the config object already - the current query attemptedQueryTime we just have to guess conservatively as they don't tell us the current session time.

supporter-product-data/src/test/scala/com/gu/Fixtures.scala
rupertbates: I put this in when the file was much bigger, I'm going to leave it now

supporter-product-data/src/test/scala/conf/ZuoraQuerierConfigSpec.scala
rupertbates: I don't think it's worth running them in the IT test runner to be honest, they are mostly there to help out when developing locally

supporter-product-data/src/test/scala/com/gu/services/ZuoraQuerierServiceSpec.scala
rupertbates: I've removed this test

supporter-product-data/src/test/scala/com/gu/services/ZuoraQuerierServiceSpec.scala
rupertbates: I've removed this test, everything within it was already being tested somewhere else

supporter-product-data/src/test/scala/com/gu/services/DynamoDBServiceSpec.scala
rupertbates: Now checks response was a success

supporter-product-data/src/test/scala/com/gu/lambdas/UpdateDynamoSpec.scala
rupertbates: I've removed this test it was a duplicate

johnduffell (approved):
this was a very complex piece of work and has a big impact architecturally on how digital fulfilment works, so it's great that you (and I) have put so much thought into it.  I do feel like we've been around the block a few times with it so far, but the good thing is this is a technical piece of work that we own all the aspects, so it's entirely within our domain to make the decisions!  So hopefully we won't have to go back round the block again 🤞 
Thanks for this so far. On to the next card...! 🚀

supporter-product-data/src/main/scala/com/gu/services/SelectActiveRatePlansQuery.scala
mario-galic: The query name `select-active-rate-plans` might be confusing as this query will not in all cases identify only the **active** rate plans, but also past ones and the ones far in the future (if there were multiple renewals amendments scheduled for the future). However if the intention of the query is to identify active subscriptions (as opposed to rateplans) it might be good enough, but perhaps this should be reflected in the scala object and zuora query names.

PR #2943, opened 2021-02-03
Do not use localstorage for AB tests

tomrf1:
Currently we're storing AB test participations in local storage, and reading them for every page view. This is actually wrong as it may bring back tests that are no longer running.
There is no reason for storing the participations in local storage, as we can recompute on each page load, and it should be deterministic because it's based on mvt value.

This PR also ensures only contributions pages are assigned an amounts test variant.


rupertbates (commented):
LGTM 👍

rupertbates (approved):

PR #2944, opened 2021-02-03
Extend existing list component for use in print & digital checkout

i-hardy:

This is an extension/improvement of the List component built for the Guardian Weekly page, to be able to re-use it in the print and digital checkout.

[**Trello Card**](https://trello.com/c/OSsIhbE0)

## Why are you doing this?

The new checkout design calls for bulleted lists in both the order summary and the 'add the digital subscription' CTA that have the same style and use of a secondary line of text, so it makes sense to have this be a single component. We already had a list component; this PR simply adds the ability to adjust the colour and size of the bullets, and inject a different component as the list item.

## Screenshots

**In Storybook - mobile**
Screenshot 2021-02-04 at 10 47 54

**In Storybook - desktop**
Screenshot 2021-02-04 at 10 47 29

**In the OrderSummary component**
Screenshot 2021-02-04 at 10 48 14

**On Guardian Weekly landing page (no regression from what's currently there)**
Screenshot 2021-02-04 at 10 48 31

### DIFF
--- a/support-frontend/assets/components/list/list.jsx
+++ b/support-frontend/assets/components/list/list.jsx
@@ -15,6 +15,7 @@ export type ListItemText = {
 type ListBulletSize = 'small' | 'large';
 
 type ListBulletColour = 'light' | 'dark';
+
 type ListProps = {
   items: ListItemText[],
   bulletSize?: ListBulletSize,
@@ -97,6 +98,7 @@ const bulletSizes: { [key: ListBulletSize]: string } = {
   large: listItemBulletLarge,
   small: listItemBulletSmall,
 };
+
 function ListItem({ item, colour, size }: ListItemProps) {
   return (
     <li key={item.explainer} css={listItem}>
--- a/support-frontend/assets/components/list/list.jsx
+++ b/support-frontend/assets/components/list/list.jsx
@@ -8,7 +8,7 @@ import { from } from '@guardian/src-foundations/mq';
 import { body } from '@guardian/src-foundations/typography';
 
 export type ListItemText = {
-  explainer: string,
+  content: string,
   subText?: string
 }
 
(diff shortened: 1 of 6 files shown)


support-frontend/assets/components/list/list.jsx
```diff
@@ -2,16 +2,35 @@
...
-import { brandAlt } from '@guardian/src-foundations/palette';
+import { brand, brandAlt } from '@guardian/src-foundations/palette';
 import { space } from '@guardian/src-foundations';
 import { from } from '@guardian/src-foundations/mq';
 import { body } from '@guardian/src-foundations/typography';
 
-type ListPropTypes = {
-  items: Array<Object>,
+export type ListItemText = {
+  explainer: string,
```
johnduffell, line 11: for a generic list, explainer seems like a strange name. how about `content` ?

support-frontend/assets/components/list/list.jsx
```diff
@@ -31,35 +50,118 @@ const listItem = css`
...
-    ))}
-    </ul>
+    <li key={item.explainer} css={listItem}>
+      <span css={[
+        listItemBullet,
+        bulletColours[colour],
+        bulletSizes[size],
+      ]}
+      />
+      <span css={listItemContent}>{item.explainer}</span>
```
johnduffell, line 111: just to confirm - if subText is passed in it will be just ignored for this list item

johnduffell (approved):
looks good, nice work, just a query about the unfied name mainText/explainer 👍

support-frontend/assets/components/list/list.jsx
i-hardy: Yep, this component isn't supposed to show it

support-frontend/assets/components/list/list.jsx
i-hardy: That was mainly not to cause regressions with the GW landing page, but you're right it's not a great name, it's worth changing.

PR #2948, opened 2021-02-15
Enable display of custom hero title, copy and roundel for promo codes on all product pages

i-hardy:

This enables any custom title, hero section copy and roundel copy set in the promo code tool for a particular code to show up on the digital subscriptions landing page, and sets up the ability to do the same on the paper landing page once this is enabled in the tool.

It also extracts and refactors the approach to picking up promo code copy from the Guardian Weekly page, ensuring that this is done in a consistent way across all the product pages.

[**Trello Card**](https://trello.com/c/aaor71qQ)

## Why are you doing this?

Only being able to set custom promotion copy on the Weekly landing page is confusing for users of the promo code tool and hinders what they can do with the codes. It also creates a confusing difference of behaviour between pages in the front end code.

## Screenshots

**Digital subscription - promo code DISC503 (dev only)**
Screenshot_2021-02-15 Support the Guardian The Guardian Digital Subscription

**Guardian Weekly - promo code STAFFGW25 (dev only)**
Screenshot_2021-02-15 The Guardian Weekly Subscriptions The Guardian

### DIFF
--- a/support-frontend/app/controllers/DigitalSubscriptionController.scala
+++ b/support-frontend/app/controllers/DigitalSubscriptionController.scala
@@ -55,7 +55,7 @@ class DigitalSubscriptionController(
 
   def digital(countryCode: String, orderIsAGift: Boolean): Action[AnyContent] = CachedAction() { implicit request =>
     implicit val settings: AllSettings = settingsProvider.getAllSettings()
-    val title = s"Support the Guardian | The Guardian Digital ${if (orderIsAGift) "Gift "}Subscription"
+    val title = s"Support the Guardian | The Guardian Digital ${if (orderIsAGift) "Gift " else ""}Subscription"
     val mainElement = EmptyDiv("digital-subscription-landing-page-" + countryCode)
     val js = Left(RefPath("digitalSubscriptionLandingPage.js"))
     val css = Left(RefPath("digitalSubscriptionLandingPage.css"))
--- a/support-frontend/app/controllers/DigitalSubscriptionController.scala
+++ b/support-frontend/app/controllers/DigitalSubscriptionController.scala
@@ -74,9 +74,10 @@ class DigitalSubscriptionController(
       country <- countryGroup.countries.headOption
     } yield country).getOrElse(UK)
     val maybePromotionCopy = queryPromos.headOption.flatMap(promoCode =>
-      ProductPromotionCopy(promotionServiceProvider
-        .forUser(false), stage)
-        .getCopyForPromoCode(promoCode, DigitalPack, country)
+      ProductPromotionCopy(
+        promotionServiceProvider.forUser(false),
+        stage
+      ).getCopyForPromoCode(promoCode, DigitalPack, country)
     )
     val readerType = if (orderIsAGift) Gift else Direct
     val productPrices = priceSummaryServiceProvider.forUser(false).getPrices(DigitalPack, promoCodes, readerType)
--- a/support-frontend/app/controllers/DigitalSubscriptionController.scala
+++ b/support-frontend/app/controllers/DigitalSubscriptionController.scala
@@ -74,10 +74,8 @@ class DigitalSubscriptionController(
       country <- countryGroup.countries.headOption
     } yield country).getOrElse(UK)
     val maybePromotionCopy = queryPromos.headOption.flatMap(promoCode =>
-      ProductPromotionCopy(
-        promotionServiceProvider.forUser(false),
-        stage
-      ).getCopyForPromoCode(promoCode, DigitalPack, country)
+      ProductPromotionCopy(promotionServiceProvider.forUser(false), stage)
+        .getCopyForPromoCode(promoCode, DigitalPack, country)
     )
     val readerType = if (orderIsAGift) Gift else Direct
     val productPrices = priceSummaryServiceProvider.forUser(false).getPrices(DigitalPack, promoCodes, readerType)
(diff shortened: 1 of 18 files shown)


support-frontend/app/controllers/DigitalSubscriptionController.scala
```diff
@@ -51,24 +55,29 @@ class DigitalSubscriptionController(
...
-
+    val country = (for {
+      countryGroup <- CountryGroup.byId(countryCode)
+      country <- countryGroup.countries.headOption
+    } yield country).getOrElse(UK)
+    val maybePromotionCopy = queryPromos.headOption.flatMap(promoCode =>
+      ProductPromotionCopy(promotionServiceProvider
+        .forUser(false), stage)
+        .getCopyForPromoCode(promoCode, DigitalPack, country)
+    )
```
johnduffell, line 80: just indentation/formatting
```suggestion
    val maybePromotionCopy = queryPromos.headOption.flatMap(promoCode =>
      ProductPromotionCopy(
        promotionServiceProvider.forUser(false),
        stage
      ).getCopyForPromoCode(promoCode, DigitalPack, country)
    )
```

support-frontend/app/controllers/DigitalSubscriptionController.scala
```diff
@@ -51,24 +55,29 @@ class DigitalSubscriptionController(
...
       "en" -> buildCanonicalDigitalSubscriptionLink("int", orderIsAGift)
     )
-
+    val country = (for {
+      countryGroup <- CountryGroup.byId(countryCode)
+      country <- countryGroup.countries.headOption
+    } yield country).getOrElse(UK)
+    val maybePromotionCopy = queryPromos.headOption.flatMap(promoCode =>
+      ProductPromotionCopy(promotionServiceProvider
+        .forUser(false), stage)
```
johnduffell, line 78: how do we use promo codes with UAT? if someone is a test user would they want to see UAT promo copy?

support-frontend/assets/pages/paper-subscription-landing/paperSubscriptionLandingPageReducer.js
```diff
@@ -20,6 +21,7 @@ export type State = {
   common: CommonState,
   page: {
     productPrices: ?ProductPrices,
+    promotionCopy: ?PromotionCopy,
```
johnduffell, line 24: I don't think promotion copy will change while the page is open.  does it really need to be copied in the state rather than leaving it as a global const?

johnduffell (approved):
looks good, thanks for doing all this! just added a couple of comments to review.  I did look at the client side stuff but feel free to get another pair of eyes on those parts as  it's not my specialist area! 👍

support-frontend/app/controllers/DigitalSubscriptionController.scala
i-hardy: I'm not sure about that- this is just how it's currently set up for the Guardian Weekly page.

support-frontend/assets/pages/paper-subscription-landing/paperSubscriptionLandingPageReducer.js
i-hardy: There's a lot of stuff in Redux state on the landing pages that won't change while the page is open- I think it's worth considering if we need to put any of it into Redux, but I think it would be more confusing currently to have this arbitrarily not happen for some properties or on some pages but do it on others.

support-frontend/app/controllers/DigitalSubscriptionController.scala
i-hardy: It looks like scalastyle doesn't like this as the method is now > 50 lines long (weirdly IntelliJ decided not to highlight this when I checked the change there 🤔). Not sure what the best thing to do is.

support-frontend/app/controllers/DigitalSubscriptionController.scala
johnduffell: fair enough, I don't think it's essential that it works.

support-frontend/app/controllers/DigitalSubscriptionController.scala
```diff
@@ -51,24 +55,30 @@ class DigitalSubscriptionController(
...
+    val country = (for {
+      countryGroup <- CountryGroup.byId(countryCode)
+      country <- countryGroup.countries.headOption
+    } yield country).getOrElse(UK)
+    val maybePromotionCopy = queryPromos.headOption.flatMap(promoCode =>
+      ProductPromotionCopy(
+        promotionServiceProvider.forUser(false),
+        stage
+      ).getCopyForPromoCode(promoCode, DigitalPack, country)
+    )
```
johnduffell, line 81: probably the quickert reasonable option would be this
```suggestion
    val maybePromotionCopy = queryPromos.headOption.flatMap(promoCode =>
      ProductPromotionCopy(promotionServiceProvider.forUser(false), stage)
        .getCopyForPromoCode(promoCode, DigitalPack, country)
    )
```
a better long term solution would be move the whole of lines 72-81 to a new getPromotionCopy function which would be called from line 100, as it looks like country is only needed for this.

support-frontend/app/controllers/DigitalSubscriptionController.scala
johnduffell: Iv'e added another suggestion above in another comment!

I find intellij is a bit random on the scalastyle - mine seems to find wrnings in a lot of (other peoples) code that clearly passes teamcity, but other times teamcity starts complaining about things I've done that were not underlined.  I suspect there are different setting for different projects but haven't looked into it.

support-frontend/assets/pages/paper-subscription-landing/paperSubscriptionLandingPageReducer.js
johnduffell: yeah agreed.  I tend not to put things in redux if I don't have to, but I can see the value of consistency at least in this change.

PR #2950, opened 2021-02-17
Track Subscriptions Pricing Card viewability in Ophan

GHaberis:

We'd like to be able to track when certain components have been viewed by our users. This PR enables that by setting up a new helper module `useHasBeenSeen`, this is a close copy of the a module in `dotcom-rendering` that already offers the functionality we're after here, the original can be found at: https://github.com/guardian/dotcom-rendering/blob/main/src/web/lib/useHasBeenSeen.ts.`useHasBeenSeen` makes use of the native [Intersection Observer API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) to observe a component's visibility, and do something when it has become viewable. What you deem viewable can be customised using intersection observer options.

I've updated the `ProductOption` component to use `useHasBeenSeen`, when `hasbeenseen` is `true` we execute `props.onView`, which is an addition to the `props` of this component. This `onView` function calls `sendTrackingEventsOnView`, which is another new function I've set up in `helpers/subscriptions.js`.

[**Trello Card**](https://trello.com/b/TjUElW0B/subscriptions-team)

## Why are you doing this?

This means we can start tracking our components viewability on the Data Lake.

## Screenshots

You can see a console log on the lefthand side (temporarily added for demo purposes) for each of the pricing cards.

trackingView


support-frontend/assets/components/product/productOption.jsx
```diff
@@ -73,10 +76,27 @@ const productOptionHighlight = css`
   padding: ${space[2]}px ${space[3]}px;
   ${headline.xxsmall({ fontWeight: 'bold' })};
 `;
-
 function ProductOption(props: Product) {
+  const [hasBeenSeen, setNode] = useHasBeenSeen({
```
rupertbates, line 80: Could we put types on these consts?

rupertbates: Also is there a more descriptive name you could use for `setNode` I can't really tell from this what it does?

GHaberis: > Could we put types on these consts?

I'll have a go at this, I was running into issues with `flow` when typing the React Hooks (this array is the return result of a React State Hook), possibly related to why we use `// $FlowIgnore` on any React Hook import, such as `useEffect` or `useState`.

> Also is there a more descriptive name you could use for setNode?

Yeah I kept the variable name from the `dotcom-rendering` implementation, `setNode` is called when the component in this case ProductOption is rendered with a reference to the `HTMLElement` (DOM node) we want to observe. `setElementToObserve` may be a more descriptive alternative?

rupertbates: Sounds good

rupertbates (approved):
👍 LGTM

support-frontend/assets/components/product/productOption.jsx
GHaberis: OK I revisited typing React Hooks, it turns out the version of `flow-bin` we're using is quite out of date (2+ years old) and doesn't include types for React Hooks. This wasn't totally apparent to begin with as my IDE linked to type definitions for these functions, on closer look these were actually Typescript definitions that lived in `node_modules/@types/react`, which must be a child dependency for one of our dependencies!

I attempted to upgrade `flow-bin` and doing so created all kinds of issues. So I'm not going to type these variables (we don't type them anywhere else we've used hooks), and I think the proposed Typescript migration removes the incentive to upgrade flow.

support-frontend/assets/helpers/useHasBeenSeen.js
```diff
@@ -0,0 +1,46 @@
...
+  const intersectionCallback = options.debounce
+    ? libDebounce(intersectionFn, 200)
+    : intersectionFn;
+
+  useEffect(() => {
+    if (observer.current) {
+      observer.current.disconnect();
+    }
+
+    observer.current = new window.IntersectionObserver(
```
i-hardy, line 29: Are we polyfilling IntersectionObserver, or will this otherwise fail gracefully in IE?

GHaberis: Good shout. Looks like we're using a different polyfill.io path on support.theguardian.com compared to the main domain, ours unfortunately doesn't polyfill, whereas there's does. It should be resolvable by adding "IntersectionObserver. IntersectionObserverEntry" to the URL.

GHaberis: OK I've updated the polyfill script we use to match the script used on [dotcom-rendering](https://github.com/guardian/dotcom-rendering/blob/82c6327083272d2ec8f563e15706ac6a933b8224/src/web/server/document.tsx#L173) and `frontend`. This script lives on `assets.guim.co.uk`, so required a change to our Content Security Policy to enable it. This script includes the IntersectionObserver polyfill.

I've tested before/after on IE11 using Browserstack, and the new Polyfill works as expected.

I noted whilst doing this we had a random 50/50 test for loading v2/v3 of the polyfill.io script, this was set-up back in June 2019: https://github.com/guardian/support-frontend/pull/1872. I guess it's been in place this entire time. I've decided to remove it as I don't think we're still monitoring the results?

Pinging @guardian/contributions for visibility on this change.

support-frontend/app/views/main.scala.html
```diff
@@ -144,18 +144,18 @@
...
     </script>
 
-    @if(Random.nextBoolean) {
-      <script src="https://polyfill.io/v3/polyfill.min.js?features=default,Array.prototype.find,fetch,Array.prototype.includes,Number.isInteger,Object.values,Object.entries&callback=polyfillLoaded"></script>
-      <script>window.guardian.polyfillVersion = 'v3';</script>
-    } else {
-      <script src="https://polyfill.io/v2/polyfill.min.js?features=default,Array.prototype.find,fetch,Array.prototype.includes,Number.isInteger,Object.values,Object.entries&callback=polyfillLoaded"></script>
-      <script>window.guardian.polyfillVersion = 'v2';</script>
-    }
+    <script src="https://assets.guim.co.uk/polyfill.io/v3/polyfill.min.js?rum=0&features=es6,es7,es2017,es2018,es2019,default-3.6,HTMLPictureElement,IntersectionObserver,IntersectionObserverEntry,fetch,NodeList.prototype.forEach&flags=gated&callback=guardianPolyfilled&unknown=polyfill&cacheClear=1"></script>
```
GHaberis, line 152: I've tested whether the polyfills we were previously getting (`Array.prototype.find, fetch, Array.prototype.includes, Number.isInteger, Object.values, Object.entries`) work and they're all there in IE11.

PR #2951, opened 2021-02-18
Fix REMINDER_PLATFORM

tomrf1:
See https://github.com/guardian/support-reminders/blob/fe449111af2c66d055e55d39e1d3ac4486df1dfa/src/create-reminder-signup/lambda/models.ts#L25


TomPretty (approved):
Oh good spot! Are we going to update the current signups?

PR #2954, opened 2021-02-22
add instacne names so we can make sense of AWS console

johnduffell:
This makes it easier to do reservations and find unused instances
image

tested in CODE
image


rupertbates (approved):
Good idea

PR #2955, opened 2021-02-23
make the landing page use a default promo for all copy (DS only)

johnduffell:

This adds a default promo to the DS landing page, meaning that the headline and subpara can be edited in the promo code tool.

## Why are you doing this?

This was the number 1 thing that marketing want when I asked.
Also, it's 10% time, and I decided to do this.

## Screenshots
promo code tool
image
landing page
image

### DIFF
--- a/support-frontend/app/controllers/DigitalSubscriptionController.scala
+++ b/support-frontend/app/controllers/DigitalSubscriptionController.scala
@@ -61,7 +61,6 @@ class DigitalSubscriptionController(
     val description = stringsConfig.digitalPackLandingDescription
     val canonicalLink = Some(buildCanonicalDigitalSubscriptionLink("uk", orderIsAGift))
     val queryPromos = request.queryString.get("promoCode").map(_.toList).getOrElse(Nil)
-    val (productPrices, maybePromotionCopy) = landingCopyProvider.promoDetails(queryPromos, countryCode, orderIsAGift)
     val hrefLangLinks = Map(
       "en-us" -> buildCanonicalDigitalSubscriptionLink("us", orderIsAGift),
       "en-gb" -> buildCanonicalDigitalSubscriptionLink("uk", orderIsAGift),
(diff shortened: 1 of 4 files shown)


support-frontend/app/controllers/DigitalSubscriptionController.scala
```diff
@@ -184,3 +173,28 @@ class DigitalSubscriptionController(
...
+    val promoCodes: List[PromoCode] = queryPromos ++ DefaultPromotions.DigitalSubscription.all
+
+    val country = (for {
+      countryGroup <- CountryGroup.byId(countryCode)
+      country <- countryGroup.countries.headOption
+    } yield country).getOrElse(UK)
+    val promoCode = queryPromos.headOption.getOrElse(DefaultPromotions.DigitalSubscription.landing)
+    val promotionCopy = ProductPromotionCopy(promotionServiceProvider.forUser(false), stage)
+      .getCopyForPromoCode(promoCode, DigitalPack, country)
+    val readerType = if (orderIsAGift) Gift else Direct
```
rupertbates, line 195: I'm not sure it adds much putting the price stuff in here - it complicates the return type and then we immediately extract the tuple in the calling method. Why not just deal with the copy in here and leave prices outside?

rupertbates (approved):
With one suggestion 👍

support-frontend/app/controllers/DigitalSubscriptionController.scala
johnduffell: good question.  The above method is perilously close to the 50 lines limit so I was a bit enthusiastic.  However you are right that they are mostly independent so they shouldn't be mixed here either.
I was half hoping the extracted method could be used in Paper and Weekly in the next PR.
I'll rejig it in this PR along the lines you suggest.

johnduffell:
working in CODE - works both with and without the promo code present (uses default copy if not present)
Alwyas uses the "DEV" promo code in CODE and will always use the "PROD" one in PROD.

PR #2956, opened 2021-02-24
Amazon Pay for Recurring - prototype validation

paulbrown1982:
Prototype changes to validate recurring mode in Amazon Pay, as demoed on 24 Feb. For expediency it modified the single checkout rather than adds the widgets to the recurring flow, that still needs to be done. 

The Amazon docs are here: https://developer.amazon.com/docs/amazon-pay-automatic/add-address-and-wallet-widgets.html, basically we just need to do steps 2 and 3.

The Zuora docs are here: https://knowledgecenter.zuora.com/Billing/Billing_and_Payments/M_Payment_Gateways/Supported_Payment_Gateways/Amazon_Pay_Gateway but I've done most of that, you just need to get the Billing Agreement ID into the first field of a CC Reference Transaction Amazon Pay payment method.

To run locally in prod mode you need to add the prod credentials to the `AmazonPayConfig.scala` file, and I had to add the additional CORS URLs changes below to `DEV.public.conf`.

You can log into Seller Central here: https://sellercentral.amazon.com/
To test as a customer in sandbox mode you need to create a new user here: https://sellercentral.amazon.com/gp/pyop/seller/testing


rupertbates:
Closing as this is a year old now, the branch will still be there if needed.

PR #2961, opened 2021-02-26
use bionic / arm AMI for support-logs proxy

johnduffell:
This PR adds ARM instance and uses bionic ubuntu for support-logs proxy.

This is needed because ubuntu xenial isn't supported by AWS any more so the amigo bakes don't work.  Also we might as well use ARM instances because they are cheaper.

Wait for the result of https://amigo.gutools.co.uk/recipes/ubuntu-bionic-arm-with-nginx/bakes/1 before merging.

### DIFF
--- a/support-logs/cfn.yaml
+++ b/support-logs/cfn.yaml
@@ -73,7 +73,7 @@ Resources:
     Type: AWS::EC2::Instance
     Properties:
         ImageId: !Ref AMI
-        InstanceType: t3.micro
+        InstanceType: t4g.micro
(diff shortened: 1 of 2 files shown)


rupertbates (approved):
🌮

johnduffell:
there's no CODE env so let's see what happens

johnduffell:
seems to be still working.  However I do notice the elk nodes just restarted, so it may be that smalls (2GB ram) are not enough and we need the bigger instances (4gb ram)

PR #2967, opened 2021-03-02
Add delivery instructions above delivery info box on Print HD checkout

lucymonie:
Adding supporting text to the Paper home delivery checkout delivery instructions text area component.

[**Trello Card**](https://trello.com/c/0ANg5Dw7/3595-add-delivery-instructions-above-delivery-info-box-on-print-hd-checkout)

## Why are you doing this?
Our print distribution crew have flagged that delivery drivers are having difficulty finding the letterbox of many initial newspaper deliveries, especially in the winter darkness. Resulting in customer service complaints. The aim is to encourage people to add instructions and reduce the customer complaints.

## Notes on implementation
1. Please note that I have altered the supporting text very slightly to 'Crinan Street' rather than 'Crinan St' as the screen reader was reading it out as 'Crinan saint'
2. The trello card asks that this field remains optional but without the optional flag. At present the form will submit even if there is no text and without the `optional` prop passed into the component but it's not ideal as explained in [Source](https://theguardian.design/2a1e5182b/p/567182--text-area/b/42916b) so I have made it optional and passed in the prop.

Screen Shot 2021-03-02 at 12 13 45

## Screenshots
Screen Shot 2021-03-02 at 12 51 35

### DIFF
--- a/support-frontend/assets/pages/paper-subscription-checkout/components/paperCheckoutForm.jsx
+++ b/support-frontend/assets/pages/paper-subscription-checkout/components/paperCheckoutForm.jsx
@@ -267,6 +267,7 @@ function PaperCheckoutForm(props: PropTypes) {
                   css={controlTextAreaResizing}
                   id="delivery-instructions"
                   label="Delivery instructions"
+                  supporting="Please let us know any details to help us find your property (door colour, any access issues) and the best place to leave your newspaper. For example, 'Front door - red - on Crinan St, put through letterbox'"
                   maxlength={250}
                   value={props.deliveryInstructions}
                   onChange={e => props.setDeliveryInstructions(e.target.value)}
--- a/support-frontend/assets/pages/paper-subscription-checkout/components/paperCheckoutForm.jsx
+++ b/support-frontend/assets/pages/paper-subscription-checkout/components/paperCheckoutForm.jsx
@@ -271,6 +271,7 @@ function PaperCheckoutForm(props: PropTypes) {
                   maxlength={250}
                   value={props.deliveryInstructions}
                   onChange={e => props.setDeliveryInstructions(e.target.value)}
+                  optional
                 />
                 : null
             }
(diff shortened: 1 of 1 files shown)


support-frontend/assets/pages/paper-subscription-checkout/components/paperCheckoutForm.jsx
```diff
@@ -267,9 +267,11 @@ function PaperCheckoutForm(props: PropTypes) {
                   css={controlTextAreaResizing}
                   id="delivery-instructions"
                   label="Delivery instructions"
+                  supporting="Please let us know any details to help us find your property (door colour, any access issues) and the best place to leave your newspaper. For example, 'Front door - red - on Crinan Street, put through letterbox'"
                   maxlength={250}
                   value={props.deliveryInstructions}
                   onChange={e => props.setDeliveryInstructions(e.target.value)}
+                  optional
```
GHaberis, line 274: Sorry if I've missed it but I can't see the value of the `optional` prop being set anywhere, it looks like this should be a `boolean` from the Design System [documentation](https://theguardian.design/2a1e5182b/p/663879--select-box/b/59d4fd) so it should be explicitly set to `true`.

lucymonie: You're so right, but the linter yells at me if I try to do that!
Screen Shot 2021-03-02 at 14 21 53

lucymonie: I think you only explicitly set it for false in this paradigm, and not setting it is like setting it to false.

GHaberis: OK I see, if if it's present it's `truthy` and if it's absent is `falsey`

GHaberis (approved):
Looks good to me 👍

PR #2968, opened 2021-03-03
Remove autocomplete from delivery instructions field

lucymonie:
The delivery instructions text area had an autocomplete popup on it, which wasn't appropriate.

## Why are you doing this?
For a better user experience.

## Notes on implementation
This should have been a simple case of adding `autocomplete="off"` to the textarea component. However, when I added this attribute, it did appear in the html, but it didn't stop the autocomplete popup. This is apparently caused by an issue in Chrome, where they were trying to hand off decisions about autocomplete to the user (though still seemingly allowing developers to turn off autocomplete for a whole form).

There are a number of solutions suggested in this [stack overflow post](https://stackoverflow.com/questions/12374442/chrome-ignores-autocomplete-off) and the one that seemed least hacky to me was to set `autocomplete="new-password"`. This works because the browser won't try to autocomplete a field that requires a new password. However, I'm happy to discuss this if anyone wants to argue for a different solution. 

I tested this on chrome, firefox and safari and it worked on all.

### Broken  😢 
Screen Shot 2021-03-03 at 14 06 59

### Working  🎉 
Screen Shot 2021-03-03 at 14 06 30

### DIFF
--- a/support-frontend/assets/pages/paper-subscription-checkout/components/paperCheckoutForm.jsx
+++ b/support-frontend/assets/pages/paper-subscription-checkout/components/paperCheckoutForm.jsx
@@ -267,6 +267,7 @@ function PaperCheckoutForm(props: PropTypes) {
                   css={controlTextAreaResizing}
                   id="delivery-instructions"
                   label="Delivery instructions"
+                  autocomplete="new-password"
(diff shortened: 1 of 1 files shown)


GHaberis:
Nice PR description of the issue/solution, I think a comment in the code explaining why we're using `autocomplete="new-password"` in this context might also help?

i-hardy (approved):
👍

PR #2969, opened 2021-03-04
Show amount validation errors immediately

tomrf1:
Currently field errors are not displayed until the user clicks the Contribute button.
But the 'other amount' field should behave differently, and display validation errors immediately.

No input:
Screen Shot 2021-03-04 at 08 31 02

Valid input:
Screen Shot 2021-03-04 at 08 31 07

Too low:
Screen Shot 2021-03-04 at 08 31 15

Not a number:
Screen Shot 2021-03-04 at 08 31 43


support-frontend/assets/pages/contributions-landing/components/ContributionAmount.jsx
```diff
@@ -31,7 +32,7 @@ type PropTypes = {|
   selectedAmounts: SelectedAmounts,
   selectAmount: (number | 'other', CountryGroupId, ContributionType) => (() => void),
   otherAmounts: OtherAmounts,
-  checkOtherAmount: (string, CountryGroupId, ContributionType) => boolean,
```
tomrf1, line 34: no need for this to be a prop

support-frontend/assets/pages/contributions-landing/components/ContributionAmount.jsx
```diff
@@ -126,21 +126,17 @@ function withProps(props: PropTypes) {
...
-    )}
-      isValid={checkOtherAmount(otherAmount || '', props.countryGroupId, props.contributionType)}
-      formHasBeenSubmitted={(checkoutFormHasBeenSubmitted || stripePaymentRequestButtonClicked)}
+      onInput={e => props.updateOtherAmount(
+        (e.target: any).value,
+        props.countryGroupId,
+        props.contributionType,
+      )}
+      isValid={checkAmount(otherAmount || '', props.countryGroupId, props.contributionType)}
+      formHasBeenSubmitted={(checkoutFormHasBeenSubmitted || stripePaymentRequestButtonClicked || otherAmount)}
```
tomrf1, line 137: The generic `ContributionTextInputDs` component does not display errors until `formHasBeenSubmitted`.
But for this particular field we want to display errors as soon as `otherAmount` has a value.

support-frontend/assets/pages/contributions-landing/components/ContributionAmount.jsx
```diff
@@ -126,21 +126,17 @@ function withProps(props: PropTypes) {
     <ContributionTextInputDs
       id="contributionOther"
       name="contribution-other-amount"
-      type="number"
```
tomrf1, line 129: Defining this field as type="number" gets in the way of our own validation. It prevents the onUpdate callback from being called if e.g. the user enters a char

support-frontend/assets/pages/contributions-landing/components/ContributionAmount.jsx
```diff
@@ -31,7 +32,7 @@ type PropTypes = {|
   selectedAmounts: SelectedAmounts,
   selectAmount: (number | 'other', CountryGroupId, ContributionType) => (() => void),
   otherAmounts: OtherAmounts,
-  checkOtherAmount: (string, CountryGroupId, ContributionType) => boolean,
+  // eslint-disable-next-line react/no-unused-prop-types
```
TomPretty, line 35: do we need this?

tomrf1: it complains that I'm not using it, even though I am. It's very odd

support-frontend/assets/pages/contributions-landing/components/ContributionAmount.jsx
TomPretty: I guess this is maybe a bad prop name? The input shouldn't really care if the form has been submitted or not, just whether or not it should show an error message. Perhaps (and this can be done as a follow up) we should consider renaming to e.g `shoudShowErrorMessage`?

tomrf1: Yes I wondered this. We currently have these props:
```
errorMessage: string | null,
isValid: boolean,
formHasBeenSubmitted: boolean,
```

Perhaps this would be better?
```
errorMessage: string | null,
shouldShowErrorMessage: boolean,
```

TomPretty: Yeah I think that would definitely be an improvement. In fact I'm not 100% if we need this wrapper around the design systems text input?

tomrf1: The only reason for it appears to be the `cssInvalid` override, but I just tested removing it and it makes no difference!

support-frontend/assets/pages/contributions-landing/components/ContributionAmount.jsx
TomPretty: spoke about this in a chat - but just for posterity:

this is a known "false-positive" with this rule: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-unused-prop-types.md#false-positives-sfc

seems like the suggested fix is to assign it to a local variable inside the outer component, and then reference that inside the helper function

TomPretty: or keep the eslint ignore ofc! but in that case it's probably worth adding a comment explaining it's a false positive

support-frontend/assets/pages/contributions-landing/components/ContributionAmount.jsx
tomrf1: I've just pushed a commit which deletes this component as it wasn't adding much.
And setting the error is now as simple as passing `string | null` to the `TextInput`

support-frontend/assets/pages/contributions-landing/components/ContributionAmount.jsx
tomrf1: happily, the refactor has made this issue go away because it's no longer within a helper function!

support-frontend/assets/helpers/formValidation.js
```diff
@@ -44,20 +44,23 @@ export const checkGiftStartDate: (string | null) => boolean = (rawDate) => {
...
+  contributionType: ContributionType,
+): boolean =>
+  isNotEmpty(input)
     && isLargerOrEqual(config[countryGroupId][contributionType].min, input)
     && isSmallerOrEqual(config[countryGroupId][contributionType].max, input)
     && maxTwoDecimals(input);
 
 
-export const checkAmountOrOtherAmount: (SelectedAmounts, OtherAmounts, ContributionType, CountryGroupId) => boolean = (
+export const amountOrOtherAmountIsValid = (
```
TomPretty, line 58: It would be good to refactor this at some point. I think all we actually need to do is check if the amount the user is about to contribute is valid.

tomrf1: agreed, it's messy

TomPretty (commented):
The code looks great to me and really happy to now just be using the design systems text input. I just wanted to check a UX thing with Iona as to me it feels maybe slightly worse than before in certain cases 

Screenshot 2021-03-08 at 09 13 43

tomrf1:
I've now changed it to `onBlur`, so it waits until you move on from the field before displaying the error.
This still isn't quite perfect, but we're doing a UX review of the form soon so it's best to wait for that.

PR #2970, opened 2021-03-05
Allow outbound traffic from Wazuh scanner

jorgeazevedo:

This is opening ports 1514 and 1515 for outbound traffic, which is currently restricted to port 443.

## Why are you doing this?

There's currently a major review of the guardian's security practices under the cyber security programme. One of the projects in this programme is to install a vulnerability scanner called Wazuh in guardian ec2 instances.

The scanner has already been installed in the ami image, but it's attempt to connect with the central server is being block by the instance security group. Hopefully this will fix it, otherwise I'll have to review network ACLs!

## Testing

Works in code ✅

### DIFF
--- a/support-frontend/cloud-formation/cfn.yaml
+++ b/support-frontend/cloud-formation/cfn.yaml
@@ -96,6 +96,7 @@ Resources:
       ImageId: !Ref AMI
       SecurityGroups:
         - !Ref InstanceSecurityGroup
+        - !Ref WazuhSecurityGroup
       InstanceType: !FindInMap [ StageVariables, !Ref Stage, InstanceType ]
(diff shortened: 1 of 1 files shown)


johnduffell (approved):
thanks, looks good! 👍

johnduffell:
why don't we need the DescribeEC2Policy here?

johnduffell:
Also, support-payment-api is also in this repo, so also might need changing.

jorgeazevedo:
> why don't we need the DescribeEC2Policy here? is it because it's already here

Indeed, this service already has those permissions

> Also, support-payment-api is also in this repo, so also might need changing.

Good shout! I've liaised with contributions for that one on #2971 . If you spot anything else I've missed feel free to drop me a line!

PR #2971, opened 2021-03-05
Payment api: Patch in necessary permissions for wazuh

jorgeazevedo:

* Open ports 1514 and 1515 for outbound traffic
* Give permission to the instance to call describe-tags,
   so the agent can identify itself in a human readable way with 
   the wazuh server.

## Why are you doing this?

There's currently a major review of the guardian's security practices under the cyber security programme. One of the projects in this programme is to install a vulnerability scanner called Wazuh in guardian ec2 instances.

The scanner has already been installed in the ami image, but it's attempt to connect with the central server is being block by the instance security group. Hopefully this will fix it, otherwise I'll have to review network ACLs!

## Testing

Works in code ✅

### DIFF
--- a/support-payment-api/src/main/resources/cloud-formation.yaml
+++ b/support-payment-api/src/main/resources/cloud-formation.yaml
@@ -392,6 +392,7 @@ Resources:
   WazuhSecurityGroup:
     Type: AWS::EC2::SecurityGroup
     Properties:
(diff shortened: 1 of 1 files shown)


jorgeazevedo:
@tomrf1

tomrf1 (approved):
👍

PR #2975, opened 2021-03-08
Track "pricing options" CTA on Digital Subs landing page

GHaberis:

We're not tracking clicks on the "pricing options" CTA on the Digital Subs landing page, this PR adds a tracking call in the click handler (for both the gifting and non-gifting landing pages).


i-hardy (approved):
👍

PR #2966, opened 2021-03-01
Digital Subs - updating LP with product information

lucymonie:
## What are you doing in this PR?
This is a small copy change on the hero of the digital subscriptions landing page. The change is for copy for all regions except Australia.

[**Trello Card**](https://trello.com/c/IiYiFqQN/3602-digital-subs-updating-lp-with-product-information)

## Why are you doing this?
To support the February campaign we need to update the digital subscriptions LP with product information.  

## Screenshots
Screen Shot 2021-03-01 at 18 19 43

PR #2977, opened 2021-03-09
Fix digital subscriptions hero styling

lucymonie:
Fixing some spacing that went awry in this PR: https://github.com/guardian/support-frontend/pull/2966

## Why are you doing this?
This adds css that will work for text that comes in from the promo tool and also for the default text that's in support-frontend.

### Text from the promo tool
Screen Shot 2021-03-09 at 10 47 05

### Default text
Screen Shot 2021-03-09 at 11 00 37


i-hardy (approved):
👍

PR #2979, opened 2021-03-10
Send refViewId to ophan

tomrf1:
The pageview table in datalake has a `referrerPageViewId` column, but it is rarely populated for support pageviews.
This is because the existing mechanism in tracker-js relies on attaching an event listener to links, and adding the id to localstorage, see:
https://github.com/guardian/ophan/blob/75b86abcce07369c8998521399327d436246c016/tracker-js/assets/coffee/ophan/click-path-capture.coffee#L41

We reliably add the REFPVID param to the querystring, so this PR uses that to set the localstorage item if it's not already set.

Using localstorage isn't ideal, but it works and I'd prefer not to make changes in ophan.

Now the initial ophan event includes `refViewId`:
Screen Shot 2021-03-10 at 11 01 09

### DIFF
--- a/support-frontend/assets/helpers/page/page.js
+++ b/support-frontend/assets/helpers/page/page.js
@@ -38,7 +38,7 @@ import {
   type CountryGroupId,
   detect as detectCountryGroup,
 } from 'helpers/internationalisation/countryGroup';
-import { trackAbTests } from 'helpers/tracking/ophan';
+import { trackAbTests, setRefViewId } from 'helpers/tracking/ophan';
 import { getSettings } from 'helpers/globals';
 import { getGlobal } from 'helpers/globals';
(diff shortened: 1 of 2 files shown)


TomPretty (commented):
Seems like a good quick fix to me, but will defer to @rtyley here!

TomPretty (approved):
👍

rtyley:
https://github.com/guardian/ophan/pull/4073 seems to be related?!

tomrf1:
> [guardian/ophan#4073](https://github.com/guardian/ophan/pull/4073) seems to be related?!

I don't think we could be sending empty refViewIds as we guard against it here:
https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/helpers/tracking/ophan.js#L117

PR #2980, opened 2021-03-10
Add 40 character limit to state field in billing and delivery address form

lucymonie:
Adding validation to the state / province / territory address field for billing and delivery addresses so they can't be longer than 40 characters.

[**Trello Card**](https://trello.com/c/Xn0mmPUu/3601-add-40-character-limit-to-state-field-in-billing-and-delivery-address-form-25)

## Why are you doing this?
There have been a number of issues with customer payments recently, in which state names that are longer than we can store have been submitted. We'd like to prevent this from happening.

## Screenshots
Screen Shot 2021-03-10 at 12 51 19

Screen Shot 2021-03-10 at 12 50 20

Screen Shot 2021-03-10 at 12 54 10

Screen Shot 2021-03-10 at 12 50 08

### DIFF
--- a/support-frontend/assets/components/subscriptionCheckouts/address/addressFieldsStore.jsx
+++ b/support-frontend/assets/components/subscriptionCheckouts/address/addressFieldsStore.jsx
@@ -127,7 +127,7 @@ const applyBillingAddressRules = (fields: FormFields, addressType: AddressType):
     rule: checkStateFieldLength(fields.state),
     error: formError(
       'state',
-      fields.country === 'CA' ? 'Please enter a province/territory no longer than 40 characters' : 'Please enter a state name no longer than 40 characters',
+      fields.country === 'CA' ? `Please enter a ${addressType} province/territory no longer than 40 characters` : `Please enter a ${addressType} state name no longer than 40 characters`,
     ),
   },
 ]);
--- a/support-frontend/assets/components/subscriptionCheckouts/address/addressFieldsStore.jsx
+++ b/support-frontend/assets/components/subscriptionCheckouts/address/addressFieldsStore.jsx
@@ -82,7 +82,7 @@ const isStateNullable = (country: Option<IsoCountry>): boolean =>
   country !== 'AU' && country !== 'US' && country !== 'CA';
 
 const checkStateFieldLength = (state: string | null): boolean =>
-  state.length <= 40;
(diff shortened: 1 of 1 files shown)


support-frontend/assets/components/subscriptionCheckouts/address/addressFieldsStore.jsx
```diff
@@ -81,6 +81,9 @@ const checkpostCodeLength = (input: string | null): boolean => ((input == null)
 const isStateNullable = (country: Option<IsoCountry>): boolean =>
   country !== 'AU' && country !== 'US' && country !== 'CA';
 
+const checkStateFieldLength = (state: string | null): boolean =>
+  ((state == null) || (state.length <= 40));
```
i-hardy, line 85: Is it worth making this more reusable, rather than hard-coding the number 40 (and in the error message too)? We might want to validate for different lengths on other inputs, and it would be helpful to pull out the number into a variable with a descriptive name in any case.

support-frontend/assets/components/subscriptionCheckouts/address/addressFieldsStore.jsx
```diff
@@ -81,6 +81,9 @@ const checkpostCodeLength = (input: string | null): boolean => ((input == null)
 const isStateNullable = (country: Option<IsoCountry>): boolean =>
   country !== 'AU' && country !== 'US' && country !== 'CA';
 
+const checkStateFieldLength = (state: string | null): boolean =>
```
johnduffell, line 84: how about this? 
```suggestion
const isStateFieldShortEnough = (state: string | null): boolean =>
```

johnduffell (approved):
thanks, this will make life better! 👍

support-frontend/assets/components/subscriptionCheckouts/address/addressFieldsStore.jsx
lucymonie: Thanks John! I ended up making this `checkLength` based on @i-hardy's suggestion that a more reusable function would be better.

johnduffell: fair enough, we could have followed both suggestions and called it `isFieldShortEnough` but either way it's pretty minor on such a simple function.

PR #2981, opened 2021-03-12
Remove doNotTrack logic

tomrf1:
Copying @GHaberis words in trello:

> On support-frontend we honor doNotTrack headers. The maybeTrack, doNotTrack and thirdPartyTrackingEnabled checks are quite confusing, especially at first look because they're all used side-by-side.
> 
> Wikipedia says "DNT is not widely adopted by the industry, with companies citing the lack of legal mandates for its use, as well as unclear standards and guidelines for how websites are to interpret the header. Thus, it is not guaranteed that enabling DNT will actually have any effect at all"
> 
> Therefore I think we could probably remove the doNotTrack/maybeTrack logic from the repo and let the CMP thirdPartyTrackingEnabled consent state check to rule them all.
> 
> This will simplify the code.


support-frontend/assets/helpers/tracking/ophan.js
```diff
@@ -81,17 +80,15 @@ type OphanABPayload = {
...
-      ophan.sendInitialEvent(url, referrer);
-    } catch (e) {
-      console.log(`Error in Ophan tracking: ${e}`);
-    }
-  });
+const pageView = (url: string, referrer: string) => {
+  try {
+    ophan.sendInitialEvent(url, referrer);
+  } catch (e) {
+    console.log(`Error in Ophan tracking: ${e}`);
```
johnduffell, line 89: relevant to the current situation but not this PR - presumably this doesn't end up anywhere we can track it (e.g. cloudwatch)

paulbrown1982 (approved):
The topic of the DNT header has been raised on the Transparency and Consent OKR backlog to make a determination to be used company-wide.

👍  w/r Ophan in that it's important support and www sites treat that consistently.
👍  w/r Google Tag Manager, which is not on the www site, and is currently configured as a custom vendor requiring specific purposes outlined in the Sourcepoint TCFv2-based consent banner.

PR #2984, opened 2021-03-16
Show active navigation link even with query params in url 

GHaberis:

I noticed when clicking on traffic drivers (eg. Banners) from the main site to the Support site that the navigation didn't indicate the initial active page on which I had landed. This is because the logic to set the active state on a navigation link doesn't take into account query string parameters, which all our traffic driving CTAs use for tracking purposes. This PR addresses that so when you click on a traffic driving CTA the nav will show the active state (the yellow bar) on the appropriate link.

## Why are you doing this?

Gives users landing on our site a better indication of where they are in relation to the navigation.

## Screenshots

Screenshot 2021-03-16 at 12 09 18

### DIFF
--- a/support-frontend/assets/components/headers/links/links.jsx
+++ b/support-frontend/assets/components/headers/links/links.jsx
@@ -92,10 +92,14 @@ function internationalisationID(countryGroupId: ?CountryGroupId = null): ?string
 
 // Export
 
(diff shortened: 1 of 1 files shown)


lucymonie (approved):
Nice one! 👍

PR #2985, opened 2021-03-16
Remove Google Analytics conditional check in-front of Tag Manager

GHaberis:

This PR simplifies the conditional that checks whether we can use Google Tag Manager by removing the part that checks whether Google Analytics has been granted user consent. We can do this as we're now passing Google Analytics consent state to Tag Manager and making that conditional check in tag manager.


rupertbates (approved):
👍

PR #2986, opened 2021-03-16
make support frontend use SSM tunnel for SSH access

johnduffell:
Co authored with @i-hardy and @lucymonie 

This PR updates support-frontend to use SSM Tunnel
https://github.com/guardian/ssm-scala#enabling-ssm-tunnel

https://trello.com/c/mr8yEEdp/3634-switch-support-frontend-to-ssm

## Why are you doing this?

It's latest security best practice, it only relies on built in AWS infrastructure and AWS credentials.

### DIFF
--- a/support-frontend/README.md
+++ b/support-frontend/README.md
@@ -1,8 +1,8 @@
 Frontend for the new [supporter platform](https://support.theguardian.com/).
 
 ### SSH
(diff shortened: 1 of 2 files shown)


kelvin-chappell (approved):
👍

PR #2987, opened 2021-03-17
Set referrer url

tomrf1:
Previous attempt: https://github.com/guardian/support-frontend/pull/2979

We're now sending `refViewId` with the initial ophan event.
However, we also need to send `ref` (which has the url), otherwise ophan ignores it. See here:
https://github.com/guardian/ophan/blob/75b86abcce07369c8998521399327d436246c016/the-slab/app/extractors/ReferrerExtractor.scala#L180

This change uses the `referrerAcquisitionData` object to get referrerUrl + referrerPageviewId, which we already construct for the acquisition event.

Example event in CODE:
Screen Shot 2021-03-17 at 10 07 52

### DIFF
--- a/support-frontend/assets/helpers/page/page.js
+++ b/support-frontend/assets/helpers/page/page.js
@@ -38,11 +38,12 @@ import {
   type CountryGroupId,
   detect as detectCountryGroup,
 } from 'helpers/internationalisation/countryGroup';
-import { trackAbTests, setRefViewId } from 'helpers/tracking/ophan';
+import { trackAbTests, setReferrerDataInLocalStorage } from 'helpers/tracking/ophan';
 import { getSettings } from 'helpers/globals';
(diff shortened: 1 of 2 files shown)


paulbrown1982 (commented):
This looks OK to me but I'm struggling to see how it all pieces together in code that hasn't changed. Have you confirmed that there won't be an unsafe increase in volume of what gets stored in Local Storage?

tomrf1:
> This looks OK to me but I'm struggling to see how it all pieces together in code that hasn't changed. Have you confirmed that there won't be an unsafe increase in volume of what gets stored in Local Storage?

It only sets the `ophan_follow` local storage item if it's not already there. The ophan tracker script then deletes `ophan_follow` as soon as it sends the first event.

TomPretty (approved):
👍

PR #3001, opened 2021-03-29
Amazon Pay for recurring contributions

tomrf1:
Currently we only support Amazon Pay for one-off contributions, in the US only.
This PR adds Amazon Pay for recurring, in US only.
To begin with the querystring must include `amazon-pay-recurring=true`.

## The existing flow
1. User selects Amazon Pay, and the `AmazonPayLoginButtonComponent` is rendered, which displays a login button.
2. User clicks the button and logs in through the Amazon popup.
3. The accessToken becomes available and we render the `AmazonPayWalletComponent`, which loads the wallet widget.
4. User selects a payment method from their wallet and clicks 'Contribute' button.

## Changes for recurring
For recurring the main change is that we display a consent widget below the wallet widget. The user must check the checkbox in this widget before we allow them to click 'Contribute'.

Behind the scenes, the other main difference is that whereas one-off contributions gets an `orderReferenceId` and sends this to payment-api, recurring gets a `billingAgreementId` and sends this to support-workers.

## UX change
Previously we would hide the Contribute button until the user has selected a payment method from their wallet.
Now, with recurring we also have to get them to check the consent checkbox.

Hiding the Contribute button goes against the UX of the rest of the form, where we always display the Contribute button and display errors if they click it.
So I've updated the UX for both one-off and recurring Amazon Pay to always show the Contribute button as soon as the wallet is displayed. Appropriate error messages are displayed above the consent or wallet widgets if user user has not completed the necessary steps.

### Before sign in:
Screen Shot 2021-03-29 at 13 29 56

### After sign in:
Screen Shot 2021-03-29 at 13 30 05

### No payment method error:
Screen Shot 2021-03-29 at 13 24 40

### No consent error:
Screen Shot 2021-03-29 at 13 24 48

### Invalid payment method error:
Screen Shot 2021-03-30 at 13 06 10

### Fatal error:
Screen Shot 2021-03-30 at 13 38 39


support-frontend/assets/helpers/paymentIntegrations/readerRevenueApis.js
```diff
@@ -166,7 +169,8 @@ export type ExistingDirectDebitAuthorisation = {|
 |};
 export type AmazonPayAuthorisation = {|
   paymentMethod: typeof AmazonPay,
-  orderReferenceId: string
+  orderReferenceId?: string,
+  amazonPayBillingAgreementId?: string
```
tomrf1, line 173: It's not ideal to be using this type for both one-off and recurring. This payment authorisation logic could do with a refactor to split one-off and recurring

support-frontend/assets/pages/contributions-landing/contributionsLandingActions.js
```diff
@@ -505,12 +525,14 @@ const onPaymentResult = (paymentResult: Promise<PaymentResult>, paymentAuthorisa
...
             ));
           } else {
-            if (result.error === 'amazon_pay_fatal') {
-              dispatch(setAmazonPayFatalError);
-            }
-            if (result.error === 'amazon_pay_try_other_card') {
-              // Must re-render the wallet widget in order to display amazon's error message
-              dispatch(setAmazonPayWalletWidgetReady(false));
+            if (paymentAuthorisation.paymentMethod === 'AmazonPay') {
+              if (result.error === 'amazon_pay_try_other_card' || result.error === 'amazon_pay_try_again') {
```
tomrf1, line 529: Both payment-api and support-workers now return these codes. Anything else can be considered 'fatal' and so we disable amazon pay on the client.

support-models/src/main/scala/com/gu/support/workers/CheckoutFailureReasons.scala
```diff
@@ -51,6 +54,18 @@ object CheckoutFailureReasons {
     override def asString = "production_test_account_mismatch"
   }
 
+  case object AmazonPayTryAnotherCard extends CheckoutFailureReason {
```
tomrf1, line 57: these failure reasons are copied from the old logic in payment-api (see AmazonPayApiError.scala)

michaelbjacobson (approved):
Great stuff! 🚀

PR #3005, opened 2021-03-31
Supporter product data deletes

rupertbates:
This PR contains all the fixes I've made to the supporter-product-data project to make its output consistent with the output of members-data-api.

Comments on individual changes are inline.

[**Trello Card**](https://trello.com/c/vLFmYSPI/3510-improve-retrieval-of-supporter-information-from-zuora-members-data-api-members-data-api-work-part-4-0)


support-models/src/main/scala/com/gu/support/catalog/Product.scala
```diff
@@ -68,8 +68,8 @@ case object Contribution extends Product {
   lazy val ratePlans: Map[TouchPointEnvironment, List[ProductRatePlan[Contribution.type]]] =
     Map(
       PROD -> List(
-        ProductRatePlan("2c92a0fb4edd70c8014edeaa4eae220a", Monthly, NoFulfilmentOptions, NoProductOptions, "Monthly Contribution"),
-        ProductRatePlan("2c92a0fb4edd70c8014edeaa4e972204", Annual, NoFulfilmentOptions, NoProductOptions, "Annual Contribution"),
+        ProductRatePlan("2c92a0fc5aacfadd015ad24db4ff5e97", Monthly, NoFulfilmentOptions, NoProductOptions, "Monthly Contribution"),
```
rupertbates, line 71: I discovered that we have had the wrong product rate plan ids specified here for a very long time. They are not actually used in the app though - contributions uses values from config. It would be good to standardise on these at some point.

supporter-product-data/cloudformation/create-table-stack.sh
```diff
@@ -0,0 +1,8 @@
+# aws s3 rm s3://supporter-product-data-export-prod/ --recursive --profile membership
```
rupertbates, line 1: This script is a quick way to recreate the Dynamo tables used by this project. That is the quickest way to drop all the data if needed.

supporter-product-data/cloudformation/dynamo-tables.yaml
```diff
@@ -15,66 +15,128 @@ Conditions:
...
       SSESpecification:
         SSEEnabled: true
       TableName: !Sub SupporterProductData-${Stage}
       TimeToLiveSpecification:
-        AttributeName: "termEndDate"
+        AttributeName: "expiryDate"
         Enabled: true
-  OnDemandTable:
+
+  WriteCapacityScalableTarget:
```
rupertbates, line 45: I have made the write capacity of the prod table autoscaling. This is because when we run a full sync it uses a massive amount of write capacity but in normal use when it is only running incremental updates 5 is easily enough.

supporter-product-data/data-extracts/query-csv.sh
```diff
@@ -0,0 +1,5 @@
+ # To run this you will need to install Q http://harelba.github.io/q/
```
rupertbates, line 1: Q is a tool which allows you to run sql queries against a csv file, it's been very useful in this work

supporter-product-data/src/main/scala/com/gu/services/DynamoDBService.scala
```diff
@@ -44,26 +57,43 @@ class DynamoDBService(client: DynamoDbAsyncClient, tableName: String) {
...
-object DynamoDBService{
+object DynamoDBService {
   lazy val CredentialsProvider = AwsCredentialsProviderChain.builder.credentialsProviders(
     ProfileCredentialsProvider.builder.profileName(ProfileName).build,
     InstanceProfileCredentialsProvider.builder.asyncCredentialUpdateEnabled(false).build,
     EnvironmentVariableCredentialsProvider.create()
   ).build
 
-  val dynamoDBClient = DynamoDbAsyncClient.builder
+  val clientOverrideConfiguration = ClientOverrideConfiguration.builder.retryPolicy(
```
rupertbates, line 80: This config makes sure we keep on retrying updates until they succeed, the default retry policy gives up quite quickly so we can end up failing because of a `ProvisionedThroughputExceededException` before the autoscaling kicks in.

supporter-product-data/cloudformation/dynamo-tables.yaml
johnduffell: I don't think the autoscaling quite works or at least not quick enough, Mario G and I were investigating alarms in manage-frontend which turned out to be due to a full sync that was running.  It had scaled up once already but I had to scale again manually to stop the alarms.  Not sure if it's better to document it (somewhere?) as something to amend manually during a full sync.

supporter-product-data/src/main/scala/com/gu/lambdas/UpdateDynamoLambda.scala
```diff
@@ -103,19 +107,34 @@ object UpdateDynamoLambda extends StrictLogging {
...
+    // 'Batch' supporterRatePlanItems up into groups which we can update in Dynamo concurrently.
+    // For this to be safe we need to make sure that no group has more than one item with the same subscription name in it because if it does
+    // then order of execution is important and we can't guarantee this with parallel executions.
+
+    def batchAlreadyHasAnItemForThisSubscription(batch: ListBuffer[(SupporterRatePlanItem, Int)], subscriptionName: String) =
+      batch.exists {
+        case (item, _) => item.subscriptionName == subscriptionName
+      }
+
+    // Using mutable state for performance reasons, it is many times faster than the immutable version
```
johnduffell, line 126: Shouldn't be the case fundamentally, I don't think it's the mutability that makes a difference here.
were you appending to a linked `List` as that would involve copying the whole list so be O(n) each time and O(n^2) overall i.e. very bad
The usual way would be prepend which is O(1) and then reverse the list if necessary at the end which would be O(n) meaning the whole thing would be O(n) i.e. good
 ListBuffer append would be O(log n) which would make it O(n log n) overall which is also generally acceptable.
If you really want to append rather than prepend and reverse, then immutable Vector is also log n for append.

supporter-product-data/src/main/scala/com/gu/services/DynamoDBService.scala
```diff
@@ -20,21 +24,30 @@ class DynamoDBService(client: DynamoDbAsyncClient, tableName: String) {
 
   def writeItem(item: SupporterRatePlanItem)(implicit executionContext: ExecutionContext) = {
     val beneficiaryIdentityId = item.gifteeIdentityId.getOrElse(item.identityId)
+
+    // Dynamo will delete expired subs at the start of the day, whereas the subscription actually lasts until the end of the day
```
johnduffell, line 28: agree with adding a day.  although zuora and dates are always a fun topic - in zuora's mind it doesn't actually last until the end of the day, all dates in zuora are 00:00 midnight at the start of the day.  So they are actually out of date by the time the renewal amendment is created.

supporter-product-data/src/main/scala/com/gu/services/SelectActiveRatePlansQuery.scala
```diff
@@ -5,35 +5,73 @@ import com.gu.model.Stage
...
+                             this data is to provide product information in members-data-api which is keyed off identityId, we're not
+                             interested in any accounts which don't have one
+
+    - is not DS gift or is redeemed DS gift - Until digital subscription gifts are redeemed they should not show up in members-data-api and
+                                              when they do they should show up against the giftee's identity id not the purchasers
+
+   * Order By *
+   Rate plans for the same user id are ordered in ASC order by subscription name and then subscription version
+   this makes sure that cancellations are written after the active version of the sub in case we have both (we shouldn't but you never know!)
+   */
```
johnduffell, line 53: that's helpful documentation! 👏

supporter-product-data/cloudformation/dynamo-tables.yaml
rupertbates: The autoscaling works fine now that I have updated the retry config of the Dynamo client to back off more aggressively and retry for longer. 
I've run multiple full syncs since then and they all completed successfully without errors.

johnduffell (approved):
good work getting through all these niggles! 👏

PR #3006, opened 2021-04-01
User attribute cookies

rupertbates:
The new architecture for members-data-api involves maintaining a DynamoDB table of all currently active contributors and subscribers. This is updated every 5 minutes by a step function but this means that there is potentially a period of up to 5 minutes from the point at which a user takes out a subscription or makes a contribution when members-data-api will be out of date for that user.

To avoid this situation occurring and then being [cached for 24 hours by frontend](https://github.com/guardian/frontend/blob/c1bdb643c2bfd3534575ee5dfd8b94ada6dd9dd6/static/src/javascripts/projects/common/modules/commercial/user-features.js#L69) this PR writes the attribute values related to recurring contributions and digital subs into the relevant cookies and sets a refresh time of 1 day.

[**Trello Card**](https://trello.com/c/vLFmYSPI/3510-improve-retrieval-of-supporter-information-from-zuora-members-data-api-members-data-api-work-part-4-0)


support-frontend/app/cookies/DigitalSubscriptionCookies.scala
```diff
@@ -0,0 +1,36 @@
...
+    def cookie(name: String, value:String) = Cookie(
+      name = name,
+      value = value,
+      secure = true,
+      httpOnly = false,
+      domain = Some(domain.value)
+    )
+    List(
+      cookie(
+        name = "gu_user_features_expiry",
```
johnduffell, line 18: would be nice to have a link to the user feaatures JS file in dotcom, so that people know it can be set in more than one place

johnduffell (approved):
👍 that's a good idea

PR #3007, opened 2021-04-06
Update marketing consent test to target LP + TYP

TomPretty:
Update the marketing consent test to target the LP + TYP. Before it just targeted the TYP, but as the routing between LP -> TYP is done client side, we weren't registering the new ab test participation.

### DIFF
--- a/support-frontend/assets/helpers/abTests/abtestDefinitions.js
+++ b/support-frontend/assets/helpers/abTests/abtestDefinitions.js
@@ -3,8 +3,13 @@ import type { Tests } from './abtest';
 
 // ----- Tests ----- //
 
+// Note: When setting up a test to run on the contributions thank you page
+// you should always target both the landing page *and* the thank you page.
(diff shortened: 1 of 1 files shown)


support-frontend/assets/helpers/abTests/abtestDefinitions.js
```diff
@@ -4,7 +4,6 @@ import type { Tests } from './abtest';
 // ----- Tests ----- //
 
 const allLandingPagesAndThankyouPages = '/contribute|thankyou(/.*)?$';
-const allThankYouPages = '/thankyou(/.*)?$';
```
TomPretty, line 7: Perhaps it's worth having a comment here in the code that mentions that we shouldn't target just the TYP for these tests?

tomrf1: 👍

tomrf1 (approved):

PR #3009, opened 2021-04-07
Update `sendTrackingEvent` component event data: "products" to "products"

GHaberis:

This PR update the "component event" data model we send via Ophan. The model for a component event listed in the [Thrift definition](https://dashboard.ophan.co.uk/docs/thrift/componentevent.html#Struct_ComponentEvent) states there should be a non optional key `products`, wjich could be an empty list. In the component event data we were sending the list with the key `product`, this PR corrects that. Interestingly I've checked the pageview table in the lake and the affected component events have been registered using the current incorrect model.


rupertbates (approved):
Looks good to me.

PR #2794, opened 2020-10-22
Remove the `userDigitalVoucher` flag

rupertbates:
## Why are you doing this?

Now that Subscription Card (AKA digital voucher or iMovo) is live we no longer need a flag for whether to use it or not. This PR removes that flag.

[**Trello Card**](https://trello.com/c/qMSYti7t/3400-post-subs-card-launch-remove-usedigitalvoucher-switch)

### NB. Do not merge until after discussion with SX on Dec 10th

PR #3010, opened 2021-04-07
Remove use digital flag

rupertbates:
## Why are you doing this?

Now that Subscription Card (AKA digital voucher or iMovo) is live we no longer need a flag for whether to use it or not. This PR removes that flag.

[**Trello Card**](https://trello.com/c/qMSYti7t/3400-post-subs-card-launch-remove-usedigitalvoucher-switch)

This PR is an update of https://github.com/guardian/support-frontend/pull/2794 - that had hung around for so long that it was easier to start again than update it.


GHaberis (approved):
Looks good 👍

PR #2994, opened 2021-03-24
Accordion AB test

rupertbates:
## What are you doing in this PR?
This PR sets up the digital subscription 'accordion' AB test. The aim of the test is to see whether having the information accordions open by default will result in an increase in checkout starts.

[**Optimize test**](https://optimize.google.com/optimize/home/#/accounts/1368065808/containers/6827913/experiments/470)
[**Trello Card**](https://trello.com/c/pjBs0cNe/3405-increase-conversion-on-digital-subs-landing-page-by-testing-optimisation-of-accordions)

PR #3014, opened 2021-04-14
Roll back changes made to support accordion AB test

GHaberis:

This PR rolls back the changes put in place in https://github.com/guardian/support-frontend/pull/2994 to support the Accordion AB test. The original aim of the test is to see whether having the information accordions open by default will result in an increase in checkout starts. The results show the original implementation (closed by default) was 61% probability to be best, so we can roll back to that now.


i-hardy (approved):
👍

PR #3016, opened 2021-04-15
remove references to bintray

johnduffell:

This PR removes references to bintray, and also updates the SSM configuration library to use one that is in normal sonatyle.

https://trello.com/c/cyWdxhry/3659-remove-support-frontend-dependency-on-bintray

## Why are you doing this?

Bintray is going away on 1st may https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/


kelvin-chappell (approved):
👍

PR #3017, opened 2021-04-16
Restart marketing consent test

TomPretty:
Restart marketing consent test to start collecting data again. We had updated the name of the test `thankyouPageMarketingConsentTest` -> `thankyouPageMarketingConsentTestR2` in the definition, but not where it was being used (in the `ContributionThankYouMarketingConsent` component). This meant that users we segmented into different variants, but would have all seen the control copy. I think when we upgrade to typescript we should try and catch this sort of error.

### control
Screenshot 2021-04-16 at 11 46 12

### v1
Screenshot 2021-04-16 at 11 46 27

### v2
Screenshot 2021-04-16 at 11 46 38

### DIFF
--- a/support-frontend/assets/helpers/abTests/abtestDefinitions.js
+++ b/support-frontend/assets/helpers/abTests/abtestDefinitions.js
@@ -77,7 +77,7 @@ export const tests: Tests = {
     seed: 12,
   },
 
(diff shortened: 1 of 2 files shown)


tomrf1 (approved):
👍 
ideally the type checker could catch this! Perhaps one for after the TS migration...

PR #3019, opened 2021-04-19
update survey link on TY page

Aracho1:

Update the SurveyMonkey link on TY page and turn on the survey action. TBC if we need to put in an end date for the survey (currently none).

[**Trello Card**](https://trello.com/c/ZxgaUQW3/2610-surveymonkey-update-on-ty-page)

## Screenshots

Screenshot 2021-04-19 at 09 34 38


TomPretty (approved):
Looks good to me! I think we can just go live with this - it will be quick to take the survey down when Paul is happy we've gathered enough data!

tomrf1 (approved):
great!

PR #3020, opened 2021-04-19
Mj local currency test

michaelbjacobson:
### What does this PR do?
Adds the ability for users to view one-off contribution amounts in their local currency, if they are in one of the following countries:
- Sweden (SE)
- Switzerland (CH)
- Norway (NO)
- Denmark (DK)

To manually enable the option to view in local currency, append the following query to the URL: `local-currency-country=se`

### Screenshots
_Contributions landing page_
Screenshot 2021-04-19 at 15 00 17
Screenshot 2021-04-19 at 15 00 29

_Contribution thank you page_
Screenshot 2021-04-19 at 14 59 28

_Contribution email_
Screenshot 2021-04-16 at 15 56 49


support-frontend/assets/components/legal/termsPrivacy/termsPrivacy.jsx
```diff
@@ -33,6 +33,10 @@ function TermsPrivacy(props: PropTypes) {
     AUD: 185,
     CAD: 167,
     NZD: 200,
+    SEK: 0,
```
tomrf1, line 36: 👍 because we're using `fromCountryGroupId` below, these will never be used

TomPretty: should we have a comment to that effect? Or maybe they can be removed?

tomrf1: We have to make flow happy somehow...
The alternative is to leave them out, and handle the case where `regionalAmounts[currency]` doesn't exist.
How about turning `patronText` into a function to handle this, e.g.:
```
const getRegionalAmountString = (): ?string => ...
const patronText = () => {
  const amount = getRegionalAmountString();
  if (amount) {
    return (<div...
  }
  return null;
}
```

tomrf1: This should never happen of course, but it's nice to defend against bugs creeping in later

support-frontend/assets/helpers/internationalisation/localCurrencyCountry.js
```diff
@@ -0,0 +1,61 @@
+// @flow
+
+import type { IsoCountry } from './country';
+import type { IsoCurrency } from './currency';
+import type { ContributionAmounts } from '../contributions';
+
+export type LocalCurrencyCountry = {
```
tomrf1, line 7: 👍

support-frontend/assets/components/legal/termsPrivacy/termsPrivacy.jsx
michaelbjacobson: @TomPretty @tomrf1 @Aracho1 what do you guys think of the change I've made to address this?

support-frontend/assets/pages/contributions-landing/contributionsLanding.jsx
```diff
@@ -115,14 +122,33 @@ const router = (
         <Route
           exact
           path="/:countryId(uk|us|au|eu|int|nz|ca)/contribute/"
-          render={() => contributionsLandingPage()
-          }
+          render={(props) => {
+            const { search } = props.location;
+            const queryParams = new URLSearchParams(search);
+            const localCurrencyCountryId = queryParams.get('local-currency-country') || '';
+            const localCurrencyCountry = localCurrencyCountries[localCurrencyCountryId.toUpperCase()];
```
tomrf1, line 129: Instead of doing this here and passing the `localCurrencyCountry` down through the components, we can set it in `buildInitialState` of `page.js`. This field will only be set once so I think it makes sense to initialise there.
(we already do some querystring stuff there)
That should simplify ContributionForm.jsx as well

support-frontend/assets/helpers/page/commonReducer.js
```diff
@@ -56,6 +60,43 @@ function createCommonReducer(initialState: CommonState): (state?: CommonState, a
...
+      case 'SET_LOCAL_CURRENCY_COUNTRY':
+        return {
+          ...state,
+          internationalisation: {
+            ...state.internationalisation,
+            localCurrencyCountry: action.localCurrencyCountry,
+          },
+        };
+
+      case 'SET_USE_LOCAL_CURRENCY_FLAG':
```
tomrf1, line 83: Can we achieve the same with only this action? It can set `currencyId` and `amounts` based on `action.useLocalCurrency`

tomrf1: discussed IRL:
we need to store whether or not the user has asked to see local currency (for switching between contribution types).
And we need to update amounts + currency based on the contributionType + useLocalCurrency

So we'll keep them as separate actions

michaelbjacobson: 👍🏼

support-frontend/assets/pages/contributions-landing/contributionsLanding.jsx
michaelbjacobson: I've made this refactor now ✅

support-frontend/assets/helpers/page/commonReducer.js
```diff
@@ -56,6 +60,43 @@ function createCommonReducer(initialState: CommonState): (state?: CommonState, a
...
+          ...state,
+          internationalisation: {
+            ...state.internationalisation,
+            currencyId: (action.useLocalCurrency && state.internationalisation.localCurrencyCountry)
+              ? state.internationalisation.localCurrencyCountry.currency
+              : state.internationalisation.defaultCurrency,
+          },
+        };
+
+      case 'SET_LOCAL_CURRENCY_COUNTRY':
```
tomrf1, line 74: this action can now be removed

support-frontend/assets/pages/contributions-landing/contributionsLanding.jsx
tomrf1: nice!

support-frontend/assets/helpers/page/commonReducer.js
michaelbjacobson: Thanks for reminding me @tomrf1, that's done now ✅

tomrf1 (approved):
brilliant work!

michaelbjacobson:
> brilliant work!

Cheers @tomrf1, I'm gonna go ahead and merge this into main now. Shall we try to make a local currency contribution at afternoon check-in?

support-frontend/assets/components/legal/termsPrivacy/termsPrivacy.jsx
michaelbjacobson: I'm marking this as resolved now ✅

michaelbjacobson:

PR #3022, opened 2021-04-21
Fix click tracking on the accordions

GHaberis:

This fixes an issue where `sendTrackingEventsOnClick` was being called on the initial render of an accordion component. This was happening as we'd made use of `useEffect` to call `sendTrackingEventsOnClick` when the `expanded` state of the accordion was updated (via a click event). `useEffect` however is called on initial render too. 

To prevent this I'm making use  of `useRef` as it can be used to store a mutable value (`initialRender`) that persists across re-renders, the value of this is `true` at first, and on initial render we set it to `false`, so any subsequent change to the `expanded` state will result in `sendTrackingEventsOnClick` being called.


lucymonie (approved):
LGTM 👍🏻

PR #3024, opened 2021-04-21
bintray part 2

johnduffell:
Second go at removing bintray - first is here: https://github.com/guardian/support-frontend/pull/3016

The first one failed due to SSM working differently in AWS v2, so the exception needed catching, see @tomrf1 's PR https://github.com/guardian/simple-configuration/pull/22


kelvin-chappell (approved):
👍

PR #3025, opened 2021-04-21
Hide digi subs feedback widget until user has scrolled past the hero

i-hardy:

This ensures the feedback widget on the digital subs landing page won't display until the user has scrolled part way through the product information block, at which point it will gently animate in.

## Why are you doing this?

The feedback widget takes up a lot of above-the-fold space on small mobile devices. It also seems a bit odd to be asking users how helpful they find the page when they've barely seen any of it!

## Screenshots

Apologies for low gif quality!

**Mobile - iPhone SE**
feedback-widget-iphoneSE

**Tablet - Samsung Galaxy Tab S7**
feedback-widget-galaxyTabS7

**Desktop - Firefox/macOS**
feedback-widget-desktop

### DIFF
--- a/support-frontend/assets/pages/digital-subscription-landing/components/feedbackWidget/feedbackWidget.jsx
+++ b/support-frontend/assets/pages/digital-subscription-landing/components/feedbackWidget/feedbackWidget.jsx
@@ -8,18 +8,18 @@ import { ThemeProvider } from 'emotion-theming';
 
 import { Button, LinkButton, buttonBrand } from '@guardian/src-button';
 import { sendTrackingEventsOnClick } from 'helpers/subscriptions';
-import { clickedCss, wrapper, widgetTitle, buttonStyles, feedbackLink, header } from './feedbackWidgetStyles';
+import { clickedCss, hideWidget, wrapper, widgetTitle, buttonStyles, feedbackLink, header } from './feedbackWidgetStyles';
 import { SvgThumbsUp } from './thumbsUp';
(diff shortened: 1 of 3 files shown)


johnduffell:
> The feedback widget takes up a lot of above-the-fold space on small mobile devices.

fair point and this might outweigh what I write below...

> It also seems a bit odd to be asking users how helpful they find the page when they've barely seen any of it!

actually that was part of the point, that people might see the page, and immediately give up because they didn't see what they expected, rather than just people who actually had a go at reading it and then gave up.  This might give more ammo for moving the pricing cards or checkout to the top of the page.

GHaberis (approved):
Looks good to me 👍

i-hardy:
I just think that fundamentally this is way, way too aggressive, especially when we already have a serious problem with conversions on mobile:
Screenshot 2021-04-19 at 16 05 38

It's also going to directly interfere with the upcoming A/B test on the hero by blocking the CTAs- it gets in the way on anything smaller than the desktop breakpoint.

I'm not moving it to the bottom of the page, it just won't appear until you're past the hero. The exact point it appears is adjustable (via the threshold passed to `useHasBeenSeen`), so we can have it appear a little sooner if you think what the gifs are demonstrating is too slow.

johnduffell (approved):
fair enough, sounds like you've got a good grasp of all the issues here, so I will go along with those considerations 👍 
would also be worth looping in Paul L once this goes live, just in case the kind of responses we receive changes (assuming we even had any yet!)

PR #3026, opened 2021-04-22
Update supporter product data from support-workers

rupertbates:

This PR adds a new lambda to support-workers which will update the SupporterProductData DynamoDB table whenever a new subscription is purchased.

The reason why this is needed is that the SupporterProductData table is now the source of truth for what products a reader currently has - this is updated every five minutes by the supporter-product-data step function which runs a stateful Zuora query to pick up new subscriptions but I have noticed that there are often requests to members-data-api relating to new customers **before** this synchronisation has happened. 

In order to give new customers the best possible experience of our products it is important that we make it as seamless as possible and don't, for instance,  tell them that they don't have a subscription when they try to sign into the edition for the first time.

[**Trello Card**](https://trello.com/c/vLFmYSPI/3510-improve-retrieval-of-supporter-information-from-zuora-members-data-api-members-data-api-work-part-4-10)

It looks like there are a lot of changes in this PR but the majority of them are just restructuring the code to allow support-workers to use the same Dynamo service class that the supporter-product-data step function uses to update the data store.

This involved adding a new module `supporter-product-data-dynamo` into the support-modules folder and then adding a dependency on this to the `support-workers` and `supporter-product-data` modules.


support-services/src/main/scala/com/gu/support/catalog/CatalogService.scala
```diff
@@ -14,8 +13,16 @@ object CatalogService {
 
 class CatalogService(val environment: TouchPointEnvironment, jsonProvider: CatalogJsonProvider) extends LazyLogging {
 
-  private[this] def getRatePlanId(billingPeriod: BillingPeriod, fulfilmentOptions: FulfilmentOptions)
-    = getProductRatePlan(environment, billingPeriod, fulfilmentOptions, NoProductOptions).map(_.id).getOrElse("")
+  def getProductRatePlan(
```
rupertbates, line 16: This is to allow us to work out what product rate plan id to add into the SupporterProductData Dynamo table

support-workers/scripts/update-code-cloudformation.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+./build-cloudformation.sh
+
+aws --region eu-west-1 cloudformation validate-template --template-body file://../cloud-formation/target/cfn.yaml --profile membership
+
+aws cloudformation update-stack \
```
rupertbates, line 7: This allows us to update the cloudformation stack on code without having to push to github, wait for a build and then deploy via RiffRaff

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -333,7 +333,7 @@ object DigitalSubscriptionGiftRedemption {
     val zuoraUpdater = new ZuoraUpdater(zuoraService, catalogService)
     for {
       codeValidation <- codeValidator.getStatus(redemptionData.redemptionCode, Some(state.requestId.toString))
-      termDates <- codeValidation match {
+      (subscriptionNumber, termDates) <- codeValidation match {
```
rupertbates, line 336: We need to know the subscriptionNumber for gift redemptions to create the data row so it's been added to `SendThankYouEmailDigitalSubscriptionGiftRedemptionState`

support-workers/src/main/scala/com/gu/support/workers/lambdas/UpdateSupporterProductData.scala
```diff
@@ -0,0 +1,132 @@
...
+import com.gu.support.workers.states.{SendAcquisitionEventState, SendThankYouEmailState}
+import com.gu.support.zuora.api.ReaderType.{Corporate, Direct, Gift}
+import com.gu.supporterdata.model.SupporterRatePlanItem
+import com.gu.supporterdata.services.SupporterDataDynamoService
+import com.gu.threadpools.CustomPool.executionContext
+
+import java.time.LocalDate
+import scala.concurrent.Future
+
+class UpdateSupporterProductData(serviceProvider: ServiceProvider)
```
rupertbates, line 18: This is the new lambda in support-workers

build.sbt
```diff
@@ -113,7 +114,14 @@ lazy val `support-workers` = (project in file("support-workers"))
...
     libraryDependencies ++= commonDependencies
-  ).dependsOn(`support-services`, `support-models` % "test->test;it->test;compile->compile", `support-config`, `support-internationalisation`, `acquisition-event-producer`, `module-bigquery`)
+  ).dependsOn(
+  `support-services`,
+  `support-models` % "test->test;it->test;compile->compile",
+  `support-config`,
+  `support-internationalisation`,
+  `acquisition-event-producer`,
+  `module-bigquery`,
+  `supporter-product-data-dynamo`)
```
johnduffell, line 124:
```suggestion
  ).dependsOn(
    `support-services`,
    `support-models` % "test->test;it->test;compile->compile",
    `support-config`,
    `support-internationalisation`,
    `acquisition-event-producer`,
    `module-bigquery`,
    `supporter-product-data-dynamo`
  )
```

support-services/src/main/scala/com/gu/support/catalog/CatalogService.scala
```diff
@@ -77,10 +86,11 @@ class CatalogService(val environment: TouchPointEnvironment, jsonProvider: Catal
     currency: Currency,
     billingPeriod: BillingPeriod,
     fulfilmentOptions: FulfilmentOptions,
-    productOptions: ProductOptions
+    productOptions: ProductOptions,
+    readerType: ReaderType = Direct
   ): Option[Price] = {
     for {
-      productRatePlan <- product.getProductRatePlan(environment, billingPeriod, fulfilmentOptions, productOptions)
+      productRatePlan <- product.getProductRatePlan(environment, billingPeriod, fulfilmentOptions, productOptions, readerType)
```
johnduffell, line 93: was this a bug before? the missing readerType.

support-models/src/main/scala/com/gu/support/catalog/Product.scala
```diff
@@ -17,12 +17,14 @@ sealed trait Product {
...
-    productOptions: ProductOptions
+    productOptions: ProductOptions,
+    readerType: ReaderType = Direct,
   ): Option[ProductRatePlan[Product]] =
     ratePlans(environment).find(prp =>
       prp.billingPeriod == billingPeriod &&
         prp.fulfilmentOptions == fulfilmentOptions &&
-        prp.productOptions == productOptions
+        prp.productOptions == productOptions &&
+        prp.readerType == readerType
```
johnduffell, line 27: was this a bug before? the missing readerType check.

support-workers/scripts/update-code-cloudformation.sh
johnduffell: sounds good, I'd be tempted to put it alongsidde the cfn, rather than separately just so people actually notice it and use it

support-workers/src/main/scala/com/gu/services/Services.scala
```diff
@@ -43,5 +46,14 @@ class Services(isTestUser: Boolean, val config: Configuration) {
   lazy val catalogService = CatalogService(TouchPointEnvironments.fromStage(stage, isTestUser))
   lazy val giftCodeGenerator = new GiftCodeGeneratorService
   lazy val bigQueryService = new BigQueryService(bigQueryConfigProvider.get(isTestUser))
+  val supporterDynamoStage = Configuration.stage match {
+    case DEV if isTestUser => DynamoStageUAT
+    case DEV => DynamoStageDEV
+    case CODE if isTestUser => DynamoStageUAT
+    case CODE => DynamoStageDEV
+    case PROD if isTestUser => DynamoStageUAT
+    case PROD => DynamoStagePROD
```
johnduffell, line 55:
```suggestion
  val supporterDynamoStage = (Configuration.stage, isTestUser) match {
    case (_, true) => DynamoStageUAT
    case (DEV, false) => DynamoStageDEV
    case (CODE, false) => DynamoStageDEV
    case (PROD, false) => DynamoStagePROD
```

support-workers/src/main/scala/com/gu/support/workers/lambdas/UpdateSupporterProductData.scala
```diff
@@ -0,0 +1,125 @@
...
+            ))
+      case SendThankYouEmailDigitalSubscriptionCorporateRedemptionState(user, product, _, subscriptionNumber) =>
+        catalogService
+          .getProductRatePlan(DigitalPack, product.billingPeriod, NoFulfilmentOptions, NoProductOptions, Corporate)
+          .map(productRatePlan =>
+            supporterRatePlanItem(
+              subscriptionName = subscriptionNumber,
+              identityId = user.id,
+              productRatePlanId = productRatePlan.id,
+              productRatePlanName = "support-workers added Gift Digital Subscription",
```
johnduffell, line 78: actualy corporate.  can we just `.describe` the product instead or toString on the State class simpleName? as the one below is also wrong

support-workers/src/main/scala/com/gu/support/workers/lambdas/UpdateSupporterProductData.scala
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+  def supporterRatePlanItem(subscriptionName: String, identityId: String, productRatePlanId: String, productRatePlanName: String) =
+    SupporterRatePlanItem(
+      subscriptionName = subscriptionName,
+      identityId = identityId,
+      gifteeIdentityId = None,
+      productRatePlanId = productRatePlanId,
+      productRatePlanName = productRatePlanName,
+      termEndDate = LocalDate.now.plusYears(1),
```
johnduffell, line 121: will this record get overwritten automatically when the query happens?

supporter-product-data/src/main/scala/com/gu/model/StageHelpers.scala
```diff
@@ -0,0 +1,18 @@
+package com.gu.model
+
+import com.gu.supporterdata.model.Stage
+import com.gu.supporterdata.model.Stage.{DEV, PROD, UAT}
+
+object StageHelpers {
```
johnduffell, line 6: I'm not a big fan of "Helpers" as it doesn't really add any meaning, I can see it's renamed from `Stage` so if that's necessary would it be we be more specific?
```suggestion
object StageConstructors {
```

johnduffell (approved):
looks good, the only concern I have is about things going wrong with the main job - with a 1 year TTL, it will get confusing if there are issues, as they won't show up for a year.  Since we are only trying to cover a 5 minute period, could we make these records last a day at the most?

support-models/src/main/scala/com/gu/support/catalog/Product.scala
rupertbates: Only a theoretical one as this method was never used in the gifting or corporate code I don't think

support-workers/src/main/scala/com/gu/support/workers/lambdas/UpdateSupporterProductData.scala
rupertbates: Yes

support-workers/src/main/scala/com/gu/support/workers/lambdas/UpdateSupporterProductData.scala
rupertbates: describe is a good idea

support-services/src/main/scala/com/gu/support/catalog/CatalogService.scala
rupertbates: I've removed this method because it was only used by test code. It's now in CatalogServiceSpec

PR #3028, opened 2021-04-26
Use delivery address country to determine currency in GW checkout

rupertbates:
The Guardian weekly checkout currently decides which currency a user will pay in based on their billing address. 

However the current pricing structure uses currency as a proxy for delivery cost - so for instance the price in New Zealand dollars is nearly twice the price in GBP to reflect the higher cost of delivery to New Zealand.

This means that we need to customers to pay in the currency of the country they want their sub delivered to, even if this is not the same as their billing country.

[**Trello Card**](https://trello.com/c/MGzDyodg/3649-gw-should-use-pricing-for-delivery-address-country-not-billing-address)

### DIFF
--- a/support-frontend/assets/helpers/subscriptionsForms/submit.js
+++ b/support-frontend/assets/helpers/subscriptionsForms/submit.js
@@ -290,22 +290,28 @@ function trackSubmitAttempt(
   trackCheckoutSubmitAttempt(componentId, productType, paymentMethod, productType);
 }
 
+function getPricingCountry(product, addresses) {
+  if (product === GuardianWeekly)
+    return addresses.deliveryAddress.country;
+  return addresses.billingAddress.country;
+}
+
 function submitForm(
   dispatch: Dispatch<Action>,
   state: AnyCheckoutState,
 ) {
-  const addresses = getAddresses(state);
-  const billingCountry = addresses.billingAddress.country;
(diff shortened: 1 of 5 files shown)


support-frontend/assets/helpers/subscriptionsForms/submit.js
```diff
@@ -290,22 +290,29 @@ function trackSubmitAttempt(
   trackCheckoutSubmitAttempt(componentId, productType, paymentMethod, productType);
 }
 
+function getPricingCountry(product, addresses) {
+  if (product === GuardianWeekly && addresses.deliveryAddress) {
```
lucymonie, line 294: Does this also work for gifting?

lucymonie (approved):
LGTM - just one question about gifting

support-frontend/assets/helpers/subscriptionsForms/submit.js
rupertbates: Great question, I'll check

rupertbates: Ok gifting is working correctly now

lucymonie: Looks good! 👍🏻

PR #3029, opened 2021-04-26
Mj local currency follow up

michaelbjacobson:
### What does this PR do?
This is a follow up for [this PR](https://github.com/guardian/support-frontend/pull/3020). This improves the way currency amounts are displayed on the post-contribution thank you page, eg. `10 kr` instead of `kr10`.

Also allows local currency min and max values to override defaults, without altering existing functionality.

### DIFF
--- a/support-frontend/assets/helpers/internationalisation/localCurrencyCountry.js
+++ b/support-frontend/assets/helpers/internationalisation/localCurrencyCountry.js
@@ -13,7 +13,7 @@ export type LocalCurrencyCountry = {
 };
 
 export const localCurrencyCountries: {
-  [?string]: LocalCurrencyCountry
+  [?IsoCountry]: LocalCurrencyCountry
 } = {
   SE: {
     countryCode: 'SE',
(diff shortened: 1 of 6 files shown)


support-frontend/assets/helpers/internationalisation/localCurrencyCountry.js
```diff
@@ -24,6 +25,15 @@ export const localCurrencyCountries: {
         defaultAmount: 50,
       },
     },
+    config: {
+      ONE_OFF: {
```
tomrf1, line 29: I'm surprised flow doesn't require monthly/annual...

michaelbjacobson: It didn't complain about it, but it _did_ about the `default` value here. 🤷🏼

tomrf1: It's really a flaw in flow because the following can be unsafe:
`localCurrencyCountry.config[contributionType].min`
There's a thread about it here: https://github.com/facebook/flow/issues/5276

When we migrate to TS this will not be possible! I guess we can live with it until then as we're not planning to do the work for recurring any time soon

tomrf1 (approved):
great!

michaelbjacobson:

PR #3032, opened 2021-04-28
Use currency to determine available payment methods not country

rupertbates:
This PR changes the logic around available payment methods to use currency not billing country. 

This is because for Guardian Weekly it is possible to have  a billing address in the UK but a delivery address overseas, and since #3028  this results in the payment being in a foreign currency which Direct Debit does not support. 

This also fixes the tests to match - so that it doesn't try to use DD where it's not available.

[**Trello Card**](https://trello.com/c/MGzDyodg/3649-gw-should-use-pricing-for-delivery-address-country-not-billing-address-99)


lucymonie (approved):
LGTM 👍🏻

johnduffell (approved):
nice work 👏

PR #3033, opened 2021-04-28
Mj local currency ab test

michaelbjacobson:
### What does this PR do?
Adds an A/B test for local currencies. 

The local currency option can be manually shown by appending the following query to the page URL: `local-currency-country={ISOCountry}`.


support-frontend/assets/helpers/abTests/abtestDefinitions.js
```diff
@@ -58,4 +59,26 @@ export const tests: Tests = {
...
+        id: 'variant',
+      },
+    ],
+    audiences: {
+      DE: {
+        offset: 0,
+        size: 1,
+      },
+    },
+    canRun: () => ['SE', 'CH', 'NO', 'DK'].includes(countryCode),
```
TomPretty, line 78: Can we do this all just with the audiences object? i.e 

```js
   audiences: {
      SE: {
        offset: 0,
        size: 1,
      },
      CH: {
        offset: 0,
        size: 1,
      },
     // ...
    },
```

michaelbjacobson: Unfortunately not @TomPretty, the country value here always defaults to 'DE' for European countries.

michaelbjacobson: Although now I'm wondering if this is because certain cookies aren't set when running locally, this might work in production. I'll check now!

TomPretty: 😨

TomPretty: looks like we get the country here: https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/helpers/internationalisation/country.js#L587 but I'm not sure why that would fallback to `DE` for European countries

michaelbjacobson: @TomPretty I think the `GU_country` and `GU_geo_country` cookies aren't set when running locally, so European countries fall back to Germany here: https://github.com/guardian/support-frontend/blob/a78a62b1839512e1cbde8aa952e0032624e5bf6e/support-frontend/assets/helpers/internationalisation/country.js#L560

michaelbjacobson: I've changed this now because it seems to work fine when in CODE or PROD, because those cookies are set. It does make it fiddly to test locally though, you need to use a VPN and/or manually set/update the cookie values.

support-frontend/assets/helpers/page/page.js
```diff
@@ -82,10 +83,20 @@ function buildInitialState(
...
+
+  const localCurrencyCountryFromQuery = localCurrencyCountries[
     (getQueryParameter('local-currency-country') || '').toUpperCase()
   ];
 
+  if (localCurrencyCountryFromQuery) {
+    localCurrencyCountry = localCurrencyCountryFromQuery;
+  } else if (abParticipations.localCurrencyTest === 'variant') {
+    localCurrencyCountry = localCurrencyCountries[
+      countryFromGeolocation() || 'GB'
```
TomPretty, line 96: Can we use the `countryId` here?

support-frontend/assets/helpers/abTests/abtestDefinitions.js
tomrf1: As discussed in the chat, the handling of country is overcomplicated currently.
If you're not on vpn then fastly's country cookie will not match with the CountryGroup, hence the default of germany.
For testing we can use the `country` querystring param

support-frontend/assets/helpers/page/page.js
```diff
@@ -82,10 +83,20 @@ function buildInitialState(
 ): CommonState {
   const excludedParameters = ['REFPVID', 'INTCMP', 'acquisitionData'];
   const otherQueryParams = getAllQueryParamsWithExclusions(excludedParameters);
-  const localCurrencyCountry = localCurrencyCountries[
+  let localCurrencyCountry;
```
tomrf1, line 86: it would be nice to call a function here instead, e.g.
```
function getLocalCurrencyCountry(
  countryId: IsoCountry,
  abParticipations: Participations
): ?LocalCurrencyCountry {
  const queryParam = getQueryParameter('local-currency-country');
  if (queryParam) {
    return localCurrencyCountries[queryParam.toUpperCase()];
  }

  if (abParticipations.localCurrencyTest === 'variant') {
    return localCurrencyCountries[countryId];
  }

  return null;
}
```

michaelbjacobson: Much better!

support-frontend/assets/helpers/page/page.js
michaelbjacobson: We sure can!

tomrf1 (approved):
great!

michaelbjacobson:

PR #3034, opened 2021-04-29
Distinguish between failures and null operations in UpdateSupporterProductData

rupertbates:
Digital Subscription gift purchases should not be added into the SupporterProductData Dynamo store because they do not give the purchaser any digisub benefits. 

This is correct behaviour, but it is currently being treated as an error by the UpdateSupporterProductData lambda. This PR changes that behaviour so that only actual failures are treated as errors.

[**Trello Card**](https://trello.com/c/vLFmYSPI/3510-improve-retrieval-of-supporter-information-from-zuora-members-data-api-members-data-api-work-part-4-80)

### DIFF
--- a/support-workers/src/main/scala/com/gu/support/workers/lambdas/UpdateSupporterProductData.scala
+++ b/support-workers/src/main/scala/com/gu/support/workers/lambdas/UpdateSupporterProductData.scala
@@ -5,6 +5,7 @@ import com.gu.supporterdata
 import com.gu.services.{ServiceProvider, Services}
 import com.gu.support.catalog._
 import com.gu.support.workers.RequestInfo
+import com.gu.support.workers.lambdas.UpdateSupporterProductData.getSupporterRatePlanItemFromState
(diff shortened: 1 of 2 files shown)


support-workers/src/main/scala/com/gu/support/workers/lambdas/UpdateSupporterProductData.scala
```diff
@@ -38,14 +39,19 @@ class UpdateSupporterProductData(serviceProvider: ServiceProvider)
     supporterDataDynamoService: SupporterDataDynamoService
   ) = {
     getSupporterRatePlanItemFromState(state, catalogService) match {
-      case Some(supporterRatePlanItem) => supporterDataDynamoService.writeItem(supporterRatePlanItem)
-      case None => Future.failed(
+      case Right(Some(supporterRatePlanItem)) => supporterDataDynamoService.writeItem(supporterRatePlanItem).map(_ => ())
+      case Right(None) => Future.successful(())
+      case Left(_) => Future.failed(
```
johnduffell, line 44: I wouldn't go back and add it now, but might have been a nice place to alias Right/None with some vals

johnduffell (approved):
great getting on that so quickly. 👍

PR #3035, opened 2021-04-29
Always take currencyId from delivery address in weeklyCheckoutForm.jsx

rupertbates:
I've noticed a bug in the way that the Guardian Weekly checkout works now that it uses currencyId to work out which payment methods to allow - the currencyId is updated in the state when **either** of the billing or delivery address is changed, whereas we always want to use the currency associated with the delivery address as this is the currency we will ultimately charge the customer in.

[**Trello Card**](https://trello.com/c/MGzDyodg/3649-gw-should-use-pricing-for-delivery-address-country-not-billing-address-99)


GHaberis (approved):
👍

PR #3036, opened 2021-04-29
Add back GA tracking of cta clicks on digi subs page

i-hardy:

This adds back Google Analytics tracking of CTA clicks on the digital subs landing page, which was accidentally removed in #3014


rupertbates (approved):
👍

PR #3039, opened 2021-04-30
Properly fix hero price card overflow

i-hardy:

Actually for real this time fixing the overflow issue with the roundel in the hero CTA A/B test variant.


rupertbates (approved):
👍

PR #3042, opened 2021-04-30
Comment out supporter product data updates for gift redemptions

rupertbates:
The UpdateSupporterProductData lambda is failing for all digital subscription redemptions at the moment because the billing period is incorrect in the state. This is triggering 'support-workers execution failed' alarms on a regular basis which is wasting developers time so I'm commenting out the part of the lambda that is failing until a proper fix can be put in place.


rupertbates:
I've merged this without a review because I want to prevent a load of alarms going off over the long weekend.

PR #3043, opened 2021-05-04
Use the correct giftee rate plan in the supporter product data table

rupertbates:

The `UpdateSupporterProductData` lambda has been failing for Digital Subscription gift redemptions because the billing period passed in from support-frontend is incorrect (it is hard coded to Monthly [here](https://github.com/guardian/support-frontend/blob/cd118e23f281746a2e03d00eb69ee52f8b7f8a58/support-frontend/assets/pages/subscriptions-redemption/api.js#L95) because we don't actually know the real billing period at this point).

We need the correct billing period in order to find the product rate plan id to insert into the SupporterProductData Dynamo DB table so this PR retrieves it in the `CreateZuoraSubscription` step and updates the value in the state it passes on.

[**Trello Card**](https://trello.com/c/DfOjhcq1/3693-fix-updatesupporterdata-lambda-for-digital-subscription-gifts)


build.sbt
```diff
@@ -122,7 +122,7 @@ lazy val `support-workers` = (project in file("support-workers"))
     `acquisition-event-producer`,
     `module-bigquery`,
     `supporter-product-data-dynamo`
-  ).aggregate(`support-services`, `support-models`, `support-config`, `support-internationalisation`, `stripe-intent`, `acquisition-event-producer`)
+  ).aggregate(`support-services`, `support-models`, `support-config`, `support-internationalisation`, `stripe-intent`, `acquisition-event-producer`, `supporter-product-data-dynamo`)
```
johnduffell, line 125: always a tricky one haaving to do dependsOn for the classpath and aggregate so that the tests run

support-workers/src/test/scala/com/gu/support/workers/lambdas/UpdateSupporterProductDataSpec.scala
```diff
@@ -86,1053 +139,13 @@ object UpdateSupporterProductDataSpec {
...
-                    }
-                ]
-            }
-        ],
-        "productFeatures": []
-      }
-    ]
-  }
-  """
+  lazy val catalog: String = Source.fromURL(getClass.getResource("/catalog.json")).mkString
```
johnduffell, line 148: good idea, that file was a bit on the big side for source code inlining

johnduffell (approved):
yep that looks like about what we thought, not ideal to have the complexity but keeps it in the code and out of the data so at least we can improve it later. 👍

PR #3045, opened 2021-05-05
Add a helper for timed release of new landing page copy, and copy for G200 special edition

i-hardy:
In order to both release new copy on the digital subs landing page on Saturday morning, when the G200 special edition is released, and yet also avoid having to do work on a Saturday, this adds a small helper to enable us to set up copy to be displayed only after a certain date and time, and potentially with a pre-set end date.

The helper has been set up with the G200 copy to release on Saturday.

[**Trello Card**](https://trello.com/c/uVEdjCyp)

## Why are you doing this?

This specific copy is intended to be released on Satuday, but having new hero copy that should go out on a certain day is a common requirement and this should save us having to sit around with un-merged PRs that have a specific go-live date.

## Screenshots

**System time set to current date - existing copy**
Screenshot 2021-05-05 at 16 18 03

**System time set to Saturday afternoon - new G200 copy**
Screenshot 2021-05-08 at 16 16 27

**Mobile - Google Pixel 5**
Screenshot 2021-05-05 at 16 27 51

**Tablet - iPad Mini**
Screenshot 2021-05-05 at 16 28 39

### DIFF
--- a/support-frontend/assets/helpers/timeBoundedCopy/timeBoundedCopy.jsx
+++ b/support-frontend/assets/helpers/timeBoundedCopy/timeBoundedCopy.jsx
@@ -1,0 +1,47 @@
+// @flow
+
+import React, { type Node } from 'react';
+
+type LandingPage = 'digitalSubscription' | 'newspaper' | 'guardianWeekly';
+
+type TimeBoundCopy = {|
+  start: string;
+  end?: string;
+  copy: Node;
+|}
+
+type TimedCopyCollection = { [LandingPage]: TimeBoundCopy[] }
+
+const timedCopy: TimedCopyCollection = {
+  digitalSubscription: [
+    {
+      start: '2021-05-08',
+      copy: <>
+        <p>
+          <strong>With two innovative apps and ad-free reading,</strong> a digital subscription gives
+          you the richest experience of Guardian journalism. It also sustains the independent reporting you love.
+        </p>
+        <p>
+          Plus celebrate our 200th birthday with our special edition, We were there, available for a limited time.
+        </p>
+      </>,
+    },
+  ],
+};
+
+export function getTimeboundCopy(
+  page: LandingPage,
+  currentDate: Date,
+  timedCopyData: TimedCopyCollection = timedCopy,
+): Node {
+  const copyListForPage = timedCopyData[page];
+  if (copyListForPage) {
+    const copyToUse = copyListForPage.find((copyForPage) => {
+      const startDate = new Date(copyForPage.start);
+      const endDate = copyForPage.end ? new Date(copyForPage.end) : null;
+      return startDate <= currentDate && (!endDate || endDate >= currentDate);
+    });
+    return copyToUse ? copyToUse.copy : null;
+  }
+  return null;
+}
--- a/support-frontend/assets/helpers/timeBoundedCopy/timeBoundedCopy.jsx
+++ b/support-frontend/assets/helpers/timeBoundedCopy/timeBoundedCopy.jsx
@@ -1,12 +1,13 @@
 // @flow
 
 import React, { type Node } from 'react';
+import { type Option } from 'helpers/types/option';
 
 type LandingPage = 'digitalSubscription' | 'newspaper' | 'guardianWeekly';
 
 type TimeBoundCopy = {|
-  start: string;
-  end?: string;
+  startShowingOn: string;
+  stopShowingOn?: string;
   copy: Node;
 |}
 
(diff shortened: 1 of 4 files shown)


GHaberis:
Nice idea! One suggestion would be a way of forcing the timebound copy to appear via a query string parameter eg. `forceTimeboundCopy=2021-01-01`, this would mean we could preview the copy without waiting on the date?

rupertbates:
This is very similar to the functionality in [flashsale.js](https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/helpers/flashSale.js). We should probably only have one or the other in the codebase

support-frontend/assets/helpers/timeBoundedCopy/timeBoundedCopy.jsx
```diff
@@ -0,0 +1,47 @@
...
+type TimeBoundCopy = {|
+  start: string;
+  end?: string;
+  copy: Node;
+|}
+
+type TimedCopyCollection = { [LandingPage]: TimeBoundCopy[] }
+
+const timedCopy: TimedCopyCollection = {
+  digitalSubscription: [
```
johnduffell, line 16: nice! Just wondering, would it make sense to have a test that fails the build if there is a timed copy that has ended? a bit like the dotcom feature switches - to make sure things are cleaned up.

support-frontend/assets/helpers/timeBoundedCopy/timeBoundedCopy.jsx
```diff
@@ -0,0 +1,47 @@
...
+  page: LandingPage,
+  currentDate: Date,
+  timedCopyData: TimedCopyCollection = timedCopy,
+): Node {
+  const copyListForPage = timedCopyData[page];
+  if (copyListForPage) {
+    const copyToUse = copyListForPage.find((copyForPage) => {
+      const startDate = new Date(copyForPage.start);
+      const endDate = copyForPage.end ? new Date(copyForPage.end) : null;
+      return startDate <= currentDate && (!endDate || endDate >= currentDate);
```
johnduffell, line 42: can we call them something like `firstDay` and `lastDay` as it's not totally clear if endDate is inclusive or exclusive

support-frontend/assets/helpers/timeBoundedCopy/timeBoundedCopy.test.jsx
```diff
@@ -0,0 +1,44 @@
...
+describe('Time bounded copy', () => {
+  const copy = 'Hello this is a test';
+  let testCopy;
+
+  beforeEach(() => {
+    testCopy = {
+      newspaper: [
+        {
+          start: '2021-01-01',
+          end: '2021-12-31',
```
johnduffell, line 12: can we call them something like firstDay and lastDay as it's not totally clear if endDate is inclusive or exclusive

support-frontend/assets/helpers/timeBoundedCopy/timeBoundedCopy.jsx
johnduffell, line 42: might be useful in future to pass in a hash from the url so people can preview the copy?

johnduffell (approved):
what a great idea, much nicer than needing a dev around.  The only tiny risk is it's a production change in the middle of the night, so we need to be sure that the new copy does actually fit the space.

support-frontend/assets/helpers/timeBoundedCopy/timeBoundedCopy.jsx
i-hardy: I think for the moment we can probably do without- I don't think it's a huge issue if we have expired copy hanging around for a bit, and we'd likely just delete anything out of date once we added something new- but it could be useful to add something like that in the future!

support-frontend/assets/helpers/timeBoundedCopy/timeBoundedCopy.jsx
```diff
@@ -0,0 +1,61 @@
...
+  page: LandingPage,
+  currentDate: Date,
+  timedCopyData: TimedCopyCollection = timedCopy,
+): Node {
+  const copyListForPage = timedCopyData[page];
+  if (copyListForPage) {
+    const copyToUse = copyListForPage.find((copyForPage) => {
+      const startShowingOnDate = new Date(copyForPage.startShowingOn);
+      const stopShowingOnDate = copyForPage.stopShowingOn ? new Date(copyForPage.stopShowingOn) : null;
+      return startShowingOnDate <= currentDate && (!stopShowingOnDate || stopShowingOnDate >= currentDate);
```
johnduffell, line 56: this is not quite right, we actually stop showing it the day after the `stopShowingOn` date according to this logic, as it's using `>=`

PR #3046, opened 2021-05-06
Fix countryCode in bigquery for contributions

tomrf1:
Currently all single contributions in the new realtime table (fact_acquisition_events) have a country_code of 'GB'.
This is because it's trying to lookup by country name rather than country code


TomPretty (approved):
👍

PR #3047, opened 2021-05-06
Use the correct stripe account for GW gift

GHaberis:

This PR fixes an issue in the checkout form for Guardian Weekly Gift Subscriptions where we were using the stripe account linked to the billing address country, when in fact we need to use the stripe account associate to the delivery address, even if this is not the same as the billing country (As already implemented for non-gift Guardian Weekly subscriptions).

### DIFF
--- a/support-frontend/assets/pages/weekly-subscription-checkout/components/weeklyCheckoutFormGifting.jsx
+++ b/support-frontend/assets/pages/weekly-subscription-checkout/components/weeklyCheckoutFormGifting.jsx
@@ -351,7 +351,7 @@ function WeeklyCheckoutFormGifting(props: PropTypes) {
             title="Your card details"
           >
             <StripeProviderForCountry
-              country={props.country}
+              country={props.deliveryCountry}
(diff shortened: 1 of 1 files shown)


i-hardy (approved):
LGTM!

support-frontend/assets/pages/weekly-subscription-checkout/components/weeklyCheckoutFormGifting.jsx
```diff
@@ -351,7 +351,7 @@ function WeeklyCheckoutFormGifting(props: PropTypes) {
             title="Your card details"
           >
             <StripeProviderForCountry
-              country={props.country}
+              country={props.deliveryCountry}
```
johnduffell, line 354: actually the backend chooses the account based on the currency (AUD vs Other), not the country https://github.com/guardian/support-frontend/blob/5ddfca4d311ad85a3bfb3d32cba37353f0448246/support-config/src/main/scala/com/gu/support/config/StripeConfig.scala#L18

PR #3050, opened 2021-05-07
update local currency country amounts

michaelbjacobson:
### What does this PR do?
Updates values for local currencies.


paulbrown1982 (approved):
👍

michaelbjacobson:
> 👍

Cheers @paulbrown1982!

tomrf1:
👍 For the analysis we'll have to use data starting from the 8th May. It's best to change the test name when restarting a test like this

PR #3053, opened 2021-05-10
Always show general error message on landing page

tomrf1:
When a user clicks Contribute but there is an issue with the form (e.g. missing email address), we have a generic error message above the Contribute button. But - this message is hidden above mobileMedium, which is most users. This is very confusing UX because the error message next to the specific field is often out of view.

E.g. clicking Contribute with an empty email address field:

### Before
Screen Shot 2021-05-10 at 09 41 59

### After
Screen Shot 2021-05-10 at 09 41 29


TomPretty (approved):
👍

PR #3054, opened 2021-05-10
Update localCurrencyTest name

michaelbjacobson:
Because the local currency values have been updated the test parameters have changed, therefore it makes sense to update the test name and run queries using the updated name _only_.


TomPretty (commented):
I think there might be some typescript migration stuff here! 

Screenshot 2021-05-10 at 10 58 40

michaelbjacobson:
> I think there might be some typescript migration stuff here!
> 
> Screenshot 2021-05-10 at 10 58 40

Yup, haha! I'm fixing it now!

PR #3056, opened 2021-05-10
Update `localCurrencyTest` test name to `localCurrencyTestV2`

michaelbjacobson:
### What does this PR do?
Because the local currency values have been updated the test parameters have changed, therefore it makes sense to update the test name and run queries using the updated name _only_.


tomrf1 (approved):
thanks!

PR #3061, opened 2021-05-12
Upgrade webpack

tomrf1:
because of a dependency flagged by snyk


support-frontend/package.json
```diff
@@ -81,7 +81,7 @@
     "@emotion/react": "^11.1.5",
     "@emotion/serialize": "^0.11.16",
     "@guardian/consent-management-platform": "^6.7.4",
-    "@guardian/pasteup": "^1.0.0-alpha.7",
+    "@guardian/pasteup": "1.0.0-alpha.7",
```
tomrf1, line 84: the next version of this is not commonJS and so it breaks webpack.
We should aim to get rid of pasteup/sass from this codebase soon

PR #3062, opened 2021-05-12
Set up the editorial voice A/B test on the digital subscriptions landing page

i-hardy:

This adds the new 'editorial voice' section to the landing page, including a new Quote component and a reusable BlockLabel component for the black labels used throughout the page.

[**Trello Card**](https://trello.com/c/wJ4tCJuX)

## Why are you doing this?

This is the second A/B test for this quarter, testing the impact of including comment from the editorial team in driving digital subscriptions.

There is also a small accessibility improvement of the existing product block content by reordering the headings.

## Screenshots

**Mobile - Motorola Moto G7 (Android)**
Screenshot 2021-05-17 at 13 07 57

**Tablet - iPad 8th gen**
Screenshot 2021-05-17 at 13 08 55

**Desktop - Chrome on macOS**
Support-the-Guardian-The-Guardian-Digital-Subscription (2)


support-frontend/assets/helpers/abTests/abtestDefinitions.js
```diff
@@ -17,6 +17,7 @@ export const pageUrlRegexes = {
   subscriptions: {
     subsShowcaseAndDigiSubPages: '(/??/subscribe(\\?.*)?$|/??/subscribe/digital(\\?.*)?$)',
     digiSubLandingPages: '(/??/subscribe/digital/gift(\\?.*)?$|/??/subscribe/digital(\\?.*)?$)',
+    digiSubLandingPagesNotGift: '(/??/subscribe/digital(\\?.*)?$)',
```
rupertbates, line 20: We need this regex to include the checkout as well or we don't track the abtest participations correctly in the datalake acquisitions table

lucymonie:
I assume that the duplicate appearance of the feedback widget on the desktop screenshot is a result of the way those screenshots are generated, and not indicative of any issue with the widget?

i-hardy:
Yes, it's a Chrome extension and doesn't work as well as the built-in Firefox screenshot tool.

support-frontend/assets/helpers/abTests/abtestDefinitions.js
i-hardy: I've decided to overhaul the regexes a bit and use actual regex literals rather than strings (mostly because I find them easier to compose and test). The new one should explicitly include the checkout as well.

support-frontend/assets/components/blockLabel/blockLabel.jsx
```diff
@@ -0,0 +1,47 @@
...
+  }
+`;
+
+type PropTypes = {|
+  children: Node;
+  tag?: string;
+  cssOverrides?: string | string[];
+|};
+
+function BlockLabel({ children, tag = 'div', cssOverrides }: PropTypes) {
```
GHaberis, line 33: A question regarding `defaultProps` and the default values of the params. The default prop/param for `tag` both match, so is there a reason why we'd not set the default parameter value for `cssOverrides` to `''` when we state the default prop value is `''`?

i-hardy: It seems to be a Flow/ESLint/something else issue. Setting the `defaultProps` is required for optional props, but if you don't provide a default value when destructuring or write conditional logic around an optional prop (despite it having a non-falsey default value), you get a Flow error, because it can't see the `defaultProps`.

I've had to do this in a few other places in the codebase, I am 99% sure we can get rid of it once we migrate to Typescript.

GHaberis (approved):
All looks good to me 👍

PR #3063, opened 2021-05-13
Update ssri/lodash/y18n dependencies

tomrf1:
These were flagged by snyk.

We tried to do this by updating webpack to v5 (as ssri/y18n are pulled in via webpack). But webpack v5 breaks storybook, so we'll have to wait until storybook catches up.
Our attempt to upgrade webpack is [here](https://github.com/guardian/support-frontend/tree/upgrade-webpack-5)


TomPretty (approved):
👍

PR #3064, opened 2021-05-13
Change PaymentMethod field names to match zuora api 

tomrf1:
The zuora api expects the PaymentMethod field names to be capitalised.
Currently our scala classes are not capitalised, and the Encoder does this transformation.

This PR changes the classes to have names which match the zuora api field names.
This is because it's becoming quite fiddly, e.g. currently the zuora field `IPAddress` would need to be `iPAddress` in the scala.

Tested stripe/direct debit/paypal with recurring contributions in CODE.


support-models/src/main/scala/com/gu/support/workers/PaymentMethods.scala
```diff
@@ -33,71 +34,71 @@ object StripePaymentType {
...
-  paymentGateway: PaymentGateway,
-  `type`: String = "CreditCardReferenceTransaction",
-  stripePaymentType: Option[StripePaymentType]
+  TokenId: String, //Stripe Card id
+  SecondTokenId: String, //Stripe Customer Id
+  CreditCardNumber: String,
+  CreditCardCountry: Option[Country],
+  CreditCardExpirationMonth: Int,
+  CreditCardExpirationYear: Int,
+  CreditCardType: Option[String] /*TODO: strip spaces?*/ ,
```
johnduffell, line 43: always love a good ancient TODO!

support-models/src/test/scala/com/gu/support/encoding/PaymentMethodEncoderSpec.scala
```diff
@@ -26,7 +26,7 @@ class PaymentMethodEncoderSpec extends AsyncFlatSpec with Matchers with LazyLogg
       "StreetNumber" : null,
       "BankTransferType" : "DirectDebitUK",
       "Type" : "BankTransfer",
-      "paymentGateway": "GoCardless"
+      "PaymentGateway": "GoCardless"
```
johnduffell, line 29: was this wrong before?

johnduffell (approved):
nice, thanks, good start on destroying the encoder completely 👍

support-models/src/test/scala/com/gu/support/encoding/PaymentMethodEncoderSpec.scala
tomrf1: Yes! The tests were passing anyway because the `paymentGateway` field in the case classes had default values, so decoding still succeeded!

johnduffell: oh dear! good spot.

PR #3067, opened 2021-05-14
Remove price cards A/B test and refactor hero components

lucymonie:
- Remove the price cards in hero ab test code from `support-frontend`
- Refactor hero code
- Fix an issue with the mobile image for Aus and gift subs so the image is centred
- Remove the yellow roundel on mobile breakpoints for Aus

[**Trello Card**](https://trello.com/c/Tq767EZf/3702-turn-off-pricing-card-a-b-test)

## Why are you doing this?
The price cards in hero A/B test recently completed with a positive result for the variant. Because of this, we're removing the test and resolving the remaining code in favour of the variant.

There are still two scenarios where we continue to have an image in the hero rather than the price cards:
- In Australia
- Gift subs

Because of this, I have refactored the code so there are two versions of the hero - one with price cards, and one with an image.

### DIFF
--- a/support-frontend/assets/helpers/abTests/abtestDefinitions.js
+++ b/support-frontend/assets/helpers/abTests/abtestDefinitions.js
@@ -17,7 +17,6 @@ export const pageUrlRegexes = {
   subscriptions: {
     subsShowcaseAndDigiSubPages: '(/??/subscribe(\\?.*)?$|/??/subscribe/digital(\\?.*)?$)',
     digiSubLandingPages: '(/??/subscribe/digital/gift(\\?.*)?$|/??/subscribe/digital(\\?.*)?$)',
-    digiSubLandingPagesNotAus: '(/(uk|us|ca|eu|nz|int)/subscribe/digital(\\?.*)?$)',
   },
 };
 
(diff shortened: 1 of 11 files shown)


support-frontend/assets/helpers/abTests/abtestDefinitions.js
```diff
@@ -17,7 +17,6 @@ export const pageUrlRegexes = {
   subscriptions: {
     subsShowcaseAndDigiSubPages: '(/??/subscribe(\\?.*)?$|/??/subscribe/digital(\\?.*)?$)',
     digiSubLandingPages: '(/??/subscribe/digital/gift(\\?.*)?$|/??/subscribe/digital(\\?.*)?$)',
-    digiSubLandingPagesNotAus: '(/(uk|us|ca|eu|nz|int)/subscribe/digital(\\?.*)?$)',
```
rupertbates, line 20: I often think it's worth leaving these in for the next time someone wants to use it. Don't mind too much either way though.

support-frontend/assets/pages/digital-subscription-landing/digitalSubscriptionLanding.jsx
```diff
@@ -132,13 +132,16 @@ function LandingPage() {
...
       footer={footer}
     >
-      <DigitalHero
-        orderIsAGift={orderIsAGift}
-        countryGroupId={countryGroupId}
-        promotionCopy={sanitisedPromoCopy}
-        showPriceCards={showPriceCardsInHero}
-        priceList={heroPriceList}
-      />
+      {countryGroupId === AUDCountries || orderIsAGift ?
```
rupertbates, line 135: There's a question about whether we should roll this out to the gift page too I guess?

rupertbates (approved):
Looks good to me!

support-frontend/assets/pages/digital-subscription-landing/digitalSubscriptionLanding.jsx
lucymonie: Yes, that's a good point - I'll check.

support-frontend/assets/helpers/abTests/abtestDefinitions.js
lucymonie: Yes, great point. Thanks Rupert

support-frontend/assets/pages/digital-subscription-landing/digitalSubscriptionLanding.jsx
lucymonie: After a long discussion, we concluded that we would keep the gift version of the page as it is (with an image rather than price cards).

PR #3070, opened 2021-05-18
Move PRB object out of redux

tomrf1:
This is the object returned by the Stripe SDK.
This type of thing doesn't really belong in redux, and we're trying to tidy up/remove redux use in support-frontend.
I had hoped to do this as a custom hook but it's fiddly because it has to happen inside an `Elements` component.


TomPretty (approved):
LGTM 👍 One less thing in redux! 🎉

PR #3071, opened 2021-05-18
Move and remove dependencies

GHaberis:

Tidy up our `package.json` so our `devDependencies` contain modules which are only required during development/build time and our `dependencies` contains modules only used at runtime. 

We also had some dependencies which weren't used anyhwere, which I've removed. I've run all the scripts in our package.json and everything works as expected.

### Moved from `dependencies` to `devDependencies`
- `cssnano`
- `fast-sass-loader`
- `webpack-bundle-analyzer`

### Moved from `devDependencies` to `dependencies` 
- `classnames`

### Removed because not used
- `es-to-primitive`
-  `lodash` (we only use `lodash.debounce` which is installed independently)
-  `react-loadable`
-  `sass-loader`


i-hardy (approved):
👍

PR #3072, opened 2021-05-19
Remove the Redux dependency from all subscriptions landing pages

i-hardy:

This completely removes Redux from all subs landing pages, replacing the static elements of the state such as product price information with top-level props, and using hooks for the mutable state on the paper landing page.

As a consequence the Redux dependency has also been removed from the country switcher component, and various versions of the footer.

## Why are you doing this?

Redux can be complicated and difficult to work with and creates a maintenance burden for these pages. We're also not really making use of its intended purpose of managing complex state changes- the only landing page where the state changes _at all_ is the paper landing page, and it's simply to switch between the Collection and Home Delivery tabs.

Removing Redux also results in a significant reduction in the size of the bundle for every page:
Screenshot 2021-05-19 at 16 06 22


support-frontend/assets/components/countryGroupSwitcher/countryGroupSwitcherContainer.jsx
```diff
@@ -28,16 +27,12 @@ export default function (
...
-      countryGroupIds: listOfCountries,
-      selectedCountryGroup: state.common.internationalisation.countryGroupId,
-      subPath,
-      onCountryGroupSelect,
-    };
-  }
-
-  return connect(mapStateToProps)(CountryGroupSwitcher);
-
+  return () => (
```
GHaberis, line 30: I wonder do we need the `CountryGroupSwitcherContainer` anymore? Could we replace instances where we use it by using `CountryGroupSwitcher` directly instead?

support-frontend/assets/pages/digital-subscription-landing/components/paymentSelection/paymentSelection.jsx
```diff
@@ -1,13 +1,13 @@
...
 
 import React from 'react';
-import { connect } from 'react-redux';
 // components
 import ProductOption, {
   type Product,
 } from 'components/product/productOption';
 import {
-  mapStateToProps,
+  getPaymentOptions,
```
GHaberis, line 9: I definitely prefer that the exports from modules such as `helpers/paymentSelection` are now named in a way which indicates their purpose, rather than a redux api method name, much more readable!

support-frontend/assets/pages/digital-subscription-landing/digitalSubscriptionLandingProps.js
```diff
@@ -0,0 +1,39 @@
...
+  participations: Participations;
+  productPrices: ?ProductPrices;
+  promotionCopy: ?PromotionCopy;
+  orderIsAGift: ?boolean;
+|}
+
+const countryGroupId = detectCountryGroup();
+const countryId = detectCountry();
+
+export const digitalLandingProps: DigitalLandingPropTypes = {
```
GHaberis, line 31: Just a thought regarding testability, would this file be more testable if it exported a function that returned an object rather than exporting an object? Importing this module as a dependency will trigger the side affect of calling `detectCountryGroup`, `detectCountry`, `detectCurrency` etc, wish I believe would make them harder to mock?

GHaberis:
@i-hardy I've been through this and have left a few minor comments, but over all this looks good to me, I found the readability of the updated modules much improved. I think we could open to this to the team regarding merging these changes, nice work!

lucymonie (approved):
I agree with @GHaberis that the readability is much improved, and the reduction in size is great.

PR #3075, opened 2021-05-21
SEPA backend support

tomrf1:
With @TomPretty 

Adds Stripe SEPA as a payment method for recurring contributions.
Stripe requires the client's IP address and user agent, so we pass this through from support-frontend's backend.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -11,7 +11,7 @@ lazy val integrationTestSettings: Seq[Def.Setting[_]] = Defaults.itSettings ++ S
   scalaSource in IntegrationTest := baseDirectory.value / "src" / "test" / "scala",
   javaSource in IntegrationTest := baseDirectory.value / "src" / "test" / "java",
   resourceDirectory in IntegrationTest := baseDirectory.value / "src" / "test" / "resources",
-  testOptions in Test += Tests.Argument(TestFrameworks.ScalaTest, "-l", "com.gu.test.tags.annotations.IntegrationTest"),
+  testOptions in Test += Tests.Argument(TestFrameworks.ScalaTest, "-l", "com.gu.test.tags.annotations.IntegrationTest", "-oI"),
   libraryDependencies += scalatest % "it",
 )
 
--- a/build.sbt
+++ b/build.sbt
@@ -210,7 +210,7 @@ lazy val `acquisition-event-producer` = (project in file("acquisition-event-prod
     publishMavenStyle := true,
     scalacOptions += "-Ymacro-annotations",
     libraryDependencies ++= Seq(
-      "com.gu" %% "ophan-event-model" % "0.0.17" excludeAll ExclusionRule(organization = "com.typesafe.play"),
+      "com.gu" %% "ophan-event-model" % "0.0.23" excludeAll ExclusionRule(organization = "com.typesafe.play"),
       "com.gu" %% "fezziwig" % "1.3",
       "com.typesafe.play" %% "play-json" % "2.7.4",
       "io.circe" %% "circe-core" % "0.12.1",
--- a/scalastyle-config.xml
+++ b/scalastyle-config.xml
@@ -82,11 +82,6 @@
    <parameter name="doubleLineAllowed"><![CDATA[false]]></parameter>
   </parameters>
  </check>
- <check level="error" class="org.scalastyle.scalariform.MethodLengthChecker" enabled="true">
-  <parameters>
-   <parameter name="maxLength"><![CDATA[50]]></parameter>
-  </parameters>
- </check>
  <check level="error" class="org.scalastyle.scalariform.MethodNamesChecker" enabled="true">
   <parameters>
    <parameter name="regex"><![CDATA[^[`]?[a-z][A-Za-z0-9]*[`]?$]]></parameter>
--- a/support-models/src/main/scala/com/gu/support/workers/PaymentMethods.scala
+++ b/support-models/src/main/scala/com/gu/support/workers/PaymentMethods.scala
@@ -5,7 +5,7 @@ import com.gu.i18n.Country
 import com.gu.i18n.Country.UK
 import com.gu.support.encoding.Codec
 import com.gu.support.encoding.Codec.deriveCodec
-import com.gu.support.zuora.api.{DirectDebitGateway, PayPalGateway, PaymentGateway}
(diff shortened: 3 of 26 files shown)


support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
```diff
@@ -59,6 +61,12 @@ object DigitalSubscriptionEmailAttributes {
       default_payment_method: String = "Direct Debit",
     ) extends PaymentFieldsAttributes
 
+    case class SepaAttributes(
+      // TODO
```
paulbrown1982, line 65: Is this TODO still necessary?

support-workers/src/main/scala/com/gu/emailservices/PaperFieldsGenerator.scala
```diff
@@ -105,6 +105,7 @@ class PaperFieldsGenerator(
     case _: CreditCardReferenceTransaction => Future.successful(List("payment_method" -> "Credit/Debit Card"))
     case _: PayPalReferenceTransaction => Future.successful(List("payment_method" -> "PayPal"))
     case _: AmazonPayPaymentMethod => Future.successful(List("payment_method" -> "AmazonPay"))
+    case _: SepaPaymentMethod => Future.successful(List("payment_method" -> "Sepa"))  // TODO
```
paulbrown1982, line 108: Is this `//TODO` still necessary?

support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
tomrf1: Perhaps not. This and the other one in `PaperFieldsGenerator.scala` are subscriptions-specific code - we've updated them to make the pattern matching exhaustive but they will not be used for contributions.
I guess if we do add support for subs, we'll test these paths anyway, so it's safe to remove the TODOs

paulbrown1982 (approved):
👍  LGTM

build.sbt
```diff
@@ -11,7 +11,7 @@ lazy val integrationTestSettings: Seq[Def.Setting[_]] = Defaults.itSettings ++ S
   scalaSource in IntegrationTest := baseDirectory.value / "src" / "test" / "scala",
   javaSource in IntegrationTest := baseDirectory.value / "src" / "test" / "java",
   resourceDirectory in IntegrationTest := baseDirectory.value / "src" / "test" / "resources",
-  testOptions in Test += Tests.Argument(TestFrameworks.ScalaTest, "-l", "com.gu.test.tags.annotations.IntegrationTest"),
+  testOptions in Test += Tests.Argument(TestFrameworks.ScalaTest, "-l", "com.gu.test.tags.annotations.IntegrationTest", "-oI"),
```
johnduffell, line 14: what does `-oI` do?

scalastyle-config.xml
```diff
@@ -82,11 +82,6 @@
    <parameter name="doubleLineAllowed"><![CDATA[false]]></parameter>
   </parameters>
  </check>
- <check level="error" class="org.scalastyle.scalariform.MethodLengthChecker" enabled="true">
-  <parameters>
-   <parameter name="maxLength"><![CDATA[50]]></parameter>
```
johnduffell, line 87: this check can be annoying but it is usually pointing to something that needs looking at

support-models/src/main/scala/com/gu/support/workers/PaymentMethods.scala
```diff
@@ -86,17 +86,32 @@ case class ClonedDirectDebitPaymentMethod(
...
+) extends PaymentMethod
+
 case class AmazonPayPaymentMethod(
   TokenId: String,
   Type: String = "CreditCardReferenceTransaction",  // This is how amazon pay works in zuora - as a credit card
   PaymentGateway: PaymentGateway
 ) extends PaymentMethod
 
 object PaymentMethod {
+  import io.circe.generic.auto._
```
johnduffell, line 109: why auto rather than semiauto?  I'm sure there was a reason we use semiauto everywhere but I can't remember why at the moment.

support-workers/src/main/scala/com/gu/emailservices/ContributionEmailFields.scala
```diff
@@ -53,6 +53,7 @@ class ContributionEmailFields(
       ))
       case _: PayPalReferenceTransaction => Future.successful(List("payment method" -> "PayPal"))
       case _: CreditCardReferenceTransaction => Future.successful(List("payment method" -> "credit / debit card"))
+      case _: SepaPaymentMethod => Future.successful(List("payment method" -> "Sepa"))  // TODO
```
johnduffell, line 56: any more detail on that TODO?  does that just mean the thank you email in braze needs doing?

build.sbt
tomrf1: this makes the failed test output appear at the end. We were struggling to view failed tests because the output is so long!

scalastyle-config.xml
tomrf1: In the case it was complaining about I thought it was a reasonable function

support-models/src/main/scala/com/gu/support/workers/PaymentMethods.scala
tomrf1: I added it for the `GatewayOptionData` type, which the `sepaPaymentMethodCodec` requires.
I've used `io.circe.generic.auto._` a lot in the past!

support-workers/src/main/scala/com/gu/emailservices/ContributionEmailFields.scala
tomrf1: Oops, missed this one! One of the remaining tasks for SEPA is to make the emails work (the email is currently blank), so we'll address this soon

build.sbt
johnduffell: that's handy! normally I end up using testOnly to work out what happened.

support-models/src/main/scala/com/gu/support/workers/PaymentMethods.scala
johnduffell: Maybe we do semiauto in other places just to be more explicit about the implicits(!) so that it doesn't use the wrong/default codec by mistake.  But it's a classic tradeoff of extra code vs magic happens, I was just keeping it consistent.

PR #3077, opened 2021-05-21
Check localstorage available

tomrf1:
to prevent errors if localstorage is not available

redux still has the data:
Screen Shot 2021-05-21 at 16 42 29


TomPretty (approved):
👍

johnduffell:
thanks Tom for fixing this so quickly! 👏 
what point would it have caused the client side error for users? at the end of the checkout or on page load?

tomrf1:
> thanks Tom for fixing this so quickly! 👏
> what point would it have caused the client side error for users? at the end of the checkout or on page load? and all products?

I think it may have prevented the page from loading. I'm not sure if it affected specific pages or browsers, I'm going to have a dig through the api gateway logs

PR #3079, opened 2021-05-24
Github action to measure compressed size changes to JS on PRs

GHaberis:
This PR introduces a new Github action, "Compressed Size", to `support-frontend`.

Compressed Size runs on every pull request and uses the action https://github.com/preactjs/compressed-size-action. This action checkouts the Pull Request's target branch, runs a build and records the file sizes of the generated JS, it then does the same to the branch under review and generates a comparison of the sizes which it adds to the Pull Request.

To enable this I introduced new `yarn` script `build-github-action` to the `package.json` in this PR: https://github.com/guardian/support-frontend/pull/3082


i-hardy (approved):
👍

PR #3085, opened 2021-05-25
Aus pricing card component

lucymonie:
This work is to update the text in the Australian digital subscriptions hero, and switch it over to the price cards version of the hero (formerly it had packshots). The text has been reduced to make it fit better with the price cards layout.

[**Trello Card**](https://trello.com/c/GLhVZjeO/3723-aus-pricing-card-component)

In addition, I have fixed a small layout issue where the DS gift heading was rolling over the line on desktop (screenshot below).

## Why are you doing this?
The Australian version of the page had been left behind when we switched the hero of this page to the new version with the price cards. The reason (as mentioned above) was that the Australian text was a bit long to fit comfortably with the price cards layout. With this new shorter text, it now fits.

This means that all non-gift digital subscriptions pages now have the price card version of the hero.

## Screenshots
Screen Shot 2021-05-25 at 10 53 45

Screen Shot 2021-05-25 at 12 58 36

Screen Shot 2021-05-25 at 10 52 53

## Layout fix for desktop gift version
Screen Shot 2021-05-25 at 10 35 58

Screen Shot 2021-05-25 at 11 20 53

### DIFF
--- a/support-frontend/assets/components/page/heroRoundel.jsx
+++ b/support-frontend/assets/components/page/heroRoundel.jsx
@@ -27,7 +27,7 @@ const heroRoundelStyles = css`
   border-radius: 50%;
   ${headline.xxsmall({ fontWeight: 'bold' })};
 
-  ${from.tablet} {
+  ${from.desktop} {
     width: calc(100% + ${space[6]}px);
     transform: translateY(-50%);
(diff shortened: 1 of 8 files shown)


rupertbates (approved):
Looks good to me.

PR #3096, opened 2021-06-03
add "and then $blah after" copy to paper promo cards

johnduffell:
co authored with @i-hardy in the mob programming

This PR ads some extra copy to the cards for HD and subs card so that people know what they will pay after the promo.
image
and
image

https://trello.com/c/f6MEbgZQ/3706-print-campaign-print-checkout-order-summary-does-not-include-promotion-discount-copy

## Why are you doing this?

We are having a june sale which discounts paper for 12 months, but this has never happened before, so we need to update the page to include the extra copy.


lucymonie (approved):
LGTM

PR #3104, opened 2021-06-07
Remove stripe fraud detection

TomPretty:
Turn off stripe fraud detection to remove reliance on sourcepoint. If a user's add blocker blocks source point (`cdn.privacy-mgmt.com`) we fail to read consents and therefore never load the stripe sdk. We were relying on purpose 1 (`Store and/or access information on a device`) before loading stripe fraud detection which sets a couple of cookies. As a quick fix, we are turning off fraud detection so that we don't need to read consents. 

[**Trello Card**](https://trello.com/c/bopeWCba/2682-ad-blocker-disables-sourcepoint-affecting-stripe)


tomrf1 (approved):
👍 great

PR #3106, opened 2021-06-07
GW monthly product mappings

rupertbates:
We would like to test a monthly billing period for Guardian Weekly rather than Quarterly. New product rate plans have been set up for this in Zuora and this PR adds product rate plan mappings to enable us to use those on the site.

[**Trello Card**](https://trello.com/c/Slzlwxeq/3616-create-monthly-rateplan-in-zuora-for-dev-uat-prod-add-in-the-backend-catalog-rateplan-mappings)


johnduffell (approved):
haven't checked the IDs but they are all different so must be right! 👍

PR #3110, opened 2021-06-09
Fix issue with discount rounding

rupertbates:
We have an issue whereby the amount a user will pay for a subscription after a promotion is applied is slightly different on support-frontend to how it appears in the promo code tool.

It turns out that this is because the discount percentage (after adjustment for the billing period length, see [here for more details](https://github.com/guardian/support-frontend/blob/2978de06f4adb86d99473571775eef7ec24eac46/support-services/src/main/scala/com/gu/support/pricing/PriceSummaryService.scala#L127)) was being rounded to two decimal places whereas the promo code tool and most importantly Zuora support at least 3, which is what is needed to be able to target prices at a penny level of granularity.

[**Trello Card**](https://trello.com/c/Dkmk4eNZ/3752-discount-rounding-issue-on-support-frontend)


lucymonie (approved):
LGTM - I assume the weird-looking discount is to get a specific price?

rupertbates:
> LGTM - I assume the weird-looking discount is to get a specific price?

Yes exactly

PR #3111, opened 2021-06-09
GW Monthly rate plan

rupertbates:
We would like to trial a monthly option for Guardian Weekly as there is a belief that it may reduce churn rates from customers coming to the end of the 6 for 6 period.

[**Trello Card**](https://trello.com/c/jShvc0c9/3455-guardian-weekly-replace-quarterly-with-monthly-test)

This PR introduces a new variable `postIntroductorySixForSixBillingPeriod` in both the Javascript code and the Scala code - this are used to determine:
- what rate plan to show alongside Annual
- what rate plan 6 for 6 customers go on to after the introductory offer period

I have tested monthly and 6 for 6 purchases locally and checked the resulting subscription record in Zuora, the thank you email and how it appears in manage my account are all correct.

## Screenshots

Screen Shot 2021-06-09 at 16 16 01
Screen Shot 2021-06-09 at 16 15 42
Screen Shot 2021-06-09 at 16 15 11
Screen Shot 2021-06-09 at 16 17 27

### DIFF
--- a/support-frontend/assets/helpers/productPrice/billingPeriods.js
+++ b/support-frontend/assets/helpers/productPrice/billingPeriods.js
@@ -12,11 +12,11 @@ const SixForSixBillingPeriod: 'Monthly' = 'Monthly';
 export type BillingPeriod = typeof SixWeekly | typeof Annual | typeof Monthly | typeof Quarterly;
 export type DigitalBillingPeriod = typeof Monthly | typeof Annual;
 export type DigitalGiftBillingPeriod = typeof Annual | typeof Quarterly;
-export type WeeklyBillingPeriod = typeof SixWeekly | typeof SixForSixBillingPeriod | typeof Annual;
+export type WeeklyBillingPeriod = typeof SixWeekly | typeof Monthly | typeof Quarterly | typeof Annual;
 
 export type ContributionBillingPeriod = typeof Monthly | typeof Annual;
 
-const weeklyBillingPeriods = [SixWeekly, SixForSixBillingPeriod, Annual];
+const weeklyBillingPeriods: WeeklyBillingPeriod[] = [SixWeekly, SixForSixBillingPeriod, Annual];
 const weeklyGiftBillingPeriods: WeeklyBillingPeriod[] = [Quarterly, Annual];
 
 function billingPeriodNoun(billingPeriod: BillingPeriod, fixedTerm: boolean = false) {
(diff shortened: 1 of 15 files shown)


support-frontend/assets/helpers/productPrice/billingPeriods.js
```diff
@@ -5,14 +5,18 @@ const Monthly: 'Monthly' = 'Monthly';
 const Quarterly: 'Quarterly' = 'Quarterly';
 const SixWeekly: 'SixWeekly' = 'SixWeekly';
 
+// The value for six for six billing period here must match
+// the value in support-models/src/main/scala/com/gu/support/catalog/Product.scala
+const SixForSixBillingPeriod: 'Monthly' = 'Monthly';
```
i-hardy, line 10: Having the monthly rate called `SixForSixBillingPeriod` feels _really_ confusing to me. I can easily envision getting tripped up on this in the future when making changes to the GW page.

rupertbates: This is just a variable so that we can easily switch between using monthly or quarterly as the billing period for six 4 six, once we decide the long term solution it can be removed.

Would it be less confusing if it was:
```javascript
const SixForSixBillingPeriod = Monthly;
```

i-hardy: Honestly it's the tying of it to six for six that I think would be a source of confusion- if I was looking at another file referring to this variable, I would assume that `SixForSixBillingPeriod`  meant the initial six week period of the offer rather than 'the recurring rate which you get transferred onto when the offer is up'. I know that effectively it _is_ tied to six for six in terms of churn etc, but I don't know that we ought to be reflecting that directly in the code.

I think something like `ShortRecurringGWBillingPeriod` might be a bit clunkier but more effectively expresses 'this is perhaps Quarterly and perhaps Monthly, but is not Annual and not the six week offer period'.

rupertbates: How about `postIntroductorySixForSixBillingPeriod`javascript - it's very verbose but as this is a temporary thing I think it's fine.

i-hardy: That sounds good to me!

i-hardy (approved):
👍 LGTM

PR #3118, opened 2021-06-15
Fix issue where the US quarterly price for Guardian Weekly was displayed to ROW on the subs landing page

i-hardy:

This fixes an issue where the incorrect rate plan was being used to provide the display price for Guardian Weekly on the subscriptions landing page to users in ROW. The rate plan for GW Domestic only includes the US$75 quarterly rate for the USA, and there is a completely separate rate plan for ROW which was not being retrieved.

[**Trello Card**](https://trello.com/c/sWRc0WGy)

### DIFF
--- a/support-frontend/app/controllers/SubscriptionsController.scala
+++ b/support-frontend/app/controllers/SubscriptionsController.scala
@@ -65,7 +65,8 @@ class SubscriptionsController(
     else
       Map.empty
 
-    val weekly = service.getPrices(GuardianWeekly, Nil)(countryGroup)(Domestic)(NoProductOptions)(Quarterly)(countryGroup.currency)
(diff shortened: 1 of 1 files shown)


support-frontend/app/controllers/SubscriptionsController.scala
```diff
@@ -65,7 +65,8 @@ class SubscriptionsController(
     else
       Map.empty
 
-    val weekly = service.getPrices(GuardianWeekly, Nil)(countryGroup)(Domestic)(NoProductOptions)(Quarterly)(countryGroup.currency)
+    val weeklyFulfilmentCountry = if (countryGroup == CountryGroup.RestOfTheWorld) RestOfWorld else Domestic
```
rupertbates, line 68: I would just call this `fulfilmentOptions`

rupertbates (approved):
One suggestion but nice work!

PR #3120, opened 2021-06-16
MJ tweak support workers for sepa emails

michaelbjacobson:
### What does this PR do?
This PR, which should be merged alongside changes to relevant Braze campaign, tweaks the `support-workers` repo to further make it compatible with SEPA direct debit recurring contributions.

### DIFF
--- a/support-workers/src/main/scala/com/gu/emailservices/ContributionEmailFields.scala
+++ b/support-workers/src/main/scala/com/gu/emailservices/ContributionEmailFields.scala
@@ -9,9 +9,9 @@ import org.joda.time.DateTime
 import scala.concurrent.{ExecutionContext, Future}
 
 class ContributionEmailFields(
-     getMandate: String => Future[Option[String]],
-     created: DateTime,
-   ) {
+   getMandate: String => Future[Option[String]],
+   created: DateTime,
+ ) {
 
   def build(contributionProcessedInfo: SendThankYouEmailContributionState)(implicit ec: ExecutionContext): Future[EmailFields] = {
     getPaymentFields(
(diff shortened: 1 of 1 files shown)


support-workers/src/main/scala/com/gu/emailservices/ContributionEmailFields.scala
```diff
@@ -9,9 +9,9 @@ import org.joda.time.DateTime
 import scala.concurrent.{ExecutionContext, Future}
 
 class ContributionEmailFields(
-  getMandate: String => Future[Option[String]],
-  created: DateTime,
-) {
+                               getMandate: String => Future[Option[String]],
+                               created: DateTime,
+                             ) {
```
TomPretty, line 14: I guess we don't want this?

michaelbjacobson: Good shout. I just left this alone, but yeah, it's only being used in unit tests now. @tomrf1 are you happy for this to be removed?

TomPretty: Oh I just meant the formatting!

TomPretty: would definitely advocate for removing unused code though!

michaelbjacobson: I was just about to comment again saying that I missed another use of the type! I'm not firing on all cylinders this week! Yeah, I'll fix the inconsistent formatting.

tomrf1: looks like there's still an extra space there!

tomrf1 (approved):
👍

PR #3121, opened 2021-06-17
Put SEPA behind query string param

tomrf1:


michaelbjacobson (approved):
🚀

TomPretty (commented):
so now we need the switch on + query param to see sepa as an option?

tomrf1:
> so now we need the switch on + query param to see sepa as an option?

Correct, the switch needs to be on as well

PR #3122, opened 2021-06-17
Upgrade to play 2.8

johnduffell:

This PR updates play to 2.8 plus associated stuff.

https://trello.com/c/w4MQtvcQ/3763-update-support-frontend-to-play-28

## Why are you doing this?

We need to keep things up to date for security reasons, this means we can update the identity libs more easily, also saves the effort in future when 2.7 becomes unsupported.


project/LibraryVersions.scala
```diff
@@ -1,13 +1,13 @@
 
 object LibraryVersions {
-  val circeVersion = "0.12.1"
+  val circeVersion = "0.14.1"
   val awsClientVersion = "1.11.827"
   val catsVersion = "2.1.1"
   val jacksonVersion = "2.10.5"
-  val jacksonDatabindVersion = "2.10.5.1"
+  val jacksonDatabindVersion = "2.11.4"
```
johnduffell, line 7: now we are pulling in 2.11 anyway, so pin the version with no vulnerability https://snyk.io/vuln/maven:com.fasterxml.jackson.core%3Ajackson-databind

project/LibraryVersions.scala
```diff
@@ -1,13 +1,13 @@
...
   val okhttpVersion = "3.10.0"
   val scalaUriVersion = "2.2.2"
-  val playCirceVersion = "2712.0"
+  val playCirceVersion = "2814.2"
```
johnduffell, line 10: this has to change from "27.." to "28.." as per the play version

support-frontend/conf/routes
```diff
@@ -1,20 +1,20 @@
 # ----- System ----- #
 
-GET /healthcheck                                                   controllers.Application.healthcheck
-GET /error500                                                      lib.ErrorController.error500
-GET /cookies                                                       controllers.DiagnosticsController.cookies
+GET /healthcheck                                                   controllers.Application.healthcheck()
```
johnduffell, line 3: the auto eta expansion is deprecated as of scala 2.13 (and will work differently in scala 3), so there were a load of warnings.
This also caused some strange compiler errors in the reverse routing in play 2.8, so it needed to be fixed now.

support-lambdas/stripe-intent/build.sbt
```diff
@@ -13,6 +13,7 @@ riffRaffManifestBranch := Option(System.getenv("BRANCH_NAME")).getOrElse("unknow
 riffRaffBuildIdentifier := Option(System.getenv("BUILD_NUMBER")).getOrElse("DEV")
 assemblyMergeStrategy in assembly := {
   case x if x.endsWith("module-info.class") => MergeStrategy.discard
+  case str if str.contains("simulacrum") => MergeStrategy.first
```
johnduffell, line 16: there is a simulacrum-scalafix in cats, and the normal simulacrum also.  This causes a conflict due to version differences at the assembly staage, for both this and support-workers jar.

support-models/src/main/scala/com/gu/support/encoding/Codec.scala
```diff
@@ -1,13 +1,13 @@
 package com.gu.support.encoding
 
+import com.gu.support.encoding.JsonHelpers._
+import com.gu.support.encoding.StringExtensions._
 import io.circe.Decoder._
+import io.circe._
 import io.circe.generic.decoding.DerivedDecoder
-import io.circe.generic.encoding.DerivedObjectEncoder
+import io.circe.generic.encoding.DerivedAsObjectEncoder
```
johnduffell, line 8: these changes are from the circe version bump.

build.sbt
```diff
@@ -211,7 +211,7 @@ lazy val `acquisition-event-producer` = (project in file("acquisition-event-prod
     scalacOptions += "-Ymacro-annotations",
     libraryDependencies ++= Seq(
       "com.gu" %% "ophan-event-model" % "0.0.23" excludeAll ExclusionRule(organization = "com.typesafe.play"),
-      "com.gu" %% "fezziwig" % "1.3",
+      "com.gu" %% "fezziwig" % "1.4",
```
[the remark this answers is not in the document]
tomrf1: Another reason to stop sending acquisition events to ophan is that we can stop using fezziwig!

tomrf1 (approved):
great work!

PR #3125, opened 2021-06-21
Acquisitions firehose transformer

TomPretty:
## Why are you doing this?
Moved `acquisitions-firehose-transformer lambda` from https://github.com/guardian/acquisitions-firehose-transformer into the `support-frontend` mono repo. The old repo can now be deprecated. Additionally, we've refactored the lambda to consume json events from the kinesis stream instead of thrift. 


tomrf1:
Note - the schema and the values and slightly different now. This is to keep it in line with the model that we send to bigquery.
Querying historical data in S3 (produced by the old version of this lambda) is not important, so it's worth making the data consistent now.

tomrf1 (approved):
👍 
build is working now.
integration testing in AWS can come later, also a readme

PR #3127, opened 2021-06-21
Turn off PRB design test

TomPretty:
Turn off the stripe PRB design test. There was no significant difference between the control and the variant, but we've decided to roll the variant out to 100%.


tomrf1 (approved):
thanks!

PR #3128, opened 2021-06-21
Remove editorial voice AB test from digital subscriptions landing page

lucymonie:
This is to remove the editorial voice AB test (v2) from the digital subscriptions landing page.

[**Trello Card**](https://trello.com/c/Lj2KF0zH/3766-end-and-remove-ab-test-editorial-voice-v2)

## Why are you doing this?
The test will be complete in one more day, and it appears that the component has not led to improved subscription rates, so we are removing this module.


rupertbates (approved):
Once the build passes 👍

PR #3129, opened 2021-06-22
Disable thrift events being sent to acquisitions stream

TomPretty:
Disable thrift events being sent to acquisitions stream. We are updating this stream to consume json so for now we are stopping sending thrift. We've deleted the `KinesisService` from `acquisitions-event-producer` project as the project is very `thrift`y. We will reintroduce a new json kinesis service (e.g in `support-modules`) in a future PR.


acquisition-event-producer/src/main/scala/com/gu/acquisition/services/DefaultAcquisitionService.scala
```diff
@@ -27,12 +27,10 @@ class DefaultAcquisitionService(config: DefaultAcquisitionServiceConfig)(implici
   override def submit[A: AcquisitionSubmissionBuilder](a: A)(implicit ec: ExecutionContext) = {
     val ov = ophanService.submit(a).value
     val gv = gAService.submit(a).value
-    val kv = kinesisService.submit(a).value
```
tomrf1, line 30: We can go further and actually remove the kinesisService from the `acquisition-event-producer`.
This module is tied to the thrift model. We can later re-implement the kinesis writing logic elsewhere

TomPretty: Okay - I've removed that now!

support-modules/acquisitions-stream/src/main/scala/com/gu/support/acquisitions/AcquisitionsStreamService.scala
```diff
@@ -0,0 +1,56 @@
...
+    val promise = Promise[Either[String, Unit]]()
+
+    val request: PutRecordRequest = {
+      val buffer = ByteBuffer.wrap {
+        acquisition.asJson.noSpaces.getBytes
+      }
+
+      new PutRecordRequest()
+        .withStreamName(config.streamName)
+        .withPartitionKey(acquisition.identityId.getOrElse(acquisition.amount.toString))
```
TomPretty, line 44: `amount` here is also an option, so I guess we shouldn't just use `.toString` on it here?

tomrf1: Good spot.
The partition key is not especially important here (see [here](https://docs.aws.amazon.com/streams/latest/dev/key-concepts.html) for explanation) as the stream is low throughput.
We could just stringify the `eventTimeStamp` here instead?

TomPretty: Yeah sure thing, so we're essentially going to be putting each event into its own shard?

tomrf1: the stream's cloudformation actually defines a `ShardCount` of 1, so whatever we put here will make no difference! If there were more shards then the partition key would determine which one kinesis picks

tomrf1 (approved):

support-modules/acquisitions-stream/src/main/scala/com/gu/support/acquisitions/AcquisitionsStreamService.scala
TomPretty: ahhh gotcha!

tomrf1:
👍 Looks great! We can begin using the new kinesis service in a follow-up PR

PR #3130, opened 2021-06-22
support-lambdas build

tomrf1:
The mono repo has a support-lambdas/ directory, and we're about to add another lambda here.
Currently the 2 lambdas there are built during the support-workers build.
This is a bit confusing, particularly because `stripe-intent` is picked up by sbt with `aggregate`, whereas `it-test-runner` is in the `build-tc` script.

We've decided to create a new teamcity build for these lambdas.
There's now a build-tc script under support-lambdas/.
This does mean another teamcity build will be kicked off for all updates to the mono repo, but at least we can now add all new lambdas to this one build.

Whether support-workers also belongs to this build is an open question, but for now we're keeping it as a separate build.


TomPretty (commented):
LGTM 👍 - but maybe we want another pair of eyes on too?

johnduffell (approved):
nice work, thanks Tom for thinking this through.  Will be interesting to get a review from @rupertbates also as he has a good eye for keeping things understandable

johnduffell:
we should also make it required once we are confident this is working 
image

rupertbates:
I think this is good. 
- I would agree that we should leave support-workers out of it.
- Possibly supporter-product-data should sit within it, I'll have a think about it.
- For future reference - this does not affect the way that projects are deployed. They still show up in RiffRaff in the same way and continuous deployment is triggered in the same way.

rupertbates (commented):
💯 good work

PR #3131, opened 2021-06-22
Restore missing paper subs blue colour

i-hardy:

This restores the `paperSubscriptionsBlue` colour that was accidentally deleted, and is required for the paper subs landing page. It also makes the Webpack settting [output.strictModuleErrorHandling](https://webpack.js.org/configuration/output/#outputstrictmoduleerrorhandling) `true` in production builds, in order to catch issues like this in future.

### DIFF
--- a/support-frontend/webpack.common.js
+++ b/support-frontend/webpack.common.js
@@ -95,7 +95,7 @@ module.exports = (cssFilename, jsFilename, minimizeCss) => ({
     chunkFilename: `webpack/${jsFilename}`,
     filename: `javascripts/${jsFilename}`,
     publicPath: '/assets/',
-    strictModuleErrorHandling: process.env.NODE_ENV === 'production',
+    strictModuleExceptionHandling: process.env.NODE_ENV === 'production',
   },
 
   resolve: {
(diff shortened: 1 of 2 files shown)


support-frontend/webpack.common.js
```diff
@@ -95,6 +95,7 @@ module.exports = (cssFilename, jsFilename, minimizeCss) => ({
     chunkFilename: `webpack/${jsFilename}`,
     filename: `javascripts/${jsFilename}`,
     publicPath: '/assets/',
+    strictModuleExceptionHandling: process.env.NODE_ENV === 'production',
```
rupertbates, line 98: Is this why the build didn't break?

rupertbates (approved):
👍

support-frontend/webpack.common.js
i-hardy: Yes, this should throw an error if an import can't be resolved, so would fail the build rather than just generate a warning.

lucymonie: Yes, I was getting a warning about it... very weird that it didn't break

PR #3133, opened 2021-06-23
java 8 corretto for it test runner

johnduffell:

Make the IT tests use the corretto runtime, to check for issues.

## Why are you doing this?

AWS is switching over from July 19th, so we should test in advance.
https://aws.amazon.com/blogs/compute/announcing-migration-of-the-java-8-runtime-in-aws-lambda-to-amazon-corretto/

The IT test runner is about the least consenquential thing we have, however it does use a lot of our code and exercise it, so it should be a good test.

We would need to follow up with support-workers, etc.


paulbrown1982 (approved):
👍

PR #3134, opened 2021-06-23
update contributions FAQs link

johnduffell:

Point to the help centre instead of the outdated FAQs page

## Why are you doing this?

> It was originally created as something to include in our auto-response, but now our auto-response isn't working anyway, plus cases are going to TP in the first instance most of the time.

> The page is still getting some traffic, mostly referrals from other Guardian pages, would it make sense to set up a redirect on the URL to send people directly to the help centre and to take this page down?

> The page is here: https://www.theguardian.com/help/users/faq

------

>This page is not being migrated as it mostly contains information that is already elsewhere on the help centre. I would advise sending people to the help centre instead.

----
re this particulaar page 
> Worth updating that to point directly at the help centre, I'd say.

### DIFF
--- a/support-frontend/assets/components/footerCompliant/ContributionsFooter.jsx
+++ b/support-frontend/assets/components/footerCompliant/ContributionsFooter.jsx
@@ -40,7 +40,7 @@ const alignmentStyles = css`
 `;
 
 function ContributionsFooter() {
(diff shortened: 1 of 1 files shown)


jambo1 (approved):
Looks good!

PR #3135, opened 2021-06-23
Set jackson versions for acquisitions-stream module

tomrf1:
flagged by snyk

Also bumped the kinesis library version


TomPretty (approved):
it's a good thing that we have this `jackson` library else we'd have nothing to fill our `snyk` time with!

PR #3136, opened 2021-06-23
Delete unused React components

i-hardy:

This deletes various React components that are now unused, moving some types and styles that need to be retained into the files they were previously just being imported.

[**Trello Card**](https://trello.com/c/VzEGpNwV)

## Why are you doing this?

This is a continuation of the client-side tidy up/audit work.


lucymonie (approved):
LGTM 👍🏻

PR #3137, opened 2021-06-24
Send acquisition events

TomPretty:
co-authored by: @tomrf1 

As part of our [super mode](https://trello.com/c/wYpok20S/2639-super-mode-project-card) we've been migrating our old acquisitions stream. We moved the firehose transformer into this repo and updated it to consume json events (#3125). We also stoped sending the old thrift events (#3129). This PR finishes up that migration by updating `support-workers` and `support-payment-api` to start sending json events to the acquisitions stream.


support-payment-api/src/main/scala/conf/AcquisitionsStremServiceConfigLoader.scala
```diff
@@ -0,0 +1,34 @@
+package conf
```
tomrf1, line 1: typo in the filename!

TomPretty: ahhhh, not even in `Acquisitions` either!

support-payment-api/src/main/scala/backend/BackendError.scala
```diff
@@ -39,6 +42,7 @@ object BackendError {
   final case class AmazonPayApiError(err: AmazonPayError) extends BackendError
   final case class SubscribeWithGoogleDuplicateEventError(error: SubscribeWithGoogleDuplicateInsertEventError) extends BackendError
   final case class Email(error: EmailService.Error) extends BackendError
+  final case class TrackingError(err: Throwable) extends BackendError
   final case class MultipleErrors(errors: List[BackendError]) extends BackendError
```
tomrf1, line 46: `MultipleErrors` is now only being used by SubscribeWithGoogle, which needs to be removed. But we'll do this in a follow up PR

support-payment-api/build.sbt
```diff
@@ -53,6 +53,8 @@ libraryDependencies ++= Seq(
   "org.scala-lang.modules" %% "scala-xml" % "1.2.0"
 )
 
+dependencyOverrides += "com.fasterxml.jackson.core" % "jackson-databind" % jacksonDatabindVersion
```
tomrf1, line 56: not sure why this has happened in this branch, but without this we get the following error:
`com.fasterxml.jackson.databind.JsonMappingException: Scala module 2.11.4 requires Jackson Databind version >= 2.11.0 and < 2.12.0`

support-payment-api/src/main/scala/backend/AmazonPayBackend.scala
```diff
@@ -113,34 +114,16 @@ class AmazonPayBackend(
     }
   }
 
-  private def trackContribution(payment: AuthorizationDetails, acquisitionData: AmazonPayAcquisition, email: String, identityId: Option[Long], clientBrowserInfo: ClientBrowserInfo): EitherT[Future, BackendError, Unit] = {
+  private def trackContribution(payment: AuthorizationDetails, acquisitionData: AmazonPayAcquisition, email: String, identityId: Option[Long], clientBrowserInfo: ClientBrowserInfo): Future[List[BackendError]] = {
```
tomrf1, line 117: For tracking we only care about the failures, for logging. So we've refactored `trackContribution` in all payment backends to return just a List of errors

support-payment-api/src/main/scala/backend/PaymentBackend.scala
```diff
@@ -0,0 +1,45 @@
...
+  val databaseService: ContributionsStoreService
+
+  private def insertContributionDataIntoDatabase(contributionData: ContributionData)(implicit pool: DefaultThreadPool): EitherT[Future, BackendError, Unit] = {
+    // log so that if something goes wrong we can reconstruct the missing data from the logs
+    logger.info(s"about to insert contribution into database: $contributionData")
+    databaseService.insertContributionData(contributionData)
+      .leftMap(BackendError.fromDatabaseError)
+  }
+
+  def track[A : AcquisitionSubmissionBuilder](legacyAcquisition: A, acquisition: AcquisitionDataRow, contributionData: ContributionData)(implicit pool: DefaultThreadPool): Future[List[BackendError]] = {
```
tomrf1, line 28: previously each payment backend had its own implementation for sending these events

support-workers/src/main/scala/com/gu/support/workers/lambdas/SendAcquisitionEvent.scala
```diff
@@ -53,11 +53,19 @@ class SendAcquisitionEvent(serviceProvider: ServiceProvider = ServiceProvider)
...
+    val streamFuture = services.acquisitionsStreamService.putAcquisition(acquisition)
+    val biqQueryFuture = services.bigQueryService.tableInsertRow(acquisition)
+
+    val result = for {
+      streamResult <- streamFuture
+      bigQueryResult <- biqQueryFuture
+    } yield ()
+
+    result.value.map {
       case Left(errorMessage) => throw new RetryUnlimited(errorMessage)
```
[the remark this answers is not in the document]
tomrf1: I think we should change this to `RetryNone`. An alarm should go off if it fails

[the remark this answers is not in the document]
johnduffell: ooh that's really bad, I don't think we thought about that one at the time. I agree with retry None.

support-payment-api/build.sbt
johnduffell: that is very annoying - there seems to be a one way restriction and so noawadays you have to force an earlier version when play 2.8 and AWS SDK 2 are involved.

rupertbates (approved):
Good work

support-workers/src/main/scala/com/gu/support/workers/lambdas/SendAcquisitionEvent.scala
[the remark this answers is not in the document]
tomrf1: Actually - I've just noticed this morning that the `SendAcquisitionEvent` failed because of an issue on BigQuery's side, and succeeded on the retry.
So it seems we do want to retry, but I'm not sure how to do this besides splitting BigQuery and kinesis into separate lambdas, which would be a shame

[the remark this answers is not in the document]
johnduffell: another option is to just retry in scala if we think the error would be simple and transient
The benefit of step functions retries is the lambda can terminate between each ones so it can go until zuora comes back up rather than just 15 minutes, and also aany fatal errors that cause the jvm to quit are alaso covered.  But that may not apply for this situation?

support-workers/src/main/scala/com/gu/config/Configuration.scala
```diff
@@ -44,4 +44,6 @@ case class Configuration(config: Config) {
   val bigQueryConfigProvider = new BigQueryConfigProvider(config, stage)
 
   val contributionThanksQueueName = config.getString("email.thankYou.queueName")
+
+  val kinesisStreamName = config.getString("kinesis.streamName")
```
johnduffell, line 48: which kinesis stream? does this need a better name?

support-modules/acquisitions-stream/src/main/scala/com/gu/support/acquisitions/AcquisitionsStreamService.scala
```diff
@@ -21,16 +21,16 @@ sealed trait AcquisitionsStreamServiceConfig {
   val streamName: String
 }
 
-case class Ec2OrLocalConfig(val streamName: String, val credentialsProvider: AWSCredentialsProviderChain) extends AcquisitionsStreamServiceConfig
-case class LambdaConfig(val streamName: String) extends AcquisitionsStreamServiceConfig
+case class AcquisitionsStreamEc2OrLocalConfig(val streamName: String, val credentialsProvider: AWSCredentialsProviderChain) extends AcquisitionsStreamServiceConfig
```
johnduffell, line 24: if module-acquisitions-stream is a new thing, it would be good to use AWS SDK v2 if possible, as we should be moving over aanyway (for snyk reasons and also it's more maaintaained by amazon)

johnduffell (approved):
looks good, although I know less about payment API but nothing sticks out as being confusing! 👍 
I also updted the PR title a bit but I might have missed the point so please update it further to claarify if necessary

support-workers/src/main/scala/com/gu/support/workers/lambdas/SendAcquisitionEvent.scala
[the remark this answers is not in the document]
tomrf1: That's a good idea @johnduffell 
Let's implement retry logic in the scala for both bigquery and kinesis, and make this lambda return RetryNone when it fails.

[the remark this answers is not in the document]
tomrf1: Also, ideally that retry logic can be shared with payment-api

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/utils/retry.scala
```diff
@@ -0,0 +1,17 @@
+package com.gu.support.acquisitions.utils
+
+import scala.concurrent.{Future, ExecutionContext}
+import cats.data.EitherT
+import cats.instances.future._
+
+object Retry {
+  def apply[T](max: Int)(f: => EitherT[Future, String, Unit])(implicit ec: ExecutionContext): EitherT[Future, String, Unit] = {
```
tomrf1, line 8: unused type parameter here, was it intended for the result type?
It's not important as the two uses of this function have the same type, but you could make the Left and Right types generic

TomPretty: oh yeah! must have missed that - I'll make the `Left` and `Right` types generic as that feels quite nice!

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/utils/Retry.scala
```diff
@@ -0,0 +1,17 @@
...
+
+import scala.concurrent.{Future, ExecutionContext}
+import cats.data.EitherT
+import cats.instances.future._
+
+object Retry {
+  def apply[A, B](max: Int)(f: => EitherT[Future, A, B])(implicit ec: ExecutionContext): EitherT[Future, A, B] = {
+    def attempt(triesLeft: Int): EitherT[Future, A, B] = {
+      f.leftFlatMap(error => {
+        if (triesLeft > 0) attempt(triesLeft - 1) else EitherT.fromEither(Left(error))
```
johnduffell, line 11: would it be possible to (also/instead?) return the first error, or maybe a list of the errors, so that if the first attempt locks us out, we can see what went wrong that time?  Sometimes the first can time out, then the later ones can fail with "duplicate record", so you need to know both to be sure.

johnduffell: you might be able to do some kind of lazy list to avoid complex recursion, along these lines
```scala
val tries = Stream.from(1 to 3).map(f).takeWhile(_.isLeft).toList
tries.find(_.isRight).getOrElse(Left(tries.map(_.left.get)))
```

support-payment-api/src/main/scala/services/CloudWatchService.scala
```diff
@@ -66,6 +67,16 @@ class CloudWatchService(cloudWatchAsyncClient: AmazonCloudWatchAsync, environmen
...
   }
 
+  def recordPostPaymentTasksErrors(paymentProvider: PaymentProvider, errors: List[BackendError]): Unit = {
+    errors match {
+      case Nil =>
+      case errors => recordPostPaymentTasksError(
+        paymentProvider,
+        s"unable to track contribution due to error: ${errors.mkString(" & ")}"
+      )
+    }
```
johnduffell, line 77:
```suggestion
    if (errors.nonEmpty)
      recordPostPaymentTasksError(
        paymentProvider,
        s"unable to track contribution due to error: ${errors.mkString(" & ")}"
      )
```

support-workers/src/main/scala/com/gu/config/Configuration.scala
johnduffell:
```suggestion
  val acquisitionKinesisStreamName = config.getString("acquisition.kinesisStreamName")
```

johnduffell (approved):
as before nice work and it looks good based on my lack of knowledge of payment API!  Just a couple of comments though. 👍

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/utils/Retry.scala
TomPretty: Yeah I think that would be a good idea. I had a quick go at your suggested code, the only issue is with `takeWhile(_.isLeft)` - as we're using an `EitherT` with a `Future`, the type of `_.isLeft` is `Future[Boolean]`, I wasn't quite sure how to get around that?

Just for comparison, I had a go at the recursive solution: 

```scala
    def attempt(triesLeft: Int, errors: List[A]): EitherT[Future, List[A], B] = {
      f.leftFlatMap(error => {
        if (triesLeft > 0) attempt(triesLeft - 1, error +: errors) else EitherT.fromEither(Left(error +: errors))
      })
    }

    attempt(max, Nil)
```

I've used an additional `errors` argument to the recursive function to act as an accumulator for the errors. 

I definitely argree though - if there's a simpler non-recursive version we should definitely go for it!

support-payment-api/src/main/scala/services/CloudWatchService.scala
TomPretty: yeah makes sense!

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/utils/Retry.scala
johnduffell: got it, I did write my suggestion without trying it so expected to miss something!
I'll have a quick play with your branch though as I'm intrigued now!  But I don't want to slow things down so happy to go with the recursive one.

TomPretty: No not at all! I'm still a scala newbie so these kind of discussions are incredibly helpful for me!

johnduffell: I did have a quick play and I don't think there's a function to do it.  We would need a Future aware version of `takeWhile` that returns a `Future[List[A]]`
`Future.sequence` is future aware in the right way, but it would take the whole list, so we would need `Future.sequenceWhile` or `Future.sequenceUntil` which would have to be implemented by us (using fold or recursion) so would not be a great benefit!
Quite interesting to think about it!

rupertbates (approved):
Looks good to me, I'd agree that it's probably best to avoid recursion if we can but I don't feel very strongly about it if it proves too difficult.

Overall this is a real improvement so well done!

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/utils/Retry.scala
tomrf1: I've just been having a think about this and it's really hard to implement without recursion!
Why do we want to avoid recursion here? I think Tom's recursive solution is quite neat, and a natural way to reason about this particular problem.

johnduffell: yeah I agree there's no well known abstraction for sequenceWhile, and I agree it is fine, I just think that recursion is very low level, and inductive reasoning isn't totally natural if you haven't used it a lot (same goes for folds).

Since all we are trying to describe is "retry a future 3 times/until it succeeds" the recursion is an implementation detail.  We could implement anything with a fold but we would tend to use map/filter/etc because they are simpler abstractions.

In this case though, I do think the implementation is perfectly clear beyond the recursion!

johnduffell: Not going to affect this PR, but since we're on the topic, I had a quick google for "traverseWhile" and apparently it's really easy in fs2 (which is a whole other topic that I should learn one day!)
https://gitter.im/typelevel/cats?at=5aac3041c3c5f8b90d7e57ac

tomrf1: @johnduffell that fs2 implementation looks nice!

I just had an attempt at a non-recursive version but it's horrible:
```
def apply[A,B](max: Int)(f: => EitherT[Future,A,B])(implicit ec: ExecutionContext): EitherT[Future, List[A], B] =
    // iterates through to max, but only calls f until first success
    (0 to max).foldLeft[EitherT[Future,List[A],B]](EitherT.fromEither(Left(Nil))) { case (acc, _) =>
      acc.leftFlatMap(errors => {
        f.biflatMap[List[A], B](
          err => EitherT.fromEither(Left(err +: errors)),
          value => EitherT.fromEither(Right(value))
        )
      })
    }
```

johnduffell: haha nice work Tom! well in my mind fold is recursion using function parameters rather than in primitives, but I'm still impressed.  Also terminating folds early is always a problem in scala, because even with a LazyList, fold is strict in its arguments.
I'm really tempated to implement traverseUntil now, but it's turning into scala wars on @TomPretty 's PR! 😆

PR #3139, opened 2021-06-24
Add av calculation to firehose

TomPretty:
Add av calculation to acquisitions firehose transfromer. This will give us realtime access to AV for use in our article super mode project.

[**Trello Card**](https://trello.com/c/Vm8BRxYZ/2665-super-mode-realtime-av-calculation)


tomrf1 (approved):
Great!

Note - with this change, any failure (decoding a record or requesting AV) will cause it to throw an exception, which means the entire batch will be retried

PR #3142, opened 2021-06-24
Use webpack's environment variables

GHaberis:

This PR updates how we define environment variables for webpack, introducing consistency by adopting webpack's [environment variables](https://webpack.js.org/guides/environment-variables/). 

I've done this as I'd like to use this approach when updating the build scripts as part of the Typescript migration. The changes made include:

- Removed `env.prod` from the `build-github-action`/`build-prod` script in the `package.json` as this variable isn't used anywhere.
- Replacing `CI_ENV=github` with `--env.ci=github` in the `build-github-action` script in the `package.json`.
- Updating `webpack.prod.js` to accept the `env` parameter as documented in the webpack guides. 
- Setting the default value of `env.ci` in ` `webpack.prod.js` to be "teamcity".
- Moving the `BundleAnalyzerPlugin` out of `webpack.common.js` and into `webpack.prod.js`  as it only ever runs for builds using the `prod` configuration.


GHaberis:
I've tested build on `CODE`/`DEV` and everything is working as expected.

lucymonie (approved):
LGTM 👍🏻

PR #3143, opened 2021-06-25
end local currency test

tomrf1:
pending further exploration.
There's quite a lot to unpick to remove this properly, and we may use it again later, so for now I'll leave it inactive


TomPretty (approved):
Yeah makes sense 👍 we should clean it up if we decide we're never going to use this, but for now just turning it off is the way to go

PR #3144, opened 2021-06-25
Add tests to firehose transformer

TomPretty:
Add tests to firehose transformer.


support-lambdas/acquisitions-firehose-transformer/src/main/scala/com/gu/acquisitionFirehoseTransformer/Lambda.scala
```diff
@@ -32,7 +32,11 @@ case class AcquisitionWithRecordId(val acquisition: AcquisitionDataRow, recordId
 
 object Lambda extends LazyLogging {
 
-  def handler(event: KinesisFirehoseEvent, context: Context): KinesisAnalyticsInputPreprocessingResponse = {
+  def handler(event: KinesisFirehoseEvent): KinesisAnalyticsInputPreprocessingResponse = {
+    processEvent(event, new AnnualisedValueServiceImpl())
+  }
+
+  def processEvent(event: KinesisFirehoseEvent, avService: AnnualisedValueServiceWrapper) = {
```
TomPretty, line 39: We've factored out a `processEvent` method so we can make use of dependency injection in our tests to provide a simple mock av service

support-lambdas/acquisitions-firehose-transformer/src/test/scala/com/gu/acquisitionFirehoseTransformer/LambdaSpec.scala
```diff
@@ -0,0 +1,109 @@
...
+import org.joda.time.DateTime
+import com.gu.i18n.{Country, Currency}
+import com.gu.support.zuora.api.ReaderType
+import java.nio.ByteBuffer
+import collection.JavaConverters._
+import com.gu.acquisitionsValueCalculatorClient.model.AcquisitionModel
+import scala.concurrent.ExecutionContext
+
+
+class LambdaSpec extends AnyFlatSpec with Matchers {
```
tomrf1, line 19: 💯 shall we delete the old test file now? I think the new tests cover serialization just as well

tomrf1 (approved):
👍 great!

support-lambdas/acquisitions-firehose-transformer/src/test/scala/com/gu/acquisitionFirehoseTransformer/LambdaSpec.scala
TomPretty: yeah makes sense - I'll remove that now! 👍

PR #3132, opened 2021-06-22
Add thank you content for events module

lucymonie:
## What are you doing in this PR?
This adds some content about the events and masterclasses tickets to the digital subscriptions thank you page for the events AB test. 

In addition, this turns on the AB test in support-frontend by updating the abTestDefinitions test audience size to 1.

[**Figma**](https://www.figma.com/file/KNabD3Ns9G7SG9rXiBvX6r/Events?node-id=171%3A1126&viewport=1920%2C83%2C0.2595961093902588)
[**Trello Card**](https://trello.com/c/3FAknLiC/3774-ab-test-events-modules-thank-you-page-module)

## Why are you doing this?
To complete the work started in this PR: https://github.com/guardian/support-frontend/pull/3117

## Accessibility test checklist
 - [x] [Tested with screen reader](https://accessibility.gutools.co.uk/testing/web/screen-readers/)
 - [x] [Navigable with keyboard](https://accessibility.gutools.co.uk/testing/web/keyboard-navigation/)
 - [x] [Colour contrast passed](https://accessibility.gutools.co.uk/testing/web/colour-contrast/)

## Screenshots
support thegulocal com_subscribe_digital_checkout_promoCode=DK0NT24WG period=Monthly(wide) (3)

PR #3147, opened 2021-06-28
Update events test to omit US and Aus

lucymonie:
- Removing the US and Australia from the events test
- Updating the Optimize id to a new one, as the original test wasn't working correctly

## Why are you doing this?
This fixes incorrect setup from this PR: https://github.com/guardian/support-frontend/pull/3132


GHaberis (approved):
👍

PR #3149, opened 2021-06-29
Add prompt to check the Optimize setup if the PR is an AB test

lucymonie:
Adding a prompt to the pull request template for AB tests.

## Why are you doing this?
We recently had to restart an Optimize test because the setup wasn't correct. This adds a prompt for reviewers to check that Optimize has been set up correctly.


rupertbates (approved):
👍  good idea

johnduffell:
nice idea, always good to hve a link/reminder to the ab test docs now we're doing so much more of them!

PR #3150, opened 2021-06-29
Extract tracking and A/B test setup from Redux setup and restore them to the subscriptions landing pages

i-hardy:

This puts the tracking and CMP back on the subscriptions landing pages where they were mistakenly removed after removing those pages' Redux dependency. I've done a little bit of reorganisation too and given the 'stateless init' function a hopefully clearer name and moved it into a separate file, which should keep Redux out of the bundle for pages that aren't using it.

## Why are you doing this?

We noticed a drop off in the analytics at around the time the PR removing the Redux dependency was merged; there may be other causes but it seems pretty likely that this issue is at least part of it.


GHaberis (approved):
👍

PR #3151, opened 2021-06-29
Restore initialise step for pages without Redux

GHaberis:

In https://github.com/guardian/support-frontend/pull/3072 we removed Redux from the following landing pages:

- pages/digital-subscription-landing/digitalSubscriptionLanding.jsx 
- pages/paper-subscription-landing/paperSubscriptionLandingPage.jsx 
- pages/showcase/showcase.jsx 
- pages/subscriptions-landing/subscriptionsLanding.jsx 
- pages/weekly-subscription-landing/weeklySubscriptionLanding.jsx 

In doing so we removed `import { init as pageInit } from 'helpers/page/page'` from each of these files as the Redux set-up steps in the function `pageInt` were no longer required. However `pageInt` not only contained Redux set-up steps it also initialised the CMP. By no longer calling `pageInt` the CMP was no longer initialised and as a result none of the tracking dependent on the CMP being initialised was triggered, this includes calls to GA.

To address this I've pulled out the code which initialises the CMP in `helpers/page/page.js` into a function called `cmpInitialisation`, this function will be called from either `init` (used for pages with Redux) or `initPageWithoutRedux` (used for pages without Redux and FKA as `statelessInit`).

I'm now importing `initPageWithoutRedux` in the 5 pages listed above, and I've also renamed the import in the following pages, which previously referred to it as `statelessInit`:

- support-frontend/assets/pages/error/error404.jsx 
- support-frontend/assets/pages/error/error500.jsx 
- support-frontend/assets/pages/paypal-error/payPalError.jsx 

This means the CMP is now initialised on pages with and without Redux.

### DIFF
--- a/support-frontend/assets/helpers/page/page.js
+++ b/support-frontend/assets/helpers/page/page.js
@@ -62,6 +62,21 @@ export type ReduxState<PageState> = {|
 
 // ----- Functions ----- //
 
+function cmpInitialisation(country: IsoCountry) {
+  /**
+   * Dynamically load @guardian/consent-management-platform
+   * on condition we're not server side rendering (ssr) the page.
+   * @guardian/consent-management-platform breaks ssr otherwise.
+   */
+  if (!getGlobal('ssr') && !isPostDeployUser()) {
+    import('@guardian/consent-management-platform').then(({ cmp }) => {
+      cmp.init({
+        country,
+      });
+    });
+  }
+}
+
 // Sets up GA and logging.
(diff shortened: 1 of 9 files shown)


GHaberis:
> I've raised #3150 to tackle this too- we should probably extract these functions out of the existing page.js file otherwise we'll still have the Redux dependencies from that file in the build.

Ah ok, great! I hadn't considered this would lead to Redux being re-bundled into the landing pages which use `initPageWithoutRedux`, but it's a v.good catch, I'm happy to close this and use your PR if you've addressed this.

PR #3153, opened 2021-07-01
Enable digi subs product blocks to be shown in a different order depending on the country

i-hardy:

This is a refactor of the 'product block' component, that makes up the bulk of the digital subscription landing page, in order to enable the different blocks to be shown in a different order depending on the user's country. I've also used the new code to display the Australia-facing content about the Australia Weekend Edition, which means we can remove the Aus-specific product block component and significantly reduce code duplication.

[**Trello Card**](https://trello.com/c/15mJ6uYC)

## Why are you doing this?

Colleagues in Guardian US have raised concerns about users from the USA landing on the page and seeing as the first item a lot of copy about the UK Daily Edition, which may give the impression that the digital subscription is geared at UK users only. This emphasises the premium features of the main Guardian app instead, and also sets up a better framework for potentially varying this content in other regions too.

## Is this an AB test?
- [ ] Yes
- [x] No

## Screenshots

This change doesn't affect styling, so screenshots are from desktop (Firefox/macOS) only to demonstrate the different ordering/content:

**UK**
Screenshot 2021-07-01 at 15-29-44 Support the Guardian The Guardian Digital Subscription

**USA**
Screenshot 2021-07-01 at 15-31-06 Support the Guardian The Guardian Digital Subscription

**Australia**
Screenshot 2021-07-01 at 15-31-43 Support the Guardian The Guardian Digital Subscription


lucymonie (approved):
I think you've done a great job here of refactoring this just enough - it must have been tricky not to get rid of all the sass, so well done for leaving it.

PR #3154, opened 2021-07-02
fix SendAcquisitionEventSpec 

tomrf1:
The `AcquisitionsStreamService` was introduced as a service in a previous PR, and this test is now failing.
I've added a mock implementation that always succeeds


support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/acquisitions-stream/AcquisitionsStreamService.scala
```diff
@@ -25,7 +25,13 @@ sealed trait AcquisitionsStreamServiceConfig {
 case class AcquisitionsStreamEc2OrLocalConfig(val streamName: String, val credentialsProvider: AWSCredentialsProviderChain) extends AcquisitionsStreamServiceConfig
 case class AcquisitionsStreamLambdaConfig(val streamName: String) extends AcquisitionsStreamServiceConfig
 
-class AcquisitionsStreamService(config: AcquisitionsStreamServiceConfig, region: String = "eu-west-1") {
+trait AcquisitionsStreamService {
```
tomrf1, line 28: I made it a trait so that we can define a mock implementation, which I think is preferable to using mockito

TomPretty (approved):
LGTM - thanks for fixing this! 👍

PR #3156, opened 2021-07-05
Enable amount selection from the url

tomrf1:
e.g.
`?selected-amount=15&selected-contribution-type=monthly`

If the selected amount is available in the set of configured amounts then this will be selected. If it's not available, we set it as the 'other' amount.

E.g. an amount is available:
Screen Shot 2021-07-05 at 12 15 31

An amount is not available:
Screen Shot 2021-07-05 at 12 15 42


TomPretty (approved):
Looks great 🚀

PR #3157, opened 2021-07-05
Update payment selection logic to use country rather than currency

lucymonie:
This changes the payment selection logic so we show direct debit as an option if the country is GB rather than if the currency is GBP.

[**Trello Card**](https://trello.com/c/OaMMxRYL/3782-only-enable-direct-debit-where-countryid-gb)

## Why are you doing this?
We've had a few errors where people from the Isle of Man try to pay with direct debit (which is an option presented to them) and their payments can't be processed because they're not in GB.

## Is this an AB test?
- [ ] Yes
- [x] No


rupertbates (approved):
👍 Great, well done!

PR #3159, opened 2021-07-06
Set isRecurringContributor using MDAPI

tomrf1:
The `gu_recurring_contributor` cookie is used for 2 things:
1. on dotcom: to hide support messaging,
2. on the contributions landing page: to prevent people from creating more than one recurring contribution.

The second use is problematic because the cookie is set when support-workers runs. This means if it fails during/after then the user cannot retry unless they delete the cookie.
With this PR, the landing page instead queries MDAPI to find out if the user has a recurring contribution. This does mean that non-signed-in users will not have this check.

Separately, we should delay setting the cookie until support-workers completes or hits the timeout.

### DIFF
--- a/support-frontend/assets/helpers/user/user.js
+++ b/support-frontend/assets/helpers/user/user.js
@@ -130,10 +130,6 @@ const init = (dispatch: Function, actions: UserSetStateActions = defaultUserActi
     dispatch(setPostDeploymentTestUser(true));
   }
 
-  if (getCookie('gu_recurring_contributor') === 'true') {
-    dispatch(setIsRecurringContributor());
-  }
-
   if (getCookie('gu.contributions.contrib-timestamp')) {
     dispatch(setIsReturningContributor(true));
   }
@@ -149,6 +145,17 @@ const init = (dispatch: Function, actions: UserSetStateActions = defaultUserActi
     dispatch(setStateField(window.guardian.user.address4 || window.guardian.geoip.stateCode));
     dispatch(setIsSignedIn(true));
     dispatch(setEmailValidated(getEmailValidatedFromUserCookie(cookie.get('GU_U'))));
+
+    fetch(`${window.guardian.mdapiUrl}/user-attributes/me`, {
+      mode: 'cors',
+      credentials: 'include',
+    }).then(response => response.json())
+      .then((attributes) => {
+        if (attributes.recurringContributionPaymentPlan) {
+          dispatch(setIsRecurringContributor());
+        }
+      });
+
   } else if (userAppearsLoggedIn) { // TODO - remove in another PR as this condition is deprecated
     fetch(routes.oneOffContribAutofill, { credentials: 'include' }).then((response) => {
       if (response.ok) {
--- a/support-frontend/assets/pages/contributions-landing/components/ContributionFormFields.jsx
+++ b/support-frontend/assets/pages/contributions-landing/components/ContributionFormFields.jsx
@@ -41,7 +41,6 @@ type PropTypes = {|
   billingState: StateProvince | null,
   checkoutFormHasBeenSubmitted: boolean,
   isSignedIn: boolean,
-  isRecurringContributor: boolean,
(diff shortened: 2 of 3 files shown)


support-frontend/assets/helpers/abTests/__tests__/abtestTest.js
```diff
@@ -602,10 +602,10 @@ it('targetPage matching', () => {
   const withAcquisitionParams = '/uk/subscribe?INTCMP=header_support_subscribe&acquisitionData=%7B"componentType"%3A"ACQUISITIONS_HEADER"%2C"componentId"%3A"header_support_subscribe"%2C"source"%3A"GUARDIAN_WEB"%2C"referrerPageviewId"%3A"k8heft91k5c3tnnnmwjd"%2C"referrerUrl"%3A"https%3A%2F%2Fwww.theguardian.com%2Fuk"%7D';
   expect(targetPageMatches(withAcquisitionParams, subsShowcaseAndDigiSubPages)).toEqual(true);
   expect(targetPageMatches('/us/subscribe/digital?test=blah', subsShowcaseAndDigiSubPages)).toEqual(true);
- // Test nonGiftLandingAndCheckout regex
```
tomrf1, line 605: linter

support-frontend/assets/pages/contributions-landing/components/ContributionFormFields.jsx
```diff
@@ -41,7 +41,6 @@ type PropTypes = {|
   billingState: StateProvince | null,
   checkoutFormHasBeenSubmitted: boolean,
   isSignedIn: boolean,
-  isRecurringContributor: boolean,
```
tomrf1, line 44: unused

support-frontend/assets/pages/contributions-landing/contributionsLandingActions.js
```diff
@@ -355,7 +355,6 @@ const sendFormSubmitEventForPayPalRecurring = () =>
       flowPrefix: 'npf',
       form: getForm('form--contribution'),
       isSignedIn: state.page.user.isSignedIn,
-      isRecurringContributor: state.page.user.isRecurringContributor,
```
tomrf1, line 358: unused

johnduffell:
I always have to think about whether "dropped" means "drop it in" or like "drop table", perhaps we could standardise on a less ambiguous phrase e.g. set?

support-frontend/assets/helpers/user/user.js
```diff
@@ -149,6 +145,17 @@ const init = (dispatch: Function, actions: UserSetStateActions = defaultUserActi
...
     dispatch(setIsSignedIn(true));
     dispatch(setEmailValidated(getEmailValidatedFromUserCookie(cookie.get('GU_U'))));
+
+    fetch(`${window.guardian.mdapiUrl}/user-attributes/me`, {
+      mode: 'cors',
+      credentials: 'include',
+    }).then(response => response.json())
+      .then((attributes) => {
+        if (attributes.recurringContributionPaymentPlan) {
+          dispatch(setIsRecurringContributor());
```
johnduffell, line 155: the default value is false https://github.com/guardian/support-frontend/blob/5ddfca4d311ad85a3bfb3d32cba37353f0448246/support-frontend/assets/helpers/user/userReducer.js#L42
so that means if we have issues with mdapi or this call stops working, we will still allow recurring contributions
I think is is correct, just wanted to write it down to confirm.

johnduffell (approved):
great stuff 👍

support-frontend/assets/helpers/user/user.js
tomrf1: Yep, there are certainly cases where a user could create 2 recurring contribution, and this PR makes it more likely. But I think it's still pretty unlikely, and enabling users to retry is more important

tomrf1:
> I always have to think about whether "dropped" means "drop it in" or like "drop table", perhaps we could standardise on a less ambiguous phrase e.g. set?

I don't know why we use that term, I've changed it to 'set' in the description!

PR #3161, opened 2021-07-07
Update aus map colours

tomrf1:
just changes colours on the territories and their labels, including hover + selected states.
Also changed the social share icon colour to blue

Screen Shot 2021-07-07 at 15 15 33
Screen Shot 2021-07-07 at 15 15 36
Screen Shot 2021-07-07 at 15 09 19


TomPretty (approved):
👍

PR #3162, opened 2021-07-08
Update `timestamp` format

TomPretty:
Update the format of the `timestamp` field that we save to s3 as the current format doesn't play nice with athena.


tomrf1 (approved):
👍

PR #3163, opened 2021-07-09
Track dateTime in ophan component events

tomrf1:
See ophan change: https://github.com/guardian/ophan/pull/4204

### DIFF
--- a/support-frontend/assets/helpers/tracking/ophan.js
+++ b/support-frontend/assets/helpers/tracking/ophan.js
@@ -86,13 +86,13 @@ type OphanABPayload = {
 const utcNow = () => {
   const now = new Date();
   return Date.UTC(
-    now.getFullYear(),
-    now.getMonth(),
-    now.getDate(),
-    now.getHours(),
-    now.getMinutes(),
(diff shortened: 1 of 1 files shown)


support-frontend/assets/helpers/tracking/ophan.js
```diff
@@ -82,7 +83,23 @@ type OphanABPayload = {
...
+  const now = new Date();
+  return Date.UTC(
+    now.getFullYear(),
+    now.getMonth(),
+    now.getDate(),
+    now.getHours(),
+    now.getMinutes(),
+    now.getSeconds(),
+    now.getMilliseconds()
+  );
```
TomPretty, line 96: I've been searching around and I can't seem to find a better way. I think we want to use the `utc` versions of these methods though? e.g `now.getUTCFullYear`

tomrf1: you're quite right!

tomrf1: `now.getHours()` etc were giving us local time

PR #3145, opened 2021-06-25
Digital subscriptions landing page: comparison table AB test

lucymonie:
## What are you doing in this PR?
Adding a comparison table to the digital subscriptions landing page for AB testing.

[**Trello Card**](https://trello.com/c/glrYCqiO/3771-comparison-table-ab-test-ux-100-ui-100-dev-0)
[**Figma**](https://www.figma.com/file/n2oBySn1jfa55DWb1uC8Ck/DS-LP-Q1---UI-UX-testing?node-id=2101%3A2915)

## Why are you doing this?
During qual testing, the comparison table has been well liked by readers, who say it's a helpful summary of what they get if they pay for a subscription.

## Is this an AB test?
- [x] yes
- [ ] no

The audience size in the test is currently set to zero, so this code could be merged in and the test started later by changing the audience size (and switching on the test in Optimize).

If this is an AB test, PR reviewers should open and check the Optimize test.
[**Optimize link**](https://optimize.google.com/optimize/home/#/accounts/1368065808/containers/6827913/experiments/491)

## Accessibility test checklist
 - [x] [Tested with screen reader](https://accessibility.gutools.co.uk/testing/web/screen-readers/)
 - [x] [Navigable with keyboard](https://accessibility.gutools.co.uk/testing/web/keyboard-navigation/)
 - [x] [Colour contrast passed](https://accessibility.gutools.co.uk/testing/web/colour-contrast/)

## Screenshots
Screen Shot 2021-07-13 at 10 20 27

Screen Shot 2021-07-13 at 10 20 05

PR #3169, opened 2021-07-14
Remove tickets offer from comparison table

lucymonie:
Removing the comparison table line that says that we are offering six free tickets to events, as this was not meant to be in the test version, this was just for the qual tests.

This is to fix an issue from this PR: https://github.com/guardian/support-frontend/pull/3145

## Why are you doing this?
To correct the content of the new comparison table.

## Screenshots
Screen Shot 2021-07-14 at 16 30 42

Screen Shot 2021-07-14 at 16 31 20


i-hardy (approved):
👍

PR #3171, opened 2021-07-15
Swap out marked for snarkdown

marialani:
This pr removes the markdown library `marked` and replaces it with `snarkdown`.

[**Trello Card**](https://trello.com/c/NCxiJcmU/3719-swap-out-marked-for-snarkdown-on-support-frontend)

## Why are you doing this?
We rely on the marked library to parse markdown coming from the promo code tool. However we only support a limited amount of markdown styling, and don't need to include a full markdown parser on every page. Marked is about 10KiB gzipped and is one of the largest single dependencies on our landing pages whilst Snarkdown is considerably smaller.

## Is this an AB test?
- [ ] Yes
- [x] No

### DIFF
--- a/support-frontend/assets/helpers/productPrice/__tests__/promotionsTests.js
+++ b/support-frontend/assets/helpers/productPrice/__tests__/promotionsTests.js
@@ -1,0 +1,28 @@
+import { getPromotionCopy } from '../promotions';
+
+const sanitisablePromotionCopy = {
+  title: 'The Guardian Weekly',
+  description: 'The Guardian Weekly magazine:\n- is a round-up of the [world news opinion and long reads that have shaped the week.](https://www.theguardian.com/about/journalism)\n- with striking photography and insightful companion pieces, all handpicked from The Guardian and The Observer.',
+  roundel: '**Save _25%_ for a year!**',
+};
+
+describe('getPromotionCopy', () => {
+  it('should return promotion copy if it exists', () => {
+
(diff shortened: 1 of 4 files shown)


johnduffell (approved):
nice to get things smaller! I presume you have tested all the variations and they look fine?
👍

GHaberis:
Woah 😮  ~10% reduction in bundle size for all affected pages, this is great to see!

GHaberis (approved):

GHaberis:
> nice to get things smaller! I presume you have tested all the variations and they look fine?
> 👍

If we want to test we're correctly handling multiple variations of markdown coming from the promo code tool then I think unit tests for the function `getPromotionCopy` (exported by `promotions.jsx` and the only function that calls `getSanitisedHtml`) would be the best method to do this.

marialani:
> > nice to get things smaller! I presume you have tested all the variations and they look fine?
> > 👍
> 
> If we want to test we're correctly handling multiple variations of markdown coming from the promo code tool then I think unit tests for the function `getPromotionCopy` (exported by `promotions.jsx` and the only function that calls `getSanitisedHtml`) would be the best method to do this.

I'd tested the variations of markdown being fed through by using some test urls and they all looked fine. Though adding unit tests for this is a very good shout, will update this branch to include some unit tests.

johnduffell (approved):
nice always good to have some lightweight tests for this kind of thing 👍

PR #3173, opened 2021-07-16
Replace font loader

jamie-lynch:

This PR removes the `fontLoaderBundle` and replaces it with a `font.scss` file containing font-face declarations via the following steps:

* Add a new `fonts.scss` file containing font declarations in `support-frontend/assets/stylesheets/skeleton`
* Import this into `support-frontend/assets/stylesheets/skeleton/skeleton.scss` so that it is loaded (once) on the page
* Remove the `non-critical-webfonts.scss` file and import (as Titlepiece is included in the list of declarations)
* Remove the `font-loader.js` file from `support-frontend/assets/rendering/` 
* Remove reference to the `fontLoaderBundle` and wiring throughout the application

## Why are you doing this?

This change is being made as part of an effort to consolidate the usage of fonts across Guardian applications, including where fonts are loaded from and how they are loaded. 

cc @guardian/client-side-infra 

## Is this an AB test?
- [ ] Yes
- [x] No


jamie-lynch:
AFAICT the `support-frontend-mono` build is failing due to `scalastyle` errors but I don't think they're related to any changes we've made. When I run it on main though, `scalastyle` is happy so now I'm confused. If anyone has got time to take a quick look, that would be much appreciated.

support-frontend/assets/stylesheets/gu-sass/typography.scss
```diff
@@ -8,9 +8,9 @@
 // ----- Font Families ----- //
 
 $gu-titlepiece: 'GT Guardian Titlepiece', Georgia, serif !default;
-$gu-headline: 'Guardian Egyptian Web', Georgia, serif !default;
-$gu-text-egyptian-web: 'Guardian Text Egyptian Web', Georgia, serif !default;
-$gu-text-sans-web: 'Guardian Text Sans Web', 'Helvetica Neue', Helvetica, Arial, 'Lucida Grande', sans-serif !default;
+$gu-headline: 'GH Guardian Headline', Georgia, serif !default;
+$gu-text-egyptian-web: 'GuardianTextEgyptian', Georgia, serif !default;
+$gu-text-sans-web: 'GuardianTextSans', 'Helvetica Neue', Helvetica, Arial, 'Lucida Grande', sans-serif !default;
```
i-hardy, line 13: How would we replicate this in Emotion? (or is it handled by Source?) Sass is a legacy part of the codebase, it would be helpful to know how we migrate away from this in future.

tomrf1:
> AFAICT the `support-frontend-mono` build is failing due to `scalastyle` errors but I don't think they're related to any changes we've made. When I run it on main though, `scalastyle` is happy so now I'm confused. If anyone has got time to take a quick look, that would be much appreciated.

Scalastyle was unhappy with the trailing commas, I've pushed commits to fix this and the tests.

support-frontend/assets/stylesheets/gu-sass/typography.scss
GHaberis: @i-hardy I believe the typography exports from Source map to font-families that reference the newer font names as the primary font and use the old font name as a fallback if the new font in unavailable.

jamie-lynch:
> > AFAICT the `support-frontend-mono` build is failing due to `scalastyle` errors but I don't think they're related to any changes we've made. When I run it on main though, `scalastyle` is happy so now I'm confused. If anyone has got time to take a quick look, that would be much appreciated.
> 
> Scalastyle was unhappy with the trailing commas, I've pushed commits to fix this and the tests.

Amazing, thanks Tom! 🙌

support-frontend/assets/stylesheets/gu-sass/typography.scss
jamie-lynch: Here's a link to where Source sets the font values which backs up what George has said above: https://github.com/guardian/source/blob/main/src/core/foundations/src/theme.ts#L27

You can then use the Typography methods that it provides:
https://theguardian.design/2a1e5182b/p/95d5d0-code

sndrs: these are [the font family name source uses](https://github.com/guardian/source/blob/0566752b3ce5851bc1a907db017e74819fcdcec1/src/core/foundations/src/theme.ts#L27-L35), so in theory you shouldn't need to do anything if you're using source.

cc @SiAdcock to make sure i've got that right!

GHaberis:
Thanks for making these changes @jamie-lynch. This is a much simpler font-loading strategy, so should be much easier for anyone new to the codebase to understand.

GHaberis:
The code looks good, although I did a quick search of the codebase and found a few hardcoded references amongst the styles to the old font names.

jamie-lynch:
> The code looks good, although I did a quick search of the codebase and found a few additional hardcoded references amongst the styles to the old font names.

Ah yes! I've updated those now.

GHaberis (approved):
All looks good to me! 👍

jamie-lynch:
Thanks for the review! I've added the `Do Not Merge` as, following a chat this morning, we want to make a couple of changes to the recommended declarations so it'd be good to get those in here before we merge.

GHaberis:
> Thanks for the review! I've added the `Do Not Merge` as, following a chat this morning, we want to make a couple of changes to the recommended declarations so it'd be good to get those in here before we merge.

No problem! @jamie-lynch  I also think it'd be worth putting this up on our `CODE` environment for a quick check before releasing to `PROD`, @i-hardy should be available to coordinate this.

sndrs:
i've updated the set of font-faces to match the recommended ones/source ones.

but agree about checking on CODE 👍

GHaberis:
@sndrs @jamie-lynch can this be merged now?

sndrs:
@GHaberis if your team is happy with it we are. I don't if it's been checked on code yet? i don't think we know how to do that...

PR #3182, opened 2021-07-21
Allow spaces in IBAN

tomrf1:
IBANs are sometimes entered with spaces within. This PR allows users to include spaces, and only removes them when it posts the form to the backend

Both of these are ok:
Screen Shot 2021-07-21 at 10 12 58
Screen Shot 2021-07-21 at 10 13 21


paulbrown1982 (approved):
👍  LGTM

PR #3183, opened 2021-07-21
Disable high contributor message for paypal/one-off

tomrf1:
Currently it uses the default values, which can cause it to incorrectly show the message about a "generous contribution".
A proper fix would involve setting the amount in session storage before the redirect


TomPretty (approved):
👍

PR #3184, opened 2021-07-22
trim name and email before sending to backend

tomrf1:
We've had some recurring contributions fail because identity doesn't allow whitespace before or after the name fields.
This change trims these fields client-side before posting to the backend.

### DIFF
--- a/support-frontend/assets/pages/contributions-landing/contributionsLandingActions.js
+++ b/support-frontend/assets/pages/contributions-landing/contributionsLandingActions.js
@@ -449,9 +449,9 @@ function regularPaymentRequestFromAuthorisation(
   const { billingCountry, billingState } = getBillingCountryAndState(authorisation, state);
 
   return {
-    firstName: state.page.form.formData.firstName || '',
-    lastName: state.page.form.formData.lastName || '',
-    email: state.page.form.formData.email || '',
+    firstName: (state.page.form.formData.firstName || '').trim(),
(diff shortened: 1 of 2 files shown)


kelvin-chappell (approved):
👍

support-frontend/assets/pages/contributions-landing/contributionsLandingActions.js
```diff
@@ -449,9 +449,9 @@ function regularPaymentRequestFromAuthorisation(
   const { billingCountry, billingState } = getBillingCountryAndState(authorisation, state);
 
   return {
-    firstName: state.page.form.formData.firstName || '',
-    lastName: state.page.form.formData.lastName || '',
-    email: state.page.form.formData.email || '',
+    firstName: (state.page.form.formData.firstName || '').trim(),
+    lastName: (state.page.form.formData.lastName || '').trim(),
+    email: (state.page.form.formData.email || '').trim(),
```
TomPretty, line 454: Is this only an issue for recurring? (single only requires `email`, but is that ever an issue?)

tomrf1: I've not seen it happen for single, we should get a payment-api alarm if it does.
I trimmed email here for completeness, but it was actually the name fields that had whitespace - perhaps some devices do this when autocompleting or something

TomPretty (approved):
Looks good 👍

PR #3185, opened 2021-07-22
Fix issue where direct debit payment option was shown when delivering outside GB or not paying in GBP

i-hardy:

This fixes the issue where we were offering the direct debit payment option in the subscriptions checkouts to users not paying in GBP. I've also done a little clean-up of the logic so we only determine available payment options once, and pass these into the PaymentMethodSelector component, and added some component testing on the checkout forms to verify the logic around when we do and don't offer the direct debit option.

[**Trello Card**](https://trello.com/c/MqGNzD3g)

## Why are you doing this?

This issue was previously addressed in #3032 and #3157- the former caused an issue with delivery or billing to the Isle of Man, which uses GBP but does not count as part of GB for payment purposes, and the latter caused an issue with the Guardian Weekly checkout where the price and currency depend on the _delivery_ country, so users based in the UK but requesting delivery to eg. Spain were being offered the opportunity to pay in EUR via direct debit.

I've added the tests to verify the behaviour of the form depending on the delivery/billing countries chosen by the user, and as a potential base for further tests we may want to write for the checkouts.

## Is this an AB test?
- [ ] Yes
- [x] No

### Billing & delivery in UK - direct debit option shown
Screenshot 2021-07-21 at 13-02-54 Support the Guardian Guardian Weekly Subscription

### Billing in Isle of Man, delivery in UK- direct debit option not shown
Screenshot 2021-07-21 at 13-01-36 Support the Guardian Guardian Weekly Subscription

### Billing in UK, delivery in Sweden- direct debit option not shown
Screenshot 2021-07-21 at 13-02-21 Support the Guardian Guardian Weekly Subscription

### DIFF
--- a/support-frontend/assets/pages/digital-subscription-checkout/components/digitalCheckoutFormGift.test.jsx
+++ b/support-frontend/assets/pages/digital-subscription-checkout/components/digitalCheckoutFormGift.test.jsx
@@ -1,0 +1,91 @@
+/* eslint-disable react/prop-types */
+import React from 'react';
+import { applyMiddleware, createStore, combineReducers, compose } from 'redux';
+import { Provider } from 'react-redux';
+import thunk from 'redux-thunk';
+import { fireEvent, render, screen } from '@testing-library/react';
+import { digitalProducts } from '__mocks__/productInfoMocks';
+import { createCheckoutReducer } from 'helpers/subscriptionsForms/subscriptionCheckoutReducer';
+import { createCommonReducer } from 'helpers/page/commonReducer';
+import { DigitalPack } from 'helpers/productPrice/subscriptions';
+import DigitalCheckoutFormGift from './digitalCheckoutFormGift';
+
+const pageReducer = commonState => createCheckoutReducer(
+  commonState.internationalisation.countryId,
+  DigitalPack,
+  '',
+  null, null, null,
+);
+
+function setUpStore(initialState) {
+  return createStore(
+    combineReducers({ page: pageReducer(initialState.common), common: createCommonReducer(initialState.common) }),
+    initialState,
+    compose(applyMiddleware(thunk)),
+  );
+}
+
+function renderWithStore(
+  component,
+  {
+    initialState,
+    store = setUpStore(initialState),
+    ...renderOptions
+  } = {},
+) {
+  function Wrapper({ children }) {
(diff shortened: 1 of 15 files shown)


support-frontend/assets/pages/digital-subscription-checkout/components/digitalCheckoutFormGift.test.jsx
```diff
@@ -0,0 +1,91 @@
...
+  } = {},
+) {
+  function Wrapper({ children }) {
+    return <Provider store={store}>{children}</Provider>;
+  }
+
+  return render(component, { wrapper: Wrapper, ...renderOptions });
+}
+
+describe('Digital gift checkout form', () => {
```
rupertbates, line 43: This is great 👍

support-frontend/assets/pages/weekly-subscription-checkout/components/weeklyCheckoutForm.test.jsx
```diff
@@ -0,0 +1,108 @@
...
+      expect(await screen.queryByText('Direct debit')).not.toBeInTheDocument();
+    });
+
+    it('does not show the direct debit option when the billing address is outside the UK', async () => {
+      const addressIsNotSame = await screen.findByRole('radio', { name: 'No' });
+      fireEvent.click(addressIsNotSame);
+      const allCountrySelects = await screen.findAllByLabelText('Country');
+      fireEvent.change(allCountrySelects[1], { target: { value: 'IE' } });
+      expect(await screen.queryByText('Direct debit')).not.toBeInTheDocument();
+    });
```
rupertbates, line 106: Could test 'does not show the direct debit option when currency is EUR even if billing address and delivery address are in the UK'

rupertbates (approved):
Great work, 💯 for adding tests!

support-frontend/assets/pages/weekly-subscription-checkout/components/weeklyCheckoutForm.test.jsx
i-hardy: How would a user get into that scenario? Setting the delivery address automatically switches the payment currency, and if you enter the checkout from the EU landing page your delivery country is initially set to be Germany.

PR #3189, opened 2021-07-27
change IT test alarms to 15 minute period

johnduffell:
Following on from https://github.com/guardian/support-frontend/pull/3188
The alarm period could be 15 minutes, this will make it group together data points from slower runs, this making it less likely to trigger.
I ahve also extended the range of the "not running" alarm to make sure it isn't running for 2 hours before going off.


rupertbates (approved):
Seems sensible

johnduffell:
looks ok
image

PR #3191, opened 2021-07-28
Csr mode

rupertbates:
We want to be able to identify Customer Service Representatives (CSRs) on support-frontend to enable them to purchase subscriptions for telephone customers. We would like to use their Salesforce logins for identification and authentication to avoid having a separate user management system.

[This document](https://docs.google.com/document/d/1okvPyyyNZUF854R5DmRSyb_RAzIrJACIC4BsBcHcGDg/edit#) describes how this will work in more detail. 

This PR adds code to identify when the site is running in CSR mode and the identity of the signed in CSR. It also has a banner which displays when in CSR mode.

## Is this an AB test?
- [ ] Yes
- [x] No

## Screenshots
Screen Shot 2021-07-29 at 14 21 38


support-frontend/assets/components/csr/csrMode.js
```diff
@@ -0,0 +1,15 @@
+const isSalesforceDomain = domain => domain.startsWith('https://gnmtouchpoint');
```
rupertbates, line 1: This will need to be updated with full domains once these are finalised

support-frontend/assets/components/csr/csrMode.js
```diff
@@ -0,0 +1,17 @@
...
+
+const listenForCsrDetails = callback =>
+  window.addEventListener(
+    'message',
+    (event) => {
+      if (isSalesforceDomain(event.origin)) {
+        callback(event.data);
+      }
+    },
+  );
```
i-hardy, line 15: I think this could be simplified if implemented as a custom hook, which also gives us easy cleanup of the event listener on `window`- something like
```suggestion
import { useEffect, useState } from 'react';

export function useCsrDetails() {
	const [csrUsername, setCsrUsername] = useState('');
	
	function checkForParentMessage (event) {
		if (isSalesforceDomain(event.origin)) {
	    	setCsrUsername(event.data);
	    }
	}

    useEffect(() => {
	    window.addEventListener('message', checkForParentMessage);
	    return () => window.removeEventListener('message', checkForParentMessage);
    }, []);

    return csrUsername;
}
```

and then used like

```javascript
const CsrBanner = () => {
  const csrUsername = useCsrDetails();

//etc
```

i-hardy (approved):
This looks good, I had a small suggestion about the hooks but it's not a blocker to merging.

support-frontend/assets/components/csr/csrMode.js
```diff
@@ -0,0 +1,17 @@
+
+const domains = ["https://gnmtouchpoint.my.salesforce.com", "https://gnmtouchpoint--dev1--c.cs88.visual.force.com"]
```
paulbrown1982, line 2: Woudn't there be 3 SF domains overall?

support-frontend/assets/components/csr/csrMode.js
GHaberis: I like this suggestion @i-hardy. Could we then do away with the  `isInCsrMode` export and instead have a conditional in `CsrBanner` that checks `csrUsername`, which the default value is `''` and therefore falsy eg.

```js
const CsrBanner = () => {
  const csrUsername = useCsrDetails();

  return csrUsername ? (
      <div css={container}>
        <div css={banner}>You are in customer support mode. Signed in as: {csrUsername}</div>
      </div>
    ) : null;
};
```

rupertbates: I've updated to Imogen's suggestion. The reason that these are separated out into different function in a different module is that I know they are going to also be needed in the checkout to get details of the currently signed in CSR. I haven't done that work in this PR though as this is a spike for now.

support-frontend/assets/components/csr/csrMode.js
rupertbates: I was trying to figure out whether UAT would be useful, the interaction with test users could be complex. I'll revisit this when I do the rest of the work to integrate with the checkout, this is mostly a proof of concept for now.

PR #3192, opened 2021-07-30
Document the TS migration process and make all new config files available

i-hardy:

This adds a new /migration directory containing all the new config files (or new versions of existing files) and scripts that we need to migrate the codebase to Typescript, and a README outlining all the necessary steps to undertake the migration.

It also removes a couple of bits of Flow syntax in Storybook config files that were causing issues in the TS migration; we weren't really gaining anything from that usage and this removes a manual step from the migration process.

[**Trello Card**](https://trello.com/c/mEeTwobC)

## Why are you doing this?

Up until now we had the migration documented in a variety of GDocs and experimental branches, with lots of different notes and comments about issues and fixes, which was becoming pretty hard to follow. This brings together the final versions that we've hashed out and a streamlined set of instructions

## Is this an AB test?
- [ ] Yes
- [x] No


GHaberis:
@i-hardy thanks for pulling all of this together and writing up the README.

I think we should now consider steps to reach our desired endpoint, and how we arrive there. Below is some thoughts regarding this. Apologies it's a bit of brain dump, but there are a few questions in there.

## Final desired state

Our desired objective is for all `.ts` and `.tsx` files within `/assets` and `/storybook` to pass our Typescript, Eslint and Prettier code validation checks and for any corresponding unit tests to pass. 

The sooner we reach this objective the better as developers can have confidence in the Type system and DevX will also be improved by standardisation of coding styles.

These validation checks should also be required for the `build-prod` command in our `package.json` to successfully complete. 

## Intermediary state following migration

Once we've carried out the migration steps outlined we'll have a number of `.ts` and `.tsx` files within `/assets` and `/storybook` that won't pass one or more of the Typescript, Eslint and Prettier code validation checks.

We will also have  a build step that `build-prod` that compiles our `.ts` and `.tsx` files to `.js` as per the configuration in our webpack, babel and tsconfig files. The `transpileOnly: true` setting in the [ts-loader] settings within `webpack.common.js` will prevent type errors from breaking the build, ideally once all Typescript errors have been resolved we can remove this setting.

**Question:** At this point will we be able to build and deploy our JavaScript bundles?

## Next steps

To arrive in our desired state we will need to fix any Typescript, Eslint and Prettier errors in every `.ts` and `.tsx` file  within `/assets` and `/storybook`.

Regarding Eslint and Prettier errors, we have the `lint:fix` and `prettier: fix` commands that fix errors in bulk over multiple files in the target directories. We have two options:

### Option 1, Fix errors in bulk

We run the  `lint:fix` and `prettier: fix` commands and fix as many files as we can in bulk. Pros for this are it involves little dev time, Cons are it's harder to identify the changes these commands have made and whether they're desired.

Even if we went with Option 1 we will still need to manually fix Eslint and Prettier errors that can't be automatically fixed via the `--fix` option.

### Option 2, Fix errors on a file-by-file basis

We run targeted commands against each file to fix Eslint and Prettier errors. Pros are it's easier to identify and review the changes these commands have made and whether they're desired, Cons are it's more manual dev steps,

**Question:** Which of these are preferable?

Whether we choose Option 1 or 2 there will always be a large number of files that need a developer to manually fix Typescript, Eslint and Prettier errors within the file.

**Question:** Can we identify all those files that contain one or more Typescript, Eslint and/or Prettier errors? If so, we can list all files that need developer attention, filtering out the `.ts` and `.tsx` files that have no errors and won't require developer attention. 

If we have a list of all files that need attention could we divide the list amongst the developers in the Contributions and Subscriptions team, assigning each developer a list of modules to fix. We could provide instructions on commands to run to see Typescript, Eslint and/or Prettier errors on an individual file and developers can fix errors on a file-by-file basis, removing them from the list of files that need attention as they do so.

If we can do this we could raise a pull-request per file, which should be easier to review. Once we have cleared the lists of files with errors we should be able to restore the validation checks into the build pipeline.

**Question:** How do we create a list of files with errors? We could do this once following the migration, but will this list remain an accurate reflection of files that need attention as the codebase changes? EG. How do we ensure we don't introduce new type errors in files which import modules where previously a type error didn't exist. Ideally we will prevent any new type errors being introduced into files. Perhaps we could use the Github action to check no files which previous were error free have now got errors?

tomrf1 (approved):
brilliant work!
I've run through this locally and it all looks good

support-frontend/migration/README.md
```diff
@@ -0,0 +1,59 @@
...
+3. Run `node scripts/ts-migrate.js --dir=assets` to migrate the assets directory (this may take a while!)
+4. Run `node scripts/ts-migrate.js --dir=stories` to migrate the stories directory
+
+### Post-migration manual fixes
+
+There are a few manual syntax fixes that need to be applied. Search and replace in both /assets and /stories for the following:
+
+- Replace “import type” with “import”
+- Replace “, type “ with “, “
+- Replace “{ type” with “{“
```
TomPretty, line 29: Discussed already but just for posterity: we might not need these, and the last one affects a few comments in the code

support-frontend/migration/README.md
```diff
@@ -0,0 +1,59 @@
...
+### VSCode
+
+In order to get the most out of the new Prettier and ESLint setup, VSCode should be configured to run them and fix problems on save. To set up Prettier, follow this guide on [How to use Prettier with ESLint and TypeScript in VSCode](https://khalilstemmler.com/blogs/tooling/prettier/#Formatting-using-VSCode-on-save-recommended). For ESLint, add the following to your settings.json:
+
+```json
+"editor.codeActionsOnSave": {
+  "source.fixAll.eslint": true
+},
+"eslint.validate": ["typescript", "typescriptreact"]
+```
```
TomPretty, line 57: At the moment we've got `prettier` set up as an `eslint` rule so having vscode fix linting errors on save runs prettier too, so you don't need the prettier plugin installed. However, it looks like the [prettier docs](https://prettier.io/docs/en/integrating-with-linters.html) now recommend against these, and instead just suggest running prettier standalone. Maybe we should update the config to do that instead? (In which case, that guide is 👌 for setting it up)

TomPretty (approved):
Amazing work team. Really looking forward to this migration and very keen to help out fixing up the errors after we do. 

On @GHaberis's point about linting, I think fixing in bulk is the way to go. It's pretty cheap to run formatting so it's not too painful to tweak the config and then reformat if we're not happy with how it looks

TomPretty (commented):
Do we want to go full [@guardian/configs](https://github.com/guardian/configs) and use the `.editorconfig` + `eslint` plugin?

i-hardy:
Running Prettier across the codebase seems to work pretty well the times I've tried it, so I'll add that in to the migration guide. With ESLint, running the fix command across all the code at once always causes some pretty severe problems- I'm not sure exactly why, but it's wrecked some files so badly that I've had to completely scrap the migrated code and start again. I think it might be better to take an incremental approach with that too- we'll be touching most of the files anyway to fix type issues, so as long as we have ESLint auto-fixing on save it'll happen more or less automatically.

GHaberis:
> Do we want to go full [@guardian/configs](https://github.com/guardian/configs) and use the `.editorconfig` + `eslint` plugin?

I'd back this suggestion, it'd take remove the overhead from the team regarding deciding which rules to apply.

TomPretty:
> Running Prettier across the codebase seems to work pretty well the times I've tried it, so I'll add that in to the migration guide. With ESLint, running the fix command across all the code at once always causes some pretty severe problems- I'm not sure exactly why, but it's wrecked some files so badly that I've had to completely scrap the migrated code and start again. I think it might be better to take an incremental approach with that too- we'll be touching most of the files anyway to fix type issues, so as long as we have ESLint auto-fixing on save it'll happen more or less automatically.

That makes sense, if letting `eslint` loose is actually quite risky then sounds like the easiest thing is to just make sure it's happy with the files after we've touched them

PR #3196, opened 2021-08-03
Turn off comparison table A/B test

marialani:
This turns off the comparison table A/B test. We may use it again later, so for now the test and related components will remain in the codebase in an inactive state.


GHaberis (approved):
👍

PR #3198, opened 2021-08-04
AB test the ticker

tomrf1:
the ticker is currently displaying on the landing page for Aus users.
We'd like to AB test its impact


TomPretty (approved):
Looks good 🚀

PR #3203, opened 2021-08-05
Identify PPC sources in acquisitions data

paulbrown1982:

[**Trello Card**](https://trello.com/c/AJVCpH2a)

## Why are you doing this?

The Support website does not attribute inbound links from search engine pay-per-click sources (e.g. Google Ads) as having a `source` of `PPC`, even though that enum is supported. As a consequence, acquisitions made via this journey are attributed the same as if the user had visited the site via organic search or by just typing support.theguardian.com/* in their address bar.

I've solved this by:

- Including the value from &CMP= in the hierarchy of values which identify a campaign code. This adds only additive tracking benefits from what I can understand.
- Adding a tiny parser of that to identify the link as coming from a search engine pay-per-click. This is driven by the assumption that inbound links from Bing have an `&CMP=ppc_*` param and inbound links from Google have `&gclid=*`. Chris is going to try and add `&CMP=` to future Google links, but that may take some time. I don't expect this parser to generate false positives, it's more likely to miss, and be no different to now.

## Is this an AB test?
- [ ] Yes
- [x] No

## Screenshots
N/A

### DIFF
--- a/support-frontend/assets/helpers/tracking/acquisitions.js
+++ b/support-frontend/assets/helpers/tracking/acquisitions.js
@@ -290,8 +290,6 @@ function getReferrerAcquisitionData(): ReferrerAcquisitionData {
     ? deserialiseReferralData(getQueryParameter(REFERRAL_DATA_PARAM) || '')
     : deserialiseJsonObject(getQueryParameter(ACQUISITIONS_PARAM) || '');
 
-  debugger;
-
   // Read from param, or read from sessionStorage, or build minimal version.
   const referrerAcquisitionData = buildReferrerAcquisitionData(paramData || readReferrerAcquisitionData() || {});
   storeReferrerAcquisitionData(referrerAcquisitionData);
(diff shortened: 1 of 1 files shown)


support-frontend/assets/helpers/tracking/acquisitions.js
```diff
@@ -188,13 +188,15 @@ function buildReferrerAcquisitionData(acquisitionData: Object): ReferrerAcquisit
     acquisitionData.queryParameters ||
     toAcquisitionQueryParameters(getAllQueryParamsWithExclusions(parameterExclusions));
 
+  const establishedSource = (/\bPPC\b/i.test(campaignCode) || getQueryParameter('glcid')) ? 'PPC' : acquisitionData.source;
```
tomrf1, line 191: flow isn't happy - `Cannot call `/\bPPC\b/i.test` with `campaignCode` bound to `string` because null or undefined [1] is incompatible with`
How about: `(campaignCode && /\bPPC\b/i.test(campaignCode)) || getQueryParameter('glcid')`

paulbrown1982: 🤦

tomrf1 (approved):
looks good!

PR #3205, opened 2021-08-09
Update eslint config in TypeScript migration instructions

GHaberis:
Whilst going through the TypeScript migration instructions in https://github.com/guardian/support-frontend/pull/3192 I managed to recreate an issue where the automated application of `eslint` rules to a file was breaking the syntax, in this case it introduced closing parentheses where there was no corresponding opening parentheses! This in turn completely broke the build steps.

I decided to follow-up on a suggestion by @TomPretty to use the `@guardian/eslint-config-typescript` [package](https://github.com/guardian/configs/tree/main/packages/eslint-config-typescript) instead of the configuration initially proposed. I've replaced the `.eslintrc.js` configuration with a much simpler set-up that extends `@guardian/eslint-config-typescript`. This also meant we could remove a number of `eslint-*` dependencies.

I ran into two issues along the way 1) I had to install `eslint-import-resolver-webpack` and include an `import/resolver` setting in the `.eslintrc.js` that used this 2) `flow-to-ts` would replace some Flow utility types are replaced by the `utility-types` [package](https://www.npmjs.com/package/utility-types) so I had to add this dependency to the `package.json`.


support-frontend/migration/README.md
```diff
@@ -22,11 +22,7 @@ All files should replace any existing versions of the same file in their destina
 
 ### Post-migration manual fixes
 
-There are a few manual syntax fixes that need to be applied. Search and replace in both /assets and /stories for the following:
-
-- Replace “import type” with “import
-
-Then do the following file-specific changes:
+There are a few manual syntax fixes that need to be applied:
```
GHaberis, line 25: Interestingly after updating the `.eslintrc.js` to use `@guardian/eslint-config-typescript` every time I attempted to manually replace  “import type” with “import” my IDE would apply our lint rules on saving the file and this began reintroducing the `type` keyword in import statements! 

I believe the `@typescript-eslint/consistent-type-imports` rule from `@guardian/eslint-config-typescript`'s [configuration](https://github.com/guardian/configs/blob/main/packages/eslint-config-typescript/index.js#L40) is driving this. The use of the `type` keyword in these import statements is correct syntax in Typescript.

I therefore attempted to test and build (ssr, dev and prod) the source code with the `type` keyword in place and all the commands worked. I also ran the site locally and it all behaved as expected. I therefore think we can remove this step.

tomrf1 (approved):
nice work!

PR #3212, opened 2021-08-13
Remove snyk check from branches

tomrf1:
It marks all branches as failed, which isn't helpful.
We have a snyk rota for addressing these issues

### DIFF
--- a/.github/workflows/snyk.yml
+++ b/.github/workflows/snyk.yml
@@ -7,6 +7,8 @@ on:
   schedule:
     - cron: '0 6 * * *'
   push:
+    branches:
+      - main
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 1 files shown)


ripecosta (commented):
You can make this entire Action run only on main if you add the branch name to line 9 of this doc. Avoids having the check on every step.

You can specific the branch name by editing line 9 like this:
```
push:
    branches:
      - main
```

tomrf1:
> You can make this entire Action run only on main if you add the branch name to line 9 of this doc. Avoids having the check on every step.
> 
> You can specific the branch name by editing line 9 like this:
> 
> ```
> push:
>     branches:
>       - main
> ```

nice, thanks!

ripecosta (approved):
👍

PR #3213, opened 2021-08-13
Add snyk as dev dep

tomrf1:
so we can run `yarn snyk` (my local version of the snyk cli doesn't work)
also deleted the .snyk as it's not doing anything


ripecosta (approved):
BOOM! 💪

PR #3216, opened 2021-08-18
Document processes for managing linting and type errors in the Typescript migration

i-hardy:

This adds documentation around the processes for auto-fixing errors in the initial migrated codebase, and managing and fixing errors in the following weeks as we bring the new TS code up to scratch. It also adds a git hook to run ESLint on changed files and also regenerate the list of TS errors on every commit that contains .ts or .tsx files, so that we can easily keep track of what's been fixed and if anything new has been introduced via the diff.

You can see an example of this 'error log' file on my experimental branch [here](https://github.com/guardian/support-frontend/blob/ih-new-ts-migration/support-frontend/typescript-errors.json).

[**Trello Card**](https://trello.com/c/2G5P39Uh)

## Why are you doing this?

This is part of the ongoing prelimary work preparing to migrate the codebase- we've been aware for a while that the code is going to have a lot of type errors to begin with, so this should help us understand and follow a process for working through them and getting them fixed in the longer term.

## Is this an AB test?
- [ ] Yes
- [x] No


GHaberis:
Looks like the link to the error log on your experimental branch is broken, it should be: https://github.com/guardian/support-frontend/blob/ih-new-ts-migration/support-frontend/typescript-errors.json

rupertbates (commented):
Generally docs for support-frontend go into the wiki rather than in md files. Should we be putting this there? If not are there elements of this that should go into the wiki?

i-hardy:
The majority of what's in this doc is going to be temporary and only applies to the few hours in which we do the migration and testing. The longer term stuff concerning how we bring down the errors over time could probably go in the wiki though- I'm happy to do this now, if you think that's appropriate, or it can wait until we've migrated the code.

GHaberis:
This is great. Nice to see the use of Git hooks and GitHub actions here! A couple comments regarding the Typescript Monitor Github action. This action reports on any increase to the amount of type/lint errors and will be a merge blocking validation step. What feedback does the action provide in the PR? Does it provide any indication of where any new errors have been introduced? This would be helpful for devs looking to address any flagged increase in errors, as these errors may not always be in the files changed. For example, they could be introduced to files that import the changed files.

Regarding `typescript-errors.json` this looks useful for grouping errors by code and identifying common errors, however if we're looking to work through errors on a file-by-file basis would it be useful to also include something that allows a quick view of all the files that contain errors? For example, a list of the file paths that contain errors, potentially with an error count to give a rough idea of the scale of work required in each file? EG:

```
{
  assets/components/directDebit/directDebitProgressiveDisclosure/directDebitForm.tsx:  1,
  assets/components/headers/roundelHeader/header.tsx: 5,
  assets/components/subscriptionCheckouts/stripeForm/stripeForm.tsx" 8
}
```

i-hardy:
Listing the error counts by file path too is a great idea, I've updated the script to do this and run it on my other branch so you can see the result in that `typescript-errors.json` file.

Currently the Typescript Monitor action reports in a PR comment the amount by which TS and ESLint errors have increased/decreased (and will fail on any increase). It doesn't report where any errors have arisen. Since we'll also be re-generating the log file on every commit, the diff should help identify where any new TS errors have been introduced. With lint issues I'm not sure it's possible to introduce a lint issue into a file you haven't touched (aside from issues which break imports, which will cause the build to break), so `lint-staged` in the commit hook should also help with this and will prevent commits with lint issues.

GHaberis:
> Listing the error counts by file path too is a great idea, I've updated the script to do this and run it on my other branch so you can see the result in that `typescript-errors.json` file.

Looks great 👍 

> Currently the Typescript Monitor action reports in a PR comment the amount by which TS and ESLint errors have increased/decreased (and will fail on any increase). It doesn't report where any errors have arisen. Since we'll also be re-generating the log file on every commit, the diff should help identify where any new TS errors have been introduced. 

Good point regarding the diff! I agree this should be enough to help devs identify new issues.

> With lint issues I'm not sure it's possible to introduce a lint issue into a file you haven't touched (aside from issues which break imports, which will cause the build to break), so `lint-staged` in the commit hook should also help with this and will prevent commits with lint issues.

Sounds good, and yes regarding introducing lint errors on untouched files, I agree this shouldn't be possible (other than breaking imports as mentioned)

GHaberis (approved):
Nice work 👍

PR #3218, opened 2021-08-19
Snyk version bumps

johnduffell:

This PR updates a load of versions needed for snyk, plus a couple that are just good to bump.

## Why are you doing this?

We want to keep our stuff secure and look after our customer data.


GHaberis (approved):
👍

tomrf1:
A couple of builds have failed, but looks like we just need extra sbt merge strategies for `javax`

johnduffell:
thanks Tom I was struggling with that on friday as it seems to work locally!  I will carry on this morning (I might just add a relevant merge strategy, rather than trying to reproduce locally)

johnduffell:
so just to update on this, it worked locally because if you run `sbt assembly` and it fails to merge, trying again seems to succeed without errors in some cases.
However if you then delete the `target` directory and try again, it does fail.
This is a bit concerning, but not enough that I fancy looking into it now. We do a clean build every time in teamcity because of possible issues of this kind, so I am happy enough that it won't affect anything productionny even if it's confusing.

PR #3220, opened 2021-08-24
 Removing the Krux bypass within our CSP as we no longer use that tag

paulbrown1982:

Removing the Krux bypass within our CSP as we no longer use that tag.

[**Trello Card**](https://trello.com/c/fz7M4Brc)

- [ ] Tested in CODE


GHaberis (approved):
👍

PR #3221, opened 2021-08-25
Add alarm for "no Guardian Weekly checkouts"

johnduffell:

This PR adds an alarm so we know if GW conversion is broken.
https://trello.com/c/9uB91SVP/2909-add-an-alarm-for-no-gw-acquisitions-in-x-hours-like-we-have-for-paper-and-digital

## Why are you doing this?

We won't know until userhelp get in touch, if the GW checkout is not allowing people to complete a purchase in any payment method at all.  This should be a cheap way to find out quicker.

### DIFF
--- a/support-workers/cloud-formation/src/templates/cfn-template.yaml
+++ b/support-workers/cloud-formation/src/templates/cfn-template.yaml
@@ -345,7 +345,7 @@ Resources:
     Properties:
       AlarmActions:
         - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:reader-revenue-dev
-      AlarmName: !Sub URGENT 9-5 - ${Stage} support-workers No successful digipack checkouts in 3h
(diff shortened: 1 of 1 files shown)


rupertbates (approved):
👍 lgtm

support-workers/cloud-formation/src/templates/cfn-template.yaml
```diff
@@ -403,7 +403,7 @@ Resources:
           ReturnData: false
       ComparisonOperator: LessThanOrEqualToThreshold
       Threshold: 0
-      EvaluationPeriods: 36
```
johnduffell, line 406: having reviewed the stats, I think 2h should be ample to avoid false alarms.  I reviewed all the rest including contributions and they all look ok.
There may even be scope for a no digital subs via Direct debit / paypal/ stripe separately.

PR #3204, opened 2021-08-09
Refactored how we establish the page's referrer acquisition data

paulbrown1982:
Refactored how we establish the page's referrer acquisition data. Mostly renames and functionisation. 

The main difference in logic is a fix where presence of a `&gclid=` value now clears all other referrer acquisition data and sets its own campaign code too.

- [x] Tested locally

PR #3222, opened 2021-08-25
convert query params in GCL case to the correct format for the backend

johnduffell:

This PR converts the queryParameters in the referrerAcquisitionData into the correct format.
Before it was
image
...which caused a 400 bad request due to a backend deserialisation error

This will change it to 
```json
[
  {
     "name": "thekey",
     "value":"thevalue"
  }
... more pairs here
]
```

This is a fix to this PR: https://github.com/guardian/support-frontend/pull/3204

## Why are you doing this?

Stop the failed signups from happening.

### DIFF
--- a/support-frontend/assets/helpers/tracking/acquisitions.js
+++ b/support-frontend/assets/helpers/tracking/acquisitions.js
@@ -303,7 +303,7 @@ function getAcquisitionDataFromPPCParams(): ?Object {
     return {
       source: 'PPC',
       campaignCode: 'Google_Adwords',
-      queryParameters: getAllQueryParams(),
+      queryParameters: toAcquisitionQueryParameters(getAllQueryParams()),
     };
   }
   return null;


support-frontend/assets/helpers/tracking/acquisitions.js
```diff
@@ -303,7 +303,7 @@ function getAcquisitionDataFromPPCParams(): ?Object {
     return {
       source: 'PPC',
       campaignCode: 'Google_Adwords',
-      queryParameters: getAllQueryParams(),
+      queryParameters: toAcquisitionQueryParameters(getAllQueryParams()),
```
johnduffell, line 306: we still need to know whether to add the exclusions to this
https://github.com/guardian/support-frontend/blob/7e5384152a8fe6168126fd44b9df445e5ab8051f/support-frontend/assets/helpers/tracking/acquisitions.js#L176
hope @paulbrown1982 can help with this decision

rupertbates (approved):
👍

johnduffell:
we have decided to ship this with the outstanding question, as it works (better) in CODE, however we should revisit this later to make sure we exclude the right parameters from `queryParameters` field if necessary

PR #3224, opened 2021-08-31
Update product price tests

marialani:

This pr adds tests for a number of files in the `productPrice` folder that handle the way we access pricing information and the text about them.

## Why are you doing this?

These tests will prevent regressions when modifying code within the `productPrice` folder. We will soon be separating the work of getting the correct price from the text about the price, as they are currently intertwined in a way that makes it difficult to make the nimble text changes that we increasingly need to make. 

[**Trello Card**](https://trello.com/c/t6CiFUA6/3753-write-tests-for-price-text-and-price-calculation-code-90)

## Is this an AB test?
- [ ] Yes
- [x] No

### DIFF
--- a/support-frontend/assets/helpers/productPrice/__tests__/paperProductPricesTests.js
+++ b/support-frontend/assets/helpers/productPrice/__tests__/paperProductPricesTests.js
@@ -1,0 +1,178 @@
+// @flow
+
+// ----- Imports ----- //
+import { getMaxSavingVsRetail } from 'helpers/productPrice/paperProductPrices';
+
+const productPrices = {
+  'United Kingdom': {
+    Collection: {
+      SixdayPlus: {
+        Monthly: {
+          GBP: {
+            price: 27.36, savingVsRetail: 35, currency: 'GBP', fixedTerm: false, promotions: [],
+          },
+        },
+      },
+      SundayPlus: {
+        Monthly: {
+          GBP: {
+            price: 22.06, savingVsRetail: 15, currency: 'GBP', fixedTerm: false, promotions: [],
+          },
+        },
+      },
+      SaturdayPlus: {
+        Monthly: {
+          GBP: {
+            price: 21.62, savingVsRetail: 16, currency: 'GBP', fixedTerm: false, promotions: [],
+          },
+        },
+      },
+      Saturday: {
+        Monthly: {
+          GBP: {
+            price: 10.36, savingVsRetail: 13, currency: 'GBP', fixedTerm: false, promotions: [],
+          },
+        },
+      },
+      Sixday: {
+        Monthly: {
+          GBP: {
+            price: 41.12, savingVsRetail: 26, currency: 'GBP', fixedTerm: false, promotions: [],
+          },
+        },
+      },
+      Weekend: {
+        Monthly: {
+          GBP: {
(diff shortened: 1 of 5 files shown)


support-frontend/assets/helpers/productPrice/__tests__/promotionsTests.js
```diff
@@ -1,28 +1,152 @@
...
+    expect(hasIntroductoryPrice(promotionWithoutIntroductoryPrice)).toEqual(false);
+    expect(hasIntroductoryPrice()).toEqual(false);
+  });
+});
+
+describe('promotionHTML', () => {
+  it('should return promotion html if present', () => {
+    expect(promotionHTML()).toEqual(null);
+    expect(promotionHTML('Get 25% off').props).toEqual({ __EMOTION_TYPE_PLEASE_DO_NOT_USE__: 'span', css: '', dangerouslySetInnerHTML: { __html: 'Get 25% off' } });
+    expect(promotionHTML('Get 20% off', { tag: 'div' }).props).toEqual({ __EMOTION_TYPE_PLEASE_DO_NOT_USE__: 'div', css: '', dangerouslySetInnerHTML: { __html: 'Get 20% off' } });
```
johnduffell, line 150: I don't understand a lot about client side tests, but is the `__EMOTION_TYPE_PLEASE_DO_NOT_USE__` something that has to be correct here?

support-frontend/assets/helpers/productPrice/__tests__/paperProductPricesTests.js
```diff
@@ -0,0 +1,223 @@
...
+      EverydayPlus: {
+        Monthly: {
+          GBP: {
+            price: 29.19, currency: 'GBP', fixedTerm: false, promotions: [],
+          },
+        },
+      },
+    },
+  },
+};
```
johnduffell, line 167: just a general comment as I can see that you have moved this from elsewhere.  If it was a new test, I would suggest using a smaller, more focussed data structure for each test, to make it clearer which values were needed.

johnduffell (approved):
looks great, always good to have more reliable test coverage.   I have made some general comments/questions but nothing that needs changes as far as I can see. 👍

rupertbates (approved):
Great work, this gives us a lot more coverage!

support-frontend/assets/helpers/productPrice/__tests__/promotionsTests.js
marialani: When we call the `promotionHTML` function the result from the shallow render is wrapped in a ForwardRef e.g. `<ForwardRef(render) __EMOTION_TYPE_PLEASE_DO_NOT_USE__="div" css="" dangerouslySetInnerHTML={{"__html": "Get 20% off"}} />` which I wasn't too sure how to get around

Though I could also update these tests to focus on the `dangerouslySetInnerHTML` object but thought as the function also allows you to specify which tag to use it might be most useful to see all of the props.

support-frontend/assets/helpers/productPrice/__tests__/paperProductPricesTests.js
marialani: Yeah this data structure was moved and I had then used the same data structure for subsequent tests in this file, which I can change if you think it would be clearer? Though for the original test that was moved i.e. `getMaxSavingVsRetail` all of the data is being used to do the calculation.

PR #3225, opened 2021-08-31
change DS alarm back to 3h

johnduffell:

This PR changes the DS alarm back to 3h

## Why are you doing this?
I changed it to 2h as an opportunity in this PR
https://github.com/guardian/support-frontend/pull/3221/files#r695752789

However, it seems it went off fairly soon after, so I was a bit optimistic (I only checked back a week or so to ensure no false alarms would happen)


kelvin-chappell (approved):
👍

PR #3227, opened 2021-09-01
Update contributions SSR

tomrf1:
I tried to include empty buttons and text fields to make it look more like a form, but getting emotion to work with our SSR is a _nightmare_.
Instead it now renders what it can without the design system, which doesn't look too bad.
I've deleted most of the old SSR components.

Screen Shot 2021-09-01 at 15 11 15
Screen Shot 2021-09-01 at 15 11 32


michaelbjacobson (approved):
LGTM 🙂

PR #3229, opened 2021-09-02
Add Saturday plan back into Paper product

rupertbates:

## Is this an AB test?
- [ ] Yes
- [ ] No


support-frontend/assets/components/product/productOption.jsx
```diff
@@ -95,34 +182,48 @@ function ProductOption(props: Product) {
...
         <ThemeProvider theme={buttonReaderRevenue}>
           <LinkButton
+            css={button}
             href={props.href}
             onClick={props.onClick}
-            aria-label={`${props.title}- ${props.buttonCopy}`}
+            aria-label={`${props.title}- ${props.buttonCopy} now`}
           >
-            {props.buttonCopy}
+            {props.buttonCopy}<span css={buttonCopySpan}>&nbsp;now</span>
```
GHaberis, line 226: Is this `<span>` appending "now" to the button copy necessary? I think it'd be easier to find where the button copy string is defined, before it's passed down in the props, and to just add "now" to the button copy string. Plus, this is a shared Component, so hard coding "now" on the end here can have unintended side affects. EG. The Guardian Weekly CTA on the pricing cards becomes:

Screenshot 2021-09-20 at 09 08 20

support-frontend/assets/components/packshots/paper-packshot.jsx
```diff
@@ -5,31 +5,13 @@ import GridImage from 'components/gridImage/gridImage';
...
-      srcSizes={[756, 500, 140]}
-      sizes="(max-width: 740px) 100%,
-             (max-width: 980px) 204px,
-             (max-width: 1140px) 227px,
-             252px"
+      srcSizes={[500, 140]}
+      sizes="(max-width: 739px) 140px,
+             (max-width: 979px) 500px,
+             (max-width: 1140px) 500px,
+             500px"
```
GHaberis, line 14: Is the 1px difference in some of the media-queries here necessary?

support-frontend/assets/components/product/productOption.jsx
marialani: Ah yes I'd missed where it was shared. With the latest design the text for the button now changes for different screen sizes so had opted to conditionally include the "now" copy using media queries but will change this now.

support-frontend/assets/components/packshots/paper-packshot.jsx
marialani: Here I'd followed the convention in the codebase but I do believe the 1px difference may not be necessary due to cascading

GHaberis: OK great, if the previous implementation went against the convention then we should retain the changes you've made.

support-frontend/assets/components/product/productOption.jsx
GHaberis: OK I see, good to know! I'd question whether the inclusion of "now" is really essential for the objective of this piece of work - to launch this Newspaper campaign, I'd say the current wording on the CTA without the "now" is fine, and if we do want to implement different wording at different breakpoints we create a separate card/PR to do this.

GHaberis (approved):
👍

PR #3233, opened 2021-09-07
Turn off emailDigiSubEventsTest AB test

GHaberis:

The `emailDigiSubEventsTest` has come to an end, this PR turns off the test in `abtestDefinitions.js` so members of the public will no longer be able to enter the test. A follow-up PR will remove the Event component and any other associated logic required to support the test.


marialani (approved):
👍

PR #3234, opened 2021-09-08
choose your payment date AB test for Digital Subscriptions

johnduffell:

This PR is a hack day PR and adds a new wiedgt for 50% audience (once the AB test is enabled) that lets people choose their payment date for Digi sub.

image

It also updates relevant copy to say "14+ day trial" or "at least 14 day trial" as appropriate.

## Why are you doing this?

This is to test whether choosing a payment date reduces churn and payment failure.
There will be a small financial loss due to extended free trial, but this may be mitigated in the long term.
There is also some work to look into adjusting payment dates later in people's subscription.

## Is this an AB test?
YES
PR reviewers should open and check the Optimize test.

## Screenshots

image

### DIFF
--- a/support-frontend/assets/helpers/abTests/abtestDefinitions.js
+++ b/support-frontend/assets/helpers/abTests/abtestDefinitions.js
@@ -88,6 +88,27 @@ export const tests: Tests = {
     targetPage: pageUrlRegexes.contributions.auLandingPage,
     seed: 1,
   },
+  freeTrialLengthTest: {
+    variants: [
+      {
+        id: 'control',
+      },
+      {
+        id: 'variant',
+      },
+    ],
+    audiences: {
+      ALL: {
+        offset: 0,
+        size: 1,
+      },
+    },
+    isActive: false,
+    referrerControlled: false,
+    targetPage: pageUrlRegexes.subscriptions.digiSub.nonGiftLandingAndCheckout,
+    seed: 5,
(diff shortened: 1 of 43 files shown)


support-frontend/assets/helpers/abTests/abtestDefinitions.js
```diff
@@ -88,6 +88,27 @@ export const tests: Tests = {
...
+        id: 'control',
+      },
+      {
+        id: 'variant',
+      },
+    ],
+    audiences: {
+      ALL: {
+        offset: 0,
+        size: 1,
```
GHaberis, line 103: We should probably consider any other ab tests we intended to run on the Digi Subs landing page in the near future, such as the upcoming Comparison table AB test. We might want to set the `size` to `0.5` so the audience under test doesn't over lap?

support-frontend/assets/helpers/forms/paymentIntegrations/readerRevenueApis.js
```diff
@@ -147,6 +147,7 @@ export type RegularPaymentRequest = {|
   telephoneNumber: Option<string>,
   promoCode?: Option<string>,
   deliveryInstructions?: Option<string>,
+  freeTrialLength?: Option<number>,
```
paulbrown1982, line 150: For the record, `freeTrialLength` has the same net effect now within Digital Subscriptions as `firstDeliveryDate` does within paper subscriptions - they both effectively delay the customer acceptance date in the subscription order.

PR #3235, opened 2021-09-10
Add spacing in the bottom of patrons text

Aracho1:
Add a line break after the patrons text to give spacing at the bottom

**before**
image

**after**
image


tomrf1 (approved):
thanks!

PR #3237, opened 2021-09-13
Track 3ds result for single contributions

tomrf1:
We already send an ophan event when the 3DS popup appears, but not when it succeeds/fails.
With this PR we now have all 3 events:

3DS popup:
{"component":{"componentType":"ACQUISITIONS_OTHER","id":"stripe-3ds"},"action":"VIEW"}

Fail:
{"component":{"componentType":"ACQUISITIONS_OTHER","id":"stripe-3ds-failure"},"action":"CLICK"}

Succeed:
{"component":{"componentType":"ACQUISITIONS_OTHER","id":"stripe-3ds-success"},"action":"CLICK"}

This applies for single contributions only. For recurring the appearance of the 3DS popup is managed by the stripe sdk, so we cannot have these events (https://stripe.com/docs/js/deprecated/handle_card_setup).
Stripe have provided us with a dashboad for monitoring 3DS events so hopefully we'll still have the full picture


TomPretty (approved):
LGTM 👍

PR #3242, opened 2021-09-17
Do not require sign in on contributions page

tomrf1:
Existing behaviour:
If the provided email address has a password, require the user to sign in before letting them create a recurring contribution.

New behaviour:
Never require sign in.


rupertbates (approved):
Looks good, is there any action or reducer code which can be deleted, or do we still need that?

tomrf1:
> Looks good, is there any action or reducer code which can be deleted, or do we still need that?

Thanks. I'm pretty sure that's it, the `isSignedIn` and `userTypeFromIdentityResponse` fields are still required elsewhere

PR #3246, opened 2021-09-21
if a create request fails for any reason, alarm immediately

johnduffell:

The server side create endpoint is the final step in creating a subscription, after the client side work has been done.
This PR makes us get an alarm if there's any non 2xx response from the create endpoint on the server side.

Tested locally with Digi sub and contribution (positive and negative cases)

https://trello.com/c/RzkGDEbI/3849-alarm-on-a-single-deserialisation-failure-on-acquisition

## Why are you doing this?

This is needed in response to an incident where some subscribe requests did not validate on the server side, but no alarms were able to detect that, as it was rejected immediately by the parser.  See retro doc here https://docs.google.com/document/d/10K8VFm2oUv6SOu6U49B1rUPtHQPwGyZB2pV8Cfam_dE/edit#
This tries to capture all errors between the request header being read in to the server and available as a ByteString, and the response being returned.

image
https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#metricsV2:graph=~(metrics~(~(~'support-frontend~'ServerSideCreateFailure~'Stage~'DEV))~view~'timeSeries~stacked~false~region~'eu-west-1~start~'-PT3H~end~'P0D~stat~'Sum~period~60~yAxis~(left~(min~0)));query=~'*7bsupport-frontend*2cStage*7d

### DIFF
--- a/support-frontend/app/actions/CustomActionBuilders.scala
+++ b/support-frontend/app/actions/CustomActionBuilders.scala
@@ -1,13 +1,18 @@
 package actions
 
-import io.lemonlabs.uri.typesafe.dsl._
+import com.gu.aws.{AwsCloudWatchMetricPut, AwsCloudWatchMetricSetup}
+import com.gu.monitoring.SafeLogger
+import com.gu.monitoring.SafeLogger.Sanitizer
+import com.gu.support.config.Stage
 import config.Configuration.IdentityUrl
+import io.lemonlabs.uri.typesafe.dsl._
 import play.api.mvc.Results._
 import play.api.mvc.Security.AuthenticatedRequest
 import play.api.mvc._
 import play.filters.csrf._
 import services.{AsyncAuthenticationService, AuthenticatedIdUser}
 import utils.FastlyGEOIP
+
 import scala.concurrent.ExecutionContext
 
 object CustomActionBuilders {
@@ -23,7 +28,8 @@ class CustomActionBuilders(
   cc: ControllerComponents,
   addToken: CSRFAddToken,
   checkToken: CSRFCheck,
-  csrfConfig: CSRFConfig
+  csrfConfig: CSRFConfig,
+  stage: Stage
 )(implicit private val ec: ExecutionContext) {
 
   import CustomActionBuilders._
@@ -77,6 +83,18 @@ class CustomActionBuilders(
 
   def maybeAuthenticatedAction(identityClientId: String = membersIdentityClientId): ActionBuilder[OptionalAuthRequest, AnyContent] =
     PrivateAction andThen maybeAuthenticated(onUnauthenticated(identityClientId))
+
+  def alarmOnFailure[A](action: Action[A]): EssentialAction =
+    (v1: RequestHeader) => action.apply(v1).map { result =>
+      if (result.header.status.toString.head != '2') {
+        SafeLogger.error(scrub"pushing alarm metric - non 2xx response ${result.toString()}")
+        val cloudwatchEvent = AwsCloudWatchMetricSetup.serverSideCreateFailure(stage)
+        AwsCloudWatchMetricPut(AwsCloudWatchMetricPut.client)(cloudwatchEvent)
+        result
+      } else {
+        result
+      }
+    }
 
   val CachedAction = new CachedAction(cc.parsers.defaultBodyParser, cc.executionContext)
 
--- a/support-frontend/app/controllers/CreateSubscriptionController.scala
+++ b/support-frontend/app/controllers/CreateSubscriptionController.scala
@@ -46,8 +46,8 @@ class CreateSubscriptionController(
 
   type ApiResponseOrError[RES] = EitherT[Future, CreateSubscriptionError, RES]
 
-  def create: Action[CreateSupportWorkersRequest] =
-    maybeAuthenticatedAction().async(new LoggingCirceParser(components).requestParser) {
+  def create: EssentialAction =
+    alarmOnFailure(maybeAuthenticatedAction().async(new LoggingCirceParser(components).requestParser) {
       implicit request =>
         request.user match {
           case Some(user) =>
@@ -57,7 +57,7 @@ class CreateSubscriptionController(
             SafeLogger.info(s"Guest user ${request.body.email} is attempting to create a new ${request.body.product} subscription [${request.uuid}]")
             createGuestUserAndHandleRequest.getOrElse(InternalServerError)
         }
-    }
+    })
 
   def createGuestUserAndHandleRequest(implicit request: OptionalAuthRequest[CreateSupportWorkersRequest]): EitherT[Future, String, Result] =
     for {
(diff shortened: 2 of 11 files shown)


support-frontend/app/actions/CustomActionBuilders.scala
```diff
@@ -78,6 +84,18 @@ class CustomActionBuilders(
   def maybeAuthenticatedAction(identityClientId: String = membersIdentityClientId): ActionBuilder[OptionalAuthRequest, AnyContent] =
     PrivateAction andThen maybeAuthenticated(onUnauthenticated(identityClientId))
 
+  def alarmOnFailure[A](action: Action[A]): EssentialAction =
+    (v1: RequestHeader) => action.apply(v1).map { result =>
+      if (result.header.status.toString.head != '2') {
+        SafeLogger.error(scrub"pushing alarm metric - non 2xx response ${result.toString()}")
```
johnduffell, line 90: the error log contains something like this
```
2021-09-21T16:38:40,122+0000 [application-akka.actor.default-dispatcher-17] INFO  com.gu.monitoring.SafeLogger$ - incoming POST: {"firstName:"kaejgrkwe","lastName":"kaejgrkwe","email":"kaejgrkwe@gu.com","billingAddress":{"lineOne":null,"lineTwo":null,"city":null,"state":null,"postCode":null,"country":"GB"},"deliveryAddress":null,"product":{"productType":"Contribution","amount":20,"currency":"GBP","billingPeriod":"Monthly"},"firstDeliveryDate":null,"paymentFields":{"paymentMethod":"pm_0JcBsZItVxyc3Q6nCLSfvaXk","stripePaymentType":"StripeCheckout"},"ophanIds":{"pageviewId":"ktuao3s12md8dttf58ww","browserId":null,"visitId":null},"referrerAcquisitionData":{"queryParameters":[]},"supportAbTests":[{"name":"productSetTest","variant":"control"},{"name":"GBP_COUNTRIES_AMOUNTS_TEST","variant":"CONTROL"}],"telephoneNumber":null,"debugInfo":"contributions does not collect redux state"}
2021-09-21T16:38:40,122+0000 [application-akka.actor.default-dispatcher-18] ERROR com.gu.monitoring.SafeLogger$ - pushing alarm metric - non 2xx response Result(400, TreeMap())

```
and the client side gets something like this
```
< HTTP/1.1 400 Bad Request
[....]
ParsingFailure: expected : got 'kaejgr...' (line 1, column 14)
```

support-frontend/app/controllers/CreateSubscriptionController.scala
```diff
@@ -91,14 +91,6 @@ class CreateSubscriptionController(
...
       respondToClient(result, createSupportWorkersRequest.product.billingPeriod)
     } else {
-      val productName = createSupportWorkersRequest.product match {
-        case _: Contribution => "Contribution"
-        case _: DigitalPack => "DigitalPack"
-        case _: Paper => "Paper"
-        case _: GuardianWeekly => "GuardianWeekly"
-      }
-      val cloudwatchEvent = AwsCloudWatchMetricSetup.serverSideValidationFailure(stage, productName)
-      AwsCloudWatchMetricPut(AwsCloudWatchMetricPut.client)(cloudwatchEvent)
```
johnduffell, line 101: this seemed to have a metric there, but no alarm!  And it has triggered tens of times in the last 6 months in PROD, only for Digital subscriptions

support-frontend/cloud-formation/cfn.yaml
```diff
@@ -545,3 +545,30 @@ Resources:
...
+          - !FindInMap [ Constants, Alarm, Process ]
+      MetricName: ServerSideCreateFailure
+      Namespace: support-frontend
+      Dimensions:
+        - Name: Stage
+          Value: PROD
+      ComparisonOperator: GreaterThanOrEqualToThreshold
+      Threshold: 1
+      Period: 60
+      EvaluationPeriods: 1
```
johnduffell, line 572: this is the same as support-workers failure in prod, and should probably be reacted to similarly, in fact it's even more likely to be a "real" issue as test users shouldn't trigger it.

support-frontend/app/controllers/CreateSubscriptionController.scala
```diff
@@ -46,8 +46,8 @@ class CreateSubscriptionController(
 
   type ApiResponseOrError[RES] = EitherT[Future, CreateSubscriptionError, RES]
 
-  def create: Action[CreateSupportWorkersRequest] =
```
tomrf1, line 49: It's a shame to have lost the type info here, `EssentialAction` doesn't tell us much.
Do you think there's any value in separating it from the alarmOnFailure wrapping? E.g.
```
private def create: Action[CreateSupportWorkersRequest] = ...
def createWithAlarm = alarmOnFailure(create)
```

tomrf1 (approved):
looks good, besides a question about readability of the controllers

support-frontend/app/controllers/CreateSubscriptionController.scala
rupertbates: Why did you choose not to just alarm on this metric?

rupertbates (approved):
👍 LGTM

support-frontend/app/controllers/CreateSubscriptionController.scala
johnduffell: thanks, see what you mean. Actually from a type checking point of view I don't think the type matters at this point in terms of outward facing, other than as documentation.   This is because the JSON parsing happens "inside" the `Action`. What matters is that the `request` on line 51 is the right type, which would be checked elsewhere.
I'll have a think about what makes sense here though, as I agree it looks strange to not have that.

support-frontend/app/controllers/CreateSubscriptionController.scala
johnduffell: do you mean 1) why did I change the way the metric is triggered or 2) why did I rename the metric?
1) We didn't see any data coming in due to the outage as it was a JSON parse error.  The new metric is for any failure in the route, rather than just specifically our validation of the data content.  This means it should catch any issue here, including JSON parse error, so we would have seen some data coming in during the failure.
2) I renamed it because it specifically said "Validation" but we are covering all "create" errors however they happen.

support-frontend/app/controllers/CreateSubscriptionController.scala
johnduffell: so I had a look and thought a bit more and still don't see a benefit from the type info
however when I created the existing LoggingCirceParser I didn't really understand Actions and BodyParsers, now I've done a bit more, we could probably change 
```scala
alarmOnFailure(maybeAuthenticatedAction().async(new LoggingCirceParser(components).requestParser) {
```
to
```scala
alarmOnFailure(maybeAuthenticatedAction().async(loggingRequestParser[CreateSupportWorkersRequest]) {
```
or even
```scala
alarmOnFailureWithLogging(maybeAuthenticatedAction().async(circe.json[CreateSupportWorkersRequest]) {
```
but that last one might be another day or two of head scratching!

johnduffell:
I've merged for now, but I'm happy to do a followup PR for any further changes of course.

PR #3247, opened 2021-09-22
also alarm on failed futures on the create endpoints

johnduffell:

Follwing on from https://github.com/guardian/support-frontend/pull/3246
It became clear that failed futures can be returned by an action and cause a 5xx in place of the Result object.
This PR traps them too and also pushes an alarm if it fails.

## Why are you doing this?

We want to know when things are going wrong right at the end of the process of purchasing, so we can make the experience better. And avoid losing money.

### DIFF
--- a/support-frontend/app/actions/CustomActionBuilders.scala
+++ b/support-frontend/app/actions/CustomActionBuilders.scala
@@ -13,7 +13,7 @@ import play.filters.csrf._
 import services.{AsyncAuthenticationService, AuthenticatedIdUser}
 import utils.FastlyGEOIP
 
-import scala.concurrent.ExecutionContext
+import scala.concurrent.{ExecutionContext, Future}
 
 object CustomActionBuilders {
   type AuthRequest[A] = AuthenticatedRequest[A, AuthenticatedIdUser]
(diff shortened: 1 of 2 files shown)


support-frontend/app/services/HttpIdentityService.scala
```diff
@@ -180,7 +180,7 @@ class IdentityService(apiUrl: String, apiClientToken: String)(implicit wsClient:
       wsClient.url(s"$apiUrl/$endpoint")
         .withHttpHeaders(headers: _*)
         .withQueryStringParameters(parameters: _*)
-        .withRequestTimeout(1.second)
+        .withRequestTimeout(3.seconds)
```
johnduffell, line 183: we were getting occasional timeouts, and 1s seemed a bit tight for idapi responses.

support-frontend/app/actions/CustomActionBuilders.scala
```diff
@@ -94,7 +94,12 @@ class CustomActionBuilders(
       } else {
         result
       }
-    }
+    }.recoverWith({ case throwable: Throwable =>
+      SafeLogger.error(scrub"pushing alarm metric - 5xx response caused by ${throwable}")
+      val cloudwatchEvent = AwsCloudWatchMetricSetup.serverSideCreateFailure(stage)
+      AwsCloudWatchMetricPut(AwsCloudWatchMetricPut.client)(cloudwatchEvent)
+      Future.failed(throwable)
```
johnduffell, line 101: this makes it "recover" back to the original failure, as we're returning what was passed in.

rupertbates:
We have another alarm for support-frontend serving 500s don't we? "PROD support-frontend instances are returning 5XX errors" is adding this going to result in two alarms per 500?

johnduffell:
@rupertbates unlikely, because the existing 5xx alarm has a much higher threshold, because it covers all the random endpoints we have that we don't care too much about
https://github.com/guardian/support-frontend/blob/fa9466e1c3920b4b965fff7d4cb7021b200f54fd/support-frontend/cloud-formation/cfn.yaml#L422-L424
This one is only targetting the critical endpoint for actually paying, so it tells us it's urgent to look into

rupertbates (approved):
👍

PR #3248, opened 2021-09-22
don't submit stripe errors to the backend as a payment method

johnduffell:

This PR makes sure a stripe error is logged as a rejected promise rather than the error being used to resolve the promise.  This means it doesn't get submitted to the back end as a valid payment id.

https://trello.com/c/hR0Sn4K3/3870-dont-submit-client-side-stripe-error-as-payment-details-to-back-end

## Why are you doing this?

This is needed to reduce the number of alarms that go off, and make the error message clearer.

Before the user used to see:
image
and this extra request:
image

Now they will see:
image


i-hardy (approved):
👏 Good spot, let's get this merged

PR #3249, opened 2021-09-23
Paypal one click

rupertbates:
This PR tests adding a one click PayPal checkout option to the digital subscription product page.
There will be three variants; `control`, `guestCheckout` and `payPal`. 

The `guestCheckout` variant removes the requirement to sign in before purchasing a digital subscription, and will enable us to determine the potential uplift in conversion rate by removing friction from the flow 
The `payPal` variant also has guest sign in but in addition it adds a PayPal button to the digital subscriptions landing page which will enable users to purchase a sub without going through the checkout form. This will enable us to determine potential uplift from a 'one click' checkout option.

[**Test Plan**](https://docs.google.com/presentation/d/1B9wQI_5Vnttj8DpjYXu_OO6mD_2IgL521ohgsaxH7us/edit#slide=id.gef697baf35_6_0)
[**Trello Card**](https://trello.com/c/IuBSmC31/3858-implement-subscribe-in-1-click-with-paypal-on-the-product-page-a-b-test)

## Is this an AB test?
- [x] Yes
- [ ] No

[**Optimize Link**](https://optimize.google.com/optimize/home/#/accounts/1368065808/containers/6827913/experiments/504)

## Screenshots
*Hero section*
Screen Shot 2021-09-23 at 15 24 32
*Price cards*
Screen Shot 2021-09-23 at 15 24 51
*Error message*
Screen Shot 2021-09-23 at 15 23 59


support-frontend/app/controllers/DigitalSubscriptionController.scala
```diff
@@ -35,38 +38,47 @@ class DigitalSubscriptionController(
     if (orderIsAGift) "subscribe/digital/gift" else "subscribe/digital"
   )
 
-  def digital(countryCode: String, orderIsAGift: Boolean): Action[AnyContent] = CachedAction() { implicit request =>
-    implicit val settings: AllSettings = settingsProvider.getAllSettings()
+  def digital(countryCode: String, orderIsAGift: Boolean): Action[AnyContent] = {
+    maybeAuthenticatedAction() {
```
rupertbates, line 42: All that has changed in this file is that everything has been indented because of the `maybeAuthenticatedAction`

support-frontend/app/controllers/PayPalRegular.scala
```diff
@@ -54,6 +54,16 @@ class PayPalRegular(
     }
   }
 
+  def createAgreementAndRetrieveUser: Action[Token] = maybeAuthenticatedAction().async(circe.json[Token]) { implicit request =>
```
rupertbates, line 57: This is a new endpoint for the client side to hit, it retrieves the PayPal user details as well as the baid token which is needed by Zuora to bill the customer.

support-frontend/app/services/PayPalNvpService.scala
```diff
@@ -79,6 +80,7 @@ class PayPalNvpService(apiConfig: PayPalConfig, wsClient: WSClient) extends Touc
 
   // Sets up a payment by contacting PayPal and returns the token.
   def retrieveToken(returnUrl: String, cancelUrl: String)(billingDetails: PayPalBillingDetails): Future[Option[String]] = {
+    val noShipping = if(billingDetails.requireShippingAddress) "0" else "1"
```
rupertbates, line 83: We don't want to ask the user to pick an address when using PayPal on the regular checkouts, only when using the 'one click' button. This controls that.

support-frontend/app/views/subscriptionCheckout.scala.html
```diff
@@ -63,7 +63,6 @@
         default: "@defaultPayPalConfig.payPalEnvironment",
         uat: "@uatPayPalConfig.payPalEnvironment"
       };
-      window.guardian.csrf = { token: "@CSRF.getToken.value" };
```
rupertbates, line 66: This already exists in the `main` template so this was a duplicate

support-frontend/assets/components/paypalExpressButton/PayPalExpressButton.jsx
```diff
@@ -9,14 +9,12 @@ import { connect } from 'react-redux';
...
 import { type IsoCurrency } from 'helpers/internationalisation/currency';
-import { type PayPalAuthorisation } from 'helpers/forms/paymentIntegrations/readerRevenueApis';
 import type { BillingPeriod } from 'helpers/productPrice/billingPeriods';
-import { PayPal } from 'helpers/forms/paymentMethods';
 import { type Action, updatePayPalButtonReady } from 'pages/contributions-landing/contributionsLandingActions';
 import AnimatedDots from 'components/spinners/animatedDots';
 
 type PropTypes = {|
-  onPaymentAuthorisation: Function,
+  onPayPalCheckoutCompleted: Function,
```
rupertbates, line 17: Renamed to make the ordering clearer

support-frontend/app/controllers/DigitalSubscriptionController.scala
johnduffell: if you view with this link, the whitespace is ignored, so it's easier to review https://github.com/guardian/support-frontend/pull/3249/files?w=1

support-frontend/app/controllers/PayPalRegular.scala
```diff
@@ -54,6 +54,16 @@ class PayPalRegular(
...
+    withPaypalServiceForRequest(request) { service =>
+      service.createAgreementAndRetrieveUser(request.body)
+    }.map { maybePayPalCheckoutDetails =>
+      maybePayPalCheckoutDetails
+        .map(details => Ok(details.asJson))
+        .getOrElse(BadRequest("We were unable to create an agreement for this request (missing user details or baid)"))
```
johnduffell, line 63: if this endpoint fails do we need to alarm?

support-frontend/app/services/PayPalNvpService.scala
johnduffell: lint missing space
```suggestion
    val noShipping = if (billingDetails.requireShippingAddress) "0" else "1"
```

support-frontend/app/services/PayPalNvpService.scala
```diff
@@ -59,7 +60,7 @@ class PayPalNvpService(apiConfig: PayPalConfig, wsClient: WSClient) extends Touc
 
   // Takes an NVP response and retrieves a given parameter as a string.
   private def retrieveNVPParam(response: QueryString, paramName: String) =
-    response.paramMap(paramName).headOption match {
+    response.paramMap.getOrElse(paramName, Nil).headOption match {
```
johnduffell, line 63: did this change to be an Option[List[ ... ]] for some reason?

support-frontend/app/services/PayPalNvpService.scala
```diff
@@ -90,14 +92,48 @@ class PayPalNvpService(apiConfig: PayPalConfig, wsClient: WSClient) extends Touc
...
-      "NOSHIPPING" -> "1"
+      "NOSHIPPING" -> noShipping
     )
 
     nvpRequest(paymentParams).map { resp =>
       retrieveNVPParam(resp, "TOKEN")
     }
   }
 
+  // Sets up a payment by contacting PayPal and also fetches the users details.
```
johnduffell, line 103: I vaguely remember we were using a very out of date paypal something, was it the library here or something else?

support-frontend/app/services/PayPalNvpService.scala
```diff
@@ -90,14 +92,48 @@ class PayPalNvpService(apiConfig: PayPalConfig, wsClient: WSClient) extends Touc
...
+  def createAgreementAndRetrieveUser(token: Token): Future[Option[PayPalCheckoutDetails]] =
+    createBillingAgreement(token).flatMap(
+      maybeBaid =>
+      retrieveUserInformation(token).map(maybeUserDetails =>
+        maybeBaid match {
+          case Some(baid) => Some(PayPalCheckoutDetails(baid, maybeUserDetails))
+          case _ => None
+        }
+      )
+    )
```
johnduffell, line 113: is this the same as this:
```suggestion
    for {
      maybeBaid <- createBillingAgreement(token)
      maybeUserDetails <- retrieveUserInformation(token)
    } yield maybeBaid.map(baid => PayPalCheckoutDetails(baid, maybeUserDetails))
```

support-frontend/app/services/PayPalNvpService.scala
```diff
@@ -90,14 +92,48 @@ class PayPalNvpService(apiConfig: PayPalConfig, wsClient: WSClient) extends Touc
...
+      for {
+        firstName <- retrieveNVPParam(resp, "FIRSTNAME")
+        lastName <- retrieveNVPParam(resp, "LASTNAME")
+        email <- retrieveNVPParam(resp, "EMAIL")
+        shipToStreet <- retrieveNVPParam(resp, "PAYMENTREQUEST_0_SHIPTOSTREET")
+        shipToCity <- retrieveNVPParam(resp, "PAYMENTREQUEST_0_SHIPTOCITY")
+        shipToState <- retrieveNVPParam(resp, "PAYMENTREQUEST_0_SHIPTOSTATE")
+        shipToZip <- retrieveNVPParam(resp, "PAYMENTREQUEST_0_SHIPTOZIP")
+        shipToCountryCode <- retrieveNVPParam(resp, "PAYMENTREQUEST_0_SHIPTOCOUNTRYCODE")
+      } yield PayPalUserDetails(
```
johnduffell, line 131: if any are missing then we will get a None. is there any logging that's going to let us know what's going on, before we return a BadRequest?

support-frontend/app/services/PayPalNvpService.scala
rupertbates: It is a map returning a vector, so we should have always been calling getOrElse. What was happening was that for missing params it was throwing an element not found exception instead of being handled properly by the match statement. This fixes that.

support-frontend/app/services/PayPalNvpService.scala
rupertbates: It is the framework that is out of date not a specific library. Unfortunately we can't upgrade because Zuora doesn't support any other method of integration (as far as I'm aware, we could re investigate this I suppose)

support-frontend/app/services/PayPalNvpService.scala
rupertbates: No I don't think it is. In the case where we have a baid but no user information (this will happen for contributions for instance) the first piece of code will return 
`Some(PayPalCheckoutDetails(String, None))`
the second will return `None`.

support-frontend/app/services/PayPalNvpService.scala
rupertbates: Good point I've added some logging

support-frontend/app/controllers/PayPalRegular.scala
rupertbates: We never have done previously. Currently the user will see this error message:
Screen Shot 2021-09-27 at 17 45 13
I think that's sufficient?

support-frontend/app/services/PayPalNvpService.scala
johnduffell: I don't think the second will return None.  It would have if I used an OptionT, but I could see that behaviour wasn't what was implemented.

support-frontend/app/services/PayPalNvpService.scala
johnduffell: got it, that's useful information 👍

support-frontend/app/controllers/PayPalRegular.scala
johnduffell: for the AB test it probably is okay, but we should return a specific error message if there is a known case where the user needs to take action. And for everything else, we should be aware of the issue and look into it.  This comment is mainly in the light of the bugs we found this last week where people are seeing this error message due to gaps in the logic.

support-frontend/assets/components/paypalExpressButton/PayPalOneClickCheckoutButton.jsx
```diff
@@ -0,0 +1,139 @@
...
+        errorHeading={submissionErrorHeading}
+      />}
+      <PayPalExpressButton
+        onPayPalCheckoutCompleted={props.onPayPalCheckoutCompleted}
+        csrf={props.csrf}
+        currencyId={props.currencyId}
+        hasLoaded={props.hasLoaded}
+        canOpen={() => true}
+        onClick={() => props.onClick(props.billingPeriod, props.trackingId, props.product)}
+        formClassName="form--contribution"
```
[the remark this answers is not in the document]
rupertbates: Yes I don't really know why it's there. I've removed it.

support-frontend/app/services/PayPalNvpService.scala
```diff
@@ -90,14 +92,51 @@ class PayPalNvpService(apiConfig: PayPalConfig, wsClient: WSClient) extends Touc
...
+        shipToStreet <- retrieveNVPParam(resp, "PAYMENTREQUEST_0_SHIPTOSTREET")
+        shipToCity <- retrieveNVPParam(resp, "PAYMENTREQUEST_0_SHIPTOCITY")
+        shipToState = retrieveNVPParam(resp, "PAYMENTREQUEST_0_SHIPTOSTATE") // State/County may not be present
+        shipToZip <- retrieveNVPParam(resp, "PAYMENTREQUEST_0_SHIPTOZIP")
+        shipToCountryCode <- retrieveNVPParam(resp, "PAYMENTREQUEST_0_SHIPTOCOUNTRYCODE")
+      } yield PayPalUserDetails(
+        firstName, lastName, email, shipToStreet, shipToCity, shipToState, shipToZip, shipToCountryCode
+      )
+      if (userDetails.isEmpty)
+        SafeLogger.error(scrub"Unable to retrieve user details from PayPal with token $token, nvp response was $resp")
```
johnduffell, line 135: I think you said that for contributions this will be expected to be None?  So we will end up with a lot of worrying log spam if we use it there in future.
I would be tempted to return an Either and then log it in the caller at the point we deem that it's an actual error.

support-frontend/assets/components/paypalExpressButton/PayPalOneClickCheckoutButton.jsx
```diff
@@ -0,0 +1,140 @@
...
+  dispatch(setEmail(payPalUserDetails.email));
+  dispatch(setFirstName(payPalUserDetails.firstName));
+  dispatch(setLastName(payPalUserDetails.lastName));
+  dispatch(setAddressLineOne(payPalUserDetails.shipToStreet));
+  dispatch(setTownCity(payPalUserDetails.shipToCity));
+  if (payPalUserDetails.shipToState) {
+    dispatch(setState(payPalUserDetails.shipToState));
+  }
+  dispatch(setPostcode(payPalUserDetails.shipToZip));
+  // $FlowIgnore stoopid flow
```
johnduffell, line 58: not for much longer!

support-frontend/assets/components/paypalExpressButton/PayPalOneClickCheckoutButton.jsx
```diff
@@ -0,0 +1,140 @@
...
+}
+
+const payPalButton = css`
+  box-sizing: border-box;
+  margin-top: ${space[3]}px;
+`;
+
+function PayPalOneClickCheckoutButton(props: PropTypes) {
+  const submissionErrorHeading = 'Sorry there was a problem';
+  const errorReason = props.submissionError && 'amazon_pay_fatal';
```
johnduffell, line 116: why amazon pay?

rupertbates: It has the right message and I thought I'd just reuse it given this is a test

support-frontend/assets/components/paypalExpressButton/PayPalOneClickCheckoutButton.jsx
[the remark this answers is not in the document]
rupertbates: Actually I'm going to add it back in because it's a required prop and I even though it doesn't look like it's actually used, I don't want to get into removing it in this PR

support-frontend/app/services/PayPalNvpService.scala
rupertbates: Oh yes I see what you're saying, sorry I had OptionT in my head.

support-frontend/app/services/PayPalNvpService.scala
rupertbates: Yeah this is a bit difficult to do unfortunately because all the different PayPal call sites call the same function and we don't know on the server when it's ok for there not to be user and when it's required. Again, this is a test and we're not expecting to leave it in place so I'm just going to remove this logging for now. If we include something like this in future we'll need to have a look at refactoring.

GHaberis (approved):
I've reviewed the client-side code and it all looks good to me 👍

PR #3250, opened 2021-09-24
Remove querystring switch for SEPA

tomrf1:
currently you can only see SEPA if you add `?sepa=true` to the url.
This PR removes this.
After this PR is merged all EUR users will see SEPA. The RRCP switchboard has a switch for SEPA if we need to disable it.

To be merged on Monday.


TomPretty (approved):
LGTM 👍

PR #3252, opened 2021-09-27
fix ipaddress field in zuora request

tomrf1:
A sepa recurring contribution failed at the zuora step with:
`The length of IP Address can not be longer than 15 characters`
This doesn't work for ipv6

### DIFF
--- a/support-workers/src/main/scala/com/gu/support/workers/lambdas/CreatePaymentMethod.scala
+++ b/support-workers/src/main/scala/com/gu/support/workers/lambdas/CreatePaymentMethod.scala
@@ -150,7 +150,7 @@ class CreatePaymentMethod(servicesProvider: ServiceProvider = ServiceProvider)
       BankTransferAccountName = sepa.accountHolderName,
       BankTransferAccountNumber = sepa.iban,
       Email = user.primaryEmailAddress,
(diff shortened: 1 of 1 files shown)


support-workers/src/main/scala/com/gu/support/workers/lambdas/CreatePaymentMethod.scala
```diff
@@ -150,7 +150,7 @@ class CreatePaymentMethod(servicesProvider: ServiceProvider = ServiceProvider)
       BankTransferAccountName = sepa.accountHolderName,
       BankTransferAccountNumber = sepa.iban,
       Email = user.primaryEmailAddress,
-      IPAddress = ipAddress,
+      IPAddress = ipAddress.take(15), // zuora doesn't support ipv6
```
TomPretty, line 153: It feels a little weird to give them a truncated ipv6 string but I guess there's nothing else we really can do

tomrf1: Indeed. I'll raise a ticket with zuora later

TomPretty (approved):
Interested to hear what zuora have to say here - they need to get with the times! 👍

PR #3253, opened 2021-09-28
Fix PRB display issue

TomPretty:
Fix PRB display issue. When toggling between `Single` -> `Recurring` the `PaymentRequestButton` component is recreated. However, we only set the `prbType` the first time (during `initialisePaymentRequest`). This meant on future renders `prbType === 'NONE'` which meant we didn't display the `Apple Pay`/`Google Pay` buttons when we should have (it falls back to our custom PRB in this case).

[**Trello Card**](https://trello.com/c/OrCQkGoG/2795-fix-issues-with-stripe-payment-request-button).

**NB**: This does cause a slight delay in rendering the button as we need to call `canMakePayment` inside a `useEffect` hook. 

A potentially better solution (but one requiring more work) would be to update [this type](https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/helpers/forms/paymentIntegrations/readerRevenueApis.js#L42) to add e.g `StripeGooglePay` as we already have this data in redux. This would simplify the `PaymentRequestButton` component, but add complexity to support this new payment method. We would need to:

1. switch off the ophan route for acquisitions, which just needs to be done!
2. decide if we want to start track google pay separately from PRB. This has implications for dashboards though
3. do we need a separate switch for google pay? probably not


tomrf1 (approved):
good spot!

PR #3232, opened 2021-09-06
Guest checkout test

rupertbates:
## What are you doing in this PR?
This PR 
- activates the guest checkout test on the newspaper checkout
- adds a 'confirm email' input onto the checkout when it is in guest mode to catch typos in the email address
- TODO: updates the thank you page to explain next steps to the user
- excludes CSRs from being included in the AB test but ensures that they will always go via the guest checkout

[**Trello Card**](https://trello.com/c/LnR9QOqv/3854-activate-guest-checkout-test)
[**Test document**](https://docs.google.com/document/d/1VecRQ91a1ezwb0WSzj0JVqN00e5-cn7G2JbfLCtXD5I/edit#)

## Is this an AB test?
- [x] Yes
- [ ] No

If this is an AB test, PR reviewers should open and check the Optimize test.
[**Optimize Link**](https://optimize.google.com/optimize/home/#/accounts/1368065808/containers/6827913/experiments/496)

## Accessibility test checklist
 - [ ] [Tested with screen reader](https://accessibility.gutools.co.uk/testing/web/screen-readers/)
 - [ ] [Navigable with keyboard](https://accessibility.gutools.co.uk/testing/web/keyboard-navigation/)
 - [ ] [Colour contrast passed](https://accessibility.gutools.co.uk/testing/web/colour-contrast/)

## Screenshots
**Guest user checkout with confirm email box:**
Screen Shot 2021-09-07 at 10 54 24
**Guest user checkout with mismatched email error message:**
Screen Shot 2021-09-07 at 10 55 34

PR #3254, opened 2021-09-28
Remove newspaper guest checkout test

rupertbates:

This PR removes the [ab test for newspaper guest checkout](https://github.com/guardian/support-frontend/pull/3232) and rolls guest checkout out to 100% of users.

[**Trello Card**](https://trello.com/c/j4hh5Kw4/3887-release-the-newspaper-guest-checkout-to-100-of-the-audience-stop-test)

## Is this an AB test?
- [ ] Yes
- [x] No

[**Link to the test results in Optimize**](https://optimize.google.com/optimize/home/#/accounts/1368065808/containers/6827913/experiments/496/report)

### DIFF
--- a/support-frontend/app/controllers/PaperSubscriptionForm.scala
+++ b/support-frontend/app/controllers/PaperSubscriptionForm.scala
@@ -40,21 +40,7 @@ class PaperSubscriptionForm(
   implicit val a: AssetsResolver = assets
 
 
-  def displayForm(): Action[AnyContent] =
-    authenticatedAction(subscriptionsClientId).async { implicit request =>
-      implicit val settings: AllSettings = settingsProvider.getAllSettings()
-      identityService.getUser(request.user.minimalUser).fold(
-        error => {
-          SafeLogger.error(scrub"Failed to display paper subscriptions form for ${request.user.minimalUser.id} due to error from identityService: $error")
-          Future.successful(InternalServerError)
-        },
-        user => {
-          Future.successful(Ok(paperSubscriptionFormHtml(Some(user))))
-        }
-      ).flatten.map(_.withSettingsSurrogateKey)
-    }
-
-  def displayGuestForm(): Action[AnyContent] = {
+  def displayForm(): Action[AnyContent] = {
     implicit val settings: AllSettings = settingsProvider.getAllSettings()
     maybeAuthenticatedAction().async{ implicit request =>
         val maybeIdUser: EitherT[Future, String, Option[IdUser]] = request.user match {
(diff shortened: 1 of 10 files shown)


support-frontend/app/controllers/PaperSubscriptionForm.scala
```diff
@@ -40,21 +40,7 @@ class PaperSubscriptionForm(
   implicit val a: AssetsResolver = assets
 
 
-  def displayForm(): Action[AnyContent] =
```
rupertbates, line 43: We don't need two separate urls for regular and guest checkout any more so this removes the one called `displayGuestForm` and changes `displayForm` to show the guest checkout.

support-frontend/assets/pages/paper-subscription-checkout/components/paperCheckoutForm.jsx
```diff
@@ -313,7 +311,7 @@ function PaperCheckoutForm(props: PropTypes) {
               setTelephone={props.setTelephone}
               formErrors={props.formErrors}
               signOut={props.signOut}
-              isUsingGuestCheckout={isUsingGuestCheckout}
+              isUsingGuestCheckout
```
rupertbates, line 314: We still need to pass this prop for the time being because this component is shared between checkouts

support-frontend/assets/pages/weekly-subscription-checkout/components/weeklyCheckoutForm.jsx
```diff
@@ -182,7 +182,7 @@ function WeeklyCheckoutForm(props: PropTypes) {
     props.setBillingCountry(props.deliveryCountry);
   };
   const paymentMethods = supportedPaymentMethods(props.currencyId, props.billingCountry);
-  const isUsingGuestCheckout = props.participations.subscriptionsGuestCheckoutTest === 'variant';
+  const isUsingGuestCheckout = false;
```
rupertbates, line 185: Roll this out to GW in a separate PR

johnduffell (approved):
great,:+👍 
 handy to have the link to optimise, but will it always work?
Just to confirm that it shows the variant and control are both equally likely to be better.
This means we can ship it because we want to confirm no negative effect on conversions, but we do want to implement it for other reasons.

rupertbates:
> great,:+👍 handy to have the link to optimise, but will it always work? Just to confirm that it shows the variant and control are both equally likely to be better. This means we can ship it because we want to confirm no negative effect on conversions, but we do want to implement it for other reasons.

Yes we do want to ship it for security purposes as well, it is strategic for the Identity team.

PR #3255, opened 2021-09-29
Unset title field to null rather than empty string in checkout

rupertbates:
We have a bug caused by users setting a title and then unsetting it which results in a value of empty string rather than null if it was never set in the first place. 

This causes problems when deserialising the json body of the create request to the backend.

[**Trello Card**](https://trello.com/c/2VqcjyCh/3881-title-field-as-an-empty-string-does-not-pass-server-side-validation-on-the-susbscription-create-endpoint)

## Is this an AB test?
- [ ] Yes
- [x] No


johnduffell (approved):
great work on fixing so quickly

PR #3260, opened 2021-09-29
better name for alarm

johnduffell:

This PR makes the alarm name and desc more accurate

## Why are you doing this?

@GHaberis pointed out it is misleading, and he's right


GHaberis (approved):
👍

tomrf1 (approved):

PR #3261, opened 2021-09-30
Stop sending acquisitions to ophan

tomrf1:
Currently, support-workers and payment-api both use the acquisition-event-producer module to publish events to ophan and GA. We no longer need to send events to ophan as they are now sent directly to BigQuery. But we do still need to send events to GA.

The old acquisition-event-producer module is being replaced by the acquisition-events module. The latter already sends events to BigQuery and the kinesis stream (used by super mode). It also uses a scala model instead of the ophan thrift model, and doesn't use complicated typeclasses.

1. Migrates the GA service from acquisition-event-producer to the acquisition-events module. The new service takes the scala acquisition model, rather than the thrift.
2. Removes the `AnalyticsService` from payment-api, which now uses the new GA service.
3. Removes the `AnalyticsService` from support-workers. Deletes the `SendOldAcquisitionEvent` lambda, and updates the `SendAcquisitionEvent` to use the new GA service

We're not quite ready to delete the acquisition-event-producer because its models (which use the thrift model) are still used in support-workers and support-frontend. But this PR is huge enough, so it'll be a follow up (see https://github.com/guardian/support-frontend/pull/3268).


support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/ga/GoogleAnalyticsServiceImpl.scala
```diff
@@ -0,0 +1,255 @@
...
+import java.io.IOException
+import java.util.UUID
+import scala.concurrent.{ExecutionContext, Future, Promise}
+import scala.util.control.NonFatal
+import scala.util.matching.Regex
+
+object GoogleAnalyticsServiceImpl {
+  val clientIdPattern: Regex = raw"GA\d\.\d\.(\d+\.\d+)".r
+
+  // e.g. GUARDIAN_WEEKLY becomes GuardianWeekly
```
tomrf1, line 25: this is for where we used to use the `name` field on the thrift models.
E.g. `acquisition.paymentFrequency.name` is `OneOff`

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/ga/models/ConversionCategory.scala
```diff
@@ -0,0 +1,23 @@
+package com.gu.support.acquisitions.ga.models
+
+sealed trait ConversionCategory {
```
tomrf1, line 3: copied from acquisition-event-producer

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/ga/models/GAData.scala
```diff
@@ -0,0 +1,25 @@
+package com.gu.support.acquisitions.ga.models
```
tomrf1, line 1: copied from acquisition-event-producer

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/ga/models/GAError.scala
```diff
@@ -0,0 +1,24 @@
+package com.gu.support.acquisitions.ga.models
```
tomrf1, line 1: copied from acquisition-event-producer

support-workers/src/main/scala/com/gu/support/workers/lambdas/SendAcquisitionEvent.scala
```diff
@@ -50,17 +48,39 @@ class SendAcquisitionEvent(serviceProvider: ServiceProvider = ServiceProvider)
 
   }
 
+  private def buildGaData(state: SendAcquisitionEventState, requestInfo: RequestInfo): Either[String, GAData] = {
```
tomrf1, line 51: adapated from the now-deleted `SendOldAcquisitionEvent.scala`

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/ga/GoogleAnalyticsServiceImpl.scala
```diff
@@ -0,0 +1,255 @@
+package com.gu.support.acquisitions.ga
```
tomrf1, line 1: This file is similar to the original GAService, but it receives the scala acquisitions model instead of thrift

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/ga/GoogleAnalyticsServiceImpl.scala
johnduffell: I don't understand this.  Do you mean, the thrift models used to magically convert the case, and now you have written a function to repliacate that?

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/ga/GoogleAnalyticsServiceImpl.scala
```diff
@@ -0,0 +1,255 @@
...
+      .fold(
+        error => {
+          logger.warn(s"Couldn't retrieve annualised value for this acquisition: $error")
+          buildPayload(acquisition, 0, gaData) // We still want to record the acquisition even if we can't get the AV
+        },
+        value => buildPayload(acquisition, value, gaData)
+      )
+      .map { maybePayload =>
+        maybePayload.map { payload =>
+          logger.debug(s"GA payload: $payload")
```
johnduffell, line 56: does debug appear anywhere? or would this be better as Info?

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/ga/GoogleAnalyticsServiceImpl.scala
```diff
@@ -0,0 +1,255 @@
...
+    getAnnualisedValue(acquisition)
+      .fold(
+        error => {
+          logger.warn(s"Couldn't retrieve annualised value for this acquisition: $error")
+          buildPayload(acquisition, 0, gaData) // We still want to record the acquisition even if we can't get the AV
+        },
+        value => buildPayload(acquisition, value, gaData)
+      )
+      .map { maybePayload =>
+        maybePayload.map { payload =>
```
johnduffell, line 55: for clarity, there's no need to repeat the buildPayload call
I'm not totally sure of the types here, but something like this would be equivalent
```suggestion
      .fold(
        error => {
          logger.warn(s"Couldn't retrieve annualised value for this acquisition: $error")
         0 // We still want to record the acquisition even if we can't get the AV
        },
        value => value
      )
      .map { maybeValue =>
        buildPayload(acquisition,maybeValue, gaData).map { payload =>
```

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/ga/GoogleAnalyticsServiceImpl.scala
```diff
@@ -0,0 +1,255 @@
...
+        )
+        EitherT(AnnualisedValueService.getAsyncAV(acquisitionModel, "ophan"))
+
+      case None => EitherT.fromEither[Future](Left("No amount for acquisition"))
+    }
+  }
+
+  private[ga] def getSuccessfulSubscriptionSignUpMetric(conversionCategory: ConversionCategory) =
+    conversionCategory match {
+      case _: ConversionCategory.ContributionConversion.type => ""
```
johnduffell, line 80: since it's a case object, this should work
```suggestion
      case ConversionCategory.ContributionConversion => ""
```

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/ga/GoogleAnalyticsServiceImpl.scala
```diff
@@ -0,0 +1,255 @@
...
+        val acquisitionModel = AcquisitionModel(amount.toDouble,
+          acquisition.product.value,
+          acquisition.currency.iso,
+          acquisition.paymentFrequency.value,
+          acquisition.paymentProvider.map(_.toString),
+          acquisition.printOptions.map(printOptions => PrintOptionsModel(printOptions.product.value, printOptions.deliveryCountry.alpha2))
+        )
+        EitherT(AnnualisedValueService.getAsyncAV(acquisitionModel, "ophan"))
+
+      case None => EitherT.fromEither[Future](Left("No amount for acquisition"))
```
johnduffell, line 74: what's the difference between "amount" and "annualised value"?

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/ga/GoogleAnalyticsServiceImpl.scala
tomrf1: Exactly!
The scrooge-generated scala classes look like this:
```
case object OneOff extends ophan.thrift.event.PaymentFrequency {
    val value: Int = 1
    val name: String = "OneOff"
    val originalName: String = "ONE_OFF"
    val annotations: immutable$Map[String, String] = immutable$Map.empty
  }
```

So wherever we were using `name` for GA, I've had to camel case, because the new scala model uses e.g. `ONE_OFF`

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/ga/GoogleAnalyticsServiceImpl.scala
tomrf1: I suspect not! This was copied across from the old GAService

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/ga/GoogleAnalyticsServiceImpl.scala
```diff
@@ -0,0 +1,255 @@
...
+      case _ => Right(maybeId) // Otherwise assume that the caller has passed in the client id correctly
+    }
+  }
+
+  private[ga] def getProductCheckout(acquisition: AcquisitionDataRow): Option[String] =
+    acquisition.product match {
+      case Contribution | RecurringContribution => Some("Contribution")
+      case DigitalSubscription => Some("DigitalPack")
+      case Paper => getProductCheckoutForPrint(acquisition.printOptions)
+      case default => None
```
johnduffell, line 162: shoudl we have added GuardianWeekly to here at some point?

johnduffell, line 162: I can actually see Guardian weekly was added below - so what other products are possible?

johnduffell (approved):
nice work! and handy to have broken it down a bit from a reviewing point of view.
 most of my comments were on code I think is probably existing, so don't worry too much about them, but it's a good chance to look at some old code! 👍

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/ga/GoogleAnalyticsServiceImpl.scala
tomrf1: It's an odd one because `getAnnualisedValue` returns an `EitherT`, `fold` gives us just a `Future`, then we end up wrapping the result in an `EitherT` anyway!
I'm just having a play...

tomrf1: I've pushed this up

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/ga/GoogleAnalyticsServiceImpl.scala
tomrf1: Good question. I'm not sure I understand the existing logic.
The original thrift `Product` type had the following:
```
    ophan.thrift.event.Product.Contribution,
    ophan.thrift.event.Product.RecurringContribution,
    ophan.thrift.event.Product.MembershipSupporter,
    ophan.thrift.event.Product.MembershipPatron,
    ophan.thrift.event.Product.MembershipPartner,
    ophan.thrift.event.Product.DigitalSubscription,
    ophan.thrift.event.Product.PaperSubscriptionEveryday,
    ophan.thrift.event.Product.PaperSubscriptionSixday,
    ophan.thrift.event.Product.PaperSubscriptionWeekend,
    ophan.thrift.event.Product.PaperSubscriptionSunday,
    ophan.thrift.event.Product.PrintSubscription,
    ophan.thrift.event.Product.AppPremiumTier
```
So `GuardianWeekly` wasn't in this set.
With the new model, it is:
```
  case object Contribution extends AcquisitionProduct("CONTRIBUTION")
  case object RecurringContribution extends AcquisitionProduct("RECURRING_CONTRIBUTION")
  case object DigitalSubscription extends AcquisitionProduct("DIGITAL_SUBSCRIPTION")
  case object Paper extends AcquisitionProduct("PRINT_SUBSCRIPTION")
  case object GuardianWeekly extends AcquisitionProduct("PRINT_SUBSCRIPTION")
```

But `GuardianWeekly` is also a valid `PrintProduct`, so I'm not sure what the correct behaviour should be here.

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/ga/GoogleAnalyticsServiceImpl.scala
```diff
@@ -0,0 +1,255 @@
...
+          "pr1pr" -> annualisedValue, // Product Price - we are tracking the annualised value here as this is what goes into the revenue metric
+          "pr1qt" -> "1", // Product Quantity
+          "pr1cc" -> acquisition.promoCode.getOrElse(""), // Product coupon code.
+          "pr1cm15" -> acquisition.amount.toString, // Custom metric 15 - purchasePrice
+          "cu" -> acquisition.currency.toString // Currency
+        )
+
+        body
+          .filter { case (key, value) => value != "" }
+          .map { case (key, value) => s"$key=$value" }
```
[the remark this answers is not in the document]
tomrf1: It is, but that's a good point. This is building the request to GA. But I wonder about the other acquisition services as well - BigQuery?

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/ga/GoogleAnalyticsServiceImpl.scala
tomrf1: I've pushed up a possible solution: https://github.com/guardian/support-frontend/pull/3261/commits/83115e09958aa0ddb9dfd39fc4f86799e3d6fb0a

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/ga/GoogleAnalyticsServiceImpl.scala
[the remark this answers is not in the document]
johnduffell: yeah could be in so many places.  On a positive side, generally anything that causes an action somewhere important e.g. Zuora, we would use a POST and a proper JSON serialiser, but building URLs is too tempting where it comes up.
We do use https://github.com/guardian/support-frontend/search?q=scala-uri in places which is a proper url serialiser and I'm sure there are others.

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/ga/GoogleAnalyticsServiceImpl.scala
johnduffell: that looks much better, however we are returning an Option, which breaks the types I think.  I will go and comment on the relevant line.

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/ga/GoogleAnalyticsServiceImpl.scala
```diff
@@ -0,0 +1,255 @@
...
+          "pr1nm" -> productName, // Product Name
+          "pr1ca" -> conversionCategory.description, // Product category
+          "pr1pr" -> annualisedValue, // Product Price - we are tracking the annualised value here as this is what goes into the revenue metric
+          "pr1qt" -> "1", // Product Quantity
+          "pr1cc" -> acquisition.promoCode.getOrElse(""), // Product coupon code.
+          "pr1cm15" -> acquisition.amount.toString, // Custom metric 15 - purchasePrice
+          "cu" -> acquisition.currency.toString // Currency
+        )
+
+        body
```
johnduffell, line 137: body is a Map[String, Any] which means we are comparing an Any with `""` and calling toString on something of unknown type, which is probably not what we want.

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/ga/GoogleAnalyticsServiceImpl.scala
johnduffell: looks good. 👍

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/ga/GoogleAnalyticsServiceImpl.scala
```diff
@@ -0,0 +1,255 @@
...
+    val tid = transactionId.getOrElse(UUID.randomUUID().toString)
+    val goExp = buildOptimizeTestsPayload(acquisition.abTests)
+
+    // clientId cannot be empty or the call will fail
+    sanitiseClientId(gaData.clientId).map {
+      clientId =>
+        val productName = getProductName(acquisition)
+        val conversionCategory = getConversionCategory(acquisition)
+        val productCheckout = getProductCheckout(acquisition)
+        val body = Map(
```
johnduffell, line 92:
```suggestion
        val body = Map[String, String](
```
adding that in intellij shows the first line which has the issues so it's easier to resolve, there are actually a few

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/ga/GoogleAnalyticsServiceImpl.scala
tomrf1: this is the case with the original code, there are a few bools and an option. I don't think it would hurt to stringify these to make it a `Map[String,String]

johnduffell: maybe not, and if it matches the existing functionality we can keep it.  personally I'd prefer to have the toString explicitly on the relevant lines, and also not to toString scala types to put into external systems in case they are overridden or refactored to use another type later.
Regarding the option, actually I was thinking about whether we can get rid of it completely, but if we are relying on the downstream being `"Some(blah)"` then that may not be possible

tomrf1: I've pushed up a change to stringify those values. I've defaulted the option to `""`, which is what we do in various other places

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/ga/GoogleAnalyticsServiceImpl.scala
tomrf1: amount is the price of the contribution/subscription. Annualised value is the expected value of 12 months

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/ga/GoogleAnalyticsServiceImpl.scala
johnduffell: perfect (as long as it's not going to break existing consumers of the GA data)

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/ga/GoogleAnalyticsServiceImpl.scala
johnduffell: got it, so the getAsyncAV is what converts the input data into an AV, and we would always expect the amount to exist. 👍

johnduffell (approved):
great work and it's great to improve the code a bit on the way through, thanks for doing thist 👏

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/ga/GoogleAnalyticsServiceLive.scala
```diff
@@ -206,6 +198,25 @@ class GoogleAnalyticsServiceImpl(client: OkHttpClient) extends GoogleAnalyticsSe
...
       .mkString(",")
+}
+
+// For test specs, does nothing
+object GoogleAnalyticsServiceMock extends GoogleAnalyticsService {
+  def submit(acquisition: AcquisitionDataRow, gaData: GAData, maxRetries: Int)(implicit ec: ExecutionContext): EitherT[Future, List[GAError], Unit] =
+    EitherT.fromEither(Right(()))
+}
+
+// For test conversions, logs the payload but doesn't send to GA
```
johnduffell, line 209: good idea, the fewer differences between test and normal, the better

support-workers/src/main/scala/com/gu/support/workers/lambdas/SendAcquisitionEvent.scala
```diff
@@ -50,17 +48,39 @@ class SendAcquisitionEvent(serviceProvider: ServiceProvider = ServiceProvider)
...
   private def sendAcquisitionEvent(state: SendAcquisitionEventState, requestInfo: RequestInfo, services: Services) = {
     sendPaymentSuccessMetric(state)
 
     val acquisition = AcquisitionDataRowBuilder.buildFromState(state, requestInfo)
 
     val streamFuture = services.acquisitionsStreamService.putAcquisitionWithRetry(acquisition, maxRetries = 5)
     val biqQueryFuture = services.bigQueryService.tableInsertRowWithRetry(acquisition, maxRetries = 5)
+    val gaFuture = for {
+      gaData <- EitherT.fromEither(buildGaData(state, requestInfo)).leftMap(err => List(err))
+      result <- services.gaService.submit(acquisition, gaData, maxRetries = 5).leftMap(gaErrors => gaErrors.map(_.getMessage))
```
johnduffell, line 77: we had no idea at the time, but it turns out the acquisition from the `AcquisitionDataRowBuilder` labels are subtley different in terms of upper/lower case from the ones in SendOldAcqisitionState buildLabels function, which broke the gift reporting in GA as it checks the exact string
https://github.com/guardian/support-frontend/blob/13b0de701b1f7b6ef81a8b6ea61df8701f9b3211/support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/ga/GoogleAnalyticsService.scala#L95

PR #3263, opened 2021-09-30
update footer faq link to the help centre

johnduffell:

This PR updates the footer FAQs link to point to the Help Centre.

## Why are you doing this?

We got this message pointing out the change needed

>I've noticed that through the footer on the support.theguardian.com platform you can access this set of FAQs.

>This should really be pointing at the help centre instead. The long list of Subscription FAQs that's currently on there has a very high bounce rate and is also difficult to maintain as it's managed on a separate platform. 

>If there's anything on the long list of Subscription FAQs that's currently not on the help centre can you please let me know so that I can add it, so that everything is in the same place. Then we should change all FAQ links on the support.theguardian.com to point to the help centre.

### DIFF
--- a/support-frontend/assets/components/footerCompliant/DigitalFooter.jsx
+++ b/support-frontend/assets/components/footerCompliant/DigitalFooter.jsx
@@ -69,7 +69,7 @@ const GiftLinks = (props: LinkTypes) => {
 
 function DigitalFooter({ productPrices, orderIsAGift, country }: PropTypes) {
   const faqsLink = orderIsAGift ? 'https://www.theguardian.com/help/2020/nov/23/guardian-gift-digital-subscription-faqs' :
-    'https://www.theguardian.com/subscriber-direct/subscription-frequently-asked-questions';
+    'https://manage.theguardian.com/help-centre';
   const termsConditionsLink = orderIsAGift ?
     'https://www.theguardian.com/help/2020/nov/24/gift-digital-subscriptions-terms-and-conditions' :
     'https://www.theguardian.com/info/2014/aug/06/guardian-observer-digital-subscriptions-terms-conditions';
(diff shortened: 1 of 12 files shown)


support-frontend/assets/components/footerCompliant/DigitalFooter.jsx
```diff
@@ -68,14 +68,11 @@ const GiftLinks = (props: LinkTypes) => {
 };
 
 function DigitalFooter({ productPrices, orderIsAGift, country }: PropTypes) {
-  const faqsLink = orderIsAGift ? 'https://www.theguardian.com/help/2020/nov/23/guardian-gift-digital-subscription-faqs' :
```
johnduffell, line 71: this redirects to help centre already anyway

support-frontend/assets/pages/digital-subscription-checkout/thankYouExisting.jsx
```diff
@@ -38,10 +38,7 @@ const content = (
     <Page
       header={<Header countryGroupId={countryGroupId} />}
       footer={
-        <Footer
-          faqsLink="https://www.theguardian.com/subscriber-direct/subscription-frequently-asked-questions"
```
rupertbates, line 42: I thought the idea was to send the faqs to the help centre, not to delete them completely?

johnduffell: true I didn't quite read the email properly, maybe we should hold off merging until someone? customer operations? has said they are happy no content is lost?

rupertbates: Yes, I think we still need a link in all these places, just to the help centre rather than the faqs

rupertbates (approved):
👍

PR #3264, opened 2021-09-30
Fix issue with query string param on Newspaper LP

GHaberis:

Investigated an issue with @rupertbates and @johnduffell where the Newspaper checkout form was always displaying Subs Card as selected product, even if Home Delivery had been selected. This was down to an issue with the way we construct the URL and query params.  

Incorrect: https://support.theguardian.com/subscribe/paper/checkout?fulfilmentOption=HomeDelivery&productOptions=Everyday

Correct: https://support.theguardian.com/subscribe/paper/checkout?fulfilment=HomeDelivery&product=Everyday


johnduffell (approved):
👍

PR #3266, opened 2021-10-01
Remove unnecessary user type check

rupertbates:

We have some code in the subscriptions checkouts to check what type a user is (`'new' | 'guest' | 'current'`) when they fill in their email address. 

This code was only needed if we were going to force existing users to sign in before checking out - now that we are no longer doing this it can be deleted, which this PR does.

@tomrf1 I think the equivalent code can also be removed from contributions would you agree? If so I'm happy to remove that as well in this PR.


tomrf1:
> @tomrf1 I think the equivalent code can also be removed from contributions would you agree? If so I'm happy to remove that as well in this PR.

We still use the userType on the contributions thank you page.
I'm not sure there's any more clean up to be done after https://github.com/guardian/support-frontend/pull/3242/files

PR #3268, opened 2021-10-04
Delete acquisition event producer + remove thrift

tomrf1:


acquisition-event-producer/src/main/scala/com/gu/acquisition/model/ReferrerAcquisitionData.scala
```diff
@@ -14,8 +13,8 @@ case class ReferrerAcquisitionData(
   referrerPageviewId: Option[String],
   referrerUrl: Option[String],
   componentId: Option[String],
-  componentType: Option[ComponentType],
-  source: Option[AcquisitionSource],
```
tomrf1, line 18: these are now just strings in the scala model

support-frontend/app/codecs/CirceDecoders.scala
```diff
@@ -1,19 +0,0 @@
-package codecs
```
tomrf1, line 1: this file just adds thrift codecs

support-models/src/main/scala/com/gu/support/acquisitions/AcquisitionData.scala
```diff
@@ -0,0 +1,70 @@
...
+
+case class QueryParameter(
+  name: String,
+  value: String
+)
+
+/**
+ * Model for acquisition data passed by the referrer.
+ */
+case class ReferrerAcquisitionData(
```
tomrf1, line 21: Was defined in acquisition-event-producer

support-models/src/main/scala/com/gu/support/acquisitions/AcquisitionData.scala
```diff
@@ -0,0 +1,70 @@
...
+import com.gu.support.encoding.Codec.deriveCodec
+import io.circe.generic.semiauto.{deriveDecoder, deriveEncoder}
+import io.circe.{Decoder, Encoder}
+
+case class AbTest(
+  name: String,
+  variant: String
+)
+
+case class QueryParameter(
```
tomrf1, line 13: AbTest and QueryParameter were defined in both the thrift and scala acquisition models.
They're actually equivalent.
I've moved them into support-models because support-frontend needs them as well, and it shouldn't have to depend on the acquisition-events module which is a separate concern.

support-models/src/main/scala/com/gu/support/acquisitions/AcquisitionData.scala
```diff
@@ -0,0 +1,70 @@
...
+object OphanIds {
+  import io.circe._
+  import io.circe.generic.semiauto._
+
+  implicit val decoder: Decoder[OphanIds] = deriveDecoder[OphanIds]
+
+  implicit val encoder: Encoder[OphanIds] = deriveEncoder[OphanIds]
+}
+
+case class AcquisitionData(
```
tomrf1, line 59: from `support-models/src/main/scala/com/gu/support/workers/AcquisitionData.scala`
but this file is now more general than support-workers

support-models/src/main/scala/com/gu/support/workers/AcquisitionData.scala
```diff
@@ -1,19 +0,0 @@
-package com.gu.support.workers
```
tomrf1, line 1: moved to support-models/src/main/scala/com/gu/support/acquisitions/AcquisitionData.scala

support-modules/acquisition-events/build.sbt
```diff
@@ -9,6 +9,8 @@ libraryDependencies ++= Seq(
   "io.circe" %% "circe-core" % circeVersion,
   "io.circe" %% "circe-generic" % circeVersion,
   "com.amazonaws" % "aws-java-sdk-kinesis" % "1.12.10",
+  "com.gu" %% "acquisitions-value-calculator-client" % "2.0.6",
+  "com.squareup.okhttp3" % "okhttp" % okhttpVersion,
```
tomrf1, line 13: these deps were coming via support-models/acquisition-event-producer

support-payment-api/src/main/scala/model/AcquisitionData.scala
```diff
@@ -13,8 +12,8 @@ case class AcquisitionData(
   referrerUrl: Option[String],
   componentId: Option[String],
   campaignCodes: Option[Set[String]],
-  componentType: Option[ComponentType],
-  source: Option[AcquisitionSource],
```
tomrf1, line 17: consistent with the scala acquisition model, these are now just strings

support-modules/acquisition-events/build.sbt
johnduffell: there is an sbt plugin that flags up when you rely on transitive dependencies, although it's probably not worth using for us

support-payment-api/src/main/scala/model/AcquisitionData.scala
johnduffell: in general I'm not a massive fan of strings everywhere, because they're so easy to mix up and therefore in this case mess up the analytics, but consistency trumps that and I think you added more tests in the last PR which should help.

johnduffell (approved):
nice!  Big pr but it looks like mostly cleanup so it's not too hard to review. 👍

support-payment-api/src/main/scala/model/AcquisitionData.scala
tomrf1: That model was introduced for BigQuery integration and has now replaced the thrift model. It wouldn't be hard to re-introduce e.g. a `ComponentType` type if we want to. Now that we've escaped thrift it's much easier to add new values to enums

support-modules/acquisition-events/build.sbt
rupertbates: I think we should get rid of the code which retrieves AV for GA. We never really use it in our tests and it would allow us to drop these dependencies.

PR #3271, opened 2021-10-04
Override no-default-export rule in our ESLint config for the migration

i-hardy:

This tweaks our future ESLint config, which otherwise just extends the Guardian TS config, to allow default exports from files.

## Why are you doing this?

We are currently using default exports all over the codebase, and fixing errors arising from this rule will mean having to touch multiple other files every time you try and fix the TS/lint errors in a single file, which will then cause any other lint errors in _those_ files to fail the ESLint check in the pre-commit hook.

Overriding this rule for now will make the gradual fix process a lot easier. We can then revisit this in future to better align with the existing shared config, perhaps switching to a warning first.

### DIFF
--- a/support-frontend/migration/.eslintrc.js
+++ b/support-frontend/migration/.eslintrc.js
@@ -1,5 +1,9 @@
 module.exports = {
   extends: '@guardian/eslint-config-typescript',
+  rules: {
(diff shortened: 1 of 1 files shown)


rupertbates (approved):
👍  makes sense

PR #3273, opened 2021-10-05
Default amount to 0 for GA AV calculation

tomrf1:
This was the old behaviour: https://github.com/guardian/support-frontend/blob/6fbf84dda131dc5eb01f8bf11437ee3cb54cfd16/support-workers/src/main/scala/com/gu/support/workers/lambdas/SendOldAcquisitionEvent.scala#L161


rupertbates (approved):
👍

johnduffell (approved):
👍

PR #3294, opened 2021-10-21
Fix TS and ESLint errors in tabs.tsx

GHaberis:

Fixes TS and ESLint errors in `assets/components/tabs/tabs.tsx`.


TomPretty:
> Looks like we should update `ts-log-errors.js` to omit everything before `/support-frontend` in the files paths written to `typescript-errors.json`.

haha yeah good for the github stats, but agree we should remove that part

PR #3302, opened 2021-10-22
Fix useLastOneOffContribution bug

TomPretty:

Fix `useLastOneOffContribution` bug where it would attempt to deserialize the response even if it was a `401`. Now we 

1. only send the request if we're signed in
2. check the response is ok before converting to json

Additionally, we now also explicitly `catch` the promise incase it rejects.


GHaberis (approved):
lgtm 👍

PR #3310, opened 2021-10-26
combine create subscription and contribution server side controllers

johnduffell:
…t logic changes

This PR combines the contribution and subscription create code on the server side, so that there is no code duplication any more.
The process I went through was
1. opening both in Diff mode,
1. find a difference in firstly naming, subsequently coding style, or finally actual logic
1. resolve them to be the same
1. repeat the process until they were identical
1. delete the contributions controller and redirect the route to point to the subs controller
1. test locally with no issues - contributions, DS, DS gift, Weekly with paypal, CC, DD

## Why are you doing this?

This is just code duplication and makes it harder to maintain.
I need to fix a bug where guest checkout sends the GU identity cookies to the backend, even if they are not valid. It doesn't make sense to fix in two places, so this PR preceds it.


tomrf1 (approved):
looks good - thanks for tidying up!

PR #3319, opened 2021-11-01
give people 7 weeks of 6 for 6 when we are in the christmas zone

johnduffell:

This PR gives people 7 weeks of 6 for 6, to guarantee that they get enough issues over christmas

NOTE we need to change the Specific Weeks for 6 for 6 in zuora (for ROW AND Domestic) to 7 at the moment this goes to PROD, and we need to revert this after or just before christmas (including the zuora change).

https://trello.com/c/IzbPWEkh/17-guardian-weekly-6-for-6-christmas-plan-change

## Why are you doing this?

Avoid people missing an issue, or the retention team having loads of subs to fix manually.

### DIFF
--- a/support-models/src/main/scala/com/gu/support/promotions/Promotion.scala
+++ b/support-models/src/main/scala/com/gu/support/promotions/Promotion.scala
@@ -61,6 +61,6 @@ object Promotions {
     starts = new DateTime(1971, 2, 20, 12, 0, 0, 0),
     expires = None,
     discount = None, freeTrial = None, incentive = None,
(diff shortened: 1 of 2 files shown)


kelvin-chappell (approved):
Thanks @johnduffell This will save us a lot of work 🎉

johnduffell:
changes made in zuora
image
image
The real time acquisitions table in bigquery doesn't show any GW subs (yet) since merge

johnduffell:
I checked earlier and a sub had come through and looks correct.  There were a number subs already which will need remedial work, but in future they should be fine.

PR #3320, opened 2021-11-02
Add Quantum Metric Script [POC]

GHaberis:

This PR loads Quantum Metric's tag on support.theguardian.com as part of a trial. The script is loaded if all of the following conditions have been met:

- The feature switch `enableQuantumMetric` has been turned ON.
- The page path matches the Contribution's landing page and thank you page (all regions) OR  the (Non-Gifting) Digital Subscriptions landing page, checkout and thank you page (all regions).
- The user has consented to the following TCFv2 purposes via the CMP: `Store and/or access information on a device`, `Measure content performance` and `Develop and improve products`.

**Note:** `switches.json` on S3 has already been updated to include the `enableQuantumMetric` switch, this has been set to "Off" on `PROD`.


support-frontend/assets/helpers/page/analyticsAndConsent.ts
```diff
@@ -9,23 +9,30 @@ import {
...
 import { isPostDeployUser } from 'helpers/user/user';
-import * as logger from 'helpers/utilities/logger';
+import { init as initLogger } from 'helpers/utilities/logger';
 import 'helpers/internationalisation/country';
+
 // ----- Functions ----- //
+
 // Sets up GA and logging.
-export function analyticsInitialisation(
+const analyticsInitialisation = (
```
rupertbates, line 20: Given the discussion elsewhere around a preference for function declarations, should we stick with that approach here?

support-frontend/assets/helpers/tracking/quantumMetric.ts
```diff
@@ -0,0 +1,70 @@
...
+	 * Dynamically load @guardian/consent-management-platform
+	 * on condition we're not server side rendering (ssr) the page.
+	 * @guardian/consent-management-platform breaks ssr otherwise.
+	 */
+	import('@guardian/consent-management-platform')
+		.then(({ onConsentChange }) => {
+			onConsentChange((state) => {
+				// Check users consent state
+				if (
+					state.ccpa?.doNotSell === false ||
```
rupertbates, line 55: Slightly struggling with the logic here, so if `doNotSell` is true we DO load QM? And also if the three consents are true.

support-frontend/conf/CODE.public.conf
```diff
@@ -17,7 +17,7 @@ paymentApi.url="https://payment.code.dev-guardianapis.com"
...
...ypal.com js.stripe.com https://payments-sandbox.amazon.com https://api-cdn.amazon.com static-na.payments-amazon.com pasteup.guim.co.uk ophan.theguardian.com j.ophan.co.uk media.guim.co.uk uploads.guim.co.uk www.google-analytics.com www.googletagmanager.com tagmanager.google.com assets.guim.co.uk www.googleadservices.com googleads.g.doubleclick.net www.google.com www.google.co.uk optimize.google.com static.ads-twitter.com bat.bing.com bid.g.doubleclick.net t.co analytics.twitter.com stats.g.doubleclick.net www.youtube-nocookie.com connect.facebook.net www.facebook.com secure.adnxs.com checkout.stripe.com fonts.googleapis.com ssl.gstatic.com www.gstatic.com fonts.gstatic.com sentry.io www.dwin1.com *.quantummetric.com data: wss: 'unsafe-inline' q.stripe.com payment.code.dev-guardianapis.com https://interactive.guim.co.uk/ https://www.theguardian.com/ https://theguardian.com/ https://contribution-referrals-code.support.guardianapis.com/referral-code; frame-ancestors https://gnmtouchpoint--dev1--c.cs88.visual.force.com https://*.lightning.force.com; worker-src blob:; child-src https://www.facebook.com/ https://www.google.com/ https://js.stripe.com/ https://cdn.privacy-mgmt.com/ blob:;"
```
rupertbates, line 20: It's so impossible to read the diff on the CSP, could you flag the changes please (just in a github comment is fine)

rupertbates (approved):
Looks good, just a couple of requests for clarification

support-frontend/assets/helpers/tracking/quantumMetric.ts
GHaberis: Yeah the various potential return values for consent can make things confusing, they are explained here: https://github.com/guardian/consent-management-platform.

We only want to load the script if the condition on the lefthand side OR righthand side of the double pipe operator (`||`) is met.

We first check the conditional on the lefthand side. `state.ccpa` should only be returned for users in the USA. `state.ccpa.doNotSell` reports whether the user has withdrawn consent to sell their data in the USA, so a `true` value indicates they have withdrawn consent and a `false` value indicates they have NOT withdrawn consent. We only want to load the script for user's in the USA who have NOT withdrawn consent. 

If the condition on the lefthand side isn't met we then check the conditional on the righthand side. `state.tcfv2` should be defined for non-US users. `state.tcfv2.consents` reports the user's preferences for each of the TCFv2 purposes, a `true` value indicates they have consented to a purpose, we want to check that they've consented to the 3 purposes listed.

support-frontend/conf/CODE.public.conf
GHaberis: I agree regarding CSP changes being difficult to diff, I'm going to create a card to see if it's possible to split the string across multiple lines in the source code, this would make it far easier to manage/review! 

Regarding the changes to the CSP in this PR, we've included a new source `*.quantummetric.com` under the `default-src` directive. We also had to explicitly include 3 new directives `worker-src`, `child-src` and `frame-src`, because we're explicitly including these now we've had to duplicate some of the sources previously covered by `default-src` directive. 

This diff illustrates the difference...

Screenshot 2021-11-09 at 12 26 38

rupertbates (changes requested):
This branch caused PayPal to fail with a CSP error on CODE, can we fix that before merging and check that the same won't happen on PROD please?

GHaberis:
> This branch caused PayPal to fail with a CSP error on CODE, can we fix that before merging and check that the same won't happen on PROD please?
> 
> Error was:
> 
> ```
> Refused to frame 'https://www.sandbox.paypal.com/' because it violates the following Content Security Policy directive: "child-src https://www.facebook.com/ https://www.google.com/ https://js.stripe.com/ https://cdn.privacy-mgmt.com/ blob:". Note that 'frame-src' was not explicitly set, so 'child-src' is used as a fallback.
> ```

This has been resolved by removing the explicit addition of `frame-src` in the CSP and instead we'll continue to use the `default-src` fallback (which includes `frame-src`)

support-frontend/conf/CODE.public.conf
GHaberis: The only change to the CSPs is the addition of `*.quantummetric.com` and `blob:` to the `default-src` directive.

support-frontend/typescript-errors.json
```diff
@@ -4085,6 +4103,14 @@
...
             "path": "assets/helpers/tracking/googleTagManager.ts",
             "message": "Could not find a declaration file for module 'uuid'. 'support-frontend/node_modules/uuid/index.js' implicitly has an 'any' type."
+        },
+        {
+            "path": "assets/pages/contributions-landing/components/ContributionFormContainer.tsx",
+            "message": "Could not find a declaration file for module 'react-router-dom'. 'support-frontend/node_modules/react-router-dom/index.js' implicitly has an 'any' type."
+        },
+        {
+            "path": "assets/pages/contributions-landing/contributionsLanding.tsx",
+            "message": "Could not find a declaration file for module 'react-router-dom'. 'support-frontend/node_modules/react-router-dom/index.js' implicitly has an 'any' type."
```
TomPretty, line 4113: I think running `yarn` will probably fix up these ones. I think you'll then need to manually run `yarn ts-errors` to regenerate this file

TomPretty: (might also need to rebase/merge main first too)

rupertbates (approved):
👍

TomPretty (approved):
🔥

PR #3114, opened 2021-06-14
Add special "previous giving" message to the LP

TomPretty:
## What are you doing in this PR?
Add a special message to the contributions landing page for users that clicked through from the "support again" header (https://github.com/guardian/support-dotcom-components/pull/469)

[**Trello Card**](https://trello.com/c/2jUr2U7V/2637-poc-display-a-custom-message-on-the-lp-based-on-readers-previous-giving)

## Screenshots

Screenshot 2021-06-14 at 14 54 16

PR #3321, opened 2021-11-03
Fix previous giving bug on contribs LP

TomPretty:
# What are you doing in this PR?

Fix bug with previous giving message on the contribs LP. It was rendering both the previous giving message & the regular blurb, we should only render one of the two.

This was the original behaviour: https://github.com/guardian/support-frontend/pull/3114/files#diff-b1538b0b6f3bada5135caf512f68ffc675bc2cc9d6c2d873da27db9ff1522c83R125 but we must have lost it through running a few different AB tests in this area.


rupertbates (approved):
👍

PR #3322, opened 2021-11-03
Add missing apostrophe and fix errors

samanthagottlieb:
Co-authored-by: Tom Pretty <TomPretty@users.noreply.github.com>

Adds missing apostrophe to the article count opt-out option. (N.B. we're not actually rendering this component at the moment).

The rest of this PR is eslint and typescript error fixes.

**UPDATE** This PR is now just the typo fix as the ts errors have now already been fixed since!


support-frontend/assets/pages/contributions-landing/components/ContributionArticleCount.tsx
```diff
@@ -295,149 +291,140 @@ export const ArticleCountWithToggle = ({
...
+		function openCmp() {
+			cmp.showPrivacyManager();
 		}
-	}
 
-	useEffect(() => {
-		import('@guardian/consent-management-platform').then(({ cmp }) => {
-			setConsentManagementPlatform(cmp);
-		});
-	}, []);
```
samanthagottlieb, line 321: We have replaced this by using a regular import to satisfy the type checker. However, we'd like to know if there was an important reason it was done like this?

support-frontend/assets/pages/contributions-landing/components/ContributionArticleCount.tsx
```diff
@@ -295,149 +291,140 @@ export const ArticleCountWithToggle = ({
...
+										on this device. Can we continue showing your article count?
+									</div>
+									<div css={articleCountCtasContainerStyles}>
+										<Button
+											priority="primary"
+											size="xsmall"
+											cssOverrides={articleCountDefaultCtaStyles}
+											onClick={onStayInClick}
+										>
+											Yes, that's OK
```
samanthagottlieb, line 368: Here's where we've added the missing apostrophe.

support-frontend/assets/pages/contributions-landing/components/ContributionArticleCount.tsx
GHaberis: @samanthagottlieb when we first started using `@guardian/consent-management-platform` elsewhere in the codebase we were using regular imports, however we found these import statements broke our scripts that pre-render some of our pages for server side rendering. To overcome this we replaced the regular imports with dynamic import statements (such as the one used here) for `@guardian/consent-management-platform`. These dynamic imports would only run on condition we're not server side rendering the page. 

For typical React applications `useEffect` would not run on the server if server side rendering on a Node server using `ReactDOMServer.renderToString`, we have what I'd describe as a non-typical approach to server side rendering where we pre-render the pages using JSDOM at build-time and save them to disc, so I'm not sure whether `useEffect` runs at this point.

GHaberis: If `ContributionArticleCount.tsx` isn't imported elsewhere replacing the dynamic import with a regular import shouldn't cause issues at build time, however once if we do begin to use `ContributionArticleCount.tsx` you may find the regular import for `@guardian/consent-management-platform` breaks our server side rendering scripts. You can test this by running `yarn build-ssr`.

i-hardy:
Just spotted that this is still hanging around- @GHaberis @samanthagottlieb are we happy with the SSR issue with the CMP? Otherwise this looks fine to me.

support-frontend/assets/pages/contributions-landing/components/ContributionArticleCount.tsx
samanthagottlieb: Thanks for the reminder about this, @i-hardy, it's been on my to do list!

I ran `yarn build-ssr` and it ran ok. Is this just checking if `ContributionArticleCount.tsx` is imported elsewhere @GHaberis? Would it be better to keep the CMP dynamic import statement in case we do begin to import `ContributionArticleCount.tsx` somewhere in the future?

Tagging @TomPretty as it was his suggestion to use a regular import statement and he may have some thoughts.

GHaberis: I'd opt for the dynamic import. As `ContributionArticleCount.tsx` isn't imported anywhere I'd not expect `yarn build-ssr` to have any issues for now, however if we do begin to import `ContributionArticleCount.tsx` somewhere in the future I suspect we'd run into issues with a regular import.

TomPretty: Okay sounds like dynamic import is best here (side note: perhaps we should factor out a `useCmp` custom hook to share this behaviour?). When we looked at this briefly the main reason for not doing it was that we'd need a type for the `cmp` object to make ts happy and that seemed like it was going to be a headache (I don't think the library ships a type for that)

i-hardy: I was thinking earlier that a `useCmp` hook would be a nice way to wrap this, as we've got the same dynamic loading behaviour elsewhere. It looks like the library has its own TS types that we now have access to, so that shouldn't be an issue anymore. It might be worth creating an extra card for that though.

PR #3324, opened 2021-11-09
Format amounts in contribs checkout and ty pages

marialani:
This PR updates the formatting of the contribution amounts for the checkout and thank you pages.

[**Trello Card**](https://trello.com/c/bOfcNwkV/122-correct-formatting-of-contribution-amounts-on-the-thank-you-page)

## Why are you doing this?
If users chose to give a custom amount which was not a multiple of a pound (e.g £2.50) using the 'other' field, their contribution amount would not translate correctly to the cta button and the thank you page.

## Is this an AB test?
- [ ] Yes
- [x] No

## Screenshots
| Before  | After |
| ------------- | ------------- |
| image  | image  |
| Screenshot 2021-11-08 at 13 46 48 | Screenshot 2021-11-08 at 13 45 28 |


support-frontend/assets/helpers/forms/checkouts.ts
```diff
@@ -135,7 +136,7 @@ function getAmountFromUrl(): number | 'other' | null {
 		return 'other';
 	}
 
-	const amount = parseInt(selected, 10);
+	const amount = parseInt(selected as string, 10);
```
johnduffell, line 139: I'm no way an expert on typescript, but does `as string` add a compile time type where it's `any`?  I'm just wondering whether `getQueryParameter` should return a `string` in the first place, if it doesn't already.  Any thoughts?

johnduffell (approved):
great description on this PR
and I honestly don't follow typescript completely, so feel free to wait for a more detailed review if you want someone who knows more  👍

support-frontend/assets/helpers/forms/checkouts.ts
marialani: I'm type casting the `selected` argument here because the the first parameter in `getQueryParameter `

The `parseInt` function only accepts a string so `as string` here will type cast the result of calling `getQueryParameter` to a string at compile time vs the given return types of the function i.e. `string | null | undefined`.

And looking at how the `getQueryParameter` function is used I agree we could include a default value for the parameter of an empty string and therefore the return type would be a string.

support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -97,7 +97,7 @@ function getParticipationsFromUrl(): Participations | null | undefined {
 }
 
 function getServerSideParticipations(): Participations | null | undefined {
-	if (window && window.guardian && window.guardian.serversideTests) {
+	if (!(Object.keys(window.guardian.serversideTests).length === 0)) {
```
marialani, line 100: As `window.guardian.serversideTests` is an object checking for the presence of keys here as an empty object is truthy

support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -112,8 +112,8 @@ function getIsRemoteFromAcquisitionData(): boolean {
 	}
 
 	try {
-		const data = JSON.parse(queryString);
-		return data && !!data.isRemote;
+		const data = JSON.parse(queryString) as Record<string, unknown>;
```
marialani, line 115: A bit unsure on how to type this but open to suggestions!

support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -200,7 +199,7 @@ function userInTest(
 }
 
 function randomNumber(mvtId: number, seed: number): number {
-	const rng = seedrandom(mvtId + seed);
+	const rng = seedrandom(`${mvtId + seed}`);
 	return Math.abs(rng.int32());
```
marialani, line 203: Using string interpolation here as `@types/seedrandom` requires a string though `rng` is of type number

support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -327,10 +326,6 @@ function getAmountsTestParticipations(
 	countryGroupId: CountryGroupId,
 	settings: Settings,
 ): Participations | null | undefined {
-	if (!settings.amounts) {
-		return null;
-	}
```
marialani, line 332: This condition was unnecessary as the value was always falsy

support-frontend/assets/helpers/abTests/__tests__/abtestTest.ts
```diff
@@ -10,16 +10,18 @@ import {
...
 const { nonGiftLandingNotAusNotUS, nonGiftLandingAndCheckoutWithGuest } =
 	digiSub;
+
 jest.mock('ophan', () => ({
 	record: () => null,
 }));
+
 // ----- Tests ----- //
-const emptySettings: Settings = {
+const emptySettings = {
```
marialani, line 24: This object doesn't conform to the `Settings` type so removed it here and am type casting the variable when it is called in the tests

support-frontend/assets/helpers/abTests/abtest.ts
marialani: Unexpectedly converting the argument passed to the `seedrandom` function to a string caused one of our tests to fail. We have looked into the library and though seedrandom accepts and converts an argument of type number, the eventual seed is slightly different than if it is passed a string. 

To account for this quirk from the library, the seed of the failing test which allocates a user to a `variant` group has been modified and to confirm functionality a test to allocate a user to the `control` group has been included.

support-frontend/assets/helpers/forms/checkouts.ts
```diff
@@ -262,9 +263,11 @@ const formatAmount = (
 		}`;
 	}
 
+	const amountText = /^(\d+\.\d)$/.test(`${amount}`) ? `${amount}0` : amount;
+
 	const valueWithGlyph = currency.isSuffixGlyph
-		? `${amount}${glyph}`
-		: `${glyph}${amount}`;
+		? `${amountText}${glyph}`
+		: `${glyph}${amountText}`;
```
marialani, line 270: Formatting the displayed contribution amount here

support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -88,7 +100,7 @@ function getParticipationsFromUrl(): Participations | null | undefined {
 
 	if (hashUrl.startsWith('#ab-')) {
 		const [testId, variant] = decodeURI(hashUrl.substr(4)).split('=');
-		const test = {};
+		const test: Record<string, string> = {};
```
TomPretty, line 103: Not necessary, but I wonder if this would be more straightforward as:

```ts
return { [testId]: variant };
```

support-frontend/assets/helpers/forms/checkouts.ts
johnduffell: ok I see so it's kind of like an optional string coming back, and we wanted to tell the type checker, actually we know it will be present.  Makes more sense in that case, but I think your suggestion of a default makes sense also because we get even more comprehensive type checking.

support-frontend/assets/helpers/abTests/abtest.ts
GHaberis: @marialani thanks for pointing me towards this. If we alter our use of `seedrandom` to pass a string then as you pointed out the eventual seed returened by that function will be slightly different than if it is passed a `number`. The issue with releasing this is that we'll potentially reassign users to different cohorts in any existing AB tests. 

I'd much rather use `"@types/seedrandom"` and pass a string as `seedrandom` intends and as you've done, but I'm not sure we can release this whilst we have any ab tests running?

support-frontend/assets/helpers/abTests/abtest.ts
marialani: Yeah this would be a nice simplification

support-frontend/assets/helpers/abTests/abtest.ts
marialani: Yeah completely agree, I can remove the ab test related changes from this PR and maybe we can time making the change to the code between ab tests?

support-frontend/assets/helpers/abTests/abtest.ts
GHaberis: How about `const data = JSON.parse(queryString) as { isRemote?: boolean };` , this is more explicit about the expected return type.

support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -123,18 +135,17 @@ function getIsRemoteFromAcquisitionData(): boolean {
...
 	const acquisitionDataParam = getQueryParameter('acquisitionData');
 
-	if (acquisitionDataParam == null) {
+	if (!acquisitionDataParam) {
 		return null;
 	}
 
 	try {
-		const acquisitionData = JSON.parse(acquisitionDataParam);
+		const acquisitionData = JSON.parse(acquisitionDataParam) as Record<
```
GHaberis, line 143: Based on the original code `abTest` is optional, could the below work instead? 

```
const acquisitionData = JSON.parse(acquisitionDataParam) as {
    abTest?: AcquisitionABTest;
};

if (acquisitionData.abTest?.variant) {
  return acquisitionData.abTest;
}
```

support-frontend/assets/helpers/abTests/abtest.ts
TomPretty: [here's](https://github.com/guardian/support-dotcom-components/blob/main/packages/shared/src/lib/tracking.ts#L34) the model in sdc that is added to the query params. We could try and match this type, but it might be easiest just to leave it like this!

support-frontend/assets/helpers/abTests/abtest.ts
GHaberis: OK I think that seems sensible, I'd say remove the `abtest.ts` changes from here and we can release them as a separate PR that doesn't risk compromising any live abtests.

GHaberis: Alternatively could we ignore the TS errors related to `seedrandom` and have a `TODO` to fix these separately in a follow-up PR?

support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -158,10 +169,10 @@ function userInBreakpoint(audience: Audience): boolean {
 
 	const minWidthMediaQuery = minWidth
 		? `(min-width:${breakpoints[minWidth]}px)`
-		: null;
+		: '';
 	const maxWidthMediaQuery = maxWidth
 		? `(max-width:${breakpoints[maxWidth]}px)`
```
GHaberis, line 174: By changing the fallback value of the ternary operators for `minWidthMediaQuery` and `maxWidthMediaQuery` from `null` to an empty string it alters what this function will return if these fallbacks are used, as it'll potentially mean the `mediaQuery` variable is assigned empty string as a value instead of `null`. 

The console log below demonstrates how this will affect the return value of `window.matchMedia`:

Screenshot 2021-11-15 at 16 35 38

I believe if we get to the point where we're calling `window.matchMedia` and `mediaQuery` is an invalid media query (such as an empty string) we should return `false`, you could put a conditional infront of the final return statement which would return `false` if `mediaQuery` is an empty string (or falsy)...

```
if (!mediaQuery) {
  return false;
}

return window.matchMedia(mediaQuery).matches;
```

support-frontend/assets/helpers/urls/url.ts
```diff
@@ -12,15 +12,12 @@ const DOMAINS: Record<Env, Domain> = {
 };
 
 // ----- Functions ----- //
-const getQueryParameter = (
-	paramName: string,
-	defaultValue?: string,
-): string | null | undefined => {
-	const params = new URL(window.location).searchParams;
+const getQueryParameter = (paramName: string, defaultValue = ''): string => {
```
johnduffell, line 15: for future reference, what we didn't spot is that the old defaultValue was `undefined` whereas now it's `''`,  this caused a subtle promo link regression that is now fixed https://github.com/guardian/support-frontend/pull/3402

PR #3328, opened 2021-11-12
Patrons free digital subscription

marialani:
Co-authored-by: Samantha Gottlieb <samanthagottlieb@users.noreply.github.com>
Co-authored-by: Rupert Bates <rupertbates@users.noreply.github.com>

This PR updates the digisub payment journey for patrons to reflect their free subscription and fixes some typing errors.

[**Trello Card**](https://trello.com/c/JG8pGEGk/59-create-100y-digi-sub-promo-code-for-patrons)

## Is this an AB test?
- [ ] Yes
- [x] No

## Screenshots
| Landing page  | Checkout |
| ------------- | ------------- |
|   |  |


support-frontend/assets/pages/digital-subscription-checkout/components/orderSummary/orderSummary.tsx
```diff
@@ -18,6 +19,8 @@ type PropTypes = {
 	orderIsAGift?: boolean;
 };
 
+export let isPatron: boolean;
```
i-hardy, line 22: It's a really, really, really bad idea to export mutable variables, and especially so that a parent component can access information determined by the child. You don't want to couple components together through data at all outside of what a parent passes to a child as props (or what a child may transmit back up through calling function props). `DigitalSubscriptionForm` shouldn't be reliant on `OrderSummary` being rendered and doing its own internal logic in order to access this information.

I'd suggest either 
a) Determine the value of `isPatron` in `DigitalCheckoutForm`, and pass it into `OrderSummary` as a prop
b) Create a utility function that does the logic to determine if a promo code means a user is a patron, and use it independently in both components

I think a) is preferable because it means you're only determining the value of `isPatron` once, but a named utility function would be nice either way, just for clarity.

i-hardy (changes requested):
Mostly this is looking good, but we need to sort out the way `isPatron` gets determined.

support-frontend/assets/pages/digital-subscription-checkout/components/orderSummary/orderSummary.tsx
marialani: Thanks Imogen, have gone for option a) as like you'd mentioned we're only determining the value of `isPatron` once here

i-hardy (approved):
Looking good!

PR #3330, opened 2021-11-15
Tidy up csp

rupertbates:

The config we were using for our CSP is deprecated so I have moved it to the new structure. It is also a bit clearer as it separates out the individual directives into separate config keys.

As a next step it would be great to break up `default-src` into more specific directives


support-frontend/app/wiring/AppComponents.scala
```diff
@@ -22,13 +24,14 @@ trait AppComponents extends PlayComponents
...
   with CORSComponents
+  with CSPComponents
   with HttpFiltersComponents {
   self: BuiltInComponentsFromContext =>
 
   private lazy val customHandler: CustomHttpErrorHandler = new CustomHttpErrorHandler(
     environment,
     configuration,
-    sourceMapper,
+    devContext.map(_.sourceMapper),
```
rupertbates, line 34: This is to fix a deprecation warning

support-frontend/app/wiring/AppLoader.scala
```diff
@@ -36,7 +36,7 @@ class AppLoader extends ApplicationLoader with StrictLogging {
         ComposedConfigurationLocation(List(privateConfigLocal, privateConfigSSM, publicConfig))
     }
 
-    initialConfiguration ++ Configuration(loadedConfig)
+    Configuration(loadedConfig).withFallback(initialConfiguration)
```
rupertbates, line 39: another deprecation warning

johnduffell (approved):
nice work although hopefully the headers returned and whatnot are still identical :)

PR #3332, opened 2021-11-15
Typescript error squashing

TomPretty:
I did some error squashing in:

- support-frontend/assets/helpers/campaigns/campaigns.ts
- support-frontend/assets/helpers/tracking/acquisitions.ts
- support-frontend/assets/pages/contributions-landing/components/ContributionFormContainer.tsx
- support-frontend/assets/pages/contributions-landing/components/ContributionThankYou/ContributionThankYou.tsx

We should probably do some more work to refactor `acquisitions.ts` as the types there are quite awkward - lots of `any`s & `A | null | undefined` which we would be better of to commiting to `A | null` or `A?`


support-frontend/assets/helpers/tracking/acquisitions.ts
```diff
@@ -111,26 +125,17 @@ function storeReferrerAcquisitionDataInSessionStorage(
...
-		response.push({
-			name: participation,
-			variant: participations[participation as any],
-		});
-	});
-	return response;
+	return Object.entries(participations).map(([name, variant]) => ({
+		name,
+		variant,
+	}));
```
TomPretty, line 138: I'm not quite sure the history of these functions, but I think now they're best of being `map`s!

support-frontend/typescript-errors.json
```diff
@@ -4881,6 +4784,12 @@
             "message": "A required parameter cannot follow an optional parameter."
         }
     ],
+    "TS2344": [
+        {
+            "path": "assets/helpers/page/page.ts",
+            "message": "Type 'A' does not satisfy the constraint 'Action<any>'."
+        }
+    ],
```
TomPretty, line 4792: I'm not clear on how I introduced this one

support-frontend/assets/helpers/tracking/acquisitions.ts
```diff
@@ -139,12 +144,15 @@ function buildReferrerAcquisitionData(
 ): ReferrerAcquisitionData {
 	// This was how referrer pageview id used to be passed.
 	const referrerPageviewId =
-		acquisitionData.referrerPageviewId || getQueryParameter('REFPVID');
+		(acquisitionData.referrerPageviewId as string | undefined) ??
+		getQueryParameter('REFPVID');
+
```
TomPretty, line 149: I think the real solution here is to get rid of the `Record<string, any>` type, but that feels like a real rabbit whole to fall down that I didn't want to tackle right now!

support-frontend/assets/helpers/tracking/acquisitions.ts
tomrf1: 💯

tomrf1 (approved):
great work! That `acquisitionData` type needs some thought...

PR #3335, opened 2021-11-16
Fully disable 6/6

tomrf1:
The decision has been made to completely remove 6/6 for Guardian Weekly.
But we want the ability to bring it back at some point, but it might be a bit different, and we don't know if it'll be an AB test.
In the absence of requirements I'm just removing the control from the AB test. This means all users will _not_ see 6/6.
We can revisit this later and remove the AB test once we know how it's going to work.

Screen Shot 2021-11-16 at 14 37 21


i-hardy (approved):
👍

johnduffell:
nice.  unless you've done that also, I think CSRs will still be able to sign people up, via salesforce and also via subsscribe.theguardian.com.

PR #3337, opened 2021-11-17
Add @types/seedrandom

marialani:
This PR includes the types package for seedrandom, passes the seedrandom function the required type i.e. a string and updates the relevant tests which make use of seedrandom to allocate users into abtest cohorts.

[**Trello Card**](https://trello.com/c/sXj0lS3s/150-update-seeded-random-aspect-of-abtest-allocation)

## Why are you doing this?
We had been passing seedrandom a value of type number and when this value is converted to a string the eventual seed output is different. To account for this difference the seed values in the relevant tests have also been updated.


GHaberis (commented):
Looks good 👍

GHaberis (approved):

PR #3339, opened 2021-11-17
Turn off contribs rich lp test

TomPretty:

Turn off contribs rich lp test. There's a future TODO to remove all the actual rich lp code.


rupertbates (approved):
👍

PR #3342, opened 2021-11-18
Check result in getAvailablePaymentRequestButtonPaymentMethod

GHaberis:

The `result` argument in `getAvailablePaymentRequestButtonPaymentMethod` was incorrectly marked as non-optional (this precedes the Typescript migration), because of this https://github.com/guardian/support-frontend/pull/3324 replaced the `result && result.applePay` with a `result.applePay` check. 

`result` is the return value of `canMakePayment` function call, as the stripe docs show this can return `null`: https://stripe.com/docs/js/payment_request/can_make_payment. We therefore need to check if `result` is truthy (not null). 

This issue came to our attention in Sentry, and is also visible in the logs on the Contributions landing page. We don't believe it will have caused any negative impact for users wanting to Contribute.

Screenshot 2021-11-18 at 17 41 03


TomPretty (approved):
👍

PR #3347, opened 2021-11-23
Log better info on create failures

johnduffell:

This PR improves the logging for both the deseralisation and the validation failures on create.

https://trello.com/c/3gAHSJVt/164-improve-logging-server-side-validation-errors-on-create-endpoint

## Why are you doing this?

This should make it easier to get to the bottom of the alarms that keep going off.

Here are examples of the extra lines that would be logged:

`WARN  com.gu.monitoring.SafeLogger$ - circe decode failure: DecodingFailure(Unrecognised title Select a title, List(DownField(title)))`

and you can see that it combines together the failures with "and"

`WARN  com.gu.monitoring.SafeLogger$ - validation of the request body failed with first name was empty and last name was empty - body was CreateSupportWorkersRequest(Some(Title(Mr)),,,Address(Some(sdfsdfs),None,So.............etcetc`


support-frontend/app/controllers/CreateSubscriptionController.scala
```diff
@@ -146,6 +149,11 @@ class CreateSubscriptionController(
 
 class LoggingCirceParser(controllerComponents: ControllerComponents) extends Circe {
 
+  override protected def onCirceError(e: io.circe.Error): Result = {
+    SafeLogger.warn(s"circe decode failure: $e")
+    super.onCirceError(e)
+  }
```
johnduffell, line 155: this is all that's needed to get the circe failure logged, so I popped it in as an extra commit

support-frontend/app/utils/CheckoutValidationRules.scala
```diff
@@ -6,84 +6,116 @@ import com.gu.support.catalog.{Collection, Domestic, FulfilmentOptions, HomeDeli
...
-  def currencyIsSupportedForCountry(countryFromRequest: Country, currencyFromRequest: Currency): Boolean = {
+  def currencyIsSupportedForCountry(countryFromRequest: Country, currencyFromRequest: Currency): Result = {
     val countryGroupsForCurrency = CountryGroup.allGroups.filter(_.currency == currencyFromRequest)
-    countryGroupsForCurrency.flatMap(_.countries).contains(countryFromRequest)
+    val isValid = countryGroupsForCurrency.flatMap(_.countries).contains(countryFromRequest)
+
+    val validCurrenciesForCountry = CountryGroup.allGroups.filter(_.countries.contains(countryFromRequest)).flatMap { countryGroup =>
+      countryGroup.currency :: countryGroup.additionalCurrencies
+    }
+    isValid.otherwise(s"currency $currencyFromRequest is not supported for country $countryFromRequest, can only have ${validCurrenciesForCountry}")
```
johnduffell, line 118: this logic was a bit upside down given the function name. It actually finds the currency and checks the country is valid for the currency.  I have kept that logic but done the calculation the other way round for the error message, rather than refactoring the logic in this PR.  However - opinions welcome!

marialani (approved):
Looks great!

PR #3350, opened 2021-11-24
Fix Typescript errors on digital subscription landing page and friends

i-hardy:

This addresses the Typescript errors on the digi subs landing page and also applies fixes to related files, including a couple of components and the `page.ts` file where we set up Redux and various other things.

I've extracted a reusable `CSSOverridable` type that can be extended by the various components we have that accept a `cssOverrides` prop that can be an Emotion style or array of Emotion styles. I've also (I hope) made the generic types used in the `initRedux` function more understandable as well as not broken any more.

Fixing `initRedux` seems to have created some additional errors in other checkouts, which will need to be resolved individually and/or by fixing the underlying type issues with our various checkout states and actions.

[**Trello Card**](https://trello.com/c/JKgpJvGz)


i-hardy:
[Typescript Error Busting - Digi subs landing page](https://trello.com/c/JKgpJvGz/176-typescript-error-busting-digi-subs-landing-page)

support-frontend/assets/helpers/tracking/acquisitions.ts
```diff
@@ -150,9 +150,7 @@ function buildReferrerAcquisitionData(
 	const campaignCode =
 		getCampaignCode() ??
 		(acquisitionData.campaignCode as string | undefined) ??
-		getQueryParameter('INTCMP') ??
-		getQueryParameter('CMP') ??
-		undefined;
+		(getQueryParameter('INTCMP') || getQueryParameter('CMP') || undefined);
```
i-hardy, line 153: A bunch of lint errors in this file initially blocked my merge commit from `main`, so I fixed them- this isn't otherwise related to the DS landing page. I think this was an instance where we were previously using logical OR (`||`) as _both_ a nullish coalescer (`??`) to deal with `null` or `undefined`, _and_ to circumvent falsey values that are _not_ `null` or `undefined`.

`getQueryParameter` will always return a string, but if it doesn't find a value for the passed parameter key name it will return an empty string. The nullish coalescing operator thinks an empty string is fine so will resolve to it and ignore the right hand side of the expression. However in this case if `'INTCMP'` returns empty string we want to look for `'CMP'` instead, so we want logical OR, which will resolve to the right hand side of the expression if the left hand side is any kind of falsey value.

rupertbates (approved):
👍

PR #3353, opened 2021-11-29
Disable promo message in roundel

tomrf1:
The UX is quite bad if the roundel copy is set from the promo tool:
1. It's not clear which cta the roundel copy is referring to (it's actually the second here)
2. The size and formatting of the copy can break the UI

Screen Shot 2021-11-29 at 11 02 15

For now the solution is to stop taking the copy from the promo tool.
The roundel will always say "14 day free trial", which is what the default page says.
Screen Shot 2021-11-29 at 11 05 29


KaliedaRik (approved):
Seems like a sensible change for the immediate issue.

PR #3354, opened 2021-11-29
Move turkey into international country group

TomPretty:
Co-authored-by: Maria Olanipekun <marialani@users.noreply.github.com>
Co-authored-by: Samantha Gottlieb <samanthagottlieb@users.noreply.github.com>

We've moved Turkey into the international country group on the frontend to match up with the backend logic. Previously users in Turkey couldn't create a subscription as the frontend would select euros and the backend would reject that currency. 

[**Trello Card**](https://trello.com/c/d8ABGbGH/161-update-frontend-currency-selection-for-turkish-users)


TomPretty:
[Update frontend currency selection for Turkish users](https://trello.com/c/d8ABGbGH/161-update-frontend-currency-selection-for-turkish-users)

support-frontend/typescript-errors.json
```diff
@@ -3114,36 +3110,6 @@
...
-        },
-        {
-            "path": "assets/helpers/rendering/ssrPages.ts",
-            "message": "Could not find a declaration file for module 'react-dom/server'. 'support-frontend/node_modules/react-dom/server.js' implicitly has an 'any' type."
-        },
-        {
-            "path": "assets/helpers/tracking/googleTagManager.ts",
-            "message": "Could not find a declaration file for module 'uuid'. 'support-frontend/node_modules/uuid/index.js' implicitly has an 'any' type."
-        }
-    ],
```
TomPretty, line 3146: I think these were caused by generating the error history but missing a few dependencies in node_modules

support-frontend/typescript-errors.json
```diff
@@ -3710,6 +3676,12 @@
             "message": "Object is possibly 'null'."
         }
     ],
+    "TS7016": [
+        {
+            "path": "assets/helpers/rendering/render.ts",
+            "message": "Could not find a declaration file for module 'preact/debug'. 'support-frontend/node_modules/preact/debug/dist/debug.js' implicitly has an 'any' type."
+        }
+    ],
```
TomPretty, line 3684: I'm not sure if this is something we added? We made sure to run `yarn` before generating the error file

support-frontend/typescript-errors.json
samanthagottlieb: Hi Tom, just to clarify what you mean - would these have not been introduced had someone run yarn before generating the error history?

support-frontend/typescript-errors.json
samanthagottlieb: Is this the same error that was removed on lines 3135 & 3136 from the 'before' diff?

support-frontend/typescript-errors.json
marialani: Yes I believe for the most part these would not have been introduced if yarn was run prior to committing

support-frontend/typescript-errors.json
marialani: Yeah it looks that way, I'm not sure why these errors were previously removed because it looks like we don't have the corresponding types module for `preact/debug`

i-hardy (approved):
👍

PR #3355, opened 2021-11-29
Fix title dropdown in subscriptions checkouts

marialani:
This PR ensures the value of the title field in the redux state is set to null when no title has been selected from the dropdown list. 

[**Trello Card**](https://trello.com/c/811Op89V/171-fix-title-dropdown-in-subscriptions-checkouts)

## Why are you doing this?
Previously if a user selected a title then set the title back to the initial value i.e. "Select a title". This would set the value of the title field to the string "Select a title" which would then cause a deserialisation error in the backend during the create product call.


marialani:
[Fix title dropdown in subscriptions checkouts](https://trello.com/c/811Op89V/171-fix-title-dropdown-in-subscriptions-checkouts)

support-frontend/assets/helpers/subscriptionsForms/formActions.ts
```diff
@@ -167,7 +167,7 @@ const setUserTypeFromIdentityResponse = (
 const formActionCreators = {
 	setTitle: (title: string): Action => ({
 		type: 'SET_TITLE',
-		title: title !== '' ? title : null,
+		title: title !== 'Select a title' ? title : null,
```
rupertbates, line 170: Is there never a point where we need to check for an empty string?

marialani: I don't think there would be, as the user can only select their title from the predetermined list in the dropdown and on the initial load the value is set to null. Although as a failsafe I can modify this to check for the string above as well as an empty string?

rupertbates (approved):
👍

support-frontend/assets/helpers/subscriptionsForms/formActions.ts
rupertbates: No that's fine, just wanted to check!

PR #3358, opened 2021-12-01
Migrate to switches v2

tomrf1:
We are migrating the switches json config to a new format. This is to make the S3 file the single source of truth. See switchboard PR: https://github.com/guardian/support-admin-console/pull/256

This PR migrates the scala model and uses the new file, switches_v2.json.
It does some transformation to the json to make the scala case classes more convenient for our app.
The app will fail to come up if the file is missing/invalid or any switches are missing (same behaviour as before)

### DIFF
--- a/support-frontend/app/admin/settings/Switches.scala
+++ b/support-frontend/app/admin/settings/Switches.scala
@@ -6,15 +6,9 @@ import io.circe.generic.extras.auto._
 import io.circe.{Decoder, Encoder, Json}
 import io.circe.optics.JsonPath._
 
-sealed trait SwitchState {
-  def isOn: Boolean
-}
-case object On extends SwitchState {
-  def isOn: Boolean = true
-}
-case object Off extends SwitchState {
-  def isOn: Boolean = false
-}
+sealed abstract class SwitchState(val isOn: Boolean)
+case object On extends SwitchState(true)
+case object Off extends SwitchState(false)
 
 object SwitchState {
   import io.circe.generic.extras.semiauto.{ deriveEnumerationDecoder, deriveEnumerationEncoder }
--- a/support-frontend/app/admin/settings/Switches.scala
+++ b/support-frontend/app/admin/settings/Switches.scala
@@ -17,44 +17,44 @@ object SwitchState {
 }
 
 case class FeatureSwitches(
-  enableQuantumMetric: SwitchState = Off,
-  usStripeAccountForSingle: SwitchState = On
+  enableQuantumMetric: SwitchState,
+  usStripeAccountForSingle: SwitchState
 )
 
 case class CampaignSwitches(
-  enableContributionsCampaign: SwitchState = Off,
-  forceContributionsCampaign: SwitchState = Off
+  enableContributionsCampaign: SwitchState,
+  forceContributionsCampaign: SwitchState
 )
 
 case class SubscriptionsSwitches(
-  enableDigitalSubGifting: SwitchState = On,
-  useDotcomContactPage: SwitchState = On,
-  checkoutPostcodeLookup: SwitchState = On
+  enableDigitalSubGifting: SwitchState,
+  useDotcomContactPage: SwitchState,
+  checkoutPostcodeLookup: SwitchState
 )
 
 case class RecaptchaSwitches(
-  enableRecaptchaBackend: SwitchState = On,
-  enableRecaptchaFrontend: SwitchState = On
+  enableRecaptchaBackend: SwitchState,
+  enableRecaptchaFrontend: SwitchState
 )
 
 case class OneOffPaymentMethodSwitches(
-  stripe: SwitchState = On,
-  stripeApplePay: SwitchState = On,
-  stripePaymentRequestButton: SwitchState = On,
-  payPal: SwitchState = On,
-  amazonPay: SwitchState = On
+  stripe: SwitchState,
+  stripeApplePay: SwitchState,
+  stripePaymentRequestButton: SwitchState,
+  payPal: SwitchState,
+  amazonPay: SwitchState
 )
 
 case class RecurringPaymentMethodSwitches(
-  stripe: SwitchState = On,
-  stripeApplePay: SwitchState = On,
-  stripePaymentRequestButton: SwitchState = On,
-  payPal: SwitchState = On,
-  directDebit: SwitchState = On,
-  existingCard: SwitchState = On,
-  existingDirectDebit: SwitchState = On,
-  amazonPay: SwitchState = Off,
-  sepa: SwitchState = Off
+  stripe: SwitchState,
+  stripeApplePay: SwitchState,
+  stripePaymentRequestButton: SwitchState,
+  payPal: SwitchState,
+  directDebit: SwitchState,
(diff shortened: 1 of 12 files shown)


support-frontend/app/admin/settings/Switches.scala
```diff
@@ -1,21 +1,121 @@
...
+)
 
 case class Switches(
-  oneOffPaymentMethods: PaymentMethodsSwitch,
-  recurringPaymentMethods: PaymentMethodsSwitch,
-  enableDigitalSubGifting: SwitchState,
-  useDotcomContactPage: Option[SwitchState],
-  enableRecaptchaBackend: SwitchState,
-  enableRecaptchaFrontend: SwitchState,
-  experiments: Map[String, ExperimentSwitch],
```
tomrf1, line 13: we no longer have an `experiments` Map. All switches must be properly defined in one of the switch groups in this file

support-frontend/app/admin/settings/Switches.scala
```diff
@@ -1,21 +1,121 @@
 package admin.settings
 
 import com.gu.support.encoding.Codec
-import com.gu.support.encoding.Codec.deriveCodec
+import io.circe.generic.extras.Configuration
+import io.circe.generic.extras.auto._
+import io.circe.{Decoder, Encoder, Json}
+import io.circe.optics.JsonPath._
+
+sealed trait SwitchState {
```
tomrf1, line 9: the `SwitchState` model is more or less as before, but with simpler encoding/decoding

support-frontend/app/admin/settings/Switches.scala
```diff
@@ -1,21 +1,121 @@
...
+   *  }
+   *  To:
+   *  {
+   *    "oneOffPaymentMethods" : {
+   *      "stripe": "On"
+   *    }
+   *  }
+   */
+  private def flattenSwitchState(json: Json): Json =
+    root.state.string
```
tomrf1, line 97: see https://circe.github.io/circe/optics.html if this is unfamiliar

support-frontend/app/admin/settings/Switches.scala
```diff
@@ -1,21 +1,121 @@
...
+            .deepMerge(flattenedSwitches)
+            .remove("switches")
+            .remove("description")
+        }
+        .getOrElse(switchGroup)
+    }
+
+  implicit private val customConfig: Configuration = Configuration.default.withDefaults
+  private val switchesEncoder = Encoder[Switches]
+  private val switchesDecoder = Decoder[Switches].prepare(_.withFocus(flattenAllSwitches))
```
tomrf1, line 118: here we auto-derive a Decoder, but first simplify the json a bit using `prepare`

support-frontend/app/admin/settings/Switches.scala
```diff
@@ -1,21 +1,121 @@
...
+
+sealed trait SwitchState {
+  def isOn: Boolean
+}
+case object On extends SwitchState {
+  def isOn: Boolean = true
+}
+case object Off extends SwitchState {
+  def isOn: Boolean = false
+}
```
johnduffell, line 17: could even simplify further, assuming deriveEnumerationDecoder can handle it

```suggestion
sealed abstract class SwitchState(val isOn: Boolean)
case object On extends SwitchState(true)
case object Off extends SwitchState(false)
```

tomrf1: Nice, it still works!

support-frontend/app/admin/settings/Switches.scala
johnduffell: that is both neat and scary at the same time, I didn't know you could call a dynamic method name in scala!

support-frontend/build.sbt
```diff
@@ -25,6 +25,7 @@ libraryDependencies ++= Seq(
   "io.circe" %% "circe-generic" % circeVersion,
   "io.circe" %% "circe-generic-extras" % circeVersion,
   "io.circe" %% "circe-parser" % circeVersion,
+  "io.circe" %% "circe-optics" % circeVersion,
```
johnduffell, line 28: this isn't going into support-workers so not too important, but I'm interested how much this increases the artifact size.  It seems pull in monocole, cats-macro based on my look at the maven metadata.

support-frontend/test/admin/settings/SwitchesSpec.scala
```diff
@@ -0,0 +1,150 @@
...
+          recurringPaymentMethods = RecurringPaymentMethodSwitches(),
+          subscriptionsSwitches = SubscriptionsSwitches(),
+          featureSwitches = FeatureSwitches(On, On),
+          campaignSwitches = CampaignSwitches(),
+          recaptchaSwitches = RecaptchaSwitches()
+        )
+      ))
+    }
+  }
+}
```
johnduffell, line 150: is it worth having an IT-test to read the actual switches file also?

support-frontend/app/admin/settings/Switches.scala
tomrf1: It's a little [scary](https://github.com/scala/scala/blob/v2.10.1/src/library/scala/Dynamic.scala#L23).
But - here it's not really any less safe than doing it the standard way, which can be a bit fiddly.
If `state` doesn't exist in the json then it quietly leaves it unchanged

support-frontend/build.sbt
tomrf1: Good question.
It's gone from 154M to 155M. I think we can live with this given it's not a lambda?

support-frontend/test/admin/settings/SwitchesSpec.scala
tomrf1: With this implementation, these are the possible scenarios:
1. no json file in S3 - server does not come up.
2. malformed json, or a missing switchgroup in the json - server does not come up.
3. a switch is missing - server comes up and uses the default for that switch.

As I write this, I'm doubting the usefulness of the 3rd point here. If a switch is missing, should it instead fail to come up?

support-frontend/build.sbt
johnduffell: 👍

support-frontend/test/admin/settings/SwitchesSpec.scala
johnduffell: I feel like fail to come up is slightly better for the 3rd point - I remember the time someone accidentally deployed a really old version of frontend, which used default values for a lot of switches and broke the website quite badly.

In our case we could deploy and thought we have provided a switch but it has a typo in the name, and whoever switches it on in the US or wherever would find the switch had no effect.  So not terrible, but not ideal.

The other side is if the switches file is broken in some way we wouldn't be able to deploy or replace unhealthy instances until it was fixed.

johnduffell (approved):
looks good, thanks for doing this redesign of the switches 👍

support-frontend/test/admin/settings/SwitchesSpec.scala
tomrf1: 👍 I'll remove the defaults and treat missing switches as an error

PR #3362, opened 2021-12-03
Refactor StripePaymentRequestButton

TomPretty:

As part of the [`in-epic payments test`](https://trello.com/c/pg579xFK/34-build-one-click-in-epic-payments-test) we need to make the `StripePaymentRequestButton` more reusable. It needs to be rendered both in the `ContributionsForm` and a new, simpler form that we will iframe in to an epic.

The current PRB component combines both core stripe logic, as well as logic specific to the contributions checkout page (e.g don't display it for UK recurring contribs). It also has styles & markup baked in that are specific to the contributions checkout page. 

To decouple the stripe logic from the landing page logic/styles/markup I've applied the [render props](https://reactjs.org/docs/render-props.html) pattern. The `StripePaymentRequestButton` now accepts a render prop (`renderPrb`). It will pass data & callbacks to the render prop, which is then responsible for actually rendering the PRB. This way different consumers of the `StripePaymentRequestButton` can choose to render the button in any way they like.

I've moved the `StripePaymentRequestButton` folder out of the `contributionsLanding` page folder and into the top level shared `components` folder.


support-frontend/assets/components/StripePaymentRequestButton/StripePaymentRequestButton.tsx
```diff
@@ -65,16 +60,33 @@ import {
...
+	paymentRequestError: ErrorReason | null;
+	onStripePrbClick: (
+		event: StripePaymentRequestButtonElementClickEvent,
+	) => void;
+	onCustomPrbClick: (
+		event: StripePaymentRequestButtonElementClickEvent,
+	) => void;
+}
+
+export type RenderPrb = (input: RenderPrbInput) => JSX.Element;
```
TomPretty, line 81: I feel this is maybe slightly ugly, but I added this type & made the input to the render prop a single object so consumers can selectively destructure what they need

support-frontend/assets/components/StripePaymentRequestButton/StripePaymentRequestButton.tsx
```diff
@@ -465,28 +455,17 @@ function PaymentRequestButton(props: PropTypes) {
...
 				}
 
-				const isUkOrEuRecurring =
-					props.contributionType !== 'ONE_OFF' &&
-					(props.countryGroupId === 'GBPCountries' ||
-						props.countryGroupId === 'EURCountries');
-				const shouldShowPrb =
-					Boolean(result?.applePay) ||
-					Boolean(result?.googlePay) ||
-					!isUkOrEuRecurring;
```
TomPretty, line 475: This logic is now in the new `StripePaymentRequestButton` component that is specific to the contribs checkout page

support-frontend/assets/pages/contributions-landing/components/StripePaymentRequestButton.tsx
```diff
@@ -0,0 +1,167 @@
+import { css } from '@emotion/core';
```
TomPretty, line 1: This is the new component that is specific to the `ContributionsForm` component

support-frontend/assets/pages/contributions-landing/components/StripePaymentRequestButton.tsx
```diff
@@ -0,0 +1,167 @@
...
+			selectedAmounts={selectedAmounts}
+			otherAmounts={otherAmounts}
+			renderPrb={({
+				type,
+				paymentRequest,
+				paymentRequestError,
+				onStripePrbClick,
+				onCustomPrbClick,
+			}) => (
+				<Prb
```
TomPretty, line 50: I factored out this `Prb` helper component just to make it a little easier to handle the conditional logic around when to show the button & also when to render the stripe button or a custom one

support-frontend/assets/pages/contributions-landing/components/StripePaymentRequestButton.tsx
```diff
@@ -0,0 +1,167 @@
...
+					onCustomPrbClick={onCustomPrbClick}
+				/>
+			)}
+		/>
+	);
+}
+
+// ---- Helper component ---- //
+
+interface PrbProps extends RenderPrbInput {
```
TomPretty, line 66: Is it lazy to just extend this type here? it's convenient but I don't feel amazing about it

support-frontend/assets/components/StripePaymentRequestButton/StripePaymentRequestButton.tsx
rupertbates: I'm not a big fan of the Prb acronym, I think it's confusing for people unfamiliar with the code base - could we call it `RenderPaymentRequestButton`? It's a bit verbose but I think probably worth it for the clarity

support-frontend/assets/pages/contributions-landing/components/StripePaymentRequestButton.tsx
rupertbates: Same comment about naming

support-frontend/assets/components/StripePaymentRequestButton/StripePaymentRequestButton.tsx
TomPretty: Yeah that's a good suggestion. I'll comb through and replace uses of `Prb`

rupertbates (changes requested):
Looks great, just the comment on using prb as an acronym, I don't think it's a common enough term to abbreviate. Willing to discuss though if you feel strongly!

TomPretty:
> Looks great, just the comment on using prb as an acronym, I don't think it's a common enough term to abbreviate. Willing to discuss though if you feel strongly!

No don't disagree at all! I've updated all the places I used prb 👍

support-frontend/assets/pages/contributions-landing/components/StripePaymentRequestButton.tsx
```diff
@@ -0,0 +1,168 @@
...
+	contributionType,
+	type,
+	paymentRequest,
+	paymentRequestError,
+	onStripeButtonClick,
+	onCustomButtonClick,
+}: PaymentRequestButtonProps) {
+	const countryGroupId = fromCountry(country);
+
+	const isUkOrEuRecurring =
```
tomrf1, line 82: perhaps time to reconsider this logic if we're testing PRB on the epic?

tomrf1 (approved):
great!

support-frontend/assets/pages/contributions-landing/components/StripePaymentRequestButton.tsx
TomPretty: Yeah definitely agree!

rupertbates (approved):
👍 Looks good

PR #3363, opened 2021-12-06
Small PRB props refactor

TomPretty:
## What are you doing in this PR?

Small refactor to the `StripePaymentRequestButtonContainer`s props. Currently it takes the `selectedAmounts` and `otherAmounts` and computes the `amount`. All it actually needs to know is the `amount`. Really `selectedAmounts` & `otherAmounts` should really just be the concern of the form components that render frequency & amount choice cards. I've refactored the ``StripePaymentRequestButtonContainer` to accept just the `amount` as a prop which makes it easier to share going forward (i.e in the new in epic checkout form).

PR #3364, opened 2021-12-08
Stripe payment request button `amount` & `contributionType` props

TomPretty:

In #3363 we added the `amount` prop to the payment request button. However, I missed that the component was still using the `selectedAmounts` and `otherAmounts` from the redux state as it was _also_ getting them from redux directly! I've now deleted them from the `mapStateToProps` function so the component no longer has access and updated where they we used. I've also added `contributionType` to be a prop that is passed by the parent instead of from redux.


support-frontend/typescript-errors.json
```diff
@@ -617,15 +617,15 @@
         },
         {
             "path": "assets/components/subscriptionCheckouts/billingPeriodSelector.tsx",
-            "message": "Type '{ children: Element[]; legend: string; role: string; }' is not assignable to type 'IntrinsicAttributes & Pick<RadioGroupProps, \"label\" | \"children\" | \"id\" | \"error\" | \"className\" | \"name\" | \"cssOverrides\" | \"supporting\"> & Partial<...> & Partial<...>'."
+            "message": "Type '{ children: Element[]; legend: string; role: string; }' is not assignable to type 'IntrinsicAttributes & Pick<RadioGroupProps, \"label\" | \"children\" | \"id\" | \"className\" | \"name\" | \"error\" | \"cssOverrides\" | \"supporting\"> & Partial<...> & Partial<...>'."
```
TomPretty, line 620: Looks like all the diffs here are because the ordering is inconsistent :( I'm not sure if there's much we can do about that?

support-frontend/assets/components/StripePaymentRequestButton/StripePaymentRequestButton.tsx
```diff
@@ -86,22 +84,20 @@ interface PropsFromParent {
...
 	amount: number;
+	contributionType: ContributionType;
 	stripeAccount: StripeAccount;
 	stripeKey: string;
 	renderPaymentRequestButton: RenderPaymentRequestButton;
 }
 
 const mapStateToProps = (state: State, ownProps: PropsFromParent) => ({
-	selectedAmounts: state.page.form.selectedAmounts,
-	otherAmounts: state.page.form.formData.otherAmounts,
```
TomPretty, line 96: before #3363 it was getting this prop both from it's parent & from redux! I think that means it just uses the redux one? either way, it doesn't need it so now it's gone!

rupertbates (approved):
👍

PR #3365, opened 2021-12-08
fix the GA code so gifts are correctly tagged plus fix warnings

johnduffell:

This PR fixes the GA code so it uses a type safe way to determine gifts and reused payment methods.

https://trello.com/c/abNDbB5k/200-reinstate-gifting-tag-for-gifting-campaign-analysis

## Why are you doing this?

We piggy back the GA code via a common `Acquisition` type.  This used to be based on a thrift model in ophan.  Due to the inflexible way we had to define it, the gift subscription flag was stored in an arbitrary text field.  In the GA code we just searched for the string to detect if it was a gift.  Eventually the data ended up in the athena data lake.

For the bigquery table, although the format was retained in general, those arbitrary strings were changed to be more consistent in terms of case.  The GA code was still working off the old strings.

When the old code was removed, the GA code was moved to piggyback off the new types.  Unfortunately at that point the string no longer matched and all subs were detected as non-gift.

This PR uses a more type safe way, because there was no way for the compiler to detect the issue, and it was not noticable during development and review.

## Other ways to make it safer in future

It's hard to detect these kind of issues until they have been in production for a while.  It's cheaper to detect them before too much data has been lost, because backfilling is expensive and sometimes impossible.

- Possible future work to refactor the code to not go via several types along the way.
- Also a test could possible have detected this.
- Also some kind of export from GA into bigquery, followed by a DBT job to check they are similar enough?

### DIFF
--- a/support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/ga/GoogleAnalyticsService.scala
+++ b/support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/ga/GoogleAnalyticsService.scala
@@ -10,6 +10,7 @@ import com.gu.support.acquisitions.ga.models.{ConversionCategory, GAData, GAErro
 import com.gu.support.acquisitions.models.AcquisitionProduct.{Contribution, DigitalSubscription, GuardianWeekly, Paper, RecurringContribution}
 import com.gu.support.acquisitions.models.{AcquisitionDataRow, AcquisitionProduct, PrintOptions, PrintProduct}
 import com.gu.support.acquisitions.utils.Retry
+import com.gu.support.zuora.api.ReaderType
 import com.typesafe.scalalogging.LazyLogging
 import okhttp3.{Call, Callback, HttpUrl, OkHttpClient, Request, RequestBody, Response}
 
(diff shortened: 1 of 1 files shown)


johnduffell:
[Reinstate 'gifting' tag for gifting campaign analysis](https://trello.com/c/abNDbB5k/200-reinstate-gifting-tag-for-gifting-campaign-analysis)

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/ga/GoogleAnalyticsService.scala
```diff
@@ -91,10 +99,10 @@ object GoogleAnalyticsService extends LazyLogging {
           "cd16" -> buildABTestPayload(acquisition.abTests), //'Experience' custom dimension
           "cd17" -> camelCase(acquisition.paymentProvider.map(_.value).getOrElse("")), // Payment method
           "cd19" -> acquisition.promoCode.getOrElse(""), // Promo code
-          "cd25" -> acquisition.labels.exists(_.contains("REUSED_EXISTING_PAYMENT_METHOD")).toString, // usedExistingPaymentMethod
-          "cd26" -> acquisition.labels.exists(_.contains("gift-subscription")).toString, // gift subscription
+          "cd25" -> acquisition.reusedExistingPaymentMethod.toString,
+          "cd26" -> (acquisition.readerType == ReaderType.Gift).toString,
           "cd27" -> productCheckout.getOrElse(""),
-          "cd30" -> acquisition.labels.exists(_.contains("corporate-subscription")).toString, // corporate subscription
+          "cd30" -> (acquisition.readerType == ReaderType.Corporate).toString,
```
johnduffell, line 105: these are the actual important changes

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/ga/GoogleAnalyticsService.scala
```diff
@@ -34,25 +35,27 @@ object GoogleAnalyticsService extends LazyLogging {
...
+        .map(av => buildPayload(acquisition, av, gaData))
+        .map { maybePayload =>
+          maybePayload.map { payload =>
+            logger.info(s"GA payload: $payload")
+            RequestBody.create(null, payload)
+          }
         }
-      }
-  }
+    }
```
johnduffell, line 53: for the linter

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/ga/GoogleAnalyticsService.scala
```diff
@@ -68,7 +71,12 @@ object GoogleAnalyticsService extends LazyLogging {
       case _ => "1"
     }
 
-  private[ga] def buildPayload(acquisition: AcquisitionDataRow, annualisedValue: Double, gaData: GAData, transactionId: Option[String] = None): Either[BuildError, String] = {
+  private[ga] def buildPayload(
+    acquisition: AcquisitionDataRow,
+    annualisedValue: Double,
+    gaData: GAData,
+    transactionId: Option[String] = None
+  ): Either[BuildError, String] = {
```
johnduffell, line 79: for the linter

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/ga/GoogleAnalyticsService.scala
```diff
@@ -124,7 +132,7 @@ object GoogleAnalyticsService extends LazyLogging {
         )
 
         body
-          .filter { case (key, value) => value != "" }
+          .filter { case (_, value) => value != "" }
```
johnduffell, line 135: unused val

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/ga/GoogleAnalyticsService.scala
```diff
@@ -242,22 +250,25 @@ class GoogleAnalyticsServiceLive(client: OkHttpClient) extends GoogleAnalyticsSe
...
+          .post(requestBody)
+          .build()
+
+      for {
+        requestBody <- buildBody(acquisition, gaData)
+        request = buildRequest(requestBody)
+        result <- executeRequest(request)
+      } yield result
+
+    }
```
johnduffell, line 273: for the linter

rupertbates (approved):
👍

PR #3367, opened 2021-12-08
Add `renderFallback` prop to the stripe payment request button

TomPretty:

Adds a new (optional) `renderFallback` prop to the stripe payment request button. The current useage doesn't require one, but for the in epic checkout page we will need render a fallback cta.

I've refactored the type we were using for the `paymentRequestObject`s. Initially we were just using

```ts
PaymentRequest | null
```

but I've now introduced a tagged union type to better model the three states we might be in: `NOT_LOADED` (we haven't created tried to create a `PaymentRequest` object yet), `NOT_AVAILABLE` (we created a `PaymentRequest` object, but its `canMakePayment` method returned `null`), and `AVAILABLE` we're ready to take a payment. Here's the new type:

```ts
export type PaymentRequestObject =
	| { status: 'NOT_LOADED' }
	| { status: 'NOT_AVAILABLE' }
	| { status: 'AVAILABLE'; paymentRequest: PaymentRequest };
```

I also deleted the only reference to `props.stripePaymentRequestButtonData.paymentMethod` that I could find in the code base, so also deleted that from the redux store. @tomrf1 do you know of any places that this is used elsewhere? I removed it in a separate commit so can easily just drop that commit if we do actually need it.

## Screenshots

I rebased my `tp-contrib-checkout-in-epic` branch on top of this one to get some screenshots.

This is the checkout with the payment request available and therefore rendering the `renderPaymentRequestButton` prop:
Screenshot 2021-12-08 at 16 52 12

This is the checkout with the payment request **not** available and therefore rendering the `renderFallback` prop:
Screenshot 2021-12-08 at 16 51 37


support-frontend/assets/components/StripePaymentRequestButton/StripePaymentRequestButton.tsx
```diff
@@ -193,8 +190,11 @@ const onComplete = (res: PaymentResult) => {
 
 function updateTotal(props: PropTypes) {
 	// When the other tab is clicked, the value of amount is NaN
-	if (!Number.isNaN(props.amount) && props.paymentRequestObject) {
-		props.paymentRequestObject.update({
+	if (
+		!Number.isNaN(props.amount) &&
+		props.paymentRequestObject.status === 'AVAILABLE'
```
TomPretty, line 195: Adding the tagged union hasn't added to much extra code. The main difference is that where we were checking that `props.paymentRequestObject` wasn't `null`, we now have to check `props.paymentRequestObject.status === 'AVAILABLE'`. It's more verbose, but I'd say the trade off is worth it

support-frontend/assets/components/StripePaymentRequestButton/StripePaymentRequestButton.tsx
```diff
@@ -476,42 +473,43 @@ function PaymentRequestButton(props: PropTypes) {
 				return stripe.handleCardAction(clientSecret);
 			});
 		}
-
-		props.setPaymentRequestObject(paymentRequestObject);
```
TomPretty, line 480: Before we were always setting the `PaymentRequest` object even if we couldn't make a payment with it. With the new state we only save it if it's available, else we save `{ status: 'NOT_AVAILABLE' }`

support-frontend/assets/components/StripePaymentRequestButton/StripePaymentRequestButton.tsx
```diff
@@ -476,42 +473,43 @@ function PaymentRequestButton(props: PropTypes) {
...
-		props.stripePaymentRequestButtonData.paymentMethod === 'none'
-	) {
+	if (props.paymentRequestObject.status === 'NOT_LOADED') {
 		return null;
 	}
 
+	if (props.paymentRequestObject.status === 'NOT_AVAILABLE') {
+		return props.renderFallback ? props.renderFallback() : null;
+	}
+
```
TomPretty, line 507: I tried initially just keeping the two states (`PaymentRequest | null`) but that meant even if the payment request would eventually be available, we would flash up `renderFallback` prop and then replace it with the `renderPaymentRequestButton` one. Now we can explicitly handle the three separate states.

support-frontend/assets/components/StripePaymentRequestButton/StripePaymentRequestButton.tsx
tomrf1: 👍

support-frontend/assets/pages/contributions-landing/contributionsLandingReducer.ts
```diff
@@ -43,7 +42,6 @@ interface FormData extends UserFormData {
 	checkoutFormHasBeenSubmitted: boolean;
 }
 export interface StripePaymentRequestButtonData {
-	paymentMethod: 'none' | StripePaymentMethod;
```
tomrf1, line 46: presumably tsc would complain if this was still being used anywhere else, so should be fine!

tomrf1 (approved):
Looks good!

So the default epic cta will be rendered by support-frontend - does this mean the epic passes tracking data through into the iframe?

TomPretty:
> So the default epic cta will be rendered by support-frontend - does this mean the epic passes tracking data through into the iframe?

Yeah that was my plan. I was thinking that all that data could go in the query string to the actual `iframe`. In dotcom-components we'd have something like:

```tsx
            <iframe
                src="https://support.thegulocal.com/contribute-in-epic?testName=epic-test-name"
                css={styles.iframeContainer}
                style={{ height: iframeHeight }}
                allow="payment *"
            />
```

but we wouldn't just have `testName`, I guess we'd need the whole `acquisitionData` object?

PR #3371, opened 2021-12-09
Fix checkout validation issue with certain unicode characters

i-hardy:

This amends the 'silly characters' regex in the subscriptions checkout validation to use a specific [unicode property escape](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Unicode_Property_Escapes) for the emoji presentation unicode set, which includes emojis, and characters usually presented as emojis such as the black right-pointing triangle `⏩`

[**Trello Card**](https://trello.com/c/BffkLXIi)

## Why are you doing this?

This fixes a bug affecting users on iOS attempting to enter an apostrophe in their names or addresses. Since iOS 11 the default iOS behaviour is to insert the right single quotation character `’` when the user types an apostrophe instead of the ASCII apostrophe `'`, unless you disable this in your settings.

We have previously experienced issues with Zuora rejecting data for new contributions and subscriptions where supporters had included emoji characters in their personal details. The initial fix for this was to add a checkout validation rule to limit input data to only the ASCII character set. However this is obviously causing issues for users who are simply using default iOS behaviour, and Zuora _do_ accept a wider range of unicode characters including the right single quotation mark.


i-hardy:
[Allow apostrophe in address fields](https://trello.com/c/BffkLXIi/192-allow-apostrophe-in-address-fields)

marialani:
Wondering if we should now change the function from `nonSillyCharacters` to something more emoji specific?

support-frontend/assets/helpers/subscriptionsForms/__tests__/validationTest.ts
```diff
@@ -7,34 +7,41 @@ import {
...
-		it('should return false if string contains a silly character', () => {
+		it('should return false if string contains an emoji', () => {
 			expect(nonSillyCharacters('😊')).toBe(false);
 			expect(nonSillyCharacters('jane✅')).toBe(false);
-			expect(nonSillyCharacters('𝒢℞à©ℨ')).toBe(false);
 		});
+
 		it('should return true if string does not contain silly characters', () => {
 			expect(nonSillyCharacters('joe')).toBe(true);
+			expect(nonSillyCharacters("King's Place")).toBe(true);
```
johnduffell, line 21:
```suggestion
			expect(nonSillyCharacters("King's Place")).toBe(true); // this is a special iOS apostrophe
```

i-hardy: It isn't, but I've just realised there's an issue with the component test that is using a right single quotation character.

johnduffell (approved):
great, this will help these users a lot! :clap:

support-frontend/assets/helpers/subscriptionsForms/__tests__/validationTest.ts
johnduffell: fair enough! in that case don't use that comment :)

PR #3372, opened 2021-12-09
Update payment request button click handler

TomPretty:

Make a small tweak to the PRB click handler. Currently if you set an invalid amount with too many decimals (e.g `£22.222`) and then click the payment request button, it will try to call the `update` method on the `PaymentRequest` object, which will throw an error.

(screenshot from PROD)

Screenshot 2021-12-09 at 16 46 04

This PR changes the click handler so that we run validation on the amount first, before calling `update`.


support-frontend/assets/components/StripePaymentRequestButton/StripePaymentRequestButton.tsx
```diff
@@ -189,11 +189,7 @@ const onComplete = (res: PaymentResult) => {
 };
 
 function updateTotal(props: PropTypes) {
-	// When the other tab is clicked, the value of amount is NaN
-	if (
-		!Number.isNaN(props.amount) &&
-		props.paymentRequestObject.status === 'AVAILABLE'
-	) {
```
TomPretty, line 196: we're now doing the validation before calling this function

support-frontend/assets/components/StripePaymentRequestButton/StripePaymentRequestButton.tsx
```diff
@@ -223,14 +218,15 @@ function getClickHandler(props: PropTypes, isCustomPrb: boolean) {
...
 
-		if (
-			isCustomPrb &&
-			isValid &&
-			props.paymentRequestObject.status === 'AVAILABLE'
-		) {
-			props.paymentRequestObject.paymentRequest.show();
-		} else if (!isCustomPrb && !isValid) {
+		if (!isValid) {
 			event.preventDefault();
```
TomPretty, line 222: If it's invalid we call `event.preventDefault` and then return. This is needed to prevent the stripe elements payment request button from opening the card popup. It's not necessary for a custom payment request button but I think it's fine to call it in that case too anyway

TomPretty:
Was superceded by #3403

PR #3376, opened 2021-12-13
Fix TS errors in user helpers

marialani:
This fixes Typescript and linting errors in the `helpers/user` subfolder.

[**Trello Card**](https://trello.com/c/ouSbxhGi/206-typescript-error-busting-helpers-user)


marialani:
[Typescript Error Busting - helpers/user](https://trello.com/c/ouSbxhGi/206-typescript-error-busting-helpers-user)

support-frontend/assets/helpers/urls/externalLinks.ts
```diff
@@ -12,13 +12,16 @@ import type { ReferrerAcquisitionData } from 'helpers/tracking/acquisitions';
 import { deriveSubsAcquisitionData } from 'helpers/tracking/acquisitions';
 import type { Option } from 'helpers/types/option';
 import { getBaseDomain, getOrigin } from 'helpers/urls/url';
+
 // ----- Types ----- //
-export type MemProduct = 'patrons' | 'events';
```
marialani, line 16: Removed 'patrons' here to fix a typing issue but I believe this is also a remnant from when patrons was on the membership subdomain

support-frontend/assets/helpers/user/__tests__/userReducerTest.ts
```diff
@@ -1,109 +1,127 @@
...
+	const nonExistentAction = {
+		type: 'DOES_NOT_EXIST',
+	};
+
 	it('should return the initial state', () => {
-		expect(createUserReducer(GBPCountries)(undefined, {})).toMatchSnapshot();
+		expect(
+			createUserReducer()(undefined, nonExistentAction as Action),
+		).toMatchSnapshot();
 	});
```
marialani, line 14: `createUserReducer` no longer takes a countryGroupId as a parameter for the initial state following [this pr](https://github.com/guardian/support-frontend/pull/1945)

support-frontend/window.d.ts
```diff
@@ -8,25 +9,28 @@ declare global {
 	 */
 	interface Window {
 		guardian: {
```
marialani, line 11: the diff is a little hard to read as I've alphabetised the properties, here I'd added `user`, `email` and `uatmode`

support-frontend/assets/helpers/user/user.ts
```diff
@@ -143,43 +137,45 @@ const init = (
...
-			setStateField(
-				window.guardian.user.address4 || window.guardian.geoip.stateCode,
-			),
-		);
+		if (address4) {
+			dispatch(setStateField(address4));
+		} else {
+			window.guardian.geoip?.stateCode &&
+				dispatch(setStateField(window.guardian.geoip.stateCode));
+		}
```
johnduffell, line 156: now if the stateCode and address4 are missing it won't dispatch anything, before it would have dispatched undefined or null - is this change important?

support-frontend/window.d.ts
```diff
@@ -8,25 +9,28 @@ declare global {
...
-				uat: string;
-			};
+			productPrices: ProductPrices;
+			recaptchaEnabled?: boolean;
+			serversideTests?: Participations | null;
 			stripeKeyAustralia: StripeKey;
-			stripeKeyUnitedStates: StripeKey;
 			stripeKeyDefaultCurrencies: StripeKey;
+			stripeKeyUnitedStates: StripeKey;
+			uatMode: unknown;
```
johnduffell, line 31: just went to have a look - uatMode should be `true` or `false` by the looks, does unknown mean that it's sometimes missing from the page?

johnduffell (approved):
thanks for busting those errors! 👏

support-frontend/assets/helpers/user/user.ts
marialani: I don't believe so because in our user reducer [here](https://github.com/guardian/support-frontend/blob/c1584b68f25f1a9c87eb463fbb940f88df146b06/support-frontend/assets/helpers/user/userReducer.ts#L47) if the type doesn't match one of the predefined strings then we'd return the default state

support-frontend/window.d.ts
marialani: `unknown` is the type-safe counterpart of any so before we used it we'd have to narrow the type. Wasn't too sure what type to go for here when initially looking at this but will update it to a boolean!

PR #3379, opened 2021-12-14
Remove Christmas eve from GW delivery dates

rupertbates:
We are not delivering Guardian Weekly on the 24th of December so we shouldn't list it as a potential delivery date for regular or gift purchasers


GHaberis (approved):

GHaberis (approved):
👍

PR #3380, opened 2021-12-15
Fix Home Delivery checkout

rupertbates:
Home delivery is currently broken on the paper checkout because there was some code which checks whether home delivery is enabled which is failing.

It is expecting to find a value at `settings.switches.experiments.paperHomeDeliveryEnabled` which does not exist. This may be something to do with the changes to switches which happened recently but for now I am just removing this check to get it working again.


GHaberis (approved):
👍

tomrf1:
Sorry I missed this one, `experiments` doesn't exist anymore. Does this feature still need a switch?

PR #3381, opened 2021-12-16
Fix tracking helpers TS errors

marialani:
This fixes Typescript and linting errors in the `helpers/tracking` subfolder.

[**Trello Card**](https://trello.com/c/USbbS4E3/209-typescript-error-busting-helpers-tracking)


support-frontend/assets/helpers/tracking/acquisitions.ts
```diff
@@ -36,33 +36,33 @@ export type OphanIds = {
...
-	referrerUrl: string | null | undefined;
-	campaignCodes: string[] | null | undefined;
-	componentId: string | null | undefined;
-	componentType: string | null | undefined;
-	source: string | null | undefined;
-	abTests: AcquisitionABTest[] | null | undefined;
-	gaId: string | null | undefined;
-	queryParameters: AcquisitionQueryParameters | null | undefined;
-	labels: string[] | null | undefined;
+	visitId?: string | null;
```
rupertbates, line 53: Do we still need all the nulls here? Could we restrict it to `?: string`?

marialani: I'm not too sure, I'd need to go on a bit of a hunt in the codebase for anywhere that we may explicitly be setting one of these values to null

rupertbates (approved):
👍  Just one question about types

support-frontend/assets/helpers/tracking/acquisitions.ts
marialani: We could get rid a fair few of the nulls here 👍, we only need a few for props that call a shared helper function

PR #3382, opened 2021-12-16
Update nested dependency immer

GHaberis:

Upgrade `immer` to version `9.0.6` to address dependabot alert.

This is a nested dependency via `@storybook/react@5.3.21` which requires `immer@1.10.0` via a transitive dependency on `react-dev-utils@9.1.0`, none of the dependencies in this nested tree have updated `immer` to the version with the resolution (`9.0.6`), so I'm adding this to the `resolutions` section in our `package.json`.


rupertbates (approved):
👍

PR #3383, opened 2021-12-16
upgrade handlebars in support-workers

rupertbates:
Upgrade the handlebars dependency in the support-workers script used to build the cloud formation. 
This gets rid of a critical severity vulnerability which would not actually affect us but it keeps our dependabot report clean


marialani (approved):
👍

PR #3387, opened 2021-12-16
update prismjs

GHaberis:

Updates nested dependency `prismjs` which as a nested dependency of various `storybook` packages. none of the dependencies in this nested tree have updated `prismjs`, so I'm adding this to the `resolutions` section in our `package.json`. I've tested storybook and it looks OK.

```
@storybook/addon-knobs@5.3.21 requires prismjs@~1.17.0 via a transitive dependency on refractor@2.10.1
@storybook/addon-storysource@5.3.21 requires prismjs@~1.17.0 via a transitive dependency on refractor@2.10.1
@storybook/react@5.3.21 requires prismjs@~1.17.0 via a transitive dependency on refractor@2.10.1
```

### DIFF
--- a/support-frontend/package.json
+++ b/support-frontend/package.json
@@ -213,7 +213,8 @@
     "ssri": "^6.0.2",
     "y18n": "^4.0.1",
     "immer": "^9.0.6",
-    "serialize-javascript": "^4.0.0"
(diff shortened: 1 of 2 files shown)


rupertbates (approved):
👍

PR #3388, opened 2021-12-16
Fix TS errors in form helpers

marialani:
This fixes Typescript and linting errors in the `helpers/form` subfolder.

[**Trello Card**](https://trello.com/c/eNEmGna7/212-typescript-error-busting-template-helpers-forms)


marialani:
[Typescript Error Busting Template - helpers/forms](https://trello.com/c/eNEmGna7/212-typescript-error-busting-template-helpers-forms)

support-frontend/assets/helpers/forms/existingPaymentMethods/existingPaymentMethods.ts
```diff
@@ -34,7 +40,11 @@ export type ExistingPaymentMethod =
 function isUsableExistingPaymentMethod(
 	existingPaymentMethod: ExistingPaymentMethod,
 ): boolean {
-	return !!existingPaymentMethod.billingAccountId;
+	if ('billingAccountId' in existingPaymentMethod) {
+		return !!existingPaymentMethod.billingAccountId;
+	} else {
```
GHaberis, line 45: We could do away with the `else` block here and simply `return false` after the `if` block.

support-frontend/assets/helpers/forms/formValidation.ts
```diff
@@ -26,52 +26,73 @@ import type { LocalCurrencyCountry } from '../internationalisation/localCurrency
...
 	!isEmpty(input);
+
 export const isNotTooFarInTheFuture: (arg0: Date | null) => boolean = (
 	date,
 ) => {
 	const rangeDate = new Date();
 	rangeDate.setDate(rangeDate.getDate() + daysFromNowForGift);
+
+	if (!date) {
+		return false;
```
GHaberis, line 43: Is there anywhere we call `isNotTooFarInTheFuture` with `null` currently? It feels like the type signature should prevent this and only allow a `Date` to be passed as a non-optional argument?

support-frontend/assets/helpers/forms/formValidation.ts
```diff
@@ -108,11 +130,10 @@ export const amountOrOtherAmountIsValid = (
 		selectedAmounts[contributionType] &&
 		selectedAmounts[contributionType] === 'other'
 	) {
-		if (
-			otherAmounts[contributionType] &&
-			otherAmounts[contributionType].amount
-		) {
-			amt = otherAmounts[contributionType].amount;
+		if (otherAmounts[contributionType].amount) {
```
GHaberis, line 133: We've removed the check for `otherAmounts[contributionType]` being `undefined`, are we sure there's no way it could be `undefined`?

support-frontend/assets/helpers/forms/paymentIntegrations/oneOffContributions.ts
```diff
@@ -150,13 +176,15 @@ const postOneOffAmazonPayExecutePaymentRequest = (data: AmazonPayData) =>
 const processStripePaymentIntentRequest = (
 	data: CreateStripePaymentIntentRequest,
 	handleStripe3DS: (clientSecret: string) => Promise<Stripe3DSResult>,
-) =>
+): Promise<PaymentResult> =>
 	handleOneOffExecution(
 		postToPaymentApi(data, '/contribute/one-off/stripe/create-payment').then(
 			(createIntentResponse) => {
-				if (createIntentResponse.type === 'requiresaction') {
+				const _createIntentResponse =
```
GHaberis, line 183: Why the use of the proceeding underscore in this variable name?

support-frontend/assets/helpers/forms/paymentIntegrations/oneOffContributions.ts
```diff
@@ -178,33 +206,41 @@ const processStripePaymentIntentRequest = (
...
+	res: Record<string, unknown>,
 ): CreatePayPalPaymentResponse {
-	if (res.data && res.data.approvalUrl) {
-		return {
-			type: 'success',
-			data: {
-				approvalUrl: res.data.approvalUrl,
-			},
-		};
+	if ('data' in res) {
```
GHaberis, line 220: Why `if ('data' in res)` rather than `if (res.data)`?

support-frontend/assets/helpers/forms/paymentIntegrations/payPalRecurringCheckout.ts
```diff
@@ -39,23 +42,24 @@ export type PayPalCheckoutDetails = {
 // ----- Functions ----- //
 function loadPayPalRecurring(): Promise<void> {
 	return new Promise((resolve) => {
-		const script = document.createElement('script');
-		script.onload = resolve;
+		const script: HTMLScriptElement = document.createElement('script');
+		script.onload = () => resolve;
```
GHaberis, line 46: Should this be `script.onload = () => resolve();`, as we want to call `resolve` on `onload`?

support-frontend/assets/helpers/async/fetch.ts
```diff
@@ -4,23 +4,24 @@ import 'helpers/csrf/csrfReducer';
...
+	const resp = await fetch(endpoint, settings);
+	return (await resp.json()) as Record<string, unknown>;
 }
 
 /** Builds a `RequestInit` object for use with GET requests using the Fetch API */
 function getRequestOptions(
 	credentials: Credentials,
 	csrf: CsrfState | null,
-): Record<string, any> {
+): Record<string, unknown> {
```
TomPretty, line 19: Given that we're in control of what we're returning, it might make sense to be more specific with this type? e.g.

```ts
interface GetRequestOptions {
  method: 'GET';
  headers: /* ... */;
  credentials: /* ... */;
}
```

support-frontend/assets/helpers/forms/paymentIntegrations/payPalRecurringCheckout.ts
```diff
@@ -221,21 +225,24 @@ function setupPayment(
 	};
 }
 
-function getPayPalEnvironment(isTestUser: boolean): string {
-	return isTestUser
+const getPayPalEnvironment = (isTestUser: boolean): string =>
```
GHaberis, line 228: There's a mix of function declarations and expressions here, should we settle on one approach for consistency, I believe within the team there was a consensus to go with declarations?

support-frontend/assets/helpers/forms/existingPaymentMethods/existingPaymentMethods.ts
TomPretty: Yeah that makes sense, or maybe 

```ts
('billingAccountId' in existingPaymentMethod) && !!existingPaymentMethod.billingAccountId
```

But that might be a little hard to read

support-frontend/assets/helpers/forms/paymentIntegrations/payPalRecurringCheckout.ts
```diff
@@ -244,14 +251,17 @@ function getPayPalOptions(
 	onPayPalCheckoutCompleted: (arg0: PayPalCheckoutDetails) => void,
 	canOpen: () => boolean,
 	onClick: () => void,
-	formClassName: string,
+	_formClassName: string,
```
GHaberis, line 254: If `_formClassName` isn't used within `getPayPalOptions` then can we stop passing it in?

support-frontend/assets/helpers/forms/recaptcha.ts
```diff
@@ -1,14 +1,12 @@
-const loadRecaptchaV2 = () =>
+const loadRecaptchaV2 = (): Promise<void> =>
 	new Promise<void>((resolve, reject) => {
 		const recaptchaScript = document.createElement('script');
 		recaptchaScript.src =
 			'https://www.google.com/recaptcha/api.js?onload=v2OnloadCallback&render=explicit';
 		recaptchaScript.onerror = reject;
-		recaptchaScript.onload = resolve;
+		recaptchaScript.onload = () => resolve;
```
GHaberis, line 7: Should this be script.onload = () => resolve();, as we want to call resolve on onload?

support-frontend/assets/helpers/forms/paymentIntegrations/payPalRecurringCheckout.ts
```diff
@@ -209,7 +213,7 @@ function setupPayment(
 	billingPeriod: BillingPeriod,
 	setupPayPalPayment: SetupPayPalRequestType,
 ) {
-	return (resolve, reject) => {
+	return (resolve: (arg0: string) => void, reject: (arg0: Error) => void) => {
```
TomPretty, line 216: It might be worth using some more descriptive names here e.g

`reject: (error: Error) => void`?

support-frontend/assets/helpers/forms/recaptcha.ts
TomPretty, line 7: Like above, I think this should probably be 

`() => resolve()`

support-frontend/assets/helpers/forms/formValidation.ts
marialani: `isNotTooFarInTheFuture` is only called by [`checkGiftStartDate`](https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/helpers/forms/formValidation.ts#L69) which can be refactored

support-frontend/assets/helpers/forms/formValidation.ts
marialani: Need to dig into this one again

support-frontend/assets/helpers/forms/paymentIntegrations/oneOffContributions.ts
marialani: Created an intermediary variable to type cast the parameter and went with an underscore as I noticed that convention used elsewhere in the codebase

support-frontend/assets/helpers/forms/paymentIntegrations/oneOffContributions.ts
marialani: Here they can be used interchangably, happy to go with one over the other. With `data' in res` we'd get a boolean but with `res.data` we'd also get a truthy or falsy value

support-frontend/assets/helpers/forms/formValidation.ts
marialani: According to the type information it shouldn't be null but the `OtherAmounts` type had been changed into a Record by the initial migration

support-frontend/assets/helpers/forms/paymentIntegrations/oneOffContributions.ts
marialani: The only place we'd _need_ to use the "in" method is for `if ('billingAccountId' in existingPaymentMethod)`. As the type of the `existingPaymentMethod` parameter is made up of a union type. Might be better to use this alternative method only where necessary so have used dot notation wherever possible.

support-frontend/assets/helpers/forms/paymentIntegrations/payPalRecurringCheckout.ts
marialani: Did we decide on function declarations for react components and constants for non react components?

GHaberis:
@marialani is it possible to resolve the feedback you've responded to and fix the conflicts, then I think this will be good to ship 🚀

GHaberis (approved):
looks good 👍

PR #3389, opened 2021-12-20
Update `animatedDots` component to use emotion

TomPretty:
Update `animatedDots` component to use emotion. I'm using this component on the new in epic checkout page and as part of that work am updating it to use emotion instead of sass.

I was inspired by [this article](https://itnext.io/how-to-increase-css-in-js-performance-by-175x-f30ddeac6bce) on css-in-js performance to make use of data attributes in the html instead of creating functions that dynamically generate the css. Whilst it will have basically zero impact here, I wonder if going forward it would be better for us to adopt this style of approach more?

## Screenshots

Screenshot 2021-12-20 at 12 41 04


support-frontend/assets/components/spinners/animatedDots.tsx
```diff
@@ -1,26 +1,71 @@
...
+import { neutral } from '@guardian/src-foundations';
 import React from 'react';
-import { classNameWithModifiers } from 'helpers/utilities/utilities';
-import './animatedDots.scss';
+
 // ----- Component ----- //
+
 type PropTypes = {
 	appearance: 'light' | 'medium' | 'dark';
-	modifierClasses: Array<string | null | undefined>;
```
TomPretty, line 8: This prop was unused

support-frontend/assets/components/spinners/animatedDots.tsx
```diff
@@ -1,26 +1,71 @@
...
+
+		&[data-bounce='1'] {
+			animation-delay: -0.32s;
+		}
+		&[data-bounce='2'] {
+			animation-delay: -0.16s;
+		}
+		&[data-bounce='3'] {
+			animation-delay: -0.00s;
+		}
```
TomPretty, line 57: I believe by doing it this way instead of e.g 

```ts
bounce: (appearance, offset) => css`
// ...
`
```

Emotion will only have to parse the generated css once, hence the performance boost. In this case, there will be essentially no difference, but perhaps worth structuring our emotion a little more like this going forward?

GHaberis: This is a nice idea, it'd be great to be able to measure any performance gains of this approach vs the function call approach we make now. Like you said it'll probably be minimal in this case, but if we were to adopt it as a convention across the codebase it could have a noticeable difference.

GHaberis (approved):
Looks good!

PR #3392, opened 2021-12-20
Remove hero roundel for free patrons digisub

marialani:
Removing the roundel from the digisub landing page when a free patron digital subscription has been applied.

## Why are you doing this?
The copy for the roundel was previously dictated by the copy that was being fed through from the promocode tool, but as there were some UX issues across different breakpoints the roundel text is now hardcoded.

## Screenshots
| Before | After |
| -- | -- |
| Screenshot 2021-12-20 at 16 07 44 | Screenshot 2021-12-20 at 16 07 14  |


TomPretty (approved):
👍

GHaberis (commented):
Looks good 👍

PR #3398, opened 2022-01-04
add public flag to s3 assets

johnduffell:

This PR adds the publicReadAcl: true parameter to the s3 upload for the assets.

## Why are you doing this?

This is needed because these are served from the web s3 interface direct to fastly.
There is a coming change to riffraff to make the default "false" instead of "true" https://github.com/guardian/riff-raff/pull/664

### DIFF
--- a/support-workers/riff-raff.yaml
+++ b/support-workers/riff-raff.yaml
@@ -26,3 +26,4 @@ deployments:
     parameters:
       bucket: support-workers-dist
       cacheControl: private
+      publicReadAcl: false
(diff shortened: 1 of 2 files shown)


tomrf1 (approved):

support-workers/riff-raff.yaml
```diff
@@ -26,3 +26,4 @@ deployments:
     parameters:
       bucket: support-workers-dist
       cacheControl: private
+      publicReadAcl: false
```
johnduffell, line 29: this is ok because it's granted explicit access
https://github.com/guardian/support-frontend/blob/4afd67abb4a8adb11ce974db31dd36ff180df7fe/support-lambdas/it-test-runner/cfn.yaml#L59

and uses the S3 client to fetch it
https://github.com/guardian/support-frontend/blob/5ddfca4d311ad85a3bfb3d32cba37353f0448246/support-lambdas/it-test-runner/src/main/scala/com/gu/RunITTests.scala#L40

PR #3399, opened 2022-01-04
Update card expiry date for post deploy tests

TomPretty:

We're now using a date that's in the past!


johnduffell (approved):
nice one, thanks! (it's always the date....)

PR #3400, opened 2022-01-06
Add `SELECT_AMOUNTS` action type to contributions LP reducer

TomPretty:

Add `SELECT_AMOUNTS` action type to the contributions LP reducer. I introduced this new action as part of the in epic payment work, but thought I'd break it out into a separate PR and make the refactor to the original contributions LP reducer. It's good practice with redux to minimise the number of actions we dispatch (each dispatch can potentially cause a rerender). On page init, we select default amounts for `ONE_OFF`, `MONTHLY`, and `ANNUAL` contributions. Previously this was done as three separate dispatches of the `SELECT_AMOUNT` action. This new action allows us to do it as a single dispatch of `SELECT_AMOUNTS`.

The action itself actually kicks off a couple of other actions (ie it's a Thunk, not a simple action) so cutting them down saves even more dispatches:

**before**
Screenshot 2022-01-06 at 12 01 58

**after**
Screenshot 2022-01-06 at 12 33 08

I also fixed the 'simple' ts errors ie import order, optional chaining, and nullish coalescing. There are still errors in this file but I think having a go at those too would have made the PR less clear.


support-frontend/assets/pages/contributions-landing/contributionsLandingActions.ts
```diff
@@ -397,6 +405,15 @@ const selectAmount =
...
 
+const selectAmounts =
+	(amounts: SelectedAmounts) =>
+	(dispatch: Dispatch, getState: () => State): void => {
+		setFormSubmissionDependentValue(() => ({
+			type: 'SELECT_AMOUNTS',
+			amounts,
+		}))(dispatch, getState);
+	};
+
```
TomPretty, line 416: I find this really hard to read but I think it's a bigger chunk of work to refactor this pattern. I've just copied the `selectAmount` action and tweaked it. I'll have a go at explaining it for clarity though.

`selectAmounts` is a Thunk creator. The body of the created thunk then calls another Thunk creator `setFormSubmissionDependentValue` (which happens to take a function as it's input) and then immediately invokes the newly created Thunk. 

I think we can definitely improve/simplify this, but that's a bigger job!

support-frontend/assets/pages/contributions-landing/contributionsLandingInit.ts
```diff
@@ -167,32 +164,47 @@ function selectInitialAmounts(
...
+	// The logic in 'amountForContributionType' will select 'other'
+	// in this case. Additionally, we also want to prefill the 'other'
+	// field with the amount in the url.
+	if (
+		amountFromUrl &&
+		amountFromUrl !== 'other' &&
+		!amounts[selectedContributionType].amounts.includes(amountFromUrl)
+	) {
+		dispatch(updateOtherAmount(`${amountFromUrl}`, selectedContributionType));
+	}
```
TomPretty, line 200: By factoring out the `amountForContributionType` function, we now have to have this separate check + dispatch, but I think it makes it a bit more explicit as to what is going on.

johnduffell: fair enough and I agree it makes the dispatches clearer, my only concern here is we have duplicated between line 197 and 172 now, so we need to make sure they keep in step in future.
Not a major thing, but I wonder if there's a way of returning a boolean when the check is done the first time, and once we know it's "other" then we don't need to redo the check?

support-frontend/assets/pages/contributions-landing/contributionsLandingInit.ts
```diff
@@ -167,32 +164,47 @@ function selectInitialAmounts(
...
-				}
+				return amountFromUrl;
 			}
-		} else {
-			const { defaultAmount } = amounts[contributionType];
-			dispatch(selectAmount(defaultAmount, contributionType));
+			return 'other';
 		}
-	});
+		return amounts[contributionType].defaultAmount;
```
johnduffell, line 178: This is not something to stop this PR, but I was trying to understand the wider code and thinking about it, and my comment might be useful:
this code does feel a bit upside down (not directly from your changes, I think just the evolution)
I think what it's trying to do is return the default amount for all other than the "selected-contribution-type" from the url, and for the one that does match, it would read the amount from the url.
I feel like it might be clearer if this just mapped through all the `amounts` array to either pick out the defaultAmount, or if the selected type, then get from the url.
Would it even make sense to just dispatch that object, and then the receiving end could decide whether that is in the amounts for that contribution type?  I'm not sufficiently up on design patterns to know if that makes sense.

johnduffell (approved):
great simplification, I've made some comments, but of course they may be possible future improvements rather than detracting from this PR 👍

support-frontend/assets/pages/contributions-landing/contributionsLandingInit.ts
TomPretty: Okay I had a go at addressing this one and the other comment about duplicating logic and have just pushed up a change. Let me know what you think!

PR #3403, opened 2022-01-14
In Epic checkout form

TomPretty:

Add the new `contribute-in-epic` page. This is a stand-alone slimmed down version of the contributions checkout page that is designed to be embedded into an Epic. This will enable us to take payments directly in Epics, which we will be A/B testing shortly. 

We have a [go-live plan](https://docs.google.com/document/d/1FbnjdkntdUhvYNOP8VLYdWaHNSFV0tOKsjSOjipr5nE/edit#) for releasing the A/B test. The first stage involves merging this PR and testing it. This can be done independently of the work on SDC.

### Saved card - mobile
Screenshot 2022-01-25 at 15 40 12

### Saved card - desktop
Screenshot 2022-01-25 at 15 39 53

### Fallback - mobile
Screenshot 2022-01-25 at 15 41 41

### Fallback - mobile
Screenshot 2022-01-25 at 15 40 49


support-frontend/app/controllers/Application.scala
```diff
@@ -234,6 +239,45 @@ class Application(
...
+    val uatMode = testUsers.isTestUser(request)
+    val guestAccountCreationToken = request.flash.get("guestAccountCreationToken")
+
+    Ok(
+    views.html.contributions(
+      title = "Contributions checkout in Epic",
+      id = s"contributions-landing-page-in-epic-$countryCode",
+      mainElement = assets.getSsrCacheContentsAsHtml("contributions-checkout-in-epic", "contributions-checkout-in-epic.html"),
+      js = Left(RefPath("contributionsCheckoutInEpic.js")),
+      css = Right(StyleContent(Html(""))),
```
TomPretty, line 256: All the styles are done via `emotion` so I don't need any style sheets. This does result in an empty `style` tag being added to the page

TomPretty: I don't love that it does that, but this felt like the quickest/easiest solution

support-frontend/app/controllers/Application.scala
```diff
@@ -234,6 +239,45 @@ class Application(
...
+  def contributionsCheckoutInEpic(countryCode: String): Action[AnyContent] = PrivateAction { implicit request =>
+    implicit val settings: AllSettings = settingsProvider.getAllSettings()
+
+    val geoData = request.geoData
+    val serversideTests = generateParticipations(Nil)
+    val uatMode = testUsers.isTestUser(request)
+    val guestAccountCreationToken = request.flash.get("guestAccountCreationToken")
+
+    Ok(
+    views.html.contributions(
```
TomPretty, line 251: Currently we're just reusing the contribs template, potentially we could create a custom one. It's not straightforward to see exactly what is/isn't required here though

support-frontend/app/controllers/Application.scala
```diff
@@ -234,6 +239,45 @@ class Application(
     )()).withSettingsSurrogateKey
   }
 
+  def contributionsCheckoutInEpic(countryCode: String): Action[AnyContent] = PrivateAction { implicit request =>
```
TomPretty, line 242: This is currently a `PrivateAction` because we were running into issues with the `SetCookiesCheck` filter. However given the potential number of views this will get, perhaps we want to figure out a better solution here?

support-frontend/assets/components/StripePaymentRequestButton/StripePaymentRequestButton.tsx
```diff
@@ -223,14 +218,15 @@ function getClickHandler(props: PropTypes, isCustomPrb: boolean) {
...
-			isValid &&
-			props.paymentRequestObject.status === 'AVAILABLE'
-		) {
-			props.paymentRequestObject.paymentRequest.show();
-		} else if (!isCustomPrb && !isValid) {
+		if (!isValid) {
 			event.preventDefault();
+			return;
+		}
+
```
TomPretty, line 225: This logic is a little confusing because of our branching based on if it's a custom PRB or not. The previous `if/else` block was doing:

```ts
if (isValidCustomPaymentRequest) {
   /* ... */
} else if (isInvalidNonCustomPaymentRequest) {
   /* ... */
}
```

To simplify that a little we now just have a single `isValid` check which returns early if it's not valid. Technically we only need to call `event.preventDefault` when it's not a custom PRB, but it's fine to call for a custom PRB too. 

I've then moved the call to `updateTotal` after this, so that we only every call it when the amount is valid.

Finally we then have the logic for showing the payment request interface for a custom PRB.

support-frontend/assets/components/StripePaymentRequestButton/StripePaymentRequestButton.tsx
```diff
@@ -189,11 +189,7 @@ const onComplete = (res: PaymentResult) => {
 };
 
 function updateTotal(props: PropTypes) {
-	// When the other tab is clicked, the value of amount is NaN
-	if (
-		!Number.isNaN(props.amount) &&
-		props.paymentRequestObject.status === 'AVAILABLE'
-	) {
+	if (props.paymentRequestObject.status === 'AVAILABLE') {
```
TomPretty, line 192: I've simplified this check as now we don't call `updateTotal` until after we've checked that the amount is valid

support-frontend/assets/pages/contributions-checkout-in-epic/ContributionsCheckout.tsx
```diff
@@ -0,0 +1,166 @@
...
+	updateOtherAmount,
+} from 'pages/contributions-landing/contributionsLandingActions';
+import type { State } from 'pages/contributions-landing/contributionsLandingReducer';
+import { ContributionsCheckoutForm } from './ContributionsCheckoutForm';
+import { ContributionsCheckoutSubmitting } from './ContributionsCheckoutSubmitting';
+import { ContributionsCheckoutThankYou } from './ContributionsCheckoutThankYou';
+
+// ---- Component ---- //
+
+type Status = 'FORM' | 'SUBMITTING' | 'SUCCESS';
```
TomPretty, line 26: Perhaps this could be better named - `FORM` really means `The user is interacting with the form`. But I was struggling to think of something nice and succinct!

support-frontend/assets/pages/contributions-checkout-in-epic/ContributionsCheckout.tsx
```diff
@@ -0,0 +1,166 @@
...
+	const { observe, height } = useDimensions({
+		useBorderBoxSize: true,
+		polyfill: ResizeObserver,
+	});
+
+	useEffect(() => {
+		onChange(height);
+	}, [height]);
+
+	return observe;
```
TomPretty, line 151: Previously I had my own custom hook here. According to caniuse `ResizeObserver` has 91.32% which I don't think is high enough for us to use without polyfilling. It also has less support if you want the borderbox sizing. I started extending my custom hook to support optionally polyfilling when `ResizeObserver` isn't available but it became a bit of an async, `useRef`, `useCallback` nighmare. 

I've gone completely in the other direction currently: using a library for the hook & always polyfilling `ResizeObserver`. This has the downside of increasing page weight though. I'd be interested to hear thoughts on this one!

support-frontend/assets/pages/contributions-checkout-in-epic/ContributionsCheckout.tsx
```diff
@@ -0,0 +1,166 @@
...
+	contributionType,
+	amounts,
+	selectedAmounts,
+	otherAmounts,
+	setSelectedContributionType,
+	setSelectedAmount,
+	setOtherAmount,
+}: ContributionsCheckoutProps): JSX.Element {
+	const ref = useOnHeightChangeEffect((height) => {
+		window.parent.postMessage({ type: 'IFRAME_HEIGHT', height }, '*');
```
TomPretty, line 104: Ideally we'd maybe lock the `targetOrigin` here down to `www.theguardian.com` but it makes it a little tricky to test locally in storybook or dcr

support-frontend/assets/pages/contributions-checkout-in-epic/ContributionsCheckoutThankYou.tsx
```diff
@@ -0,0 +1,43 @@
...
+	contributionType: ContributionType;
+	amount: number;
+}
+
+export function ContributionsCheckoutThankYou({
+	currency,
+	contributionType,
+	amount,
+}: ContributionsCheckoutThankYouProps): JSX.Element {
+	// TODO - check this is working in CODE
```
TomPretty, line 23: And PROD!

support-frontend/app/controllers/Application.scala
TomPretty: This is no longer empyt as it was a convenient place to remove the default margins added to the body element!

support-frontend/app/controllers/Application.scala
TomPretty: Having thought about this a little more I think it's probably simplest & easiest to just reuse the existing template

support-frontend/app/controllers/Application.scala
TomPretty: (ie we want it to be cached?)

support-frontend/assets/pages/contributions-checkout-in-epic/ContributionsCheckoutForm.tsx
```diff
@@ -0,0 +1,266 @@
...
+
+export const styles = {
+	prbContainer: css`
+		margin-top: ${space[6]}px;
+	`,
+	ctasContainer: css`
+		display: flex;
+		flex-direction: column;
+
+		@media (min-width: 520px) {
```
TomPretty, line 211: I've used specific media queries here which is pretty gross. I had to do this to accomodate the width of the CTA that contains the payment icons. More generally the media queries in this page don't actually with the breakpoints on dotcom when it's inside an epic. As this page is embedded within the epic inside of an iframe, the widths are dictated by the width of the container that holds the iframe. We effectivley get css container queries here, but that makes the source breakpoints generally less helpful

support-frontend/assets/pages/contributions-checkout-in-epic/ContributionsCheckout.tsx
i-hardy: Possibly something like `INPUT` or `ENTRY`? I don't think `FORM` is a huge problem though.

support-frontend/assets/pages/contributions-checkout-in-epic/ContributionsCheckout.tsx
```diff
@@ -0,0 +1,171 @@
...
+					amount={getAmount(selectedAmounts, otherAmounts, contributionType)}
+				/>
+			)}
+		</div>
+	);
+}
+
+// ---- Hooks ---- //
+
+function useOnHeightChangeEffect(onChange: (height: number) => void) {
```
i-hardy, line 145: I think it would be good to move this into `assets/helpers/customHooks` rather than having it inline in this file- makes it more discoverable if we might want to reuse it elsewhere.

support-frontend/assets/pages/contributions-checkout-in-epic/ContributionsCheckoutForm.tsx
i-hardy: I think it would help to have a named variable for the width that gives this context, maybe with a comment too if you feel like it's necessary, rather than just having a magic 520px in the code.

i-hardy (approved):
I've suggested a couple of improvements but nothing major, otherwise this is looking really good!

support-frontend/assets/pages/contributions-checkout-in-epic/ContributionsCheckoutForm.tsx
TomPretty: Yeah I've done just that - thanks!

support-frontend/assets/pages/contributions-checkout-in-epic/ContributionsCheckout.tsx
TomPretty: Yeah great suggestion - I've moved it over and added just a little comment to describe what it does

PR #3407, opened 2022-01-18
Fix paypal issue

GHaberis:

In https://github.com/guardian/support-frontend/pull/3388 we removed an unused argument `formClassName` from the `getPayPalOptions` function signature, however we continued passing it in where `getPayPalOptions` is called.

This reinstates it with an underscore to denote it's unused. We should look at whether we can stop passing it when calling `getPayPalOptions` after this fix is deployed.


tomrf1 (approved):
👍

TomPretty (approved):
Great spot 👍

PR #3408, opened 2022-01-18
Prevent street names longer than 100 characters

i-hardy:
Co-authored with @frederickobrien 

This PR adds a validation rule for the length of the first and second lines of the address (either billing or delivery) in the subscriptions checkouts, to prevent values that are longer than 100 characters from being entered. The validation function is flexible so should be able to be re-used with other max length limits in future if necessary.

[**Trello Card**](https://trello.com/c/o2tH7mJH)

## Why are you doing this?
Zuora does not allow the value of `StreetName` to be longer than 100 characters; this validation should prevent users from entering address values that are too long and causing a failure in the back end.


i-hardy:
[StreetName should not be longer than 100 characters](https://trello.com/c/o2tH7mJH/160-streetname-should-not-be-longer-than-100-characters)

rupertbates (approved):
👍 If you wanted to be double sure you could add validation on the backend as well here: https://github.com/guardian/support-frontend/blob/861e43966a92ee72503729a15770e4568fee2eb3/support-frontend/app/utils/CheckoutValidationRules.scala#L89

PR #3409, opened 2022-01-19
remove code related to the old non guest checkout and remove redundant IDAPI calls

johnduffell:

This PR removes a lot of redundant code to do with non-guest checkout.  It also tries to simplify what's left to make it easier to understand, reduce the amount of `Action`s and whatever else in the mix.
It also removes unnecessary duplicate calls to identity on many of the calls.

[**Trello Card**](https://trello.com)](https://trello.com/c/YbcRkg3Y/128-fix-guest-checkout-identity-cookie-create-bug)

## Why are you doing this?

Make things easier to understand, and ultimately fix the identity API bug where there is an invalid cookie on the headers to IDAPI.
The duplicate calls are because we used to validate the identity cookie locally, and then call identity for necessary information.  Now we have to call identity to validate the cookie, so we may as well reuse the data retrieved in the later code.

## Notes
Tested in DEV
- contribution new user CC
- contribution guest user DD
- contribution logged in user paypal
- digi sub guest paypal (with phone no)
- paper delivery guest credit card (with phone no)
- gw gift logged in DD
- DS gift logged in paypal
- ds gift redemption logged in


support-frontend/app/actions/AsyncAuthenticatedBuilder.scala
```diff
@@ -17,31 +15,22 @@ import scala.util.{Failure, Success}
 // to authenticate the user data.
 // TODO: considering porting this to identity-play-auth.
 // scalastyle:on
-class AsyncAuthenticatedBuilder[U](
-    userinfo: RequestHeader => Future[U],
-    defaultParser: BodyParser[AnyContent],
-    onUnauthorized: RequestHeader => Result
```
johnduffell, line 23: the actual changes in this file are removing the unnneeded `onUnauthorized`, and hard coding the type `U`.  The rest is just tidyup

support-frontend/app/actions/CustomActionBuilders.scala
```diff
@@ -1,95 +1,63 @@
...
-
-  def maybeAuthenticatedAction(identityClientId: String = membersIdentityClientId): ActionBuilder[OptionalAuthRequest, AnyContent] =
-    PrivateAction andThen maybeAuthenticated(onUnauthenticated(identityClientId))
-
-  def alarmOnFailure[A](action: Action[A]): EssentialAction =
-    (v1: RequestHeader) => action.apply(v1).map { result =>
-      if (result.header.status.toString.head != '2') {
-        SafeLogger.error(scrub"pushing alarm metric - non 2xx response ${result.toString()}")
+  val MaybeAuthenticatedAction: ActionBuilder[OptionalAuthRequest, AnyContent] =
+    PrivateAction andThen new AsyncAuthenticatedBuilder(asyncAuthenticationService.tryAuthenticateUser, cc.parsers.defaultBodyParser)
```
johnduffell, line 32: the diff is not really lining up, but this top part of the file is mostly deleting authenticated action stuff that's unused.  Below this point I have combined the old `alarmOnFailure` action, plus the `LoggingCirceParser` from the CreateSubscriptionController.  I have also simplified it and made it follow the examples in the documentation in terms of syntax/usage https://www.playframework.com/documentation/2.8.x/ScalaActionsComposition

support-frontend/app/controllers/Application.scala
```diff
@@ -132,10 +130,7 @@ class Application(
     val guestAccountCreationToken = request.flash.get("guestAccountCreationToken")
 
     implicit val settings: AllSettings = settingsProvider.getAllSettings()
-    request.user.traverse[Attempt, IdUser](user => identityService.getUser(user.minimalUser)).fold(
```
johnduffell, line 135: we don't need to getUser again as we already got it as part of the maybe authenticated action

support-frontend/app/controllers/CreateSubscriptionController.scala
```diff
@@ -1,30 +1,37 @@
...
 
+object CreateSubscriptionController {
+
+  sealed abstract class CreateSubscriptionError(message: String)
+  case class ServerError(message: String) extends CreateSubscriptionError(message)
+  case class RequestValidationError(message: String) extends CreateSubscriptionError(message)
+
+  type ApiResponseOrError[RES] = EitherT[Future, CreateSubscriptionError, RES]
+
+}
```
johnduffell, line 33: moved from below as they needn't be in the class itself

support-frontend/app/controllers/CreateSubscriptionController.scala
```diff
@@ -36,63 +43,52 @@ class CreateSubscriptionController(
...
-  ): Future[Result] = {
+        toHttpResponse(errorOrStatusResponse, request.body.product)
 
-    val normalisedTelephoneNumber = NormalisedTelephoneNumber.fromStringAndCountry(request.body.telephoneNumber, request.body.billingAddress.country)
+      }
+    }
 
-    val createSupportWorkersRequest = request.body.copy(
-      telephoneNumber = normalisedTelephoneNumber.map(asFormattedString)
-    )
```
johnduffell, line 74: this normalisation happens in the buildSupportWorkersUser code now

support-frontend/app/controllers/CreateSubscriptionController.scala
```diff
@@ -36,63 +44,52 @@ class CreateSubscriptionController(
...
+      case None => s"Guest User ${request.body.email}"
+      case Some(idAndEmail) => s"User ${idAndEmail.primaryEmailAddress}"
+    }
+    SafeLogger.info(s"$userDesc is attempting to create a new ${request.body.product.describe} [${request.uuid}]")
+  }
 
-    CheckoutValidationRules.validate(createSupportWorkersRequest) match {
-      case Valid =>
-        val result = for {
-          user <- identityService.getUser(idMinimalUser).leftMap(ServerError)
```
johnduffell, line 79: we don't need to getUser any more as we can reuse the one retrieved as part of auth

support-frontend/app/controllers/CreateSubscriptionController.scala
```diff
@@ -168,25 +169,3 @@ class CreateSubscriptionController(
   }
 
 }
-
-class LoggingCirceParser(controllerComponents: ControllerComponents) extends Circe {
```
johnduffell, line 172: when I wrote this code I was just delighted that it actually worked, but having come back I have implemented it as an EssentialAction which is more appropriate, and combined with the AlarmOnError code.

support-frontend/app/controllers/DigitalSubscriptionFormController.scala
```diff
@@ -43,24 +36,8 @@ class DigitalSubscriptionFormController(
 
   def displayForm(orderIsAGift: Boolean): Action[AnyContent] = {
     implicit val settings: AllSettings = settingsProvider.getAllSettings()
-    maybeAuthenticatedAction().async { implicit request =>
-      val maybeIdUser: EitherT[Future, String, Option[IdUser]] = request.user match {
-        case Some(user) =>
-          identityService.getUser(user.minimalUser).map(Some(_))
```
johnduffell, line 49: getUser call is redundant here also

support-frontend/app/controllers/DigitalSubscriptionFormController.scala
```diff
@@ -43,24 +36,8 @@ class DigitalSubscriptionFormController(
...
-        error => {
-          val maybeIdentityId = request.user.map(_.minimalUser.id).getOrElse("unknown identity id")
-          SafeLogger.error(scrub"Failed to display digital subscriptions form for ${maybeIdentityId} due to error from identityService: $error")
-          InternalServerError
-        },
-        user =>
-          if (orderIsAGift)
-            Ok(digitalSubscriptionFormHtml(user, orderIsAGift = true))
-          else
-            Ok(digitalSubscriptionFormHtml(user, orderIsAGift = false))
```
johnduffell, line 62: always nice to find an if(something) then true else false hiding in there! :)

support-frontend/app/controllers/PaperSubscriptionFormController.scala
```diff
@@ -42,20 +37,8 @@ class PaperSubscriptionFormController(
 
   def displayForm(): Action[AnyContent] = {
     implicit val settings: AllSettings = settingsProvider.getAllSettings()
-    maybeAuthenticatedAction().async{ implicit request =>
-        val maybeIdUser: EitherT[Future, String, Option[IdUser]] = request.user match {
-          case Some(user) =>
-            identityService.getUser(user.minimalUser).map(Some(_))
```
johnduffell, line 48: don't need to call idapi, just use the maybeAuth one

support-frontend/app/controllers/UserDigitalSubscription.scala
```diff
@@ -1,37 +0,0 @@
...
-import com.gu.monitoring.SafeLogger
-import play.api.mvc.Result
-import play.api.mvc.Results._
-import play.mvc.Http.Status.FOUND
-import services.{AccessCredentials, AuthenticatedIdUser, MembersDataService}
-import SafeLogger._
-import scala.concurrent.{ExecutionContext, Future}
-
-object UserDigitalSubscription {
-  def userHasDigitalSubscription(
```
johnduffell, line 13: we don't actually use this code any more, I think this was to stop duplicate purchases

support-frontend/app/services/AuthenticationService.scala
```diff
@@ -27,45 +28,37 @@ object AccessCredentials {
...
-      .map { case (credentials, user) => buildAuthenticatedUser(credentials, user) }
-      .unsafeToFuture()
-
-  def tryAuthenticateUser(requestHeader: RequestHeader): Future[Option[AuthenticatedIdUser]] =
-    authenticateUser(requestHeader)
-      .map(user => Option(user))
-      .handleError { err =>
-        logUserAuthenticationError(err)
-        None
+          .map { case (_, user) => user }
```
johnduffell, line 38: we used to need the credential object, but this is not used any more, so I flattened out the AuthenticatedIdUser object completely

support-frontend/app/services/AuthenticationService.scala
```diff
@@ -27,45 +28,37 @@ object AccessCredentials {
...
+          .unsafeToFuture()
+      .map(user => Some(user))
+      .recover {
+        case _: UserCredentialsMissingError => None // user not signed in
+        case IdentityClient.Errors(List(Error("Access Denied"))) => None // invalid SC_GU_U cookie?
```
johnduffell, line 43: this situation happens sometimes with users and it means they are not logged in/their cookies has been invalidated

support-frontend/app/services/AuthenticationService.scala
```diff
@@ -27,53 +27,41 @@ object AccessCredentials {
...
+      .recover {
+        case _: UserCredentialsMissingError =>
+          // user not signed in https://github.com/guardian/identity/pull/1578
+          None
+        case IdentityClient.Errors(List(Error("Access Denied"))) =>
+          // scalastyle:off
+          // invalid SC_GU_U cookie?
+          // https://github.com/guardian/identity/blob/424b5170f1b892778fe94f9b3d9a540fb5181e9d/identity-model/src/main/scala/com/gu/identity/model/Errors.scala#L60
+          // scalastyle:on
+          None
```
johnduffell, line 47: we now detect the invalidated session and don't log that situation

support-frontend/app/services/AuthenticationService.scala
```diff
@@ -27,53 +27,41 @@ object AccessCredentials {
...
 
 class AsyncAuthenticationService(identityPlayAuthService: IdentityPlayAuthService)(implicit ec: ExecutionContext) {
 
-  import AsyncAuthenticationService._
-
-  def authenticateUser(requestHeader: RequestHeader): Future[AuthenticatedIdUser] =
+  def tryAuthenticateUser(requestHeader: RequestHeader): Future[Option[User]] =
     identityPlayAuthService.getUserFromRequest(requestHeader)
-      .map { case (credentials, user) => buildAuthenticatedUser(credentials, user) }
+      .map { case (_, user) => user }
```
johnduffell, line 35: we now keep the whole `User` for reuse in the controllers, rather than throwing most of it away

support-frontend/app/services/HttpIdentityService.scala
```diff
@@ -105,32 +105,17 @@ class IdentityService(apiUrl: String, apiClientToken: String)(implicit wsClient:
       .subflatMap(resp => resp.json.validate[CreateSignInTokenResponse].asEither.leftMap(_.mkString(",")))
   }
 
-  private def getHeaders(request: RequestHeader): List[(String, String)] = List(
-    "X-GU-ID-Client-Access-Token" -> Some(s"Bearer $apiClientToken"),
-    "X-GU-ID-FOWARDED-SC-GU-U" -> request.cookies.get("SC_GU_U").map(_.value),
```
johnduffell, line 110: now we only call getUserIdFromEmail, so the user would not be logged in anyway (guest mode), so forwarding the SC_GU_U could only cause problems, not benefits

support-frontend/app/services/MembersDataService.scala
```diff
@@ -1,88 +0,0 @@
-package services
```
johnduffell, line 1: this was unused in the end

support-frontend/app/services/stepfunctions/SupportWorkersClient.scala
```diff
@@ -32,7 +33,7 @@ object CreateSupportWorkersRequest {
 
   implicit val giftRecipientCodec: Codec[GiftRecipientRequest] = deriveCodec
 
-  implicit val codec: Codec[CreateSupportWorkersRequest] = deriveCodec
+  implicit val codec: Decoder[CreateSupportWorkersRequest] = deriveDecoder[CreateSupportWorkersRequest]
```
johnduffell, line 36: CreateSupportWorkersRequest is the data that comes from typescript to the backend, so there was no need for an encoder

support-frontend/app/utils/CheckoutValidationRules.scala
```diff
@@ -75,6 +78,10 @@ object PaidProductValidation {
     case stripeDetails: StripeSourcePaymentFields => stripeDetails.stripeToken.nonEmpty.otherwise("stripe token missing")
     case payPalDetails: PayPalPaymentFields => payPalDetails.baid.nonEmpty.otherwise("paypal BAID missing")
     case existingDetails: ExistingPaymentFields => existingDetails.billingAccountId.nonEmpty.otherwise("existing billing account id missing")
+    case AmazonPayPaymentFields(amazonPayBillingAgreementId) => amazonPayBillingAgreementId.nonEmpty.otherwise("amazonPayBillingAgreementId missing")
+    case SepaPaymentFields(accountHolderName, iban) =>
+      accountHolderName.nonEmpty.otherwise("sepa account holder name missing") and
+        iban.nonEmpty.otherwise("sepa iban empty")
```
johnduffell, line 84: this was a non exhaustive pattern match so I fixed it up

support-frontend/app/utils/NormalisedPhoneNumber.scala
```diff
@@ -11,36 +11,27 @@ case class NormalisedTelephoneNumber(countryCode: String, localNumber: String)
 
 object NormalisedTelephoneNumber {
 
-  def asFormattedString(normalisedTelephoneNumber: NormalisedTelephoneNumber): String =
+  private def asFormattedString(normalisedTelephoneNumber: NormalisedTelephoneNumber): String =
```
johnduffell, line 14: tidy up in this file to make it a single thing to call from the one place it's used in the code.  Now we just have to call `formatFromStringAndCountry` once and we're done.  The last method is used in tests so I kept it mostly as is.

support-frontend/app/controllers/CreateSubscriptionController.scala
```diff
@@ -142,16 +139,20 @@ class CreateSubscriptionController(
...
       title = request.title,
       firstName = request.firstName,
       lastName = request.lastName,
       billingAddress = request.billingAddress,
       deliveryAddress = request.deliveryAddress,
-      telephoneNumber = request.telephoneNumber,
+      telephoneNumber = for {
+          phoneNo <- request.telephoneNumber
+          updatedNo <- NormalisedTelephoneNumber.formatFromStringAndCountry(phoneNo, request.billingAddress.country)
+            .tap(_.left.foreach(SafeLogger.warn)).toOption
```
johnduffell, line 154: we now get an Either from the normaliser, and can emit the warning at the point we discard the error, which makes more sense to me.

support-frontend/app/wiring/Controllers.scala
```diff
@@ -28,10 +26,9 @@ trait Controllers {
     appConfig.amazonPayConfigProvider,
     appConfig.recaptchaConfigProvider,
     paymentAPIService,
-    membersDataService,
+    appConfig.membersDataServiceApiUrl,
```
johnduffell, line 29: we still need the URL, as we embed it in the contributions page, then the client side can call members data api for 
- sendGetExistingPaymentMethodsRequest
- fetchLatestOneOffContribution
- recurringContributionPaymentPlan

support-frontend/app/actions/AsyncAuthenticatedBuilder.scala
rupertbates: I'm a big fan of removing unnecessary generics 👏

support-frontend/app/controllers/Application.scala
```diff
@@ -46,10 +43,9 @@ class Application(
   amazonPayConfigProvider: AmazonPayConfigProvider,
   recaptchaConfigProvider: RecaptchaConfigProvider,
   paymentAPIService: PaymentAPIService,
-  membersDataService: MembersDataService,
+  mdapiUrl: String,
```
rupertbates, line 46: Could we call this `membersDataApiUrl` just you don't have to mentally unpack the acronym whilst scanning code?

johnduffell: we could and normally I'd prefer that - I only used that strange version because it is what the client side uses - making it easier to search the code.  What do you think?  (other option is to also change the client side at the same time but maybe not for this PR!)

support-frontend/app/services/stepfunctions/SupportWorkersClient.scala
rupertbates: maybe call the val `decoder` rather than `codec`? This is a nitpick though.

rupertbates (approved):
Great tidy up, well done!

support-frontend/app/controllers/Application.scala
rupertbates: Yes I did notice that. I'd probably still change it but it's up to you

johnduffell:
I checked back after a weekend and it looked like the remaining ones had had checkouts also.  Hopefully no further issues!

PR #3410, opened 2022-01-20
Fix redemption telephone field

TomPretty:
Fix redemption telephone field - we were updating the confirm email!


rupertbates (approved):
Nice and simple!

PR #3411, opened 2022-01-20
don't send PROD PaymentSuccess for test users

johnduffell:

This Pr prevents us sending PROD PaymentSuccess for test users.
We will send them as stage `PROD-UAT` instead

## Why are you doing this?

We have a bunch of alarms looking for PROD payment success events, however it turns out test users (e.g. post deploy tests) will trigger the same event, causing the alarms to not be fired if we have done deploys recently!

### DIFF
--- a/support-services/src/main/scala/com/gu/aws/AwsCloudWatchMetricSetup.scala
+++ b/support-services/src/main/scala/com/gu/aws/AwsCloudWatchMetricSetup.scala
@@ -25,15 +25,17 @@ object AwsCloudWatchMetricSetup {
       )
     )
 
-  def paymentSuccessRequest(stage: Stage, paymentProvider: PaymentProvider, productType: ProductType): MetricRequest =
+  def paymentSuccessRequest(stage: Stage, isTestUser: Boolean, paymentProvider: PaymentProvider, productType: ProductType): MetricRequest = {
+    val qualifiedStage = stage.toString + (if (isTestUser) "-UAT" else "")
     getMetricRequest(
       MetricName("PaymentSuccess"),
       Map(
         MetricDimensionName("PaymentProvider") -> MetricDimensionValue(paymentProvider.name),
         MetricDimensionName("ProductType") -> MetricDimensionValue(productType.toString),
-        MetricDimensionName("Stage") -> MetricDimensionValue(stage.toString)
+        MetricDimensionName("Stage") -> MetricDimensionValue(qualifiedStage)
       )
     )
+  }
 
   def createSetupIntentRequest(stage: Stage, mode: String): MetricRequest =
     getMetricRequest(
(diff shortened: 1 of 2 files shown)


support-services/src/main/scala/com/gu/aws/AwsCloudWatchMetricSetup.scala
```diff
@@ -25,15 +25,17 @@ object AwsCloudWatchMetricSetup {
       )
     )
 
-  def paymentSuccessRequest(stage: Stage, paymentProvider: PaymentProvider, productType: ProductType): MetricRequest =
+  def paymentSuccessRequest(stage: Stage, isTestUser: Boolean, paymentProvider: PaymentProvider, productType: ProductType): MetricRequest = {
+    val qualifiedStage = stage.toString + (if (isTestUser) "-UAT" else "")
```
johnduffell, line 29: happy to use something other than UAT.  Also, I considered adding another dimension, but we already have payment method, product, and stage.  Also it would mean updating all the alarms.

But opinions welcome.

support-workers/src/main/scala/com/gu/support/workers/lambdas/SendAcquisitionEvent.scala
```diff
@@ -66,7 +67,7 @@ class SendAcquisitionEvent(serviceProvider: ServiceProvider = ServiceProvider)
   }
 
   private def sendAcquisitionEvent(state: SendAcquisitionEventState, requestInfo: RequestInfo, services: Services) = {
-    sendPaymentSuccessMetric(state)
+    sendPaymentSuccessMetric(state).toEither.left.foreach(SafeLogger.error(scrub"failed to send PaymentSuccess metric", _))
```
johnduffell, line 70: I noticed we were ignoring failed trys here

support-workers/src/main/scala/com/gu/support/workers/lambdas/SendAcquisitionEvent.scala
```diff
@@ -78,9 +79,9 @@ class SendAcquisitionEvent(serviceProvider: ServiceProvider = ServiceProvider)
     } yield result
 
     val result = for {
-      streamResult <- streamFuture
-      bigQueryResult <- biqQueryFuture
-      gaResult <- gaFuture
+      _ <- streamFuture
+      _ <- biqQueryFuture
+      _ <- gaFuture
```
johnduffell, line 84: unrelated - fix unused vals

support-services/src/main/scala/com/gu/aws/AwsCloudWatchMetricSetup.scala
rupertbates: Why not just UAT?

rupertbates (approved):
Just one question

support-services/src/main/scala/com/gu/aws/AwsCloudWatchMetricSetup.scala
johnduffell: since these are technically coming from the PROD environment I would not want to remove that completely, otherwise we would send the same metric from CODE and PROD if they are a test user.  The touchpoint stage vs code stage is already confusing enough!

johnduffell: although having merged it, on reflection we would get hardly any traffic for UAT especially in CODE, and even less likely to be digging into it to debug something.  But anyway it's merged now, happy to +1 a later PR though.

johnduffell:
seems to be working, this is what it looks like after a deploy
image

PR #3415, opened 2022-01-24
use iso country instead of the text to fix braze issue

johnduffell:

Change contributions thank you email to send the currency field as iso (e.g. AUD) instead of the text version (e.g. AU$)

https://trello.com/c/iuiXcPqV/258-make-thank-you-email-send-eg-gbp-instead-of-as-currency

## Why are you doing this?

We noticed the thank you emails were missing the currency symbol.  The code our end didn't seem to have changed in any way for several years, and was expecting the contents of the currency field to be used in the email.  It was previously the ISO code but was changed in 2018 to be the "glyph" as a request from someone https://github.com/guardian/support-workers/pull/114
Current logic in braze is to assume it's an iso currency code and do a look up, so we will match that.

No one on the team that looks after braze seems to know about it so the suggestion is to just fix our end to suit.


kelvin-chappell (approved):
👍

PR #3417, opened 2022-01-26
Fix basic TS errors, warnings, and formatting issues

frederickobrien:
This is my first go at fixing TS errors in the code base. To start with I've focused on replacing the defunct `Node` import with `ReactNode` in files where that's the only error. 

Other changes are limited are for formatting and explicit return types on those same files. The PR reduces the error count by 16.


support-frontend/assets/components/checkoutExpander/checkoutExpander.tsx
```diff
@@ -24,20 +24,23 @@ const copyStyleLabel = css`
...
+function CheckoutExpander({ copy, children }: ExpanderPropTypes): JSX.Element {
+	return (
+		<details css={expander}>
+			<summary css={summary}>
+				<strong css={copyStyleLabel}>{copy}</strong>
+			</summary>
+			<div css={copyStyle}>{children}</div>
+		</details>
+	);
+} // ----- Exports ----- //
```
TomPretty, line 44: Tiny thing, but maybe this should be on its own line? (Realise this is how it was before)

TomPretty (approved):
Great stuff! 🚀

support-frontend/assets/components/checkoutExpander/checkoutExpander.tsx
frederickobrien: Ah you're right, good catch. Will update now!

PR #3418, opened 2022-01-28
More TypeScript error fixes and tidying

frederickobrien:
Continuing to chip away at TypeScript errors and formatting issues. Much like [the first round of fixes](https://github.com/guardian/support-frontend/pull/3417) this focuses mainly on updating defunct `Node` imports, replacing them with `ReactNode`. Again return types have been added and formatting tidied up in all files with fixed errors.

There are a couple of slightly bigger changes concerning types - I will annotate them.

This PR reduces the TS error count by 12. Only around 560 to go!

PR #3419, opened 2022-01-31
Address SVG attribute errors

frederickobrien:
Several TS errors were related to `title` being used as an SVG attribute, [which doesn't exist](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute). Instead I've followed the recommendations of [this article on accessible SVGs](https://css-tricks.com/accessible-svgs/) which says to use `<title>` and `<desc>` element inside the `<svg>`. The title is shown as a tooltip when hovered over, which seemed a bit much for what is essentially an alt text description, so I've opted for `<desc>`. 

I've also added the `role="img"` attribute as [it helps screen readers to see SVGs as single cohesive images](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/img_role).

As in previous TS error busting PRs (https://github.com/guardian/support-frontend/pull/3417, https://github.com/guardian/support-frontend/pull/3418) this adds return types where they're missing in the affected files. 

This reduces the TS error count by seven.


i-hardy (approved):
Fantastic 👍

PR #3420, opened 2022-01-31
TS error squashing

TomPretty:

Fix TS errors in:

- `ContributionFormFields.tsx`
- `ContributionState.tsx`
- `Signout.tsx`

[**Trello Card**](https://trello.com/c/5TuheIMI/263-typescript-error-busting-contributionformfieldstsx)


support-frontend/assets/pages/contributions-landing/components/ContributionFormFields.tsx
```diff
@@ -22,22 +23,20 @@ import {
...
 	checkoutFormHasBeenSubmitted: boolean;
 	isSignedIn: boolean;
-	updateFirstName: (arg0: Event) => void;
-	updateLastName: (arg0: Event) => void;
-	updateEmail: (arg0: Event) => void;
-	updateBillingState: (arg0: Event) => void;
+	updateFirstName: (firstName: string) => void;
+	updateLastName: (lastName: string) => void;
+	updateEmail: (email: string) => void;
+	updateBillingState: (billingState: string) => void;
```
TomPretty, line 39: I think it's a nicer interface if these just take strings rather than the entire event object

support-frontend/assets/pages/contributions-landing/components/ContributionFormFields.tsx
```diff
@@ -115,18 +121,20 @@ function ContributionFormFields(props: PropTypes) {
...
 					error={
 						checkoutFormHasBeenSubmitted && !checkEmail(email)
 							? 'Please provide a valid email address'
-							: null
+							: undefined
 					}
 					disabled={isSignedIn}
 				/>
 			</div>
-			<Signout isSignedIn />
```
TomPretty, line 128: There's a couple of things going on here that made me a little confused.

First by attempting to just pass the name of a boolean variable, this should be the same as 

```ts
<Signout isSignedIn={true} />
```

which isnt' what we want. But, the component is actually getting its `isSignedIn` prop from `mapStateToProps`, which overrides what's passed from the parent anyway. So I've deleted the prop here, but nothing changes in the behaviour

support-frontend/assets/pages/contributions-landing/components/ContributionFormFields.tsx
frederickobrien: Agreed, this is so much easier to read

support-frontend/assets/pages/contributions-landing/components/ContributionFormFields.tsx
```diff
@@ -115,18 +121,20 @@ function ContributionFormFields(props: PropTypes) {
 					type="email"
 					autoComplete="email"
 					supporting="example@domain.com"
-					onInput={props.updateEmail}
+					onChange={(e) => updateEmail(e.target.value)}
```
tomrf1, line 124: is there any functional impact from changing onInput -> onChange?

TomPretty: Good point - I was going to highlight this one! So in `React`, `onChange` is the same as `onInput` (triggers on every keystroke), which is different to the default DOM `onchange` event (only triggers after exiting the field)

tomrf1 (approved):
great!

PR #3422, opened 2022-01-31
TS error busting

frederickobrien:
A bite-sized PR of routine TS error fixes, whittling it down slowly but surely. It also adds return types and addresses a small camel case typo, changing `viewbox` to `viewBox` for an SVG.


TomPretty (approved):
Nice!

PR #3426, opened 2022-02-01
Interactive table TS error fixes

frederickobrien:
This PR tries to address errors in `interactiveTables.tsx` and its child components. There are a few more changes than in previous error busting efforts, though each quite small.

- Add several `grid-column` and `grid-row` properties to `interactiveTableRow.tsx` as the styling plugin says it improves browser compatability
- The values for `colSpan` and `aria-colspan` have been changed from strings to numbers 

As in previous PRs I have also added return types where possible, though there were a couple of places (`getRows()` and `getLocalisedRows()`) where what's being returned doesn't fit the nice neat `JSX.Element` or similar. Rather than get bogged down I've left them without return types for now.


TomPretty (approved):
Great work!

PR #3427, opened 2022-02-01
TS error squashing - contributionsLanding.tsx

TomPretty:

Fix the errors in `contributionsLanding.tsx`. The errors here were actually all very simple. The most involved was casting the `store.dispatch` to the `ThunkDispatch` type we've been using around the code base.


TomPretty:
> This looks great! Though curious why the GitHub Action giving the error/lint count hasn't run?

Great question - I'm not sure! I'll take a look

PR #3432, opened 2022-02-03
Allow test users on in-Epic checkout page

TomPretty:

I'd left this hardcoded to false! Just a quick fix to get the actual value from redux.


rupertbates (approved):
👍

PR #3433, opened 2022-02-04
Update comparison table and ty page for free patron digisub

marialani:
This PR modifies the interactive comparison table on the digisub landing page to more accurately reflect the free benefits included in the digisub for patrons as well as fixing some related typescript errors. 

Reference to the standard 14 day free trial period has been removed from the table and all subsections within the table will now be ‘checked’ for patrons. The description on the corresponding thank you page has also been updated to remove reference to the trial period and payment in general. 

Helper functions to determine patronage have been included and the relevant parameters have been pulled in from redux state for the landing and checkout pages.

[**Trello Card**](https://trello.com/c/prtTGyth/236-free-patrons-digisub-comparison-table)

## Screenshots
| Landing Page | Thank you Page |
| -- | -- |
| support thegulocal com_uk_subscribe_digital_promoCode=DIGIPATRON (2) | support thegulocal com_subscribe_digital_checkout_promoCode=DIGIPATRON period=Monthly (2) |


marialani:
[Free patrons digisub - comparison table](https://trello.com/c/prtTGyth/236-free-patrons-digisub-comparison-table)

marialani:
[Remove reference to free trial from patrons thank you page](https://trello.com/c/3RNCB14G/218-remove-reference-to-free-trial-from-patrons-thank-you-page)

support-frontend/assets/helpers/patrons.ts
```diff
@@ -0,0 +1,34 @@
...
+	productPrices?: ProductPrices,
+	currencyId?: IsoCurrency,
+): Promotion[] | undefined {
+	if (!productPrices || !currencyId) return;
+
+	const countryGroupName: CountryGroupName = countryGroups[countryGroupId].name;
+	const billingPeriods =
+		productPrices[countryGroupName][NoFulfilmentOptions][NoProductOptions];
+
+	return Object.values(billingPeriods)[0][currencyId].promotions;
```
TomPretty, line 23: I'm slightly unsure how this line works. Is it grabbing either the Annual or Monthly value?

marialani: Yeah here we access the promotions for either the annual or monthly plan dependant on which is the first element in the array. We then pass the output of this function into the `userIsPatron` function where we're only concerned if the `numberOfDiscountedPeriods` is 100 or more, which will be the case for any plan if a user is a patron

TomPretty: Ah okay that makes sense - given the use of this function we don't actually care if we grab the monthly or annual promotions, so it's fine to get either. That being said, I think it might be worth refactoring a little here as that's not captured in the name of the function. I could imagine a few months from now trying to reuse this in a scenario where it is important if we use annual/monthly. Maybe this function should also take an argument that is the frequency?

marialani: Have updated this function to take the billing period as an arg

TomPretty: Nice - yeah I think that makes it align better with the name and makes it more reusable later on!

support-frontend/typescript-errors.json
```diff
@@ -1658,6 +1612,10 @@
         {
             "path": "assets/pages/paper-subscription-checkout/helpers/orderSummaryText.ts",
             "message": "Property 'NoProductOptions' is missing in type '{ Saturday: string; SaturdayPlus: string; Sunday: string; SundayPlus: string; Weekend: string; WeekendPlus: string; Sixday: string; SixdayPlus: string; Everyday: string; EverydayPlus: string; }' but required in type 'Record<ProductOptions, string>'."
+        },
+        {
+            "path": "assets/pages/subscriptions-redemption/subscriptionsRedemption.tsx",
+            "message": "Property 'billingPeriod' is missing in type '{ countryGroupId: CountryGroupId; marketingConsent: Element; }' but required in type 'Omit<Pick<PropTypes, \"paymentMethod\" | \"productPrices\" | \"billingPeriod\" | \"marketingConsent\" | \"countryGroupId\" | \"currencyId\" | \"includePaymentCopy\"> & Partial<...> & Partial<...>, \"paymentMethod\" | ... 1 more ... | \"includePaymentCopy\">'."
```
TomPretty, line 1618: Is this an error we need to fix?

marialani: I've now made the billing period argument optional for the patrons `getPromotion` function. This is one example where we use the a related component but don't need to pass through the billing period!

TomPretty (approved):
Great work! 🚀

PR #3428, opened 2022-02-01
Fix TS errors and refactor the subs checkout direct debit form

i-hardy:
## What are you doing in this PR?

This fixes Typescript errors in the direct debit form component used on the subscriptions checkouts, and refactors it from a class component into a function component. I've also added component tests to the form to support the refactoring, and made some accessibility improvements to the 'playback' stage so a screen reader user will hear the confirmation and be more easily able to confirm or edit their payment details.

[**Trello Card**](https://trello.com/c/qWqpCvbc)

## Why are you doing this?

These files had a large amount of errors, and were previously quite difficult to work with- I've tried to simplify both the client-side input validation and the process of checking account validity so it's easier to follow.

## Accessibility test checklist
 - [x] [Tested with screen reader](https://accessibility.gutools.co.uk/testing/web/screen-readers/)
 - [x] [Navigable with keyboard](https://accessibility.gutools.co.uk/testing/web/keyboard-navigation/)
 - [x] [Colour contrast passed](https://accessibility.gutools.co.uk/testing/web/colour-contrast/)

PR #3434, opened 2022-02-04
Get test user status from user state when submitting a direct debit payment

i-hardy:

This fixes an issue caused by #3428 where a function shared between subscriptions and contributions checkouts was relying on an item in state that only exists on the subscriptions checkout.


tomrf1 (approved):
👍

PR #3436, opened 2022-02-07
Show a more specific error message for users with disallowed email

rupertbates:
There are some email providers which are disallowed by Identity eg. `@0815.ru` addressses. Currently when a user tries to sign up for a contribution or subscription with one of these addresses we return a generic 500 error and trigger an alarm which then needs to be investigated.

This PR takes advantage of some [work done by Identity]( https://github.com/guardian/identity/pull/2023) to return a better error message to users and to avoid triggering an alarm for these cases. 

Unfortunately there are still a lot of email addresses which will trigger the alarm, for example `test@gmail.com123`. This is because Identity's email validation regex doesn't pick up the error in this email address, so it is trying to create the user but failing when it tries to create it in Okta as their email validation is better than what's in identity and these errors are still returned to support-frontend as a generic 500. 

Identity have an outstanding card to return this type of error to us in the same way as they do disallowed email errors, at which point we can test whether this PR works for the new errors as well.

[**Trello Card**](https://trello.com/c/IJC7qudC/147-handle-identity-blocked-email-domain-errors-on-create-endpoint)

## Is this an AB test?
- [ ] Yes
- [x] No

## Screenshots
**The new error message:**
Screen Shot 2022-02-07 at 17 52 47

### DIFF
--- a/support-frontend/app/controllers/CreateSubscriptionController.scala
+++ b/support-frontend/app/controllers/CreateSubscriptionController.scala
@@ -11,6 +11,8 @@ import config.Configuration.GuardianDomain
 import controllers.CreateSubscriptionController._
 import io.circe.syntax._
 import lib.PlayImplicits._
+import models.identity.responses.IdentityErrorResponse.IdentityError
+import models.identity.responses.IdentityErrorResponse.IdentityError.InvalidEmailAddress
 import org.joda.time.DateTime
 import play.api.libs.circe.Circe
 import play.api.mvc._
@@ -54,7 +56,7 @@ class CreateSubscriptionController(
         val errorOrStatusResponse = for {
           userAndEmail <- maybeLoggedInIdentityIdAndEmail match {
             case Some(identityIdAndEmail) => EitherT.pure[Future, CreateSubscriptionError](identityIdAndEmail)
-            case None => getOrCreateIdentityUser(request.body).leftMap(ServerError)
+            case None => getOrCreateIdentityUser(request.body).leftMap(mapIdentityErrorToCreateSubscriptionError)
           }
           _ <- validate(request)
           supportWorkersUser = buildSupportWorkersUser(userAndEmail, request.body, testUsers.isTestUser(request))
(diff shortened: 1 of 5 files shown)


TomPretty (approved):
Nice one!

support-frontend/app/models/identity/responses/IdentityErrorResponse.scala
```diff
@@ -0,0 +1,47 @@
...
+  object IdentityError {
+    object InvalidEmailAddress {
+      val message = "Registration Error"
+      val description = "Please sign up using an email address from a different provider"
+      val errorReasonCode = "invalid_email_address"
+    }
+
+    def isDisallowedEmailError(identityError: IdentityError): Boolean =
+      identityError.message == InvalidEmailAddress.message &&
+      identityError.description == InvalidEmailAddress.description
```
johnduffell, line 43: @rupertbates you only need to look at the "message"="Registration Error", the description is flexible and we designed it to be returned to the client directly

support-frontend/app/controllers/CreateSubscriptionController.scala
```diff
@@ -94,7 +102,7 @@ class CreateSubscriptionController(
       { error =>
         SafeLogger.error(scrub"[${request.uuid}] Failed to create new ${request.body.product.describe}, due to $error")
         error match {
-          case _: RequestValidationError => BadRequest
+          case err: RequestValidationError => BadRequest(err.message)
```
johnduffell, line 105: I think the "create failed" alarm will still happen for a bad request.  Not 100% sure what the best solution is there though.

TomPretty: Yeah I've just investigated a create product failed alarm that was caused by using an email from a blocked domain

Screenshot 2022-03-10 at 10 54 26

So it's still causing an alarm!

TomPretty: What do you think is the best way to prevent the alarm for blocked domains?

TomPretty: (I've created a card for this https://trello.com/c/cipTaMhA/333-prevent-alarm-when-user-signs-up-with-a-blocked-email-domain)

rupertbates: I've changed this now to go back to doing all the alarming in the action. To avoid alarming for disallowed emails I'm passing the error code for 400s through the `result.header.reasonPhrase`

PR #3438, opened 2022-02-08
Fix SBT deprecation warnings

rupertbates:
This PR fixes up all the SBT deprecations which caused a wall of warnings when starting the project up

An example of the warnings:
Screen Shot 2022-02-08 at 14 42 16


tomrf1 (approved):
great!

PR #3439, opened 2022-02-08
emotion 11/source 4 upgrade

TomPretty:

Upgrades the codebase to use emotion 11 & source 4.

[**Trello Card**](https://trello.com/c/GfrvXQ5j/274-upgrade-support-frontend-to-use-emotion-11)

There were a few steps involved, here was the process that I followed:

1. We already had emotion 11 in the project (`@emotion/react`), so first I updated all of the emotion imports in the code base to use this package instead of the old one (`@emotion/core`). I did this following the instructions in the [emotion 11 upgrade guide](https://emotion.sh/docs/emotion-11). It involved addding a new eslint plugin and rule that fixes the imports.

2. Update all the source deps. I did this folling the [source 3 migration guide](https://docs.google.com/document/d/1qmHSCCCRs98LI9vN2B8I0yjqBAsRfJclEqcPma0wqbI/edit#)

3. Everything looked good, but this added lots of typescript errors as emotion 11 updated the way it handles the types for the `css` prop. The best way to handle this was to update the way we handle `jsx`. Previously we were telling typescript to `preserve` the `jsx` and then using `babel` with the `@babel/preset-react` to transform the `jsx` to calls to `React.createElement`. It seems the best way to handle this is to use the [new jsx transform](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html). This involved updating the `jsx` and `jsxImportSource` properties in `tsconfig.json`.

4. This fixed all the type errors, but meant we now had lots of unused `import React from "react"` which created a load of new errors. To remove these I used `react-codemod` and ran the `update-react-imports` mod. 

5. At this point I thought I was all done, but I'd broken all the tests. When `jest` runs all the tests, it doesn't run `tsc`, just `babel`. To fix the tests I then also updated the `babelrc` to use the new `jsx` transform. 

And now it's all good! 

**UPDATE**

6. Based on feedback I then updated from source 3 -> 4. This involved adding some eslint plugins to auto fix the imports and then manually fixing files with errors caused by renames in source 4 (e.g `buttonBrand` -> `buttonThemeBrand`)

...And now hopefully it is all good!


i-hardy:
Couple of things

1) I assume deletion of `import React from 'react'` is due to using Emotion's custom JSX pragma instead. Is this automatically compatible with Preact?
2) CSIT recommend that we upgrade straight to Source 4 rather than upgrading to 3 first- this is a bit more work as they've rejigged the packages, but IIRC there's a codemod script that can help.

TomPretty:
>     1. I assume deletion of `import React from 'react'` is due to using Emotion's custom JSX pragma instead. Is this automatically compatible with Preact?
> 

Yeah exactly - we're now using the `jsx` pragma from `@emotion/react`. That's a good question! I'm assuming it's automatically compatible with Preact given that it's working, but actually struggling to find some explicit documentation that says it will. 

>     2. CSIT recommend that we upgrade straight to Source 4 rather than upgrading to 3 first- this is a bit more work as they've rejigged the packages, but IIRC there's a codemod script that can help.

Oh great - I'll take a look at that now

.nvmrc
```diff
@@ -1 +1 @@
-12.14.1
+12.22.0
```
TomPretty, line 1: This was required for the eslint-plugin to fix the source import errors

support-frontend/babel.config.json
```diff
@@ -1,14 +1,13 @@
 {
   "presets": [
-    "@babel/preset-react",
+    ["@babel/preset-react", { "runtime": "automatic", "importSource": "@emotion/react" }],
```
TomPretty: This was updated for the tests - now we're using `tsc` to transform the `jsx` through the webpack build, but for the tests jest only runs babel, so we need to update the config here too

support-frontend/package.json
```diff
@@ -5,6 +5,7 @@
   "scripts": {
     "clean": "echo 'Cleaning Public Folder' && rimraf public/compiled-assets",
     "validate": "echo 'Validating JS' && npm-run-all 'lint --fix' flow && yarn orphans",
+    "lint:nibble": "eslint-nibble ./assets --ext .ts,.tsx",
```
TomPretty: I've left this is as it was really useful whilst doing this work. It gives a nice interface to running eslint with just one specific rule. It helped me fix up all the errors I was introducing without fixing up all the existing errors too and bloating this PR even more!

support-frontend/tsconfig.json
```diff
@@ -13,7 +13,8 @@
     "isolatedModules": true, // Unconditionally emit imports for unresolved files
     "resolveJsonModule": true, // Include modules imported with .json extension
     "noEmit": true, // Do not emit output (meaning do not compile code, only perform type checking)
-    "jsx": "preserve", // Support JSX in .tsx files (https://github.com/emotion-js/emotion/issues/1748)
+    "jsx": "react-jsx",
+    "jsxImportSource": "@emotion/react",
```
TomPretty: This is what tells `tsc` to now transform the `jsx`, using the special jsx runtime exported from `@emotion/react`

GHaberis (approved):
Nice work, thanks for detailed breakdown of the changes and why they were necessary! 👍

PR #3440, opened 2022-02-09
Remove TS errors from `content.tsx`

frederickobrien:
Tiny PR to remove a couple of errors from `content.tsx`, as well as tidy up formatting.


support-frontend/assets/components/content/content.tsx
```diff
@@ -22,17 +23,17 @@ export const Sides = {
 export type Appearance = $Keys<typeof Appearances>;
 type PropTypes = {
 	appearance: Appearance;
-	id?: Option<string>;
-	children: Node;
-	image: Option<Node>;
+	id?: string | undefined;
```
marialani, line 26: If you wanted to replicate the previous type here I think you'd want 
```suggestion
	id?: string | null;
```

support-frontend/assets/components/content/content.tsx
```diff
@@ -22,17 +23,17 @@ export const Sides = {
...
+	children: ReactNode;
+	image: Option<ReactNode>;
```
marialani, line 28: You could replace this with 
```suggestion
	image: ReactNode | null;
```
then get rid of the Option type in this file

support-frontend/assets/components/content/content.tsx
frederickobrien: Yeah, I wasn't sure how to go about this. The HTML `id` attribute expects the type `string | undefined` or else it kicks up a fuss. This is the error I get with `null` instead of `undefined`:

```
JSX attribute) HTMLAttributes<HTMLDivElement>.id?: string | undefined
Type 'string | null | undefined' is not assignable to type 'string | undefined'.
  Type 'null' is not assignable to type 'string | undefined'.ts(2322)
[index.d.ts(1836, 9): ]()The expected type comes from property 'id' which is declared here on type 'DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>'
```

What do we do in cases like this where the attribute type is set in stone?

support-frontend/assets/components/content/content.tsx
frederickobrien: Ah nice, thank you! Will do

support-frontend/assets/components/content/content.tsx
marialani: Ah okay in that case you should be able to just get rid of the null part and having the question mark should suffice to fit the expected type i.e.
```suggestion
	id?: string;
```

marialani (approved):
LGTM 👍

PR #3442, opened 2022-02-09
Add post-deployment test for contributing with direct debit

i-hardy:

This adds a post-deployment test for setting up a recurring contribution with direct debit.

## Why are you doing this?

There was an issue introduced in #3428 (fixed in #3434 ) which prevented users from contributing via direct debit, and was only noticed the following day due to an alarm. This is the kind of issue that should get caught immediately by post-deployment tests, but we didn't have any covering that payment method.

### DIFF
--- a/support-frontend/project/build.properties
+++ b/support-frontend/project/build.properties
@@ -1,0 +1,1 @@
+sbt.version=1.6.2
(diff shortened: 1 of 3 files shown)


tomrf1 (approved):
great!

PR #3444, opened 2022-02-09
Replace scalastyle with scalafmt

rupertbates:
This PR removes the `scalastyle` plugin which is no longer maintained and replaces it with [scalafmt](https://scalameta.org/scalafmt/)

Scalafmt integrates with both [Intellij and VS Code (via Metals)](https://scalameta.org/scalafmt/docs/installation.html#vs-code) and provides reformatting capabilities (similar to prettier) as well as style checking. It is also the tool used on frontend, so this is good from a consistency point of view. There is a doc with some analysis on the two tools [here](https://docs.google.com/document/d/1u5Wk8htF9MJN2hHpnPdVEd-qykqvZ06c06LOAonI0zo/edit)

The PR contains [one commit with all the actual config changes](https://github.com/guardian/support-frontend/pull/3444/commits/91c97d1bc5940ce5ba49737f5b127dd8a06f654b), and then a second commit which is the result of running the formatter on the entire repo, this has affected 489 files so will take a very long time to check thoroughly but might be worth skimming.

One style point to note is that I have turned on trailing commas so from now on the convention will be for code such as the following:
```scala
    GetAddressIOConfig(
      config.getString("get-address-io-api.url"),
      config.getString("get-address-io-api.key"),
    )
```
Instead of 
```scala
    GetAddressIOConfig(
      config.getString("get-address-io-api.url"),
      config.getString("get-address-io-api.key")
    )
```
This is consistent with our style rules for Typescript and it also matches with what Frontend do.


build.sbt
```diff
@@ -15,6 +15,12 @@ lazy val integrationTestSettings: Seq[Def.Setting[_]] = Defaults.itSettings ++ S
   libraryDependencies += scalatest % "it",
 )
 
+lazy val scalafmtSettings = Seq(
+  (Test / test) := ((Test / test) dependsOn (Test / scalafmtCheckAll)).value,
+  (Test / testOnly) := ((Test / testOnly) dependsOn (Test / scalafmtCheckAll)).evaluated,
+  (Test / testQuick) := ((Test / testQuick) dependsOn (Test / scalafmtCheckAll)).evaluated,
+)
```
kelvin-chappell, line 22: I like this integration of scalafmt checking with the test cycle

kelvin-chappell (approved):
We should do this with all our Scala repos 👍

TomPretty (approved):
It's a good PR when it breaks githubs review UI for me! Looks great to me, I'm a big fan of having a formatter that I can run on save and not have to think about formatting much at all!

build.sbt
johnduffell, line 22: if you want to add this to every project, rather than doing it individually, you can put it in the inThisBuild block on lline 39.  Or just change this to inThisBuild too
```suggestion
InThisBuild(Seq(
  (Test / test) := ((Test / test) dependsOn (Test / scalafmtCheckAll)).value,
  (Test / testOnly) := ((Test / testOnly) dependsOn (Test / scalafmtCheckAll)).evaluated,
  (Test / testQuick) := ((Test / testQuick) dependsOn (Test / scalafmtCheckAll)).evaluated,
))
```

PR #3445, opened 2022-02-10
Fix for running selenium tests against code

rupertbates:
When I tried to run the selenium tests against CODE to test some changes they failed due to an error writing cookies. This PR fixes that and puts a comment in the `DEV.public.conf` file to make it clear how to run the tests against CODE.


TomPretty (approved):
👍

PR #3447, opened 2022-02-14
Fix deprecations in support-frontend

rupertbates:
I'm working my way through the huge number of deprecation warnings in the support-frontend Scala to fix them so that we can see what's going on a bit more during builds. I'm breaking the changes up into a few PRs to reduce the risk and the time it takes to review them!

I've tested this PR on CODE by running the post-deploy tests agains it and everything worked as expected.

[**Trello Card**](https://trello.com/c/vs7jhw7Z/294-fix-scala-deprecation-warnings-in-support-frontend)


johnduffell (approved):
nice I've been thinking we should tidy these warnings up for a while but never got around to it 👏 thanks!

PR #3448, opened 2022-02-14
Fix in-Epic checkout sentry error

TomPretty:

Fixes the in-Epic checkout sentry error when loading the page. This is the second time in quick succession that I've been investigating something and the root issue has been side-effects caused by importing from another file (see [this PR for SDC](https://github.com/guardian/support-dotcom-components/pull/630)). I'm not sure if there's something we should be doing about this to avoid it happening? 

Here is the stack trace of the current issue:

Screenshot 2022-02-14 at 14 40 24

I was a little surprised to see that the `contributionsCheckoutInEpic` page was trying to render the `contributionsLanding` page. 

Turns out that the `ContributionsCheckoutThankYou.tsx` component was importing the `setOneOffContributionCookie` from `contributionsLanding.tsx`, which as a side effect was then attempting to render the page!

What I dislike the most about these kind of bugs is that there so easy to introduce. Should we be doing more to capture which files are pure and therefore safe to import from, and which are side-effecty and not intended to be imported from?


GHaberis (approved):
👍

GHaberis:
```
What I dislike the most about these kind of bugs is that there so easy to introduce. Should we be doing more to capture which files are pure and therefore safe to import from, and which are side-effecty and not intended to be imported from?
```

It's easy to see how this could happen, it looks like the module [contributionsLanding.tsx](https://github.com/guardian/support-frontend/pull/3448/files#diff-1a4787d291658fece3a4f56c9cd7213790c5f5f347a4b9d444e8791b321ba1fa) was intended to have a side effect ([renderPage](https://github.com/guardian/support-frontend/blob/3da83df06b75f6fa63beaf39ab485ce2f1b14a2c/support-frontend/assets/pages/contributions-landing/contributionsLanding.tsx#L138)), and that an export statement was only added to `setCookie` at later date. 

Thinking about how we could we have prevent this...🤔 Could we keep modules pure by preventing function calls at module scope? Are there tools that could warn us if a module contains side effects? And what if some modules have intended side effects, such as bundle-entry-point modules that render a page? In this case is there a way we could restrict the way these module isareused, eg. Prevent imports from them in certain parts of the codebase?

PR #3449, opened 2022-02-14
Set up Redux Toolkit

i-hardy:

This swaps over our Redux setup to use [Redux Toolkit](https://redux-toolkit.js.org/), removing the need to manually set up and configure redux-thunk and the Redux devtools. This doesn't affect anything else about our use of Redux for now.

## Why are you doing this?

This is the first baby step of migrating all our Redux code over to using Toolkit to improve its maintainability and usability for the future.


rupertbates (approved):
👍

PR #3450, opened 2022-02-14
Fix in-Epic checkout fallback cta layout

TomPretty:

Fix in-Epic checkout fallback cta layout

## Screenshots

**before**

Screenshot 2022-02-14 at 16 04 05

**after**

Screenshot 2022-02-14 at 16 03 29


support-frontend/assets/pages/contributions-checkout-in-epic/ContributionsCheckoutForm.tsx
```diff
@@ -313,9 +315,22 @@ export const styles = {
 	`,
 	fallbackCtasContainer: css`
 		display: flex;
+		flex-direction: column;
```
marialani, line 318: Might be worth trying out `flex-wrap: wrap` here instead, think that should save you from having to use the custom cta breakpoint

TomPretty: Yeah that actually works great here! Just to check, is it best to add e.g an 8px top margin to all the children of a `flex-wrap: wrap` container and then add a -8px top margin to the container?

TomPretty: Okay I had a little play around and in the end I went for adding a top & left margin to all the children, and then offsetting those margins with a negative top & left margin on the container. I think that's nice because the effects of all the margins are just "within" the parent and none of the child margins "leak" out of it

TomPretty: What I'm actually trying to recreate here is using the `gap` property in a flexbox, which actually has [more support](https://caniuse.com/flexbox-gap) than I expected, but probably not quite enough? 

If I were to use that though, all we would require is 

```ts
fallbackCtasContainer: css`
  display: flex;
  flex-wrap: wrap;
  gap: ${space[2]}px;
`
```

which would be beaut!

marialani (approved):
LGTM 👍

PR #3452, opened 2022-02-15
Remove some Scala logging which is unnecessary

rupertbates:
The title says it all really, this log line is filling up the logs and doesn't add any value


TomPretty (approved):
👍

PR #3454, opened 2022-02-15
Remove the unused mail permission flags

rupertbates:
We have some marketing consent flags in the data which is sent through to support-workers. These are no longer used as Identity is updated directly and they have been hardcoded to false for a long time.

I am about to look at moving the code which creates an identity account for test users into support-workers so I'm clearing these out now in preparation.


support-services/src/test/scala/com/gu/support/promotions/PromotionServiceSpec.scala
```diff
@@ -19,7 +19,10 @@ class PromotionServiceSpec extends AsyncFlatSpec with Matchers {
     serviceWithFixtures
       .findPromotion(duplicatedPromoCode)
       .left
-      .map({ case code: DuplicateCode => code.copy(debug = "") }) should be(
+      .map({
```
rupertbates, line 22: Fix an incomplete match warning

support-workers/src/test/scala/com/gu/acquisitions/BigQuerySpec.scala
```diff
@@ -38,17 +39,14 @@ class BigQuerySpec extends AsyncFlatSpec with Matchers with LazyLogging {
       .getService
 
   "BigQuery" should "be able to run a query" in {
+
+    val yesterday = ISODateTimeFormat.dateTime().print(DateTime.now().minusDays(1))
     val query =
-      s"""select * from ${AcquisitionEventTable.datasetName}.${AcquisitionEventTable.tableName} where amount = 9999 and event_timestamp > TIMESTAMP("2020-12-14 00:20:00");"""
```
rupertbates, line 42: This test was selecting a massive number of records and then spamming the logs with them. I've updated it to just select records later than yesterday and not to log the values

TomPretty (commented):
LGTM 👍

PR #3455, opened 2022-02-15
Patrons helper function runtime error fix

marialani:
This PR fixes a runtime error highlighted by sentry, which was caused by one of the patrons related helper functions. 

This also fixes the digisub gifting page which was not rendering as a result of the incorrect billing period being passed into the `getPromotions` helper function.


i-hardy (approved):
👍

PR #3460, opened 2022-02-18
Retry failed Identity calls

rupertbates:
We have been seeing frequent timeouts from Identity when creating guest users which has been causing checkouts to fail. This PR introduces some retry functionality so that when a create attempt times out we try again before giving up.

We had considered moving the create guest account step into the support-workers step function but a potential issue would be if we get a user who gives us a bad email address when Identity is down for some reason we will keep retrying until we eventually find out that it's bad, but by that point the user will have gone away, probably thinking that their purchase has succeeded. It's not clear what we can actually do at that point, we won't be able to email them as we normally would on failure because their email is invalid. While this is a bit of an edge case but would almost certainly happen at some point. 

The changes required to move the logic into support-workers are also quite extensive so as a first step we are just introducing some simple retry logic into support-frontend to see if that solves the problem.

[**Trello Card**](https://trello.com/c/xvGTbXUa/295-move-identity-call-at-create-into-support-workers-step-functions)

### DIFF
--- a/support-frontend/app/models/identity/requests/CreateGuestAccountRequestBody.scala
+++ b/support-frontend/app/models/identity/requests/CreateGuestAccountRequestBody.scala
@@ -10,7 +10,7 @@ import play.api.libs.ws.{BodyWritable, InMemoryBody}
 // Example request:
 // =================
 // {
-//   "email": "a@b.com",
+//   "primaryEmailAddress": "a@b.com",
 //   "privateFields": {
 //     "firstName": "a",
 //     "secondName": "b"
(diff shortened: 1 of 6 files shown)


support-frontend/app/models/identity/requests/CreateGuestAccountRequestBody.scala
```diff
@@ -10,7 +10,7 @@ import play.api.libs.ws.{BodyWritable, InMemoryBody}
 // Example request:
 // =================
 // {
-//   "email": "a@b.com",
+//   "primaryEmailAddress": "a@b.com",
```
rupertbates, line 13: This comment was wrong before

PR #3462, opened 2022-02-24
Typescript error busting - contributionsLandingInit.ts

TomPretty:

Fix the errors in `contributionsLandingInit.ts`.


support-frontend/assets/pages/contributions-landing/contributionsLandingInit.ts
```diff
@@ -49,6 +50,8 @@ import {
 } from './contributionsLandingActions';
 import type { State } from './contributionsLandingReducer';
 
+type Action = ContribAction | CommonAction;
+
```
TomPretty, line 54: Is there a better way of doing this? I don't think I've actually had to fix a file before that dispatched actions for reducers defined in multiple files. Here I've had to combine the common & contributionsLanding reducer actions into a single type.

support-frontend/assets/pages/contributions-landing/contributionsLandingInit.ts
```diff
@@ -261,8 +266,8 @@ function selectInitialContributionTypeAndPaymentMethod(
 	return contributionType;
 }
 
-const init = (store: Store<State, Action, (...args: any[]) => any>) => {
-	const { dispatch } = store;
+const init = (store: Store<State, Action>): void => {
+	const dispatch = store.dispatch as ThunkDispatch<State, void, Action>;
```
TomPretty, line 270: I'm not sure if there's a better way of doing this. The `store.dispatch` is of type `Dispatch<Action>`, but we want it to be a `ThunkDispatch`. I couldn't easily find a way of getting that into the `Store<State, Action>` type signature

support-frontend/assets/pages/contributions-landing/contributionsLandingInit.ts
i-hardy: Type issues around `Dispatch` and `Action` are one of the things Redux Toolkit will help us with- it essentially enables you to [infer types for these things from your base store](https://redux-toolkit.js.org/usage/usage-with-typescript). I agree that this is pretty clunky, but I think it's clear and functional enough that it's probably not worth spending more time on if we're going to be adopting Toolkit in the near future.

i-hardy (approved):
👍

support-frontend/assets/pages/contributions-landing/contributionsLandingInit.ts
TomPretty: Oooo just read through the bit about getting the types for `Dispatch` and `Action` and that looks great! Very excited that we're working on adopting this

PR #3525, opened 2022-03-02
Fix integration tests

rupertbates:

Fix a few issues with the integration tests:
- an expired test card
- a failure caused by a new product rate plan being added to Guardian Weekly in DEV


michaelbjacobson (approved):
Looks good to me!

PR #3530, opened 2022-03-03
TS erros fixing - asyncronously.ts

marialani:
This fixes TS errors in the asyncronously.ts component and associated files where it is being used. This component was previously using a class based approach and has also been refactored to use react hooks. 

[**Trello Card**](https://trello.com/c/3Q6tc8jn/328-typescript-error-busting-template-asyncronouslyts)

### DIFF
--- a/support-frontend/assets/components/asyncronously/asyncronously.ts
+++ b/support-frontend/assets/components/asyncronously/asyncronously.ts
@@ -1,5 +1,5 @@
 import type { ComponentType } from 'preact';
-import { useEffect, useState } from 'preact/hooks';
+import { useEffect, useRef, useState } from 'preact/hooks';
 
 type PropTypes<T> = {
 	loader: Promise<{
@@ -10,19 +10,22 @@ type PropTypes<T> = {
 };
 
 function Asyncronously<T>(props: PropTypes<T>): JSX.Element | null {
-	const [component, setComponent] = useState<ComponentType<T>>();
+	const [loaded, setLoaded] = useState(false);
+
+	const componentRef = useRef<ComponentType<T>>();
 
 	useEffect(() => {
 		const { loader } = props;
 		void loader.then((imported) => {
-			setComponent(imported.default);
+			componentRef.current = imported.default;
+			setLoaded(true);
 		});
 	}, []);
 
 	const { loading, render } = props;
 
-	if (component) {
-		return render(component);
+	if (loaded && componentRef.current) {
+		return render(componentRef.current);
 	}
 
 	return loading;
--- a/support-frontend/assets/components/asyncronously/asyncronously.ts
+++ b/support-frontend/assets/components/asyncronously/asyncronously.ts
@@ -1,47 +1,35 @@
-import type { ComponentType, Node } from 'react';
-import { Component } from 'react';
+import type { ComponentType } from 'preact';
+import { useEffect, useState } from 'preact/hooks';
 
-type PropTypes = {
+type PropTypes<T> = {
 	loader: Promise<{
-		default: ComponentType<any>;
+		default: ComponentType<T>;
 	}>;
-	loading: Node;
-	children: (arg0: ComponentType<any>) => Node;
-};
-type State = {
-	loaded: boolean;
+	loading: JSX.Element;
+	render: (MktConsent: ComponentType<T>) => JSX.Element;
 };
 
-class Asyncronously extends Component<PropTypes, State> {
-	static defaultProps = {
-		loading: null,
-	};
-	state = {
-		loaded: false,
-	};
+function Asyncronously<T>(props: PropTypes<T>): JSX.Element | null {
+	const [component, setComponent] = useState<ComponentType<T>>();
 
-	componentDidMount() {
-		const { loader } = this.props;
-		loader.then((imported) => {
-			this.component = imported.default;
-			this.setState({
-				loaded: true,
-			});
+	useEffect(() => {
+		const { loader } = props;
+		void loader.then((imported) => {
+			setComponent(imported.default);
 		});
+	}, []);
+
+	const { loading, render } = props;
+
+	if (component) {
+		return render(component);
 	}
 
-	component = null;
-
-	render() {
-		const { loading, children } = this.props;
-		const { loaded } = this.state;
-
-		if (loaded && this.component) {
-			return children(this.component);
-		}
-
-		return loading;
-	}
+	return loading;
 }
+
+Asyncronously.defaultProps = {
+	loading: null,
+};
 
 export default Asyncronously;
--- a/support-frontend/assets/pages/paper-subscription-checkout/components/thankYou.tsx
+++ b/support-frontend/assets/pages/paper-subscription-checkout/components/thankYou.tsx
@@ -1,6 +1,7 @@
 // ----- Imports ----- //
 import { css } from '@emotion/react';
 import { from, headline, space, textSans } from '@guardian/source-foundations';
+import type { ConnectedProps } from 'react-redux';
 import { connect } from 'react-redux';
 import Asyncronously from 'components/asyncronously/asyncronously';
 import Content from 'components/content/contentSimple';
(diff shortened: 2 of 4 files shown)


support-frontend/assets/components/asyncronously/asyncronously.ts
```diff
@@ -1,47 +1,38 @@
...
-			this.setState({
-				loaded: true,
-			});
-		});
-	}
+function Asyncronously<T>(props: PropTypes<T>): JSX.Element | null {
+	const [loaded, setLoaded] = useState(false);
 
-	component = null;
+	const componentRef = useRef<ComponentType<T>>();
```
marialani, line 15: Replacing class based instance variable `component = null` with a ref

support-frontend/assets/pages/paper-subscription-checkout/components/thankYou.tsx
```diff
@@ -25,19 +25,18 @@ import AppsSection from './appsSection';
 import { HeroPicture } from './heroPicture';
 import SubscriptionsSurvey from './subscriptionSurvey';
 
-type MarketingConsent =
-	typeof import('components/subscriptionCheckouts/thankYou/marketingConsentContainer').default;
```
marialani, line 29: No longer necessary due to generic proptypes in `asyncronously.ts`

support-frontend/typescript-errors.json
```diff
@@ -2,25 +2,25 @@
     "errorCounts": {
         "byCode": {
```
marialani, line 3: Not too sure why this file has jumped around so much but a scan of the errors by code will show that the codes that have changed have all seen a decrease in number

support-frontend/assets/components/asyncronously/asyncronously.ts
```diff
@@ -1,47 +1,38 @@
-import type { ComponentType, Node } from 'react';
-import { Component } from 'react';
+import type { ComponentType } from 'preact';
```
rupertbates, line 1: We don't generally import from `preact`, is there a reason why we need to here?

marialani: No there's no particular reason to do so here. Will revert this change but wondering for future reference why we might not want to import from preact?

support-frontend/assets/components/asyncronously/asyncronously.ts
```diff
@@ -1,47 +1,38 @@
-import type { ComponentType, Node } from 'react';
-import { Component } from 'react';
+import { useEffect, useRef, useState } from 'preact/hooks';
```
TomPretty, line 1: Probably worth just importing this from `react`

support-frontend/assets/components/asyncronously/asyncronously.ts
TomPretty: does this need to be a ref? could we use `useState` instead?

TomPretty: The reason for this is because `imported` _is_ already a function. When you pass a function to `setState` react will invoke the function and then set the returned value as the new state.

marialani: Had taken a similar approach to your suggestion at first but came across some mention online that storing a component in state is bad practice, though I suppose in this case using a ref is only marginally better

support-frontend/assets/components/asyncronously/asyncronously.ts
marialani: Ah yeah that makes sense, though you lost me a little bit at the end there what did you mean by
> ... I suppose it would be possible to drop `preact` and just `react` by just removing the alias.

support-frontend/assets/components/asyncronously/asyncronously.ts
marialani: Ideally we could have used React lazy and suspense here instead but it's still experimental with preact. 

From what I can tell, storing a component in state is mainly discouraged because it's an 'anti pattern'. Though I suppose both approaches would be to some degree. Do you think one approach is more beneficial than the other?

TomPretty: Yeah it does feel like we're rolling our own suspense mode here!

Did anywhere mention why it's an 'anti pattern'? I can see lots of people saying not to do this, but none of their reasons apply to our use case (showing something else whilst we async load in the actual component). 

I'd be tempted to use `useState` over `useRef`. Mainly because that's how we do it for the [epics & banners in DCR](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering%2Fsrc%2Fweb%2Fcomponents%2FStickyBottomBanner%2FReaderRevenueBanner.tsx#L289).

It would also allow to drop the `loaded` state too. I think that's nice because it's generally good to structure your state such that there are no invalid states. Just looking at the state (i.e ignoring the logic in the component), it's possible to have `loaded = false` but have `componentRef.current` set, which is an invalid config. 

If instead we just have some state which is of type `ComponentType | undefined`, it's now no longer possible to have this invalid config. 

This is less explicit though. We're now implicitly treating `undefined` as our loading state. If you want to be more explicit, tagged unions are great for this. You could do something like:

```ts
type Status = { type: 'loading' } | { 'type': 'loaded', component: ComponentType }
```

marialani: Anti pattern wise this is one of the better [explanations I came across](https://stackoverflow.com/a/62913227), essentially if what we return from the render function has side effects then it may produce unexpected results. Having tried both approaches they both seemed to do what we want so if you feel strongly about the changeover that's all good with me. 

Yeah the logic and the state is quite closely coupled in this component at the moment. It did just seem like there was a lot of push back out there for not storing components in state but not much on alternate approaches when you're not using lazy and suspense.

I reckon as it's a fairly small/readable component we don't think we need to be super explicit about the state?

support-frontend/assets/components/asyncronously/asyncronously.ts
```diff
@@ -1,47 +1,34 @@
...
-		const { loading, children } = this.props;
-		const { loaded } = this.state;
+	const { loading, render, loader } = props;
 
-		if (loaded && this.component) {
-			return children(this.component);
-		}
+	useEffect(() => {
+		void loader.then((imported) => {
+			setComponent(imported.default);
```
TomPretty, line 19: I think we'll need to do 

```ts
setComponent(() => imported.default);
```

This is because passing a function to `setComponent` will cause it to call the function and store the return value. So if we don't wrap it, we'll be storing a `JSX.Element`, not the component. 

If we make this change we can then use it like `<Component />` instead of `{Component}`

marialani: Did you try this out locally? It doesn't work for me

TomPretty: I didn't! I only tried it in a fresh `create-react-app` project - I'll have a go locally

TomPretty: Okay - it did work but I had to pass in some extra arguments that are currently not being passed. This worked: 

```ts
render={(MktConsent) => (
  <MktConsent requestPending={false} error={false} />
)}
```

marialani: Are there particular disadvantages you have in mind in regards to returning a JSX.Element vs a component in this instance? It seems to me having to add extra arguments to get this to work may be unnecessary

TomPretty: Yeah I think it's probably best to store the component instead of the jsx.

Setters from `useState` can be passed two type of arguments, either: a new state; or a function mapping from old state to new state. So if I have


```ts
const [count, setCount] = useState(0);
```

We can write

```ts
const increment = () => setCount(count + 1);
```

or 

```ts
const increment = () => setCount((c) => c + 1);
```

The second one allows us to remove a dependence on `count` e.g we could do 

```ts
const increment = useCallback(() => setCount((c) => c + 1), []);
```

If we wanted to just call `setCount(count)`, we'd have to add count to the dependency array, which would mean we recreate the `increment` function every time. 

I don't think we should rely on this behaviour to call `imported.default`. It's quite subtle and easy to miss and also not really what it's intended for. I also think this will break if we import a class component instead of a function one. If we import a class component, I think we'll actually be storing the component instead of jsx (this will break if we try and render it like `{MktConsent}` instead of `<MktConsent />`).

I think storing the component also makes it more flexible and reusable e.g if I have a component

```tsx
function Hello({ name }) {
  return <div>Hello, {name}<div />
}
```

Then we could do 

```tsx
render={(Hello) => <Hello name="Maria" />}
```

or

```tsx
render={(Hello) => <Hello name="Tom" />}
```

depending on the context we're using the component.

marialani: In this instance as the set state function is only called once after the initial render I wouldn’t have thought we’d need to take the callback approach for performance reasons.

Not used imported.default too much myself, would be good to read into its proper usage if you have a reference handy. From what I can tell from this components history, I think this has always been a slightly hacky way to asynchronously render only the marketing consent component on thank you pages but I suppose could be used for other reasons and this possibly breaking if it were to be used with a class component is concerning. 

From one of your earlier messages I think you’d managed to get this working by passing some extra props. I’m not too sure how you’d changed the component to use those props but am more than happy for you to push those changes up to this branch and then we can then rope in another reviewer to have a go over this

TomPretty (approved):
Great work! 🔥 299 errors now!

support-frontend/assets/components/asyncronously/asyncronously.ts
marialani: Happy to merge, I'll sort these conflicts then get this in!

PR #3531, opened 2022-03-03
Move support-frontend build to Github actions

rupertbates:

This PR adds a github action to run the full support-frontend build. This will replace the Teamcity build which we currently have and has the benefit of moving all of our CI logic into source control rather than having it only accessible through the UI of Teamcity.

It also adds a new action which will test PRs from the [Scala steward fork](https://github.com/scala-steward/support-frontend) which is the method Scala steward uses to open PRs against our codebase. Currently these are not tested by CI which makes dealing with them a pain.

[**Trello Card**](https://trello.com/c/s6l5ywJu/330-move-support-fronted-build-to-github-actions)

### DIFF
--- a/.github/workflows/support-frontend-build.yml
+++ b/.github/workflows/support-frontend-build.yml
@@ -1,0 +1,83 @@
+name: Build support-frontend
+
+on:
+  pull_request:
+    branches: [ main ]
+
+env:
+  GU_SUPPORT_WORKERS_LOAD_S3_CONFIG: false
+
+jobs:
+  support_frontend_build:
+    if: github.event.pull_request.head.repo.owner.login == 'guardian'
+
+    # Required by actions-assume-aws-role
+    permissions:
+      id-token: write
+      contents: read
+
+    name: support-frontend build
+    runs-on: ubuntu-latest
+    steps:
+      - name: Env
+        run: env
+
+      - name: Checkout repo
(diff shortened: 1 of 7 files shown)


.github/workflows/support-frontend-build.yml
```diff
@@ -0,0 +1,62 @@
...
+on:
+  pull_request:
+    branches: [ main ]
+
+env:
+  GU_SUPPORT_WORKERS_LOAD_S3_CONFIG: false
+
+jobs:
+  support_frontend_build:
+    if: github.actor != 'scala-steward'
```
kelvin-chappell, line 12: Perhaps this should exclude any PR from a fork, not just scala-steward?  I'm not sure how to identify that at the moment though.
The workflow will fail if the PR is coming from a fork anyway because it won't have access to secrets but would be better if it just skipped the workflow.

kelvin-chappell, line 12: Doing some experiments in guardian/payment-failure-comms#95 where facing the same problem, this seems to work:
```suggestion
    if: github.event.pull_request.head.repo.owner.login == 'guardian'
```

rupertbates: Thanks, that's useful

kelvin-chappell (approved):
👍

PR #3536, opened 2022-03-04
Fix it tests2

rupertbates:

Fixes one integration test and sets another to be ignored

### DIFF
--- a/support-config/src/main/resources/reference.conf
+++ b/support-config/src/main/resources/reference.conf
@@ -5,3 +5,7 @@ include classpath("touchpoint.PROD.conf")
 
 get-address-io-api.url = "https://api.getAddress.io/v2/uk/"
 get-address-io-api.key = "hyj46rv0Uk2D1BtcW5M8jQ17434" //Test key, overridden in PROD
+AVCalculator {
+  roleArn=""
+  lambdaArn=""
+}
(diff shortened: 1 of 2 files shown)


support-config/src/main/resources/reference.conf
```diff
@@ -5,3 +5,7 @@ include classpath("touchpoint.PROD.conf")
 
 get-address-io-api.url = "https://api.getAddress.io/v2/uk/"
 get-address-io-api.key = "hyj46rv0Uk2D1BtcW5M8jQ17434" //Test key, overridden in PROD
+AVCalculator {
```
tomrf1, line 8: in case anyone else wonders, this is expected by a dependency - https://github.com/guardian/acquisitions-value-calculator-client

tomrf1 (approved):
thanks!

support-workers/src/test/scala/com/gu/zuora/ZuoraITSpec.scala
```diff
@@ -50,7 +50,9 @@ class ZuoraITSpec extends AsyncFlatSpec with Matchers {
     }
   }
 
-  it should "retrieve subscription redemption information from a redemption code" in {
+  // I'm going to ignore this test for now since gifts expire after a year and the one we are currently
+  // testing with has just expired
```
johnduffell, line 54: 😂 good old dates! is the long term fix to pass "today" in to `getSubscriptionFromRedemptionCode`

PR #3544, opened 2022-03-07
Add validation to check for emojis in users' names

TomPretty:

Add additional validation on the contrib checkout page to check a user hasn't entered an emoji as part of their name.

[**Trello Card**](https://trello.com/c/OUpj0TPD/306-client-side-validation-of-name-fields-on-contribution-checkout-page)


TomPretty:
[Client side validation of name fields on contribution checkout page](https://trello.com/c/OUpj0TPD/306-client-side-validation-of-name-fields-on-contribution-checkout-page)

tomrf1:
what does the user see if validation fails in this way?

TomPretty:
> what does the user see if validation fails in this way?

Yeah good question! So there's currently a hardcoded message for any errors in this field: https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/pages/contributions-landing/components/ContributionFormFields.tsx#L152

I guess the message (`Please provide your first name`) implicitly assumes the validation is just checking if the string is empty or not. 

It would be nice to show a different message in this case, although I'm not sure I can think of a great way of doing that without a larger refactor

TomPretty:
> Perhaps we could use a more general message, e.g. "Please enter a valid first name"?

Yeah that makes sense to me - I'll push that up now!

PR #3550, opened 2022-03-08
Fix ts errors in `formValidation.ts`

TomPretty:

Fix errors in `formValidation.ts` test file. 

The main error was coming from calling `checkStateIfApplicable` with only two arguments, instead of 3. I chose `"MONTHLY"` as the appropriate option as the code path is trivial for `"ONE_OFF"` (`"ANNUAL"` would also have worked just as well).


rupertbates (approved):
👍

PR #3558, opened 2022-03-10
Run generate-history-csv.sh with yarn ts-errors

tomrf1:
this would be useful after a rebase against main

### DIFF
--- a/support-frontend/package.json
+++ b/support-frontend/package.json
@@ -27,7 +27,7 @@
     "jest-update-snapshot": "echo 'Updating JS tests' && jest -u",
     "orphans": "echo 'Checking for orphan modules' && node scripts/orphans.js",
     "check-images": "echo 'Purging unused images' && node scripts/gridImages.js",
-    "ts-errors": "node scripts/ts-log-errors.js"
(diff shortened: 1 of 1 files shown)


support-frontend/package.json
```diff
@@ -27,7 +27,7 @@
     "jest-update-snapshot": "echo 'Updating JS tests' && jest -u",
     "orphans": "echo 'Checking for orphan modules' && node scripts/orphans.js",
     "check-images": "echo 'Purging unused images' && node scripts/gridImages.js",
-    "ts-errors": "node scripts/ts-log-errors.js"
+    "ts-errors": "node scripts/ts-log-errors.js && ./scripts/generate-history-csv.sh > ./ts-error-history.csv"
```
johnduffell, line 30: I think there can be some issue with execute permission hence we used bash directly, although I forgot what now
```suggestion
    "ts-errors": "node scripts/ts-log-errors.js && bash scripts/generate-history-csv.sh > ./ts-error-history.csv"
```

johnduffell (approved):

support-frontend/package.json
tomrf1: 👍

PR #3563, opened 2022-03-11
Only run Typescript linting on Typescript files

rupertbates:
Running TS linting on non Typescript files causes an error in Intellij (and probably elsewhere). This PR our eslint config to only run `eslint-config-typescript` against TS & TSX files


support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -204,7 +204,7 @@ function userInTest(
 	}
 
 	if (referrerControlled) {
-		return acquisitionDataTest && acquisitionDataTest.name === testId;
+		return !!acquisitionDataTest && acquisitionDataTest.name === testId;
```
marialani, line 207: Did you want this in here?

support-frontend/.eslintrc.js
```diff
@@ -1,27 +1,34 @@
 module.exports = {
-  extends: [
-    '@guardian/eslint-config-typescript',
```
marialani, line 3: Could we also remove this from our node modules

support-frontend/assets/helpers/abTests/abtest.ts
rupertbates: Oops no, I think it will disappear when you merge your PR though!

TomPretty:
This looks like it's added lots of `eslint` errors?

support-frontend/.eslintrc.js
rupertbates: As discussed it is still needed it's just moved lower in this file

marialani (approved):
Looks good!

PR #3564, opened 2022-03-11
Fix ts errors in `localCurrencyCountry`

TomPretty:

Fix ts errors in `localCurrencyCountry`. We only supported local currencies for one off contributions, but that's not reflected in the type. I updated the types with the `Omit` helper type to reflect the fact that we only support one off. This caused some knock on errors that needed cleaning up.


support-frontend/assets/helpers/internationalisation/localCurrencyCountry.ts
```diff
@@ -6,11 +6,11 @@ export type LocalCurrencyCountry = {
 	countryCode: IsoCountry;
 	countryName: string;
 	currency: IsoCurrency;
-	amounts: ContributionAmounts;
-	config: Config;
+	amounts: Omit<ContributionAmounts, 'MONTHLY' | 'ANNUAL'>;
+	config: Omit<Config, 'MONTHLY' | 'ANNUAL'>;
```
i-hardy, line 10: Really nice to have these so clear 👍

support-frontend/assets/helpers/page/commonReducer.ts
```diff
@@ -97,7 +97,10 @@ function createCommonReducer(initialState: CommonState) {
 					amounts:
 						action.useLocalAmounts &&
 						state.internationalisation.localCurrencyCountry
-							? state.internationalisation.localCurrencyCountry.amounts
+							? {
+									...state.defaultAmounts,
+									...state.internationalisation.localCurrencyCountry.amounts,
+							  }
```
i-hardy, line 103: This will need moving over into the new Redux Toolkit reducer- I think it should be pretty straightforward as this isn't a big change, but give me a shout if you need any help.

i-hardy (approved):
Looks good!

support-frontend/assets/helpers/page/commonReducer.ts
TomPretty: Yeah it was super straightforward! Gosh that new slice is good isn't it?

PR #3568, opened 2022-03-14
Get /subscribe/weekly promo copy for AU

tomrf1:
### AU with promo copy:
Screen Shot 2022-03-14 at 07 43 53

### UK with no promo copy:
Screen Shot 2022-03-14 at 07 51 07

### DIFF
--- a/support-frontend/app/controllers/WeeklySubscriptionController.scala
+++ b/support-frontend/app/controllers/WeeklySubscriptionController.scala
@@ -46,7 +46,7 @@ class WeeklySubscriptionController(
       request.queryString
         .get("promoCode")
         .map(_.toList)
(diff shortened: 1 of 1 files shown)


support-frontend/app/controllers/WeeklySubscriptionController.scala
```diff
@@ -46,7 +46,7 @@ class WeeklySubscriptionController(
       request.queryString
         .get("promoCode")
         .map(_.toList)
-        .getOrElse(List(DefaultPromotions.GuardianWeekly.NonGift.jan21Promotion))
```
tomrf1, line 49: `jan21Promotion` needs to be removed completely but I'll do that later

KaliedaRik (approved):
LGTM

PR #3569, opened 2022-03-14
Update abtests to handle array from the aquisition data

marialani:
Co-authored-by: Tom Pretty <TomPretty@users.noreply.github.com>

Update the ab test logic to handle an array ab tests in the `acquisitionData` query param. Since [this PR](https://github.com/guardian/support-dotcom-components/pull/541) in `support-dotcom-components` the `acquisitionData` object in the url generated by the modules now contains an array of `abTests` instead of a singular `abTest`.


support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -130,26 +130,25 @@ function getIsRemoteFromAcquisitionData(): boolean {
...
 
 	try {
 		const acquisitionData = JSON.parse(acquisitionDataParam) as {
+			abTests?: AcquisitionABTest[];
 			abTest?: AcquisitionABTest;
 		};
 
-		if (acquisitionData.abTest?.variant) {
-			return acquisitionData.abTest;
-		}
```
TomPretty, line 147: We were effectively doing a bit of validation here on the shape of the object coming from the URL. Do we want to have that validation on the whole array of tests? If so I think it might be worth using a proper validation library instead of rolling our own (e.g [zod](https://github.com/colinhacks/zod))

support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -320,15 +337,21 @@ function getParticipations(
...
+				testId,
+				acquisitionDataTests,
 			);
-			participations[testId] = test.variants[variantIndex].id;
+
+			if (variantIndex === -1) {
+				participations[testId] = notintest;
+			} else {
+				participations[testId] = test.variants[variantIndex].id;
+			}
```
TomPretty, line 354: Not checking if the `variantIndex` is `-1` actually breaks the page due to the `.id` call. We've now wrapped the assignment in this check

support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -261,8 +276,10 @@ function assignUserToVariant(
 		}
 
 		return index;
-	} else if (referrerControlled && acquisitionDataTest === null) {
+	} else if (referrerControlled && acquisitionDataTests === undefined) {
 		console.error('A/B test expects acquistion data but none was provided');
+
+		return -1;
```
TomPretty, line 282: Before, we weren't returning from this branch. This means we would fall through to line 285 and return a random variant. I think this probably isn't the best behaviour. The test has been set up as `referrerControlled` but we're assigning them randomly. I think if it's `referrerControlled` but we can't find the assignment in `acquisitionData` we shouldn't assign them to that test

tomrf1: This makes sense.
Side note - your change didn't introduce this, but giving `-1` a special meaning always feels a little unnatural/brittle. I wonder if it could return `number | notintest` instead? Not something that needs addressing now!

tomrf1 (approved):
looks good!

support-frontend/assets/helpers/abTests/abtest.ts
TomPretty: Yeah I'm having a go at some more refactors now to clean this file up a bit, but will do that as a follow up PR

TomPretty: Remvoing the `-1` is definitely on the todo list!

PR #3570, opened 2022-03-14
Convert common reducer to use Redux Toolkit

i-hardy:

This converts the common reducer, used across all checkouts, to a [Redux Toolkit slice](https://redux.js.org/usage/usage-with-typescript#application-usage). It also updates all pages that rely on types and actions relating to common state, in particular various checkout tests.

There's quite a lot of noise in the diff due to fixing the test files; I've commented on the files that have the most significant changes.

[**Trello Card**](https://trello.com/c/8eMQl3pd)

## Why are you doing this?

This is part of work to overhaul our usage of Redux and migrate to using Toolkit to simplify and streamline working on Redux-related code. The common reducer is used on every page and this change is part of refactoring our store setup process to enable us to use some of the utility types that Toolkit provides.

## Is this an AB test?
- [ ] Yes
- [x] No


support-frontend/assets/components/subscriptionCheckouts/address/addressFieldsStore.ts
```diff
@@ -236,11 +236,13 @@ const setFormErrorsFor =
 const addressActionCreatorsFor = (scope: AddressType) => ({
 	setCountry:
 		(countryRaw: string) =>
-		(dispatch: Dispatch<SetCountryChangedAction | SetCountryAction>) => {
+		(
+			dispatch: Dispatch<SetCountryChangedAction | PayloadAction<IsoCountry>>,
```
i-hardy, line 240: When we're further along with the migration we'll be able to get a `Dispatch` type inferred out of the store type, but for the time being in order to harmonise with our existing action types I've had to use the Redux Toolkit `PayloadAction` type explicitly in a few places.

support-frontend/assets/helpers/redux/commonState/reducer.ts
```diff
@@ -0,0 +1,73 @@
...
+	const currencyId =
+		fromCountryGroupId(countryGroupId) ?? internationalisation.currencyId;
+	return {
+		countryGroupId,
+		currencyId,
+		countryId,
+	};
+}
+
+export const commonSlice = createSlice({
```
i-hardy, line 26: This is the replacement for both the old switch statement reducer and action creator functions. The `state` argument passed to each reducer function is a mutable copy of the previous state object which is then replaced under the hood, enabling us to write simpler code without losing the benefits of immutable state.

support-frontend/assets/helpers/redux/commonState/state.ts
```diff
@@ -0,0 +1,64 @@
...
+	otherQueryParams: Array<[string, string]>;
+	abParticipations: Participations;
+	settings: Settings;
+	amounts: ContributionAmounts;
+	defaultAmounts: ContributionAmounts;
+	internationalisation: Internationalisation;
+	existingPaymentMethods?: ExistingPaymentMethod[];
+};
+
+export type CommonStateSetupData = {
```
i-hardy, line 37: I added this, along with the matching `setInitialState` reducer, to square the circle of us previously dynamically generating the reducer at runtime with variable initial state values. Having the initial state and reducer be static exports, and then updating them with the page-specific data after the store is created, will enable us to get a lot more benefits out of Toolkit's utility types in the future.

i-hardy:
[Redux Toolkit migration - refactor Redux setup](https://trello.com/c/8eMQl3pd/320-redux-toolkit-migration-refactor-redux-setup)

support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -379,7 +379,7 @@ function getAmountsTestParticipations(
 		return null;
 	}
 
-	const { test } = settings.amounts[countryGroupId];
+	const { test } = settings.amounts?.[countryGroupId] ?? {};
 
 	if (!test || !test.isLive) {
```
rupertbates, line 384:
```suggestion
	if (!test.isLive) {
```

support-frontend/assets/helpers/redux/commonState/reducer.ts
TomPretty: I was slightly dubious about making code that looks like it's doing mutations actually still give us an immutable state. But, it's hard to argue with how much simpler the code is.

Also removing the massive switch statement and all the manually defined action creators is amazing!

i-hardy: It's a bit weird to write, but the types help make it clear what `state` here actually is. It's all based on Immer under the hood- pretty cool stuff, I'm thinking I should probably write some additions to our wiki about this https://redux-toolkit.js.org/usage/immer-reducers

support-frontend/assets/helpers/redux/commonState/state.ts
```diff
@@ -0,0 +1,64 @@
...
+	campaign: Campaign | null | undefined;
```
rupertbates, line 38: This is more a general point and not for this PR probably but I really feel we should be getting rid of these `| null | undefined` types from the codebase, and standardise on one (undefined probably) there shouldn't be any need to allow both.

PR #3573, opened 2022-03-15
Fix SFD_V2 assignment bug

TomPretty:

Fix a bug in the `SFD_V2` assignment logic. We're currently not checking if the current `testId` is `SFD_V2`. This means any test with `referrerControlled` set to `true` would be assigned if a single front door test was in the `acquisitionData`. Luckily the `SFD_V2` test is the only active `referrerControlled` test atm!

I also added an explicit return type of `boolean` and changed the `.find` calls to `.some` so that we only return a boolean instead of `boolean | AcquistionAbTest | undefined`.


marialani (approved):
👍

PR #3575, opened 2022-03-15
Refactor `abTest` and corresponding `abTestTest` test file

TomPretty:

Refactor `abTest.ts` and the corresponding test file `abTestTest.ts`. I've made quite a few changes so I've left comments on the code to try and explain rather than summarising here.


support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -72,11 +71,44 @@ export type Test = {
...
+
+export function init(
+	country: IsoCountry,
+	countryGroupId: CountryGroupId,
+	settings: Settings,
+	abTests: Tests = tests,
+	mvt: number = getMvtId(),
+	acquisitionDataTests: AcquisitionABTest[] = getTestFromAcquisitionData() ??
+		[],
+): Participations {
```
TomPretty, line 84: I've moved this function to the top of the file as it's the most important one. I've also added a couple extra optional arguments (`mvt` and `acquisitionDataTests`) to make it easier to test

support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -114,20 +200,30 @@ function getServerSideParticipations(): Participations | null | undefined {
...
 }
 
-function getIsRemoteFromAcquisitionData(): boolean {
-	const queryString = getQueryParameter('acquisitionData');
-
-	if (!queryString) {
-		return false;
+function getAmountsTestParticipations(
+	countryGroupId: CountryGroupId,
+	settings: Settings,
```
TomPretty, line 205: No change here, just moved higher up in the file to reflect importance

support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -181,52 +277,71 @@ function userInBreakpoint(audience: Audience): boolean {
...
+type NoParticipation = { type: 'NO_PARTICIPATION' };
+
+type ActiveParticipation =
+	| { type: 'ACTIVE_PARTICIPATION' }
+	| {
+			type: 'REFERRER_CONTROLLED_ACTIVE_PARTICIPATION';
+			acquisitionDataTest: AcquisitionABTest;
+	  };
+
+type UserParticipation = NoParticipation | ActiveParticipation;
```
TomPretty, line 289: I've split out the `NoParticipation` and `ActiveParticipation` so that `assignUserToVariant` can explicitly accept an `ActiveParticipation`. The main benefit though is that for a `referrerControlled` test we store the matching `acquisitionDataTest`. This means we don't have to search for it again in `assignUserToVariant` and deal with the fact that the types signature of `.find` returns something that might be `undefined`

support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -181,52 +277,71 @@ function userInBreakpoint(audience: Audience): boolean {
...
+			return NO_PARTICIPATION;
 		}
 
-		return (
-			!!acquisitionDataTests &&
-			acquisitionDataTests.some(
-				(acquisitionDataTest) => acquisitionDataTest.name === testId,
-			)
-		);
+		return referrerControlledActiveParticipation(acquisitionDataTest);
```
TomPretty, line 324: I thought we could remove the special case logic for the single front door test and just make it a feature of `referrerControlled` tests to support test campaigns

support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -181,52 +277,71 @@ function userInBreakpoint(audience: Audience): boolean {
...
+
+const PARTICIPATING: UserParticipation = { type: 'ACTIVE_PARTICIPATION' };
+
+const referrerControlledActiveParticipation = (
+	acquisitionDataTest: AcquisitionABTest,
+): UserParticipation => ({
+	type: 'REFERRER_CONTROLLED_ACTIVE_PARTICIPATION',
+	acquisitionDataTest,
+});
+
```
TomPretty, line 344: These are just some helpers for returning `UserParticipation`s

support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -251,44 +366,45 @@ const trackOptimizeExperiment = (
 	);
 };
 
+type VariantAssignment =
+	| { type: 'NOT_ASSIGNED' }
+	| { type: 'ASSIGNED'; variantIndex: number };
+
```
TomPretty, line 372: Again, I've updated this function to return a tagged union. It's mainly to remove the magic `-1` we had before when it just returned a `number`. It's maybe a little overkill but I like the uniformity with `getUserParticipation`

support-frontend/assets/helpers/tracking/ophan.ts
```diff
@@ -74,7 +74,7 @@ type OphanABEvent = {
 	campaignCodes?: string[];
 };
 
-type OphanABPayload = Record<TestId, OphanABEvent>;
+type OphanABPayload = Record<string, OphanABEvent>;
```
TomPretty, line 77: `TestId` was just an alias for `string` and I don't think it was adding anything

support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -251,44 +366,45 @@ const trackOptimizeExperiment = (
...
-			t.name.startsWith(testId),
-		);
+	// For referrrer controlled tests the assignment comes from the acquisition data.
+	// If the variant in the acquisition data doesn't match up with any defined in
+	// the test, we log an error and don't assign the user to the test.
+	const acquisitionVariant = participation.acquisitionDataTest.variant;
 
-		if (!acquisitionDataTest) {
-			return -1;
-		}
```
TomPretty, line 269: This is the check we had to do before, but is actually redundant as we know if the user is participating in this test that it must be in the array of tests. Using this new type we no longer need to do this.

support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -25,22 +22,24 @@ const breakpoints = {
...
-type Breakpoint = $Keys<typeof breakpoints>;
+type Breakpoint = keyof typeof breakpoints;
 
 type BreakpointRange = {
 	minWidth?: Breakpoint;
 	maxWidth?: Breakpoint;
 };
 
-export type Participations = Record<TestId, string>;
+export type Participations = Record<string, string>;
```
TomPretty, line 32: `TestId` was just an alias for `string` which I don't think was adding value

support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -95,6 +127,60 @@ function getMvtId(): number {
...
 }
 
+function getParticipations(
+	abTests: Tests,
+	mvtId: number,
+	country: IsoCountry,
+	countryGroupId: CountryGroupId,
+	acquisitionDataTests?: AcquisitionABTest[],
+): Participations {
+	const participations: Participations = {};
```
TomPretty, line 137: This function has been moved up in the file to reflect its importance. It's also changed slighty due to renaming `userInTest` to `getUserParticipation` and the new return types of `getUserParticipation` and `assignUserToVariant`. I've also fully commited to `if` guards e.g

```ts
if (condition) {
  return;
}
```

Which has removed all nesting. I think this makes it easier to read

support-frontend/assets/helpers/abTests/abtest.ts
TomPretty: I've added a new test case for this behaviour

support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -251,44 +366,45 @@ const trackOptimizeExperiment = (
...
+	return assigned(index);
+}
 
-		return -1;
-	}
+const NOT_ASSIGNED: VariantAssignment = { type: 'NOT_ASSIGNED' };
 
-	return randomNumber(mvtId, seed) % test.variants.length;
+function assigned(variantIndex: number): VariantAssignment {
+	return { type: 'ASSIGNED', variantIndex };
```
TomPretty, line 404: Just some helpers for returning `VariantAssignment`

support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -299,133 +415,10 @@ function targetPageMatches(
...
-	const acquisitionDataTests: AcquisitionABTest[] | undefined =
-		getTestFromAcquisitionData();
-	const isRemote = getIsRemoteFromAcquisitionData();
-
-	// This is a temporary addition to help us compare remote and locally rendered
-	// epics on Frontend (as part of testing out the new Contributions Service).
-	// It will be removed once the new service is shown to be working correctly.
-	if (isRemote) {
-		participations.RemoteEpicVariants = 'remote';
-	}
```
TomPretty, line 318: I deleted this!

support-frontend/assets/helpers/abTests/__tests__/abtestTest.ts
```diff
@@ -21,840 +18,303 @@ jest.mock('ophan', () => ({
...
-	JSON.stringify(acquisitionDataMockTest2Control),
-)}`;
-
-const mockTest2Variant = `/test.html?acquisitionData=${encodeURI(
-	JSON.stringify(acquisitionDataMockTest2Variant),
-)}`;
-
-describe('basic behaviour of init', () => {
+
+describe('init', () => {
```
TomPretty, line 22: I made quite a lot of changes here! The main philosophy was simplifying the test code and removing redundant tests. I've tried to remove as much noise as possible from the test cases so that it's easier to see at a glance the important parts that we're actually testing

support-frontend/assets/helpers/abTests/__tests__/abtestTest.ts
```diff
@@ -21,840 +18,303 @@ jest.mock('ophan', () => ({
...
-	JSON.stringify(acquisitionDataMockTestVariant),
-)}`;
-
-const mockTest2Control = `/test.html?acquisitionData=${encodeURI(
-	JSON.stringify(acquisitionDataMockTest2Control),
-)}`;
-
-const mockTest2Variant = `/test.html?acquisitionData=${encodeURI(
-	JSON.stringify(acquisitionDataMockTest2Variant),
-)}`;
```
TomPretty, line 84: I've removed all of these by making `acquisitionAbTests` an optional argument to `init`. This means we don't need to create these objects and manually call `window.history.push`

support-frontend/assets/helpers/abTests/__tests__/abtestTest.ts
```diff
@@ -21,840 +18,303 @@ jest.mock('ophan', () => ({
...
+	const emptySettings = {
+		switches: {
+			experiments: {},
+		},
+		amounts: {
+			GBPCountries: {},
+			UnitedStates: {},
+		},
+	} as Settings;
+
```
TomPretty, line 42: I've pulled out all of these common arguments so that the only arguments that are defined in the test cases are actually important to what we're testing

support-frontend/assets/helpers/abTests/__tests__/abtestTest.ts
```diff
@@ -21,840 +18,303 @@ jest.mock('ophan', () => ({
...
 	afterEach(() => {
 		window.localStorage.clear();
 	});
-	it('The referrerControlled user should be allocated to the control bucket', () => {
-		document.cookie = 'GU_mvt_id=12346';
-		window.history.pushState({}, 'Test Title', mockTestControl);
+
+	it('assigns a user to a variant', () => {
```
TomPretty, line 47: I changed the names of all of the tests so that they read like prose with the `it` function name.

I've also removed test cases that just changed the `seed` to test assignment to `control`/`variant`. I think they're too subtle and what they're actually testing is that for a given combination of test name, `mvt`, and `seed` you're assigned to a specific variant. I think we get more value out of tests  by testing deterministic behaviour. This means most of the remaining tests are testing the targeting is working correctly.

support-frontend/assets/helpers/abTests/__tests__/abtestTest.ts
```diff
@@ -21,840 +18,303 @@ jest.mock('ophan', () => ({
...
-					GB: {
-						offset: 0,
-						size: 1,
-					},
-				},
-				isActive: true,
-				referrerControlled: true,
-				seed: 0,
-			},
+			t: buildTest({ variants: [buildVariant({ id: 'control' })] }),
```
TomPretty, line 49: I've factored out a number of `build` functions. This has two benefits. First, it cuts down on the number of lines of code. Secondly (and more importantly), it allows us to just specify the salient parts of the test that we want. Here, the only important part is that we have a test called `t` with a single variant `control`. This is important because we're going to explicitly assert that we get assigned to this test and variant.

support-frontend/assets/helpers/abTests/__tests__/abtestTest.ts
```diff
@@ -21,840 +18,303 @@ jest.mock('ophan', () => ({
...
+			t1: 'control',
+			t2: 'variant',
 		};
+
 		expect(participations).toEqual(expectedParticipations);
 	});
 
-	it('The user should be allocated to the control bucket', () => {
-		document.cookie = 'GU_mvt_id=12346';
+	it('uses the variant assignment in the acquisitionData for referrerControlled tests belonging to a campaign', () => {
```
TomPretty, line 107: This is a new test case I added to test the `campaign` assignment behaviour. Campaigns are configured in the RRCP with a common prefix. We can then set up a corresponding test on the support site by creating a test named with that prefix.

support-frontend/assets/helpers/abTests/__tests__/abtestTest.ts
```diff
@@ -21,840 +18,303 @@ jest.mock('ophan', () => ({
...
-		);
-		expect(participations.RemoteEpicVariants).toBe('remote');
+		deleteCookie();
 	});
-});
 
-describe('Correct allocation in a multi test environment', () => {
-	afterEach(() => {
-		window.localStorage.clear();
-	});
```
TomPretty, line 622: I deleted all of these tests. They're quite complex and hard to figure out what they're actually testing. I belive, when you drill down what they're actually doing is testing that the audience targeting works with the `offset` and `size`. I've replaced them all with two simpler tests that explicitly test that behaviour

support-frontend/assets/helpers/abTests/__tests__/abtestTest.ts
```diff
@@ -927,3 +387,43 @@ it('targetPage matching', () => {
...
+
+function buildAcquisitionAbTest({
+	name = 't',
+	variant = 'control',
+}: Partial<AcquisitionABTest>): AcquisitionABTest {
+	return {
+		name,
+		variant,
+	};
+}
```
TomPretty, line 429: These are the new `build` functions I added. The `Partial` utility type from typescript is really helpful here. It allows us to easily define functions that you can just provide with the arguments you care about for the test and have all the others fallback to sensible defaults.

TomPretty:
[Refactor AB test logic in support-frontend](https://trello.com/c/LHo1XYdb/349-refactor-ab-test-logic-in-support-frontend)

michaelbjacobson (approved):
Looks great! 😄

PR #3576, opened 2022-03-15
Fix TS errors in country switcher components

i-hardy:

This fixes the Typescript errors in the `CountryGroupSwitcher` component and associated components. These components use CSS modules (via SCSS) that I've slightly fudged the types for; these ultimately need to be migrated to Emotion but doing so right now is a bit too far out of scope.


i-hardy:
[Fix TS errors in countryGroupSwitcher](https://trello.com/c/k58znSIc/340-fix-ts-errors-in-countrygroupswitcher)

PR #3585, opened 2022-03-17
remove sepa query param

michaelbjacobson:
This PR removes the need for a query string to allow SEPA transactions.

## Is this an AB test?
- [ ] Yes
- [x] No


i-hardy (approved):
👍

PR #3595, opened 2022-03-22
Refactor Redux initialisation to enable use of utility types

i-hardy:
This overhauls the way we set up Redux for the various checkouts, in order to enable us to start to make use of Redux Toolkit's utility types.

I've created specific Redux structures for the three different kinds of page state that we actually use- for subscriptions checkouts, contributions, and digi sub gift redemption, with 'init' functions that set up common state properties and allow the recreation of the store if necessary to smooth over the differences between subs checkouts. I've replaced all uses of the old generics-based `initRedux` function with these.

We now have access to types for `State` and `Dispatch` that are inferred from the structure of the store; I've started to incorporate these in a few instances as a demonstration of how they work and to clean up some of the messier typings created by #3570.

I've also refactored the promo code terms & conditions page and the 'Thank you' page for existing digital subscribers to not rely on Redux; neither page is dynamic and we weren't even using Redux for props injection so it was just a lot simpler to get rid.

[**Trello Card**](https://trello.com/c/8eMQl3pd)

## Why are you doing this?
This is the groundwork for migrating fully to Redux Toolkit and simplifying the way we handle Redux in general.

By no means is this a finished position! Ideally we will end up with a single state structure that can handle all products; however in my opinion the benefit of separating store setup for now is that: 
- Having the utility types in place will make refactoring the rest of the Redux state much easier- anywhere that uses the `SubscriptionsDispatch` type will automatically be updated when we migrate parts of that state to Toolkit, rather than having to manually fix `Dispatch` and `Action` types everywhere.
- It becomes much, much clearer what the existing differences are between the state structures, and we'll be able to tell much more easily when we've reached parity as the types will match!

## Is this an AB test?
- [ ] Yes
- [x] No


i-hardy:
[Redux Toolkit migration - refactor Redux setup](https://trello.com/c/8eMQl3pd/320-redux-toolkit-migration-refactor-redux-setup)

support-frontend/assets/helpers/subscriptionsForms/subscriptionCheckoutReducer.ts
```diff
@@ -12,33 +12,38 @@ import { addressReducerFor } from 'components/subscriptionCheckouts/address/addr
...
 	directDebit: DirectDebitState;
 }>;
+
 export type WithDeliveryCheckoutState = ReduxState<{
 	checkout: FormState;
 	csrf: CsrfState;
 	marketingConsent: MarketingConsentState;
 	billingAddress: AddressState;
 	deliveryAddress: AddressState;
-	hasDeliveryAddress: true;
```
i-hardy, line 41: `hasDeliveryAddress` was not being read anywhere in the code, and created an unnecessary type conflict between these two state variations.

support-frontend/assets/helpers/redux/subscriptionsStore.ts
```diff
@@ -0,0 +1,60 @@
...
+const subscriptionsStore = configureStore({
+	reducer: baseReducer,
+});
+
+export type SubscriptionsStore = typeof subscriptionsStore;
+
+export function addPageReducer(
+	newReducer?: SubscriptionsReducer,
+): SubscriptionsStore {
+	return configureStore({
```
i-hardy, line 35: Creating the store twice is obviously not ideal, but my initial approach using `store.replaceReducer` caused some odd problems when not in dev mode.

TomPretty: Why is it that we need to do this?

support-frontend/assets/pages/contributions-landing/contributionsLandingReducer.ts
```diff
@@ -63,6 +63,7 @@ export interface StripeCardFormData {
 	formComplete: boolean;
 	setupIntentClientSecret: string | null;
 	recurringRecaptchaVerified: boolean;
+	// TODO: No non-serialisable values should be in Redux state!! This needs to be refactored
```
TomPretty, line 66: Oh it's a bit of a disaster how much non-serialisable state is in this reducer

support-frontend/assets/pages/digital-subscription-checkout/thankYouExisting.tsx
```diff
@@ -1,42 +1,42 @@
...
 import ThankYouExistingContent from './thankYouExistingContent';
+import 'stylesheets/skeleton/skeleton.scss';
 import './digitalSubscriptionCheckout.scss';
 
 setUpTrackingAndConsents();
+
 // ----- Redux Store ----- //
-const store = initRedux();
-const { countryGroupId } = store.getState().common.internationalisation;
+const countryGroupId = detect();
```
TomPretty, line 19: Just a small one - I guess this header comment is a little out of data now?

support-frontend/typescript-errors.json
```diff
@@ -651,6 +630,10 @@
             "path": "assets/components/directDebit/helpers/ajax.ts",
             "message": "'isTestUser' is declared but its value is never read."
         },
+        {
+            "path": "assets/helpers/redux/subscriptionsStore.ts",
+            "message": "'combineReducers' is declared but its value is never read."
+        },
```
TomPretty, line 636: This looks like a simple one to fix!

support-frontend/assets/helpers/redux/subscriptionsStore.ts
i-hardy: The subscriptions stores currently rely on creating a brand new page-specific reducer from scratch for each checkout, with specific initial state indicating the product, fulfilment option, etc etc, and we can't get around this without completely rewriting how all the checkout reducer code works. This means having no access to utility types because you don't know the structure of the store until runtime.

I initially took the approach of creating the static store instance and then using the [replaceReducer](https://redux.js.org/usage/code-splitting#using-replacereducer) utility to just update the store with the checkout-specific page reducer at runtime, but for god only knows what reason this did not work in browsers that do not have Redux devtools installed, hence tearing my hair out for half the day yesterday because the Selenium tests were failing with pages not even loading properly but I couldn't reproduce the failures! Creating a brand new store instance from scratch solves the problem.

In the future we can overhaul the store setup to not rely on creating reducers at runtime, but for the time being this means having access to the utility types without the checkouts breaking.

I should probably add or link to this explanation in a code comment actually, this is one of those 'I know this is a stupid decision but we have to' things that requires a bit of extra context.

support-frontend/assets/pages/contributions-landing/contributionsLandingReducer.ts
i-hardy: Well at least now we're going to have a lot of angry red reminders in the console during development about how bad it is 😂

support-frontend/assets/helpers/redux/subscriptionsStore.ts
TomPretty: Ah I see - so creating the `subscriptionsPageReducer` on line 10 is just so that we can extract the utility types which we can then share across the checkout pages? And the utility types work across all those pages despite them having slightly different reducers?

Yeah I'd say it's worth a comment, as someone who knows much less about the checkout pages it was a little tricky to understand what was going on!

support-frontend/conf/DEV.public.conf
```diff
@@ -13,8 +13,8 @@ identity.useStub=false
 guardianDomain=".thegulocal.com"
 
 # To run selenium post-deployment tests against code use the config below
-# support.url="https://support.code.dev-theguardian.com"
-support.url="https://support.thegulocal.com"
+support.url="https://support.code.dev-theguardian.com"
+# support.url="https://support.thegulocal.com"
```
TomPretty, line 17: I guess we don't want to commit this change?

TomPretty (approved):
Really great work here! This looks like a huge improvement and I'm very excited to get using these utility types. 

I can only imagine how much of a nightmare some of the debugging must have been here!

support-frontend/assets/helpers/redux/subscriptionsStore.ts
i-hardy: Yes, exactly- the overall type structure of the checkout state is similar enough that the utility types should work in most/all cases, but it's the provision of the initial state for a given checkout with the relevant product details that's currently impossible to do without recreating the reducer from scratch.

I've added a link to this thread as a comment- should hopefully be helpful in future!

support-frontend/conf/DEV.public.conf
i-hardy: Whoops! Fixed.

PR #3597, opened 2022-03-23
CDK migration phase 1

TomPretty:

This PR does the minimal amount of work to add `cdk` to our project. We're now using `cdk` to generate the cloudformation for our project, but just via the `CfnInclude` construct which loads our existing cloudformation file. In [phase two](https://github.com/guardian/support-frontend/pull/3598) of the migration we will build up the cloudformation directly with `cdk` constructs.

This is the same process we did for [support-dotcom-components](https://github.com/guardian/support-dotcom-components/pull/510).

There a quite a few files changed, but almost all of those changes have been auto generated by the project generator. The only files we've touched in the `cdk` directory are: `.nvmrc` (added because the node version used by `support-frontend` is too old) and `cdk/lib/frontend.ts` which is the file where we actually build our `cdk` constructs. The big lines diff is due to the snapshot test and a new `yarn.lock`.


cdk/lib/frontend.ts
```diff
@@ -0,0 +1,22 @@
+import { join } from "path";
+import { CfnInclude } from "@aws-cdk/cloudformation-include";
```
TomPretty, line 2: This is the file that we'll actually be editing to replace our old cloudformation with `cdk` constructs. For now, we're just using `CfnInclude` to directly load our existing cloudformation template

support-frontend/build-tc
```diff
@@ -20,13 +20,23 @@ if [ ! -d "$YARN_LOCATION" ]; then
...
+yarn run build-prod
+
+(
+  cd ../cdk
+  nvm install
+  nvm use
+  # nvm alters PATH so we need to ensure the ci script uses the correct version of node
+  export PATH
+  ./script/ci
+)
```
TomPretty, line 42: This was a little fiddly. If we can update the node version used by support-frontend we should be able to just do

```bash
(
  cd ../cdk
  ./script/ci
)
```

TomPretty:
[support-frontend cdk migration - phase 1](https://trello.com/c/OEX1nAen/356-support-frontend-cdk-migration-phase-1)

support-frontend/build-tc
rupertbates: Updated it beyond the update in this PR?

rupertbates (approved):
Looks good!

PR #3598, opened 2022-03-23
CDK migration phase 2

TomPretty:

In this phase we use cdk to create a 'dual stack' of all of our infrastructure. We retain the `CfnInclude` construct to directly import our original cloudformation template but also recreate all of our resources directly with cdk (mainly with the GuEc2App construct). This means we will have a second load balancer and autoscaling group etc. We can deploy this to production and then, when we're happy, we can point fastly at the new load balancer to send traffic to the new stack. In a follow up PR, we can drop `CfnInclude` and delete our old cloudformation template.

To deploy this we need to:

1. Manually deploy the cloudformation change (it adds new parameters so riffraff will fail)
2. Merge PR and wait for build and deploy
3. Point fastly at the new load balancer

There are a number of `TODO` comments in the PR that are a result of DevEx highlighting certain areas of our infrastructure. These can be addressed as follow ups. It's safest to recreate our existing infrastructure exactly with cdk rather than trying to fix up any issues at the same time.


cdk/lib/frontend.ts
```diff
@@ -18,5 +34,384 @@ export class Frontend extends GuStack {
...
+    mkdir /etc/gu
+    aws --region ${this.region} s3 cp s3://membership-dist/${this.stack}/${this.stage}/${app}/support-frontend_1.0-SNAPSHOT_all.deb /tmp
+    dpkg -i /tmp/support-frontend_1.0-SNAPSHOT_all.deb
+    /opt/cloudwatch-logs/configure-logs application ${this.stack} ${this.stage} ${app} /var/log/support-frontend/application.log '%Y-%m-%dT%H:%M:%S,%f%z'`;
+
+    const membershipSubPromotionsTable = this.withStageDependentValue({
+      app,
+      variableName: "MembershipSubPromotions",
+      stageValues: {
+        // TODO: Should these dynamo tables be region specific?
```
TomPretty, line 50: DevEx pointed out that these tables aren't region specific and that it might be preferable if they were

cdk/lib/frontend.ts
```diff
@@ -18,5 +34,384 @@ export class Frontend extends GuStack {
...
+    });
+
+    const policies = [
+      // TODO: can we 'standardise' the way we load config to use the default permissons from GuEc2App?
+      new GuAllowPolicy(this, "SSMGet", {
+        actions: ["ssm:GetParametersByPath"],
+        resources: [
+          `arn:aws:ssm:${this.region}:${this.account}:parameter/${this.stack}/${app}/${this.stage}`,
+        ],
+      }),
```
TomPretty, line 73: The `GuEc2App` construct will automatically give us permissions to `CODE/support/frontend` but the library that we use for config stores config in `support/frontend/CODE`

cdk/lib/frontend.ts
```diff
@@ -18,5 +34,384 @@ export class Frontend extends GuStack {
...
+          minimumInstances: 1,
+          maximumInstances: 2,
+        },
+      },
+      instanceType: InstanceType.of(InstanceClass.T4G, InstanceSize.SMALL),
+    });
+
+    // TODO: remove this tag after the migration
+    const ec2AppAsg = ec2App.autoScalingGroup;
+    Tags.of(ec2AppAsg).add("gu:riffraff:new-asg", "true");
```
TomPretty, line 171: We only need this tag whilst we have a dual stack

cdk/lib/frontend.ts
```diff
@@ -18,5 +34,384 @@ export class Frontend extends GuStack {
...
+      // TODO: should we move logs to kinesis?
+      new GuAllowPolicy(this, "CloudwatchMetrics", {
+        actions: [
+          "logs:CreateLogGroup",
+          "logs:CreateLogStream",
+          "logs:PutLogEvents",
+          "logs:DescribeLogStreams",
+        ],
+        resources: ["arn:aws:logs:*:*:*"],
+      }),
```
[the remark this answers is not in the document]
TomPretty: (This would save money)

support-frontend/conf/riff-raff.yaml
```diff
@@ -23,6 +23,7 @@ deployments:
     dependencies: [cfn, assets-static]
     parameters:
       bucket: membership-dist
+      asgMigrationInProgress: true
```
TomPretty, line 26: We only need this whilst we're migrating

TomPretty:
[support-frontend cdk migration - phase 2](https://trello.com/c/NnQsFrKC/358-support-frontend-cdk-migration-phase-2)

cdk/lib/frontend.ts
akash1810: Out of interest, what library are you using?

cdk/lib/frontend.ts
```diff
@@ -18,5 +34,384 @@ export class Frontend extends GuStack {
...
+        metricName: "HealthyHostCount",
+        namespace: "AWS/ApplicationELB",
+        dimensionsMap: {
+          LoadBalancer: ec2App.loadBalancer.loadBalancerFullName,
+          TargetGroup: ec2App.targetGroup.targetGroupFullName,
+        },
+        statistic: "Average",
+        period: Duration.seconds(60),
+      }),
+      snsTopicName: "reader-revenue-dev",
```
akash1810, line 199: Would be good to get this from parameter store (see also https://github.com/guardian/cdk/pull/884).

akash1810 (commented):
The `GuAlarm`s being created are fairly similar - could we make their creation more terse by use of a for loop?

cdk/lib/frontend.ts
TomPretty: Yeah makes sense - would this be a follow up to do after that cdk PR has been merged?

TomPretty:
> The `GuAlarm`s being created are fairly similar - could we make their creation more terse by use of a for loop?

I had a go at removing some duplication. Mainly that was factoring out the repeated

```ts
this.withStageDependentValue({
      app,
      variableName: "AlarmEnabledInProd",
      stageValues: {
        [Stage.CODE]: false,
        [Stage.PROD]: true,
      },
    });
```

into a single `const` that can be reused. 

I'm not sure if we can squeeze it down much more given the other fields vary between alarms - unless you had something specific in mind?

cdk/lib/frontend.ts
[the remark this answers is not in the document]
i-hardy: I would be strongly in favour of this; I find it really hard to pick out details with Cloudwatch logs whereas Kibana has a much nicer UI.

i-hardy:
I feel like I don't have enough familiarity with the existing setup to be the one to sign off on this, but I really, really love how much simpler and more understandable this is 🙏

cdk/lib/frontend.ts
```diff
@@ -18,5 +34,360 @@ export class Frontend extends GuStack {
...
+          hostedZoneId: "Z1E4V12LQGXFEC",
+        },
+        [Stage.PROD]: {
+          domainName:
+            "support.theguardian.com.origin.membership.guardianapis.com",
+          hostedZoneId: "Z1E4V12LQGXFEC",
+        },
+      },
+      monitoringConfiguration: {
+        noMonitoring: true,
```
jacobwinch, line 150: At the moment we are opting out of the 'standard' alarms that the pattern wants to create; we might be able to use some alarms provided via the pattern to reduce the number of custom alarms defined here.

cdk/lib/frontend.ts
```diff
@@ -18,5 +34,360 @@ export class Frontend extends GuStack {
...
+          LoadBalancer: ec2App.loadBalancer.loadBalancerFullName,
+          TargetGroup: ec2App.targetGroup.targetGroupFullName,
+        },
+        statistic: "Average",
+        period: Duration.seconds(300),
+      }),
+      snsTopicName: "reader-revenue-dev",
+    });
+
+    new GuAlarm(this, "High5XXRateAlarm", {
```
jacobwinch, line 238: FYI - the pattern provides a helper for creating a high 5XX alarm. There is an example of minimal configuration for that [here](https://github.com/guardian/cdk/blob/d1de578b5691d83afa591a5b6c47cb91349ea70d/src/patterns/ec2-app/base.ts#L65-L75). The custom descriptions look important here, but I think you should be able to pass that in via the relevant alarm's [props](https://guardian.github.io/cdk/interfaces/constructs_cloudwatch.Http5xxAlarmProps.html).

At the moment, the pattern uses a % based alarm and combines two of your alarms (`High5XXRateAlarm` and `HighELB5XXRateAlarm`) into a single alarm. If using absolute thresholds and retaining the granularity of two alarms is important for your team, this might not be helpful for you.

The advantages of using the alarm provided via the pattern would be:

* reduction of boilerplate in this stack definition
* you will pick up improvements 'for free' if/when we establish best practices for monitoring/alerting in this area
* you can standardise this alarm across all of your apps more easily

However, as this is not a like for like replacement for you, it might be something to consider in a future PR.

cdk/lib/frontend.ts
```diff
@@ -18,5 +34,360 @@ export class Frontend extends GuStack {
...
+        dimensionsMap: {
+          LoadBalancer: ec2App.loadBalancer.loadBalancerFullName,
+        },
+        statistic: "Sum",
+        period: Duration.seconds(60),
+      }),
+      snsTopicName: "reader-revenue-dev",
+    });
+
+    new GuAlarm(this, "LatencyNotificationAlarm", {
```
jacobwinch, line 285: Interesting to see that you are alerting on latency; I wonder if we should provide this as standard @nicl @akash1810 @michaelwmcnamara?

cdk/lib/frontend.ts
```diff
@@ -18,5 +34,360 @@ export class Frontend extends GuStack {
...
+          LoadBalancer: ec2App.loadBalancer.loadBalancerFullName,
+          TargetGroup: ec2App.targetGroup.targetGroupFullName,
+        },
+        statistic: "Average",
+        period: Duration.seconds(60),
+      }),
+      snsTopicName: "reader-revenue-dev",
+    });
+
+    new GuAlarm(this, "ReducedHealthyInstancesAlarm", {
```
jacobwinch, line 213: IIUC there is some overlap between the goal of this alarm and the goal of the [`GuUnhealthyInstances`](https://guardian.github.io/cdk/classes/constructs_cloudwatch.GuUnhealthyInstancesAlarm.html) alarm that the pattern is [able to create for you](https://github.com/guardian/cdk/blob/d1de578b5691d83afa591a5b6c47cb91349ea70d/src/patterns/ec2-app/base.ts#L73). It'd be interesting to know if the latter could work for you, but I think it makes sense to retain your existing alarm for now.

cdk/lib/frontend.ts
```diff
@@ -18,5 +34,360 @@ export class Frontend extends GuStack {
...
+      app,
+      alarmName: alarmName(
+        "reduced number healthy instances for support-frontend"
+      ),
+      alarmDescription: alarmDescription(
+        "Imminent issue cannot sell any subscriptions or contributions products"
+      ),
+      actionsEnabled: alarmEnabledInProd,
+      threshold: 2,
```
jacobwinch, line 222: I presume that this threshold is set to 2 because you want to run [at least 3 instances at all times in `PROD`](https://github.com/guardian/support-frontend/blob/c3c8289e84a301a72864ddf55cbddaa765594b48/cdk/lib/frontend.ts#L158). Now that you're defining all of this in TS, you could encode that logic (i.e. threshold = `minimumProdInstances - 1`), which means that this will continue working as desired if you ever scale up.

cdk/lib/frontend.ts
```diff
@@ -18,5 +34,360 @@ export class Frontend extends GuStack {
...
+    const alarmEnabledInProd = this.withStageDependentValue({
+      app,
+      variableName: "AlarmEnabledInProd",
+      stageValues: {
+        [Stage.CODE]: false,
+        [Stage.PROD]: true,
+      },
+    });
+
+    new GuAlarm(this, "NoHealthyInstancesAlarm", {
```
jacobwinch, line 190: I wonder if you really need a dedicated alarm for this; I'd have thought that `HighELB5XXRateAlarm` would also be triggered under this scenario (as long as you have a small amount of traffic).

cdk/lib/frontend.ts
jacobwinch: I think that [this code](https://github.com/guardian/support-frontend/blob/main/support-frontend/app/wiring/AppLoader.scala#L33) is [overriding](https://github.com/guardian/simple-configuration#custom-ssm-location) the library's [default path](https://github.com/guardian/simple-configuration#configurationloaderload).

TomPretty: Ahh so it looks like the default path from `simple-configuration` does align with what's generated for us

TomPretty: I'll create a health card to investigate this and see if we can clean it up (I'm guessing we can!)

cdk/lib/frontend.ts
TomPretty: I think this would be a great improvement for all of those reasons that you stated. As it's not like for like I'll add a health card to address as a follow up so as to not block this PR going out

cdk/lib/frontend.ts
TomPretty: I'll create a health card for this one too. I think it'd be great if we can fallback on as many defaults as possible from the `GuEc2App` pattern. I'm not convinced we need two separate alarms (`NoHealthyInstances` and `ReducedHealthInstances`), so if we can make this one work it would save quite a lot of boilerplate

cdk/lib/frontend.ts
TomPretty: Yeah makes sense, again I've added something to chase this one up

cdk/lib/frontend.ts
TomPretty: Great suggestion! I've pushed that up

cdk/lib/frontend.ts
rupertbates: By this do you mean tagged to a specific AWS region?

cdk/lib/frontend.ts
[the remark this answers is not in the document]
rupertbates: The issue has always been around PII going into the central ELK stack, we'd need to check where we are with this currently

cdk/lib/frontend.ts
rupertbates: What is the advantage of getting it from parameter store? Wouldn't it just add another level of indirection making it harder to maintain and debug?

cdk/lib/frontend.ts
```diff
@@ -18,5 +34,384 @@ export class Frontend extends GuStack {
...
+          TargetGroup: ec2App.targetGroup.targetGroupFullName,
+        },
+        statistic: "Average",
+        period: Duration.seconds(60),
+      }),
+      snsTopicName: "reader-revenue-dev",
+    });
+
+    // TODO: Do we still need this?
+    new GuAlarm(this, "CatalogLoadingFailureAlarm", {
```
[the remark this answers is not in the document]
rupertbates: That would explain why we didn't get an alarm when the catalog was broken recently. I think we should probably fix this as it will bring the whole of subs down if we can't load the catalog.

cdk/lib/frontend.ts
TomPretty: Yeah that's right. I would have to defer to DevX for exactly the benefits that would give us. I think it was that keeping everything region specific would isolate our infrastructure to just a single region. This would e.g make it easier to run another version of the app in another region (although I doubt we'd ever do it?). I think it might not give us much (any?) benefit here. I think it's more that 'keep things region specific' is on the DevX best practice checklist so they highlighted it as we went along

rupertbates (approved):
👍 nice one!

cdk/lib/frontend.ts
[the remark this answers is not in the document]
TomPretty: That's good to know - I've created a [health card](https://trello.com/c/ARL9jV8g/367-send-support-frontend-logs-to-central-elk-stack) to investigate this, and specifically highlighted the issue around PII.

cdk/lib/frontend.ts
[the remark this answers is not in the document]
TomPretty: I've created a health card for this: https://trello.com/c/vUo5sO3R/368-fix-the-catalog-loading-failed-alarm-in-support-frontend

cdk/lib/frontend.ts
nicl: Yeah, this makes sense as latency is a form of availability. Frontend do something similar:

https://github.com/guardian/platform/blob/main/provisioning/cloudformation/frontend-service.yaml#L401

PR #3602, opened 2022-03-24
Phase 3 of the cdk migration

TomPretty:
In this phase we clean up all our old infrastructure. This will leave us with a single set of resources, but now being specified with cdk constructs. This involves dropping the `CfnInclude` construct, deleting the old cloudformation files, and cleaning up the temporary changes we needed whilst having a dual stack.

To deploy this we need to:

1. Wait for #3598 to be deployed
2. Deploy the cloudformation manually (It will delete the old infrastructure which riff-raff will block)
3. Merge PR

And then tada! We've finished migrating our infrastructure to cdk!


TomPretty:
[support-frontend cdk migration - phase 3](https://trello.com/c/bBsTujNE/359-support-frontend-cdk-migration-phase-3)

tomrf1 (approved):
👍

cdk/lib/frontend.ts
```diff
@@ -1,19 +1,17 @@
-import { join } from "path";
 import {
   ComparisonOperator,
   Metric,
   TreatMissingData,
 } from "@aws-cdk/aws-cloudwatch";
 import { InstanceClass, InstanceSize, InstanceType } from "@aws-cdk/aws-ec2";
 import { FilterPattern, LogGroup, MetricFilter } from "@aws-cdk/aws-logs";
-import { CfnInclude } from "@aws-cdk/cloudformation-include";
```
jacobwinch, line 9: You should be able to remove `@aws-cdk/cloudformation-include` from your `package.json` too. We recently discovered that this pulls in _a lot_ of transitive dependencies and significantly slows down builds (https://github.com/guardian/security-hq/pull/435), so it's definitely worth removing it if you can.

TomPretty: Oh that's a great spot - thanks! I'll remove that dep now

TomPretty: Yep that's given us lots of red in the `yarn.lock`!

PR #3614, opened 2022-03-29
Use default 5xx alarms provided by the GuEc2App pattern

TomPretty:
Currently we have two custom 5xx alarms defined: one for the load balancer; and one for the instances. This change replaces both of those with a single alarm provided to us by the GuEc2App pattern. This alarm looks at the total percentage of 5xx responses:

((5xxs from load balancer + 5xxs from instances) / total number of requests)

The benefits of switching to this alarm are:

  - reduction of boilerplate in this stack definition
  - picking up improvements 'for free' if/when DevX establish best practices for monitoring/alerting in this area
  - standardising this alarm across all of our apps more easily

[**Trello Card**](https://trello.com/c/eWqxmJre/363-use-default-5xx-alarm-generated-by-cdk-in-support-frontend)


TomPretty:
[Use default 5XX alarm generated by cdk in support-frontend](https://trello.com/c/eWqxmJre/363-use-default-5xx-alarm-generated-by-cdk-in-support-frontend)

cdk/lib/frontend.ts
```diff
@@ -220,53 +230,6 @@ export class Frontend extends GuStack {
...
 
-    new GuAlarm(this, "High5XXRateAlarm", {
-      app,
-      alarmName: alarmName(
-        "support-frontend instances are returning 5XX errors"
-      ),
-      alarmDescription: alarmDescription(
-        "Some or all actions on support website are failing"
-      ),
-      actionsEnabled: alarmEnabledInProd,
```
TomPretty, line 231: Hey @jacobwinch we have this stage dependent value `alarmEnabledInProd` used for all of our custom alarms. I can see that introducing the default 5xx alarm in the `GuEc2App` pattern has added a very similar `alarmActionsEnabled` variable - is there a way to get access to that to avoid redefining it?

cdk/lib/frontend.ts
```diff
@@ -136,7 +142,17 @@ export class Frontend extends GuStack {
         },
       },
       monitoringConfiguration: {
-        noMonitoring: true,
+        snsTopicName: "reader-revenue-dev",
+        http5xxAlarm: {
+          alarmName: alarmName(
+            "support-frontend instances are returning 5XX errors"
```
jacobwinch, line 148: Minor: I would suggest `support-frontend is returning 5XX errors` (because technically it might be the load balancer rather than the instances!)

cdk/lib/frontend.ts
jacobwinch: It's not trivial to access this variable at the moment and as it happens we are about to move away from the current approach (i.e. using mappings) entirely. There's a bit more context on that [here](https://github.com/guardian/cdk/blob/main/docs/architecture-decision-records/006-stage-awareness.md).

Given all of the above, I think this duplication is fine for now!

cdk/lib/frontend.ts
jacobwinch: Conveniently the suggested name is also unique, which means AWS will allow you to create it 😀

jacobwinch:
This all looks good to me from a `cdk` perspective; I won't formally approve this as it's probably better for others in your team to review the threshold you've chosen, since they will be the people responding to any alarms!

cdk/lib/frontend.ts
TomPretty: Yeah great suggestion - I've pushed that up now!

cdk/lib/frontend.ts
TomPretty: Sounds good & thanks for linking to that doc - it was an interesting read!

TomPretty:
> This all looks good to me from a `cdk` perspective; I won't formally approve this as it's probably better for others in your team to review the threshold you've chosen, since they will be the people responding to any alarms!

Thanks for taking a look - much appreciated!

rupertbates (approved):
Looks good 👍

PR #3615, opened 2022-03-29
Remove the SFD_V2 test

TomPretty:
This change removes the SFD_V2 test which has now finished running. I've removed all the references to the test, but kept the BenefitsMessage component for now.

[**Trello card**](https://trello.com/c/pW6tKPHb/366-clean-up-benefits-test)


marialani:
Do you think it'd be worth keeping in the test config but setting `isActive` to false? I know we have another test planned but having been off I'm not sure how similar it'll be

marialani (approved):
Left a quick question but this all looks good!

TomPretty:
> Do you think it'd be worth keeping in the test config but setting `isActive` to false? I know we have another test planned but having been off I'm not sure how similar it'll be

Good question! I'd say it's probably worth removing for now as it should be quite easy to set up for a new test

PR #3622, opened 2022-03-30
Webpack 5 upgrade

rupertbates:

This PR upgrades Webpack to version 5 as well as a load of other dependency updates.

Changes
- Upgrade webpack
- Migrate a load of scss files to remove deprecated syntax (division operator)
- Remove the storybook-static step from the production build - firstly because this is currently broken (see below) and also because it slows the build down and part of the reason for this work is to speed it up.

There are a couple of outstanding issues which it would be good to resolve in future PRs
- [ ] Storybook no longer works - we will need to upgrade it to version 6 which I have had a go at but ran into a large number of issues and so have left for a future PR.
- [ ] There are a number of noisy deprecation warnings which come from `sass-mq`. Version 6 should fix these but causes the build to fail as it can no longer find the correct path to mq - I spent quite a while debugging this but couldn't find a resolution so I'm sticking with the warnings for now.

[**Trello Card**](https://trello.com/c/I47w9e83/347-speed-up-support-frontend-build)


support-frontend/package.json
```diff
@@ -151,62 +153,63 @@
...
     "prettier-eslint": "^11.0.0",
     "react-svg-loader": "^3.0.3",
     "rimraf": "^2.6.2",
+    "sass": "^1.49.10",
     "sass-mq": "^5.0.0",
     "search-in-file": "^1.2.2",
     "ts-loader": "8.2.0",
     "tsc-files": "^1.1.2",
     "typescript": "^4.3.2",
-    "uglifyjs-webpack-plugin": "^2.2.0",
```
rupertbates, line 194: No longer required, Webpack 5 uses Terser internally

i-hardy (approved):
This looks great, really glad to have this upgraded 👍

PR #3629, opened 2022-04-01
Fix TS errors in linkTo.tsx

GHaberis:

Typescript error squashing in `linkTo.tsx`.


support-frontend/assets/pages/paper-subscription-landing/components/content/linkTo.tsx
```diff
@@ -34,14 +37,16 @@ function LinkTo({
...
 	isPricesTabLink?: boolean;
-}) {
+}): JSX.Element {
+	console.log('tab --->', tab);
+
 	return (
 		<Link
 			css={isPricesTabLink ? linkStyles(tab, activeTab) : linkColor}
-			href={paperSubsUrl(tab === 'delivery')}
+			href={paperSubsUrl(tab === 'HomeDelivery')}
```
GHaberis, line 49: Interestingly I thought I'd picked up a bug here. The code was incorrect as `tab` is the type `PaperFulfilmentOptions` which can only ever hold the value "HomeDelivery" or "Collection". I fixed the code and the went looking for the bug, but I found it never manifests itself to the user...

`linkTo` is used for the bottom tabs on the Newspaper landing page, the bottom tabs are anchor tags but we don't use the `href` property of these anchors to route a user to a new page when they click on them, instead we have an `onClick` handler that take control, the actions of the handler are correct - the correct Home Delivery or Collection content is displayed in the page. This differs from the way the top tabs on the Newspaper landing page behave, these are also anchor tags but they route a user through to either `/subscribe/paper` or `subscribe/paper/delivery`.

**Top tabs**
Screenshot 2022-04-01 at 14 48 50


**Bottom tabs**
Screenshot 2022-04-01 at 14 49 04

i-hardy: Weird 🤔 I feel like it's not great if the two separate sets of tabs have slightly different behaviour, but as long as the user sees the correct thing I suppose it's not a big deal.

i-hardy (approved):
Looks good, just a stray log statement to clear up

support-frontend/assets/pages/paper-subscription-landing/components/content/linkTo.tsx
GHaberis: I know, I don't think we should have 2 sets of behaviours either. I'm going to create a card to look into consolidating one way or the other.

PR #3633, opened 2022-04-04
Fix TS errors in summary.tsx

GHaberis:

Fix Typescript errors in `Summary.tsx`.

This `Summary.tsx` module returns the main `Summary` component. This Summary component is composed of a number of local components to the scope of this module: `DataList`, `PromotionDiscount`, `ChangeSubscription`, `DropDown`, `DropDownButton`, `TabletAndDesktop`, `Mobile`, `HideDropDown`, `ShowDropDown`.

For each component I've made the following changes:
- Given each components props a "PropTypes" type definition. Before we were passing around the `Summary` component's props like so `{...props}` on all components, even if only a few values from the `Summary` component's props needed to be passed.  I felt it really wasn't clear what props where actually being used and where they were being used, so decided to be more explicit about what we're passing to each component.
- I've also destructured the props passed to each component to be more explicit in the function signature about what's being passed to the component.


support-frontend/assets/components/priceLabel/priceLabel.tsx
```diff
@@ -8,31 +8,25 @@ import { getAppliedPromo, hasDiscount } from 'helpers/productPrice/promotions';
...
 	const description = getPriceDescription(productPrice, billingPeriod, true);
 	const promotion = getAppliedPromo(productPrice.promotions);
+
 	return (
 		<span {...props}>
 			{hasDiscount(promotion) && (
 				<del aria-hidden="true">{showPrice(productPrice)}</del>
 			)}
-			<span className={orderIsAGift && giftStyles}>{` ${description}`}</span>
+			<span>{` ${description}`}</span>
```
GHaberis, line 27: `giftStyles` are not defined in `summary.module.scss` so I've removed this.

support-frontend/assets/components/subscriptionCheckouts/summary.tsx
```diff
@@ -1,41 +1,113 @@
...
+	billingPeriod: BillingPeriod;
+	productPrice: ProductPrice;
+	title: string;
+	onClick: (event: React.MouseEvent<HTMLButtonElement>) => void;
+	paper: boolean;
+	showDropDown: boolean;
+};
+
+// Variable declarations
+const styles = moduleStyles as {
```
GHaberis, line 77: I followed this pattern of using a Type assertion for the styles in a `.scss` module as I've seen it elsewhere, it's not ideal as it it makes this object the source of truth for what styles are available, rather than the `.scss` module. Ideally we'd convert all these over to emotion `css` variables.

support-frontend/assets/components/subscriptionCheckouts/summary.tsx
```diff
@@ -1,41 +1,113 @@
...
 import type { SubscriptionProduct } from 'helpers/productPrice/subscriptions';
 import SvgDropdownArrowUp from './dropDownArrowUp.svg';
-import styles from './summary.module.scss';
+import moduleStyles from './summary.module.scss';
 
-type GridImageType = typeof import('components/gridImage/gridImage').default;
-// Types
-export type DataListItem = {
+// Type declarations
+type DataListItem = {
```
TomPretty, line 15: minor: I think we've so far had a slight preference for `interface` when it's applicable

support-frontend/ts-error-history.csv
```diff
@@ -160,4 +160,5 @@
 1648126201000,212,f502411126879af64724ef46a858be9704617e8a
 1648569026000,212,2f2bec26da28e32b493bf99fa96f4308d1cf30bf
 1648641260000,212,0f4c531e6f1be7056177066c3492499644c59389
-1649068104000,202,HEAD
+1649079015000,202,bd0117a6293e20fd9a057a0b1bcfd81965294f0f
+1649086374000,197,HEAD
```
TomPretty, line 164: nice! dropping below 200!

support-frontend/typescript-errors.json
```diff
@@ -107,6 +104,8 @@
             "assets/helpers/polyfills/layout.ts": 1,
             "assets/pages/digital-subscription-checkout/components/endSummary/endSummarySelector.ts": 1,
             "assets/pages/paper-subscription-checkout/helpers/orderSummaryText.ts": 1,
+            "assets/pages/weekly-subscription-checkout/components/weeklyCheckoutForm.tsx": 1,
+            "assets/pages/weekly-subscription-checkout/components/weeklyCheckoutFormGifting.tsx": 1,
```
TomPretty, line 108: I think these two new ones are because `datalist` is required in the `SummaryPropTypes`. I'm not quite sure why this wasn't an issue before.

I think it's quite a simple fix though, as we've already got `[]` in place as a fallback for when they're not passed so I think we can just mark `datalist` as optional?

GHaberis: Ah nice spot, fix incoming...

support-frontend/assets/components/subscriptionCheckouts/summary.tsx
GHaberis: Yeah i'm inclined to say we leave it for now and wait to migrate to emotion, this styles object does at least provide a quick overview of all the styles actually used in the module for whoever picks up the migration.

TomPretty: Yeah makes sense 👍

TomPretty (approved):
Looks great! 👍

PR #3636, opened 2022-04-05
France gw offer

tomrf1:
1. Moves GW to the top of /uk/subscribe and /int/subscribe
2. Makes FR2022 the default promo for UK/INT/EU, pulling in pricing and roundel copy
3. Sets the correct T&Cs link on the /subscribe/weekly page

/subscribe/weekly price cards:
Screen Shot 2022-04-05 at 11 13 49

/subscribe/weekly roundel copy:
Screen Shot 2022-04-05 at 11 13 38

checkout page:
Screen Shot 2022-04-05 at 11 13 19


support-frontend/app/controllers/WeeklySubscriptionController.scala
```diff
@@ -42,11 +43,16 @@ class WeeklySubscriptionController(
...
+    val defaultPromotion =
+      if (countryCode == "AU" || countryCode == "NZ") ausPromotion
+      else francePromotion
+
     val queryPromos =
       request.queryString
         .get("promoCode")
         .map(_.toList)
-        .getOrElse(List(DefaultPromotions.GuardianWeekly.NonGift.ausPromotion))
+        .getOrElse(List(defaultPromotion))
```
tomrf1, line 55: although the `landingCopyProvider` takes a `List` of promos, it calls `.headOption` and only uses the first one! So we have to pick the right one here based on countryCode

KaliedaRik (approved):
Thank you for letting me pair on this. Approved.

PR #3646, opened 2022-04-08
Upgrade to version 39 of CDK.

TomPretty:
This is a slightly bigger change as v39 introduced a breaking change around how we handle the stage parameter. We now create separate CODE and PROD stacks which allows us to do 'proper' logic with the stage parameter.

The main reason for doing this though is that we've made a [small tweak to the cdk library](https://github.com/guardian/cdk/pull/1191) to allow our instances to go back to using IMDSv1 which we currently require for shipping our logs to cloudwatch. We're doing the breaking change upgrade now so that it's simple to upgrade once our small tweak has been merged and a new version of cdk released.


cdk/lib/frontend.ts
```diff
@@ -117,71 +114,37 @@ export class Frontend extends GuStack {
       }),
     ];
 
-    const minimumProdInstances = 3;
-
```
TomPretty, line 121: we no longer need this as we have the `scaling` prop now and can get `scaling.minimumInstances` to get this value

cdk/lib/frontend.ts
```diff
@@ -117,71 +114,37 @@ export class Frontend extends GuStack {
...
-      },
+      scaling,
       instanceType: InstanceType.of(InstanceClass.T4G, InstanceSize.SMALL),
     });
 
     // ---- Alarms ---- //
 
-    // TODO: remove [CDK] prefix. This prefix is needed as the alarms need to have
-    // a different name from the ones in the cloudformation template. We can remove
-    // this after the migration
```
TomPretty, line 162: sneaking this in as I forgot to remove this comment when I removed the [CDK] prefix!

cdk/lib/__snapshots__/frontend.test.ts.snap
```diff
@@ -2,28 +2,6 @@
...
-        "AlarmEnabledInProd": true,
-        "MembershipSubPromotions": "arn:aws:dynamodb:*:*:table/MembershipSub-Promotions-PROD",
-        "RedemptionCodes": "arn:aws:dynamodb:*:*:table/redemption-codes-PROD",
-        "domainName": "support.theguardian.com.origin.membership.guardianapis.com",
-        "hostedZoneId": "Z1E4V12LQGXFEC",
-        "maxInstances": 6,
-        "minInstances": 3,
-      },
-    },
-  },
```
TomPretty, line 26: We end up getting rid of all of these as now we have a separate CODE and PROD template we don't have to do any lookups

cdk/bin/cdk.ts
```diff
@@ -4,4 +4,40 @@ import { Frontend } from "../lib/frontend";
 
 const app = new App();
 const cloudFormationStackName = process.env.GU_CFN_STACK_NAME;
-new Frontend(app, "Frontend", { stack: "support", cloudFormationStackName });
+
+new Frontend(app, "Frontend-PROD", {
```
TomPretty, line 8: I followed the example in [this PR](https://github.com/guardian/amiable/pull/117) for upgrading to CDK 39. They pulled out all of the stage dependent values as props to the stack, which is what I've done here

cdk/bin/cdk.ts
```diff
@@ -4,4 +4,40 @@ import { Frontend } from "../lib/frontend";
...
+  stack: "support",
+  stage: "PROD",
+  cloudFormationStackName,
+  membershipSubPromotionsTable:
+    "arn:aws:dynamodb:*:*:table/MembershipSub-Promotions-PROD",
+  redemptionCodesTable: "arn:aws:dynamodb:*:*:table/redemption-codes-PROD",
+  certificateProps: {
+    domainName: "support.theguardian.com.origin.membership.guardianapis.com",
+    hostedZoneId: "Z1E4V12LQGXFEC",
```
akash1810, line 17: As this value is static for each stage, we could probably have it in the `lib/frontend.ts` once.

akash1810 (approved):
Wonderful! I think this is the first stack outside of DevX using v39.0.0 🎉 . Any feedback welcomed.

cdk/bin/cdk.ts
TomPretty: Good spot! I'll factor that out now 👍

PR #3647, opened 2022-04-08
Use IMDSv1 on support-frontend instances.

TomPretty:

This change updates CDK to v40.0.1 (and all aws-cdk packages to v1.152.0) in order to make use of the new `withoutImdsv2` prop introduced to the GuEc2App.

Since migrating to CDK our instances have been using IMDSv2 (= Instance MetaData Service v2). This changes the way that instances get access to their own metadata and is incompatible with the cloudwatch agent that we are using. This means that we've been without logs on support-frontend since merging phase 2 of the CDK migration.

As a long term solution, we should either:

a) Use the new version of the cloud watch agent
b) Stop using cloudwatch alltogether and ship our logs to the central ELK stack

Both of those would be preferable as using IMDSv2 is considered best practice but this is the quickest way to get our logs back online for now.


cdk/lib/__snapshots__/frontend.test.ts.snap
```diff
@@ -124,9 +124,6 @@ Object {
           "Ref": "AMIFrontend",
         },
         "InstanceType": "t4g.small",
-        "MetadataOptions": Object {
-          "HttpTokens": "required",
-        },
```
TomPretty, line 129: This is the part of the cloudformation that dictates if your instances will use `IMDSv1` or `IMDSv2`. Requiring tokens = using v2, so removing this is exactly what we want to see

tomrf1 (approved):

PR #3653, opened 2022-04-11
Prices endpoint

tomrf1:
This adds a `/prices` endpoint.
It takes the internal `ProductPrices` model used by support-frontend and returns a simplified model for use by support-dotcom-components. This enables us to reference regional product prices in dotcom messages.

For now I've just included Digisub and Weekly, and Monthly/Annual rate plans.

If there are promos for a given region/product/rate-plan, then it goes with the `discountedPrice` of the first promo. Otherwise it falls back on the standard `price`.

E.g. response:
Screen Shot 2022-04-11 at 10 52 16

### DIFF
--- a/support-frontend/app/controllers/PricesController.scala
+++ b/support-frontend/app/controllers/PricesController.scala
@@ -47,20 +47,7 @@ object PricesController {
       Canada: CountryGroupPriceData,
   )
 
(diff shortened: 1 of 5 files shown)


TomPretty (approved):
Looks good to me 👍

PR #3656, opened 2022-04-11
Don't alarm for disallowed email addresses

rupertbates:

#3436 introduced better handling of disallowed email addresses, however we are still triggering an alarm every time one is used which is unnecessary now that we inform the end user and allow them to input a different email.

This PR removes the alarm for this type of email error.

[**Trello Card**](https://trello.com/c/cipTaMhA/333-prevent-alarm-when-user-signs-up-with-a-blocked-email-domain)

### DIFF
--- a/support-frontend/app/actions/CustomActionBuilders.scala
+++ b/support-frontend/app/actions/CustomActionBuilders.scala
@@ -33,7 +33,7 @@ class CustomActionBuilders(
       cc.parsers.defaultBodyParser,
     )
 
-  case class LoggingAndAlarmOnFailure[A](chainedAction: Action[A]) extends EssentialAction {
+  case class LoggingAndAlarmOnException[A](chainedAction: Action[A]) extends EssentialAction {
 
     def apply(requestHeader: RequestHeader): Accumulator[ByteString, Result] = {
       val accumulator = chainedAction.apply(requestHeader)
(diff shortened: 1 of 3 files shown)


support-frontend/app/actions/CustomActionBuilders.scala
```diff
@@ -44,16 +42,6 @@ class CustomActionBuilders(
         byteString
       })
       loggedAccumulator
-        .map { result =>
```
johnduffell, line 47: I don't think this will catch JSON deserialisation errors any more if we do this change.

I think json deserialisation error is a 4xx but it occurs at the circe level rather than in the application logic, hence why I had to put this block in at the Play level.  I think one of the original issues was a gclid serialisation issue with guardian weekly, hence the need for this particular approach.

Might be worth testing that out by POSTing an invalid json to CODE and seeing what the status code is.

rupertbates: Good idea I'll try that

rupertbates: Ok I've changed this now so that we still alarm in the action but 400s from the controller will have the `result.header.reasonPhrase` value set to the error code allowing us to filter out ones from disallowed emails

johnduffell (approved):
nice, didn't know about the reasonPhrase being customisable, a nice way to get the information across to the filter.  I think most things ignore the actual value. 👍

PR #3662, opened 2022-04-12
Update Aus map colours for upcoming Aus election moment.

TomPretty:

Update Aus map colours for upcoming Aus election moment.

## Screenshots

Screenshot 2022-04-12 at 17 02 56


marialani (approved):
👍

PR #3671, opened 2022-04-14
Enable variable pricing of digital subscriptions

rupertbates:

## Why are you doing this?
This work is to support the second benefits test. It allows us to create a digital subscription with a price set by the user reflecting the amount of their contribution.

The general idea behind the test is that for people in the test variant who pay over the required amount we would send a slightly different json body to the `/create` endpoint - replacing the `RegularContribution` type with a `DigitalSubscription` type see types here: https://github.com/guardian/support-frontend/blob/928e349c51dd997e798eb3968f535b9163b301f1/support-frontend/assets/helpers/forms/paymentIntegrations/readerRevenueApis.ts#L52-L65

From that point on the user will be set up as a digital subscriber rather than a  regular contributor. 

However, we also want to use the contribution amount as the price of the digital subscription and to support this we need to do some back end work which is in this PR. 

The `DigitalPack` `ProductType` https://github.com/guardian/support-frontend/blob/3647c46ffb6486172fffb6e86ccdb1f2a16b3330/support-models/src/main/scala/com/gu/support/workers/Products.scala#L32-L37 accepted by the backend now has an optional `amount` number type in the same way as a `Contribution`. Passing this in will set that amount as the price (monthly or annual as appropriate) of the purchased digital subscription.

## TODO
I have put a hard-coded value for this in submit.ts with a TODO next to it as a proof of concept, this shows how we need to pass the amount through to the backend but should be removed before going live as otherwise it would hard code that value for all digisub purchases!

We should probably check a user is in the test variant on the server before using the amount value, just as a precaution but this requires more extensive changes so I have left it for now.


support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionDirectPurchaseBuilder.scala
```diff
@@ -49,4 +51,22 @@ class DigitalSubscriptionDirectPurchaseBuilder(
 
   }
 
+  def overridePricingIfRequired(state: DigitalSubscriptionDirectPurchaseState) = {
+    state.product.amount
+      .map { amount =>
+        // TODO: Limit this specifically to people in the benefits test
```
TomPretty, line 57: Would it be worth also checking it's over some minimum in order to avoid a malicious user constructing a payload with a very low price?

rupertbates:
Changes are in #3695

support-frontend/assets/helpers/subscriptionsForms/submit.ts
```diff
@@ -101,6 +101,7 @@ const getProduct = (
 			currency: currencyId ?? state.common.internationalisation.currencyId,
 			billingPeriod,
 			readerType,
+			amount: 20, //TODO: Put the actual contribution amount here
```
GHaberis, line 104: Is this value correct? Does it need updating?

PR #3673, opened 2022-04-14
Set up Redux Toolkit state for personal details

i-hardy:

This adds a new Redux Toolkit slice for managing personal details state- name, email address, etc- and integrates it with the subscriptions and redemption checkouts. A second PR will follow to integrate it with the contributions checkout.

This also introduces [listener middleware](https://redux-toolkit.js.org/api/createListenerMiddleware) to handle side effects that we want to associate with certain actions.

[**Trello Card**](https://trello.com/c/8HKdZtbv)

## Why are you doing this?

We collect names and email addresses in every checkout so this was an obvious candidate for the first shared Redux slice, and adding it in a new sub-section of the existing checkout state minimises the amount that has to be changed about the existing state for other parts of the form.

Side effects of state changes are currently handled in multiple different ways in different parts of the codebase, and can be quite hard to understand and track down. The listener middleware functions in a similar way to a `useEffect` hook, allowing us to explicitly associate specific side effects with a specific action being dispatched and allowing actions themselves to remain pure while serving as a single source of truth for side effects.

This is a fairly new Redux Toolkit feature (hence the version bump) and I fully expect our usage of it and the way we organise side effect-related code to evolve as we migrate more of the codebase- this is just a first pass!


i-hardy:
[🛠 Redux Toolkit migration - unify personal details state 🛠](https://trello.com/c/8HKdZtbv/321-%F0%9F%9B%A0-redux-toolkit-migration-unify-personal-details-state-%F0%9F%9B%A0)

support-frontend/assets/helpers/redux/checkout/personalDetails/state.ts
```diff
@@ -0,0 +1,25 @@
...
+	firstName: string;
+	lastName: string;
+	email: string;
+	confirmEmail?: string;
+	isSignedIn: boolean;
+	userTypeFromIdentityResponse: UserTypeFromIdentityResponse;
+	telephone?: string;
+};
+
+const user = getUser();
```
GHaberis, line 16: By defining `user` here we'll have a side affect that `getUser()` is called when this module is imported. Could we convert `initialPersonalDetailsState` from na object to a function that returns an object, we can then define `user` within this function, so `getUser()` is only called at that point it's needed?

i-hardy: No, the initial state has to be an object literal in order to work correctly with `createSlice`. Store components in general need to be static in order to be able to get the utility types for things like `dispatch`.

We really ought to be providing this 'starter' information after store initialisation, but unfortunately due to the way we've been using Redux up until this point we have far too much code that relies on the store having a bunch of initial state in it at the point of component render. This is something we can't really fix until everything else has been migrated.

GHaberis: OK, sounds unavoidable in this case, thanks for explaining

GHaberis (approved):
Looks good to me 👍

PR #3685, opened 2022-04-22
Add some new acquisition data values to support apps

rupertbates:
The apps teams are going to start using the acquisitions api we built rather than sending events via Ophan. To enable this we need to add some missing data values into the acquisition model.

[**Trello Card**](https://trello.com/c/OkahHCdA/412-add-new-values-to-acquisition-data-model-to-support-apps)


i-hardy (approved):
👍

PR #3690, opened 2022-04-26
Update cmp library and remove dynamic imports.

TomPretty:
Previously we were importing the cmp library with dynamic imports. This was to avoid loading the cmp during SSR which lead to build errors. Those errors are no longer an issue so we removed the dynamic imports and replace them with regular imports. We also updated the version of the cmp library we were using as we were quite a few major versions behind.


support-frontend/assets/components/footerCompliant/Footer.tsx
```diff
@@ -1,14 +1,13 @@
 // ----- Imports ----- //
-// @ts-expect-error - required for hooks
```
TomPretty, line 2: I also fixed up the errors in this file as they were all pretty easy. There errors were:

- this comment
- using `Node` instead of `ReactNode`
- Not having a return type on the component

TomPretty:
[Upgrade @guardian/consent-management-platform to v10.2.2 in support-frontend](https://trello.com/c/S2A44SVC/335-upgrade-guardian-consent-management-platform-to-v1022-in-support-frontend)

i-hardy (approved):
👍 Looks good, nice to get rid of all the complexity around the dynamic imports.

PR #3692, opened 2022-04-26
Integrate new personal details state into the contributions checkout

i-hardy:

Following on from #3673, this integrates the new personal details slice into the contributions checkout in a similar manner.

[**Trello Card**](https://trello.com/c/8HKdZtbv)

## Why are you doing this?

This is the first bit of shared checkout state code between the subs and contributions checkouts- the first step towards a less confusing future 😄 

This also introduces the magic of [the object shorthand for mapDispatchToProps](https://react-redux.js.org/using-react-redux/connect-mapdispatch#defining-mapdispatchtoprops-as-an-object), which both avoids a nasty type error I encountered (see reduxjs/redux-toolkit#2285) and significantly simplifies the process of connecting component props to store actions. We should be able to make use of this in most if not all other components that are connected to the store.


i-hardy:
[🛠 Redux Toolkit migration - unify personal details state 🛠](https://trello.com/c/8HKdZtbv/321-%F0%9F%9B%A0-redux-toolkit-migration-unify-personal-details-state-%F0%9F%9B%A0)

support-frontend/assets/components/StripePaymentRequestButton/StripePaymentRequestButton.tsx
```diff
@@ -211,7 +199,7 @@ function updateTotal(props: PropTypes) {
 function getClickHandler(props: PropTypes, isCustomPrb: boolean) {
 	function onClick(event: StripePaymentRequestButtonElementClickEvent) {
 		trackComponentClick('apple-pay-clicked');
-		props.setAssociatedPaymentMethod();
+		props.setAssociatedPaymentMethod(Stripe);
```
rupertbates, line 202: What is this change doing?

rupertbates (approved):
Looks good to me, I haven't run it locally but will do tomorrow. It's a big improvement though!

support-frontend/assets/components/StripePaymentRequestButton/StripePaymentRequestButton.tsx
i-hardy: Previously, on line 127, the `setAssociatedPaymentMethod` prop didn't take an argument but always passed `Stripe` to the `updatePaymentMethod` action creator; this just passes the argument directly given that the prop is now just a reference to `updatePaymentMethod`.

This should ultimately be removed as I'm pretty sure this is one of multiple non-serialisable values being put into Redux for the contributions checkout, but that's another large chunk of work.

support-frontend/assets/helpers/redux/checkout/personalDetails/contributionsSideEffects.ts
```diff
@@ -0,0 +1,55 @@
...
+	setUserTypeFromIdentityResponse,
+} from './actions';
+
+export function addPersonalDetailsSideEffects(
+	startListening: ContributionsStartListening,
+): void {
+	startListening({
+		actionCreator: setFirstName,
+		effect(_, listenerApi) {
+			enableOrDisableForm()(
```
markerikson, line 17: :hand: It looks like `enableOrDisableForm` is a standard Redux thunk.  As such, the right way to use it is:

```js
dispatch(enableOrDisableForm())
```

and then the thunk middleware intercepts the thunk function and actually calls it with `dispatch` and `getState` as arguments:

https://redux.js.org/usage/writing-logic-thunks#thunk-usage-patterns

(also you would never need to do `getState.bind()` - it's just a simple getter function, not anything that relies on `this`)

support-frontend/assets/helpers/redux/checkout/personalDetails/contributionsSideEffects.ts
```diff
@@ -0,0 +1,55 @@
...
+				listenerApi.dispatch,
+				listenerApi.getState.bind(listenerApi),
+			);
+		},
+	});
+
+	startListening({
+		actionCreator: setLastName,
```
markerikson, line 25: :hand: It looks like you're doing the exact same side effect in multiple cases (either the `setFirstName` or `setLastName` action being dispatched).

You can use RTK's "matching" utilities to run the effect in both cases:

```js
const isNameUpdateAction = isAnyOf(setFirstName, setLastName);

startListening({
  matcher: isNameUpdateAction,
  effect: (action, listenerApi) => {
    dispatch(enableOrDisableForm())
  }
})
```

See https://redux-toolkit.js.org/api/matching-utilities#isanyof and https://redux-toolkit.js.org/api/createListenerMiddleware#startlistening

PR #3695, opened 2022-04-27
Contrib Landing page - single front door v3 ABC test 

marialani:
This PR sets up version 3 of the single front door referrer controlled ABC test on the contributions landing page which rolls benefits into a single support-led proposition. 

For the two variants of this test we will display a benefits message to users which details the additional benefits they receive if a certain predefined threshold amount is met for their recurring contribution.

The control maintains the same contributions flow apart from the removal of the blurb below “Support our journalism…”. 

The two variants follow the SFD flow with a benefits message using either a bullet point list or paragraph. 

Other changes to the flow for users in either variant who meet the benefits threshold:
- destination of the terms and conditions link
- form button submission text (Contribute -> Pay)
- thank you page header text

## Server side changes
There are a number of server side changes to support variable pricing for digital subscriptions, these are described in #3671 but these changes have been rolled into this PR along with completion of the TODOs which were left in that PR.

[**Trello Card**](https://trello.com/c/DiBWSUBp/339-build-digital-subscriptions-benefits-test-second-benefits-iteration)

## Why are you doing this?
To support the new product proposition work we are going to run an ABC test to measure the effect of the current contribute/subscribe split messaging vs a single 'support the guardian' call to action on the main site. 

Users in the single front door experience will receive supporter benefits from a supporter newsletter to digital subscription benefits, which scale in value when they support us with a higher contribution

## Is this an AB test?
- [x] Yes
- [ ] No

## Screenshots
|                     	 	| Bullet Points Variant 	| Paragraph Variant 	|
|---------------------	|-----------------------	|-------------------	|
| LP Below Threshold    	|       support thegulocal com_uk_contribute (1)                	|      support thegulocal com_uk_contribute (3)      	|
| LP Above  Threshold 	|      support thegulocal com_uk_contribute (2)           	|           support thegulocal com_uk_contribute (4)      	|
| TY Page    	|           support theguardian com_uk_thankyou  | Same as other variant |


support-services/src/main/scala/com/gu/support/abtests/BenefitsTest.scala
```diff
@@ -0,0 +1,35 @@
+package com.gu.support.abtests
+
+import com.gu.i18n.Currency
+import com.gu.i18n.Currency.{CAD, EUR, GBP, NZD, USD}
+import com.gu.support.acquisitions.AbTest
+import com.gu.support.workers.{Annual, BillingPeriod, DigitalPack, Monthly}
+
+object BenefitsTest {
+  def isValidBenefitsTestPurchase(product: DigitalPack, maybeAbTests: Option[Set[AbTest]]) =
```
johnduffell, line 9: what does it mean if `maybeAbTests` is `None`?

support-frontend/app/utils/CheckoutValidationRules.scala
```diff
@@ -155,11 +163,23 @@ object DigitalPackValidation {
         currencyIsSupportedForCountry(country, currency) and
         PaidProductValidation.noEmptyPaymentFields(paymentFields)
 
+    def isInBenefitsTest(supportAbTests: Set[AbTest]) = {
+      if (
+        isValidBenefitsTestPurchase(
+          createSupportWorkersRequest.product.asInstanceOf[DigitalPack],
```
johnduffell, line 169: rather than casting at run time, just change this line to pass in the product as a `DigitalPack`
https://github.com/guardian/support-frontend/blob/b13ecceb880dbf292b22f87524f014292732d991/support-frontend/app/utils/CheckoutValidationRules.scala#L36

support-frontend/app/utils/CheckoutValidationRules.scala
```diff
@@ -155,11 +163,23 @@ object DigitalPackValidation {
...
+        Invalid("User is not in the benefits test")
+    }
+
     val Purchase = Left
     type Redemption[A, B] = Right[A, B]
 
     (paymentFields, readerType, createSupportWorkersRequest.giftRecipient) match {
-      case (Purchase(paymentFields), ReaderType.Direct, None) => isValidPaidSub(paymentFields)
+      case (Purchase(paymentFields), ReaderType.Direct, None) =>
+        isInBenefitsTest(createSupportWorkersRequest.supportAbTests) or isValidPaidSub(paymentFields)
```
johnduffell, line 182:
```suggestion
        isValidBenefitsTestSub(createSupportWorkersRequest.supportAbTests) or isValidPaidSub(paymentFields)
```

support-services/src/main/scala/com/gu/support/abtests/BenefitsTest.scala
```diff
@@ -0,0 +1,35 @@
...
+      case (Monthly, GBP) => 12
+      case (Annual, GBP) => 119
+      case (Monthly, USD) => 20
+      case (Annual, USD) => 199
+      case (Monthly, EUR) => 15
+      case (Annual, EUR) => 149
+      case (Monthly, NZD) => 24
+      case (Annual, NZD) => 235
+      case (Monthly, CAD) => 22
+      case (Annual, CAD) => 219
```
johnduffell, line 29: I'm impressed that these line up so well!

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionDirectPurchaseBuilder.scala
```diff
@@ -49,4 +54,22 @@ class DigitalSubscriptionDirectPurchaseBuilder(
 
   }
 
+  def overridePricingIfRequired(product: DigitalPack, maybeAbTests: Option[Set[AbTest]]) =
+    if (isValidBenefitsTestPurchase(product, maybeAbTests)) {
+      product.amount
+        .map { amount =>
+          val ratePlanChargeId =
+            if (product.billingPeriod == Monthly) config.monthlyChargeId else config.annualChargeId
```
johnduffell, line 62: so we have a new "monthly" and "annual" digital sub charge under the same rateplan, which we use for setting the charge of non standard priced digital subs, correct?

johnduffell (approved):
scala looks good subject to the comments made!
I assume it's been tested in CODE with existing other products.
other than that nice work 👍 Looks like an exciting test to get out there!

support-frontend/assets/pages/contributions-landing/components/DigiSubBenefits/LiveFeedBack.tsx
```diff
@@ -0,0 +1,113 @@
...
+	countryGroupId: CountryGroupId;
+}
+
+function LiveFeedBack({
+	contributionType,
+	countryGroupId,
+}: LiveFeedBackProps): JSX.Element {
+	const showLiveFeedBack = useLiveFeedBackContext()?.showLiveFeedBack;
+	const setShowLiveFeedBack = useLiveFeedBackContext()?.setShowLiveFeedBack;
+
```
GHaberis, line 93: We call `useLiveFeedBackContext()` twice, once when defining `showLiveFeedBack` and again when defining `setShowLiveFeedBack`, could we call it once and assign the value to a variable to save running it twice?

support-frontend/app/utils/CheckoutValidationRules.scala
```diff
@@ -155,11 +163,23 @@ object DigitalPackValidation {
...
 
+    def isInBenefitsTest(supportAbTests: Set[AbTest]) = {
+      if (
+        isValidBenefitsTestPurchase(
+          createSupportWorkersRequest.product.asInstanceOf[DigitalPack],
+          Some(createSupportWorkersRequest.supportAbTests),
+        )
+      ) Valid
+      else
+        Invalid("User is not in the benefits test")
```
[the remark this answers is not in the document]
rupertbates: The thing is that this will be the case for anyone not in the variant of this test so is not actually an error condition, we are just checking that if they ARE in the test then certain conditions have been met

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionDirectPurchaseBuilder.scala
rupertbates: It is the same charge as we use for all monthly & annual digisubs currently, we're just overriding the amount.

support-services/src/main/scala/com/gu/support/abtests/BenefitsTest.scala
rupertbates: From the point of view of this method it just means that they are not in the test variant for this test so this is not a valid benefits test purchase

support-frontend/assets/pages/contributions-landing/components/ContributionForm.tsx
```diff
@@ -377,6 +406,23 @@ function ContributionForm(props: PropTypes): JSX.Element {
 				)}
 			</div>
 
+			{props.benefitsMessagingAbTestBulletVariant && !isAUDCountryGroup && (
```
GHaberis, line 409: The `!isAUDCountryGroup` check is also made in `shouldShowBenefitsMessaging()` so I'm not sure we also need to check it again here as well?

GHaberis (approved):
I've looked over the typescript and looks good 👍

support-services/src/main/scala/com/gu/support/abtests/BenefitsTest.scala
TomPretty: Amazingly well!

support-frontend/app/utils/CheckoutValidationRules.scala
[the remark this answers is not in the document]
johnduffell: late comment
I see what you mean that this is almost a pre-check rather than the actual validation.
It is indeed a good idea to avoid the "try this, and if it fails try this" as the failure messages get super comples.
However in this case if the purchase isn't detected as in the test when it should be, then I would like to know the exact reason why it thought they weren't in the benefits test.

PR #3702, opened 2022-04-29
Remove contrib specific terms and conditions copy for benefits test

marialani:
This removes contrib specific t&cs copy for v3 of the benefits test.


rupertbates:
Are we removing that copy for people in the control as well as the variant then?

marialani:
> Are we removing that copy for people in the control as well as the variant then?

No we're not, I need push up some changes!

GHaberis:
> Are we removing that copy for people in the control as well as the variant then?

I think the Contribution T&Cs needs to remain for the Control and for those users in the Variants making contributions beneath the threshold that entitles them to a Digital Sub.

PR #3704, opened 2022-04-29
Benefits test submit button, ts and cs bug fixes

marialani:
This removes contrib specific t&cs copy for v3 of the benefits test and ensures we only change the copy for the the submit button on the contrib landing page when the user is in a test variant.

Below the threshold amount for both test variants the copy on the button will read "Contribute ..."
## Screenshots
|           	| Test variant	| Control 	|
|-----------	|--------------------------------	|---------	|
| Single    	|            Screenshot 2022-04-29 at 12 07 30                     	|      Screenshot 2022-04-29 at 12 05 10   	|
| Recurring 	|                Screenshot 2022-04-29 at 12 09 31                 	|        Screenshot 2022-04-29 at 12 12 04 	|


rupertbates (approved):
LGTM

PR #3707, opened 2022-05-03
Debug scala-steward job

rupertbates:
We have a github action which should run against PRs from scala steward to make sure that the build is successful and all the tests pass. This is currently being skipped however, so this PR adds a debugging action to try to work out why.

### DIFF
--- a/.github/workflows/env.yml
+++ b/.github/workflows/env.yml
@@ -1,0 +1,15 @@
+name: Print environment variables
+
+on:
+  pull_request:
+
+jobs:
+  print_env_variables:
+    name: print environment variables
+    runs-on: ubuntu-latest
+    steps:
+      - name: Dump GitHub context
+        id: github_context_step
+        run: echo $JSON
+        env:
+          JSON: ${{ toJSON(github) }}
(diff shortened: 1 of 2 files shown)


.github/workflows/scala-steward.yml
```diff
@@ -8,9 +8,9 @@ env:
 
 jobs:
   run_sbt_tests:
-    if: github.actor == 'scala-steward'
+    if: github.event.pull_request.head.repo.full_name == 'scala-steward/support-frontend'
```
rupertbates, line 11: I think this is most likely the cause of the issue.

.github/workflows/env.yml
```diff
@@ -0,0 +1,15 @@
+name: Print environment variables
```
rupertbates, line 1: Print out all the environment variables for this job

marialani (approved):

.github/workflows/scala-steward.yml
kelvin-chappell: 👍  The difference should be that previously the condition was true for PRs when first raised but not when updated and with this it should be true for updated PRs too.  As it's now a 'what' test rather than a 'who'.  Unless the condition wasn't true for new PRs either, in which case I don't know what's going on.

PR #3721, opened 2022-05-09
Persist query parameters when changing URL via the country switcher

i-hardy:

Co-authored with @marialani, @TomPretty and @GHaberis  

This retains the current URL's query parameters on all links to other localised pages in the country switcher, in order to keep promo codes and referrer data when the user changes their locale.

This will _not_ affect referrer-controlled A/B tests as these are checked against the designated country audiences before A/B participation is determined- we have checked this in the mob programming session.

[**Trello Card**](https://trello.com/c/JoKHTCrg)


i-hardy:
[Fix: persist query string parameter when switching country via the country dropdown in the nav](https://trello.com/c/JoKHTCrg/429-fix-persist-query-string-parameter-when-switching-country-via-the-country-dropdown-in-the-nav)

TomPretty (approved):
👍

PR #3739, opened 2022-05-16
Change reader type on Patron benefit subs

paulbrown1982:

Making a change for the platform to save a Guardian Weekly or Digital Subscription with a reader type of Patron if the patron promo code is used to buy it. 

[**Trello Card**](https://trello.com) - TBC incident resolution

## Why are you doing this?
This will let me make a data change to update historical subs, and also have the reader type act as the flag for manage-frontend to display the subscription in an appropriate context (e.g. free with Patron, no next billing date etc)

## Is this an AB test?
- [ ] Yes
- [ ] No

### DIFF
--- a/support-workers/.history
+++ b/support-workers/.history
@@ -1,0 +1,4 @@
+project support-workers
+project
+support-workers
+exit
--- a/support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionDirectPurchaseBuilder.scala
+++ b/support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionDirectPurchaseBuilder.scala
@@ -4,10 +4,11 @@ import com.gu.helpers.DateGenerator
 import com.gu.support.abtests.BenefitsTest.isValidBenefitsTestPurchase
 import com.gu.support.acquisitions.{AbTest, AcquisitionData}
 import com.gu.support.config.{TouchPointEnvironment, ZuoraDigitalPackConfig}
-import com.gu.support.promotions.{PromoError, PromotionService}
+import com.gu.support.promotions.{PromoCode, PromoError, PromotionService}
 import com.gu.support.workers.ProductTypeRatePlans.digitalRatePlan
 import com.gu.support.workers.states.CreateZuoraSubscriptionProductState.DigitalSubscriptionDirectPurchaseState
 import com.gu.support.workers.{DigitalPack, Monthly}
+import com.gu.support.zuora.api.ReaderType.Patron
 import com.gu.support.zuora.api._
 import com.gu.zuora.subscriptionBuilders.ProductSubscriptionBuilders.{applyPromoCodeIfPresent, validateRatePlan}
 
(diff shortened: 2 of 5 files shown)


support-workers/.history
```diff
@@ -0,0 +1,4 @@
+project support-workers
+project
+support-workers
+exit
```
kelvin-chappell, line 4: What is this file?  Is it meant to be checked in?

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionDirectPurchaseBuilder.scala
```diff
@@ -54,6 +55,10 @@ class DigitalSubscriptionDirectPurchaseBuilder(
 
   }
 
+  def overrideReaderTypeIfRequired(promoCode: Option[PromoCode], readerType: ReaderType): ReaderType = {
+    if (promoCode.exists(_.endsWith("PATRON"))) Patron else readerType
```
kelvin-chappell, line 59: Would be good to avoid repeating the two instances of this condition

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/GuardianWeeklySubscriptionBuilder.scala
```diff
@@ -91,6 +92,16 @@ class GuardianWeeklySubscriptionBuilder(
     }
   }
 
+  private[this] def overrideReaderTypeIfRequired(state: GuardianWeeklyState): ReaderType = {
+    if (state.giftRecipient.isDefined) {
+      Gift
+    } else if (state.promoCode.exists(_.endsWith("PATRON"))) {
```
kelvin-chappell, line 98: As above.  Would be good to extract the condition and share it

support-workers/.history
johnduffell: this should be deleted - I think paul must have run `sbt` from the subfolder by mistake

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionDirectPurchaseBuilder.scala
```diff
@@ -32,11 +33,11 @@ class DigitalSubscriptionDirectPurchaseBuilder(
...
 
     val subscriptionData = subscribeItemBuilder.buildProductSubscription(
-      productRatePlanId,
-      overridePricingIfRequired(state.product, acquisitionData.map(_.supportAbTests)),
+      productRatePlanId = productRatePlanId,
+      ratePlanCharges = overridePricingIfRequired(state.product, acquisitionData.map(_.supportAbTests)),
       contractEffectiveDate = todaysDate,
       contractAcceptanceDate = contractAcceptanceDate,
-      readerType = state.product.readerType,
+      readerType = overrideReaderTypeIfRequired(state.promoCode, state.product.readerType),
```
johnduffell, line 40: would it be possible set this up front so it's consistent, rather than overriding it when we write to zuora? or would that cause other issues?

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/GuardianWeeklySubscriptionBuilder.scala
```diff
@@ -91,6 +92,16 @@ class GuardianWeeklySubscriptionBuilder(
     }
   }
 
+  private[this] def overrideReaderTypeIfRequired(state: GuardianWeeklyState): ReaderType = {
```
johnduffell, line 95: actually, DigitalPack is the only Product that has a ReaderType specified up front
https://github.com/guardian/support-frontend/blob/34a01860ac4ad467d7a471528d6f0439324e21c6/support-models/src/main/scala/com/gu/support/workers/Products.scala#L35
so this function isn't overriding at all, it's just deciding it
```suggestion
  private[this] def readerTypeFromState(state: GuardianWeeklyState): ReaderType = {
```

kelvin-chappell (approved):
👍

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/DigitalSubscriptionDirectPurchaseBuilder.scala
paulbrown1982: I get your point, however there is no way to be certain that the promo code is validated - only the backend can do that - so it would be risky for a promo code to drive any changes to the reader type at this stage of the process.

johnduffell: agree, makes sense to validate before writing to zuora.

johnduffell (approved):
great and good tht you added tests to this so we know it works reliably. 👍

PR #3740, opened 2022-05-16
Typescript Error Fixing: ProductPrices Type

GHaberis:
#### Changes:

This PR updates the `ProductPrices` type in `support-frontend/assets/helpers/productPrice/productPrices.ts`.

Product Prices are available on the client-side at `window.guardian.productPrices`, they are added to the window on the server-side. The previous `ProductPrices` type definition in our Typescript reflected a model where all nested keys in `ProductPrices` were defined as non-optional keys (`CountryGroupName`, `FulfilmentOptions`, `ProductOptions`, `BillingPeriod` and `IsoCurrency`), each value from these types was expected to be present as a key when used as key. This type definition however did not reflect the reality of the Product Prices data, in reality these keys are optional.

A result of our type incorrectly assuming all keys would be present is that our type system does not enforce conditional checks for these keys in our code when interrogating the Product Prices, so we haven’t included them in the code. In reality the data on `window.guardian.productPrices` has always contained the keys our code has referenced and we’ve not seen any errors, but there’s a chance they might not be there. 

Making these keys optional meant having to alter the code to handle the possibility they'd be `undefined`. I have done this in the following files:

**assets/helpers/productPrice/paperProductPrices.ts**
I've added optional chaining to the `price` variable assignment in `getSavingsForFulfilmentOption() `,  this function returns `0` if `price` is `undefined`.

**assets/pages/promotion-terms/weeklyTerms.tsx**
I've dded optional chaining which would make return values `shortTermPrice` and `annualTermPrice` possibly `undefined`. We  handle possible `undefined` values for these variables in the `StandardCountryPrice()` and `SixForSixCountryPrice()` functions, by not rendering anything if either are `undefined`.

**assets/helpers/patrons.ts**
I've added optional chaining to the `productOptions` variable assignment in `getPromotions()`, this function returns `undefined` if `productOptions` is `undefined`

**assets/helpers/productPrice/productPrices.ts**
I've updated `getProductPrice()` to throw an error in the event the `productPrice` lookup fails. `getProductPrice()` is called extensively throughout the codebase so to update it to possibly return `undefined` would've had significant knock on changes.  If thrown the error will enter a `catch` block in `support-frontend/assets/helpers/rendering/render.ts` and the user would see the error page.

**assets/pages/digital-subscription-landing/components/paymentSelection/helpers/paymentSelection.tsx**
I've updated `getProductPrice()` and `getProductOptions()` to throw an error in the event the `productPrice`/`productOptions` lookups fail.  To update it to possibly return `undefined` would've had significant knock on changes.  If thrown the error will enter a `catch` block in `support-frontend/assets/helpers/rendering/render.ts` and the user would see the error page.

#### Additional changes:

- Add test fixture `support-frontend/assets/pages/digital-subscription-landing/components/paymentSelection/__tests__/__fixtures__/productPrices.ts` to be used by `support-frontend/assets/pages/digital-subscription-landing/components/paymentSelection/__tests__/paymentSelectionTests.ts`
- Update `PaperHero()` in `support-frontend/assets/pages/paper-subscription-landing/components/hero/hero.tsx` to handle the possiblity `roundel` is `[]`.


GHaberis:
[Typescript Error Busting: ProductPrices Type](https://trello.com/c/YmGwy0I2/428-typescript-error-busting-productprices-type)

support-frontend/assets/pages/digital-subscription-landing/components/paymentSelection/helpers/paymentSelection.tsx
```diff
@@ -57,7 +65,15 @@ export const getProductPrice = (
...
 	currencyId: IsoCurrency,
-): ProductPrice => productOptions[billingPeriod][currencyId];
+): ProductPrice => {
+	const productPrice = productOptions[billingPeriod]?.[currencyId];
+
+	if (productPrice) {
+		return productPrice;
+	}
+
+	throw new Error('getProductPrice: product price unavailable');
```
rupertbates, line 75: This error message should be `getCurrencySymbol` I think

support-frontend/assets/pages/promotion-terms/weeklyTerms.tsx
```diff
@@ -50,11 +49,10 @@ const orderedCountryGroupNames: NameAndSaving[] = [
 		saving: null,
 	},
 ];
-const shortTermDescription =
-	postIntroductorySixForSixBillingPeriod === Quarterly
-		? 'Quarterly (13 weeks)'
-		: 'Monthly (4 weeks)';
-export default function WeeklyTerms(props: PromotionTermsPropTypes) {
+const shortTermDescription = 'Monthly (4 weeks)';
```
rupertbates, line 52: Could actually inline this now that it doesn't vary any more. Up to you though.

rupertbates (approved):
Couple of nitpicks but this is great, it's a massive improvement to have this area covered by type checks.

support-frontend/assets/pages/promotion-terms/weeklyTerms.tsx
GHaberis: I'm going to keep it as a variable as it's referenced in 2 places in the code.

PR #3741, opened 2022-05-16
Fix PayPal button bug on subs checkouts

TomPretty:

Fixes a bug with the PayPal express button that stopped the button being enabled on checkouts that require a delivery address (all bar digi subs or gift digi sub).

The issue was that the PayPal button was only being reenabled when `checkoutFormIsValid` was `true`. However, that was running the validation for a form without a delivery address which would report errors with the `billingAddress` field. We've now update the logic to check the `product` and run the appropriate validation.

We have to add an explicit `@ts-expect-error` as the types don't quite align. The expected type is `WithDeliveryCheckoutState` but that type is out of date and doesn't match any valid states. We need to clean this up after we've improved the types around address states.

[**Trello Card**](https://trello.com/c/NezRqbNh/425-paypal-not-working-after-validation-error-on-gw-and-newspaper-checkouts)


GHaberis (approved):
Looks good to me 👏

PR #3743, opened 2022-05-16
Mj new checkout cookie

michaelbjacobson:
This PR adds a new post-checkout cookie (`GU_CO_COMPLETE`), which contains the `userType` of the user and the `product` purchased:

```js
{
      userType: String, // "new" || "guest" || "current"
      product: String, // "Contribution" || "DigitalPack" || "GuardianWeekly" || "Paper"
}
```

To handle certain restrictions on cookie values imposed by the Play framework, the JSON string is [URL encoded](https://en.wikipedia.org/wiki/Percent-encoding), and will need to be URL decoded before it can be parsed as JSON.

This new cookie is reliant on the Identity API to retrieve the `userType` for a given email address. If this request fails, eg. due to a timeout, no cookie will be created.

⚠️ **Important: We can't merge this until the CMP has been updated to contain a reference to this Cookie.** ⚠️


michaelbjacobson:
[Add checkout cookie to support sign-in prompts](https://trello.com/c/v18lhNeG/283-add-checkout-cookie-to-support-sign-in-prompts)

TomPretty (approved):
LGTM 👍

PR #3747, opened 2022-05-17
Fix ts errors in `marketingConsent/helpers.ts`

TomPretty:

Fix ts errors in `marketingConsent/helpers.ts`


i-hardy (approved):
👍

PR #3749, opened 2022-05-17
Create a unified product reducer

i-hardy:

This is the result of a spike exploring how we can unify concepts around 'product' state across the subscriptions and contributions checkouts. The result is a sort of 'first pass' that can serve all existing checkouts without requiring extensive refactoring of component code.

The reducer has been added to all checkouts but is not currently connected to anything.

[**Trello Card**](https://trello.com/c/C0BfA2c2)

## Why are you doing this?

The way that we handle product-related state- ie. the specific product, cost, payment frequency and fulfilment of whatever the user is purchasing- is _wildly_ different between the subscriptions and contributions checkouts, and can also vary quite a bit between the different subs checkouts.

I had initially hoped to reduce more of the differences between subs and contribs and develop a single structure for a 'product', but this would require a huge amount of refactoring and is probably not worth either the effort and the risk of introducing more bugs at this stage. The end result re-uses the existing structures we have around product prices and contribution amounts.

However the new reducer should enable us to both
- Reduce the differences in how we handle pre-set contribution amounts and custom amounts
- Have a single source of truth for what subscription product the user has selected and what its original and discounted price are, rather than calculating this in multiple different places, that is responsive to changes in billing period, product, etc
- Have a more solid base from which to move towards a more coherent singular concept of a 'product' in the future, if we want


support-frontend/assets/helpers/redux/checkout/product/state.ts
```diff
@@ -0,0 +1,75 @@
...
+import type { FulfilmentOptions } from 'helpers/productPrice/fulfilmentOptions';
+import type { ProductOptions } from 'helpers/productPrice/productOptions';
+import type {
+	ProductPrice,
+	ProductPrices,
+} from 'helpers/productPrice/productPrices';
+import { getCurrency } from 'helpers/productPrice/productPrices';
+import type { SubscriptionProduct } from 'helpers/productPrice/subscriptions';
+
+export type GuardianProduct =
```
i-hardy, line 18: This is a bit of a fudge for now- in the future it would be good to actually define the specific product types in this file (or in another single place where other product types are defined) and have these types follow a more coherent pattern.

support-frontend/assets/helpers/redux/checkout/product/subscriptionsSideEffects.ts
```diff
@@ -0,0 +1,49 @@
...
+	setProductOption,
+	setProductPrices,
+);
+
+export function addProductSideEffects(
+	startListening: SubscriptionsStartListening,
+): void {
+	startListening({
+		matcher: mayChangeProductPrice,
+		effect(_, listenerApi) {
```
i-hardy, line 28: We currently calculate and re-calculate the actual relevant `ProductPrice` object, and a secondary version with any relevant discount applied, in multiple places in store-connected components. This listener enables us to keep that information in Redux instead, and automatically update it if the user changes billing period (possible in the GW checkout), product type (possible by 'adding the digital subscription' in the paper checkout) and so on.

TomPretty: Yeah I think it makes a lot of sense to bring this all in to the store instead of recalculating many times. 

This might be my lack of familiarity with the new rtk stuff, but it doesn't necessarily seem 'side-effecty'? i.e updating the discounted price should be a direct effect of calling `setProductType`, and therefore could happen in the reducer?

i-hardy: I put it in side effects because it may need to change based multiple variables beyond just changing the product type. Also I think it's simpler to stick to both reducers only affecting the state of a single property at a time as far as possible, and having a single way of saying 'when actions a, b or c happen, also do something else'.

TomPretty: I guess it makes the reducers themselves simpler, but perhaps makes following the overall logic less simple? If all of the state changes are within the reducer, then it's a bit easier to understand everything that happens when an action is dispatched. It's just a little bit harder to understand the effects if you have to also look for all of the side effects that are waiting to trigger on the dispatch.

i-hardy: That's fair, I suppose in this case since the change is within the same reducer slice it's much less 'side-effecty' than having to reach out to another part of state or perform some effect outside of Redux's scope like putting something in localStorage. I'll have a go at reworking this to happen within the reducer and see how it looks.

i-hardy: I've just had a go and ran into the immediate problem that we can't access the current `countryId` from inside the product slice- reducers can only see the piece of state they control, and internationalisation info is managed in the `common` slice. Listeners, however, have access to the whole store as it is at the point a given action happens. We can't determine the relevant product price without knowing what country the user is in.

TomPretty (approved):
Great work! 👍

GHaberis (approved):
Changes look good to me 👍

PR #3752, opened 2022-05-18
Aus election GW promo

tomrf1:
This PR adds a new default GW promo. This means users in AU/NZ will see the promo regardless of whether it's in the url querystring.

This means there will now be monthly _and_ annual promos on the same GW page. The existing terms and conditions in the footer does not support 2 promos. So I've refactored the DigitalFooter (which does support 2 promos) to make it work for GW as well. This works on both the GW product and checkout pages. [This is done in 1 commit.](https://github.com/guardian/support-frontend/commit/4718115c5d53fc64ab82cc6e0224538ea065c6ae)
It also fixes a bug (currently affecting DS gifting in production) where if there are no promos on the page it still displays some (broken) copy:
Screen Shot 2022-05-18 at 09 07 05

### GW footer with both promo links:
Screen Shot 2022-05-18 at 08 59 56

### Cards on /subscribe/weekly:
Screen Shot 2022-05-18 at 09 10 12

### Copy on /subscribe/weekly:
Screen Shot 2022-05-18 at 09 10 51

### Monthly promo on /subscribe:
Screen Shot 2022-05-18 at 09 11 19


support-frontend/assets/components/footerCompliant/FooterWithPromoTerms.tsx
```diff
@@ -0,0 +1,199 @@
+import type { ReactNode } from 'react';
```
tomrf1, line 1: this file is `DigitalFooter.tsx` renamed and refactored. Both DS and GW use this file now. It enables T&Cs for 2 offers on a single page

KaliedaRik (approved):
Nice to see DigiSubs and GW pages sharing the same T&C link functionality

PR #3760, opened 2022-05-19
Remove reference to gu.com

mchv:

We should not use that domain anymore.

### DIFF
--- a/support-frontend/assets/helpers/__tests__/urlTest.ts
+++ b/support-frontend/assets/helpers/__tests__/urlTest.ts
@@ -10,7 +10,7 @@ const { jsdom } = global;
 describe('url', () => {
 	describe('addQueryParamsToURL', () => {
 		it('should add a query param to an absolute URL', () => {
-			const startingUrl = 'https://gu.com/index?hello=world';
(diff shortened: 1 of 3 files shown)


i-hardy (approved):
👍

support-frontend/assets/helpers/__tests__/urlTest.ts
```diff
@@ -10,15 +10,15 @@ const { jsdom } = global;
...
+			const startingUrl = 'https://www.theguardian.com/index?hello=world';
 			const params = {
 				spam: 'eggs',
 			};
 			const expectedUrl = `${startingUrl}&spam=eggs`;
 			expect(addQueryParamsToURL(startingUrl, params)).toEqual(expectedUrl);
 		});
 		it('should add a query param to an absolute URL [undefined case]', () => {
-			const startingUrl = 'https://gu.com/index?hello=world';
+			const startingUrl = 'https:///www.theguardian.com/index?hello=world';
```
johnduffell, line 21: why do we need 3 slashes now @mchv ?

mchv: Well spotted, this is a mistake

PR #3767, opened 2022-05-20
Default promos from config

tomrf1:
Setting up a default Guardian Weekly (GW) promo requires dev work.
A default promo means users will see it on the support site regardless of whether the url includes the promoCode.
This PR introduces a way to configure default promos, starting with GW, from a json file in S3. In the longer term we can build a tool for Marketers to configure this themselves.
For each product, this file contains an ordered list of promo codes that should be considered defaults. Targeting of promos by region/rate-plan is still done per promo.

1. Adds a `DefaultPromotionsService` class, which polls the default promos data in S3. This lives inside the `PriceSummaryService`, where it is now used.
2. Moves the `PriceSummaryService` class from support-services to support-frontend. This is because both it and the new `DefaultPromotionsService` are only used by support-frontend. support-services is a dependency of support-workers, so we should aim to keep it small (in particular, I wanted to use akka for scheduling, which support-workers does not need)


cdk/lib/__snapshots__/frontend.test.ts.snap
```diff
@@ -974,6 +974,27 @@ Object {
       },
       "Type": "AWS::IAM::Policy",
     },
+    "PromoToolBucketE9531506": Object {
```
tomrf1, line 977: autogenerated

support-frontend/app/controllers/PricesController.scala
```diff
@@ -91,7 +91,7 @@ class PricesController(
   ): CountryGroupPriceData = {
     val guardianWeeklyProductPrices = priceSummaryServiceProvider
       .forUser(false)
-      .getPrices(GuardianWeekly, DefaultPromotions.GuardianWeekly.NonGift.all, Direct)
+      .getPrices(GuardianWeekly, Nil, Direct)
```
tomrf1, line 94: any default promos will now be found by the `PriceSummaryService`

support-frontend/app/controllers/WeeklySubscriptionController.scala
```diff
@@ -43,13 +41,11 @@ class WeeklySubscriptionController(
...
     implicit val settings: AllSettings = settingsProvider.getAllSettings()
     val canonicalLink = Some(buildCanonicalWeeklySubscriptionLink("uk", orderIsAGift))
-    val defaultPromo = List(ausElection)
 
     val queryPromos =
       request.queryString
-        .get("promoCode")
-        .map(_.toList)
-        .getOrElse(defaultPromo)
+        .getOrElse("promoCode", Nil)
```
tomrf1, line 47: we no longer need default promos here - the `PriceSummaryService` will get them

support-frontend/build.sbt
```diff
@@ -7,7 +7,7 @@ packageSummary := "Support Play APP"
 
 SeleniumTest / testOptions := Seq(Tests.Filter(seleniumTestFilter))
 
-Test / testOptions := Seq(Tests.Filter(unitTestFilter))
+Test / testOptions ++= Seq(Tests.Filter(unitTestFilter))
```
tomrf1, line 10: `:=` overwrites the integration test settings

support-frontend/app/controllers/WeeklySubscriptionController.scala
```diff
@@ -43,14 +41,14 @@ class WeeklySubscriptionController(
...
 
     val queryPromos =
       request.queryString
-        .get("promoCode")
-        .map(_.toList)
-        .getOrElse(defaultPromo)
-    val maybePromotionCopy = landingCopyProvider.promotionCopy(queryPromos, GuardianWeekly, countryCode)
+        .getOrElse("promoCode", Nil)
+        .toList
+    val defaultPromos = priceSummaryServiceProvider.forUser(false).getDefaultPromoCodes(GuardianWeekly)
```
johnduffell, line 49: hardly worth a commit for, but thought I'd mention it anyway
```suggestion
    val defaultPromos = priceSummaryServiceProvider.forUser(isTestUser = false).getDefaultPromoCodes(GuardianWeekly)
```

support-frontend/app/controllers/SubscriptionsController.scala
```diff
@@ -69,13 +68,14 @@ class SubscriptionsController(
     } else
       Map.empty
 
-    val fulfilmentOptions = if (countryGroup == CountryGroup.RestOfTheWorld) RestOfWorld else Domestic
-    val guardianWeeklyPromos = List(ausElection)
+    val guardianWeeklyFulfilmentOptions = if (countryGroup == CountryGroup.RestOfTheWorld) RestOfWorld else Domestic
+    // The client currently only supports monthly
+    val guardianWeeklyBillingPeriod = postIntroductorySixForSixBillingPeriod
```
rupertbates, line 73: Is this just for clarity? You could rename the import instead:
```scala
import com.gu.support.catalog.GuardianWeekly.{postIntroductorySixForSixBillingPeriod => guardianWeeklyBillingPeriod}
```

tomrf1: It's mainly to tie the fixed billing period (for GW) to the new comment. The fact that it doesn't support annual pricing/offers on the /support page is important to note (I think)

support-frontend/app/controllers/WeeklySubscriptionController.scala
tomrf1: 👍

support-frontend/app/services/pricing/DefaultPromotionsService.scala
```diff
@@ -0,0 +1,69 @@
...
+
+  private def update(): Try[Unit] = fetch().map(newPromos => defaultPromos.set(newPromos))
+
+  private def startPollingS3(): Unit =
+    system.scheduler.scheduleWithFixedDelay(0.minutes, 1.minute) { () =>
+      update()
+        .fold(
+          err => logger.error(s"Error fetching default promos from $s3Uri: ${err.getMessage}", err),
+          _ => (),
+        )
```
johnduffell, line 59: slightly clearer as a pattern match?
```suggestion
      update() match {
        case Failure(err) => logger.error(s"Error fetching default promos from $s3Uri: ${err.getMessage}", err),
        case Success => ()
      }
```

johnduffell: also, do we want an alarm for this?  I suspect it would get reported if it's failing, and isn't a big incident, but I thought I'd mention in case it's missed.

johnduffell (approved):
great, looks good and only a couple of minor comments.  I assume the bucket already exists and you are just adding the new file to it.

👍

support-frontend/app/services/pricing/DefaultPromotionsService.scala
```diff
@@ -0,0 +1,69 @@
...
+  private def fetch(): Try[DefaultPromotions] =
+    for {
+      raw <- client.fetchAsString(s3Uri)
+      defaultPromos <- decode[DefaultPromotions](raw).toTry
+    } yield defaultPromos
+
+  private def update(): Try[Unit] = fetch().map(newPromos => defaultPromos.set(newPromos))
+
+  private def startPollingS3(): Unit =
+    system.scheduler.scheduleWithFixedDelay(0.minutes, 1.minute) { () =>
```
rupertbates, line 54: I wonder if this really needs to poll every minute, these change very infrequently. I feel like we could probably push it up to 10 mins or more and just let make it clear in the tooling that it will take a while?

support-frontend/build.sbt
tomrf1: it took me _ages_!

support-frontend/app/services/pricing/DefaultPromotionsService.scala
johnduffell: interesting thought - do you think it may be not necessary on cost reasons, or reliability, clarity? or something else?

support-frontend/app/services/pricing/PriceSummaryService.scala
```diff
@@ -12,22 +12,28 @@ import org.joda.time.Months
...
+) extends TouchpointService {
   private type GroupedPriceList = Map[(FulfilmentOptions, ProductOptions, BillingPeriod), Map[Currency, PriceSummary]]
 
   def getPrices[T <: Product](
       product: T,
       promoCodes: List[PromoCode],
       readerType: ReaderType = Direct,
   ): ProductPrices = {
-    val promotions = promotionService.findPromotions(promoCodes)
+    val defaultPromos = getDefaultPromoCodes(product)
```
rupertbates, line 27:
```suggestion
    val defaultPromoCodes = getDefaultPromoCodes(product)
```
Just to make explicit the difference between a promotion and a promoCode. This probably applies elsewhere as well

support-frontend/app/services/pricing/DefaultPromotionsService.scala
```diff
@@ -0,0 +1,69 @@
...
+import com.typesafe.scalalogging.LazyLogging
+import services.pricing.DefaultPromotionsService.DefaultPromotions
+
+import java.util.concurrent.atomic.AtomicReference
+import scala.concurrent.ExecutionContext
+import scala.concurrent.duration.DurationInt
+import scala.util.Try
+
+trait DefaultPromotionsService {
+  def getPromos(product: Product): List[String]
```
rupertbates, line 20:
```suggestion
  def getPromoCodes(product: Product): List[String]
```
To make explicit that we are talking about promo *codes* here not the full promotion object

support-frontend/app/services/pricing/DefaultPromotionsService.scala
rupertbates: I guess just on the principal of not doing unnecessary work. It feels wasteful to be polling every minute for something which can remain unchanged for months at a time.

support-frontend/app/services/pricing/DefaultPromotionsService.scala
tomrf1: Good point about an alarm. I guess that would mean a new cloudwatch metric + alarm?
I notice there's a `AwsCloudWatchMetricSetup` object in support-services for this type of thing, I'll look at adding one

support-frontend/app/services/pricing/DefaultPromotionsService.scala
```diff
@@ -0,0 +1,68 @@
...
+import scala.concurrent.ExecutionContext
+import scala.concurrent.duration.DurationInt
+import scala.util.{Failure, Success, Try}
+
+trait DefaultPromotionsService {
+  def getPromoCodes(product: Product): List[String]
+}
+
+object DefaultPromotionsService {
+  case class DefaultPromotions(guardianWeekly: List[String])
```
rupertbates, line 24: Is the plan that this object will be changed to add additional products going forward? Would it be worth adding them all at the same time?

tomrf1: Yes, more products can be added here later. I think it's best to only define the product that's currently supported/required.
Adding another product will involve some more backend changes to the relevant controllers and testing (which I don't want to do now)

support-frontend/app/services/pricing/PriceSummaryService.scala
```diff
@@ -12,22 +12,28 @@ import org.joda.time.Months
 
 import scala.math.BigDecimal.RoundingMode
 
-class PriceSummaryService(promotionService: PromotionService, catalogService: CatalogService)
-    extends TouchpointService {
+class PriceSummaryService(
+    promotionService: PromotionService,
+    defaultPromotionsService: DefaultPromotionsService,
```
rupertbates, line 17: @johnduffell what are your thoughts about passing the `defaultPromotionsService` into this object as a dependency rather than just adding the defaults to the codes that are passed in to the function. 

I can see why it's been done it this way but it is a bit weird that it requires it for a single method call and the fact that it then exposes the `defaultPromotionsService.getPromoCodes` method for external code feels like a bit of a code smell?

support-frontend/app/services/pricing/DefaultPromotionsService.scala
tomrf1: I've pushed a commit with a new alarm, `DefaultPromotionsLoadingFailureAlarm`

support-frontend/app/services/pricing/DefaultPromotionsService.scala
```diff
@@ -0,0 +1,68 @@
...
+import com.gu.support.catalog.{GuardianWeekly, Product}
+import com.typesafe.scalalogging.LazyLogging
+import services.pricing.DefaultPromotionsService.DefaultPromotions
+
+import java.util.concurrent.atomic.AtomicReference
+import scala.concurrent.ExecutionContext
+import scala.concurrent.duration.DurationInt
+import scala.util.{Failure, Success, Try}
+
+trait DefaultPromotionsService {
```
rupertbates, line 21: Bit of a nitpick but I think I'd go singular with the class name so `DefaultPromotionService` I think it is more standard for the code base. Feel free to disagree though!

support-frontend/app/services/pricing/PriceSummaryService.scala
tomrf1: I see what you mean, and it did feel a bit wrong exposing `getPromoCodes`. It's used in one external place, for getting promo copy. But it seems preferable for `PriceSummaryService` to add the default promos, rather than every use of `getPrices` having to know about them

cdk/lib/frontend.ts
```diff
@@ -250,6 +250,29 @@ export class Frontend extends GuStack {
...
+      threshold: 1,
+      evaluationPeriods: 1,
+      comparisonOperator: ComparisonOperator.GREATER_THAN_OR_EQUAL_TO_THRESHOLD,
+      metric: new Metric({
+        metricName: "DefaultPromotionsLoadingFailure",
+        namespace: "support-frontend",
+        dimensionsMap: {
+          Environment: "PROD",
+        },
+        statistic: "Average",
```
johnduffell, line 270: makes no difference when the threshold is 1, but in case we change the threshold makes sense to start off as Sum
```suggestion
        statistic: "Sum",
```

cdk/lib/__snapshots__/frontend.test.ts.snap
```diff
@@ -261,6 +261,45 @@ Object {
       },
       "Type": "AWS::IAM::Policy",
     },
+    "DefaultPromotionsLoadingFailureAlarm31718167": Object {
```
johnduffell, line 264: why are we checking in generated code? is it not possible to generate as part of the build?

tomrf1: I think this is standard with cdk

support-frontend/app/services/pricing/DefaultPromotionsService.scala
johnduffell: looks good 👍

support-frontend/app/services/pricing/DefaultPromotionsService.scala
johnduffell: fair enough, as long we keep the time taken to update smaller than the time for a bug report to be made, we will be ok!

cdk/lib/frontend.ts
tomrf1: 👍

support-frontend/app/services/pricing/PriceSummaryService.scala
johnduffell: see what you mean - we are effectively partially applying all of the methods in this service with the whole DefaultPromotionsService, most of which are not relevant.  So it does make unit testing and reuse more tricky.

Structuring code is something it is easy to go round in circles with - we would need to follow how the code actually depends on each other to decide. 

 personally I'd be tempted to split the PriceSummaryService in half if `getPrices` is called from a lot of places that the other methods are not.  But if most places need most of the methods, it gets hard to justify.  But I do generally tend towards very small interfaces with minimal dependencies, which has its own limitations.
One precendent ( naming from memory) we created the ZuoraGiftingService and the ZuoraService separately to avoid creating unnecessary dependencies.

cdk/lib/__snapshots__/frontend.test.ts.snap
johnduffell: fair enough I have raised it on devex channel for more opinions

PR #3768, opened 2022-05-20
Fix TS errors in canShow and withError hocs

i-hardy:

This fixes the Typescript errors in the `canShow` and `withError` higher-order components. I've also slightly refactored the types because using `Props` for both the generics in the types and the generics in the actual implementation was giving me a headache!


marialani (approved):
Looks good!

PR #3770, opened 2022-05-20
Fix the ts errrors in `StripeCardFormContainer`

TomPretty:

This required updating to the latest version (1.8.0) of the `react-stripe-js` package as it required a type update from the package itself to fix. The issue is that react no longer adds the `children` prop to function components, which broke the types coming from `react-stripe-js`. They fixed it in 1.7.2, but the change log since then has very little in it so it seemed to make sense to upgrade all the way to the latest version.

This upgrade also fixed the errors in:

- `assets/components/StripePaymentRequestButton/StripePaymentRequestButtonContainer.tsx`
- `assets/components/subscriptionCheckouts/stripeForm/stripeProviderForCountry.tsx`


support-frontend/assets/pages/contributions-landing/components/StripeCardForm/StripeCardFormContainer.tsx
```diff
@@ -50,10 +51,7 @@ function StripeCardFormContainer(props: PropTypes) {
 						stripe={stripeObjects[stripeAccount]}
 						options={elementsOptions}
 					>
-						<StripeCardForm
-							stripeKey={stripeKey}
-							isTestUser={props.isTestUser}
```
TomPretty, line 55: `StripeCardForm` is already getting this from redux

GHaberis (approved):
Looks good 😎

PR #3771, opened 2022-05-20
Fix ts errors in AmazonPayLoginButton

TomPretty:

Fix ts errors in AmazonPayLoginButton


GHaberis (approved):
Nice 👏

PR #3772, opened 2022-05-20
Fix Typescript errors in address and postcode lookup components

i-hardy:

This fixes TS errors in the `AddressFields` and `PostcodeFinder` components, and refactors both from class to function components. It also fixes a long-standing bug where the value of the `checkoutPostcodeLookup` switch was passed to the front end as a string, but the front end treated it as a boolean- this is why attempting postcode lookups locally always causes an error to show up on the form.

This is very much a first pass as these components have a _lot_ of type complexity due to the way we connect them to the store- they can be simplified a great deal in the future if we have them just accept props, and then determine whether they're setting the billing or delivery address higher up the component tree.

Ideally this needs testing in an environment with postcode lookup enabled before merging.

[**Trello Card**](https://trello.com/c/4VbxwAxW)


i-hardy:
[Fix TS errors in address fields and postcode finder components](https://trello.com/c/4VbxwAxW/456-fix-ts-errors-in-address-fields-and-postcode-finder-components)

michaelbjacobson (approved):
Nice! 👍🏼

PR #3783, opened 2022-05-24
Fix errors in OrderSummary component tests and create generic test render function

i-hardy:

This fixes the TS errors in the Paper checkout order summary component tests, and extracts a generic `renderWithStore` function as a test utility, to replace the numerous different versions of it we have across other component tests. The function type signature doesn't work amazingly well at the moment, but this is hopefully something we can improve once we have a more coherent store type structure in the future.

[**Trello Card**](https://trello.com/c/ACGAULV1)


michaelbjacobson (approved):
Nice! 🚀

PR #3789, opened 2022-05-25
Quantum Metric: Add "blocklist" data attribute to checkouts

GHaberis:

Quantum Metric need to be able to identify fields/elements that contain PII (Personally Identifiable Information) so they can add them add to their masking blocklist. Data in the masking blocklist won't leave the users browser. We've agreed with QM that this data attribute will be used to identify these fields: `data-qm-masking="blocklist"`.

This PR updates all the fields that have already been identified as needing to be on the blocklist to include the data-attribute, I've cross-checked the rendered forms locally to confirm the attribute is present. 

I've also for one component (`<SepaForm>`) written a UI test with `@testing-library/react` to validate the presence of the attribute on the component's rendered fields. Having such a test should help guard against accidentally removing the attribute. Ideally I'd like to do this for all components that have been updated in this PR, however that's quite a large amount of work, `<SepaForm>` is not connected to Redux so it's relatively easy to write a test for it, however UI tests for components connected to the Redux store require quite a bit of boilerplate. I intend to put together tickets to address this in follow-up PRs.


support-frontend/assets/components/subscriptionCheckouts/address/addressFields.tsx
```diff
@@ -105,7 +106,7 @@ function AddressFields({ scope, traverseState, ...props }: PropTypes) {
 			</Select>
 			{props.country === 'GB' ? (
 				<PostcodeFinder
-					id={`${scope}-postcode`}
+					id={`${scope}-postcode-lookup`}
```
GHaberis, line 109: This fields ID wasn't unique, this changes makes it unique.

GHaberis:
[Quantum Metric: Add masking selector to checkout inputs](https://trello.com/c/6baNdnjc/457-quantum-metric-add-masking-selector-to-checkout-inputs)

PR #3794, opened 2022-05-26
Squash TS errors

TomPretty:

Squashes the TS errors in:
- assets/helpers/__tests__/contributionsTests.ts
- assets/helpers/__tests__/urlTest.ts
- assets/helpers/__tests__/checkoutsTest.ts
- assets/helpers/internationalisation/__tests__/countryGroupTest.ts
- assets/helpers/internationalisation/__tests__/countryTest.ts


support-frontend/assets/helpers/__tests__/contributionsTests.ts
```diff
@@ -13,52 +14,63 @@ import {
...
-				[5, 'ANNUAL', AUDCountries],
-				[0.5, 'ONE_OFF', EURCountries],
-			].forEach((a) => expect(validateContribution(...a)).toBe('TooLittle'));
+			expect(validateContribution(1, 'MONTHLY', GBPCountries)).toBe(
+				'TooLittle',
+			);
+			expect(validateContribution(5, 'ANNUAL', AUDCountries)).toBe('TooLittle');
+			expect(validateContribution(0.5, 'ONE_OFF', EURCountries)).toBe(
+				'TooLittle',
+			);
```
TomPretty, line 29: I was struggling to get typescript to be happy with the expansion, so just unrolled the loop

support-frontend/assets/helpers/__tests__/urlTest.ts
```diff
@@ -1,12 +1,18 @@
...
 	getAbsoluteURL,
 	getAllQueryParams,
 	getAllQueryParamsWithExclusions,
 } from '../urls/url';
+
 // ----- Tests ----- //
-const { jsdom } = global;
+
+// @ts-expect-error -- This is added to the global scope by the test setup
+const jsdom = global.jsdom as JSDOM;
```
TomPretty, line 14: I think it's a bit of a smell that so many of our tests directly interact with jsdom. Really here there should probably be a couple of methods to get the query string and set the query string which interact with the dom, every thing else (i.e parsing and manipulating) could just be pure functions that are much easier to test

TomPretty: (everything below this change is just adding spaces)

support-frontend/assets/helpers/__tests__/checkoutsTest.ts
```diff
@@ -11,60 +13,88 @@ jest.mock('helpers/globalsAndSwitches/globals', () => ({
 	isSwitchOn: jest.fn(),
 }));
 
-const mock = (mockFn: any) => mockFn;
+const mock = (mockFn: unknown) => mockFn as jest.Mock;
+
```
TomPretty, line 17: I'm not quite sure why we did this in the first place - effectively this function is just a typecast.

I've fixed it up to remove the errors, but it would also be possible to just replace all instances with:

```ts
(isSwitchOn as jest.Mock)
```

TomPretty: Definitely quite gross to be having to mock like this in the first place though (we need more dep injection!)

i-hardy (approved):
Looks great! And I completely agree with your comments about how we can make this stuff more testable and less reliant on mocks in future 👍

PR #3795, opened 2022-05-26
Added the ability for the /prices endpoint to include details about the promotions which determine the returned prices

paulbrown1982:

Added the ability for the `/prices` endpoint endpoint to include, when the query parameter `?include-summary` is provided, details about the promotions which determine the returned prices.

Also added currency as a price should never be referenced without its currency being alongside.

## Why are you doing this?
This is to enable the currently displayed offers on the Digital Subscription and Guardian Weekly checkouts to be shown elsewhere, for example a cancellation-save journey.

## Is this an AB test?
- [ ] Yes
- [x] No

## Screenshots
N/A

### DIFF
--- a/support-frontend/app/controllers/PricesController.scala
+++ b/support-frontend/app/controllers/PricesController.scala
@@ -22,7 +22,11 @@ import com.gu.support.workers.{Annual, Monthly}
 import com.gu.support.zuora.api.ReaderType.Direct
 
 object PricesController {
-  case class RatePlanPriceData(price: String)
+  case class RatePlanPriceData(
+      price: String, // the offered price on the site right now
+      currency: String, // never quote a price without a currency, even if grouped
+      priceSummary: Option[PriceSummary],
+  )
 
   case class ProductPriceData(
       Monthly: RatePlanPriceData,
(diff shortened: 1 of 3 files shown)


support-frontend/app/controllers/PricesController.scala
```diff
@@ -47,19 +51,26 @@ object PricesController {
...
+    val price = topPromotion
       .flatMap(_.discountedPrice)
       .getOrElse(priceSummary.price)
 
-    RatePlanPriceData(price.toString)
+    RatePlanPriceData(
+      price = price.toString,
+      currency = priceSummary.currency.identifier,
+      priceSummary =
+        Option.when(includeSummary)(priceSummary.copy(promotions = topPromotion.map(_.copy(landingPage = None)).toList)),
```
tomrf1, line 64: `Option.when` is new to me!

support-frontend/conf/routes
```diff
@@ -170,6 +170,7 @@ GET /p/:promoCode                                                 controllers.Pr
 
 # ----- Prices API ----- #
 GET /prices                                                       controllers.PricesController.getPrices()
+GET /pricesWithSummary                                            controllers.PricesController.getPricesWithSummary()
```
tomrf1, line 173: since it's almost the same it could even be `/prices?show-summary`

paulbrown1982: That was my original plan, however upon trying I assumed it wasn't possible/advisable to read query parameters with a `CachedAction`, and I didn't want to change it to a `NoCacheAction`?

tomrf1: Ah right. Perhaps we should change this to a regular `Action`? Currently [SDC polls](https://github.com/guardian/support-dotcom-components/blob/main/packages/server/src/productPrices.ts#L15) the endpoint every minute (for each instance) which isn't much. I guess it depends what it's going to be used for long term

paulbrown1982: Ok I'll do that.

tomrf1 (approved):
looks good 👍

PR #3796, opened 2022-05-26
Fix TS issues in polyfill files

i-hardy:

This fixes TS issues in our two custom polyfill files. The `ResizeObserver` polyfill has been 'fixed' by using the implementation from polyfill.io instead, and switching over the header to use `ResizeObserver` directly, meaning we can delete this file. I've tested the new version on IE11 and it's all working as expected.


TomPretty (approved):
Nice one! Great improvement 👍

PR #3797, opened 2022-05-26
Fix TS errors

TomPretty:

Fixes the typescript errors in:

- support-frontend/assets/__mocks__/imageMock.ts
- support-frontend/assets/__mocks__/stripeMock.tsx
- support-frontend/assets/__mocks__/styleMock.ts
- support-frontend/assets/images/favicons.ts

For some reason, only one of these errors would ever be reported at one time. Fixing one, another would then appear in the `typescript-errors.json` file. It's not clear to me why they couldn't all have been there simultaneously. 

Typescript was upset with all of these files as they didn't have exports which breaks the `isolatedModules` flag.


support-frontend/assets/__mocks__/stripeMock.tsx
```diff
@@ -67,3 +67,5 @@ jest.mock('@stripe/react-stripe-js', () => {
 		useElements: mockElements,
 	};
 });
+
+export {};
```
TomPretty, line 71: This file is just imported for its side effects, so doesn't actually need to export anything

support-frontend/assets/images/favicons.ts
```diff
@@ -9,3 +9,5 @@ require('./favicons/120x120.png');
 require('./favicons/144x144.png');
 
 require('./favicons/152x152.png');
+
+export {};
```
TomPretty, line 13: This file doesn't have any exports as it's an entrypoint specified in `webpack.entryPoints.js`. Adding this empty export makes typescript happy and hasn't affected the build

marialani (approved):
Looks good!

PR #3798, opened 2022-05-27
Fix ts errors in remaining test files

TomPretty:

Fix ts errors in remaining test files


support-frontend/assets/components/directDebit/__tests__/directDebitReducerTest.ts
```diff
@@ -13,16 +14,24 @@ import {
...
 describe('direct debit reducer tests', () => {
+	const nonExistentAction = {
+		type: 'DOES_NOT_EXIST',
+	} as unknown as Action;
+
 	it('should return the initial state', () => {
-		expect(reducer(undefined, {})).toMatchSnapshot();
+		expect(reducer(undefined, nonExistentAction)).toMatchSnapshot();
 	});
+
```
TomPretty, line 28: I'm not convinced that this test is that useful, given we're forcing the typesystem to let us use and invalid action so that the switch statement falls back to the default case. For now I've kept it and followed the precedent [here](https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/helpers/user/__tests__/userReducerTest.ts#L6) for typing the invalid action

support-frontend/assets/helpers/subscriptionsForms/__tests__/subscriptionCheckoutReducerTest.ts
```diff
@@ -1,28 +1,50 @@
...
+		const action = setStage(stage, 'DigitalPack', undefined);
+
+		const newState = createCheckoutReducer(
+			'GB',
+			'DigitalPack',
+			'Annual',
+			null,
+			null,
+			null,
+		)(undefined, action);
```
TomPretty, line 27: Not quite sure how we ended up with these functions being called with so many missing arguments 😂

support-frontend/assets/components/directDebit/__tests__/directDebitReducerTest.ts
i-hardy: We can probably get rid of most of these reducer tests in future once we've moved over to Redux Toolkit- at the moment they're right on the edge of testing 'does Redux do its job properly', which IMO is not our responsibility to write unit tests for!

i-hardy (approved):
👏

PR #3801, opened 2022-05-27
Fix CSRF reducer and subs payment method selector

i-hardy:

This fixes TS errors in the CSRF reducer and subscriptions payment method selector component


support-frontend/assets/components/subscriptionCheckouts/paymentMethodSelector.tsx
```diff
@@ -6,68 +6,99 @@ import {
...
-const paymentMethodIcons: Record<PaymentMethod, Node> = {
+const paymentMethodIcons: Record<PaymentMethod, ReactNode> = {
 	Stripe: <SvgCreditCard />,
 	PayPal: <SvgPayPal />,
 	DirectDebit: <SvgDirectDebit />,
+	Sepa: '',
+	ExistingCard: '',
+	ExistingDirectDebit: '',
+	AmazonPay: '',
+	None: '',
```
i-hardy, line 71: I've added the other `PaymentMethod` fields to these types as defining a more limited type for 'payment methods we only use on subscriptions' caused problems with the way this component is used. In practice we only ever pass in the first three as allowed payment methods.

marialani: If we narrow the return type for the [supportedPaymentMethods function](https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/helpers/subscriptionsForms/countryPaymentMethods.ts#L9) I think we can avoid having to add in these extra fields

i-hardy: You're right, I think that would be better 👍

PR #3803, opened 2022-05-27
Fix ts errors in `checkoutForm`

TomPretty:

I think we should revisit this file as we could probably improve/remove a lot here. To keep things simple, I just updated the type signatures and left the code (mostly) untouched. 

I had to enable the [`downlevelIteration`](https://www.typescriptlang.org/tsconfig#downlevelIteration) flag to make typescript happy with the `...` on the `HTMLCollection` type. This effectively tells typescript to transpile away the syntax.

An alternative would have been e.g `Array.from`, but then we might want to polyfill that so I thought this would be the simplest approach.


support-frontend/assets/helpers/checkoutForm/checkoutForm.ts
```diff
@@ -62,15 +62,17 @@ export const invalidReason = (form: Record<string, any> | null): string => {
 
 		return invalidReasonString;
 	} catch (e) {
-		return e;
+		return 'unknown';
```
TomPretty, line 65: I don't think we need this try/catch anymore (or maybe even this function at all?). But for now we just need to ensure it returns a string from the catch block

TomPretty:
> Why are we rooting around in the DOM like this 😱 definitely worth a revisit in future. From searching in the codebase, though, it seems like the only functions we actually use from this file are `getForm` and `invalidReason`. Could we delete everything else?
> 
> Edit: Also I'm totally happy with enabling the `downLevelIteration` flag, it seems useful, but we're already polyfilling `Array.from` so there's no problem with using that either- it was part of the ES6 spec and we've got polyfills enabled for all of ES6:
> 
> https://github.com/guardian/support-frontend/blob/bc86c9362365cac4d9bdabbc515421cb57b7f08c/support-frontend/app/views/main.scala.html#L150

Nice I'll take a look again on Monday - in terms of making the smallest possible change I'll just use `Array.from` if it's already pollyfilled. I'll also take a look at what I can delete...

i-hardy (approved):
👍

PR #3807, opened 2022-05-31
Fix logging and increase lambda timeout

rupertbates:
Fix the logging and increase the timeout for the stripe-patrons-data lambda


TomPretty (approved):
👍

PR #3809, opened 2022-05-31
Remove the revenue distribution step from support-workers

rupertbates:
Now that we have the [revenue recogniser lambda](https://github.com/guardian/support-service-lambdas/tree/main/handlers/revenue-recogniser-job) we no longer need to distribute revenue in support-workers. This also means that we no longer need the alarm which is triggered when this fails, which it does regularly due to data consistency issues with Zuora.

[**Trello Card**](https://trello.com/c/AFJ8Jojz/437-spike-can-we-remove-data-corruption-alarm)


TomPretty (approved):
Nice!

PR #3812, opened 2022-06-01
Turn off ts monitor action on dependabot prs

TomPretty:

This is the last broken action on dependabot PRs! I don't think we need to run it given it's just for small dependency updates (also give that we'll probably be dropping it very soon!)


marialani (approved):
👍

PR #3813, opened 2022-06-01
Tests for PersonalDetails Component

GHaberis:

This PR introduces some tests in [personalDetails.test.tsx](https://github.com/guardian/support-frontend/compare/gh-personalDetailsTests?expand=1#diff-1d7f030c19d80e1f6cf765c764cf91c99b35575e58a2d796e1b76430e9970a48) for the `<PersonalDetails>` component in [personalDetails.tsx](https://github.com/guardian/support-frontend/compare/gh-personalDetailsTests?expand=1#diff-e2d0f8f326dffda6f53dd3cb2ad1a550c20eb43467aa586ff0aab02585ab52d3).

The primary objective was to write some tests to validate the Quantum Metric data-attribute (`data-qm-masking="blocklist"`) was present on those elements that contain PII. I picked this component as it's a relatively easy component to write tests for as it's not connected to Redux, all it's props are passed to it. I ended up writing a few other tests for the component in the process, just to learn a little more about `@testing-library/react`.


support-frontend/yarn.lock
```diff
@@ -80,6 +80,15 @@
     jsesc "^2.5.1"
     source-map "^0.5.0"
 
+"@babel/generator@^7.18.2":
```
i-hardy, line 83: I'm not sure why there's so many changes to the lockfile as you don't seem to have added any new dependencies. Probably best to revert this file?

PR #3814, opened 2022-06-01
Fix easy eslint errors

TomPretty:

I've gone through and fixed up all the easy eslint errors. We're now left with the following breakdown:

**errors**

- @typescript-eslint/no-unsafe-assignment: 4
- @typescript-eslint/no-unsafe-return: 1
- @typescript-eslint/no-unsafe-member-access: 5
- @typescript-eslint/no-unnecessary-condition: 6

**warnings**

- @typescript-eslint/no-explicit-any: 28
- @typescript-eslint/explicit-module-boundary-types: 117

I think it'd be worth a mini push to get all of the errors out of the way so that we can turn on linting checks in our ci. We can then pick up the warnings as and when we touch the specific files (they're mostly pretty easy as typescript infers the type that we want anyway).

### DIFF
--- a/support-frontend/.eslintrc.js
+++ b/support-frontend/.eslintrc.js
@@ -20,6 +20,7 @@ module.exports = {
 					},
 				],
 				'@emotion/pkg-renaming': 'error',
+        '@typescript-eslint/no-unused-vars': ["error", { "ignoreRestSiblings": true }],
 			},
 		},
 	],
--- a/support-frontend/assets/helpers/internationalisation/country.ts
+++ b/support-frontend/assets/helpers/internationalisation/country.ts
@@ -1,9 +1,9 @@
-import type { $Keys } from 'utility-types';
 // ----- Imports ----- //
+
+import type { $Keys } from 'utility-types';
 import * as cookie from 'helpers/storage/cookie';
 import type { Option } from 'helpers/types/option';
 import { getQueryParameter } from 'helpers/urls/url';
-import 'helpers/types/option';
 import {
 	AUDCountries,
 	Canada,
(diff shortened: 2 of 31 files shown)


support-frontend/.eslintrc.js
```diff
@@ -20,6 +20,7 @@ module.exports = {
 					},
 				],
 				'@emotion/pkg-renaming': 'error',
+        '@typescript-eslint/no-unused-vars': ["error", { "ignoreRestSiblings": true }],
```
TomPretty, line 23: this stops eslint complaining with unused vars coming from destructing with the rest operator e.g

if we have

```
const { dontWant, ...doWant } = foo
```

eslint wont complain that we're not using `dontWant`

support-frontend/assets/helpers/internationalisation/country.ts
```diff
@@ -1,9 +1,9 @@
-import type { $Keys } from 'utility-types';
 // ----- Imports ----- //
+
+import type { $Keys } from 'utility-types';
```
i-hardy, line 3: We could actually replace wherever this is being used in the file with Typescript's built-in `keyof`- this dependency just helped with moving over from Flow syntax.

TomPretty: Just took a look and I think I might make a health card to fix this one up as the types are actually useless atm!

Currently we have an object of type `Record<string, string>` and then we're using `$Keys<typeof object>` which is just `string`. 

It's actually trying to be a union of US states. I reckon we could improve quite a lot here, but probably worth a separate ticket

PR #3816, opened 2022-06-06
Pass Ophan page view id and referer URL when creating an Identity guest account

i-hardy:

This adds the Ophan page view ID and referrer URL as query parameters when creating a new Identity account for a guest user during the checkout process. We already pass the page view ID as part of the acquisition data, and as we submit form data via AJAX we can access the 'last visited URL' as the `Referer` header on the request, which is automatically added by the browser.

[**Trello Card**](https://trello.com/c/pgyDZqdO)

## Why are you doing this?

This is due to a request from the Identity team to help them better track where account registrations originat


TomPretty (approved):
LGTM 👍

PR #3817, opened 2022-06-06
Strip new line characters from response body logging

rupertbates:
Http responses from the Stripe billing API have lots of newlines in them which when logged break out the message into multiple lines, this gets very confusing for large responses.

This PR strips the newline characters before logging

### DIFF
--- a/support-modules/rest/src/main/scala/com/gu/rest/WebServiceHelper.scala
+++ b/support-modules/rest/src/main/scala/com/gu/rest/WebServiceHelper.scala
@@ -97,7 +97,7 @@ trait WebServiceHelper[Error <: Throwable] {
         .toTry
       responseBody <- Try(body.string())
       codeBody = CodeBody(code, responseBody)
-      _ = SafeLogger.info(s"response $code body: ${responseBody}")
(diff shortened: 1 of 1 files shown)


TomPretty (approved):

support-modules/rest/src/main/scala/com/gu/rest/WebServiceHelper.scala
```diff
@@ -97,7 +97,7 @@ trait WebServiceHelper[Error <: Throwable] {
         .toTry
       responseBody <- Try(body.string())
       codeBody = CodeBody(code, responseBody)
-      _ = SafeLogger.info(s"response $code body: ${responseBody}")
+      _ = SafeLogger.info(s"response $code body: ${responseBody.replace("\n", "")}")
```
johnduffell, line 100: another option which works for EC2 instance logging to cloudwatch (but not laambdas) is to indent the continuation lines with whitespace
```suggestion
      _ = SafeLogger.info(s"response $code body: ${responseBody.replace("\n", "\n ")}")
```

johnduffell: I suspected so. I haven't found a way to make it work with lambdas.

PR #3820, opened 2022-06-06
Fix eslint errors

TomPretty:

## Is this an AB test?
- [ ] Yes
- [ ] No


support-frontend/assets/helpers/internationalisation/country.ts
```diff
@@ -627,16 +627,17 @@ function detect(
...
 			countryGroupId,
 		);
 
-		if (candidateCountry !== null && country === null) {
+		if (candidateCountry !== null) {
 			country = candidateCountry;
+			break;
 		}
-	});
+	}
```
TomPretty, line 640: With the original version (using the `forEach` method) eslint was complaining about the conditional on line 642. By changing to a `for...of` loop eslint can now tell `country` is getting assigned before the check. I also think this is slightly better as we can now `break` from the loop early instead of having a conditional check

support-frontend/assets/helpers/internationalisation/currency.ts
```diff
@@ -137,15 +137,8 @@ const spokenCurrencies: Record<IsoCurrency, SpokenCurrency> = {
...
-	countryGroupId: CountryGroupId,
-): IsoCurrency | null | undefined {
-	const countryGroup: CountryGroup | null | undefined =
-		countryGroups[countryGroupId];
-
-	if (!countryGroup) {
-		return null;
-	}
+function fromCountryGroupId(countryGroupId: CountryGroupId): IsoCurrency {
+	const countryGroup = countryGroups[countryGroupId];
```
TomPretty, line 141: This one makes me slightly nervous. Given the type signature, `countryGroup` is always defined so we don't need the conditional. Typescript should catch anywhere that we use it incorrectly but I'm still worried that there might be some code that depends on it

TomPretty: I looked for every instance of type assertions casting to `CountryGroupId` and they all look fine so this should be safe

support-frontend/assets/components/ticker/contributionTicker.tsx
```diff
@@ -42,9 +42,9 @@ const getInitialTickerValues = (
 		tickerCountType === 'money' ? '/ticker.json' : '/supporters-ticker.json',
 	)
 		.then((resp) => resp.json())
-		.then((data) => {
-			const total = parseInt(data.total, 10);
-			const goal = parseInt(data.goal, 10);
+		.then((data: DataFromServer) => {
+			const total = Math.floor(data.total);
+			const goal = Math.floor(data.goal);
```
TomPretty, line 47: The types from the server are already `number`s so the call to `parseInt` wasn't actually correct as it expects a `string`.

support-frontend/assets/components/ticker/contributionTicker.tsx
```diff
@@ -158,7 +158,7 @@ export default class ContributionTicker extends Component<
 			);
 		}
 
-		if (this.state.goalReached && this.props.tickerEndType === 'unlimited') {
+		if (this.props.tickerEndType === 'unlimited') {
```
TomPretty, line 161: We don't need the `goalReached` check as we've already checked `!goalReached` above and returned from the function if so

support-frontend/assets/components/ticker/contributionTicker.tsx
```diff
@@ -171,17 +171,13 @@ export default class ContributionTicker extends Component<
 			);
 		}
 
-		if (this.state.goalReached && this.props.tickerEndType === 'hardstop') {
```
TomPretty, line 174: We don't need this check at all we've already returned if `!goalReached` or `tickerEndType !== 'hardstop'`

rupertbates (approved):
Looks good, I think it would be worth testing extensively though as there are quite a few changes that could potentially break things if there is code relying on previous quirks.

TomPretty:
> Looks good, I think it would be worth testing extensively though as there are quite a few changes that could potentially break things if there is code relying on previous quirks.

Yeah just been testing now and I'm happy that we're using it correctly everywhere. I'll merge and keep a close eye on it after it's live

PR #3821, opened 2022-06-07
Remove set up of extra node version in workflow

TomPretty:

Now that we're using the same version of node for cdk and support-frontend there is no need to have two separate node setup steps


marialani (approved):
👍

PR #3822, opened 2022-06-07
Add typescript & eslint workflows

TomPretty:

Adds two new workflows: `support-frontend-typescript` and `support-frontend-eslint`. Now that we've got both typescript and lint errors down to 0, we can turn on these checks in CI to ensure they stay that way.

I've followed the style of dotcom-rendering which is to have them as separate workflows so that they can run in parallel.

**NB:** The new actions won't show on this PR as it's not touching any files in the `support-frontend` directory. I tested they were working by making a small change to a `ts` file, but have now dropped that change.

### DIFF
--- a/.github/workflows/support-frontend-eslint.yml
+++ b/.github/workflows/support-frontend-eslint.yml
@@ -1,0 +1,28 @@
+name: support-frontend - eslint
+
+on:
+  push:
+    paths:
+      - support-frontend/**
+
+jobs:
+  support_frontend_build:
+    name: support-frontend - eslint
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


i-hardy (approved):
This is great, definitely good to have them as separate workflows so they don't slow down the build. We should probably also make these required checks under the branch protection rules once this is merged: https://github.com/guardian/support-frontend/settings/branch_protection_rules/736841

TomPretty:
> This is great, definitely good to have them as separate workflows so they don't slow down the build. We should probably also make these required checks under the branch protection rules once this is merged: https://github.com/guardian/support-frontend/settings/branch_protection_rules/736841

Yeah will do!

PR #3824, opened 2022-06-07
Notify Quantum Metric of user AB test participations 

GHaberis:

We need to notify Quantum Metric of a user's AB test participations, using the `window.QuantumMetricAPI.sendEvent` function. This PR sends this relevant data to Quantum Metric.

Whilst doing this I've also updated the logic to load Quantum Metric's script by replacing some custom logic to add the script with `import { loadScript } from '@guardian/libs'`. 

[**Trello Card**](https://trello.com/c/BfI7Ru7b/462-quantum-metric-implement-sendevent-call-for-ab-test-participations)

### DIFF
--- a/support-frontend/assets/helpers/tracking/quantumMetric.ts
+++ b/support-frontend/assets/helpers/tracking/quantumMetric.ts
@@ -1,46 +1,87 @@
 import { onConsentChange } from '@guardian/consent-management-platform';
+import { loadScript } from '@guardian/libs';
+import type { Participations } from 'helpers/abTests/abtest';
 import { isSwitchOn } from 'helpers/globalsAndSwitches/globals';
+import { logException } from 'helpers/utilities/logger';
 
-function pathIsValid(): boolean {
-	const locationPath = window.location.pathname;
-	// Contribution landing page and thank you page (all regions)
-	const contributionsRegex = /\/contribute|thankyou?(\?.*)?$/;
-	// Non-Gifting Digi Subs landing page, checkout and thank you page (all regions)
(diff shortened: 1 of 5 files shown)


GHaberis:
[Quantum Metric: Implement SendEvent call for AB test participations.](https://trello.com/c/BfI7Ru7b/462-quantum-metric-implement-sendevent-call-for-ab-test-participations)

support-frontend/assets/helpers/tracking/quantumMetric.ts
```diff
@@ -1,46 +1,87 @@
...
+				crossOrigin: 'anonymous',
+			},
+		)
+			.then(() => {
+				resolve();
+			})
+			.catch(() => {
+				logException('Failed to load Quantum Metric');
+			});
+	});
```
i-hardy, line 75:
```suggestion
	return loadScript(
			'https://cdn.quantummetric.com/instrumentation/1.31.3/quantum-gnm.js',
			{
				async: true,
				integrity:
					'sha384-jL1rlM/U30WWRsDEQDalNiDV8FOjayD5RCgzkywrohMqg6oME3zbszWB31/40MAD',
				crossOrigin: 'anonymous',
			},
		)
			.catch(() => {
				logException('Failed to load Quantum Metric');
			});
	});
```

If `loadScript` returns a promise there's no need to wrap it in the promise constructor.

i-hardy (approved):
One suggestion, otherwise looks good!

support-frontend/assets/helpers/tracking/quantumMetric.ts
GHaberis: It does indeed return a Promise, good suggestion 👍

PR #3825, opened 2022-06-07
Refactor and simplify address components' connection to Redux

i-hardy:
This refactors the `AddressFields` and `PostcodeFinder` component to remove their direct reliance on Redux, and the complex scoping mechanism that was used in order to be able to use them for both delivery and billing addresses.

The components themselves now simply accept props, and the connection to the store is managed by two separate connected component wrappers in `scopedAddressFields.ts`, that talk to the billing and delivery address respectively.

[**Trello Card**](https://trello.com/c/jXtoPc0R)

## Why are you doing this?

This will make it a lot easier to eventually migrate address state to Redux Toolkit- we'll just need to update the two wrapper components, rather than unpicking the scoping mechanism in the components themselves.

### DIFF
--- a/support-frontend/assets/components/subscriptionCheckouts/address/addressFields.tsx
+++ b/support-frontend/assets/components/subscriptionCheckouts/address/addressFields.tsx
@@ -38,7 +38,7 @@ type StatePropTypes = FormFields & {
 
 type PropTypes = StatePropTypes &
 	AddressActionCreators & {
-		setPostcode: (postcode: string) => void;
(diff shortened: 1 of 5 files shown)


i-hardy:
[🛠 Redux Toolkit migration - decouple address and postcode components from state 🛠](https://trello.com/c/jXtoPc0R/463-%F0%9F%9B%A0-redux-toolkit-migration-decouple-address-and-postcode-components-from-state-%F0%9F%9B%A0)

GHaberis (approved):
Looks good 👍

PR #3826, opened 2022-06-08
Fix default promo copy on GW product page

tomrf1:
The /subscribe/weekly page renders copy from a promo.
We have a list of promos. Currently it only checks the first promo in the list. This can result in a promo showing at the bottom of the page (in the 'cards') but not in the copy at the top.
This PR changes this logic to scan the list until a valid promo is found.


KaliedaRik (approved):
Seems reasonable

PR #3827, opened 2022-06-08
Remove av calculator client as a dependency

TomPretty:

Removes the `acquisitions-value-calculator-client` from the `acquisitions-firehose-transformer` and `acquisition-events` modules. This client would kick of a lambda which would compute the AV. The AV calculation is quite simple so we've copied it across from the [client repo](https://github.com/guardian/acquisitions-value-calculator).

The motivation behind this was to reduce the size of the `acquisitions-firehose-transformer` project as the lambda deploy had started to fail due to it's file size. It's now deploying successfully to CODE.


TomPretty:
> We should monitor that lambda and if it is no longer receiving traffic we can delete it and archive the repos

Yeah good shout - I'll create a ticket to check in on it

PR #3831, opened 2022-06-09
Small tweak to the patrons data pipeline

rupertbates:
We want to filter out the patrons@theguardian.com email address from the Stripe to SupporterProductData pipeline as this is an address used by CSRs when a patron does not provide an email, however we shouldn't filter out all guardian email addresses - mostly because this makes testing difficult!

### DIFF
--- a/stripe-patrons-data/src/main/scala/com/gu/patrons/services/StripeSubscriptionsProcessor.scala
+++ b/stripe-patrons-data/src/main/scala/com/gu/patrons/services/StripeSubscriptionsProcessor.scala
@@ -34,9 +34,7 @@ class StripeSubscriptionsProcessor(
   def processSubs(list: List[StripeSubscription]) = {
     Future.sequence(
       list
-        .filterNot(sub =>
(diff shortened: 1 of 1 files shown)


stripe-patrons-data/src/main/scala/com/gu/patrons/services/StripeSubscriptionsProcessor.scala
```diff
@@ -34,9 +34,7 @@ class StripeSubscriptionsProcessor(
   def processSubs(list: List[StripeSubscription]) = {
     Future.sequence(
       list
-        .filterNot(sub =>
-          sub.customer.email.contains("@guardian.co.uk") || sub.customer.email.contains("@theguardian.com"),
-        )
+        .filterNot(sub => sub.customer.email == "patrons@theguardian.com")
```
johnduffell, line 37:
```suggestion
    val unknownEmailViaCSR = "patrons@theguardian.com"
    Future.sequence(
      list
        .filterNot(sub => sub.customer.email == unknownEmailViaCSR)
```

johnduffell (approved):
just a small bit of documentation in the suggestion

PR #3901, opened 2022-06-14
Refactor CSR mode related code

marialani:
~~This PR creates a new Redux Toolkit slice for the state related to 'CSR mode' and integrates it into the existing checkouts.~~

We’ve decided against placing CSR mode state in the store, as the CSR banner is present on pages where we don’t use Redux and therefore the component does not always have access to the store. There is an option to pass the store as a prop to a connected component but as the CSR banner is mainly used to provide a visual indication to a CSR that all is working correctly, it most likely does not need to be in the store.

[**Trello Card**](https://trello.com/c/XZClH0l6/480-%F0%9F%9B%A0-redux-toolkit-migration-refactor-csr-mode-state-%F0%9F%9B%A0)

## Why are you doing this?
~~This is part of work to overhaul our usage of Redux and migrate to using Toolkit to simplify and streamline working on Redux-related code.~~

### DIFF
--- a/support-frontend/assets/helpers/subscriptionsForms/formActions.ts
+++ b/support-frontend/assets/helpers/subscriptionsForms/formActions.ts
@@ -237,48 +237,24 @@ function setCsrCustomerData(
 	addressType: AddressType,
 	csrCustomerData: CsrCustomerData,
 ) {
-	return (
-		dispatch: SubscriptionsDispatch,
-		getState: () => CheckoutState,
-	): void => {
-		csrCustomerData.customer.email &&
-			dispatch(formActionCreators.setEmail(csrCustomerData.customer.email));
-		csrCustomerData.customer.email &&
-			dispatch(
-				formActionCreators.setConfirmEmail(csrCustomerData.customer.email),
-			);
-		csrCustomerData.customer.firstName &&
-			dispatch(
-				formActionCreators.setFirstName(csrCustomerData.customer.firstName),
-			);
+	return (dispatch: SubscriptionsDispatch): void => {
+		const { email, firstName, country, street, city, postcode, state } =
+			csrCustomerData.customer;
+
+		email && dispatch(formActionCreators.setEmail(email));
+		email && dispatch(formActionCreators.setConfirmEmail(email));
+		firstName && dispatch(formActionCreators.setFirstName(firstName));
 		dispatch(formActionCreators.setLastName(csrCustomerData.customer.lastName));
 
 		dispatch(formActionCreators.setCsrUsername(csrUserName(csrCustomerData)));
 		dispatch(formActionCreators.setSalesforceCaseId(csrCustomerData.caseId));
 
 		const addressActions = addressActionCreatorsFor(addressType);
-		csrCustomerData.customer.country &&
-			addressActions.setCountry(csrCustomerData.customer.country)(dispatch);
-		csrCustomerData.customer.street &&
-			addressActions.setAddressLineOne(csrCustomerData.customer.street)(
-				dispatch,
-				getState,
-			);
-		csrCustomerData.customer.city &&
-			addressActions.setTownCity(csrCustomerData.customer.city)(
-				dispatch,
-				getState,
-			);
-		csrCustomerData.customer.postcode &&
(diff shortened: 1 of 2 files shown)


marialani:
[🛠 Redux Toolkit migration - refactor CSR Mode state 🛠](https://trello.com/c/XZClH0l6/480-%F0%9F%9B%A0-redux-toolkit-migration-refactor-csr-mode-state-%F0%9F%9B%A0)

support-frontend/assets/components/csr/csrBanner.tsx
```diff
@@ -17,18 +17,23 @@ const banner = css`
...
-	>(undefined);
+interface CsrBannerProps {
+	csrCustomerData: CsrCustomerData;
+	setCsrCustomerData: ActionCreatorWithPayload<CsrCustomerData>;
+}
+
+function CsrBanner({
+	csrCustomerData,
+	setCsrCustomerData,
+}: CsrBannerProps): JSX.Element | null {
```
i-hardy, line 28: If we're providing data from the store as props it would be better to use `mapStateToProps` and `mapDispatchToProps`- this would also remove the need to inject `<CsrBanner />` as a prop into the `Page` component, and means that `setCsrCustomerData` will be wrapped with `dispatch` and will actually trigger a state update when it's called- passing in the action creator by itself means all you have is a function that returns an action object.

```suggestion
function mapStateToProps(state: SubscriptionsState) {
	return {
		csrCustomerData: state.page.checkoutForm.csrCustomerData,
	}
}

const mapDispatchToProps = {
	setCsrCustomerData,
}

const connector = connect(mapStateToProps, mapDispatchToProps);

type CsrBannerProps = ConnectedProps<typeof connector>;

// component code

export default connector(CsrBanner);
```

support-frontend/assets/components/csr/csrBanner.tsx
```diff
@@ -17,18 +17,23 @@ const banner = css`
...
 	useCsrCustomerData(setCsrCustomerData);
 
-	if (csrCustomerData) {
+	if (csrCustomerData.csr.firstName) {
 		return (
 			<div css={container}>
 				<div css={banner}>
 					You are in customer support mode. Signed in as:{' '}
-					{csrUserName(csrCustomerData)}
+					{csrCustomerData.csr.userName ?? csrUserName(csrCustomerData)}
```
i-hardy, line 36: Is there a reason for checking for two different properties for the CSR user?

support-frontend/assets/helpers/subscriptionsForms/formActions.ts
```diff
@@ -241,44 +239,23 @@ function setCsrCustomerData(
...
-		csrCustomerData.customer.state &&
-			addressActions.setState(csrCustomerData.customer.state)(
-				dispatch,
-				getState,
-			);
+		country && addressActions.setCountry(country)(dispatch);
+		street && addressActions.setAddressLineOne(street)(dispatch, getState);
+		city && addressActions.setTownCity(city)(dispatch, getState);
+		postcode && addressActions.setPostcode(postcode)(dispatch, getState);
+		state && addressActions.setState(state)(dispatch, getState);
```
i-hardy, line 258: This is so much easier to read 🤩 but we can also just pass these to `dispatch` rather than needing to call them again with `dispatch` and `getState`- Redux will do that for us internally:

```suggestion
		country && dispatch(addressActions.setCountry(country));
		street && dispatch(addressActions.setAddressLineOne(street));
		city && dispatch(addressActions.setTownCity(city));
		postcode && dispatch(addressActions.setPostcode(postcode));
		state && dispatch(addressActions.setState(state));
```

i-hardy (changes requested):
Changing how we provide `setCsrCustomerData` to the banner is necessary for it to actually update the state; the other stuff I've commented on is more superficial and not really a blocker to getting this merged.

support-frontend/assets/components/csr/csrBanner.tsx
marialani: We use the `csrUserName()` function to construct the csr username from the first and last names then store it in state. The function is just constructing a simple string so I'll remove the additional check

GHaberis (approved):
Looks good 👍

PR #3912, opened 2022-06-17
Remove DD from subs

rupertbates:
Follow up to #3911


tomrf1 (approved):

support-frontend/assets/helpers/subscriptionsForms/countryPaymentMethods.ts
```diff
@@ -1,9 +1,8 @@
-import { DirectDebit, PayPal, Stripe } from 'helpers/forms/paymentMethods';
+import { PayPal, Stripe } from 'helpers/forms/paymentMethods';
 import type { IsoCountry } from 'helpers/internationalisation/country';
 import type { IsoCurrency } from 'helpers/internationalisation/currency';
 
 export type SubscriptionsPaymentMethod =
-	| typeof DirectDebit
```
i-hardy, line 6: I don't think there's any need to remove it from the typedef; as long as it's not being returned from `supportedPaymentMethods` then the option won't show up in the checkout, and this will avoid the cascade of type errors elsewhere in the code.

PR #3919, opened 2022-06-17
Remove AmazonPay objects from redux store

TomPretty:

This change removes the `AmazonLoginObject` and `AmazonPaymentsObject` from the redux store, and moves them to component state. We're removing them as they contain non-serialisable state.

To do this, I've created a `useAmazonPayObjects` hook that is responsible for initialising the AmazonPay sdk and storing the objects in state. I initially used this inside of the new `AmazonPayCheckout` component that I factored out of the `ContributionsSubmit` component. However, that gets mounted/unmounted as the user changes payment method which makes it a little awkward as we only want to initialise the sdk once. Instead, I lifted the `useAmazonPayObjects` hook into the parent `ContributionsSubmit` component as that is always rendered. This makes the `AmazonPayCheckout` component less useful, but I think it's still nice to have as it removes some logic from the `ContributionsSubmit` component.

[**Trello Card**](https://trello.com/c/YMpJALyM/507-%F0%9F%9B%A0-redux-toolkit-migration-remove-amazonpay-objects-from-contributions-state-%F0%9F%9B%A0)

### DIFF
--- a/support-frontend/assets/helpers/forms/paymentIntegrations/amazonPay/setup.ts
+++ b/support-frontend/assets/helpers/forms/paymentIntegrations/amazonPay/setup.ts
@@ -1,23 +1,25 @@
-import type { Dispatch } from 'redux';
 import type { CountryGroupId } from 'helpers/internationalisation/countryGroup';
 import { isProd } from 'helpers/urls/url';
 import { logException } from 'helpers/utilities/logger';
-import {
-	setAmazonPayLoginObject,
-	setAmazonPayPaymentsObject,
-} from 'pages/contributions-landing/contributionsLandingActions';
+import type { AmazonLoginObject, AmazonPaymentsObject } from './types';
 
 // ---- Setup ---- //
 
 export const setupAmazonPay = (
 	countryGroupId: CountryGroupId,
-	dispatch: Dispatch,
+	onAmazonLoginReady: (loginObject: AmazonLoginObject) => void,
+	onAmazonPaymentsReady: (paymentsObject: AmazonPaymentsObject) => void,
 	isTestUser: boolean,
 ): void => {
 	const isSandbox = isTestUser || !isProd();
 
 	if (isSupportedCountryGroup(countryGroupId)) {
-		setupAmazonPayCallbacks(countryGroupId, dispatch, isSandbox);
+		setupAmazonPayCallbacks(
+			countryGroupId,
+			onAmazonLoginReady,
+			onAmazonPaymentsReady,
+			isSandbox,
+		);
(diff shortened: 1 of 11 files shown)


support-frontend/assets/helpers/forms/paymentIntegrations/amazonPay/setup.ts
```diff
@@ -1,23 +1,25 @@
...
-} from 'pages/contributions-landing/contributionsLandingActions';
+import type { AmazonLoginObject, AmazonPaymentsObject } from './types';
 
 // ---- Setup ---- //
 
 export const setupAmazonPay = (
 	countryGroupId: CountryGroupId,
-	dispatch: Dispatch,
+	onAmazonLoginReady: (loginObject: AmazonLoginObject) => void,
+	onAmazonPaymentsReady: (paymentsObject: AmazonPaymentsObject) => void,
```
TomPretty, line 11: The main change in this file is that the setup functions now take specific callbacks to set the objects instead of the generic dispatch function

support-frontend/assets/pages/contributions-landing/components/AmazonPay/useAmazonPayObjects.ts
```diff
@@ -0,0 +1,41 @@
...
+		const hasLoaded = loginObject !== undefined || paymentsObject !== undefined;
+
+		if (hasSelectedAmazonPay && !hasLoaded)
+			setupAmazonPay(
+				countryGroupId,
+				setLoginObject,
+				setPaymentsObject,
+				isTestUser,
+			);
+	}, [hasSelectedAmazonPay]);
```
TomPretty, line 38: I wasn't quite sure what the best practice was to do here. Technically, `loginObject` and `paymentsObject` should be in the dependencies array, as they are referenced in side of the effect. However, I explicitly don't want the transition from `undefined` to defined to retrigger the effect (the `if` statement will prevent the `setupAmazonPay` running again, but still feels slightly watsted to trigger the effect again).

TomPretty:
[🛠 Redux Toolkit migration - remove AmazonPay objects from Contributions state 🛠](https://trello.com/c/YMpJALyM/507-%F0%9F%9B%A0-redux-toolkit-migration-remove-amazonpay-objects-from-contributions-state-%F0%9F%9B%A0)

i-hardy (approved):
Like we discussed in mobbing, I think we should definitely revisit the way we manage loading various third party SDKs for payment stuff in the future to see what improvements or standardisations we can make, but this is definitely good for now and a huge improvement to not have this stuff in Redux!

PR #3942, opened 2022-06-21
add temp hackday redirect for getting bwid cookie into CODE env

johnduffell:
## What does it do?

This PR adds a temporary hack to get the bwid into CODE env for the hack days.
This is because we want to let people see what product we would recommend to them based on their actual browsing history and the propensity model.  However the main code is hacky so won't go to PROD.  See https://github.com/guardian/support-frontend/pull/3943

This PR needs to be reverted after the hack day as it won't be useful.

## Some notes on PII for review!

The browser id is PII so we need to make sure it is taken care of.  The CODE env is entirely behind https so encrypted in transit, same as PROD.
The cookie value is sent as a URL param instead of a cookie, which is slightly more likely to be copy pasted somewhere by mistake and shared.
The bwid -> product mappings will be stored in dynamodb, which is encrypted at rest and in transit.  The data will be stored behind AWS auth and only accessible via the one endpoint (which will only be deployed to CODE for the demo.
No other data relating to the browser id will be available at any point.

### DIFF
--- a/support-frontend/app/controllers/DiagnosticsController.scala
+++ b/support-frontend/app/controllers/DiagnosticsController.scala
@@ -6,7 +6,7 @@ import Results.Ok
 
 class DiagnosticsController(
     actionRefiners: CustomActionBuilders,
-) {
+) extends Results {
 
   val relevantCookies = List(
     "gu_user_features_expiry",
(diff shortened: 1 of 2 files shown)


tomrf1 (approved):

johnduffell:
I don't think we are going to use this in the end - we won't have time to assess all the browser ids against the model, as it is a time consuming process.
As such we will just run a few selected ones for the demo, and show those.
If this isn't needed this can be closed after the demos!

PR #3943, opened 2022-06-21
DO not merge: add propensity based page for recommending a product

johnduffell:
This is the hack day project.  It adds an endpoint to show the product we would recommend based on your browser id and the propensity model generated by data science.

It will be deployed to CODE for the presentations and shortly afterwards, and then removed afterwards.


johnduffell:
hack day is over, the dynamo table is deleted, and anywaya this PR needs more work if it is to make it into production!

PR #3944, opened 2022-06-22
Use global default promo for Digisub

tomrf1:
the offers are all the same, so a single promo works.
Currently on the checkout page if a user changes their country then it loses the offer

### Before - offer is lost when switching to US:
Screen Shot 2022-06-22 at 09 36 25

### After - offer is retained when switching to US:
Screen Shot 2022-06-22 at 09 36 52


KaliedaRik (approved):
LGTM

PR #3945, opened 2022-06-22
Quantum Metric: sendEvent on contribution amount updates

GHaberis:

This PR introduces a new `sendEvent` call on Quantum Metric when users toggle the contribution amount on the contributions checkout or specify an "other amount" to contribute.

### DIFF
--- a/support-frontend/assets/helpers/tracking/quantumMetric.ts
+++ b/support-frontend/assets/helpers/tracking/quantumMetric.ts
@@ -56,17 +56,6 @@ function sendEventSubscriptionCheckoutEvent(
 ): void {
 	void canRunQuantumMetric().then((canRun) => {
 		if (canRun) {
-<<<<<<< HEAD
-=======
-			const sourceCurrency = productPrice.currency;
-			const value = getAnnualValue(productPrice, billingPeriod);
-
-			if (!value) {
-				return;
-			}
-
-			const targetCurrency: IsoCurrency = 'GBP';
->>>>>>> b396f876d (check feature switch on sendEvent)
 			const sendEventWhenReady = () => {
 				const sourceCurrency = productPrice.currency;
 				const targetCurrency: IsoCurrency = 'GBP';
@@ -231,33 +220,6 @@ function addQM() {
 	});
 }
 
-<<<<<<< HEAD
-=======
-function canRunQuantumMetric(): Promise<boolean> {
-	// resolve immediately with false if the feature switch is OFF
-	if (!isSwitchOn('featureSwitches.enableQuantumMetric')) {
-		return Promise.resolve(false);
-	}
-	// checks users consent status
-	return new Promise((resolve) => {
-		onConsentChange((state) => {
-			if (
-				state.ccpa?.doNotSell === false || // check whether US users have NOT withdrawn consent
-				state.aus?.personalisedAdvertising || // check whether AUS users have consented to personalisedAdvertising
-				(state.tcfv2?.consents && // check TCFv2 purposes for non-US/AUS users
-					state.tcfv2.consents['1'] && // Store and/or access information on a device
-					state.tcfv2.consents['8'] && // Measure content performance
-					state.tcfv2.consents['10']) // Develop and improve products
-			) {
-				resolve(true);
-			} else {
-				resolve(false);
-			}
-		});
-	});
-}
-
->>>>>>> b396f876d (check feature switch on sendEvent)
 export function init(participations: Participations): void {
 	void canRunQuantumMetric().then((canRun) => {
 		if (canRun) {
--- a/support-frontend/assets/helpers/tracking/quantumMetric.ts
+++ b/support-frontend/assets/helpers/tracking/quantumMetric.ts
@@ -32,7 +32,7 @@ enum SendEventCheckoutConversion {
 	GuardianWeeklySubGift = 70,
 }
 
-enum SendEventContributionAmountToggle {
+enum SendEventContributionAmountUpdate {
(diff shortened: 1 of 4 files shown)


GHaberis:
[Quantum Metric: Implement SendEvent call for Contribution Amount Choices](https://trello.com/c/SQAfhmri/489-quantum-metric-implement-sendevent-call-for-contribution-amount-choices)

support-frontend/assets/helpers/tracking/quantumMetric.ts
```diff
@@ -170,6 +177,57 @@ export function sendEventSubscriptionCheckoutConversion(
...
+				sendEventWhenReady();
+			} else {
+				waitForQuantumMetricAPi(() => {
+					sendEventWhenReady();
+				});
+			}
+		}
+	});
+}
+
```
TomPretty, line 230: I wonder if we can reduce a little bit of the duplicated code here with the subs events? I had a go at moving all of the common stuff into the `sendEvent` method.

In that case the send event method becomes:

```ts
function sendEvent(
	id: SendEventId,
	isConversion: boolean,
	value: string,
): void {
	const send = () => {
		if (window.QuantumMetricAPI?.isOn()) {
			window.QuantumMetricAPI.sendEvent(id, isConversion ? 1 : 0, value);
		}
	};

	void canRunQuantumMetric().then((canRun) => {
		if (canRun) {
			/**
			 * Quantum Metric's script sets up QuantumMetricAPI.
			 * We need to check it is defined and ready before we can
			 * send events to it. If it is ready we call sendEventWhenReady
			 * immediately. If it is not ready we poll a function that checks
			 * if QuantumMetricAPI is available. Once it's available we
			 * call sendEventWhenReady.
			 */
			if (window.QuantumMetricAPI?.isOn()) {
				send();
			} else {
				waitForQuantumMetricAPi(() => {
					send();
				});
			}
		}
	});
}
```

In that case, the send subs event then becomes:

```ts
function sendEventSubscriptionCheckoutEvent(
	id: SendEventCheckoutStart | SendEventCheckoutConversion,
	productPrice: ProductPrice,
	billingPeriod: BillingPeriod,
	isConversion: boolean,
): void {
	const sourceCurrency = productPrice.currency;
	const targetCurrency: IsoCurrency = 'GBP';
	const value = getSubscriptionAnnualValue(productPrice, billingPeriod);

	if (!value) {
		return;
	} else if (window.QuantumMetricAPI?.isOn()) {
		const convertedValue: number =
			window.QuantumMetricAPI.currencyConvertFromToValue(
				value,
				sourceCurrency,
				targetCurrency,
			);

		sendEvent(id, isConversion, convertedValue.toString());
	}
}
```

and the contribs event:

```ts
export function sendEventContributionAmountUpdated(
	amount: number | 'other',
	sourceCurrency: IsoCurrency,
): void {
	if (amount === 'other') {
		return;
	}

	if (window.QuantumMetricAPI?.isOn()) {
		const sendEventId =
			contributionType === 'ONE_OFF'
				? SendEventContributionAmountUpdate.SingleContribution
				: SendEventContributionAmountUpdate.RecurringContribution;
		const valueInPence =
			contributionType === 'ONE_OFF' || contributionType === 'ANNUAL'
				? amount * 100
				: amount * 100 * 12;
		const targetCurrency: IsoCurrency = 'GBP';
		const convertedValue: number =
			window.QuantumMetricAPI.currencyConvertFromToValue(
				valueInPence,
				sourceCurrency,
				targetCurrency,
			);
		sendEvent(sendEventId, false, convertedValue.toString());
	}
}
```

TomPretty: I guess this does mean that we interact with the `QuantumMetricAPI` before we check `canRunQuantumMetric`. If we're not allowed to do that, we could always make the `sendEvent` function take a `getEvent` function instead, that we only invoke after the check

support-frontend/assets/pages/contributions-landing/components/ContributionAmount.tsx
```diff
@@ -187,12 +191,23 @@ function ContributionAmount(props: PropTypes) {
...
-						}
+						onBlur={() => {
+							if (otherAmount) {
+								trackComponentClick(
+									`npf-contribution-amount-toggle-${props.countryGroupId}-${props.contributionType}-${otherAmount}`,
+								);
+
+								const otherAmountNumber = parseInt(otherAmount, 10);
+
+								if (typeof otherAmountNumber === 'number') {
```
TomPretty, line 202: Is the idea here to stop `NaN`s being sent? Unfortunately `typeof NaN` is `'number'`

support-frontend/assets/helpers/tracking/quantumMetric.ts
GHaberis: Yep I spotted the duplicated code and refactored into a function `sendEventWhenReadyTrigger` in this follow-up PR: https://github.com/guardian/support-frontend/pull/3964/files

support-frontend/assets/pages/contributions-landing/components/ContributionAmount.tsx
GHaberis: Indeed it is, good spot, I'll use `isNaN()` to catch `NaN` instead!

support-frontend/assets/helpers/tracking/quantumMetric.ts
TomPretty: oh nice one!

TomPretty (approved):
LGTM 👍

PR #3949, opened 2022-06-23
Limit acquisition-events-api 4XX alarm

tomrf1:
bots are hitting `GET /`, which triggers this alarm. We're only interested in POST requests


TomPretty (approved):
👍

PR #3973, opened 2022-06-27
Use shared Guardian browserslist config

i-hardy:

This updates our Browserslist config to use the shared [@guardian/configs](https://github.com/guardian/configs) package, which draws on our actual usage stats.


GHaberis (approved):
Nice 👏

PR #3986, opened 2022-07-01
Refactor postcode finder state to use RTK

TomPretty:

Refactor the postcode finder state to use redux toolkit. I've also made a few small tweaks to the PostcodeFinder component:

- reordered the file and added some comment headers
- removed the old helper component and factored out a new one (`AddressSelector`)
- handle the case there is just a single address better (it will now say `1 address found` instead of `1 addresses found`)
- improve formatting when there's no address line two (before there was a trailing comma)
- added some margin under the address selector

[**Trello Card**](https://trello.com/c/9VWsfMMd/519-%F0%9F%9B%A0-redux-toolkit-migration-refactor-postcode-finder-state-%F0%9F%9B%A0)


support-frontend/assets/components/subscriptionCheckouts/address/addressFields.tsx
```diff
@@ -88,27 +91,24 @@ export function AddressFields({ scope, ...props }: PropTypes): JSX.Element {
...
-						if (lineTwo) {
-							props.setLineTwo(lineTwo);
-						}
-
-						if (city) {
-							props.setTownCity(city);
-						}
+						props.setLineOne(lineOne ?? '');
+						props.setLineTwo(lineTwo ?? '');
+						props.setTownCity(city ?? '');
```
TomPretty, line 111: I removed the conditional checking because of an edge case I encountered. If you first select an address that has a `lineTwo` and then select another address that doesn't, the old `lineTwo` doesn't get overwritten and is left in the form

support-frontend/assets/components/subscriptionCheckouts/address/postcodeFinder.tsx
```diff
@@ -9,123 +8,148 @@ import {
...
+			return;
+		}
+
+		if (postcode.trim() === '') {
+			onPostcodeError('Please enter a postcode');
+			return;
+		}
+
+		onFindAddress(postcode);
+	}
```
TomPretty, line 47: This behaviour is slightly different than before. First, we now do an `isLoading` check to prevent kicking off a second search whilst the first is loading (even though the button isn't displayed, you can still hit enter in the text field to kick off the search). 

Second, we now do the error check inside of this submit handler, instead of inside of the `onFindAddress` callback - this was to keep the async thunk code as simple as possible

support-frontend/assets/components/subscriptionCheckouts/address/postcodeFinder.tsx
```diff
@@ -9,123 +8,148 @@ import {
...
+			</>
+		</Select>
+	);
+}
+
+// ---- Helper functions ---- //
+
+function pluralisedAddress(numAddresses: number) {
+	return numAddresses === 1 ? 'address' : 'addresses';
+}
```
TomPretty, line 130: According to [this ONS FOI request](https://www.ons.gov.uk/aboutus/transparencyandgovernance/freedomofinformationfoi/numberofukpostcodeswhichonlyhaveonehousehold?:uri=aboutus/transparencyandgovernance/freedomofinformationfoi/numberofukpostcodeswhichonlyhaveonehousehold) there are over 55k postcodes that only have a single address. No idea how many of those are guardian readers, but now we can rest assured they won't have to see '1 addresses found' (they're probably used to that though...)

support-frontend/assets/components/subscriptionCheckouts/address/postcodeFinder.tsx
```diff
@@ -9,123 +8,148 @@ import {
...
+
+function addressOption(address: PostcodeFinderResult) {
+	if (address.lineTwo) {
+		return [address.lineOne, address.lineTwo].join(', ');
+	}
+	return address.lineOne ?? '';
+}
```
TomPretty, line 137: Previously, we just did `[address.lineOne, address.lineTwo].join(', ')` which means if we only have a `address.lineOne` we have a trailing comma which looks a little weird. 

We still have a slight issue in that if we don't have a `lineOne` we'll just show an empty string (although before we'd just show `, `, so maybe it's still better than that?)

Perhaps we should filter them out of the results? We can do this as a follow up

support-frontend/assets/components/subscriptionCheckouts/address/postcodeFinder.tsx
```diff
@@ -9,123 +8,148 @@ import {
...
+		display: flex;
+		align-items: flex-end;
+	`,
+	buttonContainer: css`
+		margin-left: ${space[3]}px;
+	`,
+	addressSelectorContainer: css`
+		margin-top: ${space[6]}px;
+	`,
+};
```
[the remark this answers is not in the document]
TomPretty: (also note the trailing comma in the address selector option in the before screenshot)

michaelbjacobson (approved):
Looks great to me! 🚀

PR #3989, opened 2022-07-04
Connect the subscriptions checkouts to the new product state

i-hardy:

This connects up the subscriptions checkouts to the new product reducer.

[**Trello Card**](https://trello.com/c/oiERsJv5)

## Why are you doing this?

This is part of our ongoing Redux Toolkit migration work, and enables a significant simplification of how we handle the remaining old Redux code and the new store instance, as we no longer need to initialise a new reducer with product-related state for every checkout.

I tried but unfortunately could not find a good way to split this across multiple PRs. The majority of the file changes are just updates to import paths, parameter types and lookup paths for `mapStateToProps`; I've commented on the more significant changes.


support-frontend/assets/helpers/productPrice/paperSavingsVsRetail.ts
```diff
@@ -0,0 +1,37 @@
...
+	prices: CountryGroupPrices,
+	fulfilmentOption: FulfilmentOptions,
+) {
+	return ActivePaperProductTypes.map((productOption) => {
+		const price = prices[fulfilmentOption]?.[productOption]?.Monthly?.GBP;
+		return price?.savingVsRetail ?? 0;
+	});
+}
+
+function getMaxSavingVsRetail(
```
i-hardy, line 22: We previously had a lot of duplicated functions for getting product prices for paper, that did the same thing (or even directly called) functions used on other checkouts. With the simplifying of product price lookup via selectors I have removed those duplicated functions and moved these remaining ones, which are specific to displaying paper price info, into a file with a more specific name.

support-frontend/assets/helpers/redux/checkout/product/reducer.ts
```diff
@@ -31,9 +41,6 @@ export const productSlice = createSlice({
 		setProductPrices(state, action: PayloadAction<ProductPrices>) {
 			state.productPrices = action.payload;
 		},
-		setSelectedProductPrice(state, action: PayloadAction<ProductPrice>) {
```
i-hardy, line 34: After tinkering around with it for a while I decided that ultimately trying to maintain the 'selected product' in Redux state was more trouble than it's worth. A subscriptions product is essentially a _function of_ the product type, billing period, fulfilment and product options, which we can express more elegantly with a selector.

support-frontend/assets/helpers/redux/checkout/product/selectors/productPrice.ts
```diff
@@ -0,0 +1,98 @@
...
+	ANNUAL: () => false,
+};
+
+const getDefaultPriceObject = (currency: IsoCurrency): ProductPrice => ({
+	price: 0,
+	currency,
+	fixedTerm: false,
+});
+
+export function selectPriceForProduct(state: SubscriptionsState): ProductPrice {
```
i-hardy, line 41: These are the new product price selectors, for selecting the base product price, price with discounts, and the corresponding price with/without digital for the paper checkout.

support-frontend/assets/helpers/redux/subscriptionsStore.ts
```diff
@@ -39,42 +43,41 @@ export const startSubscriptionsListening =
...
 	reducer: baseReducer,
-	// Makes devtools work correctly with the re-created store
-	devTools: false,
 	middleware: (getDefaultMiddleware) =>
 		getDefaultMiddleware().prepend(listenerMiddleware.middleware),
 });
 
 export type SubscriptionsStore = typeof subscriptionsStore;
 
-export function addPageReducer(
```
i-hardy, line 50: Because all of the information we used to pass into `createReducer` is now handled by the product and address slices, and the base subs checkout state is the same for all products, we can get rid of this horrible fudge and simplify the store setup.

i-hardy:
[🛠 Redux Toolkit migration - integrate new product state into subscriptions checkout 🛠](https://trello.com/c/oiERsJv5/450-%F0%9F%9B%A0-redux-toolkit-migration-integrate-new-product-state-into-subscriptions-checkout-%F0%9F%9B%A0)

support-frontend/assets/components/directDebit/directDebitProgressiveDisclosure/directDebitForm.test.tsx
```diff
@@ -12,22 +12,14 @@ import { commonReducer } from 'helpers/redux/commonState/reducer';
...
 
-const pageReducer = (initialState: WithDeliveryCheckoutState) =>
-	createReducer(
-		GuardianWeekly,
-		'Annual',
-		formatMachineDate(new Date()),
-		initialState.page.checkout.productOption,
-		initialState.page.checkout.fulfilmentOption,
-	);
+const pageReducer = () => createReducer();
```
rupertbates, line 17: What is the advantage of this over just using `createReducer` as is?

i-hardy: None, I think this is just an artefact from when we needed to pass a lot of arguments to `createReducer`. Might as well just use `createReducer` directly, I'll update these.

support-frontend/assets/helpers/redux/checkout/product/reducer.ts
```diff
@@ -52,10 +59,17 @@ export const productSlice = createSlice({
...
+			state.startDate = action.payload;
 		},
 	},
+	extraReducers: (builder) => {
+		builder.addCase(setDeliveryCountry, (state, action) => {
+			if (state.productType === GuardianWeekly) {
+				state.fulfilmentOption = getWeeklyFulfilmentOption(action.payload);
+			}
+		});
+	},
```
TomPretty, line 72: This is cool! So can we remove the address side effect now? https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/helpers/redux/checkout/address/sideEffects.ts#L43

i-hardy: We'll still need it to be able to reset the payment method when the country changes, but once we've migrated payments too then yes- I've left a comment by the `'ON_DELIVERY_COUNTRY_CHANGED'` case in the old reducer.

TomPretty: Ahhh yeah I've just got to that comment! Sounds good 👍

support-frontend/assets/helpers/productPrice/productOptions.ts
```diff
@@ -72,6 +72,11 @@ const paperProductsWithoutDigital = {
 	EverydayPlus: Everyday,
 } as Record<ProductOptions, ProductOptions>;
 
+const correspondingPaperProducts = {
```
rupertbates, line 75: Would `allPaperProducts` be a better name for this?

i-hardy: Possibly? Its purpose is to look up the product option with/without digital that corresponds to the one selected, but I don't feel great about the name.

i-hardy: I had a bit of a think about this and decided that a function was both clearer and easier to name than an object, and added a comment with it for additional clarity.

support-frontend/assets/helpers/redux/checkout/product/selectors/productPrice.ts
```diff
@@ -0,0 +1,99 @@
...
+		const selectedProductPrice = getProductPrice(
+			product.productPrices,
+			common.internationalisation.countryId,
+			product.billingPeriod,
+			product.fulfilmentOption,
+			product.productOption,
+		);
+		return selectedProductPrice;
+	}
+	return getDefaultPriceObject(common.internationalisation.currencyId);
```
rupertbates, line 54: Is it safe to return a default price here? Is there a risk that we end up charging 0 for a subscription in which case it might be better to throw and error?

support-frontend/assets/helpers/redux/checkout/product/selectors/productPrice.ts
```diff
@@ -0,0 +1,99 @@
...
+	if (requiredFieldsForProduct[product.productType](product)) {
+		return finalPrice(
+			product.productPrices,
+			common.internationalisation.countryId,
+			product.billingPeriod,
+			product.fulfilmentOption,
+			product.productOption,
+		);
+	}
+	return getDefaultPriceObject(common.internationalisation.currencyId);
```
rupertbates, line 69: Same question

rupertbates (approved):
I've left a question about the safety of providing a default price if the required state is missing but other than that this looks good.

support-frontend/assets/helpers/redux/checkout/product/selectors/productPrice.ts
i-hardy: You're right, it's highly unlikely that we'd ever hit this but throwing is a better choice.

PR #3991, opened 2022-07-05
Update other amount field UX

tompretty:

We've had a number of issue raised by users accidently creating contributions of 100x what they expected. Our theory is that they typed `.00` into the other amount field. The current behaviour of the field is to remove the `.` and thus just leave `00`. That means typing e.g `1.00` will actually fill `100` into the other amount field. 

I first explored what other similar forms do, and found the [oxfam donate form](https://www.oxfam.org.uk/donate/) to have quite nice UX. Initially it only allows you to type a number or a `.`. After typing a `.`, you can then only type at most two more digits.

I started trying to recreate that UX but the behaviour of our form fields is a little tricky. If you make an update to a field, but the state doesn't change as a result (ie. because you filtered out the newly typed character), the field will still display the newly typed character. This is because the react component isn't rerendered and we're left with the form state (ie the state maintained by the browser) out of sync with our react state. To demonstrate, I captured this interaction with the prod site:

Jul-05-2022 16-22-10

Whilst typing the `g`s, the react state isn't getting updated. This is because the `g`s are being filtered out by the call to `replace(/[^0-9]/g, '')`. Then when I type another `0`, `50ggggggg0.replace(/[^0-9]/g, '')` is `500`, so the state is then updated and we resync the form field with the react state.

Ideally this wouldn't be the behaviour and we could rely on the two states being insync. 

To get a quick fix in, I decided then that the best thing to do would be to let users type whatever they want into the form, and then validate it after. I checked what validation we've already got in place which is [here](https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/helpers/forms/formValidation.ts#L99). Which looks to be everything we need! Hence the final change is only tiny!

### DIFF
--- a/support-frontend/assets/pages/contributions-landing/components/ContributionAmount.tsx
+++ b/support-frontend/assets/pages/contributions-landing/components/ContributionAmount.tsx
@@ -142,7 +142,7 @@ function ContributionAmount(props: PropTypes) {
 						value={`${otherAmount ?? ''}`}
 						onChange={(e) =>
 							props.setOtherAmount({
-								amount: e.target.value.replace(/[^0-9]/g, ''),
+								amount: e.target.value,
 								contributionType: props.contributionType,
 							})
 						}


GHaberis (approved):
Happy to approve this as the change achieves the objective and my comments are not blockers

tompretty:
> I'm wondering whether we should include a `Number.isNaN(parseInt(input, 100))` check in the `amountIsValid` conditions? Although I can see that it'll return `false` if the input was `50gggg0` as `isLargerOrEqual` returns false, but when a non-number input is passed to it, it does feel like the inclusion of an `isNaN` condition is a bit clearer about our expectations of the `input` value.

Yeah 100% agree here - I'll push up that change now too. 

> Also, what validation error will a user see if they type `50gggg0` is it clear characters other than decimals and numbers are forbidden?

So at the moment we just have a single error message that is shown for all errors with the field which is `Please enter an amount between <MIN_AMOUNT> and <MAX_AMOUNT>`. We could definitely improve that, but it might be a slightly bigger refactor as this function is used in a few different places.

tompretty:
> Happy to approve this as the change achieves the objective and my comments are not blockers

Okay nice one! I've just pushed up the change to add the isNaN check, and then I can create a health card around improving the error messages?

PR #4007, opened 2022-07-06
Quantum Metric: Check correct consent state in US/AUS

GHaberis:

I noted that the Quantum Metric script isn't running in the US/AUS, this is because the `getConsentFor` check only returns the consent state for IAB compliant regions, the US/AUS have their own consent state that needs to be interrogated.


GHaberis:
@kenoir actually this PR can be ignored, `getConsentFor` is already returning the correct consent state in the US and AUS.

PR #4030, opened 2022-07-14
Add direct debit feature switch for subs

marialani:
This adds a direct debit feature switch for subscriptions which will allow us to switch off direct debit as a payment method in future if need be.

The corresponding JSON files have been updated in S3 and the new switch is present in the RRCP for all environments.

[**Trello Card**](https://trello.com/c/6JPATMMG/495-put-feature-switches-around-subscription-payment-methods-client-side)


rupertbates (approved):
👍

PR #4032, opened 2022-07-14
Bump jackson-databind to version 2.13.3

marialani:
This PR updates `jackson-databind` to version 2.13.3. In order to do this we also needed to update the `jackson-module-scala` transitive dependency which requires an older version of jackson-databind.

[**Trello Card**](https://trello.com/c/QWsbCtTn/506-update-jackson-databind-dependency-high-vulnerability)

## Why are you doing this?
To mitigate potential security vulnerabilities.


tompretty (approved):
👍

PR #4048, opened 2022-07-19
Set up payments reducer and direct debit slice

i-hardy:

This creates a combined payment reducer for all payment methods, and an included RTK slice for direct debit state. The two async operations- checking the validity of the details and submitting the form- have been migrated to use the [createAsyncThunk](https://redux-toolkit.js.org/usage/usage-with-typescript#createasyncthunk) utility.

[**Trello Card**](https://trello.com/c/Y5WjDH8E)

## Why are you doing this?

This is part of our ongoing migration of our Redux setup to use Redux Toolkit, and represents a blueprint for how we can handle slices for our various other payment methods.


i-hardy:
[🛠 [SPIKE] Redux Toolkit migration - define a unified payment state 🛠](https://trello.com/c/Y5WjDH8E/325-%F0%9F%9B%A0-spike-redux-toolkit-migration-define-a-unified-payment-state-%F0%9F%9B%A0)

support-frontend/assets/helpers/redux/checkout/payment/directDebit/reducer.ts
```diff
@@ -0,0 +1,88 @@
...
+		},
+		setSortCode(state, action: PayloadAction<SortCodeUpdate>) {
+			state.sortCodeArray = state.sortCodeArray.map(
+				(partialSortCode, index) => {
+					if (index === action.payload.index) {
+						return action.payload.partialSortCode;
+					}
+					return partialSortCode;
+				},
+			);
```
tompretty, line 35: Now that we can mutate in place, can we do something like

```ts
state.sortCodeArray.splice(action.payload.index, 1, action.payload.partialSortCode);
```

support-frontend/assets/components/directDebit/directDebitForm/directDebitForm.tsx
```diff
@@ -129,19 +134,19 @@ function DirectDebitForm(props: PropTypes) {
 			<PaymentButton
 				buttonText={props.buttonText}
 				phase={props.phase}
-				onPayClick={props.payDirectDebitClicked}
-				onEditClick={() => props.setDirectDebitFormPhase('entry')}
+				onPayClick={props.confirmAccountDetails}
+				onEditClick={() => props.setPhase('entry')}
 				onConfirmClick={() =>
-					props.confirmDirectDebitClicked(props.onPaymentAuthorisation)
+					props.payWithDirectDebit(props.onPaymentAuthorisation)
```
tompretty, line 140: I think `confirmAccountDetails` and `payWithDirectDebit` are much better and clearer names than `payDirectDebitClicked` and `confirmDirectDebitClicked`. I wonder if we should update the names of the `PaymentButton` props too to align with these new names?

support-frontend/assets/helpers/redux/checkout/payment/directDebit/thunks.ts
```diff
@@ -0,0 +1,81 @@
...
+export const payWithDirectDebit = createAsyncThunk<
+	void,
+	(authorisation: PaymentAuthorisation) => void,
+	{
+		state: SubscriptionsState | ContributionsState;
+		rejectValue: string;
+	}
+>(
+	'directDebit/payWithDirectDebit',
+	function payWithDirectDebit(onPaymentAuthorisation, { getState }) {
```
tompretty, line 70: Currently, the contribs page assumes this method will validate if the recaptcha has been clicked (https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/components/directDebit/directDebitActions.ts#L193). That means right now this change will let you submit the direct debit on contribs without checking the recaptcha.

It's slightly different on the subs direct debit, where the component does that check. We could maybe do that for contribs too? Or move the check here and remove it from the subs component? Either way, seems like a good opportunity to unify some code!

i-hardy: Ah, I missed that change to the old function. I'll take a look at this.

i-hardy: I'm going to move the check inside the component too- the subs checkout doesn't actually use this function, so it wouldn't be helpful to use it here.

PR #4049, opened 2022-07-19
Server side implementation of new Supporter Plus product

rupertbates:
This PR adds in the new Supporter Plus project to the support stack. It is focused on the back end code only (frontend to follow) and is made up of two main code areas:
- Catalog mapping - new objects plus product ids, product rate plan ids and product rate plan charge ids for the new product.
- Support workers logic to create subscriptions in Zuora, send emails and acquisition analytics.

[**Trello Card**](https://trello.com/c/UIEmqp21/530-create-new-support-product)

## Is this an AB test?
- [ ] Yes
- [x] No


support-workers/src/main/scala/com/gu/support/workers/lambdas/SendAcquisitionEvent.scala
```diff
@@ -105,14 +105,4 @@ class SendAcquisitionEvent(serviceProvider: ServiceProvider = ServiceProvider)
     AwsCloudWatchMetricPut(AwsCloudWatchMetricPut.client)(cloudwatchEvent)
   }
 
-  private def maybePromoCode(s: SendThankYouEmailState): Option[PromoCode] = s match {
```
rupertbates, line 108: This was unused

support-models/src/main/scala/com/gu/support/workers/Products.scala
```diff
@@ -29,6 +29,14 @@ case class Contribution(
   override def describe: String = s"$billingPeriod-Contribution-$currency-$amount"
 }
 
+case class SupporterPlus(
+    amount: BigDecimal,
+    currency: Currency,
+    billingPeriod: BillingPeriod,
+) extends ProductType {
+  override def describe: String = s"$billingPeriod-SupporterPlus-$amount-$amount"
```
tomrf1, line 37: should this be `$currency-$amount`?

support-models/src/main/scala/com/gu/support/workers/states/SendThankYouEmailState.scala
```diff
@@ -28,6 +28,14 @@ object SendThankYouEmailState {
       subscriptionNumber: String,
   ) extends SendThankYouEmailState
 
+  case class SendThankYouEmailSupportertPlusState(
```
tomrf1, line 31: typo in name

support-workers/src/main/scala/com/gu/emailservices/SupporterPlusEmailFields.scala
```diff
@@ -0,0 +1,81 @@
...
+      created,
+    ).map { paymentFields =>
+      val fields = List(
+        "EmailAddress" -> state.user.primaryEmailAddress,
+        "created" -> created.toString,
+        "amount" -> state.product.amount.toString,
+        "currency" -> state.product.currency.iso,
+        "edition" -> state.user.billingAddress.country.alpha2,
+        "name" -> state.user.firstName,
+        "product" -> s"${state.product.billingPeriod.toString.toLowerCase}-contribution",
```
tomrf1, line 34: is it supposed to be `contribution` here? Same question for line 37

rupertbates: Well spotted, this is all a bit up in the air at the moment but that was definitely wrong

tomrf1 (approved):
looks good 👍

support-config/src/main/resources/touchpoint.PROD.conf
```diff
@@ -50,6 +50,10 @@ touchpoint.backend.environments {
         monthly.productRatePlanChargeId = "2c92a0fb4edd70c9014edeaa50342192"
         annual.productRatePlanChargeId = "2c92a0fb4edd70c9014edeaa5001218c"
       }
+      supporterplus {
+        monthly.productRatePlanChargeId = "8a12865b8219d9b401822106194e64e3"
+        annual.productRatePlanChargeId = "8a12865b8219d9b40182210618c664c1"
+      }
```
johnduffell, line 56: Not one for this PR, but we're just losing compile safety by putting them in a config file rather than scala code, as we would never change these without recompiling and redeploying the whole thing anyway.

support-frontend/app/controllers/CreateSubscriptionController.scala
```diff
@@ -210,6 +210,12 @@ class CreateSubscriptionController(
           s"gu.contributions.recurring.contrib-timestamp.$billingPeriod" -> DateTime.now.getMillis.toString,
           "gu_recurring_contributor" -> true.toString,
         )
+      case _: SupporterPlus =>
+        List(
+          "gu_digital_subscriber" -> true.toString, // TODO: confirm these
```
johnduffell, line 215: good question - is editions access triggered off this field?

johnduffell: that would be a more important question on the members-data-api side.
Also, this would mean even people contributing below the thresold would get ad free and etc for a day?

support-frontend/app/utils/CheckoutValidationRules.scala
```diff
@@ -33,6 +33,7 @@ object CheckoutValidationRules {
 
   def validate(createSupportWorkersRequest: CreateSupportWorkersRequest): Result =
     (createSupportWorkersRequest.product match {
+      case _: SupporterPlus => PaidProductValidation.passes(createSupportWorkersRequest)
```
johnduffell, line 36: this should be the same as contributions but I notice contributions validation was never switched on!  So this is an improvement.

support-models/src/main/scala/com/gu/support/catalog/Catalog.scala
```diff
@@ -14,12 +14,15 @@ case class Catalog(
...
     Paper.ratePlans(PROD),
     GuardianWeekly.ratePlans(PROD),
+    SupporterPlus.ratePlans(UAT),
     DigitalPack.ratePlans(UAT),
     Paper.ratePlans(UAT),
     GuardianWeekly.ratePlans(UAT),
+    SupporterPlus.ratePlans(SANDBOX),
     DigitalPack.ratePlans(SANDBOX),
     Paper.ratePlans(SANDBOX),
     GuardianWeekly.ratePlans(SANDBOX),
```
johnduffell, line 28: this can be rolled up like this
```suggestion
  lazy val productRatePlansWithPrices: List[ProductRatePlanId] = for {
    product <- List(SupporterPlus, DigitalPack, Paper, GuardianWeekly)
    env <- List(PROD, UAT, SANDBOX)
    plan <- product.ratePlans(env)
  } yield plan.id
```

support-workers/src/main/scala/com/gu/emailservices/FailedContributionEmailFields.scala
```diff
@@ -5,6 +5,9 @@ object FailedEmailFields {
   def contribution(email: String, identityUserId: IdentityUserId): EmailFields =
     failedEmailFields("contribution-failed", email, identityUserId)
 
+  def supporterPlus(email: String, identityUserId: IdentityUserId): EmailFields =
+    failedEmailFields("supporter-plus-failed", email, identityUserId) // TODO: this email doesn't exist!
```
johnduffell, line 9: this is the email id in membership-workflow that is triggered when we end up in the support workers failure state (as opposed to a 24h timeout)

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/SupporterPlusSubcriptionBuilder.scala
```diff
@@ -0,0 +1,54 @@
...
+      ratePlanCharges = List(
+        RatePlanChargeData(
+          RatePlanChargeOverride(
+            ratePlanChargeId,
+            price = state.product.amount, // Pass the amount the user selected into Zuora
+          ),
+        ),
+      ),
+      contractEffectiveDate = todaysDate,
+      contractAcceptanceDate = todaysDate,
```
johnduffell, line 46: no free trial for the new product

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/SupporterPlusSubcriptionBuilder.scala
```diff
@@ -0,0 +1,54 @@
...
+    val ratePlanChargeId = if (state.product.billingPeriod == Monthly) config.monthlyChargeId else config.annualChargeId
+    val todaysDate = dateGenerator.today
+
+    val subscriptionData = subscribeItemBuilder.buildProductSubscription(
+      productRatePlanId = productRatePlanId,
+      ratePlanCharges = List(
+        RatePlanChargeData(
+          RatePlanChargeOverride(
+            ratePlanChargeId,
+            price = state.product.amount, // Pass the amount the user selected into Zuora
```
johnduffell, line 41: where do we make the decsision as to whether the user gets benefits?
We have told them at the checkout whether they will get them, so if we don't store it at the point of acquisition, it will be more difficult to do a "price rise" or change in what we offer on a cohort/new customer only basis.

support-workers/src/main/scala/com/gu/emailservices/SupporterPlusEmailFields.scala
```diff
@@ -0,0 +1,81 @@
...
+    ).map { paymentFields =>
+      val fields = List(
+        "EmailAddress" -> state.user.primaryEmailAddress,
+        "created" -> created.toString,
+        "amount" -> state.product.amount.toString,
+        "currency" -> state.product.currency.iso,
+        "edition" -> state.user.billingAddress.country.alpha2,
+        "name" -> state.user.firstName,
+        "product" -> s"${state.product.billingPeriod.toString.toLowerCase}-supporter-plus",
+      ) ++ paymentFields
```
johnduffell, line 35: do we need to add a field to say whether they should get the "you have benefits" or "you don't have benefits" email?

johnduffell:
interestingly, the refactor of support workers state was only completed for the email and zuora step so far, and this is the first new product added since then (after DS gifting).
Looking at this PR, most of the actual product specific logic is for email and zuora anyway, so we have probably got 75% of the value from 50% of the work.  There's some type safety missing from earlier in the step functions, but the organisation benefits are in the later steps.

Aside from that this PR looks good although I've dropped in a few questions/comments/observations.

support-config/src/main/resources/touchpoint.PROD.conf
rupertbates: Yes I agree, this is how it has been done for  the other products though and it would be quite a big refactor to switch them all over. Worth doing though at some point

support-frontend/scalastyle-config.xml
```diff
@@ -1,103 +0,0 @@
-<scalastyle>
```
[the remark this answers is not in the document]
rupertbates: It was left over from the move from scalastyle to scalafmt

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/SupporterPlusSubcriptionBuilder.scala
rupertbates: Nope

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/SupporterPlusSubcriptionBuilder.scala
rupertbates: Anyone on this product will get benefits, if they are below the benefits threshold they will be on a recurring contribution instead

support-workers/src/main/scala/com/gu/emailservices/SupporterPlusEmailFields.scala
rupertbates: As above, anyone on this product will receive benefits

support-workers/src/main/scala/com/gu/emailservices/FailedContributionEmailFields.scala
rupertbates: Correct

support-models/src/main/scala/com/gu/support/catalog/Catalog.scala
rupertbates: Hmm, requires quite a bit more thinking but does avoid the risk of missing one out so probably a good improvement 👍

support-frontend/app/controllers/CreateSubscriptionController.scala
rupertbates: Anyone on supporter plus is paying over the threshold

PR #4045, opened 2022-07-19
Move recaptcha to second stage of direct debit checkouts

tompretty:
## What are you doing in this PR?

This PR moves the recaptcha widget to the second stage of the contribs and subs direct debit checkouts. This is to prevent an issue we've been seeing with users submitting expired recaptcha tokens. 

I made a slight tweak to the contribs checkout layout. With the error message after the payment buttons in the document, the error message displayed pretty awkwardly.

Screenshot 2022-07-19 at 11 59 34

To avoid having to dig around in the old sass (why are so many things set to `height: 0`?!) I just moved the error message component above the buttons.

Screenshot 2022-07-18 at 16 40 40

[**Trello Card**](https://trello.com/c/W4VzE2UY/516-move-recaptcha-to-confirmation-step-for-direct-debit-payments-contributions-and-subscriptions)

### Contribs

Screenshot 2022-07-19 at 12 02 53

### Subs

Screenshot 2022-07-19 at 11 48 52

PR #4050, opened 2022-07-19
Fix post deploy tests

tompretty:

Fixes the post deploy tests broken by #4045. The recaptcha is now in the second stage of direct debit so we need to update the selenium interactions to handle that.


i-hardy (approved):
👍

PR #4055, opened 2022-07-20
Update GW savings percent figures

i-hardy:

This updates various bits of copy for the savings over base retail price of a GW subscription in different regions.


GHaberis (approved):
👍

PR #4056, opened 2022-07-21
Add subs payment method switches and consolidate payment method selector components

marialani:
This PR adds debit, debit/credit card and paypal feature switches for subscriptions, which will allow us to switch off those payment methods in future if need be. 

The `countryPaymentMethods.ts` file which previously determined whether or not we would show the direct debit payment method option based on the country has been updated to account for whether subs payment method switches are turned on. This file has been renamed to the more generic `supportedPaymentMethods.ts` and the corresponding JSON files for each code environment have been updated in S3.

This also removes the contributions specific `PaymentMethodSelector.tsx` in favour of the subscriptions `paymentMethodSelector.tsx` component in an effort to consolidate more of our codebase and the related subscriptions checkout tests have been updated to account for different potential switch states.

When there is only 1 payment method available on subscriptions checkout the title will read "Payment Method" as opposed to "How would you like to pay?" and the only available payment method will be preselected:

Screenshot 2022-08-01 at 03 37 43

[**Trello Card**](https://trello.com/c/6JPATMMG/495-put-feature-switches-around-remaining-subscription-payment-methods-credit-card-paypal-client-side)

## How to test (locally)
- Pull this branch and start the support site locally
- Navigate to a subscriptions checkout page
- Start the support-admin-console locally
- Turn off a subs payment method switch in the RRCP
- Refresh the subs checkout page
- The payment method you switched off should no longer be available/visible on the subs checkout page


support-frontend/assets/pages/digital-subscription-checkout/components/digitalCheckoutForm.test.tsx
```diff
@@ -38,10 +39,20 @@ function setUpStore(initialState: WithDeliveryCheckoutState) {
...
+	__esModule: true,
+	isSwitchOn: jest.fn(),
+	getSettings: jest.fn(),
+	getGlobal: jest.fn(),
+}));
+
+const mock = (mockFn: unknown) => mockFn as jest.Mock;
+
 describe('Digital checkout form', () => {
 	// Suppress warnings related to our version of Redux and improper JSX
```
rupertbates, line 52: Not related to your PR but I wonder if this comment and suppression is still needed?

support-frontend/assets/pages/digital-subscription-checkout/components/digitalCheckoutForm.tsx
```diff
@@ -198,7 +199,7 @@ function DigitalCheckoutForm(props: PropTypes) {
 					<FormSection title="Address">
 						<BillingAddress countries={countries} />
 					</FormSection>
-					{paymentMethods.length > 1 ? (
+					{paymentMethods.length > 0 ? (
```
rupertbates, line 202: Was this wrong before?

support-frontend/assets/pages/digital-subscription-checkout/components/digitalCheckoutForm.test.tsx
marialani: Good point, will check this out now!

support-frontend/assets/pages/digital-subscription-checkout/components/digitalCheckoutForm.tsx
marialani: Yeah I reckon this was wrong although I suppose technically before we'd never have less than 2 payment methods (we conditionally removed direct debit before but should have always had at least 2)

support-frontend/assets/pages/digital-subscription-checkout/components/digitalCheckoutForm.test.tsx
marialani: We still need to suppress a few warnings here related to improper JSX but I'll update the comment to remove reference to Redux

support-frontend/assets/pages/digital-subscription-checkout/components/digitalCheckoutForm.test.tsx
```diff
@@ -79,6 +90,8 @@ describe('Digital checkout form', () => {
 			client_secret: 'super secret',
 		});
 
+		mock(isSwitchOn).mockImplementation(() => true);
```
i-hardy, line 93: I don't suppose there's any chance of testing some of the different potential switch states? I know this has been a bit of a pain to get behaving so no worries if you think it'd be too complicated.

marialani: This has been a bit of a pain to make happy but I agree it would be great to explicitly test different potential switch states! I'm happy to get in a card and do a follow up PR

support-frontend/assets/pages/digital-subscription-checkout/components/digitalCheckoutForm.tsx
marialani: Ahh I'll add a conditional check on the title that we pass to the `<FormSection>` container for 'How would you like to pay?'

rupertbates: I think that will still not be right though because we'd just show the payment method selector with only one option in it and no title. What we want is to only show the payment method selector if there is more than one payment method, otherwise we should just show that payment method without the user having to select the  radio button for it.

marialani: Ah okay I see, we want to add a new component. I'll do a bit of refactoring around this

rupertbates (approved):
Looks good to me, really well done working through this!

PR #4059, opened 2022-07-22
Refactor subs form validation

tompretty:

Make a few refactors to the subs form validation logic. This was due to a seemingly unrelated piece of work to migrate the PayPal state to an RTK slice causing some errors in this file. 

I noticed that our types were slightly weaker than they needed to be. We have

```ts
type Error<T> = {
	errors: Array<FormError<T>>;
	errorAction: (arg0: any) => Action;
};
```

But the `errorAction` function takes the `errors` array if it gets called, so we should be able to use

```ts
type Error<T> = {
	errors: Array<FormError<T>>;
	errorAction: (errors: Array<FormError<T>>) => Action;
};
```

However, that introduced a new error. What we currently have is an array of 

```ts
type AnyErrorType = Error<AddressFormField> | Error<FormField>;
```

Which we then loop through to dispatch the errors

```ts
allErrors.forEach(({ errors, errorAction }) => {
  dispatch(errorAction(errors));
});
```

I haven't quite figured out why, but at this point typescript no longer knows that the `errors` and `errorAction` have the same generic type. I've created a [typescript playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAYg9nAPAFQHxQLxQN4CgpTAAWAlgHYDmAXFMgNz5QBmArmQMY0AUx51tASkzoAbnBIATBgF8GuUJCgAhAIYAnTLASIAzsDV90AHy1IyLALYAjCGtRz2cMnqgWQqtTQ+a8BXpRoAIiIIABtQuECAGkZWDm5-fj0DSiEMdGxpGNlcXAB6PNpSHSgSErgAaxUQADo63DcPGrj2Lkb1GsSBBycXdrUdL3UAbQBdTWHGXwJCUgCoYLCI6MYCFoS5pP0+NIysxn2Cab9NmgBGAAYYmeY2TigeU6hza1tdnEOofdG5AuUWYCzMqlHRkADkwAa7nUOmacDUAFEVOwiFwuFZ1O9phi1M07uiOl0ZN1cEA) that illustrates this. I'd be interested to hear if anybody has any ideas about this one!

I decided that it was then probably easiest to replace the `errorAction` function with something that was more generic and got access to the `errors` array through its closure. I changed it for a `dispatchErrors` function that just takes a `dispatch` and returns `void`. This also meant I had to replace the declarative building up of the `AnyErrorType[]` array with some imperative code instead. 

That was the main bulk of the structural changes, outside of that, I also added some comment headers to chunk up the file and reorganised it to put the most important functions at the top (i.e the ones that get exported) and moved the helper functions to the bottom.

### DIFF
--- a/support-frontend/assets/helpers/subscriptionsForms/formValidation.ts
+++ b/support-frontend/assets/helpers/subscriptionsForms/formValidation.ts
@@ -21,11 +21,46 @@ import { getFulfilmentOption } from 'helpers/subscriptionsForms/subscriptionChec
 import type { FormError } from 'helpers/subscriptionsForms/validation';
 import { applyCheckoutRules, applyDeliveryRules } from './rules';
 
+// ---- Types ---- //
+
 type Error<T> = {
 	errors: Array<FormError<T>>;
 	dispatchErrors: (dispatch: Dispatch) => void;
 };
 type AnyErrorType = Error<AddressFormField> | Error<FormField>;
+
+// ---- Validation ---- //
+
+export function validateCheckoutForm(
+	dispatch: Dispatch,
+	state: SubscriptionsState,
+): boolean {
+	const allErrors = checkoutValidation(state);
+	dispatchAllErrors(dispatch, allErrors);
+	return allErrors.length === 0;
+}
+
+export function validateWithDeliveryForm(
+	dispatch: Dispatch<Action>,
+	state: SubscriptionsState,
+): boolean {
+	const allErrors = withDeliveryValidation(state);
+	dispatchAllErrors(dispatch, allErrors);
+	return allErrors.length === 0;
+}
+
+export function checkoutFormIsValid(state: SubscriptionsState): boolean {
+	if (state.page.checkoutForm.product.productType === DigitalPack) {
+		return checkoutValidation(state).length === 0;
+	}
+	return withDeliveryValidation(state).length === 0;
+}
+
+export function withDeliveryFormIsValid(state: SubscriptionsState): boolean {
+	return withDeliveryValidation(state).length === 0;
+}
+
+// ---- Helpers ---- //
 
 function checkoutValidation(state: SubscriptionsState): AnyErrorType[] {
 	const errors: AnyErrorType[] = [];
@@ -51,9 +86,6 @@ function checkoutValidation(state: SubscriptionsState): AnyErrorType[] {
 
 	return errors;
 }
-
-const shouldValidateBillingAddress = (fields: FormFields) =>
-	!fields.billingAddressIsSame;
 
 function withDeliveryValidation(state: SubscriptionsState): AnyErrorType[] {
 	const errors: AnyErrorType[] = [];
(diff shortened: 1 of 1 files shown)


support-frontend/assets/helpers/subscriptionsForms/formValidation.ts
```diff
@@ -21,94 +21,118 @@ import { getFulfilmentOption } from 'helpers/subscriptionsForms/subscriptionChec
...
+}
+
+function getPushIfErrors(errors: AnyErrorType[]) {
+	function pushIfErrors(error: AnyErrorType) {
+		if (error.errors.length > 0) {
+			errors.push(error);
+		}
+	}
+	return pushIfErrors;
+}
```
tompretty, line 138: I factored this out because there were lots of repeated blocks checking `length > 0` and then pushing to the errors array. However, I can't quite decide if it's better or it just adds too much complexity

i-hardy: I really like the idea of extracting this re-used condition, but I think it might be a bit of a footgun to have this be causing a side effect- it's not immediately obvious when you're looking at the validation functions that `pushIfErrors` is changing the `errors` array.

It might be simpler to have the condition inside a function that simply returns an array, which may or may not be empty, eg:

```typescript
function getErrorList(error: AnyErrorType) {
	if (error.errors.length > 0) {
		return [error]
	}
	return [];
}

function checkoutValidation() {
	const checkoutErrors = getErrorList({
		errors: applyCheckoutRules(getFormFields(state)),
		dispatchErrors: (dispatch) => dispatch(setFormErrors(checkoutErrors)),
	});
	
	const billingAddressErrors = getErrorList({
		errors: applyBillingAddressRules(
			state.page.checkoutForm.billingAddress.fields,
		),
		dispatchErrors: (dispatch) => dispatch(setFormErrors(checkoutErrors)),
	});

	return [...checkoutErrors, ...billingAddressErrors];
```

tompretty: Yeah agreed the side effects make it really unclear. 

So just a small point there, the `dispatchErrors` function needs a reference to the `errors` array, so the code would have to look like:

```ts
function checkoutValidation(state: SubscriptionsState): AnyErrorType[] {
	const checkoutErrors = applyCheckoutRules(getFormFields(state));
	const checkoutErrorsList = getErrorList({
		errors: checkoutErrors,
		dispatchErrors: (dispatch) => dispatch(setFormErrors(checkoutErrors)),
	});

	const billingAddressErrors = applyBillingAddressRules(
		state.page.checkoutForm.billingAddress.fields,
	);
	const billingAddressErrorsList = getErrorList({
		errors: billingAddressErrors,
		dispatchErrors: (dispatch) =>
			dispatch(setBillingAddressFormErrors(billingAddressErrors)),
	});

	return [...checkoutErrorsList, ...billingAddressErrorsList];
}
```

tompretty: I think this is definitely an improvement. It still feels like were having to work hard to get the code to do what we want here. Perhaps there's a bigger refactor that could really clean things up.

tompretty: Okay, I've pushed that up!

i-hardy (approved):
Looks good!

PR #4064, opened 2022-07-25
Remove unnecessary scala-steward action

GHaberis:

This action was originally set-up in https://github.com/guardian/support-frontend/pull/3531 to run on pull requests raised by the `scala-steward` user from `scala-steward/support-frontend` forks. We now no longer receive these PRs, so this action is never being run, instead we get PRs from `gu-scala-steward-public-repos`. 

We can remove this action as we have various Scala build/test actions that now run on PRs from `gu-scala-steward-public-repos` that cover the tests/validation this action triggered.


rupertbates (approved):
👍

tompretty (approved):
👍

PR #4068, opened 2022-07-26
Make the patrons alarm less sensitive

rupertbates:
The alarm we currently have for the Guardian Patrons Stripe export lambda is too sensitive - it triggers after a single failure, however as it processes all records every time and is scheduled to run every 15 minutes, we only need to worry if it fails multiple times consecutively.

This PR updates the alarm so that it will only trigger after 3 failures in a row.

### DIFF
--- a/cdk/lib/stripe-patrons-data.ts
+++ b/cdk/lib/stripe-patrons-data.ts
@@ -35,7 +35,7 @@ class StripePatronsDataLambda extends GuScheduledLambda {
           alarmDescription: `Triggers if there are errors from ${appName} on ${scope.stage}`,
           snsTopicName: "reader-revenue-dev",
           toleratedErrorPercentage: 1,
-          numberOfMinutesAboveThresholdBeforeAlarm: 1,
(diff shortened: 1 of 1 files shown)


tomrf1 (approved):

tompretty (approved):
Great! 🔥

cdk/lib/stripe-patrons-data.ts
```diff
@@ -35,7 +35,7 @@ class StripePatronsDataLambda extends GuScheduledLambda {
           alarmDescription: `Triggers if there are errors from ${appName} on ${scope.stage}`,
           snsTopicName: "reader-revenue-dev",
           toleratedErrorPercentage: 1,
-          numberOfMinutesAboveThresholdBeforeAlarm: 1,
+          numberOfMinutesAboveThresholdBeforeAlarm: 46, // The lambda runs every 15 mins so alarm if it fails 3 times in a row
```
johnduffell, line 38: are you sure this will work? if it's running every 15 minutes, then it will only be above the threshold for 3 minutes in any 46 minute period and the rest will be missing data points.
I don't know what this CDK translates into but I've had enough fun with cloudwatch to be concerned :)

PR #4070, opened 2022-07-26
Change the Customer Acceptance Date to tomorrow for Recurring Contributions

paulbrown1982:
Changeing the Customer Acceptance Date for Recurring Contributions to tomorrow instead of today.

## Why are you doing this?
Before 26 July 2022 the Customer Acceptance Date for Recurring Contributions was always  the same as the Contract Effective Date (i.e. today). A change was made to push the CAS one day forward so as not to trigger an invoice and payment immediately when making a Recurring Contribution. This will affect the customer experience for users who have no money in their account when purchasing, as they will no longer be notified on the checkout, instead they will be notified by email the day after purchasing.

The reason for making this change relates to a project being undertaken by the Finance team on August 1st 2022 to re-publish revenue schedules from 4-4-5 over to calendar monthly and it requires no invoices be posted or payments or refunds be made during the change window. This change is going live a few days ahead of that date to enable the change to be monitored for any issues downstream.

## Is this an AB test?
- [ ] Yes
- [X] No


rupertbates (approved):
I've updated the catalog in DEV so that the product rate plans charge trigger condition is 'On customer acceptance' and the revenue recognition rule is 'Recognise upon invoicing' to match the settings in PROD.

PR #4077, opened 2022-07-28
Factor out other amount field component and add tests

tompretty:

Factors out a new `ContributionAmountOtherAmountField` component and adds some unit tests. I went back and forth a little bit on the best way to add tests to this component. I was weighing up:

1. Factoring out a simple, presentational component that can be tested very easily (what I did in the end)
2. Just adding tests to the more complex, `ContributionAmount` component. 

Adding the tests to `ContributionAmount` component would have more closely replicated a proper user interaction, but would have required getting things to play nice with redux. 

In the end I decided to factor out a component which would be easy to test - we can just pass in different props and assert it renders in different ways. 

I think we'd also get benefits of some higher level tests too, e.g testing the `ContributionAmount` component or testing the checkout directly. This would definitely be a separate piece of work so can be done as a follow up!

### DIFF
--- a/support-frontend/assets/pages/contributions-landing/components/ContributionAmount.tsx
+++ b/support-frontend/assets/pages/contributions-landing/components/ContributionAmount.tsx
@@ -1,17 +1,8 @@
 // ----- Imports ----- //
-import { css } from '@emotion/react';
-import { space } from '@guardian/source-foundations';
-import { TextInput } from '@guardian/source-react-components';
 import type { ConnectedProps } from 'react-redux';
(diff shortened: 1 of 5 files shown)


tompretty:
[Improve the "other amount" field's client-side validation error messages on contributions checkout page](https://trello.com/c/RlsDrD2o/525-improve-the-other-amount-fields-client-side-validation-error-messages-on-contributions-checkout-page)

i-hardy (approved):
This looks good!

PR #4078, opened 2022-07-28
disable DatePicker snapshot tests

tompretty:

Disable `DatePicker` snapshot tests. We should be able to add the tests back if we factor out a presentational component that doesn't have any hardcoded references to `Date.now`.


i-hardy (approved):
👍

PR #4081, opened 2022-07-29
Create payment method slice and make recaptcha usage more consistent across checkouts

i-hardy:

- Refactoring the Stripe card components for both subscriptions and contributions to use the `useRecaptchaV2` hook, the same as the direct debit components
- Using the recaptcha reducer in the contributions Stripe card form
- Adding a very simple Redux slice to manage the user's selected payment method, making it easier to perform side effects when the payment method selection changes or where other actions should cause it to be reset

[**Trello Card**](https://trello.com/c/LfEfXZGq)

## Why are you doing this?

This is to smooth out different behaviour around recaptcha before building a new Stripe card slice to handle payment information. Adding the slice for selected payment method as part of this work made it much easier to handle resetting the recaptcha data if the user switches payment methods.

I have **not** changed the subscriptions Stripe component to use the recaptcha reducer. Currently this component is not connected to the store and it seemed like it would only add complexity, especially as we do not require a recaptcha token for anything other than creating a Stripe payment intent, which is done instantly and within the component. The equivalent component in the contributions checkout was already connected to the store, this just changes the bit of the state it's concerned with.

I may revisit this when working directly on the new Stripe reducer; alternatively I think it might be good to encapsulate more of the repeated logic around recaptcha verification, token use and expiry into dedicated components. Either way it felt out of scope for this work.


i-hardy:
[🛠 Redux Toolkit migration - unify use of recaptcha across payment methods and checkouts 🛠](https://trello.com/c/LfEfXZGq/574-%F0%9F%9B%A0-redux-toolkit-migration-unify-use-of-recaptcha-across-payment-methods-and-checkouts-%F0%9F%9B%A0)

support-frontend/assets/helpers/redux/checkout/payment/contributionsSideEffects.ts
```diff
@@ -20,17 +21,17 @@ import {
...
-		effect(_action, listenerApi) {
+		effect(action, listenerApi) {
+			if (setPaymentMethod.match(action)) {
+				const paymentMethod = action.payload;
+				const { payPal } = listenerApi.getState().page.checkoutForm.payment;
+				storage.setSession('selectedPaymentMethod', paymentMethod);
+				if (paymentMethod === PayPal && !payPal.hasBegunLoading) {
+					void listenerApi.dispatch(loadPayPalExpressSdk());
+				}
+			}
```
tompretty, line 34: could this be in a separate call to `startListening`? eg


```ts
startListening({
		action: setPaymentMethod,
		effect(action, listenerApi) {
				const paymentMethod = action.payload;
				const { payPal } = listenerApi.getState().page.checkoutForm.payment;
				storage.setSession('selectedPaymentMethod', paymentMethod);
				if (paymentMethod === PayPal && !payPal.hasBegunLoading) {
					void listenerApi.dispatch(loadPayPalExpressSdk());
				}
			}});
```

i-hardy: I think it's preferable to limit the amount of listeners we create, especially ones that will respond to the same actions, but it might be better to pull these effects out into separate functions.

marialani:
This looks good but may need a bit of refactoring after this PR: https://github.com/guardian/support-frontend/pull/4056 where we've removed the contrib payment method selector component

marialani (approved):
LGTM 👍

PR #4084, opened 2022-08-01
Payment API CDK migration - Phase 1

tompretty:

In this phase, we introduce CDK to the payment-api project. At this stage we're just using the `CfnInclude` construct to load the existing cloudformation template. This allows us to get all the plumbing in place (eg generating the template during CI and pointing riffraff at the new location) without also generating new versions of all of the resources. In phase 2, we will spin up a duplicate set of all resources using proper cdk/gu-cdk constructs/patterns. When we are happy this new set of resources is running as expected, we can redirect traffic to the new resources. In phase 3, after the old resources are no longer receiving traffic, we can tear them down.

### DIFF
--- a/cdk/lib/payment-api.test.ts
+++ b/cdk/lib/payment-api.test.ts
@@ -1,0 +1,16 @@
+import "@aws-cdk/assert/jest";
+import { SynthUtils } from "@aws-cdk/assert";
+import { App } from "@aws-cdk/core";
+import { PaymentApi } from "./payment-api";
+
+describe("The Payment API stack", () => {
+  it("matches the snapshot", () => {
+    const app = new App();
+    const stack = new PaymentApi(app, "Frontend-PROD", {
+      stack: "support",
+      stage: "PROD",
+    });
+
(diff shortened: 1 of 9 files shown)


cdk/lib/payment-api.test.ts
```diff
@@ -0,0 +1,16 @@
+import "@aws-cdk/assert/jest";
+import { SynthUtils } from "@aws-cdk/assert";
+import { App } from "@aws-cdk/core";
+import { PaymentApi } from "./payment-api";
+
+describe("The Frontend stack", () => {
```
rupertbates, line 6: Is 'Frontend' right here?

rupertbates (approved):
Just one question

cdk/lib/payment-api.test.ts
tompretty: Good spot! Should be `The Payment API stack`

tompretty:
[Payment API: GU CDK migration - phase 1](https://trello.com/c/52FJ3puY/588-payment-api-gu-cdk-migration-phase-1)

PR #4094, opened 2022-08-03
Fix patrons export

rupertbates:
There  was a bug with the previous implementation  of this function which meant that it was returning after the first future completed rather than when there were no more records available, this is now fixed.

### DIFF
--- a/stripe-patrons-data/src/main/scala/com/gu/patrons/services/StripeSubscriptionsProcessor.scala
+++ b/stripe-patrons-data/src/main/scala/com/gu/patrons/services/StripeSubscriptionsProcessor.scala
@@ -16,18 +16,19 @@ class StripeSubscriptionsProcessor(
 ) {
 
   def processSubscriptions(pageSize: Int) =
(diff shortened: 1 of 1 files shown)


i-hardy (approved):

stripe-patrons-data/src/main/scala/com/gu/patrons/services/StripeSubscriptionsProcessor.scala
```diff
@@ -16,18 +16,19 @@ class StripeSubscriptionsProcessor(
 ) {
 
   def processSubscriptions(pageSize: Int) =
-    processFutureResponse(stripeService.getSubscriptions(pageSize))
+    processFutureResponse(pageSize, None)
 
-  final def processFutureResponse(eventualResponse: Future[StripeSubscriptionsResponse]): Future[Unit] = {
-    eventualResponse.flatMap { response =>
-      processSubs(response.data).map(_ =>
```
johnduffell, line 23: turns out the issue this should have been flatMap, not map, and so the Futures were not combined

PR #4096, opened 2022-08-03
Wait for sub processing to complete in StripeSubscriptionsProcessor

rupertbates:
#4094 had another bug in it where we wouldn't wait for the `processSubs` method  to complete before recursing, this PR fixes that.

### DIFF
--- a/stripe-patrons-data/src/main/scala/com/gu/patrons/services/StripeSubscriptionsProcessor.scala
+++ b/stripe-patrons-data/src/main/scala/com/gu/patrons/services/StripeSubscriptionsProcessor.scala
@@ -23,11 +23,12 @@ class StripeSubscriptionsProcessor(
       stripeService
         .getSubscriptions(pageSize, startingAfterId)
         .flatMap { response =>
-          processSubs(response.data)
-          if (response.hasMore)
-            processFutureResponse(pageSize, Some(response.data.last.id))
-          else
-            Future.successful(())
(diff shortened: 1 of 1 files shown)


johnduffell (approved):
that was a tough one!

stripe-patrons-data/src/main/scala/com/gu/patrons/services/StripeSubscriptionsProcessor.scala
```diff
@@ -18,19 +18,16 @@ class StripeSubscriptionsProcessor(
...
+  final def processFutureResponse(pageSize: Int, startingAfterId: Option[String]): Future[Unit] =
+    for {
+      response <- stripeService.getSubscriptions(pageSize, startingAfterId)
+      _ <- processSubs(response.data)
+      result <-
+        if (response.hasMore)
+          processFutureResponse(pageSize, Some(response.data.last.id))
+        else
+          Future.successful(())
+    } yield result
```
johnduffell, line 30:
```suggestion
 def processSubscriptions(pageSize: Int, startingAfterId: Option[String] = None): Future[Unit] =
    for {
      response <- stripeService.getSubscriptions(pageSize, startingAfterId)
      _ <- processSubs(response.data)
      _ <-
        if (response.hasMore)
          processSubscriptions(pageSize, Some(response.data.last.id))
        else
          Future.successful(())
    } yield ()
```

PR #4101, opened 2022-08-04
Payment API CDK migration phase 2

tompretty:

In this phase we use cdk to create a 'dual stack' of all of our infrastructure. We retain the `CfnInclude` construct to directly import our original cloudformation template but also recreate all of our resources directly with cdk (mainly with the `GuPlayApp` construct). This means we will have a second load balancer and autoscaling group etc. We can deploy this to production and then, when we're happy, we can direct traffic at the new load balancer. After we've redirected traffic, we can then do a follow up to drop the `CfnInclude` and delete the old cloudformation template (phase 3).

To deploy this we need to

- Manually deploy the cloudformation change (it adds new parameters so riffraff will fail)
- Test the newly created resources
- Merge PR and wait for build and deploy
- Point traffic at the new load balancer

## TODO

- [ ] test in CODE (**requires manual cloudformation deploy**)
- [ ] delete old cfn include to remove old resources - this will need a manual deploy as well


cdk/lib/payment-api.ts
```diff
@@ -1,22 +1,409 @@
...
+        KinesisStreamArn: kinesisStreamArn,
+        ContributionsStoreSqsQueueCodeArn: contributionsStoreSqsQueueCodeArn,
+        ContributionsStoreSqsQueueProdArn: contributionsStoreSqsQueueProdArn,
+        SqsKmsArn: sqsKmsArn,
+      },
+    });
+
+    // TODO: Should these remain as cloudformation parameters?
+    const projectName = "payment-api";
+    const projectVersion = "0.1";
```
tompretty, line 102: Currently these are cloudformation parameters, but I'm not sure how much value that adds, so we've just hardcoded them for now.

cdk/lib/payment-api.ts
```diff
@@ -1,22 +1,409 @@
...
+    aws --region ${this.region} s3 cp s3://membership-dist/${this.stack}/${this.stage}/${app}/${projectName}_${projectVersion}_all.deb /tmp
+    dpkg -i /tmp/${projectName}_${projectVersion}_all.deb
+    /opt/cloudwatch-logs/configure-logs application ${this.stack} ${this.stage} ${app} /var/log/${app}/application.log`;
+
+    const playApp = new GuPlayApp(this, {
+      app: "payment-api",
+      access: { scope: AccessScope.PUBLIC },
+      certificateProps: {
+        domainName: props.domainName,
+        hostedZoneId: "Z1E4V12LQGXFEC",
```
tompretty, line 121: We should check with DevEx/somebody that understands how certificates work if this is correct. I can see the current template has a [certificate arn](https://github.com/guardian/support-frontend/blob/main/support-payment-api/src/main/resources/cloud-formation.yaml#L91) as a param

cdk/lib/payment-api.ts
```diff
@@ -1,22 +1,409 @@
...
+      roleConfiguration: {
+        additionalPolicies: [
+          new GuAllowPolicy(this, "CloudwatchLogs", {
+            actions: [
+              "logs:CreateLogGroup",
+              "logs:CreateLogStream",
+              "logs:PutLogEvents",
+            ],
+            // TODO: Should we use a more specific resource here?
+            resources: ["arn:aws:logs:*:*:*"],
```
tompretty, line 136: The current template actually creates a specific log group and then just gives permisions for that specific log group. We could do that in CDK but that would require creating a new log group which I'm not sure we want to do. For now, we've just given access to all log groups

cdk/lib/payment-api.ts
```diff
@@ -1,22 +1,409 @@
...
+
+    // TODO: remove this tag after the migration
+    const playAppAsg = playApp.autoScalingGroup;
+    Tags.of(playAppAsg).add("gu:riffraff:new-asg", "true");
+
+    // ---- Alarms ---- //
+
+    new GuAlarm(this, "NoHealthyInstancesAlarm", {
+      app,
+      alarmName: `[CDK] No healthy instances for ${app} in ${this.stage}`,
```
tompretty, line 190: For now I've added `[CDK]` to the start of all of the alarm names. That's because these need to be different from the ones that are defined in the existing cloudformation template. After phase 3 (deleting the existing resources), we should be able to drop the prefix. We did the same thing with [support-frontend](https://github.com/guardian/support-frontend/pull/3639) too

cdk/lib/payment-api.ts
```diff
@@ -1,22 +1,409 @@
...
+          LoadBalancer: playApp.loadBalancer.loadBalancerFullName,
+          TargetGroup: playApp.targetGroup.targetGroupFullName,
+        },
+        statistic: "Sum",
+        period: Duration.seconds(60),
+      }),
+      snsTopicName: "contributions-dev",
+    });
+
+    new GuAlarm(this, "NoPaypalPaymentsInTwoHours247Alarm", {
```
tompretty, line 228: An unfortunate side effect of dual stacking is that we will get all of these alarms twice. The original alarm in the cloudformation and the ones in CDK will mean we receive two emails every time an alarm goes off! I suppose that's good motivation to get phase 3 done!

tompretty:
[Payment API: GU CDK migration - phase 2](https://trello.com/c/ixflyjr8/592-payment-api-gu-cdk-migration-phase-2)

rupertbates (approved):
👍 lgtm

michaelbjacobson:
Closing this as it has been superseded by PR #4269.

michaelbjacobson:
Closing this as it has been superseded by PR #4269.

PR #4109, opened 2022-08-09
Add tests for currency localisation

michaelbjacobson:
This PR adds a new test to validate that, in the Guardian Weekly checkout, the prices displayed are those for the _delivery_ country, _not_ the billing country.

### DIFF
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=1.6.2
+sbt.version=1.7.1


michaelbjacobson:
[Write tests to verify checkout price display behaviour when setting delivery and billing addresses in different countries](https://trello.com/c/cFaDzJzU/528-write-tests-to-verify-checkout-price-display-behaviour-when-setting-delivery-and-billing-addresses-in-different-countries)

PR #4112, opened 2022-08-10
Add test for guardian weekly currency internationalisation

michaelbjacobson:
This PR adds a new test to validate that, in the Guardian Weekly checkout, the prices displayed are those for the delivery country, not the billing country.


michaelbjacobson:
[Write tests to verify checkout price display behaviour when setting delivery and billing addresses in different countries](https://trello.com/c/cFaDzJzU/528-write-tests-to-verify-checkout-price-display-behaviour-when-setting-delivery-and-billing-addresses-in-different-countries)

support-frontend/assets/pages/weekly-subscription-checkout/components/weeklyCheckoutForm.test.tsx
```diff
@@ -1,44 +1,70 @@
...
+jest.mock('helpers/globalsAndSwitches/globals', () => {
+	const actualGlobalsAndSwitches = jest.requireActual(
+		'helpers/globalsAndSwitches/globals',
+	);
 
+	return {
+		...actualGlobalsAndSwitches,
+		isSwitchOn: jest.fn(),
+	};
+});
```
i-hardy, line 52: Just wondering if it's worth pulling this out into a separate file like the Stripe mock, to make it reusable.

i-hardy (approved):
Looks good, just a small comment on the `globalsAndSwitches` mock.

support-frontend/assets/pages/weekly-subscription-checkout/components/weeklyCheckoutForm.test.tsx
michaelbjacobson: Good shout! I'm off tomorrow, but could do this on Monday. Probably makes sense to do it in a separate PR though, so this can get merged sooner rather than later.

i-hardy: Yeah that makes sense 👍

PR #4117, opened 2022-08-11
Add new postcodes to valid Home Delivery postcode list

GHaberis:

We can now deliver Newspapers to the following postcode districts:

AL2
GU1
GU2
GU21
GU22
EN8
IG4
KT22
N1C

This PR adds these postcode prefixes to the list of valid postcodes, which are defined in the code in `CheckoutValidationRules.scala`. The postcodes are split over two lists in the Scala file `M25_POSTCODE_OLD_PREFIXES` and `M25_NEW_PREFIXES`, i'm not sure why this is, I looked at the history and couldn't see a reason (@johnduffell any idea?). Anyhow these 2 lists are concatenated and exposed to the client on `window.guardian.homeDeliveryPostcodes`, and we validate postcodes entered in the client using this list.

I've tested some of the new postcode prefixes before and after the change and can confirm adding them to the lists above means they now pass validation.

There's also a small change to the wording to remove the "(within M25)" wording.


GHaberis:
[Add additional postcodes for direct delivery service](https://trello.com/c/qvu6HUaU/527-add-additional-postcodes-for-direct-delivery-service)

i-hardy (approved):
👍

johnduffell:
@GHaberis good question I did a bit of github archaeology and it looks like @rupertbates asked that question at the time and there was no answer recorded
https://github.com/guardian/support-frontend/pull/2263/files/6107f5744ef6f0bbf28595de85f9f79d856ec7c8#r352707603

It looks like the resource used for the list of postcodes is here: https://drive.google.com/drive/u/0/folders/129MbqLqrMDueSQYbcfZMJqAHmbukKdR3

I would say just combine one list with all of them, as per the latest info from the supplier.

PR #4134, opened 2022-08-18
Re-introduce Guardian Weekly Quarterly for US/CA

paul-daniel-dempsey:
This PR for Guardian Weekly enabling Quarterly billing in all regions.

[**Trello Card**](https://trello.com/c/Z7iBPyE0/523-re-enable-quarterly-billing-option-for-gw-acquisitions)

## Why are you doing this?
Screenshot 2022-08-22 at 16 34 19

Screenshot 2022-08-22 at 16 31 29

Screenshot 2022-08-22 at 16 33 28

Screenshot 2022-08-22 at 16 35 51

### DIFF
--- a/support-frontend/assets/pages/weekly-subscription-checkout/components/weeklyCheckoutForm.tsx
+++ b/support-frontend/assets/pages/weekly-subscription-checkout/components/weeklyCheckoutForm.tsx
@@ -325,7 +325,6 @@ function WeeklyCheckoutForm(props: PropTypes) {
 						onChange={(billingPeriod) => props.setBillingPeriod(billingPeriod)}
 						billingPeriods={weeklyBillingPeriods(
 							props.participations.sixForSixSuppression !== 'variant',
-							props.deliveryCountry,
 						)}
 						pricingCountry={props.deliveryCountry}
 						productPrices={props.productPrices}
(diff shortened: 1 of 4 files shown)


paul-daniel-dempsey (commented):
Ready for review

support-frontend/assets/pages/weekly-subscription-checkout/components/weeklyCheckoutForm.tsx
```diff
@@ -325,6 +325,7 @@ function WeeklyCheckoutForm(props: PropTypes) {
 						onChange={(billingPeriod) => props.setBillingPeriod(billingPeriod)}
 						billingPeriods={weeklyBillingPeriods(
 							props.participations.sixForSixSuppression !== 'variant',
+							props.deliveryCountry === 'US' || props.deliveryCountry === 'CA',
```
GHaberis, line 328: We have this `props.deliveryCountry === 'US' || props.deliveryCountry === 'CA'` (or similar) logic in 2 places: `weeklyCheckoutForm.tsx` and `weeklyProductPrices.tsx`. A potential issue with this is if we want to add quarterly to another country we'd need to know to update 2 places in the code. 

As an alternative we could move this logic into the `weeklyBillingPeriods` function in `billingPeriods.ts`, we'd need to change the signature of the function to take an `IsoCountry` as it's 2nd argument rather than a `boolean`. This means there's only one place in the code we'd need to update if we wanted to introduce/remove quarterly to/from other countries in the future.

paul-daniel-dempsey:
Yep looks like regions are now all other than US+CAN, thanks

paul-daniel-dempsey:
Ok that obviously didn't review pass the branch, sorry George!

paul-daniel-dempsey (commented):
Yep looks like regions are now all other than US+CAN, thanks!

rupertbates (approved):
👍

PR #4125, opened 2022-08-16
update support-i18n to release with scala 2.13 and 3.1

johnduffell:
This PR updates support-i18n to release for scala 2.13 and 3.1., and releases v0.14

It didn't even build with scala 2.11, and it had previously been released for scala 2.13 (for support-service-lambdas)

We now use scala 3 for new product-move-api, and even though scala 3 can use scala 2.13 libraries, it does make sense to release scala 3 libs where possible.

I have done a release to confirm my changes work, so this PR is to review what's already been done!  Release is visisble here https://mvnrepository.com/artifact/com.gu/support-internationalisation_3/0.14

---

I did have a go at making it a scala 3 only module (see https://github.com/guardian/support-frontend/compare/jd-support-i18n-scala-3...jd-support-i18n-scala-3-only ), but there seems to be an issue with scalatest versions that stops the tests from compiling as follows:
```
[IJ]support-workers/Test/compile
[info] compiling 43 Scala sources and 1 Java source to /Users/john_duffell/code/support-frontend/support-workers/target/scala-2.13/test-classes ...
[error] Unsupported Scala 3 generic tuple type scala.Tuple in bounds of type MirroredElemTypes; found in  scala.deriving.Mirror.<refinement>.
[error] one error found
[error] (Test / compileIncremental) Compilation failed
[error] Total time: 18 s, completed 16-Aug-2022 10:26:45
```

PR #4136, opened 2022-08-18
upgrade support-internationalisation to scala 3

johnduffell:
This PR updates support-i18n to release for scala 2.13 and 3.1., and releases v0.14+0.15

It didn't even build with scala 2.11, and it had previously been released for scala 2.13 (for support-service-lambdas)

We now use scala 3 for new product-move-api, and even though scala 3 can use scala 2.13 libraries, it does make sense to release scala 3 libs where possible if we're ever going to use scala 3.

I have done a release to confirm my changes work, so this PR is to review what's already been done! Release is visisble here https://mvnrepository.com/artifact/com.gu/support-internationalisation_3/0.15

(Supersedes https://github.com/guardian/support-frontend/pull/4125 )

This PR does mean that it's the first thing in this mono repo to really use scala 3, but there's no reason why more things couldn't use it.  The only issue would be in the play modules, because play framework doesn't support scala 3 yet.

Of course I haven't really made any code changes to take advantage of scala 3 syntax or features, but that's an opportunity for future.  It is still cross compiled but I don't know whether we need to maintain the 2.13 version?  Membership-common is already on 2.13 so could use the 3 version via tasty-reader.

There's a chance that this might trigger this bug in intellij's scala version determiner.  I'm not sure if/when it's going to be fixed. It means intellij built in compiler tries to compile some scala 2 as scala 3 (although SBT has no issue). https://youtrack.jetbrains.com/issue/SCL-18866/sbt-module-can-wrongly-depend-on-scala-sdk-instead-of-normal-scala-library-in-multi-module-projects

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -308,6 +308,7 @@ lazy val `acquisitions-firehose-transformer` = (project in file("support-lambdas
   .disablePlugins(ReleasePlugin, SbtPgp, Sonatype)
   .settings(
     scalafmtSettings,
+    scalacOptions += "-Ytasty-reader",
     libraryDependencies ++= commonDependencies,
     mergeStrategySettings,
   )
@@ -319,6 +320,7 @@ lazy val `acquisition-events-api` = (project in file("support-lambdas/acquisitio
   .disablePlugins(ReleasePlugin, SbtPgp, Sonatype)
   .settings(
     scalafmtSettings,
+    scalacOptions += "-Ytasty-reader",
     libraryDependencies ++= commonDependencies,
     mergeStrategySettings,
   )
--- a/build.sbt
+++ b/build.sbt
@@ -87,7 +87,7 @@ lazy val releaseSettings = Seq(
 lazy val commonDependencies = Seq(
   "com.typesafe" % "config" % "1.4.2",
   scalatest % "test",
-  "com.typesafe.scala-logging" %% "scala-logging" % "3.9.5",
+  "com.typesafe.scala-logging" % "scala-logging_2.13" % "3.9.5",
 )
 
 lazy val root = (project in file("."))
@@ -128,6 +128,7 @@ lazy val `support-frontend` = (project in file("support-frontend"))
     buildInfoKeys := BuildInfoSettings.buildInfoKeys,
     buildInfoPackage := "app",
     buildInfoOptions += BuildInfoOption.ToMap,
+    scalacOptions += "-Ytasty-reader",
     scalafmtSettings,
   )
   .dependsOn(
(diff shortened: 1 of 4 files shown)


build.sbt
```diff
@@ -87,7 +87,7 @@ lazy val releaseSettings = Seq(
 lazy val commonDependencies = Seq(
   "com.typesafe" % "config" % "1.4.2",
   scalatest % "test",
-  "com.typesafe.scala-logging" %% "scala-logging" % "3.9.5",
+  "com.typesafe.scala-logging" % "scala-logging_2.13" % "3.9.5",
```
johnduffell, line 90: hard code to the 2.13 version because it would cause a binary incompatibility if we compile some modules against each one.  Ideally we would use the scala 3 version throughout but I thought I'd keep the existing status quo/minimal change

build.sbt
```diff
@@ -128,6 +128,7 @@ lazy val `support-frontend` = (project in file("support-frontend"))
     buildInfoKeys := BuildInfoSettings.buildInfoKeys,
     buildInfoPackage := "app",
     buildInfoOptions += BuildInfoOption.ToMap,
+    scalacOptions += "-Ytasty-reader",
```
johnduffell, line 131: needed to tell scala 2.13 that it is allowed to read scala 3 `.tasty` files instead of scala 2.13 "pickles" alongside the `.class` files to understand the code it's compiling against.

support-internationalisation/src/test/scala/com/gu/i18n/CountryGroupTest.scala
```diff
@@ -1,11 +1,11 @@
...
 
 import com.gu.i18n.Currency._
-import org.scalatest.Inspectors
-import org.scalatest.flatspec.AsyncFlatSpec
+import org.scalatest._
+import org.scalatest.flatspec.AnyFlatSpec
 import org.scalatest.matchers.should.Matchers
 
-class CountryGroupTest extends AsyncFlatSpec with Matchers with Inspectors {
+class CountryGroupTest extends AnyFlatSpec with Matchers with Inspectors {
```
johnduffell, line 8: this was incorrect but the scala 2.13 compiler didn't notice

rupertbates (approved):
This makes me a bit nervous but I guess we need to start somewhere!

build.sbt
rtyley: ah, that makes sense - we're saying that the main `support-frontend` project will stay on Scala 2.13 for now, but by adding the [`-Ytasty-reader`](https://docs.scala-lang.org/scala3/guides/migration/compatibility-classpath.html#a-scala-213-module-can-depend-on-a-scala-3-artifact) flag, it's allowed to depend on the `support-internationalisation` module, which is using Scala 3.

PR #4142, opened 2022-08-23
Add Nav component - checkout MVP

marialani:

This PR adds a `Nav` component for use on the new Guardian supporter checkout. This component will only be displayed on screen sizes desktop and above.

[**Trello Card**](https://trello.com/c/tKBalS2s/638-checkout-mvp-build-new-nav-component)

## Screenshots
Desktop
United Kingdom £

LeftCol
United Kingdom £

Wide
United v Kingdom £

### DIFF
--- a/support-frontend/assets/components/nav/nav.tsx
+++ b/support-frontend/assets/components/nav/nav.tsx
@@ -1,5 +1,5 @@
 import { css } from '@emotion/react';
-import { between, brand, from } from '@guardian/source-foundations';
+import { brand, from, space } from '@guardian/source-foundations';
 import { Column, Columns } from '@guardian/source-react-components';
 import CountryGroupSwitcher from 'components/countryGroupSwitcher/countryGroupSwitcher';
 import { Container } from 'components/layout/container';
@@ -13,21 +13,16 @@ const container = css`
 	}
 `;
 
-const divider = css`
-	width: 420px;
+const col1CssOverrides = css`
 	height: 28px;
 	border-right: 1px solid ${brand[600]};
-	margin-right: 11px;
-
-	${between.leftCol.and.wide} {
-		margin-right: 7px;
-	}
+	margin-right: 0;
 `;
 
-const columnCssOverrides = css`
+const col2CssOverrides = css`
 	min-height: 42px;
 	display: flex;
-	align-items: center;
+	padding: ${space[2]}px;
 `;
 
 interface NavProps {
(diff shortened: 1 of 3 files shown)


support-frontend/assets/components/nav/nav.tsx
```diff
@@ -0,0 +1,77 @@
...
+}
+
+function Nav({
+	countryGroupIds,
+	selectedCountryGroup,
+	subPath,
+}: NavProps): JSX.Element {
+	return (
+		<nav css={container}>
+			<span css={gutter('right')} />
```
i-hardy, line 63: Just wondering if we could achieve these gutters using the [Source columns](https://guardian.github.io/source/?path=/docs/packages-source-react-components-columns--playground)- it would probably help to avoid any misalignment issues with the rest of the page.

support-frontend/stories/nav/Nav.stories.tsx
```diff
@@ -0,0 +1,73 @@
...
+	Canada,
+	EURCountries,
+	GBPCountries,
+	International,
+	NZDCountries,
+	UnitedStates,
+} from 'helpers/internationalisation/countryGroup';
+
+export default {
+	title: 'SFD/Nav',
```
i-hardy, line 16: This component should probably go under the 'Checkout' or 'Checkout Layout' category

support-frontend/assets/components/nav/nav.tsx
```diff
@@ -0,0 +1,68 @@
...
+	${between.leftCol.and.wide} {
+		margin-right: 7px;
+	}
+`;
+
+const columnCssOverrides = css`
+	min-height: 42px;
+	display: flex;
+	align-items: center;
+`;
```
i-hardy, line 31:
```suggestion
const columnCssOverrides = css`
	min-height: 42px;
	display: flex;
	border-left: 1px solid ${brand[600]};
	padding: ${space[2]}px;
`;
```
I think it'll probably be simpler to put the border on the left of the main column and offset it from the country switcher with padding. Also better to use the Source units where the spacing in Figma is only off from them by a pixel or so- it helps us keep things consistent.

i-hardy (changes requested):
We should definitely be using Source columns for the spacer on the left too, as this will keep things consistent.

support-frontend/assets/components/nav/nav.tsx
marialani: The "divider" Column has a fixed height so we can have the border that doesnt stretch the full height of the container. I've made some updates but have added a cssOverride to the divider Column

i-hardy (approved):

PR #4156, opened 2022-08-26
Set up bare-bones new checkout page behind the newProduct A/B test

i-hardy:

- Setting up a new checkout/landing page for the new supporter product, to be rendered to users in the variant of the ~`newProduct`~ `supporterPlus` A/B test
- Extracting the existing contributions checkout page component into a separate file, for clarity.
- Creating a new `PageScaffold` component to replace the existing `Page` Component, with the Source CSS reset and no reliance on Sass

[**Trello Card**](https://trello.com/c/uekqnvaI)

## Why are you doing this?

This is the base for creating the new checkout page.

## Screenshots

Screenshot 2022-08-26 at 15-54-06 Support the Guardian Make a Contribution

### DIFF
--- a/support-frontend/assets/components/page/emotionPage.tsx
+++ b/support-frontend/assets/components/page/emotionPage.tsx
@@ -1,0 +1,32 @@
+import { css, Global } from '@emotion/react';
+import { FocusStyleManager, resets } from '@guardian/source-foundations';
+import type { ReactNode } from 'react';
+import CsrBanner from 'components/csr/csrBanner';
+import { useHashLink } from 'helpers/customHooks/useHashLink';
+
+type EmotionPageProps = {
+	id: string;
+	header?: ReactNode;
+	footer?: ReactNode;
+	children: ReactNode;
+};
+
+FocusStyleManager.onlyShowFocusOnTabs();
+
+export default function EmotionPage(props: EmotionPageProps): JSX.Element {
+	useHashLink();
+
+	return (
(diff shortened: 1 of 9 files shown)


i-hardy:
[Set up new MVP checkout page behind 0% A/B test](https://trello.com/c/uekqnvaI/645-set-up-new-mvp-checkout-page-behind-0-a-b-test)

GHaberis:
We're currently using the `newProduct` AB Test to release changes to the site and existing checkout in preparation for V1, we've been putting all changes behind a `participations.newProduct === 'variant'` conditional. This has been fine for now as the test is disabled and it's given us a useful mechanism for forcing changes to appear by appending the URL fragment "#ab-newProduct=variant" to page URLs. Based on this approach I imagined we'd release V1 by enabling the `newProduct` AB test and putting 100% of users into the "variant". 

Because of this I wonder whether we want to set-up a new AB test for V2 which will pitch V1 (control) vs V2 (variant)? If so, we should therefore put the new checkout behind a conditional that only shows it to the variant in this new AB test?

@rupertbates @i-hardy would be interested to hear your thoughts on this?

i-hardy:
That's a good point, I hadn't considered that we'll need to launch a lot of stuff under the `newProduct` A/B test before this is ready to go live. I'm happy to just move this to a new test.

GHaberis (approved):
Looks good 👍

PR #4157, opened 2022-08-26
Remove email max-length validation message

GHaberis:

Email should not have a max-length validation on it, this rule was incorrectly added when adding max-length validation to first and last name.


i-hardy (approved):
👍

GHaberis:
@i-hardy I've just reintroduced the max-length validation but have set it to 80 instead of 40, as this is the actual max-length Salesforce allows: https://developer.salesforce.com/forums/?id=9062I000000gF6mQAE

PR #4166, opened 2022-08-30
Work out promo code from delivery address not billing address

rupertbates:
When we work out which promo code applies in the Guardian Weekly checkout we need to use the country from the user's delivery address as this is what pricing is worked out from. Previously we weren't doing this so this PR changes the logic.


GHaberis (approved):
👍

PR #4167, opened 2022-08-30
Get paper default promos from S3

tomrf1:
We already do this for Guardian Weekly, this PR replaces hardcoded default promos for the `Paper` product.

I tested locally by:
1. setting `stage="PROD"`
2. adding `"paper": ["AUTUMN14"]` to the config file in S3
3. checking the `AUTUMN14` promo comes through on the default `/subscribe/paper` page

### Price card on product page:
Screen Shot 2022-08-31 at 08 38 29

### Payment page:
Screen Shot 2022-08-31 at 08 38 13

### DIFF
--- a/support-models/src/main/scala/com/gu/support/promotions/DefaultPromotions.scala
+++ b/support-models/src/main/scala/com/gu/support/promotions/DefaultPromotions.scala
@@ -16,10 +16,6 @@ object DefaultPromotions {
     def all: List[PromoCode] = Monthly.all ++ Annual.all
   }
 
-  object Paper {
-    val june21Promotion = "JUNE21SALE"
-  }
-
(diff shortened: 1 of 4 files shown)


support-models/src/main/scala/com/gu/support/promotions/DefaultPromotions.scala
```diff
@@ -16,10 +16,6 @@ object DefaultPromotions {
     def all: List[PromoCode] = Monthly.all ++ Annual.all
   }
 
-  object Paper {
-    val june21Promotion = "JUNE21SALE"
```
tomrf1, line 20: this has expired so no need to add to the config file

MaelGNM (approved):
Nice one!

PR #4168, opened 2022-08-30
Block contributions form submission on first name, last name and email validation errors

GHaberis:

We've seen invalid (too long) first name (max length 40), last name (max length 40) and email addresses (max length 80) submitted from the Contributions checkout which have caused errors on the server. 

This happens because the Contributions form does not currently block the form's submission on max-length validation errors, although it does show the validation error message on the form's submission.

After looking at this I found we're using different logic to show validation error messages and prevent form submission on the Contributions checkout. The rules defined in `assets/helpers/subscriptionsForms/rules.ts` determine when validation error messages show. However the rules defined in `support-frontend/assets/helpers/forms/formValidation.ts` are used when determining whether the form is submittable.

In `formValidation.ts` (is form submittable) we didn't have any checks on the length of first name, last name and email addresses, so we were allowing the form to be submitted. My short-term solution to the issue is to include a max-length check  in `formValidation.ts`, however I really don't think this ideal, the validation logic for showing error messages and preventing form submission should be shared - and this should be something we look at on the new product checkout.

I've also updated the validation checks on the email address in `rules.ts` (what errors messages to show), replacing a catch-all `checkEmail` rule with separate `nonEmptyString`, `isValidEmail` and `notLongerThan` rules as this means the error message shown will be more relevant to the actual error. 

[**Trello Card**](https://trello.com/c/sjvO6tFc/664-client-side-validation-bugs)

## Why are you doing this?

We've seen invalid (too long) first name, last name and email addresses submitted which have caused errors further downstream.

### Before 
We see the processing screen which indicates the form has posted to the `/create` end point despite validation errors, as a result we get a server side error that'll alarm

before_validation

### After 
We block form submission and display validation errors.

after_validation

### DIFF
--- a/support-frontend/assets/helpers/__tests__/formValidation.ts
+++ b/support-frontend/assets/helpers/__tests__/formValidation.ts
@@ -3,6 +3,7 @@ import {
 	amountOrOtherAmountIsValid,
 	checkStateIfApplicable,
 	maxTwoDecimals,
+	notLongerThan,
 } from '../forms/formValidation';
 import {
 	AUDCountries,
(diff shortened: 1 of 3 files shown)


GHaberis:
[Client-side validation bugs](https://trello.com/c/sjvO6tFc/664-client-side-validation-bugs)

i-hardy (approved):

PR #4175, opened 2022-09-01
Pd support link remove

paul-daniel-dempsey:
We need to remove the 'Support' page on the Support site, remove it from the nav.

This should be released behind the ab-newProduct 0% AB test so they're visible to the variant only. This AB test has already been set-up and is in use for showing changes to the checkout, you can force the AB test to be visible as so: https://support.theguardian.com/uk/subscribe#ab-newProduct=variant

[**Trello Card**](https://trello.com/c/3UI1xLUM/576-support-site-remove-the-support-landing-page)

## Why are you doing this?

image-3

Screenshot 2022-08-31 at 10 33 43

Screenshot 2022-08-31 at 10 34 16

Screenshot 2022-08-31 at 10 34 30

Screenshot 2022-09-01 at 11 39 40

## Is this an AB test?
- [x] Yes
- [ ] No

### DIFF
--- a/support-frontend/assets/components/headers/header/header.scss
+++ b/support-frontend/assets/components/headers/header/header.scss
@@ -154,7 +154,7 @@ $header-links-min-width: gu-span(12);
 	border-top: 1px solid gu-colour(brand-pastel);
 	z-index: 9;
 	white-space: nowrap;
-	margin: 0 ($gu-h-spacing * -2) 0 ($gu-h-spacing * -1);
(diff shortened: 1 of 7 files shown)


i-hardy (approved):
Looks good!

GHaberis (approved):
Logic looks good, I've left a comment on a variable name, but it's not a blocker

paul-daniel-dempsey:
> Logic looks good, I've left a comment on a variable name, but it's not a blocker

Yes makes sense, will change this now

PR #4178, opened 2022-09-02
Add sentry logging when recaptcha fails to load

i-hardy:

This adds Sentry logging when the Recaptcha script fails to load.

## Why are you doing this?

This increases visibility on problems with Recaptcha loading in certain environments.


GHaberis (approved):
👍

PR #4179, opened 2022-09-02
Pd links remove newspaper subscriptions nonuk

paul-daniel-dempsey:
The "Subscriptions" and "Newspaper" links in the navigation should only be visible to users within the UK.

This should be released behind the ab-newProduct 0% AB test so they're visible to the variant only. This AB test has already been set-up and is in use for showing changes to the checkout, you can force the AB test to be visible as so: https://support.theguardian.com/uk/subscribe#ab-newProduct=variant

image

image

image

image

NOTE : 2 menu items small enough to not require side-bar (additional code change reqd?)
image

[**Trello Card**](https://trello.com/c/d3CVEEHj/652-support-site-only-show-subscriptions-and-newspaper-links-in-the-nav-for-the-uk)

## Is this an AB test?
- [x] Yes
- [ ] No

### DIFF
--- a/support-frontend/assets/components/headers/links/links.tsx
+++ b/support-frontend/assets/components/headers/links/links.tsx
@@ -108,6 +108,8 @@ function Links({
 }: PropTypes): JSX.Element {
 	const { protocol, host, pathname } = window.location;
 	const urlWithoutParams = `${protocol}//${host}${pathname}`;
+	const internationalisationIDValue = internationalisationID(countryGroupId);
+	const isNotUk = internationalisationIDValue !== 'uk';
 	return (
 		<nav
 			className={classNameWithModifiers('component-header-links', [location])}
@@ -115,14 +117,12 @@ function Links({
 			<ul className="component-header-links__ul" ref={getRef}>
 				{links
 					.filter(({ text }) => {
-						const internationalisationIDValue =
-							internationalisationID(countryGroupId);
 						if (
 							text === 'Digital' ||
 							text === 'Support' ||
 							text === 'Contributions' ||
-							(text === 'Newspaper' && internationalisationIDValue !== 'uk') ||
-							(text === 'Subscriptions' && internationalisationIDValue !== 'uk')
+							(text === 'Newspaper' && isNotUk) ||
+							(text === 'Subscriptions' && isNotUk)
 						) {
 							if (isNewProduct) {
 								return false;
@@ -145,9 +145,6 @@ function Links({
 						return true;
 					})
 					.map((link) => {
-						const internationalisationIDValue =
-							internationalisationID(countryGroupId);
-
 						if (internationalisationIDValue == null || !link.internal) {
 							return link;
 						}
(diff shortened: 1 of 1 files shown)


paul-daniel-dempsey:
Have checked for the InternatlisationId meeting = uk. Hope that's right. 
When resizing  the non-uk screens down to iPad/Iphone size, the links header menu is wide enough to display without invoking the  side-bar (see last screenshot). Will a new link be added here in Mk1? If so, the extra space that takes could re-invoke the side-bar. When the screen width is less than the links header menu width the side-bar appears (NOT driven by set iPhone/iPad widths). This formatting behaviour is currently different between the UK and the other regions.

GHaberis:
@paul-daniel-dempsey looking at the screenshots I think we'll need to get a designer to review these changes to the nav

paul-daniel-dempsey:
> @paul-daniel-dempsey looking at the screenshots I think we'll need to get a designer to review these changes to the nav

Yes makes sense, I will look into how to force the side bar to appear in the meantime. Would this be Jade or Helene? If so, shall I request a meeting?

GHaberis:
> > @paul-daniel-dempsey looking at the screenshots I think we'll need to get a designer to review these changes to the nav
> 
> Yes makes sense, I will look into how to force the side bar to appear in the meantime. Would this be Jade or Helene? If so, shall I request a meeting?

I'd mention it in stand-up, but I think Jade should be able to help here

support-frontend/assets/components/headers/links/links.tsx
```diff
@@ -115,10 +115,14 @@ function Links({
 			<ul className="component-header-links__ul" ref={getRef}>
 				{links
 					.filter(({ text }) => {
+						const internationalisationIDValue =
```
GHaberis, line 118: Currently we're setting `internationalisationIDValue` for each item in the `links` array as we iterate over it, however it's value will always be the same so we may as well only set it once, we could move the variable definition to the top of the `Links` function along with the other variable definitions such as `urlWithoutParams` and only define it once.

GHaberis: We could also just save a variable `const isNotUk = internationalisationIDValue !== 'uk'` and then we wouldn't have to duplicate the conditional check `internationalisationIDValue !== 'uk'` multiple times.

paul-daniel-dempsey: Yes agree, much better than repeating the internationalisationID call per links item and isNotUK more readable especially. Have just put the changes in. Just a point of note, with ab-newProduct. When routing/clicking  between links they always go back to the non-variant (ie flipping between UK Subscription/NewsPaper/GWeekly always goes to the non-variant version link). When testing would the UI guys want this? Wish had asked earlier now!

GHaberis: I think when it comes to sharing links for Design/UX to preview the changes it'd be best to create a branch where we enable the `newProduct` AB test and deploy it to `CODE`.

GHaberis (approved):
Looks good 👍

paul-daniel-dempsey:
[Support site - Only show Subscriptions and Newspaper links in the Nav for the UK](https://trello.com/c/d3CVEEHj/652-support-site-only-show-subscriptions-and-newspaper-links-in-the-nav-for-the-uk)

PR #4180, opened 2022-09-05
Log error message on Quantum Metric script load error

GHaberis:

Logs the error message when the Quantum Metric script fails to load. We see this error in Sentry often and this extra information might help understand why.


marialani (approved):
👍

PR #4186, opened 2022-09-05
Make recaptcha script loading the responsibility of the Recaptcha component

i-hardy:

This moves responsibility for loading the recaptcha script and rendering the checkbox into the `Recaptcha` component.

## Why are you doing this?

When loading recaptcha in a parent component via the hook, we were depending on the placeholder element `<div id="robot_checkbox"></div>`, which the recaptcha checkbox attaches to, being available by the time the script itself loaded. Normally this would be fine, as a `useEffect` hook is guaranteed to be called _after_ the first component render.

However in the case of the Stripe payment form components, on the first render the component will be empty- we wait for the Stripe SDK to load before rendering the contents of the form, _including the recaptcha placeholder_. This was inadvertently causing a race condition between the recaptcha script and the Stripe SDK- if Stripe loaded first then the component would render, including the placeholder element, and everything would be fine, but if the recaptcha script loaded first the placeholder would not yet have rendered and it would have no element to hook onto, throwing an error.

This fixes the race condition by calling the hook from within the `Recaptcha` component that renders the placeholder element plus terms & conditions. Given that `useEffect` always runs after the first render, this guarantees that we will not be attempting to load the recaptcha script until the placeholder element has been added to the DOM, no matter where we use the component.


rupertbates (approved):
Looks good to me. A really clear explanation in the PR as well 👏

PR #4188, opened 2022-09-05
Add a debug slice to enable sending Redux action history from all checkouts

i-hardy:

This replaces the current Redux 'debug' setup, which records action history only for subscriptions checkouts, with a dedicated debug slice incorporated into all checkouts.

The debug slice has no reducers of its own but uses a [matcher](https://redux-toolkit.js.org/api/createReducer#builderaddmatcher) to determine what other actions it listens to. I've set it up to record any action that gets dispatched, but if there are actions that we find are too noisy or not relevant for debug purposes we can use this to exclude them in future.

The majority of changes in this PR are just updating state types for various contribution components now that the top level shape of the state has changed.

## Why are you doing this?

This replaces the old debug mechanism with a RTK-based one, and extends it to all checkouts, meaning we should have richer information when investigating checkout failures from contributions.

### DIFF
--- a/support-frontend/assets/__test-utils__/testStore.ts
+++ b/support-frontend/assets/__test-utils__/testStore.ts
@@ -10,6 +10,7 @@ import type {
 	ContributionsStore,
 } from 'helpers/redux/contributionsStore';
 import { initReduxForContributions } from 'helpers/redux/contributionsStore';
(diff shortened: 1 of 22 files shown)


michaelbjacobson (approved):
All looks great to me! :shipit:

PR #4189, opened 2022-09-05
Add contribution amount into the supporter product data table for contributors

rupertbates:
[Background information on the supporter product data extract is here](https://docs.google.com/document/d/1qGwmZ6LxF5Lc51O8GI2IkTuaDuyDiHqSm_QmLhWBvUo/edit#heading=h.g16rcvoydnow)

As part of the apps metering work, we want to be able to identify contributors  who are paying  more than a certain amount in members-data-api.

Unfortunately it is not possible to get this information from the aqua query which we use to extract the rest of the information from Zuora because it is held at the `RatePlanChargeTier` level while we query at the `RatePlan` level and there is no join between the two. This means that to get the information we need to make a separate call to Zuora for all recurring contributors.

The fact that we are now having to do more processing per record for a large number of our user base has led to a change in the architecture of the supporter-product-data extract - previously it was a single step function which extracted records  from Zuora and then wrote them directly to Dynamo DB:

Screen Shot 2022-09-06 at 13 48 05

This has now changed so that the step function is extracting data and writing each record to an SQS queue called `supporter-product-data-[STAGE]` which has a trigger set up to launch the `ProcessSupporterRatePlanItemLambda` lambda.

Screen Shot 2022-09-06 at 13 54 05

This lambda checks whether the record is a recurring contribution, and if it is it retrieves the contribution amount and currency from Zuora via the `/subscriptions/[Subscription Id]` endpoint. The record is then written to Dynamo DB.

### Lambda Throttling
Because the supporter product data extract can be run as a full extract of all customer records (currently around 1.7m records) we need to be careful not to use up all of our lambda concurrent execution capacity (1000 concurrent executions per account) or to overwhelm connected systems when these full extracts are run. To achieve this we use [reserved concurrency](https://docs.aws.amazon.com/lambda/latest/dg/configuration-concurrency.html) to limit the number of concurrent executions for the `ProcessSupporterRatePlanItemLambda`


support-modules/supporter-product-data-dynamo/src/main/scala/com/gu/supporterdata/services/SupporterDataDynamoService.scala
```diff
@@ -32,33 +32,49 @@ class SupporterDataDynamoService(client: DynamoDbAsyncClient, tableName: String)
...
+    val contributionAmount = "contributionAmount"
+    val contributionCurrency = "contributionCurrency"
 
     val key = Map(
       identityId -> AttributeValue.builder.s(beneficiaryIdentityId).build,
       subscriptionName -> AttributeValue.builder.s(item.subscriptionName).build,
     ).asJava
 
-    val updateExpression =
+    val nonOptionExpression =
```
rupertbates, line 43: I've split the update expression out into optional and non optional elements as contribution amount and currency are only needed for contributions

supporter-product-data/cloudformation/cfn.yaml
```diff
@@ -20,13 +20,16 @@ Mappings:
       S3Bucket: supporter-product-data-export-dev
       # Every 10 mins between 7am and 7pm Mon-Fri
       scheduleRate: "cron(*/10 7-19 ? * 1-5 *)"
+      lambdaConcurrency: 30
```
rupertbates, line 23: might tweak these values if we see downstream systems getting overwhelmed

supporter-product-data/src/main/scala/com/gu/services/ConfigService.scala
```diff
@@ -64,9 +65,14 @@ class ConfigService(stage: Stage) extends StrictLogging {
           "2c92a0ff56fe33f301572314aed277fb",
           "2c92a0fc610e738901612d83fce461fd",
           "2c92a0fe56fe33ff015723143e4778be",
+          "8a1295918021d0d2018022d4ca0c4aac",
```
rupertbates, line 68: Found some more rate plans to exclude from the extract. We should probably have a list of rate plans to include rather than one of excluded rate plans.

support-modules/supporter-product-data-dynamo/src/main/scala/com/gu/supporterdata/services/SupporterDataDynamoService.scala
```diff
@@ -32,33 +32,49 @@ class SupporterDataDynamoService(client: DynamoDbAsyncClient, tableName: String)
...
           $productRatePlanName = :$productRatePlanName,
           $termEndDate = :$termEndDate,
           $contractEffectiveDate = :$contractEffectiveDate,
           $expiryDateName = :$expiryDateName
           """
-    val attributeValues = Map(
+    val updateExpression =
+      if (item.contributionAmount.isDefined)
+        nonOptionExpression + s",\n          $contributionAmount = :$contributionAmount,\n          $contributionCurrency = :$contributionCurrency"
+      else nonOptionExpression
```
johnduffell, line 54: what is nonOption as it's not clear the meaning? is it only the required values for all products?
how about setRequiredValuesExpression.

also would it be clearer to separate the defining the parts from combining them, rather than accumulating as we go along?
Something like this...
```scala
    val setRequiredValuesExpression =
      s"""SET
          $productRatePlanId = :$productRatePlanId,
          $productRatePlanName = :$productRatePlanName,
          $termEndDate = :$termEndDate,
          $contractEffectiveDate = :$contractEffectiveDate,
          $expiryDateName = :$expiryDateName
          """

    val additionalFieldExpressions =
      if (item.contributionAmount.isDefined)
        List(
          s"$contributionAmount = :$contributionAmount",
          s"$contributionCurrency = :$contributionCurrency",
        )
      else Nil

    val updateExpression =
      (setRequiredValuesExpression ++ additionalFieldsExpression)
        .mkString(s",\n          ")
```

supporter-product-data/build.sbt
```diff
@@ -66,7 +67,7 @@ deployToProd := {
   List(
     "-SupporterProductDataQueryZuora-",
     "-SupporterProductDataFetchResults-",
-    "-SupporterProductDataUpdateDynamo-",
+    "-SupporterProductDataAddSupporterRatePlanItemToQueue-",
```
johnduffell, line 70: `SupporterProductDataProcessSupporterRatePlanItem`  is missing here

supporter-product-data/cloudformation/cfn.yaml
```diff
@@ -154,20 +168,39 @@ Resources:
...
+      FunctionName:
+        !Sub support-SupporterProductDataProcessSupporterRatePlanItem-${Stage}
+      Description: "Process new subscriptions from the SQS queue"
+      Handler: "com.gu.lambdas.ProcessSupporterRatePlanItemLambda::handleRequest"
       Role: !GetAtt [ SupporterProductDataLambdaRole, Arn ]
       Code:
         S3Bucket: supporter-product-data-dist
         S3Key: !Sub support/${Stage}/supporter-product-data/supporter-product-data.jar
       MemorySize: 4096
       Runtime: "java8.al2"
```
johnduffell, line 201: consider using java11 now - it's the recommended one for new stuff, and it should be more efficient too.

supporter-product-data/cloudformation/cfn.yaml
```diff
@@ -319,6 +352,63 @@ Resources:
...
+    Condition: CreateProdResources
+    Properties:
+      AlarmName: !Sub 'Urgent 9-5 - There was a failure processing a supporter record in the ProcessSupporterRatePlanItemLambda lambda'
+      AlarmDescription: >
+        Check the dead-letters-direct-mail-PROD SQS queue for details of the record which failed and the ProcessSupporterRatePlanItemLambda
+        Cloudwatch logs for details of what went wrong
+      Namespace: AWS/SQS
+      Dimensions:
+        - Name: QueueName
+          Value: !Sub dead-letters-direct-mail-${Stage}
```
johnduffell, line 400: oops!!! wrong queue

johnduffell: I'm sure there's a way of doing a !Ref or something in cfn?

supporter-product-data/cloudformation/cfn.yaml
```diff
@@ -319,6 +352,63 @@ Resources:
                   - states:StartExecution
                 Resource: !Ref SupporterProductData
 
+  SupporterProductDataDeadLetterQueue:
+    Type: AWS::SQS::Queue
+    Properties:
+      QueueName: !Sub dead-letters-supporter-product-data-${Stage}
```
johnduffell, line 358: do we have any docs how to deal with dead letters?

supporter-product-data/cloudformation/cfn.yaml
```diff
@@ -319,6 +352,63 @@ Resources:
...
+        maxReceiveCount: 10
+      VisibilityTimeout: 3000
+      Tags:
+        - Key: Stage
+          Value: !Ref Stage
+
+  SQSTrigger:
+    Type: AWS::Lambda::EventSourceMapping
+    Properties:
+      BatchSize: 1
```
johnduffell, line 381: any particular reason for a batch size so small or just for simplicity?  I'm sure it's not a big deal, but I'm wondering the reason.

supporter-product-data/cloudformation/cfn.yaml
```diff
@@ -319,6 +352,63 @@ Resources:
...
+      - ProcessSupporterRatePlanItemLambda
+
+  UnprocessedSupporterProductDataRecordAlarm:
+    Type: 'AWS::CloudWatch::Alarm'
+    Condition: CreateProdResources
+    Properties:
+      AlarmName: !Sub 'Urgent 9-5 - There was a failure processing a supporter record in the ProcessSupporterRatePlanItemLambda lambda'
+      AlarmDescription: >
+        Check the dead-letters-direct-mail-PROD SQS queue for details of the record which failed and the ProcessSupporterRatePlanItemLambda
+        Cloudwatch logs for details of what went wrong
```
johnduffell, line 396: would be ideal to have a "see this doc for more about what it means and how to fix it"

supporter-product-data/cloudformation/cfn.yaml
```diff
@@ -319,6 +352,63 @@ Resources:
...
+      Namespace: AWS/SQS
+      Dimensions:
+        - Name: QueueName
+          Value: !Sub dead-letters-direct-mail-${Stage}
+      MetricName: ApproximateNumberOfMessagesVisible
+      Statistic: Average
+      ComparisonOperator: GreaterThanThreshold
+      Threshold: '1'
+      Period: '60'
+      EvaluationPeriods: '60'
```
johnduffell, line 406: worth noting that because the dead letter queue doesn't expire, this will go off just once, and never again until the queue is cleared.  Not sure if it's worth also having a threshold 100 or so for "this is really going off and noone paid attention last time"

supporter-product-data/cloudformation/cfn.yaml
```diff
@@ -328,7 +418,6 @@ Resources:
       AlarmName: !Join
         - ' '
         - - 'URGENT 9-5 -'
-          - !Ref 'Stage'
           - 'Failed to extract supporter product data from Zuora'
```
johnduffell, line 421: in that case no need the join, not to mention now we have alarms mobbing there's no such thing as urgent is there?

`AlarmName: 'Failed to extract supporter product data from Zuora'`

supporter-product-data/cloudformation/cfn.yaml
```diff
@@ -398,3 +485,28 @@ Resources:
...
+    Properties:
+      AlarmActions:
+        - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:reader-revenue-dev
+      AlarmName: !Join
+        - ' '
+        - - 'URGENT 9-5 -'
+          - 'There was a failure when trying to write supporter data to the supporter-product-data-PROD sqs queue'
+      AlarmDescription: !Join
+        - ' '
+        - - 'Search for the term "Error sending batch" in the AddSupporterRatePlanItemToQueueLambda cloudwatch logs'
```
johnduffell, line 501: I like that, only thing that would make it better is a direct link!

supporter-product-data/src/main/scala/com/gu/services/AlarmService.scala
```diff
@@ -31,6 +31,16 @@ class AlarmService(stage: Stage) {
       ),
     )
   }
+
+  def triggerSQSWriteAlarm = {
```
johnduffell, line 35: this is unused!

supporter-product-data/src/main/scala/com/gu/lambdas/AddSupporterRatePlanItemToQueueLambda.scala
```diff
@@ -98,60 +105,17 @@ object UpdateDynamoLambda extends StrictLogging {
...
         s"Continuing processing with batch of ${batch.length} - time remaining: ${timeOutCheck.timeRemainingMillis / 1000} seconds, buffer: ${timeoutBufferInMillis / 1000} seconds",
       )
-
-      Await.result(writeBatch(batch, dynamoDBService, alarmService), 120.seconds)
+      try {
+        sqsService.sendBatch(batch)
+      } catch {
+        case e: Throwable =>
+          SafeLogger.error(scrub"Error sending batch to sqs\nbatch content was ${batch.asJson.spaces2}", e)
+          alarmService.triggerCsvReadAlarm
```
johnduffell, line 113: not CSV read is it? I think this is supposed to be triggerSQSWriteAlarm?

supporter-product-data/src/main/scala/com/gu/lambdas/AddSupporterRatePlanItemToQueueLambda.scala
```diff
@@ -98,60 +105,17 @@ object UpdateDynamoLambda extends StrictLogging {
       logger.info(
         s"Continuing processing with batch of ${batch.length} - time remaining: ${timeOutCheck.timeRemainingMillis / 1000} seconds, buffer: ${timeoutBufferInMillis / 1000} seconds",
       )
-
-      Await.result(writeBatch(batch, dynamoDBService, alarmService), 120.seconds)
+      try {
```
johnduffell, line 108: if you use a scala Try instead, it will only catch NonFatal
presume this is only if the whole batch fails, e.g. credential issue, rather than individual items

johnduffell: I'd be tempted to make sendBatch return a Try or similar, to minimise the reach of exceptions through the code.

supporter-product-data/src/main/scala/com/gu/lambdas/AddSupporterRatePlanItemToQueueLambda.scala
```diff
@@ -98,60 +105,17 @@ object UpdateDynamoLambda extends StrictLogging {
...
+        sqsService.sendBatch(batch)
+      } catch {
+        case e: Throwable =>
+          SafeLogger.error(scrub"Error sending batch to sqs\nbatch content was ${batch.asJson.spaces2}", e)
```
johnduffell, line 112: what are we going to do if there is some issue that makes us burn through all the batches from zuora and fail?  Are we going to have a log full of batches and no easy way to get back again?
I'm just wondering if there's a way to put the batch progress so far into the state and retry the batch until it works or it's manually recovered.

supporter-product-data/src/main/scala/com/gu/lambdas/Handler.scala
```diff
@@ -40,7 +42,11 @@ abstract class Handler[IN, OUT](implicit
 
 object StreamHandler {
   def fromStream[T](is: InputStream)(implicit decoder: Decoder[T]): Try[T] = {
-    val triedT = Try(Source.fromInputStream(is).mkString).flatMap(decode[T](_).toTry)
+    val triedT = Try {
+      val body = Source.fromInputStream(is).mkString
+      SafeLogger.info(body)
```
johnduffell, line 47:
```suggestion
      SafeLogger.info("lambda input: " + body)
```

supporter-product-data/src/main/scala/com/gu/lambdas/ProcessSupporterRatePlanItemLambda.scala
```diff
@@ -0,0 +1,132 @@
...
+      .map(_ => ())
+  }
+}
+
+object ProcessSupporterRatePlanItemLambda {
+  def getSupporterRatePlanItemProcessor() = {
+    val stage = StageConstructors.fromEnvironment
+    val dynamoService = SupporterDataDynamoService(stage)
+    val alarmService = AlarmService(stage)
+    val futureProcessor = ConfigService(stage).load.map { config =>
```
johnduffell, line 53: it looks like we will load the config on every invocation of the lambda.  If we make getSupporterRatePlanItemProcessor a val instead, it will only do it once, on cold start/class loading.

supporter-product-data/src/main/scala/com/gu/lambdas/ProcessSupporterRatePlanItemLambda.scala
```diff
@@ -0,0 +1,132 @@
...
+      Future.successful(supporterRatePlanItem)
+
+  def fetchContributionAmountFromZuora(supporterRatePlanItem: SupporterRatePlanItem) =
+    zuoraService
+      .getSubscription(supporterRatePlanItem.subscriptionName)
+      .map { sub =>
+        SafeLogger.info(
+          s"Contribution amount for supporter ${supporterRatePlanItem.identityId} is ${sub.contributionAmount}",
+        )
+        supporterRatePlanItem.copy(contributionAmount = sub.contributionAmount)
```
johnduffell, line 85: `sub.contributionAmount` Option is not the same Option as `supporterRatePlanItem.contributionAmount` Option.  This can cause subtle bugs

The former means "missing charge or plan" wheras the latter means "not a contribution or supporter plus"

Personally I'd like to see us flag it in some way if we expected an amount and zuora returned missing plans etc rather than just omitting.  This isn't as bad as the one for one option collapsing but it still worries me.

supporter-product-data/src/main/scala/com/gu/lambdas/ProcessSupporterRatePlanItemLambda.scala
```diff
@@ -0,0 +1,132 @@
...
+        dynamoService,
+        alarmService,
+        new ContributionProcessor(stage, config),
+      )
+    }
+    Await.result(futureProcessor, 10.seconds)
+  }
+}
+
+class ContributionProcessor(stage: Stage, config: ZuoraQuerierConfig) {
```
johnduffell, line 64: Processor is a bit generic, this could be called ContributionZuoraAmountFetcher

supporter-product-data/src/main/scala/com/gu/lambdas/ProcessSupporterRatePlanItemLambda.scala
```diff
@@ -0,0 +1,132 @@
...
+        SafeLogger.info(s"Successfully wrote item to Dynamo - ${supporterRatePlanItem.asJson.noSpaces}")
+        ()
+      }
+    } catch {
+      case t: Throwable =>
+        SafeLogger.error(scrub"Error writing item to Dynamo - ${supporterRatePlanItem.asJson.noSpaces}", t)
+        alarmService.triggerDynamoWriteAlarm match {
+          case Success(_) => Future.successful(())
+          case Failure(exception) =>
+            SafeLogger.error(scrub"Failed to trigger dynamo alarm", exception)
```
johnduffell, line 114: it's getting bad once we're at this point!! 🔥

supporter-product-data/src/main/scala/com/gu/lambdas/ProcessSupporterRatePlanItemLambda.scala
```diff
@@ -0,0 +1,132 @@
...
+      .flatMap(writeItemToDynamo)
+
+  def writeItemToDynamo(supporterRatePlanItem: SupporterRatePlanItem) = {
+    SafeLogger.info(s"Attempting to write item to Dynamo - ${supporterRatePlanItem.asJson.noSpaces}")
+    try {
+      dynamoService.writeItem(supporterRatePlanItem).map { _ =>
+        SafeLogger.info(s"Successfully wrote item to Dynamo - ${supporterRatePlanItem.asJson.noSpaces}")
+        ()
+      }
+    } catch {
```
johnduffell, line 108: this won't do much as writeItem returns a Future - we need to be looking out for a Failed future which I think will happen at the handlerFuture level.  Ideally we can catch all exceptions at the lowest level, and deal in Futures and Trys everywhere else.

supporter-product-data/src/main/scala/com/gu/lambdas/ProcessSupporterRatePlanItemLambda.scala
```diff
@@ -0,0 +1,132 @@
...
+object SupporterRatePlanItemCodec {
+  implicit val codec: Codec[SupporterRatePlanItem] = deriveCodec
+  implicit val contributionCodec: Codec[ContributionAmount] = deriveCodec
+}
+
+object ContributionIds {
+  def forStage(stage: Stage) = stage match {
+    case PROD => List("2c92a0fc5aacfadd015ad24db4ff5e97", "2c92a0fc5e1dc084015e37f58c200eea")
+    case UAT => List("2c92c0f85ab269be015acd9d014549b7", "2c92c0f95e1d5c9c015e38f8c87d19a1")
+    case DEV => List("2c92c0f85a6b134e015a7fcd9f0c7855", "2c92c0f85e2d19af015e3896e824092c")
```
johnduffell, line 130: the ids proliferate! these are rate plan ids specifically.

I've recently done some work to extract the ones we have in support-service-lambdas, but that doesn't necessarily help support-frontend or elsewhere https://github.com/guardian/support-service-lambdas/pull/1698

supporter-product-data/src/main/scala/com/gu/model/dynamo/SupporterRatePlanItemCodecs.scala
```diff
@@ -25,7 +25,28 @@ object SupporterRatePlanItemCodecs {
...
+          subscriptionName,
+          identityId,
+          gifteeIdentityId,
+          productRatePlanId,
+          productRatePlanName,
+          termEndDate,
+          contractEffectiveDate,
+          None,
+        ),
+    )
```
johnduffell, line 48: no wonder json decoders are always full of macros 😞

supporter-product-data/src/main/scala/com/gu/model/zuora/response/MinimalZuoraSubscription.scala
```diff
@@ -0,0 +1,28 @@
...
+  implicit val decoder: Decoder[MinimalZuoraSubscription] = deriveDecoder
+}
+
+object RatePlan {
+  implicit val decoder: Decoder[RatePlan] = deriveDecoder
+}
+
+object RatePlanCharge {
+  implicit val decoder: Decoder[RatePlanCharge] = deriveDecoder
+}
```
johnduffell, line 28: personally I'd put all these in the companion object of ZuoraSubscriptionService to keep them together, as this is pretty specific to the situation

supporter-product-data/src/main/scala/com/gu/model/zuora/response/MinimalZuoraSubscription.scala
```diff
@@ -0,0 +1,28 @@
+package com.gu.model.zuora.response
+
+import com.gu.supporterdata.model.ContributionAmount
+import io.circe.Decoder
+import io.circe.generic.semiauto.deriveDecoder
+
+case class MinimalZuoraSubscription(ratePlans: List[RatePlan]) {
+  def contributionAmount = for {
+    ratePlans <- ratePlans.headOption
+    charges <- ratePlans.ratePlanCharges.headOption
```
johnduffell, line 10: we are assuming exactly one charge and plan - how about if there are multiple? can we return an Either and complain about it as clearly either the sub is broken, or someone has reused this code in future and it might quietly break.

supporter-product-data/src/main/scala/com/gu/services/ConfigService.scala
johnduffell: yeah what a pain!

supporter-product-data/src/main/scala/com/gu/lambdas/AddSupporterRatePlanItemToQueueLambda.scala
```diff
@@ -98,60 +105,17 @@ object UpdateDynamoLambda extends StrictLogging {
...
-    val futures = list.map { case (supporterRatePlanItem, index) =>
-      logger.info(
-        s"Attempting to write item index $index - ${supporterRatePlanItem.productRatePlanName} to Dynamo - ${supporterRatePlanItem.asJson.noSpaces}",
-      )
-      dynamoDBService.writeItem(supporterRatePlanItem)
-    }
-    Future.sequence(futures)
-  }
-
-  def batchItemsWhichCanUpdateConcurrently(
```
johnduffell, line 124: so we no longer can do anything concurrently?

supporter-product-data/src/main/scala/com/gu/services/ZuoraSubscriptionService.scala
```diff
@@ -0,0 +1,30 @@
...
+import com.gu.rest.WebServiceHelper
+import io.circe.syntax.EncoderOps
+
+import java.time.format.DateTimeFormatter
+import java.time.{LocalDateTime, ZonedDateTime}
+import scala.collection.immutable.Map.empty
+import scala.concurrent.{ExecutionContext, Future}
+
+class ZuoraSubscriptionService(val config: ZuoraQuerierConfig, client: FutureHttpClient)(implicit ec: ExecutionContext)
+    extends WebServiceHelper[BatchQueryErrorResponse] {
```
johnduffell, line 18: BatchQueryErrorResponse sounds unlikely I think, is this a copy and paste?

support-modules/supporter-product-data-dynamo/src/main/scala/com/gu/supporterdata/model/SupporterRatePlanItem.scala
```diff
@@ -10,4 +10,7 @@ case class SupporterRatePlanItem(
     productRatePlanName: String, // Name of the product in this rate plan
     termEndDate: LocalDate, // Date that this subscription term ends
     contractEffectiveDate: LocalDate, // Date that this subscription started
+    contributionAmount: Option[ContributionAmount],
 )
+
+case class ContributionAmount(amount: BigDecimal, currency: String)
```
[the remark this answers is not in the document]
tomrf1: There is also a `Price` model in membership-common, but it uses Float, which I think is more susceptible than BigDecimal? - https://github.com/guardian/membership-common/blob/main/src/main/scala/com/gu/memsub/Price.scala#L7

supporter-product-data/src/main/scala/com/gu/services/SqsService.scala
```diff
@@ -0,0 +1,57 @@
...
+import com.gu.supporterdata.model.{ContributionAmount, Stage, SupporterRatePlanItem}
+import io.circe.Encoder
+import io.circe.generic.semiauto.deriveEncoder
+import io.circe.syntax.EncoderOps
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.jdk.CollectionConverters._
+
+class SqsService(queueName: String)(implicit val executionContext: ExecutionContext) {
+  implicit val encoder: Encoder[SupporterRatePlanItem] = deriveEncoder
```
tomrf1, line 23: does some of this belong in the companion object?

rupertbates: `SupporterRatePlanItem` is defined in a project which doesn't  have a dependency on Circe so to avoid adding it we need to define codecs here.

support-modules/supporter-product-data-dynamo/src/main/scala/com/gu/supporterdata/model/SupporterRatePlanItem.scala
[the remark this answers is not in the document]
johnduffell: oh interesting, having read into it, bigdecimal actually scales the number in base 10 at whatever precision is necessary (rather than base 2 with limited precision), and it would throw an exception if it couldn't represent the number exactly.
In our case, it basically does the trick I mentioned for free.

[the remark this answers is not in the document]
johnduffell: and tom yes I think it was back in the membership-common days that we had trouble with discounting or something due to floats.

supporter-product-data/src/main/scala/com/gu/services/SqsService.scala
johnduffell: nice, I'm heartened that dependencies (aka liabilities!) are a consideration for code structure

supporter-product-data/cloudformation/cfn.yaml
rupertbates: Just copied and pasted from elsewhere and didn't change it. I'm not sure if there are any real advantages to a larger batch size but I've upped it to 10

supporter-product-data/src/main/scala/com/gu/lambdas/ProcessSupporterRatePlanItemLambda.scala
rupertbates: It also contains the logic to determine whether a rate plan is a recurring contribution though, it just generally knows about and deals with contributions.

rupertbates: I've refactored to make it all a bit clearer

supporter-product-data/src/main/scala/com/gu/lambdas/ProcessSupporterRatePlanItemLambda.scala
rupertbates: I see what you mean. I'm not sure what would be an appropriate action to take here though, I wouldn't be at all surprised to find out that we have contributions with no rate plan charges but what should we do other than just treating it as missing? We could alert but it would just get triggered every time we ran a full refresh unless we were going to go in and investigate each failing subscriptions.

supporter-product-data/src/main/scala/com/gu/services/SqsService.scala
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    val batchRequest = new SendMessageBatchRequest(queueUrl, batchRequestEntries.asJava)
+    val result = sqsClient.sendMessageBatch(batchRequest)
+    result.getFailed.asScala.toList.foreach { error =>
+      val failedItem = supporterRatePlanItems(error.getId.toInt)
+      SafeLogger.error(
+        scrub"Failed to write a supporterRatePlanItem to the queue\nItem body: ${failedItem.asJson}\nError message: ${error.getMessage}",
+      )
+    }
```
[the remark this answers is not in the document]
rupertbates: Ok, I've tidied this up a bit. We now log the result an trigger an alarm, I'm not sure if it would be better to just fail at this point but I'm going to go with the alarm for now and we can see whether it ever gets triggered

supporter-product-data/src/main/scala/com/gu/lambdas/AddSupporterRatePlanItemToQueueLambda.scala
rupertbates: We could throw at this point  and it will just keep retrying but I think I'd rather trigger an alarm and keep going for now. We can always run a full sync at any time so we're not losing any data permanently.

supporter-product-data/src/main/scala/com/gu/lambdas/AddSupporterRatePlanItemToQueueLambda.scala
rupertbates: I've consolidated this into the SqsService as there were two separate bits of code both handling errors from sqs

supporter-product-data/src/main/scala/com/gu/lambdas/AddSupporterRatePlanItemToQueueLambda.scala
rupertbates: We can actually do everything concurrently, I worked through it and the ordering doesn't actually matter, although I thought  it did when I originally wrote this code.

supporter-product-data/cloudformation/cfn.yaml
```diff
@@ -319,21 +352,77 @@ Resources:
...
+    DependsOn:
+      - SupporterProductDataQueue
+      - ProcessSupporterRatePlanItemLambda
+
+  UnprocessedSupporterProductDataRecordAlarm:
+    Type: 'AWS::CloudWatch::Alarm'
+    Condition: CreateProdResources
+    Properties:
+      AlarmName: There was a failure processing a supporter record in the ProcessSupporterRatePlanItemLambda lambda
+      AlarmDescription: >
```
tomrf1, line 395: I like all the helpful alarm messages here!

tomrf1 (approved):
Looks great!

PR #4192, opened 2022-09-06
Add a debug slice for Redux action history - with serialisation error protection

i-hardy:

This is a re-do of #4188 - please see that PR description for details on the change.

## Why are you doing this?

The previous PR caused a problem on the contributions checkout with the async thunk action that checks the validity of direct debit details, which contains a `meta` property with an unserializable, circular reference. This issue has been corrected by only serialising the `type` and `payload` values from actions, and additionally adding a try/catch to the debug reducer to guard against future edge cases.

### DIFF
--- a/support-frontend/assets/helpers/redux/debug/reducer.ts
+++ b/support-frontend/assets/helpers/redux/debug/reducer.ts
@@ -1,0 +1,23 @@
+import { createSlice } from '@reduxjs/toolkit';
+
+const isAnyAction = () => true;
+
+export const debugSlice = createSlice({
+	name: 'debug',
+	initialState: {
+		actionHistory: '',
+	},
+	reducers: {},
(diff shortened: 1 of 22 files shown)


support-frontend/assets/helpers/redux/debug/reducer.ts
```diff
@@ -0,0 +1,23 @@
...
+			try {
+				const { type, payload } = action as { type: string; payload: unknown };
+				state.actionHistory += ` ${JSON.stringify({ type, payload })}\n`;
+			} catch (error) {
+				return;
+			}
+		});
+	},
+});
+
```
michaelbjacobson, line 22: Looks good!

michaelbjacobson (approved):
👍🏼

PR #4203, opened 2022-09-08
Quantum Metric: sendEvent on Single Contribution Conversions w/ Paypal

GHaberis:

We were informed by Quantum Metric (QM) that they were not seeing any conversion events for Single Contributions where Paypal was the payment method. We send an event to notify QM of a Contribution conversion using the function `sendEventContributionCheckoutConversion`, which was being called from [ContributionFormContainer.tsx](https://github.com/guardian/support-frontend/compare/gh-qm-paypal?expand=1#diff-be89a142511eb13fcfbad9516c2d5dea6042e18103f0b48721945033f3a8d41d). 

On investigation we found `sendEventContributionCheckoutConversion` was only being called for Single Contribution conversions w/ Credit Card and Recurring Contribution conversions w/ all payment methods. The reason it wasn't being called for Single Contribution conversions w/ Paypal is we take these users off the support site to Paypal's site to complete their payment, and return them **directly** to the Contributions Thank You page, whereas all other users remain on Contributions Landing Page whilst their payment is processed, or in the case of Recurring Contribution conversions w/ Paypal they're returned to the Contributions Landing Page before being redirected to the Contributions Thank You page. In short the code block within which the `sendEventContributionCheckoutConversion` call was made is never executed for Single conversions w/ Paypal.

To get around this we've settled on the solution of storing the information we need to send QM (`selectedContributionType`, `contributionAmount` and `currency`) in session storage. `selectedContributionType` and `currency` are already set in session storage in separate side affects, so we only needed to additionally set `contributionAmount`. as a side affect of the 'SET_CHECKOUT_FORM_HAS_BEEN_SUBMITTED' redux action.

This information is retrieved on the  the Contributions Thank You page and we send it to QM from here rather now. We have to use session storage as the state does not persist when going from the Contributions Landing Page to Paypal and then back to the Contributions Thank You page.

### DIFF
--- a/support-frontend/assets/helpers/page/analyticsAndConsent.ts
+++ b/support-frontend/assets/helpers/page/analyticsAndConsent.ts
@@ -25,6 +25,7 @@ function analyticsInitialisation(
 	setReferrerDataInLocalStorage(acquisitionData);
 	void googleTagManager.init(participations);
 	ophan.init();
+	console.log('*** initQuantumMetric ***');
 	initQuantumMetric(participations);
 	trackAbTests(participations);
 	// Sentry logging.
--- a/support-frontend/assets/helpers/page/analyticsAndConsent.ts
+++ b/support-frontend/assets/helpers/page/analyticsAndConsent.ts
@@ -25,7 +25,6 @@ function analyticsInitialisation(
 	setReferrerDataInLocalStorage(acquisitionData);
 	void googleTagManager.init(participations);
 	ophan.init();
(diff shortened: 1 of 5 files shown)


GHaberis:
[Quantum Metric: SendEvent for Single Contribution conversions paid with PayPal.](https://trello.com/c/dQHMaHB8/529-quantum-metric-sendevent-for-single-contribution-conversions-paid-with-paypal)

marialani (approved):
LGTM 👍

PR #4208, opened 2022-09-09
Add Benefits List - checkout MVP

marialani:
This adds the checkout benefits list for use on the v2 Guardian supporter checkout. 

[**Trello Card**](https://trello.com/c/C6NvlSs3/643-checkout-v2-build-your-extras-list-component)

## Screenshots

Screenshot 2022-09-09 at 10 44 23


support-frontend/assets/components/checkoutBenefits/svgCrossRound.tsx
```diff
@@ -0,0 +1,54 @@
+import { css } from '@emotion/react';
+import type { EmotionJSX } from '@emotion/react/types/jsx-namespace';
+import { iconSize, visuallyHidden } from '@guardian/source-foundations';
+
+// We should replace this component with 'SvgCrossRound'
+// from source once we're on the up to date version
```
i-hardy, line 6: I updated all our Source deps recently so worth checking if we still need this

i-hardy (approved):
This looks great! Just worth checking if we can use the Source SVG now we're up to date on `main`

PR #4226, opened 2022-09-16
Remove the print + digital component from newspaper checkout

rupertbates:
The amounts being shown for print + digital on the newspaper checkout are incorrect so we are going to switch it off. Assuming that the new proposition launches and  is successful we can delete it at a later date as it will no longer be needed.


i-hardy (approved):
Interesting conditional 😆

PR #4237, opened 2022-09-22
PR to remove the country dropdown on the Subscriptions Landing Page

LAKSHMIRPILLAI:
This is  a PR to remove the country dropdown on the Subscriptions Landing Page

This should be released behind the ab-newProduct 0% AB test so they're visible to the variant only. 

[**Trello Card**](https://trello.com/c/whARz87c/706-remove-the-country-dropdown-on-the-subscriptions-landing-page)

## Why are you doing this?
After removing Digital Subscriptions from sale the Subscriptions Landing Page ([Support the Guardian | Get a Subscription](https://support.theguardian.com/uk/subscribe)) will only become relevant to the UK only, as Guardian Weekly will be the only sub on offer to non-UK regions.So we have to remove the country dropdown on the Subscriptions Landing Page

## Is this an AB test?
- [ ] Yes

Before Change
Screenshot 2022-09-22 at 14 12 54

After change
Screenshot 2022-09-22 at 14 13 27

### DIFF
--- a/support-frontend/assets/pages/subscriptions-landing/subscriptionsLanding.tsx
+++ b/support-frontend/assets/pages/subscriptions-landing/subscriptionsLanding.tsx
@@ -1,5 +1,6 @@
 // ----- Imports ----- //
 import Footer from 'components/footerCompliant/Footer';
+import Header from 'components/headers/header/header';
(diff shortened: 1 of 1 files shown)


i-hardy (approved):
Nice and simple, looks good!

PR #4256, opened 2022-10-04
Add Upsell CTA - v2 checkout 

marialani:

## Is this an AB test?
- [ ] Yes
- [ ] No

### DIFF
--- a/support-frontend/assets/components/button/upsellButton.tsx
+++ b/support-frontend/assets/components/button/upsellButton.tsx
@@ -1,0 +1,47 @@
+import type { SerializedStyles } from '@emotion/react';
+import { css, ThemeProvider } from '@emotion/react';
+import { neutral, space } from '@guardian/source-foundations';
+import {
+	Button,
+	buttonThemeReaderRevenueBrand,
+} from '@guardian/source-react-components';
+
+const button = css`
+	width: 100%;
+	justify-content: space-around;
+	margin: ${space[5]}px 0;
+	color: ${neutral[7]};
+`;
+
+interface UpsellButtonProps {
+	buttonCopy: string | null;
+	handleButtonClick: () => void;
+	cssOverrides?: SerializedStyles;
+}
+
+function UpsellButton({
+	buttonCopy,
+	handleButtonClick,
+	cssOverrides,
+}: UpsellButtonProps): JSX.Element {
+	return (
+		<>
+			{buttonCopy && (
+				<ThemeProvider theme={buttonThemeReaderRevenueBrand}>
+					<Button
+						iconSide="left"
+						priority="primary"
+						size="default"
+						css={[button, cssOverrides]}
+						onClick={() => handleButtonClick()}
+						cssOverrides={cssOverrides}
+					>
+						{buttonCopy}
+					</Button>
+				</ThemeProvider>
+			)}
+		</>
+	);
+}
+
+export default UpsellButton;
--- a/support-frontend/assets/components/checkoutBenefits/checkoutBenefitsList.tsx
+++ b/support-frontend/assets/components/checkoutBenefits/checkoutBenefitsList.tsx
@@ -6,6 +6,7 @@ import {
 	space,
 	textSans,
 } from '@guardian/source-foundations';
+import UpsellButton from 'components/button/upsellButton';
 import type { CheckListData } from './checkoutBenefitsListContainer';
 
 const container = css`
@@ -41,7 +42,6 @@ const checkListText = css`
 
 const table = css`
 	padding-top: ${space[4]}px;
-	margin-bottom: 28px;
 
 	& tr:not(:last-child) {
 		border-bottom: 10px solid transparent;
@@ -54,13 +54,17 @@ const table = css`
 	}
 `;
 
+const buttonCssOverrides = css`
+	& button {
+		margin: 20px 0 12px;
+	}
+`;
+
 const hr = (margin: string) => css`
 	border: none;
 	height: 1px;
 	background-color: #dcdcdc;
 	margin: ${margin};
-
-	margin-bottom: ${space[4]}px 0;
 `;
 
 const para = css`
@@ -70,11 +74,15 @@ const para = css`
 export type CheckoutBenefitsListProps = {
 	title: string;
 	checkListData: CheckListData[];
+	buttonCopy: string | null;
+	handleButtonClick: () => void;
 };
 
 export function CheckoutBenefitsList({
 	title,
 	checkListData,
+	buttonCopy,
+	handleButtonClick,
 }: CheckoutBenefitsListProps): JSX.Element {
 	return (
 		<div css={container}>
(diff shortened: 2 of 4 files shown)


marialani:
[V2 Checkout - build 'upsell CTA' component](https://trello.com/c/lXuc01lP/690-v2-checkout-build-upsell-cta-component)

support-frontend/assets/components/button/upsellButton.tsx
```diff
@@ -0,0 +1,56 @@
...
+import {
+	Button,
+	buttonThemeReaderRevenueBrand,
+} from '@guardian/source-react-components';
+
+const button = css`
+	width: 100%;
+	justify-content: space-around;
+	margin: ${space[3]}px 0;
+	color: ${neutral[7]};
```
GHaberis, line 13: Do we need to set style rules on `@guardian/source-react-components` such as the `<Button>`, I thought these would all be defined by the Button's Theme?

support-frontend/assets/components/checkoutBenefits/checkoutBenefitsList.tsx
```diff
@@ -54,13 +54,17 @@ const table = css`
 	}
 `;
 
+const buttonCssOverrides = css`
```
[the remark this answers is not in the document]
marialani: Good spot! These styles should be removed

support-frontend/assets/components/checkoutBenefits/checkoutBenefitsListContainer.tsx
```diff
@@ -106,9 +30,20 @@ function getBenefitsListTitle(
...
+	higherTier: boolean,
+	thresholdPriceWithCurrency: string,
+	selectedAmount: number,
+) =>
+	higherTier || Number.isNaN(selectedAmount)
+		? null
+		: `Switch to ${thresholdPriceWithCurrency} to unlock all extras`;
+
 export function CheckoutBenefitsListContainer({
 	renderBenefitsList,
```
GHaberis, line 43: I'm wondering why this `CheckoutBenefitsListContainer` component follows this pattern of taking a function (`renderBenefitsList`) that returns a rendered `CheckoutBenefitsList` component? Is this a common pattern we use? I've not seen it elsewhere, and it's not clear what benefits it offers?

It feels like it'd be cleaner if we replaced the `<CheckoutBenefitsListContainer>` in the [amountAndBenefits.tsx](https://github.com/guardian/support-frontend/blob/df6d7a7889b8d316197f49d41dde64d79feef802/support-frontend/assets/pages/supporter-plus-landing/formSections/amountAndBenefits.tsx)  with`<CheckoutBenefitsList>` and to moved all the logic that gets the values for the `benefitsListProps` out of the `CheckoutBenefitsListContainer` and into the `CheckoutBenefitsList` component, where it's used? We could then get rid of the `CheckoutBenefitsListContainer` which seems like an unnecessary component.

GHaberis: This suggestion above ☝️ would mean  `CheckoutBenefitsList` would not take any props, as the logic to generate the `benefitsListProps` would sit within the `checkoutBenefitsList.tsx` module. Would this make it harder to render in StoryBook? Is this why we pass in props to it via a `CheckoutBenefitsListContainer`?

i-hardy: This is a presentational and container component approach- the benefit of it is that we separate the logic of talking to Redux, doing calculations, sending tracking, etc etc from the purely presentational UI component. This means:

- It's much easier to write stories for purely presentational components, and to demonstrate (and this have regression tests for) their different variants in Storybook
- It's much easier to swap out one presentational component for a different one- or test two variations of a component against each other- without having to rework any of the underlying business logic

I suggested this as an approach to move towards way back at the start of the year when we were more focused on ways to unify the checkouts, but it will also make A/B testing much less burdensome and generally make our components smaller. I also wrote something up for it on the wiki [here](https://github.com/guardian/support-frontend/wiki/Presentational-and-Container-Components).

GHaberis: Ah OK thanks for explaining @i-hardy, this makes sense, and I did suspect my suggestion would make it harder to test after I posted my first message! So if we retain the container `<CheckoutBenefitsListContainer>` component and continue to do all the  talking to Redux, calculations, tracking logic in this component, could we not just return the presentational `<CheckoutBenefitsList>` component rather than the result of the `renderBenefitsList` function? Abridged example...

```
export function CheckoutBenefitsListContainer() {
   const benefitsListProps = { ... } // do all calculations to generate benefitsListProps here
   return <CheckoutBenefitsList {...benefitsListProps} />
}
```

i-hardy: We certainly could do that, but the benefit of using a render prop is that it's up to the consumer of `CheckoutBenefitsListContainer` to decide how to use those props, rather than tying this container specifically to that presentational component. We could swap out `CheckoutBenefitsList` for anything else without having to touch this component.

i-hardy: If it's helpful, here's where the component is actually being used (along with several other container/presentational components) https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/pages/supporter-plus-landing/formSections/amountAndBenefits.tsx

support-frontend/assets/components/button/upsellButton.tsx
marialani: These styles were previously being passed to the button using the css prop but should've been passed to the cssOverrides prop instead. With theming I don't think we have a [button theme](https://theguardian.design/2a1e5182b/p/435225-button/t/85cf86) for the button that would match the desired background and text colour, so I'd opted for the closest one with a text colour overwrite

support-frontend/assets/components/checkoutBenefits/checkoutBenefitsListContainer.tsx
GHaberis: OK yes I see, it's the first time i've seen the container/presentational component pattern in action, so thanks for explanation!

support-frontend/assets/components/checkoutBenefits/checkoutBenefitsListData.tsx
```diff
@@ -0,0 +1,81 @@
...
+	return [
+		{
+			icon: getSvgIcon(lowerTier),
+			text: (
+				<p>
+					<span css={boldText}>Uninterrupted reading. </span> No more yellow
+					banners
+				</p>
+			),
+			maybeGreyedOut: lowerTier ? null : greyedOut,
```
GHaberis, line 49: You could define 2 variables at the top of this `checkListData` function :

```
const maybeGreyedOutLowerTier = lowerTier ? null : greyedOut`;
const maybeGreyedOutHigherTier = higherTier ? null : greyedOut`;
```

Then you could use the variables values, it'd save making the same calculations multiple times in this function.

GHaberis (approved):
Looks good to me 👍

PR #4261, opened 2022-10-05
Set the correct invoice template ID when creating a subscription in Zuora

paulbrown1982:
Inserting the correct invoice template ID, based on the sold-to country and currency, into every account creation API call to Zuora.

[**Trello Card**](https://trello.com/c/LFp15ME6)

## Why are you doing this?
This change replicates, and improves upon, what used to happen in Membership Common (https://github.com/guardian/membership-common/pull/539). It ensures any new customer having their product supplied to Australia, paying AUD, who wants to download (within Manage which calls https://github.com/guardian/invoicing-api/tree/main/src/main/scala/com/gu/invoicing/pdf) their invoice from Zuora, has the correct company name, and correct layout for GST (Goods and Sales Tax), rather than showing the UK company name and VAT.

There will also be some follow-up changes to:

1. Backfill
2. Create a Zuora Workflow to ensure this happens for subscriptions created outside of this process.
3. Create a Zuora Workflow to set the invoice correctly if someone moves, either to, or away from, Australia

Invoice templates are Word documents entered into Zuora's Billing admin portal:

Screenshot 2022-10-05 at 17 14 06

## Screenshots

Screenshots of the two templates are here:

Default:

Screenshot 2022-10-05 at 17 16 56

Australia:

Screenshot 2022-10-05 at 17 17 09

### DIFF
--- a/support-config/src/main/scala/com/gu/support/config/ZuoraConfigProvider.scala
+++ b/support-config/src/main/scala/com/gu/support/config/ZuoraConfigProvider.scala
@@ -76,7 +76,7 @@ class ZuoraConfigProvider(config: Config, defaultStage: Stage)
   )
 
   private def invoiceTemplatesFromConfig(config: Config) = ZuoraInvoiceTemplatesConfig(
-    defaultTemplateId = config.getString("default"),
-    auTemplateId = config.getString("au"),
+    defaultTemplateId = config.getString("defaultTemplateId"),
+    auTemplateId = config.getString("auTemplateId"),
   )
 }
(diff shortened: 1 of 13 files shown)


support-config/src/main/scala/com/gu/support/config/ZuoraConfigProvider.scala
```diff
@@ -67,4 +74,9 @@ class ZuoraConfigProvider(config: Config, defaultStage: Stage)
     monthlyChargeId = config.getString("monthly.productRatePlanChargeId"),
     annualChargeId = config.getString("annual.productRatePlanChargeId"),
   )
+
+  private def invoiceTemplatesFromConfig(config: Config) = ZuoraInvoiceTemplatesConfig(
+    defaultTemplateId = config.getString("default"),
```
KentGrm, line 79: Just a question on readability and Guardian standards: should `defaultTemplateId` map to `defaultId` with the addition of the `Id` suffix. Same for `au`

KentGrm (approved):

rupertbates (approved):
👍  Looks good, remember to set up config in all environments!

PR #4268, opened 2022-10-07
Update broken accessibility testing links in our PR template to point at useful resources

i-hardy:

This updates our PR template so the links for the accessibility checks go to useful resources rather than the broken accessibility.gutools.co.uk site.


marialani (approved):
👍

PR #4283, opened 2022-10-13
Add skip links and test user banner to new checkout

i-hardy:

This PR adds two [skip links](https://webaim.org/techniques/skipnav/) to the new checkout, and a single 'skip to main content' link to all other pages on the site using the `Page` template component. The component is based on [a similar component in dotcom-rendering](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/web/components/SkipTo.tsx). The links are invisible unless they currently have keyboard focus.

It also adds a test user banner to the new checkout.

## Why are you doing this?

Skip links are a useful aid for users who navigate with a keyboard or switch, and for screen reader users- they make it quicker to get to key content on a page. We currently don't have them anywhere on the site, so adding them to existing pages as well is a significant accessibility improvement that has no effect on existing designs and visuals.

## Is this an AB test?
- [ ] Yes
- [x] No

### New checkout
Screenshot 2022-10-13 at 15 33 11

### Newspaper landing page
Screenshot 2022-10-13 at 15 33 38

### Test user banner
Screenshot 2022-10-13 at 16-15-12 Support the Guardian Make a Contribution

### DIFF
--- a/support-frontend/assets/components/layout/container.tsx
+++ b/support-frontend/assets/components/layout/container.tsx
@@ -1,6 +1,7 @@
 import { css } from '@emotion/react';
 import { from, neutral } from '@guardian/source-foundations';
 import { Container as SourceContainer } from '@guardian/source-react-components';
+import type { HTMLAttributes } from 'react';
 import React from 'react';
 
 type ContainerElement =
@@ -12,7 +13,7 @@ type ContainerElement =
 	| 'nav'
 	| 'section';
 
-type Props = {
+interface ContainerProps extends HTMLAttributes<HTMLElement> {
 	children: React.ReactNode;
 	element?: ContainerElement;
 	sidePadding?: boolean;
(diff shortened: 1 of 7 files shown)


i-hardy:
[Add skip links for keyboard navigation on the V2 checkout page](https://trello.com/c/kyMI5YhK/768-add-skip-links-for-keyboard-navigation-on-the-v2-checkout-page)

support-frontend/assets/components/layout/container.tsx
```diff
@@ -12,15 +13,15 @@ type ContainerElement =
 	| 'nav'
 	| 'section';
 
-type Props = {
+interface ContainerProps extends HTMLAttributes<HTMLElement> {
```
marialani, line 16: So this gives us the ability to pass any additional html attribute to the Container without having to explicitly add it to the prop types?

marialani (approved):
LGTM! Test user banner is also a great addition 👍

support-frontend/assets/components/layout/container.tsx
i-hardy: Yes, it just makes it easier to add things like an ID if necessary.

PR #4288, opened 2022-10-14
Reinstate Upsell CTA - v2 Checkout

marialani:
This PR adds the up-sell CTA to the v2 checkout. A previous [PR](https://github.com/guardian/support-frontend/pull/4256) doing this was merged and [reverted](https://github.com/guardian/support-frontend/pull/4284) yesterday due to a bug caused by changes to the default product state in redux (see below). 

Most all of the changes in this PR are the same as the original, apart from the changes necessary to fix this bug, which are as follows:
- globals.ts - where the `getGlobal` function now has the ability to take a full path i.e. one prefixed with the string `window.guardian`
- globals.test.ts - where we include a test to ensure the `getGlobal` function can handle a string prefixed with `window.guardian` 

Whilst working on the original PR we'd encountered a bug in storybook whereby the accompanying stories for the Upsell CTA and the Supporter Plus page checkout would fail to render. This was down to the fact it was looking for a property on the redux product state that was not accessible at that point of execution for the Upsell CTA component. Though interestingly other storybook components accessing the same redux state did not seem to be afffected, leading us to conclude the root cause may be down to how webpack bundles the code.

The original changes to the product state fixed the bug in storybook and left the contributions pages unaffected. Although, the Subscriptions checkouts were unintentionally affected by this change, causing a bug in production down to the fact that we were trying to read a property on the product state which was no longer being accessed correctly. 

[**Trello Card**](https://trello.com/c/lXuc01lP/690-v2-checkout-build-upsell-cta-component)

### DIFF
--- a/support-frontend/assets/components/button/upsellButton.tsx
+++ b/support-frontend/assets/components/button/upsellButton.tsx
@@ -1,0 +1,53 @@
+import type { SerializedStyles } from '@emotion/react';
+import { css, ThemeProvider } from '@emotion/react';
+import { from, neutral, space } from '@guardian/source-foundations';
+import {
+	Button,
+	buttonThemeReaderRevenueBrand,
+} from '@guardian/source-react-components';
+
+const button = css`
+	width: 100%;
+	justify-content: space-around;
+	color: ${neutral[7]};
+	margin: ${space[5]}px 0 ${space[3]}px;
+
+	${from.mobileMedium} {
+		margin-bottom: ${space[4]}px;
+	}
+
+	${from.desktop} {
+		margin: ${space[6]}px 0 ${space[5]}px;
+	}
+`;
+
+interface UpsellButtonProps {
+	buttonCopy: string | null;
+	handleButtonClick: () => void;
+	cssOverrides?: SerializedStyles;
+}
+
+function UpsellButton({
+	buttonCopy,
+	handleButtonClick,
+}: UpsellButtonProps): JSX.Element {
+	return (
+		<>
+			{buttonCopy && (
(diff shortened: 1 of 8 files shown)


marialani:
[V2 Checkout - build 'upsell CTA' component](https://trello.com/c/lXuc01lP/690-v2-checkout-build-upsell-cta-component)

paul-daniel-dempsey (commented):
All looks great, thanks for taking me through my first review. Looks like test credit card payments are going through fine now with the  state change (and looks like it will strip out window.guardian if was supplied in future plus an associated test). It was mentioned the upsell button is now needs to be hidden behind a (another?) variant,  would you like to complete that here or in a new request?  If a new variant would imagine it would use the participation object to test for another variant like  this? 
const isNewProduct = participations.newProduct === 'variant';

marialani:
> All looks great, thanks for taking me through my first review. Looks like test credit card payments are going through fine now with the state change (and looks like it will strip out window.guardian if was supplied in future plus an associated test). It was mentioned the upsell button is now needs to be hidden behind a (another?) variant, would you like to complete that here or in a new request? If a new variant would imagine it would use the participation object to test for another variant like this? const isNewProduct = participations.newProduct === 'variant';

I'd thought it was mentioned yesterday that the Upsell Button may or may not be on the final v2 supporter plus checkout page as opposed to it potentially being included in another variant. Let's check in on this at stand up. If we know for sure we won't be including it then I'll push up another commit to remove it from the supporter plus page storybook file

paul-daniel-dempsey (approved):
Can see lowerTier selected if equal to minimum contribution now too. Upsell component now disabled including need for text and click handler.

PR #4297, opened 2022-10-18
Fix integration tests

rupertbates:
#4261 introduced a mechanism to set the correct invoice template on a subscription, however it broke some integration tests because it used dummy data in the fixtures used by those tests to make calls made to Zuora whereas Zuora expects the invoice id which is passed in to actually exist.

[**Trello Card**](https://trello.com/c/PjWsKxcH/773-fix-it-test-regression)


paulbrown1982 (approved):
👍 Sorry! Maybe we should retro how to improve the alerts - in the old membership-frontend days failing post-deployment tests would "blame" a commit number.

PR #4298, opened 2022-10-19
Patron cancellations

rupertbates:
Currently we have a system to add Patrons customers into the SupporterProductData data store when they take out a subscription, but no way to remove them when they cancel. This PR provides that mechanism.

## Technical details
This PR consists of a lambda `PatronCancelledLambda` which is triggered by an API Gateway endpoint which in turn is called by a Stripe webhook whenever a subscription is cancelled (the webhook type is `customer.subscription.deleted`). 

This then finds the appropriate record in the Dynamo table and sets a cancellation date attribute to today.

[Details of how to test this are here](https://docs.google.com/document/d/1IbU_r-h71cIHbtSpWkm43E_TRTxoKNQtjb4-PekXRkI/edit#heading=h.z5ca9etiqjqp)

### DIFF
--- a/cdk/lib/__snapshots__/stripe-patrons-data.test.ts.snap
+++ b/cdk/lib/__snapshots__/stripe-patrons-data.test.ts.snap
@@ -47,7 +47,7 @@ exports[`The Stripe patrons data stack matches the snapshot 1`] = `
     },
   },
   "Resources": {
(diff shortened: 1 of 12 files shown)


pawelkrupinski (approved):
Scala part looks good to me.

PR #4305, opened 2022-10-20
Use default unhealthy alarm support frontend

LAKSHMIRPILLAI:

Enabling the GuUnhealthyInstancesAlarm which can be generated for us by the GuEc2App pattern

[**Trello Card**](https://trello.com/c/fONpZcAp/364-use-default-unhealthy-instances-alarm-generated-by-cdk-in-support-frontend)

## Why are you doing this?

http5xxAlarm that we have is good enough to capture most of the issues, so may be we can replace the custom alarms -no healthy and reduced number healthy instances for support-frontend with the default unhealthyInstancesAlarm one.So that would be sufficient for the project leading to reduction of noises.

We will utilise the GuUnhealthyInstancesAlarm which can be generated for us by the GuEc2App pattern and run it for one month and check  how well it captures issues compared to the other two custom one.Then we can decide on which one to continue with and discard the other.


michaelbjacobson (approved):
Good stuff 👍🏼

PR #4317, opened 2022-10-24
V2 Checkout: Sepa Form

GHaberis:

This PR introduces 2 new components: `SepaFormContainer` and `SepaForm`. The `SepaFormContainer` component manages state and the rendering of the presentational `SepaForm` component.  The `SepaForm` component is a copy of the existing `SepaForm`. I've copied it and co-located it with the `SepaFormContainer`, when copying it I also stripped the inline validation  in the old component as this will be addressed in a follow-up PR.

This PR does not include:
- Client side validation on the SEPA form.
 - Embedding the SEPA form in the new checkout, this is dependant on the creation of a PaymentMethod component that will wrap this SEPA form that hasn't been built yet.

[**Trello Card**](https://trello.com/c/3Urmo2zT/759-v2-checkout-build-sepa-form-component)

### DIFF
--- a/support-frontend/assets/components/sepaForm/SepaForm.tsx
+++ b/support-frontend/assets/components/sepaForm/SepaForm.tsx
@@ -6,9 +6,8 @@ import {
 	Select,
 	TextInput,
 } from '@guardian/source-react-components';
-import { isValidIban } from 'helpers/forms/formValidation';
 import { countries } from 'helpers/internationalisation/country';
-import { sortedOptions } from '../../../components/forms/customFields/sortedOptions';
+import { sortedOptions } from '../forms/customFields/sortedOptions';
 
 // -- Styles -- //
 const containerStyles = css`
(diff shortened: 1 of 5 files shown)


GHaberis:
[V2 Checkout - Build SEPA form component](https://trello.com/c/3Urmo2zT/759-v2-checkout-build-sepa-form-component)

support-frontend/assets/components/sepaForm/SepaForm.tsx
```diff
@@ -0,0 +1,120 @@
...
+						optional={false}
+						hideLabel={false}
+						label="Bank account holder name"
+						maxLength={40}
+						value={accountHolderName ?? undefined}
+						onChange={(e) => updateAccountHolderName(e.target.value)}
+					/>
+				</div>
+
+				<div />
```
i-hardy, line 69: Can we get rid of these empty divs?

support-frontend/assets/components/sepaForm/SepaForm.tsx
```diff
@@ -0,0 +1,120 @@
...
+const headerStyles = css`
+	${headline.xxxsmall({
+		fontWeight: 'bold',
+	})}
+`;
+const fieldsContainerStyles = css`
+	margin-top: ${space[4]}px;
+	> * + * {
+		margin-top: ${space[3]}px;
+	}
```
i-hardy, line 25: I'm not wild about this kind of complex selector. If we just need things in a column with regular space in between we can use the [Stack](https://guardian.github.io/source/?path=/docs/packages-source-react-components-stack--playground) component from Source.

i-hardy (approved):
I think this is okay to go in, but it'd be nice to clean up a few things carried over from the old form.

support-frontend/assets/components/sepaForm/SepaForm.tsx
GHaberis: Yep, done!

support-frontend/assets/components/sepaForm/SepaForm.tsx
GHaberis: Done!

PR #4318, opened 2022-10-24
Added first payment date to payment methods

LAKSHMIRPILLAI:
PR to include the first payment date to the payment methods like Credit card, Amazon Pay and Paypal

[**Trello Card**](https://trello.com/c/3zLebulV/789-supporter-plus-ty-confirmation-email-add-firstpaymentdate)

## Is this an AB test?
- [ ] Yes
- [x] No

### DIFF
--- a/support-workers/src/main/scala/com/gu/emailservices/SupporterPlusEmailFields.scala
+++ b/support-workers/src/main/scala/com/gu/emailservices/SupporterPlusEmailFields.scala
@@ -71,9 +71,16 @@ class SupporterPlusEmailFields(
             "payment method" -> "SEPA",
           ),
         )
-      case _: PayPalReferenceTransaction => Future.successful(List("payment method" -> "PayPal"))
-      case _: CreditCardReferenceTransaction => Future.successful(List("payment method" -> "credit / debit card"))
-      case _: AmazonPayPaymentMethod => Future.successful(List("payment method" -> "AmazonPay"))
+      case _: PayPalReferenceTransaction =>
+        Future.successful(List("payment method" -> "PayPal", "first payment date" -> formatDate(created.toLocalDate)))
+      case _: CreditCardReferenceTransaction =>
+        Future.successful(
+          List("payment method" -> "credit / debit card", "first payment date" -> formatDate(created.toLocalDate)),
+        )
+      case _: AmazonPayPaymentMethod =>
+        Future.successful(
+          List("payment method" -> "AmazonPay", "first payment date" -> formatDate(created.toLocalDate)),
+        )
     }
   }
 }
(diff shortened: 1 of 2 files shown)


support-workers/src/main/scala/com/gu/emailservices/SupporterPlusEmailFields.scala
```diff
@@ -71,9 +71,16 @@ class SupporterPlusEmailFields(
             "payment method" -> "SEPA",
           ),
         )
-      case _: PayPalReferenceTransaction => Future.successful(List("payment method" -> "PayPal"))
-      case _: CreditCardReferenceTransaction => Future.successful(List("payment method" -> "credit / debit card"))
-      case _: AmazonPayPaymentMethod => Future.successful(List("payment method" -> "AmazonPay"))
+      case _: PayPalReferenceTransaction =>
+        Future.successful(List("payment method" -> "PayPal", "first payment date" -> formatDate(created.toLocalDate)))
```
rupertbates, line 75: I think this is supposed to be `first_payment_date` @GHaberis can you confirm?

GHaberis: Not sure, the info in the ticket is "To fix this we need to add the API trigger property first_payment_date for ALL payment methods, currently we just do this for Direct Debit and SEPA where it is sent as 'first payment date'. Doing this should feed it through to the template". This was taken from the email chain where we discussed the issue. The other fields seem to contain spaces, eg. "payment method" so "first payment date" does not look unusual. 

@LAKSHMIRPILLAI did you get to test this in `CODE`? That should verify everything's working as expected and that the data is getting to Braze.

LAKSHMIRPILLAI: Yes, we tested this change in CODE .It successfully adds the first payment date in the email template.But for the payment method,I think  template in Braze was using "default_payment_method " where as it should be "payment_method".Michael would send a mail regarding that.

GHaberis: Ah ok, so it sounds like `first payment date` in the triggers maps to `first_payment_date` in the template and `payment method` maps to `payment_method`, so the template substitutes spaces for underscores. If we need to update the template to use `payment_method` instead of `deafult_payment_method` this can be arranged to the appropriate team.

LAKSHMIRPILLAI: WE tested the change in CODE and as a result the  Supporter plus TY Confirmation email now shows first payment date as well as payment method as in the screenshot below
image

GHaberis: That's great, thanks for confirming @LAKSHMIRPILLAI 👍

rupertbates (approved):
👍

PR #4323, opened 2022-10-25
Add validation mechanisms for payment method and other amount field

i-hardy:

This adds validation mechanisms for payment method- so we can display an error if no payment method is selected- and the 'other amount' field, including checking the max/min contribution amount.

## Why are you doing this?

This is part of building a more consistent and resilient validation system for the V2 checkout. I've broken out this PR largely because updating the type of the payment method state involved touching a lot of files with minor changes


GHaberis (approved):
Nice work 👍

PR #4327, opened 2022-10-26
Remove the gu_supporter_plus cookie for now

rupertbates:
We want to add a cookie (gu_supporter_plus) which enables the website to identify and treat a signed-out browser as hosting a recently subscribed "Supporter plus" product holder, however the new cookie has not yet been cleared by the data protection team and added to the consent manangement platform so we are commenting it out until that work is complete.

[**Trello Card**](https://trello.com/c/MgspqDBS/802-dont-set-gusupporterplus-cookie-until-it-is-added-to-the-consent-management-platform)


worldofpablo (approved):
LGTM

PR #4331, opened 2022-10-27
Add error summary component and form validation on submit for new checkout

i-hardy:

This adds

- An error summary component which will appear at the top of the form on validation and automatically receive keyboard focus, listing all errors as anchor links to the relevant input
- A `useFormValidation` hook to wrap any payment function that will submit the form, which will run form validation and only continue with form submission if the form state is valid.
- Some tweaks to error display based on validating the Stripe form (work in progress)

## Why are you doing this?

This is a key part of building the new form validation system. Having an error summary with anchor links makes it much, much, much easier for all users to tackle problems that prevent them from submitting the form. The summary is announced to screen readers and auto-focused to make it easy to immediately navigate to fields with errors.

The `useFormValidation` hook provides a simpler and more generic way to make payment dependent on form validation without having to include validation logic in every single payment button component.

## Is this an AB test?
- [ ] Yes
- [x] No

### Error summary
Screenshot 2022-10-27 at 11-26-23 Support the Guardian Make a Contribution

### Validation on the Stripe form
Screenshot 2022-10-27 at 11 27 16

### DIFF
--- a/support-frontend/assets/components/stripeCardForm/stripePaymentButton.tsx
+++ b/support-frontend/assets/components/stripeCardForm/stripePaymentButton.tsx
@@ -6,6 +6,7 @@ import {
 import type { StripeError } from '@stripe/stripe-js';
 import { useEffect, useState } from 'preact/hooks';
 import { DefaultPaymentButtonContainer } from 'components/paymentButton/defaultPaymentButtonContainer';
+import { useFormValidation } from 'helpers/customHooks/useFormValidation';
 import { Stripe } from 'helpers/forms/paymentMethods';
 import {
 	useContributionsDispatch,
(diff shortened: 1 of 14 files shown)


support-frontend/assets/components/stripeCardForm/stripePaymentButton.tsx
```diff
@@ -37,6 +38,21 @@ export function StripePaymentButton(): JSX.Element {
 		(state) => state.page.checkoutForm.payment.stripe,
 	);
 
+	const payWithStripe = useFormValidation(function pay() {
```
GHaberis, line 41: I know this works, but it seems unnecessary to pass a named function declaration as a parameter here, this could simply be an anonymous arrow function, and that feels the typical way we'd do this? Just interested if there's any reason we'd do it as you've done?

GHaberis: I can see a case that using a named function is more readable, eg. use form validate then pay.

support-frontend/assets/helpers/redux/checkout/checkoutSelectors.ts
```diff
@@ -42,35 +31,33 @@ export function getAllErrorsForContributions(
...
 		return {
 			email,
-			...getPaymentMethodErrors(state),
+			...nonPersonalDetailsErrors,
 		};
 	}
 	return {
 		firstName,
 		lastName,
 		email,
```
GHaberis, line 54: If we're in the US should 'state' be part of the personal details errors? As we collect this with the personal details?

i-hardy: Yes, I'm going to tackle this in a future PR as we also need to validate other address details in certain contexts.

support-frontend/assets/components/stripeCardForm/stripePaymentButton.tsx
i-hardy: Named functions create much more readable stack traces. Because the actual execution of the payment function is deferred until after validation, it'll be a bit clearer what's happening if the payment function throws an error.

support-frontend/assets/helpers/redux/checkout/checkoutSelectors.ts
GHaberis: Ok sounds good

GHaberis (approved):
Big improvement vs the current error messages 👍

PR #4334, opened 2022-10-27
Update 2nd Argument in Quantum Metric sendEvent call

GHaberis:

QM contacted us to request we update the 2nd argument in the `window.QuantumMetricAPI.sendEvent` call. Previously this had been `0` or `1`. `0` would indicate the event was not a conversion event, and 1 would indicate it was a conversion event. They've requested we change this, so now we still pass `1` for a conversion event, and non-conversion events will now be either `0` or `64`. `0` will indicate a non cart value event, and `64` will indicate a cart value event. "cart value" is the term QM use to indicate the value of the subscription/contribution a user is going to make.

I've also updated the QM script source.

[**Trello Card**](hhttps://trello.com/c/25ul9iPp/803-quantum-metric-cart-value-changes)

### DIFF
--- a/support-frontend/assets/helpers/tracking/quantumMetric.ts
+++ b/support-frontend/assets/helpers/tracking/quantumMetric.ts
@@ -318,11 +318,11 @@ function sendEventABTestParticipations(participations: Participations): void {
 
 function addQM() {
 	return loadScript(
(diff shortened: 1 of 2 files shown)


GHaberis:
[Quantum Metric cart value changes](https://trello.com/c/25ul9iPp/803-quantum-metric-cart-value-changes)

michaelbjacobson (approved):
👍🏼

PR #4342, opened 2022-10-28
Add validation to the payment request button

i-hardy:

This adds form validation behaviour to the payment request button. Validation is performed before the Apple/Google pay interface appears, to check any custom amount errors, and then again after completion, to make sure the user's details won't cause any problems downstream.

It also adds initial Zod-based address field validation for the state, country and zip/postal code, as we collect some of this information via the PRB, and refactors the validation selectors into a more sensible file structure.

## Is this an AB test?
- [ ] Yes
- [x] No

### Validation on button click
Support-the-Guardian-Make-a-Contribution (1)

NB. The issue with the error state on the custom amount field is addressed in guardian/source#1610

### Validation post-completion
Support-the-Guardian-Make-a-Contribution (2)


support-frontend/assets/helpers/internationalisation/shouldCollectStateForContribs.ts
```diff
@@ -0,0 +1,33 @@
+import type { CountryGroup, CountryGroupId } from './countryGroup';
+import {
+	AUDCountries,
+	Canada,
+	countryGroups,
+	UnitedStates,
+} from './countryGroup';
+
+export function shouldCollectStateForContributions(
```
i-hardy, line 9: This is the same logic that's inside the `ContributionsState` component to determine whether to display anything; we should be able to re-use it there in the future.

support-frontend/assets/components/subscriptionCheckouts/address/addressFields.tsx
```diff
@@ -149,7 +149,7 @@ export function AddressFields({ scope, ...props }: PropTypes): JSX.Element {
 				id={`${scope}-stateProvince`}
 				data-qm-masking="blocklist"
 				label={props.country === 'CA' ? 'Province/Territory' : 'State'}
-				value={props.state ?? ''}
+				value={props.state}
```
i-hardy, line 152: These changes are just because the Zod schema means these fields are no longer nullable, and ESLint was yelling.

support-frontend/assets/helpers/customHooks/useFormValidation.ts
```diff
@@ -1,28 +1,32 @@
...
  * Validates the form, and will only run the handler if the form is valid.
  */
-export function useFormValidation(
-	paymentHandler: (event: React.MouseEvent<HTMLButtonElement>) => void,
-): (event: React.MouseEvent<HTMLButtonElement>) => void {
-	const [clickEvent, setClickEvent] =
-		useState<React.MouseEvent<HTMLButtonElement> | null>(null);
+export function useFormValidation<
+	EventType extends PreventableEvent = React.MouseEvent<HTMLButtonElement>,
+>(paymentHandler: (event: EventType) => void): (event: EventType) => void {
```
i-hardy, line 20: I had to change the type signature here to handle Stripe's `StripePaymentRequestButtonElementClickEvent`, which for some reason does not inherit from `Event` or anything else. Most of the time we'll just default to the React button click event.

GHaberis (approved):
looks good to me 👍

PR #4352, opened 2022-11-01
Only perform validation on the currently selected payment method

i-hardy:

This amends the `validateForm` action to optionally include a payload containing the current `PaymentMethod`. This means that in payment method specific reducers we can avoid performing validation unless that method is the currently selected one.

This also adds an initial component to display payment failure errors at the bottom of the form, however this requires some further testing once we have all payment methods properly implemented.

## Why are you doing this?

This is to avoid the experience of the user failing to pay via one payment method due to validation issues, and then opening another payment method form and finding it's already full of error messages when they haven't even used it yet.

## Is this an AB test?
- [ ] Yes
- [x] No

## Screenshots

In this screenshot, the selected payment method is `None` (hence the 'please select a payment method' validation message). Accordingly, the Stripe card form has not been validated.

Screenshot 2022-11-01 at 15-05-29 Support the Guardian Make a Contribution


GHaberis (approved):
Looks good 👍

PR #4353, opened 2022-11-01
Add V2 Payment Method Selector

marialani:
This adds the new payment method selector component to the v2 checkout. The selector is made up of a source `<Accordion>` component placed inside a `<RadioGroup>` source component. Both of which come along with accessibility benefits baked in from the source design system. 

In order to fit our specific use case we have then created a bespoke 'accordion row' component. This is so each 'accordion row' will be able to house a source `<Radio>` input element as opposed to plain text. The v2 `paymentMethodData` object now has an optional `accordionBody` property which will render the optional accompanying accordion body as a dropdown. If there is not an  accordion body for a selected payment method we will not display a dropdown and only update the payment method redux state to reflect the newly selected payment method. 

[**Trello Card**](https://trello.com/c/7DcCUNj0/646-v2-checkout-build-the-payment-method-selector)

## Why are you doing this?
This adds the v2 payment method selector component in preparation for the upcoming supporter plus v2 checkout AB test.

## Screenshots
| mobile medium | desktop |
| -- | -- |
| Screenshot 2022-11-01 at 17 22 13 | Screenshot 2022-11-01 at 17 22 43 |

> Storybook Build: [Payment Method Selector Component](https://62e115310aef0868687b2322-xihousiojp.chromatic.com/?path=/story/checkouts-paymentmethodselector--default)
>
> *Things to note*: 
> - We have a myriad of payment methods and are displaying radio buttons for all possible payment methods
> - The Stripe elements do not currently appear in storybook but this is something we can fix later
>


marialani:
[V2 checkout - Build the payment method selector](https://trello.com/c/7DcCUNj0/646-v2-checkout-build-the-payment-method-selector)

support-frontend/assets/components/paymentMethodSelector/paymentMethodSelector.tsx
```diff
@@ -0,0 +1,170 @@
...
+}: PaymentMethodSelectorProps): JSX.Element {
+	const dispatch = useContributionsDispatch();
+
+	useEffect(() => {
+		availablePaymentMethods.length === 1 &&
+			dispatch(setPaymentMethod(availablePaymentMethods[0]));
+	}, []);
+
+	const fullExistingRecurringPaymentMethods =
+		contributionTypeIsRecurring && fullExistingPaymentMethods;
```
i-hardy, line 83:
```suggestion
	const fullExistingRecurringPaymentMethods =
		contributionTypeIsRecurring ? fullExistingPaymentMethods : [];
```
Logical AND means this variable could also end up being `false`- usually better to avoid mixing up types too much.

support-frontend/assets/components/paymentMethodSelector/paymentMethodSelector.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+export interface PaymentMethodSelectorProps {
+	cssOverrides?: SerializedStyles;
+	availablePaymentMethods: PaymentMethod[];
+	paymentMethod: PaymentMethod | null;
+	validationError: string | undefined;
+	fullExistingPaymentMethods?: RecentlySignedInExistingPaymentMethod[];
+	contributionTypeIsRecurring?: boolean;
+	existingPaymentMethod?: RecentlySignedInExistingPaymentMethod;
+	existingPaymentMethods?: ExistingPaymentMethod[];
```
i-hardy, line 63:
```suggestion
	fullExistingPaymentMethods: RecentlySignedInExistingPaymentMethod[];
	contributionTypeIsRecurring?: boolean;
	existingPaymentMethod?: RecentlySignedInExistingPaymentMethod;
	existingPaymentMethods: ExistingPaymentMethod[];
```
I'd suggest making the array props non-optional, and require the container to pass an empty array if necessary. Just makes the logic of mapping over them a little bit simpler.

support-frontend/assets/components/paymentMethodSelector/PaymentMethodSelectorContainer.tsx
```diff
@@ -0,0 +1,64 @@
...
+function PaymentMethodSelectorContainer({
+	render,
+}: {
+	render: (
+		paymentMethodSelectorProps: PaymentMethodSelectorProps,
+	) => JSX.Element;
+}): JSX.Element {
+	const contributionType = useContributionsSelector(
+		(state) => state.page.checkoutForm.product.productType,
+	);
```
i-hardy, line 18:
```suggestion
	const contributionType = useContributionsSelector(getContributionType);
```
The `getContributionType` selector guarantees that you'll receive a `ContributionType`, so no need to use an additional `isContribution` check later.

support-frontend/assets/components/paymentMethodSelector/paymentMethodSelectorAccordionRow.tsx
```diff
@@ -0,0 +1,206 @@
...
+				<div hidden={!expanded}>
+					i am the expanded body i am the expanded body i am the expanded body i
+					am the expanded body i am the expanded body Lorem ipsum dolor sit
+					amet, consectetur adipisicing elit. Ut expedita autem neque sed in
+					unde impedit molestiae? Asperiores quo accusamus exercitationem ex
+					omnis ipsum voluptatum accusantium culpa! Accusamus, repellat
+					consectetur! Lorem ipsum dolor sit amet consectetur adipisicing elit.
+					Quae, aspernatur et quas vero pariatur magnam dolor velit minus
+					temporibus ad veniam quaerat! Vero molestias quod modi fugit nihil et
+					ad.
```
i-hardy, line 152: Remove this?

i-hardy (approved):
This is looking fantastic, just a few suggestions to simplify the logic and types a bit.

support-frontend/assets/components/paymentMethodSelector/paymentMethodSelectorAccordionRow.tsx
marialani: This was the temp accordion body for existing payment methods. Has now been updated

PR #4364, opened 2022-11-03
Fix payment method selector bug

marialani:
[This PR](https://github.com/guardian/support-frontend/pull/4353/files#diff-7b8647d698721301523125caa4577622c2098c3a91ab14090f6e2d237787b92a) added the payment method selector for the v2 checkout and did some refactoring of the existing component. This PR reverts the existing payment method selector back to it's original state and removes a visual bug that had been introduced.

### DIFF
--- a/support-frontend/assets/components/subscriptionCheckouts/paymentMethodSelector.tsx
+++ b/support-frontend/assets/components/subscriptionCheckouts/paymentMethodSelector.tsx
@@ -6,10 +6,9 @@ import {
 	SvgDirectDebit,
 	SvgPayPal,
 } from '@guardian/source-react-components';
+import type { ReactNode } from 'react';
 import React from 'react';
 import Rows from 'components/base/rows';
-import { PaymentMethodLabel } from 'components/paymentMethodSelector/paymentMethodLabel';
-import { RadioWithImage } from 'components/paymentMethodSelector/radioWithImage';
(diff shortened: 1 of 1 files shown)


support-frontend/assets/components/paymentButton/defaultPaymentButtonContainer.tsx
```diff
@@ -4,6 +4,7 @@ import { currencies } from 'helpers/internationalisation/currency';
 import { getContributionType } from 'helpers/redux/checkout/product/selectors/productType';
 import { getUserSelectedAmount } from 'helpers/redux/checkout/product/selectors/selectedAmount';
 import { useContributionsSelector } from 'helpers/redux/storeHooks';
+import { shouldShowBenefitsMessaging } from 'pages/contributions-landing/components/DigiSubBenefits/helpers';
```
i-hardy, line 7: Doesn't need to be done right now, but we should really move this out of the digi subs page folder.

i-hardy (approved):

support-frontend/assets/components/paymentButton/defaultPaymentButtonContainer.tsx
marialani: These were actually from some incorrectly stashed changes but totally agree!

PR #4370, opened 2022-11-04
Add validation message to payment method selector

i-hardy:

This PR

- Adds validation to the payment method selector when the user has not chosen a payment method
- Updates Source dev kitchen to fix a small visual bug with the error state of the other amount field
- Only checks recaptcha validation for Stripe, as we are still using the old direct debit form

## Why are you doing this?

This is (hopefully) the last piece of form validation work for the new checkout (at least until we can add a better direct debit form).

## Is this an AB test?
- [ ] Yes
- [x] No

### Validating the payment method field
Screenshot 2022-11-04 at 11-24-32 Support the Guardian Make a Contribution

### DIFF
--- a/support-frontend/assets/components/paymentMethodSelector/paymentMethodSelector.tsx
+++ b/support-frontend/assets/components/paymentMethodSelector/paymentMethodSelector.tsx
@@ -97,6 +97,7 @@ export function PaymentMethodSelector({
 		<div css={container}>
 			<PaymentMethodSelectorLegend />
 			<RadioGroup
+				id="paymentMethod"
 				label="Select payment method"
 				hideLabel
 				error={validationError}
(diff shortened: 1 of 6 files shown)


support-frontend/assets/components/paymentMethodSelector/paymentMethodSelector.tsx
```diff
@@ -97,6 +97,7 @@ export function PaymentMethodSelector({
 		<div css={container}>
 			<PaymentMethodSelectorLegend />
 			<RadioGroup
+				id="paymentMethod"
```
marialani, line 100: Adding an id is great here, could also add `role='radiogroup'` which hadn't been ported over initially

marialani (approved):
This looks great! 👏

support-frontend/assets/components/paymentMethodSelector/paymentMethodSelector.tsx
i-hardy: Will do that in my next PR!

PR #4374, opened 2022-11-04
Fix issues around existing payment methods in the new checkout

i-hardy:

This PR:
- Fixes an issue where we were showing a permanent loading state for existing payment methods even when not retrieving them
- Simplfies the logic around whether we should show a user's existing payment methods or a sign-in link to reauthenticate if their payment methods are stale
- Fixes visual problems with the display of those payment method options and the reauthenticate link
- Adds payment buttons for the `ExistingCard` and `ExistingDirectDebit` payment methods so users can actually check out with these payment types

## Why are you doing this?

This is a virtually undocumented and very poorly understood aspect of the contributions checkout which we had largely neglected in building the new checkout so far. This PR should fix things ready for launch, but a lot of the code this behaviour relies on needs refactoring post-launch and we desperately need to thoroughly document what this feature is, how it works, and how to test it.

## Is this an AB test?
- [ ] Yes
- [x] No

### Existing card payment methods displayed in the payment selector
Screenshot 2022-11-04 at 13 30 27

### Sign-in link for users with a 'stale' existing payment method
Screenshot 2022-11-04 at 13 09 00

### DIFF
--- a/support-frontend/assets/components/paymentMethodSelector/paymentMethodSelectorAccordionRow.tsx
+++ b/support-frontend/assets/components/paymentMethodSelector/paymentMethodSelectorAccordionRow.tsx
@@ -83,9 +83,7 @@ const collapsedBody = css`
 interface ExistingPaymentMethodAccordionRowProps {
 	expanded: boolean;
 	preExistingPaymentMethod: RecentlySignedInExistingPaymentMethod;
-	updateExistingPaymentMethod?: (
-		existingPaymentMethod: RecentlySignedInExistingPaymentMethod,
-	) => void;
+	onChange: () => void;
 	paymentMethod: PaymentMethod | null;
 	existingPaymentMethod: RecentlySignedInExistingPaymentMethod | undefined;
 	accordionBody?: JSX.Element;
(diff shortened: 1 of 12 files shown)


support-frontend/assets/components/paymentMethodSelector/paymentMethodSelectorAccordionRow.tsx
```diff
@@ -96,53 +94,45 @@ export function ExistingPaymentMethodAccordionRow({
...
-
-				<div css={explainerListContainer}>
-					Used for your{' '}
-					{subscriptionsToExplainerList(
+					cssOverrides={[
+						radio,
+						...(checked && accordionBody ? [borderBottom] : []),
+						existingPaymentMethodOverrides,
+					]}
+					isSupporterPlus={true}
```
rupertbates, line 127: I can't quite see what this is doing, is it right that it's hardcoded?

rupertbates (approved):
Just one question about a hardcoded boolean but otherwise 👍

support-frontend/assets/components/paymentMethodSelector/paymentMethodSelectorAccordionRow.tsx
i-hardy: Seems to affect something about the component styling, I'm not entirely sure why it's there either 🤷

PR #4378, opened 2022-11-04
Fix issue with validation on the other amount field

i-hardy:

This fixes an issue where validating the other amount field could produce duplicate error messages in the summary, when the amount failed both the regex condition and the 'is not `NaN` when parsed as a float' condition.

I've removed the NaN check altogether as it's not actually necessary- the regex will permit _only_ a numeric string with a maximum of one decimal place, which will always successfully parse to a number.

## Screenshots

Screenshot 2022-11-04 at 14 41 38


GHaberis (approved):
Nice 👍

PR #4379, opened 2022-11-04
Fix Supporter Plus Single Contrib T&Cs styling

marialani:
This fixes the styling for the terms and conditions for a single contribution on the Supporter Plus checkout.

## Screenshots
Screenshot 2022-11-04 at 15 09 54


i-hardy (approved):
👍

PR #4380, opened 2022-11-04
Disable pre-render of the contribution landing page

i-hardy:

This disables pre-rendering (aka SSR, although it's not really) of the contributions landing page with the `EmptyContributionsLandingPage` component, ahead of A/B testing the new checkout.

[**Trello Card**](https://trello.com/c/dNBJJElZ)

## Why are you doing this?

Showing the pre-rendered version of the old checkout at first and then flashing to the very different new design creates a weird user experience that might unduly influence the A/B test.

After the conclusion of the test we'll need to think about what a pre-rendered version of the new checkout would need to look like before restoring this feature.


i-hardy:
[Turn off server side rendering of Contributions checkout for V2 AB test](https://trello.com/c/dNBJJElZ/682-turn-off-server-side-rendering-of-contributions-checkout-for-v2-ab-test)

marialani (approved):
LGTM

PR #4383, opened 2022-11-04
Use isSupporterPlusPurchase in new checkout

GHaberis:

Users checking out above the threshold whilst in the `supporterPlus` AB test's `variant` will still be allocated as a Recurring Contributor. I thought this change to `isSupporterPlusPurchase` would fix that: https://github.com/guardian/support-frontend/pull/4382, however it didn't. That's because `isSupporterPlusPurchase` is not used at the point we submit the new checkout form and assign the `productOptions` in the `getProductOptionsForSupporterPlusTest` function.

In this PR i've renamed `getProductOptionsForBenefitsTest` to `getProductOptionsForSupporterPlusTest`, and updated it to use `isSupporterPlusPurchase` as they do exactly the same checks.

i've also updated `isSupporterPlusPurchase` to return `false` immediately if the user is not in the `supporterPlus` AB test's `variant` cohort, this just saves is doing all the amount calculations to just return `false` if the test/variant.


marialani (approved):
👍

PR #4393, opened 2022-11-07
Fix Direct Debit overlay bug

marialani:
This fixes an issue with the payment processing overlay, which would appear immediately when attempting to pay via direct debit. The overlay would sit above the modal and therefore block the user from completing the form. 

The form validation hook now takes an optional `dispatchPaymentWaiting` argument and to prevent recaptcha being called on component load the `accordionBody` of the payment method selector is now a function.

Co-authored by: @GHaberis @i-hardy


rupertbates (approved):
👍

PR #4394, opened 2022-11-07
Fix issue with PayPal checkout not triggering the correct pending state

i-hardy:

This fixes a bug where completing the recurring checkout with PayPal on the new checkout did not trigger the 'payment processing' state before navigating to the thank you page on payment completion.


GHaberis (approved):
Great! 👍

PR #4395, opened 2022-11-07
Treat supporter plus as a contribution when considering payment method switches

rupertbates:
Supporter plus has the same available payment methods as a contribution so it should be treated as one when evaluating payment method switches. 

Prior to this PR it wasn't - it was being treated as a subscription. However subscriptions do not have a switch for Sepa or Apple pay payment methods so supporter plus purchases were failing with an 'Invalid Payment Method' error.


GHaberis (approved):
Looks good 👍

PR #4396, opened 2022-11-07
Add aria labels to app store logo links on the thank you page

i-hardy:

This is a small fix to address an accessibility issue on the thank you pages that was being flagged up in Storybook. The app store logo links now have screen reader accessible text.


marialani (approved):
👍

GHaberis (approved):
Looks good 👍

PR #4402, opened 2022-11-08
Fix issue with tab panel focus and issue with loading state after exiting a PRB interface

i-hardy:

This fixes:
- An issue where the `PaymentFrequencyTabs` tab panel was receiving manual keyboard focus, when it should only be able to receive focus programmatically
- An issue where opening a payment request interface (ie. Apple or Google Pay) and then closing it without paying left the checkout stuck in the 'processing' state


rupertbates (approved):
LGTM

marialani (approved):
👍

PR #4404, opened 2022-11-08
Prevent users who already have a recurring contribution from taking out another on the new checkout

i-hardy:

This fixes an issue where we were not preventing users who already have a recurring contribution from checking out with a new one. This is very much a quick-n-dirty patch as this information is stored in the user reducer which has not yet been migrated to Redux Toolkit and requires a significant refactor.

## Screenshots
Screenshot 2022-11-08 at 14 41 23

### DIFF
--- a/support-frontend/assets/helpers/redux/selectors/formValidation/personalDetailsValidation.ts
+++ b/support-frontend/assets/helpers/redux/selectors/formValidation/personalDetailsValidation.ts
@@ -40,3 +40,16 @@ export function getPersonalDetailsErrors(
 		...stateOrProvinceErrors,
 	};
 }
+
+export function getUserCanTakeOutContribution(
+	state: ContributionsState,
+): boolean {
+	const contributionType = getContributionType(state);
+	if (contributionType === 'ONE_OFF') {
+		return true;
+	}
+	const userIsAlreadyARecurringContributor =
+		state.page.user.isRecurringContributorError;
+
+	return !userIsAlreadyARecurringContributor;
+}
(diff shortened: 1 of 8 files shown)


support-frontend/assets/helpers/redux/selectors/formValidation/personalDetailsValidation.ts
```diff
@@ -40,3 +40,16 @@ export function getPersonalDetailsErrors(
 		...stateOrProvinceErrors,
 	};
 }
+
+export function getUserCanTakeOutContribution(
```
rupertbates, line 44: would be good to call this `getUserCanTakeOutRecurringContribution` because they are still able to take out a single contribution

support-frontend/assets/pages/supporter-plus-landing/components/existingRecurringContributorMessage.tsx
```diff
@@ -0,0 +1,35 @@
...
+export function ExistingRecurringContributorMessage(): JSX.Element | null {
+	const isNotAlreadyARecurringContributor = useContributionsSelector(
+		getUserCanTakeOutContribution,
+	);
+
+	function onClick() {
+		trackComponentClick('send-to-mma-already-contributor');
+	}
+
+	if (!isNotAlreadyARecurringContributor) {
```
rupertbates, line 19:
```suggestion
	const isAlreadyARecurringContributor = !useContributionsSelector(
		getUserCanTakeOutContribution,
	);

	function onClick() {
		trackComponentClick('send-to-mma-already-contributor');
	}

	if (isAlreadyARecurringContributor) {
```

rupertbates (approved):
Looks good, couple of suggestions.

support-frontend/assets/helpers/redux/selectors/formValidation/personalDetailsValidation.ts
i-hardy: The function accounts for this by always returning `true` for one-off contributions.

PR #4412, opened 2022-11-09
Clear other amount errors when switching product type

i-hardy:

This fixes a bug where error messages associated with the other amount field were not cleared when switching contribution frequency.


i-hardy:
['Other' amount errors aren't cleared down when switching between tabs](https://trello.com/c/XTtT7BwH/845-other-amount-errors-arent-cleared-down-when-switching-between-tabs)

PR #4414, opened 2022-11-09
Increase audience size of supporterPlus AB test

GHaberis:

This will allocate 20% of the whole audience into the test, with half of that ie. 10% seeing the variant


i-hardy (approved):
👍

PR #4416, opened 2022-11-09
Retrieve user type from identity on the contributions checkout

i-hardy:

This PR:
- Migrates our retrieval of user type based on email address to an RTK thunk using `createAsyncThunk`, and removes the old code
- Retrieves the user type on the contributions checkout once we have an email address, either from user input or via the payment request button

[**Trello Card**](https://trello.com/c/8wjsR0bL)

## Why are you doing this?

This should address an issue where we are not correctly displaying components on the contribution thank you page for new users, as we never actually try to determine the user type after initial page load.

## Is this an AB test?
- [ ] Yes
- [x] No

## Screenshots
Screenshot 2022-11-09 at 14-19-09 Support the Guardian Make a Contribution


support-frontend/assets/helpers/redux/checkout/personalDetails/thunks.ts
```diff
@@ -0,0 +1,44 @@
...
+		condition: (email, thunkApi) => {
+			const { userTypeFromIdentityResponse, isSignedIn } =
+				thunkApi.getState().page.checkoutForm.personalDetails;
+			const emailInvalid = !emailRules.safeParse(email).success;
+			const requestInFlight = userTypeFromIdentityResponse === 'requestPending';
+
+			if (isSignedIn || emailInvalid || requestInFlight) {
+				return false;
+			}
+		},
```
i-hardy, line 42: This `condition` option will prevent any actions being dispatched if it returns `false`, which effectively debounces getting the user type based on an entered email. See https://redux-toolkit.js.org/api/createAsyncThunk#canceling-before-execution

i-hardy:
[Components not quite right on the Thank You page](https://trello.com/c/8wjsR0bL/835-components-not-quite-right-on-the-thank-you-page)

GHaberis (approved):
Looks good to me 👍

PR #4417, opened 2022-11-09
supporterPlus-variant: send contribution conversion event to QM

GHaberis:

This PR makes an update so we now call `sendEventContributionCheckoutConversion` when the new `supporterPlus-variant` TY page loads. This `sendEventContributionCheckoutConversion` function sends a conversion event to QuantumMetric (QM) with the `amount`, `contributionType` and `currencyId`.

I've also had to dispatch an additional `setCheckoutFormHasBeenSubmitted` action when a user clicks the Paypal payment button for one off contributions. This action has a side affect of setting the `contributionAmount` value in Session Storage, which is retrieved from Session Storage on the TY page, before being sent to QM.

### DIFF
--- a/support-frontend/assets/components/payPalPaymentButton/payPalOneOffContainer.tsx
+++ b/support-frontend/assets/components/payPalPaymentButton/payPalOneOffContainer.tsx
@@ -9,6 +9,7 @@ import { payPalCancelUrl, payPalReturnUrl } from 'helpers/urls/routes';
 import {
 	createOneOffPayPalPayment,
 	paymentWaiting,
+	setCheckoutFormHasBeenSubmitted,
 } from 'pages/contributions-landing/contributionsLandingActions';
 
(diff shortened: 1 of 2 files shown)


support-frontend/assets/components/payPalPaymentButton/payPalOneOffContainer.tsx
```diff
@@ -32,6 +33,8 @@ export function PayPalButtonOneOffContainer(): JSX.Element {
 				cancelURL: payPalCancelUrl(countryGroupId),
 			}),
 		);
+
+		dispatch(setCheckoutFormHasBeenSubmitted());
```
paul-daniel-dempsey, line 37: can see uses reducer to set page.form.formData.checkoutFormHasBeenSubmitted

paul-daniel-dempsey (approved):

PR #4428, opened 2022-11-11
Fix some issues with form validation and the payment request button

i-hardy:

This tackles a couple of bugs related to validation and the payment request button.

We originally dispatched `validateForm` when clicking the PRB like we do for every other payment button, as we need to verify that the user hasn't entered an invalid custom amount. However this had the undesirable side effect of causing validation messages to show up beside personal details fields that the user doesn't actually need to fill out if they're paying with Apple/Google Pay (although it would not block them from making a payment and submitting the form). This changes the validation behaviour for the PRB so it _only_ validates the other amount field.

There was also an edge case where if the user opened the PRB interface but changed their mind, exited it, and tried to complete the form with another payment method, validation would behave abnormally as the form remained in a 'using the PRB' state. This adds a handler for the payment request `'cancel'` event so we can correctly respond to the user exiting the interface.

A lot of the Redux code relating to both payment request button state and the other amount field is still pretty sub-optimal, but this is something we can look at improving once we know the outcome of the test and can (hopefully) cease having to accommodate the old checkout code.

[**Trello Card**](https://trello.com/c/soqnkYlh)

### DIFF
--- a/support-frontend/assets/components/paymentRequestButton/hooks/usePaymentRequestEvent.ts
+++ b/support-frontend/assets/components/paymentRequestButton/hooks/usePaymentRequestEvent.ts
@@ -1,7 +1,11 @@
 import type { PaymentMethod, PaymentRequest } from '@stripe/stripe-js';
 import { useEffect, useState } from 'react';
 import { validateForm } from 'helpers/redux/checkout/checkoutActions';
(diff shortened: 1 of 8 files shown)


i-hardy:
[Fix validation issues with the payment request button](https://trello.com/c/soqnkYlh/868-fix-validation-issues-with-the-payment-request-button)

marialani (approved):
looks good 👍

PR #4445, opened 2022-11-16
Support for joint patrons

rupertbates:
Some patrons subscriptions have a joint account holder. Currently these are only recorded in Salesforce however we want to ensure they receive their digital benefits when they sign into the apps or website.

This PR adds a mechanism for adding joint patrons to the SupporterProductData Dynamo db table so that they receive benefits. That mechanism is as follows:
- add two metadata fields `joint_patron_email` (mandatory) and `joint_patron_name` (optional) to the metadata of the existing patron customer record
- this metadata is picked up by the stripe-patrons-data lambda and the identity account for the email is either retrieved from identity or a new account created
- a patron record is created in SupporterProductData for this identity id 

[There are more details in this document](https://docs.google.com/document/d/1IbU_r-h71cIHbtSpWkm43E_TRTxoKNQtjb4-PekXRkI/edit#)

[**Trello Card**](https://trello.com/c/oo6Qj3wi/59-patron-joint-accounts)

### DIFF
--- a/stripe-patrons-data/src/main/scala/com/gu/patrons/lambdas/PatronCancelledEventLambda.scala
+++ b/stripe-patrons-data/src/main/scala/com/gu/patrons/lambdas/PatronCancelledEventLambda.scala
@@ -12,14 +12,13 @@ import com.gu.patrons.model.{
   InvalidJsonError,
   InvalidRequestError,
   StageConstructors,
+  StripeCustomer,
   UserNotFoundIdentityError,
   UserNotFoundStripeError,
 }
-import com.gu.patrons.services.PatronsIdentityService
+import com.gu.patrons.services.{PatronsIdentityService, PatronsStripeService}
 import com.gu.supporterdata.model.Stage
 import com.gu.supporterdata.services.SupporterDataDynamoService
-import com.stripe.Stripe
-import com.stripe.model.Customer
 import com.stripe.net.Webhook
 import com.typesafe.scalalogging.StrictLogging
 import io.circe.Decoder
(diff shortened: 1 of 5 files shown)


stripe-patrons-data/src/main/scala/com/gu/patrons/lambdas/PatronCancelledEventLambda.scala
```diff
@@ -125,10 +142,21 @@ object PatronCancelledEventLambda extends StrictLogging {
     EitherT.fromEither(decode[PatronCancelledEvent](payload).left.map(e => InvalidJsonError(e.getMessage)))
   }
 
-  def getCustomer(config: PatronsStripeConfig, customerId: String): EitherT[Future, Error, Customer] = {
+  def getCustomer(config: PatronsStripeConfig, customerId: String): EitherT[Future, Error, StripeCustomer] = {
     logger.info(s"Attempting to retrieve customer with id $customerId from Stripe")
-    Stripe.apiKey = config.apiKey
-    EitherT.fromEither(Try(Customer.retrieve(customerId)).toEither.left.map(e => UserNotFoundStripeError(e.getMessage)))
+    val stripeService = new PatronsStripeService(config, runner)
+    val futureCustomer = stripeService.getCustomer(customerId).transform {
```
paulbrown1982, line 148: I don't understand if the Failure is being handled why this needs to be wrapped in a Try?

paulbrown1982 (approved):
👍 LGTM.

It didn't dawn on me until I read this that the process creates Identity accounts for Patrons who signs and are not a Guardian user. I can see where you've had to apply this to paid-for Patrons too!

stripe-patrons-data/src/main/scala/com/gu/patrons/lambdas/PatronCancelledEventLambda.scala
rupertbates: Because transform maps from a Try to a Try, it just allows us to use a custom error type for easier handling later on.

PR #4447, opened 2022-11-16
Add supporter plus support to acquisitions-firehose-transformer

tjmw:

Adding support for Supporter Plus to the acquisitions-firehose-transformer lambda.

[**Trello Card**](https://trello.com/c/hOVJ6jSz/926-ticker-include-supporter)

## Why are you doing this?

We want to include Supporter Plus in the ticker and RRCP super mode, which both rely on the acquisitions-firehose-lambda.

## Is this an AB test?
- [ ] Yes
- [x] No


tomrf1 (approved):
👍

PR #4470, opened 2022-11-23
Fix error in supporter-product-data caused by compression of a zuora http response

rupertbates:
Recently executions of the supporter-product-data-DEV (for example [this one](https://eu-west-1.console.aws.amazon.com/states/home?region=eu-west-1#/v2/executions/details/arn:aws:states:eu-west-1:865473395570:execution:supporter-product-data-DEV:15e359bc-8cef-46f3-782d-a37e9fb768a3)) step function have started failing with the error `The request signature we calculated does not match the signature you provided. Check your key and signing method.`
It turns out that this was actually because we were passing the wrong content length for the Zuora CSV results file we were trying to transfer to S3 and that the reason that the content length was wrong was because Zuora had stopped returning the `Content-length` header on the file response. 

As this was only happening in the sandbox environment it is not obvious why this started happening, also it only occurred when making a request via OkHttp, not when using Curl.
There are some differences in the Zuora response between production and sandbox, most noticably that sandbox is using Http 2 whereas prod is using Http 1.1.

I found [an issue on the OkHttp repo](https://github.com/square/okhttp/issues/1542) which described the same issue caused by a compressed response and provides a solution which is to add an additional header to the request. My best guess for why this error has suddenly started happening is that some change Zuora has made to their sandbox servers has caused responses to start being compressed by default whereas previously they weren't.

### DIFF
--- a/.sdkmanrc
+++ b/.sdkmanrc
@@ -1,0 +1,3 @@
+# Enable auto-env through the sdkman_auto_env config
+# Add key=value pairs of SDKs to use below
+java=11.0.16-amzn
--- a/supporter-product-data/cloudformation/cfn.yaml
+++ b/supporter-product-data/cloudformation/cfn.yaml
@@ -48,6 +48,8 @@ Resources:
           - Id: DeleteAllOldFiles
             Status: Enabled
             ExpirationInDays: 14
+            AbortIncompleteMultipartUpload:
+              DaysAfterInitiation: 1
 
   SupporterProductDataLambdaRole:
     Type: AWS::IAM::Role
--- a/supporter-product-data/src/main/scala/com/gu/services/S3Service.scala
+++ b/supporter-product-data/src/main/scala/com/gu/services/S3Service.scala
@@ -1,18 +1,16 @@
 package com.gu.services
 
-import com.amazonaws.event.ProgressEventType._
-import com.amazonaws.event.{ProgressEvent, ProgressListener}
 import com.amazonaws.regions.Regions
 import com.amazonaws.services.s3.AmazonS3ClientBuilder
 import com.amazonaws.services.s3.model.{GetObjectRequest, ObjectMetadata, PutObjectRequest}
 import com.amazonaws.services.s3.transfer.TransferManagerBuilder
 import com.gu.aws.CredentialsProvider
 import com.gu.supporterdata.model.Stage
+import com.typesafe.scalalogging.StrictLogging
 
 import java.io.InputStream
-import scala.concurrent.Promise
 
-object S3Service {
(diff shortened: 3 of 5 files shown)


supporter-product-data/cloudformation/cfn.yaml
```diff
@@ -48,6 +48,8 @@ Resources:
           - Id: DeleteAllOldFiles
             Status: Enabled
             ExpirationInDays: 14
+            AbortIncompleteMultipartUpload:
```
rupertbates, line 51: In my investigations I noticed that Amazon recommend this setting

supporter-product-data/src/main/scala/com/gu/services/S3Service.scala
```diff
@@ -24,32 +22,17 @@ object S3Service {
...
     objectMetadata.setContentLength(length)
+
     val putObjectRequest = new PutObjectRequest(bucketName(stage), filename, inputStream, objectMetadata)
-    val progressListener = new TransferProgressListener(filename)
-    putObjectRequest.setGeneralProgressListener(progressListener)
-    transferManager.upload(putObjectRequest)
-    progressListener.future
+    val transfer = transferManager.upload(putObjectRequest)
+
+    transfer.waitForCompletion()
```
rupertbates, line 32: The changes in this file are because the previous implementation was swallowing the error so it was hard to diagnose what was going wrong

supporter-product-data/src/main/scala/com/gu/services/ZuoraQuerierService.scala
```diff
@@ -22,6 +22,7 @@ class ZuoraQuerierService(val config: ZuoraQuerierConfig, client: FutureHttpClie
   val authHeaders = Map(
     "apiSecretAccessKey" -> config.password,
     "apiAccessKeyId" -> config.username,
+    "Accept-Encoding" -> "identity", // Required to ensure that response content-length header is available. We need this when we transfer results to S3. See https://github.com/square/okhttp/issues/1542 for details
```
rupertbates, line 25: This is the new header that fixes the problem

.sdkmanrc
```diff
@@ -0,0 +1,3 @@
+# Enable auto-env through the sdkman_auto_env config
```
tomrf1, line 1: did you mean to commit this?

tomrf1 (approved):
👍 good investigation

.sdkmanrc
rupertbates: Yes it's been hanging around for ages in my  folder and I think it's useful

PR #4473, opened 2022-11-24
Remove newProduct AB test and add switch for suppressing digi sub pages

i-hardy:

This removes all aspects of the old `newProduct` AB test from the contributions checkout, and moves the aspect of that test that hid links to the digi sub landing page to be behind a new feature switch instead.

The switch has been added to the switchboard in all environments, so can be tested with this branch in CODE.

[**Trello Card**](https://trello.com/c/IsN6u3MO)

## Why are you doing this?

We're no longer running any testing for offering benefits on the old contributions checkout, but need to retain the ability to easily hide links to the digital subscription pages.

### DIFF
--- a/support-frontend/assets/pages/subscriptions-landing/subscriptionsLanding.tsx
+++ b/support-frontend/assets/pages/subscriptions-landing/subscriptionsLanding.tsx
@@ -3,6 +3,7 @@ import Footer from 'components/footerCompliant/Footer';
 import Header from 'components/headers/header/header';
 import headerWithCountrySwitcherContainer from 'components/headers/header/headerWithCountrySwitcher';
 import Page from 'components/page/page';
+import { isSwitchOn } from 'helpers/globalsAndSwitches/globals';
 import {
 	AUDCountries,
 	Canada,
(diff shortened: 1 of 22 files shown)


i-hardy:
[Remove the newProduct AB test](https://trello.com/c/IsN6u3MO/876-remove-the-newproduct-ab-test)

support-frontend/assets/pages/subscriptions-landing/subscriptionsLanding.tsx
```diff
@@ -39,13 +40,13 @@ function SubscriptionsLandingPage({
 			NZDCountries,
 			International,
 		],
-		isNewProduct,
+		isNewProduct: hideDigiSub,
```
GHaberis, line 43: Would it be better to update the prop name `isNewProduct` to `hideDigiSub`?

i-hardy: Yes, I think I wanted to avoid rooting around in too many files but it'd be better to just go ahead and change the prop name to avoid future confusion.

GHaberis (approved):
Looks good

PR #4474, opened 2022-11-24
Supporter Plus V2

rupertbates:
This PR sets up a 0% AB test  which will allow us to create Supporter Plus V2 subs in production by opting into the test with a hash url eg. https://support.thegulocal.com/uk/contribute#ab-supporterPlusV2=variant

This means that we need to be able to create both V1 and V2 subs so there is quite a bit of code which we can strip out when we switch over to V2 completely.

## Some related docs:
[High level presentation on what V2 is and why it is needed](https://docs.google.com/presentation/d/1bTl2CT4w_ZiliBeMdLZLYauLexJLganGYKkJMsySXUc/edit#slide=id.p3)
[Spike doc](https://docs.google.com/document/d/1WeI2uPMX3UpB5BxYXtKu_CVX9sQnMou7DNAmzM91yCc/edit#)
[More detailed info](https://docs.google.com/document/d/1bZfMtPZBGmtk-IBm941xsrg79UmZ5T_4ajrHkMwP5is/edit#)


support-config/src/main/resources/touchpoint.PROD.conf
```diff
@@ -57,6 +57,10 @@ touchpoint.backend.environments {
       supporterplus {
         monthly.productRatePlanChargeId = "8a12865b8219d9b401822106194e64e3"
         annual.productRatePlanChargeId = "8a12865b8219d9b40182210618c664c1"
+        v2 {
```
rupertbates, line 60: v2 product ids, we can make these default and remove the old ones at some point

support-frontend/assets/helpers/abTests/abtestDefinitions.ts
```diff
@@ -59,4 +59,24 @@ export const tests: Tests = {
 		targetPage: pageUrlRegexes.contributions.allLandingPagesAndThankyouPages,
 		seed: 15,
 	},
+	supporterPlusV2: {
```
rupertbates, line 62: The ab test definition

support-models/src/main/scala/com/gu/support/catalog/Product.scala
```diff
@@ -41,28 +41,35 @@ case object SupporterPlus extends Product {
...
       id: String,
       billingPeriod: BillingPeriod,
+      versionOptions: SupporterPlusVersionOptions,
   ) =
     ProductRatePlan(
       id,
       billingPeriod,
       NoFulfilmentOptions,
-      NoProductOptions,
+      versionOptions,
```
rupertbates, line 50: I am using the `ProductOptions` concept to model two version of the same product. This is used by newspaper to differentiate different flavours of sub so works well here.

support-models/src/main/scala/com/gu/support/workers/ProductTypeRatePlans.scala
```diff
@@ -49,13 +49,27 @@ object ProductTypeRatePlans {
           (productRatePlan.readerType == Corporate && product.readerType == Corporate), // We don't care about the billing period for corporates
       )
 
-  def supporterPlusRatePlan(
+  def supporterPlusRatePlanV1(
```
rupertbates, line 52: Convenience constructors for supporter plus product rate plan objects

support-models/src/main/scala/com/gu/support/workers/states/SendThankYouEmailState.scala
```diff
@@ -34,6 +35,7 @@ object SendThankYouEmailState {
       paymentMethod: PaymentMethod,
       accountNumber: String,
       subscriptionNumber: String,
+      abTests: Option[Set[AbTest]],
```
rupertbates, line 38: We need to know whether the user who has created a sub is in the ab test variant so that we can tell which version they have purchased.

support-workers/cloud-formation/src/templates/cfn-template.yaml
```diff
@@ -31,6 +31,10 @@ Mappings:
       RedemptionTables:
       - arn:aws:dynamodb:*:*:table/redemption-codes-DEV
       - arn:aws:dynamodb:*:*:table/redemption-codes-UAT
+      S3Files:
```
rupertbates, line 34: support-workers now needs to be able to load the product catalog from S3 to be able to work out the price of a supporter plus sub and from that how much of the purchase price is actually a contribution.

support-workers/src/main/scala/com/gu/helpers/SupportWorkersV2Helper.scala
```diff
@@ -0,0 +1,8 @@
+package com.gu.helpers
+
+import com.gu.support.acquisitions.AbTest
+
+object SupportWorkersV2Helper {
+  def isV2SupporterPlus(abTests: Option[Set[AbTest]]) =
+    abTests.flatMap(_.find(test => test.name == "supporterPlusV2" && test.variant == "variant")).isDefined
```
rupertbates, line 7: This is needed in a few places so I've centralised the logic here

support-frontend/test/services/pricing/PriceSummaryServiceSpec.scala
```diff
@@ -33,10 +33,10 @@ class PriceSummaryServiceSpec extends AsyncFlatSpec with Matchers {
       )
 
     val paper = service.getPrices(Paper, List(discountPromoCode))
```
rupertbates, line 35: This file is just full of price updates in tests because the catalog fixtures have been updated for the first time in ages, ignore it

support-services/src/test/scala/com/gu/support/catalog/CatalogServiceSpec.scala
```diff
@@ -40,36 +40,36 @@ class CatalogServiceSpec extends AsyncFlatSpec with Matchers {
       Monthly,
       HomeDelivery,
       Everyday,
-    ) shouldBe Some(Price(67.99, GBP))
+    ) shouldBe Some(Price(73.99, GBP))
```
rupertbates, line 43: More price updates

undergroundquizscene (approved):

support-models/src/main/scala/com/gu/support/catalog/ProductOptions.scala
```diff
@@ -30,8 +30,15 @@ case object EverydayPlus extends PaperProductOptions(true)
 
 case object Everyday extends PaperProductOptions(false)
 
+sealed abstract class SupporterPlusVersionOptions extends ProductOptions
```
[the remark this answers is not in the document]
rupertbates: Hmm, interesting, I guess conceptually you'd want anything in this ADT to 'be' a version option rather than just mixing it in so maybe slightly more correct from that point of view? But functionally no it's identical.

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/SupporterPlusSubcriptionBuilder.scala
```diff
@@ -49,6 +56,46 @@ class SupporterPlusSubcriptionBuilder(
...
+    val todaysDate = dateGenerator.today
+
+    val contributionAmount = state.product.amount - getBaseProductPrice(productRatePlanId, state.product.currency)
+    val subscriptionData = subscribeItemBuilder.buildProductSubscription(
+      productRatePlanId = productRatePlanId,
+      ratePlanCharges = List(
+        RatePlanChargeData(
+          RatePlanChargeOverride(
+            contributionRatePlanChargeId,
+            price = contributionAmount, // Pass the amount of the cost which counts as a contribution into Zuora
```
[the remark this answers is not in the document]
rupertbates: The other charge, the supporter plus subscription charge, inherits the price from the catalog automatically when the sub is created. We override the contribution charge amount because it is variable depending on what the user has chosen to pay.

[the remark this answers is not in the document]
undergroundquizscene: Ah of course – thanks for the explanation

PR #4479, opened 2022-11-25
Handle bad email errors from Identity

LAKSHMIRPILLAI:

PR to handle bad email error from Identity

[**Trello Card**](https://trello.com/c/dmVAQEHA/873-handle-bad-email-error-from-identity)

## Why are you doing this?

We have the issue of "bad" email addresses being submitted on our checkouts /create endpoint.We frequently get alarms triggered by users unsuccessfully attempting to checkout on our recurring contributions / subscriptions checkouts. A lot of these alarms are down to users entering email addresses which Identity considers valid but Okta doesn't: [test@gmail.com](mailto:test@gmail.com)123.This issue occurs when we're posting to the endpoint https://idapi.code.dev-theguardian.com/guest when creating a guest identity account as part of the checkout process.

The issue we've had is in the past we've received a generic 500 from Identity which triggers a generic 500 alarm on our endpoint that calls Identity, and these generic alarms require investigation.
However Identity have recently made a change so now the /guest endpoint will return a 400 response with exactly the same body as the one returned by IDAPI's internal email validator: [guardian/identity: Pull Request 2157](https://github.com/guardian/identity/pull/2157)

This PR will
1) Report a meaningful validation error to the user and
 2) Not trigger an alarm in this scenario.
## Is this an AB test?
- [ ] Yes
- [x ] No

### DIFF
--- a/support-frontend/app/models/identity/responses/IdentityErrorResponse.scala
+++ b/support-frontend/app/models/identity/responses/IdentityErrorResponse.scala
@@ -29,12 +29,14 @@ object IdentityErrorResponse {
   case class IdentityError(
       message: String,
       description: String,
+      context: String = "primaryEmailAddress",
   )
 
   object IdentityError {
     object InvalidEmailAddress {
-      val message = "Registration Error"
-      val description = "Please sign up using an email address from a different provider"
+      val message = "Invalid emailAddress:"
+      val description = "Please enter a valid email address"
+      val context = "primaryEmailAddress"
       val errorReasonCode = "invalid_email_address"
     }
 
--- a/support-frontend/app/models/identity/responses/IdentityErrorResponse.scala
+++ b/support-frontend/app/models/identity/responses/IdentityErrorResponse.scala
@@ -29,20 +29,27 @@ object IdentityErrorResponse {
   case class IdentityError(
       message: String,
       description: String,
-      context: String = "primaryEmailAddress",
   )
 
   object IdentityError {
     object InvalidEmailAddress {
-      val message = "Invalid emailAddress:"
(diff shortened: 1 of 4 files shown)


support-frontend/app/models/identity/responses/IdentityErrorResponse.scala
```diff
@@ -29,12 +29,14 @@ object IdentityErrorResponse {
   case class IdentityError(
       message: String,
       description: String,
+      context: String = "primaryEmailAddress",
```
GHaberis, line 32: I'm not clear what this `IdentityError` object represents, but it looks generic enough to handle any Identity Error, so I'm not sure what impact adding a context here has?

support-frontend/app/models/identity/responses/IdentityErrorResponse.scala
```diff
@@ -29,12 +29,14 @@ object IdentityErrorResponse {
...
   )
 
   object IdentityError {
     object InvalidEmailAddress {
```
GHaberis, line 36: Are we categorising the bad email errors that come from Identity as `InvalidEmailAddress` errors? When I test the form now I don't see this `message` "Registration Error" or the `description` "Please sign up using an email address from a different provider", I see a generic error in the client...

Screenshot 2022-11-25 at 16 53 45

GHaberis: I think the `InvalidEmailAddress` error was originally set-up to handle email addresses that were on Identity's disallowed email provider list, rather than email errors originating from OKTA.

PR #4484, opened 2022-11-28
Fix warning about a discarded value

undergroundquizscene:
I believe what’s happening here is that the `true` doesn’t typecheck (as the type of the expression is `EitherT[Future, ..., Unit]`), and so the compiler allows the `true` to be discarded and replaced with unit, in a similar way to the description in [this stack overflow question](https://stackoverflow.com/questions/28564324/in-scala-why-can-i-use-unit-as-return-type-here).

Replacing `true` with `()` removes the warning.


rupertbates (approved):
👍 looks good

PR #4496, opened 2022-11-30
Enable SnapStart on support-workers lambdas

tomrf1:
we need to upgrade java first - `java8.al2 is not supported for SnapStart enabled functions.`
https://aws.amazon.com/blogs/aws/new-accelerate-your-lambda-functions-with-lambda-snapstart/
*Edit*: java has been upgraded

*Edit*: tested this in CODE - no significant difference in warm-up times.
But - looking at how each lambda spends its time, they all fetch `support-workers.private.conf` from S3 during a cold start and this takes 6-7 seconds. The bad part of the cold start for these lambdas is not the init phase, it's the s3 fetch.
Some ideas -
- don't use s3, maybe parameter store?
- load config once in the first lambda and pass it through the state machine

### DIFF
--- a/support-modules/aws/src/main/scala/com/gu/aws/AwsS3Client.scala
+++ b/support-modules/aws/src/main/scala/com/gu/aws/AwsS3Client.scala
@@ -14,7 +14,7 @@ object AwsS3Client extends AwsS3Client {
     AmazonS3ClientBuilder
       .standard()
       .withRegion(Regions.EU_WEST_1)
-      .withCredentials(CredentialsProvider)
+//      .withCredentials(CredentialsProvider)
       .build()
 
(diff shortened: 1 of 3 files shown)


rupertbates:
Loading the config once is an interesting idea but it would mean either having secrets visible in the state or encrypting them on the fly. 

6-7 seconds seems a long time to load a file from S3, I wonder why that is taking so long?

I'm not sure cold start is a massive issues for support-workers as it is in use so much that is usually warm but still would be good to improve it if there is a straightforward way.

i-hardy:
Hey Tom, same question as on the other PR- is this work something you're likely to get back to or should Contributions add this to our backlog?

PR #4498, opened 2022-11-30
Create new SupporterPlus Alarms

LAKSHMIRPILLAI:
Create New Supporter Plus alarms

## Is this an AB test?
- [ ] Yes
- [ ] No


support-workers/cloud-formation/src/templates/cfn-template.yaml
```diff
@@ -222,22 +222,51 @@ Resources:
...
+                  Value: Contribution
+                - Name: Stage
+                  Value: !Ref Stage
+              MetricName: PaymentSuccess
+              Namespace: support-frontend
+            Period: 300
+            Stat: Sum
+            Unit: Count
+          ReturnData: false
+        - Id: m2
```
rupertbates, line 253: So what is happening here is that we're adding a new metric to alarm on with a different `ProductType` dimension, is that right?

rupertbates (approved):
Looks good to me

support-workers/cloud-formation/src/templates/cfn-template.yaml
undergroundquizscene: The intention is to have the alarms that currently fire when there are no recurring contributions also count SupporterPlus – i.e. they should only alarm if there are no recurring contributions _and_ no SupporterPlus checkouts.

PR #4507, opened 2022-12-01
Fix - Digi Sub gifting client-side datepicker validation

paul-daniel-dempsey:
Fix - Digi Sub gifting client-side datepicker validation

Bug 1
The Digi Subs Gifting checkout allows the gifter to select a date in the future on which the giftee will receive the gift via email. However the form's client side validation is not blocking invalid dates and dates in the past from being submitted. eg. I can submit the form with 01/01/2022 as my chosen gift date, or 01/01/202.

If a date in the past is accidentally submitted the email will be sent immediately - Retention have received a bug report from the contact centre about such an incident.

Bug 2
There's also another bug on the form related to the date picker. If I click the "Check date" button we see a message "Your gift will be delivered on NaN undefined Nan"

Bug 3 
Calendar should not allow dates to be selected before the current day.

## Is this an AB test?
- [ ] Yes
- [x] No

Before
Screenshot_2022-11-30_at_15 08 12

After
Screenshot 2022-12-01 at 15 16 25

### DIFF
--- a/support-frontend/assets/components/datePicker/datePicker.tsx
+++ b/support-frontend/assets/components/datePicker/datePicker.tsx
@@ -77,7 +77,15 @@ class DatePickerFields extends Component<PropTypes, StateTypes> {
 	}
 
 	getDateString = (): string =>
-		`${this.state.year}-${this.state.month}-${this.state.day}`;
+		`${this.state.year}-${this.padZero(this.state.month)}-${this.padZero(
+			this.state.day,
+		)}`;
+
+	padZero = (singleDigit: string): string => {
+		const doubleDigit =
+			(singleDigit.length === 1 && +singleDigit < 10 ? '0' : '') + singleDigit;
+		return doubleDigit;
+	};
 
 	getDateConfirmationText = (): string => {
 		const { value } = this.props;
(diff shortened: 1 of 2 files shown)


support-frontend/assets/components/datePicker/datePicker.tsx
```diff
@@ -77,7 +77,15 @@ class DatePickerFields extends Component<PropTypes, StateTypes> {
...
-		`${this.state.year}-${this.state.month}-${this.state.day}`;
+		`${this.state.year}-${this.padZero(this.state.month)}-${this.padZero(
+			this.state.day,
+		)}`;
+
+	padZero = (singleDigit: string): string => {
+		const doubleDigit =
+			(singleDigit.length === 1 && +singleDigit < 10 ? '0' : '') + singleDigit;
+		return doubleDigit;
+	};
```
marialani, line 88: We could use the `padStart` method here
```suggestion
		`${this.state.year}-${this.state.month.padStart(2, '0')}-${this.state.day.padStart(2, '0')}`;
```

paul-daniel-dempsey: Ah yes, failed to find that, much better to use inbuilt. All changed, thanks.

PR #4518, opened 2022-12-06
Upgrade sbt-native-packager to 1.9.11

LAKSHMIRPILLAI:

## Is this an AB test?
- [ ] Yes
- [ ] No

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -22,7 +22,9 @@ addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.4.6")
 
 addSbtPlugin("com.gu" % "sbt-riffraff-artifact" % "1.1.18")
 
-addSbtPlugin("com.github.sbt" % "sbt-native-packager" % "1.9.9")
+addSbtPlugin("com.github.sbt" % "sbt-native-packager" % "1.9.11")
+//Fix for sbt-native-packager 1.9.11 upgrade errors
+ThisBuild / libraryDependencySchemes += "org.scala-lang.modules" %% "scala-xml" % VersionScheme.Always
 
 // needed for riffraff as it uses AWS S3 SDK 1.11 which uses some base64 methods that were removed in JDK11
 // this is not needed for the current build server, but it is a good reason to move to AWS SDK v2


undergroundquizscene:
I’m finding the problem a little hard to wrap my head around – would you mind explaining a bit more? I’ll outline my understanding so far.

It seems like we want to update `com.github.sbt:sbt-native-packager`, and the newer version depends on a newer version of `scala-xml`. However, other dependencies of ours (including the scala compiler!) depend on older versions of `scala-xml`. In erroring, sbt is warning us that version 2.1.0 of `scala-xml` may not be binary-compatible with the versions that our older dependencies request, so when the older dependencies call into that library things may error unexpectedly. However, sbt tells whether things are compatible using version numbers (as described [here](https://www.scala-lang.org/blog/2021/02/16/preventing-version-conflicts-with-versionscheme.html)), and we’re fairly confident from the person who made the GitHub issues that the older versions are actually binary-compatible. Therefore the change in this PR tells sbt ”don’t warn us about binary incompatibility for scala-xml, we’re confident we know what we’re doing”.

Have I understood correctly? If so, it worries me a little: how confident are we that the versions of scala-xml _are_ actually binary-compatible? Is there any alternative, such as providing our dependencies with different versions of scala-xml to use, or upgrading those dependencies?

undergroundquizscene (approved):

LAKSHMIRPILLAI:
> I’m finding the problem a little hard to wrap my head around – would you mind explaining a bit more? I’ll outline my understanding so far.
> 
> It seems like we want to update `com.github.sbt:sbt-native-packager`, and the newer version depends on a newer version of `scala-xml`. However, other dependencies of ours (including the scala compiler!) depend on older versions of `scala-xml`. In erroring, sbt is warning us that version 2.1.0 of `scala-xml` may not be binary-compatible with the versions that our older dependencies request, so when the older dependencies call into that library things may error unexpectedly. However, sbt tells whether things are compatible using version numbers (as described [here](https://www.scala-lang.org/blog/2021/02/16/preventing-version-conflicts-with-versionscheme.html)), and we’re fairly confident from the person who made the GitHub issues that the older versions are actually binary-compatible. Therefore the change in this PR tells sbt ”don’t warn us about binary incompatibility for scala-xml, we’re confident we know what we’re doing”.
> 
> Have I understood correctly? If so, it worries me a little: how confident are we that the versions of scala-xml _are_ actually binary-compatible? Is there any alternative, such as providing our dependencies with different versions of scala-xml to use, or upgrading those dependencies?

It seems, it is ok to use "org.scala-lang.modules" %% "scala-xml" % VersionScheme.Always at the project/plugins.sbt .Probably riskier to use it at the top-level build.sbt file. When it's in project/plugins.sbt, it only affects the compilation of the build.sbt file itself, not the build of the application code itself, so there's no risk of runtime problems when used at that level.

PR #4523, opened 2022-12-07
Fix - Thank You page

marialani:

## Is this an AB test?
- [ ] Yes
- [ ] No

### DIFF
--- a/support-frontend/assets/components/test-user-banner/testUserBanner.tsx
+++ b/support-frontend/assets/components/test-user-banner/testUserBanner.tsx
@@ -1,6 +1,8 @@
 import { css } from '@emotion/react';
 import { error, neutral, space, textSans } from '@guardian/source-foundations';
+import { useLocation } from 'react-router-dom';
 import { isTestUser } from 'helpers/user/user';
+import { ThankYouUserTypeSelector } from './thankYouUserTypeSelector';
 
 const testUserBannerStyles = css`
 	${textSans.large()};
@@ -8,15 +10,20 @@ const testUserBannerStyles = css`
 	color: ${neutral[100]};
 	text-align: center;
 	padding: ${space[2]}px 0;
+	font-weight: bold;
 `;
 
 export function TestUserBanner(): JSX.Element | null {
 	const testUser = isTestUser();
+	const location = useLocation();
+
+	const isThankYouPage = location.pathname.includes('thankyou');
 
 	if (testUser) {
 		return (
 			<div css={testUserBannerStyles}>
 				<p>You are a test user</p>
+				{isThankYouPage && <ThankYouUserTypeSelector />}
 			</div>
 		);
 	}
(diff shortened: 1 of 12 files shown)


support-frontend/assets/components/test-user-banner/testUserBanner.tsx
```diff
@@ -1,22 +1,29 @@
...
 	background-color: ${error[400]};
 	color: ${neutral[100]};
 	text-align: center;
 	padding: ${space[2]}px 0;
+	font-weight: bold;
 `;
 
 export function TestUserBanner(): JSX.Element | null {
 	const testUser = isTestUser();
+	const location = useLocation();
```
GHaberis, line 18: We could move the variable definitions for `location` and `isThankYouPage` inside the `if (testUser)` block, as we only use these 2 variables if this conditional is `true`.

marialani:
[Fix thank you page module copy for brand new supporter plus customers](https://trello.com/c/HEjsWiid/863-fix-thank-you-page-module-copy-for-brand-new-supporter-plus-customers)

support-frontend/assets/components/test-user-banner/testUserBanner.tsx
i-hardy: If `useLocation` is a hook then it can't be used inside a conditional block; hooks won't work properly if they're not called on every render.

support-frontend/assets/helpers/redux/checkout/personalDetails/thunks.ts
```diff
@@ -33,10 +33,12 @@ export const getUserTypeFromIdentity = createAsyncThunk<
 		condition: (email, thunkApi) => {
 			const { userTypeFromIdentityResponse, isSignedIn } =
 				thunkApi.getState().page.checkoutForm.personalDetails;
+			const { isStorybookUser } = thunkApi.getState().page.user;
+
```
i-hardy, line 37: I don't feel great about effectively baking our Storybook dependency into real application code- I think we should simply avoid running this request in the Storybook context- but I appreciate that might require too significant a refactor at this point.

i-hardy (approved):
I think this is okay to go in now, but I'd like to take some time in the near future to review the way we handle Redux dependencies in Storybook as it's obviously become a bit too messy. Something for next year!

support-frontend/assets/components/test-user-banner/testUserBanner.tsx
marialani: `useLocation` is a hook so have moved this out of the conditional block

support-frontend/assets/helpers/redux/checkout/personalDetails/thunks.ts
marialani: This sounds like a great approach, will look into avoiding running this request in the Storybook context

marialani:
> I think this is okay to go in now, but I'd like to take some time in the near future to review the way we handle Redux dependencies in Storybook as it's obviously become a bit too messy. Something for next year!

Definitely want to revisit the logic in the stories even if just to remove the duplication!

PR #4541, opened 2022-12-13
Batch Scala Steward updates & set their frequency

rupertbates:
This PR configures Scala Steward dependency update PRs to be batched (multiple dependencies per PR) rather than a separate PR being raised for each update.

It also set the PR frequency to one per day at 10am, this should allow us to then configure this pull request to be merged automatically if it passes tests.

### [Trello card](https://trello.com/c/AsUzwQdv/959-batch-scala-steward-updates)

### DIFF
--- a/.scala-steward.conf
+++ b/.scala-steward.conf
@@ -1,4 +1,4 @@
-pullRequests.frequency = "0 10 * * MON-FRI" #10 am Monday to Friday
+pullRequests.frequency = "0 10 * * 1-5" #10 am Monday to Friday
 pullRequests.grouping = [
   { name = "aws", "title" = "AWS Scala dependency updates", "filter" = [{"group" = "software.amazon"}, {"group" = "com.amazonaws"}] },
   { name = "non_aws", "title" = "Non AWS Scala dependency updates", "filter" = [{"group" = "*"}] }
--- a/.scala-steward.conf
+++ b/.scala-steward.conf
@@ -1,0 +1,5 @@
+pullRequests.frequency = "0 10 * * MON-FRI" #10 am Monday to Friday
+pullRequests.grouping = [
+  { name = "aws", "title" = "AWS Scala dependency updates", "filter" = [{"group" = "software.amazon"}, {"group" = "com.amazonaws"}] },
+  { name = "non_aws", "title" = "Non AWS Scala dependency updates", "filter" = [{"group" = "*"}] }
+]


undergroundquizscene (approved):
Looks good!

undergroundquizscene:
Although: I tested the cron expression online and it said that it’s non-standard. I think that’s due to the `MON-FRI` – do you know whether scala steward supports that?

rupertbates:
> Although: I tested the cron expression online and it said that it’s non-standard. I think that’s due to the `MON-FRI` – do you know whether scala steward supports that?

Where di you test it? I was using this site: https://crontab.guru/#0_10_*_*_MON-FRI

.scala-steward.conf
```diff
@@ -0,0 +1,5 @@
+pullRequests.frequency = "0 10 * * 1-5" #10 am Monday to Friday
```
rtyley, line 1: Just so you know, it looks like Scala Steward doesn't like this cron expression! You can see this by searching for `InvalidCron` in the [_Public Repos Scala Steward_](https://github.com/guardian/scala-steward-public-repos/actions/workflows/public-repos-scala-steward.yml) GitHub Action job logs:

image

The error in the logs is _"Either DayOfMonth and DayOfWeek must have a ? expression"_:

```
2023-01-24 12:20:47,380 INFO  Refresh cache of guardian/support-frontend
2023-01-24 12:20:47,464 INFO  Failed to parse .scala-steward.conf: cron4s.InvalidCron: Either DayOfMonth and DayOfWeek must have a ? expression: DownField(frequency),DownField(pullRequests)
```

`DayOfMonth` is the 4th value, and `DayOfWeek` is the 5th, instead of:

```
0 10 * * 1-5
```

...I think you probably want:

```
0 10 * ? 1-5
```

An asterisk `*` in the `DayOfMonth` field means "run _every_ day of the month", but that's incompatible with only running on weekdays (the `1-5`), so the question mark `?` is the approved way of denoting that you'll just accept whatever days of the month work with the day-of-week rule you've set, I think.

rtyley: @LAKSHMIRPILLAI points out in https://github.com/guardian/contributions-platform/pull/263 that `0 10 * ? 1-5` fails too!

I do think trying a few things out with the cron4s library in a little project might be the best way to work out what a valid cron is. It's unfortunate that cron syntax isn't completely standardised- there's even inconstancies about how many fields should be used. I can see that we've been using 5 fields, eg `0 10 * * 1-5`, but the examples in the cron4s docs all seem to use 6 fields: https://www.alonsodomin.me/cron4s/userguide/index.html#parsing - 

- Seconds
- Minutes
- Hour Of Day
- Day Of Month
- Month
- Day Of Week

PR #4542, opened 2022-12-13
Set up reducer for existing payment methods

i-hardy:

This adds an RTK slice for existing payment methods- card or direct debit payment methods that a user has already used to pay for another Guardian product, which we may offer them the option to re-use when they land on the support site from MMA. The conditional logic around when and how to fetch this data has been refactored into a thunk for better encapsulation. The supporter plus checkout code now uses this new slice.

I've also:
- Refactored the `PaymentMethodSelector` component to remove its explicit dependency on Redux and use the same component for regular and existing payment methods
- Fixed an accessibility issue with our payment SVG icons that was being flagged up in Storybook
- Added an existing payment to the `PaymentMethodSelector` stories, and a story for the component in an error state

**NOTE:** Existing payment methods are currently turned off via the switchboard as we can't use them for products other than contributions. There's a separate card to address that.

[**Trello Card**](https://trello.com/c/xjRHZXmA)

## Why are you doing this?

This functionality needed a refactor to include it properly with other payment methods and better encapsulate the fetching behaviour.

### DIFF
--- a/support-frontend/assets/components/paymentMethodSelector/PaymentMethodSelectorContainer.tsx
+++ b/support-frontend/assets/components/paymentMethodSelector/PaymentMethodSelectorContainer.tsx
@@ -65,7 +65,7 @@ function PaymentMethodSelectorContainer({
 		(state) => state.page.form,
 	);
 	const { existingPaymentMethods } = useContributionsSelector(
-		(state) => state.common,
+		(state) => state.page.checkoutForm.payment,
 	);
 	const { switches } = useContributionsSelector(
 		(state) => state.common.settings,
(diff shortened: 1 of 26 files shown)


support-frontend/assets/components/paymentMethodSelector/PaymentMethodSelectorContainer.tsx
```diff
@@ -76,28 +70,31 @@ function PaymentMethodSelectorContainer({
...
-			contributionType,
-			existingPaymentMethods,
-		),
 		validationError: errors?.[0],
-		pendingExistingPaymentMethods,
-		showReauthenticateLink: showReauthenticationLink(
-			contributionType,
-			existingPaymentMethods,
-		),
+		...getExistingPaymentMethodProps(existingPaymentMethods),
```
marialani, line 96: This is much nicer!

support-frontend/assets/helpers/redux/checkout/payment/existingPaymentMethods/reducer.ts
```diff
@@ -0,0 +1,59 @@
...
+		});
+
+		builder.addCase(getExistingPaymentMethods.fulfilled, (state, action) => {
+			const usablePaymentMethods = getUsableExistingPaymentMethods(
+				action.payload,
+			);
+			state.paymentMethods = usablePaymentMethods;
+			state.status = Completed;
+
+			// The user has at least one existing payment method, but it's not currently usable for payment
```
marialani, line 50: Could see a payment method wouldn't be usable if there is no billing account id. Was wondering what other factors might make the payment method usable?

i-hardy: The billing account ID is the only think that we actually _need_ in order to submit a payment with an existing payment method, and is how we can distinguish between the two types of payload MDAPI might return for a user's payment methods- see the `RecentlySignedInExistingPaymentMethod` vs `NotRecentlySignedInExistingPaymentMethod` types in the state file for this slice.

marialani (approved):
This looks great! And makes how we handle existing payment methods a lot less mysterious 🎉

PR #4553, opened 2022-12-14
Remove the supporterPlus AB test and only serve the supporter plus landing page

i-hardy:

This PR:

- Removes the `supporterPlus` AB test, and makes the supporter plus landing page the only thing rendered by the existing contributions entry point
- Amends the CSS reset and adds font-face declarations to the global CSS added by the `PageScaffold` component, to remove the need for the old SCSS code which handled these things
- Simplifies many of our Redux side effects by removing all uses of the `enableOrDisableForm` thunk, which was part of the old form validation system, and the implicit loading of the PayPal SDK when switching payment method, which is now handled by the `usePayPal` hook.

[**Trello Card**](https://trello.com/c/U6lfxNCp)

## Why are you doing this?

This means we are no longer shipping the old contributions checkout code, laying the groundwork to moving the entry point to the supporter plus page folder and entirely deleting all unused contributions code.


i-hardy:
[Supporter Plus to 100%: Remove SupporterPlus AB test.](https://trello.com/c/U6lfxNCp/961-supporter-plus-to-100-remove-supporterplus-ab-test)

GHaberis (approved):
Looks good! 👍

PR #4555, opened 2022-12-15
Use configured amounts in new checkout page

tomrf1:
currently the amounts are hardcoded. This PR changes it to use the amounts tests config from the RRCP (which the old contributions form was using)


GHaberis (approved):
Looks good 👍

PR #4558, opened 2022-12-15
Amend Selenium tests to work correctly with new checkout

i-hardy:

This updates the selectors in the contributions Selenium tests to reflect the new structures, IDs and other selectors on the SupporterPlus checkout, and adds a consistent test ID to the payment button(s).


rupertbates (approved):
👍

PR #4567, opened 2022-12-19
Increase the memory allocation of two lambdas

rupertbates:
Increase the memory allocation of two lambdas in the stripe-patrons-data project.

These lambdas are called by Stripe webhooks and occasionally the webhook times out so the call appears to have failed in Stripe. These usually succeed on retry but increasing the memory allocation of the lambda should speed the execution up and allow it to complete more reliably.


marialani (approved):
👍

PR #4568, opened 2022-12-19
Move supporter plus checkout to new Webpack entry point

i-hardy:

This PR:

- Creates a new Webpack entry point for the support plus landing page in place of the contributions landing page
- Creates new versions of key 'set up' files, such as the reducer set up, for the supporter plus landing page, with unused old code removed
- Moves the benefits threshold helper code out of the contributions landing folder and into a dedicated helper folder, and refactors it to separate concerns

[**Trello Card**](https://trello.com/c/tHYjrBaX)

## Why are you doing this?

This is a key part of the work to roll out the new checkout and remove the old contributions checkout code. The supporter plus checkout should now no longer be dependent on reaching into the `contributions-landing` folder.


support-frontend/assets/helpers/redux/checkout/product/contributionsSideEffects.ts
```diff
@@ -75,7 +75,7 @@ export function addProductSideEffects(
 	});
 
 	startListening({
-		type: 'SET_CHECKOUT_FORM_HAS_BEEN_SUBMITTED',
+		type: 'PAYMENT_WAITING',
```
i-hardy, line 78: I've completely removed the `setCheckoutFormHasBeenSubmitted` action as we no longer use that field in Redux; however it had an important side effect of setting the contribution amount in session storage for us to pick up and send to Quantum Metric following a one-off PayPal contribution (which involves a redirect via PayPal, meaning we lose all Redux state).

I've attached this side effect to the `paymentWaiting` action creator for now, as that is called before we redirect to PayPal. However it might be better to just make this an explicit effect of clicking the one-off PayPal payment button?

support-frontend/assets/components/payPalPaymentButton/payPalRecurringContainer.tsx
```diff
@@ -55,7 +63,13 @@ export function PayPalButtonRecurringContainer({
...
 		isTestUser: isTestUser ?? false,
 		setValidationControls,
-		onClick: () => dispatch(sendFormSubmitEventForPayPalRecurring()),
+		onClick: () =>
+			trackCheckoutSubmitAttempt(
+				`PayPal-${contributionType}-submit`,
+				`npf-allowed-for-user-type-${userTypeFromIdentityResponse}`,
+				'PayPal',
+				isSupporterPlus ? 'SupporterPlus' : 'Contribution',
+			),
```
i-hardy, line 72: This was the only thing the `sendFormSubmitEventForPayPalRecurring` action was actually doing as far as I can tell.

support-frontend/assets/pages/supporter-plus-landing/setup/legacyActionCreators.ts
```diff
@@ -0,0 +1,583 @@
+// ----- Imports ----- //
```
i-hardy, line 1: This is [contributionsLandingActions.ts](https://github.com/guardian/support-frontend/blob/9d8786c148917d6a399dd5c4a6ccd7ea1b960531/support-frontend/assets/pages/contributions-landing/contributionsLandingActions.ts) with unused action creators removed. Everything that remains is related to payment and form submission, which will need some focused work in the new year.

support-frontend/assets/pages/supporter-plus-landing/setup/legacyReducer.ts
```diff
@@ -0,0 +1,103 @@
+// ----- Imports ----- //
```
i-hardy, line 1: This is [contributionsLandingReducer.ts](https://github.com/guardian/support-frontend/blob/9d8786c148917d6a399dd5c4a6ccd7ea1b960531/support-frontend/assets/pages/contributions-landing/contributionsLandingReducer.ts), with unused state fields removed.

i-hardy:
[SupporterPlus to 100%: Create a new Webpack entry point for the Supporter Plus checkout](https://trello.com/c/tHYjrBaX/945-supporterplus-to-100-create-a-new-webpack-entry-point-for-the-supporter-plus-checkout)

support-frontend/assets/helpers/redux/checkout/product/contributionsSideEffects.ts
GHaberis: > However it might be better to just make this an explicit effect of clicking the one-off PayPal payment button?

Would this be to simply run this within an `onClick` handler to the PayPal payment button?

GHaberis (approved):
Looks good 👍

support-frontend/assets/helpers/redux/checkout/product/contributionsSideEffects.ts
i-hardy: Yes, although on further investigation I couldn't find anywhere we were actually retrieving the `contributionAmount` key from session storage- I'm not sure if this is intentional? If so we could remove this side effect entirely.

GHaberis: It's retrieved here https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/pages/contributions-landing/components/ContributionThankYou/ContributionThankYouPage.tsx#L35, so we still need it, but moving the logic to a `onClick` handler seems reasonable if this is the only event which has this side effect.

i-hardy: That page is no longer in use! I assume that needs to be replicated on the new thank you pages? We really do need all this Quantum Metric stuff documented somewhere.

If we're going to assume that `contributionAmount` will _always_ be in session storage- as that code does- we need to make sure to always put it there, so it _shouldn't_ be set purely in a click handler for one-off PayPal payment. We could use it as a fallback for retrieving the amount from Redux, but that would involve a bit more complexity than just copying and pasting over that old code.

support-frontend/assets/helpers/redux/checkout/product/selectors/isSupporterPlus.ts
```diff
@@ -0,0 +1,27 @@
...
+		state.common.internationalisation.countryGroupId,
+		contributionType,
+	);
+	const amount = getAmount(
+		state.page.checkoutForm.product.selectedAmounts,
+		state.page.checkoutForm.product.otherAmounts,
+		contributionType,
+	);
+
+	const amountIsHighEnough = !!(thresholdPrice && amount >= thresholdPrice);
```
paul-daniel-dempsey, line 24: Presume always best to explicitly define the return parameter, casting forces boolean return type

support-frontend/assets/pages/supporter-plus-landing/components/paymentTsAndCs.tsx
```diff
@@ -76,24 +85,14 @@ export function PaymentTsAndCs({
 		  )}`;
 
 	const currencyGlyph = glyph(detect(countryGroupId));
-	const thresholdPrice =
-		getThresholdPrice(countryGroupId, contributionType) ?? 1;
+	const thresholdPrice = getThresholdPrice(countryGroupId, contributionType);
```
paul-daniel-dempsey, line 88: Is this used anywhere, maybe kept for use later?

support-frontend/assets/pages/supporter-plus-landing/setup/setUserStateActions.ts
```diff
@@ -0,0 +1,39 @@
...
+import { stateProvinceFieldFromString } from 'helpers/internationalisation/country';
+import type { CountryGroupId } from 'helpers/internationalisation/countryGroup';
+import type { ContributionsState } from 'helpers/redux/contributionsStore';
+import { defaultUserActionFunctions } from 'helpers/user/defaultUserActionFunctions';
+import type { Action, UserSetStateActions } from 'helpers/user/userActions';
+
+// ----- Actions Creators ----- //
+// This action creator seems to be overridden in order to prevent setting an invalid string as the user's state
+// as we get this from window.guardian - see helpers/user/user.ts#140 on
+// It would make more sense to put this logic in the reducer itself when we refactor to RTK
```
paul-daniel-dempsey, line 11: Yes, can see most likely overwrites window.guardian.geoip.stateCode setup in user.init. Might be worth moving there at some point in future.

support-frontend/assets/helpers/redux/checkout/product/contributionsSideEffects.ts
GHaberis: OK it looks like there's a bug on [supporterPlusThankYou.tsx]( the https://github.com/guardian/support-frontend/blob/00468e02176ea94faed4a3982c2a585befcf8527/support-frontend/assets/pages/supporter-plus-thank-you/supporterPlusThankYou.tsx) for One Off Paypal payments, where it should be pulling the `contributionAmount` from session storage, but it isn't and therefore it's reporting the wrong amount...

GHaberis: When defining `amount` we should check `isOneOffPayPal` and if `isOneOffPayPal` is `true` we use the `contributionAmount` from session storage rather than the `getAmount` helper.

PR #4581, opened 2022-12-22
Fix bundle analyser gh action

marialani:
Our bundle analyser github action was being skipped for some time due to a conditional check in the yaml file. This conditional had been introduced to prevent certain actions running on a fork. Although, in the bundle analyser check the value we were attempting to access i.e. `github.event.pull_request.head.repo.owner.login` was undefined. This uses an alternative github context value `github.repository_owner` for the repo owner conditional checks in our github actions.

[**Trello Card**](https://trello.com/c/cEiFq7u9/964-support-frontend-fix-bundle-analyser-action)

### DIFF
--- a/.github/workflows/bundle-analyser.yml
+++ b/.github/workflows/bundle-analyser.yml
@@ -5,35 +5,48 @@ on:
       - main
 
 jobs:
-    build_check:
-      if: github.event.pull_request.head.repo.owner.login == 'guardian'
-      name: Bundle Analyser
-      defaults:
-        run:
+  context_check:
+    steps:
+      - name: Head owner
+        run: echo "head owner = ${{ github.event.pull_request.head.repo.owner.login }}"
+
+      - name: Base owner
+        run: echo "base owner = ${{ github.event.pull_request.base.repo.owner.login }}"
+
+      - name: Pull request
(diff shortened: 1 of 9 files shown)


marialani:
[Support-Frontend: Fix "Bundle Analyser" action](https://trello.com/c/cEiFq7u9/964-support-frontend-fix-bundle-analyser-action)

marialani:
@kelvin-chappell this may be of interest to you following https://github.com/guardian/payment-failure-comms/pull/95

michaelbjacobson (changes requested):
Looks great, but just one small thing: you've changed single quotes to double quotes in all relevant files, but `github.repository_owner == 'guardian'` is using single quotes. It would be nice to adjust this for consistency, especially as you've done so for other strings 🙂

michaelbjacobson (approved):
Fab!

marialani:
> Looks great, but just one small thing: you've changed single quotes to double quotes in all relevant files, but `github.repository_owner == 'guardian'` is using single quotes. It would be nice to adjust this for consistency, especially as you've done so for other strings 🙂

Turns out we do need to use the single quotes with string comparison here

PR #4596, opened 2022-12-30
Don’t do postcode lookup for long inputs

undergroundquizscene:
For certain very long input values, getAddress.io’s postcode lookup returns a 404 instead of the expected error. To guard against this, this change avoids calling the API for values too long to be a valid postcode.

I believe the longest valid postcodes are 8 characters; this check checks for inputs longer than 10 characters in case I’m mistaken or in case that changes.

This issue came up from an alarm investigation.

### DIFF
--- a/support-frontend/app/controllers/GetAddress.scala
+++ b/support-frontend/app/controllers/GetAddress.scala
@@ -3,6 +3,7 @@ package controllers
 import actions.CustomActionBuilders
 import com.gu.monitoring.SafeLogger
 import com.gu.monitoring.SafeLogger._
+import com.gu.rest.{CodeBody, WebServiceHelperError}
 import com.gu.support.getaddressio.GetAddressIOService
 import io.circe.syntax._
 import play.api.libs.circe.Circe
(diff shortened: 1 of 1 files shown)


undergroundquizscene:
[Don’t do postcode lookup for long inputs](https://trello.com/c/5rj97Ajx/976-dont-do-postcode-lookup-for-long-inputs)

undergroundquizscene:
It occurs to me that I’ve assumed we’re only calling this for UK postcodes: is that correct?

GHaberis:
> It occurs to me that I’ve assumed we’re only calling this for UK postcodes: is that correct?

Yes this is correct: https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/components/subscriptionCheckouts/address/addressFields.tsx#L92

undergroundquizscene:
> > It occurs to me that I’ve assumed we’re only calling this for UK postcodes: is that correct?
> 
> Yes this is correct: https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/components/subscriptionCheckouts/address/addressFields.tsx#L92

Thanks!

GHaberis (approved):
Looks good to me 👍

PR #4597, opened 2023-01-02
Propagate 4xx errors when getting user type

undergroundquizscene:
Before now, if we got a 4xx error from Identity (for example, if the email address we sent them was invalid), we would return a 500 to the client.

There are two problems with this. The first is that an invalid email address here is not a *server-side* error, so a 500 has the wrong meaning.

The second is that this triggers an alert we have set up to detect 5xx errors. We call this on the digital subscription checkout page when the user moves focus out of the email address input, and don’t show the user the resulting error immediately, so 5xx errors here produce unnecessary alerts.

This change instead returns the 4xx error we got from Identity to the frontend, avoiding alerting and returning a more semantically appropriate error.

### DIFF
--- a/support-frontend/app/controllers/IdentityController.scala
+++ b/support-frontend/app/controllers/IdentityController.scala
@@ -11,6 +11,7 @@ import io.circe.{Decoder, Encoder}
 import play.api.libs.circe.Circe
 import play.api.mvc._
 import services.IdentityService
+import services.GetUserTypeError._
 
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.Try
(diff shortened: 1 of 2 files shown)


support-frontend/app/controllers/IdentityController.scala
```diff
@@ -57,9 +57,23 @@ class IdentityController(
       identityService
         .getUserType(email)
         .fold(
-          err => {
-            SafeLogger.error(scrub"Failed to retrieve user type for $email: ${err.toString}")
-            InternalServerError
+          _ match {
+            case GetUserTypeGotErrorResponse(response) =>
+              if (response.status >= 400 && response.status < 500) {
+                SafeLogger.info(s"4xx error when retrieving user type for $email: ${response.body}")
```
undergroundquizscene, line 63: Is it ok not to be scrubbing this info message here? I’ve changed it to info because it’s not an error, but I’m slightly worried that using `s""` instead of `scrub""` is an information leak.

undergroundquizscene:
I have tested on code, confirming that the email address validation on submission still works with the changed response code while writing. The response passed back to the frontend from Identity looks like this:

```json
{"status":"error","errors":[{"message":"Bad Request","description":"Invalid email format"}]}
```

support-frontend/app/controllers/IdentityController.scala
michaelbjacobson: My concern here is that our logging level for application logs on AWS is `error`, hence why we scrub strings logged as errors. While this isn't technically an error, we need to think about whether it's something we want to have logged in production. I'd suggest keeping it as an error-type log message, and ensure we scrub the string.

michaelbjacobson (changes requested):
Looks good, but please see my comment about logging

michaelbjacobson (approved):

support-frontend/app/controllers/IdentityController.scala
undergroundquizscene: We discussed this on a video call: on reflection we’re pretty sure we’ve seen info-level logs in AWS, so this explanation isn’t quite right. However, it may be that we only _alert_ on levels higher than info, and so that’s why. I will check with others who may know more.

We also agreed that `warn` is probably a more correct error level for this situation than `info`, which I’ll change now.

PR #4599, opened 2023-01-03
Update card expiry in IT tests

tomrf1:
the tests are failing because of this


undergroundquizscene:
There’s a similar CreditCardReferenceTransaction with an expiration date of December 2022 in `support-models/src/test/scala/com/gu/support/zuora/api/Fixtures.scala` – should that be changed as well?

tomrf1:
> There’s a similar CreditCardReferenceTransaction with an expiration date of December 2022 in `support-models/src/test/scala/com/gu/support/zuora/api/Fixtures.scala` – should that be changed as well?

not sure it's relevant to these tests, but I've pushed an update!

undergroundquizscene (approved):

PR #4602, opened 2023-01-03
Update card expiry in IT tests (again)

tomrf1:
Changes the test credit card expiry to 2041, for consistency with the one in Stripe.
Changes the `PreparePaymentMethodForReuseSpec` expiry back to 2022, because we can't change the expiry date in zuora. It's only testing cloning of a payment method so the expiry doesn't matter.


graham228221 (approved):
Thanks Tom, we got there eventually! (hopefully)

PR #4610, opened 2023-01-04
Fix problem with Patron cancellations

rupertbates:
We have been getting a number of alarms caused by incomplete data for Patrons subscriptions in the SupporterProductData dynamo table. Upon investigation it appears that what is happening is the following:

- All subscriptions have a [TTL (time to live)](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/TTL.html) setting in Dynamo which is set to the term end date of the subscription. This means that when this date is reached the record will automatically be deleted from the data store.
- There is also a Stripe webhook which triggers when a subscription is cancelled to enable us to mark these subs as cancelled by adding a cancellation date to the record - this is mostly to take care of accounts which are cancelled mid-term
- It appears that in some (all?) cases the webhook triggers after the record has been deleted which due to the nature of Dynamo creates a new record with just the partition key, the sort key and the cancellation date - this then causes an error when members-data-api tries to read this record as it has required fields missing.

This PR fixes this issue by checking that a subscription actually exists in Dynamo before attempting to update it with a cancellation date. There will also be some cleanup required for corrupted records which I will do manually.

### DIFF
--- a/stripe-patrons-data/src/test/scala/com/gu/patrons/services/SupporterDataDynamoServiceSpec.scala
+++ b/stripe-patrons-data/src/test/scala/com/gu/patrons/services/SupporterDataDynamoServiceSpec.scala
@@ -1,0 +1,37 @@
+package com.gu.patrons.services
+
+import com.gu.supporterdata.model.Stage.DEV
+import com.gu.supporterdata.services.SupporterDataDynamoService
+import com.gu.test.tags.annotations.IntegrationTest
+import org.scalatest.flatspec.AsyncFlatSpec
+import org.scalatest.matchers.should.Matchers
+
+@IntegrationTest
+class SupporterDataDynamoServiceSpec extends AsyncFlatSpec with Matchers {
+  "SupporterDataDynamoService" should "return false if a subscription doesn't exist" in {
+    val identityId = "non_existent_identity_id"
+    val subscriptionId = "non_existent_subscription_id"
+    val dynamoService = SupporterDataDynamoService(DEV)
+    dynamoService.subscriptionExists(identityId, subscriptionId).flatMap {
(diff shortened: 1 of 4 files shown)


stripe-patrons-data/src/test/scala/com/gu/patrons/services/SupporterDataDynamoServiceSpec.scala
```diff
@@ -0,0 +1,37 @@
...
+      case Right(userExists) if userExists =>
+        fail(s"Subscription $subscriptionId should not exist")
+      case Right(_) =>
+        succeed
+      case Left(errorMessage) => fail(errorMessage)
+    }
+  }
+  "SupporterDataDynamoService" should "return true if a subscription does exist" ignore {
+    // This test relies on a record being present in the dev data store so
+    // I'm marking it as ignored in case that record gets deleted
```
undergroundquizscene, line 25: What does the ignored test mean? Is it that we won’t be notified of a test failure, or something less strong than that?

undergroundquizscene: Ah, I looked it up: the test will always report as Ignored. Why have the test if it’ll always be ignored?

rupertbates: It just means that it won't run unless the `ignored` keyword is switched back to `in`, so it won't be run by the scheduled integration test runner. It's arguable whether I should just delete it if it's not going to be run regularly but I like to have tests around which allow me to run individual bits of code easily so on balance I think it's worth keeping.

PR #4615, opened 2023-01-05
Quantum Metric: Get contribution amount from session storage for one off paypal contributions

GHaberis:

There's a bug on [supporterPlusThankYou.tsx](https://github.com/guardian/support-frontend/blob/00468e02176ea94faed4a3982c2a585befcf8527/support-frontend/assets/pages/supporter-plus-thank-you/supporterPlusThankYou.tsx) for One Off Paypal contributions, it's passing the wrong `amount` to Quantum Metric via the `sendEventContributionCheckoutConversion` function.

 It should be pulling the `contributionAmount` from session storage if `isOneOffPayPal` is true, but we're not making the check so it looks like it's pulling the `amount` from Redux, which is set to the default and therefore it's reporting the wrong amount to Quantum Metric.

This PR gets the`contributionAmount` from session storage if `isOneOffPayPal` is true and assigns that to the `amount` variable on the `supporterPlusThankYou.tsx` page. 

We have to retrieve the amount from session storage as the Thank You page is displayed via a new page load following a redirect via the Paypal site for  for one off paypal contributions, so the amount contributed is not held in the redux store as it is for all other payment types/frequency combinations.

[**Trello Card**](https://trello.com/c/WEXCLqCj/970-quantum-metric-sendevent-for-paypal-single-contribution-conversions-reporting-wrong-contribution-amount)

### DIFF
--- a/support-frontend/assets/pages/supporter-plus-thank-you/components/thankYouHeader/heading.tsx
+++ b/support-frontend/assets/pages/supporter-plus-thank-you/components/thankYouHeader/heading.tsx
@@ -17,7 +17,7 @@ const amountText = css`
 interface HeadingProps {
 	name: string | null;
 	isOneOffPayPal: boolean;
-	amount: number;
+	amount: number | undefined;
 	currency: IsoCurrency;
 	contributionType: ContributionType;
 }
--- a/support-frontend/assets/pages/supporter-plus-thank-you/components/thankYouHeader/thankYouHeader.tsx
+++ b/support-frontend/assets/pages/supporter-plus-thank-you/components/thankYouHeader/thankYouHeader.tsx
@@ -40,7 +40,7 @@ type ThankYouHeaderProps = {
 	showDirectDebitMessage: boolean;
 	isOneOffPayPal: boolean;
 	contributionType: ContributionType;
-	amount: number;
+	amount: number | undefined;
 	currency: IsoCurrency;
 	shouldShowLargeDonationMessage: boolean;
 	amountIsAboveThreshold: boolean;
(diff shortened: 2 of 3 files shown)


GHaberis:
[Quantum Metric: sendEvent for Paypal Single Contribution conversions reporting wrong contribution amount](https://trello.com/c/WEXCLqCj/970-quantum-metric-sendevent-for-paypal-single-contribution-conversions-reporting-wrong-contribution-amount)

undergroundquizscene (approved):
Why is it that the amount is in the session storage for one-off paypal contributions only?

paul-daniel-dempsey (approved):

paul-daniel-dempsey (approved):
Ditto Emily's question, why does Paypal on-off comes from session storage rather than redux? Just not been converted over yet am guessing?

GHaberis:
@paul-daniel-dempsey @undergroundquizscene users making one-off contributions with Paypal are actually taken from the support site onto Paypal's site where they complete their transaction, when they complete the transaction on Paypal's site they're redirected back to the support site, however the Redux store in memory in which all this info typically sits, doesn't persist. To get around this we store useful bits of info in session storage on the checkout, which does persist for the duration of that session, and we can retrieve this info on the Thank You page.

paul-daniel-dempsey:
Ah ok, thanks. 

Can see the equivalent setSession(‘contibutionsAmount’) is some sort of listener/side-effect setup on the redux store, am guessing when a value is changed there, the session updates too.

GHaberis:
> Ah ok, thanks. Can see the equivalent setSession(‘contibutionsAmount’) is some sort of listener/side-effect setup on the redux store, am guessing when a value is changed there, the session updates too.

Yes that's correct, that's when we store the value in Session Storage.

PR #4621, opened 2023-01-06
Update github actions to use Node 16

GHaberis:

We're seeing Github Actions fail with a warning related to dependent actions they use: `actions/setup-node@v2` and `actions/setup-node@v2`. Both these actions used Node 12 which is now deprecated (see screenshot below). Updates to https://github.com/actions/checkout and https://github.com/actions/setup-node to there latest `v3` versions should resolve this.

Screenshot 2023-01-06 at 14 32 24


michaelbjacobson (approved):
Looks good to me ✅

PR #4622, opened 2023-01-06
Update github actions to use Node 16

GHaberis:

Follow up to https://github.com/guardian/support-frontend/pull/4621, we also need to update the dependent action `actions/setup-java@v2` to `actions/setup-java@v3` (https://github.com/actions/setup-java), as `v2` uses a deprecated version of Node.


marialani (approved):
👍

PR #4623, opened 2023-01-06
Quantum Metric payment method

paul-daniel-dempsey:
Quantum Metric set-up a Payment Method Click event, which should be called when a user clicks on a payment method. The syntax required for supporting this event is:

QuantumMetricAPI.sendEvent(103,0,"Insert Name of Payment Method Here");

eg : DirectDebit , Stripe, PayPal (selection passed through to Quantum Metric)
Screenshot 2023-01-06 at 19 15 33

[**Trello Card**](https://trello.com/c/vWt8G9MY/954-quantum-metric-send-payment-method-event-to-qm)

## Why are you doing this?
On the Supporter Plus (variant) checkout It would be useful to send the name of the payment method selected to Quantum Metric.

## Is this an AB test?
- [ ] Yes
- [x] No

### DIFF
--- a/support-frontend/assets/components/paymentMethodSelector/PaymentMethodSelectorContainer.tsx
+++ b/support-frontend/assets/components/paymentMethodSelector/PaymentMethodSelectorContainer.tsx
@@ -14,6 +14,7 @@ import {
 	useContributionsDispatch,
 	useContributionsSelector,
 } from 'helpers/redux/storeHooks';
+import { sendEventContributionPaymentMethod } from 'helpers/tracking/quantumMetric';
 import type { PaymentMethodSelectorProps } from './paymentMethodSelector';
 
 function getExistingPaymentMethodProps(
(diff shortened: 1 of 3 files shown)


GHaberis (approved):
Code looks good to me 👍

GHaberis:
@paul-daniel-dempsey this looks good, regarding how to verify it's working, I've just been adding a log to the `sendEvent` function here https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/helpers/tracking/quantumMetric.ts#L93 when developing locally to confirm the correct values are passed to `window.QuantumMetricAPI.sendEvent`. Once released we can notify QM who'll confirm they're seeing the data as expected.

paul-daniel-dempsey:
> @paul-daniel-dempsey this looks good, regarding how to verify it's working, I've just been adding a log to the `sendEvent` function here https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/helpers/tracking/quantumMetric.ts#L93 when developing locally to confirm the correct values are passed to `window.QuantumMetricAPI.sendEvent`. Once released we can notify QM who can then confirm they're seeing the data as expected.

Ok, yes much better than my temporary console logs there. UI tests seem to have locked up, will sort & check that then get put into prod.

paul-daniel-dempsey:
[Quantum Metric: send payment method event to QM](https://trello.com/c/vWt8G9MY/954-quantum-metric-send-payment-method-event-to-qm)

PR #4628, opened 2023-01-09
Change lambda scheduling

rupertbates:
Costs for running  the supporter product data update step function have increased out AWS  bill noticeably and are the single largest step function cost by a long way.

While it is important that the SupporterProductData dynamo table (which is used by members-data-api to work out which products a user has) is kept up to date, it is already updated separately by support-workers when a purchase is made via the support site so it is only for purchases made via CSRs that we are wholly reliant on this step function. Since these are much lower volume and due to their nature (via a phone call) are probably less time sensitive I think it is reasonable to decrease the frequency of the execution of this step function. I have also decreased the frequency in the DEV and UAT environments to once a day.

I have also reduced the frequency of the stripe-patrons-data scheduled lambda as acquisition of patrons is very low volume so we are probably wasting money buy running it more frequently - ultimately we should change this to a Stripe webhook and removed the scheduling completely.

### DIFF
--- a/cdk/lib/__snapshots__/stripe-patrons-data.test.ts.snap
+++ b/cdk/lib/__snapshots__/stripe-patrons-data.test.ts.snap
@@ -756,9 +756,9 @@ exports[`The Stripe patrons data stack matches the snapshot 1`] = `
       },
       "Type": "AWS::IAM::Role",
     },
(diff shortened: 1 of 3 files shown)


tomrf1 (approved):
👍

PR #4630, opened 2023-01-10
Fix supporter-product-data schedule cron

undergroundquizscene:
Based on [the docs](https://docs.aws.amazon.com/lambda/latest/dg/services-cloudwatchevents-expressions.html), I believe either day-of-week or day-of-month should be a question mark. I think builds are currently failing in UAT.

### DIFF
--- a/supporter-product-data/cloudformation/cfn.yaml
+++ b/supporter-product-data/cloudformation/cfn.yaml
@@ -19,11 +19,11 @@ Mappings:
     DEV:
       S3Bucket: supporter-product-data-export-dev
       # Run once at 6am Mon-Fri
-      scheduleRate: "cron(0 6 * * 1-5)"
+      scheduleRate: "cron(0 6 ? * 1-5 *)"
       lambdaConcurrency: 30
     UAT:
       S3Bucket: supporter-product-data-export-uat
-      scheduleRate: "cron(0 6 * * 1-5)"
+      scheduleRate: "cron(0 6 ? * 1-5 *)"
       lambdaConcurrency: 30
     PROD:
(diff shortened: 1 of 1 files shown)


supporter-product-data/cloudformation/cfn.yaml
```diff
@@ -19,11 +19,11 @@ Mappings:
     DEV:
       S3Bucket: supporter-product-data-export-dev
       # Run once at 6am Mon-Fri
-      scheduleRate: "cron(0 6 * * 1-5)"
+      scheduleRate: "cron(0 6 ? * 1-5)"
```
rupertbates, line 22: This needs to be `cron(0 6 ? * 1-5 *)` I just tested it

undergroundquizscene: Ah yes, of course! I’ll update that now.

undergroundquizscene: How did you test it? I’d like to remember to do that in future

rupertbates: Go to the stack in the cloudformation console: https://eu-west-1.console.aws.amazon.com/cloudformation/home?region=eu-west-1#/stacks/stackinfo?stackId=arn%3Aaws%3Acloudformation%3Aeu-west-1%3A865473395570%3Astack%2Fsupport-DEV-supporter-product-data%2F0e889890-917b-11eb-a530-0acd9480bd5b

Click 'Update' and select your cloudformation file, then hit next a few times and it will run the cloudformation update which would normally happen through RiffRaff

rupertbates (approved):
Good work

PR #4634, opened 2023-01-10
Create prerendered page for the new checkout

i-hardy:

This adds an 'empty' version of the supporter plus checkout page to be prerendered, to improve time to first paint on the checkout pages.

It requires a slightly awkward approach to including the Emotion styles due to the way our pre-rendering setup uses Preact's `renderToString` in a pseudo-browser environment with JSDOM. Emotion [does not currently support using its SSR utilities in the browser](https://github.com/emotion-js/emotion/issues/2691). This means that although we are rendering our JSX as a string rather than as HTML elements, Emotion will insert `<style>` elements into the JSDOM document head, and the `Global` utility component that we use to add fonts and our CSS reset does not work correctly.

In order to fix this I've updated `ssrScript.js` to extract the contents of any Emotion-created style nodes inserted into the document head and include this in the output file, and added a dedicated component that creates a `<style>` element with our required global styles. 

Ideally we should try to use Emotion's intended approach to server-side rendering and not our hybrid approach, but currently even the minimal 'empty' checkout code is too dependent on browser APIs to ditch JSDOM entirely, and the work required to fix this is probably not worth the time unless it's in service of a much broader server-side rendering project.

[**Trello Card**](https://trello.com/c/8T3Nkpnv)

## Screenshots

With JS disabled to show only the pre-rendered page

Screenshot 2023-01-10 at 16-45-36 Support the Guardian


i-hardy:
[SupporterPlus to 100%: Create a new pre-rendered page scaffold for the Supporter Plus checkout](https://trello.com/c/8T3Nkpnv/946-supporterplus-to-100-create-a-new-pre-rendered-page-scaffold-for-the-supporter-plus-checkout)

GHaberis (approved):
Nice work 👍

PR #4644, opened 2023-01-12
Supporter Product Data Step function Fixes

rupertbates:
We recently noticed some issues with the [supporter-product-data-UAT](https://eu-west-1.console.aws.amazon.com/states/home?region=eu-west-1#/statemachines/view/arn:aws:states:eu-west-1:865473395570:stateMachine:supporter-product-data-UAT) step function which meant it had been failing repeatedly for many months. This PR makes some changes to try to prevent this happening again.

## What happened
The supporter product data step function extracts data from Zuora on a scheduled basis using a stateful Aqua query: https://knowledgecenter.zuora.com/Zuora_Central_Platform/API/AB_Aggregate_Query_API/BA_Stateless_and_Stateful_Modes

This means that it has two modes of running - Full or Incremental specified by the value that is passed into it eg
{
  "queryType": "incremental"
}

when a full extract is required the SupporterProductDataQueryZuora lambda passes in the query statement to Zuora eg.
```
SELECT RatePlan.AmendmentType, Subscription.Id, Account.IdentityId__c, Subscription.GifteeIdentityId__c, RatePlan.Id, ProductRatePlan.Id, ProductRatePlan.Name, Subscription.TermEndDate 
FROM rateplan 
WHERE Subscription.TermEndDate >= '2021-03-16' 
AND (Subscription.Status = 'Active' OR (Subscription.Status = 'Cancelled' 
AND ProductRatePlan.Id != '2c92c0f85ab269be015acd9d014549b7' 
AND ProductRatePlan.Id != '2c92c0f95e1d5c9c015e38f8c87d19a1')) 
AND (RatePlan.AmendmentType is null OR RatePlan.AmendmentType = 'NewProduct' OR RatePlan.AmendmentType = 'UpdateProduct') 
AND ProductRatePlan.Id != '2c92c0f953078a5601531299dae54a4d' 
AND Account.IdentityId__c like '_%' 
AND ((Subscription.RedemptionCode__c = '' OR Subscription.RedemptionCode__c is null) OR (Subscription.RedemptionCode__c like '_%' AND Subscription.GifteeIdentityId__c like '_%'))
```
however when an incremental update is required it only passes the last query time and the query from the last full extract is used.

What  this means is that if the fields returned by the query change we MUST run a full sync in all environments or we will be continuing to use the old query in our executions. In this case a change in the select fields led a CSV parsing error when we tried to read the results.

## The fixes
- I have added  code to make the `AddSupporterRatePlanItemToQueueLambda` fail if it comes across an empty CSV file or if it is unable to read any items from the CSV. Previously we would keep processing and alarm if there were read errors but in all likelihood if one row can't be read, none will be and it is better to exit the lambda and terminate the step function execution.
- I have added a comment where the zuora query is defined to make it clear that any change to the query will require a full data refresh in all environments.

I have also reverted the scheduling frequency of the PROD version of the lambda to 5 minutes as this was reduced to try to reduce the processing cost but it turned out that the cost was actually down to the errors in the DEV and UAT configurations, not the amount of processing in PROD

### DIFF
--- a/supporter-product-data/cloudformation/cfn.yaml
+++ b/supporter-product-data/cloudformation/cfn.yaml
@@ -27,8 +27,8 @@ Mappings:
       lambdaConcurrency: 30
     PROD:
       S3Bucket: supporter-product-data-export-prod
-      # Every 15 mins
-      scheduleRate: "cron(*/15 * ? * * *)"
+      # Every 5 mins
+      scheduleRate: "cron(*/5 * ? * * *)"
       lambdaConcurrency: 50 # May need to reduce this number if we see downstream systems getting overwhelmed during a full refresh, I've tested with 30 and that was fine
 
 Resources:
--- a/supporter-product-data/cloudformation/cfn.yaml
+++ b/supporter-product-data/cloudformation/cfn.yaml
@@ -447,7 +447,7 @@ Resources:
         - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:reader-revenue-dev
       AlarmName: There was a csv read failure when loading supporter product data into DynamoDB
       AlarmDescription: >
-        Search for 'CSV read failures' in the AddSupporterRatePlanItemToQueueLambda Cloudwatch logs:
+        Search for 'CSV read failure' in the AddSupporterRatePlanItemToQueueLambda Cloudwatch logs:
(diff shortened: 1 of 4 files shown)


undergroundquizscene (approved):
Looks good! I don’t know much about the integration tests though, so my review of that has not been very thorough.

PR #4652, opened 2023-01-13
Make supporter-product-data-dynamo library available via Maven

rupertbates:
This PR makes the supporter-product-data-dynamo library publishable via Maven to allow us to use it from support-service-lambdas. This is necessary to support the product switching work currently taking place.

One  day it would be nice to put all of this code into a shared support mono-repo but today is not that day!

[**Trello Card**](https://trello.com/c/8aWm1Yop/3775-make-supporter-product-data-dynamo-library-available-via-maven)


support-internationalisation/README.md
```diff
@@ -10,7 +10,10 @@ Run `sbt +publishLocal`.
 Releasing to maven
```
rupertbates, line 10: I've updated the readme on support-internationalisation as well because it was pretty terse before!

support-internationalisation/version.sbt
```diff
@@ -1 +1 @@
-version := "0.16-SNAPSHOT"
```
rupertbates, line 1: I ran a test release of support-internationalisation to make sure my local setup was working, this is the result of that

jonflynng:
Make sure to amend the ReadMe and add the line needed to add the dependency to other projects depending on their Scala version. e.g.:

```
("com.gu" %% "support-product-data-dynamo" % "0.2").cross(CrossVersion.for3Use2_13) exclude("com.typesafe.scala-logging", "scala-logging_2.13")
```

jonflynng (approved):

PR #4653, opened 2023-01-13
update monthly MIN value for other amount field

GHaberis:

This PR sets the MIN value users are allowed to submit in the "Other amount" field on the monthly tab at https://support.theguardian.com/contribute to `2` in all regions. 

## Why are you doing this?

We've seen when AB testing choice card amounts that the MIN value in the code is sometimes greater than the lowest value choice card, so isn't really the MIN, this leads to confusing UX, as shown below:

Screenshot 2023-01-13 at 15 44 30

The current MIN values also contradict some support messaging on www.theguardian.com that states "you can give from as little as £/$2 a month".

It's been agreed that the smallest monthly choice card in an amounts AB test will not be less than 2.


tomrf1 (approved):

paul-daniel-dempsey (commented):
Looks good

PR #4657, opened 2023-01-16
Show discount vs retail for newspaper

tomrf1:
Currently in the newspaper product page the price cards only show the normal online vs retail discount, even if a promo is applied.
We only have the `savingVsRetail`, so the logic is a bit fiddly.

### collection before (the price is with the promo discount, but the % is for normal online price vs retail)
Screenshot 2023-01-16 at 15 47 25

### collection after
Screenshot 2023-01-16 at 15 46 49

### delivery before
Screenshot 2023-01-16 at 15 48 14

### delivery after
Screenshot 2023-01-16 at 15 48 08

### DIFF
--- a/support-frontend/assets/helpers/productPrice/productPrices.ts
+++ b/support-frontend/assets/helpers/productPrice/productPrices.ts
@@ -126,6 +126,22 @@ function getCurrency(country: IsoCountry): IsoCurrency {
 	return currency;
 }
 
+/**
+ * @param discountedPrice - price after promo discount applied to online price
+ * @param onlineVsRetailPerc - % discount of normal online price vs retail price
+ * @param discountedVsOnlinePerc - % discount of discountedPrice against normal online price
+ */
+const getDiscountVsRetail = (
+	discountedPrice: number,
+	onlineVsRetailPerc: number,
+	discountedVsOnlinePerc: number,
+): number => {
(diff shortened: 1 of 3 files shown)


tomrf1:
For home delivery (for WINTER20), the sat/sun are blank because the discount is negative:
Screenshot 2023-01-16 at 15 42 39

support-frontend/assets/helpers/productPrice/productPrices.ts
```diff
@@ -126,6 +126,22 @@ function getCurrency(country: IsoCountry): IsoCurrency {
...
+ */
+const getDiscountVsRetail = (
+	discountedPrice: number,
+	onlineVsRetailPerc: number,
+	discountedVsOnlinePerc: number,
+): number => {
+	const onlinePrice = discountedPrice / (1 - discountedVsOnlinePerc / 100);
+	const retailPrice = onlinePrice / (1 - onlineVsRetailPerc / 100);
+	const totalSavingVsRetail = (1 - discountedPrice / retailPrice) * 100;
+	return Math.round(totalSavingVsRetail);
```
MaelGNM, line 142: I agree that it's a bit fiddly but that works and actually having thought about this calculation before I like how you've written it!

MaelGNM (approved):
Looks good to me!

PR #4662, opened 2023-01-17
Fix UpdateSupporterProductData lambda

rupertbates:
I noticed an issue with the UpdateSupporterProductData lambda which updates the Dynamo cache when a user purchases a product. When we were working out what type of record to insert, we returned an Option to represent the fact that for Digital Subscription gift purchases we don't insert a record (because gift purchases don't receive the benefits, the giftee does). However the code was actually returning `None` for any unknown product not just DS gifts and because the function wasn't updated when Supporter Plus was released we were returning `None` for all S+ purchases and not writing a record to the data store.

The impact of this would be that purchasers of Supporter Plus would not have been recognised by members-data-api until up to 5 minutes after their purchase. In some cases if a user received a response from mdapi during those 5 minutes, that response would have been cached for a day so in the worst case a user may not have been recognised as a S+ subscriber until a day after their purchase.

I have also removed some unnecessary AWS permissions from the support-workers lambda role

### DIFF
--- a/support-workers/src/main/scala/com/gu/support/workers/lambdas/UpdateSupporterProductData.scala
+++ b/support-workers/src/main/scala/com/gu/support/workers/lambdas/UpdateSupporterProductData.scala
@@ -1,7 +1,7 @@
 package com.gu.support.workers.lambdas
 
 import com.amazonaws.services.lambda.runtime.Context
-import com.gu.supporterdata
+import com.gu.monitoring.SafeLogger
 import com.gu.services.{ServiceProvider, Services}
 import com.gu.support.catalog._
 import com.gu.support.workers.RequestInfo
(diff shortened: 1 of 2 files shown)


undergroundquizscene (approved):

support-workers/src/main/scala/com/gu/support/workers/lambdas/UpdateSupporterProductData.scala
```diff
@@ -65,61 +76,88 @@ object UpdateSupporterProductData {
...
+                identityId = user.id,
+                productRatePlanId = productRatePlan.id,
+                productRatePlanName = s"support-workers added ${product.describe}",
+                Some(ContributionAmount(product.amount, product.currency.iso)),
+              ),
             ),
           )
-          .toRight(())
-          .map(Some(_))
+          .toRight(s"Unable to create SupporterRatePlanItem from state $state")
```
tomrf1, line 95: I found this quite hard to follow in github because it creates nested Options then turns the outer one into an Either! Perhaps swapping them round might help
```
.getProductRatePlan(...
.toRight(...
.map(...
```

tomrf1 (approved):

support-workers/src/main/scala/com/gu/support/workers/lambdas/UpdateSupporterProductData.scala
rupertbates: In the way it was previously? I changed it to this because I thought it was clearer!

tomrf1: yeah, we must reason about this in different ways!

PR #4663, opened 2023-01-17
Supporter-plus nudge

paul-daniel-dempsey:
Nudging component, encouraging single to recurring contributions. 
Untitled 2

Features->
- 2 coloured header lines, 
- close button, 
- body copy containing currencyGlyph and monthly minimum contribution amount
- monthly recurring link, 
- background-image (using GUIM)

3-cohort A/B test for implementation ->
https://support.thegulocal.com/uk/contribute (Nudge NOT displayed)
https://support.thegulocal.com/uk/contribute#ab-singleToRecurring=control (Nudge NOT displayed)
https://support.thegulocal.com/uk/contribute#ab-singleToRecurring=variantA
https://support.thegulocal.com/uk/contribute#ab-singleToRecurring=variantB

unnamed

[**Trello Card**](https://trello.com/c/gnwhW92Q/982-nudging-single-to-recurring-a-b-test-figma-designs-for-implementation)

## Is this an AB test?
- [x] Yes
- [ ] No


support-frontend/assets/components/checkoutNudge/checkoutNudge.tsx
```diff
@@ -0,0 +1,150 @@
...
+					<h2 css={styles.heading(brand[500])}>{nudgeTitleTopCopy}</h2>
+				</div>
+				<CheckoutNudgeCloseButton onClose={onNudgeClose} />
+			</div>
+			<h2 css={styles.heading(neutral[7])}>{nudgeTitleCopy}</h2>
+			<p css={styles.para}>
+				Regular, reliable support powers Guardian journalism in perpetuity. If
+				you can, please consider setting up a monthly payment today from just{' '}
+				{currencyGlyph}
+				{minAmount} – it takes less than a minute.
```
undergroundquizscene, line 141: What’s the reason this is broken over two lines? I assume the linter has passed it (or even done it), but I’m surprised – I would’ve expected it all on one line.

undergroundquizscene (approved):
Looks good to me, though I’m not a frontend expert! In particular, I don’t know how to check Optimize, which is mentioned in the description because it’s an AB test – when I follow that link I just end up at the Optimize homepage, I think.

Might be worth waiting for a review from someone with more frontend knowledge, but I’ll leave that up to you.

GHaberis:
>  I don’t know how to check Optimize, which is mentioned in the description because it’s an AB test – when I follow that link I just end up at the Optimize homepage, I think.

Thank you for highlighting this @undergroundquizscene, we're no longer using Optimize to monitor our AB tests, so we should look at removing these instructions from the repo, this can be done in a follow-up PR, but for now these instructions should be ignored.

support-frontend/assets/components/checkoutBenefits/checkoutBenefitsList.tsx
```diff
@@ -72,34 +76,58 @@ const para = css`
 export type CheckoutBenefitsListProps = {
 	title: string;
 	checkListData: CheckListData[];
+	contributionType: ContributionType;
+	countryGroupId: CountryGroupId;
 	buttonCopy: string | null;
+	nudgeTitleTopCopy: string;
+	nudgeTitleCopy: string;
```
marialani, line 83: How about something like?

```suggestion
	nudgeTitleCopySection1: string;
	nudgeTitleCopySection2: string;
```
Something that indicates that these are 2 parts of a whole would be great here

support-frontend/assets/components/checkoutBenefits/checkoutBenefitsList.tsx
```diff
@@ -72,34 +76,58 @@ const para = css`
...
+						))}
+					</table>
+					<hr css={hr(`${space[5]}px 0 ${space[4]}px`)} />
+					<p css={para}>Cancel anytime</p>
+				</>
+			)}
 			{/* <UpsellButton
 				buttonCopy={buttonCopy}
 				handleButtonClick={handleButtonClick}
 			/> */}
```
marialani, line 130: We could delete this as the supporter plus nudge supersedes this old component

support-frontend/assets/components/checkoutBenefits/checkoutBenefitsList.tsx
```diff
@@ -72,34 +76,58 @@ const para = css`
...
-				{checkListData.map((item) => (
-					<tr>
-						<td css={[checkListIcon, item.maybeGreyedOut]}>
-							<div css={iconContainer}>{item.icon}</div>
-						</td>
-						<td css={[checkListText, item.maybeGreyedOut]}>{item.text}</td>
-					</tr>
-				))}
-			</table>
+			{contributionType === 'ONE_OFF' && displayNudge && (
```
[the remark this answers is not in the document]
marialani: Have seen you'd updated this 👍

support-frontend/assets/components/checkoutBenefits/checkoutBenefitsListContainer.tsx
```diff
@@ -41,21 +46,25 @@ export function CheckoutBenefitsListContainer({
...
 	const minimumContributionAmount = useContributionsSelector(
 		getMinimumContributionAmount,
 	);
 
 	const currency = currencies[currencyId];
 
-	const thresholdPrice = getThresholdPrice(countryGroupId, contributionType);
+	const thresholdPrice = isRecurring(contributionType)
+		? getThresholdPrice(countryGroupId, contributionType)
+		: 0;
```
marialani, line 67: When do we want the thresholdPrice at 0?

support-frontend/assets/helpers/abTests/abtestDefinitions.ts
```diff
@@ -115,4 +115,23 @@ export const tests: Tests = {
 		seed: 10,
 		optimizeId: 'dQCXBc3QQIW7M1Di_qSCHw',
 	},
+	singleToRecurring: {
+		variants: [
```
GHaberis, line 119: I believe based on the spec for this AB test there should be 3 cohorts: `control`, `variantA` and `variantB`? The copy inside the container will differ `variantA` and `variantB`.

paul-daniel-dempsey:
Wondered that, control is no-nudge displayed at all? Then I need to define a VariantA and VariantB? 33% chance of each?

GHaberis:
> control is no-nudge displayed at all? Then I need to define a VariantA and VariantB? 33% chance of each?

Exactly, `control` is current behaviour, so no-nudge displayed at all.

support-frontend/assets/helpers/abTests/abtestDefinitions.ts
```diff
@@ -115,4 +115,23 @@ export const tests: Tests = {
...
+		],
+		audiences: {
+			ALL: {
+				offset: 0,
+				size: 1,
+			},
+		},
+		isActive: false,
+		referrerControlled: false,
+		seed: 15,
```
GHaberis, line 135: I think we should add a `targetPage: pageUrlRegexes.contributions.allLandingPagesAndThankyouPages` to this AB test, this will mean we only assign users on the "contributions" checkout to this AB test (which is what we want)

support-frontend/assets/components/checkoutNudge/checkoutNudge.tsx
paul-daniel-dempsey: Yes its just too long for the linter am afraid

support-frontend/assets/components/checkoutBenefits/checkoutBenefitsList.tsx
paul-daniel-dempsey: Yes better group naming

support-frontend/assets/components/checkoutBenefits/checkoutBenefitsListContainer.tsx
paul-daniel-dempsey: Its used for the handleButtonClick (event handler line86), it need's a valid threshold price even when the contributionType is ONE_OFF, otherwise I get  'possibly undefined' in handleButtonClick. Can look into a better way however tomorrow.

support-frontend/assets/helpers/abTests/abtestDefinitions.ts
paul-daniel-dempsey: Yes, implemented and redescribed correctly on PR

support-frontend/assets/helpers/abTests/abtestDefinitions.ts
paul-daniel-dempsey: Ah yes, have done and will use this going forwards to reduce the pages affected by ab-tests.

support-frontend/assets/components/checkoutBenefits/checkoutBenefitsListContainer.tsx
GHaberis: Did you get a typescript error prompting you to make the change? This code would have never actually run when the `contributionType` was ONE_OFF because of the return statement `if (isOneOff(contributionType)) { return null }` that's been removed from this module. I'd be very wary of this change. I don't think we should be repurposing the `checkoutBenefitsListContainer.tsx` component to also double up as the nudge component. I'd create a brand new component for the nudge component that only renders when `isOneOff(contributionType)` and the user is in the appropriate AB test variant.

paul-daniel-dempsey: Yes, agree this would be much neater and may help or/simplfy the display side effects have seen, will bring up the nudge component out of the benefits list.

support-frontend/assets/components/checkoutBenefits/checkoutBenefitsList.tsx
marialani: Should be good to delete the component as the nudge component replaces the upsell button

support-frontend/assets/components/checkoutBenefits/checkoutBenefitsListContainer.tsx
paul-daniel-dempsey: Ok, finished completing this, would you mind having a look it's acceptable? Have asked Maria to look too. Doing this completely got rid of the display side-effects so will remember components go as high as they can in future.

support-frontend/assets/components/checkoutNudge/checkoutNudge.tsx
```diff
@@ -0,0 +1,150 @@
...
+	nudgeDisplay,
+	nudgeTitleCopySection1,
+	nudgeTitleCopySection2,
+	nudgeParagraphCopy,
+	onNudgeClose,
+	onNudgeClick,
+}: CheckoutNudgeProps): JSX.Element {
+	const displayValid = contributionType === 'ONE_OFF' && nudgeDisplay;
+	return (
+		<div css={styles.container}>
```
GHaberis, line 128: Is it common practise to define CSS rules as props within a `styles` object rather than defining them as individual variables? I can see it works, but my concern is ESLint would not warn us if rules within this `styles` object are no longer used, which it would do if they were individual variables? eg. 

```
const styles = { 
  container: css``,
  top: css``
 }
```

VS 

```
const container = css: ``;
const top = css: ``;
```

This isn't a blocker, but it's worth considering/thinking about.

support-frontend/assets/components/checkoutNudge/checkoutNudge.tsx
```diff
@@ -0,0 +1,150 @@
...
+			{displayValid && (
```
GHaberis, line 129: Could we not put this conditional on the before the parent `<div css={styles.container}>` and if `displayValid` is `false` we just return `null` rather than an empty div? We'd need to update the return type of `CheckoutNudge` to `JSX.Element | null`. We do this quite often in the code for conditionally rendered components.

support-frontend/assets/components/checkoutNudge/checkoutNudgeContainer.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+export function CheckoutNudgeContainer({
+	renderNudge: renderNudge,
+}: CheckoutNudgeContainerProps): JSX.Element | null {
+	const contributionType = useContributionsSelector(getContributionType);
+	const dispatch = useContributionsDispatch();
+	const { countryGroupId, countryId } = useContributionsSelector(
+		(state) => state.common.internationalisation,
+	);
+	const abParticipations = initAbTests(
```
GHaberis, line 26: I don't think we should be calling `initAbTests` to get `abParticipations` they should be available in the Redux state in `state.common. abParticipations`.

support-frontend/assets/helpers/abTests/abtestDefinitions.ts
```diff
@@ -115,4 +115,27 @@ export const tests: Tests = {
...
+				id: 'variantB',
+			},
+		],
+		audiences: {
+			ALL: {
+				offset: 0,
+				size: 1,
+			},
+		},
+		isActive: false,
```
GHaberis, line 136: We'll need to set `isActive` to `true` to turn it on.

paul-daniel-dempsey: Ok, turned on now for use on prod

support-frontend/assets/components/checkoutNudge/checkoutNudgeContainer.tsx
paul-daniel-dempsey: Ah I see it in Redux now, will do retrieval from Redux now

GHaberis: For reference this is where we assign `abParticipations` it's value when setting up the initial Redux state: https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/helpers/redux/utils/setup.ts#L88.

support-frontend/assets/components/checkoutNudge/checkoutNudge.tsx
paul-daniel-dempsey: Yes makes sense, split them out

support-frontend/assets/components/checkoutNudge/checkoutNudge.tsx
paul-daniel-dempsey: Completed - Yes returning null much better than an empty div

support-frontend/assets/components/checkoutNudge/checkoutNudge.tsx
paul-daniel-dempsey: Completed- yes makes sense for linter to spot, have split them out now

support-frontend/assets/helpers/abTests/abtestDefinitions.ts
paul-daniel-dempsey: Completed - turned on now for use on prod

support-frontend/assets/components/checkoutNudge/checkoutNudgeContainer.tsx
paul-daniel-dempsey: Completed - makes sense to use redux wherever possible, saves repeat getting of the same participations

support-frontend/assets/components/checkoutBenefits/checkoutBenefitsListContainer.tsx
paul-daniel-dempsey: Completed

GHaberis (approved):
Looks good 👍

PR #4664, opened 2023-01-17
Validate postCode length on server side

LAKSHMIRPILLAI:

Validate PostCode  length on server side for Guardian Weekly and Paper 

[**Trello Card**](https://trello.com/c/hgXWkPCo/663-validate-length-on-postcode-field)

## Why are you doing this?
Salesforce cannot accept postcodes greater than 20 characters in length, but we currently do not validate the length of this field on the server side. The client side validation for postCode field is in place.
We use Salesforce for all Subscriptions (Newspaper, Guardian Weekly) and Recurring Contributions, so any checkout for these products which has a postcode field should adopt this rule. As we're removing Digital Subs making the change this checkout is not a priority.
## Is this an AB test?
- [ ] Yes
- [x ] No

### DIFF
--- a/support-frontend/app/utils/CheckoutValidationRules.scala
+++ b/support-frontend/app/utils/CheckoutValidationRules.scala
@@ -32,9 +32,9 @@ object CheckoutValidationRules {
   case class Invalid(message: String) extends Result
   case object Valid extends Result
   def checkSubscriptionPaymentMethodEnabled(
-      switches: SubscriptionsPaymentMethodSwitches,
-      paymentFields: Either[PaymentFields, RedemptionData],
-  ) = paymentFields match {
+                                             switches: SubscriptionsPaymentMethodSwitches,
+                                             paymentFields: Either[PaymentFields, RedemptionData],
+                                           ) = paymentFields match {
     case Left(_: PayPalPaymentFields) =>
       if (switches.paypal.isOn) Valid else Invalid("Invalid Payment Method")
     case Left(_: DirectDebitPaymentFields) =>
@@ -46,9 +46,9 @@ object CheckoutValidationRules {
   }
 
   def checkContributionPaymentMethodEnabled(
-      switches: RecurringPaymentMethodSwitches,
-      paymentFields: Either[PaymentFields, RedemptionData],
-  ) = paymentFields match {
+                                             switches: RecurringPaymentMethodSwitches,
+                                             paymentFields: Either[PaymentFields, RedemptionData],
+                                           ) = paymentFields match {
     case Left(_: PayPalPaymentFields) =>
       if (switches.payPal.isOn) Valid else Invalid("Invalid Payment Method")
     case Left(_: DirectDebitPaymentFields) =>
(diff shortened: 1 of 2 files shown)


support-frontend/app/utils/CheckoutValidationRules.scala
```diff
@@ -180,6 +180,17 @@ object AddressAndCurrencyValidationRules {
     ) {
       stateFromRequest.isDefined.otherwise(s"state is required for $countryFromRequest")
     } else Valid
+  def hasValidPostcodeLength(postcodeFromRequest: Option[String], addressType: String): Result = {
```
undergroundquizscene, line 183: What would you think about naming this more specifically? Something like `postcodeIsShortEnoughForSalesforce`. I know it’s a bit long, but I think it’d be useful for it to be clear from the code (both definition and uses) that this is why we’re limiting it to 20 characters.

support-frontend/app/utils/CheckoutValidationRules.scala
```diff
@@ -180,6 +180,17 @@ object AddressAndCurrencyValidationRules {
...
+    val validPostCode = postcodeFromRequest match {
+      case Some(postCode) if (postCode.length <= 20) => true
+      case None => true
+      case _ => false
+    }
+    if (validPostCode)
+      Valid
+    else
+      Invalid(s"$addressType PostCode length must be less than 20 characters")
```
undergroundquizscene, line 192: I think this message should actually say “… must be 20 characters or fewer”, or “… must be less than 21 characters”, since 20 characters is an acceptable length.

support-frontend/app/utils/CheckoutValidationRules.scala
```diff
@@ -180,6 +180,17 @@ object AddressAndCurrencyValidationRules {
     ) {
       stateFromRequest.isDefined.otherwise(s"state is required for $countryFromRequest")
     } else Valid
+  def hasValidPostcodeLength(postcodeFromRequest: Option[String], addressType: String): Result = {
+    val validPostCode = postcodeFromRequest match {
+      case Some(postCode) if (postCode.length <= 20) => true
+      case None => true
+      case _ => false
```
undergroundquizscene, line 187: I wonder about simplifying this into two lines?

```scala
case Some(postCode) if (postCode.length > 20) => false
case _ => true
```

I don’t think it’s important to draw attention to the extra cases, and then you can simplify the whole thing into:

```scala
postcodeFromRequest match {
  case Some(postCode) if (postCode.length > 20) =>
    Invalid(s"$addressType PostCode length must be less than 20 characters")
  case _ => Valid
}
```

(If you think it is important to have all three cases, do say! I may have missed something)

undergroundquizscene (changes requested):

support-frontend/app/utils/CheckoutValidationRules.scala
LAKSHMIRPILLAI: Have refactored the function as per the comment

support-frontend/app/utils/CheckoutValidationRules.scala
LAKSHMIRPILLAI: Changed the Invalid message to PostCode length must not be greater than 20 characters

support-frontend/app/utils/CheckoutValidationRules.scala
LAKSHMIRPILLAI: I have kept hasValidPostcodeLength in order to match with the other functions  in the same file.I have  added a comment to make it clear

undergroundquizscene: Ok, that sounds good :)

undergroundquizscene (approved):

PR #4667, opened 2023-01-17
Try to fix existing payment methods

undergroundquizscene:

We enabled existing card payment methods, which worked in testing but then ran into errors in production. This PR is to fix the errors that have cropped up.

~~Specifically, the errors we’ve had have the following form: `Failed to create new Annual-SupporterPlus-USD-120, due to RequestValidationError(validation of the request body failed with existing billing account id missing...`. From looking at the [failed step function](https://eu-west-1.console.aws.amazon.com/states/home?region=eu-west-1#/v2/executions/details/arn:aws:states:eu-west-1:865473395570:execution:support-workers-PROD:a6266d8b-83c4-4fa2-919e-ee30ef8ae5dc) that produced that error, and examining [the code the error came from](https://github.com/guardian/support-frontend/blob/6dd1b1268557cec9cb94919fd15e80779e3a1c53/support-frontend/app/utils/CheckoutValidationRules.scala#L156), I have concluded that the client side isn’t sending the billing account ID to the server.~~

~~Looking at the redux state in the browser for the contributions form, I can see that selecting an existing payment method sets the payment methods details at `state.page.checkoutForm.payment.existingPaymentMethods.selectedPaymentMethod`, which is not where [the mapping from state to props](https://github.com/guardian/support-frontend/blob/6dd1b1268557cec9cb94919fd15e80779e3a1c53/support-frontend/assets/pages/contributions-landing/components/ContributionForm.tsx#L82) is looking for it.~~

~~So I have updated the mapping, but I’m not super confident this is the correct change to make: it works, but I don’t understand how the data flow all fits together, so there may still be some redundancy in there.~~

[**Trello Card**](https://trello.com/c/4DYyBs1a/993-server-side-validation-failing-for-users-trying-to-checkout-with-an-existing-card-for-recurring-contributions-supporterplus)

### DIFF
--- a/support-frontend/assets/components/existingMethodPaymentButton/existingCardPaymentButton.tsx
+++ b/support-frontend/assets/components/existingMethodPaymentButton/existingCardPaymentButton.tsx
@@ -21,5 +21,10 @@ export function ExistingCardPaymentButton(): JSX.Element {
 		);
 	});
 
-	return <DefaultPaymentButtonContainer onClick={payWithExistingCard} />;
+	return (
+		<DefaultPaymentButtonContainer
+			onClick={payWithExistingCard}
+			disabled={!!selectedPaymentMethod?.billingAccountId}
+		/>
+	);
 }
--- a/support-frontend/assets/components/paymentButton/defaultPaymentButton.tsx
+++ b/support-frontend/assets/components/paymentButton/defaultPaymentButton.tsx
@@ -15,16 +15,23 @@ export type DefaultPaymentButtonProps = {
 	id?: string;
 	buttonText: string;
 	onClick: (event: React.MouseEvent<HTMLButtonElement>) => void;
+	disabled?: boolean;
 };
 
 export function DefaultPaymentButton({
 	id,
 	buttonText,
 	onClick,
+	disabled = false,
 }: DefaultPaymentButtonProps): JSX.Element {
 	return (
 		<ThemeProvider theme={buttonThemeReaderRevenueBrand}>
-			<Button id={id} cssOverrides={buttonOverrides} onClick={onClick}>
+			<Button
+				id={id}
+				cssOverrides={buttonOverrides}
+				onClick={onClick}
+				disabled={disabled}
+			>
 				{buttonText}
 			</Button>
 		</ThemeProvider>
--- a/support-frontend/assets/components/paymentButton/defaultPaymentButtonContainer.tsx
+++ b/support-frontend/assets/components/paymentButton/defaultPaymentButtonContainer.tsx
@@ -23,6 +23,7 @@ type ButtonTextCreator = (
 export type DefaultPaymentContainerProps = {
 	onClick: (event: React.MouseEvent<HTMLButtonElement>) => void;
 	createButtonText?: ButtonTextCreator;
+	disabled?: boolean;
 };
 
 function getButtonText(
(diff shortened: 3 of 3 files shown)


i-hardy:
This is unlikely to be the problem as this is dead code (now deleted) that isn't in use on the supporter plus checkout. I think the easiest place for a quick fix would be in the payment button component for paying with an existing card, preventing submission if there's no `selectedPaymentMethod` or if `billingAccountId` is an empty string: https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/components/existingMethodPaymentButton/existingCardPaymentButton.tsx

However it would be worth investigating how exactly we're ending up in the situation where an existing payment method is selectable in the payment method list, but doesn't have a valid `billingAccountId`. This whole system is a bit fragile and unpleasantly difficult to test, so it probably needs a separate spike.

undergroundquizscene:
[Server Side validation failing for users trying to checkout with an Existing Card for Recurring Contributions / SupporterPlus](https://trello.com/c/4DYyBs1a/993-server-side-validation-failing-for-users-trying-to-checkout-with-an-existing-card-for-recurring-contributions-supporterplus)

undergroundquizscene:
> This is unlikely to be the problem as this is dead code (now deleted) that isn't in use on the supporter plus checkout. I think the easiest place for a quick fix would be in the payment button component for paying with an existing card, preventing submission if there's no `selectedPaymentMethod` or if `billingAccountId` is an empty string: https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/components/existingMethodPaymentButton/existingCardPaymentButton.tsx
> 
> However it would be worth investigating how exactly we're ending up in the situation where an existing payment method is selectable in the payment method list, but doesn't have a valid `billingAccountId`. This whole system is a bit fragile and unpleasantly difficult to test, so it probably needs a separate spike.

Ah, thank you! I realise now that in my testing I never reproduced the error: it worked after I made my change because it was working before I made my change. I’ll take a look where you suggest and see what I can work out – it is a bit alarming that you might have selectable existing payment methods that don’t have a valid billingAccountId.

undergroundquizscene:
Hmmm, I wasn’t trying to close this – I guess it’s because I rebased on main and pushed a version without the changes I’d made (since those files have now been deleted).

undergroundquizscene:
I have now gone with the quick fix of disabling the button if the billing account id is falsy, and can reopen this PR

undergroundquizscene:
I haven’t been able to test this – I had hoped I could use the redux devtools to issue a custom event, but I can’t get around CSP on code and I can’t seem to get a logged in user locally. Any help/advice would be appreciated :)

undergroundquizscene:
If/Once merged, I will need to re-enable the existing card payment switch in the RRCP

undergroundquizscene:
I have also created a trello card (https://trello.com/c/jNf1CZB9/1020-find-out-why-billing-account-ids-are-invalid-for-some-existing-cards) to look into the issue further

i-hardy (approved):
Looks good!

GHaberis:
> If/Once merged, I will need to re-enable the existing card payment switch in the RRCP

@undergroundquizscene yes you will, message me if you haven't done this before and would like me to show you how to access the switchboard.

undergroundquizscene:
> > If/Once merged, I will need to re-enable the existing card payment switch in the RRCP
> 
> @undergroundquizscene yes you will, message me if you haven't done this before and would like me to show you how to access the switchboard.

Thanks, I should be fine! Someone (Maria, I think?) showed me how to do this before :)

PR #4671, opened 2023-01-18
Use CSS media queries to show/hide mobile menu on Support site Landing pages

GHaberis:

This PR fixes some inconsistencies in the way the nav was appearing across regions on the GW landing page, we had some logic in the JS that would decide whether to show the veggie-burger or non-veggie-burger menu. In all regions other than the UK this logic was concluding to show the non-veggie-burger menu on mobile devices (a side effect of removing some links when deprecating Digi Subs). 

The issue is the non-veggie-burger doesn't look great on Mobile and shifts the contents of the page down, as shown below:

Screenshot 2023-01-18 at 13 20 24

**Ideally we'd be consistent and show the veggie-burger menu in ALL regions at the mobile breakpoint.**

I've therefore replaced this JavaScript logic that calculates whether to add the CSS class name modifiers to the `.component-header` element, with something more simple. 

Before this PR we were using JS to calculate the widths of navigation links vs the overall width of the menu to decide whether to display the "one-row" or "veggie-burger" variations of the menu, or the default version. Now we always show the "veggie-burger" until the tablet breakpoint. If we're NOT in the UK we show the "one row" navigation from the `tablet` breakpoint up, and if we're in the UK we show the  the "one row" navigation from the `leftCol` breakpoint up.

### Veggie Burger (All regions until tablet)

Screenshot 2023-01-18 at 13 20 00

#### Non-UK regions

Screenshot 2023-01-18 at 13 40 00

#### UK

Screenshot 2023-01-18 at 13 40 12

### Default (UK only, from tablet until leftCol)

Screenshot 2023-01-18 at 13 41 29

### DIFF
--- a/support-frontend/assets/components/headers/header/header.tsx
+++ b/support-frontend/assets/components/headers/header/header.tsx
@@ -101,6 +101,9 @@ export default class Header extends Component<PropTypes, State> {
 		return (
 			<header
 				className={classNameWithModifiers('component-header', [
+					countryGroupId !== 'GBPCountries'
+						? 'one-row-from-tablet'
+						: 'one-row-from-desktop',
 					display === 'navigation' ? 'display-navigation' : null,
 					display === 'checkout' ? 'display-checkout' : null,
 				])}
--- a/support-frontend/assets/components/headers/header/header.tsx
+++ b/support-frontend/assets/components/headers/header/header.tsx
@@ -15,36 +15,7 @@ export type PropTypes = {
 	display?: 'navigation' | 'checkout' | 'guardianLogo' | void;
 };
 export type State = {
-	fitsLinksInOneRow: boolean;
-	fitsLinksAtAll: boolean;
 	isTestUser: boolean | null | undefined;
-};
-
-// ----- Metrics ----- //
-const getMenuStateMetrics = ({
-	menuRef,
-	logoRef,
-	containerRef,
-}: {
-	menuRef: Element;
-	logoRef: Element;
-	containerRef: Element;
-}): State => {
-	const [logoLeft, menuWidth, containerLeft, containerWidth] = [
-		logoRef.getBoundingClientRect().left,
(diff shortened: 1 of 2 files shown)


GHaberis:
[Support site navigation layout broken on non-UK Guardian Weekly Landing Page @ Mobile Breakpoint](https://trello.com/c/tEhI7KwW/989-support-site-navigation-layout-broken-on-non-uk-guardian-weekly-landing-page-mobile-breakpoint)

GHaberis:
I need to fix a small layout bug flagged by the UI tests

support-frontend/assets/components/headers/header/header.tsx
```diff
@@ -104,13 +73,9 @@ export default class Header extends Component<PropTypes, State> {
...
-				this.setState(
-					getMenuStateMetrics({
-						menuRef,
-						logoRef,
-						containerRef,
-					}),
-				);
+				this.setState({
+					isTestUser: getGlobal<boolean>('isTestUser'),
+				});
```
undergroundquizscene, line 78: Is it important to set this on window resize, or could it be set once earlier in the process?

undergroundquizscene (approved):

support-frontend/assets/components/headers/header/header.tsx
GHaberis: Good point, we wouldn't need to wait to call `setState` on `ResizeObserver` now we're not using JS to adjust the menu items.

paul-daniel-dempsey (approved):
Looks good to me, worked exactly as described  when ran PR. Maybe setState can just move up out of the resizeObserver then?

support-frontend/assets/components/headers/header/header.tsx
undergroundquizscene: Perfect, thanks for making that change!

PR #4673, opened 2023-01-19
Nudge variantA copy change

paul-daniel-dempsey:
Copy Change VariantA

From->
"Regular, reliable funding from readers is vital for our future. It protects our independence long term, so we can report with freedom and truth. Support us monthly from just $2."

To->
"Regular, reliable funding from readers is vital for our future. It protects our independence long term, so we can report **freely and without outside influence**. Support us monthly from just $2.

Screenshot 2023-01-19 at 08 52 25

## Is this an AB test?
- [x] Yes
- [ ] No


marialani (approved):
👍

PR #4675, opened 2023-01-19
Remove old abTests from abtestDefinitions

GHaberis:

I noticed when querying the the datalake we're still bucketing users (on the Contributions checkout in AUS) into an old `tickerTest` AB test that's no longer used, so we're tracking and recording them in this AB test that isn't running. Therefore I'm removing it from `abtestDefinitions`. 

Whilst doing this I noted we have 2 tests `localCurrencyTestV2` and `emailDigiSubEventsTest` which are no longer active (`isActive:false`). So I removed the tests and associated code paths. For `localCurrencyTestV2`this was just a couple lines of code, for `emailDigiSubEventsTest` there was a little more involved, I had to delete some components previously used on the now deprecated Digi Subs landing and thank you page.


i-hardy (approved):
Nice!

PR #4684, opened 2023-01-20
Add single contributions into the supporter product data Dynamo store

rupertbates:
This  PR adds code to the payment-api to write single contributions into the Supporter Product Data Dynamo store.

Doing this will allow mean that  we don't need to query the contributions data store from members-data-api whenever we get a request to retrieve a users single contributions - we can get both single contributions and subscriptions from the same store.  

This will be both more performant and reduce the load on the postgres database  as well as simplifying the code base.

### DIFF
--- a/support-payment-api/src/main/scala/services/SupporterProductDataService.scala
+++ b/support-payment-api/src/main/scala/services/SupporterProductDataService.scala
@@ -24,17 +24,16 @@ class SupporterProductDataService(environment: Environment) {
         contributionData.identityId
           .map(idAsLong =>
             SupporterRatePlanItem(
-              subscriptionName = contributionData.paymentId,
+              subscriptionName = s"${contributionData.paymentProvider.entryName} - ${contributionData.paymentId}",
               identityId = idAsLong.toString,
               gifteeIdentityId = None,
               productRatePlanId = "single_contribution",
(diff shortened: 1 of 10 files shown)


support-payment-api/src/main/scala/services/SupporterProductDataService.scala
```diff
@@ -0,0 +1,58 @@
...
+  })
+  def insertContributionData(
+      contributionData: ContributionData,
+  )(implicit executionContext: ExecutionContext): EitherT[Future, String, Unit] =
+    for {
+      item <- EitherT.fromEither[Future](
+        contributionData.identityId
+          .map(idAsLong =>
+            SupporterRatePlanItem(
+              subscriptionName = s"${contributionData.paymentProvider.entryName} - ${contributionData.paymentId}",
```
tomrf1, line 27: is `subscriptionName` likely to be used for anything?

tomrf1 (approved):
👍

support-payment-api/src/main/scala/services/SupporterProductDataService.scala
rupertbates: It's basically just to have a human readable field so we can make sense of the data in the dynamo table

PR #4699, opened 2023-01-24
Use /contribute over /support in geoRedirect

undergroundquizscene:
When running locally, we now redirect to `/support` in a couple of places, while our site is actually at `/contribute`. This PR updates those places to redirect straight to `/contribute`. Since we’re redirecting with Fastly anyway (in code and prod), this isn’t essential, but it’s a small change and makes things more consistent.


michaelbjacobson (approved):
Looks good to me! 👍🏼

PR #4705, opened 2023-01-26
Country drop down supporter plus fix

LAKSHMIRPILLAI:
Supporter Plus checkout: Fix layout bug with the country dropdown

[**Trello Card**](https://trello.com/c/CeJ4lNs4/932-supporter-plus-checkout-fix-layout-bug-with-the-country-dropdown)

## Why are you doing this?
Dropdown of country picker doesn't align properly when it's open.
Also, a small tweaks needs to be done on the 320 breakpoint to align the chevron with the link.

## Screenshots
Before Change
Frame_2077

After Change
image

image

### DIFF
--- a/support-frontend/assets/components/countryGroupSwitcher/countryGroupSwitcher.tsx
+++ b/support-frontend/assets/components/countryGroupSwitcher/countryGroupSwitcher.tsx
@@ -84,7 +84,7 @@ function CountryGroupSwitcher({
 			>
 				<Menu
 					style={{
-						top: bounds.top,
+						top: bounds.top + 30,
 						left: bounds.left,
 						position: 'absolute',
 					}}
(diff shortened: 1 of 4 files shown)


support-frontend/assets/components/countryGroupSwitcher/countryGroupSwitcher.tsx
```diff
@@ -84,7 +84,7 @@ function CountryGroupSwitcher({
 			>
 				<Menu
 					style={{
-						top: bounds.top,
+						top: bounds.top + 30,
```
paul-daniel-dempsey, line 87: might be better to remove 30 and use bounds.bottom (add it to the bounds useState)

paul-daniel-dempsey (approved):
Small change on countryGroupSwitcher and looks good to go

support-frontend/assets/components/countryGroupSwitcher/countryGroupSwitcher.tsx
marialani: We just had a catch up on this one and the `bounds` state here is used to store the outcome of the [getBoundingClientRect()](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect) function. We use that function to tell us how far the 'reference' element is from the top and the left of the page so I reckon adding 30 to the resolution of bounds.top here is correct

PR #4706, opened 2023-01-27
hideChooseYourAmount refactor

tomrf1:
I got quite confused by `currentAmountsLen`, `amountOfAmounts` etc - much of this logic is for deciding if the final button is full width


KaliedaRik (approved):
These make sense. I'll merge into my branch and test.

PR #4719, opened 2023-01-30
Add lambda for patron sign-up stripe webhook

undergroundquizscene:
At the moment, there is a scheduled lambda which gets information from stripe about newly-signed-up patrons and adds it to the SupporterProductDataPROD DynamoDB table. This PR makes a new lambda which can do the same job but is triggered by a stripe webhook rather than a schedule.

[**Trello Card**](https://trello.com/c/1LKM0HEu/998-change-stripe-patrons-lambda-into-a-webhook-rather-than-a-scheduled-lambda)

In more detail, this change:

- Makes that new lambda
- Adds an API gateway endpoint for calling it
- Adds necessary config for it to be called from both patron stripe accounts (GNM Patron Scheme and GNM Patron Scheme AUS)
- Adds necessary config for the existing cancel-event lambda to be called from GNM Patron Scheme AUS
- Changes existing stripe service calls to specify which account they’re calling (so that, for example, we can look up a customer in the right account, depending on which account’s webhook called the lambda)

We’ve tested this by deploying it to CODE, creating the necessary parameter store config for CODE, and triggering events on the stripe accounts in test mode, which show the calls succeeding.

## Post-Deployment Steps

Once this change is deployed, there are some post-deployment steps, which need to happen in order:

- Create a (non-test-mode) webhook (initially disabled) to call the sign-up event lambda:
  - [ ] On the GNM Patron Scheme account
  - [ ] On the GNM Patron Scheme AUS account
- Create a (non-test-mode) webhook (initially disabled) to call the cancel event lambda:
  - [ ] On the GNM Patron Scheme AUS account
- Add the following AWS Parameter Store entries:
  - [ ] /PROD/support/stripe-patrons-data/stripe-config/api-key-au
  - [ ] /PROD/support/stripe-patrons-data/stripe-config/cancelled-hook-signing-secret
  - [ ] /PROD/support/stripe-patrons-data/stripe-config/cancelled-hook-signing-secret-au
  - [ ] /PROD/support/stripe-patrons-data/stripe-config/sign-up-hook-signing-secret
  - [ ] /PROD/support/stripe-patrons-data/stripe-config/sign-up-hook-signing-secret-au
- [ ] Enable the new webhooks
- [ ] Remove the schedule from the existing lambda
- Remove these AWS Parameter Store entries:
  - [ ] /CODE/support/stripe-patrons-data/stripe-config/signing-secret
  - [ ] /PROD/support/stripe-patrons-data/stripe-config/signing-secret

### DIFF
--- a/cdk/lib/__snapshots__/stripe-patrons-data.test.ts.snap
+++ b/cdk/lib/__snapshots__/stripe-patrons-data.test.ts.snap
@@ -211,12 +211,15 @@ exports[`The Stripe patrons data stack matches the snapshot 1`] = `
       "Type": "AWS::IAM::Role",
       "UpdateReplacePolicy": "Retain",
     },
-    "RestApiDeployment180EC5031415bce23b484655df855d90ebafb1d3": {
+    "RestApiDeployment180EC5033fc0c8ea808241777537b04b239b12f5": {
       "DependsOn": [
         "RestApipatron1DDE9314",
         "RestApipatronsubscriptioncancelcountryIdPOSTFA2082F3",
         "RestApipatronsubscriptioncancelcountryId57A3F9F4",
         "RestApipatronsubscriptioncancel0F56DD38",
+        "RestApipatronsubscriptioncreatecountryIdPOST1E68C8CD",
+        "RestApipatronsubscriptioncreatecountryId97F3026B",
+        "RestApipatronsubscriptioncreateC44D8A79",
         "RestApipatronsubscriptionAE58DC90",
       ],
       "Properties": {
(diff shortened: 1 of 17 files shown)


undergroundquizscene:
Question for reviewers: the only test we’ve added is the deserialisation test of the stripe event – is there anything else you can think of that would be good to add tests for?

undergroundquizscene:
[Change Stripe Patrons lambda into a webhook rather than a scheduled lambda](https://trello.com/c/1LKM0HEu/998-change-stripe-patrons-lambda-into-a-webhook-rather-than-a-scheduled-lambda)

undergroundquizscene:
@rupertbates Requesting review from you as you made the trello card – hope that’s ok :)

michaelbjacobson (approved):
Looks good to me, great work! Would be good to get @rupertbates review as well though!

undergroundquizscene:
I’m going to wait a week before deleting the signing-key parameter store entries, to allow time for CODE and people’s branches to catch up

PR #4722, opened 2023-01-31
Refactor user reducer and migrate to RTK

i-hardy:

This PR migrates the user reducer to a Redux Toolkit slice and removes many unused properties. It also now stores all the product information we retrieve from MDAPI when checking the user's recurring contributor status, so we can check their ownership of various products.

[**Trello Card**](https://trello.com/c/R2CmOyBD)

## Why are you doing this?

The user reducer was a complex bit of our legacy Redux code, and this simplifies it down to the information we actually need and use while aligning it with the rest of our RTK code.


i-hardy:
[Migrate the user reducer to Redux Toolkit](https://trello.com/c/R2CmOyBD/875-migrate-the-user-reducer-to-redux-toolkit)

support-frontend/assets/helpers/user/user.ts
```diff
@@ -45,6 +41,14 @@ function isTestUser(): boolean {
 	return isDefined(uatMode) || isDefined(testCookie);
 }
 
+function getUserStateField(): string | undefined {
+	const user = getGlobal<User>('user');
+	if (user) {
+		return user.address4 ?? window.guardian.geoip?.stateCode;
+	}
+	return window.guardian.geoip?.stateCode;
```
undergroundquizscene, line 49: Is there a difference between this and the following? It looks to me like the should be the same, so the shorter form seems preferable, but I’m not very confident in my typescript.

```suggestion
	return user?.address4 ?? window.guardian.geoip?.stateCode;
```

support-frontend/assets/components/subscriptionCheckouts/thankYou/marketingConsentContainerGift.ts
```diff
@@ -7,15 +7,14 @@ import { sendTrackingEventsOnClick } from 'helpers/productPrice/subscriptions';
...
-import type { Action } from 'helpers/user/userActions';
 
 const mapStateToProps = (state: CheckoutState) => ({
 	confirmOptIn: state.page.checkoutForm.marketingConsent.confirmOptIn,
 	email: getEmail(state),
 	csrf: state.page.checkoutForm.csrf,
 });
 
-function mapDispatchToProps(dispatch: Dispatch<Action>) {
+function mapDispatchToProps(dispatch: Dispatch) {
```
paul-daniel-dempsey, line 17: These look like a simplification of dispatching actions would guess?

support-frontend/assets/helpers/redux/checkout/personalDetails/state.ts
```diff
@@ -39,6 +39,8 @@ export const personalDetailsSchema = z
 		),
 		email: emailRules,
 		confirmEmail: z.optional(z.string()),
+		// isSignedIn is maintained in personal details state purely for the email matching check
```
paul-daniel-dempsey, line 42: Great comment, this answers my previous question about (personalDetails.isSignedIn) already!

support-frontend/assets/helpers/user/user.ts
```diff
@@ -1,19 +1,15 @@
...
-import { routes } from 'helpers/urls/routes';
-import { defaultUserActionFunctions } from 'helpers/user/defaultUserActionFunctions';
-import type { Action, UserSetStateActions } from 'helpers/user/userActions';
 
 export type User = {
 	firstName: Option<string>;
 	lastName: Option<string>;
 	email: Option<string>;
 	isSignedIn: boolean;
+	address4?: string;
```
paul-daniel-dempsey, line 12: Not sure what this is? Some sort of 4th address row field (for testing)? If empty, we load with stateCode? Is there a better name or perhaps need to go changing Scala code perhaps?

support-frontend/assets/pages/supporter-plus-landing/setup/setUpRedux.ts
```diff
@@ -179,28 +182,29 @@ function selectInitialContributionTypeAndPaymentMethod(
...
-
 export function setUpRedux(store: ContributionsStore): void {
 	const dispatch = store.dispatch;
 	const state = store.getState();
 	// TODO - move these settings out of the redux store, as they only change once, upon initialisation
 	const contributionTypes = getContributionTypes(state);
 	dispatch(setContributionTypes(contributionTypes));
-	getStoredEmail(dispatch);
+
+	setUpUserState(dispatch);
```
paul-daniel-dempsey, line 192: My understanding is this where the thunk (asynchronously) initialises User.supporterStatus from the user-attributes json file? From a recurring contributor.

paul-daniel-dempsey (approved):
Much neater with the User Slice, sorry for my comments, mostly to aid my understanding of Redux Toolkit am afraid (hope thats ok). Could not figure out how to test the User.SupporterStatus fields (maybe I need an identity test account setup?) If you have time would be great to understand how this is tested?

support-frontend/assets/components/subscriptionCheckouts/thankYou/marketingConsentContainerGift.ts
i-hardy: No, it's just a simplification of the type because that particular `Action` type no longer exists and `mapDispatchToProps` doesn't play very well with RTK's store-specific `Dispatch` types.

support-frontend/assets/helpers/user/user.ts
i-hardy: `address4` should be the fourth line of the address if we have it, so state/province/county. This is information that is added to `window.guardian` from the back end, we don't update it or set it in client-side code. I'm not sure where the names of the properties come from but it's likely MDAPI or Identity, so not trivial to change it.

support-frontend/assets/pages/supporter-plus-landing/setup/setUpRedux.ts
i-hardy: `getRecurringContributorStatus` is the thunk, yes, it talks to MDAPI to get the user's current product status.

support-frontend/assets/components/payPalPaymentButton/payPalRecurringContainer.tsx
```diff
@@ -61,7 +61,7 @@ export function PayPalButtonRecurringContainer({
 
 	const buttonProps = getPayPalButtonProps({
 		csrf,
-		isTestUser: isTestUser ?? false,
+		isTestUser,
```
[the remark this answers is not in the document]
i-hardy: Yep!

i-hardy:
Yes, the supporter status data requires an Identity account with an existing product or products; this is probably easiest to test on CODE.

support-frontend/assets/helpers/user/user.ts
paul-daniel-dempsey: Ok

PR #4733, opened 2023-02-02
Add a new integration test to create a supporter plus sub in Austria

rupertbates:
Because they tax digital news products so it is useful to have some being created in Zuora on a regular basis.

I've also added a git pre-commit hook to run `scalafmtCheckAll` if the commit contains Scala files, to avoid the annoying situation where you push files and they then fail the build with a formatting error.

### DIFF
--- a/support-frontend/.husky/pre-commit
+++ b/support-frontend/.husky/pre-commit
@@ -2,3 +2,16 @@
 . "$(dirname "$0")/_/husky.sh"
 
 cd support-frontend && yarn lint-staged
+
+if git diff-index --cached  --diff-filter=AM --name-only  HEAD | grep --quiet .scala$
+then
+  echo "Running scalafmtCheckAll to check for Scala formatting errors"
+  cd ..
+  if ! sbt scalafmtCheckAll
+  then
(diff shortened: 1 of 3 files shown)


support-frontend/.husky/pre-commit
```diff
@@ -2,3 +2,11 @@
 . "$(dirname "$0")/_/husky.sh"
 
 cd support-frontend && yarn lint-staged
+
+if git diff-index --cached  --diff-filter=AM --name-only  HEAD | grep --quiet .scala$
+then
+  echo "Running scalafmtCheckAll"
+  cd ..
+  sbt scalafmtCheckAll
```
undergroundquizscene, line 10: Would it be possible/worth it to change from checking the formatting to actually performing the formatting? I can see it might be annoying, but my next step after this fails will always be manually running the formatting, so getting the hook to do it would save me time (though I’ll probably still need to stage the changes – I guess that provides a bit of safety in case I didn’t want the formatting).

rupertbates: Ok I've updated it to run the format if the check fails

undergroundquizscene (approved):

PR #4741, opened 2023-02-07
Remove client-side code for digital subscription landing page and checkout

i-hardy:
## What are you doing in this PR?

This PR:
- Removes the code for the digi subs landing page and checkout, both regular and gift
- Removes files 'orphaned' by this and the contributions code deletion, ie. modules and components no longer in use
- Moves some thank-you page components still in use by the digi subs gift redemption form to that page folder

[**Trello Card**](https://trello.com/c/GrdAFt7G)

## Why are you doing this?

As digi sub purchases have now been discontinued we no longer need this code.

PR #4746, opened 2023-02-08
Restore setting a cookie when a user makes a one-off contribution

i-hardy:

This restores setting the `gu.contributions.contrib-timestamp` cookie via the thank you page for users who have made a one-off contribution.

## Why are you doing this?

While clearing up orphaned code in #4741 I noticed that the `contributionsCookies.ts` module was no longer being imported anywhere; however this code serves an important purpose in setting a cookie enabling us to identify recent one-off contributors. This cookie is regarded as essential so there are no consent issues around setting it again.


marialani (approved):
👍

paul-daniel-dempsey (commented):
all good

PR #4759, opened 2023-02-10
Fix for NaN error with  Other Amount Field

LAKSHMIRPILLAI:

Click on single, then click on other amount button. Then (without touching the exposed other amount field) click on monthly, and then click on single again. We get a NaN error in the console.This PR fixes that

[**Trello Card**](https://trello.com/c/DrwJf6ix/1054-nan-error-with-other-amount-field)

## Screenshots
image


marialani:
Could we now also remove the comment from `selectedAmounts.ts` about handling the other amount when it's NaN?

i-hardy (approved):
Great catch!

PR #4760, opened 2023-02-10
Validate Zuora Product Catalog JSON using Circe 

paul-daniel-dempsey:
Co-authored-by: Emily Bourke

Explicitly specifies the schema of the catalog json which we can then decode with circe to the `ZuoraCatalog` object.
Map/Convert `ZuoraCatalog` object to the required `Catalog` object.
Using the original faulty catalog.json (`catalog-bad.json`), create a test decode case, assert error is raised upon failure.
CatalogService decode Json to the `ZuoraCatalog` and convert onto `Catalog`. 

[**Trello Card**](https://trello.com/c/PWbn6bJN/458-use-circe-to-validate-zuora-product-catalog-json)

## Why are you doing this?

Support-Frontend validation of the S3 Zuora catalog JSON is not very strict and did allow for invalid JSON to pass.

## Is this an AB test?
- [ ] Yes
- [x] No


support-models/src/main/scala/com/gu/support/catalog/ZuoraCatalog.scala
```diff
@@ -0,0 +1,31 @@
+package com.gu.support.catalog
+
+import io.circe._
+import io.circe.generic.semiauto.{deriveDecoder, deriveEncoder}
+
+case class ZuoraCatalog(
+    products: Array[ZuoraProduct],
```
rupertbates, line 7: Use List rather than Array for all these iterable types. Arrays in Scala are mutable which we don't want

support-models/src/main/scala/com/gu/support/catalog/Catalog.scala
```diff
@@ -20,54 +20,41 @@ object Catalog {
...
     // have their own price list. To get the total prices for these products therefore
     // we need to sum all of the price lists
-    priceLists.flattenJsonArrays
-      .flatMap(_.as[Price].toOption) // convert the Json to Price objects as they're easier to work with
+    priceLists
       .groupBy(_.currency)
-      .map(sumPrices)
-      .map({ case (_, price) => price.asJson }) // convert back to Json
+      .toArray
+      .map(sumPrices(_)._2)
```
rupertbates, line 32:
```suggestion
      .map(sumPrices)
      .map({ case (_, price) => price })
```
This is a nitpick, but to me this is easier to read because it shows that we are returning the price, `_2` could be anything

support-models/src/main/scala/com/gu/support/catalog/ZuoraCatalog.scala
undergroundquizscene: Ah, thanks – I hadn’t realised Arrays are mutable. We chose it over List because I was thinking that linked lists would lead to poor cache locality for large collections like this. For this reason, I’m now leaning towards Vector, but since I know Scala types less well I reckon I should ask: do you think that’s worthwhile?

support-services/src/main/scala/com/gu/support/catalog/CatalogService.scala
```diff
@@ -61,7 +61,7 @@ class CatalogService(val environment: TouchPointEnvironment, jsonProvider: Catal
   private lazy val catalog = {
 
     jsonProvider.get.flatMap { json =>
-      val attempt = json.as[Catalog]
+      val attempt = json.as[ZuoraCatalog].map { Catalog.convert }
```
rupertbates, line 64:
```suggestion
      val attempt = json.as[ZuoraCatalog].map(Catalog.convert)
```
Another nitpick but you only need curly brackets for multi-line code blocks, otherwise use standard braces

rupertbates (approved):
Good work, just change the Arrays to Lists and then good to go 👍

support-models/src/main/scala/com/gu/support/catalog/ZuoraCatalog.scala
rupertbates: I would still say use List since all we are doing is mapping over it. Whilst Vector is more performant for quite a few operation types, realistically performance in an application of this kind will not be affected by a choice of a collection implementation so keeping consistency with the rest of the codebase (which uses List almost exclusively) is likely of more value.

undergroundquizscene: Ok, that sounds good

paul-daniel-dempsey: Great, ported to List now and will do going forwards.

support-models/src/main/scala/com/gu/support/catalog/Catalog.scala
```diff
@@ -20,54 +20,41 @@ object Catalog {
...
-      .getField("Saving__c")
-      .getOrElse(Json.fromString("null"))
-      .asString
-      .getOrElse("null") match {
-      case "null" => Json.Null
-      case i: String => Json.fromInt(i.toInt)
-    }
-
-  def sumPriceLists(priceLists: List[Json]): Iterable[Json] = {
+  def sumPriceLists(priceLists: Array[Price]): Array[(Price)] = {
```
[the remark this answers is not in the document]
paul-daniel-dempsey: Yep, done.

support-services/src/main/scala/com/gu/support/catalog/CatalogService.scala
paul-daniel-dempsey: Yep, done, will remember going forwards

support-models/src/main/scala/com/gu/support/catalog/Catalog.scala
paul-daniel-dempsey: Yep, agree easier to see price taken out. Changed.

undergroundquizscene (approved):

undergroundquizscene:
(Approval pending final change discussed on video call)

PR #4764, opened 2023-02-13
Restore contribution validation

undergroundquizscene:
This validation is currently disabled for recurring contributions below the threshold for supporter plus. It was disabled [in October 2021](https://github.com/guardian/support-frontend/commit/0ef5a8277d5b45209752e123a8c4a36b51b5adbd#diff-a6dbd081ad177ddb469e4501b720b5bf9bbb45801fa6ad382c6c0be4e28c25daL18), and I don’t think it should stay disabled any longer, as it would now be called if enabled.

I’ve tested in CODE, verified that the validation is called (by curling the endpoint with an empty billing account ID), and that I can still get through some normal user journeys (new recurring contribution with existing card, new recurring contribution with new card, new recurring contribution with new direct debit). The plan is to merge this and keep an eye out for any errors, though I don’t _think_ there will be any.


undergroundquizscene:
[Restore contribution validation](https://trello.com/c/91DGX15z/1085-restore-contribution-validation)

GHaberis (approved):
Looks good 👍

PR #4766, opened 2023-02-13
Only render existing payment methods that have a valid billingAccountId

GHaberis:
After we enabled existing payment methods on the new checkout we began seeing failed checkouts in our Step Function executions. These were caused by an empty string for the `paymentMethod`'s `billingAccountId` in the Step Function's input JSON, which causes the failure.

In response we introduced a change (https://github.com/guardian/support-frontend/pull/4667) that aimed to block users from checking out if the existing payment method they selected had an invalid `billingAccountId`.  It would block checking out by disabling the submit button on the checkout.

Unfortunately there was an issue with the condition used for disabling/enabling the button, and we were disabling it for users with a valid `billingAccount` and enabling for those without.

Rather than fix I've removed that logic as I'm not sure disabling the submit button is the best in terms of UX as visually there's nothing to indicate it's disabled or why it's disabled, it just doesn't respond to clicks. Instead I've made a change that filters the  
`existingPaymentMethodList` list in the `paymentMethodSelector` and removes existing payment methods with an invalid `billingAccountId`, so they shouldn't even be rendered.

### DIFF
--- a/support-frontend/assets/components/existingMethodPaymentButton/existingCardPaymentButton.tsx
+++ b/support-frontend/assets/components/existingMethodPaymentButton/existingCardPaymentButton.tsx
@@ -13,6 +13,7 @@ export function ExistingCardPaymentButton(): JSX.Element {
 	);
 
 	const payWithExistingCard = useFormValidation(function pay() {
+		console.log('payWithExistingCard --->', selectedPaymentMethod);
 		void dispatch(
 			onThirdPartyPaymentAuthorised({
 				paymentMethod: 'ExistingCard',
--- a/support-frontend/assets/components/existingMethodPaymentButton/existingCardPaymentButton.tsx
+++ b/support-frontend/assets/components/existingMethodPaymentButton/existingCardPaymentButton.tsx
@@ -21,6 +21,8 @@ export function ExistingCardPaymentButton(): JSX.Element {
 		);
 	});
 
+	console.log('selectedPaymentMethod --->', selectedPaymentMethod);
+
 	return (
 		<DefaultPaymentButtonContainer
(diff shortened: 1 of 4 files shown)


GHaberis:
[Only render existing payment methods that have a valid billingAccountId](https://trello.com/c/jgqllPON/1089-only-render-existing-payment-methods-that-have-a-valid-billingaccountid)

michaelbjacobson (approved):
👍🏼

undergroundquizscene (commented):
This looks good and covers the case where the billingAccountId is falsy, but what if it’s the `selectedPaymentMethod` instead that’s falsy? (https://github.com/guardian/support-frontend/pull/4766/files#diff-a610ea58492a416ab2c0406e273a4c8e36eb23667e1b2d953a17156f687d23ceL11)

Is that possible? I can see the type system doesn’t rule it out, which is frustrating: feels like it should be possible for us to require it to be non-null if the button is clicked.

GHaberis:
> This looks good and covers the case where the billingAccountId is falsy, but what if it’s the `selectedPaymentMethod` instead that’s falsy? (https://github.com/guardian/support-frontend/pull/4766/files#diff-a610ea58492a416ab2c0406e273a4c8e36eb23667e1b2d953a17156f687d23ceL11)
> 
> Is that possible? I can see the type system doesn’t rule it out, which is frustrating: feels like it should be possible for us to require it to be non-null if the button is clicked.

It's a good question, in reality I can't see a way it'd be possible for `selectedPaymentMethod` to be `undefined` here, but the type states otherwise. I'd be inclined to release the change and monitor alarms for any further instances of submissions where billingAccountId is an empty string.

PR #4768, opened 2023-02-14
Remove schedule from StripePatronsDataLambda

undergroundquizscene:
Now that sign-up is being handled by a webhook, we don’t need this schedule. We do want to keep the lambda though, in case we want to run it again sometime.

This PR is post-deployment work needed for [this Trello card](https://trello.com/c/1LKM0HEu/998-change-stripe-patrons-lambda-into-a-webhook-rather-than-a-scheduled-lambda), as detailed in https://github.com/guardian/support-frontend/pull/4719

### DIFF
--- a/cdk/lib/__snapshots__/stripe-patrons-data.test.ts.snap
+++ b/cdk/lib/__snapshots__/stripe-patrons-data.test.ts.snap
@@ -901,43 +901,6 @@ exports[`The Stripe patrons data stack matches the snapshot 1`] = `
       },
       "Type": "AWS::IAM::Role",
     },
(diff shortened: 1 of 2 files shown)


undergroundquizscene:
[Change Stripe Patrons lambda into a webhook rather than a scheduled lambda](https://trello.com/c/1LKM0HEu/998-change-stripe-patrons-lambda-into-a-webhook-rather-than-a-scheduled-lambda)

michaelbjacobson (approved):
Nice!

undergroundquizscene:
I’m not sure why prout reckons it’s not on prod – I can see that the riff raff build succeeded, and sure enough the lambda no longer has a schedule in the AWS console.

PR #4773, opened 2023-02-15
Add QM event to track payment method at the point of conversion

i-hardy:

This adds a Quantum Metric event recording the user's selected payment method at the point of conversion, enabling us to better track conversions via payment method. It also refactors some repeated code in our various QM tracking functions to make it easier to establish when a call can be sent.

[**Trello Card**](https://trello.com/c/F1P7jkM6)

### DIFF
--- a/support-frontend/assets/helpers/tracking/quantumMetric.ts
+++ b/support-frontend/assets/helpers/tracking/quantumMetric.ts
@@ -41,6 +41,7 @@ enum SendEventContributionAmountUpdate {
 
 enum SendEventContributionPaymentMethodUpdate {
 	PaymentMethod = 103,
+	PaymentMethodAtConversion = 110,
 }
 
 enum SendEventContributionCheckoutConversion {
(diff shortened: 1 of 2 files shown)


i-hardy:
[Quantum Metric: Enable tracking of payment method at conversion on S+ checkout](https://trello.com/c/F1P7jkM6/1055-quantum-metric-enable-tracking-of-payment-method-at-conversion-on-s-checkout)

support-frontend/assets/helpers/tracking/quantumMetric.ts
```diff
@@ -264,46 +269,52 @@ function sendEventContributionCartValue(
...
 	}
 }
 
+function sendEventConversionPaymentMethod(paymentMethod: PaymentMethod): void {
+	void ifQmPermitted(() => {
+		sendEvent(
+			SendEventContributionPaymentMethodUpdate.PaymentMethodAtConversion,
+			false,
+			paymentMethod.toString(),
+		);
```
undergroundquizscene, line 314: Do you not need to use `sendEventWhenReadyTrigger` here? I guess the thinking is that since it’s called late enough in the process, we can assume quantum metric is ready. Is there any harm in using the check function anyway, just in case?

undergroundquizscene (approved):

PR #4775, opened 2023-02-15
Fix inconsistency in populating window.guardian.user

i-hardy:

The user reducer migration saga continues from #4774. This PR fixes an inconsistency between the way we populate `window.guardian.user` on the subscriptions and contributions checkouts, in order to overcome an issue where logged-out users were incorrectly being identified as logged in on the subscriptions checkouts.


marialani (approved):
LGTM

support-frontend/app/views/subscriptionCheckout.scala.html
```diff
@@ -40,11 +40,25 @@
...
+      @idUser.map { user =>
+        window.guardian.user = {
+          id: "@user.id",
+          email: "@user.primaryEmailAddress",
+          @user.publicFields.displayName.map { displayName =>
+            displayName: "@displayName",
+          }
+          @for(firstName <- user.privateFields.firstName; lastName <- user.privateFields.secondName) {
+            firstName: "@firstName",
+            lastName: "@lastName",
```
undergroundquizscene, line 52: Given a user who only has one name (first or last) set, will this leave out the name from the response? If so, does that matter?

i-hardy: It shouldn't matter, the user can enter their name manually into the form.

PR #4776, opened 2023-02-15
Fix Subscribe GW firstName field not limited to 40chars

LAKSHMIRPILLAI:
This is to place a validation for lengths of FirstName and LastName field  within the Guardian Weekly checkout ,It should not be more than 40 characters.

[**Trello Card**](https://trello.com/c/zFFAjsWu/1024-subscribe-gw-firstname-field-not-limited-to-40chars)

Validation not appearing for FirstNames greater than 40 characters (SalesForce limit) in length within the Guardian Weekly checkout - see screenshot for example.

## Screenshots

Before Change

image

After Change

image
image


i-hardy (approved):
Can't believe we had this lurking around unvalidated all this time, good fix!

PR #4789, opened 2023-02-17
Supporter plus nudge annual (variant) copy changes

paul-daniel-dempsey:
Nudging component, encouraging single to monthly or modified annual recurring contributions
Rename test  to 'ab-singleToRecurringV3'

[**Trello Card**](https://trello.com/c/0bPcisad/1082-ab-test-copy-change-for-the-single-nudge-on-the-checkout)

Annual Features->
- copy change variant (control stays monthly)
- minimum weekly contribution rounded (p = GPB Countries, displayCcy = remainder)

2-cohort A/B test for implementation ->
https://support.thegulocal.com/uk/contribute#ab-singleToRecurringV3=control (Monthly)
https://support.thegulocal.com/uk/contribute#ab-singleToRecurringV3=variant (NEW Annual)

## Is this an AB test?
- [x] Yes
- [ ] No

### DIFF
--- a/support-frontend/assets/components/checkoutNudge/checkoutNudgeContainer.tsx
+++ b/support-frontend/assets/components/checkoutNudge/checkoutNudgeContainer.tsx
@@ -28,8 +28,8 @@ export function CheckoutNudgeContainer({
 	const [displayNudge, setDisplayNudge] = useState(true);
 
 	const recurringType =
-		abParticipations.singleToRecurringV2 === 'control' ||
-		!abParticipations.singleToRecurringV2
+		abParticipations.singleToRecurringV3 === 'control' ||
+		!abParticipations.singleToRecurringV3
 			? 'MONTHLY'
 			: 'ANNUAL';
 	const currencyGlyph = glyph(detect(countryGroupId));
--- a/support-frontend/assets/components/checkoutNudge/checkoutNudgeContainer.tsx
+++ b/support-frontend/assets/components/checkoutNudge/checkoutNudgeContainer.tsx
@@ -34,9 +34,15 @@ export function CheckoutNudgeContainer({
 			: 'ANNUAL';
 	const currencyGlyph = glyph(detect(countryGroupId));
 	const minAmount = config[countryGroupId][recurringType].min;
-	const paragraphCopyVariant = `Regular, reliable support powers Guardian journalism in perpetuity. If you can, please consider setting up ${
-		recurringType === 'MONTHLY' ? 'a' : 'an'
-	} ${recurringType.toLowerCase()} payment today from just  ${currencyGlyph}${minAmount} – it takes less than a minute.`;
+
+	const minWeeklyAmount =
+		countryGroupId === 'GBPCountries'
+			? Math.ceil((minAmount * 100) / 52).toString() + `p`
+			: currencyGlyph +
+			  (Math.ceil((minAmount * 100) / 52) / 100).toFixed(2).toString();
+
+	const paragraphCopyMonthly = `Regular, reliable support powers Guardian journalism in perpetuity. If you can, please consider setting up a ${recurringType.toLowerCase()} payment today from just  ${currencyGlyph}${minAmount} – it takes less than a minute.`;
+	const paragraphCopyAnnual = `Regular, reliable support powers Guardian journalism in perpetuity. If you can, please consider setting up an ${recurringType.toLowerCase()} payment from ${currencyGlyph}${minAmount} – that’s just ${minWeeklyAmount} a week.`;
 
 	function onNudgeClose() {
 		setDisplayNudge(false);
@@ -52,7 +58,8 @@ export function CheckoutNudgeContainer({
 		nudgeTitleCopySection2: `Support us every ${
 			recurringType === 'MONTHLY' ? 'month' : 'year'
 		}`,
-		nudgeParagraphCopy: paragraphCopyVariant,
+		nudgeParagraphCopy:
+			recurringType === 'MONTHLY' ? paragraphCopyMonthly : paragraphCopyAnnual,
 		nudgeLinkCopy: `See ${recurringType.toLowerCase()}`,
 		onNudgeClose,
 		onNudgeClick,
--- a/support-frontend/assets/helpers/abTests/abtestDefinitions.ts
+++ b/support-frontend/assets/helpers/abTests/abtestDefinitions.ts
@@ -39,7 +39,7 @@ export const pageUrlRegexes = {
 	},
 };
 export const tests: Tests = {
-	singleToRecurringV2: {
+	singleToRecurringV3: {
 		variants: [
 			{
 				id: 'control',


paul-daniel-dempsey:
[AB test: copy change for the single nudge on the checkout](https://trello.com/c/0bPcisad/1082-ab-test-copy-change-for-the-single-nudge-on-the-checkout)

support-frontend/assets/components/checkoutNudge/checkoutNudge.tsx
```diff
@@ -45,52 +46,50 @@ const top = css`
 const topheading = css`
 	margin-top: ${space[3]}px;
 `;
-const heading = (backColor: string, marginBottom: number) => css`
-	margin-left: 10px;
-	color: ${backColor};
-	${headline.xxsmall({ fontWeight: 'bold', lineHeight: 'tight' })};
-	line-height: 108%;
+const heading = (
+	backColor: string,
```
marialani, line 50: Had assumed this would be for a background colour so maybe something like this instead?

```suggestion
	color: string,
```

support-frontend/assets/components/checkoutNudge/checkoutNudge.tsx
```diff
@@ -45,52 +46,50 @@ const top = css`
...
-		margin-left: 12px;
+  ${from.mobileMedium} {
 		margin-bottom: ${marginBottom}px;
-		${headline.xsmall({ fontWeight: 'bold', lineHeight: 'tight' })};
-		line-height: 108%;
-	}
+    ${
+			shrinkHeadline
+				? headline.xxsmall({ fontWeight: 'bold', lineHeight: 'tight' })
+				: headline.xsmall({ fontWeight: 'bold', lineHeight: 'tight' })
```
marialani, line 63: You could change the font-size like this:
```suggestion
			font-size: ${shrinkHeadline ?  ...px : ...px}
```

support-frontend/assets/components/checkoutNudge/checkoutNudge.tsx
```diff
@@ -128,13 +131,21 @@ export function CheckoutNudge({
...
 							{nudgeTitleCopySection1}
 						</h2>
 					</div>
 					<CheckoutNudgeCloseButton onClose={onNudgeClose} />
 				</div>
-				<h2 css={heading(neutral[7], 0)}>{nudgeTitleCopySection2}</h2>
+				<h2
+					css={heading(
+						neutral[7],
+						0,
```
marialani, line 143: This could also be an optional prop for the `heading` css function if you were to use a ternary within the function

support-frontend/assets/components/checkoutNudge/checkoutNudgeContainer.tsx
```diff
@@ -28,15 +28,24 @@ export function CheckoutNudgeContainer({
...
+		countryGroupId === 'GBPCountries'
+			? Math.ceil((minAmount * 100) / 52).toString() + `p`
+			: currencyGlyph +
+			  (Math.ceil((minAmount * 100) / 52) / 100).toFixed(2).toString();
+
+	const title2CopyMonthly = `Support us every month`;
+	const paragraphCopyMonthly = `Regular, reliable support powers Guardian journalism in perpetuity. If you can, please consider setting up a ${recurringType.toLowerCase()} payment today from just  ${currencyGlyph}${minAmount} – it takes less than a minute.`;
+
+	const title2CopyAnnual = `From just ${currencyGlyph}${minAmount} (${minWeeklyAmount} a week)`;
+	const paragraphCopyAnnual = `Funding Guardian journalism every year doesn’t need to be expensive. Make a bigger impact today, and protect our independence long term. Please consider annual support.`;
```
marialani, line 48: This is a bit confusing, it's quite difficult to tell from the code what the resultant string will be. It would probably be best to use a function here as determining the string is more complicated. You could also use a switch statement inside the function

support-frontend/assets/components/checkoutNudge/checkoutNudge.tsx
paul-daniel-dempsey: Ok have passed through color

support-frontend/assets/components/checkoutNudge/checkoutNudge.tsx
paul-daniel-dempsey: Ah yes, much neater, have replaced with suggestion

support-frontend/assets/components/checkoutNudge/checkoutNudge.tsx
paul-daniel-dempsey: Struggled with 2 optional parameters, then accessing one or the other. Ended up creating an object with 2 optional parameters inside (line 51). Think it does what you suggest but perhaps there is a better/simpler way?

support-frontend/assets/components/checkoutNudge/checkoutNudgeContainer.tsx
paul-daniel-dempsey: Ah yes, have placed in a neater function (line 50) so easier to read I hope.

support-frontend/assets/components/checkoutNudge/checkoutNudge.tsx
marialani: If you make it optional you could do something like this inside the fn: 
`margin-bottom: ${marginBottom ? marginBottom : 0}px;`

marialani: Also don't think these need to passed in as an object

support-frontend/assets/components/checkoutNudge/checkoutNudgeContainer.tsx
```diff
@@ -48,12 +79,12 @@ export function CheckoutNudgeContainer({
...
 		contributionType,
 		nudgeDisplay: displayNudge,
-		nudgeTitleCopySection1: 'Make a bigger impact',
-		nudgeTitleCopySection2: `Support us every ${
-			recurringType === 'MONTHLY' ? 'month' : 'year'
-		}`,
-		nudgeParagraphCopy: paragraphCopyVariant,
+		nudgeTitleCopySection1: titleNudge,
+		nudgeTitleCopySection2: title2Nudge,
+		nudgeParagraphCopy: paragraphNudge,
```
marialani, line 84: Just a suggestion but could also simplify these prop names 

```suggestion
		nudgeTitle: ...,
		nudgeSubtitle: ...,
		nudgeParagraphCopy: ...,
```

support-frontend/assets/components/checkoutNudge/checkoutNudgeContainer.tsx
marialani: That function looks great! Just one small note on the switch statement

support-frontend/assets/components/checkoutNudge/checkoutNudgeContainer.tsx
```diff
@@ -28,16 +28,47 @@ export function CheckoutNudgeContainer({
...
+		minWeeklyAmount: string,
+	): string[] {
+		switch (recurringType) {
+			case 'ANNUAL': {
+				return [
+					`Support us every year`,
+					`from just ${minAmount} (${minWeeklyAmount} a week)`,
+					`Funding Guardian journalism every year doesn’t need to be expensive. Make a bigger impact today, and protect our independence long term. Please consider annual support.`,
+				];
+			}
```
marialani, line 62: It would be a good idea to explicitly add a case for MONTHLY to be explicit on what is returned for the MONTHLY case. This function should never hit the default case but the default is needed regardless for the switch.

support-frontend/assets/components/checkoutNudge/checkoutNudgeContainer.tsx
paul-daniel-dempsey: Ah yes, subtitle was the name I was looking for, thanks, changed to match

support-frontend/assets/components/checkoutNudge/checkoutNudgeContainer.tsx
paul-daniel-dempsey: Yep, have added monthly and default now

support-frontend/assets/components/checkoutNudge/checkoutNudge.tsx
paul-daniel-dempsey: Yep, removed object (was unnecessarily complicated), and applied suggestion

marialani (approved):
Looks good! 👍

PR #4791, opened 2023-02-21
Set up first mobile optimisation test for the supporter plus checkout

i-hardy:

This PR sets up the first mobile optimisation test. I've also refactored a few components used in the checkout that were relevant to the test to make them a little more flexible and easier to apply conditional styling to.

[**Trello Card**](https://trello.com/c/zdqCSfuk)

## Why are you doing this?

This test aims to start to streamline the user flow on mobile.

## Screenshots
Screenshot 2023-02-22 at 10-57-53 Support the Guardian

### DIFF
--- a/support-frontend/assets/components/checkoutBenefits/checkoutBenefitsList.tsx
+++ b/support-frontend/assets/components/checkoutBenefits/checkoutBenefitsList.tsx
@@ -8,6 +8,7 @@ import {
 } from '@guardian/source-foundations';
 import CancelAnytimeTooltip from 'components/tooltip/Tooltip';
 import type { CountryGroupId } from 'helpers/internationalisation/countryGroup';
+import type { CSSOverridable } from 'helpers/types/cssOverrideable';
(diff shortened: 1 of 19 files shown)


marialani (approved):
Looking good!

marialani:
Only note from me on this might be on styling but is more whether we want to lean towards using [static vs dynamic styles](https://emotion.sh/docs/best-practices#use-the-style-prop-for-dynamic-styles) with the css prop on support frontend in future. We could discuss briefly at the catch up!

GHaberis:
@i-hardy just letting you know @paul-daniel-dempsey removed his restriction on the current `singleToRecurringV3` test last week = so we could get learnings on all breakpoints for as long as possible (see https://github.com/guardian/support-frontend/pull/4805). The restriction will need to be restored once your test goes live, it may be easiest to add it in this PR?

paul-daniel-dempsey (approved):
All good too

PR #4794, opened 2023-02-22
Fix the Bug on error message for State / province dropdown (for US / AUS / CA pages)

LAKSHMIRPILLAI:

Fix the Bug on error message for State / province dropdown (for US / AUS / CA pages)

[**Trello Card**](https://trello.com/c/br9VGdKF/1100-bug-on-error-message-for-state-province-dropdown-for-us-aus-nz-pages)

## Why are you doing this?
Error messages is not dismissed when the users chooses an item from the dropdown for the State or Province dropdown on US / AUS and CA pages.

## Is this an AB test?
- [ ] Yes
- [x ] No

## Screenshots
**Before Change**
image
image

**After Change**
image
image

### DIFF
--- a/support-frontend/assets/helpers/redux/checkout/address/reducer.ts
+++ b/support-frontend/assets/helpers/redux/checkout/address/reducer.ts
@@ -68,6 +68,12 @@ function getAddressFieldsSlice(type: AddressType) {
 			setState(state, action: PayloadAction<string>) {
 				state.state = action.payload;
 				state.errors = removeError('state', state.errors);
+				const validationResult = addressFieldsSchema.safeParse(state);
+				if (!validationResult.success) {
+					state.errorObject = getSliceErrorsFromZodResult(
+						validationResult.error.format(),
+					);
+				}
 			},
 			setPostcode(state, action: PayloadAction<string>) {
 				state.postCode = action.payload;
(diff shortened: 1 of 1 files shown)


support-frontend/assets/helpers/redux/checkout/address/reducer.ts
```diff
@@ -68,6 +68,12 @@ function getAddressFieldsSlice(type: AddressType) {
 			setState(state, action: PayloadAction<string>) {
 				state.state = action.payload;
 				state.errors = removeError('state', state.errors);
+				const validationResult = addressFieldsSchema.safeParse(state);
+				if (!validationResult.success) {
+					state.errorObject = getSliceErrorsFromZodResult(
+						validationResult.error.format(),
+					);
+				}
```
i-hardy, line 76: Rather than re-validating every time the state is updated, we should just delete any errors from the last Zod validation pass, as the validation will run again when the user next tries to submit the form.

```suggestion
				delete state.errorObject?.state;
```

i-hardy (changes requested):

support-frontend/assets/helpers/redux/checkout/address/reducer.ts
LAKSHMIRPILLAI: Have updated the code as per suggestion

i-hardy (approved):
👍

PR #4796, opened 2023-02-22
Simplify basis of detecting when a user is a test user

i-hardy:

This simplifies the conditions around detecting when a user is using test credentials, to solve a bug that's arisen from merging user detection and setup logic between the contribution and subscriptions checkouts.


GHaberis (approved):
👍

PR #4810, opened 2023-02-28
Fail validation for create requests containing characters that would break Zuora

undergroundquizscene:
Trello Card: https://trello.com/c/47aV3LRb/1102-block-entering-emojis-in-the-delivery-instructions-field-in-newspaper-home-delivery-checkout

## Why

Some characters cause Zuora's API to break, or at least the `/subscribe` endpoint – it’ll return a 200 with an error in the response body. Validating the input first avoids weird behaviour in our step functions (like retrying for a day because the response from Zuora doesn't look like an error), and by validating on the client as well we can help users catch/correct errors more quickly.

## How

The characters that break Zuora are those whose UTF-8 representation contains more than 3 bytes. [According to Wikipedia](https://en.wikipedia.org/wiki/UTF-8#Encoding), that’s those code points above U+FFFF.

I stole this regex method [from Identity](https://github.com/guardian/identity/blob/77488775d4a9f4db34e3b795dbcd28e2b96be4a1/identity-api/src/main/scala/com/gu/identity/api/services/UserCleaningService.scala#L11), so it’s consistent with how things are done elsewhere.

I considered a couple of alternatives to the regex:

- testing for surrogate characters (as JVM strings are UTF-16):

  If we check for surrogate characters under the assumption that Java’s Strings are encoded as UTF-16, what happens if that representation changes?

- explicitly encoding to UTF-8 and testing for a byte whose digits start
  with 11110:

  Converting to UTF-8 seems like wasted effort, especially when the result of the conversion is thrown away, and we later re-encode when converting the fields to JSON.

I’m not sure if the regex example is any faster/better than converting to UTF-8: we avoid the re-encoding I reckon, but it seems we’re then relying on the regex compiler to work out it can do a linear scan of the input. I think that’s probably fine.

## Testing

To test (local/code/prod):

- Insert a character over U+FFFF (e.g. 𠀀 or 𐀀 or 😊) into a field (for example, delivery instructions or company name – first name will be caught by Identity) and checkout
- Confirm that with these changes, the client-side validation rejects these characters, whereas on current prod it only rejects the emoji
- Make a successful call to create via the browser with something in the delivery instructions field, copy it as a curl from the network tab
- Re-do the curl with one of the bad characters in the delivery instructions field, see that the call to create fails with the new changes (and succeeds without, except for when the bad character is an emoji)
- Remember to stop any looping support-workers step function executions that you create while testing – better to save the money and energy (though I’m sure it’s not much)

To test against Zuora directly, here’s a sample subscribe request:

<details>
<summary>Expand for JSON</summary>
<p>

```json
{
  "subscribes": [
    {
    "Account": {
      "AutoPay": false,
      "Batch": "Batch1",
      "BillCycleDay": 1,
      "Currency": "USD",
      "Name": "West ✅Corporation",
      "PaymentTerm": "Due Upon Receipt"
    },
    "BillToContact": {
      "Address1": "312 2nd Ave W",
      "City": "Seattle",
      "Country": "United States",
      "FirstName": "Sarah✅",
      "LastName": "Smith",
      "PostalCode": "98119",
      "State": "Washington",
      "WorkEmail": "sarah@example.com",
      "SpecialDeliveryInstructions__c" : "✅"
    },
    "PaymentMethod": {
      "CreditCardAddress1": "312 2nd Ave W",
      "CreditCardCity": "Seattle",
      "CreditCardCountry": "United States",
      "CreditCardExpirationMonth": 12,
      "CreditCardExpirationYear": 2023,
      "CreditCardHolderName": "Ms Sarah Smith",
      "CreditCardNumber": "4111111111111111",
      "CreditCardPostalCode": "98119",
      "CreditCardState": "Washington",
      "CreditCardType": "Visa",
      "Type": "CreditCard"
    },
    "SubscribeOptions": {
      "GenerateInvoice": true,
      "ProcessPayments": false
    },
    "SubscriptionData": {
      "RatePlanData": [
        {
        "RatePlan": {
          "ProductRatePlanId": "8ad08e018499b108018499f163ac32e3"
        },
        "RatePlanChargeData": [
          {
          "RatePlanCharge": {
            "ChargeModel": "Discount-Percentage",
            "DiscountPercentage": "6.75",
            "ProductRatePlanChargeId": "8ad08e018499b108018499f163e132ef"
          }
        }
        ]
      }
      ],
      "Subscription": {
        "AutoRenew": true,
        "ContractAcceptanceDate": "2019-02-15",
        "ContractEffectiveDate": "2019-02-15",
        "InitialTerm": 12,
        "RenewalTerm": 6,
        "ServiceActivationDate": "2019-02-15",
        "TermType": "TERMED"
      }
    }
  }
  ]
}
```

</p>
</details>

And here’s how you could call the Zuora dev api with it:

```sh
curl -i -XPOST -d @subscribe.json -H "apiAccessKeyId: your.zuora.account.email@guardian.co.uk.dev" -H "apiSecretAccessKey: $yourZuoraDevPassword" -H "Content-Type: application/json; charset=utf-8"  https://rest.apisandbox.zuora.com/v1/action/subscribe
```

### DIFF
--- a/support-frontend/assets/components/directDebit/directDebitProgressiveDisclosure/directDebitForm.tsx
+++ b/support-frontend/assets/components/directDebit/directDebitProgressiveDisclosure/directDebitForm.tsx
@@ -17,7 +17,7 @@ import {
 	setRecaptchaToken,
 } from 'helpers/redux/checkout/recaptcha/actions';
 import type { CheckoutState } from 'helpers/subscriptionsForms/subscriptionCheckoutReducer';
-import { nonSillyCharacters } from 'helpers/subscriptionsForms/validation';
+import { zuoraCompatibleString } from 'helpers/subscriptionsForms/validation';
 import Form from './components/form';
 import Playback from './components/playback';
 import type { DirectDebitFieldName } from './types';
@@ -72,7 +72,7 @@ const fieldValidationFunctions: {
 	[key in DirectDebitFieldName]: (fieldValue: string) => boolean;
 } = {
 	accountHolderName: (fieldValue) =>
-		!!/^\D+$/.exec(fieldValue) && nonSillyCharacters(fieldValue),
+		!!/^\D+$/.exec(fieldValue) && zuoraCompatibleString(fieldValue),
 	sortCodeString: (fieldValue) => !!/^\d{6}$/.exec(fieldValue),
 	accountNumber: (fieldValue) => !!/^\d{6,8}$/.exec(fieldValue),
 	accountHolderConfirmation: (fieldValue) => !!fieldValue,
--- a/support-frontend/assets/helpers/subscriptionsForms/rules.ts
+++ b/support-frontend/assets/helpers/subscriptionsForms/rules.ts
@@ -9,10 +9,10 @@ import type { FormField, FormFields } from './formFields';
 import {
 	formError,
 	nonEmptyString,
-	nonSillyCharacters,
 	notLongerThan,
 	notNull,
 	validate,
+	zuoraCompatibleString,
 } from './validation';
 import type { FormError } from './validation';
 
@@ -30,7 +30,7 @@ function applyPersonalDetailsRules(
 			error: formError('firstName', 'Please enter a first name.'),
 		},
 		{
-			rule: nonSillyCharacters(fields.firstName),
+			rule: zuoraCompatibleString(fields.firstName),
 			error: formError(
 				'firstName',
 				'Please use only letters, numbers and punctuation.',
@@ -45,7 +45,7 @@ function applyPersonalDetailsRules(
 			error: formError('lastName', 'Please enter a last name.'),
 		},
 		{
-			rule: nonSillyCharacters(fields.lastName),
+			rule: zuoraCompatibleString(fields.lastName),
 			error: formError(
 				'lastName',
 				'Please use only letters, numbers and punctuation.',
@@ -56,7 +56,7 @@ function applyPersonalDetailsRules(
 			error: formError('lastName', 'Last name is too long'),
 		},
 		{
-			rule: nonSillyCharacters(fields.telephone),
+			rule: zuoraCompatibleString(fields.telephone),
 			error: formError(
 				'telephone',
 				'Please use only letters, numbers and punctuation.',
@@ -94,7 +94,7 @@ function applyCheckoutRules(fields: FormFields): Array<FormError<FormField>> {
 			error: formError('firstName', 'Please enter a first name.'),
 		},
 		{
-			rule: nonSillyCharacters(fields.firstName),
+			rule: zuoraCompatibleString(fields.firstName),
 			error: formError(
 				'firstName',
 				'Please use only letters, numbers and punctuation.',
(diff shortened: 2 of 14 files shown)


undergroundquizscene:
[Block entering emojis. in the 'Delivery instructions' field in Newspaper Home delivery checkout](https://trello.com/c/47aV3LRb/1102-block-entering-emojis-in-the-delivery-instructions-field-in-newspaper-home-delivery-checkout)

support-frontend/assets/pages/paper-subscription-checkout/components/paperCheckoutForm.test.tsx
```diff
@@ -234,8 +234,8 @@ describe('Newspaper checkout form', () => {
 			expect(
 				screen.queryAllByText(
 					'Please use only letters, numbers and punctuation.',
-				),
-			).toBeTruthy();
+				).length,
+			).toBeGreaterThan(0);
```
undergroundquizscene, line 238: This test wasn’t actually working correctly, I’ve updated it to match the others (and verified that it fails when expected to)

support-frontend/assets/helpers/subscriptionsForms/rules.ts
```diff
@@ -265,6 +265,13 @@ function applyDeliveryRules(fields: FormFields): Array<FormError<FormField>> {
 				'Please indicate whether the billing address is the same as the delivery address',
 			),
 		},
+		{
+			rule: zuoraCompatibleString(fields.deliveryInstructions),
+			error: formError(
+				'deliveryInstructions',
+				'Please use only letters, numbers and punctuation.',
+			),
+		},
```
undergroundquizscene, line 274: We weren’t testing the delivery instructions for emojis. I haven’t conclusively reviewed all fields on the frontend to make sure we’re doing this validation for them, though I have attempted to on the backend.

support-frontend/test/utils/CheckoutValidationRulesTest.scala
```diff
@@ -370,6 +370,24 @@ class SimpleCheckoutFormValidationTest extends AnyFlatSpec with Matchers {
...
   }
 
+  it should "reject invalid characters in strings" in {
+    SimpleCheckoutFormValidation.noFourByteUtf8Characters("test1", "hello😊") shouldBe an[Invalid];
+    SimpleCheckoutFormValidation.noFourByteUtf8Characters("test2", "𐀀goodbye") shouldBe an[Invalid];
+    SimpleCheckoutFormValidation.noFourByteUtf8Characters("test3", "wa𠀀it") shouldBe an[Invalid];
+  }
+
+  it should "not reject valid characters in strings" in {
+    SimpleCheckoutFormValidation.noFourByteUtf8Characters("test1", "hello✅") shouldBe Valid;
```
undergroundquizscene, line 380: We were preventing this checkmark from being sent before (as it’s an emoji), but Zuora can actually handle it fine, as it’s under U+FFFF.

support-frontend/assets/components/directDebit/directDebitProgressiveDisclosure/directDebitForm.tsx
```diff
@@ -17,7 +17,7 @@ import {
 	setRecaptchaToken,
 } from 'helpers/redux/checkout/recaptcha/actions';
 import type { CheckoutState } from 'helpers/subscriptionsForms/subscriptionCheckoutReducer';
-import { nonSillyCharacters } from 'helpers/subscriptionsForms/validation';
+import { zuoraCompatibleString } from 'helpers/subscriptionsForms/validation';
```
undergroundquizscene, line 20: I’ve renamed this to be more accurate, especially because some of the characters that Zuora can’t handle (and that we are now blocking with this validation) are perfectly valid characters (e.g. to have in a name), we just don’t support them.

i-hardy (approved):
This is great, really good to have this issue properly understood, documented and tackled in the code 👍

PR #4839, opened 2023-03-10
Add request ID to acquisition-events-api logs

undergroundquizscene:
It can be hard to tell when looking at these logs which messages are from which request. The request ID is mentioned in the start-end-report logs (which I think AWS provides?), but not in our logging in between. Adding the request ID like this will allow us to filter to a specific request more easily, hopefully helping to debug issues.


rupertbates (approved):
👍

undergroundquizscene:
Huh, we actually haven’t had a Riff Raff build on this project (support:lambdas:acquisition-events-api) in prod since June?

undergroundquizscene:
So it says seen on prod now, but there’s still no riff raff build and the logs don’t include the extra data. I’ll try running a riff raff build manually.

undergroundquizscene:
That worked, I can now see the results on prod. I will look into why the deployment didn’t happen automatically.

PR #4849, opened 2023-03-14
stripe-intent: cdk

tomrf1:
...on second thoughts, why don't we just move this logic into support-frontend?


tomrf1:
> Hey Tom, just looking at this as another older PR we have kicking around and same questions as before- are you planning to pick this back up at some point and get it merged, or should we put this in the Contributions backlog?

I'll close this PR, I realised the time would be better spent moving this code into support-frontend

PR #4853, opened 2023-03-16
Single Less Prominent Contribute Checkout

paul-daniel-dempsey:
Build of the single less prominent designs. ab-test name 'singleLessProminent'

- Desktop web - designs below
- Mobile web - re-based on mobile optimisation variantA 
- Initial ONE_OFF only contribution types via URL query parameters or session storage need to be removed from the abtest group (not control or variant).

Variant Features/Assumptions->
1. Support Once entire component clickable (not just arrow).
2. Once clicked, Single tab remains visible (a refresh required to reset).

A ClickableCard component may be a more accessible way to contain multiple elements that can be mouse clicked (or the card itself can be tabbed onto and enter pressed to re-produce mouse-click). A working example of this is shown here (2nd solution) ->
https://css-tricks.com/creating-animated-clickable-cards-with-the-has-relational-pseudo-class/

Pre (Single,Monthly&Annual)->
| mobile <740px | dekstop >740px -|
|--------|--------|
|Screenshot 2023-03-20 at 09 46 00|Screenshot 2023-03-20 at 09 45 12|

Post (Annual&Monthly)->
| mobile <740px | dekstop >740px -|
|--------|--------|
| Screenshot 2023-03-20 at 09 46 19|Screenshot 2023-03-20 at 09 44 45|

[**Trello Card**](https://trello.com/c/cbOAHNBc/1167-make-single-less-prominent-build)

## Is this an AB test?
- [x] Yes
- [ ] No

https://support.thegulocal.com/uk/contribute#ab-singleLessProminent=variant
https://support.thegulocal.com/uk/contribute#ab-singleLessProminent=control


undergroundquizscene (commented):
I’m a bit confused by the state handling, and I’m not sure it’s correct. Have I got the following walkthrough right, and is it the intended behaviour?

# State Walkthrough

## Control

- `supportOnceDisplay` starts out as `false`, because `participations.singleLessProminent` will not be `"variant"`
- so `hideOneOff` is set to `false`, and the single tab won’t be hidden
- because `supportOnceDisplay` is `false`, the `CheckoutSupportOnce` element will be displayed to offer people the option to give a single contribution
- if the user clicks the `CheckoutSupportOnce` element, then `supportOnceDisplay` will be set to `true` because `participations.singleLessProminent` is not `"variant"`

## Variant

- `supportOnceDisplay` starts out as `true`, because `participations.singleLessProminent` is `"variant"`
- so `hideOneOff` is set to `true`, and the single tab is hidden
- because `supportOnceDisplay` is `true`, the `CheckoutSupportOnce` element won’t be displayed
- so the user has no option to change back to support once

## Other Bits

- Why do we need to check the variant/control status again in the callback?
- Does `onSupportOnceContainerClick` need to take an argument, if it already has access to the value of `supportOnceDisplay`?

  Also, if its argument is always `false`, this suggests the argument isn’t doing anything: you could remove it and replace it in the body with `false` and get the same behaviour. So then `(supportOnceDisplay && displaySupportOnce)` becomes `(supportOnceDisplay && false)`, and since `x && false` is always false for any x, you could replace it with just `(false)`. But then the whole expression can be reduced to just the first part, because `x || false` is always equal to x. So the expression is equivalent to just `participations.singleLessProminent !== 'variant'`, which definitely doesn’t seem right.

- Would it be possible to rename `supportOnceDisplay` to something like `shouldShowSingleTab`? I think that would make the state logic clearer.

support-frontend/assets/components/checkoutSupportOnce/checkoutSupportOnce.tsx
```diff
@@ -0,0 +1,98 @@
...
+export function CheckoutSupportOnce({
+	contributionType,
+	supportOnceDisplay,
+	supportOnceTitle,
+	supportOnceParagraph,
+	onSupportOnceClick,
+}: CheckoutSupportOnceProps): JSX.Element | null {
+	if (contributionType !== 'ONE_OFF' && supportOnceDisplay) {
+		return (
+			<Box onClick={onSupportOnceClick} cssOverrides={box}>
```
GHaberis, line 83: @paul-daniel-dempsey I think we need to consider accessibility when making a `div` a clickable element. As it is it's essentially just a layout element (a `div`) with a click handler, I'm pretty this would be an issue for users using screen readers or keyboard navigation. This article covers this well: https://tech.boxed.com/accessible-ui-stop-making-clickable-divs-264115938716. They're suggestion is to use an HTML `button` element rather than a `div`.

support-frontend/assets/components/checkoutSupportOnce/checkoutSupportOnceButton.tsx
```diff
@@ -0,0 +1,36 @@
...
+	Button,
+	SvgChevronRightSingle,
+} from '@guardian/source-react-components';
+
+const button = css`
+	color: #606060;
+	background-color: transparent;
+	&:hover {
+		background-color: transparent;
+	}
```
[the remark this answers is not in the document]
undergroundquizscene: Ah yes, that makes sense!

support-frontend/assets/components/paymentFrequencyTabs/paymentFrequencyTabsContainer.tsx
```diff
@@ -60,6 +62,10 @@ export function PaymentFrequencyTabsContainer({
 		},
 	);
 
+	if (hideOneOff) {
+		delete tabs[0];
```
GHaberis, line 66: We shouldn't use the `delete` keyword to remove elements from the `tabs` array, this [article](https://bytearcher.com/articles/how-to-delete-value-from-array/#:~:text=Whatever%20you%20do%2C%20don't,t%20update%20the%20length%20property.) provides a good summary of why and alternatives...

> Whatever you do, don't use delete to remove an item from an array. JavaScript language specifies that arrays are sparse, i.e., they can have holes in them. Using delete creates these kinds of holes. It removes an item from the array, but it doesn't update the length property. This leaves the array in a funny state that is best avoided.

I actually think you could put a conditional in the `map` above `if (hideOneOff && contributionType === 'ONE_OFF') { return; }` This would leave `undefined` in the `tabs` array instead of the one off tab's props if `hideOneOff` is `true`, so you'll need to chain a `.filter(item => item != undefined) to the end of the `map` block to get remove this from `tabs`.

support-frontend/assets/components/checkoutSupportOnce/checkoutSupportOnce.tsx
```diff
@@ -0,0 +1,98 @@
...
+				<div css={container}>
+					<div css={top}>
+						<h2 css={heading(brand[500])}>{supportOnceTitle}</h2>
+					</div>
+					<div css={bottom}>
+						<p css={para}>{supportOnceParagraph}</p>
+						<CheckoutSupportOnceButton onClick={onSupportOnceClick} />
```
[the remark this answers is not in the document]
GHaberis: This https://github.com/guardian/support-frontend/pull/4853#discussion_r1142418069 comment might negate this issue if we choose to make the entire container a `<button>` element.

support-frontend/assets/pages/supporter-plus-landing/supporterPlusLanding.tsx
```diff
@@ -150,6 +163,24 @@ export function SupporterPlusLandingPage({
 	};
 	const heading = <LandingPageHeading />;
 
+	const participations = initAbTests(
```
GHaberis, line 166: You shouldn't need to call `initAbTests` here, the user's AB test participations have already been allocated and added to the Redux state. You should be able to use the `isUserInAbVariant` function from 'helpers/redux/commonState/selectors' to determine if a user in a test cohort rather than looking up `participations`. EG.

Replace...

```
const [supportOnceDisplay, setSupportOnceDisplay] = useState(
		participations.singleLessProminent === 'variant',
	);
```

With...

```
const [supportOnceDisplay, setSupportOnceDisplay] = useState(
		isUserInAbVariant('singleLessProminent', 'variant')
	);
```

support-frontend/assets/pages/supporter-plus-landing/supporterPlusLanding.tsx
```diff
@@ -150,6 +163,24 @@ export function SupporterPlusLandingPage({
...
+		participations.singleLessProminent === 'variant',
+	);
+
+	const dispatch = useContributionsDispatch();
+	function onSupportOnceContainerClick(displaySupportOnce: boolean) {
+		setSupportOnceDisplay(
+			participations.singleLessProminent !== 'variant' ||
+				(supportOnceDisplay && displaySupportOnce),
+		);
+		dispatch(setProductType('ONE_OFF'));
```
GHaberis, line 181: I think this can be simplified. 

Firstly, I'm not sure it's possible for `onSupportOnceContainerClick` to be called where `displaySupportOnce` will be `true`? I can't see a scenario where a click event on this element would lead us to `setSupportOnceDisplay` to `false`. Therefore I don't think then we need this `displaySupportOnce` parameter.

I'm also not sure why we're checking `participations.singleLessProminent !== 'variant'` here? When will a user who is not in the `variant` of `singleLessProminent` be able to click on the `SupportOnceContainer`? It should only be visible to users in the  the `variant` of `singleLessProminent`?

I think we could just be calling `setSupportOnceDisplay(false)` instead of...

```
setSupportOnceDisplay(
  participations.singleLessProminent !== 'variant' ||
    (supportOnceDisplay && displaySupportOnce),
);
```

support-frontend/assets/pages/supporter-plus-landing/supporterPlusLanding.tsx
```diff
@@ -222,8 +253,15 @@ export function SupporterPlusLandingPage({
 						<Box
 							cssOverrides={optimisedMobileLayout ? shorterBoxMargin : css``}
 						>
-							<AmountAndBenefits />
+							<AmountAndBenefits hideOneOff={supportOnceDisplay} />
 						</Box>
+						<CheckoutSupportOnceContainer
```
GHaberis, line 258: Instead of passing `supportOnceDisplay` all the way down to the component tree to make the decision whether to render it or not we could do so here, eg.

```
(supportOnceDisplay && <CheckoutSupportOnceContainer>)
```

This simplifies the prop signatures for `CheckoutSupportOnceContainer` and `CheckoutSupportOnce`, and it means we're executing less code needlessly as we're making the further up the component tree without having to needlessly run code for `CheckoutSupportOnceContainer` and `CheckoutSupportOnce` if `supportOnceDisplay` is `false`.

support-frontend/assets/components/paymentFrequencyTabs/paymentFrequencyTabsContainer.tsx
paul-daniel-dempsey: Yes understand, no delete as leaves holes of space inside arrays, will use filtering undefined going forwards. In my  '[abtest check for multiple variants amended]' commit, I had to type assert the filter even though I defined a user defined type guard, hopefully this change is thorough enough now.

support-frontend/assets/pages/supporter-plus-landing/supporterPlusLanding.tsx
paul-daniel-dempsey: Ok, quicker from redux store than participations

support-frontend/assets/pages/supporter-plus-landing/supporterPlusLanding.tsx
paul-daniel-dempsey: Ahhh, if you get the SupportOnceContainerClick event you must have the SupportOnce component open therefore no need at all to check for the variant. Once depressed once simply set to false why pass arguments through. Ok changes made.

support-frontend/assets/pages/supporter-plus-landing/supporterPlusLanding.tsx
paul-daniel-dempsey: changes made

support-frontend/assets/pages/supporter-plus-landing/supporterPlusLanding.tsx
paul-daniel-dempsey: Yes should spotted that, have seen this done many times reducing need for unnecessary props. Have corrected/simplified this as suggested.

support-frontend/assets/components/checkoutSupportOnce/checkoutSupportOnce.tsx
paul-daniel-dempsey: Hmmm, really struggling to get a button to follow the formatting of the original div. Buttons do not to appear to follow identical css rules as div's am finding. Is a label be an acceptable replacement as this does and its used in Choice Cards for example? Have put label version in and will continue attempting to figure out the button method. My CSS needs more practice!

support-frontend/assets/components/checkoutSupportOnce/checkoutSupportOnce.tsx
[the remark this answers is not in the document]
paul-daniel-dempsey: See replacement  comment below

support-frontend/assets/components/checkoutSupportOnce/checkoutSupportOnce.tsx
GHaberis: @paul-daniel-dempsey I'd be happy to pair on this if you'd like?

i-hardy: My two cents on this:

- In the future we should resist this kind of pattern at the design stage. 'Clickable cards' are always a problem, and if we do not have a standardised UI pattern in Source for this kind of thing (or we are not creating the thing that designers want to make the standardised pattern) we should strenuously avoid doing it.
- Given that this element is sending users to another 'resource' (ie. the version of the page with single contributions available), the semantically correct element here is a link, not a button (buttons are for changing something or submitting information on the resource you're already on). There's a good article from CSS Tricks here on a few different ways to make a 'clickable card' with an internal `<a>` tag: https://css-tricks.com/creating-animated-clickable-cards-with-the-has-relational-pseudo-class/. However note that none of them are particularly great- hence why we should try and avoid this pattern if possible!

support-frontend/assets/components/checkoutSupportOnce/checkoutSupportOnceContainer.tsx
```diff
@@ -0,0 +1,41 @@
...
+	function onSupportOnceClick() {
+		onSupportOnceContainerClick();
+	}
+
+	return renderSupportOnce({
+		contributionType,
+		supportOnceTitle: title,
+		supportOnceParagraph: paragraph,
+		onSupportOnceClick,
+	});
```
undergroundquizscene, line 40: Could you remove `onSupportOnceClick`? I.e. would the following work?

```suggestion
	return renderSupportOnce({
		contributionType,
		supportOnceTitle: title,
		supportOnceParagraph: paragraph,
		onSupportOnceContainerClick,
	});
```

support-frontend/assets/components/checkoutSupportOnce/checkoutSupportOnceButton.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+export function CheckoutSupportOnceButton({
+	onClick,
+}: CheckoutSupportOnceButtonProps): JSX.Element {
+	return (
+		<Button
+			onClick={onClick}
+			icon={<SvgChevronRightSingle size="xsmall" />}
+			size="small"
+			hideLabel
```
i-hardy, line 31: What label is being hidden here? A button should _always_ have some text content, even if it's not visible, for screen reader users.

support-frontend/assets/components/checkoutSupportOnce/checkoutSupportOnce.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+export function CheckoutSupportOnce({
+	contributionType,
+	supportOnceTitle,
+	supportOnceParagraph,
+	onSupportOnceClick,
+}: CheckoutSupportOnceProps): JSX.Element | null {
+	if (contributionType !== 'ONE_OFF') {
+		return (
+			<a onClick={onSupportOnceClick} css={[mainStyles, box]}>
```
i-hardy, line 99: Having a `<button>` inside an `<a>`, which both have the same click handler, is a huge anti-pattern. Do not do this. I'm not sure that it's a good idea to contain the entire component inside an anchor- I would want to run through it with VoiceOver to check but this is potentially going to make this link confusing and/or meaningless to anyone using the rotor menu to look for links on the page, and confusing to interact with for anyone using tab navigation.

The CSS Tricks article I linked on the previous PR suggests using an anchor (or button) _inside_ the 'clickable card', and then using some hacky positioning to extend an interactive area across the whole containing component. Please have a read of it in full: https://css-tricks.com/creating-animated-clickable-cards-with-the-has-relational-pseudo-class/ While I still don't think we should be doing this at all, I think that would be the best approach if we are committed to this component- it keeps the HTML structure as sensible as possible for assistive technologies while still allowing the desired interaction.

support-frontend/assets/helpers/redux/commonState/selectors.ts
```diff
@@ -55,3 +55,13 @@ export function isUserInAbVariant(abTestName: string, variantName: string) {
...
 	};
 }
+
+export function isUserInAbVariants(abTestNames: string[], variantName: string) {
+	return function getAbTestStatus(state: ContributionsState): boolean {
+		const participations = state.common.abParticipations;
+		const foundAbTests = abTestNames.filter(
+			(abTestName) => participations[abTestName] === variantName,
+		);
+		return foundAbTests.length > 0;
```
i-hardy, line 65:
```suggestion
		const matchingAbTest = abTestNames.find(
			(abTestName) => participations[abTestName] === variantName,
		);
		return !!matchingAbTest
```
It doesn't really matter with such tiny arrays, but if you have a condition where 'as long as at least one of these array items matches the condition, go ahead', `find` is more efficient because it will return as soon as it finds the first match rather than continuing to loop through the whole array.

I'd also suggest tweaking the function name from `isUserInAbVariants` to something that more clearly communicates that we're checking if the user is in `variantName` for _any one of_ the passed tests; I feel like the existing name could also be taken for a function that's checking if the user is in _all_ of the passed tests.

i-hardy (changes requested):
I'm afraid the 'clickable card' component still needs a lot of work, and to be thoroughly checked to make sure that whatever solution we come up with is accessible. There are links in the PR template on how to get started testing with a screen reader and keyboard-only navigation- this is a critically important aspect of building new interactive controls.

support-frontend/assets/components/checkoutBox/checkoutBox.tsx
```diff
@@ -22,12 +22,16 @@ const mainStyles = css`
...
 }
 
 export function Box(props: BoxProps): JSX.Element {
 	const TagName = props.tag ?? 'section';
 	return (
-		<TagName css={[mainStyles, props.cssOverrides ?? '']}>
+		<TagName
+			css={[mainStyles, props.cssOverrides ?? '']}
+			onClick={props.onClick}
+		>
```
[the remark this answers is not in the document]
paul-daniel-dempsey: Yes, not required and removed. Was left from the first of the 3-ways looked at as clickable components.

support-frontend/assets/helpers/redux/commonState/selectors.ts
paul-daniel-dempsey: Done

support-frontend/assets/components/checkoutSupportOnce/checkoutSupportOnceContainer.tsx
paul-daniel-dempsey: This is how the SupportOnce component indicates to the container its been clicked, its passed up from there to SupporterPlusLanding to notify redux to re-display the single tab.  From this point it acts as 3 frequency tabs (single/monthly/annual) as usual. It's needed I believe.

support-frontend/assets/components/checkoutSupportOnce/checkoutSupportOnceButton.tsx
paul-daniel-dempsey: hideLabel is incorrect, it's also not required to be clickable if inside a clickable component. Have renamed to 'CheckoutSupportOnceChevronRight' component which contains just a div and chevron right styled. Possibly does not need to be a component?

support-frontend/assets/components/checkoutSupportOnce/checkoutSupportOnce.tsx
paul-daniel-dempsey: re: <button> inside an <a>  am sorry I didn't know it was a huge no-no. Am compiling a list of 'don't do' rules in my notes for future reference, this is in there. Ok, the button has been removed from the component inside the anchor at least now.

Have gone through the article and think the approach is fundamentally different from the nudge. Agree there is much more complexity in building this the right way as a standalone clickable component and testing it. 

From reading this closer, would guess a component useEffect would be needed so the JS in the article can convert mouse pressing within any hypertext labeled 'card's to 'a' click (which is inside the card). If working, would the approach still be acceptable as its mentioned as being a bit hacky?

I don't think I can complete this alone, then test it and in time before the end of the week. 

Perhaps I close this card as it stands, without putting it into prod, and we refer to it in future if needed? Then create a new 'investigate clickable component' trello card to do as suggested above (and test plus decide if acceptable). Followed by another trello card to integrate it into supportJustOnce (perhaps referring to parts of this PR if necessary). Would that work?

support-frontend/assets/components/checkoutSupportOnce/checkoutSupportOnceContainer.tsx
paul-daniel-dempsey: Sorry I see, removed click event, and passed through now as 'onSupportOnceClick'

support-frontend/assets/components/checkoutSupportOnce/checkoutSupportOnce.tsx
i-hardy: I think rather than trying to compile a set of rules, the most helpful thing is usually to take a step back from being deep in the weeds of the problem and have a think about the actual HTML structure you're creating, outside of the design and the business needs. Good semantic HTML really, really does get you most of the way there with accessibility (which is one of the reasons why if we're asked to build something that doesn't have an obvious HTML implementation it should give us pause).

In this case, try thinking about what an anchor tag and a button actually _are_, and what we're trying to communicate by using them. If for some reason our styles failed to work on this page, what would it look like? Would that kind of interface look or feel odd to use? If you were reading that raw HTML, would it seem a bit odd to you?

Don't feel like you have to rush, or that you have to jump immediately to follow mine or anyone else's suggestions without fully comprehending the problem for yourself. It's _always_ better to take a little bit more time to have more confidence in your code. If we have to push the deadline a bit then so be it.

paul-daniel-dempsey: Ok, have created a clickable card component that produces a click event upon mouse pressing any html element within a div labelled 'card' (by listening & timing mouseup/mousedown events). IMoe generic and its based upon  https://css-tricks.com/creating-animated-clickable-cards-with-the-has-relational-pseudo-class/ Obviously subject to tomorrows meeting over accesibility and whether we should do this at all but at least its there now.

support-frontend/assets/components/cardClickable/cardClickable.tsx
```diff
@@ -0,0 +1,182 @@
...
+	focus,
+	from,
+	headline,
+	neutral,
+	space,
+	textSans,
+	until,
+} from '@guardian/source-foundations';
+import { SvgChevronRightSingle } from '@guardian/source-react-components';
+import { useEffect } from 'preact/hooks';
```
marialani, line 13:
```suggestion
import { useEffect } from 'react';
```

support-frontend/assets/components/cardClickable/cardClickable.tsx
```diff
@@ -0,0 +1,182 @@
...
+	}
+`;
+const chevronCss = css`
+	color: #606060;
+	background-color: transparent;
+	&:hover {
+		background-color: transparent;
+	}
+	& svg {
+		width: 24px;
```
marialani, line 83:
```suggestion
		width: ${space[6]}px;
```

support-frontend/assets/components/cardClickable/cardClickable.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+export function CardClickable({
+	cardTitle,
+	cardParagraph,
+	onCardClick,
+}: CardClickableProps): JSX.Element | null {
+	const cardClickableProp = { onCardClick };
+
+	useEffect(() => {
+		const cardClick = document.querySelector('.card');
```
marialani, line 102: This var name slightly confused me on first look

```suggestion
		const cardElement = document.querySelector('.card');
```

support-frontend/assets/components/cardClickable/cardClickable.tsx
```diff
@@ -0,0 +1,182 @@
...
+				// mousedown: Log timestamp
+				element.addEventListener('mousedown', (event) => {
+					const card = (event.target as HTMLElement).closest('.card');
+					if (card) {
+						const mousedownTime = new Date().getTime();
+						card.setAttribute('time-mousedown', mousedownTime.toString());
+					}
+				});
+
+				// mouseup: Determine whether to raise click event
```
marialani, line 118: Could we add a brief text explanation the logic you have here (aka the code requirements of this clickable card) to the PR description? Think that background info would be really useful in future

support-frontend/assets/components/cardClickable/cardClickableContainer.tsx
```diff
@@ -0,0 +1,34 @@
...
+	const currencyGlyph = glyph(detect(countryGroupId));
+	const minAmount = config[countryGroupId]['ONE_OFF'].min;
+
+	const [cardTitle, cardParagraph] = [
+		`Support us just once`,
+		`We welcome support of any size, any time, whether you choose to give ${
+			currencyGlyph + minAmount.toString()
+		} or much more`,
+	];
+
```
marialani, line 28: Small note but splitting these into separate variables is slightly easier on the eye

support-frontend/assets/pages/supporter-plus-landing/supporterPlusLanding.tsx
```diff
@@ -12,8 +12,11 @@ import {
 	FooterLinks,
 	FooterWithContents,
 } from '@guardian/source-react-components-development-kitchen';
+import { useState } from 'preact/hooks';
```
marialani, line 15:
```suggestion
import { useState } from 'react';
```

support-frontend/assets/components/cardClickable/cardClickable.tsx
paul-daniel-dempsey: Done

support-frontend/assets/components/cardClickable/cardClickable.tsx
paul-daniel-dempsey: Done

support-frontend/assets/components/cardClickable/cardClickable.tsx
paul-daniel-dempsey: Yes, done, much better named

support-frontend/assets/components/cardClickable/cardClickable.tsx
paul-daniel-dempsey: Yes, done, have updated the PR with background information on ClickableCards for reference and placed a small comment inside the code here too.

support-frontend/assets/components/cardClickable/cardClickableContainer.tsx
paul-daniel-dempsey: Ok, will follow in future, done

support-frontend/assets/pages/supporter-plus-landing/supporterPlusLanding.tsx
paul-daniel-dempsey: Done

support-frontend/assets/components/cardClickable/cardClickable.tsx
marialani: The code here details a few different events (e.g. focus, keypress, mouseup etc) and what should happen when each of these events occur. This bit of the code looks to be the bit that you'd grappled with in terms of accessibility? If time allows, a write up of what events you had to hand-roll and why for this clickable card would be very helpful

marialani:
[Single less prominent build](https://trello.com/c/cbOAHNBc/1167-single-less-prominent-build)

support-frontend/assets/components/cardClickable/cardClickable.tsx
```diff
@@ -0,0 +1,188 @@
...
+
+export function CardClickable({
+	cardTitle,
+	cardParagraph,
+	onCardClick,
+}: CardClickableProps): JSX.Element | null {
+	const cardClickableProp = { onCardClick };
+
+	useEffect(() => {
+		const cardElement = document.querySelector('.card');
```
i-hardy, line 102: I really hate to say this because I can tell this has been a lot of work, but I think you have seriously over-engineered this. The moment you find yourself using browser selection APIs inside a React component, such as `document.querySelector`, it should give you serious pause because you're stepping outside React's rendering behaviour. If you absolutely _have_ to access a DOM node directly in React code you should use [refs](https://react.dev/learn/manipulating-the-dom-with-refs), but this should be avoided if at all possible. The point of using React is that we don't have to think about rooting around in the DOM manually- we use JSX to describe the DOM that we want to create, and React handles actually rendering it and keeping it up to date.

I would honestly favour a pure CSS approach- using an actuall `<a>` tag with succinct but descriptive text, (in this case probably using the title) and applying an `::after` pseudo-element to it that covers the whole 'card' component. This is the approach used on dotcom fronts and gov.uk. It does mean the text in the card is not selectable (which is why I just strongly dislike the 'clickable card' as a component type in the first place), but it's _much_ less code to document and maintain and means we are actually using proper semantic HTML rather than attaching event listeners to elements like `<div>`s that have no semantic meaning.

When it comes to web accessibility, proper semantic HTML should always be the first thing you reach for, and anything extra that we need to add with ARIA attributes, extra JS, etc should be kept to the absolute minimum that we can feasibly get away with. Browsers are already built to communicate a page's structure to screenreaders and other kinds of assistive technology, and we should strive to rely as much as possible on what the browser can do for us for free, rather than trying to create DIY solutions that we then have to maintain ourselves.

support-frontend/assets/components/paymentFrequencyTabs/paymentFrequencyTabsContainer.tsx
```diff
@@ -34,6 +36,14 @@ export function PaymentFrequencyTabsContainer({
 	);
 	const productType = useContributionsSelector(getContributionType);
 
+	// Force to monthly if hideOneOff, OneOff and refresh pressed
+	if (
+		hideOneOff &&
+		useContributionsSelector(getContributionType) === 'ONE_OFF'
+	) {
```
i-hardy, line 43:
```suggestion
 const contributionType = useContributionsSelector(getContributionType);
 
	if (
		hideOneOff &&
		contributionType === 'ONE_OFF'
	) {
```
The right-hand side of a logical operator in JS (ie. `&&`, `||`) is lazily evaluated- this means that if the left hand side is falsey for `&&` or truthy for `||`, any code on the right hand side will not be run. Try running this in your console to see:

```javascript
false && console.log("you won't see this");
true || console.log('or this');
```

Because of the way React hooks work it's really important that they get called _every single time_ a component renders, so that React can associate the correct bits of state with the correct components. We should always use hooks at the top level, and then use their results in conditions if we need to.

support-frontend/assets/components/paymentFrequencyTabs/paymentFrequencyTabsContainer.tsx
```diff
@@ -50,15 +60,23 @@ export function PaymentFrequencyTabsContainer({
...
+				return;
+			} else {
+				return {
+					id: contributionType,
+					labelText: toHumanReadableContributionType(contributionType),
+					selected: contributionType === productType,
+				};
+			}
+		})
+		.filter(isTab) as TabProps[];
```
i-hardy, line 79: Just curious- what's the TS error here if you remove `as TabProps[]`? It's better to avoid using type assertions if we can.

i-hardy (dismissed):

support-frontend/assets/components/paymentFrequencyTabs/paymentFrequencyTabsContainer.tsx
paul-daniel-dempsey: Ok, don't put React  hook call into a conditional where it may not be called. Change applied.

support-frontend/assets/components/paymentFrequencyTabs/paymentFrequencyTabsContainer.tsx
paul-daniel-dempsey: Will try to remove as many as possible going forwards. Removed the need for type assertion by changing the return type of line72 function 'toHumanReadableContributionType' to string (from 'Single' | 'Monthly' | 'Annual' this did'nt associate with a string)

paul-daniel-dempsey: Change applied

i-hardy: I'm not sure that that's the correct solution- there's no type conflict at the moment between using a function that returns `'Single' | 'Monthly' | 'Annual'` to create the value of a field of type `string`, and generally it's better to avoid making types _less_ specific.

You may get a more informative error message if you move the type guard function inline, like this:

```typescript
.filter((maybeTab): maybeTab is TabProps => !!maybeTab);
```

paul-daniel-dempsey: Placing the filter first in the EDIT avoided the type conflict in the error guard, have implemented like above and amended the function back to be more type specific ('Single' | 'Monthly' | 'Annual') again. Thanks.

support-frontend/assets/components/cardClickable/cardClickable.tsx
paul-daniel-dempsey: New approach taken for clickable card now.

support-frontend/assets/pages/supporter-plus-landing/supporterPlusLanding.tsx
```diff
@@ -149,6 +160,27 @@ export function SupporterPlusLandingPage({
...
+			isUserInAbVariant('singleLessProminent', 'variant'),
+		),
+	);
+
+	const dispatch = useContributionsDispatch();
+	function onCardClick() {
+		setSupportOnceDisplay(false);
+
+		const paymentMethodToSelect = getPaymentMethodToSelect(
+			contributionType,
```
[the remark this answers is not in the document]
paul-daniel-dempsey: Understand, I should of made this match the following (underneath) dispatch line which selects 'ONE_OFF'.  Have tested, can see it's Stripe with 'ONE_OFF' and 'DIRECT DEBIT' if incorrectly left as 'contributionType'. Corrected as suggested, thanks.

support-frontend/assets/pages/supporter-plus-landing/supporterPlusLanding.tsx
```diff
@@ -149,6 +160,27 @@ export function SupporterPlusLandingPage({
 	};
 	const heading = <LandingPageHeading />;
 
+	const [supportOnceDisplay, setSupportOnceDisplay] = useState(
+		useContributionsSelector(
+			isUserInAbVariant('singleLessProminent', 'variant'),
+		),
+	);
```
GHaberis, line 167: @paul-daniel-dempsey I think an edge case you'll need to account for is if a user has an initial `contributionType` of 'ONE_OFF'. This can happen in 2 scenarios: 

1) If they're referred to the page after clicking Single in the choice cards in the Epic or Banner then a query string param `?selected-contribution-type=ONE_OFF` will be appended to the URL, if this query param is present them initial the `contributionType` in our Redux store is 'ONE_OFF'.

2) If they're a returning to the page and they were on the Single tab before we'll have saved this in session storage (`selectedContributionType`) and we'll set the initial `contributionType` in our Redux store based on that value.

So, if the initial `contributionType`  is `ONE_OFF` I think we'll want to set `supportOnceDisplay` to `false`, we could make that check when we define it like so ...

```
const [supportOnceDisplay, setSupportOnceDisplay] = useState(() => contributionType === 'ONE_OFF' ? false : useContributionsSelector(
		isUserInAbVariant('singleLessProminent', 'variant'),
	));
```

You can test this locally by appending the `?selected-contribution-type=ONE_OFF` query param to the URL.

https://support.thegulocal.com/uk/contribute?selected-contribution-type=ONE_OFF#ab-singleLessProminent=variant

paul-daniel-dempsey: Ah yes, I can see the  ?selected-contribution-type=ONE_OFF Query Param would force you onto Monthly otherwise. Not what is need. Will try to remember query parameters can change your starting point, can see there is an amountFromURL too but thats not relevant here. Thanks, changed as recommended.

support-frontend/assets/helpers/redux/commonState/selectors.ts
```diff
@@ -55,3 +55,16 @@ export function isUserInAbVariant(abTestName: string, variantName: string) {
 		return participations[abTestName] === variantName;
 	};
 }
+
+export function isUserInAnyAbVariant(
```
GHaberis, line 59: I believe we agreed this `isUserInAnyAbVariant` function can be removed as it's not used anywhere in the code?

support-frontend/assets/helpers/abTests/abtestDefinitions.ts
```diff
@@ -100,4 +101,29 @@ export const tests: Tests = {
...
+				offset: 0,
+				size: 1,
+			},
+		},
+		isActive: true,
+		referrerControlled: false,
+		targetPage: pageUrlRegexes.contributions.allLandingPagesAndThankyouPages,
+		seed: 11,
+		canRun: () => {
+			// If URL query string parameter contains annual or monthly contributionType apply abTest
```
GHaberis, line 124: The comment isn't quite right, If `contributionTypesFromUrl` is not ONE_OFF we apply the abTest, so it could be `null` as well as "annual" or "monthly" and we'd apply the abTest.

GHaberis (approved):
Looks good 👍

support-frontend/assets/components/paymentFrequencyTabs/paymentFrequencyTabsContainer.tsx
```diff
@@ -34,6 +33,12 @@ export function PaymentFrequencyTabsContainer({
 	);
 	const productType = useContributionsSelector(getContributionType);
 
+	// Force to monthly if hideOneOff, OneOff and refresh pressed
+	const contributionType = useContributionsSelector(getContributionType);
+	if (hideOneOff && contributionType === 'ONE_OFF') {
```
GHaberis, line 38: @paul-daniel-dempsey I'm finding it hard to work out when this conditional will be `true`? Is this `if` block needed?

support-frontend/assets/helpers/abTests/abtestDefinitions.ts
paul-daniel-dempsey: Ok, comments should follow the logic, done.

PR #4861, opened 2023-03-23
Remove all gaEvent calls and unused GTM triggers

GHaberis:
We have tracking calls to send GA Events and Google Tag Manager (GTM) [Triggers](https://support.google.com/tagmanager/answer/7679316?hl=en) in the support site codebase, these tracking calls are fired on various events such as page loads, component clicks, component views and conversions. Following an investigation we've seen the calls are inconsistent between pages, are sporadically used and are not working or needed. 

In this PR I've made the following changes...

- In `assets/helpers/tracking/googleTagManager.ts` I've deleted the `gaEvent` function and the GTM triggers ~`successfulConversion`~ and `appStoreCtaClick` plus any associated types/exports if unused elsewhere. I've also deleted `gaPropertyId` from  `assets/helpers/tracking/googleTagManager.ts` as wasn't used anywhere.
- Removed all `gaEvent` imports and function calls from the codebase.
- Removed all `optimizeId` logic from `assets/helpers/abTests/abtest.ts`, as we no longer used Optimize and it is being [subset](https://support.google.com/optimize/answer/12979939?hl=en) by Google.

[**Trello Card**](https://trello.com/c/XGLfhKFs/1183-remove-ga-events-and-associated-code-from-support-site)

## Why are you doing this?
Simplify code and less network requests on page

### DIFF
--- a/support-frontend/yarn.lock
+++ b/support-frontend/yarn.lock
@@ -5707,11 +5707,6 @@ caniuse-lite@^1.0.30001317:
   resolved "https://registry.yarnpkg.com/caniuse-lite/-/caniuse-lite-1.0.30001466.tgz#c1e6197c540392e09709ecaa9e3e403428c53375"
   integrity sha512-ewtFBSfWjEmxUgNBSZItFSmVtvk9zkwkl1OfRZlKA8slltRN+/C/tuGVrF9styXkN36Yu3+SeJ1qkXxDEyNZ5w==
 
-caniuse-lite@^1.0.30001359:
-  version "1.0.30001469"
-  resolved "https://registry.yarnpkg.com/caniuse-lite/-/caniuse-lite-1.0.30001469.tgz#3dd505430c8522fdc9f94b4a19518e330f5c945a"
-  integrity sha512-Rcp7221ScNqQPP3W+lVOYDyjdR6dC+neEQCttoNr5bAyz54AboB4iwpnWgyi8P4YUsPybVzT4LgWiBbI3drL4g==
-
 capture-exit@^2.0.0:
   version "2.0.0"
   resolved "https://registry.yarnpkg.com/capture-exit/-/capture-exit-2.0.0.tgz#fb953bfaebeb781f62898239dabb426d08a509a4"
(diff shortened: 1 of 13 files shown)


GHaberis:
> Will this not break the GTM remarketing tags?

No, we're still firing the trigger (`DataLayerReady`) which drops the AdWords/remarketing tags. The `successfulConversion` trigger isn't fired at all from the Subscriptions checkouts and even though we do fire it from the Support checkout there's an exception rule on the trigger handler in GTM that means no actions are taken by GTM (tags dropped etc.) when this event is triggered.  The `appStoreCtaClick` trigger was not being used/called anywhere.

GHaberis:
> > Will this not break the GTM remarketing tags?
> 
> No it shouldn't as we're still firing the trigger (`DataLayerReady`) which drops the AdWords/remarketing tags. The `successfulConversion` trigger isn't currently fired at all from the Subscriptions checkouts and even though we do fire it from the Support checkout there's an exception rule on the trigger handler in GTM that means no actions are taken by GTM (tags dropped etc.) when this event is triggered. The `appStoreCtaClick` trigger was not being used/called anywhere.
> 
> @rupertbates this ticket links to further info: [Trello](https://trello.com/c/FvLP1Du1/1169-spike-google-tag-manager-google-analytics-review-on-support-site)

@rupertbates OK my original change would have removed the AdWords Conversion Tracking which is required! I was running the change past Marketing before releasing and they told me they report on AdWords Conversion Tracking for the Paper and Guardian Weekly checkouts, I was surprised given  my initial tests on the print checkouts hadn't shown the `SuccesfulConversion` trigger being fired. So I ran some further tests on `PROD` using GTM's debugger and I did see the `SuccesfulConversion` trigger firing on the print checkouts on conversion. I've therefore restored this trigger!

The other changes still stand. I've tested these changes with GTM's debugger and all the AdWords/Remarketing tags are dropped as expected.

support-frontend/yarn.lock
undergroundquizscene, line 1: These yarn.lock updates are a little surprising: should they be included? There aren’t any package.json changes, are there?

undergroundquizscene (approved):
None of the removed tracking is worth porting to ophan tracking, is it? E.g. the polyfill loading or having the payment method and product type in the checkout submit attempt.

support-frontend/yarn.lock
GHaberis: Yeah not entirely sure about that, I've rebased with `main` and removed those changes, so it mates `main` now.

GHaberis:
> None of the removed tracking is worth porting to ophan tracking, is it? E.g. the polyfill loading or having the payment method and product type in the checkout submit attempt.

Not entirely sure about tracking polyfill's status, it's not something we've used before. I do think interactions such as submit attempts should be tracked in the data-lake, we have card to begin documenting what's being tracked already and then to add missing calls we think would be useful.

PR #4862, opened 2023-03-23
Add masking data-atrribute to existing payment method label

GHaberis:

Before we turn on existing payment methods on the new checkout we have a task to add the `data-qm-masking="blocklist"` data attribute (explained here: https://github.com/guardian/support-frontend/pull/3789 ) to the payment method label. This will block QM for recording this info.

## Screenshot

Screenshot 2023-03-24 at 14 45 28


michaelbjacobson (approved):
👍🏼

PR #4869, opened 2023-03-29
stripe-patrons-data: snapstart

tomrf1:
Enables snapstart for the stripe-patrons lambdas. This is to speed up lambda runtimes by initialising + caching the code at deploy time.
See [here](https://aws.amazon.com/about-aws/whats-new/2022/11/aws-lambda-snapstart-java-functions/) for details of snapstart.

Invocation time isn't particularly important for these lambdas, but it was a simple project to experiment with.

Changes in this PR:
1. The build now uses _lambda versioning_ (riffraff does not use this). This is necessary for enabling snapstart.
2. The build now sets the artifact zip file name based on the build number. This is because cdk will not create a new lambda version unless we give it a different filename for each build.
3. Migrates to using the `actions-riff-raff` action for riffraff because it enables us to set the artifact name based on the build number
4. Refactors the 3 lambdas to load config during initialisation by moving the logic into the top-level class scope. This means config is now loaded synchronously, and if this fails then the riffraff deploy will fail (the lambda cloudwatch logs will contain the error)
5. Adds the `EC2ContainerCredentialsProviderWrapper` credentials provider, which is necessary for snapstart.

By loading config during lambda initialisation (at deploy time) we reduce the cold start invocation duration by about 7 seconds.
This does mean that if we need to change the config, we need to do a new riffraff deploy _with a different build number_.

### Before
Screenshot 2023-07-03 at 10 17 17

### After
Screenshot 2023-07-03 at 10 26 45


build.sbt
```diff
@@ -185,6 +185,7 @@ lazy val `supporter-product-data-dynamo` = (project in file("support-modules/sup
     libraryDependencies ++= commonDependencies,
     releaseSettings,
     scalafmtSettings,
+    mergeStrategySettings
```
tomrf1, line 188: for some reason this is necessary now we're running sbt assembly

.github/workflows/stripe-patrons-data.yml
```diff
@@ -57,6 +56,50 @@ jobs:
             ~/.coursier
           key: sbt
 
-      - name: Build and upload to RiffRaff
+      - name: Build
         run: |
-          sbt "project stripe-patrons-data" riffRaffUpload
+          sbt "project stripe-patrons-data" assembly
+
+      - name: Upload to RiffRaff
```
tomrf1, line 63: Migrate to using `actions-riff-raff` action for riffraff because it enables us to set the artifact name based on the build number

stripe-patrons-data/src/main/scala/com/gu/patrons/lambdas/ProcessStripeSubscriptionsLambda.scala
```diff
@@ -22,29 +22,35 @@ import scala.concurrent.duration.{Duration, DurationInt, MILLISECONDS}
 
 class ProcessStripeSubscriptionsLambda extends RequestHandler[Unit, Unit] {
 
+  private val stage = StageConstructors.fromEnvironment
+  private val stripeConfig = PatronsStripeConfig.fromParameterStoreSync(stage)
+  private val identityConfig = PatronsIdentityConfig.fromParameterStoreSync(stage)
```
tomrf1, line 27: By loading config synchronously in the class here we cause it to happen during lambda initialisation (at deploy time), rather than during invocation. This reduces invocation duration by about 8 seconds

stripe-patrons-data/src/main/scala/com/gu/patrons/services/ParameterStoreService.scala
```diff
@@ -59,7 +68,7 @@ object ParameterStoreService {
   lazy val client = AWSSimpleSystemsManagementAsyncClientBuilder
     .standard()
     .withRegion(Regions.EU_WEST_1)
-    .withCredentials(CredentialsProvider)
+//    .withCredentials(CredentialsProvider)
```
tomrf1, line 71: passing `CredentialsProvider` is unnecessary for a lambda, and breaks snapstart

stripe-patrons-data/src/main/scala/com/gu/patrons/services/ConfigService.scala
```diff
@@ -7,9 +7,11 @@ import com.typesafe.scalalogging.StrictLogging
 trait ConfigService extends StrictLogging {
 
   protected def findParameterOrThrow(name: String, params: List[Parameter]) =
-    findParameterValue(name, params).getOrElse(
-      throw new RuntimeException(s"Missing config value for parameter $name"),
-    )
+    findParameterValue(name, params).getOrElse {
+      val error = s"Missing config value for parameter $name"
+      logger.error(error)
```
tomrf1, line 12: this log doesn't show up in cloudwatch if it fails during the initialisation stage

support-modules/aws/src/main/scala/com/gu/aws/aws.scala
```diff
@@ -14,6 +15,7 @@ package object aws {
     new ProfileCredentialsProvider(ProfileName),
     new InstanceProfileCredentialsProvider(false),
     new EnvironmentVariableCredentialsProvider(),
+    new EC2ContainerCredentialsProviderWrapper(), // for use with lambda snapstart
```
tomrf1, line 18: see https://docs.aws.amazon.com/lambda/latest/dg/snapstart-activate.html#snapstart-credentials for how credentials are loaded with snapstart

support-services/src/main/scala/com/gu/aws/aws.scala
```diff
@@ -1,19 +0,0 @@
-package com.gu
-
-import com.amazonaws.auth.profile.ProfileCredentialsProvider
-import com.amazonaws.auth.{
-  AWSCredentialsProviderChain,
-  EnvironmentVariableCredentialsProvider,
-  InstanceProfileCredentialsProvider,
-}
-
-package object aws {
```
tomrf1, line 10: this file was an unused duplicate of another aws.scala

cdk/package.json
```diff
@@ -17,7 +17,7 @@
     "diff": "cdk diff --path-metadata false --version-reporting false"
   },
   "devDependencies": {
-    "@guardian/cdk": "50.7.0",
+    "@guardian/cdk": "50.10.0",
```
tomrf1, line 20: supports the `enableVersioning` field for lambdas

.github/workflows/stripe-patrons-data.yml
```diff
@@ -57,6 +56,50 @@ jobs:
             ~/.coursier
           key: sbt
 
-      - name: Build and upload to RiffRaff
+      - name: Build
         run: |
-          sbt "project stripe-patrons-data" riffRaffUpload
+          sbt "project stripe-patrons-data" assembly
```
jacobwinch, line 61: I _think_ that `riffRaffUpload` is going to [run the tests automatically](url), whereas IIUC `assembly` [is not doing this](https://github.com/guardian/support-frontend/actions/runs/5542018268/jobs/10116159745). Consequently, I think we might need to explicitly add `sbt test` to this command?

tomrf1: good spot!

stripe-patrons-data/README.md
```diff
@@ -1,6 +1,22 @@
-## stripe-patrons-data
+This stack consists of 3 Lambdas.
 
-This project is a simple, scheduled lambda function which will query Stripe for data on active Guardian Patrons and insert them into the SupporterProductData DynamoDB table
+The lambdas are [versioned](https://docs.aws.amazon.com/lambda/latest/dg/configuration-versions.html) so that they can use [Snapstart](https://aws.amazon.com/about-aws/whats-new/2022/11/aws-lambda-snapstart-java-functions/) for faster start-ups.
+
+This means that the lambdas are initialised during the riffraff deploy, resulting in faster invocations.
+
+Config is fetched from SSM during initialisation, not during invocation. This means that if you want to change the lambda config you need to also deploy a different build number - otherwise a new lambda version will not be created and the new config will not be picked up.
```
jacobwinch, line 7: I think this could be achieved by adding the `workflow_dispatch:` trigger ([example](https://github.com/guardian/slo-alerts/blob/3d11e60c359982fa1523f4e59f5d26238467a663/.github/workflows/ci.yaml#L2-L7)) to your GitHub Actions config. This would allow developers to easily [run a new build for `main` via the GH Actions UI](https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow) and I presume this would give you an identical build with a different `GITHUB_RUN_NUMBER` (and therefore the opportunity to reload config!).

Obviously there are other ways to do this quickly (e.g. editing the `README` and deploying a branch!) but as GH Actions is relatively new it might be worth documenting the recommended approach here?

tomrf1:
Just confirming that the triggers are pointing at the lambda version (rather than `LATEST`):
Daily schedule:
Screenshot 2023-07-13 at 15 13 23

Api gateway lambda integration:
Screenshot 2023-07-13 at 15 12 49

jacobwinch (approved):
Great work! 👍

.github/workflows/stripe-patrons-data.yml
```diff
@@ -57,6 +57,50 @@ jobs:
             ~/.coursier
           key: sbt
 
-      - name: Build and upload to RiffRaff
+      - name: Build
         run: |
-          sbt "project stripe-patrons-data" riffRaffUpload
+          sbt "project stripe-patrons-data" "test; assembly"
```
johnduffell, line 62: just check that if the tests fail, the assembly etc doesn't continue regardless

stripe-patrons-data/src/main/scala/com/gu/patrons/lambdas/PatronCancelledEventLambda.scala
```diff
@@ -40,9 +40,13 @@ import scala.concurrent.duration.{Duration, DurationInt, MINUTES}
 import scala.jdk.CollectionConverters.MapHasAsScala
 import scala.util.{Failure, Success, Try}
 
-object PatronCancelledEventLambda extends StrictLogging {
+class PatronCancelledEventLambda extends StrictLogging {
   val runner = configurableFutureRunner(60.seconds)
 
+  implicit val stage = StageConstructors.fromEnvironment
+  private val stripeConfig = PatronsStripeConfig.fromParameterStoreSync(stage)
+  private val identityConfig = PatronsIdentityConfig.fromParameterStoreSync(stage)
```
johnduffell, line 48: by making these Sync, it means they will run one after the other instead of in parallel.  I'm not sure how much benefit of running them in parallel, but if there is we could always sequence and then resolve the Futures into a val using Await or similar.

On the flip side, if only warming up AWS client is important, and subsequent fetches are cheap, we could just fetch one of the config and throw it away, and then refetch it all once the lambda is woken up again.  This would make things easier if we update the config, as we wouldn't have to redeploy/restart the lambda to pick it up.

stripe-patrons-data/src/main/scala/com/gu/patrons/lambdas/PatronSignUpEventLambda.scala
```diff
@@ -40,26 +40,26 @@ import scala.concurrent.duration.{Duration, DurationInt, MINUTES}
 import scala.jdk.CollectionConverters.MapHasAsScala
 import scala.util.{Failure, Success, Try}
 
-object PatronSignUpEventLambda extends StrictLogging {
+class PatronSignUpEventLambda extends StrictLogging {
   val runner = configurableFutureRunner(60.seconds)
 
+  implicit val stage = StageConstructors.fromEnvironment
+  private val stripeConfig = PatronsStripeConfig.fromParameterStoreSync(stage)
+  private val identityConfig = PatronsIdentityConfig.fromParameterStoreSync(stage)
```
johnduffell, line 48: I am wondering in my mind how easy it would be to write an automated test for this.  It would be very easy to regress this if it were implemented in a critical lambda and I'm wondering how we would know before we deploy - it could get expensive quickly plus slow for users.

stripe-patrons-data/src/main/scala/com/gu/patrons/services/ConfigService.scala
johnduffell: good point - where do the logs/stderr go from the initial startup/bake at deploy time?

tomrf1: sorry I should have replied to my original comment!
I reported this issue to AWS and they've since fixed it.
So errors during initialisation _do_ now show in cloudwatch logs as normal.
So if a riffraff deploy fails, you can find the relevant logs fairly easily

.github/workflows/stripe-patrons-data.yml
tomrf1: Good point.
I just checked locally by breaking a test, and it does _not_ continue with the `assembly` step

stripe-patrons-data/src/main/scala/com/gu/patrons/lambdas/PatronSignUpEventLambda.scala
tomrf1: I see what you mean. The difference between code that gets executed during initialisation or not is quite subtle.
But we don't want the tests to be talking to AWS. I will keep thinking about this one...

stripe-patrons-data/src/main/scala/com/gu/patrons/lambdas/PatronCancelledEventLambda.scala
tomrf1: >> we could just fetch one of the config and throw it away
I hadn't considered this! It's a bit hacky :)
I think I'm coming round to the idea of tying a snapshot of the config to a deploy. In a way it's closer to how ec2 apps work

stripe-patrons-data/src/main/scala/com/gu/patrons/lambdas/PatronSignUpEventLambda.scala
tomrf1: It's not as good as a test, but we could probably create cloudwatch alarms that are triggered when a lambda's duration exceeds some threshold, which would suggest something has stopped happening at initialisation time

stripe-patrons-data/build.sbt
```diff
@@ -42,6 +44,13 @@ assembly / assemblyMergeStrategy := {
     oldStrategy(y)
 }
 
+// We also have to put the build number in the .jar, because AWS refuses to create a new lambda version if the jar is the same!
+resourceGenerators in Compile += Def.task {
+  val file = (resourceManaged ).value / "build.number"
+  IO.write(file, buildNumber)
+  Seq(file)
+}.taskValue
```
tomrf1, line 52: I've had to add this additional task in the build.sbt because the deploy was failing if the contents of the jar didn't change between deploys - even though the filename was different! AWS is very strict.
This task generates a file with the build number, to ensure each jar is unique

cdk/lib/stripe-patrons-data.ts
```diff
@@ -39,10 +43,10 @@ function parameterStorePolicy(scope: GuStack, appName: string) {
 }
 
 class StripePatronsDataLambda extends GuScheduledLambda {
-  constructor(scope: GuStack, id: string, appName: string) {
+  constructor(scope: GuStack, id: string, appName: string, buildNumber: string) {
     super(scope, id, {
       app: appName,
-      fileName: `${appName}.jar`,
+      fileName: `${buildNumber}.jar`,
```
rupertbates, line 49:
```suggestion
      fileName: `${appName}-${buildNumber}.jar`,
```

cdk/lib/stripe-patrons-data.ts
```diff
@@ -78,51 +86,62 @@ class StripePatronsDataLambda extends GuScheduledLambda {
 }
 
 class PatronSignUpLambda extends GuLambdaFunction {
-  constructor(scope: GuStack, id: string, appName: string) {
+  constructor(scope: GuStack, id: string, appName: string, buildNumber: string) {
     super(scope, `${appName}-sign-up`, {
       app: appName,
-      fileName: `${appName}.jar`,
+      fileName: `${buildNumber}.jar`,
```
rupertbates, line 92:
```suggestion
      fileName: `${appName}-${buildNumber}.jar`,
```

cdk/lib/stripe-patrons-data.ts
```diff
@@ -78,51 +86,62 @@ class StripePatronsDataLambda extends GuScheduledLambda {
...
   }
 }
 
 class PatronCancelledLambda extends GuLambdaFunction {
-  constructor(scope: GuStack, id: string, appName: string) {
+  constructor(scope: GuStack, id: string, appName: string, buildNumber: string) {
     super(scope, `${appName}-cancelled`, {
       app: appName,
-      fileName: `${appName}.jar`,
+      fileName: `${buildNumber}.jar`,
```
rupertbates, line 113:
```suggestion
      fileName: `${appName}-${buildNumber}.jar`,
```

stripe-patrons-data/build.sbt
```diff
@@ -30,7 +30,9 @@ riffRaffArtifactResources += (
 riffRaffArtifactResources += (
   file("cdk/cdk.out/StripePatronsData-CODE.template.json"), "cfn/StripePatronsData-CODE.template.json"
 )
-assemblyJarName := s"${name.value}.jar"
+// We use the buildNumber to set the lambda fileName, because lambda versioning requires a new fileName each time
+val buildNumber = sys.env.getOrElse("GITHUB_RUN_NUMBER", "DEV")
+assemblyJarName := s"$buildNumber.jar"
```
rupertbates, line 35:
```suggestion
assemblyJarName := s"${name.value}-$buildNumber.jar"
```

cdk/lib/stripe-patrons-data.ts
tomrf1: 👍 I've pushed these changes up in a single commit - the workflow file needed a change as well

tomrf1:
Duration of the stripe-patrons-data lambda over the last week -

Screenshot 2023-10-02 at 08 28 10

it's run on a schedule, so the duration isn't very important!
but it's shaved off about 8 seconds, so potentially quite useful for other types of lambdas

stripe-patrons-data/build.sbt
```diff
@@ -42,20 +44,9 @@ assembly / assemblyMergeStrategy := {
     oldStrategy(y)
 }
 
-lazy val deployToCode =
```
johnduffell, line 45: @tomrf1 why did you have to take this out? was it fundamentally broken, or you just didn't see the value in fixing it up?

tomrf1: The conclusion was that it's not compatible with snapstart, which involves creating versions of lambdas.
I found this old chat message:
Screenshot 2024-04-10 at 15 04 35

johnduffell: ah got it so we would have to change this to reflect the new way riffraff works - first create and upload the assembly jar, and then either run the cloudformation update, or whatever the `aws` commands are to create a version and then use it from the state machine

PR #4871, opened 2023-03-30
sticky CTA mobile optimisation test2

paul-daniel-dempsey:
Sticky CTA Support at tablet breakpoints and below only.

sticky CTA (plus scrolling) operation from left to right as shown below ->
Screenshot 2023-03-30 at 14 54 18

[**Trello Card**](https://trello.com/c/xkv6bagO/1107-mobile-optimisation-variant-b-build-for-global-ab-test-build)

## Is this an AB test?
- [x] Yes
- [ ] No

https://support.thegulocal.com/uk/contribute#ab-supporterPlusMobileTest2=variant
https://support.thegulocal.com/uk/contribute#ab-supporterPlusMobileTest2=control


support-frontend/assets/pages/supporter-plus-landing/supporterPlusLanding.tsx
```diff
@@ -156,147 +208,234 @@ export function SupporterPlusLandingPage({
...
+		!optimisedMobileLayout2,
+	);
+	const [buttonIsStuck, setButtonIsStuck] = useState(false);
+
+	// Use IntersectionObserver to detect when button is 'stuck' at the bottom
+	// of the viewport. The bottom of the observable area is set to -1px so that
+	// when the button is stuck it is not considered to be fully visible. The
+	// top edge is similarly extended upwards so the button is considered fully
+	// visible when scrolling off the top of the screen.
+
```
i-hardy, line 227: Is there a reason you opted for this kind of approach over using `position: sticky`? I feel like using primarily CSS for this would make the code a lot simpler to follow and to either clean up or ship following the test, but I don't know if there are some technical restrictions in play here.

support-frontend/assets/pages/supporter-plus-landing/supporterPlusLanding.tsx
```diff
@@ -156,147 +213,234 @@ export function SupporterPlusLandingPage({
...
+								) : (
+									heading
+								)}
+							</Hide>
+							<Box
+								cssOverrides={optimisedMobileLayout ? shorterBoxMargin : css``}
+							>
+								<AmountAndBenefits />
+							</Box>
+							<div css={displayContributionsStripe(stripeDisplayed)}>
```
i-hardy, line 331: I think the sticky button should probably go here in the HTML structure rather than at the bottom above the footer. This _should_ mean that a screen reader user will be able to move on to the next part of the form without us having to do anything extra to manage reading order. However we'll probably also need to do something to indicate that more content is being revealed when you hit the button, I'm just not sure what.

support-frontend/assets/pages/supporter-plus-landing/supporterPlusLanding.tsx
paul-daniel-dempsey: I took this from manage-frontend (SwitchOptions.tsx) which had a corresponding storybook with an example here -> https://61eee3dabb99df003a4090fd-qlolqzjxkh.chromatic.com/?path=/story/pages-switchoptions--below-threshold

support-frontend/assets/pages/supporter-plus-landing/supporterPlusLanding.tsx
paul-daniel-dempsey: Yes, thought this method might cause issues, think I struggled to get the button to scroll on-top-off of <PageScaffold> this way. Will give it another try now. Thanks.

support-frontend/assets/pages/supporter-plus-landing/supporterPlusLanding.tsx
paul-daniel-dempsey: Will see if I can research, do a simpler way now. If you know of an article that might help perhaps drop it in chat?

paul-daniel-dempsey: Yes, all related isStuck useState/useEffects were redundant in this case. Could of checked this more closely. Have removed it all.

support-frontend/assets/pages/supporter-plus-landing/supporterPlusLanding.tsx
paul-daniel-dempsey: Hmmm, have managed at least to have a single footer now. Looking at a few articles like this -> https://elad.medium.com/css-position-sticky-how-it-really-works-54cd01dc2d46, it says when you define an element with 'position: sticky' you’re automatically defining the parent element as a sticky container. So, in order to get the button to 'scroll' the only place I could put it seems to be above <PageScaffold> in the viewport. I cant seem to specify the parent container explicitly and thus place it where I would want. Maybe something I have missed?

support-frontend/assets/components/page/pageScaffold.tsx
```diff
@@ -13,7 +13,7 @@ const container = css`
 	flex-direction: column;
 	min-height: 100vh;
 	max-width: 100%;
-	overflow-x: hidden;
+	overflow-x: clip;
```
paul-daniel-dempsey, line 16: clip in conjunction with sticky, is how you force the sticky button to move up on the horizontal axis when the bottom of the view is dragged up. Thus you can put the sticky button inside the page scaffold now. Well done!

support-frontend/assets/pages/supporter-plus-landing/supporterPlusLanding.tsx
```diff
@@ -298,7 +272,22 @@ export function SupporterPlusLandingPage({
 								>
 									<AmountAndBenefits />
 								</Box>
-								<div css={displayContributionsStripeCss(stripeDisplayed)}>
+								{optimisedMobileLayout2 && (
+									<StickyCta
```
paul-daniel-dempsey, line 276: Can see moved to exactly where its displayed on screen as suggested would be best

support-frontend/assets/pages/supporter-plus-landing/components/stickyCta.tsx
```diff
@@ -0,0 +1,65 @@
...
+	return (
+		<section css={stickyContainerCss(isVisible)}>
+			<ThemeProvider theme={buttonThemeReaderRevenueBrand}>
+				<LinkButton
+					cssOverrides={buttonOverrides}
+					size="small"
+					onClick={onCtaClick}
+					href={ctaLink}
+					role="button"
+					aria-controls={ariaControls}
```
paul-daniel-dempsey, line 57: Presume aria? controls and roles for screen readers?

paul-daniel-dempsey (commented):
Perfect, just comments to aid my understanding really. Thankyou for the help completing Imogen. Ah I cant approve however, doh!

support-frontend/assets/pages/supporter-plus-landing/components/stickyCta.tsx
i-hardy: [aria-controls](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-controls) and [aria-expanded](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded) help to give information to a screenreader about the fact that a particular element manages the visibility of something else on the page.

Overriding the role is _generally_ something you want to avoid, but in this case it's useful to have a link that can navigate the user to the ID of the expanded element, but that can also indicate its expanded/collapsed state to a screenreader (which seems to be behaviour reserved for buttons). So this indicates to assistive technologies that this link should be treated as a button.

paul-daniel-dempsey:
[Mobile optimisation: variant B build for global AB test: build](https://trello.com/c/xkv6bagO/1107-mobile-optimisation-variant-b-build-for-global-ab-test-build)

support-frontend/assets/helpers/redux/commonState/selectors.ts
```diff
@@ -55,3 +55,16 @@ export function isUserInAbVariant(abTestName: string, variantName: string) {
 		return participations[abTestName] === variantName;
 	};
 }
+
+export function isUserInAnyAbVariant(
```
GHaberis, line 59: Is this function `isUserInAnyAbVariant` used anywhere?

paul-daniel-dempsey: No, was umming/ahhing whether to keep this. Thought was useful to check 2+ variants at the same time. Can remove however as it is clutter. Presume if not used, just remove is the rule?

undergroundquizscene (approved):

GHaberis (approved):

support-frontend/assets/helpers/redux/commonState/selectors.ts
GHaberis: Yeah i'd generally remove unused code to keep the codebase tidy, if we did keep it I'd suggest we add a comment, or even better unit tests to demonstrate it's purpose. If we're unsure I'd err on the side of removing unused code, and then revisit the decision if the need for the export arises in the future.

support-frontend/assets/pages/supporter-plus-landing/supporterPlusLanding.tsx
```diff
@@ -146,119 +170,165 @@ export function SupporterPlusLandingPage({
...
-			}
-			footer={
-				<FooterWithContents>
-					<FooterLinks></FooterLinks>
-				</FooterWithContents>
+	const [fullFormDisplayed, setFullFormDisplayed] = useState(
+		!optimisedMobileLayout2,
+	);
+
+	function onStickyButtonClick() {
```
GHaberis, line 177: I think we need to add a tracking call on this button using the `sendTrackingEventsOnClick` function so we measure click through rate when evaluating the test...

```
sendTrackingEventsOnClick({
    id: 'sticky_cta',
    componentType: 'ACQUISITIONS_BUTTON',
})();
```

PR #4877, opened 2023-04-02
Switching to braze-emails queue (from contribution-thanks).

pawel-krupinski-guardian:
We decided to retire all SQS queues that send emails in favour of single queue - braze-emails.
This change switches support-frontend to use braze-emails queue rather than contribution-thanks.
It also provides the queue name through CloudFormation / system environment rather than config.

### DIFF
--- a/support-workers/cloud-formation/src/templates/cfn-template.yaml
+++ b/support-workers/cloud-formation/src/templates/cfn-template.yaml
@@ -77,12 +77,8 @@ Resources:
               Action:
               - sqs:GetQueueUrl
               - sqs:SendMessage
-              Resource: arn:aws:sqs:eu-west-1:865473395570:contributions-thanks
-            - Effect: Allow
-              Action:
-              - sqs:GetQueueUrl
-              - sqs:SendMessage
-              Resource: arn:aws:sqs:eu-west-1:865473395570:contributions-thanks-dev
+              Resource:
+                Fn::ImportValue: !Sub "comms-${Stage}-EmailQueueArn"
         - PolicyName: CloudWatchLogging
           PolicyDocument:
             Version: '2012-10-17'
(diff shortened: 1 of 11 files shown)


pawel-krupinski-guardian:
[Switch producers of braze email queues other than braze-emails](https://trello.com/c/9Qbo1x1G/4019-switch-producers-of-braze-email-queues-other-than-braze-emails)

support-workers/cloud-formation/src/templates/cfn-template.yaml
```diff
@@ -77,12 +77,9 @@ Resources:
               Action:
               - sqs:GetQueueUrl
               - sqs:SendMessage
-              Resource: arn:aws:sqs:eu-west-1:865473395570:contributions-thanks
-            - Effect: Allow
-              Action:
-              - sqs:GetQueueUrl
-              - sqs:SendMessage
-              Resource: arn:aws:sqs:eu-west-1:865473395570:contributions-thanks-dev
```
undergroundquizscene, line 85: Is this not necessary any more?

undergroundquizscene:
I can see the contributions-thanks-dev queue is also mentioned in [support-lambdas/it-test-runner/cfn.yaml](https://github.com/guardian/support-frontend/blob/main/support-lambdas/it-test-runner/cfn.yaml) – I think we’ll need to update that as part of this PR as well?

pawel-krupinski-guardian:
Good spot, fixed now.

support-workers/cloud-formation/src/templates/cfn-template.yaml
undergroundquizscene: Ah, I see what I missed: there used to be two queues here and now there’s only one. We still have the Allow for the queue.

undergroundquizscene (approved):

PR #2378, opened 2020-03-04
Fix state mismatch and console logging bugs

paulbrown1982:
## Why are you doing this?
This change fixes 3 bugs and has 1 refactor:

1) A mismatch between the state and country when the user's billing data is sent to Zuora. Before, when somebody chose not to transact in USD, CAD or GBP, but was based in USA or CA (where a state is required), we would read their billing country from their GEO-IP, but then we'd have no state to add alongside the country, as the form for the other currencies do not have a state drop-down, and would submit a blank state which would error in Zuora. This solution also reads the state value from GEO-IP. It's not a perfect solution as the GEO-IP read may fail. A full solution would be to include a billing country and optional state drop-down on all checkouts. That may be a later PR if we still get some of these 'on holiday' errors. This was not introduced by https://github.com/guardian/support-frontend/pull/2372 but it does rename the `'UPDATE_STATE_OR_PROVINCE'` action which was also renamed in that PR.

2) Fixed a bug in the StripePaymentRequestButton.jsx where we would log that we were missing a state value for non US or CA checkouts.

3) Fixed a bug where if the payment from the PRB fails, the country from the PRB would remain in the formData, and be selected with highest precedence. If the user decided to move to using card entry, and it was the US or CA checkout which required a state, then the transaction would be permanently blocked because there was no way to set the billingCountry back to the page internationalisation.

4) Refactored the 2 Actions.js and Reducer.js and formValidation.js classes to name the real-world (not React) 'state' variable 'billingState' in order to distinguish it and be named similar to billingCountry.

[**Trello Card**](https://trello.com/c/qNyxy3in)

## Screenshots
Screen Shot 2020-03-04 at 13 33 47

PR #4881, opened 2023-04-03
Fix apple pay country & state logic

undergroundquizscene:
Trello card: https://trello.com/c/PBBDfgCy/1159-check-on-client-side-validation-for-state-field-when-using-apple-pay-or-google-pay

# Background

We’ve had some examples recently of users paying via Apple Pay where the currency is GBP and the credit card country is GB, but the billing address country is US and the `state` field is the empty string. This then fails in support-workers because the state isn’t allowed to be empty when the country is US. Some users then try another payment, and unfortunately the country and state data from Apple Pay is reused on subsequent submissions, preventing them from continuing.

# Changes

## Use correct country to check if state is required

At the moment, if a user is on the UK page but uses Apple Pay and we get a billing address that’s in the US, we don’t require a non-empty state on the client-side. We then send the request to the backend, which fails.

To fix this, this commit uses the same logic for determining the country when checking whether a state is required as is used when sending the info to the backend. This way the client-side and server-side validation should be consistent.

## Add stripePaymentMethod to state, only use payment-request-button-provided country and state when the payment request button is being used

Another option might be to change the logic in [getBillingCountryAndState](https://github.com/guardian/support-frontend/blob/a24606d40e5ab5cf920af265cd67dbc00d5ee69f/support-frontend/assets/pages/supporter-plus-landing/setup/legacyActionCreators.ts#L139) to check the current payment method. Conceptually I’m more comfortable with that option, because it doesn’t require co-ordinating the logic in two places, but I don’t think the payment method detail (i.e. more specific than the `PaymentMethod` type) is easily available there.

At the moment, if a payment via Apple Pay (or presumably other stripe request button payment methods) fails, the country and state will get reused for subsequent card payments, potentially causing further failure. I saw this recently where we got US as the country and no state from Apple Pay, and when that failed the user retried with card, which also failed. Retrying with direct debit fixed the problem for this user, but it shouldn’t be necessary.

Paul Brown fixed this bug for some payment types in [PR 2378](https://github.com/guardian/support-frontend/pull/2378) (in commit 14b6a2b351fdcd5a198d1958b359ad0467daae34) by clearing the country and state on payment failures. However, since the else branch isn’t taken when the payment method is Apple Pay, the bug can still happen in that case.

Rather than deleting the country and state when payment fails, this change keeps track of whether the current/most recently used payment method is the payment request button, and only sends the country and state obtained from the button if it is.

This should fix the bug for Apple Pay users, but I have not verified that it does – I don’t know how to test it, because Apple Pay won’t let me add an invalid card, or one that has an invalid billing address, or a test card. I have confirmed that the happy path still works: using a US Apple Pay card on the UK page or vice-versa works fine (and the billing address is set correctly) if the address has a state when needed.

I wanted to set the `stripePaymentMethod` correctly when using Stripe without the payment request button as well, but it was hard to determine whether the correct value should be `StripeCheckout` or `StripeElements`, or a mix of the two. Currently we use a mix: mostly `StripeCheckout` but with `StripeElements` used once.

I believe the original intended meaning was that `StripeCheckout` meant Stripe’s low-code integration (https://stripe.com/docs/payments/checkout), and `StripeElements` meant the prebuilt UI components that we later switched to (https://stripe.com/docs/payments/elements). However, we aren’t using the types that way: as far as I can see we don’t use the low-code integration at all anymore, but we do still use `StripeCheckout`.

I wanted to clarify the situation (and maybe get rid of one of the two), but don’t want to spend more time on this. Maybe we can update this in the future.

### DIFF
--- a/support-frontend/assets/components/paymentFrequencyTabs/paymentFrequencyTabsContainer.tsx
+++ b/support-frontend/assets/components/paymentFrequencyTabs/paymentFrequencyTabsContainer.tsx
@@ -47,7 +47,7 @@ export function PaymentFrequencyTabsContainer({
 		);
 
 		dispatch(setProductType(contributionType));
-		dispatch(setPaymentMethod(paymentMethodToSelect));
+		dispatch(setPaymentMethod({ paymentMethod: paymentMethodToSelect }));
 	}
 
 	const tabs: TabProps[] = contributionTypes[countryGroupId].map(
--- a/support-frontend/assets/components/paymentMethodSelector/PaymentMethodSelectorContainer.tsx
+++ b/support-frontend/assets/components/paymentMethodSelector/PaymentMethodSelectorContainer.tsx
@@ -109,7 +109,7 @@ function PaymentMethodSelectorContainer({
 		if (event === 'select') {
 			trackComponentClick(trackingId);
 			sendEventContributionPaymentMethod(paymentMethodDescription);
-			dispatch(setPaymentMethod(paymentMethod));
+			dispatch(setPaymentMethod({ paymentMethod }));
 			existingPaymentMethod &&
 				dispatch(selectExistingPaymentMethod(existingPaymentMethod));
 		} else {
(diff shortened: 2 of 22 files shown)


i-hardy (approved):
This looks good, a much more coherent way of managing this information 👍

PR #4883, opened 2023-04-04
Support Acquisition API GuCDK Migration-Recreate existing CFN template using CDK constructs

LAKSHMIRPILLAI:
Recreate existing CFN template using CDK constructs, while retaining existing infrastructure for Acquisition-events-api.This is done for Support Acquisition API GuCDK Migration.

[**Trello Card**](https://trello.com/c/UMwD5nX3/1222-support-acquisition-api-gucdk-migration-recreate-existing-cfn-template-using-cdk-constructs-while-retaining-existing-infrastruct)

## Why are you doing this?

For apps where avoiding downtime is critical, a dual-stack migration strategy forGuCDK Migration is recommended . The idea is to duplicate the key parts of the stack using the pattern, test, and, once confident, switch DNS to point to the new stack.
In this phase  we create a new version of your app's infrastructure, according to CDK’s best practices. So, we should have a functional copy of the application, which can be used for testing purposes. Riff-Raff will have been configured to deploy to both versions of the infrastructure, so you can operate in this state for as long as you need to.

### DIFF
--- a/cdk/lib/acquisition-events-api.ts
+++ b/cdk/lib/acquisition-events-api.ts
@@ -1,11 +1,12 @@
 import { join } from "path";
 import { GuApiLambda } from "@guardian/cdk";
 import type { GuStackProps } from "@guardian/cdk/lib/constructs/core";
-import { GuStack, GuStringParameter } from "@guardian/cdk/lib/constructs/core";
+import { GuStack } from "@guardian/cdk/lib/constructs/core";
 import type { App } from "aws-cdk-lib";
 import { Duration } from "aws-cdk-lib";
+import { CfnBasePathMapping, CfnDomainName } from "aws-cdk-lib/aws-apigateway";
 import { Runtime } from "aws-cdk-lib/aws-lambda";
-import { CfnRecordSetGroup } from "aws-cdk-lib/aws-route53";
+import {CfnRecordSet} from "aws-cdk-lib/aws-route53";
 import { CfnInclude } from "aws-cdk-lib/cloudformation-include";
 
 export interface AcquisitionEventsApiProps extends GuStackProps {
@@ -15,6 +16,7 @@ export interface AcquisitionEventsApiProps extends GuStackProps {
   certificateId: string;
   domainName: string;
   hostedZoneName: string;
+  hostedZoneId: string;
 }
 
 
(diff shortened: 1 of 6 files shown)


cdk/lib/acquisition-events-api.ts
```diff
@@ -1,23 +1,135 @@
 import { join } from "path";
```
undergroundquizscene, line 1:
```suggestion
```

cdk/lib/acquisition-events-api.ts
```diff
@@ -1,23 +1,135 @@
...
+import { GuApiLambda } from "@guardian/cdk";
 import type { GuStackProps } from "@guardian/cdk/lib/constructs/core";
 import { GuStack } from "@guardian/cdk/lib/constructs/core";
 import type { App } from "aws-cdk-lib";
+import { Duration } from "aws-cdk-lib";
+import { CfnBasePathMapping, CfnDomainName } from "aws-cdk-lib/aws-apigateway";
+import {Effect, Policy, PolicyStatement} from "aws-cdk-lib/aws-iam";
+import { Runtime } from "aws-cdk-lib/aws-lambda";
+import {CfnRecordSet} from "aws-cdk-lib/aws-route53";
 import { CfnInclude } from "aws-cdk-lib/cloudformation-include";
```
undergroundquizscene, line 11:
```suggestion
```

support-lambdas/acquisition-events-api/riff-raff.yaml
```diff
@@ -6,7 +6,9 @@ deployments:
     type: aws-lambda
     parameters:
       bucket: membership-dist
-      functionNames: [acquisition-events-api-]
+      functionNames:
+       -acquisition-events-api-
+       -acquisition-events-api-cdk-
```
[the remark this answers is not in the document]
LAKSHMIRPILLAI: updated as per suggestions

cdk/lib/acquisition-events-api.ts
LAKSHMIRPILLAI: I think it's better to leave it here till I remove the existing API Gateway Instance in next phase

undergroundquizscene (approved):

michaelbjacobson (approved):
Looks good! 👍🏼

cdk/lib/acquisition-events-api.ts
```diff
@@ -1,23 +1,159 @@
...
+    new GuAlarm(this, "ApiGateway4XXAlarmCDK", {
+      app,
+      alarmName: alarmName("API gateway 4XX response"),
+      alarmDescription: alarmDescription(
+        "Acquisition Events API received an invalid request"
+      ),
+      evaluationPeriods: 1,
+      threshold: 1,
+      snsTopicName: "contributions-dev",
+      actionsEnabled: true,
```
michaelbjacobson, line 92: I suggest we make this conditional, based on the stage/environment, as we don't want to deploy alarms in `CODE`.

michaelbjacobson (approved):
Looks great!

cdk/lib/acquisition-events-api.ts
michaelbjacobson: Eg. `actionsEnabled: isProd()`

LAKSHMIRPILLAI: I deployed alarms to CODe for testing.So now disabling it now as per the suggestion.Instead of using isProd() function,. I have used actionsEnabled: this.stage === "PROD"

PR #4888, opened 2023-04-11
Disambiguate error logs

undergroundquizscene:
We’ve had some errors recently that are coming from one of these lines, but we’re not entirely sure which. Rather than stripping the context (i.e. which kind of error we matched in the case), this change restores that info to the logs. Hopefully this should help us understand what’s causing the timeouts we’re seeing.


LAKSHMIRPILLAI (approved):
LGTM 👍

PR #4892, opened 2023-04-13
Allow taking out multiple recurring contributions

emdash-ie:

Trello Card: https://trello.com/c/idQFlaUj/1173-remove-client-side-validation-that-prevents-recurring-contributors-from-taking-out-supporter-plus-or-another-recurring-contribut

This PR removes the restriction on taking out multiple recurring contributions, allowing users to take out as many recurring contributions (above or below the supporter plus threshold) as they like.

## Why are you doing this?

Motivation summarised from the Trello card:

- we’ve seen a high number of users prevented from checking out by this validation error

- the endpoint we use to determine whether users have a recurring contribution subscription does not immediately return false for users that have cancelled their subscription, preventing users from taking out replacement subscriptions

## Checks

- [ ] Inform user help and Customer Experience when we make this change
- [ ] Let Georgia and Charles know too

## Is this an AB test?

- [ ] Yes
- [X] No


undergroundquizscene:
[Remove client-side validation that prevents recurring contributors from taking out Supporter Plus or another recurring contribution.](https://trello.com/c/idQFlaUj/1173-remove-client-side-validation-that-prevents-recurring-contributors-from-taking-out-supporter-plus-or-another-recurring-contribut)

rupertbates (approved):
Looks good 👍

PR #4893, opened 2023-04-13
Fix it tests broken by removing UAT

rupertbates:

#4891 Broke some integration tests, this fixes them.

### DIFF
--- a/support-workers/src/test/scala/com/gu/salesforce/SalesforceSpec.scala
+++ b/support-workers/src/test/scala/com/gu/salesforce/SalesforceSpec.scala
@@ -58,19 +58,6 @@ class SalesforceSpec extends AsyncFlatSpec with Matchers with LazyLogging {
     }
   }
 
-  it should "get a different auth token for each stage" in {
-    val config = Configuration.load().salesforceConfigProvider.get()
-
-    val futureAuths = for {
-      devAuth <- AuthService.getAuth(Configuration.load().salesforceConfigProvider.get())
-      uatAuth <- AuthService.getAuth(Configuration.load().salesforceConfigProvider.get(true))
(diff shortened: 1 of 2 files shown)


support-workers/src/test/scala/com/gu/salesforce/SalesforceSpec.scala
```diff
@@ -58,19 +58,6 @@ class SalesforceSpec extends AsyncFlatSpec with Matchers with LazyLogging {
     }
   }
 
-  it should "get a different auth token for each stage" in {
```
rupertbates, line 61: This is not applicable any more as the UAT environment no longer exists

support-workers/src/test/scala/com/gu/zuora/ZuoraITSpec.scala
```diff
@@ -21,15 +21,15 @@ import scala.concurrent.duration._
 @IntegrationTest
 class ZuoraITSpec extends AsyncFlatSpec with Matchers {
 
-  def uatService: ZuoraService =
+  def codeService: ZuoraService =
     new ZuoraService(
-      Configuration.load().zuoraConfigProvider.get(true),
```
undergroundquizscene, line 26: What’s the reason for making this no longer a test user?

undergroundquizscene (approved):

support-workers/src/test/scala/com/gu/zuora/ZuoraITSpec.scala
rupertbates: It doesn't make any difference either way because test users and non-test users will use the SANDBOX environment in these tests so I thought it was less confusing to new comers to have to think about what that `true` meant.

PR #4904, opened 2023-04-19
Remove checkout-in-epic leftovers

undergroundquizscene:
We’ve been getting some 500s recently in support-frontend because this checkout-in-epic route has seen some requests, but we don’t provide the `contributionsCheckoutInEpic.js` file anymore:

> AssetNotFoundException: Cannot find asset RefPath(contributionsCheckoutInEpic.js

The checkout-in-epic was disabled in https://github.com/guardian/support-dotcom-components/pull/648, and the support-dotcom-components component `ContributionsEpicCheckout` which sends requests to this path is no longer used anywhere. We suspect that a user has kept open a tab from a year ago, and that that’s why we’re getting these requests now.

This PR removes a few things introduced in [commit 7c2bf2462ad189f8d961b87b6c99d1da3b08be0a](https://github.com/guardian/support-frontend/commit/7c2bf2462ad189f8d961b87b6c99d1da3b08be0a), including some dependencies that don’t seem to be used anymore.


tomrf1 (approved):
👍 thanks!

PR #4905, opened 2023-04-19
Realign the build numbers for Acquisitions events api

michaelbjacobson:
This is to update acquisition-events-api.yml to realign the build numbers  for Acquistion-events-api .

This is to resolve the issue 
### Unable to locate cloudformation template s3://riffraff-artifact/support:lambdas:acquisition-events-api/16122/cfn/AcquisitionEventsAPI-CODE.template.json

that occurred during the riff-raff deployment caused due to a mismatch in the build numbers 

image

### DIFF
--- a/.github/workflows/acquisition-events-api.yml
+++ b/.github/workflows/acquisition-events-api.yml
@@ -62,4 +62,6 @@ jobs:
 
       - name: Build and upload to RiffRaff
         run: |
+          export LAST_TEAMCITY_BUILD=6000
+          export GITHUB_RUN_NUMBER=$(( $GITHUB_RUN_NUMBER + $LAST_TEAMCITY_BUILD ))
           sbt "project acquisition-events-api" riffRaffUpload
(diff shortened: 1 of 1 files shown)


.github/workflows/acquisition-events-api.yml
```diff
@@ -62,4 +62,6 @@ jobs:
 
       - name: Build and upload to RiffRaff
         run: |
+          export LAST_TEAMCITY_BUILD=20000
+          export GITHUB_RUN_NUMBER=$(( $GITHUB_RUN_NUMBER + $LAST_TEAMCITY_BUILD ))
```
LAKSHMIRPILLAI, line 66: I think, without these lines, the acquisition-events-api was getting the build number generated by the  support-lambdas-build.yml

undergroundquizscene: How did you get the 20,000 value, out of curiosity? Picked a value bigger than the most recent TeamCity build?

undergroundquizscene (approved):

.github/workflows/acquisition-events-api.yml
LAKSHMIRPILLAI: > How did you get the 20,000 value, out of curiosity? Picked a value bigger than the most recent TeamCity build?

yes

PR #4908, opened 2023-04-20
Migrate delivery instructions and separate billing address toggle to RTK

i-hardy:

This creates an `addressMeta` slice that manages a couple of additional bits of address-related information:

- Delivery instructions for newspaper home delivery
- Whether or not the user's billing address is different to their delivery address

This also fixes a small bug where errors relating to the delivery instructions field- if users entered characters that aren't compatible with Zuora's database- were not being passed to the text area component to be displayed inline as well as in the error summary at the bottom of the form.

[**Trello Card**](https://trello.com/c/FeSSCuFe)

## Why are you doing this?

This is a bit of address-related state that we missed in the original address migration.

### DIFF
--- a/support-frontend/assets/components/directDebit/directDebitProgressiveDisclosure/directDebitForm.test.tsx
+++ b/support-frontend/assets/components/directDebit/directDebitProgressiveDisclosure/directDebitForm.test.tsx
@@ -75,7 +75,7 @@ describe('Direct debit form', () => {
 			page: {
 				checkout: {
 					formErrors: [],
-					billingAddressIsSame: true,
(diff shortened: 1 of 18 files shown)


i-hardy:
[🛠 Redux Toolkit migration - add additional address state 🛠](https://trello.com/c/FeSSCuFe/1095-%F0%9F%9B%A0-redux-toolkit-migration-add-additional-address-state-%F0%9F%9B%A0)

marialani (approved):
This looks great! 👍 

A non-blocking suggestion if simple enough, `billingAddressIsSame` &rarr; `billingAddressMatchesDelivery`

PR #4921, opened 2023-04-26
Revert billing country logic change

undergroundquizscene:
## Revert billing country logic change

I changed this logic to fix a bug where incorrect payment request button country info was getting reused on subsequent payment attempts (with different payment methods). However, it caused another bug: [the fallback logic](https://github.com/guardian/support-frontend/blob/7c692d7212d169d8d026f3767debaa43e96f2613/support-frontend/assets/pages/supporter-plus-landing/setup/legacyActionCreators.ts#L183-L195) prioritises geolocation over detection of region based on the path of the page, unlike [the initial detection of the billing country on page load](https://github.com/guardian/support-frontend/blob/7c692d7212d169d8d026f3767debaa43e96f2613/support-frontend/assets/helpers/redux/checkout/address/state.ts#L50).

This meant that users trying to use a currency other than that of their current (inferred) location, e.g. U.S. users attempting to pay in Canadian dollars or vice-versa, were unable to checkout successfully: the geolocation would detect the other country, and the state they entered would no longer be valid. The only way they could correct the error was by returning to the form that matched the country the geolocation thought they were in (and the geolocation may have been wrong, making things extra confusing for them).

Since this case appears to be more frequent than the Apple Pay problems we’ve had (based on the rate of errors), we’ve decided to revert pending a comprehensive fix.

This PR partially reverts commit c4fae6dcf24baf017d53c7499729f08f70bca4d7.

## Missed renaming: billingCountry -> fallbackCountry

The variable `fallbackCountry` here used to be called `billingCountry`, and was renamed in [commit cc42b3734141947383233909287bac2a17b8b425](https://github.com/guardian/support-frontend/commit/cc42b3734141947383233909287bac2a17b8b425#diff-6479026456bf36aed9635d37b78393c9dfe3ba4de67ba0efdf75d7346e2a5e2cR266-R274). However, this usage wasn’t renamed, meaning that when we get to this fallback case the state will likely fail to parse, as it’s being parsed based on the wrong country.

### DIFF
--- a/support-frontend/assets/helpers/redux/selectors/formValidation/personalDetailsValidation.ts
+++ b/support-frontend/assets/helpers/redux/selectors/formValidation/personalDetailsValidation.ts
@@ -11,9 +11,8 @@ export function getStateOrProvinceError(
 	const contributionType = getContributionType(state);
 	const billingCountryGroup = fromCountry(
 		getBillingCountryAndState(
-			state,
 			state.page.checkoutForm.payment.paymentMethod.name,
-			state.page.checkoutForm.payment.paymentMethod.stripePaymentMethod,
+			state,
 		).billingCountry,
 	);
 	if (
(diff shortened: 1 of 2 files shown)


GHaberis (approved):
Looks good 👍

PR #4923, opened 2023-04-26
BigqueryAcquisitionsPublisher

tomrf1:
A new stack, consisting of an API Gateway, an Eventbridge event bus, a lambda and an SQS queue.

Currently several backends send acquisition events directly to BigQuery.
In future acquisition events will instead be added to the new queue, and the lambda will be the only system responsible for writing acquisitions to BigQuery.

[Doc with more details is here](https://docs.google.com/document/d/161e602tyFTtD33P06ZqAVHjdkBIQzA6rnV6Jf5Noyds/edit)

## Progress so far:
This PR has introduced a new stack `support-PROD-bigquery-acquisitions-publisher` which contains the Eventbridge event bus `acquisitions-bus-PROD` and a rule which forwards events from this bus to a new lambda `support-PROD-bigquery-acquisitions-publisher` which carries out the task of writing events to BigQuery.

The previous api gateway and associated lambda have been removed and a new HttpApi  api gateway set up with an integration which just forwards events to the new event bus.

## Release Details
Releasing this PR required deleting the previous stack before deploying this version. This was because it was not possible (I couldn't find a way!) to update the mapping between the custom domain and the new api gateway. 
This took place before the merging of this PR and completed at 2023-09-06 15:05:32 UTC+0100


support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/bigquery/BigQueryService.scala
```diff
@@ -26,42 +27,21 @@ import com.gu.support.config.Stage
...
 import java.util.concurrent.Executors
 import scala.concurrent.{ExecutionContext, Future, Promise}
 import scala.jdk.CollectionConverters._
 
-class BigQueryService(config: BigQueryConfig) {
+class BigQueryService(credentials: Credentials) {
   private val stage: Stage = sys.env.get("STAGE").flatMap(Stage.fromString).getOrElse(CODE)
+  private val projectId = s"datatech-platform-${stage.toString.toLowerCase}"
 
-  lazy val bigQuery =
```
tomrf1, line 36: this client wasn't being used

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/bigquery/BigQueryService.scala
```diff
@@ -103,6 +83,23 @@ class BigQueryService(config: BigQueryConfig) {
 }
 
 object BigQueryService {
+
```
tomrf1, line 86: There are now 2 ways of constructing a `BigQueryService`.
The legacy way takes the `BigQueryConfig`, which contains secrets.
The new way takes a String containing json with just the credentials required for federated auth with Google Cloud.

rupertbates (approved):

PR #4925, opened 2023-04-27
Fix accessibility issues with the Tooltip component and make it more reusable

i-hardy:

This PR fixes an issue where the activation button for the `Tooltip` component had no label, and the contents of the tooltip were not read by screenreaders. It also moves the specific cancellation-related text out of the component itself, so it's more reusable if we need it elsewhere in the future.

I've implemented a [toggletip](https://inclusive-components.design/tooltips-toggletips/) pattern here as it makes the interaction make the most sense for screen reader users- the tooltip will not appear or be read until the button is actually _activated_, rather than just focused. After experimenting with the accessibility options provided by the floating-ui/react library I couldn't find one that would assign a role that actually had the tooltip contents be read aloud when the tooltip is inserted, so instead I've manually assigned the `status` role to the element that gets inserted.

[**Trello Card**](https://trello.com/c/TsWbZTNW)

## Is this an AB test?
- [ ] Yes
- [x] No

### DIFF
--- a/support-frontend/assets/components/checkoutBenefits/checkoutBenefitsList.tsx
+++ b/support-frontend/assets/components/checkoutBenefits/checkoutBenefitsList.tsx
@@ -6,7 +6,7 @@ import {
 	space,
 	textSans,
 } from '@guardian/source-foundations';
-import CancelAnytimeTooltip from 'components/tooltip/Tooltip';
+import Tooltip from 'components/tooltip/Tooltip';
 import type { CountryGroupId } from 'helpers/internationalisation/countryGroup';
 import type { CSSOverridable } from 'helpers/types/cssOverrideable';
 import type { CheckListData } from './checkoutBenefitsListData';
@@ -97,7 +97,14 @@ export function CheckoutBenefitsList({
 				))}
 			</table>
 			<hr css={hr(`${space[5]}px 0 ${space[4]}px`)} />
-			<CancelAnytimeTooltip countryGroupId={countryGroupId} />
+			<Tooltip promptText="Cancel anytime">
+				<p>
+					You can cancel
+					{countryGroupId === 'GBPCountries' ? '' : ' online'} anytime before
+					your next payment date. If you cancel in the first 14 days, you will
+					receive a full refund.
+				</p>
+			</Tooltip>
 		</div>
 	);
 }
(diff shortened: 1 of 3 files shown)


i-hardy:
[Fix tooltip component accessibility issues](https://trello.com/c/TsWbZTNW/1235-fix-tooltip-component-accessibility-issues)

undergroundquizscene:
> the tooltip will not appear or be read until the button is actually _activated_, rather than just focused

In my testing I’m finding that the tooltip appears if I load the page and then tab until the info icon is focused. From this description I expected I would need to press space or enter at that point, but that doesn’t seem to be the case. Am I misunderstanding the intention?

i-hardy:
> In my testing I’m finding that the tooltip appears if I load the page and then tab until the info icon is focused. From this description I expected I would need to press space or enter at that point, but that doesn’t seem to be the case. Am I misunderstanding the intention?

I can't recreate that, it requires activation in all three browsers I've checked it in. Are you definitely on the correct branch? The current behaviour on `main` is to show the tooltip on focus, but I've removed that behaviour in this branch.

undergroundquizscene:
> > In my testing I’m finding that the tooltip appears if I load the page and then tab until the info icon is focused. From this description I expected I would need to press space or enter at that point, but that doesn’t seem to be the case. Am I misunderstanding the intention?
> 
> I can't recreate that, it requires activation in all three browsers I've checked it in. Are you definitely on the correct branch? The current behaviour on `main` is to show the tooltip on focus, but I've removed that behaviour in this branch.

I’m definitely on the right branch, but I’ll kill everything and restart: I was on the wrong branch at first, maybe it hasn’t reloaded!

undergroundquizscene:
> I’m definitely on the right branch, but I’ll kill everything and restart: I was on the wrong branch at first, maybe it hasn’t reloaded!

Works as expected now. I hadn’t actually run `yarn devrun`, so presumably it was an old frontend build – sorry about that!

undergroundquizscene (approved):

undergroundquizscene:
One other thing: in the storybook if you open the tooltip and then hit tab, the close button on the tooltip is focused next. On the support page though, it isn’t (presumably because the tooltip is above the icon instead of below). Moving backwards with shift-tab doesn’t seem to get to the close icon either: is it possible to focus it with the keyboard?

i-hardy:
I haven't been able to get focus into the tooltip on the page itself, but it's dismissible with esc so I think that should be fine as far as keyboard support goes.

PR #4927, opened 2023-04-28
Store payment method country/state separately

undergroundquizscene:

Trello Card: https://trello.com/c/ILKQJnOC/1262-support-checkout-client-side-validation-not-enforcing-required-state-field-for-recurring-supporters-in-usa-and-ca-paying-with-cr

This PR attempts to fix issues we’ve had with apple pay users by storing the payment-request-button-provided country and state separately from the form-entered or guessed ones, so that our code can choose to only use the payment-request-button-provided info when the payment request button is the current payment method. I’ve split the description of the changes for the three commits in the PR:

### Unify country-choosing logic

Rather than have two places in which country-guessing logic lives, this commit consolidates them into one: the country is guessed on page load, and not when sending data to the backend.

I’ve also removed the special case for direct debit: the new logic will already prioritise the page base country over geolocation, and I don’t think a more explicit case is needed.

### Store payment method country/state separately

Rather than overwriting the form values, this commit stores the country and state obtained from a payment request button wallet in a separate part of the redux state. Then when working out the billing country and state for validation or for submission to the backend, the code only uses the payment-method-provided country and state if the current payment method is the payment request button.

This should fix the following error we’ve encountered: If a user has invalid country/state info set in their Apple Pay details (e.g. the country is US but the state is empty), the payment will fail on the backend. If the user then tries another payment method, the country/state info from Apple Pay will still be in the redux state, and will be sent to the backend to fail again.

Another way to fix that issue would be to clear the payment-method-provided info on error: we already do that for errors earlier in the process, and we could plausibly do that if the support-workers execution fails. However, now that we’re doing the country determination at page load, we would’ve lost that info when overwriting it with the payment-method-provided info. We could save it somewhere and restore it, or run the determination function again, but I prefer this approach. One reason for that is that it puts the choice logic all in one place: the other approaches require an insertion and deletion to be co-ordinated in different parts of the code, but this approach only requires the insertions.

Something else I’m wondering about: if we clear a user’s email due to a payment failure with the payment request button, what if they’re signed in? Will they then have a disabled but empty email input on the form, with no way to tell us their email address? Should we ignore the email address from a payment method if the user is already signed in (and then we have an email address we can have high confidence in)? I suppose they can always sign out to make the email address editable again.

### Use country instead of country group

Before now, shouldCollectStateForContribs took a country group ID, then checked for some exception countries using window.guardian.geoip.countryCode. We have the country available everywhere shouldCollectStateForContribs is called, so this commit changes the function to take the country ID instead, meaning the geolocation access and the exception logic is no longer required.

The similar function `checkStateIfApplicable` is unused, so I’ve removed it.

## Testing

I have tested the happy path of using google pay on CODE for successful single and recurring payments.

### DIFF
--- a/support-frontend/assets/components/paymentRequestButton/hooks/payerDetails.ts
+++ b/support-frontend/assets/components/paymentRequestButton/hooks/payerDetails.ts
@@ -3,20 +3,17 @@ import type {
 	PaymentRequestPaymentMethodEvent,
 } from '@stripe/stripe-js';
 import {
-	detect as detectCountry,
 	findIsoCountry,
 	stateProvinceFromString,
 } from 'helpers/internationalisation/country';
-import {
-	setBillingCountry,
-	setBillingState,
-} from 'helpers/redux/checkout/address/actions';
+import { setPaymentMethodCountryAndState } from 'helpers/redux/checkout/payment/paymentMethod/actions';
 import {
 	setEmail,
 	setFirstName,
 	setLastName,
 } from 'helpers/redux/checkout/personalDetails/actions';
 import type { ContributionsDispatch } from 'helpers/redux/contributionsStore';
+import * as storage from 'helpers/storage/storage';
 import { logException } from 'helpers/utilities/logger';
 
 function setPayerName(
@@ -49,7 +46,7 @@ function setPayerEmail(
 	}
 }
 
-function setBillingCountryAndState(
+function dispatchPaymentMethodCountryAndState(
 	dispatch: ContributionsDispatch,
 	billingDetails: PaymentMethod.BillingDetails,
 ): void {
@@ -63,8 +60,12 @@ function setBillingCountryAndState(
 			state ?? undefined,
 		);
 
-		dispatch(setBillingCountry(validatedCountry));
-		dispatch(setBillingState(validatedState ?? ''));
+		dispatch(
+			setPaymentMethodCountryAndState([
+				validatedCountry,
+				validatedState ?? undefined,
+			]),
+		);
 	}
 }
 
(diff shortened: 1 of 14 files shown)


undergroundquizscene:
[Support Checkout: client-side validation not enforcing required "state" field for Recurring supporters in USA and CA paying with Credit Card.](https://trello.com/c/ILKQJnOC/1262-support-checkout-client-side-validation-not-enforcing-required-state-field-for-recurring-supporters-in-usa-and-ca-paying-with-cr)

support-frontend/assets/helpers/redux/checkout/address/state.ts
```diff
@@ -46,9 +46,10 @@ export type AddressFieldsState = AddressFields & {
 };
 
 export function getInitialAddressFieldsState(): AddressFieldsState {
+	const country = detect();
 	return {
-		country: detect(),
-		state: '',
+		country,
```
marialani, line 51: Tiny note but could revert creating a country var here for consistency with the line below

support-frontend/assets/helpers/redux/checkout/payment/paymentMethod/reducer.ts
```diff
@@ -12,8 +16,17 @@ export const paymentMethodSlice = createSlice({
...
+			state.state = undefined;
 			state.errors = [];
 		},
+		setPaymentMethodCountryAndState(
+			state,
+			action: PayloadAction<[IsoCountry, StateProvince | undefined]>,
+		) {
+			state.country = action.payload[0];
+			state.state = action.payload[1];
+		},
```
marialani, line 29: Good shout grouping these together!

marialani (approved):
This looks great! 🥳  
A thorough approach down to the commit history and accompanying explanations! Made for a pleasant review, nicely done!

support-frontend/assets/helpers/redux/checkout/address/state.ts
undergroundquizscene: I’m sorry, I don’t follow – are you suggesting removing the `country` var and calling `detect()` twice?

marialani: Ahh I'd missed if there was a second use of `country` var! Best not to call `detect()` twice 👍

support-frontend/assets/components/paymentRequestButton/hooks/payerDetails.ts
```diff
@@ -74,13 +75,9 @@ export function addPayerDetailsToRedux(
...
+	dispatchPaymentMethodCountryAndState(dispatch, paymentMethod.billing_details);
 }
 
 export function resetPayerDetails(dispatch: ContributionsDispatch): void {
-	dispatch(setEmail(''));
-	dispatch(setFirstName(''));
-	dispatch(setLastName(''));
-	dispatch(setBillingCountry(''));
-	dispatch(setBillingState(''));
+	dispatch(setEmail('')); // TODO: what if the user was signed in?
```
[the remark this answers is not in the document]
undergroundquizscene: Ah, I didn’t think of that! That sounds good – would that be using `storage.getSession('gu.email')`? I should take out the TODO message either way, as it’s not very clear.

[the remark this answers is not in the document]
undergroundquizscene: I’ve made this change now.

PR #4889, opened 2023-04-12
Rename Sandbox TouchPointEnvironment to CODE

rupertbates:
## What are you doing in this PR?

This PR follows on from #4885 and continues the work described [in this doc](https://docs.google.com/document/d/1wg1kMk7U7ht3AmPpsPUimtz1FkVbu5BWjijjme2598s/edit#). Specifically it renames the non-standard `Sandbox` environment to `CODE` and updates any related usages 

- [x] update support-workers private config in S3
- [x] update support-frontend config in parameter store
- [x] test on code

PR #4938, opened 2023-05-02
Rename DEV to CODE in supporter product data

rupertbates:

This PR continues the work started in #4889 and described in [this document](https://docs.google.com/document/d/1wg1kMk7U7ht3AmPpsPUimtz1FkVbu5BWjijjme2598s/edit#) to standardise environment names across the supporter revenue stacks.

The supporter product data stack (and stripe patrons data stacks) have their own definition of stages (to avoid pulling in a large number of dependencies from the support-frontend stack I think?), which was not changed in previous PRs but is updated here.

### DIFF
--- a/supporter-product-data/cloudformation/create-table-stack.sh
+++ b/supporter-product-data/cloudformation/create-table-stack.sh
@@ -3,6 +3,6 @@
 # aws cloudformation delete-stack --stack-name supporter-product-data-tables-PROD
 aws cloudformation create-stack \
   --capabilities CAPABILITY_IAM  \
-  --stack-name supporter-product-data-tables-PROD \
+  --stack-name supporter-product-data-tables-CODE \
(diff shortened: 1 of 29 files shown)


supporter-product-data/cloudformation/create-table-stack.sh
```diff
@@ -3,6 +3,6 @@
 # aws cloudformation delete-stack --stack-name supporter-product-data-tables-PROD
 aws cloudformation create-stack \
   --capabilities CAPABILITY_IAM  \
-  --stack-name supporter-product-data-tables-PROD \
+  --stack-name supporter-product-data-tables-CODE \
```
rupertbates, line 6: Not sure that this script is actually needed but I'm definitely not keen on it pointing to PROD by default

LAKSHMIRPILLAI (approved):
LGTM 👍

PR #4943, opened 2023-05-04
Set-up Emotional benefit AB Test

LAKSHMIRPILLAI:
Set-up Emotional benefit AB Test

[**Trello Card**](https://trello.com/c/a2iskxQ7/1219-set-up-emotional-benefit-ab-test)

### Objective:
Test whether adding in an emotional benefit in to the benefits bullet points will increase conversion
### Context:
Following the latest copy changes and review, we have decided to change this to make more sense in a flow
### Changes:
Remove “you will” in title
change all benefits copy for variant

[**Figma link here**](https://www.figma.com/file/MQu0jaD3GR1C0CpdLLozOJ/MVP---optimisation---Q4?node-id=1672-87106&t=sIKUwyRJh6ZKQin9-0)

## Is this an AB test?
- [ x] Yes
- [ ] No
Test details:
this will be AB tested on desktop only, at the same time as mobile variant B
Metrics: This will be to test conversion rate

### (Control-Desktop)

image

### (Variant-Desktop-Below Threshold)

image

### (Variant-Desktop-Above Threshold)

image

### (Control-Mobile)

image

### (Variant-Mobile-same as control)

image

### (Control-Desktop-Australia)

image

### (Variant-Desktop-Australia-same as control)
image

### DIFF
--- a/support-frontend/assets/components/checkoutBenefits/checkoutBenefitsListData.tsx
+++ b/support-frontend/assets/components/checkoutBenefits/checkoutBenefitsListData.tsx
@@ -51,110 +51,114 @@ export const checkListData = ({
 }: CheckListDataProps): CheckListData[] => {
 	const maybeGreyedOutHigherTier = higherTier ? undefined : greyedOut;
 
+	const emotionalBenefits = [
+		{
+			icon: getSvgIcon(true),
+			text: (
+				<p>
+					<span css={boldText}>Fund fearless, independent reporting </span>
+					that's open and free for everyone
+				</p>
+			),
+		},
+		{
+			icon: getSvgIcon(true),
+			text: (
+				<p>
+					<span css={boldText}>Get a regular supporter newsletter </span> with
+					exclusive insights from our newsroom
+				</p>
+			),
+		},
+		{
+			icon: getSvgIcon(true),
+			text: (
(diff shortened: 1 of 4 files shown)


support-frontend/assets/components/checkoutBenefits/checkoutBenefitsListData.tsx
```diff
@@ -46,64 +47,114 @@ export const getSvgIcon = (isUnlocked: boolean): JSX.Element =>
...
-					<span css={boldText}>Ad-free reading. </span>Avoid ads on all your
-					devices
-				</p>
-			),
-			maybeGreyedOut: maybeGreyedOutHigherTier,
-		},
-	];
+	return isEmotionalBenefitTestVariant && !isAustralia
+		? [
+				{
```
marialani, line 56: This ternary is a little hard to read, could the conditional items be spread into the existing array instead? There's a good example on line 122

LAKSHMIRPILLAI: The new set of benefits are entirely different from the old one, that's why we cannot use the spread operator.

marialani (approved):
Looks great! 👍

PR #4945, opened 2023-05-05
Don’t send currency glyphs to braze

undergroundquizscene:
Trello Card: https://trello.com/c/V294lIz4/1258-send-currencies-to-braze-as-three-letter-codes-instead-of-currency-glyphs

We recently discovered that our braze template [SV_CN_WelcomeDay0_MASTER_v3](https://dashboard-01.braze.eu/engagement/templates_and_media/email_templates/5b75933b36dc781650d85641?id=5efc9e5b65ea093e62f8d7e7&locale=en), which is the template for the day-0 welcome emails for single contributions, was assuming the currency it was sent was [a three-letter currency code like USD or AUD](https://en.wikipedia.org/wiki/ISO_4217#List_of_ISO_4217_currency_codes), while we were sending a currency symbol (like $ or AU$). The template was switching on the currency symbol, and so was always falling through to the default cases, which meant for example that the UK terms and conditions were always sent instead of the regionalised terms and conditions that were intended to be sent to Australian and US users.

This PR updates payment API to send the three-letter currency codes instead.

This problem also affected recurring contributions, but was fixed in [PR 7336c0003774cbcf44774db708f9330703c294b5), merged in January 2022. That PR links to [a Trello card](https://trello.com/c/iuiXcPqV/258-make-thank-you-email-send-eg-gbp-instead-of-as-currency) which suggests that there was a change on the Braze side “around October 2021”, so I think recurring contributions were affected from October 2021 to January 2022.

I’ve checked all the relevant templates, and the others don’t contain this kind of currency switching logic, so I don’t believe this bug still exists anywhere else.


undergroundquizscene:
[Send currencies to braze as three-letter codes (instead of currency glyphs)](https://trello.com/c/V294lIz4/1258-send-currencies-to-braze-as-three-letter-codes-instead-of-currency-glyphs)

marialani (approved):
LGTM! Thanks for digging into this 👍

PR #4954, opened 2023-05-10
Narrow IsoCountry type to union of literals

undergroundquizscene:
The IsoCountry type is currently inferred as `string`, rather than the more specific union of string literals that it looks like was intended. From talking to other developers, I think this is leftover from the switch from flow to typescript, and that the type was correctly inferred with flow.

This PR updates it to be inferred as a union of literals, and propagates that change throughout the codebase. There are a few commits, which I’ve described separately:

## Restore emailRegexPattern comment beside variable

This comment used to be alongside this variable, but was displaced during the typescript migration (in commit 8dc45498b79671d86b699b65c8cf9e78612edc76).

## Correct typo: AcVanuatu -> Vanuatu

I noticed this typo while testing: it was added in commit ed1f79dc90d1a12dcde0eacdd3d972fd78923fbb, and from looking at [the Trello card](https://trello.com/c/MQCaMWce/3080-gw-list-of-countries-update-affected-by-covid19-request-v5) and [linked spreadsheet](https://docs.google.com/spreadsheets/d/1l0RhEYebyprf98eG86dTlgBvgDoCVVbAMkD1_AzzVDc/edit#gid=72359606) I don’t believe it was a deliberate change. I also can’t find anything elsewhere on the internet listing this as a correct/valid name.

## Narrow IsoCountry type to union of literals

Using an array of the possible values like this allows it to be used with zod as a zod enum – I initially tried a straightforward union type but that didn’t work with it.

I’ve tested locally that the supporter plus checkout, newspaper subscription checkout, and guardian weekly checkout all function correctly for a variety of countries and send valid requests to the backend.

## Correct IsoCountry errors

I’ve split these errors out from the later commit so they can be merged more easily, as the later commit is quite large and will take some more work. Valid IsoCountry values should all come from the list in helpers/internationalisation/country.ts, and these don’t.


i-hardy (approved):
This looks excellent, it'll be really good to have a much more reliable type for `IsoCountry` and to have the validation properly beefed up 👍

PR #4955, opened 2023-05-10
Enable soft opt-ins for single contributors

jonflynng:
## Why are we doing this?

We want to enable a customer's soft opt-ins when they make a single contribution so we can keep in touch with them and send them personalised emails.

## Changes

- Adds a new service in the `services` package of the support-payments-api to initialise an SQS client and send a message to the soft opt-ins queue
- Sends the message in parallel alongside other jobs in the PaymentBackend.scala file in [this line here](https://github.com/guardian/support-frontend/blob/main/support-payment-api/src/main/scala/backend/PaymentBackend.scala#L65](https://github.com/guardian/support-frontend/blob/main/support-payment-api/src/main/scala/backend/PaymentBackend.scala#L65) )
- Updates unit tests
- Add queue permissions to the CDK cloudformation

## to-do before merging
 - [x] Test in CODE
 - [x] Add feature flag
 - [x] Add feature switch JSON file to dev, code and prod


support-payment-api/src/main/scala/services/SoftOptInsService.scala
```diff
@@ -0,0 +1,97 @@
...
+class SoftOptInsService(sqsClient: AmazonSQSAsync, queueUrlResponse: Future[Either[SoftOptInsServiceError, String]])
+    extends StrictLogging {
+  def sendMessage(
+      message: Message,
+  )(implicit executionContext: ExecutionContext): EitherT[Future, SoftOptInsServiceError, Unit] = {
+    logger.info(s"Preparing to send message: ${message.asJson.noSpaces}")
+
+    message.identityId match {
+      case None =>
+        val error = SoftOptInsServiceError("Identity ID is None, soft opt-ins cannot be set")
```
jonflynng, line 27: @rupertbates  Will this happen often?

cdk/lib/payment-api.ts
```diff
@@ -131,13 +131,20 @@ export class PaymentApi extends GuStack {
...
             resources:
               this.stage === "PROD"
                 ? [emailSqsCodeArn.valueAsString, emailSqsProdArn.valueAsString]
                 : [emailSqsCodeArn.valueAsString],
           }),
+          new GuAllowPolicy(this, "SoftOptInsSqsMessages", {
+            actions: ["sqs:GetQueueUrl", "sqs:SendMessage"],
+            resources:
+              this.stage === "PROD"
+                ? [`arn:aws:sqs:${this.region}:${this.account}:soft-opt-in-consent-setter-queue-PROD`,]
```
jonflynng, line 145: is it better to import an exported ARN instead?

support-payment-api/src/main/scala/services/SoftOptInsService.scala
rupertbates: I don't think it should ever happen

cdk/lib/payment-api.ts
michaelbjacobson: I don't think there's anything wrong with this, and I _think_ it's fairly idiomatic GuCDK use.

michaelbjacobson (approved):
Looks good to me! 👍🏼

support-payment-api/src/main/scala/backend/AmazonPayBackend.scala
```diff
@@ -47,7 +38,8 @@ class AmazonPayBackend(
     val acquisitionsStreamService: AcquisitionsStreamService,
     val databaseService: ContributionsStoreService,
     val supporterProductDataService: SupporterProductDataService,
-    switchService: SwitchService,
```
jonflynng, line 50: this line without `val` was giving compiler errors?

rupertbates (approved):
Looks good 👍

PR #4956, opened 2023-05-11
Stop handling any visitId cookie / acquisitionData value

paulbrown1982:
Not passing around the visitId value, whether that comes from &acquisitionData, or picked up from Ophan's Tag's cookie set.

- [On Engine team board](https://trello.com/c/U1dNMW9r)

## Why are you doing this?
Ophan Tag's visitId is not actually passed onwards or back to the server via the AcquisitionsDataRow, so no app in this monorepo actually needs to be handling it. This ensures SR is not a blocker if Ophan decide to deprecate their visitId cookie.

## Is this an AB test?
- [ ] Yes
- [X] No

There are no design-related changes in this PR, just data collection and orchestration.

## Testing
- [x] Deployment to CODE and exploratory testing thereover, shows no errors


rtyley:
This sounds like a good idea to me! cc @guardian/ophan for awareness

rupertbates (approved):
👍

paulbrown1982:
The setting of a vsid cookie was removed in  https://github.com/guardian/ophan/pull/5452

PR #4960, opened 2023-05-11
mobile optimisation variantb footer line restore

paul-daniel-dempsey:
Before the Mobile Optimisation Variant 2 AB test went live their was a horizontal rule above the “Privacy Policy | Contact us | Help centre” container. 
It was actually a 1px border bottom on the main > section element above the footer element.

## Is this an AB test?
- [ ] Yes
- [x] No

### DIFF
--- a/support-frontend/assets/pages/supporter-plus-landing/supporterPlusLanding.tsx
+++ b/support-frontend/assets/pages/supporter-plus-landing/supporterPlusLanding.tsx
@@ -198,143 +198,138 @@ export function SupporterPlusLandingPage({
 	}
 
 	return (
-		<>
-			<>
-				<PageScaffold
-					id="supporter-plus-landing"
-					header={
-						<>
-							<Header>
-								<Hide from="desktop">
-									<CountrySwitcherContainer>
-										<CountryGroupSwitcher {...countrySwitcherProps} />
-									</CountrySwitcherContainer>
-								</Hide>
-							</Header>
-							<Nav {...countrySwitcherProps} />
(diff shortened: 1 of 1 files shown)


paul-daniel-dempsey:
[Restore missing horizontal rule above Mobile footer](https://trello.com/c/oh9pLTvZ/1294-restore-missing-horizontal-rule-above-mobile-footer)

undergroundquizscene (approved):
Looks good! Why did the footer need to be moved in the page structure to make it work?

Also I’ve just noticed that it doesn’t seem to be possible to focus any of the footer elements by tabbing. It’s the same before and after these changes, so it’s not relevant to this PR, but I think that’s something we should fix.

paul-daniel-dempsey:
> Looks good! Why did the footer need to be moved in the page structure to make it work?
> 
> Also I’ve just noticed that it doesn’t seem to be possible to focus any of the footer elements by tabbing. It’s the same before and after these changes, so it’s not relevant to this PR, but I think that’s something we should fix.

Footer was just a cleanup I noticed was originally there to avoid using empty tags around the whole component (looked neater). Maybe raise a separate PR for tabbing if not caused by the mobile optimisation PR's.

undergroundquizscene:
> Footer was just a cleanup I noticed was originally there to avoid using empty tags around the whole component (looked neater).

Ah, that makes sense – thanks!

> Maybe raise a separate PR for tabbing if not caused by the mobile optimisation PR's.

Yes, addressing it separately sounds good :)

PR #4965, opened 2023-05-15
Update scalafmt to 3.7.3

undergroundquizscene:
Spun off from Scala Steward PR https://github.com/guardian/support-frontend/pull/4919

Builds are failing on that PR, so I’m spinning off each dependency manually to see if it’ll pass on its own.


support-payment-api/src/main/scala/services/StripeService.scala
```diff
@@ -81,31 +81,32 @@ class SingleAccountStripeService(config: StripeAccountConfig)(implicit pool: Str
...
-        err => StripeApiError.fromThrowable(err, Some(config.publicKey)),
-        paymentIntent => {
-          logger.info(
-            s"Created Stripe Payment Intent with id ${paymentIntent.getId}, status ${paymentIntent.getStatus}",
-          )
-          paymentIntent
-        },
-      )
+    } else
+      {
```
paul-daniel-dempsey, line 85: Presume just formatting

paul-daniel-dempsey (approved):
All good.

support-payment-api/src/main/scala/services/StripeService.scala
undergroundquizscene: It’s an interesting change actually

PR #4974, opened 2023-05-15
Only group minor and patch non-aws PRs

undergroundquizscene:
Grouping all non-AWS PRs leads to big failing PRs like [4919](https://github.com/guardian/support-frontend/pull/4919) where it’s unclear which dependency is causing the failure. Major version changes in dependencies are less likely to pass without attention, so let’s remove those from the group to improve the odds of getting large PRs with successful builds. If no group matches, individual PRs will be raised.

This configuration is described [here](https://github.com/scala-steward-org/scala-steward/blob/main/docs/repo-specific-configuration.md). I think it’s just minor, patch, and major that we need to worry about, but I’m not certain: maybe we should include pre-release and build-metadata? Should be easy to see if that happens anyway, and we can fix then.


marialani (approved):
This looks good to me! 👍 
Also agree that we'd most likely only need to worry about minor and ~~major~~ patch versions

PR #5002, opened 2023-05-18
tooltip hidden paragraph overlaying components below

paul-daniel-dempsey:

Tooltip paragraph content/text box overlaps any content below it.
This appears to cause an un-selectable region below, as highlighted in the screenshot in blue.
ie. If you were to click in this region there would be no response, it could also possibly be an email box, GooglePay or ApplePay.

TooltipFocusOverlay

[**Trello Card**](https://trello.com/c/48pUZOtu/1318-tooltip-paragraph-text-box-overlay-repair)

## Solution

`Tooltip.tsx` component 'open' state checked and containig div component css set to display:none if closed.

## Is this an AB test?
- [ ] Yes
- [x] No

### DIFF
--- a/support-frontend/assets/components/tooltip/Tooltip.tsx
+++ b/support-frontend/assets/components/tooltip/Tooltip.tsx
@@ -59,8 +59,8 @@ const tooltipContainer = css`
 	}
 `;
 
-const tooltipCss = (hide: boolean) => css`
-	display: ${hide ? 'none' : 'block'};
+const tooltipCss = css`
+	overflow: hidden;
 	${textSans.small()};
 	background-color: #606060;
 	color: white;
@@ -242,7 +242,7 @@ export default function Tooltip({
 						}}
 						{...getFloatingProps()}
 					>
-						<div css={tooltipCss(!open)}>
+						<div css={tooltipCss}>
 							{children}
 							<Button
 								onClick={() => setOpen(false)}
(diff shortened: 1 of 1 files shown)


paul-daniel-dempsey:
[Tooltip : paragraph text box overlay repair](https://trello.com/c/48pUZOtu/1318-tooltip-paragraph-text-box-overlay-repair)

support-frontend/assets/components/tooltip/Tooltip.tsx
```diff
@@ -59,7 +59,8 @@ const tooltipContainer = css`
 	}
 `;
 
-const tooltipCss = css`
+const tooltipCss = (hide: boolean) => css`
```
GHaberis, line 62: I believe screen readers generally ignore anything with `display: none`, so this could affect the accessibility. An alternative is to apply `overflow: hidden` on the `<p>` paragraph's parent `<div>` element, as the issues here is the the paragraph overflows beyond the dimensions of it's parent element. By applying `overflow: hidden` this overflow is clipped.

GHaberis: You shouldn't need to apply this rule conditionally either.

GHaberis (approved):
Looks good 👍

paul-daniel-dempsey:
> Just wanted to check that this was tested with a screenreader and keyboard navigation? I'm sure it'll be fine, it's just best to make sure.

Thank's for the reminder, have done this now. Will do so going forwards in PR's. Used Safari (option F5 voice-over, tabbing around as per guide).  Checked against current production for differences.

PR #5009, opened 2023-05-19
support-payment-api: update java version to 11 (from 8)

undergroundquizscene:

I have updated the github action for the CI, as well as which amigo recipe is used by riff raff. There’s also a debian package specified in the build.sbt: I’m not sure what this is used for, as I would’ve expected the JRE to be provided by amigo.


undergroundquizscene:
I tried deploying this change to CODE but the new instances wouldn’t come up properly. I haven’t been able to work out why yet (and can’t see any logs which explain the problem), but I’ll try another couple of things to figure it out.

LAKSHMIRPILLAI:
Closing this PR as we already made the change in another PR

PR #5013, opened 2023-05-22
Remove Zuora Account Number from the AcquisitionDataRow

paulbrown1982:
The Zuora Account number is not used in any downstream process of `datalake.fact_acquisition_event`. As the Zuora Subscription ID is already there, there should be no benefit to having the Billing Account number too, so this PR removes it to reduce complexity.

Hopefully just not providing the `zuora_account_number` key in the BigQuery post will still work, as it's always been optional - and so not sent in the JSON payload for non-Zuora acquisitions. 

FYI - I staged a PR to remove the column but it reported that the whole table would be dumped to achieve that, so I'll check in with Data Tech - see: https://github.com/guardian/datatech-tf-modules/pull/164. 

- This PR also paves the way in another PR to add the Zuora Payment ID during checkouts which take payment immediately (e.g. Recurring Contribution and Supporter+; but not Paper or Weekly)

## Screenshots
- My changes were deployed at 11:54 BST, these are the CODE acquisitions since then. I did a S+ at 11:55, Single, then Newspape, then Weekly acquisition at 11:01, then finally the local integration test at the top at 12:17.
Screenshot 2023-05-24 at 12 32 21

## Testing
- [x] Tested in CODE

### DIFF
--- a/support-workers/src/test/scala/com/gu/acquisitions/BigQuerySpec.scala
+++ b/support-workers/src/test/scala/com/gu/acquisitions/BigQuerySpec.scala
@@ -81,7 +81,6 @@ class BigQuerySpec extends AsyncFlatSpec with Matchers with LazyLogging {
       Direct,
       Purchase,
       Some("subscription number"),
-      Some("account number"),
       Some("contributionId"),
       Some("paymentId1234"),
       List(QueryParameter("foo", "bar")),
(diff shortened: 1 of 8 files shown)


support-workers/src/test/scala/com/gu/acquisitions/BigQuerySpec.scala
```diff
@@ -81,7 +81,6 @@ class BigQuerySpec extends AsyncFlatSpec with Matchers with LazyLogging {
       Direct,
       Purchase,
       Some("subscription number"),
-      Some("account number"),
```
paulbrown1982, line 84: This local integration test also worked as expected.

paulbrown1982: See `paymentId12345` in the attached screenshot of CODE acquisitions.

rupertbates (approved):
👍

PR #5018, opened 2023-05-22
DO_NOT_MERGE: Always fail support workers, for testing

undergroundquizscene:
We’d like to see what step function failure looks like to the user in the fastest possible case, as part of a spike into changes related to direct debit. This change should not be merged into main, it’s just for testing in CODE.


undergroundquizscene:
This PR is no longer needed

PR #5022, opened 2023-05-25
Remove duplicate acquistion-events-api build

undergroundquizscene:
At the moment, there are frequently two deployments being attempted for the acquisition-events-api at the same time: one from the acquisition-events-api github action, and one from the support-lambdas github action. The latter always fails, I think because the acquisition-events-api now uses CDK but the support-lambdas github action doesn’t do the necessary CDK setup.

I don’t _think_ it makes sense for us to have this being built by two separate github actions, so I propose we remove it from the support-lambdas project (and thereby from the action).


michaelbjacobson (approved):
Looks good to me! 👍🏼

PR #5023, opened 2023-05-25
Fall back to inferred state when payment request button state is null

undergroundquizscene:
If the payment request button state is null, we may still have worked out a sensible state from geolocation. This makes a reasonable fallback if the inferred country matches the payment method country.


marialani (approved):
LGTM!

PR #4997, opened 2023-05-17
Perform silent authentication flow

kelvin-chappell:
## What are you doing in this PR?
This PR adds two new endpoints to populate cookies on the support domain with ID and access tokens if the user is signed in.
The endpoints and new code are isolated from the rest of the code at the moment, so it's safe to deploy and should have no effect on existing functionality.

https://trello.com/c/3zXUK4Pj/4388-support-oauth2-setup

## Why are you doing this?
This is for step (6) in the [new auth flow](https://github.com/guardian/support-frontend/wiki/Proposed-auth-flow).

## To test
Go to `/oauth/authorize`
  If signed in => redirected to `/` and have two new cookies called GU_ID_TOKEN and GU_ACCESS_TOKEN
  if not signed in => redirected to `/` and no new cookies

Tested in Code env.

PR #5024, opened 2023-05-25
Support Okta auth in MaybeAuthenticatedAction

kelvin-chappell:
This is the implementation of the [proposed new auth flow](https://github.com/guardian/support-frontend/wiki/Proposed-auth-flow) that uses Okta ID and access token cookies instead of legacy custom Identity cookies.  Its effect at this stage is to make some new auth cookies available on the support subdomain and pre-fill some form fields when the user is signed in.

[**Trello Card**](https://trello.com/c/UnCwidmM/4389-support-migrate-internal-routes-functionality-to-use-access-id-tokens)

The implementation is behind a feature switch so that it can be tested in Code and quickly switched back in Prod if anything goes wrong.  The feature switch can't be switched on in Prod until the next round of cookie re-permissioning next month enables the new cookies to be used.

## To test
Smoke test is:

1. Feature switch `Auth is by Okta tokens rather than legacy Identity cookies` off:
All functionality is same as before.  Forms pre-fill and submit correctly.  Subscriptions and contributions can be made successfully.

2. Feature switch `Auth is by Okta tokens rather than legacy Identity cookies` on:
When user is signed in, on requesting form view endpoints, new cookies `GU_ID_TOKEN` and `GU_ACCESS_TOKEN` are generated and form fields are pre-filled.
When user isn't signed in, forms can be viewed and submitted successfully without pre-filled fields and no new cookies generated.

3. No new errors in browser console or in Cloudwatch logs.

## Still to do

- [x] Configure feature switch in Prod **(required before merging this PR!)**
- [ ] Complete Okta config to give access to scopes that will provide ID claims
- [ ] Clientside support of calls to members-data-api
- [ ] Comprehensive test plan

## Related PRs

- #4997


support-frontend/app/controllers/AuthCodeFlowController.scala
```diff
@@ -61,7 +61,10 @@ class AuthCodeFlowController(cc: ControllerComponents, authService: AsyncAuthent
       "prompt" -> "none",
     )
     val authorizeUrl = s"${config.oauthAuthorizeUrl}?${toQuery(queryParams)}"
-    Redirect(authorizeUrl).addingToSession(SessionKey.state -> state, SessionKey.codeVerifier -> codeVerifier)
+    Redirect(authorizeUrl).addingToSession(
+      SessionKey.state -> state,
+      SessionKey.codeVerifier -> codeVerifier,
+    )
```
undergroundquizscene, line 67: Would it be much work to put the formatting changes in a separate commit? It’s not super important, but if it’s easy I’d prefer it.

support-frontend/conf/PROD.public.conf
```diff
@@ -6,9 +6,21 @@
...
+  api.url="https://idapi.theguardian.com"
+  production.keys=true
+
+  oauth {
+    client.id="0oa79m3ecvbJXDSqg417"
+    issuer.url="https://profile.theguardian.com/oauth2/aus3xgj525jYQRowl417"
+    audience="https://profile.theguardian.com/"
+    authorize.url="https://profile.theguardian.com/oauth2/aus3xgj525jYQRowl417/v1/authorize"
+    token.url="https://profile.theguardian.com/oauth2/aus3xgj525jYQRowl417/v1/token"
+    callback.url="https://support.theguardian.com/oauth/callback"
```
undergroundquizscene, line 20: How come the oath config wasn’t in this file already (like it was for CODE/DEV)? Has it not been used in production?

undergroundquizscene (approved):

support-frontend/conf/PROD.public.conf
kelvin-chappell: The Code/Dev config was introduced in #4997 and I should have added the Prod config then.  But it hasn't been used in Prod yet.

support-frontend/app/controllers/AuthCodeFlowController.scala
undergroundquizscene: Thanks!

support-frontend/conf/PROD.public.conf
undergroundquizscene: Perfect, that makes sense

pvighi (approved):
the code looks good to me.
Although I'm not super up to date on more client side things like our cookies and the auth redirection expected behaviour. If you need someone to verify those aspects you might want to get somebody to take a quick look as well

PR #5027, opened 2023-05-26
Two-tier geographical targeting for amounts tests

KaliedaRik:
## What does this change?
Marketing have asked us to supply them with the ability to set up a contribution amounts test for a given country, in addition to the tests for existing regions.

+ Note that this request represents a major refactor of the data model used to store amounts tests, which previously restricted amounts test data to just our 7 named regions.

Associated PRs in other repos
+ SAC: https://github.com/guardian/support-admin-console/pull/477
+ SDC: https://github.com/guardian/support-dotcom-components/pull/940
+ Support Frontend: https://github.com/guardian/support-frontend/pull/5027

This PR updates the default Choice card component in the Supporter+ landing page, to use the new V2 amounts model data. More importantly it makes changes on the server side to first check to see if an amounts test exists for the user's reported country code (eg: `FR` for France) and only falls back to the applicable regional amounts test if no country-specific amounts test can be found in the data.

For users redirected to the landing page from an Epic or Banner messaging test on the main site, the system should detect which amounts test variant the user saw in the Banner/Epic from data supplied in the URL string and display the same variant in the Choice card component

## Navigating through the PR
This PR touches 17 files, as follows:

**/support-frontend/app/admin/settings folders** (3 files)
Necessary changes to the Scala backend so that it can decode the new amounts-v2 model (stored in S3)

**/support-frontend/test/controllers folder** (1 files)
An additional Scala file with updates to stop a test failing

**/support-frontend/assets/helpers/contributions.ts file** (1 file)
We define the types for the new amounts-v2 model here

**/support-frontend/assets/helpers/abTests folders** (3 files)
The key `getAmounts` function is defined in the `helpers.ts`file - this is where we check to see if an amounts test has been defined in the url, or alternatively the user needs to be assigned to an amounts test dependent on their country/region location and, if necessary, assigned to a variant of that test.

**/support-frontend/assets/helpers/redux folders** (2 files)
The redux state for the page is initialised here

**/support-frontend/assets/pages folders** (2 files)
Additional redux management functionality here

**Other /support-frontend/assets/helpers folders** (5 files)
Mainly tweaking files so they recognise the new V2 amounts model

 ## How to test
Testing is in 3 parts:
+ Storybook
+ End-to-end testing in CODE to make sure that a country-specific amounts test displays correctly in Epics and Banners for a user in a country with an associated amounts test, that details of the correct amounts test get coded into the support site URL, and that clicking on the CTA will take the user to the support site which displays the same amounts data and choice as the user saw/made in the Epic/Banner. [Link to testing document](https://docs.google.com/document/d/1P-MS1Do_5Xqro6E6KKiw_7gID8c4U3L_CE_nAn-j2VI/edit).
+ A check to make sure the correct data is logged in the correct places and can be found using the normal data analysis tools.

## Screenshots
This PR does not change the display of the Supporter+ landing page


support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -224,16 +224,39 @@ function getAmountsTestParticipations(
...
+	}
+
+	const AMOUNTS = settings.amounts;
+
+	// Two-tier amounts - check for country test, then region test
+	let testArray = AMOUNTS.filter((t) => t.target === country);
+	// If there is a country level test we need to discount it if its isLive boolean is false
+	if (!testArray.length || !testArray[0].isLive) {
+		testArray = AMOUNTS.filter((t) => t.target === countryGroupId);
+	}
```
tjmw, line 238: Similar to the SDC PR, I wonder if we could use `find` here and lose the mutation:

```js
const matchedTest = AMOUNTS.find((t) => t.target === country && t.isLive) || AMOUNTS.find((t) => t.target === countryGroupId);
```

You could even create named functions for the two finds, which I think makes the country > region precedence nice and clear:

```js
const findMatchingCountryTest = (amounts, country) => amounts.find(t => t.target === country && t.isLive);

const findMatchingRegionTest = (amounts, countryGroupId) => amounts.find(t => t.target === countryGroupId);

const matchedTest = findMatchingCountryTest(amounts, country) || findMatchingRegionText(amounts, countryGroupId);
```

support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -224,16 +224,39 @@ function getAmountsTestParticipations(
...
 
-	const { test } = settings.amounts?.[countryGroupId] ?? {};
+	if (!settings.amounts) {
+		return null;
+	}
+
+	const AMOUNTS = settings.amounts;
+
+	// Two-tier amounts - check for country test, then region test
+	let testArray = AMOUNTS.filter((t) => t.target === country);
```
tjmw, line 234: I think `find` is the way to go here, but if you want to stick with `filter` you could always use array destructuring to get the first matched test, to avoid having to index into and array later, e.g.

```js
const [matchedTest] = AMOUNTS.filter((t) => t.target === country);
```

tjmw (commented):
Hey this is looking good! Sorry I've not had a chance to look over the whole thing yet, but leaving my comments so far in case they're useful.

support-frontend/assets/helpers/abTests/abtest.ts
KaliedaRik: I'll experiment for clarity

support-frontend/assets/helpers/abTests/abtest.ts
KaliedaRik: Will go with find, as in the other PRs

support-frontend/assets/helpers/abTests/helpers.ts
```diff
@@ -1,130 +1,335 @@
...
-		return FALLBACK_AMOUNTS[countryGroupId];
+		return getFallbackAmounts(countryGroupId);
+	}
+
+	const AMOUNTS = settings.amounts;
+
+	// Has user arrived at landing page with an URL detailing a specific amounts test variant?
+	if (Object.keys(abParticipations).length) {
+		for (const [test, variant] of Object.entries(abParticipations)) {
+			const testArray = AMOUNTS.filter((t) => {
```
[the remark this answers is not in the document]
KaliedaRik: Will recode

support-frontend/assets/helpers/abTests/abtest.ts
KaliedaRik: I've updated with the same `find` code as used in SDC (not the `findMatching functions` suggestion).

support-frontend/assets/helpers/abTests/abtest.ts
KaliedaRik: Code updated

support-frontend/assets/helpers/abTests/helpers.ts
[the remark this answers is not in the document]
KaliedaRik: Code updatd

support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -224,16 +224,37 @@ function getAmountsTestParticipations(
...
+	}
+
+	const { testName, liveTestName, isLive, seed, variants } = targetTest;
+
+	if (!variants.length) {
+		return null;
+	}
 
-	if (!test || !test.isLive) {
+	if (!isLive || variants.length === 1) {
```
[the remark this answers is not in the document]
KaliedaRik: The array of amounts tests generated in RRCP, stored in S3 and retrieved by SDC/SF will always include one-and-only-one test for each region (countryGroup). For these region tests there will always be a `control` variant, alongside 0-or-more AB test variants. Region tests are "always on"; the `isLive` boolean in these tests tells us whether an AB test is running (true) in which case there will be a choice to be made about which variant the user falls into, or whether there's no AB test running in which case the user should see the control variant. Every region test will have an obligatory "control" variant which is always the first variant in the test variants array.

This differs for country-specific tests. The `isLive` boolean in these tests tells us whether to use the country-specific data (true) in an AB test, or to fall back to that country's region test (false).

The purpose of the code here is to determine whether there's an active AB amounts test running for that user, given their geographical location (country or region). If there is an active AB test then the test name/variant gets coded as a key:value pair into an object which then amalgamates into a `participations` object which includes details of all the current (often hardcoded) tests being run for that user across the entire support site.

Note that this code is very similar to the code in the sibling `helpers.ts` file, which is where the data actually gets returned to the SF `priceCards` component. I have no idea why our predecessors chose to do things this way, but I am not going to argue with their wisdom!

support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -224,16 +224,37 @@ function getAmountsTestParticipations(
...
 
-	const { test } = settings.amounts?.[countryGroupId] ?? {};
+	if (!settings.amounts) {
+		return null;
+	}
+
+	const AMOUNTS = settings.amounts;
+
+	// Two-tier amounts - check for live country test, else get a region test
+	let targetTest = AMOUNTS.find((t) => t.target === country && t.isLive);
```
GHaberis, line 234: So it looks like the model now has an array of tests. Question, can a country test be configured in the RRCP to target multiple countries? If so will that result in multiple items in the array of tests, one for each country the test is targeted towards? It appears so from this look up.

If this is the case a suggestion for a future change to the model would be for the test to have a single entry in the array of tests and for the `target` property to be an array of country codes.

So the lookup could be

```
let targetTest = AMOUNTS.find((t) => t.target.includes(country) && t.isLive);
```

I'm suggesting this as I'd be concerned that this amounts array could become pretty big if we start targeting multiple countries (eg. VAT countries), and we're shipping all of these tests to the browser in the HTML.

support-frontend/assets/helpers/abTests/helpers.ts
```diff
@@ -1,130 +1,333 @@
...
-		},
-		ANNUAL: {
-			amounts: [60, 120, 240, 480],
-			defaultAmount: 120,
-		},
+export const FALLBACK_AMOUNTS: AmountsTest[] = [
+	{
+		testName: 'FALLBACK_AMOUNTS__GBPCountries',
+		liveTestName: '',
+		isLive: false,
```
GHaberis, line 14: Is the purpose of these defaults so we fallback and render something if the amounts data from config isn't available? If so, should we be setting `isLive` to `false`? Won't this block them being rendered?

support-frontend/assets/helpers/abTests/abtest.ts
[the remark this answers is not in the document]
KaliedaRik: I would argue strongly against that - it is what we had before and what this series of PRs is moving away from. Also, there is no guarantee that regions will remain static going forward and it makes sense to me to keep hardcoded regions as far away as possible from the data model. Does that make sense?

support-frontend/assets/helpers/abTests/abtest.ts
KaliedaRik: `can a country test be configured in the RRCP to target multiple countries?` - no. An amounts test will target either a single region, or a single country.

`and we're shipping all of these tests to the browser in the HTML` - yes. They all end up in the window.guardian.settings.amounts object/array. I can't see a way around this, given that the Supporter+ landing page includes a region selector where users can change the currency they wish to donate in. I think there may be a way to restrict the data we push into window.guardian.settings.amounts to just the regions and the user's current country if an amounts test exists for that country?

support-frontend/assets/helpers/abTests/helpers.ts
KaliedaRik: We only have hardcoded default amounts for each region. And for region amounts test, setting the `isLive` flag to false tells the system to use the control variant for that test.

support-frontend/assets/helpers/abTests/abtest.ts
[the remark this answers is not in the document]
tjmw: > I would argue strongly against that - it is what we had before and what this series of PRs is moving away from. Also, there is no guarantee that regions will remain static going forward and it makes sense to me to keep hardcoded regions as far away as possible from the data model. Does that make sense?

I don't follow I'm afraid! I just meant that a JS object feels like a more appropriate data structure than an array if you want to enforce some kind of uniqueness, and if you're often looking up that item in the data structure based on a given property. But happy to continue to conversation offline as it doesn't sound like this is something that'll change in this PR.

support-frontend/assets/helpers/abTests/abtest.ts
GHaberis: > `can a country test be configured in the RRCP to target multiple countries?` - no. An amounts test will target either a single region, or a single country.

Given a Country Test can only target one country then my original optimisation suggestion may be a moot point.

> `and we're shipping all of these tests to the browser in the HTML` - yes. They all end up in the window.guardian.settings.amounts object/array. I can't see a way around this, given that the Supporter+ landing page includes a region selector where users can change the currency they wish to donate in. I think there may be a way to restrict the data we push into window.guardian.settings.amounts to just the regions and the user's current country if an amounts test exists for that country?

I did think about whether we need to ship all tests to the browser, or whether they could be filtered on the server. Originally I suspected they'd all need be available on `window.guardian.settings.amounts` so we can handle when users come via amounts tests in the Banner/Epic, as I'd assumed the amounts test allocation on the checkout was based on the test/cohort present in the `acquisitionData` and we did a lookup using these values, however it looks like we just replicate the AB test allocation logic on the checkout and the banner/epic - so as long as they get the same input (amounts config and user MVT id) the outcome will be the same.

support-frontend/assets/helpers/abTests/helpers.ts
GHaberis: Got it 👍

support-frontend/assets/helpers/abTests/abtest.ts
GHaberis: I think we could do with assessing the impact of shipping a large list of amount tests to the browser and whether it can be optimised, I can see this list getting pretty big (particularly with the VAT proposal), one impact is it'll bloat the data in our cloudwatch logs significantly.

support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -224,16 +224,37 @@ function getAmountsTestParticipations(
 		return urlTest;
 	}
 
-	const { test } = settings.amounts?.[countryGroupId] ?? {};
+	if (!settings.amounts) {
+		return null;
+	}
+
+	const AMOUNTS = settings.amounts;
```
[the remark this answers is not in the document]
KaliedaRik: Code updated

support-frontend/assets/helpers/abTests/abtest.ts
KaliedaRik: Hi George. Question: how much of this problem will go away if/when we get into a world of 2-step checkout, where users coming from a Banner/Epic interaction will go directly to the step-2-checkout page (where we won't have to care about all the amounts tests) and the amounts tests only have to be supplied for users who go directly to the step-1-landing page?

I do like the idea of filtering tests on the server, but I don't know how to implement such a filter in Scala

`I think we could do with assessing the impact of shipping a large list of amount tests to the browser` - grateful for details on how I could perform this testing

GHaberis: @KaliedaRik I'm not sure the 2 step checkout will help here, as looking at designs even if we drop a user on Page 2 there'll be the option to adjust their choice by stepping back to Page 1 where the amounts will need to be available. Of course. a solution here is to remove that ability to adjust the amounts for users landing directly on Page 2, but that's a whole discussion.

It'd be interesting to asses the impact of delivering something like the Contributions Only Country Targeted tests in the Amounts Test config, as there'll be around ~45 country targeted tests here. I was thinking we could test on CODE by mocking the `amounts-v2` JSON in S3 to have this volume of tests. We could then test this branch (using `amounts-v2`  vs main (using `amounts-v1`) using Chrome performance profiler?

support-frontend/assets/helpers/abTests/abtest.ts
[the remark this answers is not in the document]
tomrf1: I think an array is the correct choice, despite the additional complexity.
As Rik says, we're moving away from a simple one-test-per-region lookup, to something with more sophisticated targeting.
Though I wonder if the ordering of the array of tests should be used, instead of scanning the array twice. And the RRCP puts regional tests last in the priority order. In theory other types of targeting would then be possible in future.

tomrf1 (approved):
👍 
I'd still like to revisit how we handle the array of tests - I think the ordering within the array is going to be quite confusing for people maintaining this codebase.
But if it works then 🚢 it

PR #4917, opened 2023-04-24
Add tote bag to benefits in Australia

LAKSHMIRPILLAI:
## What are you doing in this PR?
Add Tote bag to benefits in Australia 

[**Trello Card**](https://trello.com/c/AyN6IYu4/1237-🇦🇺-tote-bag-added-to-benefits-in-australia-engineering)

## Why are you doing this?
Copy change on checkout - only in AUS
Copy is:
Limited-edition Guardian Australia tote bag. Offer ends 31 May
Context:
Australia 10 years

## Is this an AB test?
- [ ] Yes
- [ x] No

## Before change(Below Threshold)

image

## Before change(Above Threshold)
image

## After change(Below Threshold -for mobile)

image

## After change(Above Threshold -for mobile)

image

## After change(Below Threshold -for tablet and desktop)

image

## After change(Above Threshold -for tablet and desktop)

image

PR #5033, opened 2023-05-30
Revert "Add tote bag to benefits in Australia"

undergroundquizscene:
Reverts guardian/support-frontend#4917, as the 10-year anniversary campaign is coming to an end. Should be merged shortly before 3pm UTC+1 on the 31st of May.


paul-daniel-dempsey (approved):
Looks good, thanks linking the original PR to check, nice and quick to double check.

PR #5034, opened 2023-05-30
Increase request timeout when calling Identity

undergroundquizscene:
Spoke with Identity about our recent timeouts (https://chat.google.com/room/AAAAFdv9gK8/nb8k33IJzV0). They felt that 3 seconds might not be long enough for Okta, and increasing to 5 seconds sounds like a reasonable next step.

I’m worried our retries are flawed in some way, and can never succeed if the initial call fails. I haven’t had a chance to look at it in more detail yet though.


GHaberis (approved):
👍

GHaberis:
@undergroundquizscene I noticed a `payment-api` test is failing (`should return successful payment response with guestAccountRegistrationToken if available`), this looks like it could be linked to the changes here, possibly something needs updating in the `mockIdentityService`?

undergroundquizscene:
> @undergroundquizscene I noticed a `payment-api` test is failing (`should return successful payment response with guestAccountRegistrationToken if available`), this looks like it could be linked to the changes here, possibly something needs updating in the `mockIdentityService`?

Hmmm, I’m not sure – I’ll take a look. I think it’s more likely to be the transient payment-api build failure we saw across a number of PRs the other day. I'll rebase on main and see if it goes away.

PR #5038, opened 2023-05-31
Remove some countries from the GW checkout drop-downs

paulbrown1982:
The business have asked us to remove the following countries from the GW checkout drop-downs.


GHaberis (approved):
Looks good 👍

PR #5045, opened 2023-06-02
Update support-frontend to NODE 18

LAKSHMIRPILLAI:

This PR updates support-frontend from Node 16-> Node 18. This will allow us to take updates of dependencies where latest versions are no longer compatible with older versions of Node.

[**Trello Card**](https://trello.com/c/a8SnsqZ1/1342-update-support-frontend-to-node-18)

Along with this , we have updated 
1. @guardian/consent-management-platform to  13.1.0 from ^13.0.2
2. @guardian/libs to 14.0.0 from ^5.1.0


undergroundquizscene (approved):
The [recommendation on JS dependencies](https://github.com/guardian/recommendations/blob/main/dependencies.md) was recently updated to suggest pinning dependencies in deployed applications. Should we take this chance to pin these dependencies (remove the caret so we depend on one specific version rather than a range)?

PR #5061, opened 2023-06-08
Document uncertainty about sbt dependencyOverrides

emdash-ie:
I’m a little nervous about our use of dependencyOverrides, and whether
it’s definitely safe for us to do. However, we do it in a few places: it
seems to be a standard practice of ours. I wanted to document this here
following a discussion with a senior engineer who shared my concerns:
to record that we’re uncertain about it, but it has worked fine so far.

### DIFF
--- a/support-frontend/README.md
+++ b/support-frontend/README.md
@@ -1,8 +1,30 @@
+# Support-frontend
+
 Frontend for the new [supporter platform](https://support.theguardian.com/).
 
-### SSH
+## SSH
+
 Use SSM Tunnel - [ssm-scala](https://github.com/guardian/ssm-scala#enabling-ssm-tunnel):
 
 ```
 ssm ssh --profile membership --tags frontend,support,CODE -a -x --newest --ssm-tunnel
 ```
+
+## Dependency Overrides in build.sbt
+
+We use [sbt’s
+dependencyOverrides](https://www.scala-sbt.org/1.x/docs/Library-Management.html#Overriding+a+version)
+in a couple of places to raise the version of a dependency deep in our
+dependency tree, where we don’t want to raise the versions of intermediate
+dependencies.
+
+Using `dependencyOverrides` puts the onus on us to guarantee that our
+dependencies are compatible with the newer version of the package we’re forcing,
+which we do by relying on our builds to fail, or our testing to reveal issues.
+It’s unclear if this is always enough: it seems plausible that overriding
+dependencies like this might cause subtle bugs that we don’t discover for some
+time.
+
+That said, we haven’t encountered any issues with this yet: everything seems to
+have worked fine. This notice is here to document our uncertainty about the risk
+involved in overriding dependencies like this.


support-frontend/README.md
```diff
@@ -1,8 +1,30 @@
...
+Using `dependencyOverrides` puts the onus on us to guarantee that our
+dependencies are compatible with the newer version of the package we’re forcing,
+which we do by relying on our builds to fail, or our testing to reveal issues.
+It’s unclear if this is always enough: it seems plausible that overriding
+dependencies like this might cause subtle bugs that we don’t discover for some
+time.
+
+That said, we haven’t encountered any issues with this yet: everything seems to
+have worked fine. This notice is here to document our uncertainty about the risk
+involved in overriding dependencies like this.
```
kelvin-chappell, line 30: Actually there's never any guarantee of what transitive dependency version you're going to get with SBT.
By default, there's an eviction process that removes all but the latest version of common transitive dependencies.  So if you have two dependencies `A` and `B` and they have a common dependency `C`, and `A` is dependent on `C v1` while `B` is dependent on `C v2`, `C v1` will be evicted and the program will use `C v2`, regardless of what effect that might have on dependency `A`.
By using `dependencyOverrides` you can tell the program explicitly to use `C v1` and evict other versions of the same dependency, which means that the unpredictable behaviour shifts from dependency `A` to dependency `B`.  So `dependencyOverrides` doesn't necessarily make the behaviour of the system more unstable; it just makes it potentially different.

undergroundquizscene: > You mean if you're specifying a version that no dependency actually declares it's dependent on?

Yes, exactly. We’ve done this a couple of times where the declared dependency has a vulnerability, and we want a quicker way of updating the transitive dependency than updating the intermediate dependencies on the dependency chain. (E.g. we depend on `A v1` which depends on `B v1` which depends on `C v1`, Snyk identifies a vulnerability in `C v1` so we use dependencyOverrides to increase the version of `C` used without increasing that of `A` or `B`.)

> In that case it's still not necessarily less stable because the effect depends on the worst combination of dependencies.

I don’t know what you mean by this – would you mind expanding on it?

> I do think you make an important point though that any change to explicit dependencies can change the ultimate list of dependencies and the behaviour of the program in ways that might not be obvious and needs careful testing. Perhaps running `dependencyList` before and after changing dependencies would also help.

Yes, I had forgotten about `dependencyList` – definitely helpful to see the full list of transitive dependencies before and after.

kelvin-chappell: > > In that case it's still not necessarily less stable because the effect depends on the worst combination of dependencies.
> 
> I don’t know what you mean by this – would you mind expanding on it?

I mean if for example the interaction between `A` and `C v2` has a fatal effect on the program because `A` is actually depending on `C v1` which was evicted, then swapping `C v2` for `C v0.5` won't necessarily have a worse effect.  What I'm really saying is that `dependencyOverrides` for a transitive dependency isn't necessarily worse than a change to a direct dependency even when it's specifying a dependency that no direct dependency actually depends on.

undergroundquizscene: Ok, I follow! Using dependency overrides to specify a newer version of a non-common transitive dependency (one that appears only once in the dependency tree) has the same effect as specifying the newer version as a direct dependency. So it’s not the mechanism used (dependencyOverrides vs direct dependencies), it’s having the dependency in multiple places in the tree at different versions at all.

So I’ll rewrite this message in the readme to make that clear: what I’m worried about is us forcing a newer version of a dependency deep in the tree (rather than upgrading the intermediate dependencies), whatever mechanism we use to do it.

kelvin-chappell: 👍

emdash-ie:
Closed as I’ve moved out of the team and I don’t feel this is important enough to hand over to someone else.

PR #5062, opened 2023-06-09
Enter FailState if Zuora reports missing postcode

emdash-ie:
A [recent support-workers execution](https://eu-west-1.console.aws.amazon.com/states/home?region=eu-west-1#/v2/executions/details/arn:aws:states:eu-west-1:865473395570:execution:support-workers-PROD:1578622a-578c-484f-b2bb-3a66c4f8c69a) showed us that errors are sometimes happening where Zuora complains about a partially completed address without a postcode, and that we may not be very aware of them because they’re entering the CheckoutFailure state rather than the FailState state.

However, they’re definitely caused by a problem with our code: we have full control over what address information is sent and aren’t currently requesting a postcode from the user. (We think this is happening because we’re sending a “state” value inferred from geolocation, and it may be limited to Direct Debit payments.)

I haven’t found a way to easily query support-workers executions to see how often this is happening, so I want to make them count as failures so we see them when they happen. This’ll help us work out how many people are affected, and possibly narrow down exact criteria for reproducing the error.


paul-daniel-dempsey (approved):
Looks good, isolating those postcode failures will really help, thanks

GHaberis (approved):
Good plan 👍

PR #5064, opened 2023-06-12
Use better link in support-workers failure alarms

undergroundquizscene:
The old link gives me a 404, and our first step in alarm investigations is always to look at the list of failed executions – having a link in the alert should help investigate more quickly.


undergroundquizscene:
[Put typical log search links in alert emails](https://trello.com/c/rw0QCEew/1379-put-typical-log-search-links-in-alert-emails)

LAKSHMIRPILLAI (approved):

PR #5065, opened 2023-06-13
Don’t send state to backend unless required

undergroundquizscene:
We’ve noticed in at least [one recent support-workers execution](https://eu-west-1.console.aws.amazon.com/states/home?region=eu-west-1#/v2/executions/details/arn:aws:states:eu-west-1:865473395570:execution:support-workers-PROD:1578622a-578c-484f-b2bb-3a66c4f8c69a) that Zuora sometimes complains if the state in the address is set but the postcode isn’t.

Since we only _need_ the state on the backend for certain country values, this change only sends it when it’s needed. It’s not clear yet how many people are affected (from testing it seems to be just UK users attempting to use direct debit where Fastly’s geolocation has determined a state for them), but I reckon it’s worth fixing the bug even though we’re not certain of the impact.

### DIFF
--- a/support-frontend/assets/pages/supporter-plus-landing/setup/legacyActionCreators.ts
+++ b/support-frontend/assets/pages/supporter-plus-landing/setup/legacyActionCreators.ts
@@ -44,6 +44,7 @@ import type {
 	IsoCountry,
 	StateProvince,
 } from 'helpers/internationalisation/country';
+import { shouldCollectStateForContributions } from 'helpers/internationalisation/shouldCollectStateForContribs';
 import { Annual, Monthly } from 'helpers/productPrice/billingPeriods';
 import {
 	setAmazonPayFatalError,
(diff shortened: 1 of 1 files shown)


undergroundquizscene:
[Don’t send inferred state in /create request if payment method is Direct Debit](https://trello.com/c/2FihV4mD/1377-dont-send-inferred-state-in-create-request-if-payment-method-is-direct-debit)

undergroundquizscene:
I wrote a script to query support-workers executions to see how frequently this was happening. From 31/5/2023 to 13/6/2023 inclusive it happened 73 times, an average of 5.2 per day.

I haven’t looked at all of the executions in detail yet, but it looks like at least a couple of them are re-attempts by the same person, and it’s likely that at least some people successfully paid using a different payment method after getting the error. However, it’s still a pretty high frequency for this to be happening, and I wonder if there’s anything else we might be missing.

LAKSHMIRPILLAI (approved):

undergroundquizscene:
> I wrote a script to query support-workers executions to see how frequently this was happening. From 31/5/2023 to 13/6/2023 inclusive it happened 73 times, an average of 5.2 per day.
> 
> I haven’t looked at all of the executions in detail yet, but it looks like at least a couple of them are re-attempts by the same person, and it’s likely that at least some people successfully paid using a different payment method after getting the error. However, it’s still a pretty high frequency for this to be happening, and I wonder if there’s anything else we might be missing.

I checked how many unique email addresses there were in this group, to screen out some re-attempts: there were 35.

PR #5076, opened 2023-06-20
Fix support-workers failed execution error message

undergroundquizscene:
I forgot to include the “!Sub” when I added the Stage interpolation! :facepalm:


rupertbates (approved):
Is this still needed?

undergroundquizscene:
> Is this still needed?

Yes, I believe so – thanks for the approval! We could plausibly take the URLs out and just put them in our alarm spreadsheet, but I see no harm in keeping it

PR #5077, opened 2023-06-20
Don’t guess state from geolocation data

undergroundquizscene:
At the moment, we guess a user’s state on page load based on geolocation data provided by Fastly. This allows users to avoid needing to fill out the state field on the form.

However, the guessed state is wrong relatively often, and has been causing us trouble. For example, a US user may have a geolocation-guessed state that is in Canada (presumably if they are near the border), and then our client-side validation needs to check that the current state value fits with the current country value.

Since it doesn’t seem like too much effort to ask the user to go to to enter their state along with the other info we ask them for, this commit stops the site guessing the user’s state from Fastly’s geolocation data.

I tested this by hardcoding the geolocation data on the backend (see commit a2d337309a36c148c6765619ebc28d0119629540) and confirming that the state wasn’t filled in the redux state (at page > checkoutForm > billingAddress > fields > state). I also submitted without selecting a state and confirmed that I get a client-side validation error.

### DIFF
--- a/support-frontend/assets/helpers/internationalisation/country.ts
+++ b/support-frontend/assets/helpers/internationalisation/country.ts
@@ -913,14 +913,9 @@ function detect(
 	return country;
 }
 
-function detectState(country: IsoCountry): Option<StateProvince> {
-	return stateProvinceFromString(country, window.guardian.geoip?.stateCode);
-}
-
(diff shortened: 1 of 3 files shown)


i-hardy (approved):
Looks good 👍

PR #5087, opened 2023-06-21
Remove legacy localised amounts code

KaliedaRik:
A long time ago we ran a hardcoded AB test to see if using local currencies in the contributions landing page amounts card could improve conversions. This PR removes that legacy code.

[**Trello Card**](https://trello.com/c/2GvdrD7f)

## Why are you doing this?
To simplify the code base and remove technical debt

## Is this an AB test?
- [ ] Yes
- [x] No


support-frontend/assets/helpers/redux/commonState/reducer.ts
```diff
@@ -88,40 +64,11 @@ export const commonSlice = createSlice({
 		setContributionTypes(state, action: PayloadAction<ContributionTypes>) {
 			state.settings.contributionTypes = action.payload;
 		},
-		setCurrencyId(state, action: PayloadAction<boolean>) {
+		setCurrencyId(state) {
 			state.internationalisation.currencyId = getLocalisedCurrencyId(
 				state.internationalisation,
```
i-hardy, line 69: Given that `getLocalisedCurrencyId` is now so simple, I'd suggest removing it altogether
```suggestion
		setCurrencyId(state) {
			state.internationalisation.currencyId = state.internationalisation.defaultCurrency
```
It's also worth thinking about whether we need this reducer at all anymore, given that it's just setting information in one bit of the state to another bit of the state.

i-hardy (approved):
Just one small suggestion but otherwise this looks great, thanks so much for excising this extremely confusing old feature!

support-frontend/assets/helpers/redux/commonState/reducer.ts
KaliedaRik: Many thanks Imogen! I've checked the usage of `internationalisation.currencyId` - it appears 13 times across 8 files, which include files involved in paper subscription checkout, subscription forms for Guardian Weekly, etc. On reflection, I don't want to touch those files now but instead just implement your suggested change above and leave it at that.

PR #5097, opened 2023-06-28
Upgrade cdk

tomrf1:
This causes some changes to support-frontend and payment-api cloudformation templates, which leads to some resources being recreated.
I've tested in CODE by making single and recurring contributions.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.ts.snap linguist-generated=true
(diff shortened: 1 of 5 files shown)


.gitattributes
```diff
@@ -0,0 +1 @@
+*.ts.snap linguist-generated=true
```
tomrf1, line 1: this tells github to collapse the snapshot files by default

undergroundquizscene (approved):

PR #5098, opened 2023-06-29
Do not auto-select a payment method when switching contribution frequency

i-hardy:

This PR removes the existing feature where we automatically pre-select a payment method for users when they switch their contribution frequency, eg. switching from a monthly to a single contribution. It also reverts the selected payment method back to 'None' when the user switches contribution frequency, in order to avoid users attempting to pay with a disallowed payment method, eg. direct debit with a single contribution.

[**Trello Card**](https://trello.com/c/R3YkoxDd)

## Why are you doing this?

This is to reduce user confusion when selecting a payment method, and improve the overall user journey.


i-hardy:
[Remove default payment method on Support checkout](https://trello.com/c/R3YkoxDd/1242-remove-default-payment-method-on-support-checkout)

undergroundquizscene (approved):

PR #5115, opened 2023-07-07
Print debug info earlier

undergroundquizscene:
# Summary

Recently, I’ve noticed that some errors mean we don’t see the nicely formatted debug info. I think this is because the errors happen before we get to the line that’s currently doing that logging. Moving that earlier in the execution flow should make our debugging a bit nicer.

# Testing

To test this, run locally or deploy to CODE and check out using an email address whose domain is `qq.com`. In the support-frontend logs in CloudWatch, you should be able to see the log message that I’ve moved where the debug info is pretty printed. Without the change, the debug info would only be accessible in the request logging, where it is escaped inside a JSON string.


rupertbates (approved):
👍

PR #5116, opened 2023-07-07
Provide better feedback to user when their email provider is rejected

undergroundquizscene:
### Fix detection of specific identity errors

I broke this in commit 2020275389c3812d25992d8bca7c3655b017fa16, when I modified the error messages to add information about which endpoint was called. I forgot or didn’t realise that we were later checking those messages to check the specific kind of error.

To reduce the chance of this happening again, I’ve moved the specific error recognition to the JSON parsing, and used specific case classes for the specific errors (avoiding redundant info and checking error messages multiple times).

### Improve naming of specific identity errors

“bad_email_address” and “invalid_email_address” were too similar, and the latter seemed like the wrong name for the email provider having been rejected.

### Add comma to error message

These error messages read a bit weirdly without the comma, I reckon. Might be worth checking if this is the tone the site should strike at all, though?

## Testing

Run locally or deploy to CODE, and attempt to check out with an email address whose domain is `@qq.com`. You should see an error message in the frontend which tells you to use an email address from a different provider.

### DIFF
--- a/support-frontend/app/actions/CustomActionBuilders.scala
+++ b/support-frontend/app/actions/CustomActionBuilders.scala
@@ -8,7 +8,7 @@ import com.gu.aws.{AwsCloudWatchMetricPut, AwsCloudWatchMetricSetup}
 import com.gu.monitoring.SafeLogger
 import com.gu.monitoring.SafeLogger.Sanitizer
 import com.gu.support.config.Stage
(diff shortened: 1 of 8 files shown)


undergroundquizscene:
[Provide better feedback to user when their email address provider is rejected by Identity/Okta](https://trello.com/c/UQEnDXlh/1364-provide-better-feedback-to-user-when-their-email-address-provider-is-rejected-by-identity-okta)

PR #5129, opened 2023-07-12
acquisition-events-api: Try increasing memory to prevent API Gateway timeouts

undergroundquizscene:
Increasing the memory available to the lambda might help it run within the 30-second API Gateway timeout.

### DIFF
--- a/cdk/lib/__snapshots__/acquisition-events-api.test.ts.snap
+++ b/cdk/lib/__snapshots__/acquisition-events-api.test.ts.snap
@@ -6,6 +6,7 @@ exports[`The Acquisition Events API stack matches the snapshot 1`] = `
     "gu:cdk:constructs": [
       "GuDistributionBucketParameter",
       "GuApiLambda",
+      "GuApiGateway5xxPercentageAlarm",
       "GuAlarm",
     ],
     "gu:cdk:version": "TEST",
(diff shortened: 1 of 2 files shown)


undergroundquizscene:
I should deploy this to CODE and test it there

undergroundquizscene:
Testing in CODE seems to show that doubling the memory roughly halves the execution time. This is a bit surprising, because the amount of memory _used_ by the lambda doesn’t really change. Maybe the memory affects the startup procedure? I’ll check the logs.

Screenshot 2023-07-13 at 11 00 07

undergroundquizscene:
The logs show the following approximate execution times:

| Memory (MiB) | Time to first app log (s) | Time from first app log to last app log (s) |
|--------------|---------------------------|---------------------------------------------|
| 512          | 22.7                      | 12.1                                        |
| 512          | 20.5                      | 11.3                                        |
| 1024         | 12                        | 5.8                                         |
| 1024         | 12.4                      | 5.6                                         |

So, oddly, the increased memory seems to improve both the startup and the app execution times, both made twice as fast.

tomrf1 (approved):
This is a feature of aws lambda -
>> Adding more memory proportionally increases the amount of CPU
https://docs.aws.amazon.com/lambda/latest/operatorguide/computing-power.html

👍 Good fix while we wait to decommission this lambda

undergroundquizscene:
I should also reinstate the alarm, since I believe this will fix our issue with false positives.

undergroundquizscene:
[Increase acquisition-events-api lambda memory to avoid API Gateway timeouts](https://trello.com/c/Q2c9KoKc/1452-increase-acquisition-events-api-lambda-memory-to-avoid-api-gateway-timeouts)

undergroundquizscene:
Tried out higher memory values (2048 and 4096 MiB), and they lead to a decrease in execution time (to approximately 10s and 8s), but an increase in cost (to about 22 GB seconds and 34 GB seconds per execution). Evaluating whether this cost increase would be worth it (monthly cost for these lambda executions in June was $0.75, so even doubling cost would be cheap) would involve checking with apps teams about how much of a problem the length of these requests is, and I’m not sure I want to go to that effort for this.

For now let’s go with the 1024 increase, which avoids timeouts.

PR #5132, opened 2023-07-13
Add endpoints for calling PaperRound’s API

undergroundquizscene:
To expand paper delivery outside the M25 using PaperRound, we’ll need to offer users a choice of delivery agents for their postcode. This change makes a prototype API on the support-frontend backend which calls PaperRound’s test API to check the agents available for a certain postcode. (I’ve also added the other API endpoints to the service while here.)

### Model PaperRound API and write JSON decoders

- Fixtures taken from responses from the test API

### Write separate types for support-frontend

I wanted to change the structure received by the frontend, so I’ve used separate types for the API called by the frontend. I didn’t want to have decoders and encoders for the PaperRound types that don’t round-trip. It also seems a good idea to isolate the dependency on PaperRound’s types from the frontend, but I’m less confident about that one.

### Write custom json encoding for frontend

The derived encoding of the `GetAgentsResponse` ADT doesn’t seem very nice to handle from TypeScript, so I’ve written a custom one that I think will be nicer to consume from TypeScript.

### Add extra test step to CI workflow

We weren’t running all our tests in CI, which I noticed when I added the deserialisation tests. I think we should be running them all, so I’ve added the `test` task.

## Questions for Reviewers

- Is this the right place to put all this stuff? Should the API definitions be somewhere more reusable?
- Is using a new controller the right way to go? It seems like a lot of boilerplate.
- Should there be more logging?
- Is it clear what the new types mean?

### DIFF
--- a/support-services/src/main/scala/com/gu/support/paperround/PaperRoundService.scala
+++ b/support-services/src/main/scala/com/gu/support/paperround/PaperRoundService.scala
@@ -1,0 +1,193 @@
+package com.gu.support.paperround
+
+import com.gu.okhttp.RequestRunners.FutureHttpClient
+import com.gu.rest.WebServiceHelper
+import com.gu.support.config.PaperRoundConfig
+import java.time.ZonedDateTime
+import io.circe._
+import io.circe.generic.semiauto._
+import io.circe.generic.extras.semiauto._
+import io.circe.generic.extras.decoding.ConfiguredDecoder
+import io.circe.generic.extras.Configuration
+
+import scala.concurrent.{ExecutionContext, Future}
+
+import PaperRoundService.{AgentsEndpoint, CoverageEndpoint, ChargeBandsEndpoint}
+
+class PaperRoundService(config: PaperRoundConfig, client: FutureHttpClient)(implicit
+    executionContext: ExecutionContext,
+) extends WebServiceHelper[PaperRoundService.Error] {
+  override val wsUrl: String = config.apiUrl
+  override val httpClient: FutureHttpClient = client
+
+  def coverage(body: CoverageEndpoint.RequestBody): Future[CoverageEndpoint.Response] = {
+    postForm[CoverageEndpoint.Response](
+      endpoint = "coverage",
+      data = Map("postcode" -> List(body.postcode)),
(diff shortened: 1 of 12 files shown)


undergroundquizscene:
[Create support-frontend backend for getting the list of PaperRound delivery agents](https://trello.com/c/gN131PMt/1445-create-support-frontend-backend-for-getting-the-list-of-paperround-delivery-agents)

undergroundquizscene:
The following change suggestions were made in review:

- [ ] It would be good to have deserialisation tests for the error cases, as they’re currently untested. There’re example values in the swagger documentation that can be used.
- [ ] I need to add the API URL and key to the parameter store for PROD before merging
- [ ] I should include the documentation I’ve been preparing explaining the API and what I’ve learned doing this work

support-services/src/main/scala/com/gu/support/paperround/PaperRoundService.scala
```diff
@@ -0,0 +1,193 @@
...
+      data = Map.empty,
+      headers = Map("x-api-key" -> config.apiKey),
+    )
+  }
+}
+
+object PaperRoundService {
+  val snakeCase: Configuration = Configuration.default.withSnakeCaseMemberNames
+  val lowercase: Configuration = Configuration.default.copy(transformMemberNames = _.toLowerCase)
+  case class Error(statusCode: Integer, message: String, errorCode: ZonedDateTime)
```
LAKSHMIRPILLAI, line 51: Are we planning to ask PaperRound why is there the ZonedDateTime  passed as an errorCode?

undergroundquizscene: Yes, I’ll email Chris today

undergroundquizscene:
I can see that [the deployment succeeded](https://riffraff.gutools.co.uk/deployment/view/20b3726d-a91d-46db-8164-84d55801fa7a) a little while ago, and it references the right commit. Not sure why prout thinks it’s not on prod yet.

undergroundquizscene:
The commit visible in the comment in the html on the site matches the commit on the riff raff deployment: commit 38a971a72c44862705863cc6bdf2a7ba5173937f

PR #5070, opened 2023-06-15
Use Okta access token to authorise call to Mdapi

kelvin-chappell:
## What are you doing in this PR?

This modifies calls out to the members' data API so that they include an authorization header containing an Okta access token when the feature switch `authenticateWithOkta` is on.  When the feature switch is on but a user isn't signed in, calls to members' data API won't occur at all as they are bound to fail without authorisation.  When the switch is off, the authorization header is omitted and the legacy `SC_GU_U` cookie is used for authorisation.

[**Trello Card**](https://trello.com/c/yun441Ni/4437-support-migrate-clientside-calls-to-mdapi-to-use-okta-access-tokens)

## To test

Smoke test is:

1. Feature switch `authenticateWithOkta` off:
All functionality is same as before. Calls out to members' data API give a successful response in browser console network tab when user is signed in with a legacy Identity cookie.

2. Feature switch `authenticateWithOkta` on:
When user is signed in, calls out to members' data API have an authorization request header and give a successful response in browser console network tab.
When user isn't signed in, there are no calls out to members' data API appearing in the browser console network tab.

No new errors in browser console or in Cloudwatch logs.

## Still to do

- [x] Complete Okta config to give access to scopes that will provide ID claims

## Related PRs

- #4997 
- #5024
- guardian/identity-platform#648

PR #5133, opened 2023-07-13
Add new action that tries to authenticate using cookies only

kelvin-chappell:
Previously, we had a [UserFromAuthCookiesActionBuilder](https://github.com/guardian/support-frontend/blob/fd08fbdc015eece6c8193bb6ad9b0bd93dc49653/support-frontend/app/actions/UserFromAuthCookiesActionBuilder.scala) that checked for cookies containing auth tokens and tried to fetch the tokens if the cookies didn't exist.
In this PR, we optimise the use of auth token cookies by renaming `UserFromAuthCookiesActionBuilder` to `UserFromAuthCookiesOrAuthServerActionBuilder` and introducing a new action builder called `UserFromAuthCookiesActionBuilder`.  Now,  `UserFromAuthCookiesOrAuthServerActionBuilder` either uses the cookies to authenticate or to tries to fetch them.  And the new `UserFromAuthCookiesActionBuilder` only tries to use the cookies to authenticate.  With this new action builder if the cookies aren't present or are invalid, we allow the action to continue but without a user - we don't bother trying to fetch auth tokens.

## Why are you doing this?
This is to avoid making calls to the auth server unnecessarily.  If, on page load, we find that the user isn't signed in, there doesn't seem much point in checking again on form submission.  It also means that where form submission is triggered from the browser we don't have security issues with async calls from the browser to the auth domain and back.

## Related PRs
- #4997 
- #5024 
- #5070

### DIFF
--- a/support-frontend/app/actions/CustomActionBuilders.scala
+++ b/support-frontend/app/actions/CustomActionBuilders.scala
@@ -19,6 +19,7 @@ import scala.concurrent.{ExecutionContext, Future}
 
 class CustomActionBuilders(
     val asyncAuthenticationService: AsyncAuthenticationService,
+    userFromAuthCookiesOrAuthServerActionBuilder: UserFromAuthCookiesOrAuthServerActionBuilder,
     userFromAuthCookiesActionBuilder: UserFromAuthCookiesActionBuilder,
     cc: ControllerComponents,
     addToken: CSRFAddToken,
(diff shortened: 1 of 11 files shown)


coldlink (approved):
looks fine to me, worth having someone with better scala knowledge have a look though

support-frontend/app/actions/CustomActionBuilders.scala
```diff
@@ -32,6 +33,15 @@ class CustomActionBuilders(
     new PrivateActionBuilder(addToken, checkToken, csrfConfig, cc.parsers.defaultBodyParser, cc.executionContext)
 
   def MaybeAuthenticatedAction: ActionBuilder[OptionalAuthRequest, AnyContent] =
+    if (featureSwitches.authenticateWithOkta.isOn)
+      PrivateAction andThen userFromAuthCookiesOrAuthServerActionBuilder
+    else
+      PrivateAction andThen new AsyncAuthenticatedBuilder(
+        asyncAuthenticationService.tryAuthenticateUser,
+        cc.parsers.defaultBodyParser,
+      )
```
pvighi, line 42: minor: I don't know if it's too hard to do, but it might be nice to extract this duplicated code into a function that takes the action to taken when `featureSwitches.authenticateWithOkta` is on as a parameter.
not really necessary but it might make the code slightly easier to read

pvighi (approved):
looks good to me

support-frontend/app/actions/CustomActionBuilders.scala
kelvin-chappell: Refactored in daa7699661e6f5cbb9af9d2835c634e3b8253990

PR #5149, opened 2023-07-26
Add new zuora product for national delivery

undergroundquizscene:
This PR makes some of the changes in support-frontend required to handle new National Delivery products in Zuora. All Zuora changes I’ve made have been in the api sandbox, not in production.

Trello card: https://trello.com/c/4gIH3UcY/43-product-build-zuora-product-catalog-summary-journal

So far, I have:

- Created a [new product “Newspaper - National Delivery”](https://apisandbox.zuora.com/apps/Product.do?method=view&id=8ad096ca8992481d018992a35483159d) in CODE Zuora (the api sandbox) by cloning the existing [“Newspaper Delivery”](https://apisandbox.zuora.com/apps/Product.do?method=view&id=2c92c0f955c3cf0f0155c5d9ddc53bc3) product and making some changes.
  - Removed the …+, Legacy, and Echo-Legacy product rate plans, because we no longer sell these
  - Removed the Saturday and Sunday product rate plans, because we will not offer these for national delivery
- Added a new value “Newspaper - National Delivery” to [the ProductType custom field](https://apisandbox.zuora.com/apps/CustomField.do?method=viewDetail&id=2c92c0f845fed47e0146057243b161cd&objectType=Product) and used it for the new product
- Added a new `FulfilmentOptions` case
- Added a custom field in Zuora on the subscription for the “delivery agent” chosen by the customer, as well as a corresponding Salesforce field
  - Added this to the Paper product type and thereby the `CreateSupportWorkersRequest`
  - Is “delivery agent” an ok name for this? The designs done by Rob for the frontend use the phrase “delivery provider”, which I think I prefer.
- Modified the frontend to submit NationalDelivery products (so I could test them in CODE). (This change should not be merged, and has been removed from the PR.)
  - Made these zuora subscriptions in this way: https://apisandbox.zuora.com/platform/subscriptions/8ad081c689c0d7620189c1576e0e7bc3 and https://apisandbox.zuora.com/platform/subscriptions/8ad081c689de67b50189dfbe96ed661b
- Added the new chart of account entries
  - I added National Delivery, National Delivery M-F, National Delivery SAT, National Delivery SUN, and deferred versions of each of those, copying from the Home Delivery equivalents in production zuora.
  - I also set each product rate plan charge for national delivery to use these entries
  - I copied the Home Delivery tax code from production zuora as it didn’t exist in the sandbox: I assume we will use the existing one when we do this in production, rather than creating a new one

I have not yet:

- Added the holiday stop discounts

I would like to find out whether there’s a way to test that the Zuora setup is all correct, e.g. by generating a financial report as would be done in production.

### DIFF
--- a/support-frontend/app/utils/CheckoutValidationRules.scala
+++ b/support-frontend/app/utils/CheckoutValidationRules.scala
@@ -5,7 +5,15 @@ import com.gu.i18n.Currency.GBP
 import com.gu.i18n.{Country, CountryGroup, Currency}
 import com.gu.support.abtests.BenefitsTest.isValidBenefitsTestPurchase
 import com.gu.support.acquisitions.AbTest
-import com.gu.support.catalog.{Collection, Domestic, FulfilmentOptions, HomeDelivery, NoFulfilmentOptions, RestOfWorld}
+import com.gu.support.catalog.{
+  Collection,
+  Domestic,
+  FulfilmentOptions,
+  NationalDelivery,
+  HomeDelivery,
+  NoFulfilmentOptions,
+  RestOfWorld,
+}
 import com.gu.support.redemptions.RedemptionData
 import com.gu.support.workers._
 import com.gu.support.zuora.api.ReaderType
@@ -368,7 +376,8 @@ object PaperValidation {
         deliveredToUkAndPaidInGbp(address.country, createSupportWorkersRequest.product.currency)
       val deliveryAddressHasAddressLine1AndCity = hasAddressLine1AndCity(address)
       val validPostcode = fulfilmentOptions match {
-        case HomeDelivery => postcodeIsWithinDeliveryArea(postCode)
+        case NationalDelivery => postcodeIsWithinNationalDeliveryArea(postCode)
+        case HomeDelivery => postcodeIsWithinHomeDeliveryArea(postCode)
         case Collection => Valid
         case Domestic => Invalid("domestic is not valid for paper")
         case RestOfWorld => Invalid("rest of world is not valid for paper")
@@ -398,7 +407,11 @@ object PaperValidation {
 
   }
 
-  def postcodeIsWithinDeliveryArea(postcode: String): Result =
+  def postcodeIsWithinNationalDeliveryArea(postcode: String): Result = {
+    Valid // TODO: determine valid postcodes
+  }
+
+  def postcodeIsWithinHomeDeliveryArea(postcode: String): Result =
     M25_POSTCODE_PREFIXES.contains(getPrefix(postcode)).otherwise(s"postcode $postcode is not within M25")
 
   val M25_POSTCODE_OLD_PREFIXES = List(
--- a/support-frontend/test/utils/CheckoutValidationRulesTest.scala
+++ b/support-frontend/test/utils/CheckoutValidationRulesTest.scala
@@ -562,7 +562,8 @@ class PaperValidationTest extends AnyFlatSpec with Matchers {
   }
 
   it should "fail if the currency is USD" in {
-    val requestDeliveredToUs = validPaperRequest.copy(product = Paper(Currency.USD, Monthly, HomeDelivery, Everyday))
+    val requestDeliveredToUs =
+      validPaperRequest.copy(product = Paper(Currency.USD, Monthly, HomeDelivery, Everyday, None))
     PaperValidation.passes(requestDeliveredToUs, Collection) shouldBe an[Invalid]
   }
 
@@ -641,7 +642,8 @@ class GuardianWeeklyValidationTest extends AnyFlatSpec with Matchers {
   }
 
   it should "succeed if the currency is USD" in {
-    val requestDeliveredToUs = validWeeklyRequest.copy(product = Paper(Currency.USD, Monthly, HomeDelivery, Everyday))
+    val requestDeliveredToUs =
+      validWeeklyRequest.copy(product = GuardianWeekly(Currency.USD, Monthly, HomeDelivery))
     GuardianWeeklyValidation.passes(requestDeliveredToUs) shouldBe Valid
   }
 
@@ -755,7 +757,7 @@ object TestData {
     title = None,
     firstName = "grace",
     lastName = "hopper",
-    product = Paper(Currency.GBP, Monthly, HomeDelivery, Everyday),
+    product = Paper(Currency.GBP, Monthly, HomeDelivery, Everyday, Some("delivery agent (oho)")),
     firstDeliveryDate = Some(someDateNextMonth),
     paymentFields =
       Left(StripePaymentMethodPaymentFields(PaymentMethodId("test_token").get, Some(StripePaymentType.StripeCheckout))),
--- a/support-models/src/test/scala/com/gu/support/workers/SendThankYouEmailStateSerialisationSpec.scala
+++ b/support-models/src/test/scala/com/gu/support/workers/SendThankYouEmailStateSerialisationSpec.scala
@@ -78,7 +78,11 @@ object ProductTypeCreatedTestData {
   )
   val paperCreated = SendThankYouEmailPaperState(
     user = User("111222", "email@blah.com", None, "bertha", "smith", Address(None, None, None, None, None, Country.UK)),
-    Paper(fulfilmentOptions = Collection, productOptions = Saturday),
+    Paper(
+      fulfilmentOptions = Collection,
+      productOptions = Saturday,
+      deliveryAgent = Some("A delivery agent ID"),
+    ),
     PayPalReferenceTransaction("baid", "email@emaail.com"),
     PaymentSchedule(List(Payment(new LocalDate(2020, 6, 16), 1.49))),
     None,
--- a/support-workers/src/main/scala/com/gu/acquisitions/AcquisitionDataRowBuilder.scala
+++ b/support-workers/src/main/scala/com/gu/acquisitions/AcquisitionDataRowBuilder.scala
@@ -140,6 +140,9 @@ object AcquisitionDataRowBuilder {
         case (HomeDelivery, SaturdayPlus) => HomeDeliverySaturdayPlus
         case (HomeDelivery, Sunday) => HomeDeliverySunday
         case (HomeDelivery, SundayPlus) => HomeDeliverySundayPlus
+        case (NationalDelivery, Everyday) => NationalDeliveryEveryday
+        case (NationalDelivery, Sixday) => NationalDeliverySixday
+        case (NationalDelivery, Weekend) => NationalDeliveryWeekend
         case (Collection, Everyday) => VoucherEveryday
         case (Collection, EverydayPlus) => VoucherEverydayPlus
         case (Collection, Sixday) => VoucherSixday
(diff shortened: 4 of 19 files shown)


support-workers/src/main/scala/com/gu/acquisitions/AcquisitionDataRowBuilder.scala
```diff
@@ -149,7 +154,13 @@ object AcquisitionDataRowBuilder {
...
         case (Collection, SaturdayPlus) => VoucherSaturdayPlus
         case (Collection, Sunday) => VoucherSunday
-        case _ => VoucherSundayPlus
+        case (Collection, SundayPlus) => VoucherSundayPlus
+        case (NoFulfilmentOptions, _) | (_, NoProductOptions) | (Domestic, _) | (RestOfWorld, _) |
+            (NationalDelivery, EverydayPlus) | (NationalDelivery, SixdayPlus) | (NationalDelivery, WeekendPlus) |
+            (NationalDelivery, SaturdayPlus) | (NationalDelivery, SundayPlus) =>
+          throw new Exception(
+            s"Unrecognised (fulfilmentOptions, productOptions) combination: ($fulfilmentOptions, $productOptions)",
+          )
```
undergroundquizscene, line 163: Is `Exception` the right exception type to use here?

undergroundquizscene: I guess an alternative here would be to add a new case object `InvalidPrintProduct` and use that here: then at least the acquisitions would make it to the DB ok (and nothing would crash). It would mean we could have quiet failures though.

johnduffell: if you want to do something in response to the exception, it makes sense to have a proper type, otherwise I generally just throw a built in (usually runtime exception).  Although from a "functional programming" point of view, catching exceptions is frowned upon, so it depends how far away from FP the surrounding code is.

undergroundquizscene: At the moment there isn’t anything I want to do in response to the exception, as this case shouldn’t actually happen: I just want it to be a bit more visible than it was. At the moment incorrect combinations all go in as VoucherSundayPlus, which happened when I added the new fulfilment option: explicitly matching on all the cases means adding a new fulfilment option without updating this code gives a compiler error, which makes it harder to miss, but raises the question of what to do in these cases.

As of this morning, I’ve gone with an `InvalidPrintProduct` case class, which means nothing will crash with an unhandled exception and it should be clear from the acquisition data when something’s gone wrong.

johnduffell:
just taking a look at this one - great description from a technical point of view.

> Is “delivery agent” an ok name for this? The designs done by Rob for the frontend use the phrase “delivery provider”, which I think I prefer.

it's always great if we can use the business terminology - but so many times it's been changed after the code has been written!  Not sure who can confirm the final name at this stage - sorry I'm probably more unsure than you about it.

> Modified the frontend to submit NationalDelivery products (so I could test them in CODE). (This change should not be merged, and has been removed from the PR.)

That's totally fine now, although thinking forward I wonder if we'll need a switch or something later, so people can test in PROD/using test users?  I think we've used URL parameters or a separate route, and wrapped things in guardian google auth, which might work ok here?

support-frontend/app/utils/CheckoutValidationRules.scala
```diff
@@ -398,7 +407,11 @@ object PaperValidation {
 
   }
 
-  def postcodeIsWithinDeliveryArea(postcode: String): Result =
+  def postcodeIsWithinNationalDeliveryArea(postcode: String): Result = {
+    Valid // TODO: determine valid postcodes
```
johnduffell, line 411: I guess we'll have to check with paper round?  I was wondering myself whether they have a mapping from postcode to providers that we can download to S3 on a schedule or whether we always have to hit their API.

support-frontend/test/utils/CheckoutValidationRulesTest.scala
```diff
@@ -641,7 +642,8 @@ class GuardianWeeklyValidationTest extends AnyFlatSpec with Matchers {
   }
 
   it should "succeed if the currency is USD" in {
-    val requestDeliveredToUs = validWeeklyRequest.copy(product = Paper(Currency.USD, Monthly, HomeDelivery, Everyday))
+    val requestDeliveredToUs =
+      validWeeklyRequest.copy(product = GuardianWeekly(Currency.USD, Monthly, HomeDelivery))
```
johnduffell, line 646: good spot

support-models/src/test/scala/com/gu/support/workers/SendThankYouEmailStateSerialisationSpec.scala
```diff
@@ -78,7 +78,11 @@ object ProductTypeCreatedTestData {
   )
   val paperCreated = SendThankYouEmailPaperState(
     user = User("111222", "email@blah.com", None, "bertha", "smith", Address(None, None, None, None, None, Country.UK)),
-    Paper(fulfilmentOptions = Collection, productOptions = Saturday),
+    Paper(
+      fulfilmentOptions = Collection,
+      productOptions = Saturday,
+      deliveryAgent = Some("A delivery agent with an emoji: 📮"),
```
johnduffell, line 84: that's brave - zuora historically doesn't handle emojis well at all, often returning SQL errors!.  Not sure if it's changed.

support-workers/src/main/scala/com/gu/acquisitions/AcquisitionDataRowBuilder.scala
```diff
@@ -149,7 +152,12 @@ object AcquisitionDataRowBuilder {
         case (Collection, Saturday) => VoucherSaturday
         case (Collection, SaturdayPlus) => VoucherSaturdayPlus
         case (Collection, Sunday) => VoucherSunday
-        case _ => VoucherSundayPlus
+        case (Collection, SundayPlus) => VoucherSundayPlus
+        case (NoFulfilmentOptions, _) | (_, NoProductOptions) | (Domestic, _) | (RestOfWorld, _) |
+            (NationalDelivery, Saturday) | (NationalDelivery, Sunday) | (NationalDelivery, EverydayPlus) |
+            (NationalDelivery, SixdayPlus) | (NationalDelivery, WeekendPlus) | (NationalDelivery, SaturdayPlus) |
+            (NationalDelivery, SundayPlus) =>
+          InvalidPrintProduct(fulfilmentOptions, productOptions)
```
johnduffell, line 160: good spot fixing this up, that could really skew things otherwise.
Regarding the exception idea, we would presumably get either a support workers failure or timeout in that case, whereas in this case we wouldn't know about it.  If we want this behaviour, is it worth sending a metric to cloudwatch and alarming on it?  I'm not sure either way but wanted to raise it.

johnduffell (approved):
looks fab, appears like you've been very rigorous on this one and although tedious this PR looks like it's left things in a better state than before hand! 👍

undergroundquizscene:
> > Is “delivery agent” an ok name for this? The designs done by Rob for the frontend use the phrase “delivery provider”, which I think I prefer.
> 
> it's always great if we can use the business terminology - but so many times it's been changed after the code has been written! Not sure who can confirm the final name at this stage - sorry I'm probably more unsure than you about it.

I think I’ll leave it as “delivery agent” for now.

> > Modified the frontend to submit NationalDelivery products (so I could test them in CODE). (This change should not be merged, and has been removed from the PR.)
> 
> That's totally fine now, although thinking forward I wonder if we'll need a switch or something later, so people can test in PROD/using test users? I think we've used URL parameters or a separate route, and wrapped things in guardian google auth, which might work ok here?

In https://github.com/guardian/support-frontend/pull/5169 and https://github.com/guardian/support-frontend/pull/5201 we’ve added the ability to do this but put it behind a disabled AB test – that way people can force themselves into the variant with the query parameter to test it. I wondered about restricting it more, but we decided that the consequences of someone accessing it before we’re ready are pretty minor, and that therefore it’s not worth it.

support-models/src/test/scala/com/gu/support/workers/SendThankYouEmailStateSerialisationSpec.scala
undergroundquizscene: Hmm, good point – I wanted to make sure we tested with emojis, but we don’t actually run the IT tests until after merge, do we? I’d better put through an application with an emoji to test it out in the sandbox before I merge.

I believe whether zuora can handle emojis depends on the field (and the emoji: some have low enough codepoints that they work fine, but this isn’t one of them) – maybe it’ll be fine? 😅

support-workers/src/test/scala/com/gu/zuora/Fixtures.scala
```diff
@@ -97,19 +97,31 @@ object Fixtures {
     ReferrerAcquisitionData(None, None, None, None, None, None, None, None, None, None, None, None, None)
 
   val touchpointEnvironment = TouchPointEnvironments.fromStage(Configuration.stage)
-  val everydayHDProductRatePlanId =
+  val everydayHomeDeliveryProductRatePlanId =
     catalog.Paper.getProductRatePlan(touchpointEnvironment, Monthly, HomeDelivery, Everyday) map (_.id)
+  val everydayNDProductRatePlanId =
```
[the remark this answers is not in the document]
undergroundquizscene: Yes, I absolutely did 🤦🏻

support-workers/src/main/scala/com/gu/acquisitions/AcquisitionDataRowBuilder.scala
undergroundquizscene: You’re right, this is too hidden. If the support-workers execution fails at this point (the `SendAcquisitionEvent` lambda within the parallel steps at the end), will it be straightforward to retry the lambda? I’ll test that now. I was worried that if it threw an exception it might be hard to recover the acquisition data to retry it, but I had forgotten this was within support-workers.

undergroundquizscene: Tested this: it’s quite straightforward to copy the input from the step function step and trigger a new execution with it. I think I’ll return to throwing a runtime exception in this case, and remove the `InvalidPrintProduct` type.

support-frontend/app/utils/CheckoutValidationRules.scala
undergroundquizscene: I believe we’ll always have to hit their API: we could possibly take a look at caching it our side, I suppose. There’s [another Trello card](https://trello.com/c/r557VThC/65-product-build-support-frontend-backend-validates-paperround-coverage-of-provided-postcode-and-chosen-agent) in the backlog to add this validation.

support-models/src/test/scala/com/gu/support/workers/SendThankYouEmailStateSerialisationSpec.scala
undergroundquizscene: Sure enough, sending a delivery agent ID with an emoji in it breaks zuora. At the moment it’s just delivery agent IDs we’re storing, which will presumably be numeric – I think we can probably assume those won’t contain emojis, so I’ll remove them from the tests. However, if we add more custom fields (which is a possibility under consideration), we’ll need to remember to exclude characters above U+FFFF and probably raise an alarm in those cases.

undergroundquizscene: I’ve removed the emojis

undergroundquizscene:
@johnduffell and I agreed that he would add the holiday stop and delivery problem discount product rate plans as part of another piece of work he’s doing, as it’s more closely related to that

PR #5152, opened 2023-07-27
Point prout to subscribe page

undergroundquizscene:
The /contribute page now does an oauth redirect, which prout seems to be having trouble with. ([The logs](https://prout-bot.herokuapp.com/view/guardian/support-frontend) report `Could not read from this url, got java.net.ProtocolException: Too many follow-up requests: 21`.)

We think this is because the oauth redirects require cookies to be set in order to function correctly, while Prout doesn’t set cookies.

Handily, the /subscribe endpoint does not (currently) require an oauth redirect, and so this should fix the problem.

### DIFF
--- a/.prout.json
+++ b/.prout.json
@@ -1,7 +1,7 @@
 {
     "checkpoints": {
         "PROD": {
-            "url": "https://support.theguardian.com/uk",
+            "url": "https://support.theguardian.com/uk/subscribe",
             "overdue": "16M",
             "messages": {
                 "seen": "prout/seen.md"


rtyley:
> The /contribute page now does an oauth redirect, which prout seems to be having trouble with

Yeah, Prout does not currently maintain cookies as it hits a url, so it gets told to redirect - which it _does_ follow, but it doesn't gain the cookies from that redirect that everyone else does, so it gets into a redirect loop!

```
$ curl -I https://support.theguardian.com/uk/contribute
HTTP/2 303
location: /oauth/authorize
```

Pointing to https://support.theguardian.com/uk/subscribe should work fine for now!

```
$ curl -I https://support.theguardian.com/uk/subscribe
HTTP/2 200
content-type: text/html; charset=UTF-8
```

rtyley:
I've also opened https://github.com/guardian/prout/issues/111, which proposes a long-term solution to the problem, ie Prout supporting cookies - could make a nice small PR on Prout if anyone fancies working on it, I'd be happy to pair if that helps!

undergroundquizscene:
> I've also opened [guardian/prout#111](https://github.com/guardian/prout/issues/111), which proposes a long-term solution to the problem, ie Prout supporting cookies - could make a nice small PR on Prout if anyone fancies working on it, I'd be happy to pair if that helps!

Ah yes, that’d be excellent! I’ll see if I can find some time soon, and get on to you about pairing :)

rtyley:
Note that immediately on merge, with the new config available on the default branch, Prout [now sees accurate commit id information](https://prout-bot.herokuapp.com/view/guardian/support-frontend)!

image

It has labelled all recently-merged PRs as 'Seen' now:

image

undergroundquizscene:
Yes, it was impressively quick!

PR #5153, opened 2023-07-28
Describe manual acquisition event insertion

undergroundquizscene:
Sometimes the acquisition-events-api fails when sending events to BigQuery. In these cases we need to manually insert the events that failed.

This wasn’t too tricky to figure out, but having a description of the process might save someone some guesswork.


rBangay (approved):
Thanks for doing this Emily 👍

PR #5162, opened 2023-08-01
acquisition-events-api: Close JsonStreamWriter and BigQueryWriteClient

undergroundquizscene:
The code wasn’t closing the JsonStreamWriter or BigQueryWriteClient when finished with them, which is what’s been causing the OutOfMemoryErrors that we’ve seen in production. With this change we should stop seeing those.


undergroundquizscene:
Tested this in CODE by sending 1200 events close together, no errors – I reckon this has fixed the problem.

rBangay (approved):
👍

undergroundquizscene:
I think this has broken the payment API: gonna test in CODE to check.

PR #727, opened 2018-06-15
Track where the Payment Request API is available

joelochlann:
## Why are you going this?
We'd like to know how many visitors to the site could make use of the Payment Request API if we integrated it

The Payment Request API allows the browser to store & retrieve card details, resulting in a much slicker checkout experience, especially on mobile. But browser support is still limited and there are a few other caveats to its availability. See here for more info:  https://stripe.com/docs/payment-request-api

## How?
We created a new custom dimension in GA (`paymentRequestApiStatus`, number `21`) and we populate with one of the following values: 

* 'PaymentRequestAPINotAvailable': the PaymentRequest is not defined. 
* 'CanMakePaymentNotAvailable': the PaymentRequest's CanMakePaymentNotAvailable function is not availble.
* 'AvailableNotInUse: The payment API is available but the user is not using it.
* 'AvailableInUse': The payment API is available and the user is using it.
* 'PaymentRequestAPIError': There was an exception in the block which is testing the payment request API.

PR #5165, opened 2023-08-01
Remove getPaymentAPIStatus logic from Google Tag Manager events

GHaberis:

Removes some information no longer necessary from the support sites Google Tag Manager events. Introduced over 5 years ago this was a dimension introduced to GA so we could report on whether the Payment Request API was available (see: https://github.com/guardian/support-frontend/pull/727), this isn't data we need to record anymore.

## Why are you doing this?

Cleaning out unused data from GTM tracking events ahead of a refactor.


paulbrown1982 (approved):
👍 keeping the GA/GTM implementation as simple and as textbook as possible is wise at this stage, whilst we are assessing whether to upgrade or remove GA.

PR #5168, opened 2023-08-02
Only run scalafmt on files which have been changed

rupertbates:
Scalafmt was previously being run on all scala files if any had been updated, this PR changes it to only run on changed files.

[Docs are here](https://scalameta.org/scalafmt/docs/installation.html#settings)


undergroundquizscene (approved):
Excellent!

PR #5176, opened 2023-08-08
Hide benefits list when a user cannot contribute above the threshold

i-hardy:

This adds a condition to the rendering of the benefits list to hide it if it is not possible for a user to contribute an amount over the threshold for receiving supporter plus benefits- ie. they cannot enter a custom amount, and the pre-set amounts are all under the threshold.

[**Trello Card**](https://trello.com/c/H63ql4Ua)

## Why are you doing this?

We need to be able to not sell supporter plus in specific countries or regions for tax reasons. Amounts can now be controlled on a per-country basis from the RRCP, so this simply avoids users seeing a list of benefits that they can't actually purchase.

## Is this an AB test?
- [ ] Yes
- [x] No

## Screenshots

Screenshot 2023-08-08 at 12-27-08 Support the Guardian

### DIFF
--- a/support-frontend/assets/components/checkoutBenefits/checkoutBenefitsListContainer.tsx
+++ b/support-frontend/assets/components/checkoutBenefits/checkoutBenefitsListContainer.tsx
@@ -2,6 +2,7 @@ import type { ContributionType } from 'helpers/contributions';
 import { simpleFormatAmount } from 'helpers/forms/checkouts';
 import { currencies } from 'helpers/internationalisation/currency';
 import { setSelectedAmount } from 'helpers/redux/checkout/product/actions';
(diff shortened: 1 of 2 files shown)


marialani (approved):
LGTM! 👍

GHaberis:
@i-hardy we discussed this after stand-up today and agreed we need the cancel anytime message and tooltip to remain visible, as although it sits structurally within the same container element as the Benefits, it actually serves a separate purpose - which is still required whether we show the benefits or not.

Screenshot 2023-08-09 at 11 52 31

i-hardy:
Made the amendments with the tooltip and updated the screenshot.

GHaberis (approved):
Looks good @i-hardy 👍

PR #5177, opened 2023-08-08
Increase timeout for calling idapi to 7 seconds

undergroundquizscene:
We’ve had a couple more timeouts recently (4 in the last 2 months). We’d like to increase the timeout length slightly more to see if that resolves them, as we understand that the calls can take a while because it involves calling Okta.


johnduffell:
deploy failed with `S3Exception We encountered an internal error. Please try again.` so I have redeployed it and it's working better

PR #5179, opened 2023-08-10
Allow site to continue when identity gateway is down

kelvin-chappell:
This change ensures that when [gateway](https://profile.theguardian.com/) is down readers are still able to create subscriptions and make contributions.  It does this by making a server-side preflight call to the auth server to check that it's responding appropriately before starting the auth flow.  If the auth server is down the request will continue without user data, in the same way it would do if there were no signed-in user.

The preflight call will only take place if a call to the auth server is necessary for authentication.  In many cases, the cookies that are already present will be enough to authenticate the user.

## How to test
1. Go to /uk/contribute in Code in an Incognito window and inspect network tab to see that call to auth flow is being made.
2. Disable Code gateway and then go to /uk/contribute in Code in an Incognito window and inspect network tab to see that call to auth flow is not being made.  Failure to authenticate should be logged in Cloudwatch.

[**Trello Card**](https://trello.com/c/B50MXOR6/4493-support-site-should-continue-to-work-when-gateway-is-down)


raphaelkabo (approved):
This looks like a nice, elegant solution to the problem!

PR #5180, opened 2023-08-10
Migrate contribs/S+ direct debit components to use the Redux hooks pattern

i-hardy:

This migrates our existing direct debit components to use the `useContributionsSelector` and `useContributionsDispatch` hooks rather than the deprecated connected props pattern.

[**Trello Card**](https://trello.com/c/DoAUCR1g)

## Why are you doing this?

This lays useful groundwork for ongoing work to replace this component with something more in line with the current design and UX of the checkout.


marialani:
This looks fine to go into main but I think may lead to merge conflicts later on down the line. The current DD base branch has also had phasing removed. Maybe instead of a base branch it might be more helpful to open up a draft PR with a WIP label?

PR #5183, opened 2023-08-11
No Benefits AB test

GHaberis:

This PR hides the benefits list if a user is participating in the `2023-08-08_BENEFITS_GONE` RRCP configured Amounts test and has been assigned to either the `V1` or `V3` cohort.

Note: When looking at this I also noted an issue we introduced when moving the Tooltip out of the benefits list ( https://github.com/guardian/support-frontend/pull/5176), this inadvertently meant the "Cancel anytime" Tooltip started appearing on the Single tab. To address this I've put in an `isOneOff` check when deciding to render this "Cancel anytime" Tooltip, I've also put this in a `TooltipContainer` module to keep the logic out of the presentational `Tooltip` component.

[**Trello Card**](https://trello.com/c/j4eWBkHY/1471-removing-benefits-on-the-current-1-step-checkout-a-b-test-with-amounts)


i-hardy (approved):

paul-daniel-dempsey (commented):
Nice and neat, looks good

PR #5186, opened 2023-08-11
Iterate through possible hide benefits test names

GHaberis:

It turns out the hide benefits test name in the RRCP is not '2023-08-08_BENEFITS_GONE', but that they're actually multiple tests (one per region) with the region code appended on the test name by convention when setting the test up in RRCP. So the actual test names are:

2023-08-08_BENEFITS_GONE__UK
2023-08-08_BENEFITS_GONE__US
2023-08-08_BENEFITS_GONE__CA
2023-08-08_BENEFITS_GONE__EU
2023-08-08_BENEFITS_GONE__AU
2023-08-08_BENEFITS_GONE__NZ
2023-08-08_BENEFITS_GONE__ROW

I've therefore had to update the check so we iterate over the possible test names.

### DIFF
--- a/support-frontend/assets/helpers/redux/checkout/product/selectors/isSupporterPlus.ts
+++ b/support-frontend/assets/helpers/redux/checkout/product/selectors/isSupporterPlus.ts
@@ -1,3 +1,4 @@
+import type { Participations } from 'helpers/abTests/abtest';
 import { getAmount } from 'helpers/contributions';
 import { getContributionType } from 'helpers/redux/checkout/product/selectors/productType';
 import type { ContributionsState } from 'helpers/redux/contributionsStore';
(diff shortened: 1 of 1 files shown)


support-frontend/assets/helpers/redux/checkout/product/selectors/isSupporterPlus.test.ts
```diff
@@ -0,0 +1,33 @@
+import { _ } from './isSupporterPlus';
+
+describe('isInHideBenefitsTest', () => {
```
paul-daniel-dempsey, line 3: Jest looks extremely useful testing standalone calls, will use this in future for standalone function testing. Have 'run' the test manually/successfully from visual studio, this brings up a jest console. Are these jest lines added to the build automatically as well?

support-frontend/assets/helpers/redux/checkout/product/selectors/isSupporterPlus.ts
```diff
@@ -26,6 +27,33 @@ export function isSupporterPlusPurchase(state: ContributionsState): boolean {
...
+		'UK',
+		'US',
+		'CA',
+		'EU',
+		'AU',
+		'NZ',
+		'ROW',
+	];
+
+	const result = regionalSuffixList.reduce((hideBenefits, suffix): boolean => {
```
paul-daniel-dempsey, line 41: Yep can see how hideBenefits gradually reduces through the list of regional suffix's, nice!

support-frontend/assets/helpers/redux/checkout/product/selectors/isSupporterPlus.test.ts
GHaberis: @paul-daniel-dempsey yes unit tests are a nice way of validating your code does what you expect it to, and yes we have a Github action that runs these tests on our Pull Requests too!

PR #5138, opened 2023-07-17
[RB] - update GW landing page discount copy for EU countries

rBangay:
## What are you doing in this PR?

This pr updates the copy on the Guardian Weekly landing page specifically for EU countries. For those countries the discount should be 87% not the default (excluding Australia) of 35%.

#### This change should be live between 17th July - 14th Aug.

[**Trello Card**](https://trello.com/c/Li371ilc/1408-guardian-weekly-12-for-12-eu-copy-change-on-landing-page)

## Screenshot
Screenshot 2023-07-17 at 10 52 36

PR #5187, opened 2023-08-11
remove countrygroup specific benefits  (EU GW landing page)

paul-daniel-dempsey:
Reverts https://github.com/guardian/support-frontend/pull/5138

Removes the copy on the Guardian Weekly landing page specifically for EU countries. For those countries the discount should be 35% not 87% now.

CountryGroupId not required for benefits currently so removed.

### DIFF
--- a/support-frontend/assets/pages/weekly-subscription-landing/components/content/benefits.tsx
+++ b/support-frontend/assets/pages/weekly-subscription-landing/components/content/benefits.tsx
@@ -17,10 +17,6 @@ const coreBenefits = [
 	},
 ];
 
-function getBenefits() {
-	return coreBenefits;
-}
-
 function Benefits(): JSX.Element {
 	return (
 		<BenefitsContainer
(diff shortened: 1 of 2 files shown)


support-frontend/assets/pages/weekly-subscription-landing/components/content/benefits.tsx
```diff
@@ -38,7 +30,7 @@ function Benefits({
 					content: (
 						<>
 							<BenefitsHeading text="As a subscriber you’ll enjoy" />
-							<List items={getBenefits(countryGroupId)} />
+							<List items={getBenefits()} />
```
GHaberis, line 33: Looks like we don't need a `getBenefits` function / function call anymore, we could just pass down the `coreBenefits` list in the `items` prop here.

paul-daniel-dempsey: ah yes, done.

marialani (approved):
Looks good! 👍

PR #5196, opened 2023-08-18
go away webpack error overlay

rBangay:
stop webpack from showing an error overlay in the webpack dev config. (errors are still logged in the console and by default it doesn't do this in production)

## Screenshot
Screenshot 2023-08-18 at 15 53 11


paul-daniel-dempsey (commented):
Looks good

PR #5197, opened 2023-08-18
Restyle the first page of the two step checkout

rBangay:
Restyling the first page of the new two step checkout.

## Screenshots
Control single contribution             |  Two step single contribution
:-------------------------:|:-------------------------:
  |  

Control recurring contribution             |  Two step recurring contribution
:-------------------------:|:-------------------------:
  |  

-----------------------------

Control single contribution             |  Two step single contribution
:-------------------------:|:-------------------------:
  |  

Control recurring contribution             |  Two step recurring contribution
:-------------------------:|:-------------------------:
  |


support-frontend/assets/components/checkoutBenefits/checkoutBenefitsList.tsx
```diff
@@ -40,18 +51,34 @@ export type CheckoutBenefitsListProps = {
 	checkListData: CheckListData[];
 	buttonCopy: string | null;
 	handleButtonClick: () => void;
+	withBackground?: boolean;
+	twoStepBenefitsList?: boolean;
```
i-hardy, line 55: I would favour making this prop more generic versus a name like this that's very specific to the test. If this doesn't get cleaned up after the test (whichever variant we ship) this creates a bit of an artefact in the code that may be harder to understand in six months or a year. I'd suggest just calling the prop `checklistStyle` and passing `'compact' | 'standard'` directly, or retaining the boolean but making this `isCompactList` or something.

i-hardy (approved):
Looking good, just a couple of suggestions.

PR #5200, opened 2023-08-22
[RB] - only show refund copy in cancel tooltip if supporter plus thre…

rBangay:
Only show refund copy in cancel tooltip if supporter plus threshold is met

[**Trello Card**](https://trello.com/c/dBwhFzhT/1514-change-copy-on-cancel-anytime-tooltip-checkout)

### before
Screenshot 2023-08-22 at 12 40 32

### after
below threshold            |  above threshold
:-------------------------:|:-------------------------:
  |


paul-daniel-dempsey (approved):
Nice and neat, great

PR #5205, opened 2023-08-24
Update server-side validation to block support-workers executions with no state

LAKSHMIRPILLAI:

Update server-side validation to  block support-workers executions with no state.

[**Trello Card**](https://trello.com/c/x9y6amch/1500-find-out-why-server-side-validation-didnt-block-support-workers-executions-with-no-state)
[The support-workers executions failed are attached to the trello card]
## Why are you doing this?

In alarm investigations, we found a few step functions where the support workers successfully went through CreatePaymentMethod step when the billing address details had an empty string in the State field for US .But later this caused a Step function failure at the CreateZuoraSubscription step  with the error "Taxation Requirement: State is required for Sold To Contact if the country is United States or Canada.\\\". Countries like US and CA should not have the empty State field.
Upon investigation, it was found that server side validation checks considers an empty string added in the Optional State field as Valid.

This issue was flagged  when the user was utilising the 'Payment Request Button"  and server-side checks  allowed form submission with an empty string in State field in billing address even when the client side validation  showed the error (if the State field was not selected from the checkout form) to the end user. Otherwise the client side validation checks should have prevented the form submission.

### DIFF
--- a/support-frontend/app/utils/CheckoutValidationRules.scala
+++ b/support-frontend/app/utils/CheckoutValidationRules.scala
@@ -246,8 +246,6 @@ object AddressAndCurrencyValidationRules {
     ) {
       {
         val validState = stateFromRequest match {
-          case Some(validState) if (validState.isEmpty) =>
-            Invalid(s"state is required for $countryFromRequest")
           case None =>
             Invalid(s"state is required for $countryFromRequest")
           case Some("") =>
--- a/support-frontend/app/utils/CheckoutValidationRules.scala
+++ b/support-frontend/app/utils/CheckoutValidationRules.scala
@@ -244,15 +244,12 @@ object AddressAndCurrencyValidationRules {
     if (
       countryFromRequest == Country.US || countryFromRequest == Country.Canada || countryFromRequest == Country.Australia
     ) {
-      {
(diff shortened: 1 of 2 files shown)


support-frontend/app/utils/CheckoutValidationRules.scala
```diff
@@ -244,10 +244,19 @@ object AddressAndCurrencyValidationRules {
...
+      {
+        val validState = stateFromRequest match {
+          case None =>
+            Invalid(s"state is required for $countryFromRequest")
+          case Some("") =>
+            Invalid(s"state is required for $countryFromRequest")
+          case _ => Valid
+        }
+        validState
+      }
```
undergroundquizscene, line 256: I think `match` produces a single expression, so I believe you should be able to remove the val and braces:

```suggestion
      stateFromRequest match {
        case None =>
          Invalid(s"state is required for $countryFromRequest")
        case Some("") =>
          Invalid(s"state is required for $countryFromRequest")
        case _ => Valid
      }
```

undergroundquizscene (approved):
Looks great! Glad to see the tests as well. Have a minor code suggestion:

PR #5207, opened 2023-08-25
National Delivery - Add Validation for Delivery Provider

andrewHEguardian:

This PR adds validation to the paper checkout form field for selecting a delivery provider when one is available (based on postcode), as well as a side effect to clear the form error when one is selected. 

The validation itself is not behind the `nationalDelivery` AB test but the logic checks whether a delivery provider is available - which it won't be unless the user is in the AB test.

[**Trello Card**](https://trello.com/c/rpO50Bym/82-support-frontend-error-message-on-no-delivery-partner-chosen)

## Why are you doing this?

When inputting an address that requires a delivery partner, if a user doesn’t choose a partner, red error message should appear.

## Is this an AB test?
- [ ] Yes
- [ ] No

## Screenshots
image

### DIFF
--- a/support-frontend/assets/helpers/redux/checkout/addressMeta/subscriptionsSideEffects.ts
+++ b/support-frontend/assets/helpers/redux/checkout/addressMeta/subscriptionsSideEffects.ts
@@ -1,0 +1,42 @@
+import type { AnyAction } from '@reduxjs/toolkit';
+import { isAnyOf } from '@reduxjs/toolkit';
+import type { SubscriptionsStartListening } from 'helpers/redux/subscriptionsStore';
+import { enableOrDisableForm } from 'helpers/subscriptionsForms/checkoutFormIsSubmittableActions';
+import type { FormField } from 'helpers/subscriptionsForms/formFields';
+import type { AnyCheckoutState } from 'helpers/subscriptionsForms/subscriptionCheckoutReducer';
+import { removeError } from 'helpers/subscriptionsForms/validation';
+import { setDeliveryAgent } from './actions';
+
+function removeErrorsForField(
+	fieldName: FormField,
+	state: AnyCheckoutState,
+): AnyAction {
+	return {
+		type: 'SET_FORM_ERRORS',
+		errors: removeError(fieldName, state.page.checkout.formErrors),
+	};
+}
+
+const shouldCheckFormEnabled = isAnyOf(setDeliveryAgent);
+
+const actionCreatorFieldNames: Record<string, FormField> = {
+	[setDeliveryAgent.type]: 'deliveryProvider',
+};
+
+export function addAddressMetaSideEffects(
+	startListening: SubscriptionsStartListening,
+): void {
+	startListening({
+		matcher: shouldCheckFormEnabled,
+		effect(action, listenerApi) {
+			listenerApi.dispatch(
+				removeErrorsForField(
+					actionCreatorFieldNames[action.type],
+					listenerApi.getState(),
+				),
+			);
+
+			listenerApi.dispatch(enableOrDisableForm());
+		},
+	});
+}
--- a/support-frontend/assets/helpers/redux/checkout/addressMeta/subscriptionsSideEffects.ts
+++ b/support-frontend/assets/helpers/redux/checkout/addressMeta/subscriptionsSideEffects.ts
@@ -1,5 +1,4 @@
 import type { AnyAction } from '@reduxjs/toolkit';
-import { isAnyOf } from '@reduxjs/toolkit';
 import type { SubscriptionsStartListening } from 'helpers/redux/subscriptionsStore';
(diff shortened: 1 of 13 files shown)


support-frontend/assets/helpers/subscriptionsForms/__tests__/validationTest.ts
```diff
@@ -69,6 +70,65 @@ describe('validation', () => {
...
+			expect(
+				requiredDeliveryAgentChosen('Collection', { isLoading: false }),
+			).toBe(true);
+		});
+
+		it('should return true if the fulfilment option is HomeDelivery and no agents are available', () => {
+			expect(
+				requiredDeliveryAgentChosen('HomeDelivery', { isLoading: false }),
+			).toBe(true);
+		});
```
undergroundquizscene, line 84: This case was a bit unexpected. I guess we don’t want this error to fire in this situation, because we’ll have another error that prevents the user from continuing?

I suppose this can happen when we actually want the user to continue with our standard within-M25 home delivery, as well as when they’re outside the M25 and we can’t deliver to their address, and we only want to stop them in the latter case.

support-frontend/assets/helpers/subscriptionsForms/validation.ts
```diff
@@ -69,6 +86,18 @@ function validate<Err>(rules: Array<Rule<Err>>): Err[] {
...
+	deliveryAgent: DeliveryAgentState,
+): boolean {
+	return deliveryAgent.chosenAgent ? true : false;
+}
+
+function deliveryAgentsAreAvailable(
+	deliveryAgent: DeliveryAgentState,
+): boolean {
+	return (deliveryAgent.response?.agents?.length ?? 0) > 0;
+}
```
undergroundquizscene, line 99: I see this function is now defined in two places: does that matter?

support-frontend/assets/helpers/redux/checkout/addressMeta/subscriptionsSideEffects.ts
```diff
@@ -0,0 +1,42 @@
...
+				removeErrorsForField(
+					actionCreatorFieldNames[action.type],
+					listenerApi.getState(),
+				),
+			);
+
+			listenerApi.dispatch(enableOrDisableForm());
+		},
+	});
+}
```
undergroundquizscene, line 42: I don’t really understand this bit: would you mind summarising? I assume it’s similar to things we do for other fields elsewhere.

undergroundquizscene (approved):

support-frontend/assets/helpers/subscriptionsForms/__tests__/validationTest.ts
andrewHEguardian: yes, exactly - we want this to fire only when the choice is available but they haven't chosen. Could the naming be more clear that this is what the validation is doing?

support-frontend/assets/helpers/subscriptionsForms/validation.ts
andrewHEguardian: I wasn't sure where the best place to share it would be so I thought a little duplication was better (it's a relatively simple function)

support-frontend/assets/helpers/subscriptionsForms/__tests__/validationTest.ts
undergroundquizscene: I guess you could call it `deliveryAgentChosenIfRequiredAndAvailable` 🫣 I think it’s probably fine as-is

support-frontend/assets/helpers/subscriptionsForms/validation.ts
undergroundquizscene: Yeah, sounds good

support-frontend/assets/helpers/redux/checkout/addressMeta/subscriptionsSideEffects.ts
undergroundquizscene: Perfect, thank you for the docs pointer!

support-frontend/assets/helpers/redux/checkout/addressMeta/subscriptionsSideEffects.ts
```diff
@@ -0,0 +1,42 @@
...
+
+const actionCreatorFieldNames: Record<string, FormField> = {
+	[setDeliveryAgent.type]: 'deliveryProvider',
+};
+
+export function addAddressMetaSideEffects(
+	startListening: SubscriptionsStartListening,
+): void {
+	startListening({
+		matcher: shouldCheckFormEnabled,
```
marialani, line 30: As there is a single action to be matched could swap out the `matcher` for the `actionCreator` option

```suggestion
		actionCreator: setDeliveryAgent,
```

support-frontend/assets/helpers/subscriptionsForms/validation.ts
marialani: I might suggest defining once and importing

PR #5215, opened 2023-08-31
Refactor direct debit styling for expanding form

marialani:
This updates the styling for the supporter plus direct debit section styling, including updating the underlying css from scss to emotion.

In order to visually update the existing styling a temporary container component has been passed to the direct debit section of the dropdown. This is only a visual change, not a functional one, as the underlying direct debit functionality, including hooking up the payment button is set to be updated separately.

[**Trello Card**](https://trello.com/c/yW2SDyGx/1502-refactor-direct-debit-scss-styling-to-emotion)

## Screenshots
| Mobile | Desktop |
| -- | -- |
| Screenshot 2023-09-01 at 14 01 28 | Screenshot 2023-09-01 at 14 02 29 |


support-frontend/assets/components/directDebit/directDebitForm/directDebitForm.tsx
```diff
@@ -111,12 +122,16 @@ export default function DirectDebitForm(props: PropTypes): JSX.Element {
...
+					css={css`
+						margin-top: 8px;
+					`}
+				>
 					<ErrorMessage message={formError} svg={<SvgExclamationAlternate />} />
 				</div>
 			)}
 
-			<PaymentButton onConfirmClick={onSubmit} />
+			{/* <PaymentButton onConfirmClick={onSubmit} /> */}
```
undergroundquizscene, line 134: Removing this because once we’re no longer popping up the modal we won’t need it, I assume. Why comment out rather than deleting?

undergroundquizscene (approved):
Looks good :)

support-frontend/assets/components/directDebit/directDebitForm/directDebitForm.tsx
marialani: Yeah that's right, we wouldn't need the component but the onSubmit handler will probably be moved around slightly. The reason for commenting out is that there's another card in play that will be touching the same code and most likely moving around that button handler. Whilst conflicts can't be completely avoided in this case I thought commenting out may potentially make life easier

undergroundquizscene: Ah yeah, that makes sense :)

PR #5216, opened 2023-08-31
Fix - Direct Debit popup form close button

marialani:
This is a small fix to the DD form popup close button. The redux state had recently been refactored to use the newer redux toolkit approach and therefore the function needed to close the button now need to be 'called'.

## Why are you doing this?
The popup form close button does not currently close the form.


undergroundquizscene (approved):
Nice, thanks for fixing that!

PR #5221, opened 2023-09-01
Pass referrer through auth flow

kelvin-chappell:
When an auth flow takes place on page load, we lose the referrer for that page.  To get around that, in this PR we store the referrer of the URL when the auth flow begins.  Then, at the end of the auth flow, we expose it as the value of a new query parameter called `pre-auth-ref` so that it can be used for attribution and other analytics.

See also https://github.com/guardian/ophan/pull/5528


rupertbates (approved):
looks good 👍

PR #5226, opened 2023-09-05
Fix client-side validation for the country and state provided by the payment request button

marialani:
As of [PR 4927](https://github.com/guardian/support-frontend/pull/4927), payment method address data is stored separately from the supporter plus form address data. Our existing address validation enforces that a state must be present for applicable countries but this validation is based on the form address data which has led to a recent series of errors. This PR adds a payment details side effect that updates the billing country and state whenever the payment method country and state is changed.

[**Trello Card**](https://trello.com/c/ChpoH4Qd/1501-fix-client-side-validation-for-the-country-and-state-provided-by-the-payment-request-button)

### DIFF
--- a/support-frontend/assets/helpers/redux/checkout/payment/contributionsSideEffects.ts
+++ b/support-frontend/assets/helpers/redux/checkout/payment/contributionsSideEffects.ts
@@ -1,6 +1,13 @@
+import {
+	setBillingCountry,
+	setBillingState,
+} from 'helpers/redux/checkout/address/actions';
+import {
+	setPaymentMethod,
+	setPaymentMethodCountryAndState,
+} from 'helpers/redux/checkout/payment/paymentMethod/actions';
 import type { ContributionsStartListening } from 'helpers/redux/contributionsStore';
 import * as storage from 'helpers/storage/storage';
-import { setPaymentMethod } from './paymentMethod/actions';
 
 export function addPaymentsSideEffects(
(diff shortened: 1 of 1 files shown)


support-frontend/assets/helpers/redux/checkout/payment/contributionsSideEffects.ts
```diff
@@ -11,4 +18,13 @@ export function addPaymentsSideEffects(
...
 	});
+
+	startListening({
+		actionCreator: setPaymentMethodCountryAndState,
+		effect(action, listenerApi) {
+			listenerApi.dispatch(setBillingCountry(action.payload[0]));
+			action.payload[1] &&
+				listenerApi.dispatch(setBillingState(action.payload[1]));
+		},
+	});
```
i-hardy, line 29: If we do need to retain the separate `setPaymentMethodCountryAndState` action, I would suggest that rather than using a listener we use `extraReducers` instead, since all we're doing here is setting more bits of Redux state. cf. the `extraReducer` for the `validateForm` action: https://github.com/guardian/support-frontend/blob/eb4fa9e84801fdd8ddbf69a6adc93d9125aa8487/support-frontend/assets/helpers/redux/checkout/address/reducer.ts#L90

marialani: The plan now is to go with the second approach and get rid of the separate payment method country and state!

PR #5231, opened 2023-09-06
scala-steward: Ignore simple-configuration-ssm

undergroundquizscene:
This [will require a manual update](https://trello.com/c/mvIwywtl/308-support-frontend-support-frontend-update-comgusimple-configuration-ssm-to-158). Until we do that manual update, let’s get scala steward to ignore it so it doesn’t break our scala steward PRs.


marialani (approved):
👍

PR #5236, opened 2023-09-08
Remove existing pre-auth-ref values before adding

kelvin-chappell:
Previously, if a URL already had a `pre-auth-ref` query param another one would be added.  This creates a confusing situation and the first value would probably be taken as the referrer.

In this PR the new value replaces any existing values so that it's very clear what the referrer actually is.

This PR also bumps the version of `ophan-tracker-js` in use for the sake of clarity, although it would actually be covered by the existing specification because only the patch version has changed.

For more context, see:
* #5221 
* https://github.com/guardian/ophan/pull/5529

### DIFF
--- a/support-frontend/app/controllers/AuthCodeFlowController.scala
+++ b/support-frontend/app/controllers/AuthCodeFlowController.scala
@@ -99,16 +99,18 @@ class AuthCodeFlowController(cc: ControllerComponents, authService: AsyncAuthent
      * of the auth flow as it will be the same as the referrer of the page that triggered the auth flow.
      */
     lazy val redirect = {
+      /* This has to correspond with the param name at
+       * https://github.com/guardian/ophan/blob/7cb4283cb6543c009a9e7a0d3d2486744facc902/tracker-js/assets/coffee/ophan/core.coffee#L41
+       */
+      val referrerParamName = "pre-auth-ref"
       val origin = originUrl.getOrElse("/")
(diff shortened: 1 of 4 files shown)


support-frontend/app/controllers/AuthCodeFlowController.scala
```diff
@@ -99,16 +99,18 @@ class AuthCodeFlowController(cc: ControllerComponents, authService: AsyncAuthent
      * of the auth flow as it will be the same as the referrer of the page that triggered the auth flow.
      */
     lazy val redirect = {
+      /* This has to correspond with the param name at
+       * https://github.com/guardian/ophan/blob/7cb4283cb6543c009a9e7a0d3d2486744facc902/tracker-js/assets/coffee/ophan/core.coffee#L41
```
rupertbates, line 103: 👍

support-frontend/test/controllers/AuthCodeFlowControllerTest.scala
```diff
@@ -183,4 +183,31 @@ class AuthCodeFlowControllerTest extends AnyWordSpec with Matchers {
       flash(result).get(FlashKey.authTried) must be(None)
     }
   }
+
+  "AuthCodeFlow.replaceParam" should {
```
rupertbates, line 187: 💯 for adding tests

rupertbates (approved):
Good stuff

PR #5239, opened 2023-09-08
Add alarm and deadletter queue to bigquery-acquisitions-publisher.ts

rupertbates:
#4923 added a new stack for publishing acquisition events to BigQuery but there were a couple of TODOs in it
- add an alarm for the publisher lambda
- add a deadletter queue for the publisher SQS queue
this PR does that.


tomrf1 (approved):
👍

PR #5241, opened 2023-09-08
Validate user’s choice of delivery agent on backend

undergroundquizscene:
Trello Cards: 
  - National Delivery: https://trello.com/c/r557VThC/65-support-frontend-backend-validates-paperround-coverage-of-provided-postcode-and-chosen-agent
  - Conversion: https://trello.com/c/Z29Pxo2C/1525-validate-paperround-coverage-of-users-postcode-by-users-chosen-delivery-agent-on-server-side

The user’s choice of delivery agent needs to be validated on the server side (to guard against requests that haven’t come from the client): we need to validate that a delivery agent was provided, and that according to PaperRound it covers the user’s postcode.

### DIFF
--- a/support-frontend/app/controllers/CreateSubscriptionController.scala
+++ b/support-frontend/app/controllers/CreateSubscriptionController.scala
@@ -8,6 +8,8 @@ import cats.data.EitherT
 import cats.implicits._
 import com.gu.monitoring.SafeLogger
 import com.gu.monitoring.SafeLogger._
+import com.gu.support.catalog.NationalDelivery
+import com.gu.support.paperround.PaperRoundService
 import com.gu.support.workers._
 import config.Configuration.GuardianDomain
 import config.RecaptchaConfigProvider
@@ -26,7 +28,7 @@ import services.AsyncAuthenticationService.IdentityIdAndEmail
 import services.stepfunctions.{CreateSupportWorkersRequest, StatusResponse, SupportWorkersClient}
 import services.{IdentityService, RecaptchaService, StripeSetupIntentService, TestUserService}
 import utils.CheckoutValidationRules.{Invalid, Valid}
-import utils.{CheckoutValidationRules, NormalisedTelephoneNumber}
+import utils.{CheckoutValidationRules, NormalisedTelephoneNumber, PaperValidation}
 
 import java.net.URLEncoder
 import scala.concurrent.{ExecutionContext, Future}
@@ -52,6 +54,7 @@ class CreateSubscriptionController(
     testUsers: TestUserService,
     components: ControllerComponents,
     guardianDomain: GuardianDomain,
+    paperRoundService: PaperRoundService,
 )(implicit val ec: ExecutionContext, system: ActorSystem)
     extends AbstractController(components)
     with Circe {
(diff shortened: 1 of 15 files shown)


support-frontend/app/controllers/CreateSubscriptionController.scala
```diff
@@ -177,24 +179,22 @@ class CreateSubscriptionController(
   private def validate(
       request: Request[CreateSupportWorkersRequest],
       switches: Switches,
-  ): EitherT[Future, CreateSubscriptionError, Unit] = {
+  ): EitherT[Future, CreateSubscriptionError, Unit] = EitherT(for {
 
-    val paymentMethodEnabledValidation =
+    validationRulesResult <- CheckoutValidationRules.validate(request.body, paperRoundService)
```
johnduffell, line 184: here we are changing the validate function from a pure function to one that needs a service.
Of course if it happens to be a national delivery, we don't know if it's valid without the help of paperround, so it's a tricky one.
I wonder whether having a separate check and validate specifically for the postcode if it's national delivery would keep the validate code cleaner?  What do you think?

support-frontend/app/utils/CheckoutValidationRules.scala
```diff
@@ -103,18 +109,23 @@ object CheckoutValidationRules {
...
-    }) match {
-      case Invalid(message) =>
-        Invalid(s"validation of the request body failed with $message - body was $createSupportWorkersRequest")
-      case Valid => Valid
+  def validate(
+      createSupportWorkersRequest: CreateSupportWorkersRequest,
+      paperRound: PaperRoundAPI,
+  )(implicit ec: ExecutionContext): Future[Result] = for {
+    result <- createSupportWorkersRequest.product match {
+      case _: SupporterPlus => Future { PaidProductValidation.passes(createSupportWorkersRequest) }
```
johnduffell, line 117: this change is just to stop it spinning off the contained code in a separate thread
```suggestion
      case _: SupporterPlus => Future.successful { PaidProductValidation.passes(createSupportWorkersRequest) }
```

support-frontend/app/utils/CheckoutValidationRules.scala
```diff
@@ -413,8 +423,27 @@ object PaperValidation {
...
+    case Some(agent) =>
+      paperRound
+        .coverage(RequestBody(postcode))
+        .map(response =>
+          response.data.status match {
+            case CO if response.data.agents.map(_.agentId).contains(agent) => Valid
+            case _ =>
+              Invalid(
+                s"User’s postcode $postcode wasn’t covered by their chosen delivery agent $agent"
+                  + s": PaperRound response was $response",
```
johnduffell, line 443: is this returned to the client or just logged on the server?  We should be a bit careful about including arbitrary content if it might go back to the client

johnduffell (commented):
Thanks for this, well done on getting through all those EitherT's. I've made some comments, the Future.successful(...) one should be simple enough to change them all.  The other two are more things to consider rather than something that definitely can need changing.

support-frontend/app/utils/CheckoutValidationRules.scala
undergroundquizscene: Ah perfect, I wondered about that alright!

johnduffell (approved):

support-frontend/app/controllers/CreateSubscriptionController.scala
undergroundquizscene: I’ve now split the future-requiring validation out and removed the Future from the `validate` call here: would you mind taking a look to see if this looks better?

support-frontend/app/utils/CheckoutValidationRules.scala
undergroundquizscene: Done :)

support-frontend/app/utils/CheckoutValidationRules.scala
undergroundquizscene: Ah ok, that’s a good point – I’ll log the paper round response instead of returning it to the client.

PR #5242, opened 2023-09-08
Omit pre-auth-ref param when page has no referrer

kelvin-chappell:
Previously, going direct to a checkout page, instead of following a link, would redirect to a URL with an empty `pre-auth-ref` parameter in the query string.
eg. `https://support.theguardian.com/uk/contribute?pre-auth-ref=`

With this change, if the page has no referrer the URL will have no `pre-auth-ref` parameter.
eg. `https://support.theguardian.com/uk/contribute`, as expected.

Referrer reporting to Ophan should be unaffected.

Thanks to @GHaberis for spotting this.


GHaberis (approved):
👍

PR #5243, opened 2023-09-08
Add thread sleep time before thank you page

LAKSHMIRPILLAI:
Increasing the sleep time before thank you page.
This is an attempt to fix the intermittently failing post deployment tests


undergroundquizscene (approved):
Hopefully this improves the flakiness we’re seeing 🤞🏻

PR #5244, opened 2023-09-08
Synchronise deliveryAgent on frontend and backend

undergroundquizscene:
These were out of sync, which I hadn’t noticed: the delivery agent returned from PaperRound’s API is a number, whereas the backend expected it from the frontend as a string – I’ve changed this to be a number everywhere.

Also, the backend expects it in the product but the frontend was sending it at the top level of the create request – I’ve moved this to the product.

### DIFF
--- a/support-frontend/assets/helpers/forms/paymentIntegrations/readerRevenueApis.ts
+++ b/support-frontend/assets/helpers/forms/paymentIntegrations/readerRevenueApis.ts
@@ -81,6 +81,7 @@ export type PaperSubscription = {
 	billingPeriod: BillingPeriod;
 	fulfilmentOptions: FulfilmentOptions;
 	productOptions: ProductOptions;
+	deliveryAgent?: number;
 };
 export type GuardianWeeklySubscription = {
 	productType: typeof GuardianWeekly;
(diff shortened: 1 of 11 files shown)


andrewHEguardian (approved):
Looks good 👍  makes sense

johnduffell:
oooh I'd be a bit worried about that, unless paperround specifically say they can be treated as integers.  If they return it as a number in their json we could treat it as a long or a BigDecimal or similar, but String (or a specific type e.g. `case class DeliveryAgentId(value: String) extends AnyVal`) would be safer as we can just treat it opaquely.

undergroundquizscene:
> oooh I'd be a bit worried about that, unless paperround specifically say they can be treated as integers. If they return it as a number in their json we could treat it as a long or a BigDecimal or similar, but String (or a specific type e.g. `case class DeliveryAgentId(value: String) extends AnyVal`) would be safer as we can just treat it opaquely.

Ah, is the range of an integer in Scala less than that of a number in json? I didn’t think to check that. Treating them as an opaque type does seem conceptually nice, but I’m not sure how much value there really is in it: the only advantage I can see is preventing us using number operations on the ID, which seems like something we’re unlikely to do accidentally. (Though using a string to avoid the sizing issues makes total sense!)

PR #5249, opened 2023-09-11
Two step checkout page two nudge

rBangay:
This pr adds to the two step checkout a/b test to add another variant to the second page of the checkout where users are asked to top up the amount they choose on the first page.

[**Trello Card**](https://trello.com/c/NNBogC6e/1520-2-step-checkout-build-order-summary-component-with-the-nudge-for-page-2-of-2-step-checkout)

## Is this an AB test?
- [x] Yes
- [ ]

## Screenshots
Page 2 before | Page 2 after
--- | ---
 | 

With amount selected
mobile_amount_selected

Contributing above threshold (page 1) | Contributing above threshold (page 2)
--- | ---
 | 

Contributing above threshold - starting below (page 1) | Contributing above threshold - starting below (page 2)
--- | ---
 | 

Page 2 - Desktop breakpoint
Screenshot 2023-09-11 at 11 41 18

Page 2 - Desktop breakpoint US edition
Screenshot 2023-09-11 at 12 01 31

Page 2 - Desktop breakpoint Europe edition
Screenshot 2023-09-11 at 12 01 31

### DIFF
--- a/support-frontend/assets/components/checkoutTopUpAmounts/checkoutTopUpAmounts.tsx
+++ b/support-frontend/assets/components/checkoutTopUpAmounts/checkoutTopUpAmounts.tsx
@@ -1,8 +1,3 @@
-import {
-	Button,
-	SvgArrowLeftStraight,
-	SvgArrowRightStraight,
-} from '@guardian/source-react-components';
 import { useState } from 'react';
 import {
 	amountOption,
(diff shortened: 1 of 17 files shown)


support-frontend/assets/components/checkoutTopUpAmounts/checkoutTopUpAmounts.tsx
```diff
@@ -0,0 +1,82 @@
...
+	timePeriod,
+	amounts,
+	isWithinThreshold,
+	handleAmountUpdate,
+	customMargin,
+}: CheckoutTopUpAmountsProps): JSX.Element {
+	const [selectedTopUpamount, setSelectedTopUpamount] = useState<number>(0);
+	const amountsLength = amounts.length;
+
+	return isWithinThreshold ? (
```
i-hardy, line 35: I'd suggest something like the following rather than a ternary

```typescript
if (!isWithinThreshold) return <></>

return (
	<section>
	// blah blah
```

It's just a little bit easier to read and makes it clear immediately that if the condition is false we return nothing, rather than the reader having to scroll all the way to the end of the component to find that out.

support-frontend/assets/components/orderSummary/contributionsOrderSummary.tsx
```diff
@@ -13,14 +13,18 @@ import {
...
+import { simpleFormatAmount } from 'helpers/forms/checkouts';
+import type { Currency } from 'helpers/internationalisation/currency';
 
 const componentStyles = css`
 	${textSans.medium()}
 `;
 
-const summaryRow = css`
-	display: flex;
+const summaryRow = (withFlexWrap?: boolean) => css`
```
i-hardy, line 26:
```suggestion
const summaryRow = (withFlexWrap = false) => css`
```
Very nitpicky, but I think it's preferable to always have a known type and use a default parameter than to have to deal with the ambiguity of `T | undefined` if we don't absolutely have to.

i-hardy (approved):
Couple of small suggestions but otherwise this is looking really good, great work!

PR #5250, opened 2023-09-11
Fix bug resulting in incorrect stage being used by BigQueryService

rupertbates:
https://github.com/guardian/support-frontend/pull/4923 had a bug which meant that any code base which uses the BigQueryService will need to have an environment variable set to specify the stage, if it does not then the code will default to writing events to the CODE bigquery project.

Although the new BigqueryPublisher lambda has this environment variable set, unfortunately support-workers and the payment-api do not and so since the PR was merged production acquisition events for all products other than in app purchases have been going into the CODE biguery project

This PR fixes the bug by requiring the stage which is used by the BigQueryService to be explicitly provided by the calling code.

## Testing
Tested by deploying 
- support-workers
- payment-api
- acquisition-events-api
- bigquery-acquisitions-publisher

to CODE 

then purchased a 
- supporter plus - through the code support site
- one off contribution - through the code support site
- recurring contribution - by pushing an acquisition to the api with postman


support-lambdas/acquisition-events-api/src/main/scala/com/gu/acquisitionEventsApi/Lambda.scala
rupertbates, line 1: This and other deleted files are not used now that we are routing from api gateway to eventbridge

tomrf1 (approved):
👍

PR #5251, opened 2023-09-12
AB Test configuration - how to handle users viewing S+ only variant of checkout

LAKSHMIRPILLAI:

Handle users viewing S+ only variant of checkout.
This change also includes the changes to check the VAT_COMPLIANCE test. On the 2-step checkout, we're going to need to do something to account for this VAT compliant (contributions only) checkout. Without doing anything users in the VAT compliant country list could be allocated to the 2-step checkout with the nudge component - which will allow them to contribute over the  threshold, and acquire S+.

## Why are you doing this?

Currently we bucket all users who land on /contribute into our [abtestDefinitions](https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/helpers/abTests/abtestDefinitions.ts) tests, including those landing on the Supporter Plus Only variant of the checkout.
This isn’t ideal as a lot of the tests we design are for the standard checkout w/ Single Contributions, Recurring Contributions and S+ Contributions. This could lead to misleading results when evaluating tests if we’re bucketing users landing on the Supporter Plus Only variant of the checkout.
We should have a way configure [abtestDefinitions](https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/helpers/abTests/abtestDefinitions.ts) tests so we can choose whether to bucket users on the Supporter Plus Only version of the checkout or not.

## Is this an AB test?
- [x ] Yes
- [ ] No

## Screenshots

###Support site
image

### Supporter Plus in variant
image

### Supporter Plus in control
image

### DIFF
--- a/support-frontend/assets/helpers/abTests/abtestDefinitions.ts
+++ b/support-frontend/assets/helpers/abTests/abtestDefinitions.ts
@@ -90,10 +90,10 @@ export const tests: Tests = {
 		audiences: {
 			ALL: {
 				offset: 0,
-				size: 1,
+				size: 0,
 			},
 		},
-		isActive: true,
+		isActive: false,
 		referrerControlled: false,
 		targetPage: pageUrlRegexes.contributions.allLandingPagesAndThankyouPages,
 		seed: 3,
--- a/support-frontend/assets/helpers/abTests/abtestDefinitions.ts
+++ b/support-frontend/assets/helpers/abTests/abtestDefinitions.ts
@@ -90,10 +90,10 @@ export const tests: Tests = {
 		audiences: {
 			ALL: {
 				offset: 0,
-				size: 0,
+				size: 1,
 			},
 		},
-		isActive: false,
+		isActive: true,
 		referrerControlled: false,
 		targetPage: pageUrlRegexes.contributions.allLandingPagesAndThankyouPages,
 		seed: 3,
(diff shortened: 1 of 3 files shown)


GHaberis:
OK @LAKSHMIRPILLAI I've just had a chance to give this a look. I think there are 2 issues here:

1. This approach will mean users assigned to the `supporterPlusOnly` AB test can **only** be assigned to this one AB test. If we wanted to run an AB test on  the Supporter Plus only checkout in the future we'd therefore be unable to. What I thought we could do is set a rule on an AB test's definition (in `support-frontend/assets/helpers/abTests/abtestDefinitions.ts`) such as `ignoreSupporterPlusOnly`. When allocating user's participations we could check if tghe user was on the Supporter Plus only checkout and we wouldn't assign them to any test with `ignoreSupporterPlusOnly: true` in it's definition.

2. I'm not sure this reducer is the best place to alter the `participations`, I think this logic should sit in the same place where we do the initial participation allocation logic, which is `support-frontend/assets/helpers/abTests/abtest.ts`

i-hardy (approved):
👍 Happy to let this go in, it all looks good.

I think it'd be worth thinking about how we could make this more generic and reusable in the future- say, a field in the A/B test definition that denotes the test as exclusive, so that we wouldn't have to amend this for specific named tests in the future. However that's something we can tackle later.

GHaberis (approved):
Looks good 👍

PR #5254, opened 2023-09-12
Send new Ophan page view event when navigating between page 1 and 2 on the 2-step checkout

GHaberis:

When a user navigates from page 1 to 2 (and vice-versa) of the 2 step checkout this is done via React Router, so no additional network request is made by the browser. This means both page views are covered by a single page view event in the Page View table of the Data Lake.

We'd like each page view to have a separate page view event in the Page View table of the Data Lake, to do this we can use the `sendInitialEvent` function exposed by Ophan's tracker-js library to manually trigger a second Page View event when a user navigates from Page 1 to 2 (and vice-versa). 

This PR introduces a new function to our Ophan module `navigateWithPageView`, that takes care of setting the referral data used by Ophan in localStorage, navigating to the next page and triggering the new page view event.

I've also bumped `ophan` to `1.4.3` to get a change that generates a new page view id when `sendInitialEvent` is called: https://github.com/guardian/ophan/pull/5537

[**Trello Card**](https://trello.com/c/wGDOSGBd/1532-2-step-checkout-manually-trigger-new-page-view-on-react-router-navigate)

### DIFF
--- a/support-frontend/assets/helpers/abTests/abtestDefinitions.ts
+++ b/support-frontend/assets/helpers/abTests/abtestDefinitions.ts
@@ -93,7 +93,7 @@ export const tests: Tests = {
 		audiences: {
 			ALL: {
 				offset: 0,
-				size: 1,
+				size: 0,
 			},
 		},
 		isActive: false,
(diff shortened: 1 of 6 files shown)


rBangay (approved):
👍

PR #5255, opened 2023-09-12
Client-side: Set up A/B test for making zip code a mandatory field for US contributors

i-hardy:

This sets up an A/B test for making the zip code field mandatory for all contribution types and payment methods in the US, instead of only required when paying by card.

The zip code field is moved into the personal details section alongside the state field, and validation messaging is shown for it regardless of payment method if the user is in the A/B variant.

Postcode/zipcode data will be included in the payload for a recurring contribution/S+, but the client side **DOES NOT** currently send this field for one-off contributions. The back end will need to accept this as part of the payload before this can be changed client-side.

[**Trello Card**](https://trello.com/c/pufBPFJv)

## Why are you doing this?

This test is for improved targeting with US contributors.

### Single contribution
Screenshot 2023-09-12 at 12-23-34 Support the Guardian

### Recurring contribution
Screenshot 2023-09-12 at 12-23-59 Support the Guardian

### Form validation
Screenshot 2023-09-12 at 12-24-17 Support the Guardian

### DIFF
--- a/support-frontend/assets/components/personalDetails/personalDetails.tsx
+++ b/support-frontend/assets/components/personalDetails/personalDetails.tsx
@@ -49,6 +49,7 @@ export type PersonalDetailsProps = {
 	onLastNameChange: (lastName: string) => void;
 	signOutLink: React.ReactNode;
 	contributionState: React.ReactNode;
+	contributionZipcode?: React.ReactNode;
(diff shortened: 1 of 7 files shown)


undergroundquizscene (approved):
Looks great :)

PR #5259, opened 2023-09-14
Add omitCountries check on abTestDefinition

GHaberis:

Paired on with @LAKSHMIRPILLAI 

We added a rule in this PR https://github.com/guardian/support-frontend/pull/5251 that would make sure users allocated to the `VAT_COMPLIANCE` test were not included in any other test, implemented as we wanted to be sure they weren't allocated to the `twoStepCheckout` AB test. 

The change was implemented as we assumed the `VAT_COMPLIANCE` would appear in a user's `abParticipations` in state...unfortunately we were mistakenwith this assumption.

We therefore decided to revert back to an original plan for making sure users who experience the `VAT_COMPLIANCE` test are not allocated to the `twoStepCheckout` AB test. We did this by adding a new check when deciding on AB test participations in `abTest.ts`, this check checks a user's country against an `omitCountries` list that can be optionally assigned to an ab test definition. If the user's country is in that list they won't be allocated to the test.

We've set this up on the `twoStepCheckout` test, all the country codes listed are those in the `VAT_COMPLIANCE` list.

### DIFF
--- a/support-frontend/assets/helpers/abTests/abtestDefinitions.ts
+++ b/support-frontend/assets/helpers/abTests/abtestDefinitions.ts
@@ -80,12 +80,12 @@ export const tests: Tests = {
 	},
 	twoStepCheckout: {
 		variants: [
-			// {
-			// 	id: 'control',
-			// },
-			// {
-			// 	id: 'variant_a',
-			// },
+			{
+				id: 'control',
+			},
+			{
+				id: 'variant_a',
+			},
 			{
 				id: 'variant_b',
 			},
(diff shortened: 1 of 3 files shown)


support-frontend/assets/helpers/abTests/abtestDefinitions.ts
```diff
@@ -100,6 +100,45 @@ export const tests: Tests = {
 		referrerControlled: false,
 		targetPage: pageUrlRegexes.contributions.allLandingPagesAndThankyouPages,
 		seed: 3,
+		omitCountries: [
```
rBangay, line 103: Hey, is it alright to call this omitCountries, or should we be more specific as to why we are omitting these countries, incase there is a need to omit a different list of countries for another reason.

rBangay (approved):

PR #5260, opened 2023-09-14
Amend checkout benefits title break points

rBangay:
Amend checkout benefits title to break at a sensible width and … stretch to a single line where possible.

This only affects the two step checkout view, the mutli page (original/control) version stays the same.

## Why are you doing this?
Because Helene asked me to :p

## Screenshots
before | after
--- | ---
 | 

before | after
--- | ---
 | 

before | after
--- | ---
 |


paul-daniel-dempsey (approved):
Perfect

PR #5261, opened 2023-09-15
Checkout page 2 lower threshold check

rBangay:
Do not show the checkout second page top up/nudge amounts to users that select an amount below the supporter plus threshold on the first page

## Screenshots
page 1 before | page 2 before
--- | ---
 | 

page 1 after | page 2 after
--- | ---
 |


GHaberis (approved):
Looks good 👍

paul-daniel-dempsey (approved):
Perfect

PR #5270, opened 2023-09-18
Update payment api to java-11

LAKSHMIRPILLAI:

Update Payment API to JAVA 11
Along with that updating the below dependencies
1) support-payment-api: Update com.dripower:play-circe to 2814.4 from 2814.2
2) support-payment-api: Update mockito-core to 5.3.1 from 4.11.0
3) support-payment-api: Update scaffeine dependency to 5.2.0 from 4.1.0

[**Trello Card**](https://trello.com/c/Lj7Wu48p/1322-update-java-version-in-support-payment-api-from-8-to-11)

## Why are you doing this?

Have updated the github action for the CI, as well as which amigo recipe is used by riff raff. There’s also a debian package specified in the build.sbt: I’m not sure what this is used for, as I would’ve expected the JRE to be provided by amigo.

### DIFF
--- a/support-payment-api/src/main/resources/riff-raff.yaml
+++ b/support-payment-api/src/main/resources/riff-raff.yaml
@@ -10,7 +10,7 @@ deployments:
     parameters:
       amiParameter: AMIPaymentapi
       amiTags:
-        Recipe: jammy-membership-java8
+        Recipe: jammy-membership-java
         AmigoStage: PROD
       amiEncrypted: true
       templateStagePaths:
--- a/support-payment-api/src/main/resources/riff-raff.yaml
+++ b/support-payment-api/src/main/resources/riff-raff.yaml
@@ -10,7 +10,7 @@ deployments:
     parameters:
       amiParameter: AMIPaymentapi
       amiTags:
-        Recipe: jammy-membership-java
+        Recipe: jammy-membership-java11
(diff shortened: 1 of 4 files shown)


LAKSHMIRPILLAI:
These changes were deployed to Payment API CODE successfully

image

support-payment-api/src/main/resources/riff-raff.yaml
```diff
@@ -10,7 +10,7 @@ deployments:
     parameters:
       amiParameter: AMIPaymentapi
       amiTags:
-        Recipe: jammy-membership-java8
+        Recipe: jammy-membership-java
```
undergroundquizscene, line 13: Based on https://amigo.gutools.co.uk/recipes, I would’ve expected this to be [jammy-membership-java11](https://amigo.gutools.co.uk/recipes/jammy-membership-java11). I assume I’ve missed something, since this works in CODE – do you know what?

undergroundquizscene (approved):
After deploying to CODE, did you test by submitting some single contributions? Just in case there are errors that only surface when the code is run.

support-payment-api/src/main/resources/riff-raff.yaml
undergroundquizscene: Ah perfect, that explains it!

PR #380, opened 2017-11-20
UK and US Recurring amounts tests

jranks123:
This PR is a combination of 2 tests which tests different suggested amounts for US and UK recurring contributions.

The UK has the following amount variants:
Control: 5, 10, 20
Lower: 2, 5, 10
Wildcard: 5, 7 20

The US has the following variants:
Control: 5, 10, 20
Lower: 2, 5, 10
Higher: 7, 15, 30

 @guardian/contributions 
Screenshots:

Control UK
ukcontrol

Lower UK:
loweruknew

Wildcard UK:
wildcarduk

Control US:
uscontrol

Lower US:
uslowernew

Higher US:

ushigher

PR #5271, opened 2023-09-18
Require zip code for US contributions (backend changes)

undergroundquizscene:
We want to collect zip codes for US contributions, and this change adds an optional postal code field to requests to the payment API to hold these values. It also stores users’ addresses in Zuora for supporter plus and for contributions. Finally, I’ve changed the frontend to send the postal code to the backend for single contributions. This isn’t limited to the US: if we’ve collected a postal code it’ll be sent.

I’ve gone with “postal code” rather than “zip code” because even though we’re only storing it for US contributions at the moment, we may expand that in future, and “postal code” appears to be the more general term, judging by [the Wikipedia page](https://en.wikipedia.org/wiki/Postal_code).

- Trello Card: https://trello.com/c/VSe61QY5/1436-support-checkout-back-end-changes-card-move-zip-code-out-of-the-card-payment-section-across-both-single-recurring-in-us-and-make

### DIFF
--- a/support-frontend/assets/helpers/tracking/__tests__/__snapshots__/acquisitionsTest.ts.snap
+++ b/support-frontend/assets/helpers/tracking/__tests__/__snapshots__/acquisitionsTest.ts.snap
@@ -30,6 +30,7 @@ Object {
   "labels": undefined,
   "pageviewId": "123456",
   "platform": "SUPPORT",
+  "postalCode": null,
   "queryParameters": Array [
     Object {
       "name": "param1",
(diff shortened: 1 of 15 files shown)


undergroundquizscene:
Tested single contributions by deploying support-frontend and the payment-api to CODE, making stripe and paypal contributions and confirming that the corresponding record in contributions-store-CODE contains the postcode entered.

Tested recurring contributions by deploying to support-workers and verifying that the postcode makes it to the sold-to/bill-to section on the right-hand-side of the customer view in Zuora.

undergroundquizscene:
Before release, I will need to connect to the contributions-store-PROD database and add the new “postal_code” field, and then merge [contributions-platform PR #380](https://github.com/guardian/contributions-platform/pull/380/files).

i-hardy:
Client side changes all look good to me 👍

undergroundquizscene:
> Client side changes all look good to me 👍

Perfect, thanks for taking a look :)

LAKSHMIRPILLAI (approved):
👍🏼 LGTM

AndreaDiotallevi (approved):

PR #5274, opened 2023-09-19
DO_NOT_MERGE: Hardcode some test agents

emdash-ie:
PaperRound’s test API is currently not returning any agents for the postcodes we’ve been using. So that we can keep testing in CODE, this commit hardcodes responses for those postcodes.

This PR should not be merged.


emdash-ie:
Closing this as it’s now old – I figure its usefulness is past

PR #5275, opened 2023-09-19
Make storage of country and state consistent for Apple/Google Pay payments

i-hardy:

This switches over the storage of country and state information for Apple and Google Pay payments from a separate section in the payment method reducer to using the address reducer that's used when a user fills out this information manually.

[**Trello Card**](https://trello.com/c/ChpoH4Qd)

## Why are you doing this?

This should help to address an issue where client-side validation is not correctly catching issues with a user's address information obtained from the payment request button interface. Using the address reducer means our existing validation for a user's country and state will run and should prevent form submission if there are any issues.

### DIFF
--- a/support-frontend/assets/components/paymentRequestButton/hooks/payerDetails.ts
+++ b/support-frontend/assets/components/paymentRequestButton/hooks/payerDetails.ts
@@ -3,6 +3,7 @@ import type {
 	PaymentRequestPaymentMethodEvent,
 } from '@stripe/stripe-js';
 import {
+	detect as detectCountry,
 	findIsoCountry,
 	stateProvinceFromString,
 } from 'helpers/internationalisation/country';
@@ -80,4 +81,8 @@ export function addPayerDetailsToRedux(
 
 export function resetPayerDetails(dispatch: ContributionsDispatch): void {
 	dispatch(setEmail(storage.getSession('gu.email') ?? ''));
+	dispatch(setFirstName(''));
+	dispatch(setLastName(''));
+	dispatch(setBillingCountry(detectCountry()));
+	dispatch(setBillingState(''));
 }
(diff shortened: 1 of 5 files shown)


undergroundquizscene (approved):
Looks good!

I’d also like to note that this is undoing [changes I originally made](https://github.com/guardian/support-frontend/pull/4927/commits/d8758db0c1706d48c4c308fb178dbd59e610f738) where I didn’t realise I’d broken the validation – sorry about that :(

In that PR I also removed a lot of [resetPayerDetails](https://github.com/guardian/support-frontend/blob/33ee3f781bad764030a1fa4f64b01125b596be94/support-frontend/assets/components/paymentRequestButton/hooks/payerDetails.ts#L82) – would it be worth restoring [these lines](https://github.com/guardian/support-frontend/pull/4927/commits/d8758db0c1706d48c4c308fb178dbd59e610f738#diff-2c06498093b6d0f083a30585ce19e77f183f614962c0353cf5dc713d07ea1c05L82-L86), and detecting the state value instead of setting it to an empty string? That should prevent users running into trouble when a PRB payment fails.

i-hardy:
I've restored those lines but I think we're fine to just reset the state- if the user attempts a PRB payment again the value will just get overwritten by whatever's in their Apple/Google Pay data, and otherwise we expect users to enter this information manually and don't auto-populate the field in the same way as we try to detect the country/region.

undergroundquizscene:
> I've restored those lines but I think we're fine to just reset the state- if the user attempts a PRB payment again the value will just get overwritten by whatever's in their Apple/Google Pay data, and otherwise we expect users to enter this information manually and don't auto-populate the field in the same way as we try to detect the country/region.

Ah yes, good point! I’d forgotten we stopped collecting an auto-detected state.

PR #5279, opened 2023-09-20
Update java 8 to 11 for payment-api

LAKSHMIRPILLAI:

Update Payment API  to JAVA 11 
Along with that updating the below dependencies
1) support-payment-api: Update mockito-core to 5.3.1 from 4.11.0
2) support-payment-api: Update scaffeine dependency to 5.2.0 from 4.1.0

[**Trello Card**](https://trello.com/c/Lj7Wu48p/1322-update-java-version-in-support-payment-api-from-8-to-11)

## Why are you doing this?

Have updated the github action for the CI, as well as which amigo recipe is used by riff raff. There’s also a debian package specified in the build.sbt: I’m not sure what this is used for, as I would’ve expected the JRE to be provided by amigo.

### DIFF
--- a/.github/workflows/payment-api-build.yml
+++ b/.github/workflows/payment-api-build.yml
@@ -50,11 +50,11 @@ jobs:
           role-to-assume: ${{ secrets.GU_RIFF_RAFF_ROLE_ARN }}
           aws-region: eu-west-1
 
-      - name: Setup Java 8
+      - name: Setup Java 11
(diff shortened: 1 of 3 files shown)


undergroundquizscene (approved):

LAKSHMIRPILLAI:
I deployed changes to support-frontend and payment api CODE and they went in without issues.Also tried to raise one single contribution and  one  recurring contribution and both were successful

PR #5282, opened 2023-09-21
Send acquisition events to BigQuery via the event bus

rupertbates:
Following on from #4923 which created a new acquisition event bus, sqs queue and lambda to handle writing acquisition data to BigQuery, this PR update the support-workers `SendAcquisitionEvent` class to also use the event bus rather than writing to BigQuery directly.

The next step will be to do the same for the payment-api and then to trigger writes to the Kinesis stream from the event bus.

I have left a number of TODOs in the code, these will be removed in the upcoming PRs

### DIFF
--- a/support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/eventbridge/AcquisitionsEventBusService.scala
+++ b/support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/eventbridge/AcquisitionsEventBusService.scala
@@ -56,22 +56,12 @@ class AcquisitionsEventBusService(source: String, stage: Stage, client: AmazonEv
 }
 
 object AcquisitionsEventBusService {
-//  lazy val CredentialsProvider = AwsCredentialsProviderChain.builder
-//    .credentialsProviders(
-//      ProfileCredentialsProvider.builder.profileName(ProfileName).build,
-//      InstanceProfileCredentialsProvider.builder.asyncCredentialUpdateEnabled(false).build,
-//      EnvironmentVariableCredentialsProvider.create(),
-//    )
-//    .build
 
   lazy val eventBridgeClient = AmazonEventBridgeClient
     .builder()
     .withRegion(Regions.EU_WEST_1)
     .withCredentials(CredentialsProvider)
     .build
-//    .region(Region.EU_WEST_1)
-//    .credentialsProvider(CredentialsProvider)
-//    .build
 
   /** @param source
     *   \- A string which is passed to the source value of Eventbridge to identify the system that this event comes from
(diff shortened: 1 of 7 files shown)


support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/eventbridge/AcquisitionsEventBusService.scala
```diff
@@ -0,0 +1,86 @@
...
+        val errorMessage = s"There was an exception writing ${acquisitionJson.spaces2} to Eventbridge: ${e.getMessage}"
+        SafeLogger.error(scrub"$errorMessage", e)
+        promise.success(Left(errorMessage))
+    }
+    promise.future
+  }
+}
+
+object AcquisitionsEventBusService {
+//  lazy val CredentialsProvider = AwsCredentialsProviderChain.builder
```
tomrf1, line 59: can this be removed?

rupertbates: Yes it can, done now!

tomrf1 (approved):
looks good!

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/eventbridge/AcquisitionsEventBusService.scala
```diff
@@ -0,0 +1,76 @@
...
+    entry.setEventBusName(eventBusName)
+    entry.setSource(source)
+    entry.setDetailType(detailType)
+    entry.setDetail(acquisitionJson.noSpaces)
+    entry.setTime(Date.from(Instant.now))
+
+    val putEventsRequest = new PutEventsRequest
+    putEventsRequest.setEntries(List(entry).asJava)
+
+    val promise = Promise[Either[String, Unit]]
```
johnduffell, line 36: promis is a type?  I'm not sure how to read this code - is the try on the next line a parameter to the Promise.apply?

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/eventbridge/AcquisitionsEventBusService.scala
```diff
@@ -0,0 +1,76 @@
...
+
+import java.time.Instant
+import java.util.Date
+import scala.concurrent.Promise
+import scala.jdk.CollectionConverters.SeqHasAsJava
+
+class AcquisitionsEventBusService(source: String, stage: Stage, client: AmazonEventBridge) {
+  val eventBusName = s"acquisitions-bus-${stage.toString}"
+  val detailType = "AcquisitionsEvent"
+  def putAcquisitionEvent(acquisition: AcquisitionDataRow) = {
```
johnduffell, line 23: is that return type a future?
```
def putAcquisitionEvent(acquisition: AcquisitionDataRow): Future[Either[String,Unit]] = {
```

PR #5287, opened 2023-09-22
Payment-api event bus

rupertbates:

## Is this an AB test?
- [ ] Yes
- [ ] No


support-payment-api/src/main/scala/backend/AmazonPayBackend.scala
```diff
@@ -268,14 +268,11 @@ object AmazonPayBackend {
...
-        .map(config =>
-          BigQueryService.build(
-            if (env == Live) PROD else CODE,
-            config,
-          ),
-        ): InitializationResult[BigQueryService],
+      // I think this initialisation code is correct because at this point the env variable has been set
+      // correctly to live or test based on whether the environment is PROD or CODE and whether the user is
+      // a test user. The AcquisitionEventBusService class has some logic within it to deal with test users,
+      // but that is not needed in this case so we just pass in `isTestUser = false`
```
tomrf1, line 274: This makes sense, but maybe it would be less confusing to add an alternative constructor?
`def apply(source: String, stage: Stage)`

rupertbates: Yes, I think you're right

tomrf1 (approved):
great!

PR #5293, opened 2023-09-26
Update version of CDK

undergroundquizscene:
We think the version mismatch may be the cause of a current
outage (issue with S3 not accepting our signatures) in the supporter
product data step function: updating this to see if it fixes the
problem.

-------

## Is this an AB test?
- [ ] Yes
- [ ] No

## Screenshots


undergroundquizscene:
Closing this as it was not the cause of the error

PR #5294, opened 2023-09-26
fix: Only pass content length to S3 if positive

undergroundquizscene:
This should fix errors writing the zuora batch request results to S3 when Zuora doesn’t include a Content-Length header in its response.


undergroundquizscene:
For anyone finding this PR in the future: Zuora [stopped including a Content-Length header in compressed responses last November](https://github.com/guardian/support-frontend/pull/4470), and recently stopped including the header on uncompressed responses as well, hence this fix. The underlying HTTP library returns a -1 when the content-length is unknown, which we were passing on to S3 when uploading a file, resulting in a misleading error about checking our access key.

undergroundquizscene:
We also [had to deal with this problem in support-service-lambdas in March](https://github.com/guardian/support-service-lambdas/pull/1888), which suggests we may need to check that now as well.

PR #5297, opened 2023-09-28
National Delivery - swap Home Delivery and Sub card tabs

andrewHEguardian:

Swapping the Home delivery and Sub card tabs on the paper subscriptions landing page, as well as the header links to that page to ensure the HD tab opens first. (This is all behind a 0% AB Test).

[**Trello Card**](https://trello.com/c/jzG8cDMw/61-support-frontend-flip-hnd-and-sub-card-options)

## Why are you doing this?

To give more prominence to home delivery once it has opened up outside the M25.

## Screenshots

Before
image

After
image

### DIFF
--- a/support-frontend/assets/components/headers/header/header.tsx
+++ b/support-frontend/assets/components/headers/header/header.tsx
@@ -14,7 +14,7 @@ export type PropTypes = {
 	utility?: JSX.Element;
 	countryGroupId: CountryGroupId;
 	display?: 'navigation' | 'checkout' | 'guardianLogo' | void;
-	participations: Participations;
+	participations?: Participations;
 };
 export type State = {
 	isTestUser: boolean | null | undefined;
--- a/support-frontend/assets/components/headers/header/header.tsx
+++ b/support-frontend/assets/components/headers/header/header.tsx
@@ -1,6 +1,7 @@
 // ----- Imports ----- //
 import { Component } from 'react';
 import SvgGuardianLogo from 'components/svgs/guardianLogo';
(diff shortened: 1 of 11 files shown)


support-frontend/assets/components/headers/header/header.tsx
```diff
@@ -13,6 +14,7 @@ export type PropTypes = {
 	utility?: JSX.Element;
 	countryGroupId: CountryGroupId;
 	display?: 'navigation' | 'checkout' | 'guardianLogo' | void;
+	participations?: Participations;
```
andrewHEguardian, line 17: Because this is optional, not every use of the header is directing to the `/delivery` version of the url (eg error pages). Maybe I should go back and add this in but I focused on the other landing pages for now

support-frontend/assets/pages/paper-subscription-landing/paperSubscriptionLandingPage.tsx
```diff
@@ -52,6 +52,7 @@ const pageQaId = 'qa-paper-subscriptions';
 function PaperLandingPage({
 	productPrices,
 	promotionCopy,
+	participations,
 }: PaperLandingContentPropTypes) {
 	const sanitisedPromoCopy = getPromotionCopy(promotionCopy);
 	const fulfilment: PaperFulfilmentOptions = window.location.pathname.includes(
```
andrewHEguardian, line 58: an alternative solution might to check the test and swap the fulfilment option here, but not sure if swapping the URL when the page loads is a great idea

PR #5298, opened 2023-09-28
National Delivery - swap Home Delivery and Sub card tabs

andrewHEguardian:

Swapping the order of Home delivery and Sub card tabs on the paper subscriptions landing page, and appending`/delivery` when landing on this page. (This is behind a 0% AB Test).

I tried a slightly different approach where I updated the links to this page to use `paper/delivery`, as seen in this WIP PR #5297, but the approach in this PR touches a lot fewer places.

[**Trello Card**](https://trello.com/c/jzG8cDMw/61-support-frontend-flip-hnd-and-sub-card-options)

## Why are you doing this?

To give more prominence to home delivery once it has opened up outside the M25.

## Screenshots

Before
image

After
image

### DIFF
--- a/support-frontend/assets/pages/paper-subscription-landing/paperSubscriptionLandingPage.tsx
+++ b/support-frontend/assets/pages/paper-subscription-landing/paperSubscriptionLandingPage.tsx
@@ -52,6 +52,7 @@ const pageQaId = 'qa-paper-subscriptions';
 function PaperLandingPage({
 	productPrices,
 	promotionCopy,
+	participations,
 }: PaperLandingContentPropTypes) {
 	const sanitisedPromoCopy = getPromotionCopy(promotionCopy);
 	const fulfilment: PaperFulfilmentOptions = window.location.pathname.includes(
@@ -80,7 +81,9 @@ function PaperLandingPage({
 	return (
 		<Page
 			id={pageQaId}
-			header={<Header countryGroupId={GBPCountries} />}
+			header={
+				<Header countryGroupId={GBPCountries} participations={participations} />
+			}
 			footer={paperSubsFooter}
 		>
 			<PaperHero
@@ -94,6 +97,7 @@ function PaperLandingPage({
 							<Tabs
 								selectedTab={selectedTab}
 								setTabAction={handleSetTabAction}
+								participations={participations}
 							/>
 						</div>
 					</Block>
@@ -106,6 +110,7 @@ function PaperLandingPage({
 						productPrices={productPrices}
 						tab={selectedTab}
 						setTabAction={setSelectedTab}
+						participations={participations}
 					/>
 				</CentredContainer>
 			</FullWidthContainer>
--- a/support-frontend/assets/pages/paper-subscription-landing/paperSubscriptionLandingPage.tsx
+++ b/support-frontend/assets/pages/paper-subscription-landing/paperSubscriptionLandingPage.tsx
@@ -59,20 +59,20 @@ function PaperLandingPage({
 	const isNationalDeliveryAbTestVariant =
 		participations.nationalDelivery === 'variant';
 
-	const [redirectToHomeDelivery, setRedirectToHomeDelivery] = useState<boolean>(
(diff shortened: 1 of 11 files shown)


support-frontend/assets/pages/paper-subscription-landing/paperSubscriptionLandingPage.tsx
```diff
@@ -52,13 +52,29 @@ const pageQaId = 'qa-paper-subscriptions';
...
+	const [shouldRedirectToHomeDelivery, setShouldRedirectToHomeDelivery] =
+		useState<boolean>(isNationalDeliveryAbTestVariant);
+
+	if (
+		shouldRedirectToHomeDelivery &&
+		!window.location.pathname.includes('delivery')
+	) {
+		window.history.replaceState({}, '', paperSubsUrl(true));
+		setShouldRedirectToHomeDelivery(false);
+	}
```
undergroundquizscene, line 71: So the logic here is: if we’ve landed on the paper landing page and the national delivery AB test participation is in the variant, change the user’s URL to include `delivery` at the end, and register that we’ve done that so we don’t do it twice – right?

I dislike the weirdness of landing on `…/paper/delivery` (after redirection) and clicking the other tab to go to `…/paper`, but I’m not sure there’s a straightforward alternative. To my mind the fulfilment option makes more sense as a query parameter (e.g. `…/paper?fulfilmentOption=HomeDelivery`), and then if it’s not set we’d just assume a value, the AB test changing the order of the tabs and which value we assume. But I don’t know this code well enough to be confident that there isn’t a good reason it’s in the path, and I’m not sure I want us to get sucked in to changing it now.

What do you think?

support-frontend/assets/pages/paper-subscription-landing/paperSubscriptionLandingPage.tsx
```diff
@@ -52,13 +52,29 @@ const pageQaId = 'qa-paper-subscriptions';
...
+		participations.nationalDelivery === 'variant';
+
+	const [shouldRedirectToHomeDelivery, setShouldRedirectToHomeDelivery] =
+		useState<boolean>(isNationalDeliveryAbTestVariant);
+
+	if (
+		shouldRedirectToHomeDelivery &&
+		!window.location.pathname.includes('delivery')
+	) {
+		window.history.replaceState({}, '', paperSubsUrl(true));
```
undergroundquizscene, line 69: Mind if we include the argument name, to make it clearer what it means?

```suggestion
		window.history.replaceState({}, '', paperSubsUrl(withDelivery = true));
```

undergroundquizscene (approved):

support-frontend/assets/pages/paper-subscription-landing/paperSubscriptionLandingPage.tsx
andrewHEguardian: yep: that is the logic 👍 

I agree with the weirdness, although that was the existing functionality. Could definitely spend more time refactoring the page to use query params instead, but I don't the code either

support-frontend/assets/pages/paper-subscription-landing/paperSubscriptionLandingPage.tsx
andrewHEguardian: alas this isn't a Javascript feature (unless I'm mistaken) but I wish it was

undergroundquizscene: Oh I see! Glad one of us knows 🙃

undergroundquizscene: Probably "works" because assignment returns the value, a la C? Anyway, no worries

support-frontend/assets/pages/paper-subscription-landing/paperSubscriptionLandingPage.tsx
undergroundquizscene: Let’s leave it

PR #5300, opened 2023-09-28
Add National Delivery to PROD catalog

undergroundquizscene:
I added the new National Delivery CODE product rate plans (from the api sandbox) to the catalog a while ago, but did not remember to add the PROD ones: this commit does that. (These IDs were taken from the HTML source of the zuora web UI, specifically from the “export product” button link.)

To update the catalog files I ran the following command, which sorts the products by their ID and will hopefully minimise diffs in future:

```sh
aws s3 cp --profile membership s3://gu-zuora-catalog/PROD/Zuora-PROD/catalog.json - | jq ".products |= sort_by(.id)" > catalog-prod.json
```


support-models/src/main/scala/com/gu/support/catalog/Product.scala
```diff
@@ -195,6 +195,9 @@ case object Paper extends Product {
         homeDelivery("2c92a0ff560d311b0156136f2afe5315", Sixday, "Home Delivery Sixday"),
         homeDelivery("2c92a0fd560d132301560e43cf041a3c", EverydayPlus, "Home Delivery Everyday+"),
         homeDelivery("2c92a0fd560d13880156136b72e50f0c", Everyday, "Home Delivery Everyday"),
+        nationalDelivery("8a12999f8a268c57018a27ebe868150c", Weekend, "National Delivery Weekend"),
+        nationalDelivery("8a12999f8a268c57018a27ebfd721883", Sixday, "National Delivery Sixday"),
+        nationalDelivery("8a12999f8a268c57018a27ebe31414a4", Everyday, "National Delivery Everyday"),
```
johnduffell, line 200: why is voucher separate (`prodCollection`) but home delivery and national delivery combined?
should we either combine them all, or separate them all (probably separate all makes a bit more sense)

johnduffell (approved):
looks good but see comment about inconsistency of grouping products

support-models/src/main/scala/com/gu/support/catalog/Product.scala
undergroundquizscene: Ah, I hadn’t noticed that! Sure, happy to separate them

PR #5301, opened 2023-09-29
Fix SupporterDataDynamoService permission for stripe-patrons-data

tomrf1:
The previous PR to use snapstart needed to add a new credentials provider - [see here.](https://github.com/guardian/support-frontend/pull/4869/files#diff-2b20bde739cbd13a3d9c392f38c53ca2196a6531f4a25800cf013e5e824a5096R18)

But these lambdas also use a dynamodb service which defines its credentials provider differently (using the newer aws sdk). This means we get a permissions error when using dynamodb.

Tested in CODE by creating a test user with patron subscription and running the `stripe-patrons-data-CODE` lambda:
Screenshot 2023-09-29 at 14 01 02


rupertbates:
It would be great if we could start to move all the aws stuff over to V2, it's a bit of a mess at the moment when you have both in the same project.

PR #5304, opened 2023-09-29
Add .trim() to name validation on checkout

iainjchambers-guardian:
Adding the `.trim()` command to validation for both first and last name fields on the checkout.

[**Trello Card**](https://trello.com/c/kvnhy64j/1575-last-name-field-validation-in-the-checkout-to-block-an-empty-space)

## Why are you doing this?
Currently the client side validation does not complain when those two text fields contain only blank spaces.

## Is this an AB test?
- [ ] Yes
- [x] No


GHaberis (approved):
Nice 👍

PR #5311, opened 2023-10-02
Fix national delivery checkout: set single delivery agent in reducer, not component

andrewHEguardian:

@undergroundquizscene spotted a bug in the national delivery checkout - when choosing a postcode with a single provider option the page hangs (and was infinitely rerendering). 

We investigated and found that the cause was the `addAddressMetaSideEffects` listener listening for changes to the `setDeliveryAgent` redux action and also calling this action in the component (when only one provider is available).

This PR instead sets the chosen delivery agent when only one option is available in the reducer.

## Is this an AB test?
- [ ] Yes
- [x] No

### DIFF
--- a/support-frontend/assets/helpers/redux/checkout/addressMeta/reducer.ts
+++ b/support-frontend/assets/helpers/redux/checkout/addressMeta/reducer.ts
@@ -17,11 +17,8 @@ export const addressMetaSlice = createSlice({
 		setDeliveryAgent(state, action: PayloadAction<number | undefined>) {
 			state.deliveryAgent.chosenAgent = action.payload;
 		},
-		setDeliveryAgentResponse(
-			state,
-			action: PayloadAction<DeliveryAgentsResponse | undefined>,
-		) {
-			state.deliveryAgent.response = action.payload;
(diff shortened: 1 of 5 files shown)


support-frontend/assets/helpers/redux/checkout/addressMeta/reducer.ts
```diff
@@ -17,11 +17,8 @@ export const addressMetaSlice = createSlice({
 		setDeliveryAgent(state, action: PayloadAction<number | undefined>) {
 			state.deliveryAgent.chosenAgent = action.payload;
 		},
-		setDeliveryAgentResponse(
-			state,
-			action: PayloadAction<DeliveryAgentsResponse | undefined>,
-		) {
-			state.deliveryAgent.response = action.payload;
+		clearDeliveryAgentResponse(state) {
```
undergroundquizscene, line 20: For the record: We renamed this to make it clearer what it’s being used for, because we initially added code here instead of in the “thunk fulfilled” reducer below.

undergroundquizscene (approved):

PR #5312, opened 2023-10-02
Reinstate indexing after auth flow on page load

kelvin-chappell:
On pages where an auth flow takes place on page load, the response from Gateway has a response header `X-Robots-Tag: noindex,nofollow`, which might be giving Google a hint not to index content at the end of the redirect chain.

This is an attempt to see if we can give an opposing hint in the final redirect to the content.

The test will be if the page can now be indexed.

Alternatively, bots could be being sent in a loop.  But let's see if this works first.


GHaberis (approved):
Looks good 👍

kelvin-chappell:
Unfortunately, this didn't fix the indexing.  A request to re-index in the Google search gave an error response complaining about redirects.

PR #5314, opened 2023-10-03
National Delivery - update validation message on paper postcode

andrewHEguardian:

Updating the validation message in Paper checkout when a user selects a postcode we cannot deliver to. 

[**Trello Card**](https://trello.com/c/4TtvEoqc/62-support-frontend-error-message-on-postcode-validation)

## Why are you doing this?

This is part of the National Delivery project, but the message applies to normal checkouts because we no longer advertise vouchers.

## Is this an AB test?
- [ ] Yes
- [x] No


johnduffell (approved):
ha I just saw this error about 1 minute ago too 👍

PR #5317, opened 2023-10-04
Revert payment-api java back to 8 from 11

LAKSHMIRPILLAI:

Downgrading the payment-api java from 11 to 8

[**Trello Card**][(https://trello.com](https://trello.com/c/NvMKiklp/1574-amazonpay-amountnotsetthe-seller-has-not-set-the-amount-for-the-order-reference))

## Why are you doing this?
The payment-api was upgraded to 11 previously in the [PR](https://github.com/guardian/support-frontend/pull/5279)
Post this change, we started getting a few ALARM: "[CDK] payment-api PROD Amazon Pay payment error for one-off contribution..." in EU (Ireland) which showed  Encountered marshalling error while marshalling data  and AmountNotSet:The seller has not set the amount for the Order Reference errors for a few users.
So we stopped offering Amazon Pay as a payment method for One-Off Contributions in US.

Later upon investigation , it was found that this was caused due to the upgrade of payment-api from 8 to 11 and there was  this dependency amazon-pay-java-sdk with a version 3.6.2  which is incompatible with java 11.

Upgrading this to a version higher than 3.6.3 caused our payment-api related tests to fail with the error below
image

After further investigation it was decided that it was best to downgrade payment api from 11 to 8 instead of upgrading the  amazon-pay-java-sdk dependency

## Tests 
I deployed  these  changes to payment-api and also deployed to support-frontend (latest main ) to CODE and tested different payment options, all are working fine for me

After merging , we have to Switch On Amazon Pay for One-off contributions in RRCP


undergroundquizscene:
> Later upon investigation , it was found that […] there was this dependency amazon-pay-java-sdk with a version 3.6.2 which is incompatible with java 11.

It’s worth noting for the record that we only _think_ this is the case: we’re not entirely sure where the error is coming from, and it being a marshalling error suggests it’s related to json or xml serialisation somehow. (Though we do know that the upgrade to java 11 was the cause.)

undergroundquizscene (approved):

jamesgorrie (approved):
This looks good if the tests are passing! But I am still unfamiliar with the stack.

Should we be adding a health issue to update to 11 while ensuring we update the `payment-api` => `3.6.3`?

PR #5319, opened 2023-10-04
Remove code which writes to Kinesis

rupertbates:
## What does this change?
As part of the work described in [this document](https://docs.google.com/document/d/161e602tyFTtD33P06ZqAVHjdkBIQzA6rnV6Jf5Noyds/edit) we want to stop sending acquisition events to Kinesis from support-workers and payment-api and instead use the EventBridge event bus which these events are now being sent to.

This PR depends on [this contributions-platform PR](https://github.com/guardian/contributions-platform/pull/392) which sets up an EventBridge rule to send any events on the `acquisitions-bus` where the source is support-workers or payment-api to the `acquisitions-stream` Kinesis stream.

This PR  removes the existing code that writes to Kinesis and changes the `source` property of acquisition events from `support-workers` and `payment-api` to `support-workers.1` and `payment-api.1`. 
The reason we are changing the source is so that we can put the contributions-platform PR live without worrying about duplicates; the source of acquisition events will only match the rule in that PR when this PR which removes the old write mechanism is live.


tomrf1 (approved):
great!

PR #5321, opened 2023-10-05
Accept variant in mandatoryZipCode AB test

undergroundquizscene:
When we merged the PR with these changes, we forgot to enable the test! This PR removes the test, accepting the variant.


rBangay (approved):
👍

PR #5327, opened 2023-10-06
Give post-deployment availability test access to cookies

kelvin-chappell:
I had thought that the reason the availability test that runs before the Selenium tests was failing was something to do with the interaction with bots.  But it's actually because the OkHttp client doesn't store cookies by default.  So I've fixed that but it hasn't helped to understand the Google indexing problem.


GHaberis (approved):
Looks good 👍

PR #5328, opened 2023-10-06
Attempt 2 to reinstate indexing after auth flow on page load

kelvin-chappell:
This is the second attempt to get indexing working after the auth flow on checkout pages.
Reading the [documentation](https://developers.google.com/search/docs/crawling-indexing/robots-meta-tag#xrobotstag) on the `X-Robots-Tag` header, I realise that I may have given it invalid values on the first attempt.  I've now corrected that.

> all: There are no restrictions for indexing or serving. This rule is the default value

The first attempt was at #5312.


GHaberis (approved):
👍

kelvin-chappell:
Unfortunately this didn't work either, apparently because Google indexer doesn't like the number of redirects involved in the auth flow.
Now taking a new approach in #5337.

PR #5330, opened 2023-10-06
WIP: National Delivery: thank you email changes

undergroundquizscene:
- use new extension name
- include delivery agent details


undergroundquizscene:
I have verified in CODE that sending the thank you email succeeds (once the parameter store config for support-frontend has been updated, as well as the S3 config for support-workers), as visible from this support-workers execution: https://eu-west-1.console.aws.amazon.com/states/home?region=eu-west-1#/v2/executions/details/arn:aws:states:eu-west-1:865473395570:execution:support-workers-CODE:0d7c34f6-5321-42a9-b53f-95ce4ef720b3

undergroundquizscene:
I have updated the CODE and PROD configs in both the parameter store and the support-workers config S3 buckets

undergroundquizscene:
Though I should possibly fix it to remove that `Some(1730)`

undergroundquizscene:
Ok, I’ve removed the delivery agent ID:

```
{
  "To" : {
    "Address" : "emily.bourke@guardian.co.uk",
    "ContactAttributes" : {
      "SubscriberAttributes" : {
        "subscription_rate" : "£56.23 every month for 12 months, then £73.99 every month",
        "date_of_first_paper" : "Friday, 13 October 2023",
        "subscriber_id" : "A-S00696305",
        "delivery_postcode" : "RG1 5EX",
        "delivery_agent_address1" : "Ground Floor 4A Woodley Park Estate",
        "payment_method" : "Credit/Debit Card",
        "last_name" : "Bourke",
        "delivery_agent_county" : "",
        "delivery_agent_postcode" : "RG5 3AW",
        "delivery_address_line_2" : "s",
        "EmailAddress" : "emily.bourke@guardian.co.uk",
        "delivery_country" : "United Kingdom",
        "first_name" : "Emily",
        "ZuoraSubscriberId" : "A-S00696305",
        "package" : "Everyday",
        "delivery_agent_town" : "Reading",
        "delivery_agent_address2" : "Reading Road, Woodley",
        "date_of_first_payment" : "Friday, 13 October 2023",
        "delivery_agent_email" : "brightonstwyford@gmail.com",
        "delivery_agent_telephone" : "01189341470",
        "delivery_address_line_1" : "s",
        "delivery_agent_name" : "Brightons News",
        "delivery_address_town" : "s"
      }
    }
  },
  "DataExtensionName" : "paper-national-delivery",
  "IdentityUserId" : "200085824"
}
```

support-config/src/main/resources/touchpoint.PROD.conf
```diff
@@ -73,5 +73,8 @@ touchpoint.backend.environments {
         campaigns = "MembershipSub-Campaigns-PROD"
       }
     }
+    paper-round-api {
+      url = "https://testapi.guardianhnd.co.uk/v1/guardian" // api key in parameter store
```
rupertbates, line 77: I assume it's correct that these are the same currently?

johnduffell:
> Only the name is needed, but I might as well include the others while I’m here.

I can see the logic, but on the flip side I'm also in favour of minimising the fields sent, as it makes it easier to refactor or update later.  It probably comes down to a judgement call on whether they're likely to be useful in the coming months.

support-workers/src/main/scala/com/gu/support/workers/lambdas/SendThankYouEmail.scala
```diff
@@ -36,12 +37,21 @@ class SendThankYouEmail(
   ): FutureHandlerResult = {
     val emailBuilder = new EmailBuilder(
       services.zuoraService.getMandateIdFromAccountNumber,
+      _ match {
```
rupertbates, line 40: maybe extract this to a named function for readability

support-workers/src/main/scala/com/gu/support/workers/lambdas/SendThankYouEmail.scala
```diff
@@ -36,12 +37,21 @@ class SendThankYouEmail(
...
+      },
       services.promotionService,
       Configuration.stage,
     )
 
     for {
-      emailFields <- emailBuilder.buildEmail(state)
+      emailFields <- emailBuilder.buildEmail(
+        state,
+      )
```
johnduffell, line 54: can we just leave formatting this how it was?

support-config/src/main/resources/touchpoint.PROD.conf
undergroundquizscene: It is: at the moment we’re pointing to the test API for everything because the production API isn’t ready. We have a Trello card in the backlog to switch to their production API before we go live

support-workers/src/main/scala/com/gu/support/workers/lambdas/SendThankYouEmail.scala
```diff
@@ -36,12 +37,21 @@ class SendThankYouEmail(
   ): FutureHandlerResult = {
     val emailBuilder = new EmailBuilder(
       services.zuoraService.getMandateIdFromAccountNumber,
+      _ match {
+        case None => Future.successful(None)
+        case Some(n) =>
+          for {
+            response <- services.paperRoundService.agents()
+          } yield response.data.agents.find(_.refId == n)
```
johnduffell, line 45: what is "n"? the reference about the agent? I'm still not keen on modelling an opaque id as an integer data type as there are too many caveats about overflow/negatives etc.

It might be worth having a trait rather than a function type here, just so it's clearer what's going on (It may have been me that did the original function type in the EmailBuilder, and I now know it's not very readable compared with a trait!)

johnduffell (approved):
looks good, no show stoppers but see my comments above 👍

support-config/src/main/resources/touchpoint.PROD.conf
undergroundquizscene: (Which I will need to update now the config is moving)

support-frontend/conf/routes
```diff
@@ -130,7 +130,7 @@ GET  /subscribe/redeem/validate/:redemptionCode                    controllers.R
 GET  /r/:redemptionCode                                            controllers.RedemptionController.redirect(redemptionCode: String)
 
 GET /postcode-lookup/:postcode                                     controllers.GetAddress.findAddress(postcode: String)
-GET /delivery-agents/:postcode                                     controllers.PaperRound.getAgents(postcode: String)
+GET /delivery-agents/:postcode                                     controllers.PaperRound.getAgents(postcode: String, isTestUser: Option[Boolean])
```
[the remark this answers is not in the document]
undergroundquizscene: Ah, that does sound better – I hadn’t realised that’s how we normally do it. I’ll find an example of that, I assume it’ll work here too.

support-workers/src/main/scala/com/gu/support/workers/lambdas/SendThankYouEmail.scala
undergroundquizscene: Yes, of course – I had originally added a field here, and when I removed it I didn’t notice I’d left the formatting change.

support-frontend/conf/routes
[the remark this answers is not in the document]
undergroundquizscene: Yup, looks good

johnduffell:
I know it's down to personal taste, but once someone has made changes after a PR review, I prefer to just look at the commits since the review so I don't have to look through it all again.  I don't think that's possible if you force push (unless you know a way?) so that is my vote in favour of keeping the commits separate during PR time.
I'm less bothered either way about squashing on merge, as I'm less likely to be interested in the individual commits at that point.

support-workers/src/main/scala/com/gu/support/workers/lambdas/SendThankYouEmail.scala
undergroundquizscene: > Thanks for your detailed reply and all good points.
> 
> I think my concerns are also about clear naming and also type checking, if you have a wrapper type it means you can't mix it up with other unrelated values, also it provides documentation/naming for the code. It's a good point that there's a cost too that has to be worth it.
> 
> For the agent ID it makes more sense to have a proper type as it goes throughout the codebase a lot more.

I’m sorry, I’ve been too stubborn about this: I’ll change it to use a wrapper type throughout. Have you a preference for the underlying type? I asked Chris from PaperRound about the range for IDs and he said a 32-bit integer should be fine, but I’m uncertain whether I’m happy to rely on that. I’m leaning towards scala’s`BigInt` at the moment, and wondering how awkward it would be to use a string instead.

> In the case of the function, because it's passed in one place and used in the target function it's not a big issue. I just took a while to go back and forth to work out what it was, I thought it was related to the DD mandate stuff on the line above at first.

Now that it’s named following Rupert’s suggestion I think it’s clearer, so I’d rather stick with it than change it to a trait, if that’s ok?
 
> I think the cases that really confuse me (in other codebases), functions are passed in several levels and by that time it's very hard to navigate to the possible implementations.

This is an interesting point: when it’s a trait you can search for all classes that implement the trait to find all the implementations in a codebase, whereas when it’s functions it’s hard to search for all implementations other than climbing up through function calls?

undergroundquizscene:
Rebasing and squashing now ahead of merge

PR #5334, opened 2023-10-09
Add alarm to deadletter queue for bigquery-acquisitions-publisher

rupertbates:
#4923 introduced a new SQS queue and lambda which handles writing acquisition events to BigQuery, while it had an alarm for unhandled runtime failures, any handled runtime failures were being returned to the queue via an `SQSBatchResponse`. The idea was that the queue would then pass the failed event to the dead letter queue where it could be investigated however there were two issues with this:
- the event source mapping between the queue and the lambda was not set up to report batch failures so these would have been lost.
- there was no alarm on the dead letter queue so events could be building up on there and we would not know

This PR fixes both of those issues.

As I have said in the description of the alarm, it is important from now on that when events arrive in the dead letter queue they are either sent back to the main queue using redrive functionality (after fixing whatever caused the failure) OR purged from the dead letter queue, otherwise the alarm will remain in an alarm state and will not send the alarm email again in future.

### DIFF
--- a/cdk/lib/bigquery-acquisitions-publisher.ts
+++ b/cdk/lib/bigquery-acquisitions-publisher.ts
@@ -1,10 +1,10 @@
+import { GuAlarm } from "@guardian/cdk/lib/constructs/cloudwatch";
 import type { GuStackProps } from "@guardian/cdk/lib/constructs/core";
 import { GuStack } from "@guardian/cdk/lib/constructs/core";
 import { GuLambdaFunction } from "@guardian/cdk/lib/constructs/lambda";
 import type { App } from "aws-cdk-lib";
 import { Duration } from "aws-cdk-lib";
 import {
-  Alarm,
   ComparisonOperator,
   TreatMissingData,
 } from "aws-cdk-lib/aws-cloudwatch";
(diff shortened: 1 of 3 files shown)


cdk/lib/bigquery-acquisitions-publisher.ts
```diff
@@ -97,16 +103,43 @@ export class BigqueryAcquisitionsPublisher extends GuStack {
...
       events: [eventSource],
       timeout: Duration.minutes(2),
       role,
       errorPercentageMonitoring: monitoring,
     });
+
+    new GuAlarm(this, "DeadLetterQueueAlarm", {
+      app: appName,
+      alarmName: `The ${props.stage} big-query-acquisitions-publisher lambda has failed`,
+      alarmDescription:
```
tomrf1, line 128: very useful 👍

tomrf1 (approved):
👍

cdk/lib/bigquery-acquisitions-publisher.ts
```diff
@@ -97,16 +103,43 @@ export class BigqueryAcquisitionsPublisher extends GuStack {
           }
         : undefined;
 
+    // SQS to Lambda event source mapping
+    const eventSource = new SqsEventSource(queue, {
+      reportBatchItemFailures: true,
```
rupertbates, line 108: This needs to be set or the queue doesn't receive the failure response from the lambda and failed messaes are not sent to the dead letter queue

PR #5335, opened 2023-10-09
Nudge below threshold

iainjchambers-guardian:
Adding a toggle under the the order summary when the contributed amount is below the threshold for subscriber benefits. This is the variant on the a/b test which will replace the currently deployed a/b/c test for one step checkout/two step checkout/two step checkout with nudge.

[**Trello Card**](https://trello.com/c/hvnysW5m/1559-2-step-checkout-nudge-below-threshold-build)

## Is this an AB test?
- [x] Yes
- [ ] No


support-frontend/assets/components/checkoutTopUpToggle/checkoutTopUpToggle.tsx
```diff
@@ -0,0 +1,49 @@
...
+	onChange(): void;
+	contributionType: RegularContributionType;
+	threshold: number;
+}
+
+export function CheckoutTopUpToggle({
+	checked,
+	onChange,
+	contributionType,
+	threshold = 10,
```
GHaberis, line 27: Not sure about defaulting to 10 here, as this is only appropriate for monthly not annual? I'm wondering why we need to define a default for threshold? In what scenario will it be undefined? 
I'd expect we'd only be rendering this component if threshold is defined further up the Component tree.

iainjchambers-guardian: Well spotted, I think that was just added to get it working quickly in storybook. Have removed and renamed the prop to be more descriptive

support-frontend/assets/helpers/abTests/abtestDefinitions.ts
```diff
@@ -141,6 +141,68 @@ export const tests: Tests = {
 			'NE',
 		],
 	},
+	twoStepCheckoutWithNudgeBelow: {
```
GHaberis, line 144: We should delete the `twoStepCheckout` ab test definition above.

GHaberis (approved):
Looks good 👍

support-frontend/assets/pages/supporter-plus-landing/supporterPlusRouter.tsx
```diff
@@ -41,10 +41,12 @@ const {
...
 
 const isInTwoStepTest =
-	abParticipations.twoStepCheckout &&
-	abParticipations.twoStepCheckout !== 'control';
-const showCheckoutTopUpAmounts =
-	abParticipations.twoStepCheckout === 'variant_b';
+	abParticipations.twoStepCheckoutWithNudgeBelow &&
+	abParticipations.twoStepCheckoutWithNudgeBelow !== 'control';
+
+console.log(abParticipations);
```
GHaberis, line 47: @iainjchambers-guardian, I missed this in the review but there's a stray console log here

PR #5336, opened 2023-10-10
Playwright post-deploy tests

rBangay:

This pr adds post deploy test framework using Playwright, browserstack automate and github actions.

The intention of writing these tests in Playwright and Typescript is to replace the existing [Scala Selenium tests](https://github.com/guardian/support-frontend/tree/main/support-frontend/test) with tests that are written in the same language as the site, hopefully encouraging developers to write more tests in the future.

The tests have been setup to run via a github action. The action triggers on `pull_request` with the additional condition of the label `Seen-on-prod` being present (this label is added by prout once a deploy has finished and the changes have been confirmed in production).

They can also be run via the scripts setup in the package.json file, which allow to you to run the tests via the playwright ui against a local running version of the site, or against the production site, or using Browserstack automate against the production site.

Screenshot 2023-10-17 at 10 24 52

More details about running the tests can be found in the README file in this pr.


playwright-tests/README.md
```diff
@@ -0,0 +1,37 @@
...
+Running the Playwright tests with Browserstack requires the following environment variables to be set:
+
+- `BROWSERSTACK_USERNAME`
+- `BROWSERSTACK_ACCESS_KEY`
+
+When the tests are run from the github action these are made available via github secrets. When running the tests locally on your machine you can either set these environment variables globally however you like, or add them to a .env file in the following location:
+
+`support-frontend/playwright-tests/.env`
+
+##### Note: The credentials you provide in the environment variables will need to asociated with a BrowserStack account that has access to [Automate](https://automate.browserstack.com).
```
jamesgorrie, line 30: Tip 🎩  - you can use [this](https://github.com/orgs/community/discussions/16925) for notes

playwright-tests/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+Scripts to run the tests are present in the [package.json](https://github.com/guardian/support-frontend/playwright-tests/package.json) file. Allowing you to also run the tests locally. Three ways of running the tests have been setup:
+
+- `pw-browserstack-test` Runs the tests via BrowserStack Automate against the production site (https://support.theguardian.com/).
+- `pw-local-test` Runs the tests via the Playwright ui against a local running version of the site (https://support.thegulocal.com/).
+- `pw-prod-test` Runs the tests via the Playwright ui against the production site (https://support.theguardian.com/).
+
+It should also be possible to run the tests via BrowserStack Automate against a local running version of the site if you have local testing setup in your BrowserStack account and amend/create a new test configuration file.
+
+### Running the tests using BrowserStack
```
jamesgorrie, line 20: Reading through this, I wonder where the value is of being able to run this locally with quite a complicated DevX? Maybe we could add it separately and maybe add some scripts to set it up?

playwright-tests/README.md
```diff
@@ -0,0 +1,37 @@
...
+When the tests are run from the github action these are made available via github secrets. When running the tests locally on your machine you can either set these environment variables globally however you like, or add them to a .env file in the following location:
+
+`support-frontend/playwright-tests/.env`
+
+##### Note: The credentials you provide in the environment variables will need to asociated with a BrowserStack account that has access to [Automate](https://automate.browserstack.com).
+
+Once the tests have been run (either via the github action or by yourself locally) you can view the results in the [Browserstack Automate dashboard](https://automate.browserstack.com/dashboard). Note again you will have to log in with your Browserstack account that has Automate access.
+
+
+### Possible future work
```
jamesgorrie, line 35: Should we capture this in a follow up issue for planning? I wonder if it'd just get lost here.

playwright-tests/browserstack.config.ts
```diff
@@ -0,0 +1,66 @@
...
+	browser: 'chrome',
+	os: 'osx',
+	os_version: 'catalina',
+	name: 'My first playwright test',
+	build: 'playwright-build',
+	'browserstack.username': process.env.BROWSERSTACK_USERNAME ?? '',
+	'browserstack.accessKey': process.env.BROWSERSTACK_ACCESS_KEY ?? '',
+	'browserstack.local': false,
+	'client.playwrightVersion': clientPlaywrightVersion,
+};
```
jamesgorrie, line 40: I wonder if it's worth being more explicit here about what properties are set, and which are configurable via calling `getCdpEndpoint`? This would also avoid a little mutation which can sometimes be hard to debug.

Keeping the names the same to avoid levels of indirection:

```suggestion
const browserStackProps: BrowserStackProps = {
	build: 'playwright-build',
	'browserstack.username': process.env.BROWSERSTACK_USERNAME ?? '',
	'browserstack.accessKey': process.env.BROWSERSTACK_ACCESS_KEY ?? '',
	'browserstack.local': false,
	'client.playwrightVersion': clientPlaywrightVersion,
};

type BrowserDetails = {
  name: string;
	browser_version: string;
	os: string;
	osVersion?: string;
}
```

and then just join them in `getCdpEndpoint`:
```ts
const cap = {
  ...browserStackProps,
  ...browserDetails
}
const cdpUrl = `wss://cdp.browserstack.com/playwright?caps=${encodeURIComponent(
  JSON.stringify(caps),
)}`;
```

playwright-tests/global-setup.ts
```diff
@@ -0,0 +1,24 @@
...
+		if (err) {
+			console.error(
+				`${redColour}Error starting BrowserStackLocal${whiteColour}`,
+			);
+		} else {
+			console.log('BrowserStackLocal Started');
+		}
+		localResponseReceived = true;
+	});
+	while (!localResponseReceived) {
```
jamesgorrie, line 21: Should we have a `tried 10 times, halting...` short circuit?

playwright-tests/playwright.browserstack.config.ts
```diff
@@ -0,0 +1,30 @@
+import { defineConfig } from '@playwright/test';
+import { getCdpEndpoint } from './browserstack.config';
+import { baseObject } from './playwright.base.config';
+
+export default defineConfig({
+	...baseObject,
+
+	globalSetup: require.resolve('./global-setup.ts'),
+	globalTeardown: require.resolve('./global-teardown.ts'),
```
jamesgorrie, line 9: What's the reason for not just putting these methods in this file? They seem pretty coupled together?

playwright-tests/tests/monthlyContribution.test.ts
```diff
@@ -0,0 +1,62 @@
...
+  const pageUrl = `${baseUrlWithFallback}/uk/contribute`;
+
+  const domain = new URL(pageUrl).hostname;
+  await setTestCookies(context, firstName, domain);
+  await page.goto(pageUrl, {waitUntil: "networkidle"});
+});
+
+test.describe("Sign up for a Recurring Contribution (New Contributions Flow)", () => {
+  test("Monthly contribution sign-up with Stripe - GBP", async ({ page }) => {
+    const monthlyTab = "#MONTHLY";
```
jamesgorrie, line 16: Happy to open the discussion wider as it's more abstract, and could be part of a second PR where we could have this discussion.

Playwright [have a list of preferred locators](https://playwright.dev/docs/locators) - from experience and the literature, it's mainly to do so that you test as closely to what the user sees as possible, allowing us to think about the accessibility as we go and hopefully reduce releasing bugs that the user sees, but we don't as it's obfuscated through invisible IDs.

Testing library cite a great quote:
> [The more your tests resemble the way your software is used, the more confidence they can give you.](https://testing-library.com/docs/guiding-principles)

As per Playwrights docs:
> To make tests resilient, we recommend prioritizing user-facing attributes and explicit contracts such as [page.getByRole()](https://playwright.dev/docs/api/class-page#page-get-by-role).

– [Playwright docs](https://playwright.dev/docs/locators)

What do we think about following a similar pattern?

playwright-tests/README.md
jamesgorrie, line 1: I wonder if we should use the default `e2e` directory for if/when we decide to want to change these files?


> Name of your Tests folder (default is tests or e2e if you already have a tests folder in your project)

-- [Playwright docs](https://playwright.dev/docs/intro#installing-playwright)

Saying that - being a stickler for consistency, should it be `support-e2e` to avoid future folk wondering why this is the special case and doesn't have the `support-` prefix?

jamesgorrie (approved):
Quite a lot of housekeeping style comments, which I would happily take offline and address in separate PRs where we might have more detailed, focussed discussions.

Great set of work, and a fantastic bit of confidence added to help release quicker!

.github/workflows/playwright.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+concurrency:
+  group: 'playwright-${{ github.head_ref }}'
+  cancel-in-progress: true
+
+jobs:
+  test:
+    if: ${{ github.event.label.name == 'Seen-on-PROD' }}
+    name: Playwright
+    timeout-minutes: 30
```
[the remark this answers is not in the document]
rBangay: yeah, shall we leave it as 30 mins here for now, let it run and figure out how much time it needs then make a change

playwright-tests/README.md
```diff
@@ -0,0 +1,37 @@
+## Playwright tests for the [supporter platform](https://support.theguardian.com/).
+
+Playwright tests written in Typescript have been added with the intention of replacing the existing [Scala Selenium tests](https://github.com/guardian/support-frontend/tree/main/support-frontend/test). The rational for this decision was primarily to allow us to write tests in the same language that the site is written in.
+
+As a first pass these tests will run alongside the existing Selenium tests while we migrate the test scenarios over to Playwright. When that work is complete we can remove the Selenium tests entirely.
```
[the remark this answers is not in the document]
rBangay: moved to the bottom of the README file under a 'Rationale' section

playwright-tests/.env.sample
```diff
@@ -0,0 +1,2 @@
+BROWSERSTACK_USERNAME='replace-me'
```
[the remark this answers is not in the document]
rBangay: Ok, we've decided to keep these for now and look into adding SSM integration alongside a script to auto generate the .env file.

playwright-tests/README.md
jamesgorrie: We've spoke about this and it would be good to support all workflows locally and in CI:
- Writing and running integration tests locally against playwright 
- For added confidence you should be able to run locally against browser stack
- Running automate in CI

playwright-tests/playwright.browserstack.config.ts
rBangay: require.resolve here is just returning a string path to the files themselves and not the file contents. This appears to be the expected behaviour of the `defineConfig` function from 'playwright/test'

PR #5337, opened 2023-10-10
Ignore auth cookies if user isn't signed in

kelvin-chappell:
Previously, the `MaybeAuthenticatedAction` was attempting to fetch auth tokens unless the user was signed out.
Now we also check that the user isn't ***signed in*** as well, by checking for the presence of the `GU_U` cookie, which is only dropped for signed-in users.

## Why are you doing this?
The reason for this is that, if a person or bot (Google bot in this instance) is not signed in they go through a redirect path of:

- https://support.theguardian.com/uk/contribute
- https://support.theguardian.com/oauth/authorize
- https://profile.theguardian.com/oauth2/{id}/v1/authorize?state={state}
- https://support.theguardian.com/oauth/callback?state={state}
- https://support.theguardian.com/uk/contribute

The issue is that https://profile.theguardian.com has the `X-Robots-Tag:noindex,nofollow` header set which means that Google defines this as `Not indexable" and we get no search traffic to the page.

Not redirecting non-signed in users should rectify this.

## Is this an AB test?
- [ ] Yes
- [x] No

Thanks to @jamesgorrie for the idea and most of the PR description.

Tested in Code and all seems to work.  Page https://support.code.dev-theguardian.com/uk/contribute is indexable by https://search.google.com/test/rich-results.

### DIFF
--- a/support-frontend/app/actions/UserFromAuthCookiesActionBuilder.scala
+++ b/support-frontend/app/actions/UserFromAuthCookiesActionBuilder.scala
@@ -1,8 +1,13 @@
 package actions
 
 import actions.AsyncAuthenticatedBuilder.OptionalAuthRequest
+import actions.UserFromAuthCookiesActionBuilder.{
+  UserClaims,
+  responseWithNoUser,
+  shouldValidateUser,
+  validateUserRemotely,
+}
 import actions.UserFromAuthCookiesActionBuilder.UserClaims.toUser
-import actions.UserFromAuthCookiesActionBuilder._
 import com.gu.identity.auth._
 import com.gu.identity.model.{PrivateFields, User}
 import config.Identity
@@ -29,7 +34,7 @@ class UserFromAuthCookiesActionBuilder(
     with Logging {
 
   override def invokeBlock[A](request: Request[A], block: OptionalAuthRequest[A] => Future[Result]): Future[Result] =
-    processRequest(config, oktaAuthService)(request, block, _ => processRequestWithoutUser(config)(request, block))
+    shouldValidateUser(config, oktaAuthService)(request, block, _ => responseWithNoUser(config)(request, block))
 }
 
 /** Tries to authenticate the user from ID and access token cookies. If there are no cookies, queries the auth server to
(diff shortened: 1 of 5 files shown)


support-frontend/app/actions/UserFromAuthCookiesActionBuilder.scala
```diff
@@ -104,7 +113,7 @@ object UserFromAuthCookiesActionBuilder {
   /** Tries to process a request with the given block. If the request doesn't have valid token cookies, returns a
     * [[ValidationError]].
     */
-  def tryToProcessRequest[A](
+  private def tryToProcessRequest[A](
```
jamesgorrie, line 116: The naming of the methods here confuse me a little, although the implementation is very explanatory. I wonder if we could be more explicit about what these methods are doing to help with the clarity of the flow for future travellers.

> [!NOTE]  
> I am new to the codebase, so it could be that I am just unfamiliar with the code

Currently we have:
```mermaid
flowchart LR
    invokeBlock-->processRequest{{processRequest}}
    processRequest-->tryToProcessRequest{{tryToProcessRequest}}
    processRequest-->processRequestWithoutUser

    tryToProcessRequest-->AuthenticatedRequest
    tryToProcessRequest-->handFailure
```


Would something like the below be more explanatory and explicit?
```mermaid
flowchart LR
    invokeBlockA-->shouldValidateUser{{shouldValidateUser}}
    shouldValidateUser-- yes -->validateUserLocally
    shouldValidateUser-- no -->responseWithNoUser
    validateUserLocally-- valid -->responseWithUser
    validateUserLocally-- invalid -->validateUserRemotely
    validateUserRemotely-- valid -->responseWithUser
    validateUserRemotely-- invalid -->responseWithNoUser
```

jamesgorrie (approved):
Looks good, looking forward to see what the results are.

Just a little question on naming, but that could be a follow up, or we just leave it.

jamesgorrie (approved):
Looking good!

kelvin-chappell:
Looks successful.  Google search console is no longer complaining about these URLs:
* https://support.theguardian.com/
* https://support.theguardian.com/uk/contribute
* https://support.theguardian.com/us/contribute

They're now in a priority queue for indexing.

GHaberis:
@kelvin-chappell @jamesgorrie amazing, thank you for your work here!

PR #5339, opened 2023-10-12
fix: include @guardian/source packages in babel

jamesgorrie:

Including `@guardian/source` packages in the babel transpilation.

Source [suggest that you use certain babel plugins to have their components work](https://github.com/guardian/csnx/blob/main/libs/%40guardian/source-foundations/src/accessibility/focus-style-manager.ts#L9C14-L9C52)[^1]. This change ensures we're passing the components through those plugins.

## Why are you doing this?
To get support and 💰 from people using < Safari 13.

We should analyse the data post-fix to see try assess what the impact of this was. And maybe a way to test for large scale breaks like this

## Screenshots

Screenshot 2023-10-12 at 12 22 16

Screenshot 2023-10-12 at 12 21 02

We are aware that this doesn't look ideal, but we have tested the functionality and will follow up visual changes once we have assessed the effort required and impact it might be having.

[^1]: I think this components doc is out of date as it works without that plugin.

## Is this an AB test?
- No

### DIFF
--- a/support-frontend/webpack.common.js
+++ b/support-frontend/webpack.common.js
@@ -117,7 +117,7 @@ module.exports = (cssFilename, jsFilename, minimizeCss) => ({
 				test: /\.([jt]sx?|mjs)$/,
 				exclude: {
 					and: [/node_modules/],
-					not: [/@guardian\/consent-management-platform/, /@guardian\/libs/],
+					not: [/@guardian\/consent-management-platform/, /@guardian\/libs/, /@guardian\/source/],
 				},
 				use: [
 					{


rBangay (approved):
👍  It's presumably source / csnx have a different list of supported browsers to us?

jamesgorrie:
> 👍 presumably source / csnx have a different list of supported browsers to us?

@rBangay Nope, as mentioned in the comment, the do say that you should be running certain babel plugins. I'm not sure this is massively effective way of letting people know about it. Maybe their readme needs a "Beware!"

Here's the list that applies to the whole of the Guardian:
https://github.com/guardian/csnx/tree/main/libs/@guardian/browserslist-config#readme

PR #5340, opened 2023-10-12
Update post-deployment tests for new DD form

undergroundquizscene:
Now that the direct debit form has been changed to an expanding form from a two-step form, the post-deployment tests need to be updated to handle the new flow.


jamesgorrie (approved):
Nice

PR #5341, opened 2023-10-12
fix: add pull-request: write permission to compressed-size action

jamesgorrie:
Adds the `pull-request: write` permissions to the compressed-size action.

I had noticed [this error](https://github.com/guardian/support-frontend/actions/runs/6494978751/job/17638995600#step:6:930) as I was hoping to have the report on #5339.

## Why are you doing this?

A clearer understanding of how my changes are affecting the bundle size.

### DIFF
--- a/.github/workflows/compressed-size.yml
+++ b/.github/workflows/compressed-size.yml
@@ -2,6 +2,9 @@ name: Compressed Size
 
 on: [pull_request]
 
+permissions:
+  pull-requests: write
+
 jobs:
   compressed_size:
     if: github.repository_owner == 'guardian'


jamesgorrie:
# Question: 

I _think_ this might append this to all PRs.
This might not be desirable and noisy on PRs that aren't JS bundle focussed.

Are we OK to re-enable this and then mitigate noise if it becomes a problem?

i-hardy (approved):
I'd suggest adding this line as well https://github.com/guardian/support-frontend/blob/ecfec07e4b9168d4fcf22e1e06c0c7093b56150b/.github/workflows/chromatic.yml#L6 - this should at least stop the action trying to run against PRs from the Scala Steward bot, which will never have any effect on bundle size.

PR #5344, opened 2023-10-13
2 step checkout validation fix

iainjchambers-guardian:
Adding a hook to all checkout reducers to reset the validation object back to its default state. 

[**Trello Card**](https://trello.com/c/9nTL2Q17/1588-2-step-checkout-client-side-validation-bug)

## Why are you doing this?
If a user causes validation issues to occur on the second step of the checkout and then navigates back to the first step to change the amount, the validation will prevent them from proceeding to the second step again. This occurs because the validation object is not empty and validation returns false solely on that condition. This PR adds an action to clear that object which is called when the user navigates back to the first step. This will allow validation to pass again on the first step and will be called again when submitting on the second. 

## Is this an AB test?
- [ ] Yes
- [x] No


support-frontend/assets/pages/supporter-plus-landing/twoStepTest/firstStepLanding.tsx
```diff
@@ -114,6 +120,10 @@ export function SupporterPlusInitialLandingPage({
 			: paymentMethodsMarginRecurring}
 	`;
 
+	useEffect(() => {
+		dispatch(resetValidation());
+	}, []);
+
```
rBangay, line 126: note, this will catch users that press the browser back button on the 2nd checkout page instead of the change button. The only downside of having it here in a useEffect hook is that this will fire after the validation error message(s) have had a chance to render on the page as well, so you may get a brief flash of error messages before they go away. This seems fine for now until a more fundamental review of the checkout form validation is complete.

rBangay (approved):

PR #5345, opened 2023-10-13
feat: add fields (postalCode, state, email) to AcquisitionDataRow

AndreaDiotallevi:

I add the optional fields `postalCode`, `state` and `email` to the `AcquisitionDataRow` case class.

These new fields are sent from:
- `payment-api` producer
- `support-workers` producer

[**Trello Card**](https://trello.com/c/c6yBnLcu/366-single-contribution-record-add-fields-to-event-body-send-by-payment-api-and-support-workers-apps)

## Why are you doing this?

The single contribution record, I need to create in Salesforce, requires those fields if they are available.

This is the case class:
```scala
case class SalesforceSingleContributionRecord(
  Amount__c: Double,
  Contribution_ID__c: String,
  Country_Code__c: String,
  Country_Subdivision_Code__c: String, // Not available in the event yet
  Currency__c: String,
  Email__c: String, // Not available in the event yet
  Identity_ID__c: String,
  Payment_Date__c: String,
  Payment_ID__c: String,
  Payment_Provider__c: String,
  Payment_Status__c: String,
  Postal_Code__c: String // Not available in the event yet
)
```

## Risks

Acquisition bus consumers to confirm this change does not break their schema / implementation.

### DIFF
--- a/support-lambdas/acquisitions-firehose-transformer/src/test/scala/com/gu/acquisitionFirehoseTransformer/LambdaSpec.scala
+++ b/support-lambdas/acquisitions-firehose-transformer/src/test/scala/com/gu/acquisitionFirehoseTransformer/LambdaSpec.scala
@@ -88,6 +88,9 @@ class LambdaSpec extends AnyFlatSpec with Matchers {
       paymentId = None,
       queryParameters = Nil,
       platform = None,
+      postalCode = Some("postalCode"),
+      state = Some("state"),
+      email = Some("email")
     )
 
     val record = new KinesisFirehoseEvent.Record()
--- a/support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/models/AcquisitionDataRow.scala
+++ b/support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/models/AcquisitionDataRow.scala
@@ -2,7 +2,6 @@ package com.gu.support.acquisitions.models
 
 import com.gu.i18n.{Country, Currency}
 import com.gu.support.acquisitions.{AbTest, QueryParameter}
-import com.gu.support.catalog.{FulfilmentOptions, ProductOptions}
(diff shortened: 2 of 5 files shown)


support-lambdas/acquisitions-firehose-transformer/src/test/scala/com/gu/acquisitionFirehoseTransformer/LambdaSpec.scala
```diff
@@ -88,6 +88,9 @@ class LambdaSpec extends AnyFlatSpec with Matchers {
       paymentId = None,
       queryParameters = Nil,
       platform = None,
+      postalCode = Some("postalCode"),
+      countrySubdivisionCode = Some("countrySubdivisionCode"),
```
rupertbates, line 92: I would rather call this `state` as that's what it's known as in most of the rest of this codebase

rupertbates: Also this class doesn't need to know about the new fields as it doesn't use them, so let's not add them in.

rupertbates (approved):
👍

support-modules/acquisition-events/src/main/scala/com/gu/support/acquisitions/models/AcquisitionDataRow.scala
```diff
@@ -42,6 +41,9 @@ case class AcquisitionDataRow(
     paymentId: Option[String],
     queryParameters: List[QueryParameter],
     platform: Option[String],
+    postalCode: Option[String],
+    countrySubdivisionCode: Option[String],
```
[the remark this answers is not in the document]
AndreaDiotallevi: Thanks Rupert - I have amended.

PR #5346, opened 2023-10-16
Set nationalDelivery AB test live at 100%

rupertbates:
Make the National Delivery functionality available to everybody via a 100% AB test with only one variant.


shtukas (approved):

andrewHEguardian (approved):

johnduffell:
the card in trello says to remove the AB test, but this PR just sets it to 100%.  I presume it's the wording on the card that's inaccurate and we didn't plan to remove the test completely yet.

PR #5353, opened 2023-10-18
Add an AV calculation for the National Delivery product

rupertbates:
As part of the work to create the new National Delivery product we need to provide an Annualised Value (AV) calculation for each configuration of the product (every day, six day and weekend).

This AV calculation is only used by the kinesis data stream which drives tickers and the Epic super mode, the calculation which is available in the data lake is specified in the [data-platform-models repo](https://github.com/guardian/data-platform-models/blob/498533bfe08ac005f95b961d87817130fb922c50/dbt/macros/annualised_value/annualised_value.sql#L4)


andrewHEguardian (approved):
👍

PR #5359, opened 2023-10-19
Read the defaultContributionType and displayContributionType from RRCP Amounts config in Support Checkout

LAKSHMIRPILLAI:

Support checkout: Read the `defaultContributionType` and `displayContributionType` from RRCP Amounts config

[**Trello Card**](https://trello.com/c/EY0v3UWy/1565-default-to-annual-remove-monthly-support-checkout-read-the-defaultcontributiontype-and-displaycontributiontype-from-rrcp-amounts)

## Why are you doing this?

Currently the payment frequency tabs and the default tab on the Support Checkout are rendered based on the values in  `common.settings.contributionTypes` from Redux state.

However, the amounts cards rendered on the support checkout for each tab are defined in amounts test config set in RRCP, this config is written to S3 as JSON, and then read by support-frontend and made available in the client on `window.guardian.settings.amounts`.

`window.guardian.settings.amounts` is an array of amounts tests. Each test can have one or more variants. Each variant has a `defaultContributionType` and `displayContributionType` property. Both of these properties can be configured in the RRCP:

- `defaultContributionType` defines the default tab a user assigned to the test should see, this can be "ONE_OFF" | "MONTHLY" | "ANNUAL"
- `displayContributionType` defines what tabs a user should see in an array of Contribution Types

In https://github.com/guardian/support-frontend/pull/5027 logic was added to select the correct amounts test/variant to render and add it to `common.settings.amounts` in Redux state.

To date we’ve not been reading `defaultContributionType` or `displayContributionType` from `common.settings.amounts` in Redux state.

However with this Pull Request we achieve our goal of doing so in the fastest and most straightforward manner.

**Notes:**

While  testing locally, to make changes to the defaultContributionType and displayContributionType  you can update the  Amazon S3 / Buckets /support-admin-console / DEV / configured-amounts-v3.json file .

Users can click on choice cards in the Epic/Banner. The choices they make are passed as query string params selected-contribution-type ("ONE_OFF", "MONTHLY", "ANNUAL") and selected-amount. If present in the URL selected-contribution-type should override the  defaultContributionType.

## Navigating through the PR

This PR touches 7 files, as follows:

**support-frontend/assets/components/paymentFrequencyTabs/paymentFrequencyTabsContainer.tsx**
Changes to display tabs in Supporter checkout based on the `displayContributionType`  from `common.settings.amounts` in Redux state.

**support-frontend/assets/helpers/redux/commonState/selectors.ts**
Changes to get the default tab selected from `defaultContributionType` from `common.settings.amounts` in Redux state.

**support-frontend/assets/pages/supporter-plus-landing/setup/setUpRedux.ts**
Necessary changes to look for initial contribution type and amount if passed as query string params `selected-contribution-type` ("ONE_OFF", "MONTHLY", "ANNUAL") and `selected-amount`, or set in session storage. 

If a selected contribution type is present as a query string param or in session storage then we cross check the value is valid by confirming it exists in `common.settings.amounts.displayContributionType`. If it's invalid we fallback to the value passed from `common.settings.amounts.defaultContributionType` in Redux state.

We've also removed the check for contribution Types from campaign settings  as this is no longer used. 

**support-frontend/assets/helpers/redux/commonState/reducer.ts.**  and 
**support-frontend/assets/helpers/redux/commonState/state.ts** and 
**support-frontend/assets/helpers/redux/utils/setup.ts**
Removing common.defaultAmounts from reducer which we concluded is not used anywhere in the code

**support-frontend/assets/pages/supporter-plus-landing/supporterPlusRouter.tsx**
Remove an unnecessary console.log statement

## Tasks for later

> The ideal solution for this is to remove the contributionTypes from `window.guardian` and Redux state, this will require client and server side changes. This is also complicated as we have Kindle checkout which references the contributionType and shares similar code as the supporter checkout. So our next step would be to remove the contributionTypes entirely after removing the Kindle (Digital-sub) code in a separate PR.

> Also we found a lot of Code related to Campaign settings that is not used anymore while working on this. We need to remove them as well in a separate PR.

## Is this an AB test?
- [ ] Yes
- [x ] No

### DIFF
--- a/support-frontend/assets/pages/supporter-plus-landing/setup/setUpRedux.ts
+++ b/support-frontend/assets/pages/supporter-plus-landing/setup/setUpRedux.ts
@@ -134,19 +134,6 @@ function selectInitialAmounts(
 	}
 }
 
-// Override the settings from the server if contributionTypes are defined in url params or campaign settings
-// function getContributionTypes(state: ContributionsState): ContributionTypes {
-// 	// const campaignSettings = getCampaignSettings();
-//   //
-// 	// if (campaignSettings?.contributionTypes) {
-// 	// 	return campaignSettings.contributionTypes;
-// 	// }
-//   //
-// 	// return getValidContributionTypesFromUrlOrElse(
-// 	// 	state.common.settings.contributionTypes,
-// 	// );
-// }
-
 function selectInitialContributionTypeAndPaymentMethod(
 	state: ContributionsState,
 	dispatch: ContributionsDispatch,
@@ -171,9 +158,6 @@ function selectInitialContributionTypeAndPaymentMethod(
 export function setUpRedux(store: ContributionsStore): void {
 	const dispatch = store.dispatch;
 	const state = store.getState();
-	// TODO - move these settings out of the redux store, as they only change once, upon initialisation
-	// const contributionTypes = getContributionTypes(state);
-	// dispatch(setContributionTypes(contributionTypes));
 
 	setUpUserState(dispatch);
 	void dispatch(getRecurringContributorStatus());
@@ -187,7 +171,6 @@ export function setUpRedux(store: ContributionsStore): void {
 	const contributionType = selectInitialContributionTypeAndPaymentMethod(
 		state,
 		dispatch,
-		// contributionTypes,
 	);
 	selectInitialAmounts(state, dispatch, contributionType);
 
--- a/support-frontend/assets/pages/supporter-plus-landing/setup/setUpRedux.ts
+++ b/support-frontend/assets/pages/supporter-plus-landing/setup/setUpRedux.ts
@@ -1,15 +1,10 @@
 // ----- Imports ----- //
-import { getCampaignSettings } from 'helpers/campaigns/campaigns';
-import type {
-	ContributionType,
-	ContributionTypes,
-} from 'helpers/contributions';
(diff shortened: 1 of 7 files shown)


support-frontend/assets/pages/supporter-plus-landing/setup/setUpRedux.ts
```diff
@@ -65,28 +59,20 @@ function getInitialPaymentMethod(
...
 		getContributionTypeFromUrl() ?? getContributionTypeFromSession();
 
-	// make sure we don't select a contribution type which isn't on the page
-	if (
-		contributionType &&
-		contributionTypes[countryGroupId].find(
-			(ct) => ct.contributionType === contributionType,
-		)
-	) {
+	// // make sure we don't select a contribution type which isn't on the page
```
jamesgorrie, line 70:
```suggestion
	// make sure we don't select a contribution type which isn't on the page
```

jamesgorrie (approved):
This looks like a really great simplification of the code!

support-frontend/assets/pages/supporter-plus-landing/setup/setUpRedux.ts
LAKSHMIRPILLAI: Thank you for reviewing this! 👍🏼

PR #5361, opened 2023-10-20
chore(deps): Upgrade @guardian/source-* packages and their linting friends

jamesgorrie:
Upgrading our [source](https://github.com/guardian/csnx/tree/main/libs/%40guardian) dependencies to newer versions for:
- security
- be able to use newer versions if we contribute towards them e.g. with the ticker

They are quite heavily coupled to the eslint rules and plugins, as those packages use them too, so there were a few conflicts the needed sorting.

We also had a typescript upgrade, which meant we need to be tighter with out typing (which is better)
- convert `void` to `Promise<void>` where necessary
- use null conditional checking
- Use a union of `string | PaymentMethod` for when that's the type we receive from Stripe responses
- There is some cyclical dependency stuff we can look at later as it's actually quite complicated and shouldn't cause things to break (but does make code potentially brittle an hard to work through)


jamesgorrie:
Closing in favour of:
https://github.com/guardian/support-frontend/pull/5367

PR #5366, opened 2023-10-24
Set up new US campaign ticker

i-hardy:

This adds a new and updated ticker component for the 2023 EOY US campaign and adds it to the checkout, behind the existing campaign switch.

[**Trello Card**](https://trello.com/c/XgWlzhwj)

## Why are you doing this?

There is a desire to revive the old campaign tickers used in previous years to track contributions during specific periods. The existing ticker component uses some difficult-to-maintain SCSS styles, is tightly coupled to the data fetching mechanism, is an older style class component, and is not especially screen reader accessible; the new component addresses these issues.

### DIFF
--- a/support-frontend/assets/components/ticker/tickerContainer.tsx
+++ b/support-frontend/assets/components/ticker/tickerContainer.tsx
@@ -1,12 +1,7 @@
 import { useEffect, useState } from 'react';
 import { fetchJson } from 'helpers/async/fetch';
 import type { TickerProps } from './ticker';
-import type {
-	TickerConfigData,
-	TickerCopy,
-	TickerCountType,
-	TickerEndType,
-} from './types';
+import type { TickerConfigData, TickerCountType, TickerEndType } from './types';
 
 function getInitialTickerValues(
 	tickerCountType: TickerCountType,
@@ -25,14 +20,14 @@ function getInitialTickerValues(
 }
 
 function getDefaultTickerEnd(_total: number, goal: number) {
-	return goal + goal * 0.15;
+	return goal;
 }
 
 type TickerContainerProps = {
 	countType: TickerCountType;
 	endType: TickerEndType;
 	currencySymbol: string;
-	copy: TickerCopy;
+	headline: string;
 	calculateEnd?: (total: number, goal: number) => number;
 	render: (props: TickerProps) => JSX.Element;
 };
@@ -42,7 +37,7 @@ export function TickerContainer({
 	countType,
 	endType,
 	currencySymbol,
-	copy,
+	headline,
 	calculateEnd = getDefaultTickerEnd,
 }: TickerContainerProps): JSX.Element {
 	const [tickerConfig, setTickerConfig] = useState<TickerConfigData>({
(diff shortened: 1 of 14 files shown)


support-frontend/assets/components/ticker/tickerContainer.tsx
```diff
@@ -0,0 +1,68 @@
...
+	TickerCopy,
+	TickerCountType,
+	TickerEndType,
+} from './types';
+
+function getInitialTickerValues(
+	tickerCountType: TickerCountType,
+): Promise<TickerConfigData> {
+	return fetchJson<TickerConfigData>(
+		tickerCountType === 'money' ? '/ticker.json' : '/supporters-ticker.json',
```
GHaberis, line 15: @i-hardy the data for a ticker can be retrieved from the support site route `/ticker/${tickerId}.json`. The `tickerId` must match the corresponding key from the ticker config from S3. This config is defined as JSON in S3 in the membership-private bucket, under `/{STAGE}/ticker.conf.json`. 

The `tickerId` for the US campaign should be "US". So the route to retrieve the data for the US will be `/ticker/US.json`. This route works on PROD and CODE, and will return the appropriate data for each environment. Each environments data is written to S3  in the contributions-ticker bucket, under `/{STAGE}/{TickerID}.json`. 

The routes above can be used whether the ticker is reporting on a money target OR a supporter target. 

For DEV if working on a ticker with a money target you can use the dummy route `/ticker.json`. If it's a ticker with a supporter target you can use the dummy route `/supporters-ticker.json`, both these routes contain dummy data that can be manipulated during dev.

This repo contains all this info: https://github.com/guardian/contributions-ticker-calculator

support-frontend/assets/pages/supporter-plus-landing/formSections/amountAndBenefits.tsx
```diff
@@ -27,6 +33,14 @@ export function AmountAndBenefits({
 	addBackgroundToBenefitsList?: boolean;
 	isCompactBenefitsList?: boolean;
 }): JSX.Element {
+	const { internationalisation } = useContributionsSelector(
+		(state) => state.common,
+	);
+	const campaignSettings = getCampaignSettings('Us_eoy_2023');
+
+	const showUSCampaignTicker =
```
jamesgorrie, line 41: I've probably missed it - but where's the switch to make sure this isn't enabled yet?

jamesgorrie (approved):
Really clear PR and code!

support-frontend/assets/pages/supporter-plus-landing/formSections/amountAndBenefits.tsx
i-hardy: It's `window.guardian.enableContributionsCampaign` or `window.guardian.settings.switches.campaignSwitches.enableContributionsCampaign`. It's driven from the RRCP the same as our other switches for things like payment methods.

Currently we only have an on/off switch for 'enable any campaign' as I think it's only ever been used for one thing at a time in the past. Improving this in the future is something George and I were discussing earlier today.

PR #5368, opened 2023-10-24
Upgrade @guardian/source-* and @guardian/libs

jamesgorrie:
This upgrades `@guardian/source-*` and `@guardian/libs`.

This fixes some security vulnerabilities.

It should also help when / if we want to push updates up to the `csnx` library e.g. the ticker so that we can use them across different repos e.g. `support-frontend`, `csti-interactives` and `support-dotcom-components`.

I've tried to keep the changes to types and not implementation. 

There is also the question of the cyclical dependencies, but given those will be larger, more complicated changes, we should keep them out of scope of this PR.

```[tasklist]
### Tasks
- [ ] Run design past spacing changes from the new source lib
```


i-hardy (approved):
I've already raised a health card for the `import/no-cycle` issues as this also came up via a Dependabot PR.

PR #5369, opened 2023-10-24
Change Single to One-Time on checkout tabs (support frontend)

LAKSHMIRPILLAI:

We need to change the wording for One-off contributions from Single to "One-time" in the checkout tab.

[**Trello Card**](https://trello.com/c/0ABxipb3/1619-changing-single-to-one-time-on-checkout-tabs-support-frontend)

## Why are you doing this?

Following UXR, We would like to test changing the word “single” on the checkout to “One-time”.

This would not be a test, this would be global.

## Is this an AB test?
- [ ] Yes
- [x ] No

##  Before Change
image

Screenshot 2023-10-24 at 17 06 20

## After Change 

image

Screenshot 2023-10-24 at 17 07 45

### DIFF
--- a/support-frontend/assets/components/orderSummary/contributionsOrderSummary.tsx
+++ b/support-frontend/assets/components/orderSummary/contributionsOrderSummary.tsx
@@ -176,7 +176,7 @@ export type ContributionsOrderSummaryProps = {
 };
 
 const supportTypes = {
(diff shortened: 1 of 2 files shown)


jamesgorrie (approved):
Great! Can'\t but help think of the Fugees song 🎶

PR #5373, opened 2023-10-26
E2E tests: 2 step checkout

jamesgorrie:
- forces the e2e tests to be running on the `ab-twoStepCheckoutWithNudgeBelow=variant_a`
- refactors a few of the [locators](https://playwright.dev/docs/locators#locate-by-text) to follow [Playwright's advice of "prioritizing user-facing attributes and explicit contracts such as page.getByRole()"](https://playwright.dev/docs/locators#locating-elements)

[**Trello Card**](https://trello.com/c/1F4DgCyB/1621-put-2-step-variant-live-to-100-of-audience)

## Why are you doing this?
This is because we're swapping to the 2 step checkout as default, but will allow us to work on the tests in parallel while doing so.

### DIFF
--- a/support-e2e/tests/monthlyContribution.test.ts
+++ b/support-e2e/tests/monthlyContribution.test.ts
@@ -15,7 +15,7 @@ test.beforeEach(async ({ page, context, baseURL }) => {
 test.describe("Sign up for a Recurring Contribution (New Contributions Flow)", () => {
   test("Monthly contribution sign-up with Stripe - GBP", async ({ page }) => {
     await page.getByRole("tab").getByText("Monthly").click();
-    await page.getByText("Continue to checkout").click();
+    await page.getByRole("button", { name: "Continue to checkout" }).click();
 
     await page.getByLabel("Email address").type(email);
     await page.getByLabel("First name").type(firstName);
--- a/support-e2e/tests/monthlyContribution.test.ts
+++ b/support-e2e/tests/monthlyContribution.test.ts
@@ -4,11 +4,11 @@ import { email, firstName, lastName } from "./utils/users";
 
 test.beforeEach(async ({ page, context, baseURL }) => {
   const baseUrlWithFallback = baseURL ?? "https://support.theguardian.com";
-  const pageUrl = `${baseUrlWithFallback}/uk/contribute`;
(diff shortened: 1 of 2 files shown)


support-e2e/tests/monthlyContribution.test.ts
```diff
@@ -4,58 +4,58 @@ import { email, firstName, lastName } from "./utils/users";
...
-  await page.goto(pageUrl, {waitUntil: "networkidle"});
+  await page.goto(pageUrl, { waitUntil: "networkidle" });
 });
 
 test.describe("Sign up for a Recurring Contribution (New Contributions Flow)", () => {
   test("Monthly contribution sign-up with Stripe - GBP", async ({ page }) => {
-    const monthlyTab = "#MONTHLY";
-    await page.locator(monthlyTab).click();
+    await page.getByRole("tab").getByText("Monthly").click();
+    await page.getByText("Continue to checkout").click();
```
jamesgorrie, line 18: This is the main difference between 1 step and 2 step checkout.

LAKSHMIRPILLAI, line 18: Why is this a page.getByText instead of  await page.getByRole('button', { name: 'Continue to checkout' }).click() ? Is it just to reduce the code?

jamesgorrie: Nice suggestion, I wasn't aware of that syntax and have updated it.

support-e2e/tests/monthlyContribution.test.ts
```diff
@@ -4,58 +4,58 @@ import { email, firstName, lastName } from "./utils/users";
...
 
-    await page.locator("#email").type(email);
-    await page.locator("#firstName").type(firstName);
-    await page.locator("#lastName").type(lastName);
+    await page.getByLabel("Email address").type(email);
+    await page.getByLabel("First name").type(firstName);
+    await page.getByLabel("Last name").type(lastName);
 
-    await page.locator("#qa-credit-card").click();
+    await page.getByLabel("Credit/Debit card").click();
```
LAKSHMIRPILLAI, line 24: And I think , we can use  await page.getByRole('radio', { name: 'Credit/Debit card' }).click(); here as it's a radio button, but this is also fine

LAKSHMIRPILLAI (approved):
👍🏼 LGTM

PR #5375, opened 2023-10-26
Make 2 step checkout (`variant_b`) the default for people

jamesgorrie:
Makes the `twoStepCheckoutWithNudgeBelow; variant_b` the default for 100% of our readers and supporters.

| Before        | After      |
| -----------   | ---------- |
| ![before-1][] | ![after-1][] |
|               | ![after-2][] |

[before-1]: https://github.com/guardian/support-frontend/assets/31692/5ce7f387-b5d5-4f8d-baeb-346c395a0f90

[after-1]: https://github.com/guardian/support-frontend/assets/31692/06af8290-e5d0-4a36-b374-80418a2d24db

[after-2]: https://github.com/guardian/support-frontend/assets/31692/5f23022a-6529-4ba5-aa7d-5bb74bf2b12e

## Follow ups
- In `variant_b` we had the `CheckoutTopUpToggle` which prompted people to top up to a larger amount, which we're leaving out, but keeping in `storybook` for now. We might want to think about how we tag this Component as unused but to keep. Maybe somewhere in storybook?
- There is another component `CheckoutTopUpAmounts` which I am less certain about what it does, but it looks like it was usurped by `CheckoutTopUpToggle` [in this PR](https://github.com/guardian/support-frontend/pull/5335). Assuming we should remove this?


support-frontend/assets/components/orderSummary/contributionsOrderSummaryContainer.tsx
```diff
@@ -64,12 +70,45 @@ export function ContributionsOrderSummaryContainer({
...
+		getUserSelectedAmountBeforeAmendment,
+	);
+	const isWithinThreshold =
+		contributionType !== 'ONE_OFF' &&
+		amountBeforeAmendments >=
+			benefitsThresholdsByCountryGroup[countryGroupId][contributionType] &&
+		amountBeforeAmendments <=
+			checkoutTopUpUpperThresholdsByCountryGroup[countryGroupId][
+				contributionType
+			];
```
i-hardy, line 97: This is extremely hard to read- any chance of refactoring this into separate named conditions?

support-frontend/assets/components/orderSummary/contributionsOrderSummaryContainer.tsx
```diff
@@ -64,12 +70,45 @@ export function ContributionsOrderSummaryContainer({
...
 		tsAndCs: getTermsConditions(contributionType, isSupporterPlus),
+		handleAmountUpdate,
+		isWithinThreshold,
 	});
 }
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars -- we should be able to use underscore variable names but that still breaks
+function dispatch(_: unknown) {
+	throw new Error('Function not implemented.');
+}
```
i-hardy, line 114: This looks like it was maybe auto-generated by VSCode?

support-frontend/stories/checkouts/CheckoutTopUpAmounts.stories.tsx
```diff
@@ -23,39 +23,36 @@ export default {
...
 export const UKMonthlyTopUpAmounts = Template.bind({});
 
 UKMonthlyTopUpAmounts.args = {
+	contributionType: 'MONTHLY',
 	currencySymbol: '£',
 	timePeriod: 'month',
 	amounts: [1, 2, 3, 4, 5],
 	isWithinThreshold: true,
-};
+} as CheckoutTopUpAmountsProps;
```
i-hardy, line 43: Would be good to avoid these type assertions if possible- what are the issues that are causing their use?

support-frontend/assets/components/orderSummary/contributionsOrderSummaryContainer.tsx
jamesgorrie: Yea - definitely a copy/pasta, I will have a look into this, but a little wary of changing implementation all that much.

support-frontend/stories/checkouts/CheckoutTopUpAmounts.stories.tsx
jamesgorrie: The issue I was having is that when setting `ComponentNameThatInheritsTemplate.args` they weren't being typed, so they were a free for all.

What's the benefit of creating the `Template` abstraction vs 
```tsx
const UKMonthlyTopUpAmounts = () => {
  return <CheckoutTopUpAmounts contributionType={"MONTHLY"} /* ... */ />
}

const UKYearlyTopUpAmounts = () => {
  return <CheckoutTopUpAmounts contributionType={"YEARLY"} /* ... */ />
}
```
?

jamesgorrie:
@i-hardy - I've actually removed the work to allow for these containers as we can actually remove the containers from the component all-together.

I hadn't realised `showTopUpAmounts` is not actually being passed down as [it was removed in this PR](https://github.com/guardian/support-frontend/commit/42c15b07621b0df7390f07bd463ddba3ab319e95#diff-2ff3d03ae505d8f9b657e5faa00762e20c2522a4a030e38582f5547869f055c0L51) so `CheckoutTopUpAmounts` would never be rendered. 

Same for `CheckoutTopUpToggle` as `showTopUpToggle` would never be true as it's [derived from `variant_b`](https://github.com/guardian/support-frontend/pull/5375/files#diff-fd63b7bdddd1f8c8a686a6ba94bee4323ebd6d88ea919d35cdcf7dbecd9a6db5L47-L49).

support-frontend/stories/checkouts/CheckoutTopUpAmounts.stories.tsx
i-hardy: The Template abstraction is what's demonstrated in the Storybook docs and as far as I'm aware is part of how its built-in documentation and controls features work- as far as my understanding goes your suggestion would destroy the automatic detection of prop types and require us to add this stuff manually for every story.

Asserting the type of the `args` property on `Template` should propagate that type to every story that derives from `Template`- this seems to be working fine when looking at other stories. If `IndividualStory.args` isn't correctly typed then I'd imagine something else is wrong somewhere and using a type assertion is simply hiding the real issue.

i-hardy:
If there's a way to enforce a design pattern like this via types then I'd love to know about it, but I'm not aware of anything. The pattern is both documented and explained in the support-frontend wiki but if there's any improvements we can make to this that would be great.

I would consider this to be something of an edge case as in most other scenarios the problem would have been made immediately apparent when trying to write the component stories- it was just that in this case the original author of the component didn't attempt to create any stories that would trigger the Redux-reliant component to render. Again, if there's better ways to enforce 'you should always exhaustively document your component in its stories' than via PR we should implement that, but I'm not sure that any such thing exists.

jamesgorrie:
@i-hardy completely agree. I suppose we hit the edge case of a few patterns colliding + the combo of the tests filtering out the offending code path.

Nice problem to think about though, thanks again for all the pointers and info!

i-hardy (approved):
I'd like to see that type assertion issue sorted out, but I don't think it should be a blocker to getting this merged and everything else looks great 👍

support-frontend/stories/checkouts/CheckoutTopUpAmounts.stories.tsx
jamesgorrie: This all seems to work as you've described, so perhaps something else got mixed up in my IDE/head.

Dropping [this link](https://storybook.js.org/tutorials/intro-to-storybook/ember/en/simple-component/) here for any future folk who fancy reading about it. Thanks for the info!

jamesgorrie:
> I'd like to see that type assertion issue sorted out

Sorted, thanks for walking me through it.

PR #5376, opened 2023-10-26
Delete redundant bigquery config

tomrf1:
This deletes the `BigQuerySpec` integration test in support-workers, which isn't particularly useful.
This was the last thing using the old method of BigQuery authentication that used secrets. We now use [identity federation](https://github.com/guardian/support-frontend/pull/4923/files#diff-1a2b4b4f84c36c9c7c27027728c73863f09a21516f27b806d556f71e9c113e9dR94).


johnduffell (approved):
👍 so this PR is deleting the code to read the config, presumably it's a manual step in AWS console to delete the actual config keys?

tomrf1:
> 👍 so this PR is deleting the code to read the config, presumably it's a manual step in AWS console to delete the actual config keys?

that's right, we can delete some Parameter Store config later

PR #5380, opened 2023-10-27
Temporary disabled AB test to test canMakePaymentsWithActiveCard

GHaberis:
## What are you doing in this PR?

Proposal: Release a disabled AB test canMakePaymentsWithActiveCard to PROD. Force myself in to the AB test to see whether the `window.ApplePaySession.canMakePaymentsWithActiveCard` resolves with `true` on `support.theguardian.com`. I've tested on CODE and it resolved as `false`. Once my test is complete I'll revert this change.

PR #5391, opened 2023-10-31
Revert "Temporary disabled AB test to test canMakePaymentsWithActiveCard"

GHaberis:
Reverts guardian/support-frontend#5380

I've run the test and concluded `support.theguardian.com` is not set-up to use `canMakePaymentsWithActiveCard`.

Tested on `PROD` using an iPhone with Apple Pay enabled, `canMakePaymentsWithActiveCard` resolved with `false`.

### DIFF
--- a/support-frontend/assets/helpers/abTests/abtestDefinitions.ts
+++ b/support-frontend/assets/helpers/abTests/abtestDefinitions.ts
@@ -94,15 +94,4 @@ export const tests: Tests = {
 		referrerControlled: false,
 		seed: 0,
 	},
-	canMakePaymentsWithActiveCard: {
-		variants: [
-			{
-				id: 'control',
-			},
-		],
-		isActive: false,
-		audiences: {},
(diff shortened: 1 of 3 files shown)


paul-daniel-dempsey (approved):
Does that mean Stripe entirely handles ApplePay detection itself? If so, thats a shame we can't be re-assured it will work. Would guess will still test in code once approval comes through?

GHaberis:
> Does that mean Stripe entirely handles ApplePay detection itself? If so, thats a shame we can't be re-assured it will work. Would guess will still test in code once approval comes through?

Stripe handles everything on the support site, it was a bit of a hopeful shot in the dark seeing if this worked - if it did we may have had more options to implementing the Epic AB test you've been looking at, at least we now know it doesn't. We can't use Stripe on the main Guardian domain though, which is why we should continue pursuing our original plan for the Epic AB test.

PR #5393, opened 2023-10-31
Round selected amount in benefits container to 2 decimal places

iainjchambers-guardian:
Rounding the amount in the benefits container to two fixed decimal places

## Why are you doing this?
Currently, if the user enters an invalid amount (i.e. with more than 2 decimal places) in the *choose amount* field, it will be shown in the benefits list below. 

## Is this an AB test?
- [ ] Yes
- [x] No

## Screenshots
| Before | After |
| --- | --- |
| image | image |


rBangay (approved):
While I'm sure this is fine, it is worth noting that toFixed will round the decimal places.

for example:
Screenshot 2023-11-02 at 23 28 49

I assume this isn't an issue

PR #5403, opened 2023-11-03
Identity retries

rupertbates:

## Is this an AB test?
- [ ] Yes
- [ ] No


LAKSHMIRPILLAI (approved):

support-modules/retry/build.sbt
```diff
@@ -0,0 +1,5 @@
+import LibraryVersions.circeVersion
+
```
jamesgorrie, line 2: For this to be picked up in the main build, don't you need to add the module name [like other modules](https://github.com/guardian/support-frontend/blob/main/support-modules/aws/build.sbt#L3-L5)?

```suggestion

name := "module-retry"

description := "Scala library to retry asynchronous methods if they fail."

```

jamesgorrie (commented):
Main question is around whether we need to add the module name to the `build.sbt`?

build.sbt
```diff
@@ -136,8 +136,15 @@ lazy val `support-frontend` = (project in file("support-frontend"))
     `support-models`,
     `support-config`,
     `support-internationalisation`,
+    `module-retry`,
+  )
+  .aggregate(
+    `support-services`,
+    `support-models`,
+    `support-config`,
+    `support-internationalisation`,
```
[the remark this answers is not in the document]
rupertbates: Yes exactly

support-modules/retry/build.sbt
jamesgorrie: Thanks for the detail!

PR #5404, opened 2023-11-03
Remove sortCodeArray and rename sortCodeString to sortCode

iainjchambers-guardian:
Removing the `sortCodeArray` property from the redux state and renaming `sortCodeString` to `sortCode`

[**Trello Card**](https://trello.com/c/nyO53aVo/1595-remove-sort-code-array-direct-debit-redux-code)

## Why are you doing this?
There was previously a need to store the sort code for direct debit transactions as an array but now that there isn't, it's best to remove it to tidy up the redux store. As it is now gone, there's no need for the extra delineation in the name of `sortCodeString` so I have renamed it to `sortCode` so as not to infer that there is another variable of a similar purpose but different type as that's no longer the case.

## Is this an AB test?
- [ ] Yes
- [x] No


jamesgorrie (approved):
A strange false positive I've seen on a few PRs now with Chromatic.

Otherwise - nice!

PR #5405, opened 2023-11-03
feat: use fastly image service for grid images

jamesgorrie:

# Same
https://i.guim.co.uk/img/media/81d6bda5f74a84b952e7162d553ae71499c0c7ed/0_9_540_324/master/500.png?dpr=1&s=none&width=500

https://media.guim.co.uk/81d6bda5f74a84b952e7162d553ae71499c0c7ed/0_9_540_324/500.png

# Quite different
https://i.guim.co.uk/img/media/94f9164e914d4783a60ac44b822872827e6e29d0/0_0_420_165/master/420.jpg?dpr=1&s=none&width=420

https://media.guim.co.uk/img/media/94f9164e914d4783a60ac44b822872827e6e29d0/0_0_420_165/420.jpg

https://gist.github.com/jamesgorrie/3387584985adb22de556ca4782d6503d

## Is this an AB test?
- [ ] Yes
- [ ] No


support-frontend/assets/helpers/__tests__/theGridTest.ts
```diff
@@ -1,23 +1,18 @@
...
-} from '../images/theGrid';
+import { GRID_DOMAIN, gridSrcset, gridUrl } from '../images/theGrid';
 // ----- Tests ----- //
 describe('theGrid', () => {
 	it('should construct a grid url', () => {
-		const gridId = 'guardianObserverOffice';
+		const gridId = 'weeklyLandingHero';
 		const size = 333;
-		const expectedUrl = `${GRID_DOMAIN}/${imageCatalogue[gridId]}/333.jpg`;
+		const expectedUrl = `${GRID_DOMAIN}/img/media/0e5187e523a6445003fae7b1bb4124a8a4a814c8/0_655_6362_2706/master/333.jpg?dpr=1&s=none&width=333`;
```
i-hardy, line 8: Is there a need to hardcode the ID here, or could we continue to use ``${imageCatalogue[gridId]``?

i-hardy (approved):
One small question but otherwise this looks good!

support-frontend/assets/helpers/__tests__/theGridTest.ts
jamesgorrie: Just habit of writing tests to not couple them to any implementation. For example the last test was actually returning `undefined`, which, while valid, felt odd.

Happy to change it back.

i-hardy: I'd consider this still tied to the implementation in that if we ever change or remove the `weeklyLandingHero` id the test will break even if there's nothing wrong with the code, but it's up to you.

jamesgorrie: I'll leave it as is as it seems both ways seem to make sense.

My logic is "Given these configuration options, `size` and `id`, I expect this answer, whereas if you lookup the URL in the expected and test cases, you're essentially ignoring it as a parameter to `theGrid`.

I think I might be getting slightly overly-philosophical on the case, but interesting to think through.

PR #5412, opened 2023-11-06
Hide decimals when they're unnecessary in benefits list

iainjchambers-guardian:
Hotfixing an issue caused by a [previous PR](https://github.com/guardian/support-frontend/pull/5393) where amounts displayed in the benfits container are rounded to two decimal places.

## Why are you doing this?
The previous PR had the unintended effect of converting whole numbers to floats with two decimal places. So **£95** became **£95.00** which is unnecessary.

## Is this an AB test?
- [ ] Yes
- [x] No

## Screenshots
| Before | After |
| ------- | ---- |
| image | image |


jamesgorrie (approved):
Noice

PR #5413, opened 2023-11-06
Add US EOY 2023 switch

jamesgorrie:

Adds a switch to be able to deliver the [US EOY 2023 campaign](https://github.com/guardian/support-frontend/pull/5414) behind. 

This should also make it easier to deploy i.e. just turn it on.

[**Trello Card**](https://trello.com/c/IkscfJ9B/1624-us-eoy-checkout-image-copy-change)

This uses an `Option[SwitchState]` and exposes it on the `window.guardian.campaigns` object as it feels slightly different to the current `CampaignSwitches` which feel more permanent and less test-y. This also then requires far fewer changes to the code where we do decoding for required objects e.g. [`SwitchesSpec.scala`](https://github.com/guardian/support-frontend/blob/main/support-frontend/test/admin/settings/SwitchesSpec.scala#L130-L142)

This might not be the most long-term resilient way of doing this. We could create a new block of switches in the tool, but that feels like overkill where this might work well enough.

It does point out that if anyone had to delete any of the expected switches from the control panel, the decoding code would bork with an `io.circe.DecodingFailure$DecodingFailureImpl: DecodingFailure at .eoyUs2023: Missing required field` error, which feels quite brittle. I might put in another PR to resolve that separately.

## `settingsMock`

As the `campaigns.ts` code now uses `guardian.windown.settings` which is `undefined` if you don't have the mocks included in:
* tests
* storybook

So we've had to add that in those places too.

```[tasklist]
### Tasks
- [x] Deploy switch config to `DEV`, `CODE` and `PROD`
``` 

CC @i-hardy as this might be useful to put your ticker behind as I _think_ it's part of this campaign?


LAKSHMIRPILLAI (approved):
👍🏼

support-frontend/assets/pages/weekly-subscription-checkout/components/weeklyCheckoutForm.test.tsx
```diff
@@ -1,6 +1,7 @@
 /* eslint-disable eslint-comments/require-description -- This is a mocks file, it is not intended to be good code! */
 /* eslint-disable @typescript-eslint/no-unsafe-return */
 /* eslint-disable @typescript-eslint/no-unsafe-assignment */
+import '__mocks__/settingsMock';
```
LAKSHMIRPILLAI, line 4: I think as we are importing  '__mocks__/settingsMock' to  testStore.ts file, importing it into individual test files may not be necessary

support-frontend/assets/components/subscriptionCheckouts/stripeForm/stripeForm.test.tsx
```diff
@@ -1,4 +1,5 @@
 /* eslint-disable @typescript-eslint/require-await -- To simplify mocking of functions that return promises */
+import '__mocks__/settingsMock';
```
LAKSHMIRPILLAI, line 2: Same with the weeklyCheckoutFrom test, as we are importing  '__mocks__/settingsMock' to  testStore.ts file,it seems  importing it into individual test files may not be necessary

support-frontend/assets/pages/weekly-subscription-checkout/components/weeklyCheckoutForm.test.tsx
jamesgorrie: Yea, this is true, but if that's the case, I would really like to fund out where the highest entrypoint is that I can do this. Do you have any ideas?

jamesgorrie: We can address this in another PR.

PR #5419, opened 2023-11-07
AN and AMP sources should only use regional amounts tests

KaliedaRik:
Adds a check to determine the source of an incoming user (`GOOGLE_AMP, APPLE_NEWS`) and enforces the requirement to show these users the appropriate regional data in the amounts card.

[**Trello Card**](https://trello.com/c/JtwmCKxH)

## Why are you doing this?
Epics shown on Google AMP and Apple News articles will always include `single`, `monthly` and `annual` payment options. We need to ensure that the amounts card shown on the landing page similarly includes all payment options as part of a seamless user journey.

## Is this an AB test?
- [ ] Yes
- [x] No

Not required - will not affect visual output

## Screenshots
Not required - will not affect visual output

### DIFF
--- a/support-frontend/assets/helpers/abTests/abtest.ts
+++ b/support-frontend/assets/helpers/abTests/abtest.ts
@@ -268,7 +268,11 @@ function getAmountsTestParticipations(
 			t.targeting.countries.includes(country),
 	);
 	let targetTest;
-	if (targetTestArray.length) {
+	const source = getSourceFromAcquisitionData() ?? '';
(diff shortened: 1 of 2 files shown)


tomrf1 (approved):
This change looks good to me.
There's a [PR to refactor the amounts test code](https://github.com/guardian/support-frontend/pull/5427), which would mean we don't have to do this in two places. But your new `getSourceFromAcquisitionData` function will be required regardless.

PR #5420, opened 2023-11-07
Add global AB test for make it annual

iainjchambers-guardian:
Adding dynamic numbers into the make it annual nudge, similar to the US version but with the current text

[**Trello Card**](https://trello.com/c/6HdIGPm8/1656-make-it-annual-nudge-global-test)

## Is this an AB test?
- [x] Yes
- [ ] No

## Screenshots
| Before | After | 
| --- | --- |
| image | image image |


jamesgorrie (approved):
This looks good.

I am just trying to work out what the tests are now.

Previously it was just to see if nudge gets traction in the US.

Now it's to see if nudge get's traction, with a slightly different implementation outside the US?

iainjchambers-guardian:
> This looks good.
> 
> I am just trying to work out what the tests are now.
> 
> Previously it was just to see if nudge gets traction in the US.
> 
> Now it's to see if nudge get's traction, with a slightly different implementation outside the US?

Yeah, you're correct that initially this was a nudge test for the US and that is currently running, but this is to implement a similar functionality with different copy for global, excluding the US.

GHaberis:
@iainjchambers-guardian  we should omit the countries below from this test. You can add the `omitCountries` key to the test definition with the following array. 

`[
    "RS",
    "EG",
    "PK",
    "MU",
    "BH",
    "MA",
    "MC",
    "OM",
    "GE",
    "NC",
    "TZ",
    "ZM",
    "AL",
    "BD",
    "KZ",
    "CW",
    "DO",
    "GP",
    "MQ",
    "PF",
    "TN",
    "BQ",
    "AX",
    "SN",
    "AM",
    "CM",
    "AO",
    "KG",
    "GA",
    "UZ",
    "MD",
    "DZ",
    "TJ",
    "LS",
    "CG",
    "TG",
    "NE"
]`

GHaberis:
@iainjchambers-guardian could you also a `targetPage` to this ab test definition, as done here on the other definitions: https://github.com/guardian/support-frontend/pull/5435

iainjchambers-guardian:
> @iainjchambers-guardian could you also a `targetPage` to this ab test definition, as done here on the other definitions: #5435

@GHaberis I've replicated the notUkLanding page regex and swapped the US for UK, figured it will add a redundancy for keeping the test off the US page but I could just add an entry for all landing pages. Let me know what you think.

support-frontend/assets/components/checkoutNudge/checkoutNudgeContainer.tsx
```diff
@@ -92,27 +98,44 @@ export function CheckoutNudgeContainer({
...
 
-	if (dynamic) {
+	const calcWeeklyAmount = (annualAmount: number, toNearest?: number) => {
+		const weeklyAmount = Math.ceil((annualAmount * 100) / 52);
+
+		// Current/control solution is to round to the nearest 10p so this may be
+		// disposable if the variant is adopted as we want the dynamic version to be
+		// rounded to the nearest 1p
+		const roundedWeeklyAmount =
+			toNearest !== undefined
```
i-hardy, line 108: I think it would be better to give `toNearest` a default value- presumably 1?- rather than to check for `undefined`.

support-frontend/assets/components/checkoutNudge/checkoutNudgeContainer.tsx
```diff
@@ -92,27 +98,44 @@ export function CheckoutNudgeContainer({
 
 	let title, subtitle, paragraph;
 
-	if (dynamic) {
+	const calcWeeklyAmount = (annualAmount: number, toNearest?: number) => {
```
i-hardy, line 101: I know the codebase is a bit all over the place on this for historic reasons, but function definitions (`function calcWeeklyAmount(//etc)`) are preferable for new code.

support-frontend/assets/components/checkoutNudge/checkoutNudgeContainer.tsx
```diff
@@ -92,27 +98,44 @@ export function CheckoutNudgeContainer({
...
+			toNearest !== undefined
+				? Math.round(weeklyAmount / toNearest) * toNearest
+				: weeklyAmount;
+
+		return countryGroupId === 'GBPCountries' && roundedWeeklyAmount < 100
+			? roundedWeeklyAmount.toString() + `p`
+			: currencyGlyph +
+					(roundedWeeklyAmount / 100).toFixed(
+						roundedWeeklyAmount % 100 === 0 ? 0 : 2,
+					);
```
i-hardy, line 117:
```suggestion
		if (countryGroupId === 'GBPCountries' && roundedWeeklyAmount < 100) {
			return `${roundedWeeklyAmount.toString()}p`;
		} else {
			const fixedDigitAmount = roundedWeeklyAmount % 100 === 0 ? 0 : 2;
			const amountString = (roundedWeeklyAmount / 100).toFixed(fixedDigitAmount);
			return `${currencyGlyph}${amountString}`
		}
```
Proper if statements are always preferable to ternaries unless the condition is very simple- this is _much_ easier to read and understand for someone new to the code. Similarly extracting individual calculations to named variables makes the code a lot easier to follow.

We should also avoid ES5-style string concatenation in favour of template literals.

support-frontend/assets/components/checkoutNudge/checkoutNudgeContainer.tsx
```diff
@@ -92,27 +98,44 @@ export function CheckoutNudgeContainer({
...
-			countryGroupId === 'GBPCountries' && weeklyMinAmount < 100
-				? weeklyMinAmount.toString() + `p`
-				: currencyGlyph +
-				  (weeklyMinAmount / 100)
-						.toFixed(weeklyMinAmount % 100 === 0 ? 0 : 2)
-						.toString();
+		if (isDynamicGlobal) {
+			subtitle = `with ${clampedAmountToCurrenyStr} (${calcWeeklyAmount(
+				clampedAmount,
+			)} per week)`;
```
i-hardy, line 129: Is it possible to avoid reassigning the `title`/`subtitle`/`paragraph` variables here? It would be nice to be able to have a separate function to handle this logic that always returned an object like this:
```
{
	title: 'xyz',
	subtitle: 'abc',
	paragraph: 'lorem impsum',
}
```
than to have `let` variables with their values changing and a possible `undefined` state.

iainjchambers-guardian: It's totally possible, just didn't think it was worth it at this point when they're conditional on A/B tests that are only temporary and when they end, all that conditional logic will presumably be removed and those variables can go back to being simple `const`s that are set once outside of all these if statements

iainjchambers-guardian: That said the outcome of the A/B tests might be that we have one title for US and one for everyone else in the world so it's probably still worth implementing your suggested solution

support-frontend/assets/components/checkoutNudge/checkoutNudgeContainer.tsx
iainjchambers-guardian: I've extracted this out of the component and made it a pure function that sits outside of it now, could be considered a utility/helper function that we could move elsewhere. Let me know what you think.

support-frontend/assets/components/checkoutNudge/checkoutNudgeContainer.tsx
i-hardy: It's been my experience fairly consistently that anything you write 'just for an A/B test' that's a little bit messy has about a 75% chance of ending up being shipped with no time to make it any better. Always better to go for the cleaner version if at all possible.

rBangay (approved):

PR #5426, opened 2023-11-08
fix: change US EOY content and do correct line breaks on heading

jamesgorrie:
- Adding line breaks to the heading for US EOY 2023 campaign
- Changing the copy

It would be good to have a think about how we might support this amount of copy control if it were configurable.

 - [x] [Tested with screen reader](https://webaim.org/articles/voiceover/) (mostly to revalidate that `<br />` which has no affect, which it doesn't)

## Screenshots

| before | after |
| --- | --- |
| ![Before][before] | ![After][after] |

[before]: https://github.com/guardian/support-frontend/assets/31692/b60e8cac-9b17-4258-84fc-e77ec2f8fae7
 
[after]: https://github.com/guardian/support-frontend/assets/31692/aeba7190-c4f2-4aca-b8ec-c6f3dc794e2d

### DIFF
--- a/support-frontend/assets/pages/supporter-plus-landing/twoStepTest/checkoutScaffold.tsx
+++ b/support-frontend/assets/pages/supporter-plus-landing/twoStepTest/checkoutScaffold.tsx
@@ -130,7 +130,16 @@ export function SupporterPlusCheckoutScaffold({
 		isUsEoy2023CampaignEnabled && countryGroupId === 'UnitedStates';
 
 	const heading = showUsEoy2023Content ? (
-		<LandingPageHeading heading="Make a year-end gift to the Guardian" />
+		<LandingPageHeading
+			heading={
+				<>
+					Make a<br />
+					year-end gift
+					<br />
+					to the Guardian
+				</>
+			}
+		/>
 	) : (
 		<LandingPageHeading heading="Support&nbsp;fearless, independent journalism" />
 	);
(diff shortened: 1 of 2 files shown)


support-frontend/assets/pages/supporter-plus-landing/twoStepTest/checkoutScaffold.tsx
```diff
@@ -130,7 +130,16 @@ export function SupporterPlusCheckoutScaffold({
...
 	const heading = showUsEoy2023Content ? (
-		<LandingPageHeading heading="Make a year-end gift to the Guardian" />
+		<LandingPageHeading
+			heading={
+				<>
+					Make a<br />
+					year-end gift
+					<br />
+					to the Guardian
+				</>
```
rupertbates, line 140: Does this work across all breakpoints?

rupertbates (approved):

support-frontend/assets/pages/supporter-plus-landing/twoStepTest/checkoutScaffold.tsx
jamesgorrie: It does indeed in the sense of that the left column is a fixed width across all breakpoints, and eventually just disappears.

## Wide
Screenshot 2023-11-08 at 10 24 44

## Widish
Screenshot 2023-11-08 at 10 24 55

## Narrow + gone
Screenshot 2023-11-08 at 10 25 03

PR #5427, opened 2023-11-08
Refactor amounts test logic

tomrf1:
### Current situation:
- `getAmountsTestParticipations` potentially assigns the user to an amounts AB test/variant. If the chosen test only has a single variant then the participation is not added, as we don't want to track it.
- later, `getAmounts` either uses the amounts test participation (from the step above), or assigns the user to an amounts AB test/variant. It then uses this to resolve the test variant to the actual amounts configuration (a `SelectedAmountsVariant`).

These two functions duplicate functionality for assigning a user to an AB test. This logic is quite confusing to maintain.
The main reason for it is that we don't always want to track an amounts AB test participation, but we do always want to select an amounts configuration.

### Proposed refactor:
This PR replaces these two functions with `getAmountsTestVariant`. This returns both:
1. the `SelectedAmountsVariant`, with the amounts config
2. optionally an AB test participation, if we actually want to track it

The optional participation is then added to the rest of the participations (if it exists).

I've also added some unit tests.


support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -197,111 +191,140 @@ function getServerSideParticipations(): Participations | null | undefined {
...
+		test: AmountsTest,
+		testName: string,
+		variantName: string,
+	): Participations | undefined => {
+		// Check if we actually want to track this test
+		const pathMatches = targetPageMatches(
+			path,
+			'/??/contribute|thankyou(/.*)?$',
+		);
+		if (pathMatches && test.variants.length > 1 && test.isLive) {
```
tomrf1, line 234: we've noticed that, in the existing code, the check for more than one variant [only happens](https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/helpers/abTests/abtest.ts#L296) if there isn't an amounts test in the url.
If there is an amounts test in the url then we currently do add it to the participations [regardless of the number of variants](https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/helpers/abTests/abtest.ts#L229).

tomrf1: This inconsistency is probably not deliberate, so we need to decide which is correct

tomrf1: We've decided this logic is correct. Participation of an amounts test is only tracked if it's a genuine AB test with more than just a control

GHaberis (approved):
Looks good to me, thanks for writing unit tests!

support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -197,115 +191,146 @@ function getServerSideParticipations(): Participations | null | undefined {
...
-	*/
-	const targetTestArray = amounts.filter(
-		(t) =>
-			t.isLive &&
-			t.targeting.targetingType === 'Country' &&
-			t.targeting.countries.includes(country),
-	);
-	let targetTest;
+	// First try country-targeted tests
 	const source = getSourceFromAcquisitionData() ?? '';
```
tomrf1, line 278: I've just pushed after a rebase that brings in the change to exclude AMP/AN from country-targeted tests

PR #5429, opened 2023-11-08
fix: remove `nbsp;` in `LandingPageHeadline`

jamesgorrie:
Removes the `nbsp;` in the LandingPageHeadline as it seems to render incorrectly on the https://support.theguardian.com/uk/kindle ppage.

It also seems to be unnecessary as it would never break between those words give the left column is a fixed width or hidden. I imagine this was here for a previous version of the UI.

I haven't quite got to the bottom of why this is the case, but thought it worth getting a hotfix in.

#### Desktop (before)
Screenshot 2023-11-08 at 13 27 42

#### Desktop (after)

Screenshot 2023-11-08 at 13 22 10

#### Mobile (after)
Screenshot 2023-11-08 at 13 22 05

---

#### Desktop 
Screenshot 2023-11-08 at 13 25 41

#### Mobile
Screenshot 2023-11-08 at 13 21 56

### DIFF
--- a/support-frontend/assets/pages/supporter-plus-landing/components/landingPageHeading.tsx
+++ b/support-frontend/assets/pages/supporter-plus-landing/components/landingPageHeading.tsx
@@ -31,7 +31,7 @@ const headingStyles = css`
 
 type Props = { heading?: string | JSX.Element };
 export function LandingPageHeading({
-	heading = 'Support&nbsp;fearless, independent journalism',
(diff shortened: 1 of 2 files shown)


jamesgorrie:
Confirmed it's undoing the nasty:
https://www.chromatic.com/test?appId=62e115310aef0868687b2322&id=654b8c3966c81f0b699ef070

iainjchambers-guardian (approved):

PR #5432, opened 2023-11-08
feat: deploy fastly snippets

jamesgorrie:
Adds the ability to deploy [Fastly's regular snippets](https://docs.fastly.com/en/guides/using-regular-vcl-snippets) via the `workflow_dispatch` of an action.

These are currently deployed manually to either `CODE` (`support.code.dev-theguardian.com`) or PROD (`support.theguardian.com`), which you can select when dispatching the workflow (this only becomes available once merged).

Screenshot 2023-11-09 at 17 14 29

---

The action has corresponding secret values for the `FASTLY_API_TOKEN` and `FASTLY_SERVICE_ID` set using [GitHub Environments](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment).

The process is the standard Fastly process that we currently do via the console (and is hopefully evident in the script):
- Find the service's active version
- Clone this version
- Copy the snippet to that version
- Active that version
- Lock that version

I was going to make the active/lock step optional, but given we have `CODE` and we're already only deploying on `workflow_dispatch`, that felt like a good balance of initial friction vs safety before we go to full continuous delivery.

If you'd like to have a whirl - I have a demo repo setup here that deploys to a demo set of Fastly services (not linked to any URLs): https://github.com/guardian/actions-deploy-fastly-snippets/actions/workflows/deploy-fastly-snippets.yml

This work was inspired by the work we are doing for adding the digital subscription product, which needs some effort around routing. By bringing in the redirects VCL this adds visibility as to how our routes work in `PROD`.


support-fastly/snippets/redirects-table.vcl
jamesgorrie, line 1: I'm not 100% sure if this should live here, or closer to the product that uses it, in this case `support-frontend`?

I keep flip-flopping between co-location to service vs having a full view of what's in Fastly.

rupertbates: I think it would be more intuitive to have it in the support-frontend project in a similar location to where you might find a play routes file

rupertbates (approved):
Really good stuff 👍

PR #5433, opened 2023-11-09
Re-support `/subscribe/digital` routes

jamesgorrie:
Mostly renaming `kindle` => `digital` to prepare for the digital subscription work coming up.

This will maintain that:
- `/kindle` => `/{country}/kindle?promoCode=DLESNCWON`
- `/digital` => `/{country}/subscribe/digital`

Both of these are rendered from the `DigitalSubscriptionController.digital` (which is what is happening now).

We also support both routes in the `ReactRouter` while we're developing this, but will remove the kindle route when we're out in `PROD`.

None of these pages are indexed by Google [as we `noindex` on them](https://github.com/guardian/support-frontend/blob/bb8b3847f605c9584a1ebb14198380f8c50795a6/support-frontend/app/controllers/DigitalSubscriptionController.scala#L88), so we're not too worried about fragmenting our google juice.


support-frontend/conf/routes
```diff
@@ -91,16 +91,13 @@ GET  /$country<(uk|us|au|eu|int|nz|ca)>/subscribe                  controllers.S
 # subscribe route. We just redirect to the subscriptions site and let its geolocation handle it.
 GET  /:country/subscribe                                           controllers.SubscriptionsController.legacyRedirect(country: String)
 
-GET  /digital                                                      controllers.DigitalSubscriptionController.digitalGeoRedirect(orderIsAGift: Boolean = false)
+# Digital subscriptions
```
paul-daniel-dempsey, line 94: top digital route removed?

paul-daniel-dempsey (approved):
Sorry late to approve, presume `/digital` doe snot make sense to keep this route.

support-frontend/conf/routes
jamesgorrie: Nice spot, thanks. I think we should be trying to keep things as close to current as possible, so I've added it back in.

PR #5435, opened 2023-11-10
Add targetPage to AB test definitions

GHaberis:

Whilst looking at Quantum Metric with @LAKSHMIRPILLAI and @paul-daniel-dempsey we noted that despite viewing the Support Checkout (https://support.theguardian.com/uk/contribute) we were being allocated to the `nationalDelivery` AB Test. The `nationalDelivery` AB test is for the Paper checkout flow only so we shouldn't allocate users across the entire site into this test - as it could potentially lead to misreporting of data. I noted the same issue for the other AB tests `supporterPlusOnly` and `makeItAnnualNudge`.

To remedy this I've add a `targetPage` prop to all these AB tests so:

- `supporterPlusOnly`/`makeItAnnualNudge` run only on the Support Checkout and Thank You page.
-  `nationalDelivery` runs only on the Newspaper Landing Page, Checkout and Thank You page.

I had to update the `paperLandingWithGuestCheckout` REGEX as it wasn't picking up https://support.theguardian.com/uk/subscribe/paper/delivery

### DIFF
--- a/support-frontend/assets/helpers/abTests/abtestDefinitions.ts
+++ b/support-frontend/assets/helpers/abTests/abtestDefinitions.ts
@@ -33,7 +33,7 @@ export const pageUrlRegexes = {
 		paper: {
 			// Requires /subscribe/paper, allows /checkout or /checkout/guest, allows any query string
 			paperLandingWithGuestCheckout:
-				/\/subscribe\/paper(\/checkout|\/checkout\/guest)?(\?.*)?$/,
+				/\/subscribe\/paper(\/delivery|\/checkout|\/checkout\/guest)?(\?.*)?$/,
 		},
 		subsWeeklyPages:
 			'(/??/subscribe(\\?.*)?$|/??/subscribe/weekly(\\/checkout)?(\\?.*)?$)',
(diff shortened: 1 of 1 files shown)


andrewHEguardian (approved):
Good to know for future! The `nationalDelivery` test wasn't really an AB test and was basically used as a feature flag, if that makes any difference to the reporting?

support-frontend/assets/helpers/abTests/abtestDefinitions.ts
```diff
@@ -58,6 +58,7 @@ export const tests: Tests = {
 		isActive: true,
 		referrerControlled: true,
 		seed: 2,
+		targetPage: pageUrlRegexes.contributions.allLandingPagesAndThankyouPages,
```
paul-daniel-dempsey, line 61: Utterly minor but would is be better to label this `allContributionsLandingAndThankyouPages` or is the wider description sufficient.

GHaberis:
> Good to know for future! The `nationalDelivery` test wasn't really an AB test and was basically used as a feature flag, if that makes any difference to the reporting?

Yeah I don't actually think it's an issue for this particular test, as it's not really a test we're reporting on 👍

paul-daniel-dempsey (approved):
Looks good

PR #5441, opened 2023-11-13
Scroll back to the top of the page when transitioning to the print thank you page

i-hardy:

This returns the user's scroll position to the top when they reach the thank you page for both newspaper and Guardian Weekly checkouts.

[**Trello Card**](https://trello.com/c/AMe5EYem)

## Why are you doing this?

This is to resolve an issue flagged up during manual testing for the newspaper national delivery rollout. We have the same issue on Guardian Weekly so I implemented it there too just for consistency's sake


rBangay (approved):

rBangay:
Thanks for doing this

paul-daniel-dempsey (approved):
Looks good.

PR #5443, opened 2023-11-14
Enable nightly backups for SupporterProductData tables

jacobwinch:

This PR allows us to start backing up these DynamoDB tables using https://github.com/guardian/aws-backup. For more details on this project see [this doc](https://docs.google.com/document/d/11EZtuVHCnjavE9AYLroiuDUsMzbp2ulFBtj4VkyCKNc/edit) and this [audit](https://docs.google.com/spreadsheets/d/1gNh_yQ3GVo_dVAemgZfRZM_MaQxep6_x-hzrhvVBGHY/edit#gid=1523470588) that @guardian/devx-reliability have been working on with Heads of Engineering.

## Why are you doing this?

We will be able to recover (most) data stored in this table in the unlikely event that it is deleted. 

## Deployment

As far as I can tell, this CloudFormation stack is deployed manually[^1].

If people are happy with this change in principle, I will:

- [x] Manually update the `CODE` stack via the console (to check that this works)
- [ ] Merge the PR
- [ ] Manually update the `PROD` stack via the console

[^1]: It looks like [some of this infrastructure is deployed via Riff-Raff](https://github.com/guardian/support-frontend/blob/8329681e585714f3334c3ffaac6effca9027b4a0/supporter-product-data/riff-raff.yaml#L11), but not [these tables](https://github.com/guardian/support-frontend/blob/main/supporter-product-data/cloudformation/dynamo-tables.yaml).

### DIFF
--- a/supporter-product-data/cloudformation/dynamo-tables.yaml
+++ b/supporter-product-data/cloudformation/dynamo-tables.yaml
@@ -40,6 +40,9 @@ Resources:
       TimeToLiveSpecification:
         AttributeName: "expiryDate"
         Enabled: true
+      Tags:
+        - Key: devx-backup-enabled
+          Value: true
 
   WriteCapacityScalableTarget:
     Condition: CreateProdResources
@@ -126,6 +129,9 @@ Resources:
       TimeToLiveSpecification:
         AttributeName: "expiryDate"
         Enabled: true
+      Tags:
+        - Key: devx-backup-enabled
+          Value: true
 Outputs:
   ProvisionedTableOutput:
     Condition: CreateProdResources


johnduffell:
I don't think we need to back these up as they are just a cache from zuora, we just repopulate them from zuora when they are lost or corrupted (which does happen sometimes)

jacobwinch:
> I don't think we need to back these up as they are just a cache from zuora, we just repopulate them from zuora when they are lost or corrupted (which does happen sometimes)

Thanks @johnduffell. We (DevX) have also opted-out of backups for a table which is used as a cache, so this makes sense to me. I will close this and have opened https://github.com/guardian/support-frontend/pull/5455 to replace it.

cc @paulbrown1982 for visibility, as I will also update the main audit spreadsheet.

johnduffell:
Paul I don't think backups would work even if they were quicker - because  it's constantly updated and once data is lost we would repopulate anyway.
I think restoration takes quite a few hours as long even if we scale up dynamo, but Rupert usually operates the restore so he may know more.
Impact if the tables.being missing would be digital benefits wouldn't be visible on the site during that time.

jacobwinch:
I've gone back and forth on this (see https://github.com/guardian/support-frontend/pull/5455!), but after a bit more discussion and testing, I think it is worth enabling this after all as the restore time (16 minutes) is much faster than the current approach (estimated at ~3 hours).

johnduffell (commented):
thanks for going back and forth on this! plus useful input from @rupertbates and @paulbrown1982 
Hopefully we won't need to use it though!
If we need a restore is that self-service or need DevX support?
👍

johnduffell (approved):

jacobwinch:
Thanks @johnduffell! 

> If we need a restore is that self-service or need DevX support?

It's self-service but there is a [general guide here](https://docs.google.com/document/d/1V_y5RDrNH_Kjw_pnbPa-gS_t6F_5LcG3ILTtCFkt7DQ/edit) which may be useful. Also happy to chat more about this specific use-case if helpful!

jacobwinch:
Unfortunately testing/deploying this was not straightforward: https://github.com/guardian/support-frontend/pull/5524.

jacobwinch:
This needs to be rebased once https://github.com/guardian/support-frontend/pull/5524 has been merged.

jacobwinch:
I've applied this change in `PROD`:

image

jacobwinch:
This is working:

image

PR #5445, opened 2023-11-14
AgentsEndpoint.Response: Remove message field

undergroundquizscene:
This field has been removed by PaperRound from the API, presumably because they didn’t intend for us to rely on it. We weren’t actually _using_ the field, so we shouldn’t have been parsing it in the first place!

At the moment this is breaking the sending of thank you emails from support-workers, so this commit should fix that by no longer parsing it.


andrewHEguardian (approved):

johnduffell (approved):
great. then we can review the rest of the fields in more detail 👍

undergroundquizscene:
Tested in CODE, verified that it fixed the problem

PR #5451, opened 2023-11-15
Fastly snippets

jamesgorrie:
A few new features to the Fastly Snippets deploy to make it more useful.

### Delete snippets before adding them
Configurable by UI, on by default.

This means that your code will always be the source of truth for the state of your Fastly service avoiding drift and inconsistencies. There are some instances in which you might not want this, so it is configurable.

This avoids deploying conflicting VCL that might exist in the service e.g. table names.

### Locking and activated newly cloned service
Configurable by UI, on by default.

This allow you to skip the lock and activate steps. As this is a new bit of functionality I would like to review the service in Fastly a few times before feeling more confident that this will do what I want it to.

### Moving snippets configuration into the JS file
This makes it slightly less reusable across repos, but will do for now. As a secondary step in trying to make this more useful, I might add an ENV var to a config file that follows the same shape.

### The rest of the snippets
Previously I had just deployed the redirects table, but none of the actual snippets that make that work. I've gone with the whole lot as a few are quite tightly coupled, so it makes sense to deploy them as 1.

[**Trello Card**](https://trello.com/c/GJ2YBWKG/1689-deploy-fastly-vcl-via-github-actions)

## Why are you doing this?

To hopefully give us a clearer understanding of our paths and routes while working on the digital subscription relaunch.

## Testing

You can run the script locally, so long as you have a service ID and Fastly API token that can modify that service set as env variables e.g.
```bash
# ~/.gu/fastly-service-env-vars

# Bash
export FASTLY_SERVICE_ID="abc123"
export FASTLY_API_TOKEN="abc123"
export DELETE_SNIPPETS="true"
export LOCK_AND_ACTIVATE_SERVICE="true"

# Fish
set -Ux FASTLY_SERVICE_ID abc123
set -Ux FASTLY_API_TOKEN abc123
set -Ux DELETE_SNIPPETS true
set -Ux LOCK_AND_ACTIVATE_SERVICE true
```

and then 

```bash
source ~/.gu/fastly-service-env-vars
node ./.github/workflows/deploy-fastly-snippets.js
```


.github/workflows/deploy-fastly-snippets.yaml
```diff
@@ -13,15 +22,15 @@ jobs:
...
     steps:
       - name: Checkout repo
         uses: actions/checkout@v4
       - uses: actions/setup-node@v3
         with:
           node-version: 18
-      - name: Fastly snippet - Redirects table
+      - name: Deploy Fastly snippets
         run: node .github/workflows/deploy-fastly-snippets.js
         env:
```
LAKSHMIRPILLAI, line 35: Can we  remove this if no value is specified for env?

LAKSHMIRPILLAI (approved):
LGTM 👍🏼

LAKSHMIRPILLAI:
Should we keep the snippet names same as that is used in the PROD?

.github/workflows/deploy-fastly-snippets.yaml
jamesgorrie: Nope, I've removed it.

jamesgorrie:
> Should we keep the snippet names same as that is used in the PROD?

@LAKSHMIRPILLAI I've tried to standardise them around the [Fastly's VCL lifecycle](https://developer.fastly.com/learning/vcl/using/#the-vcl-request-lifecycle). 

They were to some degree mostly there. 

I have also added the `support-frontend` prefix as that's where they live in the codebase, and I can see where we might want to add some for e.g. the `payments-api`. This should help working out where the script lives when debugging.

Did you have any specific reasons you thought they should remain the same?

LAKSHMIRPILLAI:
> > Should we keep the snippet names same as that is used in the PROD?
> 
> @LAKSHMIRPILLAI I've tried to standardise them around the [Fastly's VCL lifecycle](https://developer.fastly.com/learning/vcl/using/#the-vcl-request-lifecycle).
> 
> They were to some degree mostly there.
> 
> I have also added the `support-frontend` prefix as that's where they live in the codebase, and I can see where we might want to add some for e.g. the `payments-api`. This should help working out where the script lives when debugging.
> 
> Did you have any specific reasons you thought they should remain the same?

Sorry, I was not aware that we have snippets for payment-api as well.So prefixing with project name makes sense.I don't have any specific reason to think about keeping names as in PROD, just that when I checked the  diff of versions after testing this change for CODE, the names from PROD seemed much simpler

PR #5452, opened 2023-11-16
add audience to makeItAnnualNudgeGlobal AB test

GHaberis:

The `makeItAnnualNudgeGlobal` was missing an Audience so no users were allocated to it, this PR fixes that by assigning ALL users to the test. We actually want to omit the US from the test so we'd already added the US to the `omitCountries` list, however I've split the list by separating the countries affected by the VAT issue into an additional array, which I think helps readability.

We could look at the type definition for the `audiences` prop if a value is required to actually run a test, I'm not doing that here because of time.


rBangay (approved):

paul-daniel-dempsey (approved):
perfect

PR #5456, opened 2023-11-20
Read default-promos.json for Digital checkout

LAKSHMIRPILLAI:

## Is this an AB test?
- [ ] Yes
- [ ] No

### DIFF
--- a/support-frontend/app/controllers/DigitalSubscriptionController.scala
+++ b/support-frontend/app/controllers/DigitalSubscriptionController.scala
@@ -2,12 +2,11 @@ package controllers
 
 import actions.CustomActionBuilders
 import admin.settings.{AllSettings, AllSettingsProvider, SettingsSurrogateKeySyntax}
-import assets.{AssetsResolver, RefPath, StyleContent}
+import assets.{AssetsResolver}
 import com.gu.support.catalog.DigitalPack
 import com.gu.support.config.{PayPalConfigProvider, StripeConfigProvider}
 import com.gu.support.encoding.CustomCodecs._
 import services.pricing.{PriceSummaryServiceProvider, ProductPrices}
-import com.gu.i18n.Currency.{AUD}
 import com.gu.support.promotions._
 import com.gu.support.zuora.api.ReaderType.{Direct, Gift}
 import config.RecaptchaConfigProvider
@@ -22,6 +21,7 @@ import scala.concurrent.ExecutionContext
 
 class DigitalSubscriptionController(
     priceSummaryServiceProvider: PriceSummaryServiceProvider,
+    landingCopyProvider: LandingCopyProvider,
     val assets: AssetsResolver,
     val actionRefiners: CustomActionBuilders,
     testUsers: TestUserService,
@@ -50,6 +50,8 @@ class DigitalSubscriptionController(
   def digital(countryCode: String, orderIsAGift: Boolean): Action[AnyContent] = {
     MaybeAuthenticatedAction { implicit request =>
       implicit val settings: AllSettings = settingsProvider.getAllSettings()
+      val defaultPromos = priceSummaryServiceProvider.forUser(isTestUser = false).getDefaultPromoCodes(DigitalPack)
+      val queryPromos = request.queryString.get("promoCode").map(_.toList).getOrElse(Nil)
 
       if (!settings.switches.subscriptionsSwitches.enableDigitalSubGifting.isOn && orderIsAGift) {
         Redirect(routes.DigitalSubscriptionController.digitalGeoRedirect(false)).withSettingsSurrogateKey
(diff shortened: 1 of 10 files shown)


support-frontend/app/controllers/DigitalSubscriptionController.scala
```diff
@@ -50,6 +50,8 @@ class DigitalSubscriptionController(
   def digital(countryCode: String, orderIsAGift: Boolean): Action[AnyContent] = {
     MaybeAuthenticatedAction { implicit request =>
       implicit val settings: AllSettings = settingsProvider.getAllSettings()
+      val defaultPromos = priceSummaryServiceProvider.forUser(isTestUser = false).getDefaultPromoCodes(DigitalPack)
```
jamesgorrie, line 53: Should we be using the [test user value](https://github.com/guardian/support-frontend/pull/5456/files#diff-73774672ba93580adbffcdbbd2cbf93499940fa5b20598b4dec1949dab6b786bR68) here?

support-frontend/app/controllers/DigitalSubscriptionController.scala
```diff
@@ -86,13 +89,20 @@ class DigitalSubscriptionController(
             v2recaptchaConfigPublicKey,
             orderIsAGift,
             noindex = true,
-          ),
+          ) {
+            Html(
+              s"""<script type="text/javascript">
+                    window.guardian.productPrices = ${outputJson(productPrices(queryPromos, orderIsAGift))}
```
jamesgorrie, line 95: This value seems to be [set in the `subscriptionsCheckout`](https://github.com/guardian/support-frontend/blob/main/support-frontend/app/views/subscriptionCheckout.scala.html#L40). Should we just be adding the `orderAsGift` to that method?

If this is the case, it should be a matter of just setting one more prop to the `subscriptionCheckout` for `promotionCopy`.

jamesgorrie (commented):
Question on double setting the value for `window.guardian.productPrices`.

support-frontend/app/controllers/DigitalSubscriptionController.scala
```diff
@@ -63,12 +64,14 @@ class DigitalSubscriptionController(
...
-
         val testMode = testUsers.isTestUser(request)
         val promoCodes = request.queryString.get("promoCode").map(_.toList).getOrElse(Nil)
         val v2recaptchaConfigPublicKey = recaptchaConfigProvider.get(testMode).v2PublicKey
         val readerType = if (orderIsAGift) Gift else Direct
-
+        val defaultPromos = priceSummaryServiceProvider.forUser(isTestUser = false).getDefaultPromoCodes(DigitalPack)
+        val maybePromotionCopy = {
+          landingCopyProvider.promotionCopy(promoCodes ++ defaultPromos, DigitalPack, "uk", orderIsAGift)
+        }
```
jamesgorrie, line 74: Not sure if the brackets add any clarity?

```suggestion
        val maybePromotionCopy =
          landingCopyProvider.promotionCopy(promoCodes ++ defaultPromos, DigitalPack, "uk", orderIsAGift)
```

support-frontend/app/controllers/WeeklySubscriptionFormController.scala
```diff
@@ -66,6 +66,7 @@ class WeeklySubscriptionFormController(
       maybeIdUser,
       testMode,
       priceSummaryServiceProvider.forUser(testMode).getPrices(GuardianWeekly, promoCodes, readerType),
+      maybePromotionCopy = None,
```
jamesgorrie, line 69: Given this has a default value set, you don't need to pass this value through.

support-frontend/app/controllers/DigitalSubscriptionFormController.scala
```diff
@@ -71,6 +70,7 @@ class DigitalSubscriptionFormController(
       maybeIdUser,
       testMode,
       priceSummaryServiceProvider.forUser(testMode).getPrices(DigitalPack, promoCodes, readerType),
+      maybePromotionCopy = None,
```
jamesgorrie, line 73: Given this has a default value set, you don't need to pass this value through.

support-frontend/app/controllers/PaperSubscriptionFormController.scala
```diff
@@ -65,6 +65,7 @@ class PaperSubscriptionFormController(
       maybeIdUser,
       testMode,
       priceSummaryServiceProvider.forUser(testMode).getPrices(Paper, promoCodes),
+      maybePromotionCopy = None,
```
jamesgorrie, line 68: Given this has a default value set, you don't need to pass this value through.

support-frontend/app/controllers/SubscriptionsController.scala
```diff
@@ -82,7 +82,7 @@ class SubscriptionsController(
     val digitalSubscription = service
       .getPrices(
         DigitalPack,
-        List(DefaultPromotions.DigitalSubscription.Monthly.fiftyPercentOff3Months),
+        Nil,
```
jamesgorrie, line 85: Why are we changing this in this PR?

jamesgorrie (approved):
Couple of small questions, nothing blocking.

support-frontend/app/controllers/PaperSubscriptionFormController.scala
LAKSHMIRPILLAI: I passed it here, because  otherwise , it is showing me the error 
image

And if I pass just the  maybePromotionCopy , it is showing me the error 

image

support-frontend/app/controllers/SubscriptionsController.scala
LAKSHMIRPILLAI: I just changed it as support-models/src/main/scala/com/gu/support/promotions/DefaultPromotions.scala hasn't got any value for DigitalSubscription

support-frontend/app/controllers/PaperSubscriptionFormController.scala
jamesgorrie: Ah, OK - this is a nasty problem with using non-named parameters and so order counts. 
You could add `maybePromotionCopy` to the bottom of the parameters list in `subscriptionsCheckout` - or better, we could add the param names in to avoid this in future. 

Let's maybe pic this up outside of this PR.

PR #5460, opened 2023-11-23
Add client-side validation to prevent form submission if captcha hasn’t been selected with Direct Debit

LAKSHMIRPILLAI:

Add client-side validation  to prevent form submission if captcha hasn’t been selected with Direct Debit

[**Trello Card**](https://trello.com/c/VmdWC8sR/1425-look-into-captcha-validation-failures-not-preventing-support-workers-execution)

## Why are you doing this?

If the  getRecaptchaError  is not included in the  getDirectDebitFormErrors , it allows the form submission even if there is no recaptcha  selected  in Direct debit form when all the direct debit details/ fields are correct.
## Is this an AB test?
- [ ] Yes
- [x ] No

###  (Before change)-Even if the recaptcha in the Direct debit section was not selected, the user was allowed to submit the form and the support workers successfully executed the request.

image

The user was shown -"Please check the 'I'm not a robot' " checkbox error message

###  (After change)- The user won't be allowed to submit the form

image

### DIFF
--- a/support-frontend/assets/helpers/redux/selectors/formValidation/paymentValidation.ts
+++ b/support-frontend/assets/helpers/redux/selectors/formValidation/paymentValidation.ts
@@ -18,7 +18,6 @@ export function getStripeFormErrors(
 	const recaptchaErrors = getRecaptchaError(state);
 
 	if (!showErrors) return {};
-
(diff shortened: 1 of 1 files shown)


rBangay (approved):
🤖 ✅

i-hardy (approved):
Nice work!

PR #5463, opened 2023-11-24
fix: make ausTicker2023 tickerId AU

jamesgorrie:
Updating the tickerId as there is no `AUS` ticker file in S3.


GHaberis (approved):
👍

PR #5465, opened 2023-11-24
feat: digiSubThankyou heading subheading copy change

paul-daniel-dempsey:

## Is this an AB test?
- [ ] Yes
- [ ] No

### DIFF
--- a/support-frontend/assets/pages/digital-subscriber-thank-you/digitalSubscriptionThankYou.tsx
+++ b/support-frontend/assets/pages/digital-subscriber-thank-you/digitalSubscriptionThankYou.tsx
@@ -31,7 +31,7 @@ export function DigitalSubscriptionThankYou(): JSX.Element {
 		() => getCampaignSettings(campaignCode),
 		[],
 	);
-	const { countryId, countryGroupId, currencyId } = useContributionsSelector(
+	const { countryId, countryGroupId } = useContributionsSelector(
 		(state) => state.common.internationalisation,
 	);
 	const { csrf } = useContributionsSelector((state) => state.page.checkoutForm);
@@ -45,8 +45,9 @@ export function DigitalSubscriptionThankYou(): JSX.Element {
 	const paymentMethod = useContributionsSelector(
 		(state) => state.page.checkoutForm.payment.paymentMethod.name,
 	);
-	const { selectedAmounts, otherAmounts, billingPeriod } =
-		useContributionsSelector((state) => state.page.checkoutForm.product);
+	const { selectedAmounts, otherAmounts } = useContributionsSelector(
+		(state) => state.page.checkoutForm.product,
+	);
 	const { isSignedIn } = useContributionsSelector((state) => state.page.user);
 	const isNewAccount = userTypeFromIdentityResponse === 'new';
 	const { monthlyPrice, annualPrice } = useContributionsSelector(
@@ -132,9 +133,6 @@ export function DigitalSubscriptionThankYou(): JSX.Element {
 						<ThankYouHeader
 							name={firstName}
 							showDirectDebitMessage={paymentMethod === DirectDebit}
-							billingPeriod={billingPeriod}
-							amount={billingPeriod === 'Monthly' ? monthlyPrice : annualPrice}
-							currency={currencyId}
 						/>
 					</div>
 
--- a/support-frontend/assets/pages/digital-subscriber-thank-you/components/heading.tsx
+++ b/support-frontend/assets/pages/digital-subscriber-thank-you/components/heading.tsx
@@ -1,50 +1,16 @@
-import type { IsoCurrency } from 'helpers/internationalisation/currency';
-import type { BillingPeriod } from 'helpers/productPrice/billingPeriods';
-
 const MAX_DISPLAY_NAME_LENGTH = 10;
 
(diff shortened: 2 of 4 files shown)


rBangay (approved):

support-frontend/assets/pages/supporter-plus-thank-you/components/thankYouHeader/thankYouHeader.tsx
```diff
@@ -17,12 +18,14 @@ export const header = css`
 	}
 `;
 
-export const headerTitleText = css`
+export const headerTitleText = (
+	digitalFontSizeReduce?: boolean,
```
jamesgorrie, line 22: Instead of turning this into a method, which creates more CSS to push down the pipes, we could just use the `style` prop in `heading.tsx`, which not only would create less css, keeps that value colocated with the relevant component.

[It's described quite well here](https://emotion.sh/docs/best-practices#use-the-style-prop-for-dynamic-styles).

There's a little bit of me that wonders if this style should actually live in the header component as it seems like a strange coupling as it currently stands, but it works.

jamesgorrie (approved):

support-frontend/assets/pages/supporter-plus-thank-you/components/thankYouHeader/thankYouHeader.tsx
paul-daniel-dempsey: Not sure how I can get the  font to resize dependent on screen size using  this `Heading.style` method recommended? All the styles I see used in the codebase do not check the screen-size like this function does? SerializedStyles appear to be required. Have uploaded as far as I could get, however, `digital-subscriber-thank-you/thankyouHeader.tsx` line 26 is stuck in 34px regardless of screensize..

support-frontend/assets/pages/digital-subscriber-thank-you/digitalSubscriptionThankYou.tsx
```diff
@@ -45,13 +44,11 @@ export function DigitalSubscriptionThankYou(): JSX.Element {
 	const paymentMethod = useContributionsSelector(
 		(state) => state.page.checkoutForm.payment.paymentMethod.name,
 	);
-	const { selectedAmounts, otherAmounts, billingPeriod } =
-		useContributionsSelector((state) => state.page.checkoutForm.product);
+	const { selectedAmounts, otherAmounts } = useContributionsSelector(
```
GHaberis, line 47: It looks like there's quite a bit of Contributions/Supporter+ related logic copied over from the Support Checkout's Thank You page when this page was set-up for the `/kindle` route. We could probably review this and see whether any of this can be removed - not in this PR however, lets create a card to tidy this up.

GHaberis: Specifically I'm thinking do we need this code to check `isOneOff` and `amountIsAboveThreshold`, that are passed to `getThankYouModuleData`.

I'm assuming `isOneOff` is always `false`? And `amountIsAboveThreshold` is always `true` when coming through the Digi Subs checkout? Can you confirm? If so, we could remove this code that calculates the values of these variables and just pass the boolean value to `getThankYouModuleData`.

support-frontend/assets/pages/digital-subscriber-thank-you/digitalSubscriptionThankYou.tsx
```diff
@@ -106,8 +80,6 @@ export function DigitalSubscriptionThankYou(): JSX.Element {
 			'signIn',
 		),
 		'appDownloadKindle',
```
GHaberis, line 82: Could we relabel this "appDownloadEditionsHeader", currently the presence of "appDownloadKindle" here instructs `thankYouModuleData` to render the "Download the Guardian Editions app" container, the inclusion of "kindle" in the name is a hangover from when this checkout was served at the `/kindle` route, "appDownloadEditionsHeader" feels like a more relevant/accurate description of it's purpose now.

support-frontend/assets/pages/digital-subscriber-thank-you/digitalSubscriptionThankYou.tsx
paul-daniel-dempsey: Yes can confirm, digitalEdition never one-off and cannot (currently) enter other amount to get below minimum in digital checkout (ie `isOneOff`=`false`, `amountIsAboveThreshold`=`true`). Have made this change suggested.

support-frontend/assets/pages/digital-subscriber-thank-you/digitalSubscriptionThankYou.tsx
paul-daniel-dempsey: Yes, makes sense, have renamed 'kindle' references to 'editions' throughout the thank-you components.

support-frontend/assets/pages/supporter-plus-thank-you/components/thankYouHeader/thankYouHeader.tsx
paul-daniel-dempsey: Removed `Heading.style` as did not appear to  support media resizing.

PR #5473, opened 2023-11-27
Move the US into the global nudge test

i-hardy:

This removes the US-specific dynamic nudge test and adds the US to the global test, as requested.

[**Trello Card**](https://trello.com/c/w7TVgx9F)


rBangay (approved):
👍

PR #5475, opened 2023-11-27
Changes to redirect users in VAT impacted countries

LAKSHMIRPILLAI:

Any users landing on the subscription checkout route should be redirected them to the S+ checkout, where they will be allocated in the VAT compliant amounts test.

[**Trello Card**](https://trello.com/c/lzATPJUo/1698-editions-prevent-users-in-vat-impacted-countries-from-purchasing-digi-subs)

## Why are you doing this?
This is to prevent users in VAT impacted countries from purchasing Digital Subscriptions

## Is this an AB test?
- [ ] Yes
- [X ] No

### DIFF
--- a/support-frontend/conf/fastly-snippets/vat-compliance-redirects-recv.vcl
+++ b/support-frontend/conf/fastly-snippets/vat-compliance-redirects-recv.vcl
@@ -40,7 +40,7 @@
   client.geo.country_code == "NE"
 )
 {
-  if (req.url ~ "^\/[a-z]{2}/subscribe/digital" && req.url == "/subscribe/digital"   ){
+  if (req.url ~ "^\/[a-z]{2}/subscribe/digital" || req.url == "/subscribe/digital"   ){
        error 619;
     }
 }
--- a/support-frontend/conf/fastly-snippets/vat-compliance-redirects-recv.vcl
+++ b/support-frontend/conf/fastly-snippets/vat-compliance-redirects-recv.vcl
@@ -1,0 +1,46 @@
+# type: recv
+   if (
+  client.geo.country_code == "RS"||
+  client.geo.country_code == "EG"||
+  client.geo.country_code == "PK"||
+  client.geo.country_code == "MU"||
+  client.geo.country_code == "BH"||
+  client.geo.country_code == "MA"||
+  client.geo.country_code == "MC"||
+  client.geo.country_code == "OM"||
+  client.geo.country_code == "GE"||
+  client.geo.country_code == "NC"||
+  client.geo.country_code == "TZ"||
+  client.geo.country_code == "ZM"||
+  client.geo.country_code == "AL"||
+  client.geo.country_code == "BD"||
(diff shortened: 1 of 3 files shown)


jamesgorrie (approved):
Looks good!

support-frontend/conf/fastly-snippets/vat-compliance-redirects-recv.vcl
```diff
@@ -0,0 +1,46 @@
...
+  client.geo.country_code == "MD"||
+  client.geo.country_code == "DZ"||
+  client.geo.country_code == "TJ"||
+  client.geo.country_code == "LS"||
+  client.geo.country_code == "CG"||
+  client.geo.country_code == "TG"||
+  client.geo.country_code == "NE"
+)
+{
+  if (req.url ~ "^\/[a-z]{2}/subscribe/digital" && req.url == "/subscribe/digital"   ){
```
johnduffell, line 43: if `req.url == "/subscribe/digital"` then the first condition is redundant? unless `==` matches anywhere in the string?

johnduffell:
personally I would avoid putting logic/rules in fastly (unless it's performance critical) because it's harder to test and we don't always think to look in the fastly logs if something strange is happening.  But I'm not sure if there's another reason we need to do it this way?

support-frontend/conf/fastly-snippets/vat-compliance-redirects-recv.vcl
LAKSHMIRPILLAI: Thanks for spotting this, I have updated the logic

jamesgorrie (approved):
This looks good to go.

@johnduffell [we currently handle other situations in Fastly](https://github.com/guardian/support-frontend/blob/main/support-frontend/conf/fastly-snippets/compliance-redirect-recv.vcl).

If we want to change this we should do it universally. Currently we are a little split between redirection happening in [two](https://github.com/guardian/support-frontend/blob/main/support-frontend/conf/fastly-snippets/redirects-table-deliver.vcl) [places](https://github.com/guardian/support-frontend/blob/main/support-frontend/conf/routes#L25-L31) - we should probably unify that too.

I have been talking to @GHaberis about this - personally I like the split which simplifies the app some, but we do need a way to do this. It is on our health board as we need to test other Fastly logic too. This has been tested on `CODE`. 

Are you aware of any tests for the Scala redirection?

johnduffell:
Sorry that was a bit of a fire-and-leave comment and I didn't return!

> Are you aware of any tests for the Scala redirection?

I am not sure about automated tests, I actually meant test in the more general sense e.g. running it up locally, plus finding the logic (we recently had an issue with manage-frontend, and other than me the team had very little visibliity of fastly at all, let alone the logic in there - as we set ours up directly in fastly console)

The good thing is that you're putting it all in git so it's much more discoverable/reviewable that way, which is a big plus.

Overall it sounds like you've thought about it and are happy with the approach you're going for, so take my input in that vein!

PR #5476, opened 2023-11-27
Parity between compliance-redirect receive and deliver vcl

GHaberis:

I noticed the country list in `compliance-redirect-recv.vcl` was longer than `compliance-redirect-deliver.vcl`, it looks like a number of extra countries have been added to compliance-redirect-recv.vcl at some point, but not `compliance-redirect-deliver.vcl`. I've therefore updated `compliance-redirect-deliver.vcl` to include these countries.

Following @jamesgorrie feedback I've also consolidated the receive - error - deliver steps to receive - error:

**Custom Fastly Error: 620: Table redirect**
`redirects-table-recv.vcl` returns 620 "redirect" error if redirect table contains `req.url.path`
`redirects-table-error.vcl` checks if error code is 620 and sets `obj.status`, `obj.response`, and `req.http.Location` if `true`.

**Custom Fastly Error: 618: Compliance redirect**
`compliance-redirect-recv.vcl` returns 618 "redirect" error if request originates from listed geolocations.
`compliance-redirect-error.vcl` checks error code 618 and sets `obj.status`, `obj.response`, and `req.http.Location` if `true`.


jamesgorrie (approved):
Nice, @LAKSHMIRPILLAI and I noticed this yesterday. The joy of having it in code 🥳 

We have simplified how we're doing it for countries excluded for tax reasons but doing straight from `recv` -> `error` [as described in the Fastly docs](https://github.com/guardian/support-frontend/pull/5475), so we could probably do that here to (in another PR, or this one).

GHaberis:
@jamesgorrie I've applied the changes discussed to the compliance redirects and the table redirects, so this will need another look... 👀

jamesgorrie (approved):
Looks good!

On @johnduffell's point, I've left a comment [here](https://github.com/guardian/support-frontend/pull/5475#pullrequestreview-1757183691). In short, happy to discuss further with the outcome being having this logic in 1 place as opposed to the current two.

PR #5484, opened 2023-11-30
Digital edition designs

jamesgorrie:
A few updates on the digital subscriptions page

- Side panel copy updates
- Side panel image updates
- Removes the mobile heading
- Adds a "cancel anytime" note
- Removes patrons message
- Updates T&Cs 

[**Trello Card**](https://trello.com/c/M3C1dMsM/1699-editions-design-changes-to-the-checkout-change-copy-and-image)

## Why are you doing this?
In support of launching the digital subscriptions

## Is this an AB test?
- [x] No


iainjchambers-guardian (approved):
LGTM :+1:

support-frontend/assets/components/checkoutHeading/checkoutHeading.tsx
```diff
@@ -26,6 +26,14 @@ const headingContentContainer = css`
...
 `;
 
+const leftColImageEditions = css`
+	height: 129px;
+
+	img {
+		max-width: 100%;
+	}
+`;
+
```
jamesgorrie, line 36: Should we do this in [`digitalSubscriptionLandingPage`](https://github.com/guardian/support-frontend/pull/5484/files#diff-f37c28087b1494529802a18f41b7ad6c1bb6605c699e81dea0556fc3c269c656) similar to how we do in the [`/contribute` checkoutScaffold](https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/pages/supporter-plus-landing/twoStepPages/checkoutScaffold.tsx#L206-L214)?

The reason being:
- it decouples this styling specific to the digital subscriptions page so any other component that uses `checkoutHeading` doesn't inherit these styles
- it also means we won't have to remove the `<figure>` from [checkoutScaffold](https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/pages/supporter-plus-landing/twoStepPages/checkoutScaffold.tsx#L206-L214) as we're currently rendering `<figure><figure><!-- ... --></figure></figure>` on the `/contribute` page.

support-frontend/assets/pages/digital-subscriber-checkout/components/billingPeriodSelector.tsx
```diff
@@ -59,6 +59,17 @@ const subheading = css`
...
+const choiceCardContainer = css`
+	div {
+		${from.mobileLandscape} {
+			column-gap: ${space[1]}px;
+		}
+		${from.tablet} {
+			column-gap: ${space[2]}px;
+		}
+	}
+`;
```
jamesgorrie, line 71: It does feel as though this should be just part of what's used on the `/contribute` checkout, but probably worth drawing in where we try to abstract it.

support-frontend/assets/components/checkoutHeading/checkoutHeading.tsx
iainjchambers-guardian: Anything that removes unnecessary markup gets the ol' 👍 from me

paul-daniel-dempsey: Great spot @jamesgorrie , second point 2 is an important issue need to check for in future. Should of checked dependencies on `checkoutHeading` component when using local css otherwise `<figure>`  does indeed get duplicated here. Have moved it up to `digitalSubcriptionLandingPage` now as suggested. Thanks, done.

PR #5498, opened 2023-12-04
fix: remove noindex meta robots HTML tag

jamesgorrie:
Removes the `noindex` meta robots tag from digital subscriptions so that robots can find this and share it with their users.


paul-daniel-dempsey (commented):
Looks good.

PR #5499, opened 2023-12-04
fix: Add `default: true` to delete-snippets on deploy-fastly-snippets action

jamesgorrie:
Ensures that both inputs are required and default to true.

These default to true as that is the expected behaviour, with the option to opt-out for testing.


LAKSHMIRPILLAI (approved):

paul-daniel-dempsey (approved):
Yep make default explicitly true makes sense.

PR #5509, opened 2023-12-06
Install and configure circular dependency plugin

iainjchambers-guardian:
Installing and configuring the circular-dependency-plugin for webpack

[**Trello Card**](https://trello.com/c/hnt5M6GX/1719-fix-cyclical-file-dependencies)

## Why are you doing this?
This is a really helpful plugin that finds and highlights circular dependencies and outputs them in the terminal log. We have a larger task to remove those, and this is foundational to that challenge. 

## How to test

Run the dev server, you should see extra output in the terminal that highlights all circular dependencies

This should only affect the webpack dev server, it might slow build down a little but we can always remove it later when the problematic dependencies have been removed as ESLint will ensure that no new ones are created in future.

## Screenshots
image

### DIFF
--- a/support-frontend/webpack.common.js
+++ b/support-frontend/webpack.common.js
@@ -10,6 +10,7 @@ const { getClassName } = require('./scripts/css');
 const entryPoints = require('./webpack.entryPoints');
 const { BundleAnalyzerPlugin } = require('webpack-bundle-analyzer');
 const { WebpackManifestPlugin } = require('webpack-manifest-plugin');
+const CircularDependencyPlugin = require('circular-dependency-plugin');
 
 const cssLoaders = [
 	{
@@ -83,6 +84,10 @@ module.exports = (cssFilename, jsFilename, minimizeCss) => ({
 		}),
 		...(minimizeCss ? [new CssMinimizerPlugin()] : []),
 		new CleanUpStatsPlugin(),
+		new CircularDependencyPlugin({
+			// exclude detection of files based on a RegExp
+			exclude: /a\.js|node_modules/,
+		}),
 	],
 
 	context: path.resolve(__dirname, 'assets'),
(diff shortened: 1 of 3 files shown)


support-frontend/webpack.common.js
```diff
@@ -83,6 +84,10 @@ module.exports = (cssFilename, jsFilename, minimizeCss) => ({
 		}),
 		...(minimizeCss ? [new CssMinimizerPlugin()] : []),
 		new CleanUpStatsPlugin(),
+		new CircularDependencyPlugin({
+			// exclude detection of files based on a RegExp
+			exclude: /a\.js|node_modules/,
```
jamesgorrie, line 89: What's `a.js`?

jamesgorrie (commented):
Question - is this going to add a lot of noise to the local logging for engineers who are not specifically working on this problem?

This would be added for everyone, but not specifically useful to everyone unless the theory is it will prompt us to tidy it up quicker?

As a point - this is not allowed via [Guardian's eslint config](https://github.com/guardian/csnx/blob/main/libs/%40guardian/eslint-config/index.js#L78). Because of this we have had to disable it via `// eslint-disable-next-line import/no-cycle`. You can just search the codebase for this, remove that line and then fix that case individually.

support-frontend/webpack.common.js
iainjchambers-guardian: No idea, that came from the example in the docs so I didn't look too closely at it. I assumed until you pointed it out that it was filtering all .js files but now I see it isn't and I'm just as confused. I'll delete it, see what happens

iainjchambers-guardian:
> Question - is this going to add a lot of noise to the local logging for engineers who are not specifically working on this problem?
> 
> This would be added for everyone, but not specifically useful to everyone unless the theory is it will prompt us to tidy it up quicker?
> 
> As a point - this is not allowed via [Guardian's eslint config](https://github.com/guardian/csnx/blob/main/libs/%40guardian/eslint-config/index.js#L78). Because of this we have had to disable it via `// eslint-disable-next-line import/no-cycle`. You can just search the codebase for this, remove that line and then fix that case individually.

Depends on what you'd call a lot of noise, the screenshot I posted shows the total output which given the amount warnings running the devserver currently throws out, I'd say is pretty minimal.

Yes, you can find the files that have the ignored rules you mentioned but then you've got to figure out which one's are causing the individual circles from that list and installing this plugin did all of that work for me, visualised it all nicely and took as long to install as it would have taken me to isolate one of the circles so I considered it an easy win.

jamesgorrie (approved):
Great!

jamesgorrie:
> I'd say is pretty minimal.

I suppose I am just a little sensitive to it. I would prefer only warnings for things that I should be actively trying to solve, or I just ignore all of them (this is partially true now).

>  installing this plugin did all of that work for me, visualised it all nicely and took as long to install as it would have taken me to isolate one of the circles so I considered it an easy win.


Makes a tonne of sense - not to labour the point, but if it has a singular use we could have a flag that enabled for that person `yarn devrun --cyclical-deps` or something.

But by the time you've done the work I probably won't have seen any of the errors.

PR #5522, opened 2023-12-11
Big query publisher deadletter retention

rupertbates:
Currently the retention period on the BigQueryAcquisitionsPublisher dead letter queue is the default 4 days. This doesn't give much time to investigate and redrive any failures so this PR increases it to the maximum 14 days.


iainjchambers-guardian (approved):
LGTM 👍

PR #5527, opened 2023-12-11
Show dynamic nudge to all on one-time tab except those in VAT affected countries

GHaberis:

We should show the Dynamic nudge to all users in countries not affected by the VAT compliance issue, right now we're doing the opposite.


iainjchambers-guardian (approved):

paul-daniel-dempsey (approved):
Looks good

PR #5531, opened 2023-12-13
Fix isLive flag use for regional amounts tests

tomrf1:
Regional amounts test configurations are still used even if `isLive: false`. This is because there is always a regional amounts configuration, and the control should be used if the AB test is not live.
Currently, if `isLive: false`, the checkout page may still show the variants.
This PR fixes this by ensuring only the control is used in this case.


MaelGNM (approved):
LGTM!

tjmw (approved):

PR #5539, opened 2023-12-13
Editions : TsAndCs Manage My Account Url amended

paul-daniel-dempsey:

Digital Edition TsAndCs, change 'Manage My Account' from profile to manage. Profile won't send you through to manage unless you came from there.

|before|after|
|--------|---------|
|http://profile.theguardian.com/|http://manage.theguardian.com/|


andrewHEguardian (approved):
👍

PR #5543, opened 2023-12-14
Changes to Keep query parameters in URL when signing out for Newspaper

LAKSHMIRPILLAI:

Keep query parameters in URL when signing out

[**Trello Card**](https://trello.com/c/mFVI5m14/1692-support-frontend-keep-query-parameters-in-url-when-signing-out)

## Why are you doing this?

If the user is signed in on the paper checkout, they get a sign-out link below your email address. Clicking that link removes query parameters from the URL, which can change the product the user’s filling out a form for from Home Delivery or National Delivery to subscription card.
To fix this, the sign-out link should include the query parameters from the current page.

## How to test
Tested in CODE

### Newspaper checkout page before signing out

image

### Before Change- Newspaper checkout page after "Signing Out"
(Here the query parameters are removed from the url and the promotion applied to the page also changed along with the price and the customer selected changed to "Every day subscription card" instead of "Everyday paper" (before signing out)
image

### After Change - Newspaper checkout page after "Signing Out"
(Here the query parameters are retained)
image


support-frontend/assets/pages/paper-subscription-checkout/helpers/options.ts
```diff
@@ -14,24 +14,43 @@ import {
...
 	Everyday,
 } from 'helpers/productPrice/productOptions';
+import * as storage from 'helpers/storage/storage';
 import type { DateYMDString } from 'helpers/types/DateString';
 import { getQueryParameter } from 'helpers/urls/url';
 import { formatMachineDate } from 'helpers/utilities/dateConversions';
 import { getHomeDeliveryDays } from 'pages/paper-subscription-checkout/helpers/homeDeliveryDays';
 import { getVoucherDays } from 'pages/paper-subscription-checkout/helpers/voucherDeliveryDays';
 
 function getProductOption(): PaperProductOptions {
```
GHaberis, line 24: I don't think we need the changes to store the `product` in session storage anymore as the `redirectAfterSignOut` session storage technique negates the need to store `product`?

support-frontend/assets/pages/paper-subscription-checkout/helpers/options.ts
```diff
@@ -14,24 +14,43 @@ import {
...
 	return (
-		ActivePaperProductTypes.find((product) => product === productInUrl) ??
-		Everyday
+		ActivePaperProductTypes.find(
+			(product) => product === productInUrlOrSessionStorage,
+		) ?? Everyday
 	);
 }
 
 function getFulfilmentOption(): PaperFulfilmentOptions {
```
GHaberis, line 41: Same as above, I don't think we need the changes to store the `fulfilment` in session storage anymore as the `redirectAfterSignOut` session storage technique now negates the need to do this?

support-frontend/assets/helpers/user/user.ts
```diff
@@ -49,7 +50,9 @@ function getUserStateField(): string | undefined {
 const isPostDeployUser = (): boolean =>
 	cookie.get('_post_deploy_user') === 'true';
 
+export const REDIRECT_AFTER_SIGNOUT_STORAGE_KEY = 'redirectAfterSignOut';
```
GHaberis, line 53: Not entirely sure this module is the best place to have this logic yet, we want to run this as early as possible in the page's javascript, so we might need to find the entry point for the Newspaper Checkout landing page. 🤔

support-frontend/assets/pages/paper-subscription-checkout/helpers/options.ts
LAKSHMIRPILLAI: thanks for capturing this, I have removed this

support-frontend/assets/helpers/user/user.ts
GHaberis: @LAKSHMIRPILLAI the entry point for the Paper checkout is`support-frontend/assets/pages/paper-subscription-checkout/paperSubscriptionCheckout.tsx` I think this would be the most appropriate place for this logic.

jamesgorrie (approved):
This looks good. I it interesting that browsing to another URL, and then returning retain the session. That use-case [isn't well documented in the docs](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage) so it might be worth checking this across browsers if we haven't yet.

support-frontend/assets/helpers/user/user.ts
LAKSHMIRPILLAI: @GHaberis I tried to move REDIRECT_AFTER_SIGNOUT_STORAGE_KEY to paperSubscriptionCheckout.tsx , but it seems to raise some cyclical dependencies.So I think, it's best to keep it here for now

GHaberis:
@LAKSHMIRPILLAI the code looks good however I think we should hold on merging this until 2024 as there may be data privacy considerations we need to take in relation to session storage.

jamesgorrie (approved):

PR #5545, opened 2023-12-14
E2E paypal and browserstack fixes

rBangay:

This pr fixes the problem where browserstack was not reporting the status of the tests that were just run.

The paypal form filling in function has been changed to use id selectors instead of labels to address an issue where the paypal form was displaying in dutch, although this issue idsprobably also mitigated by selecting the geo location in the browserstack test config (also in this pr).

The paypal test user credentials have been updated so that these tests do not use a shared account that may be locked out if someone enters the wrong password too many times.

The browserstack automate runs are now labelled with the latest commit id and the date/time so that each run is unique and more discoverable.

## Screenshots
Screenshot 2023-12-14 at 16 59 09

### DIFF
--- a/support-e2e/browserstack.config.ts
+++ b/support-e2e/browserstack.config.ts
@@ -1,6 +1,7 @@
 import 'dotenv/config';
-import { exec } from 'child_process';
+import { execSync } from 'child_process';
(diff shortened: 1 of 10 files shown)


jamesgorrie (approved):
Fab!

PR #5559, opened 2023-12-18
bump scala compiler 2.13.12 version and fix fatal warnings

johnduffell:
Bumps scala compiler from `2.13.10` => `2.13.2`.


jamesgorrie (approved):
Woohoo!

PR #5560, opened 2023-12-18
Simplify `renderPage`

jamesgorrie:
The `renderPage` method currently takes a `elementId`, which is [the element we then replace with the client side content](https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/helpers/rendering/render.ts#L73).

We also have the `gu-render-to` classname used on error pages, and fallback to the body.

It would be good to standardise around 1 way, and the `gu-render-to` feels the best as it allows us to explicitly set where we want the render, whilst not having to pass IDs across from the scala code and client, which creates a coupling that is quite hard to work out especially when approaching the code with new eyes.

[**Trello Card**](https://trello.com/c/KnOaH4Bu/1739-remove-elementid-from-renderpage-method)

## Why are you doing this?
Reduce complexity of the code to make it easier to understand and change in future. Hopefully it is also a pattern that will make adding new pages easier.

## What about the scala references

[We still generate the `EmptyDiv`s with the ID](https://github.com/search?q=repo%3Aguardian%2Fsupport-frontend%20EmptyDiv&type=code), and this is still partially coupled to the `CSS`. If we removed the SASS that would allow us to remove the ID altogether. [See this example](https://github.com/guardian/support-frontend/blob/7c3fcea9947dc60a7654c68b532f00d31d7c4bdf/support-frontend/assets/pages/paypal-error/payPalError.scss#L15).

This is something we can look at as another PR, but I still feel like this decouples the JS rendering and Scala rendering which is a nice start.


support-frontend/app/controllers/Application.scala
```diff
@@ -220,7 +221,7 @@ class Application(
     Ok(
       views.html.main(
         title = "Guardian Supporters Map",
-        mainElement = assets.getSsrCacheContentsAsHtml("aus-moment-map", "aus-moment-map.html"),
+        mainElement = EmptyDiv("aus-moment-map"),
```
jamesgorrie, line 224: [This doesn't actually exist](https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/helpers/rendering/ssrPages.ts#L1-L9) and is another example of [having quite flexible fallbacks which make it quite hard to work out when something is or isn't doing](https://github.com/guardian/support-frontend/blob/7676de2bca6110564c9d86a4559a375ce5a3b10a/support-frontend/app/assets/AssetsResolver.scala#L62).

support-frontend/app/views/main.scala.html
```diff
@@ -126,18 +126,18 @@
...
-    </div>
-  }
+      <div
+        class="gu-render-to"
+        data-not-hydrated="true"
+        id="@ssrContent.value"
+      >
+        @ssrContent.content
+      </div>
+    }
```
jamesgorrie, line 140: This is just spacing, sorry!

GHaberis (approved):
Nice 👏

PR #5563, opened 2023-12-18
chore(deps): use java11 on all support workers and it-test-runner

jamesgorrie:
Part of: https://github.com/guardian/support-frontend/issues/5532

Upgrading all `support-lambdas` and `support-workers` to be on the same version of Java (Java11). There is an implicit coupling where `support-workers` calls `support-lambdas`. This should hopefully remove any conflict errors we're seeing because of that. 

Whilst trying to upgrade to `logback-classic@1.4.1` there are too many moving parts to try work out what is causing the issues. This should hopefully remove one of those variables.

### DIFF
--- a/.github/workflows/support-lambdas-build.yml
+++ b/.github/workflows/support-lambdas-build.yml
@@ -47,11 +47,11 @@ jobs:
           role-to-assume: ${{ secrets.GU_RIFF_RAFF_ROLE_ARN }}
           aws-region: eu-west-1
 
-      - name: Setup Java 8
+      - name: Setup Java 11
         uses: actions/setup-java@v3
         with:
-          java-version: "8"
-          distribution: "adopt"
+          java-version: "11"
+          distribution: "corretto"
       - uses: actions/cache@v3
         with:
           path: |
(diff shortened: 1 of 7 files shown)


LAKSHMIRPILLAI (approved):

johnduffell:
do these need changing too?
https://github.com/search?q=repo%3Aguardian%2Fsupport-frontend%20release%3A8&type=code

(Also not relevant here but watch out if you are updating the JRE in the AMI in case of any changes in parameters accepted by the JVM)

jamesgorrie:
> do these need changing too?

Yup - I've removed them as my understanding is it will just use the underlaying JVM and we don't need to backwards compile.

> (Also not relevant here but watch out if you are updating the JRE in the AMI in case of any changes in parameters accepted by the JVM)

Good shout.

@johnduffell a discussion for later, but I wonder in the monorepo land, are we really reaping the benefits by having so many different version of things, whereas a root `build.sbt`, especially with these system wide values, would be very useful. 

Do you know if that was the intention, or we'd prefer to keep the more granular configuration we currently have?

Issue #5571, opened 2023-12-20
Referrer set to `origin` when redirecting from `/contribute` => `/uk/contribute`


jamesgorrie:
When a reader:
- visits `https://www.theguardian.com/path/to/article`
- links through to `https://support.theguaridan.com/contribute`
- and is redirected to a geolocated `/uk/contribute`

The referrer sent from `/contribute` to `/uk/contribute` is the domain only i.e. `https://www.theguardian.com`. 

This is because [we use](https://github.com/guardian/support-frontend/blob/main/support-frontend/app/wiring/AppComponents.scala#L50) [the securityHeadersFilter](https://www.playframework.com/documentation/2.9.x/SecurityHeaders#Configuring-the-security-headers) which sets the `Referrer-Policy` header to `origin-when-cross-origin, strict-origin-when-cross-origin` for all requests in `support-frontend` including redirects.

While this _feels_ like it shouldn't make a difference as we are going from `www` -> `support` -> `support`, only the domain of the `cross-origin` `Referrer` header gets sent to the `/uk/contribute` as the referrer is carried over from the original request to the redirect, and considered `cross-origin`. Maybe this diagram can make it a little more clear

```mermaid
sequenceDiagram
    person->>www: https://www.theguardian.com/path/to/thing
    www->>person: Referrer-Policy: no-referrer-when-downgrade
    person->>support: https://support.theguardian.com/contribute
    Note over person,support: Referrer: https://www.theguardian.com/path/to/thing
    support->>person: Referrer-Policy: origin-when-cross-origin, strict-origin-when-cross-origin
    Note over person,support: 301 Location: https://support.theguardian.com/uk/contribute
    person->>support: https://support.theguardian.com/uk/contribute
    Note over person,support: Referrer: https://www.theguardian.com/
```

The integral step here is the last step. As the client has been told:
- The full referrer you have is `https://www.theguardian.com/path/to/thing`
- The URL you're being pointed to is `https://support.theguardian.com/uk/contribute` i.e. [cross-origin](https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy#definition_of_an_origin) (different host)
- The server has sent the `Referrer-Policy: origin-when-cross-origin, strict-origin-when-cross-origin`
- Next response, send only the domain for the referrer i.e. https://www.theguardian.com

[Trello card](https://trello.com/c/QJDeEMHs/1746-spike-referrer-header-altered-on-georedirects).

PR #5573, opened 2023-12-28
feat: Relax referrer-policy on redirects in support-frontend

jamesgorrie:
Relaxes the [Referrer-Policy](https://developer.mozilla.org/en-us/docs/web/http/headers/referrer-policy) to `no-referrer-when-downgrade` when we return redirects from the App.

## Why are you doing this?
This will allow the whole origin to be maintained when coming from `www.theguardian.com`, which will in turn allow us to analyse which articles or sections people are coming from when editors use geo-redirected routes e.g. `/contribute`.

The issue is described in more detail here https://github.com/guardian/support-frontend/issues/5571

[**Trello Card**](https://trello.com/c/QJDeEMHs/1746-spike-referrer-header-altered-on-georedirects)

## How to test
Follow from a site to a redirected URL, and see if the full `Referrer` header is maintained through to the redirected URL e.g. `/uk/contribute`.

I've currently `@Ignore`d the test as we have no precedence for testing filters and the new test is borking on logging with the error `java.lang.ClassCastException: class org.slf4j.helpers.NOPLoggerFactory cannot be cast to class ch.qos.logback.classic.LoggerContext (org.slf4j.helpers.NOPLoggerFactory and ch.qos.logback.classic.LoggerContext are in unnamed module of loader 'app')`. I am finding this hard to debug.

We can analyse with more granularity where traffic from `www.` => `support.` is coming from.


jamesgorrie:
I've removed the spec for now so we an start collecting this data.

jamesgorrie:
When visiting https://support.theguardian.com/contribute from [here](https://www.theguardian.com/environment/2023/nov/28/guardian-environment-pledge-2023) - the referrer is no `https://www.theguardian.com/environment/2023/nov/28/guardian-environment-pledge-2023`.

Screenshot 2024-01-04 at 10 18 14

mchv:
Nice one @jamesgorrie 💎

PR #5574, opened 2023-12-29
Update headline copy and ticker headline visibility

iainjchambers-guardian:
Updating the checkout headline copy displayed as part of the US end of year campaign and hiding the text above the ticker

[**Trello Card**](https://trello.com/c/hgJRjICX/1748-update-headline-copy-for-us-eoy-campaign)

## Why are you doing this?
With the year ended, the copy **Make a year-end gift to the Guardian** won't make sense, so it needs to be updated to something that will. Same goes for the text above the ticker that appears on mobile. 

## Screenshots
| Before | After |
| --- | --- |
| image image  | image image |

### DIFF
--- a/support-frontend/assets/components/ticker/ticker.tsx
+++ b/support-frontend/assets/components/ticker/ticker.tsx
@@ -87,7 +87,9 @@ export function Ticker(props: TickerProps): JSX.Element {
 
 	return (
 		<div>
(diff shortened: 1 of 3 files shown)


shtukas (approved):

iainjchambers-guardian:
No need to merge this as the goal has been met and the campaign switched off

PR #5580, opened 2024-01-02
Small refactor of abTest module interface

GHaberis:

I've been documenting our [client-side AB tests](https://github.com/guardian/support-frontend/wiki/Client-side-AB-Testing) set-up on the support site. In doing so I noticed a few exports that weren't actually imported anywhere, or were only imported to test, so I made the following small changes:

1) Removed the export `getSourceFromAcquisitionData`, this function is only referenced within the module.
2) Moved the `targetPageMatches` export to an `_` export which is naming convention to highlight it's only exported for testing purposes.
3) Grouped the remaining named exports `init` and `getAmountsTestVariant` at the end of the module. I find grouping the exports like this at the end of the module easier when trying to get a quick overview of a modules interface. 

Interested in feedback on points 2 and 3 as I don't believe we have agreed conventions on these approaches, but consistency through convention could be beneficial.

### DIFF
--- a/support-frontend/assets/helpers/abTests/abtest.ts
+++ b/support-frontend/assets/helpers/abTests/abtest.ts
@@ -79,7 +79,7 @@ export type Tests = Record<string, Test>;
 
 // ----- Init ----- //
 
-export function init(
+function init(
 	country: IsoCountry,
 	countryGroupId: CountryGroupId,
 	abTests: Tests = tests,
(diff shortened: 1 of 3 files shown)


support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -538,3 +541,10 @@ export function targetPageMatches(
 
 	return locationPath.match(targetPage) != null;
 }
+
+export { init, getAmountsTestVariant };
+
+// Exported for testing only
+export const _ = {
```
tomrf1, line 548: this is new to me! Are we already doing this elsewhere?

tomrf1 (approved):
looks good

PR #5581, opened 2024-01-03
VSCODE format .ts/.tsx on save

GHaberis:

I've noticed VSCODE hasn't been applying Prettier formatting on save, despite the workspace settings in .`vscode/settings.json` containng a rule `"source.fixAll.prettier": "explicit"`, the purpose of which was to format using our Prettier settings on save, having looked at it this doesn't do anything.

I've looked at various examples online and have landed on an updated configuration that now:

- Applies Prettier fixes to all `.ts` and `.tsx` files on save, I've tested and confirmed it's using our Prettier settings in `support-frontend/.prettierrc.js` when formatting on save.
- Applies ESLint updates to all `.ts` and `.tsx` files on save.
- Doesn't attempt to format non `.ts` and `.tsx` files on save.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@
 .metals
 metals.sbt
 .bloop
+.vscode/settings.json
 
 node_modules
 npm-debug.log
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,12 +1,20 @@
 {
   "typescript.format.enable": false, // disable default VS code TS formatting in favour of ESLint
-  "editor.formatOnSave": true,
-  "editor.defaultFormatter": "esbenp.prettier-vscode",
-  "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": "explicit"
+  "[typescript]": { // for .ts files
+    "editor.defaultFormatter": "esbenp.prettier-vscode",
+    "editor.formatOnSave": true,
+    "editor.codeActionsOnSave": {
+      "source.fixAll.eslint": true
+    },
+  },
+  "[typescriptreact]": { // for .tsx files
+    "editor.defaultFormatter": "esbenp.prettier-vscode",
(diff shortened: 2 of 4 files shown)


jamesgorrie (approved):
Is there a reason to not have this on `.js` files too?

Should we be following [our recommendations on .vscode](https://github.com/guardian/dotcom-rendering/blob/main/.vscode/settings.json.recommended) and have this file called `.vscode/settings.default.json` or similar? The main reason is that I have certain settings for this repo that are user/me specific and having this setup doesn't allow me to configure my editor my way + get these useful settings.

[`dotcom-rendering` uses a `required` and `recommended` config](https://github.com/guardian/dotcom-rendering/tree/main/.vscode), but I\'d argue that's overkill, but a good example.

GHaberis:
> Is there a reason to not have this on `.js` files too?
> 
> Should we be following [our recommendations on .vscode](https://github.com/guardian/dotcom-rendering/blob/main/.vscode/settings.json.recommended) and have this file called `.vscode/settings.default.json` or similar? The main reason is that I have certain settings for this repo that are user/me specific and having this setup doesn't allow me to configure my editor my way + get these useful settings.
> 
> [`dotcom-rendering` uses a `required` and `recommended` config](https://github.com/guardian/dotcom-rendering/tree/main/.vscode), but I'd argue that's overkill, but a good example.

I don't think there's a strong reason to not apply the formatting to `.js` files too. The formatting was introduced to keep the source code tidy, which I believe is all `.ts` and `.tsx` files, however we have various config defined in `.js` files for build, code quality and testing tools.

The idea of naming it `.vscode/settings.default.json` so devs can customise is good. I'll do this and add a note to the client-side set-up steps.

jamesgorrie (approved):

jamesgorrie (approved):

PR #5583, opened 2024-01-04
fix: remove symphonia logging and use ConsoleAppender

jamesgorrie:
Removes `symphonia` logging.

We're removing this as we don't use it and it is also blocking the upgrade to `logback@1.4.14`.


support-lambdas/it-test-runner/build.sbt
```diff
@@ -13,7 +13,6 @@ libraryDependencies ++= Seq(
   "com.fasterxml.jackson.core" % "jackson-databind" % jacksonDatabindVersion,
   "com.fasterxml.jackson.dataformat" % "jackson-dataformat-cbor" % jacksonVersion,
   "ch.qos.logback" % "logback-classic" % "1.2.12",
-  "io.symphonia" % "lambda-logging" % "1.0.3",
```
jamesgorrie, line 16: This wasn't actually used in this project.

tomrf1 (approved):

PR #5566, opened 2023-12-19
Take users directly to second page of checkout

rBangay:
## What are you doing in this PR?
setup ab test in order to direct some users directly to the second page of the checkout

## Screenshots
Screenshot 2023-12-19 at 11 00 55

PR #5587, opened 2024-01-05
Revert "Take users directly to second page of checkout"

GHaberis:
Reverts guardian/support-frontend#5566

This test has reached significance and the control has won, AV per impression was -12.9% lower in the variant compared to the control, this was significant.


paul-daniel-dempsey (approved):
Looks good, URL goes to first page now.

PR #5588, opened 2024-01-05
refactor: remove id from pagescaffold

jamesgorrie:
Removes the `id` from `PageScaffold` as it's unused.

## Why are you doing this?
Extra code = extra overhead when trying to refactor making things harder to change, and thus sometimes they don't get changed.


GHaberis (approved):
Nice 👍

PR #120, opened 2017-07-11
Cleaning

davidfurey:
## Why are you doing this?

We've added quite a lot of functionality to this project already, it felt like a good time to do some tidying

## Changes

* Moved lots of classes out of the very generic `lib` package
* Fewer implicits for wiring (less magic)
* Split wiring into separate files since it was getting a bit large

PR #5591, opened 2024-01-09
AppLoader renaming for consistency

jamesgorrie:
- Extend `BuiltInComponentsFromContext` from `AppLoader`. This brings us closer inline with [the official](https://www.playframework.com/documentation/2.9.x/ScalaCompileTimeDependencyInjection#Application-entry-point) docs which helps for future travellers, especially those unfamiliar with Play. This seems to be [a more common way (80 examples)](https://github.com/search?q=org%3Aguardian+%22class+AppComponents%22&type=code) vs [how it's done here (23 examples)](https://github.com/search?q=org%3Aguardian+%22trait+AppComponents%22&type=code) across the org.
- Rename `actinRefiners` => `actionBuilders`. This is just one last loose end from [this commit](https://github.com/guardian/support-frontend/pull/120/commits/8693a10c6ce0880937051285d5bde68ddcf51fe4).

## Why are you doing this?
A little but of consistency and clarity for future folks and myself when I have to revisit this code.

Both commits should be no-ops.

### DIFF
--- a/support-frontend/app/wiring/ActionBuilders.scala
+++ b/support-frontend/app/wiring/ActionBuilders.scala
@@ -7,7 +7,7 @@ import play.filters.HttpFiltersComponents
 trait ActionBuilders {
   self: Services with BuiltInComponentsFromContext with ApplicationConfiguration with HttpFiltersComponents =>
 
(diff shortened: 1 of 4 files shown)


tomrf1 (approved):
👍

PR #5596, opened 2024-01-10
Identity | Change logic of `GU_SO` cookie use

coldlink:
## Why are you doing this?

We discovered an issue in the way that the GU_SO cookie works, meaning that a user could sign out, then immediately sign in again as a new user, but the tokens for the old user had not been cleared yet, so the new user would see the old users data.

The problem and solution have been outlined in detail within this PR https://github.com/guardian/gateway/pull/2529

And the documentation was updated in this PR https://github.com/guardian/gateway/pull/2530

Essentially the fix is to read the `GU_SO` cookie and get the timestamp for when a user signs out, and compare it to the `iat` claim of the ID token. This leads to one of three scenarios:
1. If the `GU_SO` value is > `iat` then the user has signed out recently, so the tokens are invalid, get new tokens if needed
2. If the `GU_SO` value is < `iat` then the user has not signed out recently, so the tokens are valid, continue
3. If the `GU_SO` cookie is missing then the user has not signed out recently, so the tokens are valid, continue

This PR updates the logic in `UserFromAuthCookiesActionBuilder` to use this new logic regarding the `GU_SO` cookie to fix this issue. We've also documented in line as to why this is required.

### DIFF
--- a/support-frontend/app/actions/UserFromAuthCookiesActionBuilder.scala
+++ b/support-frontend/app/actions/UserFromAuthCookiesActionBuilder.scala
@@ -79,10 +79,9 @@ object UserFromAuthCookiesActionBuilder extends Logging {
       handleFailure: ValidationError => Future[Result],
   )(implicit ctx: ExecutionContext): Future[Result] = {
     def isCookiePresent(name: String) = request.cookies.get(name).isDefined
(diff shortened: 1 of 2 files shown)


kelvin-chappell (approved):
Looks good! 👍

kelvin-chappell (approved):
👍

PR #5600, opened 2024-01-12
Three tier router redirect

rBangay:
The href links to the checkout/landing pages of the checkout from banners and epics are going to `/contribute` ... for users (coming from banners of epics) that find themselves in the three tier ab test variant and are making a one off contribution we want to redirect them to the `/contribute/checkout` route so that they go straight to the 2nd page of the checkout. This should only apply to users that include the `selected-contribution-type=ONE_OFF` url param/value and for all other users the default choice in the clientside router for the checkout ([supporterPlusRouter.tsx](https://github.com/guardian/support-frontend/compare/main...three-tier-router-redirect#diff-fd63b7bdddd1f8c8a686a6ba94bee4323ebd6d88ea919d35cdcf7dbecd9a6db5)) should take precedence.

## Example
`https://support.theguardian.com/uk/contribute?selected-contribution-type=ONE_OFF&selected-amount=4#ab-threeTierCheckout=variant`

This url should redirect (on the client) to the /contribute/checkout route (keeping all the url params and hash)


paul-daniel-dempsey (approved):
Looks great, tested the url route taken from banner choicecards for one off (in three tier abest) and lands on 2nd page of checkout with params and hash details intact.

support-frontend/assets/pages/supporter-plus-landing/supporterPlusRouter.tsx
```diff
@@ -59,6 +65,22 @@ const router = () => {
 		<SupporterPlusInitialLandingPage thankYouRoute={thankYouRoute} />
 	);
 
+	const isThreeTierAndOneOffUrlParam = () => {
```
jamesgorrie, line 68: Could this just be a value rather than method?

support-frontend/assets/pages/supporter-plus-landing/supporterPlusRouter.tsx
```diff
@@ -67,12 +89,17 @@ const router = () => {
...
 							<Route
-								path={`/${countryId}/contribute/`}
-								element={firstStepLandingPage}
-							/>
-							<Route
-								path={`/${countryId}/contribute/:campaignCode`}
-								element={firstStepLandingPage}
+								path={`/${countryId}/contribute/:campaignCode?`}
+								element={
+									isThreeTierAndOneOffUrlParam() ? (
```
jamesgorrie, line 94: I know this is all part of the test, but a comment with the context the PR gives that we're doing this to avoid changing the `www` banners and go straight to checkout might help later.

jamesgorrie (approved):

support-frontend/assets/pages/supporter-plus-landing/supporterPlusRouter.tsx
rBangay: yep 👍

PR #5601, opened 2024-01-12
redirect digitaledition to contribute for VAT affected countries

GHaberis:

We have a redirect to send users affected by VAT compliance issues from `/subscribe/digital` to `/contribute`, however `/subscribe/digital` has been updated to `/subscribe/digitaledition`, so the redirect needs updating.

The list of countries affected by the VAT issue also needed updating, with many more countries added. We have multiple hardcoded references to the list at the moment, it may be worth pulling this from a central location at some point.


GHaberis:
@jamesgorrie and I had a chat about the maintentance overhead of having this list of countries in multiple places and have a proposed improvement to use the list in RRCP as the default going forward, this will be in a follow-up PR.

jamesgorrie (approved):

PR #5609, opened 2024-01-16
fix: run threeTierCheckout test on contribution or  weekly pages with relevant URL param

jamesgorrie:
# What does this do?

Updates the `canRun` function of the `threeTierCheckout` to run on contribution pages and the weekly checkout with the `isThirdTier=true` URL parameter.

# Why are you doing this?

Adding the new `isThirdTier=true` URL param allows us to

- Let the Weekly checkout know it should include the Support+ subscription
- Include a person in the test with this parameter
- Gives us a signal to pass the upstream `/create` endpoint & step functions that it should include the Subscriber+ contribution


paul-daniel-dempsey (approved):
Makes sense to me, great usage of canRun with `isThirdTier` incorporated for GW checkout.

PR #5612, opened 2024-01-17
fix: Handle error where salesforce API response status is 200 but Success is false

jamesgorrie:
The Salesforce API returns a http status of 200 even if the request fails, but with the `Success` property on the response set to `false`.

e.g.
```JSON
{
    "Success": false,
    "ErrorString": "Failed Upsert of new Contact: Upsert failed. First exception on row 0; first error: CANNOT_INSERT_UPDATE_ACTIVATE_ENTITY, ContactTrigger: execution of AfterInsert\n\ncaused by: line 78, column 9: Dependent class is invalid and needs recompilation:\n Class testUtilities : Dependent class is invalid and needs recompilation:\n Class KnowledgePublishToHelpCentreService : DML operation Update not allowed on Knowledge__kav: []",
    "ContactRecord": null
}
```

This fix handles that case and returns the `SalesforceErrorResponse` to the lambda / client to ensure it's handled correctly i.e. retried.

[**Trello Card**](https://trello.com/c/hcVE76TJ/453-fix-salesforce-retries-in-support-workers-impacts-recurring-contribution)

## How to test

There is an integration test using invalid user data to ensure the right error is thrown. I haven't quite thought of a way to test against the use case we are hoping to have covered by this i.e. a Salesforce outage.

We don't lose any contact information when there is a Salesforce API outage.

The testing isn't against the actual usecase.

### DIFF
--- a/support-models/src/main/scala/com/gu/salesforce/Salesforce.scala
+++ b/support-models/src/main/scala/com/gu/salesforce/Salesforce.scala
@@ -126,36 +126,35 @@ object Salesforce {
       MailingCountry: Option[String],
   ) extends UpsertData
 
-  trait SalesforceResponse {
-    val Success: Boolean
-  }
-
+  case class SalesforceContactResponse(
+      Success: Boolean,
+      ErrorString: Option[String],
+      ContactRecord: Option[SalesforceContactRecord],
+  )
   object SalesforceContactResponse {
     implicit val codec: Codec[SalesforceContactResponse] = deriveCodec
   }
 
-  object SalesforceContactRecords {
-    implicit val codec: Codec[SalesforceContactRecords] = deriveCodec
-  }
-
-  case class SalesforceContactResponse(
+  case class SalesforceContactSuccess(
       Success: Boolean,
       ErrorString: Option[String],
       ContactRecord: SalesforceContactRecord,
-  ) extends SalesforceResponse
-
-  case class SalesforceContactError(
-      Success: Boolean,
-      ErrorString: Option[String],
   )
+
+  object SalesforceContactSuccess {
+    implicit val codec: Codec[SalesforceContactSuccess] = deriveCodec
+  }
 
   case class SalesforceContactRecords(buyer: SalesforceContactRecord, giftRecipient: Option[SalesforceContactRecord]) {
     def recipient: SalesforceContactRecord = giftRecipient.getOrElse(buyer)
   }
+  object SalesforceContactRecords {
+    implicit val codec: Codec[SalesforceContactRecords] = deriveCodec
+  }
 
   case class SalesforceContactRecordsResponse(
-      buyer: SalesforceContactResponse,
-      giftRecipient: Option[SalesforceContactResponse],
+      buyer: SalesforceContactSuccess,
+      giftRecipient: Option[SalesforceContactSuccess],
   ) {
     def successful: Boolean = buyer.Success && giftRecipient.forall(_.Success)
 
--- a/support-modules/rest/src/main/scala/com/gu/rest/WebServiceHelper.scala
+++ b/support-modules/rest/src/main/scala/com/gu/rest/WebServiceHelper.scala
@@ -75,6 +75,7 @@ trait WebServiceHelper[Error <: Throwable] {
       rb: Request.Builder,
   )(implicit decoder: Decoder[A], errorDecoder: Decoder[Error], ctag: ClassTag[A]): Future[A] = {
     for {
+
       response <- getResponse(rb)
       codeBody <- Future.fromTry(getJsonBody(response))
       decodedResponse <- Future.fromTry(decodeBody[A](codeBody))
@@ -82,11 +83,12 @@ trait WebServiceHelper[Error <: Throwable] {
 
   }
 
-  protected def getResponse(rb: Request.Builder) = for {
-    req <- wsPreExecute(rb).map(_.build())
-    _ = SafeLogger.info(s"Issuing request ${req.method} ${req.url}")
-    response <- httpClient(req)
-  } yield response
+  protected def getResponse(rb: Request.Builder) =
+    for {
+      req <- wsPreExecute(rb).map(_.build())
+      _ = SafeLogger.info(s"Issuing request ${req.method} ${req.url}")
+      response <- httpClient(req)
+    } yield response
 
   def getJsonBody[A](response: Response)(implicit ctag: ClassTag[A]): Try[CodeBody] = {
     val code = response.code().toString
@@ -115,7 +117,7 @@ trait WebServiceHelper[Error <: Throwable] {
     responseFamily match {
       case "2xx" =>
         decode[A](responseBody).left.map { err =>
-          decodeError(responseBody).right.getOrElse(
+          decodeError(responseBody).getOrElse(
             WebServiceHelperError[A](
               codeBody,
               s"failed to parse response. Error was: $err, Response was: $responseBody",
@@ -124,7 +126,7 @@ trait WebServiceHelper[Error <: Throwable] {
           )
         }.toTry
       case "4xx" =>
-        Failure(decodeError(responseBody).right.toOption.getOrElse(WebServiceClientError(codeBody)))
+        Failure(decodeError(responseBody).toOption.getOrElse(WebServiceClientError(codeBody)))
       case statusCode =>
         Failure(WebServiceHelperError[A](codeBody, s"unrecognised response code $statusCode"))
     }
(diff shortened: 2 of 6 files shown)


AndreaDiotallevi:
I think it's a good solution!

A couple of questions:

- What is the difference between `SalesforceContactResponse` and `SalesforceContactSuccess`? Do we need both.
- Regarding the potential test for the actual use-case of Salesforce being unavailable, can you mock the Salesforce service response?

rupertbates:
A couple of things:
- This looks good as a solution for the decoding error, we could ask @graham228221 to help us with testing as he will be able to get Salesforce in CODE to return the error we are expecting
- We need to check that the error which is returned results in the step function retrying the failed step. If the error is this one: https://github.com/guardian/support-frontend/blob/7437fa11c1a33a2c6fe6634587da7344fa577dd5/support-models/src/main/scala/com/gu/salesforce/Salesforce.scala#L171 then it will, otherwise we'll need to add in details of the new error to this object

graham228221:
> * We need to check that the error which is returned results in the step function retrying the failed step. If the error is this one: https://github.com/guardian/support-frontend/blob/7437fa11c1a33a2c6fe6634587da7344fa577dd5/support-models/src/main/scala/com/gu/salesforce/Salesforce.scala#L171
>    then it will, otherwise we'll need to add in details of the new error to this object

To clarify, the expected responses during the maintenance window on Sunday I'm expecting to look some like this:

```
{
    "Success": false,
    "ErrorString": "Failed Upsert of new Contact: Upsert failed. First exception on row 0; first error: INSERT_UPDATE_DELETE_NOT_ALLOWED_DURING_MAINTENANCE, Updates can’t be made during maintenance. Try again when maintenance is complete: []",
    "ContactRecord": null
}
```

Not 100% sure what the exact `errorString` will be as we can't actually put SF into maintenance mode to test this! But it should contain `INSERT_UPDATE_DELETE_NOT_ALLOWED_DURING_MAINTENANCE`.

Interesting that previously we assumed the whole request to Salesforce would fail: https://github.com/guardian/support-frontend/pull/2332 - it's possible, it depends on whether this error can be caught or not on the Salesforce side. If it can't be caught, then a non-200 response would be provided from SF and it looks like this PR already handles this.

Based on this [documented example](https://developer.salesforce.com/docs/atlas.en-us.apexref.meta/apexref/apex_methods_system_test.htm#apex_System_Test_setReadOnlyApplicationMode) of simulating read only mode in Salesforce tests, I am _fairly_ confident that this error can be caught so Salesforce would be returning a 200 response (with `success = false`).

> * This looks good as a solution for the decoding error, we could ask @graham228221 to help us with testing as he will be able to get Salesforce in CODE to return the error we are expecting

I can get Salesforce to return the example response I provided above, if that would be helpful! But again, I can't actually put Salesforce into maintenance mode so the `errorString` may not be 100% representative.

graham228221:
> Based on this [documented example](https://developer.salesforce.com/docs/atlas.en-us.apexref.meta/apexref/apex_methods_system_test.htm#apex_System_Test_setReadOnlyApplicationMode) of simulating read only mode in Salesforce tests, I am _fairly_ confident that this error can be caught so Salesforce would be returning a 200 response (with `success = false`).

To clarify further:

> [!NOTE]
> In Salesforce, the endpoint called from Support-Workers [has a try-catch](https://github.com/guardian/salesforce/blob/master/force-app/main/default/classes/REST_RegisterCustomer.cls#L23) that tries to ensure that we always return a 200 response. 
> Certain errors in Salesforce cannot be caught and will not return a 200 response, `INVALID_SESSION_ID` and `REQUEST_LIMIT_EXCEEDED` defined [here](https://github.com/guardian/support-frontend/blob/fa53376561adf9ec7abf3d080bbecba26fa6ef3a/support-models/src/main/scala/com/gu/salesforce/Salesforce.scala#L167) are obvious examples.

I _think_ that `INSERT_UPDATE_DELETE_NOT_ALLOWED_DURING_MAINTENANCE` can be caught, so a 200 response will be returned.

jamesgorrie:
@AndreaDiotallevi 

> What is the difference between `SalesforceContactResponse` and `SalesforceContactSuccess`? Do we need both.

`SalesforceContactResponse` is what we get from the API - which can have both `ErrorString` or `ContactRecord` optional. `SalesforceContactSuccess` is what's parsed from that if we know it's a success by checking `Success` is `true` and `ContactRecord` exists.

I were going to be more descriptive I might have:
```scala
case class SalesforceContactResponse(
  Success: Boolean,
  ErrorString: Option[String],
  ContactRecord: Option[SalesforceContactRecord],
)
case class SalesforceContactSuccess(
  ContactRecord: Option[SalesforceContactRecord],
)
case class SalesforceContactFailure(
  ErrorString: String,
)
```

That - or we work on the API to return a descriptive HTTP response. @graham228221 is this something we could look at in future?

But this feels out of remit for this PR - I have added it to [an issue](https://github.com/guardian/support-frontend/issues/5615) to discuss post.

> Regarding the potential test for the actual use-case of Salesforce being unavailable, can you mock the Salesforce service response?

Definitely - @graham228221 has given some really valuable information here that will help with that.

jamesgorrie:
Thanks @rupertbates and @graham228221 - I didn't know we had created that API - I was searching for official Salesforce docs for ages 🤦.

I'll investigate further to ensure we retry on this case.

jamesgorrie:
@graham228221 great! I have added a unit test to check my logic on the way that we parse a response from Salesforce.

Not the best code I have written, but the way we have compose `SalesforceService`, our generic `WebServiceHelper` and `AuthService` (Salesforce) makes it very hard to test. 

I'll add an integration test now too!

jamesgorrie:
@rupertbates @AndreaDiotallevi @graham228221 - another set of eyes would be useful. Mainly I have
- added the logic to return the [right `errorCode`](https://github.com/guardian/support-frontend/blob/main/support-models/src/main/scala/com/gu/salesforce/Salesforce.scala#L171) to [force](https://github.com/guardian/support-frontend/blob/main/support-models/src/main/scala/com/gu/salesforce/Salesforce.scala#L185-L189) [retries](https://github.com/guardian/support-frontend/blob/main/support-workers/src/main/scala/com/gu/support/workers/exceptions/ErrorHandler.scala#L34)
- Added a unit test to test  and describe the parsing
- Added an integration test with logic to force a maintenance error

@graham228221 - the last piece, is that something that will live in `CODE` for perpetuity or just for a bit?

jamesgorrie:
I have tried to [catch some of the tech debt here](https://github.com/guardian/support-frontend/issues/5615) - after a chat with @AndreaDiotallevi and @graham228221 we thought it might be best to try and shift some of this logic upsteam to the API

graham228221:
> * Added an integration test with logic to force a maintenance error
> 
> @graham228221 - the last piece, is that something that will live in `CODE` for perpetuity or just for a bit?

That's a good question! I imagined it only being in CODE for a short while, but if it's in an integration test that's not ideal and we should productionise it somehow so the change is on our main branch.

Perhaps we could revisit this next week, and decide how we want to move forward? I'd much prefer to address https://github.com/guardian/support-frontend/issues/5615 instead of making this hacky mock error permanent!

rupertbates (approved):
Looks good to me

jamesgorrie:
@graham228221 

> I imagined it only being in CODE for a short while

[Consider it removed](https://github.com/guardian/support-frontend/pull/5612/commits/a675423e90c93c2900da31daee3f2b377132b96f), it was super useful as to be able to test though, thanks!

AndreaDiotallevi (approved):

graham228221:
I've been reviewing the errors we got during Sunday's maintenance window, and thought I'd record them here for future reference. 

The behaviour was not as described in the Salesforce docs: https://help.salesforce.com/s/articleView?id=000386834&type=1 we were expecting integrations to be able to authenticate, but fail when trying to mutate any data.

What we actually saw was simple "invalid password" authentication failures at the start of the maintenance window from 3am (the response would have returned `invalid_grant`). This occurred for 30 minutes or so, and by 3.38am support-workers was able to authenticate AND add data to SF. We saw this pattern across integrations, for example for OwnBackup which had a scheduled SF backup during the window and the single-contributions-salesforce-writes Lambda which is consistently adding data to SF. This downtime may have been extended slightly as many integrations had enough "invalid password" attempts to trigger a 30 minute password lockout on their account.

For support-workers, the authentication failure fell into the `RetryLimited` pattern with a maximum of 5 attempts. Fortunately this was enough for each support-workers run that didn't succeed first time so there was no financial/customer impact (apart from confirmation emails being a little delayed).

It would have been interesting to see whether the behaviour was as expected in the UI, maybe next time. Org migrations typically take place every 3-5 years.

PR #5343, opened 2023-10-13
fix: transpile @guardian/source-foundations for Safari < 14 support

jamesgorrie:
## What are you doing in this PR?
Redux of https://github.com/guardian/support-frontend/pull/5339.

Transpiles `@guardian/source-foundations` `node_module` to support in Safari < v14.

The previous PR introduced a bug which seems to occur when you include `@guardian/source-react-components` to the transpilation (validated locally).

I would like to further investigate:
* Why does that transpilation create that bug?
* [Why didn't Chromatic pick up on the changes](https://www.chromatic.com/build?appId=62e115310aef0868687b2322&number=4215)?

But thought a fix first would be good.

[**Trello Card**](https://trello.com/c/PUYJmq9U/1607-safari-13-js-breaks)

## screenshots

**Chrome - modern OSX** 
Screenshot 2023-10-13 at 15 22 50

**Safari 13.1**
Screenshot 2023-10-13 at 15 26 17

PR #5614, opened 2024-01-17
Extend `@guardian/tsconfig`

jamesgorrie:

- Extends [@guardian/tsconfig](https://github.com/guardian/csnx/tree/main/libs/@guardian/tsconfig)
  - [adds `target: "esnext"`](https://github.com/guardian/csnx/blob/main/libs/%40guardian/tsconfig/tsconfig.json#L13)
  - in turn enables [`downlevelIteration`](https://www.typescriptlang.org/tsconfig#downlevelIteration)
  - which is needed in Source - [specifically for column widths](https://github.com/guardian/csnx/blob/main/libs/%40guardian/source-react-components/src/columns/styles.ts#L131-L140)
- Overrides `noImplicitReturns` and `noUncheckedIndexedAccess` to be fixed later as they don't cause any user facing errors, but introduce a list of 225 errors 
- Removes repeated `compilerOptions` from our local `tsconfig` (they were the same) 
- Adds `@guardian-source*` modules to our compilation pipeline for these changes to take effect

## Why are you doing this?

- To support JS in < Safari 13 we [added all source packages to our compilation pipeline](https://github.com/guardian/support-frontend/pull/5339/files#diff-0d87b2c785c3642f9806dce67dc10f123352423c9eea722ffb9df5703169a5a3R120)
- This fixed the JS but broke our `Column` layouts so [we reverted](https://github.com/guardian/support-frontend/pull/5342)
- [We got `@guardian/source-foundations` supported](https://github.com/guardian/support-frontend/pull/5343) but we were left with the question "why does compiling guardian packages create a layout bug"?

We have now answered that question. We needed to extend Source's tsconfig, which we have done. I might think about where to expose this information for future travellers. 

Bonus points 🎁 "Size Change: -28.7 kB (-2%)"

Ping @guardian/client-side-infra for thoughts

## How to test

Browserstack

- Fewer issues raised by readers
- Potentially a tiny increase in traffic
- If we're luck, people using lower spec devices will give us loads of 💰 
- **OR** - people with spec devices have more access to Guardian content

🤷 

## Screenshots

**iPad Pro 12.9 2018 v12.1**
Screenshot 2024-01-17 at 19 15 54

**Fancy high spec mac**
Screenshot 2024-01-17 at 19 18 35


rBangay (approved):
Nice work 👍

jamesgorrie:
In the wild 
Screenshot 2024-01-18 at 11 35 38

mchv:
Great work @jamesgorrie 💎

PR #5625, opened 2024-01-22
chore(deps): upgrade node version from 18 => 20

jamesgorrie:
Upgrades Node to the latest LTS version (20). 18 is currently in maintenance.

## Why are you doing this?
I was hoping [to update the `setup-java` action](https://github.com/guardian/support-frontend/pull/5494) - but realised as part of that release they had bumped to node 20. 

This is to ensure we have parity across dev / build / prod.

## How to test

- [x] Run CDK locally
  We had to upgrade CDK, and thus `@guardian/cdk` to have [full Node 20 support](https://github.com/aws/aws-lambda-base-images/issues/91#issuecomment-1737694321)
- [x] Run `support-frontend` node app locally


GHaberis (approved):
Looks good 👍

PR #5627, opened 2024-01-22
Switch PreparePaymentMethodForReuse lambda to use standard getAccount

rupertbates:
The `PreparePaymentMethodForReuse lambda was fetching account information using a Zuora api call which required the guid id of the account rather than the more usual account number. This was a problem for the three tier work currently taking place as we do not have this id readily available.

This PR changes the lambda to use the more commonly used api call which accepts either an account id or an account number and also adds a few fields to the response of that call which were not being mapped previously.


jamesgorrie (approved):
Lovely!

PR #5631, opened 2024-01-23
chore(deps): remove outdated resolutions from package.json

jamesgorrie:
Removes the `resolutions` from `package.json` as they're not needed any more. From the `yarn.lock` - they might actually be holding some packages back.

I assumed this by looking at the git blame and saw most were added yonks ago.

## Why are you doing this?

Health and hopefully it'll avoid any upgrade hell in the future.

## Screenshots

**Snyk**
Screenshot 2024-01-23 at 13 54 11

**Git Blame**
Screenshot 2024-01-23 at 13 55 27

### DIFF
--- a/support-frontend/package.json
+++ b/support-frontend/package.json
@@ -214,17 +214,6 @@
     "webpack-stats-plugin": "^0.3.2",
     "yargs": "^17.7.2"
   },
-  "resolutions": {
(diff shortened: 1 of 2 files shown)


paul-daniel-dempsey (approved):
Nice :)

PR #5635, opened 2024-01-24
feat: add promoCode to SupporterPlusState

jamesgorrie:
- Adds `promoCode` `promoCode` to `PreparePaymentMethodForReuse` for Supporter+ bolt-on which then propagates through the system.
- Tightens the check to see if a payment is a duplicate by comparing the rate plans as well as the `requestId` as we were assuming the Supporter+ bolt on was a duplicate thus not being created.
- Adds a test for `subscribeIfApplicable`

The logic was quite hard to fathom in writing the tests for `ZuoraSubscriptionCreator`, which made me feel it could use with a little tidy, but that might be for another PR.

[**Trello Card**](https://trello.com/c/zFv8FO72/540-add-three-tier-promotion-codes)

### DIFF
--- a/support-workers/cloud-formation/src/templates/state-machine.yaml
+++ b/support-workers/cloud-formation/src/templates/state-machine.yaml
@@ -88,10 +88,10 @@ States:
     Choices:
       - Next: ThreeTierUnpackSendThankYouEmailState
         And:
-#          - Variable: $.state.acquisitionData.threeTierCreateSupporterPlusSubscription
-#            IsPresent: true
-#          - Variable: $.state.acquisitionData.threeTierCreateSupporterPlusSubscription
-#            BooleanEquals: true
+          - Variable: $.state.acquisitionData.threeTierCreateSupporterPlusSubscription
+            IsPresent: true
+          - Variable: $.state.acquisitionData.threeTierCreateSupporterPlusSubscription
+            BooleanEquals: true
           - Variable: $.state.sendThankYouEmailState.product.productType
             IsPresent: true
           - Variable: $.state.sendThankYouEmailState.product.productType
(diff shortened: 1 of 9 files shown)


support-workers/src/main/scala/com/gu/support/workers/lambdas/PreparePaymentMethodForReuse.scala
```diff
@@ -98,15 +98,15 @@ class PreparePaymentMethodForReuse(servicesProvider: ServiceProvider = ServicePr
...
-        None,
-        None,
-        None,
+        productSpecificState = productState,
+        requestId = state.requestId,
+        user = state.user,
+        product = productType,
+        analyticsInfo = state.analyticsInfo,
+        firstDeliveryDate = None,
+        promoCode = state.promoCode,
```
jamesgorrie, line 107: This is the added line - it just became hard to read without the named parameters.

support-workers/cloud-formation/src/templates/state-machine.yaml
```diff
@@ -142,18 +142,32 @@ States:
...
+  {{> supporterPlusProduct currency="AUD" amount=160 billingPeriod="Annual" promoCode="3TIER_SUPPORTERPLUS_AU_ANNUAL" }}
+
+  ThreeTierPreparePaymentMethodForReuse:
+    Type: Task
+    Resource: "${PreparePaymentMethodForReuseLambda.Arn}"
+    Next: ThreeTierCreateZuoraSubscription
+    {{> retry}}
+    {{> catch}}
+
+  ThreeTierCreateZuoraSubscription:
```
jamesgorrie, line 165: I've split these out as separate steps, although they use the same lambdas. This makes it easier for debugging via the console rather than cycling back round which can get confusing.

rupertbates (approved):
Looks good 👍

PR #5636, opened 2024-01-24
3-Tier - First Page, Benefits Tooltip

paul-daniel-dempsey:

Add tooltip to the 3-tier card benefits list as shown in screenshot at desktop breakpoints or above

The tiers benefits to apply tooltips to are the following:
tier 2: `Full access to the Guardian News app`
tier 3: `Guardian Weekly magazine delivered to your door every week`

Checkout can be viewed at the following url behind (currently disabled) ab-Test named `threeTierCheckout` ->
https://support.thegulocal.com/uk/contribute#ab-threeTierCheckout=variant

|Desktop+|
|--------|
|image|

[**Trello Card**](https://trello.com/c/35rbMrFV/446-first-page-checkout-benefits-tooltip)

## How to test & measure success?

[**Figma**](https://www.figma.com/file/xqUMHA7rP4XDtxsOHcZy3U/testing-tiers?type=design&node-id=858%3A70938&mode=dev)


support-frontend/assets/components/checkmarkList/checkmarkList.tsx
```diff
@@ -101,7 +113,24 @@ export function CheckmarkList({
...
+					<td css={[checkListTextCss, item.maybeGreyedOut]}>
+						{typeof item.text === 'string' ? (
+							<span>
+								{item.text}
+								{item.toolTip && (
+									<Tooltip
+										cssOverrides={toolTipCss}
+										children={<p>{item.toolTip}</p>}
+										xAxisOffset={108}
+										yAxisOffset={12}
```
jamesgorrie, line 125: I can't fully see why these aren't calculated by the tooltip, any why they are these specific values?

Looking at this a little closer - it looks like this offset if broken anyway:
Screenshot 2024-01-25 at 09 33 58

It also looks like the only place we use it is [on the "Cancel anytime" prompt](https://github.com/guardian/support-frontend/blob/b115bdc88c53c40829a6cf9bac1e144f0b217377/support-frontend/assets/pages/supporter-plus-landing/formSections/amountAndBenefits.tsx#L130) - and [we never actually turn it on](https://github.com/guardian/support-frontend/blob/b115bdc88c53c40829a6cf9bac1e144f0b217377/support-frontend/assets/pages/supporter-plus-landing/twoStepPages/firstStepLanding.tsx#L162-L169).

As these values work, this is good.

Outside of this PR we might want to look at how we can always have the tooltip be next to the ℹ️ icon.

support-frontend/assets/components/tooltip/Tooltip.tsx
```diff
@@ -164,28 +166,36 @@ const arrowTop = css`
 `;
 
 export type TooltipProps = {
-	promptText: string;
+	promptText?: string;
 	buttonLabel?: string;
 	children: React.ReactNode;
+	xAxisOffset?: number;
+	yAxisOffset?: number;
+	placement?: Placement;
```
jamesgorrie, line 174: Given we don't use this anywhere else, we could avoid adding this extra complexity and just set the default to `'bottom'`.

jamesgorrie (approved):

jamesgorrie:
> +13.6 kB

That tooltip sure is heavy.

support-frontend/assets/components/checkmarkList/checkmarkList.tsx
```diff
@@ -27,6 +28,16 @@ const checkListTextCss = css`
...
 
+const toolTipCss = css`
+	display: none;
+
+	${from.desktop} {
+		display: inline;
+		margin-left: ${space[1]}px;
+		vertical-align: top;
+	}
+`;
```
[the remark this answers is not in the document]
paul-daniel-dempsey: Yep,  removed cssOverride from `Tooltip`

support-frontend/assets/components/tooltip/Tooltip.tsx
paul-daniel-dempsey: `Tooltip` is referenced in the `amountsAndBenefits` and in a story. The former could be enabled onto the `firstStepLanding` by the prop `showCancelTooltip`.  To preserve this compatibility I would like to keep these parameters untie we decide to create a PR to remove this feature from the `firstStepLanding` page (then they could be defaulted then).

support-frontend/assets/components/checkmarkList/checkmarkList.tsx
paul-daniel-dempsey: Yep, agree its not perfect, here it works with the `xAxisOffset` but would be good for the  the floating container, rather than centering horizontally, was more intelligent to move over the ℹ️ icon. Vertically (`yAxisOffset`) we could default allowing no adjustment. Created PR here -> https://trello.com/c/OwGnNCjF/553-tooltip-vertical-and-horizontal-alignment-corrections

PR #5655, opened 2024-02-01
fix: update three tier pricing in JS app

jamesgorrie:
Updating the pricing to the actual pricing for the test.

Apologies for the reordering, but that's how it is in the source data making it way easier to check.


paul-daniel-dempsey (approved):
Looks perfect

PR #5661, opened 2024-02-05
Add excludeIfInReferrerControlledTest to AB test model

tomrf1:
Currently there's some special logic in the reducer to prevent any other other AB test participations if the url specifies that the user should be in the `supporterPlusOnly` AB test.
[Example](https://support.theguardian.com/uk/contribute?acquisitionData={%20%22abTest%22:{%20%22name%22:%20%22supporterPlusOnly%22,%20%22variant%22:%20%22variant%22%20}%20})

We'd like to use this type of logic again for an upcoming feature. But really we want to avoid interfering with the `threeTierCheckout` test - if the user is in a referrer-controlled AB test then we do not want them to go into the `threeTierCheckout`.

This PR adds a new field to the AB test model: `excludeIfInReferrerControlledTest`.
If a test participation has `referrerControlled: true`, we exclude any other test which has `excludeIfInReferrerControlledTest: true`.
This means AB tests like `supporterPlusOnly` can override the `threeTierCheckout` test

### DIFF
--- a/support-frontend/assets/helpers/redux/commonState/reducer.ts
+++ b/support-frontend/assets/helpers/redux/commonState/reducer.ts
@@ -4,7 +4,6 @@ import type { ContributionTypes } from 'helpers/contributions';
 import { CountryGroup } from 'helpers/internationalisation';
 import type { IsoCountry } from 'helpers/internationalisation/country';
 import { fromCountryGroupId } from 'helpers/internationalisation/currency';
-import type { Participations } from '../../abTests/abtest';
 import type { CommonStateSetupData, Internationalisation } from './state';
 import { initialCommonState } from './state';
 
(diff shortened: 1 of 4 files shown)


support-frontend/assets/helpers/redux/commonState/reducer.ts
```diff
@@ -22,24 +21,11 @@ function getInternationalisationFromCountry(
...
-
 export const commonSlice = createSlice({
 	name: 'common',
 	initialState: initialCommonState,
 	reducers: {
 		setInitialCommonState(state, action: PayloadAction<CommonStateSetupData>) {
-			action.payload.abParticipations = getABTestNames(
-				action.payload.abParticipations,
-			);
 			const {
```
GHaberis, line 29: Is this `action.payload.abParticipations =` logic not required to add the `abParticipations` to the common state?

tomrf1: as far as I can tell this was just reassigning `action.payload.abParticipations` to itself, unless it contains `supporterPlusOnly`

GHaberis: Ah I see, so were doing our `abParticipations` switcheroo in the Reducer, makes more sense to have that logic in `abtests.ts` as you've done for the new logic.

GHaberis (approved):
Nice 👍

PR #5665, opened 2024-02-06
Promo cache respects stage

jamesgorrie:

Currently the `PromomotionCache` is a singleton, meaning that any service using it will use the same cache, irrespective of which table it came from.

This means that if
- we fill the cache with `PROD` data
- Become a test user
- Try to read CODE data

We will actually get `PROD` data until we need to refresh the cache. 

```mermaid

flowchart TB
    PriceSummaryService
    PriceSummaryService --> codeService
    PriceSummaryService --> prodService
    
    subgraph PROD
        prodService --> ProdPromotionService
        ProdPromotionService --> ProdCachedDynamoPromotionCollection["new CachedDynamoPromotionCollection"]
        
    end


    subgraph CODE
        codeService --> CodePromotionService
        CodePromotionService --> CodeCachedDynamoPromotionCollection["new CachedDynamoPromotionCollection"]
    end
        
    ProdCachedDynamoPromotionCollection <-- read CODE data --> PromotionCache 
    CodeCachedDynamoPromotionCollection -- write CODE data --> PromotionCache
    
    PromotionCache
```

[**Trello Card**](https://trello.com/c/PtZJsS3h/582-get-promocodes-working-on-prod-for-test-users)

## Testing

I've omitted testing as I couldn't think of anyway to test this that was any more clear than the implementation.

### DIFF
--- a/support-services/src/main/scala/com/gu/support/promotions/PromotionCache.scala
+++ b/support-services/src/main/scala/com/gu/support/promotions/PromotionCache.scala
@@ -11,6 +11,7 @@ case class PromotionCacheResponse(fetched: DateTime, promotions: Iterable[Promot
 }
 
 class PromotionCache {
+  // this val contains the mutable cache contents
   val promotionsRef = Ref[Option[PromotionCacheResponse]](None)
 
   def get: Option[Iterable[Promotion]] = promotionsRef.single().filter(_.isFresh).map(_.promotions)
--- a/support-services/src/main/scala/com/gu/support/promotions/PromotionCache.scala
+++ b/support-services/src/main/scala/com/gu/support/promotions/PromotionCache.scala
@@ -5,18 +5,17 @@ import org.joda.time.{DateTime, Minutes}
 
 import scala.concurrent.stm.{Ref, atomic}
 
-case class PromotionCache(fetched: DateTime, promotions: Iterable[Promotion]) {
-  def isFresh: Boolean = DateTime.now.getMillis - fetched.getMillis < PromotionCache.maxAge
+case class PromotionCacheResponse(fetched: DateTime, promotions: Iterable[Promotion]) {
+  val maxAge = Minutes.ONE.toStandardDuration.getMillis
+  def isFresh: Boolean = DateTime.now.getMillis - fetched.getMillis < maxAge
 }
 
-object PromotionCache {
-  val maxAge = Minutes.ONE.toStandardDuration.getMillis
-
-  val promotionsRef = Ref[Option[PromotionCache]](None)
+class PromotionCache {
+  val promotionsRef = Ref[Option[PromotionCacheResponse]](None)
 
   def get: Option[Iterable[Promotion]] = promotionsRef.single().filter(_.isFresh).map(_.promotions)
 
   def set(promotions: Iterable[Promotion], fetched: DateTime = DateTime.now): Unit = atomic { implicit txn =>
-    promotionsRef() = Some(PromotionCache(fetched, promotions))
+    promotionsRef() = Some(PromotionCacheResponse(fetched, promotions))
   }
 }
--- a/support-services/src/main/scala/com/gu/support/promotions/PromotionCollection.scala
+++ b/support-services/src/main/scala/com/gu/support/promotions/PromotionCollection.scala
@@ -19,11 +19,12 @@ class CachedDynamoPromotionCollection(config: PromotionsTablesConfig)
     extends DynamoPromotionCollection(config)
     with PromotionCollection {
 
-  override def all: Iterator[Promotion] = PromotionCache.get.getOrElse(fetchAndCache).toIterator
+  val cache = new PromotionCache
+  override def all: Iterator[Promotion] = cache.get.getOrElse(fetchAndCache).toIterator
 
   private def fetchAndCache = {
     val promotions = super.all.toList
-    PromotionCache.set(promotions)
+    cache.set(promotions)
     promotions
   }
 }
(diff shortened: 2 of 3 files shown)


support-services/src/main/scala/com/gu/support/promotions/PromotionCache.scala
```diff
@@ -5,18 +5,16 @@ import org.joda.time.{DateTime, Minutes}
 
 import scala.concurrent.stm.{Ref, atomic}
 
-case class PromotionCache(fetched: DateTime, promotions: Iterable[Promotion]) {
-  def isFresh: Boolean = DateTime.now.getMillis - fetched.getMillis < PromotionCache.maxAge
+case class PromotionCacheResponse(fetched: DateTime, promotions: Iterable[Promotion]) {
+  def isFresh: Boolean = DateTime.now.getMillis - fetched.getMillis < Minutes.ONE.toStandardDuration.getMillis
```
johnduffell, line 9: I like the extra documentation provided by the constant name, also this line is a bit verbose already
```suggestion
  val maxAge = Minutes.ONE.toStandardDuration.getMillis
  
  def isFresh: Boolean = DateTime.now.getMillis - fetched.getMillis < maxAge
```

support-services/src/main/scala/com/gu/support/promotions/PromotionCollection.scala
```diff
@@ -19,11 +19,12 @@ class CachedDynamoPromotionCollection(config: PromotionsTablesConfig)
     extends DynamoPromotionCollection(config)
     with PromotionCollection {
 
-  override def all: Iterator[Promotion] = PromotionCache.get.getOrElse(fetchAndCache).toIterator
+  val cache = new PromotionCache
```
johnduffell, line 22: have you used printlns or something just to check that only two PromotionCollection are created during the app lifecycle?  I think it is ok, but it would be annoying if it was one per request or something!

johnduffell: haveing searched about a bit I'm a bit concerned why we have 3 places creating promotion service - see the "non-test" versions in this search
https://github.com/search?q=repo%3Aguardian%2Fsupport-frontend+%22new+PromotionService%22&type=code

support-services/src/main/scala/com/gu/support/promotions/PromotionCache.scala
```diff
@@ -5,18 +5,16 @@ import org.joda.time.{DateTime, Minutes}
...
 }
 
-object PromotionCache {
-  val maxAge = Minutes.ONE.toStandardDuration.getMillis
-
-  val promotionsRef = Ref[Option[PromotionCache]](None)
+class PromotionCache {
+  val promotionsRef = Ref[Option[PromotionCacheResponse]](None)
```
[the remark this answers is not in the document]
jamesgorrie: As far as I know this isn't mutable, but rather a `Ref[T]` that contains `val t: T`. And you can change that using a method on `Ref` that returns a new `Ref` containing another value for `t` (but still of type `T`). 

And it get's stored in memory.

johnduffell (approved):
definitely approve of the diagram in the description!
most of the comments are just minor things, maybe not important.  However I am just not 100% sure we only create two caches - it looks like there are multiple places where we might create the cache in the end.  Worth having a bit of a check, even if having twice as many caches as we expect isn't a disaster!

support-services/src/main/scala/com/gu/support/promotions/PromotionCache.scala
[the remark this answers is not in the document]
johnduffell: I don't claim to understand fully, but I checked the documentation here
https://nbronson.github.io/scala-stm/#:~:text=ScalaSTM%20provides%20a%20mutable%20cell%20called%20a%20Ref

and they call it a "mutable cell"

So the knowledge of what's in the container is mutable, rather than the container itself or the thing that is inside it.

support-services/src/main/scala/com/gu/support/promotions/PromotionCollection.scala
jamesgorrie: > have you used printlns or something

Screenshot 2024-02-06 at 16 31 24

For piece of mind on `support-frontend` - this is all I got from multiple refreshes on the app once started.

jamesgorrie: I think the intention though is that every instance of PromotionsService has it's own cache.

We could probably make this more scalable, buy having a keyed cache e.g.
```scala
val cache: Map[Env, Promotions] = ???
def get(env: Env) = cache.get(env)
```

But I'm not sure our use-case warrants this complexity.

johnduffell: sounds good thanks for the updates (sorry I was probably editing my comment while you were already reading it).  If we wanted to be all belt and braces about it in case of regressions, we could keep a counter and alarm if it goes above 10, but I don't think it's worth it.

johnduffell (approved):

support-services/src/main/scala/com/gu/support/promotions/PromotionCollection.scala
jamesgorrie: Yea, or you could have the cache generated at the start up level and injected into the services to avoid it all-together... Let's give it a whirl and see if it becomes a problem.

jamesgorrie: I'm going to leave it as is an if we fancy, we can try flesh out a better abstraction for this.

PR #5666, opened 2024-02-06
enable threeTierCheckout and temporarily have only variant cohort

GHaberis:

Enable `threeTierCheckout` and disable `emotionalBenefits` AB tests.


paul-daniel-dempsey (approved):
Looks good

PR #5667, opened 2024-02-07
Remove US EOY

tomrf1:
removing old code from the last campaign


jamesgorrie (approved):
Great - sorry for leaving this laying about!

PR #5668, opened 2024-02-07
Support plus promocode support

jamesgorrie:

Adds support for `promoCode` for `SupporterPlus` in `support-workers`.

[**Trello Card**](https://trello.com/c/ClRWHxk7/585-add-support-for-promocode-for-supporterplus-in-createzuorasubscription)

## Why are you doing this?
So people can benefit from promos on Support+.

## Testing

I've deployed the lambda to `CODE`, and gone and bought a "Digital + print" subscription. And look what's showed up in Zuora (sorry for redactions, I am uncertain as to what is sensitive, so just covered all the IDs).

Screenshot 2024-02-08 at 15 03 34


support-models/src/main/scala/com/gu/support/workers/states/CreateZuoraSubscriptionState.scala
```diff
@@ -45,6 +46,8 @@ object CreateZuoraSubscriptionProductState {
       product: SupporterPlus,
       paymentMethod: PaymentMethod,
       salesForceContact: SalesforceContactRecord,
+      promoCode: Option[PromoCode],
+      user: User,
```
jamesgorrie, line 50: This is fine to be required as it is sent from all endpoints.

support-workers/src/main/scala/com/gu/zuora/productHandlers/ZuoraSupporterPlusHandler.scala
```diff
@@ -25,9 +21,15 @@ class ZuoraSupporterPlusHandler(
...
-        supporterPlusSubscriptionBuilder.build(state, csrUsername, salesforceCaseId),
-      )
+      subscribeItem <- Future
+        .fromTry(
+          supporterPlusSubscriptionBuilder
+            .build(state, csrUsername, salesforceCaseId)
+            .leftMap(BuildSubscribePromoError)
+            .toTry,
+        )
+        .withEventualLogging("SupporterPlus subscription data")
```
jamesgorrie, line 31: Lifted from [here](https://github.com/guardian/support-frontend/blob/main/support-workers/src/main/scala/com/gu/zuora/productHandlers/ZuoraPaperHandler.scala#L25-L29).

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/SupporterPlusSubcriptionBuilder.scala
```diff
@@ -49,8 +51,16 @@ class SupporterPlusSubcriptionBuilder(
...
 
+    applyPromoCodeIfPresent(
+      promotionService,
+      state.promoCode,
+      state.user.billingAddress.country,
+      productRatePlanId,
+      subscriptionData,
+    ).map { subscriptionData =>
+      subscribeItemBuilder.build(subscriptionData, state.salesForceContact, Some(state.paymentMethod), None)
+    }
```
jamesgorrie, line 63: Lifted from [here](https://github.com/guardian/support-frontend/blob/main/support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/PaperSubscriptionBuilder.scala#L40-L60)

support-models/src/main/scala/com/gu/support/workers/states/CreateZuoraSubscriptionState.scala
johnduffell: it is true that it's sent from all the endpoints but in the tests we don't want to bloat them too much - if you just pick only the billing country then it makes it easier from that side.

A big problem we have in tests in this code is you need a ridiclous amount of data passed in, half of which isn't actually needed for your specific test, hence the desire to minimise the state.  Blue peter style, there's one I made earlier here:
https://github.com/guardian/support-frontend/pull/5673/files#diff-1eed1787d655713c7200add111db3c6809de259f3b4731166f2303eb91811e64R45

I also notice you added the params on the end, which usually makes sense, but I actually tried to keep the ordering more similar to other of the CreateZuora*State classes, just to improve consistency.  Hence my billing country is first parameter.

support-models/src/main/scala/com/gu/support/workers/states/CreateZuoraSubscriptionState.scala
```diff
@@ -3,16 +3,17 @@ package com.gu.support.workers.states
...
 import com.gu.support.encoding.Codec.deriveCodec
 import com.gu.support.promotions.PromoCode
 import com.gu.support.workers.{PaymentMethod, SalesforceContactRecord, User, _}
 import org.joda.time.LocalDate
-import com.gu.support.encoding.CustomCodecs.{decodeCountry, decodeLocalTime, encodeCountryAsAlpha2, encodeLocalTime}
 import com.gu.support.encoding.{Codec, DiscriminatedType}
 import com.gu.support.redemptions.RedemptionData
 import com.gu.support.workers.GiftRecipient.{DigitalSubscriptionGiftRecipient, WeeklyGiftRecipient}
+// These imports contain implicits that are used in this file, and aren't unused as suggested by IntelliJ
+import com.gu.support.encoding.CustomCodecs.{decodeCountry, decodeLocalTime, encodeCountryAsAlpha2, encodeLocalTime}
```
johnduffell, line 16: yeah intellij does that sometimes, it's annoying, and sometimes even the ordering matters (if there's a clash)

johnduffell (approved):
looks good, it's very very familiar (!), however despite that I'm going to approve in case of expediency, but I would probably recommend checking my PR first: https://github.com/guardian/support-frontend/pull/5673
and we could merge that in preference, as I think I've taken a couple more conventions into account there, also I've added some automated testing specifically for promo codes in the zuora create step.
The only thing is I've only tested mine locally so far, not in CODE.

Take a look and see what you think and we can work out which to merge or whether to do a combination of the two.

support-models/src/main/scala/com/gu/support/workers/states/CreateZuoraSubscriptionState.scala
jamesgorrie: The problem I find is that we are essentially creating new transformation steps for all these types to narrow down their fields, making changes quite hard to work out where the transformation has happened.

It is also a load more code for each transformation. 

Test fixtures would be easy to create by just generating 1 full version and the modifying the fields we need to to make the test make sense.

Happy to chat about this offline as I have found this issue get me a few times and worry it will for anyone new to the codebase.

jamesgorrie:
Closing in favour of https://github.com/guardian/support-frontend/pull/5673

PR #5672, opened 2024-02-08
Three tier: Update international monthly price for top tier

GHaberis:

Full price for the top tier in the international region should be $46


paul-daniel-dempsey (approved):
All good

jamesgorrie (approved):
🤑

PR #5673, opened 2024-02-08
update support-workers to handle discounts for supporter plus thank you email

johnduffell:
TODO once the email itself is modified in braze test and live, we should to a PR remove the `amount` field from the email, as it could cause confusion.  At the moment it needs to be kept for backwards compatibility and will be misleading for discounted subs.
`"amount" : "13",`

This PR updates support-workers to understand promo codes for supporter plus.

We need to confirm appropriate braze field name (it already varies between Digisub and paper/weekly)

Testing in CODE works ok with an existing subscription without a promo code.  The Zuora state returns something including the new fields, and the sub looks correct.
```
      "paymentSchedule": {
        "payments": [
          {
            "date": "2024-02-09",
            "amount": 12
          },
          {
            "date": "2024-03-09",
            "amount": 12
          },
          {
            "date": "2024-04-09",
            "amount": 12
          },
          {
            "date": "2024-05-09",
            "amount": 12
          },
          {
            "date": "2024-06-09",
            "amount": 12
          },
          {
            "date": "2024-07-09",
            "amount": 12
          },
          {
            "date": "2024-08-09",
            "amount": 12
          },
          {
            "date": "2024-09-09",
            "amount": 12
          },
          {
            "date": "2024-10-09",
            "amount": 12
          },
          {
            "date": "2024-11-09",
            "amount": 12
          },
          {
            "date": "2024-12-09",
            "amount": 12
          },
          {
            "date": "2025-01-09",
            "amount": 12
          },
          {
            "date": "2025-02-09",
            "amount": 12
          }
        ]
      },
      "promoCode": null,
```
Plus the email step sent this new field to the SQS queue
`"subscription_details" : "£12.00 every month",`

---

With the promo code, 
the email step sent this:
`"subscription_details" : "£8.32 every month for 12 months, then £13.00 every month",`

and the zuora sub looks correct 
image


build.sbt
```diff
@@ -158,7 +158,7 @@ lazy val `support-workers` = (project in file("support-workers"))
     mergeStrategySettings,
   )
   .dependsOn(
-    `support-services`,
+    `support-services` % "test->test;compile->compile",
```
johnduffell, line 161: needed so `CreateZuoraSubscriptionSupporterPlusStepsSpec` can reference `CatalogServiceSpec.serviceWithFixtures` which has a copy of the PROD catalog.  It looks like there are already 3(!) copies of the prod catalog in this repo, presumably because other people hit this same issue, but I didn't want even more copy and paste as it's wasteful.

support-models/src/main/scala/com/gu/support/workers/states/CreateZuoraSubscriptionState.scala
```diff
@@ -42,8 +42,10 @@ object CreateZuoraSubscriptionProductState {
   ) extends CreateZuoraSubscriptionProductState
 
   case class SupporterPlusState(
+      billingCountry: Country,
       product: SupporterPlus,
       paymentMethod: PaymentMethod,
+      promoCode: Option[PromoCode],
```
johnduffell, line 48: we need the country and promo code in order to look up te promotion details and check that it's valid for this country and product.  This is standard for other promoted products.

support-models/src/main/scala/com/gu/support/workers/states/SendThankYouEmailState.scala
```diff
@@ -33,6 +33,8 @@ object SendThankYouEmailState {
       user: User,
       product: SupporterPlus,
       paymentMethod: PaymentMethod,
+      paymentSchedule: PaymentSchedule,
```
johnduffell, line 36: the zuora step looks up the previewed payment schedule and passes it on to the email step so that it can confidently tell the user what they will be charged during the promo and standard periods

support-workers/src/main/scala/com/gu/emailservices/SupporterPlusEmailFields.scala
```diff
@@ -1,14 +1,17 @@
...
+import com.gu.support.config.TouchPointEnvironment
 import com.gu.support.workers._
 import com.gu.support.workers.states.SendThankYouEmailState.SendThankYouEmailSupporterPlusState
 import org.joda.time.DateTime
 
 import scala.concurrent.{ExecutionContext, Future}
 
 class SupporterPlusEmailFields(
+    paperFieldsGenerator: PaperFieldsGenerator,
     getMandate: String => Future[Option[String]],
```
johnduffell, line 13: although the paperFieldsGenerator has getMandate already, it looks like we have a modified version copied into this class already, with different field names.  Perhaps we should split PaperFieldsGenerator into the payment realted stuff and the promo related stuff, so we don't have this loss of cohesion (but it's not affecting functionality, it just adds to code complexty.)

support-workers/src/main/scala/com/gu/emailservices/SupporterPlusEmailFields.scala
```diff
@@ -20,6 +23,14 @@ class SupporterPlusEmailFields(
...
       created,
     ).map { paymentFields =>
+      import state._
+      val promotion = paperFieldsGenerator.getAppliedPromotion(
+        state.promoCode,
+        state.user.billingAddress.country,
+        ProductTypeRatePlans.supporterPlusRatePlan(state.product, touchPointEnvironment).map(_.id).getOrElse(""),
+      )
+      val subscription_details = SubscriptionEmailFieldHelpers
+        .describe(paymentSchedule, product.billingPeriod, product.currency, promotion)
```
johnduffell, line 33: this line generates the nice "£blah per month for 3 months then £hello per month" human readable from the schedule that zuora generates

support-workers/src/main/scala/com/gu/emailservices/SupporterPlusEmailFields.scala
```diff
@@ -30,6 +41,7 @@ class SupporterPlusEmailFields(
         "billing_period" -> state.product.billingPeriod.toString.toLowerCase,
         "product" -> s"${state.product.billingPeriod.toString.toLowerCase}-supporter-plus",
         "zuorasubscriberid" -> state.subscriptionNumber,
+        "subscription_details" -> subscription_details,
```
johnduffell, line 44: currently I've called it subscription_details but I noticed it was different between paper and digisub, so it could be subscription_rate instead if we prefer consistency that way

support-workers/src/test/scala/com/gu/support/workers/createZuoraSubscription/CreateZuoraSubscriptionSupporterPlusStepsSpec.scala
```diff
@@ -0,0 +1,174 @@
+package com.gu.support.workers.createZuoraSubscription
```
johnduffell, line 1: this whole test is based on the above one, which I've moved both into a new package together to group them nicely

support-workers/src/test/scala/com/gu/support/workers/integration/SendThankYouEmailSpec.scala
```diff
@@ -137,6 +138,40 @@ object SendContributionEmail extends App {
   sendSingle(ef)
 
 }
+
+object SendSupporterPlusEmail extends App {
```
johnduffell, line 142: if you run this locally with AWS creds, it sends an email and also displays all the field names and values on the console so you can check it makes sense.

support-workers/src/test/scala/com/gu/support/workers/lambdas/UpdateSupporterProductDataSpec.scala
```diff
@@ -21,33 +21,36 @@ import scala.io.Source
...
 
   "UpdateSupporterProductData" should "return a valid SupporterRatePlanItem for a digisub gift redemption" in {
-    val state = decode[SendThankYouEmailState](digitalSubscriptionGiftRedemptionState)
-    state.isRight shouldBe true
-    val supporterRatePlanItem = UpdateSupporterProductData
-      .getSupporterRatePlanItemFromState(state.toOption.get, serviceWithFixtures)
-    inside(supporterRatePlanItem) { case Right(item) =>
-      item.value.identityId shouldBe "102803446"
+    val maybeState = decode[SendThankYouEmailState](digitalSubscriptionGiftRedemptionState)
+    inside(maybeState) { case Right(state) =>
```
jamesgorrie, line 36: I know some of this is copy / paste, but I find ScalaTest's `Values` `traits` useful in creating easy to read test avoiding nested assertions. e.g. [`EitherValues`](https://www.scalatest.org/user_guide/using_EitherValues)

jamesgorrie (approved):
Loving it - really nice work!

support-workers/src/main/scala/com/gu/emailservices/SupporterPlusEmailFields.scala
johnduffell: I meant to say payment method related stuff, if you look in the class they are both there together, but S+ uses its own function for payment methods stuff (that is probably copied and modified from paperFieldsGenerator) in order to insert the payment stuff into the email.
Let's revisit this once this PR is out, as it does need a bit of naming at least as you mention.

support-workers/src/main/scala/com/gu/emailservices/SupporterPlusEmailFields.scala
```diff
@@ -20,6 +23,14 @@ class SupporterPlusEmailFields(
       state.accountNumber,
       created,
     ).map { paymentFields =>
+      import state._
```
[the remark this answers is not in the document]
johnduffell: yeah good point, it does make it harder to understand in github but makes it look more tidy in the code.  I'll have a play and decide which way to go on it. (it was copied and pasted from somewhere else hence the inconsistency)

support-workers/src/test/scala/com/gu/support/workers/lambdas/UpdateSupporterProductDataSpec.scala
johnduffell: it's not even copy and paste, the main change is indentation, I haven't actually added anything material.  I'm always up for learning more scalatest stuff though so thanks for the pointer!

PR #5676, opened 2024-02-09
Remove the URL fragment for forcing ab tests

GHaberis:

These URL fragments for forcing yourself into an AB test should not be user facing at all, they're just a testing mechanism for engineers, we don't need to include it on this link as once the test has been enabled a user will be allocated into the correct AB test & cohort via their MVT cookie.


paul-daniel-dempsey (approved):
Yes, will take note of that in future with ab-tests, do not hard code within strings in the code (url's in this case)

PR #5677, opened 2024-02-09
refactor: use EitherValues on UpdateSupporterProductDataSpec

jamesgorrie:

Using `EitherValues` to make the test read easier.

This was based on [a conversation](https://github.com/guardian/support-frontend/pull/5673#discussion_r1484120789) with @johnduffell which I thought an example would be best to describe what I was getting at.

### DIFF
--- a/support-workers/src/test/scala/com/gu/support/workers/lambdas/UpdateSupporterProductDataSpec.scala
+++ b/support-workers/src/test/scala/com/gu/support/workers/lambdas/UpdateSupporterProductDataSpec.scala
@@ -11,6 +11,7 @@ import com.gu.support.workers.lambdas.UpdateSupporterProductDataSpec.{
 import com.gu.support.workers.states.SendThankYouEmailState
 import com.gu.supporterdata.model.ContributionAmount
 import io.circe.parser._
+import org.scalatest.EitherValues
 import org.scalatest.Inside.inside
 import org.scalatest.flatspec.AnyFlatSpec
 import org.scalatest.matchers.should.Matchers.convertToAnyShouldWrapper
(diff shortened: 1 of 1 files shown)


johnduffell (approved):
fab that's much much better! 👏 got rid of the indenting
Before you merge can I just confirm that the error message is good - i.e. if it's a Left(something) that the something definitely appears in the Test failed exception in scalatest?  Because this is the reason I changed it not to use `.get` any more because I couldn't see the deserialisation error it was just a "false is not true" kind of thing.

jamesgorrie:
> can I just confirm that the error message is good

@johnduffell sure this - how does this feel for you?
Screenshot 2024-02-13 at 08 15 39

> [!CAUTION]
> The Either on which value was invoked was not defined as a Right; it was Left(DecodingFailure at .user: Missing required field). 
ScalaTestFailureLocation: com.gu.support.workers.lambdas.UpdateSupporterProductDataSpec at (UpdateSupporterProductDataSpec.scala:25)

support-workers/src/test/scala/com/gu/support/workers/lambdas/UpdateSupporterProductDataSpec.scala
```diff
@@ -212,8 +203,12 @@ object UpdateSupporterProductDataSpec {
...
     }
   """
-
-  lazy val catalog: String = Source.fromURL(getClass.getResource("/catalog.json")).mkString
+  lazy val catalogSource = Source.fromURL(getClass.getResource("/catalog.json"))
+  lazy val catalog: String = {
+    val catalogString = catalogSource.mkString
+    catalogSource.close()
+    catalogString
+  }
```
jamesgorrie, line 211: I added this as I noticed the source wasn't closed (thanks IntelliJ). We close it once we've read the value. It has to be this was as they're lazy vals.

PR #5683, opened 2024-02-12
Correct the ids used for discounts in CODE

rupertbates:
When we create acquisition discounts in support-workers, we do so by adding a specific discount product rate plan to the new subscription and overriding the amount of the discount. The product rate plan we use as the template for this in production is in the `Discounts` product and is called `Percentage`, however we haven't been using the equivalent product rate plan from the code environment even though it does exist. 

This PR updates the ids used by support-workers in the CODE environment to match the ones in PROD for the sake of consistency.

### DIFF
--- a/support-config/src/main/resources/touchpoint.CODE.conf
+++ b/support-config/src/main/resources/touchpoint.CODE.conf
@@ -65,8 +65,8 @@ touchpoint.backend.environments {
     }
     promotions {
       discount {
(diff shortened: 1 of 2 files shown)


tomrf1 (approved):

jamesgorrie (approved):
Superb

johnduffell (commented):
looks good, ids check out correctly in zuora 👍

PR #5684, opened 2024-02-12
remove redundant Amount field from S+ email

johnduffell:
This PR removes the redundant "Amount" field from the supporter plus welcome email.

It. also does a bit of tidyup because the amount field is not currently accurate for other purposes.

Tested against the PROD braze email and it seems to send correctly to me without that field.
image


support-models/src/main/scala/com/gu/support/workers/Products.scala
```diff
@@ -31,11 +31,11 @@ case class Contribution(
 }
 
 case class SupporterPlus(
-    amount: BigDecimal,
+    preDiscountAmount: BigDecimal,
```
johnduffell, line 34: renamed to make it clearer what it represents on a discounted sub.

support-workers/src/main/scala/com/gu/acquisitions/AcquisitionDataRowBuilder.scala
```diff
@@ -123,7 +123,7 @@ object AcquisitionDataRowBuilder {
 
   private def productTypeAndAmount(product: ProductType): (AcquisitionProduct, Option[BigDecimal]) = product match {
     case c: Contribution => (AcquisitionProduct.RecurringContribution, Some(c.amount.toDouble))
-    case s: SupporterPlus => (AcquisitionProduct.SupporterPlus, Some(s.amount.toDouble))
+    case s: SupporterPlus => (AcquisitionProduct.SupporterPlus, None) // we don't send S+ amount because it may be discounted
```
johnduffell, line 126: don't want to send to fact_acquisition_event as it's not correct, and the analysts are more likely to use zuora data to find out the cost of a subscription.
Not sure if we need to check this change with anyone?  We could alternatively only send it for non discounted subs (it would be hard to know what to send for a discounted sub).  But ideally keep it simple and never send it.

support-workers/src/main/scala/com/gu/emailservices/SupporterPlusEmailFields.scala
```diff
@@ -33,7 +33,6 @@ class SupporterPlusEmailFields(
       val fields = List(
         "email_address" -> state.user.primaryEmailAddress,
         "created" -> created.toString,
-        "amount" -> state.product.amount.toString,
```
johnduffell, line 36: this is the field that the email no longer needs.

support-workers/src/main/scala/com/gu/support/workers/lambdas/UpdateSupporterProductData.scala
```diff
@@ -103,7 +103,7 @@ object UpdateSupporterProductData {
                 identityId = user.id,
                 productRatePlanId = productRatePlan.id,
                 productRatePlanName = s"support-workers added ${product.describe}",
-                Some(ContributionAmount(product.amount, product.currency.iso)),
+                None, // We don't send the amount for S+, because it may be discounted
```
johnduffell, line 106: we don't actually use this even for contributions, but it's been kept "just in case" even after apps don't need it via members-data-api, see https://github.com/guardian/members-data-api/pull/850
The current MDAPI code doesn't use it at all for any product.

support-models/src/main/scala/com/gu/support/workers/Products.scala
rupertbates: I don't think this is accurate though, currently (until we have a checkout which understands S+ discounts) this amount is the actual amount that the user will pay. I would revert this change but keep the others.

support-workers/src/main/scala/com/gu/acquisitions/AcquisitionDataRowBuilder.scala
rupertbates: It would be worth checking this with Data Design in case they do rely on it for something

support-models/src/main/scala/com/gu/support/workers/Products.scala
johnduffell: the checkout itself doesn't understand S+ discounts but there will be lots of discounted S+ going in via the 3 tier checkout
top tier are going to be discounted S+ subscriptions with a separate discounted GW sub.
So the S+ will all have amount=10 but the actual amount we would charge the reader would be less than 10 pounds.  Ones that are >10 would still be correct.
I'm not sure how that relates to your comment though - perhaps I should call it `undiscountedAmount`?

support-workers/src/main/scala/com/gu/acquisitions/AcquisitionDataRowBuilder.scala
johnduffell: data design don't use amount at all, they came back and said they check the zuora data in all cases (it sounds like that also includes recurring contribution, but I'll leave that as-is for now, as it's accurate)

rupertbates (approved):

johnduffell:
just as a note, this caused a support workers failure, as the model shared between support-frontend and support-workers has changed.  Since support-workers deployed first, and someone took a S+ in the gap, it was sent with the old field name, which did not succeed.
The solution would be to 
1. turn off CD for support workers until support-frontend had deployed, and
2.  deploy support-workers manually within 15 mins of that.  
3.  turn on CD

PR #5689, opened 2024-02-13
feat: hardcode promocodes for three tier into CreateSubscriptionController

jamesgorrie:
Hardcoding the promocode we should apply based on if they've come from the three tier landing page, are ordering weekly and a their billing address.

## Why are you doing this?

There is a use case
* A person is in the `threeTierCheckout` test
* When a person hits a geo-located landing page e.g. `/uk/contribute`
* They navigate to a three tier weekly checkout e.g. `/subscribe/weekly/checkout?promoCode=UK_MONTHLY_PROMO_CODE&threeTierCreateSupporterPlusSubscription=true&period=Monthly`
* They change the country they are in

We then show the new countries promo, but don't change the `promoCode` sent to the server. This means they will be charged full price.

We could have done this, or removed the promo code from the UI & request when a person changes a country.

This felt easier, and actually a better UX. It does however add quite a lot of tech debt and hardcoded logic for functionality we don't support with our current services. 

## How to test

I am writing some tests for local, and we can test this by following the steps above.


johnduffell:
> We then show the new countries promo, but don't change the promoCode sent to the server. This means they will be charged full price.

eek this is not right, we planned to fix that as part of the 1for1 incident, the subscribe call should fail if the promo couldn't be applied by the backend

johnduffell (approved):
looks good! in its own way! 👍

support-frontend/app/controllers/CreateSubscriptionController.scala
```diff
@@ -382,3 +396,45 @@ class CreateSubscriptionController(
...
+        "int-annual" -> "3TIER_WEEKLY_INT_ANNUAL",
+        "us-monthly" -> "3TIER_WEEKLY_US_MONTHLY",
+        "us-annual" -> "3TIER_WEEKLY_US_ANNUAL",
+        "ca-monthly" -> "3TIER_WEEKLY_CA_MONTHLY",
+        "ca-annual" -> "3TIER_WEEKLY_CA_ANNUAL",
+        "nz-monthly" -> "3TIER_WEEKLY_NZ_MONTHLY",
+        "nz-annual" -> "3TIER_WEEKLY_NZ_ANNUAL",
+        "au-monthly" -> "3TIER_WEEKLY_AU_MONTHLY",
+        "au-annual" -> "3TIER_WEEKLY_AU_ANNUAL",
+      ) get s"$id-$billingPeriod").flatten
```
johnduffell, line 435: let's bomb out if we didn't find a promo code - we don't want to accidentally sell non discounted ones
```suggestion
      Some((for {
        id <- maybeId
        billingPeriod <- maybeBillingPeriod
      } yield Map(
        "uk-monthly" -> "3TIER_WEEKLY_UK_MONTHLY",
        "uk-annual" -> "3TIER_WEEKLY_UK_ANNUAL",
        "eu-monthly" -> "3TIER_WEEKLY_EU_MONTHLY",
        "eu-annual" -> "3TIER_WEEKLY_EU_ANNUAL",
        "int-monthly" -> "3TIER_WEEKLY_INT_MONTHLY",
        "int-annual" -> "3TIER_WEEKLY_INT_ANNUAL",
        "us-monthly" -> "3TIER_WEEKLY_US_MONTHLY",
        "us-annual" -> "3TIER_WEEKLY_US_ANNUAL",
        "ca-monthly" -> "3TIER_WEEKLY_CA_MONTHLY",
        "ca-annual" -> "3TIER_WEEKLY_CA_ANNUAL",
        "nz-monthly" -> "3TIER_WEEKLY_NZ_MONTHLY",
        "nz-annual" -> "3TIER_WEEKLY_NZ_ANNUAL",
        "au-monthly" -> "3TIER_WEEKLY_AU_MONTHLY",
        "au-annual" -> "3TIER_WEEKLY_AU_ANNUAL",
      ) get s"$id-$billingPeriod").flatten.get)
```

PR #5690, opened 2024-02-13
make sure JVM always generates stack traces

johnduffell:
Following on from
https://github.com/guardian/support-frontend/pull/5682

Most of the time the stack traces was missing and we could only see "ClassCastExecption: null". This made it quite hard to understand what was going on.

It turns out this is an optimisation if an error is frequently generated, the JVM uses a generic instance without any useful information, in order to improve performance.  It can be turned off with `-XX:-OmitStackTraceInFastThrow,`

https://yoshihisaonoue.wordpress.com/2021/02/07/jvm-option-xx-omitstacktraceinfastthrow/

Tested on CODE - the instances come up healthy and the checkout displays correctly.

### DIFF
--- a/support-frontend/build.sbt
+++ b/support-frontend/build.sbt
@@ -81,6 +81,7 @@ Universal / javaOptions ++= Seq(
   "-J-XX:+PrintGCDetails",
   "-J-XX:+PrintGCDateStamps",
   s"-J-Xloggc:/var/log/${packageName.value}/gc.log",
+  "-XX:-OmitStackTraceInFastThrow",
 )
 
 addCommandAlias(


jamesgorrie (approved):
Really nice catch!

I was trying to think if there might be any performance degradation here, but I liked the point made in [this article](https://yoshihisaonoue.wordpress.com/2021/02/07/jvm-option-xx-omitstacktraceinfastthrow/)

>  if exceptions are thrown a lot of times and tons of stack trace are logged, the performance can be compromised. However, in that case, there is something wrong with the system and the clue for the problem is needed. So, if the application is not supposed to throw exceptions very often, we should use this flag even in production environment.

johnduffell:
> Assuming we could check memory and CPU for any increase in usage via AWS? From my understanding this flag would only start to show problems in performance at scale.

It's a good question to ask but I don't think the instances are heavily loaded at all, we only use that size instances to ensure there's plenty of memory, and we only have 3 for redundancy.
It may be worth thinking about how many instances we actually need - if we check how often instances die off and need replacement, the downtime over a year may well be smaller than the cost of the instance (maybe less so for support-frontend than things like manage-frontend)

johnduffell:
here is the CPU usage over the last week - it's only sampled every 5 minutes and I assume the sample is the average over the preceding 5 minutes, but it shows generally the instances are lightly loaded
image

PR #5694, opened 2024-02-13
3-tier: Track Contribution conversion & Guardian Weekly conversion with GTM for Top Tier

paul-daniel-dempsey:

**Objective:** On Top Tier checkout success drop a Guardian Weekly tracking pixel (via `successfulSubscriptionConversion()`) AND a Supporter Plus tracking pixel (via `successfulContributionConversion()`).

**Changes:**

- Update `handleSubscribeResult` in `submit.ts` (which runs on successful print subs checkouts) so we now call `successfulSubscriptionConversion` _and_ `successfulContributionConversion`  if `inThreeTierTestVariant` is `true` (eg. they have checked out via Top Tier variant of the Guardian Weekly checkout). This required some additional logic to calculate the S+ value (aka `digitalPriceDiscounted`) to send with the `successfulContributionConversion` call.

- We added a new `productType` property  to be used in GTM (for targeting) when a `successfulContributionConversion` is called, this is always set to "Contribution". 

- We've also added two additional properties to be used in GTM (for targeting) when a `successfulSubscriptionConversion` is called, these are `billingPeriod` and `productType`.

[**Trello Card**](https://trello.com/c/fshH8Spw/527-3-tier-tracking-review-and-implement)

Checkout can be viewed at the following url behind (currently disabled) ab-Test named threeTierCheckout ->
https://support.thegulocal.com/uk/contribute#ab-threeTierCheckout=variant


support-frontend/assets/helpers/subscriptionsForms/submit.ts
```diff
@@ -289,6 +292,21 @@ function onPaymentAuthorised(
...
+
+			// track conversion with GTM
+			successfulSubscriptionConversion();
+			if (inThreeTierTestVariant) {
+				/**
+				 * track S+ with GTM, digital split price unavailable from config
+				 * so discounted digitalPlusPrint price applied
+				 **/
+				successfulContributionConversion(
+					digitalPlusPrintPriceDiscounted,
```
GHaberis, line 304: `digitalPlusPrintPriceDiscounted` will be the combined price of GW and S+? I'm concerned we leave ourselves open to over reporting the Conversion value with this tag as is, as `successfulContributionConversion` should only report the S+ price.

Currently we don't send a price with the GW price in `successfulSubscriptionConversion`, _however_ the stakeholder has requested we do this. The amount we send `successfulSubscriptionConversion` would therefore need to be the GW price, and the amount sent in `successfulContributionConversion` would need to be the S+ price.

Lets catch-up to discuss options.

support-frontend/assets/helpers/subscriptionsForms/submit.ts
```diff
@@ -249,46 +269,76 @@ function onPaymentAuthorised(
...
+				paymentMethod.name,
 				billingPeriod,
+				productType,
 			);
+
+			if (inThreeTierTestVariant) {
+				const tierBillingPeriodName =
+					billingPeriod === 'Annual' ? 'annual' : 'monthly';
+				const contributionType =
+					billingPeriod === 'Annual' ? 'ANNUAL' : 'MONTHLY';
```
rupertbates, line 295:
```suggestion
				const tierBillingPeriodName = billingPeriod.toLowerCase();
				const contributionType = billingPeriod.toUpperCase();
```

rupertbates (approved):
Looks good to me but I'm not massively familiar with the code

PR #5695, opened 2024-02-14
3-Tier: Remove PayPal as a payment method from the GW checkout

GHaberis:

Testing the 3-tier checkout flow picked up some issues on the Guardian Weekly checkout (used by the Top Tier) when attempting to pay with PayPal. The conclusion was this is too complex to fix ahead of the test go live, so we'll remove PayPal as an available payment method for users in the test on this checkout.


paul-daniel-dempsey (approved):
Looks good

jamesgorrie (approved):
Simple solution, great!

PR #5698, opened 2024-02-15
add `/month` to the 3 tier disclaimer

johnduffell:
This PR updates the 3 tier disclaimer to include /month, otherwise it sounds like the monthly cost covers the whole year
image

I copied the block from
https://github.com/guardian/support-frontend/blob/ca20f7f7c77d807795086571d9a4f777b7d3bd53/support-frontend/assets/pages/supporter-plus-landing/components/threeTierCard.tsx#L170-L176
because it looked a bit neater and the overall logic looks the same as what this should be


paul-daniel-dempsey (approved):
Looks good

PR #5703, opened 2024-02-16
fix: use countryCode not currency on 3tier state choices

jamesgorrie:
## What are you doing in this PR?

Use `countryCode` not `currency` to choose which promo code to apply to which Weekly purchase.

## Previously
Screenshot 2024-02-16 at 08 45 00

```ts
if (state.product.currency == "USD" && state.product.billingPeriod == "Monthly") {
  // This was chosen as it matches first, even though it would match the US version too
  // as they both use `USD`
  choose "3TIER_SUPPORTERPLUS_INT_MONTHLY"
}
``` 

## Now

We select on the completely specific `countryCode` 

Screenshot 2024-02-16 at 08 43 58

[**Trello Card**](https://trello.com)

## Accessibility test checklist

-  [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-  [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-  [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-  [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

PR #5708, opened 2024-02-19
fix: use countryCode GB not UK for three tier

jamesgorrie:

We were using the wrong country code. `UK` isn't a country code, but `GB` is. This is a little confusing as the `countryGroup` code is `UK`.

[Here's the link through to the `billingAddress`](https://github.com/guardian/support-frontend/blob/main/support-frontend/app/services/stepfunctions/SupportWorkersClient.scala#L48) - which uses [Address](https://github.com/guardian/support-frontend/blob/main/support-models/src/main/scala/com/gu/support/workers/Address.scala#L13) which in turn uses `Country`- [specifically `UK`](https://github.com/guardian/support-frontend/blob/main/support-internationalisation/src/main/scala/com/gu/i18n/Country.scala#L98).

[This was the PR that created this broken behaviour](https://github.com/guardian/support-frontend/pull/5703/files).


paul-daniel-dempsey (approved):
Looks good

PR #5709, opened 2024-02-19
Three tier lower tier price cards

paul-daniel-dempsey:

Tier1/Low-Tier leads to price cards on second page of checkout. 
- PricesCard amounts are defined within the config. 
- Amounts equal or over S+ convert to S+ checkout.
- variantB setup (variant is existing 3-tier low-tier route)

|Step1|Step2|
|-----|-----|
|image|image|

[Trello Card](https://trello.com/c/2wGwyhja/591-build-3-tier-test-2-tier-1-variable-pricing)

Checkout can be viewed at the following url behind (currently disabled) ab-Test named threeTierCheckout ->
https://support.thegulocal.com/uk/contribute#ab-threeTierCheckout=control
https://support.thegulocal.com/uk/contribute#ab-threeTierCheckoutV2=variantA
https://support.thegulocal.com/uk/contribute#ab-threeTierCheckoutV2=variantB


support-frontend/assets/components/priceCards/priceCardsContainer.tsx
```diff
@@ -33,43 +38,67 @@ const contributionTypeToPaymentInterval: Partial<
...
 	const {
 		amounts: frequencyAmounts,
 		defaultAmount,
 		hideChooseYourAmount,
-	} = amountsCardData[frequency];
+	} = getTierLowPiceCardAmounts(
+		paymentFrequency,
+		countryGroupId,
+		inThreeTierVariantB,
+	) ?? amountsCardData[paymentFrequency];
```
jamesgorrie, line 81: To keep the three tier logic out of the method and leave it to the implementing component, you could do something like:

```suggestion
	} = inThreeTierVariantB ? getTierLowPiceCardAmounts(
		paymentFrequency,
		countryGroupId,
		) : amountsCardData[paymentFrequency];
```

That way `getTierLowPiceCardAmounts` would always return something.

support-frontend/assets/components/priceCards/priceCardsContainer.tsx
```diff
@@ -33,43 +38,67 @@ const contributionTypeToPaymentInterval: Partial<
...
 		(state) => state.page.checkoutForm.product,
 	);
 	const minAmount = useContributionsSelector(getMinimumContributionAmount());
+
+	const { abParticipations } = useContributionsSelector(
+		(state) => state.common,
+	);
+	const inThreeTierVariantB = abParticipations.threeTierCheckout === 'variantB';
+
+	const getTierLowPiceCardAmounts = (
```
jamesgorrie, line 60:
```suggestion
	const getTierLowPriceCardAmounts = (
```

jamesgorrie (approved):
A couple of tiny code suggestions, but looks good.

Just to clarify my understanding though - what are we testing here? That if people come through on tier-1 and we show them "Choose more" they will?

support-frontend/assets/components/priceCards/priceCardsContainer.tsx
paul-daniel-dempsey: Ah yes, corrected, thanks

support-frontend/assets/helpers/abTests/abtestDefinitions.ts
```diff
@@ -106,10 +106,13 @@ export const tests: Tests = {
 	threeTierCheckout: {
```
GHaberis, line 106: I think we should give this test a new name as this'll make it easier to differentiate the results between test 1 and test 2, `threeTierCheckoutV2` would be fine

support-frontend/assets/components/priceCards/priceCardsContainer.tsx
paul-daniel-dempsey: Am not sure this works because the call `getTierLowPriceCardsAmounts` can also legitimately  return `undefined` and you would get an type error in this scenario with the solution above.

support-frontend/assets/components/paymentButton/defaultPaymentButtonContainer.tsx
```diff
@@ -63,8 +63,8 @@ export function DefaultPaymentButtonContainer({
 	const { abParticipations } = useContributionsSelector(
 		(state) => state.common,
 	);
-	const isInThreeTierCheckoutTest =
-		abParticipations.threeTierCheckout === 'variant';
+	const inThreeTierVariant =
+		abParticipations.threeTierCheckout.startsWith('variant');
```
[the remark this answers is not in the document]
paul-daniel-dempsey: It was supposed to allow for a super-set of `variant` and `variantB` but agree this is an open loophole which is not great. I wonder whether a single check/function held somewhere would be better than repeating these multiple lines all over the place to be honest.

support-frontend/assets/helpers/abTests/abtestDefinitions.ts
paul-daniel-dempsey: Yes agree, otherwise I cant push it through into production either without affecting current test.

jamesgorrie (approved):
A test like this seems to be incredibly far reaching. I wonder if we could find some time to work out why and if there is anything to do to make it touch less files?

Nice work!

support-frontend/assets/components/paymentButton/defaultPaymentButtonContainer.tsx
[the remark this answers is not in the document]
paul-daniel-dempsey: Done - am explicitly only checking for `variant` , `variantA` and `variantB` now.

support-frontend/assets/helpers/abTests/abtestDefinitions.ts
paul-daniel-dempsey: Done - `threeTierCheckoutV2` setup with a `variantA` and `variantB` now/.

support-frontend/assets/components/priceCards/priceCardsContainer.tsx
```diff
@@ -33,43 +42,67 @@ const contributionTypeToPaymentInterval: Partial<
...
 	const { amountsCardData } = amounts;
 	const { selectedAmounts, otherAmounts } = useContributionsSelector(
 		(state) => state.page.checkoutForm.product,
 	);
 	const minAmount = useContributionsSelector(getMinimumContributionAmount());
+
+	const inThreeTierVariantB = inThreeTierV2VariantB(
+		useContributionsSelector((state) => state.common).abParticipations,
+	);
+	const tierCards = inThreeTierVariantB ? tierCardsVariantB : tierCardsVariantA;
```
GHaberis, line 62: I was a little confused why we assign `tierCards` the value of `tierCardsVariantA` here when `inThreeTierVariantB` is `false`, as looking below I see we dom't use `tierCards` if this is the case - so the value of `tierCardsVariantA` will never be used, so why  assign it to `tierCards`? I think we could simplify things here.

support-frontend/assets/components/priceCards/priceCardsContainer.tsx
```diff
@@ -33,43 +42,67 @@ const contributionTypeToPaymentInterval: Partial<
...
+	const getTierLowPriceCardAmounts = (
+		contributionType: ContributionType,
+		countryGroup: CountryGroupId,
+		intThreeTierVariant: boolean,
+	): AmountValuesObject | undefined => {
+		const tierBillingPeriod =
+			contributionType === 'ANNUAL' ? 'annual' : 'monthly';
+		const priceCards = tierCards[`tier1`].plans[tierBillingPeriod].priceCards;
+		if (priceCards && intThreeTierVariant) {
+			return priceCards[countryGroup];
```
GHaberis, line 73: This function only returns an `AmountValuesObject` if `intThreeTierVariant` is `true` - otherwise it returns `undefined`. Rather than making this check at the end of the function we could simplify it by checking whether `intThreeTierVariant` is `true` at the very top of the and just returning immediately if it's `false`? I think it might not even be necessary to have this function

Instead of...

```
	const {
		amounts: frequencyAmounts,
		defaultAmount,
		hideChooseYourAmount,
	} = getTierLowPriceCardAmounts(
		paymentFrequency,
		countryGroupId,
		inThreeTierVariantB,
	) ?? amountsCardData[paymentFrequency];
```
Could we do...

```
const {
		amounts: frequencyAmounts,
		defaultAmount,
		hideChooseYourAmount,
	} = inThreeTierVariantB ? tierCardsVariantB[`tier1`].plans[tierBillingPeriod].priceCards[countryGroup] : amountsCardData[paymentFrequency];
```

We could then get rid of `getTierLowPriceCardAmounts` and the `tierCards` variable referenced in my comment above?

support-frontend/assets/pages/supporter-plus-landing/twoStepPages/secondStepCheckout.tsx
```diff
@@ -109,10 +125,13 @@ export function SupporterPlusCheckout({
 			<Box cssOverrides={shorterBoxMargin}>
 				<BoxContents>
 					{showPriceCards ? (
-						<ContributionsPriceCards />
+						<ContributionsPriceCards
+							paymentFrequency={contributionType}
+							inThreeTierVariant={inThreeTierVariantB}
```
GHaberis, line 130: Is this correct? We're assigning `inThreeTierVariantB` to a prop labelled `inThreeTierVariant`, the prop name implies this should be `true` if they were in variant A too?

support-frontend/assets/components/priceCards/priceCardsContainer.tsx
paul-daniel-dempsey: Ah yes, 'tierCards' will always be tierCardsVariantB`, have simplified `tierCards` to `tierCardsVaraiantB`

support-frontend/assets/components/priceCards/priceCardsContainer.tsx
paul-daniel-dempsey: Using your suggestion above,  I have simplified and remove `getTierLowPricaeCardAmounts` as requetsed with this rewrite:`	const tierBillingPeriod =
		paymentFrequency === 'ANNUAL' ? 'annual' : 'monthly';
	const tierCardData =
		tierCardsVariantB.tier1.plans[tierBillingPeriod].priceCards;
	const {
		amounts: frequencyAmounts,
		defaultAmount,
		hideChooseYourAmount,
	} = tierCardData
		? tierCardData[countryGroupId]
		: amountsCardData[paymentFrequency];`

paul-daniel-dempsey: updated as shown above

support-frontend/assets/components/priceCards/priceCardsContainer.tsx
paul-daniel-dempsey: Using Georges suggestion below,  I have simplified and removed `getTierLowPriceCardAmounts` as requested with this rewrite:
```
	const inThreeTierVariantB = inThreeTierV2VariantB(
		useContributionsSelector((state) => state.common.abParticipations),
	);
	const tierBillingPeriod =
		paymentFrequency === 'ANNUAL' ? 'annual' : 'monthly';
	const tierCardData =
		tierCardsVariantB.tier1.plans[tierBillingPeriod].priceCards;
	const {
		amounts: frequencyAmounts,
		defaultAmount,
		hideChooseYourAmount,
	} = inThreeTierVariantB && tierCardData
		? tierCardData[countryGroupId]
		: amountsCardData[paymentFrequency];
```

support-frontend/assets/components/priceCards/priceCardsContainer.tsx
GHaberis: Looks good @paul-daniel-dempsey 👍

support-frontend/assets/pages/supporter-plus-landing/twoStepPages/secondStepCheckout.tsx
paul-daniel-dempsey: Yes, its correctly used but your right, very confusing to load `inThreeTierVariantB` into prop `inThreeTierVariant`.  Have decided this `inThreeTierVariant` prop-drilling really isn't necessary and adds needless lines of code. Have amended to use it directly within `priceCards.tsx`.

PR #5715, opened 2024-02-20
fix: use currency as discriminator in 3-tier step functions

jamesgorrie:

**Previously we had `currency` as a discriminator.**

This didn't work as `US` and `INT` both shared `USD`.

**[We then changed to `countryCode`](https://github.com/guardian/support-frontend/pull/5703).**

This doesn't work for more reasons, especially for `EUR` and `INT` which have an array of country codes. (@paul-daniel-dempsey noticed this yesterday).

---

This takes both these learnings for our new solution. Use `currency` with an added discriminator for `isUS` to differentiate between `US` and `INT`.

The comment says it, but the order here is important as the `US` is actually a tighter condition that `INT`. While the implementation here leaves a bit to be desired, it now works. 

Lesson is it might have been easier to setup a new lambda here. 

---

Tested on `CODE`

**USD / INT**
Screenshot 2024-02-20 at 09 50 37

**EUR**
Screenshot 2024-02-20 at 10 03 17

**USD / US**
Screenshot 2024-02-20 at 10 06 02

**UK**
Screenshot 2024-02-20 at 10 21 23

**AU**
Screenshot 2024-02-20 at 10 23 31


rupertbates:
You could and probably should use `fulfilmentOptions` for this. If it is currency `USD` & fulfilmentOptions `Domestic` then it is the US otherwise it is international.

jamesgorrie:
> You could and probably should use `fulfilmentOptions` for this. If it is currency `USD` & fulfilmentOptions `Domestic` then it is the US otherwise it is international.

@rupertbates Is `INT` only ever applied to people outside of all the other territories? e.g. Could I be paying in `EUR` but be `RestOfWorld`?

rupertbates:
> > You could and probably should use `fulfilmentOptions` for this. If it is currency `USD` & fulfilmentOptions `Domestic` then it is the US otherwise it is international.
> 
> @rupertbates Is `INT` only ever applied to people outside of all the other territories? e.g. Could I be paying in `EUR` but be `RestOfWorld`?

Yes international is outside of all other regions, anyone paying in Euros will be domestic

paul-daniel-dempsey (approved):
Yep can see country code removed and  run from Ccy instead. Exception for US with IsUS (presume cant assume the promo-code is the differentiator)

PR #5716, opened 2024-02-20
fix: use fulfilmentOptions as discriminator in 3-tier step functions

jamesgorrie:

As per @rupertbates' suggestion - uses the `fulfilmentOptions` to discern between `US` and `INT`. I've not used it elsewhere as those should always be domestic from my understanding which them makes it more explicit as to why we're using it.

**Aus**
Screenshot 2024-02-20 at 11 41 06

**Germany**
Screenshot 2024-02-20 at 11 41 40

**UK**
Screenshot 2024-02-20 at 11 41 59

**India**
Screenshot 2024-02-20 at 11 42 23

**US**
Screenshot 2024-02-20 at 11 42 45


rupertbates (approved):
Nice

PR #5718, opened 2024-02-21
Fix webpack

tomrf1:
After a [recent webpack upgrade](https://github.com/guardian/support-frontend/pull/5706) we're unable to run locally

### DIFF
--- a/support-frontend/package.json
+++ b/support-frontend/package.json
@@ -207,7 +207,7 @@
     "web-vitals": "^2.0.0",
     "webpack": "^5.76.0",
     "webpack-bundle-analyzer": "^4.5.0",
-    "webpack-cli": "^4.9.2",
+    "webpack-cli": "^5.1.4",
     "webpack-dev-server": "^5.0.2",
     "webpack-manifest-plugin": "^5.0.0",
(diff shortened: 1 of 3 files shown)


support-frontend/package.json
```diff
@@ -207,7 +207,7 @@
     "web-vitals": "^2.0.0",
     "webpack": "^5.76.0",
     "webpack-bundle-analyzer": "^4.5.0",
-    "webpack-cli": "^4.9.2",
+    "webpack-cli": "^5.1.4",
```
tomrf1, line 210: the older version gave us the error: `options has an unknown property '_assetEmittingPreviousFiles'.`

support-frontend/webpack.dev.js
```diff
@@ -7,12 +7,12 @@ module.exports = merge(common('[name].css', '[name].js', false), {
 	devtool: 'inline-source-map',
 	devServer: {
 		allowedHosts: 'all',
-		proxy: {
+		proxy: [{
```
tomrf1, line 10: proxy must now be an array

jamesgorrie (approved):
Great catch.

PR #5721, opened 2024-02-21
Changes for VAT compliance

MaelGNM:
This PR:
-  updates the list of countries we only sell contributions to support-frontend/assets/helpers/internationalisation/country.ts`)
- updates the countries in the Fastly snippets to align with the list above
- changes how users are assigned to an amounts test so that it first checks whether they are in the `VAT_COMPLIANCE` test

## Why are you doing this?

This ensures we are up to date with our tax department on the types of products we sell to different countries (contributions only).

## How to test

This can be tested by navigating to the checkout page from different countries (browserstack can be used to do that) and checking the amounts displayed are correct.


support-frontend/assets/helpers/internationalisation/country.ts
```diff
@@ -642,175 +642,199 @@ const stripePaymentRequestAllowedCountries = [
 ];
 
 const countriesAffectedByVATStatus: IsoCountry[] = [
```
MaelGNM, line 644: 24 countries were added, this is a big change because I have also ordered the list alphabetically

support-frontend/conf/fastly-snippets/vat-compliance-redirects-recv.vcl
```diff
@@ -1,175 +1,199 @@
 # type: recv
 # fiddle: https://fiddle.fastly.dev/fiddle/6db36311
 if (
-    client.geo.country_code == "KE" ||
```
MaelGNM, line 4: Same as the other one, 24 countries added here and the list was ordered alphabetically

support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -259,6 +259,35 @@ function getAmountsTestVariant(
 		}
 	};
 
+	// Is the country in the list for contributions only checkout?
+	// This relies on the existence of an amounts test with a specific name
+	// It could be done using the countriesAffectedByVATStatus list
+	// but this would need the list and the amounts test to be in sync
+	// and this way simplifies testing as it is all set up in the RRCP
+	const contribOnlyTestName = 'VAT_COMPLIANCE';
```
johnduffell, line 267: just wondering if we should line up the name and the const, unless there's a reason for the difference?

MaelGNM: That's a good question. The difference is due to how we speak about this project. The contributions only checkout is part of the VAT compliance project.
This amounts test is used to show the contributions only checkout, which is why I named the const this way as it seemed more specific. The amounts test itself was created back in November 2023, and predates the use of the term "contributions only" in this project.

Happy to rename the const if you think it's confusing.

johnduffell: that's fine, I was mainly wanted to check it was considered (and not a copy and paste error!)

johnduffell (approved):
I don't know much about this area but as long as it tests out ok I am happy to approve! also update the PR description for future clarity (delete the template content if you don't need it for the PR)

GHaberis (approved):
Looks good

Issue #5722, opened 2024-02-21
Generic Checkout


jamesgorrie:
# Product agnostic checkout!

```[tasklist]
### Bugs
- [ ] State selector doesn't respond to url `/{geo}` but rather the `GU_Country` cookie
``` 

```[tasklist]
### Payment processing / thank you page
- [ ] Error handling
- [ ] Timeout on poll (show thank you anyway)
- [ ] Make `thank-you` route and add thank you page
``` 

```[tasklist]
### Data
- [ ] Ensure that the JS is smaller (or at worst the same) than the current JS
``` 

```[tasklist]
### Things to get back to
- [ ] Server side rendering
- [ ] Apple Pay
- [ ] ReCaptcha expiry https://github.com/guardian/support-frontend/pull/5787/files#r1539567520
- [ ] Add level 1 (`<h1>`) heading
- [ ] Tracking
- [ ] Ensure D&I are aware of change
```

PR #5704, opened 2024-02-16
Generic checkout

jamesgorrie:
WIP

PR #5723, opened 2024-02-22
refactor: decouple orderSummary from contributionType

jamesgorrie:
- Renames `CheckmarkList` and `Accordion` => `CheckList` as they are all the same thing. ✨ consistency!
- Removes `contributionType` from `ContributionsOrderSummary` (with the aim of it becoming _the_ `OrderSummary` 🧹 
- Add `description` and `paymentFrequency` to `ContributionsOrderSummary`

[**Trello Card**](https://trello.com/c/ZY1dQDdY/619-generic-checkout-mvp-using-catalog-data)

## Why are you doing this?
In the aim of making these components useful for all products ([the generic checkout](https://github.com/guardian/support-frontend/pull/5704/files#r1497743253)) we can no longer rely on `contributionType` as it varies more widely across products. 

I have decided to go for the most vague of types, `string`, for `description` and `paymentFrequency` which would give us the greatest flexibility in the checkout. We might want to tighten the types once we know what they are for all products, but would suggest we also keep an eye on being flexible for testing and new product and payment types.

## How to test

This should be a no-op

## Screenshots

| One-time | Monthly | Annual |
|---|---|---|
| Screenshot 2024-02-21 at 21 29 47 | Screenshot 2024-02-21 at 21 30 03 | Screenshot 2024-02-21 at 21 30 16 |

Part of #5722

### DIFF
--- a/support-frontend/assets/components/checkList/checkList.tsx
+++ b/support-frontend/assets/components/checkList/checkList.tsx
@@ -4,7 +4,7 @@ import { from, palette, space, textSans } from '@guardian/source-foundations';
 import { SvgCrossRound, SvgTickRound } from '@guardian/source-react-components';
 import Tooltip from 'components/tooltip/Tooltip';
 
-const checkListIconCss = (style: CheckmarkListStyle) => css`
+const checkListIconCss = (style: CheckListStyle) => css`
 	vertical-align: top;
 	padding-right: ${style === 'compact' ? '4px' : '10px'};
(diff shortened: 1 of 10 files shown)


paul-daniel-dempsey (approved):
Looks much cleaner now. Defaulting to annual on `contibutionType` in `contributionsOrderSummaryContainer.tsx` does seem like ther should be a better way but the limitation is well  documented there

PR #5726, opened 2024-02-22
Increase tolerance for slow zuora jobs

tomrf1:
We sometimes get many supporter-product-data state machine failures during the early hours. This is because zuora query results can be very slow at this time.
Currently the `FetchResults` step retries for 20 minutes.
This PR increases this to 40 minutes.

This chart (from @graham228221) suggests some queries can take up to an hour -
zuora_is_slow

### DIFF
--- a/supporter-product-data/cloudformation/cfn.yaml
+++ b/supporter-product-data/cloudformation/cfn.yaml
@@ -267,7 +267,7 @@ Resources:
                    {
                     "ErrorEquals": ["States.ALL"],
                     "IntervalSeconds": 60,
-                    "MaxAttempts": 20,
+                    "MaxAttempts": 30,
                     "BackoffRate": 1.0
                   }]
               },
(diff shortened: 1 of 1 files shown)


supporter-product-data/cloudformation/cfn.yaml
```diff
@@ -267,7 +267,7 @@ Resources:
                    {
                     "ErrorEquals": ["States.ALL"],
                     "IntervalSeconds": 60,
-                    "MaxAttempts": 20,
+                    "MaxAttempts": 40,
```
jamesgorrie, line 270: Could we maybe leave a comment for future folk that this is something we expect in the unusual (although regular) circumstances of the morning hours?

jamesgorrie (approved):
I like it.

As a thought experiment to think about risks, I asked myself "What would happen if this regularly took 39 minutes and we weren't alerted?"

I can't think of anything, but thought maybe someone with more domain knowledge might have an answer.

johnduffell (approved):

supporter-product-data/cloudformation/cfn.yaml
tomrf1: I'm not sure it's possible here as it's json! An incentive to migrate to cdk...

PR #5735, opened 2024-02-26
Disable supporter-product-data alarm in early hours

tomrf1:
Sometimes during the night zuora becomes very slow and the state machine fails.
I previously tried [increasing the tolerance](https://github.com/guardian/support-frontend/pull/5726) to 40mins, but it's not enough.

Instead, this PR changes the alarm to only trigger after 6am. We expect the state machine to fail until that time.

### DIFF
--- a/supporter-product-data/cloudformation/cfn.yaml
+++ b/supporter-product-data/cloudformation/cfn.yaml
@@ -450,16 +450,25 @@ Resources:
       AlarmDescription: >
         The supporter-product-data step function failed, check for failed executions here:
         https://eu-west-1.console.aws.amazon.com/states/home?region=eu-west-1#/statemachines/view/arn:aws:states:eu-west-1:865473395570:stateMachine:supporter-product-data-PROD
-      MetricName: ExecutionsFailed
-      Namespace: AWS/States
-      Dimensions:
-        - Name: StateMachineArn
-          Value: !Ref SupporterProductData
+      Metrics:
+        - Id: errors
+          # Do not alarm between 00:00 and 06:00 because Zuora is very slow and we often have failures during these hours
+          Expression: "IF(HOUR(m1) > 5, m1)"
+        - Id: m1
+          ReturnData: false
+          MetricStat:
+            Metric:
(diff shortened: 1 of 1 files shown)


johnduffell (approved):
I was going to say we shouldn't even kick them off if we're not going to use the result, as we're just adding to zuora's queue when it's busy, but I think it runs in a few seconds anyway (of actual run time), so it wouldn't add much?
Other than that good idea and I didn't know you could do a calculated metric for alarm using the time! 👏

supporter-product-data/cloudformation/cfn.yaml
```diff
@@ -450,16 +450,25 @@ Resources:
       AlarmDescription: >
         The supporter-product-data step function failed, check for failed executions here:
         https://eu-west-1.console.aws.amazon.com/states/home?region=eu-west-1#/statemachines/view/arn:aws:states:eu-west-1:865473395570:stateMachine:supporter-product-data-PROD
-      MetricName: ExecutionsFailed
-      Namespace: AWS/States
-      Dimensions:
-        - Name: StateMachineArn
-          Value: !Ref SupporterProductData
+      Metrics:
+        - Id: errors
```
johnduffell, line 454:
```suggestion
        - Id: errorsOutsideZuoraSlowHours
```

supporter-product-data/cloudformation/cfn.yaml
```diff
@@ -450,16 +450,25 @@ Resources:
...
+          # Do not alarm between 00:00 and 06:00 because Zuora is very slow and we often have failures during these hours
+          Expression: "IF(HOUR(m1) > 5, m1)"
+        - Id: m1
```
johnduffell, line 457:
```suggestion
        - Id: errors
```

supporter-product-data/cloudformation/cfn.yaml
johnduffell: comment below looks good, I wasn't sure where the ID is shown?  Might not be worth updating but a suggestion anyway

PR #5736, opened 2024-02-26
build: bump logback-classic

AndreaDiotallevi:
following on from https://github.com/guardian/support-frontend/pull/5737
We had to bump logback to 1.3 or 1.4 latest, but play didn't start with either of those.
Having bumped play to 3.0 in the above PR, we can update the explicit version across the board.
All tests pass and support-frontend still can start (and log stuff)


johnduffell (approved):
looks good, I've rebased and tested locally 👍

PR #5737, opened 2024-02-26
update to play 3 in order to reduce security vulns

johnduffell:
This will update play in both support frontend and payment api to 3 (from 2.8)

It mainly involves changing the word "akka" to "org.apache.pekko"!

This is needed to get a better version of logback that isn't vulnerable, but it's also a good thing in general.

I noticed that identity-auth-play depends on play 2.8 but it doesn't seem to cause an issue, perhaps because they have different package names.

### DIFF
--- a/project/LibraryVersions.scala
+++ b/project/LibraryVersions.scala
@@ -9,7 +9,7 @@ object LibraryVersions {
   val jacksonDatabindVersion = "2.15.2"
   val okhttpVersion = "3.14.9"
   val scalaUriVersion = "4.0.3"
-  val playCirceVersion = "2814.2"
+  val playCirceVersion = "3014.1"
   val stripeVersion = "21.15.0" // Supports API version 2019-05-16
   val oktaJwtVerifierVersion = "0.5.7"
 }
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=1.9.4
+sbt.version=1.9.8
--- a/support-frontend/build.sbt
+++ b/support-frontend/build.sbt
@@ -28,7 +28,7 @@ libraryDependencies ++= Seq(
   "com.gu" %% "identity-test-users" % "0.8",
   "com.google.guava" % "guava" % "32.1.1-jre",
   "io.lemonlabs" %% "scala-uri" % scalaUriVersion,
-  "com.gu.play-googleauth" %% "play-v28" % "2.2.7",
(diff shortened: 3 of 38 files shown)


project/LibraryVersions.scala
```diff
@@ -9,7 +9,7 @@ object LibraryVersions {
   val jacksonDatabindVersion = "2.15.2"
   val okhttpVersion = "3.14.9"
   val scalaUriVersion = "4.0.3"
-  val playCirceVersion = "2814.2"
+  val playCirceVersion = "3014.1"
```
johnduffell, line 12: the `30` has to match the play version

project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.9.4
+sbt.version=1.9.8
```
johnduffell, line 1: opportunistic bump

support-frontend/build.sbt
```diff
@@ -43,6 +43,10 @@ libraryDependencies ++= Seq(
 )
 dependencyOverrides += "com.fasterxml.jackson.core" % "jackson-databind" % jacksonDatabindVersion
 
+ThisBuild / libraryDependencySchemes ++= Seq(
+  "org.scala-lang.modules" %% "scala-xml" % VersionScheme.Always
+)
```
johnduffell, line 48: this was caused by conflicting versions of scala-xml which SBT worries will cause a binary incompatiblity, so it aborts at that point.  In reality it doesn't cause an issue, this also happens elsewhere, and the usual suggestion is to add this line.  E.G. https://github.com/sbt/sbt/issues/6997#issuecomment-1310637232

support-frontend/build.sbt
```diff
@@ -28,7 +28,7 @@ libraryDependencies ++= Seq(
   "com.gu" %% "identity-test-users" % "0.8",
   "com.google.guava" % "guava" % "32.1.1-jre",
   "io.lemonlabs" %% "scala-uri" % scalaUriVersion,
-  "com.gu.play-googleauth" %% "play-v28" % "2.2.7",
+  "com.gu.play-googleauth" %% "play-v30" % "3.0.6",
```
johnduffell, line 31: name has to match the version of play

tomrf1 (approved):
👍

johnduffell:
tested in CODE with a paper
tested in PROD(test user) with a paper DD and stripe contribution
Also real users seem to be going through with S+ and contribution (stripe and paypal)

PR #5739, opened 2024-02-27
Suppress support-workers alarm if test user

tomrf1:
If the state machine fails but `testUser: true`, move to the `CheckoutFailure` state, which does not trigger an alarm


rupertbates (approved):
Looks good 👏

PR #5743, opened 2024-02-27
Generic checkout: add personal details

jamesgorrie:
Adds personal fields to the generic checkout.

There's been a few choices made, mostly around trying not to use redux state for things where there are native HTML alternative (hopefully reducing complexity).

- Introduce the `form` element to be able to hook into native HTML form handling, and work with documented standard APIs
  - [client side form validation](https://developer.mozilla.org/en-US/docs/Learn/Forms/Form_validation)
  - [`FormData`](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest_API/Using_FormData_Objects)
- Start to create the shape of the request we would like to send over to `create` endpoint based on the Zuora model. This currently sends the `product`, `ratePlan` and `currency`, which should allow us to look up the cost from there e.g. Screenshot 2024-02-27 at 12 12 03

I've added some context specific comments in the code below.

[**Trello Card**](https://trello.com/c/rurWegCp/628-add-personal-details-to-generic-checkout)

Part of #5722 

Next steps: [Add payment fields to the checkout](https://trello.com/c/qHy6BAvz/645-add-payment-fields-to-generic-checkout)

### Form validation
Screenshot 2024-02-27 at 12 12 16

Screenshot 2024-02-27 at 12 12 41

### DIFF
--- a/support-frontend/assets/components/personalDetails/personalDetails.tsx
+++ b/support-frontend/assets/components/personalDetails/personalDetails.tsx
@@ -87,6 +87,7 @@ export function PersonalDetails({
 					pattern={emailRegexPattern}
 					error={errors?.email?.[0]}
 					disabled={isSignedIn}
+					name="email"
 				/>
 			</div>
 
@@ -104,6 +105,7 @@ export function PersonalDetails({
 							autoCapitalize="words"
 							onChange={(e) => onFirstNameChange(e.target.value)}
 							error={errors?.firstName?.[0]}
+							name="firstName"
 							required
 						/>
 					</div>
@@ -117,6 +119,7 @@ export function PersonalDetails({
 							autoCapitalize="words"
 							onChange={(e) => onLastNameChange(e.target.value)}
 							error={errors?.lastName?.[0]}
+							name="lastName"
 							required
 						/>
 					</div>
--- a/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
@@ -6,7 +6,12 @@ import {
 	textSans,
 	until,
 } from '@guardian/source-foundations';
-import { Column, Columns, Container } from '@guardian/source-react-components';
+import {
+	Column,
+	Columns,
+	Container,
+	TextInput,
+} from '@guardian/source-react-components';
 import {
 	FooterLinks,
 	FooterWithContents,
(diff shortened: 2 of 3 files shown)


support-frontend/assets/components/personalDetails/personalDetails.tsx
```diff
@@ -87,6 +87,7 @@ export function PersonalDetails({
 					pattern={emailRegexPattern}
 					error={errors?.email?.[0]}
 					disabled={isSignedIn}
+					name="email"
```
jamesgorrie, line 90: By adding the `name` fields this becomes available on the `FormData` [when passing the form element through to it's constructor](https://developer.mozilla.org/en-US/docs/Web/API/FormData/FormData#prepopulating_from_a_html_form_element).

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -70,11 +83,15 @@ switch (countryGroupId) {
 		break;
 }
 
+const isSignedIn = !!get('GU_U');
+const countryId: IsoCountry =
+	CountryHelper.fromString(get('GU_country') ?? 'GB') ?? 'GB';
+
```
Comment (jamesgorrie) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:89]: This is stuff we would normally manage in state, but is static on page load, so this felt like the easiest way to do it.

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -70,11 +83,15 @@ switch (countryGroupId) {
 		break;
 }
 
+const isSignedIn = !!get('GU_U');
+const countryId: IsoCountry =
+	CountryHelper.fromString(get('GU_country') ?? 'GB') ?? 'GB';
```
Comment (jamesgorrie) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:88]: I thought we might be able to infer this from `countryGroupId` but we need the actual country in the `StateSelect`.

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -250,6 +270,100 @@ export function Checkout() {
...
+								console.info('Posting data', data);
+
+								// Currently support-workers doesn't understand the new shape, and we also need to
+								// add payment data, so for now, just abort.
+								return false;
+							}}
+						>
+							<input type="hidden" name="product" value={query.product} />
+							<input type="hidden" name="ratePlan" value={currentCurrencyKey} />
+							<input type="hidden" name="currency" value={currentCurrencyKey} />
```
Comment (jamesgorrie) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:303]: We could just read the `query[name]` here , but it felt like having one way to get the data from the form was consistent and maintained the use of standard API e.g. `input => FormData`

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -250,6 +270,100 @@ export function Checkout() {
...
+
+							<Box cssOverrides={shorterBoxMargin}>
+								<BoxContents>
+									<PersonalDetails
+										email={''}
+										firstName={''}
+										lastName={''}
```
Comment (jamesgorrie) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:310]: As state isn't managed we're all good to just have these blank.

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -250,6 +270,100 @@ export function Checkout() {
...
+										isSignedIn={isSignedIn}
+										hideNameFields={query.product === 'Contribution'}
+										onEmailChange={() => {
+											//  no-op
```
Comment (jamesgorrie) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:314]: Currently we don't need to do anything on the value changing as the form doesn't react to changes real time, but rather `onSubmit`.

Comment (rupertbates) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:314]: Is this going to change in future?

Comment (jamesgorrie) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:314]: We might have to with things like validation. I was hoping to raise pieces like this as separate PRs as something we wanted to explore as part of this work was finding areas we could simplify the CS code.

support-frontend/assets/components/personalDetails/personalDetails.tsx
paul-daniel-dempsey: Presume so we can use for example 'document.getElementById("email")'

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -250,6 +270,100 @@ export function Checkout() {
...
+													state={'STATE'}
+													onStateChange={() => {
+														//  no-op
+													}}
+													error={undefined}
+												/>
+											)
+										}
+										contributionZipcode={
+											countryId === 'US' ? (
```
Comment (paul-daniel-dempsey) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:337]: Would we have a differently validated postcode/zipcode checker dependent on the region in future?

paul-daniel-dempsey (approved):
Framework looks nice and neat for the new fields.

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
Comment (jamesgorrie) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:337]: Definitely - the next step is definitely going to start needing more state management, so we can approach this there.

PR #5746, opened 2024-02-27
3-tier: all one-off price card amounts should now be fed from RRCP, test2 bug-fix

paul-daniel-dempsey:

Bug-fix so one-off price card amounts in the 3-tier variant should be fed from RRCP here ->
image

- Bug introduced in Test2 PR ( https://github.com/guardian/support-frontend/pull/5709 }
- Test1 unaffected

|From|To|
|------|------|
|image|image|


jamesgorrie (approved):
If we want to rerun the test, great simple fix

PR #5758, opened 2024-03-01
Add stripe to generic checkout

jamesgorrie:
Adding stripe to the generic checkout. A side effect of this is that we will have the initial POST object for the `/create` endpoint.

This PR:
- Changes some page setup so we can also have `countryGroupId`. There is a lot of type and value proliferation across server and client side. I think we should aim to move the client closer to the server as it's simpler, but a large TS refactor.
- Adds the `StripCardForm` with `Recatcha`
- Captures and validates the Stripe data to post to `/create`

The shape of that object is:
```JSONC
{
    "firstName": "James",
    "lastName": "Gorrie",
    "email": "XXXX.XXXX@XXXXXXX.com",
    "product": "SupporterPlus",
    "ratePlan": "Monthly",
    // This is currency as that's what we call it elsewhere, but maybe charge is better?
    "currency": "GBP",
    "paymentFields": {
        "stripePaymentType": "StripeCheckout",
        "paymentMethod": "XXXX",
        "recaptchaToken": "XXXXXXXX"
    }
}
```

https://support.thegulocal.com/uk/checkout?product=SupporterPlus&ratePlan=Monthly

Capture-2024-03-06-115429

Part of https://github.com/guardian/support-frontend/issues/5722


support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -47,39 +74,49 @@ import { CheckoutDivider } from 'pages/supporter-plus-landing/components/checkou
...
-);
+const isTestUser = true;
+const geoIds = ['uk', 'us', 'eu', 'au', 'nz', 'ca', 'int'] as const;
+const GeoIdSchema = picklist(geoIds);
+const geoId = parse(GeoIdSchema, window.location.pathname.split('/')[1]);
 
 let currentCurrency: Currency;
 let currentCurrencyKey: keyof typeof currencies;
-switch (countryGroupId) {
+let countryGroupId: CountryGroupId;
```
Comment (jamesgorrie) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:84]: I've gone with `geo` and `countryGroup`, although we should look at syncing this up with the server side, which I think it clearer, and uses the URL part and the `countryGroupId`.

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -353,13 +455,86 @@ export function Checkout() {
...
+																	},
+																	body: JSON.stringify({
+																		isTestUser,
+																		stripePublicKey,
+																		token,
+																	}),
+																},
+															)
+																.then((resp) => resp.json())
+																.then((json) =>
```
Comment (jamesgorrie) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:518]: We could add some validation here, but that doesn't currently exist.

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -202,6 +232,22 @@ const shorterBoxMargin = css`
 	}
 `;
 
+const validPaymentMethods = [
+	query.product !== 'Contribution' && countryGroupId === 'EURCountries' && Sepa,
```
Comment (paul-daniel-dempsey) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:236]: Maybe check for AplePay (safari browser or IOS with ApplePay setup)

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -353,13 +440,81 @@ export function Checkout() {
...
+																		stripePublicKey,
+																		token,
+																	}),
+																},
+															)
+																.then((resp) => resp.json())
+																.then((json) =>
+																	setStripeClientSecret(
+																		(json as Record<string, string>)
+																			.client_secret,
```
Comment (paul-daniel-dempsey) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:498]: Can see this is the end-result state used above in form submit

support-frontend/assets/helpers/forms/paymentMethods.ts
```diff
@@ -40,6 +40,25 @@ export const recaptchaRequiredPaymentMethods: PaymentMethod[] = [
...
 
+const paymentMethods = [
+	Stripe,
+	PayPal,
+	DirectDebit,
+	Sepa,
+	ExistingCard,
+	ExistingDirectDebit,
+	AmazonPay,
+	None,
```
paul-daniel-dempsey, line 51: ApplePay here too?

paul-daniel-dempsey (approved):
Looks good.

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
Comment (jamesgorrie) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:236]: I'm going to bring this into another PR as it's a bit more complicated (added to #5722)

PR #5759, opened 2024-03-04
feat: add payment config to checkout

jamesgorrie:

- Copy payment config from `contributions.html.scala` `windowGuardianPaymentConfig.scala.html` that contains the `window.guardian` JS config needed for payments
- Reference this in `contributions.html.scala`
- Create new `checkout.scala.html` page with `@windowGuardianPaymentConfig`

This should be a no-op on existing pages, and make payment config available on the new checkout page.

Part of #5722


paul-daniel-dempsey (approved):
Looks good

PR #5760, opened 2024-03-04
feat: add productPrices to window.guardian

jamesgorrie:
This puts the productPrices, including promoCodes on the to the window.guardian.productPrices field, which we will then be able to access in the React app to use promoCodes for Supporter Plus in the UI.

[**Trello Card**](https://trello.com/c/5t4LIRvo/564-promos-on-supporterplus-part-2-attach-active-promotions-to-supporterplus-thresholds-model-on-windowguardian)

## Why are you doing this?

To allow people to use promoCodes for Supporter Plus

## Screenshots
This shows an active `promoCode` in the productPrices with `promoCode` set in the URL. 

Screenshot 2024-03-04 at 13 06 36

### DIFF
--- a/support-frontend/app/views/contributions.scala.html
+++ b/support-frontend/app/views/contributions.scala.html
@@ -8,7 +8,9 @@
 @import views.{Preload, SSRContent, ReactDiv}
 @import models.GeoData
 @import controllers.PaymentMethodConfigs
-
+@import com.gu.support.encoding.CustomCodecs._
+@import services.pricing.ProductPrices
+@import views.ViewHelpers.outputJson
 @(
   title: String,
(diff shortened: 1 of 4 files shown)


support-frontend/app/views/contributions.scala.html
```diff
@@ -8,7 +8,9 @@
 @import views.{Preload, SSRContent, ReactDiv}
 @import models.GeoData
 @import controllers.PaymentMethodConfigs
-
+@import com.gu.support.encoding.CustomCodecs._
```
paul-daniel-dempsey, line 11: Did we still need this custom encoding withthe call `outputJson` now?

PR #5762, opened 2024-03-04
feat: add window.guardian parsing

jamesgorrie:
Adding `window.guardian` validation to ensure the data from the server matches the type we have on the client.

This works similarly to how you might parse data from an API (and can be co-oped if that happens).

This was in reaction to the bug release in [this PR](https://github.com/guardian/support-frontend/pull/5759) and [fixed here](https://github.com/guardian/support-frontend/pull/5761) to add confidence that our changes won't break things (tests! 🧪). [Currently we just type coerce the window object](https://github.com/guardian/support-frontend/blob/main/support-frontend/window.d.ts#L39).

We only ever include `valibot` in dev and the error is currently in the `console`, [similarly to the axe errors in the render method](https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/helpers/rendering/render.ts#L60-L65).

I would have hoped to have made the error more explicit by using the [webpack overlay](https://webpack.js.org/configuration/dev-server/#overlay) - but after 2 hours of it really not playing dice for some very unknown reason, I thought I would get this PR out first. I know [we disabled it](https://github.com/guardian/support-frontend/pull/5196), but just re-enabling it didn't work and it flashes up and then disappears making me think that it's something in out webpack config which is a little fiddly to debug at the moment.

Adds +1.47 kB


support-frontend/assets/helpers/internationalisation/classes/country.ts
```diff
@@ -195,7 +195,7 @@ export default class CountryHelper {
 	}
 
 	static fromOldGeolocation(): IsoCountry | null | undefined {
-		return this.findIsoCountry(window.guardian.geoip?.countryCode);
+		return this.findIsoCountry(window.guardian.geoip.countryCode);
```
jamesgorrie, line 198: This was an eslint thing.

support-frontend/tsconfig.tsbuildinfo
jamesgorrie, line 1: I might see if I can remove this in another PR.

paul-daniel-dempsey (approved):
Looks good, just console logs payment configs that are missing in development, nifty!

PR #5763, opened 2024-03-04
add test that realistic requests can be deserialised

johnduffell:
After this PR: https://github.com/guardian/support-frontend/pull/5684
broke the parsing contract between client and server side.

I checked if the tests are missing one for S+, but it turned out there's no test for those deserialisers.

I decided to capture 24h worth of requests, then I designed a test to work on a sanitised and deduplicated version of the real data.

If I run the test with the faulty PR in, 15 tests fail.  If I run the test on main, it passes.


jamesgorrie (approved):
Great idea! Would adding a borked test add anything?

rupertbates (approved):
Great idea

PR #5765, opened 2024-03-05
use currency before country to determine stripe account to use

johnduffell:
A reader was having trouble doing a recurring contribution.
It turns out the client side uses the AU stripe account if country=AU, and the server side does the same if currency=AUD.

On the client side a country code is taken from the geolocation cookie [1] and if it's within the country group (e.g. "KI" is within AUDCountries from "/**au**/contribute") then it will use that as the country.  The currency to offer is also taken from the country group.

This means that readers in AUD countries other than AU, would use the ROW stripe account on the client side and the AUS stripe account on the server side.

Since the idea of having an AUS stripe account is to reduce the chance of international transaction fees, the transactions will still go through even if we get it wrong.  Generally we think the currency is a good proxy for whether the bank is an AUS based one.

This PR makes it check the currency rather than the country.  For the US stripe account, it checks the country also, to maintain the old behaviour for single contributions.  This is so specific US only cards can be accepted.

Single contributions has code to POST the stripe public key, and then look it up, rather than making the same decision on client and server side separately.  We aim to follow the same pattern in recurring in the next PR.

Tested locally against CODE support-workers as follows:
billing country: GB, delivery country: AU,  AUD  GuardianWeekly
billing country: AU, delivery country: GB,  GBP  GuardianWeekly
billing country: GB,  GBP  SupporterPlus
billing country: AU,  AUD  SupporterPlus
billing country: KI,  AUD  Contribution

[1] https://github.com/guardian/support-frontend/blob/5c86d13562fa74900241b0612a152e0225389e6c/support-frontend/assets/helpers/internationalisation/classes/country.ts#L234

### DIFF
--- a/support-frontend/assets/helpers/forms/stripe.ts
+++ b/support-frontend/assets/helpers/forms/stripe.ts
@@ -5,6 +5,7 @@ import type { ContributionType } from 'helpers/contributions';
 import type { PaymentMethod } from 'helpers/forms/paymentMethods';
 import { Stripe } from 'helpers/forms/paymentMethods';
 import type { IsoCountry } from 'helpers/internationalisation/country';
+import type { IsoCurrency } from '../internationalisation/currency';
 
 const stripeCardFormIsIncomplete = (
 	paymentMethod: PaymentMethod,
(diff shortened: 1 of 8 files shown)


support-frontend/assets/helpers/forms/stripe.ts
```diff
@@ -40,24 +41,25 @@ export interface StripeKey {
...
+  switch (currency) {
+		case 'AUD': // need to match with how support-workers does it
+      account = window.guardian.stripeKeyAustralia[stripeAccount];
+      break;
+		case 'USD':
+      if (country === 'US') { // this allows support of US only cards (for single)
+        account = window.guardian.stripeKeyUnitedStates[stripeAccount];
+      } else {
+        account = window.guardian.stripeKeyDefaultCurrencies[stripeAccount];
+      }
```
johnduffell, line 57: we tried to do this as a fallthough case, but that was red underlined by typescript, which is fair enough

tomrf1 (approved):
👍

PR #5766, opened 2024-03-06
refactor: remove tsconfig.tsbuildinfo

jamesgorrie:
Removes the `tsconfig.tsbuildinfo` file. [From the docs - it can be useful for large composite projects](https://www.typescriptlang.org/tsconfig#tsBuildInfoFile), which we are not.

The reason to remove it is that it changes when running TypeScript locally, making PRs messy, and then using time to have to refactor it back to the original state.

I will see if it adds any extra build time to our CI typescript compilation.

[Built in 59 seconds](https://github.com/guardian/support-frontend/actions/runs/8173253077/job/22345319720?pr=5766), which was one of the fastest:
Screenshot 2024-03-06 at 14 27 25


rupertbates (approved):

johnduffell (approved):
oh good I wondered why it kept wanting to commit that.  Do we need to add it to gitignore or anything, or is the compiler smart enough to not regenerate it?

johnduffell:
also since it's for incremental compilation, having that file would only speed things up locally, having it would slow things down if anything in the automated build (which I presume does a clean build every time?)

jamesgorrie:
> also since it's for incremental compilation, having that file would only speed things up locally

On your previous suggestion, I'll add it to .gitignore so it still gets generated, and can work locally.

> having it would slow things down if anything in the automated build

Our builds are ~1min and not having this file doesn't seem to have made a difference (see PRs description).

PR #5767, opened 2024-03-06
fix: remove switches from getValidPaymentMethods

jamesgorrie:
Removes an unused argument.

## Why are you doing this?
Makes the code simpler.


paul-daniel-dempsey (approved):
Looks good

PR #5768, opened 2024-03-06
make client side tell support-workers which stripe account to use

johnduffell:
Further to the quick fix https://github.com/guardian/support-frontend/pull/5765

Currently the client side makes a decision about which stripe account to use, for recurring, mainly based on the currency (since the above PR)
This works well as it matches how the support-workers does it.

However it's a common source of bugs where the stripe account has a mismatch, and during the work on the above PR we noticed that payment API receives the public key from the client side in order to select the account.

This PR updates our recurring products to send the key from the client side and then use that on the server side to look it up.

In this PR it is optional, and if it's missing it will use the old method. However I will monitor the logs, and once all the products are sending the key, I will remove the old currency based code and make the field mandatory.

I also removed the old "non SCA" stripe code, as that is unused.

- Tested in CODE
- put "US" config in prod support-workers config (as duplicate of default), also delete the unused ones - old AUD and old top level key

Once this PR is merged
- we can keep an eye out for "DEPRECATED FALLBACK" appearing in the CreatePaymentMethod lambda, to see if we missed any checkouts.
- we can delete the stripe config from parameter store (CODE and PROD) and do a followup PR to remove the allow `"unused"` line


build.sbt
```diff
@@ -11,7 +11,7 @@ lazy val integrationTestSettings: Seq[Def.Setting[_]] = Defaults.itSettings ++ S
   IntegrationTest / javaSource := baseDirectory.value / "src" / "test" / "java",
   IntegrationTest / resourceDirectory := baseDirectory.value / "src" / "test" / "resources",
   Test / testOptions += Tests
-    .Argument(TestFrameworks.ScalaTest, "-l", "com.gu.test.tags.annotations.IntegrationTest", "-oI"),
+    .Argument(TestFrameworks.ScalaTest, "-l", "com.gu.test.tags.annotations.IntegrationTest", "-oIS"),
```
johnduffell, line 14: this makes us see "short" stack traces in the tests, as opposed to no stack traces

support-frontend/app/utils/CheckoutValidationRules.scala
```diff
@@ -72,9 +72,7 @@ object CheckoutValidationRules {
       if (switches.directDebit.isOn) Valid else Invalid("Invalid Payment Method")
     case Left(_: SepaPaymentFields) =>
       if (switches.sepa.isOn) Valid else Invalid("Invalid Payment Method")
-    case Left(_: StripeSourcePaymentFields) =>
-      if (switches.stripe.isOn) Valid else Invalid("Invalid Payment Method")
-    case Left(s: StripePaymentMethodPaymentFields) =>
+    case Left(s: StripePaymentFields) =>
```
johnduffell, line 75: I got rid of `StripeSourcePaymentFields` as it was for the old non-SCA checkouts which don't exist.  I renamed StripePaymentMethodPaymentFields to StripePaymentFields as it's the only one now.

support-frontend/assets/components/subscriptionCheckouts/stripeForm/stripeProviderForCountry.tsx
```diff
@@ -33,6 +38,7 @@ function StripeProviderForCountry(props: PropTypes): JSX.Element {
 	);
 	useEffect(() => {
 		if (stripeObject === null) {
+			props.setStripePublicKey(stripeKey);
```
johnduffell, line 41: I noticed that the field existed in the redux state, but wasn't getting populated for GW/GW gift/paper so it was still empty string, as they were using this `useState` mechanism instead.  I have just threaded it through to the redux to set it there as well now.

support-frontend/assets/pages/supporter-plus-landing/setup/legacyActionCreators.ts
```diff
@@ -112,12 +112,13 @@ const buildStripeChargeDataFromAuthorisation = (
...
 		state.page.checkoutForm.billingAddress.fields.postCode,
 	),
-	publicKey: getStripeKey(
-		stripeAccountForContributionType[getContributionType(state)],
-		state.common.internationalisation.countryId,
-		state.common.internationalisation.currencyId,
-		state.page.user.isTestUser,
-	),
+	publicKey:
+		/* why don't we use state.page.checkoutForm.payment.stripeAccountDetails.publicKey ?*/ getStripeKey(
```
johnduffell, line 116: this is the mechanism that sends the key over to payment api.  I wasn't quite sure why we call getStripeKey again here (with hopefully the same parameters), rather than just using the value from the state?

support-frontend/test/utils/CheckoutValidationRulesTest.scala
```diff
@@ -498,8 +515,8 @@ class DigitalPackValidationTest extends AnyFlatSpec with Matchers {
   }
 
   it should "fail if the source payment field received is an empty string" in {
-    val requestMissingState = validDigitalPackRequest.copy(paymentFields = Left(StripeSourcePaymentFields("", None)))
-    DigitalPackValidation.passes(requestMissingState, monthlyDirectUSDProduct) shouldBe an[Invalid]
+    val requestMissingState = PaymentMethodId("")
+    requestMissingState shouldBe None
```
johnduffell, line 519: for the PaymentMethodId case class I have made it so you can't even construct one if it's not passing validation.  I.e. Parse don't validate.  It would be nice to move all the validation over to a similar parsing process, so that the type checker proves that it's been validated rather than the tests.

support-models/src/main/scala/com/gu/support/encoding/CustomCodecs.scala
```diff
@@ -65,7 +66,10 @@ trait EitherCodecs {
   implicit def decodeEither[A, B](implicit
       decoderA: Decoder[A],
       decoderB: Decoder[B],
-  ): Decoder[Either[A, B]] = decoderA.either(decoderB)
+  ): Decoder[Either[A, B]] = {
+    import CodecHelpers.withClue
+    CodecHelpers.or(withClue(decoderA.map(Left.apply), "Left"), withClue(decoderB.map(Right.apply), "Right"))
```
johnduffell, line 71: when I was debugging the tests I discovered that our `or` function (below) and the circe `either` function just reported the last error message without context, rather than everything that they tried.  This makes for a useless error message.
Although it would be a nice opportunity to use a [discriminator](https://circe.github.io/circe/codecs/adt.html#the-future) that would require extra fields on the client side.  I have just patched it up with a better error message for now.  It's a bit verbose in terms of "here are all the things I tried...." but that can't be helped without the client side tweaks.

support-models/src/test/scala/com/gu/support/SerialisationTestHelpers.scala
```diff
@@ -30,7 +30,7 @@ trait SerialisationTestHelpers extends Matchers {
       objectChecks: T => Assertion = (_: T) => succeed,
   ): Assertion =
     decodeResult.fold(
-      e => fail(e.getMessage),
+      e => throw new RuntimeException(e),
```
johnduffell, line 33: "fail" causes a TestFailedException, and scalatest is smart? enough to only print the line of stack trace where the fail call is.  This makes sense if it's in your actual test, but when it's in a helper function it's not enough information.

In conjuction with the -oS option to scalatest (see build.sbt above) this provides enough information to find out where in the actual test it failed.

support-workers/src/main/scala/com/gu/stripe/StripeService.scala
```diff
@@ -19,36 +20,41 @@ import com.gu.support.zuora.api.{
...
-  def withCurrency(currency: Currency): StripeServiceForCurrency = new StripeServiceForCurrency(this, currency)
-
-}
-
-object StripeServiceForCurrency {
-
-  def chargeGateway(currency: Currency): PaymentGateway =
-    if (currency == AUD) StripeGatewayAUD else StripeGatewayDefault
+  def withCurrency(currency: Currency): StripeServiceForCurrency = {
+    logger.warn(s"DEPRECATED FALLBACK - using currency $currency to determine stripe gateway")
```
johnduffell, line 40: this is here so I can just search the support-workers log quickly and find out all/any checkouts that are not yet sending the public key in the payload.

support-workers/src/main/scala/com/gu/stripe/StripeService.scala
```diff
@@ -62,18 +68,14 @@ class StripeServiceForCurrency(val stripeService: StripeService, currency: Curre
...
-      map + ("Stripe-Version" -> version)
-    }
-
-  private def getAuthorizationHeader() =
-    Map("Authorization" -> s"Bearer ${config.forCurrency(Some(currency)).secretKey}")
+  private def getHeaders(): Map[String, String] = {
+    val authorizationHeader = "Authorization" -> s"Bearer ${stripeAccountConfig.secretKey}"
+    val versionHeader = stripeService.config.version.map("Stripe-Version" -> _)
+    List(Some(authorizationHeader), versionHeader).flatten.toMap
+  }
```
johnduffell, line 75: this didn't really need to use a foldLeft on an option

support-workers/src/main/scala/com/gu/stripe/StripeService.scala
```diff
@@ -62,18 +68,14 @@ class StripeServiceForCurrency(val stripeService: StripeService, currency: Curre
...
 
   val createCustomerFromPaymentMethod = com.gu.stripe.createCustomerFromPaymentMethod.apply(this) _
 
-  val createCustomerFromToken = com.gu.stripe.createCustomerFromToken.apply(this) _
```
johnduffell, line 75: this is no longer needed as it's for the non-sca based stripe calls.

support-workers/src/test/scala/com/gu/support/workers/lambdas/CreatePaymentMethodSpec.scala
```diff
@@ -53,15 +53,18 @@ class CreatePaymentMethodSpec extends AsyncLambdaSpec with MockContext {
...
         // Check the output
         val createSalesforceContactState = Encoding.in[CreateSalesforceContactState](outStream.toInputStream)
 
         createSalesforceContactState.isSuccess should be(true)
         createSalesforceContactState.get._1.paymentMethod match {
           case Left(stripe: CreditCardReferenceTransaction) =>
-            stripe.TokenId should be("1234")
+            withClue("stripe: " + stripe) {
+              stripe.SecondTokenId should be("12345")
+              stripe.CreditCardNumber should be("1234")
```
johnduffell, line 66: we put different information into zuora for the new stripe gateway, so this test checks different things now.

support-models/src/main/scala/com/gu/support/workers/PaymentFields.scala
```diff
@@ -11,19 +11,34 @@ sealed trait PaymentFields
...
+        (char >= 'a' && char <= 'z') ||
+        (char >= 'A' && char <= 'Z') ||
+        (char >= '0' && char <= '9') ||
+        char == '_'
+      }
+    ) Right(new StripePublicKey(value))
+    else Left(s"StripePublicKey <$value> had invalid characters")
+
+  def get(value: String): StripePublicKey =
+    parse(value).left.map(new Throwable(_)).toTry.get
```
johnduffell, line 35: by converting to a try before doing get, you get the actual error message in the exception, rather than just a generic None.get exception

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreatePaymentMethod.scala
```diff
@@ -106,42 +105,32 @@ class CreatePaymentMethod(servicesProvider: ServiceProvider = ServiceProvider)
...
-            card.exp_year,
-            card.brand.zuoraCreditCardType,
-            PaymentGateway = paymentIntentGateway(currency),
-            StripePaymentType = stripePaymentType,
-          )
-        }
+    val stripeServiceForCurrency = {
+      stripe.stripePublicKey match {
+        case Some(stripePublicKey) => stripeService.withPublicKey(stripePublicKey)
+        case None => stripeService.withCurrency(currency)
```
johnduffell, line 111: still temporarily uses the withCurrency method in case one of the checkouts isn't updated yet!

support-models/src/main/scala/com/gu/support/workers/PaymentFields.scala
jamesgorrie: Could this comment exist next to the code for the next engineer that passes through?

jamesgorrie (approved):
Nice - the fallback is useful until we've completely moved over.

support-config/src/main/scala/com/gu/support/config/StripeConfig.scala
```diff
@@ -11,8 +13,19 @@ case class StripeConfig(
...
     version: Option[String],
 ) {
+  val secretForPublic: Map[StripePublicKey, (StripeAccountConfig, PaymentGateway)] =
+    List(
+      (defaultAccount, StripeGatewayPaymentIntentsDefault),
+      (australiaAccount, StripeGatewayPaymentIntentsAUD),
+      (unitedStatesAccount, StripeGatewayPaymentIntentsDefault), // US currently uses default account for recurring
+    ).map { config =>
+      config._1.publicKey -> config
+    }.toMap
```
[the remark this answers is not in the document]
johnduffell: more DRY - reduces the chance of a copy paste error? 😂 
(although I do take your point!  I'll revisit it and try to make a simpler option)
image

support-models/src/main/scala/com/gu/support/workers/PaymentFields.scala
```diff
@@ -32,14 +33,38 @@ object StripePublicKey {
...
-    parse(value).left.map(new Throwable(_)).toTry.get
+    parse(value).left.map(new Throwable(_)).toTry.get // Try.get will rethrow the useful error
 
   implicit val decoder: Decoder[StripePublicKey] = Decoder.decodeString.emap(parse)
   implicit val encoder: Encoder[StripePublicKey] = Encoder.encodeString.contramap[StripePublicKey](_.value)
 
 }
 case class StripePublicKey private (value: String) extends AnyVal
 
+object StripeSecretKey {
```
johnduffell, line 44: I've also added a wrapper for the secret key, this means it is also parsed at the point of loading and checked that the PK and SK are the right way around and formatted correctly.  Then the type system will ensure that they are used correctly after that.

support-models/src/main/scala/com/gu/support/workers/PaymentFields.scala
johnduffell: good idea, done that now

support-config/src/main/scala/com/gu/support/config/StripeConfig.scala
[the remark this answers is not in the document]
johnduffell: thanks I've basically done your suggestion in terms of not adding unnecessary logic, although I also realised that we only use the secret key in the StripeService, so I only return that instead.  I did also wrap it in a type rather than passing it around as a String.

support-config/src/main/resources/touchpoint.CODE.conf
```diff
@@ -3,21 +3,23 @@ touchpoint.backend.environments {
     stripe {
       AU {
         api.key.public = "pk_test_m0sjR1tGM22fpaz48csa49us"
-        api.key.secret = "dummy-key"
```
johnduffell, line 6: these dummy values were just there to get the tests to pass
Now the tests have their own reference.conf

support-config/src/main/resources/touchpoint.PROD.conf
```diff
@@ -3,21 +3,23 @@ touchpoint.backend.environments {
     stripe {
       AU {
         api.key.public = "pk_live_57xV50eFjPIA990PEGWJwoHp"
-        api.key.secret = "dummy-key"
       }
       US {
-        api.key.public = "pk_live_FCXf4hYCh6eefkHDbwKvvwZR00AFIyucU3"
-        api.key.secret = "dummy-key"
+        api.key.public = "pk_live_2O6zPMHXNs2AGea4bAmq5R7Z"
```
johnduffell, line 8: this was wrong, for recurring US we use the default key not the oneoff one.  This was getting ovverriden by the correct values that were in parameter store.
This file is correct now, and we can delete all the duplicated values from parameter store once this is shipped

support-frontend/app/controllers/StripeController.scala
```diff
@@ -62,8 +61,6 @@ class StripeController(
       val cloudwatchEvent = AwsCloudWatchMetricSetup.createSetupIntentRequest(stage, "v2Recaptcha")
       AwsCloudWatchMetricPut(client)(cloudwatchEvent)
 
-      val testPublicKeys = Set(testStripeConfig.australiaAccount.publicKey, testStripeConfig.defaultAccount.publicKey)
```
johnduffell, line 65: unused

support-frontend/app/services/stepfunctions/SupportWorkersClient.scala
```diff
@@ -96,7 +96,8 @@ class SupportWorkersClient(
     stateWrapper: StateWrapper,
     supportUrl: String,
     statusCall: String => Call,
-)(implicit system: ActorSystem) {
+)(implicit system: ActorSystem)
+    extends SafeLogging {
```
johnduffell, line 100: this was a merge conflict that crept in here somehow

support-frontend/app/views/subscriptionCheckout.scala.html
```diff
@@ -67,14 +66,14 @@
...
+          default: "@defaultStripeConfig.defaultAccount.rawPublicKey",
+          test: "@testStripeConfig.defaultAccount.rawPublicKey"
         }
       };
       window.guardian.stripeKeyAustralia = {
         REGULAR: {
-          default: "@defaultStripeConfig.forCurrency(Some(AUD)).publicKey",
-          test: "@testStripeConfig.forCurrency(Some(AUD)).publicKey"
+          default: "@defaultStripeConfig.australiaAccount.rawPublicKey",
+          test: "@testStripeConfig.australiaAccount.rawPublicKey"
```
johnduffell, line 76: we don't need to look these up by currency we can just use the member fields

support-workers/src/test/scala/com/gu/support/workers/EndToEndSpec.scala
```diff
@@ -25,7 +25,7 @@ class EndToEndSpec extends AsyncLambdaSpec with MockContext {
 
   def runSignupWithCurrency(currency: Currency): Future[Assertion] = {
     val json = createStripePaymentMethodContributionJson(currency = currency)
-    SafeLogger.info(json)
+    info(json)
```
johnduffell, line 28: another merge conflict that popped out here somehow

johnduffell:
I have deleted the DEV and CODE config and tested both locally and in CODE and AU/ROW recurring/single seem all to work fine.  I will delete all the PROD config tomorrow morning and redeploy, meaning it will be using the config from the config file rather than parameter store.

johnduffell:
this is done in PROD and after a redeploy everything seems fine so far

PR #5769, opened 2024-03-07
Add tier-3-button identifier for QM

GHaberis:

The CTA for Tier 3 on the 3-Tier Landing Page was missing the identifier used by Quantum Metric for click tracking.


paul-daniel-dempsey (approved):
Looks good, ah we must have missed tier3 QM indicator?

PR #5771, opened 2024-03-07
feat: add names to address fields

jamesgorrie:
In the efforts to be reusing our form fields on the #5722 - this adds the `name` attribute to allow reading of the values `onSubmit` of the `form` [as we are for other fields](https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/pages/%5BcountryGroupId%5D/checkout.tsx#L286-L293).

[**Trello Card**](https://trello.com/c/tUUScZqY/674-add-delivery-and-billing-address-fields-to-checkout)

This is a no-op as it's completely additive. I have tested all the critical flows via playwright locally.


paul-daniel-dempsey (approved):
looks fine, thanks

PR #5778, opened 2024-03-12
Skip thank you email on e2e test email address

jamesgorrie:
Stops E2E email addresses triggering thank you from from Braze.

[**Trello Card**](https://trello.com/c/S5wJ92zV/685-skip-e2e-thank-you-emails-sending)

## Why are you doing this?
Currently we [generate email address via uuidv4](https://github.com/guardian/support-frontend/commit/c6e124bb2c65bb9a5a6fd499cd427105e7806a0f) in our playwright e2e tests.

This has created a situation where Braze is trying to send emails to non-existant email addresses and they are bouncing.

This is compounded by the fact that the `SendThankYouEmail` sends emails via the `PROD` Braze account, even for test users as the [Lambda SQS queue is set on instantiation](https://github.com/guardian/support-frontend/blob/main/support-workers/src/main/scala/com/gu/support/workers/lambdas/SendThankYouEmail.scala#L25) where we have no context as to whether the user is test as that's stored in state.

Either way - the @guardian/value team are concerned that having these emails blasted out and bouncing from `CODE` or `PROD` might have a negative impact on our email reputation from email clients, so this just skips adding the email to queue all together.

## How to test

I've added and extended the integration test for the lambda, although I _think_ this doesn't run in CI. @johnduffell do you know?

Fewer bouncing emails sent with no debugging ability lost.

The check of `startsWith` or `endsWith` could potentially block valid emails, although I really don't think this is the case.

I have also used an email that is to a specific email address, so we could move to another option discussed of setting that up as an email group and just blasting emails there.

We are also not sure if this is an issue for identity.

## Screenshots

Screenshot 2024-03-12 at 10 29 15

### DIFF
--- a/support-e2e/tests/utils/users.ts
+++ b/support-e2e/tests/utils/users.ts
@@ -13,4 +13,10 @@ function generateString(length:number) {
 
 export const firstName = () => `${generateString(5)}TestF`;
 export const lastName = () => `${generateString(5)}TestL`;
-export const email = () => `test${uuidv4()}@theguardian.com`;
+/**
+ * This email is skipped when sending thank you emails from Braze in `support-workers`.
+ * If you change this here, you need to change it there too.
+ *
+ * see: support-workers/src/main/scala/com/gu/support/workers/lambdas/SendThankYouEmail.scala
+ */
+export const email = () => `test.e2e.supporter.revenue+${uuidv4()}@theguardian.com`;
--- a/support-workers/src/main/scala/com/gu/support/workers/lambdas/SendThankYouEmail.scala
+++ b/support-workers/src/main/scala/com/gu/support/workers/lambdas/SendThankYouEmail.scala
@@ -44,6 +44,7 @@ class SendThankYouEmail(
             response <- services.paperRoundService.agents()
           } yield response.data.agents.find(_.refId == id)
       }
+
     val emailBuilder = new EmailBuilder(
       services.zuoraService.getMandateIdFromAccountNumber,
       getAgentDetails,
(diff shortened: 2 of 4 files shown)


support-e2e/tests/utils/users.ts
```diff
@@ -13,4 +13,4 @@ function generateString(length:number) {
 
 export const firstName = () => `${generateString(5)}TestF`;
 export const lastName = () => `${generateString(5)}TestL`;
-export const email = () => `test${uuidv4()}@theguardian.com`;
+export const email = () => `test.e2e.supporter.revenue+${uuidv4()}@theguardian.com`;
```
tjmw, line 16: Agree that a clash with a genuine email seems very unlikely!

support-workers/src/main/scala/com/gu/support/workers/lambdas/SendThankYouEmail.scala
```diff
@@ -44,19 +44,32 @@ class SendThankYouEmail(
...
       getAgentDetails,
       services.promotionService,
       Configuration.stage,
     )
 
-    for {
-      emailFields <- emailBuilder.buildEmail(state)
-      emailResult <- Future.sequence(emailFields.map(emailService.send))
-    } yield HandlerResult(emailResult, requestInfo)
+    /** We generate these users `testGuardianDoNotUse` as part of our e2e testing and don't want to create a braze user
```
tjmw, line 55: What does `testGuardianDoNotUse` refer to here? Is/was that a pattern we were going to use for the email?

support-workers/src/main/scala/com/gu/support/workers/lambdas/SendThankYouEmail.scala
```diff
@@ -44,19 +44,32 @@ class SendThankYouEmail(
...
+      * for them as if the emails bounce it might affect our reputation with email clients in the long run.
+      *
+      * see: support-e2e/tests/utils/users.ts
+      *
+      * We cannot use `isTestUser` as we do actually use this for testing receiving of emails and it _might_ still have
+      * a negative affect on the reputation as email clients are not aware it is a test account.
+      */
+    val email = state.user.primaryEmailAddress
+    if (email.startsWith("test.e2e.supporter.revenue") && email.endsWith("theguardian.com")) {
```
tjmw, line 64: Would it be worth extracting this pattern/prefix out to a val somewhere so it's used both by the check and the code which generates the email address? That way it'd be impossible for them to get out of sync.

tjmw (approved):
Looks really good! I've left a couple of comments/questions, see what you think (not blockers though).

support-e2e/tests/utils/users.ts
jamesgorrie: I kept the `test` prefix too as I wondered if there might be some check somewhere that uses it.

support-workers/src/main/scala/com/gu/support/workers/lambdas/SendThankYouEmail.scala
jamesgorrie: My bad - it was the old address I was using that I didn't like the look of. Changed.

support-workers/src/main/scala/com/gu/support/workers/lambdas/SendThankYouEmail.scala
jamesgorrie: I've linked to the file above and can't think of a way of sharing that between projects that doesn't add a tonne of complexity that hopefully the comment mostly addresses. 

To this end, I have added a comment to the `users.ts` file too.

johnduffell:
can we just change the e2e tests to use thegulocal.com?  This is the same as the old "gu.com" that we used to use.  Then they will all be filtered out in membership-workflow, including any other service emails that happen later for any reason.

https://github.com/guardian/membership-workflow/blob/99e2b90305f93bf35ce230f6b6c17e0c4533facb/app/model/BrazeCampaignTriggerPayload.scala#L28

johnduffell:
> I've added and extended the integration test for the lambda, although I think this doesn't run in CI. @johnduffell do you know?

the integration tests do run every hour, they are the ones that come up with the IT tests are failing alarm.

jamesgorrie:
Closed in favour of #5779

PR #5779, opened 2024-03-12
refactor: change e2e email to thegulocal.com

jamesgorrie:
Changing the email for our e2e users so emails aren't triggered from Braze.

This is to avoid blasting email addresses that bounce might potentially have a negative impact on our email sending reputation.

More context can be found here:
https://github.com/guardian/support-frontend/pull/5778

[**Trello Card**](https://trello.com/c/S5wJ92zV/685-skip-e2e-thank-you-emails-sending)


support-e2e/tests/utils/users.ts
```diff
@@ -1,16 +1,27 @@
...
+ * This email needs to end with @thegulocal.com.
+ *
+ * Those are often skipped by things like our Braze helpers to avoid
+ * us spamming non-existant email addresses which might hurt our
+ * reputation with email clients.
+ *
+ * e.g: https://github.com/guardian/membership-workflow/blob/99e2b90305f93bf35ce230f6b6c17e0c4533facb/app/model/BrazeCampaignTriggerPayload.scala#L28
+ **/
+export const email = () =>
+  `test.e2e.supporter.revenue${uuidv4()}@thegulocal.com`;
```
jamesgorrie, line 27: I've kept it this so we at least have an indicator as to where they have come from.

support-e2e/tests/utils/users.ts
```diff
@@ -1,16 +1,27 @@
...
+  return result;
 }
 
 export const firstName = () => `${generateString(5)}TestF`;
 export const lastName = () => `${generateString(5)}TestL`;
-export const email = () => `test${uuidv4()}@theguardian.com`;
+/**
+ * This email needs to end with @thegulocal.com.
+ *
+ * Those are often skipped by things like our Braze helpers to avoid
```
rupertbates, line 20:
```suggestion
 * Those are skipped by membership-workflow to avoid
```

support-e2e/tests/utils/users.ts
```diff
@@ -1,16 +1,27 @@
...
+ * us spamming non-existant email addresses which might hurt our
```
rupertbates, line 21:
```suggestion
 * Those are skipped by membership-workflow to avoid
 * us spamming non-existent email addresses which might hurt our
```

rupertbates (approved):
One tiny suggestion

support-e2e/tests/utils/users.ts
```diff
@@ -1,16 +1,27 @@
...
+  let result = "";
+  const charactersLength = characters.length;
+  for (let i = 0; i < length; i++) {
+    result += characters.charAt(Math.floor(Math.random() * charactersLength));
+  }
+  return result;
 }
 
 export const firstName = () => `${generateString(5)}TestF`;
 export const lastName = () => `${generateString(5)}TestL`;
```
johnduffell, line 16: unrelated to this PR, but it's sometimes nicer to have the fixed part at the beginning to aid ordering

support-e2e/tests/utils/users.ts
```diff
@@ -1,16 +1,27 @@
...
-export const email = () => `test${uuidv4()}@theguardian.com`;
+/**
+ * This email needs to end with @thegulocal.com.
+ *
+ * Those are skipped by membership-workflow to avoid
+ * us spamming non-existent email addresses which might hurt our
+ * reputation with email clients.
+ *
+ * e.g: https://github.com/guardian/membership-workflow/blob/99e2b90305f93bf35ce230f6b6c17e0c4533facb/app/model/BrazeCampaignTriggerPayload.scala#L28
```
johnduffell, line 24: minor thing but it's not really an example, that's the only place it's referenced as far as I know
```suggestion
 * See https://github.com/guardian/membership-workflow/blob/99e2b90305f93bf35ce230f6b6c17e0c4533facb/app/model/BrazeCampaignTriggerPayload.scala#L28
```

johnduffell (approved):
looks good nice one James

support-e2e/tests/utils/users.ts
jamesgorrie: Good point, it was because the previous comment was an example - thanks for catching!
(changed it to [@see](https://jsdoc.app/tags-see))

support-e2e/tests/utils/users.ts
jamesgorrie: Great suggestion - [what do you think of this](https://github.com/guardian/support-frontend/pull/5782)?

GHaberis: This change has meant we're generating email addresses that are greater than 80 characters long for the GW Gift tests as they also prepend 'giftee' to the email address generated by this function. 80 characters is the maximum email address length accepted by SalesForce, so the e2e tests are failing the GW Gift tests as the `CreateSalesforceContact` step in the `Support-Workers-Prod` step functions is now failing.

PR #5780, opened 2024-03-12
Stop showing top-tier TY page to standard GW checkouts

GHaberis:

`threeTierCheckoutEnabled` was returning `true` 

Instead of using this on GW checkout pages we can use the presence of the `threeTierCreateSupporterPlusSubscription` query string param as indication they're on the top-tier checkout as we no longer use the participation in a 3-tier AB test.


paul-daniel-dempsey (approved):
Yes, as per Tier3 get from the URL. Looks good.

PR #5782, opened 2024-03-12
fix: prefix names with test info in e2e

jamesgorrie:
[As per @johnduffell's good suggestion](https://github.com/guardian/support-frontend/pull/5779/files#r1521347342), let's prefix the names to help for ordering and search.

I've tried to keep the data in the name that might allow people to identify where this test user came from.

Related to #5779


johnduffell (approved):
yeah good way to test the string length limit too! 👍 
thanks for followup with a PR for this one

GHaberis:
@jamesgorrie @johnduffell this change is also causing the e2e tests to fail client side validation as the firstName and lastName are now too long

PR #5783, opened 2024-03-12
Fix threeTierCheckoutEnabled conditional check on GW checkouts

GHaberis:

We call `threeTierCheckoutEnabled` from [submit.ts](https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/helpers/subscriptionsForms/submit.ts#L246 ) when building the payload to send to the `/create` endpoint. This logic is shared between the `/contribute` checkout and the `/subscribe` checkouts.

However in the standard GW checkout `threeTierCheckoutEnabled` was returning `true` (as the checks in `threeTierCheckoutEnabled` were only relevant to the `/contribute`  checkout, the consequence was we were setting `threeTierCreateSupporterPlusSubscription: true` in the payload for standard non 3-tier checkouts.

To resolve this I've made `threeTierCheckoutEnabled` work so it'll only return true for users on the GW checkout when the `threeTierCreateSupporterPlusSubscription` is present.

### DIFF
--- a/support-frontend/assets/pages/supporter-plus-landing/setup/threeTierChecks.ts
+++ b/support-frontend/assets/pages/supporter-plus-landing/setup/threeTierChecks.ts
@@ -6,6 +6,14 @@ export const threeTierCheckoutEnabled = (
 	abParticipations: Participations,
 	countryId: IsoCountry,
 ): boolean => {
+	const isWeeklyCheckout =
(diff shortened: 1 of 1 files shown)


jamesgorrie (approved):
Nice, makes it clear where this logic is actually used!

johnduffell:
does this mean that we need to go back and find everyone that tried to take out GW only recently and cancel their supporter plus?

PR #5777, opened 2024-03-11
Generic checkout  - Billing fields

jamesgorrie:
**Depends on https://github.com/guardian/support-frontend/pull/5775**

## What are you doing in this PR?
Adding the billing address fields to the generic checkout (#5722).

This adds the control as to whether to use the `billingAddress` fields over the `deliveryAddress` fields (`billingAddressMatchesDelivery`).

Mostly copy/pasted. 

[**Trello Card**](https://trello.com/c/tUUScZqY/674-add-delivery-and-billing-address-fields-to-checkout)

## Screenshots

https://github.com/guardian/support-frontend/assets/31692/b66aa54d-576c-40cf-9a38-73eb5d605896

URL: https://support.thegulocal.com/uk/checkout?product=HomeDelivery&ratePlan=Everyday

Depends on https://github.com/guardian/support-frontend/pull/5775
Part of https://github.com/guardian/support-frontend/issues/5722

PR #5787, opened 2024-03-12
feat: directdebit in checkout

jamesgorrie:
Depends on #5777 

Adds the direct debit to the checkout.

- Adds the `DirectDebit` components
- Adds the `DirectDebit` component state
- Adds the `DirectDebit` `onSubmit` handler

[**Trello Card**](https://trello.com/c/CwcLwSxI/686-add-direct-debit-to-checkout)

## Screenshots

Screenshot 2024-03-12 at 15 53 24


support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -750,6 +777,66 @@ export function Checkout() {
...
+											) => {
+												setAccountHolderConfirmation(confirmation);
+											}}
+											recaptcha={
+												<Recaptcha
+													// We could change the parents type to Promise and uses await here, but that has
+													// a lot of refactoring with not too much gain
+													onRecaptchaCompleted={(token) => {
+														setRecaptchaToken(token);
+														void fetch(
```
Comment (GHaberis) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:809]: I don't think we need to make this `fetch` in the `onRecaptchaCompleted` function for Direct Debit, this is only necessary for Credit/Debit card (eg. w/ Stripe)

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -750,6 +777,66 @@ export function Checkout() {
...
+															.then((resp) => resp.json())
+															.then((json) =>
+																setStripeClientSecret(
+																	(json as Record<string, string>)
+																		.client_secret,
+																),
+															);
+													}}
+													onRecaptchaExpired={() => {
+														// no-op
```
Comment (GHaberis) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:832]: Eventually we'll want to call `setRecaptchaToken` here and (un)set the `recaptchaToken` when it expires.

paul-daniel-dempsey (commented):
Looks similar in setup to credit card/ stripe. George has commented whether re:recaptcha used for direct debit and the expiry? It does appears on the checkout in production though?

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
Comment (jamesgorrie) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:777]: Screenshot 2024-03-28 at 09 56 53

When the post is made in `PROD` - this value is set, so I am just replicating that. [This is validated on the server](https://github.com/guardian/support-frontend/blob/main/support-frontend/app/controllers/CreateSubscriptionController.scala#L81-L87), so we do need to send it across.

jamesgorrie:
> It does appears on the checkout in production though?

I think @GHaberis's concern was with setting the async token rather than the Recaptcha showing, which you're right is in both.

[I've replied](https://github.com/guardian/support-frontend/pull/5787#discussion_r1542628369) saying I think we do need to set it as we have server side validation on ReCaptcha on the server.

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
Comment (GHaberis) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:777]: Do we make a `POST` to `/stripe/create-setup-intent/recaptcha` in the `onRecaptchaCompleted` for the Direct Debit form? I can't see that in the network requests, I only see it when checking the Recpatcha in the Credit/Debit card section of the form.

For the current Recaptcha in the Direct Debit section of the form we define `onRecaptchaCompleted` [here](https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/components/directDebit/directDebitForm/directDebitFormContainer.tsx#L65), this in turn calls the Recaptcha [reducer](https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/helpers/redux/checkout/recaptcha/reducer.ts#L12) which sets the captcha token in state, this is then sent to the server in the `/create` payload for validation.

Comment (jamesgorrie) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:775]: Got you - I've added the second step of using the `recaptchToken` to get a `stripe-intent`, which is a silly copy/pasta.

This has been removed.

GHaberis (approved):
Looks good!

PR #5788, opened 2024-03-13
Check promoCode before triggering 3 tier step functions

jamesgorrie:
Ensure we check that the right and corresponding `promoCode` is attached to a request before going down this decision tree.

This has been done in response to us having accidentally set `threeTierCreateSupporterPlusSubscription` in the request, but the `promoCode` was something else, and so we created a subscription for that reader in error[^1]

## Why are you doing this?
To ensure we don't accidentally create subscriptions for readers who didn't ask for them.

## How to test
The request made on `PROD` was recreated on `CODE` the 3-tier steps were not to be triggered.

### With non-matching `promoCode`
Screenshot 2024-03-13 at 11 30 29

### With matching `promoCode`
Screenshot 2024-03-13 at 11 34 10

No more un-asked for subscriptions are made.

[^1]: And we have reimbursed them - thanks @johnduffell!


rupertbates (approved):
I do think a lambda might be the way to go in future!

jamesgorrie:
> I do think a lambda might be the way to go in future!

Definitely agree. I think the lesson is that combining lambda + Step function transforms is way more error prone.

PR #5789, opened 2024-03-13
Change prefix format so email addresses generated by e2e tests pass validation

GHaberis:

We made change in https://github.com/guardian/support-frontend/pull/5779 that meant we started generating email addresses that are greater than 80 characters long for the GW Gift tests as these tests prepend 'giftee-' to the email address generated by the `email()` function. 

80 characters is the maximum email address length accepted by SalesForce, so the e2e tests are failing the GW Gift tests as the `CreateSalesforceContact` step in the `Support-Workers-Prod` step functions is now failing Salesforce's validation.

### DIFF
--- a/support-e2e/tests/utils/users.ts
+++ b/support-e2e/tests/utils/users.ts
@@ -12,10 +12,8 @@ function generateString(length: number) {
   return result;
 }
 
-export const firstName = () =>
-  `test.support.frontend.e2e.firstName+${generateString(5)}`;
-export const lastName = () =>
-  `test.support.frontend.e2e.lastName+${generateString(5)}`;
+export const firstName = () => `e2e.firstName+${generateString(5)}`;
+export const lastName = () => `e2e.lastName+${generateString(5)}`;
 /**
  * This email needs to end with @thegulocal.com.
  *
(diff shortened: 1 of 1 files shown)


support-e2e/tests/utils/users.ts
```diff
@@ -25,5 +23,4 @@ export const lastName = () =>
  *
  * @see https://github.com/guardian/membership-workflow/blob/99e2b90305f93bf35ce230f6b6c17e0c4533facb/app/model/BrazeCampaignTriggerPayload.scala#L28
  **/
-export const email = () =>
-  `test.support.frontend.e2e+${uuidv4()}@thegulocal.com`;
+export const email = () => `e2e.email+${uuidv4()}@thegulocal.com`;
```
jamesgorrie, line 26: The reason support-frontend is there is so we know where these tests have come from (using a new engineer as a use case).

Maybe we could just use something like https://github.com/ai/nanoid/?

GHaberis: Ok cool, yeah nice suggestion, will look at that now

GHaberis: I've updated the tests to use the `nanoid` library instead of `uuid`. One thing to note is I'm using `nanoid@v3` instead of  the latest `nanoid@v5`. I initially went to `nanoid@v5` but ran into an issue as `nanoid@v5` works only with ESM projects, in tests or Node.js scripts - and Playwright has issues working with ESM modules, this issue in Github neatly summarises the issues I was running into with PlayWright and ESM: https://github.com/microsoft/playwright/issues/23662

For CommonJS you need `nanoid@v3` (this is outlined in Nano's [docs](https://github.com/ai/nanoid?tab=readme-ov-file#install)), so I've switched to this and it works as desired.

johnduffell: I won't suggest to change it now but I'm sure time in milliseconds since epoch would have been good enough and I'm sure you don't need a library.
Having a rough idea of when it was created is probably not important but it would be a side effect too.

johnduffell (approved):
nice one, let's see if it fixes it! 👍

PR #5792, opened 2024-03-13
update support-workers to java21 runtime

johnduffell:
This PR bumps support workers to java 21.  This is supposed to give us performance improvements.
https://aws.amazon.com/blogs/compute/java-17-runtime-now-available-on-aws-lambda/
https://aws.amazon.com/about-aws/whats-new/2023/11/aws-lambda-support-java-21/

We are using scala 2.13.12 which is compatible with JRE21
https://docs.scala-lang.org/overviews/jdk-compatibility/overview.html

I have tested in CODE and looking at one lambda (create payment method) you can clearly see the performance improvement for the cold start
image

This is replicated for the whole state machine execution, although it is diluted by the different paths it can take through the state machine.
image

### DIFF
--- a/support-workers/cloud-formation/src/templates/lambda.yaml
+++ b/support-workers/cloud-formation/src/templates/lambda.yaml
@@ -10,5 +10,5 @@
     Code:
       S3Bucket: support-workers-dist
       S3Key: !Sub support/${Stage}/support-workers/support-workers.jar
-    Runtime: "java11"
+    Runtime: "java21"
     Timeout: "60"


tomrf1 (approved):
👏

johnduffell:
in PROD you can see the improvement in cold starts, 
image
although there's one outlier which was quicker just before the deploy.  This does sometimes happen as you can see also on the extreme left of the graph.

johnduffell:
still looking good overnight 
image

PR #5793, opened 2024-03-13
Skip e2e tests for /contributions checkout

GHaberis:

These tests were written to test the old 2-step (tabbed) checkout page 1, they'll fail with the new tiered landing page which has gone to 100%. I propose we skip them for now, get the tests working and posting to the SR Dev channel and then we rewrite these.


jamesgorrie (approved):

paul-daniel-dempsey (approved):
skip looks good

PR #5798, opened 2024-03-14
Update subscription landing page e2e tests to verify presence of expected HTML element

GHaberis:

It doesn't look like the e2e tests in `landingPages.test.ts` were actually testing anything.  This line in the tests `await page.locator("id=qa-paper-subscriptions").isVisible();`, would eventually resolve with `true` (element present) or `false` (element not present) however we didn't do anything with the return value of `isVisible`, so for example I could update the test to look for an element that doesn't exist `await page.locator("id=hello-world").isVisible();` and the test would still pass. The only expectation was to confirm the URL matched what we'd specified it is in the test.

### DIFF
--- a/support-e2e/tests/landingPages.test.ts
+++ b/support-e2e/tests/landingPages.test.ts
@@ -16,7 +16,10 @@ test.describe("Paper product page", () => {
     const domain = new URL(pageUrl).hostname;
     await setTestCookies(context, firstName(), domain);
     await page.goto(pageUrl);
-    await page.locator("id=qa-paper-subscriptions").isVisible();
+    const pageRendered = await page
+      .locator("id=qa-paper-subscriptions")
+      .isVisible();
+    expect(pageRendered).toBeTruthy();
     await expect(page).toHaveURL(/\/uk\/subscribe\/paper/);
   });
 });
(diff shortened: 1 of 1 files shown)


support-e2e/tests/landingPages.test.ts
```diff
@@ -16,7 +16,10 @@ test.describe("Paper product page", () => {
     const domain = new URL(pageUrl).hostname;
     await setTestCookies(context, firstName(), domain);
     await page.goto(pageUrl);
-    await page.locator("id=qa-paper-subscriptions").isVisible();
+    const pageRendered = await page
+      .locator("id=qa-paper-subscriptions")
+      .isVisible();
+    expect(pageRendered).toBeTruthy();
```
jamesgorrie, line 22: An alternative way I've seen this written is
```suggestion
    expect(page.locator("id=qa-paper-subscriptions")).toBeVisible();
```

jamesgorrie (approved):

jamesgorrie (approved):
Looks like the [prettier PR](https://github.com/guardian/support-frontend/pull/5807) created conflicts.

jamesgorrie:
Duplicate of https://github.com/guardian/support-frontend/pull/5907

PR #5799, opened 2024-03-14
Reduce retry count in Playwright config

GHaberis:

We have set up our Playwright config to retry a failing test up to 3 times. This means if the tests fai the time taken to test & retry 3 times exceeds Github Action’s maximum execution time of 30 minutes, the consequence of this is the Action won't post about the failed test to Google Chat, which is the most visible way to notify engineers to learn the e2e Action has failed.

I therefore propose in the PR to reduce the retry count from 3 to 1. 

PRO: The Action is more likely to complete within 30 mins, if it fails we learn about a failed test quickly via the Action posting to Google Chat.

CON: We may fail tests on transient issues which may have been resolved after multiple retries? I think it's worth testing with a lower retry count and seeing if this is an issue.

### DIFF
--- a/support-e2e/.prettierrc.js
+++ b/support-e2e/.prettierrc.js
@@ -1,0 +1,3 @@
+module.exports = {
+	...require('@guardian/prettier'),
+};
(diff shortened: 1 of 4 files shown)


paul-daniel-dempsey (approved):
Looks good, ready to test reduced runs in production, presume ok to replace single with double quotes at same time.

PR #5801, opened 2024-03-14
make support-frontend use java 21

johnduffell:
Since updating support-workers to java 21, the cold start time has improved a lot.

This PR updates support-frontend to java 21 by using the new AMI that I created and baked
https://amigo.gutools.co.uk/recipes/jammy-membership-java21

This is as per the new recommendation in https://docs.google.com/document/d/1ZR-YnaXCT5_gLVmTCeGs0mWd3KPaAozPjQK8uUzHZ9w/edit

Tested in CODE and working to buy a middle tier (supporter plus)

### DIFF
--- a/support-frontend/build.sbt
+++ b/support-frontend/build.sbt
@@ -77,13 +77,17 @@ def getFiles(rootFile: File, deployName: String): Seq[(File, String)] = {
   getFiles0(rootFile)
 }
 
-Universal / javaOptions ++= Seq(
-  "-Dpidfile.path=/dev/null",
-  "-J-XX:MaxMetaspaceSize=256m",
-  "-J-XX:+PrintGCDetails",
-  s"-J-Xloggc:/var/log/${packageName.value}/gc.log",
+val jvmParameters = Def.setting(Seq(
+  "-XX:MaxMetaspaceSize=256m",
+  s"-Xlog:gc*:/var/log/${packageName.value}/gc.log", // https://docs.azul.com/prime/Unified-GC-Logging#enabling-unified-gc-logging
   "-XX:-OmitStackTraceInFastThrow",
+))
+val playParameters = Seq(
+  "-Dpidfile.path=/dev/null", // https://www.playframework.com/documentation/3.0.x/ProductionConfiguration#Changing-the-path-of-RUNNING_PID
 )
+// -J tells the packager to pass it through to the JVM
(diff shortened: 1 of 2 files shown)


support-frontend/build.sbt
```diff
@@ -77,13 +77,17 @@ def getFiles(rootFile: File, deployName: String): Seq[(File, String)] = {
...
+  "-XX:MaxMetaspaceSize=256m",
+  s"-Xlog:gc*:/var/log/${packageName.value}/gc.log", // https://docs.azul.com/prime/Unified-GC-Logging#enabling-unified-gc-logging
   "-XX:-OmitStackTraceInFastThrow",
+))
+val playParameters = Seq(
+  "-Dpidfile.path=/dev/null", // https://www.playframework.com/documentation/3.0.x/ProductionConfiguration#Changing-the-path-of-RUNNING_PID
 )
+// -J tells the packager to pass it through to the JVM
+// https://www.scala-sbt.org/sbt-native-packager/archetypes/java_app/customize.html#via-build-sbt
+Universal / javaOptions ++= playParameters ++ jvmParameters.value.map(param => "-J" + param)
```
johnduffell, line 90: I have tweaked this structure because it wasn't clear that you have to prepend `-J` in order to supply JVM parameters.  This means that the `-XX:-OmitStackTraceInFastThrow` was being passed to play (which ignored it) rather than the JVM.  The -D argument is actually a play parameter, so it's in a separate list.

johnduffell: I have tweaked this structure because it wasn't clear that [you have to prepend -J in order to supply JVM parameters](https://www.scala-sbt.org/sbt-native-packager/archetypes/java_app/customize.html#via-build-sbt). This means that the -XX:-OmitStackTraceInFastThrow was being passed to play (which actually seems to pass it on to the JVM anyway) rather than the JVM. The -D argument is actually a play parameter, so it's in a separate list.  But again, it does get passed to the JVM anyway.

I can tell this is (still) working by checking the command line to java.

image

Although as I say it was actually present before:
image

support-frontend/conf/riff-raff.yaml
```diff
@@ -21,7 +21,7 @@ deployments:
         PROD: Frontend-PROD.template.json
       amiParameter: AMIFrontend
       amiTags:
-        Recipe: jammy-membership-java11
+        Recipe: jammy-membership-java21
```
johnduffell, line 24: https://amigo.gutools.co.uk/recipes/jammy-membership-java21

PR #5803, opened 2024-03-14
Three tier: S+ promoCode non-visual optimisations PT2

paul-daniel-dempsey:

PT2 : Code changes required to prepare for the PromoCode PR, 

No visual changes are made in this PR

- Improved isSupporterPlus, Ts&Cs, TierPlan function optimisations/naming
- contributionOrderSummary total can (optionally) display crossed out originalPrice
- contributionOrderSummaryContainer can (optionally) retrieve and pass onto contributionOrderSummary the originalPrice (alongside CurrentPrice)
- redux S+ amounts can be stored as fractional numbers (likely to be generated by promoCodes in future)
- getProductPrice try-catch wrapper to ensures errors returned are undefined as getPromotion function specifies (future storybook support without PromoCodes)

[Trello Card](https://trello.com/c/5t4LIRvo/564-promos-on-supporterplus-part-2-attach-active-promotions-to-supporterplus-thresholds-model-on-windowguardian)

### DIFF
--- a/support-frontend/assets/components/orderSummary/contributionsOrderSummaryContainer.tsx
+++ b/support-frontend/assets/components/orderSummary/contributionsOrderSummaryContainer.tsx
@@ -1,6 +1,20 @@
 import { checkListData } from 'components/checkoutBenefits/checkoutBenefitsListData';
 import type { ContributionType } from 'helpers/contributions';
+import type { IsoCountry } from 'helpers/internationalisation/country';
 import { currencies } from 'helpers/internationalisation/currency';
+import type { BillingPeriod } from 'helpers/productPrice/billingPeriods';
+import {
+	type FulfilmentOptions,
+	NoFulfilmentOptions,
+} from 'helpers/productPrice/fulfilmentOptions';
+import {
+	NoProductOptions,
+	type ProductOptions,
+} from 'helpers/productPrice/productOptions';
+import {
+	getCountryGroup,
+	type ProductPrices,
+} from 'helpers/productPrice/productPrices';
 import { isSupporterPlusFromState } from 'helpers/redux/checkout/product/selectors/isSupporterPlus';
 import { getContributionType } from 'helpers/redux/checkout/product/selectors/productType';
 import { getUserSelectedAmount } from 'helpers/redux/checkout/product/selectors/selectedAmount';
@@ -45,10 +59,38 @@ export function ContributionsOrderSummaryContainer({
 }: ContributionsOrderSummaryContainerProps): JSX.Element {
 	const contributionType = useContributionsSelector(getContributionType);
 
-	const { currencyId } = useContributionsSelector(
+	const { countryId, currencyId } = useContributionsSelector(
 		(state) => state.common.internationalisation,
 	);
-	const selectedAmount = useContributionsSelector(getUserSelectedAmount);
+	const { productType } = useContributionsSelector(
+		(state) => state.page.checkoutForm.product,
+	);
+	const billingPeriod = (productType[0] +
+		productType.slice(1).toLowerCase()) as BillingPeriod;
+	const currentPrice = useContributionsSelector(getUserSelectedAmount);
+	const originalPrice = useContributionsSelector((state) =>
+		getOriginalPrice(
+			state.page.checkoutForm.product.productPrices,
+			countryId,
+			billingPeriod,
+		),
+	);
+
+	function getOriginalPrice(
+		productPrices: ProductPrices,
+		country: IsoCountry,
+		billingPeriod: BillingPeriod,
+		fulfilmentOption: FulfilmentOptions = NoFulfilmentOptions,
+		productOption: ProductOptions = NoProductOptions,
+	): number | undefined {
+		const countryGroup = getCountryGroup(country);
+		const originalPrice =
+			productPrices[countryGroup.name]?.[fulfilmentOption]?.[productOption]?.[
+				billingPeriod
+			]?.[countryGroup.currency]?.price ?? 0;
+		return originalPrice > currentPrice ? originalPrice : undefined;
+	}
+
 	const isSupporterPlus = useContributionsSelector(isSupporterPlusFromState);
 
 	const currency = currencies[currencyId];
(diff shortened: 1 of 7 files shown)


support-frontend/assets/components/orderSummary/contributionsOrderSummaryContainer.tsx
```diff
@@ -45,10 +59,38 @@ export function ContributionsOrderSummaryContainer({
...
+		fulfilmentOption: FulfilmentOptions = NoFulfilmentOptions,
+		productOption: ProductOptions = NoProductOptions,
+	): number | undefined {
+		const countryGroup = getCountryGroup(country);
+		const originalPrice =
+			productPrices[countryGroup.name]?.[fulfilmentOption]?.[productOption]?.[
+				billingPeriod
+			]?.[countryGroup.currency]?.price ?? 0;
+		return originalPrice > currentPrice ? originalPrice : undefined;
+	}
```
jamesgorrie, line 92: I'd really like a test here of sorts to describe this (not sure we have tests for this type of stuff). But if I am reading this right...

1. We have a `currentPrice` - which will be the price with a promotion applied.
2. This might be different to the price in the catalogue, which is stored in `productPrices`
3. We only do this on S+ as contributions technical price is `0`
4. We send over `originalPrice` only if it's different and use that optionality to render differently

On point 4 - I wonder if we always send current + original price and let the rendering component calculate what to do with that?

support-frontend/assets/helpers/redux/checkout/product/selectors/isSupporterPlus.ts
```diff
@@ -20,37 +20,24 @@ export function isSupporterPlus(
...
-	}
+	const thresholdPrice = getThresholdPrice(countryGroupId, contributionType);
+	const selectedAmount = getAmount(
+		selectedAmounts,
+		otherAmounts,
+		contributionType,
+	);
 
-	return selectedAmount >= benefitsThreshold;
+	return selectedAmount >= thresholdPrice;
```
jamesgorrie, line 30: This looks good to me - but I wonder if we could have some time next week writing some tests to help ensure these methods are covering all the bases and we don't change something for a use case we didn't know about

jamesgorrie (approved):
This is looking good!

There's a few refactors I wish we had tests on. for confidence - maybe something to look at in future.

Also a note on maybe a renaming we can add in future for clarity - it'd be good to hear what you think about that.

support-frontend/assets/components/orderSummary/contributionsOrderSummaryContainer.tsx
```diff
@@ -45,10 +59,38 @@ export function ContributionsOrderSummaryContainer({
...
 		(state) => state.common.internationalisation,
 	);
-	const selectedAmount = useContributionsSelector(getUserSelectedAmount);
+	const { productType } = useContributionsSelector(
+		(state) => state.page.checkoutForm.product,
+	);
+	const billingPeriod = (productType[0] +
+		productType.slice(1).toLowerCase()) as BillingPeriod;
+	const currentPrice = useContributionsSelector(getUserSelectedAmount);
+	const originalPrice = useContributionsSelector((state) =>
```
jamesgorrie, line 71: A thought here on naming.

- We have `productPrice`, which is the price from the catalogue
- We have `productPriceWithPromo` which is a _lower_ price with a promotion applied (non-applicable to contributions)
- We have `productPriceWithContribution` which is a _higher_ price with an additional contribution added.

With these three values and the thresholds, we should have everything we need to create the UIs and functionality we want. It's an initial thought, let's pair on it next week.

paul-daniel-dempsey: Yep happy to change this but as we havent introduced PromoCodes in this R I will do on next one.  It will be `currentPrice` = `productPriceWithPromo` & `originalPrice` = `productPriceWithPromo`. I don't think `productPriceWithContribution` is required here (its only associated with fixed pricing),

support-frontend/assets/helpers/redux/checkout/product/selectors/isSupporterPlus.ts
paul-daniel-dempsey: I did check the use cases but happy to learn how to write test harnesses next week in a new PR.

support-frontend/assets/components/orderSummary/contributionsOrderSummaryContainer.tsx
```diff
@@ -45,10 +59,38 @@ export function ContributionsOrderSummaryContainer({
...
+		getOriginalPrice(
+			state.page.checkoutForm.product.productPrices,
+			countryId,
+			billingPeriod,
+		),
+	);
+
+	function getOriginalPrice(
```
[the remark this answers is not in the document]
paul-daniel-dempsey: Yep, good point, have moved it out now

support-frontend/assets/components/orderSummary/contributionsOrderSummaryContainer.tsx
paul-daniel-dempsey: Hmm, I thought that but thought it looked more messy. The `originalPrice` (to be later `productPrice`) is only returned if its available AND is greater than the `currentPrice` (to be later`productPriceWithPromo`) otherwise its `undefined`. erhaps we can cleanup further if needed after this PR.

paul-daniel-dempsey:
Ok am going to put this in, re:test harness, agree would of been to do but think non of these changes will affect the checkout visually, it not until the next (apply PromoCodes) PR is completed when see/check them anyway. Can raise a seperate PR to do any further chnage you think are needed however.

support-frontend/assets/components/orderSummary/contributionsOrderSummaryContainer.tsx
paul-daniel-dempsey: First thing in 3rd PromoCode PR https://github.com/guardian/support-frontend/compare/main...three-tier-promocode-splus-visual-change

PR #5806, opened 2024-03-14
Better playright chat links

johnduffell:
This updates the E2E test failure message so that we get a deep link to the failing test runs.  And small copy updates

image

Now the link goes to the right sessions:
image


GHaberis (approved):
Nice 👏

PR #5808, opened 2024-03-14
Hide the country group dropdown if user is in a country from VAT group

MaelGNM:
Don't show the `CountryGroupSwitcher` on the checkout if the user is in a country from the `countriesAffectedByVATStatus` list.

## Why are you doing this?
There is a list of countries in which we only offer contributions (not incurring VAT). This is done by using a amounts test target at these specific countries, and only allowing amounts under supporter+ threshold.
Changing the country/currency via the dropdown in the header allowed to bypass this and so this PR removes it when the user is in one of these countries.

## How to test
Navigate to the checkout from one of the affected countries. You can force the country by adding the query string parameter `country` and a 2 letter country code (make sure the region matches the country) e.g. /uk/contribute?country=GB for UK or /int/contribute?country=TW for Taiwan. (browserstack can also be used to test this)

## Screenshots
| Before (PROD) | After (CODE) |
| --- | --- |
| Desktop Before desktop | Desktop After desktop |
| Mobile Before mobile | Mobile After mobile |

### DIFF
--- a/support-frontend/assets/pages/supporter-plus-landing/twoStepPages/checkoutScaffold.tsx
+++ b/support-frontend/assets/pages/supporter-plus-landing/twoStepPages/checkoutScaffold.tsx
@@ -28,6 +28,7 @@ import { PageScaffold } from 'components/page/pageScaffold';
 import { SecureTransactionIndicator } from 'components/secureTransactionIndicator/secureTransactionIndicator';
 import HeadlineImageDesktop from 'components/svgs/headlineImageDesktop';
 import HeadlineImageMobile from 'components/svgs/headlineImageMobile';
+import { countriesAffectedByVATStatus } from 'helpers/internationalisation/country';
 import {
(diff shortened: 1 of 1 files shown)


rupertbates:
Can we test by adding a querystring parameter? Could you put the details of that in the description?

MaelGNM:
> Can we test by adding a querystring parameter? Could you put the details of that in the description?

Yes! I've updated the description with more information.

rupertbates:
This looks good but we will also need to hide it on the Guardian Weekly landing page I think:
https://support.code.dev-theguardian.com/eu/subscribe/weekly?country=AD

MaelGNM:
> This looks good but we will also need to hide it on the Guardian Weekly landing page I think: https://support.code.dev-theguardian.com/eu/subscribe/weekly?country=AD

We don't restrict Guardian Weekly sales so that wouldn't be needed.

PR #5811, opened 2024-03-15
threeTierCreateSupporterPlusSubscription should be false in payload for Paper and Digital Edition subs

GHaberis:

`threeTierCheckoutEnabled()` is called via https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/helpers/subscriptionsForms/submit.ts#L246 when building the payload for ALL recurring subscriptions, including Paper and Digital Edition. 

Currently `threeTierCheckoutEnabled()` returns `true` on the Paper and Digital Edition checkout, meaning we pass `threeTierCreateSupporterPlusSubscription` as `true` in the payload to the `/create` endpoint, I've tested and luckily the step functions don't go down the path of creating an SupporterPlus sub for users who are purchasing Paper OR Digital Edition! But  `threeTierCheckoutEnabled()` should really return `false` for  Paper and Digital Edition checkouts.

### DIFF
--- a/support-frontend/assets/pages/supporter-plus-landing/setup/threeTierChecks.ts
+++ b/support-frontend/assets/pages/supporter-plus-landing/setup/threeTierChecks.ts
@@ -6,6 +6,17 @@ export const threeTierCheckoutEnabled = (
 	abParticipations: Participations,
 	countryId: IsoCountry,
 ): boolean => {
+	const isPaperCheckout = window.location.pathname.startsWith(
+		'/subscribe/paper/checkout',
+	);
+	const isDigitalEditionCheckout = window.location.pathname.startsWith(
+		'/subscribe/digitaledition/checkout',
+	);
+
+	if (isPaperCheckout || isDigitalEditionCheckout) {
+		return false;
+	}
+
 	const isWeeklyCheckout = window.location.pathname.startsWith(
 		'/subscribe/weekly/checkout',
 	);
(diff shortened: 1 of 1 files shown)


support-frontend/assets/pages/supporter-plus-landing/setup/threeTierChecks.ts
```diff
@@ -6,6 +6,17 @@ export const threeTierCheckoutEnabled = (
 	abParticipations: Participations,
 	countryId: IsoCountry,
 ): boolean => {
+	const isPaperCheckout = window.location.pathname.startsWith(
+		'/subscribe/paper/checkout',
+	);
+	const isDigitalEditionCheckout = window.location.pathname.startsWith(
+		'/subscribe/digitaledition/checkout',
```
paul-daniel-dempsey, line 13: should this read `/subscribe/digitaledition` to match -> https://support.theguardian.com/uk/subscribe/digitaledition

paul-daniel-dempsey (approved):
Looks good, small comment over the digital edition path used and whether it has `checkout` in it?

support-frontend/assets/pages/supporter-plus-landing/setup/threeTierChecks.ts
GHaberis: Good spot! You're correct, will update now.

PR #5813, opened 2024-03-15
Replace Recurring Contribution e2e tests with Tiered Checkout tests

GHaberis:

- Replace `recurringContributions.test.ts` with `tieredCheckout.test.ts`. The `tieredCheckout.test.ts` are closely based om the original `recurringContributions.test.ts`, but have been updated to work with the 3-tier landing page. Currently they only test Tier 1 and 2, as we essentially had 80% of the tests already written in `recurringContributions.test.ts`.
- Restore Paypal tests for `oneOffContributions.test.ts`.


support-e2e/tests/tieredCheckout.test.ts
```diff
@@ -0,0 +1,104 @@
...
+import { fillInDirectDebitDetails } from './utils/directDebitDetails';
+import { fillInPayPalDetails } from './utils/paypal';
+import { setupPage } from './utils/page';
+import { afterEachTasks } from './utils/afterEachTest';
+
+interface TestDetails {
+	paymentType: 'Credit/Debit card' | 'Direct debit' | 'PayPal';
+	tier: 1 | 2 | 3;
+	frequency: 'Monthly' | 'Annual';
+	country?: 'US' | 'AU';
```
paul-daniel-dempsey, line 15: Worth adding a European country or `EU`?

GHaberis: I don't think so, I think the only reason we test a US variant is because there are a couple additional fields on the US checkout compared to UK. We don't actually have any tests which test AU (but this was there in the original `recurringContributions.tests.ts` file so I've left it here - there'd be no benefit in testing AU as well as UK as there's no extra functionality.

PR #5814, opened 2024-03-15
Replace Recurring Contribution e2e tests with Tiered Checkout tests

GHaberis:

- Replace `recurringContributions.test.ts` with `tieredCheckout.test.ts`. The `tieredCheckout.test.ts` are closely based om the original `recurringContributions.test.ts`, but have been updated to work with the 3-tier landing page. Currently they only test Tier 1 and 2, as we essentially had 80% of the tests already written in `recurringContributions.test.ts`.
- Restore Paypal tests for `oneOffContributions.test.ts`.


support-e2e/tests/tieredCheckout.test.ts
```diff
@@ -0,0 +1,104 @@
...
+	paymentType: 'Credit/Debit card' | 'Direct debit' | 'PayPal';
+	tier: 1 | 2 | 3;
+	frequency: 'Monthly' | 'Annual';
+	country?: 'US' | 'AU';
+}
+
+const testsDetails: TestDetails[] = [
+	{ paymentType: 'Credit/Debit card', tier: 1, frequency: 'Monthly' },
+	{ paymentType: 'Credit/Debit card', tier: 2, frequency: 'Annual' },
+	{ paymentType: 'Direct debit', tier: 2, frequency: 'Monthly' },
```
paul-daniel-dempsey, line 21: Do we need an `Annual` direct debit here?

GHaberis: I don't think so, the functionality is exactly the same on Monthly w/ Direct Debit and Annual w/ Direct Debit across both tiers so I don't think the test would bring any extra benefit than what we have now.

paul-daniel-dempsey (approved):
Thanks for running through setting this up, runs perfectly on playwright but will look into understanding these tests a little more now too.

support-e2e/tests/tieredCheckout.test.ts
paul-daniel-dempsey: Makes sense, thanks

PR #5817, opened 2024-03-19
improve logging of which class logging came from and remove excessive logging

johnduffell:
I have noticed that all our logging appears as "SafeLogger" rather than the class it was logged from.

Also I noticed there's some really verbose logging which makes it hard to find anything in the logs based on timestamp (intorduced in https://github.com/guardian/support-frontend/pull/5664 )

This PR changes all logging to use a trait (first commit), and removes the above logging (second commit)

Before it has all these lines and they're hard to find in the codebase:
image

now they look better
image

and locally in intellij
image

Tested in CODE and checked out an every day newspaper successfully

### DIFF
--- a/stripe-patrons-data/src/main/scala/com/gu/patrons/conf/PatronsIdentityConfig.scala
+++ b/stripe-patrons-data/src/main/scala/com/gu/patrons/conf/PatronsIdentityConfig.scala
@@ -8,7 +8,7 @@ case class PatronsIdentityConfig(apiUrl: String, apiClientToken: String)
 
 object PatronsIdentityConfig extends ConfigService {
   def fromParameterStoreSync(stage: Stage) = {
-    SafeLogger.info(s"Loading Identity config")
+    logger.info(s"Loading Identity config")
     val params = ParameterStoreService(stage).getParametersByPathSync("identity-config")
     PatronsIdentityConfig(
       findParameterOrThrow("api-url", params),
--- a/stripe-patrons-data/src/main/scala/com/gu/patrons/lambdas/PatronCancelledEventLambda.scala
+++ b/stripe-patrons-data/src/main/scala/com/gu/patrons/lambdas/PatronCancelledEventLambda.scala
@@ -43,7 +43,7 @@ import scala.util.{Failure, Success, Try}
 class PatronCancelledEventLambda extends StrictLogging {
   val runner = configurableFutureRunner(60.seconds)
 
-  implicit val stage = StageConstructors.fromEnvironment
+  implicit val stage: Stage = StageConstructors.fromEnvironment
   private val stripeConfig = PatronsStripeConfig.fromParameterStoreSync(stage)
   private val identityConfig = PatronsIdentityConfig.fromParameterStoreSync(stage)
 
@@ -56,7 +56,7 @@ class PatronCancelledEventLambda extends StrictLogging {
 
   def cancelCustomerInDynamoDb(event: APIGatewayProxyRequestEvent): Future[APIGatewayProxyResponseEvent] = {
     val countryId = event.getPathParameters.get("countryId")
-    SafeLogger.info(s"Path is ${countryId}")
+    logger.info(s"Path is ${countryId}")
     val account = if (countryId == "au") GnmPatronSchemeAus else GnmPatronScheme
 
     (for {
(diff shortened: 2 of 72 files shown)


support-frontend/app/services/pricing/PriceSummaryService.scala
```diff
@@ -76,25 +77,7 @@ class PriceSummaryService(
...
-        promotionService
-          .validatePromotion(promo, country, productRatePlan.id, isRenewal = false),
-      )
-      .partitionMap(identity)
-
-    SafeLogger.info("Invalid promotions")
-    SafeLogger.info(invalidPromotions.map(promoError => promoError.msg).mkString(", "))
-
-    SafeLogger.info("Valid promotions")
-    SafeLogger.info(validPromotions.map(promoError => promoError.promoCode).mkString(", "))
```
johnduffell, line 97: removing these lines is the bit that makes the logging less verbose when people are hitting our page

stripe-patrons-data/src/main/scala/com/gu/patrons/lambdas/PatronCancelledEventLambda.scala
```diff
@@ -43,7 +43,7 @@ import scala.util.{Failure, Success, Try}
 class PatronCancelledEventLambda extends StrictLogging {
   val runner = configurableFutureRunner(60.seconds)
 
-  implicit val stage = StageConstructors.fromEnvironment
+  implicit val stage: Stage = StageConstructors.fromEnvironment
```
johnduffell, line 46: this was fixing a warning about implicit parameters with no explicit type

stripe-patrons-data/src/main/scala/com/gu/patrons/conf/PatronsIdentityConfig.scala
```diff
@@ -8,7 +8,7 @@ case class PatronsIdentityConfig(apiUrl: String, apiClientToken: String)
 
 object PatronsIdentityConfig extends ConfigService {
   def fromParameterStoreSync(stage: Stage) = {
-    SafeLogger.info(s"Loading Identity config")
+    logger.info(s"Loading Identity config")
```
johnduffell, line 11: now classes extend (directly or indirectly) `SafeLogging`, which has a logger with a reference to the class name.  Previously we called the object `SafeLogger` which had a logger that referenced its own name.

Overall that means we log with the correct class name, which makes the logs more useful.

stripe-patrons-data/src/test/scala/com/gu/patrons/services/StripeSubscriptionsProcessorSpec.scala
```diff
@@ -33,7 +32,7 @@ class StripeSubscriptionsProcessorSpec extends AsyncFlatSpec with Matchers {
       identityService
         .getUserIdFromEmail(subscription.customer.email)
         .map(maybeIdentityId =>
-          SafeLogger.info(s"${subscription.customer.email} - ${maybeIdentityId.getOrElse("No Identity account")}"),
+          info(s"${subscription.customer.email} - ${maybeIdentityId.getOrElse("No Identity account")}"),
```
johnduffell, line 35: this is a test - there is a scalatest built in `info` method for writing test logging

support-config/src/main/scala/com/gu/monitoring/SafeLogger.scala
```diff
@@ -1,21 +1,33 @@
 package com.gu.monitoring
```
johnduffell, line 1: this file just has a bit of a shuffle to make everything available from the trait rather than by acessing the static object.

support-frontend/app/admin/settings/SettingsProvider.scala
```diff
@@ -2,13 +2,12 @@ package admin.settings
 
 import java.util.concurrent.Executors
 import java.util.concurrent.atomic.AtomicReference
-import admin.settings.SettingsProvider._
+import admin.settings.SettingsProvider.SettingsUpdate
```
johnduffell, line 5: this had to be more specific because importing everything from SettingsProvider also pulled in the `logger`.  That causes a cyclical import issue.

support-frontend/app/wiring/AppComponents.scala
```diff
@@ -37,7 +37,6 @@ class AppComponents(context: Context)
     Some(router),
     assetsResolver,
     allSettingsProvider,
-    appConfig.stage,
```
johnduffell, line 40: this was unused

support-modules/rest/src/main/scala/com/gu/monitoring/SafeLogger.scala
```diff
@@ -1,21 +1,33 @@
 package com.gu.monitoring
```
johnduffell, line 1: this is a second copy of SafeLogger - it was probably from when the mono repo was created.  There's no obvious module at the top of the dependency tree to put a single copy in so I've just left the duplication

rupertbates (approved):
👍

PR #5818, opened 2024-03-19
safely add explicit type to prices encoder

johnduffell:
The prices encoder had no explicit type and used an auto decoder.

This meant that if you added an explicit type, it would cause it to depend on itself and therefore cause a NPE [1]

This PR changes it to a semiauto so that we can add a type.

A type is important because if we switch over to Scala 3, types are required on all implicits.

Also, it's risky to have code that misbehaves at runtime if a type is added!  Especially as this code path is used by dotcom, so unless you check the logs you won't realse you have broken it!

semi auto: https://circe.github.io/circe/codecs/semiauto-derivation.html

[1]
```
Caused by: java.lang.NullPointerException: Cannot invoke "io.circe.Encoder.apply(Object)" because "encoder" is null
 at io.circe.syntax.package$EncoderOps$.asJson$extension(package.scala:24)
 at controllers.PricesController.$anonfun$getPrices$1(PricesController.scala:132)
 at scala.Function1.$anonfun$andThen$1(Function1.scala:87)
 at actions.NoCacheActionBuilder.invokeBlock(NoCacheActionBuilder.scala:15)
 at actions.NoCacheActionBuilder.invokeBlock(NoCacheActionBuilder.scala:7)
 at play.api.mvc.ActionBuilder$$anon$9.apply(Action.scala:434)
```

output of /prices (in DEV) is identical
BEFORE
```
// 20240319132645
// http://localhost:9210/prices

{
  "GBPCountries": {
    "GuardianWeekly": {
      "Monthly": {
        "price": "15",
        "currency": "£",
        "priceSummary": null
      },
      "Annual": {
        "price": "180",
        "currency": "£",
        "priceSummary": null
      }
    },
    "Digisub": {
      "Monthly": {
        "price": "7.49",
        "currency": "£",
        "priceSummary": null
      },
      "Annual": {
        "price": "149",
        "currency": "£",
        "priceSummary": null
      }
    }
  },
  "UnitedStates": {
    "GuardianWeekly": {
      "Monthly": {
        "price": "30",
        "currency": "US$",
        "priceSummary": null
      },
      "Annual": {
        "price": "360",
        "currency": "US$",
        "priceSummary": null
      }
    },
    "Digisub": {
      "Monthly": {
        "price": "12.49",
        "currency": "US$",
        "priceSummary": null
      },
      "Annual": {
        "price": "249",
        "currency": "US$",
        "priceSummary": null
      }
    }
  },
  "EURCountries": {
    "GuardianWeekly": {
[...]

PR #5819, opened 2024-03-19
add new test for prices encoder

johnduffell:
following on from 
https://github.com/guardian/support-frontend/pull/5818#issuecomment-2007240542
@mchv pointed out a test might find the NPE issue.

This is useful to have, because this endpoint https://support.theguardian.com/prices is used by dotcom, and wouldn't be tested routinely in CODE.  The existing testing is on the data, rather than the serialisation.  The first we would know of it would be a high 5xx rate alarm.

I wrote this test against main before the above PR and it passed.  I added the type annotation and it failed
```
[info] PricesControllerSerialisationTest:
[info] - should serialise a sample prices payload *** FAILED ***
[info]   java.lang.NullPointerException: Cannot invoke "io.circe.Encoder.apply(Object)" because "encoder" is null
[info]   at io.circe.syntax.package$EncoderOps$.asJson$extension(package.scala:24)
[info]   at controllers.PricesControllerSerialisationTest.$anonfun$new$1(PricesControllerSerialisationTest.scala:49)
```

I updated main locally and it passed again.

### DIFF
--- a/support-frontend/test/controllers/PricesControllerSerialisationTest.scala
+++ b/support-frontend/test/controllers/PricesControllerSerialisationTest.scala
@@ -1,0 +1,55 @@
+package controllers
+
+import com.gu.i18n.Currency
+import com.gu.support.promotions.{
+  DiscountBenefit,
+  FreeTrialBenefit,
+  IntroductoryPeriodType,
+  IntroductoryPriceBenefit,
+  Issue,
+}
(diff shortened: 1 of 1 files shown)


mchv:
Nice one @johnduffell

jamesgorrie (approved):

PR #5820, opened 2024-03-19
fix browserstack link

johnduffell:
browserstack link wasn't populated here https://github.com/guardian/support-frontend/actions/runs/8346808348/job/22844962434

`browserStackSearch="https://automate.browserstack.com/dashboard/v2/search?query=playwright-build-&type=builds"`
should have the commit ID in there too

This PR adds it back in so the var is set on failure, rather than skipped

### DIFF
--- a/.github/workflows/playwright.yaml
+++ b/.github/workflows/playwright.yaml
@@ -37,6 +37,7 @@ jobs:
         working-directory: ./support-e2e
 
       - name: Set outputs
+        if: ${{ failure() }}
         id: vars
         run: echo "sha_short=$(git rev-parse --short HEAD)" >> $GITHUB_OUTPUT
 


.github/workflows/playwright.yaml
```diff
@@ -37,6 +37,7 @@ jobs:
         working-directory: ./support-e2e
 
       - name: Set outputs
+        if: ${{ failure() }}
```
jamesgorrie, line 40: I'm not sure I understand this - my expectations is that the outputs would always be set? And then we just [read them on failure](https://github.com/guardian/support-frontend/pull/5820/files#diff-b710b539e104e4351f3e7598f94f434b8c7aaafdba88e57b3280582c718f7b4bR45).

I can't find any docs to suggest differently? But if it works, it's just something I am not getting.

johnduffell: it seems to skip that step when the previous step failed
image

johnduffell: here are the docs, by default it's only run on success 
https://docs.github.com/en/actions/learn-github-actions/expressions#status-check-functions

jamesgorrie: Ah, so the `failure` is on the `steps` rather than the previous `step`. We live we learn - thanks for sharing!

jamesgorrie (approved):

PR #5821, opened 2024-03-20
print out supporter-workers input and output state in a much nicer way

johnduffell:
It's quite hard to pick out individual fields in the state when looking at support-workers logs.

This PR updates it so that it's nicely formatted.

I had to pull in a new library pprint https://com-lihaoyi.github.io/PPrint/

Also I had to write a custom logback appender which uses the AWS functionality that doesn't split up multi line log statements https://docs.aws.amazon.com/lambda/latest/dg/java-logging.html#:~:text=The%20logger%20class%20supports%20multiline%20logs.

before (this is in CODE)

image

After (split into two screenshots due to length)
image
image

### DIFF
--- a/support-workers/src/main/scala/com/gu/support/workers/lambdas/Handler.scala
+++ b/support-workers/src/main/scala/com/gu/support/workers/lambdas/Handler.scala
@@ -1,8 +1,9 @@
 package com.gu.support.workers.lambdas
 
-import java.io.{InputStream, OutputStream}
+import ch.qos.logback.core.OutputStreamAppender
 
-import com.amazonaws.services.lambda.runtime.{Context, RequestStreamHandler}
+import java.io.{InputStream, OutputStream}
+import com.amazonaws.services.lambda.runtime.{Context, LambdaRuntime, RequestStreamHandler}
 import com.gu.monitoring.SafeLogging
 import com.gu.support.workers.exceptions.ErrorHandler
 import com.gu.support.workers.{ExecutionError, RequestInfo}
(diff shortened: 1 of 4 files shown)


support-workers/src/main/scala/com/gu/support/workers/lambdas/Handler.scala
```diff
@@ -12,6 +13,28 @@ import scala.concurrent.duration._
 import scala.concurrent.{Await, ExecutionContext, Future}
 import scala.util.control.NonFatal
 
+/** this appender makes sure that multi line messages stay together in cloudwatch
+  */
+class LambdaAppender[E] extends OutputStreamAppender[E] {
+  override def start(): Unit = {
+    println("Starting appender " + this.getClass.getName)
```
jamesgorrie, line 20: Do we want this?

support-workers/src/main/scala/com/gu/support/workers/lambdas/Handler.scala
```diff
@@ -36,12 +59,17 @@ abstract class Handler[IN, OUT](implicit
     }
 
   def handleRequestFuture(is: InputStream, os: OutputStream, context: Context): Future[Unit] = {
+    def log[T](message: String, a: T): Unit = {
+      val prettyData = pprint.apply(a).plainText.replaceAll("\\n", "\n\t")
```
jamesgorrie, line 63: Interesting that we don't have to `import` `pprint`?

support-workers/src/main/scala/com/gu/support/workers/lambdas/Handler.scala
```diff
@@ -36,12 +59,17 @@ abstract class Handler[IN, OUT](implicit
     }
 
   def handleRequestFuture(is: InputStream, os: OutputStream, context: Context): Future[Unit] = {
+    def log[T](message: String, a: T): Unit = {
```
jamesgorrie, line 62: Maybe for a future PR - but would we want to pull this into our `SafeLogger` so we benefit from this everywhere?

jamesgorrie (approved):
Very nice and tidy!

support-workers/src/main/scala/com/gu/support/workers/lambdas/Handler.scala
johnduffell: I did that on purpose so that we know that it's working, but I could take it out.  I notice logback already logs quite a lot on startup so maybe we don't want to add to it?  or maybe take it out if it annoys us later

support-workers/src/main/scala/com/gu/support/workers/lambdas/Handler.scala
johnduffell: `pprint` is a package with an apply method in the package object!  I guess the idea was to make it available everywhere without imports?  You can also log directly to the console with pprint.log, (however in our case we want to send it to our logger)

support-workers/src/main/scala/com/gu/support/workers/lambdas/Handler.scala
johnduffell: good point.  Yes it would make so much sense to do that, I'll do that in another PR.

PR #5824, opened 2024-03-20
Add ProductCatalogService

jamesgorrie:
- Adds the ProductCatalogService from the [product-catalog API](https://product-catalog.guardianapis.com/product-catalog.json) (which is generated by https://github.com/guardian/support-service-lambdas/tree/main/handlers/generate-product-catalog )
- Adds the `CachedProductCatalogServiceProvider` to prevent having an HTTP call on each request for data that doesn't been to be mega-fresh
- Adds an integration test to help with debugging

Currently I am just using it for the endpoint that the generic checkout (#5722) uses, but I'll be making it available on the `window.guardian` object, negating the need to make an async call from the client for it.

[**Trello Card**](https://trello.com/c/IfPDMdrT/709-create-product-catalog-service-to-make-to-enable-making-it-available-on-windowguardian)

## How to test

This is available at `/api/products` (although this will be made obsolete once it's on the `window.guardian`).

## Screenshots
Screenshot 2024-03-20 at 15 26 44

### DIFF
--- a/support-frontend/app/controllers/Application.scala
+++ b/support-frontend/app/controllers/Application.scala
@@ -19,7 +19,7 @@ import play.api.libs.circe.Circe
 import play.api.libs.ws.WSClient
 import play.api.mvc._
 import services.pricing.PriceSummaryServiceProvider
-import services.{PaymentAPIService, TestUserService}
+import services.{CachedProductCatalogService, CachedProductCatalogServiceProvider, PaymentAPIService, TestUserService}
 import utils.FastlyGEOIP._
 import views.EmptyDiv
 
@@ -39,7 +39,7 @@ case class PaymentMethodConfigs(
 class Application(
     actionRefiners: CustomActionBuilders,
     val assets: AssetsResolver,
-    testUsers: TestUserService,
+    testUserService: TestUserService,
     components: ControllerComponents,
     oneOffStripeConfigProvider: StripePublicConfigProvider,
     regularStripeConfigProvider: StripePublicConfigProvider,
@@ -53,6 +53,7 @@ class Application(
     stage: Stage,
     wsClient: WSClient,
     priceSummaryServiceProvider: PriceSummaryServiceProvider,
+    cachedProductCatalogServiceProvider: CachedProductCatalogServiceProvider,
     val supportUrl: String,
 )(implicit val ec: ExecutionContext)
     extends AbstractController(components)
@@ -186,7 +187,7 @@ class Application(
     )
 
     val serversideTests = generateParticipations(Nil)
-    val testMode = testUsers.isTestUser(request)
+    val testMode = testUserService.isTestUser(request)
 
     val queryPromos =
       request.queryString
@@ -273,7 +274,7 @@ class Application(
 
     val geoData = request.geoData
     val serversideTests = generateParticipations(Nil)
-    val isTestUser = testUsers.isTestUser(request)
+    val isTestUser = testUserService.isTestUser(request)
     // This will be present if the token has been flashed into the session by the PayPal redirect endpoint
     val guestAccountCreationToken = request.flash.get("guestAccountCreationToken")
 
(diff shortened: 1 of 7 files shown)


support-frontend/app/controllers/Application.scala
```diff
@@ -39,7 +39,7 @@ case class PaymentMethodConfigs(
 class Application(
     actionRefiners: CustomActionBuilders,
     val assets: AssetsResolver,
-    testUsers: TestUserService,
+    testUserService: TestUserService,
```
jamesgorrie, line 42: I renamed this as keeping things universally named helps (me) with discoverability.

support-frontend/app/services/ProductCatalogService.scala
```diff
@@ -0,0 +1,58 @@
...
+  }
+}
+
+class CachedProductCatalogServiceProvider(
+    codeCachedProductCatalogService: CachedProductCatalogService,
+    prodCachedProductCatalogService: CachedProductCatalogService,
+) {
+  def forUser(isTestUser: Boolean) =
+    if (isTestUser) codeCachedProductCatalogService else prodCachedProductCatalogService
+}
```
jamesgorrie, line 58: I've not extended the `TouchPointService` here, as I was creating more bootstrap code than this implementation is which is often a sign of the wrong abstraction being used.

Happy to move this over if there is benefit to that.

support-frontend/app/services/ProductCatalogService.scala
```diff
@@ -0,0 +1,58 @@
...
+
+class CodeProductCatalogService(val client: FutureHttpClient) extends ProductCatalogService {
+  val wsUrl: String = "https://product-catalog.code.dev-guardianapis.com"
+}
+
+class CachedProductCatalogService(system: ActorSystem, productCatalogService: ProductCatalogService)(implicit
+    ec: ExecutionContext,
+) {
+  private val json = new AtomicReference[JsonObject](JsonObject())
```
jamesgorrie, line 39: I'm using a JsonObject as I am hoping to keep this data only relevant to the client, which [has its own parser](https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/pages/%5BcountryGroupId%5D/checkout.tsx#L138-L150).

support-frontend/app/services/ProductCatalogService.scala
johnduffell: could also be a sign that the touchpointservice could be simplified!  I think one of the strengths of this code base is its consistency (as opposed to support-service-lambdas for example)

support-frontend/app/controllers/Application.scala
```diff
@@ -300,13 +301,12 @@ class Application(
...
-        "https://raw.githubusercontent.com/guardian/support-service-lambdas/0b031ea5821c95a7f7c59e45951d2d1f0bebed9d/modules/product/src/prodCatalogMapping.json",
-      )
-      .get()
-      .map(response => Ok(response.body).withHeaders("Cache-Control" -> "max-age=30"))
+  @deprecated("We will remove this once we get the data to the client not via HTTP")
+  def products() = Action { implicit request =>
+    val isTestUser = testUserService.isTestUser(request)
+    val cachedProductCatalogService = cachedProductCatalogServiceProvider.forUser(isTestUser)
+
+    Ok(cachedProductCatalogService.get().toJson).withHeaders("Cache-Control" -> "no-cache, private")
```
johnduffell, line 309: we have a `NoCacheAction` which saves having to do this explicitly

support-frontend/app/services/ProductCatalogService.scala
jamesgorrie: Happy to try and simplify outside of this PR - do you think I should go ahead and bootstrap those abstractions then?

support-frontend/app/controllers/Application.scala
jamesgorrie: Fab, added!

support-frontend/app/services/ProductCatalogService.scala
johnduffell: happy to leave the decision up to you as you've seen the extra boilerplate, if you're happy as is then leave it if it's working

johnduffell (approved):
looks good, I can't really assess the typescript as much but overall I approve! 👍

support-frontend/app/services/ProductCatalogService.scala
jamesgorrie: I might have a go at creating a configless Touchpoint service as a refactor based on this.

PR #5825, opened 2024-03-20
Product catalog on `window.guardian`

jamesgorrie:
Adds the `ProductCatalog` to the `window.guardian` of `/contribution` and `/checkout` route.

This PR also moves the parsing from the checkout => window as it is included in both places now.

### DIFF
--- a/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
@@ -149,9 +149,7 @@ const ProductSchema = object({
 	),
 });
 
-const ProductsSchema = object({
-	products: record(ProductSchema),
-});
+const ProductsSchema = record(ProductSchema);
 type Products = Output<typeof ProductsSchema>;
 
 function describeProduct(product: string, ratePlan: string) {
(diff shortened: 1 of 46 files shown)


johnduffell (approved):
nice 👍

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -280,29 +256,22 @@ const stripePublicKey = getStripeKey(
 );
 
 export function Checkout() {
-	const [productCatalog, setProductCatalog] = useState<ProductCatalog>();
-
-	useEffect(() => {
-		void fetch('/api/products')
```
Comment (johnduffell) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:286]: I guess we will delete the server side endpoint once no client side are using the old code any more

Comment (jamesgorrie) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:286]: I thought about doing it as part of this PR, but I found it useful to reference. But maybe I should just use standard URL: https://product-catalog.guardianapis.com/product-catalog.json

Comment (jamesgorrie) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:286]: I'm going to remove it - good shout

Comment (johnduffell) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:286]: actually I was assuming it was intentional in case of a race condition - i.e. someone loaded the old client side code, but the new server side is already responding to the request.  But as a one off it's probably going to be ok, assuming that information is only fetched on initial page load.

PR #5827, opened 2024-03-20
Lint tsconfig

jamesgorrie:
Lints the tsconfig file so it doesn't get caught in another PRs changes.

This doesn't happen automatically as [we only run the linter on `ts` and `tsx` files](https://github.com/guardian/support-frontend/blob/main/support-frontend/package.json#L8-L10).

I think I might try and make sure linting happens at the root of the project as we have JS elsewhere, and autofixing in the browser often catches things which I would prefer to ignore in the `ignore-revs` where possible.


paul-daniel-dempsey (approved):
Looks good

Issue #5829, opened 2024-03-20
set noUncheckedIndexedAccess: true


jamesgorrie:
We gain far more security by having [`noUncheckedIndexedAccess: true`](https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess) in [our `tsconfig`](https://github.com/guardian/support-frontend/blob/main/support-frontend/tsconfig.json#L23).

```[tasklist]
### Tasks
- [x] https://github.com/guardian/support-frontend/pull/5828 1 assets/components/directDebit/helpers/ajax.ts:29
- [ ] 1  assets/components/forms/customFields/options.tsx:9
- [ ] 3  assets/components/forms/customFields/sortedOptions.tsx:10
- [ ] 1  assets/components/gridImage/gridImage.tsx:33
- [ ] 5  assets/components/headers/links/links.tsx:34
- [ ] 1  assets/components/paymentMethodSelector/PaymentMethodSelectorContainer.tsx:109
- [ ] 1  assets/components/paymentRequestButton/hooks/payerDetails.ts:27
- [ ] 1  assets/components/payPalPaymentButton/payPalButtonProps.ts:67
- [ ] 1  assets/components/subscriptionCheckouts/address/postcodeFinder.tsx:105
- [x] https://github.com/guardian/support-frontend/pull/5830 3 assets/components/subscriptionCheckouts/addressSearch/__tests__/addressResultsTests.ts:44
- [x] https://github.com/guardian/support-frontend/pull/5830 14 assets/components/subscriptionCheckouts/addressSearch/addressResultsHelpers.ts:9
- [ ] 1  assets/components/subscriptionCheckouts/paymentMethodSelector.tsx:189
- [ ] 1  assets/components/thankYou/signIn/signInItems.tsx:127
- [ ] 2  assets/components/thankYou/supportReminder/supportReminderItems.tsx:201
- [ ] 11 assets/helpers/abTests/abtest.ts:177 https://github.com/guardian/support-frontend/pull/5928
- [ ] 4  assets/helpers/abTests/helpers.ts:265
- [ ] 2  assets/helpers/campaigns/campaigns.ts:34
- [ ] https://github.com/guardian/support-frontend/pull/5831 23 assets/helpers/contributions.ts:181
- [ ] 1  assets/helpers/customHooks/useHasBeenSeen.ts:19
- [ ] 1  assets/helpers/forms/existingPaymentMethods/existingPaymentMethods.ts:134
[...]

PR #5830, opened 2024-03-20
chore(ts): remove addressSearch

jamesgorrie:
In aid of #5829 

These helpers don't seem to be used anymore - so instead of fixing the TS error, I've just removed them.


paul-daniel-dempsey (approved):
Less code to maintain!

PR #5823, opened 2024-03-20
use useful name for support workers executions

johnduffell:
So far the support workers executions are just a wall of UUIDs
image

This PR makes them more useful
image

The format is
```
TEST-Monthly-GuardianWeekly-Domestic-GBP-DirectDebit-1409401929736666
^^^^ ^^^^^^^^product description^^^^^^^^ ^^^^^^^^^^^ ^^^^^nanosecond rolling time (292 year rollaround)
||||                                     pay_meth
present only if test user
```

PR #5832, opened 2024-03-20
truncate execution names that are over 80 character

johnduffell:
Following on from https://github.com/guardian/support-frontend/pull/5823

It turns out that in extreme cases the execution name can be over 80 chars
`TestUser-Quarterly-GuardianWeekly-RestOfWorld-USD-StripeCheckout-1425327897029666`
is 81 characters, meaning that checkout is impossible.

This PR truncates it if necessary (and makes sure the unique number is in the non truncated section)
image

EDIT
I have updated this PR to use a the densest possible encoding of a long (see below), so the truncation problem has gone away probably.  However it could still happen unexpectedly in future, so I have added some tests.
image

example IDs as generated by the new test
```
12345678901234567890123456789012345678901234567890123456789012345678901234567890
================================================================================
123456789-123456789-123456789-123456789-123456789-123456789-12345678-gAAAAAAAAAA
123456789-123456789-123456789-123456789-123456789-123456789-123456-gAAAAAAAAAA--
123456789-123456789-123456789-123456789-123456789-123456789-12345678-f_________8
123456789-123456789-123456789-123456789-123456789-123456789-123456-f_________8--
123456789-123456789-123456789-123456789-123456789-123456789-12345678-AAAAAAAAAAA
123456789-123456789-123456789-123456789-123456789-123456789-123456-AAAAAAAAAAA--
123456789-123456789-123456789-123456789-123456789-123456789-12345678-AAV3NZMnDh8
123456789-123456789-123456789-123456789-123456789-123456789-123456-AAV3NZMn8vY--
```

### DIFF
--- a/support-frontend/app/services/stepfunctions/Client.scala
+++ b/support-frontend/app/services/stepfunctions/Client.scala
@@ -12,6 +12,8 @@ import services.aws.{AwsAsync, CredentialsProvider}
 import services.stepfunctions.StateMachineContainer.{Response, convertErrors}
 import services.stepfunctions.StateMachineErrors.Fail
 
+import java.nio.ByteBuffer
+import java.util.Base64
 import scala.concurrent.{ExecutionContext, Future}
 import scala.jdk.CollectionConverters._
 
@@ -27,6 +29,24 @@ object Client {
 
     new Client(client, arn)
   }
+
+  def generateExecutionName(name: String, l: Long = System.nanoTime()) = {
+    val uniq = {
+      val bytes = ByteBuffer.allocate(8).putLong(l).array()
+      val encodedString = Base64.getEncoder.encodeToString(bytes)
+      encodedString.replaceAll("\\+", "_").replaceAll("/", "_").replaceAll("=", "")
+    }
+    val freeCharacters = 80 - name.length - 1 - uniq.length
+    val validName =
+      if (freeCharacters >= 0)
+        name + "-" + uniq
+      else {
+        val truncatedName = name.take(80 - 1 - uniq.length - 2)
+        truncatedName + "-" + uniq + "--"
+      }
+    validName
+  }
+
 }
 
 class Client(client: AWSStepFunctionsAsync, arn: StateMachineArn) extends SafeLogging {
@@ -34,12 +54,7 @@ class Client(client: AWSStepFunctionsAsync, arn: StateMachineArn) extends SafeLo
   private def startExecution(arn: String, input: String, name: String)(implicit
       ec: ExecutionContext,
   ): Response[StartExecutionResult] = convertErrors {
-    val proposedName = name + "-" + System.nanoTime().toString
-    val validName =
-      if (proposedName.length <= 80)
-        proposedName
-      else
-        (System.nanoTime().toString + "-" + name).take(78) + "--"
+    val validName: String = Client.generateExecutionName(name)
     val startExecutionRequest =
       new StartExecutionRequest()
         .withStateMachineArn(arn)
--- a/support-frontend/app/services/stepfunctions/Client.scala
+++ b/support-frontend/app/services/stepfunctions/Client.scala
@@ -54,12 +54,11 @@ class Client(client: AWSStepFunctionsAsync, arn: StateMachineArn) extends SafeLo
   private def startExecution(arn: String, input: String, name: String)(implicit
       ec: ExecutionContext,
   ): Response[StartExecutionResult] = convertErrors {
-    val validName: String = Client.generateExecutionName(name)
     val startExecutionRequest =
       new StartExecutionRequest()
         .withStateMachineArn(arn)
         .withInput(input)
-        .withName(validName)
+        .withName(Client.generateExecutionName(name))
     AwsAsync(client.startExecutionAsync, startExecutionRequest)
       .transform { theTry =>
         logger.info(s"state machine result: $theTry")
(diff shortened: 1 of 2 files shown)


support-frontend/app/services/stepfunctions/Client.scala
```diff
@@ -34,11 +34,17 @@ class Client(client: AWSStepFunctionsAsync, arn: StateMachineArn) extends SafeLo
   private def startExecution(arn: String, input: String, name: String)(implicit
       ec: ExecutionContext,
   ): Response[StartExecutionResult] = convertErrors {
+    val proposedName = name + "-" + System.nanoTime().toString
+    val validName =
+      if (proposedName.length <= 80)
+        proposedName
+      else
+        (System.nanoTime().toString + "-" + name).take(78) + "--"
```
johnduffell, line 42: the dashes are just an extra clue that it's been truncated.  We can only use dash and underscore in the string

jamesgorrie, line 42: To avoid having the prefix changing, could we use:
```suggestion
        // maxLength - System.nanoTime().toString.size - truncationString.size
        name.take(80 - 16 - 2) + System.nanoTime().toString + "--"
```

johnduffell: yeah good idea I'd like that too, i just wanted to prioritise reliability over consistency (although test users are going to be affected before real ones)

johnduffell: actually MAX and MIN LONG are 19 and 20 characters respectively when run through toString.  Safer to toString the value first, measure it, and check the total.
```
123456789 123456789 123456789
9223372036854775807
-9223372036854775808
```

support-frontend/test/services/stepfunctions/SupportWorkersClientTest.scala
```diff
@@ -60,4 +60,27 @@ class SupportWorkersClientTest extends AnyFlatSpec with Matchers with MockitoSug
...
+      }
+    }
+    doTest(68, Long.MinValue, truncated = false)
+    doTest(69, Long.MinValue, truncated = true)
+    doTest(68, Long.MaxValue, truncated = false)
+    doTest(69, Long.MaxValue, truncated = true)
+    doTest(68, 0L, truncated = false)
+    doTest(69, 0L, truncated = true)
+    doTest(68, System.nanoTime(), truncated = false)
+    doTest(69, System.nanoTime(), truncated = true)
```
johnduffell, line 83: these generate the following names
```
12345678901234567890123456789012345678901234567890123456789012345678901234567890
================================================================================
123456789-123456789-123456789-123456789-123456789-123456789-12345678-gAAAAAAAAAA
123456789-123456789-123456789-123456789-123456789-123456789-123456-gAAAAAAAAAA--
123456789-123456789-123456789-123456789-123456789-123456789-12345678-f_________8
123456789-123456789-123456789-123456789-123456789-123456789-123456-f_________8--
123456789-123456789-123456789-123456789-123456789-123456789-12345678-AAAAAAAAAAA
123456789-123456789-123456789-123456789-123456789-123456789-123456-AAAAAAAAAAA--
123456789-123456789-123456789-123456789-123456789-123456789-12345678-AAV3NZMnDh8
123456789-123456789-123456789-123456789-123456789-123456789-123456-AAV3NZMn8vY--
```

support-frontend/app/services/stepfunctions/Client.scala
johnduffell: OK I have pushed updates so it takes (exactly) 11 characters for the ID while remaining valid for step functions, and some tests to make sure, so the overflow issue has gone away for now.  Please have a look and see what you think @jamesgorrie

support-frontend/app/services/stepfunctions/Client.scala
```diff
@@ -27,19 +29,32 @@ object Client {
 
     new Client(client, arn)
   }
+
+  def generateExecutionName(name: String, l: Long = System.nanoTime()) = {
+    val uniq = {
+      val bytes = ByteBuffer.allocate(8).putLong(l).array()
+      val encodedString = Base64.getEncoder.encodeToString(bytes)
+      encodedString.replaceAll("\\+", "_").replaceAll("/", "_").replaceAll("=", "")
+    }
```
jamesgorrie, line 38: I think I can see what's potting here, is it a base64 encoded string to an 8bytes string?

support-frontend/test/services/stepfunctions/SupportWorkersClientTest.scala
```diff
@@ -60,4 +60,28 @@ class SupportWorkersClientTest extends AnyFlatSpec with Matchers with MockitoSug
...
```
jamesgorrie, line 83: I quite like [table driven testing](https://www.scalatest.org/user_guide/table_driven_property_checks) for this type of test - just sharing as this is nice and clear as it is.

jamesgorrie (approved):
Nice tests!

support-frontend/app/services/stepfunctions/Client.scala
johnduffell: Ultimately it's turning a 64 bit long into 11 characters that are valid in an execution name.

1. turn the Long (64 bits = 8 x 8bit byte) into 8 bytes in an array.
2. encode the byte array into base64 (6 bits per character - 64 bits makes 11 characters)
3. base64 includes 0-9, a-z, A-Z, `+` and `/` i.e. 10 + 26 + 26 + 2 = 64.  Our extra characters are `-` and `_` so it's just a quick substitution.  The `=` is just padding as base64 always pads to a multiple of 4 characters i.e. 3 bytes, so.we just throw that away.

support-frontend/test/services/stepfunctions/SupportWorkersClientTest.scala
johnduffell: thanks for the pointer, that's a nice level of separation.

On first impression only thing I don't like is it separates the data from its context - as it is, the parameter names are all there, cmd-click works well, and the stack trace will nicely show you where it was working at the time.

But I am aiming to learn more about the scalatest functionality so I will play with that next time and see how it feels.

PR #5833, opened 2024-03-21
update the AWS SDK to v2 for S3 and cloudwatch metrics

johnduffell:
We are currently on AWS SDK v1.  This will go into maintenance mode in summer (critical fixes and security updates) and out of support at the end of next year.
https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-java-v1-x-on-december-31-2025/

This PR starts the process of moving to v2 in support-frontend repo.

I noticed that the cloudwatch metrics client used some wrapper classes so the AWS api didn't escape, so all the changes were in one file.  The S3 client leaked out a couple of implementation details so the changes were more extensive.

Testing in CODE

- DONE stripe-patrons-data (CD wasn't switched on)
- DONE support-frontend (promo codes, price summary service - catalog)
- DONE support-workers
- [ ] it-test-runner (load tests jar)
- DONE stripe-intent (load config)


stripe-patrons-data/src/main/scala/com/gu/patrons/services/ParameterStoreService.scala
```diff
@@ -11,9 +11,8 @@ import com.amazonaws.services.simplesystemsmanagement.{
   AWSSimpleSystemsManagementAsync,
   AWSSimpleSystemsManagementAsyncClientBuilder,
 }
-import com.gu.aws.{AwsAsync, CredentialsProvider}
+import com.gu.aws.{AwsAsync, CredentialsProviderDEPRECATEDV1}
```
johnduffell, line 14: could have deprecated the old one, but I decided to give it this snappy name instead.

support-frontend/app/admin/settings/Settings.scala
```diff
@@ -39,7 +39,7 @@ object Settings {
   def fromS3[T: Decoder](source: SettingsSource.S3)(implicit s3: AwsS3Client): Either[Throwable, T] =
     for {
       buf <-
-        s3.fetchAsString(new AmazonS3URI("s3://" + source.bucket + "/" + source.key))
+        s3.fetchAsString(S3Location(source.bucket, "/" + source.key))
```
johnduffell, line 42: the new `S3Uri` class requires an instance of the S3 client to parse an s3 uri - not sure if it makes an api call?  Also the bucket and key are optional in it.  So I switched to a normal case class where possible.

https://docs.aws.amazon.com/AmazonS3/latest/userguide/example_s3_Scenario_URIParsing_section.html

support-frontend/app/services/pricing/DefaultPromotionService.scala
```diff
@@ -36,10 +36,9 @@ class DefaultPromotionServiceS3(
     extends DefaultPromotionService
     with LazyLogging {
 
-  private val s3Uri = {
-    val env = TouchPointEnvironments.fromStage(stage)
```
johnduffell, line 40: wasn't used, maybe left over from the Great Test Env Simplification of 2022ish

support-lambdas/stripe-intent/src/main/scala/com/gu/handler/impure/S3Loader.scala
```diff
@@ -1,30 +1,12 @@
...
-  EnvironmentVariableCredentialsProvider,
-  InstanceProfileCredentialsProvider,
-}
-import com.amazonaws.services.s3.AmazonS3URI
 import com.gu.aws.AwsS3Client
+import com.gu.aws.AwsS3Client.S3Location
 import com.gu.monitoring.SafeLogging
 import com.typesafe.config.{Config, ConfigFactory}
 
-object Aws {
```
johnduffell, line 14: unused

support-lambdas/stripe-intent/src/main/scala/com/gu/stripeIntent/Handler.scala
```diff
@@ -76,7 +76,7 @@ object Handler extends ApiGatewayHandler[RequestBody, StripeIntentEnv] {
 
   def getRealConfig(stage: Stage): Map[StripePublicKey, StripePrivateKey] = {
     val config = S3Loader
-      .load(StripeConfigPath.apply(stage))
+      .load(StripeConfigPath.forStage(stage))
```
johnduffell, line 79: oppotrunistic rename

build.sbt
```diff
@@ -44,25 +44,23 @@ inThisBuild(
     // https://www.scala-sbt.org/1.x/docs/Cached-Resolution.html
     updateOptions := updateOptions.value.withCachedResolution(true),
     resolvers ++= Resolver.sonatypeOssRepos("releases"), // libraries that haven't yet synced to maven central
+    assembly / assemblyMergeStrategy := {
```
johnduffell, line 47: this moves the merge strategy to "ThisBuild" which applies to all projects, so as long as no other project overrides it, it will automatically get used.  This saves faff trying to get it into all the right places.
The reason was because some mime.types file and netty file was conflicting in some places

build.sbt
```diff
@@ -89,6 +87,7 @@ lazy val commonDependencies = Seq(
   "com.typesafe" % "config" % "1.4.2",
   scalatest % "test",
   "com.typesafe.scala-logging" % "scala-logging_2.13" % "3.9.5",
+  "ch.qos.logback" % "logback-classic" % "1.4.14",
```
johnduffell, line 90: everything that sends logs needs this at runtime, otherwise you just get a "NOP logger" message and no actual logging.  This is mostly relevant for running tests against subprojects, as we had manually added it to all the actual deployed projects.

support-frontend/app/controllers/Application.scala
```diff
@@ -248,7 +248,10 @@ class Application(
   }
 
   def healthcheck: Action[AnyContent] = PrivateAction {
-    Ok("healthy")
+    if (priceSummaryServiceProvider.forUser(false).getPrices(SupporterPlus, List()).isEmpty)
+      InternalServerError("no prices in catalog")
+    else
+      Ok("healthy")
```
johnduffell, line 254: previously there was obviously an issue with this pricing data not being populated, hence this PR was raised https://github.com/guardian/support-frontend/pull/1752
However if there's no pricing data it's probably better not to let the instance even come up, rather than trying to alert on it?

support-frontend/test/controllers/ApplicationTest.scala
```diff
@@ -38,7 +44,19 @@ class ApplicationTest extends AnyWordSpec with Matchers with TestCSRFComponents
...
+        val priceSummaryService = mock[PriceSummaryService]
+        when(
+          priceSummaryService
+            .getPrices[SupporterPlus.type](any[SupporterPlus.type], any[List[PromoCode]], any[ReaderType]),
+        )
+          .thenReturn(Map[CountryGroup, CountryGroupPrices](CountryGroup.UK -> Map.empty))
+        val priceSummaryServiceProvider = mock[PriceSummaryServiceProvider]
+        when(priceSummaryServiceProvider.forUser(anyBoolean())).thenReturn(priceSummaryService)
+        priceSummaryServiceProvider
+      }
```
johnduffell, line 59: I updated the test so it's still testing the happy case, but I didn't actually add a test for heathcheck failing, not sure if it's necessary

support-internationalisation/build.sbt
```diff
@@ -2,7 +2,7 @@ import sbtrelease.ReleaseStateTransformations._
 
 name := "support-internationalisation"
 
-scalaVersion := "3.3.3"
+scalaVersion := "2.13.13"
```
johnduffell, line 5: tests that depended on this module for some reason missed it off the classpath when running in intellij, until I changed this default version. (`sbt test` etc was fine)
Once more is updated to scala 3, we might have the same problem with other way around - but let's cross that when we get there.

support-services/src/main/scala/com/gu/aws/AwsCloudWatchMetricSetup.scala
```diff
@@ -13,22 +13,6 @@ object AwsCloudWatchMetricSetup {
       ),
     )
 
-  def catalogFailureRequest(environment: TouchPointEnvironment): MetricRequest =
```
johnduffell, line 16: the health check now fails in this situation, so no metric is needed

support-services/src/main/scala/com/gu/aws/AwsCloudWatchMetricSetup.scala
```diff
@@ -13,22 +13,6 @@ object AwsCloudWatchMetricSetup {
...
-    getMetricRequest(
-      MetricName("CatalogLoadingFailure"),
-      Map(
-        MetricDimensionName("Environment") -> MetricDimensionValue(environment.envValue),
-      ),
-    )
-
-  def revenueDistributionFailureRequest(stage: Stage): MetricRequest =
```
johnduffell, line 24: this is unused, now the revenue-recogniser-job does this

support-services/src/main/scala/com/gu/support/catalog/CatalogJsonProvider.scala
```diff
@@ -1,26 +1,22 @@
...
+  override def get: Try[Json] = {
+    val catalog = S3Location("gu-zuora-catalog", s"PROD/Zuora-$environment/catalog.json")
+    logger.info(s"Attempting to load catalog from $catalog")
     fetchJson(AwsS3Client, catalog)
   }
 }
-
-class SimpleJsonProvider(json: Json) extends CatalogJsonProvider {
-  override def get: Option[Json] = Some(json)
-}
```
johnduffell, line 26: this is for tests so I've moved it to the test code

support-services/src/main/scala/com/gu/support/catalog/CatalogService.scala
```diff
@@ -58,24 +58,16 @@ class CatalogService(val environment: TouchPointEnvironment, jsonProvider: Catal
...
-        },
-      )
+    val attempt = for {
+      json <- jsonProvider.get
+      decoded <- json.as[ZuoraCatalog].toTry
+    } yield {
+      adjustSixWeeklyPriceList(Catalog.convert(decoded))
     }
+
+    attempt.get
```
johnduffell, line 70: this will throw any exception related to loading the catalog which will log at that point

support-frontend/build.sbt
```diff
@@ -6,7 +6,7 @@ packageSummary := "Support Play APP"
 
 libraryDependencies ++= Seq(
   "com.typesafe" % "config" % "1.4.2",
-  "com.gu" %% "simple-configuration-ssm" % "1.5.7",
+  "com.gu" %% "simple-configuration-ssm" % "1.7.0",
```
johnduffell, line 9: without this update we had a 
```
java.lang.NoClassDefFoundError: software/amazon/awssdk/protocols/query/interceptor/QueryParametersToBodyInterceptor
 at software.amazon.awssdk.services.autoscaling.DefaultAutoScalingBaseClientBuilder.finalizeServiceConfiguration(DefaultAutoScalingBaseClientBuilder.java:60)

```
This is because the odler SSM was compliled against an older AWS SDK v2, which wasn't binary compatible with the updated one I bumped in this PR.

support-modules/aws/src/main/scala/com/gu/aws/AwsS3Client.scala
```diff
@@ -1,34 +1,58 @@
...
+  def fetchAsString(s3Location: S3Location): Try[String] =
+    withStream(s3Location) { is =>
+      Try(Source.fromInputStream(is).mkString)
+    }
+
+  case class S3Location(bucket: String, key: String)
+  def parseUri(uri: String): S3Location = {
+    val s3Uri = s3.utilities().parseUri(URI.create(uri))
+    S3Location(
+      s3Uri.bucket().get(), // throw if not present
```
tomrf1, line 46: elsewhere in this file we're returning `Try` - would that be more consistent here?

johnduffell: yeah why not I'll do that.
sometimes dealing with the `Try`s can get out of hand when all you want to do is throw a 500 when it goes wrong (this function is only used for reading a config value to decide which file to load from S3)

johnduffell: done - it also actually encouraged me to neaten up S3Loader.scala into a for comprehension by making that change.

tomrf1 (approved):
good progress!

PR #5842, opened 2024-03-25
feat: add new google-site-verification-token

jamesgorrie:
Adds a new `google-site-verification` token. This is for the [Google Search Console](https://search.google.com/search-console/about).

My understanding [from the docs is that you can have multiple of the same verification methods](https://support.google.com/webmasters/answer/9008080?hl=en#:~:text=Using%20multiple%20verification%20methods). 

I didn't want to replace the existing one as [the PR updating it](https://github.com/guardian/support-frontend/pull/792) doesn't describe which service it is for - so this should avoid breaking that.


paul-daniel-dempsey (approved):
Not entirely sure how the process works of generating the site verification code but will have a look at the link in conversation over this.

jamesgorrie:
This would be adding my user as an owner (As I misunderstood the settings). We don't want this, closing.

PR #5843, opened 2024-03-25
fix: use testMode for product prices on contributions

jamesgorrie:
Uses `testMode` when fetching prices and injecting them on the client. This should only affect the `/contribution` route on `PROD`.

This was noticed as we are using `PROD` promotions even when a test user.


paul-daniel-dempsey (approved):
Great to get the Playwright tests going again on Prod

PR #5837, opened 2024-03-22
feat: promotion text on thank you page

jamesgorrie:
## What are you doing in this PR?
Passes the promotion down from state into the thank you page heading.

I'm not sure we want it to look like this, but it will get us to stage 1. I wonder if a <sup>*</sup> style footnote might work better? e.g.

---

### Thank you James for supporting us with **_£5_** every month<sup>*</sup> ❤️

<sup>*</sup> For 6 months, then £10 every month.

---

☝️ has been done, I'll work with design to make sure it's all good.

## 👀 Screenshots 👀 
Screenshot 2024-03-22 at 16 33 31
Screenshot 2024-03-22 at 16 33 45

PR #5844, opened 2024-03-25
fix(tests): add promoCode to thank you URL assertion

jamesgorrie:
After [enabling the tests again](https://github.com/guardian/support-frontend/pull/5843) - I had forgotten to account for the fact that [we now pass the `promoCode` to the thank you page](https://github.com/guardian/support-frontend/pull/5837/files#diff-fd63b7bdddd1f8c8a686a6ba94bee4323ebd6d88ea919d35cdcf7dbecd9a6db5R41-R46).

This fixes that.

### DIFF
--- a/support-e2e/tests/tieredCheckout.test.ts
+++ b/support-e2e/tests/tieredCheckout.test.ts
@@ -150,7 +150,10 @@ test.describe('Subscribe (S+) incl PromoCode via the Tiered checkout', () => {
 				'Pay £([0-9]+|(([0-9]+).([0-9]+))) per (year|month)',
 			);
 			await page.getByText(paymentButtonRegex).click();
-			await expect(page).toHaveURL(`/uk/thankyou`, { timeout: 600000 });
+			await expect(page).toHaveURL(
+				`/uk/thankyou?promoCode=PLAYWRIGHT_TEST_SPLUS`,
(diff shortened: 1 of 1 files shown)


support-e2e/tests/tieredCheckout.test.ts
```diff
@@ -150,7 +150,10 @@ test.describe('Subscribe (S+) incl PromoCode via the Tiered checkout', () => {
 				'Pay £([0-9]+|(([0-9]+).([0-9]+))) per (year|month)',
 			);
 			await page.getByText(paymentButtonRegex).click();
-			await expect(page).toHaveURL(`/uk/thankyou`, { timeout: 600000 });
+			await expect(page).toHaveURL(
+				`/uk/thankyou?promoCode=PLAYWRIGHT_TEST_SPLUS`,
```
johnduffell, line 154: can/should we strip the promoCode when checking the URL?

johnduffell (approved):
👍

support-e2e/tests/tieredCheckout.test.ts
jamesgorrie: We could use something like a `startsWith`, but we do expect it to be there, and this tests captures that.

PR #5846, opened 2024-03-25
Update tiered e2e promo tests to use Annual and Monthly `promoCodes`

jamesgorrie:
We have to use two different `promoCodes` for annual vs monthly as we want them to have different durations.

This does that, makes the `promoCode` configurable, and makes the expected text configrable.

Screenshot 2024-03-25 at 17 18 37


paul-daniel-dempsey (approved):
Looks good

PR #5848, opened 2024-03-26
Read thank you page amount from session storage

jamesgorrie:
Reads the amount paid in the checkout from session storage.

This is currently read from redux state outside of the `isOneOffPayPal` condition, which resets once you refresh the page.

We had a discussion as to whether this state should be stored in the `URLParams` going forward, and are a little undecided as it might expose information users aren't happy with, so thought this the quickest fix for now.

## Tested

- run playwright tests
- tested 1 off payment
- tested recurring payment
- tested S+ payment
- tested S+ promo payment


GHaberis (approved):
Looks good 👍

PR #5852, opened 2024-03-26
fix: update e2e promo tests to use multi-stage text checks

jamesgorrie:
Adds checks at the multiple stages of buying a product with a promo.

- Asserts the promo text on the tier card
- Asserts the total text on the checkout
- Asserts the "Buy" button text
- Asserts the promo text on the thank you page

This has picked up a bug that I was using the full amount for the promo text on the thank you page. Nice ✨ .

Lastly we add the annual promo example and header.

The script names for the e2e tests now follow `test-{env}` as it was a tiny bit fiddly to remember what it was.

[Trello card](https://trello.com/c/6EA1c4B7/705-3-tier-promotion-playwright-test-improve)

## Screenshots
Screenshot 2024-03-26 at 14 41 03

### DIFF
--- a/support-e2e/tests/tieredCheckout.test.ts
+++ b/support-e2e/tests/tieredCheckout.test.ts
@@ -34,12 +34,22 @@ const testDetailsPromo = [
 		frequency: 'Monthly',
 		promoCode: 'PLAYWRIGHT_TEST_SPLUS_MONTHLY',
 		expectedPromoText: '£8/month for the first 3 months, then £10/month',
+		/** This looks a little odd as it is searching for the original price struck out with the promo price */
+		expectedCheckoutTotalText: '£10 £8/month',
+		expectedCheckoutButtonText: 'Pay £8 per month',
+		expectedThankYouText:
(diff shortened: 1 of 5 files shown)


support-e2e/tests/tieredCheckout.test.ts
```diff
@@ -34,12 +34,20 @@ const testDetailsPromo = [
 		frequency: 'Monthly',
 		promoCode: 'PLAYWRIGHT_TEST_SPLUS_MONTHLY',
 		expectedPromoText: '£8/month for the first 3 months, then £10/month',
+		expectedCheckoutTotalText: '£10 £8/month',
```
jamesgorrie, line 37: I haven't really found a way to work out how to check for <s>strikethrough<s>.

jamesgorrie: I might see if we can add some useful content and check for that similar to Option 3 in [this article](https://www.webaxe.org/strikethrough-html-accessibility/).

johnduffell (approved):
nice work on the tidyup and extra fixes 👏

paul-daniel-dempsey (commented):
Looks Good

PR #5853, opened 2024-03-26
run scala builds on non main PRs

johnduffell:
Sometimes I do one PR and while it's waiting for review, I add another PR off the first one.

This means I am not blocked if the work depends on each other, but the second PR doesn't look like it includes the first PR, making it easier to review.

I noticed that if I do that, it doesn't run the builds on the second one, so I can't test in CODE.

This PR makes the Scala builds run on all PRs, and it also allows them to be manually run from the GH UI.

### DIFF
--- a/.github/workflows/payment-api-build.yml
+++ b/.github/workflows/payment-api-build.yml
@@ -1,10 +1,8 @@
 name: Build payment-api
 
 on:
-  workflow_dispatch:
   pull_request:
-    branches:
-      - main
+  workflow_dispatch:
   push:
     branches:
       - main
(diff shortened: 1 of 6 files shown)


jamesgorrie (approved):
Nice - I've been meaning to look why we had it setup like this, but can't find much of a reason.

johnduffell:
> Nice - I've been meaning to look why we had it setup like this, but can't find much of a reason.

not sure - it was already there in the first PR adding them https://github.com/guardian/support-frontend/pull/3531 so probably just copy and paste from elsewhere?

PR #5857, opened 2024-03-26
Checkout total text makes more sense to screen readers

jamesgorrie:
As per [this article](https://www.webaxe.org/strikethrough-html-accessibility) - struck through  text is not really read out as such, so what gets read from our current implementation is "£95 £76/year" (as reflected in the tests).

This keeps the same visual treatment but adds a bit of visually-hidden text making the total read "Was £10, now £8/month".

This is a nice option of writing playwright tests [as person not a robot](https://testing-library.com/docs/guiding-principles) helps us spot holes in our accessibility.

## Playwright tests
Screenshot 2024-03-26 at 16 05 10

## DOM structure
Screenshot 2024-03-26 at 16 11 19


paul-daniel-dempsey (approved):
Invisible tags so can be read out via screen readers, great idea.

PR #5859, opened 2024-03-27
fix payment success metric

johnduffell:
I shipped a PR yesterday to update some of the AWS clients to v2, including cloudwatch.
https://github.com/guardian/support-frontend/pull/5833

This involved a rewrite to use the new API.  Unfortunately the `dimensions` method in the new SDK overwrites the set of dimensions, whereas in the old SDK the similar method it added to them.

This means that when I ported over the old code, I made it go through each dimension, replacing them all with that one,r ather than adding them all.  This meant that they all ended up only under the Stage=PROD dimension, rather than the right PaymentProvider and Product as well.
image

To exacerbate that initial error, the test I was using to check it was only sending one dimension, so it worked fine (I ran it locally and just inspected the AWS console).  In CODE testing I didn't look at the PaymentSuccess metric in cloudwatch so I missed that it went to the wrong place.

This PR fixes the PaymentSuccess metric in support-workers, and updates the test to send a multi metric as part of the test.

### DIFF
--- a/support-modules/aws/src/main/scala/com/gu/aws/AwsCloudWatchMetricPut.scala
+++ b/support-modules/aws/src/main/scala/com/gu/aws/AwsCloudWatchMetricPut.scala
@@ -1,12 +1,14 @@
 package com.gu.aws
 
+import com.typesafe.scalalogging.LazyLogging
 import software.amazon.awssdk.regions.Region
 import software.amazon.awssdk.services.cloudwatch.CloudWatchClient
 import software.amazon.awssdk.services.cloudwatch.model.{Dimension, MetricDatum, PutMetricDataRequest, StandardUnit}
 
-import scala.util.Try
+import scala.jdk.CollectionConverters._
+import scala.util.{Failure, Success, Try}
 
-object AwsCloudWatchMetricPut {
+object AwsCloudWatchMetricPut extends LazyLogging {
   val client: CloudWatchClient =
     CloudWatchClient
       .builder()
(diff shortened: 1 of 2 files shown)


support-modules/aws/src/main/scala/com/gu/aws/AwsCloudWatchMetricPut.scala
```diff
@@ -27,32 +29,40 @@ object AwsCloudWatchMetricPut {
...
     val putMetricDataRequest = PutMetricDataRequest
       .builder()
       .namespace(request.namespace.value)
       .metricData(metricDatum1)
       .build()
 
-    Try(client.putMetricData(putMetricDataRequest)).map(_ => ())
+    val attempt = Try(client.putMetricData(putMetricDataRequest))
+    attempt match {
+      case Failure(exception) => logger.warn("metric send failed with " + exception.toString)
```
jamesgorrie, line 62:
```suggestion
      case Failure(exception) => logger.error("metric send failed with " + exception.toString)
```

?

support-services/src/test/scala/com/gu/aws/AwsCloudWatchMetricPutSpec.scala
```diff
@@ -1,15 +1,25 @@
...
-    AwsCloudWatchMetricPut(AwsCloudWatchMetricPut.client)(AwsCloudWatchMetricSetup.serverSideCreateFailure(Stages.CODE))
+  "payment success request" should "be logged to CloudWatch" in {
+    val cloudwatchEvent =
+      paymentSuccessRequest(
+        Stages.CODE,
+        true,
+        DirectDebit,
+        SupporterPlus(amount = 10, currency = GBP, billingPeriod = Monthly),
+      )
+    AwsCloudWatchMetricPut(AwsCloudWatchMetricPut.client)(cloudwatchEvent)
```
jamesgorrie, line 22: Great!

jamesgorrie (approved):
Nice test too

support-modules/aws/src/main/scala/com/gu/aws/AwsCloudWatchMetricPut.scala
johnduffell: yeah makes sense, I was assuming someone further up the tree[1] would log it properly and I wanted to avoid doubling up, but I don't see the harm.
[1]
https://github.com/guardian/support-frontend/blob/22d14edf4db1e4b5a4be7fc5c1c22369af16be30/support-workers/src/main/scala/com/gu/support/workers/lambdas/SendAcquisitionEvent.scala#L44

tomrf1 (approved):

johnduffell:
tested and working in CODE

johnduffell:
it's all good now 
image

PR #5850, opened 2024-03-26
support workers full aws sdk v2

johnduffell:
this PR updates support-workers so it doesn't use AWS SDK v1 at all.  This should hopefully reduce the Jar size and improve the startup slightly.

```
% jar xf support-frontend/support-workers/target/scala-2.13/support-workers.jar
% find com/amazonaws
com/amazonaws
com/amazonaws/services
com/amazonaws/services/lambda
com/amazonaws/services/lambda/runtime
com/amazonaws/services/lambda/runtime/CognitoIdentity.class
com/amazonaws/services/lambda/runtime/LambdaRuntime.class
com/amazonaws/services/lambda/runtime/ClientContext.class
com/amazonaws/services/lambda/runtime/LambdaRuntime$1.class
com/amazonaws/services/lambda/runtime/LambdaLogger.class
com/amazonaws/services/lambda/runtime/CustomPojoSerializer.class
com/amazonaws/services/lambda/runtime/Context.class
com/amazonaws/services/lambda/runtime/Client.class
com/amazonaws/services/lambda/runtime/RequestHandler.class
com/amazonaws/services/lambda/runtime/RequestStreamHandler.class
com/amazonaws/services/lambda/runtime/LambdaRuntimeInternal.class
com/amazonaws/services/lambda/runtime/logging
com/amazonaws/services/lambda/runtime/logging/LogLevel.class
com/amazonaws/services/lambda/runtime/logging/LogFormat.class
% 
```

follow on from https://github.com/guardian/support-frontend/pull/5833 
and incoporating https://github.com/guardian/support-frontend/pull/5836

Tested in CODE support-frontend/support-workers
- newspaper
- conribution

PR #5863, opened 2024-03-27
fix IT tests

johnduffell:
After the PR to update to AWS v2 in https://github.com/guardian/support-frontend/pull/5850 , I noticed one of the IT tests are failing due to a simplified return type.

```
'send' is a *void method* and it *cannot* be stubbed with a *return value*!
Voids are usually stubbed with Throwables:
    doThrow(exception).when(mock).someVoidMethod();
If you need to set the void method to do nothing you can use:
    doNothing().when(mock).someVoidMethod();
For more information, check out the javadocs for Mockito.doNothing().
```

This PR fixes it.

### DIFF
--- a/support-workers/src/test/scala/com/gu/support/workers/lambdas/FailureHandlerSpec.scala
+++ b/support-workers/src/test/scala/com/gu/support/workers/lambdas/FailureHandlerSpec.scala
@@ -117,8 +117,6 @@ class FailureHandlerITSpec extends AsyncLambdaSpec with MockContext {
 
     val testFields = FailedEmailFields.digitalPack("test@thegulocal.com", IdentityUserId("30001643"))
 
-    when(emailService.send(any[EmailFields])).thenReturn(Future.successful(()))
-
     val failureHandler = new FailureHandler(emailService)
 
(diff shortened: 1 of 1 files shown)


support-workers/src/test/scala/com/gu/support/workers/lambdas/FailureHandlerSpec.scala
```diff
@@ -117,8 +117,6 @@ class FailureHandlerITSpec extends AsyncLambdaSpec with MockContext {
 
     val testFields = FailedEmailFields.digitalPack("test@thegulocal.com", IdentityUserId("30001643"))
 
-    when(emailService.send(any[EmailFields])).thenReturn(Future.successful(()))
```
johnduffell, line 120: this returns Unit now [1], so there's no need to tell it to return anything specific

[1] https://github.com/guardian/support-frontend/pull/5850/files#diff-54f1a2c774e69ee7474fea35d7431f5dbc1d707ced5a0b32006ebddc7612915fR26

tomrf1 (approved):

PR #5864, opened 2024-03-27
Send promo code information in the acquisition event for supporter plus

rupertbates:
We now support promotions on the supporter plus product, however we are not currently sending promo code information to BigQuery with the acquisition event. This PR fixes that.


jamesgorrie (approved):

rupertbates:
Tested on CODE and promo code made it's way into BigQuery

PR #5868, opened 2024-03-28
Wrap skip links in navigation landmark

jamesgorrie:
We are seeing a "All page content should be contained by landmarks https://dequeuniversity.com/rules/axe/4.4/region?application=axeAPI" error from the [Axe plugin](https://github.com/dequelabs/axe-core-npm) we use to surface accessibility issues.

This was due to our skip-links not being wrapped in [landmark](https://developer.mozilla.org/en-US/blog/aria-accessibility-html-landmark-roles/) - which `nav` felt like the most appropriate.


paul-daniel-dempsey (approved):
So a screen reader will say 'skip to section navigation' presume.

jamesgorrie:
> So a screen reader will say 'skip to section navigation' presume.

Correct! It becomes available in the landmarks menu

Screenshot 2024-03-28 at 10 09 45

PR #5877, opened 2024-04-02
feat: use RRCP amounts for recurring tier

jamesgorrie:
Following this diagram, this implements the red bounding boxes, namely, **the lowest amounts** in **`Monthly` and `Annual`** being reflected in **Recurring contribution tier (Tier-1)**. 

Screenshot 2024-04-02 at 22 17 12

### RRCP
Screenshot 2024-04-02 at 22 15 17

### Monthly
Screenshot 2024-04-02 at 22 13 36

### Annual
Screenshot 2024-04-02 at 22 13 39

### DIFF
--- a/support-frontend/assets/pages/supporter-plus-landing/twoStepPages/threeTierLanding.tsx
+++ b/support-frontend/assets/pages/supporter-plus-landing/twoStepPages/threeTierLanding.tsx
@@ -366,6 +366,13 @@ export function ThreeTierLanding(): JSX.Element {
 				contributionTypeKeyOverride ?? tierPlanPeriod
 			].charges[countryGroupId];
 
+		if (cardTier === 1) {
+			tierPlanCountryCharges = {
+				...tierPlanCountryCharges,
+				price: recurringAmount,
+			};
+		}
+
 		if (cardTier === 2 && promotion) {
 			tierPlanCountryCharges = {
 				...tierPlanCountryCharges,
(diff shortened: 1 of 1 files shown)


support-frontend/assets/pages/supporter-plus-landing/twoStepPages/threeTierLanding.tsx
```diff
@@ -366,6 +366,13 @@ export function ThreeTierLanding(): JSX.Element {
 				contributionTypeKeyOverride ?? tierPlanPeriod
 			].charges[countryGroupId];
 
+		if (cardTier === 1) {
+			tierPlanCountryCharges = {
+				...tierPlanCountryCharges,
+				price: recurringAmount,
+			};
+		}
```
jamesgorrie, line 374: I think we should refactor how we transform the cards, and just read them from the catalog. But this will at least give people some functionality to test out.

jamesgorrie:
Before delivering this - we should definitely let the people using the tools that changing these values will have an effect on the site.

GHaberis:
@jamesgorrie this change will mean we can AB test different amounts in the lower tier! 🥳 , Amounts Test cohort allocation is done [here](https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/helpers/abTests/abtest.ts#L227) and the allocated cohort's amount config is stored in the Redux state you're now reading [here](https://github.com/guardian/support-frontend/pull/5877/files#diff-9e9d19a3470db4ff48cd4c73865be05d061001afc0751360451d6d509f390e9bR419).

GHaberis (approved):
Nice 👍

PR #5879, opened 2024-04-03
fix: use mockdate in storybook

jamesgorrie:
Uses [mockdate](https://www.npmjs.com/package/mockdate) in our storybook to void false positives in our test when the date changes. 

This was specifically noticed in [our support reminders](https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/components/thankYou/supportReminder/supportReminderItems.tsx#L94-L96) for the thank you page.

## Screenshots

Screenshot 2024-04-03 at 09 34 26


GHaberis (approved):
Nice improvement

PR #5881, opened 2024-04-03
Move `describeProduct` to it's own module

jamesgorrie:
This moves `describeProduct` to it's own module to be shared with `ThreeTierLanding`, and adds `benefits` and `benefitsDescription` to ensure we can render `ThreeTierLanding`.

We have also added a `GuardianWeeklyAndSupporterPlus` product which is a hack we'll need while we work out how we deal with bundles.

This is to help being [`ProductCatalog` into the `ThreeTierLanding`](https://github.com/guardian/support-frontend/issues/5882)


jamesgorrie:
Closing as I think I have a better solution to this,

PR #5883, opened 2024-04-03
fix: only show promotion is supporter plus on thank you page

jamesgorrie:
This is a hotfix and probably not **_the_** sustainable way to do this.

We will be looking at how data is passed to the thank you page in future which should help with this.

This only applies the promotion if it is `SupporterPlus`. I haven't found the route to how someone got to the checkout with a contribution amount and a `promoCode` e.g. https://support.theguardian.com/uk/contribute/checkout?selected-amount=4&selected-contribution-type=monthly&promoCode=50SPRING2024

The `promoCode` isn't reflected in the UI at that point:
Screenshot 2024-04-03 at 12 39 45

But when the `promoCode` is in the thank-you page URL, it borks:
https://support.theguardian.com/uk/thankyou?promoCode=50SPRING2024

I think this is because we read amount from `localStorage` _OR_ state. Hopefully we can get around this by removing this condition. 

But for now this fixes a known bug seen in `PROD` (reported by triage unit)

[**Trello Card**](https://trello.com/c/nYPxPdsy/746-severity-2-medium-bug-report-55-system-impacted-guardian-checkout-supporttheguardiancom)

### Before
Screenshot 2024-04-03 at 12 49 48

### After
Screenshot 2024-04-03 at 12 49 48

### DIFF
--- a/support-frontend/assets/helpers/redux/checkout/product/selectors/isSupporterPlus.ts
+++ b/support-frontend/assets/helpers/redux/checkout/product/selectors/isSupporterPlus.ts
@@ -15,6 +15,7 @@ export function isSupporterPlusFromState(state: ContributionsState): boolean {
 		state.page.checkoutForm.product.otherAmounts,
 		contributionType,
 	);
+	console.info('isSupporterPlusFromState', thresholdPrice, selectedAmount);
 	return selectedAmount >= thresholdPrice;
 }
 
--- a/support-frontend/assets/helpers/redux/checkout/product/selectors/isSupporterPlus.ts
+++ b/support-frontend/assets/helpers/redux/checkout/product/selectors/isSupporterPlus.ts
@@ -15,7 +15,7 @@ export function isSupporterPlusFromState(state: ContributionsState): boolean {
 		state.page.checkoutForm.product.otherAmounts,
 		contributionType,
 	);
-	console.info('isSupporterPlusFromState', thresholdPrice, selectedAmount);
+
 	return selectedAmount >= thresholdPrice;
 }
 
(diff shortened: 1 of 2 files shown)


support-frontend/assets/helpers/redux/checkout/product/selectors/isSupporterPlus.ts
```diff
@@ -15,6 +15,7 @@ export function isSupporterPlusFromState(state: ContributionsState): boolean {
 		state.page.checkoutForm.product.otherAmounts,
 		contributionType,
 	);
+	console.info('isSupporterPlusFromState', thresholdPrice, selectedAmount);
```
GHaberis, line 18: Do we want to remove this?

support-frontend/assets/pages/supporter-plus-thank-you/supporterPlusThankYou.tsx
```diff
@@ -140,14 +141,32 @@ export function SupporterPlusThankYou(): JSX.Element {
...
+	const thresholdPrice = useContributionsSelector((state) =>
+		getThresholdPrice(contributionType, state),
 	);
+	/**
+	 * We would normally use the isSuporterPlusFromState selector here,
+	 * but the amount can actually come from `localStorage`.
+	 *
+	 * We should clear this up when refactoring
+	 */
+	const isSupporterPlus = thresholdPrice ? amount >= thresholdPrice : false;
```
GHaberis, line 153: Do we want to check `contributionType !== 'ONE_OFF'` here too? EG. I'm wondering if a one off contribution with an amount above the threshold will lead to `isSupporterPlus`  being set as `true`... Unless `thresholdPrice` is a `falsy` value if `contributionType` is 'ONE_OFF'? In which case we don't need to check `contributionType !== 'ONE_OFF'` here.

jamesgorrie: Done - hopefully we'll be able to make this more specific to product soon.

support-frontend/assets/helpers/redux/checkout/product/selectors/isSupporterPlus.ts
jamesgorrie: 👀 thanks for catching

GHaberis (approved):
Looks good 👍

PR #5887, opened 2024-04-05
[RB] - update post support survey links for AU and ROW

rBangay:

Updating the survey link for the post checkout journey, with a specific link for Australia and the same 3 links for the rest of the world dependant on whether you are making a one off contribution, recurring below the supporter plus treshold and recurring above the supporter plus threshold.


jamesgorrie (approved):
Lovely

PR #5889, opened 2024-04-08
Use type `number` for amounts

jamesgorrie:
Maintains the source data type of `number` for `amounts` rather than transforming to `string`.

This tightens the type making it easier to work with as a `number` when doing calculations, and then render to `string` at the rendering edge, rather than having to flip-flop between conversions dependant on context.

This was inspired by wanting to `round` in the `format` methods in a refactor I am doing and the `number | string` type created unnecessary checks.

[These are the examples of wanting rounding to happen at all instances](https://github.com/search?q=repo%3Aguardian%2Fsupport-frontend++promoPriceRounded&type=code).

## How to test

Compile, E2E and visual tests are passing.

## Screenshots

Screenshot 2024-04-08 at 14 10 14

### DIFF
--- a/support-frontend/assets/components/orderSummary/contributionsOrderSummary.tsx
+++ b/support-frontend/assets/components/orderSummary/contributionsOrderSummary.tsx
@@ -170,7 +170,7 @@ export function ContributionsOrderSummary({
 	const formattedTotal = simpleFormatAmount(currency, total);
 	const formattedTotalExcludingPromo = simpleFormatAmount(
 		currency,
-		totalExcludingPromo ?? '',
+		totalExcludingPromo ?? 0,
 	);
 
 	return (
--- a/support-frontend/assets/components/otherAmount/otherAmount.tsx
+++ b/support-frontend/assets/components/otherAmount/otherAmount.tsx
@@ -9,7 +9,7 @@ const topSpacing = css`
 `;
 
 export type OtherAmountProps = {
-	selectedAmount: string;
+	selectedAmount: number | 'other';
 	otherAmount: string;
 	currency: IsoCurrency;
 	minAmount: number;
--- a/support-frontend/assets/components/priceCards/priceCardsContainer.tsx
+++ b/support-frontend/assets/components/priceCards/priceCardsContainer.tsx
@@ -55,7 +55,7 @@ export function PriceCardsContainer({
 		selectedAmounts,
 		paymentFrequency,
 		defaultAmount,
-	).toString();
+	);
 	const otherAmountErrors = useContributionsSelector(getOtherAmountErrors);
 	const otherAmount = otherAmounts[paymentFrequency].amount ?? '';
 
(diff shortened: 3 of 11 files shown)


support-frontend/assets/components/orderSummary/contributionsOrderSummary.tsx
```diff
@@ -170,7 +170,7 @@ export function ContributionsOrderSummary({
 	const formattedTotal = simpleFormatAmount(currency, total);
 	const formattedTotalExcludingPromo = simpleFormatAmount(
 		currency,
-		totalExcludingPromo ?? '',
+		totalExcludingPromo ?? 0,
```
jamesgorrie, line 173: I'm not 100% sure why this is the case, but this refactor should amount to the same thing.

support-frontend/assets/components/priceCards/priceCardsContainer.tsx
```diff
@@ -91,7 +91,7 @@ export function PriceCardsContainer({
 
 	return renderPriceCards({
 		currency: currencyId,
-		amounts: frequencyAmounts.map((amount) => amount.toString()),
+		amounts: frequencyAmounts,
```
jamesgorrie, line 94: This is the source of the type change.

support-frontend/assets/helpers/forms/checkouts.ts
```diff
@@ -228,12 +228,28 @@ function getPaymentDescription(
...
+	/**
+	 * This rounds a `number` to the second decimal.
+	 *
+	 * `Number.toFixed` returns a string which is not useful for calculations
+	 * and would need unnecessary type conversions
+	 */
+	return Math.round(amount * 1e2) / 1e2;
+}
+
+const simpleFormatAmount = (currency: Currency, amount: number): string => {
```
jamesgorrie, line 241: And this is the reading type change I wanted to just be `number`.

The method itself is refactored to do rounding. Happy to remove this as that was superfluous to this refactor.

support-frontend/assets/components/otherAmount/otherAmount.tsx
```diff
@@ -9,7 +9,7 @@ const topSpacing = css`
 `;
 
 export type OtherAmountProps = {
-	selectedAmount: string;
+	selectedAmount: number | 'other';
```
rupertbates, line 12: Where does 'other' come from? Would `number | undefined` maybe work better?

paul-daniel-dempsey (approved):
This simplifies the conversions required

support-frontend/assets/components/otherAmount/otherAmount.tsx
jamesgorrie: 'other' here is a valid value.
- `number` = the actual amount
- `other` = is for "Other / free text value"

We can definitely look at refactoring this out as a separate PR for clarity.

PR #5894, opened 2024-04-09
Fix layout issue on VAT compliant checkout

GHaberis:

I noticed a small layout issue on the VAT Compliant checkout where the horizontal rule at the top of the page was missing. A change went in recently to hide the country drop down (https://github.com/guardian/support-frontend/pull/5808) on the VAT Compliant checkout however the conditional was wrapping the entire `<Nav>` element, the horizontal rule isn't visible when the `<Nav>` isn't rendered, I've moved the conditional to the country drop down element within the `<Nav>` element. 

**Before**
Screenshot 2024-04-09 at 17 10 10

**After**
Screenshot 2024-04-09 at 17 13 03

### DIFF
--- a/support-frontend/assets/components/nav/nav.tsx
+++ b/support-frontend/assets/components/nav/nav.tsx
@@ -22,7 +22,7 @@ function Nav({
 	countryGroupIds,
 	selectedCountryGroup,
 	subPath,
-	countryIsAffectedByVATStatus,
+	countryIsAffectedByVATStatus = false,
 }: NavProps): JSX.Element {
 	return (
 		<Container
--- a/support-frontend/assets/components/nav/nav.tsx
+++ b/support-frontend/assets/components/nav/nav.tsx
@@ -15,12 +15,14 @@ interface NavProps {
 	countryGroupIds: CountryGroupId[];
 	selectedCountryGroup: CountryGroupId;
 	subPath: string;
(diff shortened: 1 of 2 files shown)


jamesgorrie (approved):

support-frontend/assets/components/nav/nav.tsx
```diff
@@ -15,12 +15,14 @@ interface NavProps {
 	countryGroupIds: CountryGroupId[];
 	selectedCountryGroup: CountryGroupId;
 	subPath: string;
+	countryIsAffectedByVATStatus: boolean;
```
jamesgorrie, line 18: You might want to have this as 
```suggestion
	countryIsAffectedByVATStatus?: boolean;
```

support-frontend/assets/components/nav/nav.tsx
```diff
@@ -15,12 +15,14 @@ interface NavProps {
...
 }
 
 function Nav({
 	countryGroupIds,
 	selectedCountryGroup,
 	subPath,
+	countryIsAffectedByVATStatus,
```
jamesgorrie, line 25: And this as 

```suggestion
	countryIsAffectedByVATStatus = false,
```

To avoid having to pass it through

GHaberis: @jamesgorrie suggestions looks good, thanks!

PR #5898, opened 2024-04-10
use actions-riff-raff for support-workers

johnduffell:
use https://github.com/guardian/actions-riff-raff for uploading support-workers

This means it's more secure (don't expose the riffraff key to the whole build.
Also we can use snapstart (possibly)

I compared the riffraff.yaml generated which is semantically the same, and the file names and sizes that appear in the artifact, and they are identical 
see https://riffraff.gutools.co.uk/deployment/request/deployFiles?projectName=support%3Asupport-workers-mono&id=18823

### DIFF
--- a/.github/workflows/support-workers-build.yml
+++ b/.github/workflows/support-workers-build.yml
@@ -60,14 +60,14 @@ jobs:
         run: |
           sbt "project support-workers" it:assembly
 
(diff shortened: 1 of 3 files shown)


jamesgorrie (approved):
Grand

PR #5901, opened 2024-04-10
turn on snapstart for support-workers

johnduffell:
REMINDER: approve and merge this other PR - https://github.com/guardian/support-frontend/pull/5902 - first, as otherwise cloudformation fails to deploy.

This PR updates support-workers to use snapstart, while still allowing quick deployment to CODE from local development.

The old `deployToCode` has been replaced by `deployToCodeQuick` which takes <30 seconds.

To update the cloudformation there is `deployToCodeWithCfn` - this is needed because the jar file name changes every time.  (The other option is to run the update-cloudformation-code script with the build number of the build that is already deployed).  deployToCodeWithCfn takes 2-3 minutes as it has to create a snapshot of each lambda which is surprisingly slow for some reason.  Hopefully this is a price worth paying.  If not we can parameterise the CFN to make the Alias point to $LATEST and remove the Version (or even turn off SnapStart) if we deploy CFN from locally.
Theoretically for CFN deploys we could have a flag to omit the version (or turn off snapstart) if we were testing e.g. some alarm changes and wanted to speed it up, but if someone is in that situation they could just comment out the snapstart line temporarily.

Also riffraff deploys have been slowed down to 4-5 minutes from ~30 seconds.

It seems like cold starts have improved slightly - a full support-workers cold execution has gone from ~30s to ~20s - I haven't really put much effort into exploiting snapstart, so the only thing it does during snapshot is loading the config data from AWS (it was using a lazy val before).  I'm sure we can warm it up more thoroughly with further code changes, there looks like there's an easy ~1.4s per lambda that I'm sure I can capture.

Thoughts please!

copies from https://github.com/guardian/support-frontend/pull/4869
builds on https://github.com/guardian/support-frontend/pull/5898


.github/workflows/support-workers-build.yml
```diff
@@ -83,8 +83,11 @@ jobs:
                 app: workers
                 parameters:
                   templatePath: cfn.yaml
+                dependencies: [support-workers]
               support-workers:
                 type: aws-lambda
+                actions:
+                  - uploadLambda
```
johnduffell, line 90: we only need to upload the jar file now, and the cfn will change every time ( 😢 slow) so it will delete the old version and create a new one rather than calling update-function which only updates $LATEST.
Perhaps there's scope for an improvement to riffraff to handle that rather than hacking it with CFN?

.github/workflows/support-workers-build.yml
```diff
@@ -96,8 +99,7 @@ jobs:
                     - "-FailureHandlerLambda-"
                     - "-SendAcquisitionEventLambda-"
                     - "-PreparePaymentMethodForReuseLambda-"
-                  fileName: support-workers.jar
-                dependencies: [cfn]
```
johnduffell, line 100: now the Jar is uploaded first and cfn second, so the old cyclical dependency is gone 🌟

.github/workflows/support-workers-build.yml
```diff
@@ -96,8 +99,7 @@ jobs:
...
+                  fileName: support-workers-${{ github.run_number }}.jar
```
johnduffell, line 102: jar name has to change every time, otherwise lambda won't let us publish a new version.  I suspect if we change anything about the Lambda it would let us, perhaps even the description?

support-workers/build.sbt
```diff
@@ -49,14 +49,34 @@ IntegrationTest / assembly / assemblyMergeStrategy := {
 IntegrationTest / assembly / test := {}
 assembly / aggregate := false
 
-lazy val deployToCode =
-  inputKey[Unit]("Directly update AWS lambda code from CODE instead of via RiffRaff for faster feedback loop")
+// We also have to put the build number in the .jar, because AWS refuses to create a new lambda version if the jar is the same!
```
johnduffell, line 52: this came from the other PR but I haven't explicitly tested it (maybe TODO)

support-workers/build.sbt
```diff
@@ -49,14 +49,34 @@ IntegrationTest / assembly / assemblyMergeStrategy := {
...
+Compile / resourceGenerators += Def.task {
+  val buildNumber = lambdaVersion.value
+  val file = resourceManaged.value / "build.number"
+  IO.write(file, buildNumber)
+  Seq(file)
+}.taskValue
+
+lazy val lambdaVersion = taskKey[String]("unique value to allow us to create a new AWS lambda version every time")
```
johnduffell, line 60: sbt lets us create a "task" which is executed for each command, but it's shared between all the steps, this means every time we run deployToCodeWithCfn, it uses a new value across all stages of the build

support-workers/build.sbt
```diff
@@ -69,10 +89,26 @@ deployToCode := {
     "-FailureHandlerLambda-",
     "-SendAcquisitionEventLambda-",
     "-PreparePaymentMethodForReuseLambda-",
-  ).foreach { functionPartial =>
+  ).par.foreach { functionPartial =>
```
johnduffell, line 92: much quicker if we do them in parallel

support-workers/build.sbt
```diff
@@ -69,10 +89,26 @@ deployToCode := {
...
+  import scala.sys.process._
+  val log = streams.value.log
+  val version = lambdaVersion.value
+  val s3Bucket = "support-workers-dist"
+  val s3Path = s"support/CODE/support-workers/support-workers-$version.jar"
+  val assemblyJar = assembly.value
+  log.info(s"generated jar $assemblyJar, about to upload to S3...")
+  log.info((s"aws s3 cp $assemblyJar s3://" + s3Bucket + "/" + s3Path + " --profile membership --region eu-west-1").!!)
+  log.info("now running scripts/update-code-cloudformation.sh in order to activate the uploaded JAR")
+  log.info((s"/bin/bash ./support-workers/scripts/update-code-cloudformation.sh $version").!!)
```
johnduffell, line 113: we pass in the version to this script, which can then pass it in turn to the node script

support-workers/cloud-formation/src/templates/lambda.yaml
```diff
@@ -9,8 +9,25 @@
...
+  Properties:
+    FunctionName: !Ref {{name}}Lambda
+    FunctionVersion: !GetAtt [ {{name}}CurrentVersion, Version ]
+    Name: SNAPSTART
+
+{{name}}CurrentVersion:
+  Type: "AWS::Lambda::Version"
+  Properties:
+    FunctionName: !Ref {{name}}Lambda
+    Description: {{> GITHUB_RUN_NUMBER}}
```
johnduffell, line 31: the Alias is pointed by the state machine now, we just update the alias automatically to point to the relevant version without having to update the state machine

johnduffell, line 31: we need to update the description so that it deletes and recreates the version each time.

support-workers/cloud-formation/src/templates/lambda.yaml
```diff
@@ -9,8 +9,25 @@
...
     Architectures:
-      - arm64
+      - x86_64
+
+{{name}}AliasForLambda:
+  Type: "AWS::Lambda::Alias"
+  Properties:
+    FunctionName: !Ref {{name}}Lambda
+    FunctionVersion: !GetAtt [ {{name}}CurrentVersion, Version ]
+    Name: SNAPSTART
```
johnduffell, line 25: wasn't sure what to call the alias, it's just the latest production version, but I don't want to confuse with `$LATEST` by calling it `latest` or similar.

support-workers/src/main/scala/com/gu/services/Services.scala
```diff
@@ -23,7 +23,7 @@ import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.duration._
 
 trait ServiceProvider {
-  private lazy val config = Configuration.load()
+  private val config = Configuration.load()
```
johnduffell, line 26: this is the only optimisation (so far) to make this happen during warmup.  It already makes things quicker to cold start, but I'm sure we can do more.

johnduffell:
Closing this PR because although it was an interesting 10% time project, I went off the idea of snapstart in production. Noone had anything good to say about the complexity added ( you can see for yourself the result in this PR), and quite a few of us were dubious at the least about the benefits.

PR #5902, opened 2024-04-10
revert support-workers to x86 in preparation for snapstart

johnduffell:
Snapstart only works on x86, but we run support-workers on ARM as it's slightly cheaper (see https://github.com/guardian/support-frontend/pull/5794 )

If you enable snapstart and switch to x86_64 in the same deploy, it fails with `Resource handler returned message: "arm64 is not supported for SnapStart enabled functions.`. 

however if you do one then the other, it works.

We will lose the small saving from using ARM, but snapstart should give a better user experience (and developer expeirence for CODE)

Snapstart main PR (in progress) is here: https://github.com/guardian/support-frontend/pull/5901

(deleting the riffraff file is a tidy up that I missed from https://github.com/guardian/support-frontend/pull/5898 )


support-workers/riff-raff.yaml
```diff
@@ -1,32 +0,0 @@
-stacks: [support]
```
johnduffell, line 1: forgot to delete this in https://github.com/guardian/support-frontend/pull/5898

jamesgorrie (approved):
> We will lose the small saving from using ARM

Out of interest do we have an idea of how much this is? Useful to help make decisions about this in the future.

johnduffell:
> > We will lose the small saving from using ARM
> 
> Out of interest do we have an idea of how much this is? Useful to help make decisions about this in the future.

it's quite small, all off the top of my head - our whole lambda bill is barely into 3 figures a month, and ARM lambdas are ~25% cheaper and very marginally slower for our workload.  So even if support-workers was most of our cost, worst case it would be maybe 10 dollars a month.  But I think it's much smaller than that.

Having said all that I just popped in to close this PR because although it was an interesting 10% time project, I went off the idea of snapstart in production. Noone had anything good to say about the complexity added, and quite a few people were dubious about the benefits.

PR #5904, opened 2024-04-11
fix: remove promotion value from S+ check

jamesgorrie:
- Removes checking the promotion value in S+ threshold check.
- Remove application of promotion on non-SupporterPlus products (recurring and one-off)

This was a hangover for when we included the promotion deduction in the `selectedAmount`.

We need to look into why this wasn't caught by the e2e tests and there's a potential to refactor the [`getPromotion` code](https://github.com/guardian/support-frontend/blob/5eb34e324d78d563bb280229b7639683307ae2bb/support-frontend/assets/helpers/productPrice/promotions.tsx#L103) to account for this at the source.

### DIFF
--- a/support-frontend/assets/pages/supporter-plus-landing/twoStepPages/secondStepCheckout.tsx
+++ b/support-frontend/assets/pages/supporter-plus-landing/twoStepPages/secondStepCheckout.tsx
@@ -60,9 +60,7 @@ export function SupporterPlusCheckout({
 		getUserSelectedAmountBeforeAmendment,
 	);
 	const otherAmount = useContributionsSelector(getUserSelectedOtherAmount);
-	const amountIsAboveThreshold = useContributionsSelector(
-		isSupporterPlusFromState,
-	);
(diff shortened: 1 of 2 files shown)


support-frontend/assets/pages/supporter-plus-landing/twoStepPages/secondStepCheckout.tsx
```diff
@@ -94,13 +92,15 @@ export function SupporterPlusCheckout({
...
 	);
 
-	const promotion = useContributionsSelector((state) =>
-		getPromotion(
-			state.page.checkoutForm.product.productPrices,
-			countryId,
-			state.page.checkoutForm.product.billingPeriod,
-		),
-	);
+	const promotion = isSupporterPlus
```
paul-daniel-dempsey, line 95: Perhaps comment to say `currently promotions can only be applied to recurring S+ products only`

paul-daniel-dempsey (approved):
Looks good, small comment suggestion

PR #5905, opened 2024-04-11
Allow recurring contributions above s+ threshold for users affected by VAT issue

GHaberis:

We are restricting acquisitions in some countries using the `VAT_COMPLIANCE` amounts test, so we should only offer one-off OR recurring contributions beneath the S+ threshold.

However the amounts test does not take currency into account, it shows the same configured amounts in every currency.

The highest amount on the test is currently 12, which is under the Supporter+ threshold for USD, CAD, AUD and NZD, _but_ is above it for GBP and EUR.

Therefore if a country is in the `VAT_COMPLIANCE` test but also within the the GBP or EUR country groups, the checkout shows Supporter+ and benefits and allows them to checkout with S+, which we don't want to happen!

You can see it happen by forcing the country on the checkout like so:

https://support.theguardian.com/eu/contribute?country=BA 

This mimics Bosnia and Herzegovina, which is in the `VAT_COMPLIANCE` test,  but also in the EUR country group - where the S+ threshold is 10. 

To prevent this this PR makes the following changes:

If a user is the `VAT_COMPLIANCE` test:

- Don't assign them Supporter Plus even if they have selected to make a recurring contribution at or above the S+ threshold
- Hide the S+ benefits container

The product listed in the payload to the `/create` endpoint will be "Contribution".

Screenshot 2024-04-11 at 17 50 51

### Checkout (Note: No benefits container)
https://support.thegulocal.com/eu/contribute?country=BA (eg. Bosnia and Herzegovina)

support thegulocal com_eu_contribute_country=BA

### Thank you page (Note: Non Supporter Plus variant)

support thegulocal com_eu_thankyou

### MMA (Note: Product is a Recurring Contribution)

Screenshot 2024-04-11 at 16 33 56


jamesgorrie (approved):

GHaberis:
@jamesgorrie I picked up a bug in testing on the TY page of this checkout. When checking out at or above the S+ threshhold I was seeing the following header:

monthly-at-fail

The issue was the `isSupporterPlus` check on the TY page was resolving as `true` if the contribution value was  at or above the S+ threshhold and the frequency was not One Off, I've therefore updated the condition to also check that the user is not in the `VAT_COMPLIANCE` test.

jamesgorrie:
> I picked up a bug in testing on the TY page of this checkout

@GHaberis - it seems the fix works, but it is a strange bug. Maybe we can look at a more sustainable fix than just excluding values repeatadly.

jamesgorrie (approved):

GHaberis:
> > I picked up a bug in testing on the TY page of this checkout
> 
> @GHaberis - it seems the fix works, but it is a strange bug. Maybe we can look at a more sustainable fix than just excluding values repeatadly.

Yeah it doesn't feel great to have the same/similar logic in multiple places. I wondered whether we could be using the `isSupporterPlusFromState` function I'd already amended on the TY page too.

PR #5906, opened 2024-04-11
Prevent creation of a supporter plus sub with a negative contribution

rupertbates:
We have recently had a bug [documented here](https://docs.google.com/document/d/15P66QvoEPdt3Nd_td5D6FQ2Fm-inrzO04H3GFOnUH2k/edit) which enabled the creation of supporter plus subscriptions with a negative contribution element.

This PR prevents that by failing if the contribution amount is less than zero.

### DIFF
--- a/support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/SupporterPlusSubcriptionBuilder.scala
+++ b/support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/SupporterPlusSubcriptionBuilder.scala
@@ -7,7 +7,7 @@ import com.gu.support.config.{TouchPointEnvironment, ZuoraSupporterPlusConfig}
 import com.gu.support.promotions.{PromoError, PromotionService}
 import com.gu.support.workers.Monthly
 import com.gu.support.workers.ProductTypeRatePlans.supporterPlusRatePlan
-import com.gu.support.workers.exceptions.CatalogDataNotFoundException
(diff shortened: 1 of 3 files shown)


support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/SupporterPlusSubcriptionBuilder.scala
```diff
@@ -35,6 +35,11 @@ class SupporterPlusSubcriptionBuilder(
     val todaysDate = dateGenerator.today
 
     val contributionAmount = state.product.amount - getBaseProductPrice(productRatePlanId, state.product.currency)
+    if (contributionAmount < 0)
```
jamesgorrie, line 38: Should we block for `<= 0`?

jamesgorrie (approved):

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/SupporterPlusSubcriptionBuilder.scala
rupertbates: No because 0 is expected, it's the normal case where people are just paying the subscription amount

jamesgorrie: Gotcha, it's the contribution amount.

PR #5910, opened 2024-04-12
Use product catalog types in three tier

jamesgorrie:

PR #5913, opened 2024-04-12
Use product catalog on checkout

jamesgorrie:
Depends on https://github.com/guardian/support-frontend/pull/5910

Uses the `productCatalogDescription` on the checkout.

## Why are you doing this?
To make sure we have and use 1 way to describe the product catalog.


jamesgorrie:
Closing as got into a weird git state

PR #5914, opened 2024-04-12
Use productCatalogDescription on checkout

jamesgorrie:
Depends on https://github.com/guardian/support-frontend/pull/5910

Uses the `productCatalogDescription` on the checkout.

## Why are you doing this?
To make sure we have and use 1 way to describe the product catalog.

### DIFF
--- a/support-frontend/assets/helpers/productCatalog.tsx
+++ b/support-frontend/assets/helpers/productCatalog.tsx
@@ -1,3 +1,6 @@
+import { newspaperCountries } from './internationalisation/country';
+import { gwDeliverableCountries } from './internationalisation/gwDeliverableCountries';
+
 export const productCatalog = window.guardian.productCatalog;
 
 export type ProductDescription = {
@@ -6,6 +9,8 @@ export type ProductDescription = {
 	benefitsSummary?: Array<string | { strong: boolean; copy: string }>;
 	offers?: Array<{ copy: JSX.Element; tooltip?: string }>;
 	offersSummary?: Array<string | { strong: boolean; copy: string }>;
+	ratePlans: Record<string, { billingPeriod: 'Annual' | 'Month' | 'Quarter' }>;
+	deliverableTo?: Record<string, string>;
 };
 
 export const productCatalogDescription: Record<string, ProductDescription> = {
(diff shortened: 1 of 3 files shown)


support-frontend/assets/helpers/productCatalog.tsx
```diff
@@ -21,6 +26,21 @@ export const productCatalogDescription: Record<string, ProductDescription> = {
 				tooltip: `Guardian Weekly is a beautifully concise magazine featuring a handpicked selection of in-depth articles, global news, long reads, opinion and more. Delivered to you every week, wherever you are in the world.`,
 			},
 		],
+		ratePlans: {
+			MonthlyWithGuardianWeekly: {
+				billingPeriod: 'Month',
```
jamesgorrie, line 31: In the long term, I'd like to try and get this into [the API](https://product-catalog.guardianapis.com/product-catalog.json).

It needs a bit of thinking though as the `billingPeriod` is actually part of the `charge` which we don't really use in the UI.

I think we might be able to determine this value as in practise all `charges` for our `ratePlans` have the same `billingPeriod` so we could do some sort of validation on that, and then hoist the property up to the ratePlan if it works.

jamesgorrie: [Here's a discussion](https://github.com/guardian/support-frontend/discussions/5919)

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -391,7 +329,8 @@ export function Checkout() {
 									formData.get('billingAddressMatchesDelivery') === 'yes';
 
 								const billingAddress =
-									product.showAddressFields && !billingAddressMatchesDelivery
+									productDescription.deliverableTo &&
```
Comment (paul-daniel-dempsey) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:332]: Like this `deliverableTo` :)

paul-daniel-dempsey (approved):
Looks good, feels a little like an intermediate step with this model (changes still to come) and wondered whether we should now relocate any `BillingPeriod`  conversions to a standalone single location/file (like done with displaying the currency amounts to 2dp)?

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -271,7 +199,17 @@ export function Checkout() {
 	const currentRatePlan = currentProduct.ratePlans[query.ratePlan];
 	const currentPrice = currentRatePlan.pricing[currentCurrencyKey];
 
-	const product = describeProduct(query.product, query.ratePlan);
+	const productDescription = productCatalogDescription[query.product];
+	const ratePlanDescription = productDescription.ratePlans[query.ratePlan];
+	let paymentFrequency;
+	if (ratePlanDescription.billingPeriod === 'Annual') {
```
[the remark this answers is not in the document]
Comment (jamesgorrie) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:205]: Yea, it definitely feels like time.

We will need to get this into the API first, which I have put some pairing time in to do.

jamesgorrie:
> `BillingPeriod` conversions to a standalone single location/file

Nice idea - let's get the data into the API first and we can start consolidating our transformations too!

PR #5921, opened 2024-04-16
fix: use new button label for CMP e2e test

jamesgorrie:
Howfix to get e2e tests working. 

We'll follow up with a test that isn't as brittle given we don't control the copy of the CMP.


paul-daniel-dempsey (approved):
Looks Good

PR #5922, opened 2024-04-16
Three tier get thresholdPrices from productCatalog / Zuora

paul-daniel-dempsey:

Replace the hard-coded lower and(& upper) thesholds stored here ->
`support-frontend/assets/helpers/supporterPlus/benefitsThreshold.ts`

With those in the `productCatalog` stored here (fed from Zuora) ->
`support-frontend/assets/helpers/productCatalog.tsx`

This impacts ->
1. `supporterPlusThankyou` (no-op)
2. `supporterPlusLegal` (no-op)
3. `checkoutTopUpAmounts` stories updated (remove upper threshold dependency, does not exist in Zuora)
4. `supporterPlusThankyou` stories (provide a thresholdPrice overide as it cannot load the productCatalog from Zuora)

[**Trello Card**](https://trello.com/c/ij1KtorF/563-promos-on-supporterplus-part-4b-read-s-threshold-from-zuora-product)

## Why are you doing this?

Allow Zuora to drive the threshold prices going forwards, thus this file will not need to be updated to keep in-line.


jamesgorrie (approved):
Great - I think we should have a think about how we include this for storybook, but we can as a secondary PR.

PR #5920, opened 2024-04-15
Make data valid to `POST` to `/subscribe/create` endpoint

jamesgorrie:
Adds the following fields to the POST data sent to `/subscribe/create`

- [`productType: 'Contribution' | 'SupporterPlus' | 'DigitalPack' | 'Paper' | 'GuardianWeekly'`](https://github.com/guardian/support-frontend/blob/main/support-models/src/main/scala/com/gu/support/workers/Products.scala#L17-L66)
- `currency`
- `billingPeriod` - this is hard-coded and will be introduced properly [here](https://github.com/guardian/support-frontend/pull/5914)
- `fulfillmentOptions`

To achieve this we needed to add
- state handlers for `country`
- POST handler to /subscribe/create

PR #5923, opened 2024-04-16
refactor: make checkout onsubmit async

jamesgorrie:
No-op refactor to make the `onSubmit` a little more readable, and to make it more procedural when adding new payment methods and products.

You can see this starting to happen here #5920

### DIFF
--- a/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
@@ -262,6 +262,99 @@ export function Checkout() {
 	const [accountHolderConfirmation, setAccountHolderConfirmation] =
 		useState(false);
 
+	const formOnSubmit = async (formData: FormData) => {
+		/**
+		 * The validation for this is currently happening on the client side form validation
+		 * So we'll assume strings are not null.
+		 * see: https://developer.mozilla.org/en-US/docs/Learn/Forms/Form_validation
+		 */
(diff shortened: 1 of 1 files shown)


support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -262,6 +262,99 @@ export function Checkout() {
...
 		useState(false);
 
+	const formOnSubmit = async (formData: FormData) => {
+		/**
+		 * The validation for this is currently happening on the client side form validation
+		 * So we'll assume strings are not null.
+		 * see: https://developer.mozilla.org/en-US/docs/Learn/Forms/Form_validation
+		 */
+		const data = {
+			firstName: formData.get('firstName') as string,
```
Comment (rupertbates) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:272]: Why do we need all these `as string`s?

Comment (rupertbates) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:272]: Oh it's because they could be null right?

tomrf1 (approved):

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
Comment (jamesgorrie) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:272]: They could be null when using the `get` API - but we know they're not. 

There's probably a much nicer way to do this - I am waiting for the validation of this piece of work to help elucidate that.

PR #5924, opened 2024-04-16
Post to subscription endpoint with valid(ish) data

jamesgorrie:
Getting the `POST` data for the `/subscribe/create` endpoint into a more complete state.

I've `Omit`ted some of the properties from [`RegularPaymentRequest`](https://github.com/guardian/support-frontend/blob/9d5a6a64e40fedf57541e14409488033a8621d1b/support-frontend/assets/helpers/forms/paymentIntegrations/readerRevenueApis.ts#L153), as they will warrant their own PRs.

There's a few changes in here:
* addresses - refactors the code a little to match the way we actually collect data i.e. if an item is deliverable, we require a `deliveryAddress` and either copy that to `billingAddress`. If it's not, we just need the country in the `billingAddress`
* Appends `ophanIds` and `referrerAcquisitionData` tracking data to the request
* `POST`s the data to the endpoint, which will break until the `Omit` it removed


support-frontend/assets/helpers/forms/paymentIntegrations/readerRevenueApis.ts
```diff
@@ -135,11 +135,11 @@ export type RegularPaymentFields =
...
-	state: UsState | CaState | null;
-	lineOne: Option<string>;
-	lineTwo: Option<string>;
-	postCode: Option<string>;
-	city: Option<string>;
+	state?: UsState | CaState | null;
+	lineOne?: string;
+	lineTwo?: string;
+	postCode?: string;
+	city?: string;
```
jamesgorrie, line 142: [This is a more idiomatic TypeScript way of having optional properties](https://www.typescriptlang.org/docs/handbook/2/objects.html#optional-properties).

support-frontend/assets/helpers/forms/paymentIntegrations/readerRevenueApis.ts
```diff
@@ -135,11 +135,11 @@ export type RegularPaymentFields =
...
-	state: UsState | CaState | null;
-	lineOne: Option<string>;
-	lineTwo: Option<string>;
-	postCode: Option<string>;
-	city: Option<string>;
+	state?: UsState | CaState | null;
+	lineOne?: Option<string>;
+	lineTwo?: Option<string>;
+	postCode?: Option<string>;
+	city?: Option<string>;
```
jamesgorrie, line 142: [`?:` is a more idiomatic TypeScript way of having optional properties](https://www.typescriptlang.org/docs/handbook/2/objects.html#optional-properties).

I'll put in another PR to remove the `Option` as it's a little more far reaching.

johnduffell (approved):
looks good, and nicely broken down into a small PR 👍 
something I would mention although you probably know, is that you have to be careful with the country with GW as it's the devliery country that decides the currency (unless the two addrresses are ticked as the same). I only mention as we've had lots of bugs in the past there!

jamesgorrie:
> you have to be careful with the country with GW

Thanks @johnduffell - I'll find some time to pair on maybe writing a test for this.

PR #5925, opened 2024-04-16
Remove countriesAffectedByVATStatus list from support-frontend code

GHaberis:

We currently have an array of country codes named `countriesAffectedByVATStatus` in the support-frontend code. We search this array using a user's `countryId` in a few places to determine whether a user is subject to the VAT compliance amounts checkout (eg. pre-3-tier landing page, S+ isn't available).

The issue we've experienced is this list needs to be maintained to prevent it falling out of sync with the list of countries targeted by the `VAT_COMPLIANCE` amounts test, this list is defined in RRCP.

Instead of having two ways to check whether a user is subject to the VAT compliance issues it would be better to have one. 

This PR implements that by replacing any instances where we interrogated `countriesAffectedByVATStatus` with a check to see whether a user is participating in the `VAT_COMPLIANCE` amounts test.

I've also centralised this check in to a function `isSubjectToVatCompliantAmounts` exported by a new module `support-frontend/assets/helpers/vatCompliance.ts`, this module contains two exports:

1. vatCompliantAmountsTestName: So we don't have multiple references to "VAT_COMPLIANCE" in the code
2. isSubjectToVatCompliantAmounts: A function that checks the `testName` from `state.common.amounts`

## What are you NOT doing in this PR?

Although it's not used currently `countriesAffectedByVATStatus` was helpful to use in conjunction with the `omitCountries` property on AB test definitions, this allowed us to say we don't want users in the affected countries to be allocated to specific AB tests.

I'm planning on introducing another means of achieving this that doesn't require us to store a list of country codes in our codebase. I'll do this in a follow up PR.


jamesgorrie (approved):
I think I get this flow. Just to check:

* `state.common.amounts` is `VAT_COMPLIANCE` if you have the conditions met i.e. you're in a certain country. This value is set upstream
* We use that as the test for `isSubjectToVatCompliantAmounts`

GHaberis:
> I think I get this flow. Just to check:
> 
> * `state.common.amounts` is `VAT_COMPLIANCE` if you have the conditions met i.e. you're in a certain country. This value is set upstream
> * We use that as the test for `isSubjectToVatCompliantAmounts`

@jamesgorrie that's correct, on page set-up we call [getAmountsTestVariant](https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/helpers/abTests/abtest.ts#L227 ) and store the result in `state.common.amounts`,   `getAmountsTestVariant` checks whether the user's `countryId` is present in the RRCP defined list of countries that the `VAT_COMPLIANCE` test tragets.

PR #5926, opened 2024-04-17
Allow for 0 `discountPrice` (100% discount)

jamesgorrie:
We were applying the promotion based on `!!promotion.discountedPrice`. This would evaluate to `false` in the case that `discountPrice === 0`.

It is now the case that if the promotion exists, we will apply it. If the `discountPrice` is missing, we will fallback to full price, and if `numberOfDiscountedPeriods` is missing, we will fallback to `0`.

So without neither of those you might expect to see
> You'll pay £10/month for the first 0 months, then £10/month afterwards unless you cancel.

This is odd - but then again it's odd input. If we'd like to guard against this, I'd suggest we do it at the source.

### Landing
Screenshot 2024-04-17 at 12 21 04

### Checkout
Screenshot 2024-04-17 at 12 21 09

### Thank you
Screenshot 2024-04-17 at 12 22 21


GHaberis (approved):
Nice 👍

PR #5928, opened 2024-04-17
fix: noUncheckedIndexedAccess in abtests

jamesgorrie:
Part of https://github.com/guardian/support-frontend/issues/5829

Fixes `noUncheckedIndexedAccess` errors in `abtests`.

- Mostly adds `undefined` checks
- Sets the `@typescript-eslint/no-unnecessary-condition": "off"` eslint rule on the file. This can sometime leave unnecessary code, but never buggy code
- Adds the [non-null assertion](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator) as a bit of a cheat - but the code doesn't change so we should be good


andrewHEguardian (approved):
Looks good to me 👍

PR #5932, opened 2024-04-18
Restore ability to exclude countries subject to VAT compliant Amounts from client-side AB tests

GHaberis:

In https://github.com/guardian/support-frontend/pull/5925 I removed the array `countriesAffectedByVATStatus` from the codebase, and replaced any instances where it had been used (to identify whether a user's country was affected by VAT compliance) with a look-up that checked whether they were assigned to the `VAT_COMPLIANT` amounts instead.

`countriesAffectedByVATStatus` had previously been used in conjunction with the `omitCountries` property in our AB test definitions to make sure users viewing the VAT Compliant checkout weren't registered as participating in AB tests that didn't touch this checkout flow (eg. 3-tier tests). By removing `countriesAffectedByVATStatus`  I removed an easy way to achieve this omission.

This PR restores that by introducing a new prop to our AB test definitions `excludeCountriesSubjectToVatCompliantAmounts`, this is a required `boolean`. If it's `true` we will check whether a user has been assigned to the `VAT_COMPLIANT` amounts test and if so we will not assign a user to the AB test.

### DIFF
--- a/support-frontend/assets/helpers/abTests/abtest.ts
+++ b/support-frontend/assets/helpers/abTests/abtest.ts
@@ -78,30 +78,41 @@ export type Test = {
 	// before activating this test eg. '/(uk|us|au|ca|nz)/subscribe$'
 	targetPage?: string | RegExp;
 	omitCountries?: IsoCountry[];
+	excludeCountriesSubjectToVatCompliantAmounts: boolean;
 };
 
 export type Tests = Record<string, Test>;
 
 // ----- Init ----- //
 
-function init(
-	country: IsoCountry,
-	countryGroupId: CountryGroupId,
-	abTests: Tests = tests,
-	mvt: number = getMvtId(),
-	acquisitionDataTests: AcquisitionABTest[] = getTestFromAcquisitionData() ??
-		[],
-): Participations {
+type ABtestInitalizerData = {
+	countryId: IsoCountry;
+	countryGroupId: CountryGroupId;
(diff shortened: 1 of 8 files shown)


support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -78,30 +78,41 @@ export type Test = {
...
 
-function init(
-	country: IsoCountry,
-	countryGroupId: CountryGroupId,
-	abTests: Tests = tests,
-	mvt: number = getMvtId(),
-	acquisitionDataTests: AcquisitionABTest[] = getTestFromAcquisitionData() ??
-		[],
-): Participations {
+type ABtestInitalizerData = {
```
tomrf1, line 88: 👏

support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -78,30 +78,41 @@ export type Test = {
...
+
+function init({
+	countryId,
+	countryGroupId,
+	selectedAmountsVariant,
+	abTests = tests,
+	mvt = getMvtId(),
+	acquisitionDataTests = getTestFromAcquisitionData() ?? [],
+}: ABtestInitalizerData): Participations {
+	console.log('*** INIT ***');
```
tomrf1, line 105: log!

tomrf1 (approved):
👍

PR #5933, opened 2024-04-18
Use "Contributions Only" nomenclature

GHaberis:

We'd like to introduced "Contributions Only" nomenclature for the contributions only variant of the checkout. This PR updates naming conventions to do that.

I've deleted the module `helpers/vatCompliance` and have moved the exports into `helpers/contributions` as this feels like an appropriate location for these exports.

Naming updates:

`isSubjectToVatCompliantAmounts` -> `isSubjectToContributionsOnlyAmounts`
`vatCompliantAmountsTestName` -> `contributionsOnlyAmountsTestName`
`excludeCountriesSubjectToVatCompliantAmounts` -> `excludeCountriesSubjectToContributionsOnlyAmounts`


support-frontend/assets/helpers/contributions.ts
```diff
@@ -501,6 +501,15 @@ const contributionTypeAvailable = (
 		(settings) => settings.contributionType === contributionType,
 	);
 
+const contributionsOnlyAmountsTestName = 'VAT_COMPLIANCE';
+
+const isSubjectToContributionsOnlyAmounts = (
```
paulbrown1982, line 506: I wonder if a better name for `isSubjectToContributionsOnlyAmounts` might be something like `countriesWhereWeOfferContributionsOnly`?

support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -78,7 +78,10 @@ export type Test = {
 	// before activating this test eg. '/(uk|us|au|ca|nz)/subscribe$'
 	targetPage?: string | RegExp;
 	omitCountries?: IsoCountry[];
-	excludeCountriesSubjectToVatCompliantAmounts: boolean;
+	// Some users will see a version of the checkout that only offers
+	// the option to make contributions. We won't want to include these
+	// users in some AB tests
+	excludeCountriesSubjectToContributionsOnlyAmounts: boolean;
```
paul-daniel-dempsey, line 84: So this is generally `true` unless testing  the contibutions-only checkout

paul-daniel-dempsey (approved):
Looks good subject to PaulB's name change suggestion

support-frontend/assets/helpers/abTests/abtest.ts
GHaberis: Yes that's correct, I've made it a non-optional value so whoever's setting up a test will need to consider this

support-frontend/assets/helpers/contributions.ts
jamesgorrie: I like including `countries` in the name as this makes it clear that that's what we're basing this on. 

I also like the removal of the `amounts` bit as that's  implementation details.

My 2¢

jamesgorrie (commented):
There's the chat about the name of the method that feels like it should be resolved before committing as this is a renaming PR.

support-frontend/assets/helpers/contributions.ts
GHaberis: Suggestions taken on board, I've gone with `isContributionsOnlyCountry`

jamesgorrie (approved):

PR #5934, opened 2024-04-18
Fix discounts not showing in email for annual products

rupertbates:
Fixes a bug whereby discounted annual subscriptions were not told the correct billing information in their thank you email. 

The billing summary in the thank you email is worked out by retrieving a preview of future invoices from Zuora and then using this to generate a description such as '£5.99 every month for 3 months, then £11.99 every month'.

The bug was caused by us only fetching one preview invoice (the one with the discounted price) so the code which works out the message thought the initial price was the regular price.

[**Trello Card**](https://trello.com/c/rj9SSTDl/721-annual-s-ty-email-price-does-not-contain-discount-info)

### DIFF
--- a/support-workers/src/main/scala/com/gu/support/workers/PreviewPaymentSchedule.scala
+++ b/support-workers/src/main/scala/com/gu/support/workers/PreviewPaymentSchedule.scala
@@ -12,13 +12,12 @@ object PreviewPaymentSchedule {
 
   def preview(
       subscribeItem: SubscribeItem,
-      billingPeriod: BillingPeriod,
       zuoraService: ZuoraSubscribeService,
       singleResponseCheck: Future[List[PreviewSubscribeResponse]] => Future[PreviewSubscribeResponse],
   ): Future[PaymentSchedule] = {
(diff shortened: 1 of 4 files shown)


support-workers/src/main/scala/com/gu/support/workers/PreviewPaymentSchedule.scala
```diff
@@ -12,22 +12,16 @@ object PreviewPaymentSchedule {
 
   def preview(
       subscribeItem: SubscribeItem,
-      billingPeriod: BillingPeriod,
       zuoraService: ZuoraSubscribeService,
       singleResponseCheck: Future[List[PreviewSubscribeResponse]] => Future[PreviewSubscribeResponse],
   ): Future[PaymentSchedule] = {
-    val numberOfInvoicesToPreview: Int = billingPeriod match {
```
rupertbates, line 19: This was wrong, the number is not the number of invoices, it's the number of months to invoice within

GHaberis (approved):
Nice!

PR #5938, opened 2024-04-18
stop nuisance alarms for no acquisitions

johnduffell:
we get nuisance alarms, this should help
when I split S+ and RC I worked out a suitable interval for each, but in the end it was a bit tight (so false alarms often)

image


tomrf1 (approved):

support-workers/cloud-formation/src/templates/cfn-template.yaml
```diff
@@ -288,8 +287,8 @@ Resources:
           Value: !Ref Stage
       ComparisonOperator: LessThanOrEqualToThreshold
       Threshold: 0
-      Period: 300
-      EvaluationPeriods: 36
+      Period: 3600
```
johnduffell, line 290: I changed them all to 3600s interval (1 hour) so you can tell how long they are at a glance

PR #5939, opened 2024-04-19
Add retrying for Zuora `TEMPORARY_ERROR`

jamesgorrie:
Adds retry logic for the [Zuora `TEMPORARY_ERROR`](https://knowledgecenter.zuora.com/Zuora_Central_Platform/API/G_SOAP_API/L_Error_Handling/Errors#ErrorCode_Object).

We have seen this error raised a couple times.

### DIFF
--- a/support-workers/src/test/scala/com/gu/support/workers/errors/ErrorHandlerSpec.scala
+++ b/support-workers/src/test/scala/com/gu/support/workers/errors/ErrorHandlerSpec.scala
@@ -95,8 +95,13 @@ class ErrorHandlerSpec extends AnyFlatSpec with Matchers {
     ]
     ZuoraErrorResponse(
       false,
+      List(ZuoraError("TEMPORARY_ERROR", "Operation failed due to a temporary error, please retry later.")),
+    ).asRetryException shouldBe a[
+      RetryUnlimited,
+    ]
+    ZuoraErrorResponse(
+      false,
       List(ZuoraError("TRANSACTION_FAILED", "Your card was declined")),
     ).asRetryException shouldBe a[RetryNone]
-
   }
 }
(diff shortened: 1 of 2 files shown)


support-workers/src/test/scala/com/gu/support/workers/errors/ErrorHandlerSpec.scala
```diff
@@ -93,10 +93,15 @@ class ErrorHandlerSpec extends AnyFlatSpec with Matchers {
     ).asRetryException shouldBe a[
       RetryUnlimited,
     ]
+    ZuoraErrorResponse(
+      false,
+      List(ZuoraError("TEMPORARY_ERROR", "Operation failed due to a temporary error, please retry later.")),
+    ).asRetryException shouldBe a[
+      RetryUnlimited,
```
paul-daniel-dempsey, line 100: Went for unlimited rather than 5-times? Guessing its unlikely the a server would be back up 5 retries

paul-daniel-dempsey (approved):
This looks far simpler than we looked into previously! Like this singular `ErrorHandlerSpec` though

support-workers/src/test/scala/com/gu/support/workers/errors/ErrorHandlerSpec.scala
jamesgorrie: My logic is that it is temporary - so should come back up eventually.

I'm not to sure what the usecase for `RetryLimited` is, potentially if a service is overwhelmed and we want to backoff to allow it to stabilise. 

[We also use it for unknown errors](https://github.com/guardian/support-frontend/blob/main/support-workers/src/main/scala/com/gu/support/workers/exceptions/ErrorHandler.scala#L66-L67) - you can see the other usecases in this file.

johnduffell: @jamesgorrie  unlimited is 24h, limited is about 2 hours I think (exponential backoff)
The main issue with unlimited is that the state machine times out completly after 24h and the customer doesn't get an email afaik so they don't realise it's failed.  Also we don't get a "failed" alarm until 24h later.  The limited one they get an email telling them it's failed, so they can take action e.g. retry.

jamesgorrie: Thanks @johnduffell - I think in this instance then we probably want to `RetryLimited` - if Zuora's `TEMPORARY_ERROR` is more than 2 hours, I imagine it's worth knowing about then.

The example that inspired this PR it seems like it was out for < 5 minutes. So that fits the bill too.

[Here's the PR](https://github.com/guardian/support-frontend/pull/5947)

PR #5947, opened 2024-04-22
fix: RetryLimited on Zuora TEMPORARY_ERROR

jamesgorrie:
Makes the Zuora `TEMPORARY_ERROR` `RetryLimited`.

Based on [this conversation](https://github.com/guardian/support-frontend/pull/5939#discussion_r1574374387) it felt like knowing if a `TEMPORARY_ERROR` was lasting longer than 2 hours, we should know. The example case for this only lasted < 5 minutes, so it feels like designing to real life is a better plan.

### DIFF
--- a/support-models/src/main/scala/com/gu/support/zuora/api/response/Responses.scala
+++ b/support-models/src/main/scala/com/gu/support/zuora/api/response/Responses.scala
@@ -4,7 +4,7 @@ import com.gu.support.encoding.Codec._
 import com.gu.support.encoding.CustomCodecs.{decodeLocalTime, encodeLocalTime}
 import com.gu.support.encoding.JsonHelpers._
 import com.gu.support.encoding.{Codec, ErrorJson}
-import com.gu.support.workers.exceptions.{RetryException, RetryNone, RetryUnlimited}
+import com.gu.support.workers.exceptions.{RetryException, RetryLimited, RetryNone, RetryUnlimited}
 import com.gu.support.zuora.api.PaymentGateway
 import io.circe.Decoder.Result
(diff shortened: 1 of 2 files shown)


johnduffell (approved):
nice, I think that's sensible. 👍 
as an added note, sometimes Zuora or Salesforce have scheduled downtime overnight or on the weekend, and we always (perhaps blithely) say that the step functions will retry anyway.

jamesgorrie:
> and we always (perhaps blithely) say that the step functions will retry anyway

@johnduffell [It looks like we `RetryUnlimited` for all Zuora errors](https://github.com/guardian/support-frontend/blob/a816124d7fb69ea9ea2345ceb039bc93f3428c98/support-models/src/main/scala/com/gu/support/zuora/api/response/Responses.scala#L50).

Maybe that's not the right thing, although is hasn't raised any problems as yet. It could be worth a think about to be preventative rather than reactive though.

johnduffell:
> It looks like we `RetryUnlimited` for all Zuora errors

that's not true, only the ones specifically determined, the rest are RetryNone
https://github.com/guardian/support-frontend/blob/984dc6ea209c35fc4777f8679d9072c8e1b525fc/support-models/src/main/scala/com/gu/support/zuora/api/response/Responses.scala#L60

I suspect (GH history lets me down) the specifically determined ones were because zuora used to have outages all the time (i.e. every month) so we had a good idea of what it returned in those cases.

> Maybe that's not the right thing, although is hasn't raised any problems as yet. It could be worth a think about to be preventative rather than reactive though.

could be, things have changed a lot in the last 5+ years

PR #5951, opened 2024-04-22
Don't use `productId = 'Contribution'` to infer `ONE_OFF`

jamesgorrie:
I had mistakenly inferred the `productId = 'Contribution'` meant `ONE_OFF` payments. 

I've started dropping `TODO`s where we are going to have to think about supporting `ONE_OFF` as it's not actually in the catalog.

This means we now
* use the right stripe key
* show the state selector in relevant countries
* ask for personal details

## 👀 

Screenshot 2024-04-22 at 17 04 39


paul-daniel-dempsey (approved):
Yep contribution can be regular, just below S+ threshold value.

PR #5952, opened 2024-04-22
fix: add billing-address country to non-deliverable products

jamesgorrie:
- Fixes the field name from where we fetch billing/delivery address
- Sets the billing-address for non-deliverable products

It's interesting that we set the `billingAddresss.country` from the [Fastly geo client](https://www.fastly.com/documentation/reference/vcl/variables/geolocation/) and don't allow a person to reset it.

This would mean the Billing Country could be wrong for people using VPNs etc - so maybe something to look into.


GHaberis (approved):
Looks good 🚀

PR #5959, opened 2024-04-25
Run integration tests every 6 hours

tomrf1:
Currently these run hourly.
There's a limit on salesforce records in the sandbox, and zuora will also be enforcing limits soon.
Reducing integration test runs will help reduce our usage


graham228221 (approved):
Happy to approve!

PR #5962, opened 2024-04-26
Use react router for landing => checkout => thank-you route

jamesgorrie:
- Uses `react-router` to encapsulate the `landing => checkout => thank-you` journey
- Adds the concept `GeoIdConfig` which is config that is variable on the `/{geoId}` segment of the URL
- Changes the `checkout` name for the route to `router` in the play app
- Copy/pastes the values dependant on the `geoId` in the checkout component into the component

I'm doing this because when setting up a whole new route for the thank you page from Play => rendering, the boilerplate was becoming immense and unwieldy. This replicates the pattern we had setup for `/{geoId}/contribution` from [Play](https://github.com/guardian/support-frontend/blob/main/support-frontend/conf/routes#L57-L69) => [Router](https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/pages/supporter-plus-landing/supporterPlusRouter.tsx).

This also avoids having to spin up another `webpack.entryPoint`. A 🚤 performance-point 🚀  here 
 - this does mean that the entire journey is built into 1 file, which could also start to get too large. This is how we currently do it, but it made me think that we should be tracking this against the current checkout flow to ensure we're making improvements. I'm tracking that in [the Epic](https://github.com/guardian/support-frontend/issues/5722).

Part of #5722

```[tasklist]
### Tasks
- [x] Add ADR if we like this
```

### DIFF
--- a/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
@@ -25,7 +25,6 @@ import {
 	useStripe,
 } from '@stripe/react-stripe-js';
 import { useEffect, useRef, useState } from 'react';
-import { parse, picklist } from 'valibot'; // 1.54 kB
 import { Box, BoxContents } from 'components/checkoutBox/checkoutBox';
 import { CheckoutHeading } from 'components/checkoutHeading/checkoutHeading';
 import DirectDebitForm from 'components/directDebit/directDebitForm/directDebitForm';
@@ -66,19 +65,17 @@ import { getStripeKey } from 'helpers/forms/stripe';
 import { validateWindowGuardian } from 'helpers/globalsAndSwitches/window';
 import CountryHelper from 'helpers/internationalisation/classes/country';
 import type { IsoCountry } from 'helpers/internationalisation/country';
-import type { CountryGroupId } from 'helpers/internationalisation/countryGroup';
-import type { Currency } from 'helpers/internationalisation/currency';
-import { currencies } from 'helpers/internationalisation/currency';
 import { productCatalogDescription } from 'helpers/productCatalog';
 import { NoFulfilmentOptions } from 'helpers/productPrice/fulfilmentOptions';
 import { NoProductOptions } from 'helpers/productPrice/productOptions';
-import { renderPage } from 'helpers/rendering/render';
 import { get } from 'helpers/storage/cookie';
 import {
 	getOphanIds,
 	getReferrerAcquisitionData,
 } from 'helpers/tracking/acquisitions';
 import { trackComponentClick } from 'helpers/tracking/behaviour';
+import type { GeoId } from 'pages/geoIdConfig';
+import { getGeoIdConfig } from 'pages/geoIdConfig';
 import { CheckoutDivider } from 'pages/supporter-plus-landing/components/checkoutDivider';
 import { GuardianTsAndCs } from 'pages/supporter-plus-landing/components/guardianTsAndCs';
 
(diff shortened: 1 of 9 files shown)


support-frontend/assets/pages/[countryGroupId]/router.tsx
```diff
@@ -0,0 +1,25 @@
+import { createBrowserRouter, RouterProvider } from 'react-router-dom';
+import { renderPage } from 'helpers/rendering/render';
+import { geoIds } from 'pages/geoIdConfig';
+import { Checkout } from './checkout';
+
+const router = createBrowserRouter(
+	geoIds
+		.map((geoId) => [
+			{
```
Comment (jamesgorrie) [inline support-frontend/assets/pages/[countryGroupId]/router.tsx:9]: We'll add the `/${geoId}` soon as part of the migration from `/{geoId}/contribute` work as this URL is currently redirected via Fastly and the Play app.

support-frontend/webpack.entryPoints.js
```diff
@@ -1,26 +1,36 @@
...
-    unsupportedBrowserStyles: 'stylesheets/fallback-pages/unsupportedBrowser.scss',
-    contributionsRedirectStyles: 'stylesheets/fallback-pages/contributionsRedirect.scss',
-    promotionTerms: 'pages/promotion-terms/promotionTerms.tsx',
-    ausMomentMap: 'pages/aus-moment-map/ausMomentMap.tsx',
-  },
-  ssr: {
-    ssrPages: 'helpers/rendering/ssrPages.ts',
-  },
+	common: {
+		'[countryGroupId]/router': 'pages/[countryGroupId]/router.tsx',
```
jamesgorrie, line 3: This is change - the rest is formatting.

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -346,13 +146,104 @@ const processPayment = (statusResponse: StatusResponse) => {
...
+	price:
+		searchParamsPrice && isNumeric(searchParamsPrice)
+			? parseFloat(searchParamsPrice)
+			: undefined,
+};
+
+type Props = {
+	geoId: GeoId;
+};
+function CheckoutComponent({ geoId }: Props) {
```
Comment (jamesgorrie) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:173]: The ordering of this file has changed - that's provably the most important piece to review.

jamesgorrie:
@GHaberis - re: filesize - currently `SupporterPlusRouter` (AKA `supporterPlusLandingPage` is `284 kB` and `router` now is `+158 kB`.

It's not a fair comparison though as `supporterPlusLandingPage` includes the landing page.

Interestingly - adding `react-router` adds `15kb` - which is significant, I was looking at a more slim router like https://github.com/molefrog/wouter, or a more fully featured, but smaller https://tanstack.com/router

support-frontend/assets/pages/[countryGroupId]/router.tsx
```diff
@@ -0,0 +1,25 @@
+import { createBrowserRouter, RouterProvider } from 'react-router-dom';
+import { renderPage } from 'helpers/rendering/render';
+import { geoIds } from 'pages/geoIdConfig';
+import { Checkout } from './checkout';
+
+const router = createBrowserRouter(
+	geoIds
+		.map((geoId) => [
```
Comment (rBangay) [inline support-frontend/assets/pages/[countryGroupId]/router.tsx:8]: could you use a `flatMap` here instead of a `map` and then a `flat` later on?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap

rBangay (approved):
👍 

Nice work!

support-frontend/assets/pages/[countryGroupId]/router.tsx
Comment (jamesgorrie) [reply support-frontend/assets/pages/[countryGroupId]/router.tsx:8]: Indeed I can, and have!

PR #5968, opened 2024-04-30
feat: make support-workers work with the new third tier rate plans

AndreaDiotallevi:

[Trello Card](https://trello.com/c/6NNlzzq0/753-make-support-frontend-and-support-workers-use-new-gw-digital-rate-plan)

This PR makes `support-workers` work with the new third tier rate plans (CODE & PROD).
It also updates the PROD catalog and the tests that depend on it.

[Zuora subscription example](https://apisandbox.zuora.com/platform/subscriptions/8ad09c9f8f2edb6e018f2f558ae116c3):

Screenshot 2024-04-30 at 16 16 27

## Why are you doing this?

At the moment third tier product purchases are handled by creating two subscriptions, one for Supporter Plus and one for Guardian Weekly.

This PR is the first part of the refactor to be able to just create one subscription containing the new rate plans.

## How to test

From the root of the repository, run:
- `sbt`
- `project support-workers`
- `test`

The `support-workers` service understands the new product rate plans and is able to create Zuora subscriptions.


support-frontend/.vscode/settings.json
jamesgorrie, line 1: If we want to add settings, we should add them to settings recommended. [This has been added to the root `gitignore`](https://github.com/guardian/support-frontend/blob/main/.gitignore#L11) - assuming this was missed as you have the editor open at this sub-directory?

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/SupporterPlusSubcriptionBuilder.scala
```diff
@@ -27,6 +28,55 @@ class SupporterPlusSubcriptionBuilder(
...
+      )
+    }
+
+    val todaysDate = dateGenerator.today
+
+    val subscriptionData = subscribeItemBuilder.buildProductSubscription(
+      productRatePlanId = productRatePlanId,
+      contractEffectiveDate = todaysDate,
+      contractAcceptanceDate = todaysDate,
+      readerType = Direct,
```
jamesgorrie, line 58: Just because I still don't really know what `readerType` is - is this always `Direct`?

jamesgorrie (approved):
To clarify my understanding - if we have a `SupporterPlus` subscription, and there are `fulfilmentOptions` - we assumed `SupporterPlusWithGuardianWeekly`?

Maybe we could outline that in the test?

support-workers/src/main/scala/com/gu/zuora/subscriptionBuilders/SupporterPlusSubcriptionBuilder.scala
rupertbates: Yes it will always be `Direct` the only other value of this that we care about is `Gift` which is used for gift subs and we don't have those for this product

AndreaDiotallevi:
> To clarify my understanding - if we have a `SupporterPlus` subscription, and there are `fulfilmentOptions` - we assumed `SupporterPlusWithGuardianWeekly`?
> 
> Maybe we could outline that in the test?

@jamesgorrie Thanks for the comment. Can you elaborate a bit more what you mean exactly?

rupertbates (approved):
Looks good

PR #5973, opened 2024-05-01
feat: thank you page for generic checkout

jamesgorrie:
Adds the thank you page to the generic checkout.

This uses a explicit object in session storage:
```JSON
{
  "value": {
    "firstName": "???",
    "price": 5,
    "product": "Contribution",
    "ratePlan": "Monthly",
    "paymentMethod": "Stripe"
  }
}
```

This will last **only** for the session, and will be deleted if you visit the checkout again (the only place that sets it).

## Screenshots
Screenshot 2024-05-01 at 15 12 48

### DIFF
--- a/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
@@ -46,6 +46,7 @@ import type { PostcodeFinderResult } from 'components/subscriptionCheckouts/addr
 import { findAddressesForPostcode } from 'components/subscriptionCheckouts/address/postcodeLookup';
 import { getAmountsTestVariant } from 'helpers/abTests/abtest';
 import { isContributionsOnlyCountry } from 'helpers/contributions';
+import type { ErrorReason } from 'helpers/forms/errorReasons';
 import { loadPayPalRecurring } from 'helpers/forms/paymentIntegrations/payPalRecurringCheckout';
 import type {
 	RegularPaymentRequest,
@@ -78,6 +79,7 @@ import type { GeoId } from 'pages/geoIdConfig';
 import { getGeoIdConfig } from 'pages/geoIdConfig';
 import { CheckoutDivider } from 'pages/supporter-plus-landing/components/checkoutDivider';
 import { GuardianTsAndCs } from 'pages/supporter-plus-landing/components/guardianTsAndCs';
+import { setThankYouOrder } from './thank-you';
 
 /** App config - this is config that should persist throughout the app */
 validateWindowGuardian(window.guardian);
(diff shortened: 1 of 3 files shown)


support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -130,26 +132,36 @@ const legend = css`
...
-				.then((json) => {
-					void processPayment(json as StatusResponse, geoId);
-				});
-		}, 1000);
-	}
+/**
+ * This method removes the `pending` state by retrying,
+ * resolving on success or failure only.
+ */
+const processPayment = async (
```
Comment (jamesgorrie) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:139]: This was refactored to just return `success | failure`, removing the `pending` state.

It also leaves the side effects to the calling method.

GHaberis:
@jamesgorrie looks good, although I think we'll need to notify Data Privacy of any changes to the items we store in Session Storage, such as the introduction of the `thankYouOrder` object, there's a form we have to submit these changes to, I'll send you the link.

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -533,7 +549,27 @@ function CheckoutComponent({ geoId }: Props) {
 				.then((response) => response.json())
 				.then((json) => json as StatusResponse);
 
-			processPayment(createSubscriptionResult, geoId);
+			const processPaymentResponse = await processPayment(
+				createSubscriptionResult,
+				geoId,
+			);
+			if (processPaymentResponse.status === 'success') {
+				const order = {
```
Comment (GHaberis) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:557]: I think this idea of having a single object in Session storage encapsulating all useful info is much better than having a key/value for each property - it'll be easier to maintain and modify.

GHaberis (approved):
Looks good

jamesgorrie:
> I think we'll need to notify Data Privacy

Thanks! I will merge for now as this isn't live.

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -533,7 +549,27 @@ function CheckoutComponent({ geoId }: Props) {
...
+					firstName: personalData.firstName,
+					price: price,
+					product: productId,
+					ratePlan: query.ratePlan,
+					paymentMethod: paymentMethod as string,
+				};
+				setThankYouOrder(order);
```
Comment (paul-daniel-dempsey) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:564]: Can see the session storage call with thankyou order object here and then going through the router to the thankyou page below. Nice!

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -130,26 +132,36 @@ const legend = css`
...
+	geoId: GeoId,
+): Promise<
+	{ status: 'success' } | { status: 'failure'; failureReason?: ErrorReason }
+> => {
+	return new Promise((resolve) => {
+		const { trackingUri, status, failureReason } = statusResponse;
+		if (status === 'success') {
+			resolve({ status: 'success' });
+		} else if (status === 'failure') {
+			resolve({ status, failureReason });
```
Comment (paul-daniel-dempsey) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:150]: much neater than the console log, nice!

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:139]: Could not see a `pending` if state in the original tbh, however, I like the way it just returns a failure status object with reason  now.

support-frontend/assets/pages/[countryGroupId]/thank-you.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+type Props = {
+	geoId: GeoId;
+};
+export function ThankYou({ geoId }: Props) {
+	const countryId = CountryHelper.fromString(get('GU_country') ?? 'GB') ?? 'GB';
+	const isSignedIn = !!get('GU_U');
+	const { countryGroupId, currencyKey } = getGeoIdConfig(geoId);
+
+	const sessionStorageOrder = storage.session.get('thankYouOrder');
```
Comment (paul-daniel-dempsey) [inline support-frontend/assets/pages/[countryGroupId]/thank-you.tsx:58]: Do you need this const? only used once

paul-daniel-dempsey (commented):
Looks really good

PR #5979, opened 2024-05-07
fix: refine cmp e2e test to buttons and simpler text

jamesgorrie:
Refines the search for the "Yes/No" buttons to
* Be a button
* Have only "Yes" or "No" in

This is slightly reactive to the work going on in CMP - so if it happens again, we might have to re-think - but we should have a test that CMP is working so maybe the `isVisible` test is enough.

We've updated the chromatic action to use `v11.3.0` as [`v11.3.1`](https://github.com/chromaui/action/commit/d208304630e76d960a80b9db40a838eb8babcebe) seemed to have introduced a bug.

If we update to `v11.3.1` - we'll need to add the `storybookBaseDir: "support-frontend"` to the Chromatic action.

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -15,7 +15,7 @@ jobs:
       - name: Install Node
         uses: actions/setup-node@v4
         with:
-          node-version-file: '.nvmrc'
+          node-version-file: ".nvmrc"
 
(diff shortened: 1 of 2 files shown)


paul-daniel-dempsey (approved):
Looks good

paul-daniel-dempsey:
Getting `Failed to retrieve dependent story files` build failure from Chromatic?

PR #5980, opened 2024-05-07
Ticket tailor embed widget

jamesgorrie:
Adds a prototype ticket tailor widget to the site.

- Pretty standard implementation with little customisation
- Lives at `https://support.theguardian.com/{geo}/events` with a default ID set to my test event, but you can change that with the `?eventId={id}` QueryParam. we might want to revisit this to be `/events/{eventId}` but that was a little more work.
- Sits behind a Google Auth service so the only accessible with a Guardian Google Account
- The completion page currently keeps maintains the same Guardian page but with TT's "Thank you" messaging. We can setup a redirect URL which will allow us to add tracking via the [config of the event](https://help.tickettailor.com/en/articles/2096942-how-to-redirect-ticket-buyers-back-to-your-website-after-buying-tickets) including the event ID and other information. This is slightly laborious as it's per event rather than per box office.

### Tracking
- The Ophan tracking event is `https://ophan.theguardian.com/img/1?v=17&platform=support&url=https%3A%2F%2Fsupport.thegulocal.com%2Fuk%2Fevents%3FeventId%3D12&ref=&visibilityState=visible&tz=-60&viewId=xxxxxxxxx`
- There is no Guardian tracking on the journey through TicketTailor. As per the completion page info ☝️ if we add a completion page ourselves, we would be able to track completions in Ophan. Saying that TT has quite a lot of it's own metrics in their reporting.

### Signed in users
- We can definitely have signed in information on the support.theguardian.com pages, although I can't find a way to pass that information to TT. This information would be tracked in Ophan. 
- We could technically put this behind a sign-in gate with a few changes to [the existing AuthActions](https://github.com/guardian/support-frontend/blob/main/support-frontend/app/actions/UserFromAuthCookiesActionBuilder.scala) 

https://github.com/guardian/support-frontend/assets/31692/4b8020b9-3a33-43c7-82ef-38147e108186

```[tasklist]
### Things to think about
- [x] To avoid a popout experience, we need to setup a CNAME. Could this use `theguardian.live` e.g. `events.theguardian.live`?
- [ ] UX - what Chrome do we want?
- [x] With the data, is an Ophan pageview enough?
- [ ] Where should we send someone after they have bought a ticket?
``` 

```[tasklist]
### Depends on
- [x] https://github.com/guardian/support-frontend/pull/6012
```


support-frontend/assets/pages/[countryGroupId]/events/events.tsx
```diff
@@ -0,0 +1,55 @@
...
+						<script
+							src="https://cdn.tickettailor.com/js/widgets/min/widget.js"
+							data-url={`https://tickets.theguardian.live/checkout/new-session/id/${eventId}/chk/9fa2/?ref=support-theguardian-com`}
+							data-type="inline"
+							data-inline-minimal="true"
+							data-inline-show-logo="false"
+							data-inline-bg-fill="false"
+							data-inline-inherit-ref-from-url-param=""
+							data-inline-ref="support-theguardian-com"
+						></script>
```
Comment (jamesgorrie) [inline support-frontend/assets/pages/[countryGroupId]/events/events.tsx:49]: This is the standard code they recommend.

support-frontend/assets/pages/[countryGroupId]/events/router.tsx
Comment (jamesgorrie) [inline support-frontend/assets/pages/[countryGroupId]/events/router.tsx:1]: I've separated this into it's own Controller method, router and component to avoid bloating the existing checkout bundle. The boilerplate felt minimal enough.

support-frontend/assets/pages/[countryGroupId]/events/events.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { PageScaffold } from 'components/page/pageScaffold';
+import ThankYouFooter from 'pages/supporter-plus-thank-you/components/thankYouFooter';
+
+const checkoutContainer = css`
+	${from.tablet} {
+		background-color: ${sport[800]};
+	}
+`;
+
+ophan.init();
```
Comment (jamesgorrie) [inline support-frontend/assets/pages/[countryGroupId]/events/events.tsx:16]: @paul-daniel-dempsey this is how I'm doing Ophan on events (related to checkout tracking).

support-frontend/assets/pages/[countryGroupId]/events/router.tsx
```diff
@@ -0,0 +1,21 @@
+import ophan from 'ophan';
+import { createBrowserRouter, RouterProvider } from 'react-router-dom';
+import { renderPage } from 'helpers/rendering/render';
+import { geoIds } from 'pages/geoIdConfig';
+import { Events } from './events';
+
+ophan.init();
```
Comment (jamesgorrie) [inline support-frontend/assets/pages/[countryGroupId]/events/router.tsx:7]: @paul-daniel-dempsey I've moved this to the router, which felt more appropriate given this is where the entry point of the page actually is.

support-frontend/assets/pages/[countryGroupId]/events/router.tsx
Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/events/router.tsx:1]: Yep, can see this file referenced from the `webpack.entryPoints.js`

support-frontend/assets/pages/[countryGroupId]/events/router.tsx
Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/events/router.tsx:16]: Can see this initialisation is part of the call `setUpTrackingAndConsent()` which is placed on most pages including the `supporterPlusRouter.tsx`

support-frontend/app/controllers/Application.scala
```diff
@@ -310,6 +312,11 @@ class Application(
       ),
     ).withSettingsSurrogateKey
   }
+
+  def eventsRouter(countryGroupId: String, eventId: Option[String]) = authAction { implicit request =>
```
paul-daniel-dempsey, line 316: Not sure I understand difference between `MaybeAuthenticatedAction` used elsewhere and  `authAction` - presumably both allow us to share settings?

paul-daniel-dempsey (approved):
Have got this running no problem, small question on on `Application.scala` thats all I could find with my limited Scala knowledge

support-frontend/app/controllers/Application.scala
jamesgorrie: `MaybeAuthenticatedAction` authenticates against Okta, so used for readers and people with Guardian Identity accounts.

This `authAction` uses Google auth, so only people with Guardian Google accounts can access this.

Great to point out that it's not massively clear. We need a little work on the identity stuff.

support-frontend/assets/pages/[countryGroupId]/events/events.tsx
Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/events/events.tsx]: Thanks have put this into `assets/pages/[countryGroupId]/router.tsx` in my Ophan PR here -> (https://github.com/guardian/support-frontend/pull/6014)

PR #5982, opened 2024-05-07
Personal details form validation

jamesgorrie:
Part of #5722 

* Removes the use of `PersonalDetails` component as the error API of that component was becoming difficult to integrate with. The abstraction isn't used anywhere that the checkout isn't going to make obsolete, so it's a little bit of code now, for less later.
* Adds validation using [browser APIs](https://developer.mozilla.org/en-US/docs/Learn/Forms/Form_validation). [The `ValidityState` is good to use in browsers we support](https://developer.mozilla.org/en-US/docs/Web/API/ValidityState#browser_compatibility). I might try push this upstream to Source in a later PR once we've ironed some of this out.

It is probably easier to view this PR via [this commit](https://github.com/guardian/support-frontend/commit/fc8a1fc934bac0e44192b68124edfffa4832673c) and then [this commit](https://github.com/guardian/support-frontend/commit/fd1979a833f8f7ae34bdc43134598208c0bf8d1b).

## Show me the screenshots 

https://github.com/guardian/support-frontend/assets/31692/479a9604-a062-49f5-82a4-96eb334c6151

### DIFF
--- a/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
@@ -688,6 +688,7 @@ function CheckoutComponent({ geoId }: Props) {
 												disabled={isSignedIn}
 												name="email"
 												required
+												maxLength={80}
 												error={emailError}
 												onInvalid={(event) => {
 													preventDefaultValidityMessage(event.currentTarget);
@@ -721,6 +722,7 @@ function CheckoutComponent({ geoId }: Props) {
 													onChange={(event) => setFirstName(event.target.value)}
 													name="firstName"
 													required
+													maxLength={40}
 													error={firstNameError}
 													onInvalid={(event) => {
 														preventDefaultValidityMessage(event.currentTarget);
@@ -752,6 +754,7 @@ function CheckoutComponent({ geoId }: Props) {
 													onChange={(event) => setLastName(event.target.value)}
 													name="lastName"
 													required
+													maxLength={40}
 													error={lastNameError}
 													onInvalid={(event) => {
 														preventDefaultValidityMessage(event.currentTarget);
--- a/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
@@ -197,6 +197,12 @@ const query = {
 			: undefined,
 };
 
+/** Form Validation */
+/**
+ * This uses a Unicode character class escape
+ * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Unicode_character_class_escape
+ */
+const doesNotContainEmojiPattern = '^[^\\p{Emoji_Presentation}]+$';
 function preventDefaultValidityMessage(currentTarget: HTMLInputElement) {
 	/**
 	 * Prevents default message showing, but maintains the default validation methods occuring
@@ -724,6 +730,7 @@ function CheckoutComponent({ geoId }: Props) {
 													required
 													maxLength={40}
 													error={firstNameError}
+													pattern={doesNotContainEmojiPattern}
 													onInvalid={(event) => {
 														preventDefaultValidityMessage(event.currentTarget);
 														const validityState = event.currentTarget.validity;
@@ -756,6 +763,7 @@ function CheckoutComponent({ geoId }: Props) {
 													required
 													maxLength={40}
 													error={lastNameError}
+													pattern={doesNotContainEmojiPattern}
 													onInvalid={(event) => {
 														preventDefaultValidityMessage(event.currentTarget);
 														const validityState = event.currentTarget.validity;
--- a/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
@@ -6,6 +6,7 @@ import {
 	space,
 	textSans,
 	until,
+	visuallyHidden,
 } from '@guardian/source-foundations';
 import {
 	Column,
@@ -34,7 +35,6 @@ import { ContributionsOrderSummary } from 'components/orderSummary/contributions
 import { PageScaffold } from 'components/page/pageScaffold';
 import { DefaultPaymentButton } from 'components/paymentButton/defaultPaymentButton';
 import { PayPalButton } from 'components/payPalPaymentButton/payPalButton';
-import { PersonalDetails } from 'components/personalDetails/personalDetails';
 import { StateSelect } from 'components/personalDetails/stateSelect';
 import { Recaptcha } from 'components/recaptcha/recaptcha';
 import { SecureTransactionIndicator } from 'components/secureTransactionIndicator/secureTransactionIndicator';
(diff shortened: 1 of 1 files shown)


support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -640,55 +674,132 @@ function CheckoutComponent({ geoId }: Props) {
...
-														//  no-op
+									<div css={personalDetailsFieldGroupStyles(true)}>
+										<h2 css={personalDetailsHeader}>Your details</h2>
+										<div>
+											<TextInput
+												id="email"
+												data-qm-masking="blocklist"
+												label="Email address"
+												value={email}
+												type="email"
```
Comment (paul-daniel-dempsey) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:685]: can see standard error type `email` used here, presume the `onInvalid` is fired if not matched and the `emailError` displays what is wrong. Nice!

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -133,6 +133,23 @@ const legend = css`
...
+	& > *:not(:first-of-type) {
+		margin-top: ${space[3]}px;
+	}
+	${from.tablet} {
+		& > *:not(:first-of-type) {
+			margin-top: ${space[4]}px;
+		}
+	}
+`;
+const personalDetailsHeader = css`
```
Comment (paul-daniel-dempsey) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:149]: Is this for screen readers?

Comment (jamesgorrie) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:149]: Yup - we should generally use [fieldsets](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset) to describe form structure, and fieldsets need headings - in this case one we don't want to show visually.

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
Comment (jamesgorrie) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:685]: @paul-daniel-dempsey [the onInvalid is a standard event](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/invalid_event) fired via the client.

We tap into this and pipe it through to [Source's implementation of errors](https://github.com/guardian/csnx/blob/main/libs/%40guardian/source-react-components/src/text-input/TextInput.tsx#L65-L68).

As mentioned in the description - it could be a good exercise to use standard DOM libs.

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -640,55 +674,132 @@ function CheckoutComponent({ geoId }: Props) {
...
-										errors={{}}
-										signOutLink={<Signout isSignedIn={isSignedIn} />}
-										contributionState={
-											showStateSelect && (
-												<StateSelect
-													countryId={countryId}
-													state={'STATE'}
-													onStateChange={() => {
-														//  no-op
+									<div css={personalDetailsFieldGroupStyles(true)}>
```
Comment (GHaberis) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:677]: @jamesgorrie regarding the stricture of the form, and your [comment](https://github.com/guardian/support-frontend/pull/5982/files#r1593733388) above regarding fieldsets, is there a reason we use a `<div>` to wrap the personal details and not a `<fieldset>` element?

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -640,55 +674,132 @@ function CheckoutComponent({ geoId }: Props) {
...
+												<TextInput
+													id="firstName"
+													data-qm-masking="blocklist"
+													label="First name"
+													value={firstName}
+													autoComplete="given-name"
+													autoCapitalize="words"
+													onChange={(event) => setFirstName(event.target.value)}
+													name="firstName"
+													required
```
Comment (GHaberis) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:723]: Nice to see the built in HTML attributes being used for validation. Have we checked the rules currently applied to the equivalent form fields on the original checkout are covered here? For example, I think there are max length limits on some of these fields (as implemented here https://github.com/guardian/support-frontend/pull/4168)

Comment (jamesgorrie) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:724]: I've added a [`maxLength` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/maxlength) to the input fields.

This doesn't error however, but rather stop a person adding more input. I'm not sure this a great UX as per [some a11y docs](https://a11ysupport.io/tech/html/maxlength_attribute#:~:text=The%20attribute%20prevents%20characters%20beyond,they%20are%20in%20fact%20not.)

> The attribute prevents characters beyond the maximum length from being entered. The user needs to know when the maximum length has been reached, so that they do not think that they are entering data when they are in fact not.

I'm not sure our UX is great either, where we allow you to go on, but we know we're going to block you anyway - wasting your time.

I'll see if I can get to "better" by responding to maxlength being reached.

Comment (jamesgorrie) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:724]: @GHaberis Why can't we have emoji?

Comment (GHaberis) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:724]: > @GHaberis Why can't we have emoji?

It'll be an issue for a downstream system such as Salesforce, Braze or Zuora. I'm afraid I can't recall which, maybe the PR history helps.

Comment (GHaberis) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:724]: Looks like it was Zuora https://github.com/guardian/support-frontend/pull/3371

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
Comment (GHaberis) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:677]: As we're pushing to use as many built-in features of the standard HTML form on this page a review of the HTML seems like a good idea - should we create a card to review the HTML as suggested?

Comment (jamesgorrie) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:677]: https://trello.com/c/KYYngPNf/840-make-generic-checkout-use-standard-dom-form-fields

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
Comment (jamesgorrie) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:731]: Interesting.

I've added this as a [standard `pattern` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern) for now - but we might have to see how it scales.

I think it should be alright as most of our [current form validations](https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/helpers/forms/formValidation.ts) are covered by the `maxLength`, `required` and a simple `pattern`.

GHaberis (approved):
Looks good!

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -197,6 +197,12 @@ const query = {
 			: undefined,
 };
 
+/** Form Validation */
+/**
+ * This uses a Unicode character class escape
+ * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Unicode_character_class_escape
```
Comment (paul-daniel-dempsey) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:203]: thanks for the reference link, v.useful

PR #5985, opened 2024-05-08
Generic checkout : Contribution/S+ e2e tests

paul-daniel-dempsey:

Setup Playwright tests for the Generic/Flexible Checkout
- Multiple Countries
- Contribution and SupporterPlus
- Prices
- Stripe/DirectDebit/PayPal

image

[**Trello Card**](https://trello.com/c/OuiyeUnb/811-e2e-tests-for-generic-checkout)

## How to test
{Assumes Playwright setup/installed}

Bash window1 -> start dev Support-Frontend
1. `cd support-frontend`
2. `./devrun.sh`

Bash window2 -> start Playwright
1. `cd support-e2e`
2. `yarn run test-local`


support-e2e/tests/genericCheckout.test.ts
```diff
@@ -0,0 +1,130 @@
...
+
+interface TestDetails {
+	product: 'Contribution' | 'SupporterPlus';
+	ratePlan: 'Monthly' | 'Annual';
+	paymentType?: 'Stripe' | 'DirectDebit' | 'PayPal';
+	price: number;
+	country?: 'UK' | 'US' | 'AU' | 'EU';
+}
+
+const glyphs: Record<'UK' | 'US' | 'AU' | 'EU', string> = {
```
jamesgorrie, line 19: Why `string`?

support-e2e/tests/genericCheckout.test.ts
```diff
@@ -0,0 +1,130 @@
...
+import { fillInPayPalDetails } from './utils/paypal';
+import { setupPage } from './utils/page';
+import { afterEachTasks } from './utils/afterEachTest';
+
+interface TestDetails {
+	product: 'Contribution' | 'SupporterPlus';
+	ratePlan: 'Monthly' | 'Annual';
+	paymentType?: 'Stripe' | 'DirectDebit' | 'PayPal';
+	price: number;
+	country?: 'UK' | 'US' | 'AU' | 'EU';
```
jamesgorrie, line 16: You could make this explicit and add `UK` explicitly. Defaults create 1 extra step someone has to jump through to understand.

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -390,6 +390,12 @@ function CheckoutComponent({ geoId }: Props) {
 	const elements = useElements();
 	const cardElement = elements?.getElement(CardNumberElement);
 	const [stripeClientSecret, setStripeClientSecret] = useState<string>();
+	/**
+	 * flag to ensure the (successful) recaptcha token has been fully passed to Stripe
+	 * used to disable/set 'Loading...' message on Pay Now button (Playwright test use)
```
Comment (jamesgorrie) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:395]:
```suggestion
	 * flag that disables the submission of the form until the stripeClientSecret is 
	 * fetched from the Stripe API with the reCaptcha token
```

support-e2e/tests/genericCheckout.test.ts
```diff
@@ -0,0 +1,130 @@
...
+import { email, firstName, lastName } from './utils/users';
+import { setTestUserDetails } from './utils/testUserDetails';
+import { checkRecaptcha } from './utils/recaptcha';
+import { fillInCardDetails } from './utils/cardDetails';
+import { fillInDirectDebitDetails } from './utils/directDebitDetails';
+import { fillInPayPalDetails } from './utils/paypal';
+import { setupPage } from './utils/page';
+import { afterEachTasks } from './utils/afterEachTest';
+
+interface TestDetails {
```
jamesgorrie, line 11: I would say this is a `type` rather than an `interface`. An `interface` is normally something you implement (extend) and is never really used itself. Similar to a Scala `trait`.

We could actually get rid of this altogether and just use a [`const` assertion](https://www.typescriptlang.org/docs/handbook/typescript-in-5-minutes-func.html#readonly-and-const) on the `testDetails`.

jamesgorrie (approved):
Just a couple notes on TypeScript, but nothing blocking.

Nice work!

support-e2e/tests/genericCheckout.test.ts
jamesgorrie: But reading through [the docs](https://www.typescriptlang.org/docs/handbook/typescript-in-5-minutes.html), `interface` does get used a lot.

support-e2e/tests/genericCheckout.test.ts
paul-daniel-dempsey: `string` looks like the recognised type used to display glyphs through support-frontend?

support-e2e/tests/genericCheckout.test.ts
paul-daniel-dempsey: Ok, this approach was taken in the `tieredCheckout.test.ts` but can see its unnecessary complication, `country` not optional now, done.

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx]: Ok, applied, thats a more technical description of the use of state `stripeClientSecretInProgress`

support-e2e/tests/genericCheckout.test.ts
paul-daniel-dempsey: Have read up on `const assertion` and can see it does away with writing the interface explicitly and forces the types to only be those specified in the array const definition. I have reworked to this approach. However, am not sure its particularly readable versus defining the interface (but its only used once in this file). Thanks for pointing out and getting me to understand this concept.

PR #5993, opened 2024-05-09
fix: detect countryId which also sets the GU_Country cookie

jamesgorrie:
[Uses `CountryHelper.detect()` reads and also sets the `GU_Country` cookie](https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/helpers/internationalisation/classes/country.ts#L245-L283).


paul-daniel-dempsey (approved):
👍

jamesgorrie:
:shipit:

PR #5972, opened 2024-05-01
Add abandoned basket cookie

david-pepper:

PR #5995, opened 2024-05-09
Abandoned basket update cookie when one off amount changes

andrewHEguardian:

Updating the cookie set on the second step of the checkout (#5972) with the updated amount when a user selects a different amount in the one-off contribution component.

[**Trello Card**](https://trello.com/c/5NzPmZK0/438-abandoned-basket-update-cookie-when-one-off-amount-changes)

## Why are you doing this?

To keep the cookie as up to date as possible with the last thing a user selected on the checkout.

## How to test

Go to `/contribute/checkout?selected-contribution-type=one_off` and see the `GU_CO_INCOMPLETE` cookie update the amount when choosing different amounts, including `other`

## Screenshots

image

### DIFF
--- a/support-frontend/package.json
+++ b/support-frontend/package.json
@@ -42,7 +42,7 @@
       ".+\\.(js|jsx|ts|tsx)$": "./node_modules/babel-jest"
     },
     "transformIgnorePatterns": [
-      "/node_modules/(?!@guardian)",
+      "/node_modules/(?!@guardian|valibot)",
       "^.+\\.module\\.(css|sass|scss)$"
     ],
     "moduleFileExtensions": [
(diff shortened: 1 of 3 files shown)


tomrf1 (approved):
nice!

support-frontend/package.json
```diff
@@ -42,7 +42,7 @@
       ".+\\.(js|jsx|ts|tsx)$": "./node_modules/babel-jest"
     },
     "transformIgnorePatterns": [
-      "/node_modules/(?!@guardian)",
+      "/node_modules/(?!@guardian|valibot)",
```
andrewHEguardian, line 45: https://stackoverflow.com/a/49676319
https://jestjs.io/docs/configuration#transformignorepatterns-arraystring
some sources that might help make sense of this, before adding we got 
`    Jest encountered an unexpected token

    This usually means that you are trying to import a file which Jest cannot parse, e.g. it's not plain JavaScript.

    By default, if Jest sees a Babel config, it will use that to transform your files, ignoring "node_modules".
`

PR #5998, opened 2024-05-10
Generic checkout:  zip-code & state validation

paul-daniel-dempsey:
Part of https://github.com/guardian/support-frontend/issues/5722

State & ZipCode validation for the Generic Checkout

|State|ZIP Code|
|-----|-----|
|image|image|

[**Trello Card**](https://trello.com/c/ihabrisb/802-form-validation-for-generic-checkout)

### DIFF
--- a/support-frontend/assets/components/personalDetails/stateSelect.tsx
+++ b/support-frontend/assets/components/personalDetails/stateSelect.tsx
@@ -49,7 +49,7 @@ export function StateSelect({
 				name={'billing-state'}
 			>
 				<>
-					<Option value="" selected>
+					<Option value="">
 						{`Select your ${stateDescriptor.toLowerCase()}`}
 					</Option>
 					{Object.entries(statesList).map(([abbreviation, name]) => {
--- a/support-frontend/assets/components/personalDetails/stateSelect.tsx
+++ b/support-frontend/assets/components/personalDetails/stateSelect.tsx
@@ -1,4 +1,6 @@
+import { css } from '@emotion/react';
 import { Option, Select } from '@guardian/source-react-components';
+import type { FormEventHandler } from 'react';
 import { CountryGroup } from 'helpers/internationalisation';
 import type { IsoCountry } from 'helpers/internationalisation/country';
 import {
@@ -12,6 +14,7 @@ type StateSelectProps = {
 	countryId: IsoCountry;
 	state: string;
 	onStateChange: (newState: string) => void;
+	onInvalid?: FormEventHandler<HTMLSelectElement>;
 	error?: string;
 };
 
@@ -31,6 +34,7 @@ export function StateSelect({
 	countryId,
 	state,
 	onStateChange,
+	onInvalid,
 	error,
 }: StateSelectProps): JSX.Element | null {
 	const countryGroupId = CountryGroup.fromCountry(countryId);
@@ -45,8 +49,23 @@ export function StateSelect({
 				label={stateDescriptor}
 				value={state}
 				onChange={(e) => onStateChange(e.target.value)}
+				onInvalid={onInvalid}
 				error={error}
 				name={'billing-state'}
+				required
+				cssOverrides={
+					/**
+					 * Source applies a red border initially unlike textInput's
+					 */
+					!error
+						? css`
+								:invalid {
+									border-color: #707070;
+									border-width: 1px;
+								}
+						  `
+						: undefined
+				}
 			>
 				<>
 					<Option value="">
--- a/support-frontend/assets/components/personalDetails/stateSelect.tsx
+++ b/support-frontend/assets/components/personalDetails/stateSelect.tsx
@@ -49,7 +49,7 @@ export function StateSelect({
 				name={'billing-state'}
 			>
 				<>
-					<Option value="">
+					<Option value="" selected>
 						{`Select your ${stateDescriptor.toLowerCase()}`}
 					</Option>
 					{Object.entries(statesList).map(([abbreviation, name]) => {
--- a/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
@@ -376,17 +376,7 @@ function CheckoutComponent({ geoId }: Props) {
 	const showStateSelect =
 		countryId === 'US' || countryId === 'CA' || countryId === 'AU';
 
-	/**
-	 * Payment method: PayPal
-	 * BAID = Billing Agreement ID
-	 */
 	const [paymentMethod, setPaymentMethod] = useState<PaymentMethod>();
-	const [payPalLoaded, setPayPalLoaded] = useState(false);
-	useEffect(() => {
-		if (paymentMethod === 'PayPal' && !payPalLoaded) {
-			void loadPayPalRecurring().then(() => setPayPalLoaded(true));
-		}
-	}, [paymentMethod, payPalLoaded]);
 
 	/** Payment methods: Stripe */
 	const stripePublicKey = getStripeKey(
@@ -400,7 +390,6 @@ function CheckoutComponent({ geoId }: Props) {
 	const elements = useElements();
 	const cardElement = elements?.getElement(CardNumberElement);
 	const [stripeClientSecret, setStripeClientSecret] = useState<string>();
-
 	/**
 	 * flag that disables the submission of the form until the stripeClientSecret is
 	 * fetched from the Stripe API with using the reCaptcha token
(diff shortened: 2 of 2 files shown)


support-frontend/assets/components/personalDetails/stateSelect.tsx
```diff
@@ -49,7 +49,7 @@ export function StateSelect({
 				name={'billing-state'}
 			>
 				<>
-					<Option value="">
+					<Option value="" selected>
```
jamesgorrie, line 52: Isn't the first option selected by default?

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -376,7 +376,17 @@ function CheckoutComponent({ geoId }: Props) {
...
+	 * Payment method: PayPal
+	 * BAID = Billing Agreement ID
+	 */
 	const [paymentMethod, setPaymentMethod] = useState<PaymentMethod>();
+	const [payPalLoaded, setPayPalLoaded] = useState(false);
+	useEffect(() => {
+		if (paymentMethod === 'PayPal' && !payPalLoaded) {
+			void loadPayPalRecurring().then(() => setPayPalLoaded(true));
+		}
+	}, [paymentMethod, payPalLoaded]);
```
Comment (jamesgorrie) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:389]: Is there a reason we've separated this from `payPalBAID`?

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -446,10 +447,27 @@ function CheckoutComponent({ geoId }: Props) {
...
-	const [billingState, setBillingState] = useState('');
+	const [billingStateError, setBillingStateError] = useState<string>();
+	/**
+	 * BillingState selector initialised to undefined to hide
+	 * billingStateError message. formOnSubmit checks and converts to
+	 * empty string to display billingStateError message.
+	 */
+	const [billingState, setBillingState] = useState<string | undefined>(
+		undefined,
+	);
```
Comment (jamesgorrie) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:462]:
```suggestion
	const [billingState, setBillingState] = useState<string>();
```

If you don't pass a default value, the type defaults to option AKA `string | undefined` in this case.

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -468,6 +486,9 @@ function CheckoutComponent({ geoId }: Props) {
 	const [isProcessingPayment, setIsProcessingPayment] = useState(false);
 	const formOnSubmit = async (formData: FormData) => {
 		setIsProcessingPayment(true);
+		if (!billingState) {
+			setBillingState('');
+		}
```
Comment (jamesgorrie) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:491]: I'm a little confused by this code as it's been accessed in a few places

From my reading - we have two states:
- `billingState` is not required
- `billingState` is required
  - `billingState` is invalid if `''`
  - otherwise `billingState` is valid 🥳 

I wonder if there is a clearer way to indicate this in the code?

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -803,23 +824,42 @@ function CheckoutComponent({ geoId }: Props) {
...
 													label="ZIP code"
-													value={billingPostcode}
-													error={undefined}
+													name="billing-postcode"
 													onChange={(event) =>
 														setBillingPostcode(event.target.value)
 													}
+													maxLength={20}
+													value={billingPostcode}
+													pattern={doesNotContainEmojiPattern}
```
Comment (jamesgorrie) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:844]: Should we be using this?
https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/helpers/forms/formValidation.ts#L37-L38

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -446,10 +447,27 @@ function CheckoutComponent({ geoId }: Props) {
...
+	const [billingState, setBillingState] = useState<string | undefined>(
+		undefined,
+	);
+	useEffect(() => {
+		if (billingState === '') {
+			setBillingStateError('Please enter a state, province or territory');
+		} else {
+			setBillingStateError('');
+		}
+	}, [billingState]);
```
Comment (jamesgorrie) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:469]: This makes sense - just validating - we're doing this as we don't want to write the validation into the component as it would the double validate in other areas that we use it?

jamesgorrie (commented):
A couple of questions - main one is around code clarity on `billingState`.

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:389]: Just seemed  neater to move useEffect's underneath the associated/related useStates (ie batched together).

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:462]: Yes am trying to start in an `undefined` state for the `billingState` dropdown. If undefined I donot display the error message (you don't want it to pop up when form loads) only after you have selected the `billigState` dropdown and did not selected a state.

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:469]: When form loads, `billingState = undefined` so don't show the error message. If you have submitted or pressed the drop down and not seleceted a state, then it will display error message.

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
Comment (jamesgorrie) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:462]: Super, I think we're in agreement, I was suggesting a simpler syntax for that.

e.g.
```ts
const [optionalA, setOptionalA] = useState<string | undefined>(undefined);
// optionalA is now : string | undefined

// Which can also be expressed as...
const [optionalB, setOptionalB] = useState<string>();
// Because we didn't pass the default argument, optionalB is also of type string | undefined
```

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
Comment (jamesgorrie) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:389]: Rather than neatness I tend to group things by functionality e.g. PayPal, Stripe etc. 

The reason being is that I find myself rarely working on all side-effects, but rather updating Stripe, or similar, so having the code co-located on that basis often avoids having to bounce around the code making it a little unwieldy.

Maybe we should look at starting to split this file up?

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:462]: Ah I see, will update.

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:389]: Let me revert this re-order firstly by functionality now.

Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx]: reverted

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:841]: I dont think this is used? Its a full house format I think (https://stackoverflow.com/questions/2577236/regex-for-zip-code) so `12345-1234` rather than `AL1234` for an area (I think). Dont think much fomatting other than 20chars is done on production?

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:488]: Its possible where the `billingState` starts at undefined and you directly press 'Pay Now' to run the `formSubmit`. We would want the error message to be displayed in this scenario.  The 'formOnSubmit` checks for an undefined and, if found, sets the `billingState` to '' to display the error message.

support-frontend/assets/components/personalDetails/stateSelect.tsx
paul-daniel-dempsey: Left this in whilst playing with it, removed.

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
Comment (jamesgorrie) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:841]: OK! Sounds good.

jamesgorrie (approved):

PR #6002, opened 2024-05-14
Add legends and secure checkout SVG to checkout

jamesgorrie:
Part of #5722 

- Uses and displays legends for Contribution checkout
- Adds Secure checkout SVG to payment details `fieldset`

I know that the `1.`, `2.`, `n.` won't work when we have delivery fields, I will reach out to Helene and see if there is a generic design solution to this.

[Worth avoiding whitedspace check in the diff](https://github.com/guardian/support-frontend/pull/6002/files?diff=unified&w=1).

| before | after |
|---|---|
| Screenshot 2024-05-14 at 15 44 26 | Screenshot 2024-05-14 at 15 44 11 |

### DIFF
--- a/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
@@ -129,6 +129,10 @@ const legend = css`
 	${from.tablet} {
 		font-size: 28px;
 	}
+
+	display: flex;
+	width: 100%;
+	justify-content: space-between;
 `;
 
 const fieldset = css`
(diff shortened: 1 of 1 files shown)


support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -695,8 +696,8 @@ function CheckoutComponent({ geoId }: Props) {
 						>
 							<Box cssOverrides={shorterBoxMargin}>
 								<BoxContents>
-									<div css={personalDetailsFieldGroupStyles(true)}>
-										<h2 css={personalDetailsHeader}>Your details</h2>
+									<fieldset css={fieldset}>
```
Comment (paul-daniel-dempsey) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:699]: As this css is related to a single fieldset and not all, would its be worth giving it a longer name like `fieldsetPersonalDetails`?

paul-daniel-dempsey (approved):
Nice use of `<fieldset>` to group related items in a box and `<legend>` to apply caption. Much improvement on `<div>`, will have to remember to use these more approriate tags

jamesgorrie:
> will have to remember to use these more approriate tags

@paul-daniel-dempsey it's because forms drive so much interactivity on the web. Always work [looking at MDN for reference](https://developer.mozilla.org/en-US/docs/Learn/Forms/How_to_structure_a_web_form)!

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
Comment (jamesgorrie) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:734]: I've added this to payment fields too - but I am not sure we should as the CSS is using first-of-type, so I might try a refactor post-merge (so we can start demoing this).

PR #6003, opened 2024-05-14
Add styles to payment methods on generic checkout

jamesgorrie:
[Reviewing without whitespace is easier](https://github.com/guardian/support-frontend/pull/6003/files?w=1).

Part of #5722 

Adds the styles to the payment methods for the generic checkout.

I've avoided using [Source's `Accordion`](https://github.com/guardian/csnx/blob/main/libs/%40guardian/source-react-components/src/accordion/Accordion.tsx) as it provides none of the functionality and we override nearly all of the styles. I think this is because they are quite different use cases as you can [see in the design system](https://theguardian.design/2a1e5182b/p/38c5aa-accordion).

If we wanted to use this I imagine we would build an `AccordianSansStyles`, and then add themes. Maybe worth using the new design system process we have setup for this?

## Show me please

https://github.com/guardian/support-frontend/assets/31692/70635d78-6f23-4cc2-8376-4d36d269d487

### DIFF
--- a/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
@@ -15,6 +15,7 @@ import {
 	Radio,
 	RadioGroup,
 	TextInput,
+	textInputThemeDefault,
 } from '@guardian/source-react-components';
 import {
 	FooterLinks,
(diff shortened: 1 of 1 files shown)


paul-daniel-dempsey (approved):
Thats a lot of styling work in the checkout ported from contribute, well done

PR #6008, opened 2024-05-15
Generic Checkout: Personal Detail Fields Validate via OnChange 

paul-daniel-dempsey:

Ensure Personal Details fields validate after editing when loose focus (onChange/onStateChange).

[**Trello Card**](https://trello.com/c/ihabrisb/802-form-validation-for-generic-checkout-state-zipcode)

image

### DIFF
--- a/support-frontend/assets/components/personalDetails/personalDetailsContainer.tsx
+++ b/support-frontend/assets/components/personalDetails/personalDetailsContainer.tsx
@@ -82,7 +82,9 @@ export function PersonalDetailsContainer({
 			<StateSelect
 				countryId={countryId}
 				state={state}
-				onStateChange={onBillingStateChange}
+				onStateChange={(event) => {
+					onBillingStateChange(event.currentTarget.value);
+				}}
 				error={errorObject?.state?.[0]}
 			/>
 		),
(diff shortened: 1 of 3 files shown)


jamesgorrie (commented):
Assuming this will be different to current behavior. My understanding is that currently

* from `valid => invalid`: Only show error `onSubmit`
* from `invalid => valid`: Hide errors `onChange` and `onSubmit`

This change will show/hide errors `onSubmit` and `onChange`.

So the diff is that we now show errors `onChange` e.g. you enter an invalid value `(😄)` and then focus on another event.

Are we happy with this?

paul-daniel-dempsey:
> Assuming this will be different to current behavior. My understanding is that currently
> 
> * from `valid => invalid`: Only show error `onSubmit`
> * from `invalid => valid`: Hide errors `onChange` and `onSubmit`
> 
> This change will show/hide errors `onSubmit` and `onChange`.
> 
> So the diff is that we now show errors `onChange` e.g. you enter an invalid value `(😄)` and then focus on another event.
> 
> Are we happy with this?

Hmmm, `onSubmit` and `onChange` was the behaviour we tested earlier today on the Contribution checkout  was'nt it? 
It doesnt seem to work like that now...have changes to personalDetails today affected it's behaviour? 
Personally, after using it, I prefer the  `onSubmit` and `onChange`  approach.

jamesgorrie (approved):
Neato!

PR #6020, opened 2024-05-22
update Quantum Metric script

GHaberis:

Update QM script


paulbrown1982 (approved):
👍

jamesgorrie (approved):
Is there any docs we can link to as to why we're updating this?

GHaberis:
> Is there any docs we can link to as to why we're updating this?

No just QM stating it'll enable a number of new features, it would be useful to know what

PR #6021, opened 2024-05-22
Abandoned Basket Cookie: fix valibot parsing error message

andrewHEguardian:

Fix the error message logged when the abandoned basket cookie fails to parse.

## Why are you doing this?

image

There a lot of sentry exceptions for failed parsing of the cookie, but the error message wasn't formatted correctly so can't see the actual parsing error.

Once we see what the errors are we can fix the issue, and potentially don't want this logged as an exception 

## How to test

Running locally, go to one time checkout, manually edit the `GU_CO_INCOMPLETE` cookie to be invalid, and see error in console with readable error message.

image


tomrf1:
Do we know why there are so many parsing errors?

andrewHEguardian:
> Do we know why there are so many parsing errors?

no, but hopefully we'll see once we log the error correctly

tomrf1 (approved):

PR #6022, opened 2024-05-24
Source upgrade

jamesgorrie:
Updates to the `@guardian/source` and the @guardian/source-development-kitchen`, deprecating the rest.

There was also some linting that needed updating from this, so this is done here. Mostly it was typing or changing `||` to `??` which should be no-ops.

## The E2E test have bee run that validate this

Screenshot 2024-05-24 at 11 39 12

## Filesize

I am not 100% sure why the filesize has increased - not sure if we think this is an issue?


support-frontend/assets/helpers/forms/paymentIntegrations/readerRevenueApis.ts
```diff
@@ -135,7 +131,7 @@ export type RegularPaymentFields =
 	| RegularAmazonPayPaymentFields;
 export type RegularPaymentRequestAddress = {
 	country: IsoCountry;
-	state?: UsState | CaState | null;
+	state?: UsState | null;
```
tjmw, line 134: I wonder if there'd be a way to collapse the `null` and `undefined` here? Do they have different meanings? (Probably something for a future PR!)

support-frontend/assets/helpers/internationalisation/country.ts
```diff
@@ -602,7 +602,7 @@ const countries: Record<IsoCountry, string> = {
 export type UsState = $Keys<typeof usStates>;
 export type CaState = $Keys<typeof caStates>;
 export type AuState = $Keys<typeof auStates>;
-export type StateProvince = UsState | CaState | AuState;
```
tjmw, line 605: Similar to above - what happened to Canada and Australia? 😃

support-frontend/assets/pages/paper-subscription-checkout/components/paperCheckoutForm.tsx
```diff
@@ -567,7 +567,7 @@ function PaperCheckoutForm(props: PropTypes) {
 						/>
 					) : null}
 					<GeneralErrorMessage
-						errorReason={props.submissionError}
+						errorReason={props.submissionError ?? undefined}
```
tjmw, line 570: Is this needed because `props.submissionsError` could be `null` but `GeneralErrorMessage` wants it to be something or `undefined`?

tjmw (approved):
This looks good to me (with the caveat that I don't know this project well, so probably makes sense to get a ✅  from someone who does).

I left a couple of questions, but no blockers, just interested to understand a couple of changes better.

paul-daniel-dempsey (approved):
Checkouts work fine, no linting errors remain and all pass on playwright. Maybe put in Tuesday morning.

support-frontend/assets/helpers/internationalisation/country.ts
paul-daniel-dempsey: Hmmm, unsure where this is used either, the checkout does work with state selector CAD+AUD states, possibly set in redux but does not appear to be called upon.

support-frontend/assets/helpers/forms/paymentIntegrations/readerRevenueApis.ts
jamesgorrie: Yea, there definitely is. It would be useful to have an initiate to have 1 `None` value (`undefined` please).

I've left this outside of the scope of this PR as it's quite far reaching as is and there might be some check somewhere that breaks in different conditions.

support-frontend/assets/pages/paper-subscription-checkout/components/paperCheckoutForm.tsx
jamesgorrie: It's because I removed the `Option` type from `GeneralErrorMessage`, an unnecessary type in TS, which could have had the value as `T | undefined | null`. 

It's now optional - as in `val?: string` - so this was coalescing the `null` into `undefined`.

support-frontend/assets/components/subscriptionCheckouts/personalDetails.tsx
```diff
@@ -87,7 +87,7 @@ export default function PersonalDetails(props: PropTypes): JSX.Element {
 		e: React.FocusEvent<HTMLInputElement>,
 	) => {
 		if (props.fetchAndStoreUserType) {
-			props.fetchAndStoreUserType(e.target.value);
+			void props.fetchAndStoreUserType(e.target.value);
```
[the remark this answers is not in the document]
jamesgorrie: [From the docs](https://www.typescriptlang.org/docs/handbook/2/functions.html#return-type-void)

> function type with a void return type (type voidFunc = () => void), when implemented, can return any other value, but it will be ignored.

Which is similar to `unknown` in the sense that it can't really be used but also `undefined` as in it will not be used.

In this instance it's to indicate "This value is not meaningful and will not be used").

support-frontend/assets/helpers/internationalisation/country.ts
jamesgorrie: For both the above - it's because the type of `CaState`, `AuState` and `UsState` is actually `string`.

I've left it for now, but will follow up in a PR immanently to fix it.

support-frontend/assets/components/csr/csrMode.ts
```diff
@@ -50,7 +46,7 @@ export type CsrCustomerData = {
 		email?: string;
 		street?: string;
 		city?: string;
-		state?: UsState | CaState;
```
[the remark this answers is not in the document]
jamesgorrie: [See the comment below](https://github.com/guardian/support-frontend/pull/6022#discussion_r1621959694)

jamesgorrie:
> just interested to understand a couple of changes better.

Really good exercise in me understanding the changes too!

PR #6024, opened 2024-05-24
fix: add hack to ticket tailor iframe sizing

jamesgorrie:
Removes the TT widget JS and uses an iframe directly.

What we lose is the iframe resizing capability, but this is OK as it's currently not working. This is with TT and they are raising a case for it.

This is mainly to allow people to test the flow E2E.

Weirdly it did work ([you can see it in this PR](https://github.com/guardian/support-frontend/pull/5980)) - I am not sure what has changed.

## New slightly substandard version

https://github.com/guardian/support-frontend/assets/31692/e81910f7-1dd1-4bd1-aad2-7ded0861bbe7


GHaberis (approved):
👍

PR #6033, opened 2024-05-30
style: format cdk files and add pre-push hook

AndreaDiotallevi:

- Format all files inside the `cdk` folder with Prettier
- Add `pre-push` hook to check for formatting and tests to improve devX


rupertbates (approved):
Looks good

PR #6039, opened 2024-05-31
refactor: remove experiments from settings

jamesgorrie:
The property is not set or used anywhere, so assuming it's legacy and removing it. 

This is part of an effort to get parsing between server & client a little tighter.


paul-daniel-dempsey (approved):
Looks Good

PR #6040, opened 2024-05-31
refactor: remove option types from user

jamesgorrie:
Remove option types from User type and use the more idiomatic optional parameter.

This is part of a larger piece of work to help with parsing between server and client.


paul-daniel-dempsey (approved):
Looks good

PR #6049, opened 2024-06-03
fix: update styling of payment button and ts and cs on checkout

jamesgorrie:
[Viewing without whitespace is recommended](https://github.com/guardian/support-frontend/pull/6049/files?w=1)

- Puts the button into the white background box
- Updates text on button to be "Pay £x per Month" (with currency and frequency alternatives)
- Adds Patrons text

The Ts and Cs read differently to PROD, but I think this is because we're being more specific to the product.

## This / PROD
Screenshot 2024-06-03 at 09 39 11


paul-daniel-dempsey (approved):
Works perfectly, think TC's are ok

PR #6052, opened 2024-06-03
feat: only serve JS & CSS from path

jamesgorrie:
This removes inlining of CSS and JS. The benefits of this would be:
- We can leverage browser caching of relatively large files
- [Get closer to removing unsafe-inline CSP policy](https://content-security-policy.com/unsafe-inline/)
- We can start using other performance attributes such as [`preload`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/preload)

## Stats
This branch performs marginally better on some metrics than others. I'd say they are negligible but open up more opportunity for improvement.

#### PageSpeed insights
Screenshot 2024-06-03 at 11 12 15

#### WebPageTest
Screenshot 2024-06-03 at 11 12 39

#### PageSpeed insights
Screenshot 2024-06-03 at 11 18 11

#### WebPageTest
Screenshot 2024-06-03 at 11 20 24


rupertbates (approved):
Looks good, was the inline style actually used anywhere? It looks like it only ever had a blank string passsed to it.

PR #6053, opened 2024-06-03
Reinstate the ticket tailor widget

jamesgorrie:
Ticket Tailor has fixed the issue with cross-origin iframe communication.

This re-instates the use of the widget over just having an iframe.


paul-daniel-dempsey (approved):
Yep can see it re-instated from comment

PR #6054, opened 2024-06-03
Add Ts&Cs to price summary on checkout

jamesgorrie:
Adds the Ts & Cs to the summary on the checkout.

Capture-2024-06-03-131800


paul-daniel-dempsey (approved):
Nice and neat!

PR #6055, opened 2024-06-03
Add missing benefits to checkout

jamesgorrie:
Adds the benefits you're missing out on to the checkout.

Capture-2024-06-03-142842

### DIFF
--- a/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
@@ -708,10 +708,25 @@ function CheckoutComponent({ geoId }: Props) {
 									}
 									amount={price}
 									currency={currency}
-									checkListData={productDescription.benefits.map((benefit) => ({
-										isChecked: true,
-										text: benefit.copy,
-									}))}
+									checkListData={[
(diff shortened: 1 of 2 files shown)


support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -708,10 +708,25 @@ function CheckoutComponent({ geoId }: Props) {
...
+									checkListData={[
+										...productDescription.benefits.map((benefit) => ({
+											isChecked: true,
+											text: benefit.copy,
+										})),
+										...(productDescription.missingBenefits ?? []).map(
+											(benefit) => ({
+												isChecked: false,
+												text: benefit.copy,
+												maybeGreyedOut: css`
```
Comment (jamesgorrie) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:720]: This is the current API - to use a `css` object. This could definitely be simplified, but I'm just reimplementing what we have to avoid breaking something in `PROD`.

Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:720]: Looks neat to re-use `maybeGreyedOut` to me

paul-daniel-dempsey (approved):
Looks Good

PR #6056, opened 2024-06-04
Add ophan tracking to country select

jamesgorrie:
This adds [`data-link-name` tracking](https://github.com/guardian/ophan/blob/a69e354d1bec4c0ba206a75fc62ea77a34f19a66/tracker-js/assets/click-path-capture.js#L31-L45) to the country `select`. I want to use this data to have an indication of how often we get the country wrong to help inform UX later on.

## Network request to Ophan
Screenshot 2024-06-04 at 09 11 24

### DIFF
--- a/support-frontend/assets/components/subscriptionCheckouts/address/addressFields.tsx
+++ b/support-frontend/assets/components/subscriptionCheckouts/address/addressFields.tsx
@@ -77,7 +77,7 @@ function statesForCountry(country: Option<IsoCountry>): React.ReactNode {
 
 export function AddressFields({ scope, ...props }: PropTypes): JSX.Element {
 	return (
-		<div>
+		<div data-component={`${scope}AddressFields`}>
 			<Select
 				css={marginBottom}
 				id={`${scope}-country`}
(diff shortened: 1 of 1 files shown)


support-frontend/assets/components/subscriptionCheckouts/address/addressFields.tsx
```diff
@@ -77,7 +77,7 @@ function statesForCountry(country: Option<IsoCountry>): React.ReactNode {
 
 export function AddressFields({ scope, ...props }: PropTypes): JSX.Element {
 	return (
-		<div>
+		<div data-component={`${scope}AddressFields`}>
```
paul-daniel-dempsey, line 80: Is this linked to the `data-link-name` property in the <Select?? Or is this for readability in the inspector?

paul-daniel-dempsey (approved):
`data-link-name` properties seem nice and simple to indicate to Ophan something has been clicked. Did we use this originally (if not I wonder why)?

support-frontend/assets/components/subscriptionCheckouts/address/addressFields.tsx
jamesgorrie: `data-component` helps to differentiate or "contain" events. The in multiple containers. Probably not massively useful for this, but it felt like good practise to define a boundry.

PR #6058, opened 2024-06-04
Fix e2e tests: update the cta selector on the 3-tier landing page

GHaberis:

We've updated the CTA copy on the 3-tier landing page, so the e2e test need updating so they can find it.

### DIFF
--- a/support-e2e/tests/tieredCheckout.test.ts
+++ b/support-e2e/tests/tieredCheckout.test.ts
@@ -45,6 +45,8 @@ test.describe('Subscribe/Contribute via the Tiered checkout)', () => {
 			const testFirstName = firstName();
 			const testLastName = lastName();
 			const testEmail = email();
+			const ctaCopy = testDetails.country === 'US' ? 'Subscribe' : 'Support';
+
 			await setupPage(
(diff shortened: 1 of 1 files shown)


support-e2e/tests/tieredCheckout.test.ts
```diff
@@ -162,7 +166,7 @@ test.describe('Supporter Plus promoCodes', () => {
 				card.getByText(testDetails.expectedPromoText).first(),
 			).toBeVisible();
 			await page
-				.getByRole('link', { name: 'Subscribe' })
+				.getByRole('link', { name: ctaCopy })
```
paul-daniel-dempsey, line 169: Presume you don't want to just leave as `Support` in case the country canges to `US`?

paul-daniel-dempsey (approved):
Looks good, smal comment thats all

support-e2e/tests/tieredCheckout.test.ts
GHaberis: Yep pretty much, although we don't have regional variants of this test, we may do like we do already on the other

PR #6062, opened 2024-06-04
fix: make css RefPath optional

jamesgorrie:
Makes the CSS `RefPath` optional for pages that aren't using CSS (and shouldn't be as this is the CSS generated via SASS).

[Fixes this bug](https://support.theguardian.com/uk/checkout?product=SupporterPlus&ratePlan=Monthly) where the `RefPath` is not being found. This is a runtime error so wasn't picked up on hence I have attached the Playwright test results. 

### 100% tested
Screenshot 2024-06-04 at 16 45 38

### DIFF
--- a/support-frontend/app/views/eventsRouter.scala.html
+++ b/support-frontend/app/views/eventsRouter.scala.html
@@ -9,7 +9,7 @@
   title = "Support the Guardian | Events",
   mainElement = EmptyDiv("events"),
   mainJsBundle = RefPath("[countryGroupId]/events/router.js"),
-  mainStyleBundle = RefPath(""),
+  mainStyleBundle = None,
   description = Some("Help us deliver the independent journalism the world needs. Support the Guardian by making a contribution."),
   canonicalLink = Some("https://support.theguardian.com/events"),
   hrefLangLinks = Map(),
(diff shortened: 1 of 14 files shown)


support-frontend/app/views/eventsRouter.scala.html
```diff
@@ -9,7 +9,7 @@
   title = "Support the Guardian | Events",
   mainElement = EmptyDiv("events"),
   mainJsBundle = RefPath("[countryGroupId]/events/router.js"),
-  mainStyleBundle = RefPath(""),
+  mainStyleBundle = None,
```
jamesgorrie, line 12: Here's where we needed a `None`

support-frontend/app/views/router.scala.html
```diff
@@ -24,7 +24,7 @@
   title = "Support the Guardian | Checkout",
   mainElement = EmptyDiv("checkout"),
   mainJsBundle = RefPath("[countryGroupId]/router.js"),
-  mainStyleBundle = RefPath(""),
+  mainStyleBundle = None,
```
jamesgorrie, line 27: And here's where we needed a `None`

paul-daniel-dempsey (approved):
Looks good

PR #6070, opened 2024-06-06
feat: add QM and GA tracking to thank-you

jamesgorrie:
Adds QA and GA analytics as per [the old thank you page](https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/pages/supporter-plus-thank-you/supporterPlusThankYou.tsx#L188-L216).

I haven't added the Ophan `trackUserData` due to our discussions that we are going to trim our tracking footprint as part of this migration.

There is still the idea of the `contributionType` on the thank you page that we are going to have to unpick. I don't want to block the checkout test for that though, so this is to parity, and we will do a little more as we move onto other products.


GHaberis (approved):
Nice 👏

PR #6072, opened 2024-06-06
Add supportAbTests to checkout

jamesgorrie:
Add `supportAbTests` to checkout.
I was just going through the `TODOs` and this stood out.

### DIFF
--- a/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
@@ -49,7 +49,10 @@ import { StripeCardForm } from 'components/stripeCardForm/stripeCardForm';
 import { AddressFields } from 'components/subscriptionCheckouts/address/addressFields';
 import type { PostcodeFinderResult } from 'components/subscriptionCheckouts/address/postcodeLookup';
 import { findAddressesForPostcode } from 'components/subscriptionCheckouts/address/postcodeLookup';
-import { getAmountsTestVariant } from 'helpers/abTests/abtest';
+import {
+	init as abTestInit,
+	getAmountsTestVariant,
+} from 'helpers/abTests/abtest';
 import { isContributionsOnlyCountry } from 'helpers/contributions';
 import { simpleFormatAmount } from 'helpers/forms/checkouts';
 import type { ErrorReason } from 'helpers/forms/errorReasons';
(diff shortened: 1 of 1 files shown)


GHaberis:
I think this is already done via the `setUpTrackingAndConsents()` call in https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/pages/%5BcountryGroupId%5D/router.tsx#L8, (granted you wouldn't guess `setUpTrackingAndConsents` also initialises the client-side ab test framework)

paul-daniel-dempsey (approved):
Yep routes `supportAbTests` through fine and test ran checkout ok.

GHaberis:
> @jamesgorrie I think the ab tests are already initialised via the `setUpTrackingAndConsents()` call in https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/pages/%5BcountryGroupId%5D/router.tsx#L8, (granted you wouldn't guess `setUpTrackingAndConsents` also initialises the client-side ab test framework)

@jamesgorrie I see why you've initialised the AB test framework in this module - so you can pass the tests to the checkout in the `supportAbTests` prop, however this means we're running the ab test `init` function from 2 places on page set-up, running the same code twice.  

Maybe we could refactor and extract the ab test `init` out of `setUpTrackingAndConsents` and pass the result of calling it into setupTrackingAndConsents and into the components from the router?

**router.tsx**

```
const countryId: IsoCountry = Country.detect();
const countryGroupId: CountryGroupId = CountryGroup.detect();

const supportAbTests = abTest.init({
    countryId,
    countryGroupId,
});

setUpTrackingAndConsents(supportAbTests);

const router = createBrowserRouter(
	geoIds.flatMap((geoId) => [
		{
			path: `/${geoId}/checkout`,
			element: <Checkout geoId={geoId} supportAbTests={supportAbTests} />,
		},
		{
			path: `/${geoId}/thank-you`,
			element: <ThankYou geoId={geoId} supportAbTests={supportAbTests} />,
		},
	]),
);
```

jamesgorrie:
@GHaberis this is true - but I'm not 100% sure I see the issue with it? Looking at [the code](https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/helpers/abTests/abtest.ts#L108-L122), all it does is transform data from a few sources? 

I think there might be a naming issue here where `init` is not really `init` but rather `getParticipations`?

GHaberis:
> setUpTrackingAndConsents

I don't think there are any negative consequences of calling `init` twice other than running the code twice, it'll always assign the user to the same set of AB tests, so it's not a blocker. I think we probably called it twice in the page set-up on the pages with Redux tbh, once in setUpTrackingAndConsents and again when setting up Redux.

GHaberis (approved):
Looks good 👍

PR #6084, opened 2024-06-10
feat: use roundel text on lozenge on product cards

jamesgorrie:
Takes the text from the promo tool (called `roundel`) and, if present, uses it instead of "Recommended" on the tier card.

image

There has been no text limi imposed, we could recommend this to editors or enforce it on the tooling side.


paul-daniel-dempsey (approved):
Looks ready to display offers in the lozenge

jamesgorrie (commented):
Nice work @paul-daniel-dempsey

PR #6085, opened 2024-06-10
feat: add starts and expires to PromotionSummary

jamesgorrie:
Adds `starts` and `expires` to the `PromotionSummary` model for us to be able to use in in the rendering of the Ts and Cs for promotions.

Screenshot 2024-06-10 at 17 07 56

### DIFF
--- a/support-frontend/test/controllers/PricesControllerSerialisationTest.scala
+++ b/support-frontend/test/controllers/PricesControllerSerialisationTest.scala
@@ -9,7 +9,7 @@ import com.gu.support.promotions.{
   Issue,
 }
 import controllers.PricesController.{CountryGroupPriceData, Prices, ProductPriceData, RatePlanPriceData}
-import org.joda.time.Days
+import org.joda.time.{DateTime, Days}
 import org.scalatest.flatspec.AnyFlatSpec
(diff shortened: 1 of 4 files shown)


support-frontend/test/controllers/PricesControllerSerialisationTest.scala
```diff
@@ -31,6 +31,8 @@ class PricesControllerSerialisationTest extends AnyFlatSpec with Matchers {
       None,
       Some(IntroductoryPriceBenefit(0.5, 3, Issue)),
       None,
+      new DateTime("2020-01-01"),
```
paul-daniel-dempsey, line 34: Why this particular date? Should we place this in a global const?

jamesgorrie: It's just a date, it doesn't affect anything about the test. What would you see as a benefit of putting it in a `val`?

paul-daniel-dempsey: Would guess not a lot to gain other than lining it up with tests below as well.

paul-daniel-dempsey (approved):
Looks good, will pull it through into Ts&Cs next

PR #6088, opened 2024-06-11
fix: dynamically set isTestUser on checkout

jamesgorrie:
Sets `isTestUser` based on their cookies.

Completely missed this. No harm done, we rolled back and [reverted](https://github.com/guardian/support-frontend/pull/6087).

We missed this as we hadn't tried a transaction as a non-test user. We'll make sure we do this pre-turning the test back on.

### DIFF
--- a/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
@@ -84,7 +84,7 @@ import {
 	getSupportAbTests,
 } from 'helpers/tracking/acquisitions';
 import { trackComponentClick } from 'helpers/tracking/behaviour';
-import { getUser } from 'helpers/user/user';
+import { getUser, isTestUser as isTestUserFunc } from 'helpers/user/user';
 import type { GeoId } from 'pages/geoIdConfig';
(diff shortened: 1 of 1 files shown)


support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -96,11 +96,12 @@ import { setThankYouOrder, unsetThankYouOrder } from './thank-you';
 /** App config - this is config that should persist throughout the app */
 validateWindowGuardian(window.guardian);
 
-const isTestUser = true as boolean;
 const csrf = window.guardian.csrf.token;
 
 const user = getUser();
 const isSignedIn = user.isSignedIn;
+const isTestUser = isTestUserFunc();
```
Comment (paul-daniel-dempsey) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:103]: Why not just call `isTestUser` direct?

paul-daniel-dempsey (approved):
Small comment , looks fine

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
Comment (jamesgorrie) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:103]: Yea, couldn't quite think of a better naming as the `const` is `isTestUser` which is also the method...

PR #6090, opened 2024-06-11
New Tier Three Product

rupertbates:
This PR adds the new Tier Three product into the code base as a separate product rather than an additional rate plan in Supporter Plus. Reasons for this are explained in [this document](https://docs.google.com/document/d/1NOata7_3YNsZ97bd312QQ9MYJrmko55CaO4D77ExUhI/edit?usp=sharing)

## TODOs (for follow up PRs)
- [ ] Finish off email integration, this is just a placeholder currently 
- [ ] Any extra validation needed beyond Guardian Weekly validation

### DIFF
--- a/support-frontend/app/controllers/CreateSubscriptionController.scala
+++ b/support-frontend/app/controllers/CreateSubscriptionController.scala
@@ -339,8 +339,12 @@ class CreateSubscriptionController(
         )
       case _: SupporterPlus =>
         List(
-          "gu_digital_subscriber" -> true.toString,
-          // "gu_supporter_plus" -> true.toString, // TODO: add this and remove the digisub one when the CMP cookie list has been updated
+          "gu_supporter_plus" -> true.toString,
+          "GU_AF1" -> DateTime.now().plusDays(1).getMillis.toString,
+        )
+      case _: TierThree =>
+        List(
+          "gu_supporter_plus" -> true.toString,
           "GU_AF1" -> DateTime.now().plusDays(1).getMillis.toString,
         )
       case _: DigitalPack =>
--- a/support-frontend/app/controllers/CreateSubscriptionController.scala
+++ b/support-frontend/app/controllers/CreateSubscriptionController.scala
@@ -339,12 +339,14 @@ class CreateSubscriptionController(
         )
       case _: SupporterPlus =>
         List(
-          "gu_supporter_plus" -> true.toString,
+          "gu_digital_subscriber" -> true.toString,
+          // "gu_supporter_plus" -> true.toString, // TODO: add this and remove the digisub one now that the CMP cookie list has been updated
           "GU_AF1" -> DateTime.now().plusDays(1).getMillis.toString,
         )
       case _: TierThree =>
         List(
-          "gu_supporter_plus" -> true.toString,
+          "gu_digital_subscriber" -> true.toString,
+          // "gu_supporter_plus" -> true.toString, // TODO: add this and remove the digisub one now that the CMP cookie list has been updated
           "GU_AF1" -> DateTime.now().plusDays(1).getMillis.toString,
         )
       case _: DigitalPack =>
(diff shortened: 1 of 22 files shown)


support-frontend/app/controllers/CreateSubscriptionController.scala
```diff
@@ -339,8 +339,12 @@ class CreateSubscriptionController(
         )
       case _: SupporterPlus =>
         List(
-          "gu_digital_subscriber" -> true.toString,
-          // "gu_supporter_plus" -> true.toString, // TODO: add this and remove the digisub one when the CMP cookie list has been updated
+          "gu_supporter_plus" -> true.toString,
```
rupertbates, line 342: This is in the cmp now

support-frontend/app/utils/CheckoutValidationRules.scala
```diff
@@ -102,6 +102,7 @@ object CheckoutValidationRules {
       case d: DigitalPack => DigitalPackValidation.passes(createSupportWorkersRequest, d)
       case p: Paper => PaperValidation.passes(createSupportWorkersRequest, p.fulfilmentOptions)
       case _: GuardianWeekly => GuardianWeeklyValidation.passes(createSupportWorkersRequest)
+      case _: TierThree => GuardianWeeklyValidation.passes(createSupportWorkersRequest) // TODO: is this right?
```
jamesgorrie, line 105: I think this is - but might be worth leaving a note why it's GWValidation (basically it's the same fields).

support-models/src/main/scala/com/gu/support/catalog/Product.scala
```diff
@@ -42,6 +42,37 @@ sealed trait Product {
...
+      id: String,
+      billingPeriod: BillingPeriod,
+      fulfilmentOptions: FulfilmentOptions = NoFulfilmentOptions,
+  ) =
+    ProductRatePlan(
+      id,
+      billingPeriod,
+      fulfilmentOptions,
+      NoProductOptions,
+      s"Supporter Plus ${billingPeriod.getClass.getSimpleName}",
```
jamesgorrie, line 56: Is this right?

support-workers/src/main/scala/com/gu/emailservices/ThreeTierEmailFields.scala
```diff
@@ -0,0 +1,39 @@
...
+import com.gu.support.workers.states.SendThankYouEmailState.SendThankYouEmailTierThreeState
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class ThreeTierEmailFields(
+    paperFieldsGenerator: PaperFieldsGenerator,
+    touchPointEnvironment: TouchPointEnvironment,
+) {
+  def build(state: SendThankYouEmailTierThreeState)(implicit ec: ExecutionContext): Future[EmailFields] = {
+    // TODO: don't really know what we need here yet
```
jamesgorrie, line 14: Nice to know where this is now!

support-workers/src/main/scala/com/gu/support/workers/lambdas/FailureHandler.scala
```diff
@@ -42,6 +42,10 @@ class FailureHandler(emailService: EmailService) extends Handler[FailureHandlerS
         FailedEmailFields.contribution(email = state.user.primaryEmailAddress, IdentityUserId(state.user.id))
       case _: SupporterPlus =>
         FailedEmailFields.supporterPlus(email = state.user.primaryEmailAddress, IdentityUserId(state.user.id))
+
+      case _: TierThree =>
+        // TODO: Do we need a new one here?
+        FailedEmailFields.supporterPlus(email = state.user.primaryEmailAddress, IdentityUserId(state.user.id))
```
jamesgorrie, line 48: Is this going to be closer to GW given that we might want to pass delivery details in the email. 

This is a consideration that we can look at while looking at the email.

jamesgorrie (approved):
This looks good - there's a question that looks like a copy/paste problem.

I was also wondering if we should have the email more aligned to delivery products i.e. GW whereas we're doing it more like S+ - but this is something we'll probably look at when doing the email.

support-models/src/main/scala/com/gu/support/catalog/Product.scala
rupertbates: Nope, thanks.

PR #6091, opened 2024-06-11
Merge pull request #6089 from guardian/revert-6087-revert-6082-add-va…

jamesgorrie:
Remove a lot from `babel`.

While trying to upgrade `preact` & `@gaurdian/tsconfig` I just realised there was a tonne of moving parts that I think have been added and forgotten over the years. This starts to try and get rid of that.


support-frontend/assets/components/paymentButton/defaultPaymentButtonContainer.tsx
```diff
@@ -77,10 +80,6 @@ export function DefaultPaymentButtonContainer({
 	const currency = currencies[currencyId];
 	const amountWithCurrency = simpleFormatAmount(currency, amount);
 
-	const { countryId } = useContributionsSelector(
-		(state) => state.common.internationalisation,
-	);
```
jamesgorrie, line 82: No idea how this worked in the first place as it was required before it was declared.

PR #6094, opened 2024-06-12
refactor: use i18nId instead of geo & supportInternationalisation

jamesgorrie:
Replaces `geoId` and `supportInternationalisationId` with `i18nId`.

Even though `countryGroupId` might be the more fitting name for this if we were basing it on the Scala model e.g. [for `uk`](https://github.com/guardian/support-frontend/blob/main/support-internationalisation/src/main/scala/com/gu/i18n/CountryGroup.scala#L28) - the term is completely [overloaded through the local codebase](https://github.com/search?q=repo%3Aguardian%2Fsupport-frontend%20countryGroupId&type=code) so the refactor would be large.

I started this, but then started to wonder if I could remove it all together, but [doing an org wide search for `countryGroupId`](https://github.com/search?q=org%3Aguardian%20countryGroupId&type=code) surfaces that we seem to overload it and use it interchangeably with [`Region`](https://github.com/guardian/support-admin-console/blob/8e3e1416c1343baab40a75dd019a36f92b1ea022/app/models/Region.scala#L12).

This led to the smallest refactor I could do as both `geoId` and `supportInternationalisationId` are both local.

This work spawned because I wanted to use these values in more places, and rule of three suggests I should standardise.

Neither naming felt great, but combining them I feel has given us an easy to understand name without having it related to the overloaded `CountryGroupId` / `Region`.


GHaberis:
So the benefit of doing this is we currently have `SupportInternationalisationId` and `GeoId` types which are the same, this combines all references tp the two old types into a single type? `CountryGroupId` would've been your preferred type name, for consistency with the Scala, but we have a type named this already used extensively in the client-side code, representing something else?

Struggling with this name `i18nId`, it's not particularly descriptive unless you know the acronym already? What about alternatives, such as `InternationalizationId`? Or just keep the existing  `SupportInternationalisationId` and  update any references to `GeoId` to use that?

jamesgorrie:
> So the benefit of doing this is we currently have `SupportInternationalisationId` and `GeoId` types which are the same, this combines all references tp the two old types into a single type? `CountryGroupId` would've been your preferred type name, for consistency with the Scala, but we have a type named this already used extensively in the client-side code, representing something else?

@GHaberis Yup - this exactly - and I was going to use it in one other place - so wanted to standardise.

> `InternationalizationId`

I think I will go with this as thee support prefix isn't useful, and I hope to push this up to [`@guardian/libs`](https://github.com/guardian/csnx/tree/main/libs/@guardian/libs/src/countries)

PR #6096, opened 2024-06-12
update google auth library

johnduffell:
the /test-users page was breaking with an error
JsResultException(errors:List((/locale,List(JsonValidationError(List(error.path.missing),List())))))

This PR seems to fix it - tested locally and in CODE.
Probably due to this fix https://github.com/guardian/play-googleauth/pull/218

I also raised this PR (edit: which has now been merged and released as 8.0.1), as we shoudl not be writing exception information to the UI
https://github.com/guardian/play-googleauth/pull/233


rupertbates (approved):
Good investigation 👍

PR #6099, opened 2024-06-13
Non mandatory ZipCode in US checkout

paul-daniel-dempsey:

We would like to make the US ZipCode field non-mandatory (ie Displayed but not required to be filled in to checkout).  Currently this affects->
1. US Contributions : Contribution Checkout
2. US Recurring S+ Contributions : Contribution Checkout
3. US Recurring Contribution: Generic Checkout

This is the client-side aspect ONLY. Server-side aspects to be determined in follow-up PR.

## Why are we doing in this?

As part of a change we made in Sep’23 we changes both the position of the Zipcode and made it mandatory , that has lead to an issue with Apple and Google pay payment in the US , as in these fail when the mandory field is not filled.

[**Trello Card**](https://trello.com/c/Cpvg3n5l/896-zipcode-make-it-non-mandatory-on-the-checkout-page-for-single-recurring-and-s-for-us)

### DIFF
--- a/support-frontend/assets/pages/supporter-plus-landing/setup/legacyActionCreators.ts
+++ b/support-frontend/assets/pages/supporter-plus-landing/setup/legacyActionCreators.ts
@@ -192,7 +192,8 @@ function regularPaymentRequestFromAuthorisation(
 	state: ContributionsState,
 ): RegularPaymentRequest {
 	const { actionHistory } = state.debug;
-	const { billingCountry, billingState } = getBillingCountryAndState(state);
+	const { billingCountry, billingState, postCode } =
+		getBillingCountryAndState(state);
 	const recaptchaToken = state.page.checkoutForm.recaptcha.token;
 	const contributionType = getContributionType(state);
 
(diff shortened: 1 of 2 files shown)


jamesgorrie (approved):

support-frontend/assets/pages/supporter-plus-landing/setup/legacyActionCreators.ts
```diff
@@ -221,7 +220,7 @@ function regularPaymentRequestFromAuthorisation(
 				? billingState
 				: null,
 			// required Zuora field if country is US or CA
-			postCode: billingCountry === 'US' ? postCode : null,
+			postCode: null,
```
paulbrown1982, line 223: Can we still provide it to Zuora/Stripe if it is given?

paul-daniel-dempsey: Good point, will test that now on Code and re-enable. Thanks.

paulbrown1982: Re: https://stripe.com/en-no/guides/optimizing-authorization-rates#:~:text=In%20particular%2C%20passing%20the%20ZIP,the%20customer%20and%20how%20much.

paul-daniel-dempsey: Yes, tested  and this is back in place.

PR #6102, opened 2024-06-14
feat: add stripe payment errors to generic checkout

jamesgorrie:
Adds notification of stripe payment errors to the generic checkout.

Screenshot 2024-06-14 at 12 54 34

This is an initial step - I'll be adding some more to the page incrementally.

Testing was done with [these testing card details](https://docs.stripe.com/testing).

### DIFF
--- a/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
@@ -20,6 +20,7 @@ import {
 	textInputThemeDefault,
 } from '@guardian/source/react-components';
 import {
+	ErrorSummary,
 	FooterLinks,
 	FooterWithContents,
 } from '@guardian/source-development-kitchen/react-components';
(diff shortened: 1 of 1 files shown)


support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -699,6 +712,8 @@ function CheckoutComponent({ geoId }: Props) {
 				supportAbTests,
 				debugInfo: '',
 			};
+			setErrorMessage(undefined);
```
Comment (paul-daniel-dempsey) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:715]: Is `undefined` better than explicitly checking for empty string `''` later down here?

paul-daniel-dempsey (approved):
Looks good. Just small comment whether we use `undefined` or empty string to indocat an empty  `errorMessage/errorContext`.

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
Comment (jamesgorrie) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:715]: The type for `errorMessage` is `string | undefined`.

`undefined` is representative of "No error".

This is essentially saying "Unset the errors". If we past a string it would be saying, "Set the error to an empty message".

We're not really checking here - but rather setting.

PR #6109, opened 2024-06-17
Tier3 Price Rise - July price update  (S+ and GW)

jamesgorrie:
1. Change Tier3 S+ hard coded prices under abtest `july2024PriceRise`

|Country|Monthly Before|Monthly After|Annual Before|Annual After|
|-----|-----|-----|-----|-----|
|US|image|image|image|image|
|UK|image|image|image|image|
|EU|image|image|image|image|

[Trello](https://trello.com/c/hUNa0AG2/881-tier-3-july-supporter-plus-price-rise-on-acquisition)

3rd Tier Testing Url:
https://support.thegulocal.com/uk/contribute#ab-july2024PriceRise=variant

### DIFF
--- a/support-frontend/assets/helpers/abTests/abtestDefinitions.ts
+++ b/support-frontend/assets/helpers/abTests/abtestDefinitions.ts
@@ -145,6 +145,27 @@ export const tests: Tests = {
 		targetPage: pageUrlRegexes.contributions.allLandingPagesAndThankyouPages,
 		excludeCountriesSubjectToContributionsOnlyAmounts: true,
 	},
+	july2024PriceRise: {
+		variants: [
+			{
+				id: 'control',
+			},
+			{
+				id: 'variant',
+			},
+		],
+		audiences: {
+			ALL: {
+				offset: 0,
+				size: 1,
+			},
+		},
+		isActive: false,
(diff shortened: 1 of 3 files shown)


paul-daniel-dempsey (approved):
Agree was quite a few deep ternaries.  My hand seemed forced I think because the `ProductCatalog international rateplans pricing contained only a USD ccy where as the Domestic had them all - 2 different shapes (according to typescript) which then needed the ternaries. Have aproved even though am on this, that ok?

support-frontend/assets/helpers/abTests/abtestDefinitions.ts
```diff
@@ -145,6 +145,27 @@ export const tests: Tests = {
 		targetPage: pageUrlRegexes.contributions.allLandingPagesAndThankyouPages,
 		excludeCountriesSubjectToContributionsOnlyAmounts: true,
 	},
+	july2024PriceRise: {
+		variants: [
+			{
```
GHaberis, line 150: If we're intending on using this AB test to launch the changes then we won't need the `control`? We should probably get rid of it from here as if we leave it in and enable the test 50% of users will see the old prices on Tier 3.

PR #6115, opened 2024-06-19
Shorten alarm description

tomrf1:
This alarm is now too long, causing the deploy to fail


andrewHEguardian (approved):
great copywriting 👍

PR #6116, opened 2024-06-19
fix: use international pricing for int on tier three

jamesgorrie:
Uses international pricing for `int` on tier three of the landing page.

| before | after |
|---|---|
| Screenshot 2024-06-19 at 13 59 15 | Screenshot 2024-06-19 at 13 54 56 |


paul-daniel-dempsey (approved):
Neat!

PR #6129, opened 2024-06-26
Set abandoned basket cookie on generic checkout

andrewHEguardian:

Setting the Abandoned Basket cookie on the generic checkout.

[**Trello Card**](https://trello.com/c/fJQjsvzY/513-enable-abandoned-basket-cookie-on-new-checkout)

## Why are you doing this?

The generic checkout is being AB tested at the same time as the abandoned basket banner. Currently the cooke is not set on the generic checkout, which means we're losing out on some potential banner views.

## How to test

Go to `https://support.thegulocal.com/uk/contribute#ab-useGenericCheckout=variant`, select tier one.
Then add `#ab-abandonedBasket=variant` to the URL.
See `GU_CO_INCOMPLETE` cookie set.

The banner CTA URL goes to the existing checkout, not sure if this is a risk.

## Screenshots

image

### DIFF
--- a/support-frontend/assets/helpers/storage/abandonedBasketCookies.ts
+++ b/support-frontend/assets/helpers/storage/abandonedBasketCookies.ts
@@ -10,7 +10,6 @@ import {
 	union,
 } from 'valibot';
 import * as cookie from 'helpers/storage/cookie';
-import type { ProductCheckout } from 'helpers/tracking/behaviour';
 import { logException } from 'helpers/utilities/logger';
 
 const COOKIE_EXPIRY_DAYS = 3;
@@ -26,7 +25,7 @@ const abandonedBasketSchema = object({
 type AbandonedBasket = InferInput<typeof abandonedBasketSchema>;
 
 export function useAbandonedBasketCookie(
-	product: ProductCheckout,
+	product: string,
 	amount: number,
 	billingPeriod: string,
 	region: string,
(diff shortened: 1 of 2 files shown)


support-frontend/assets/helpers/storage/abandonedBasketCookies.ts
```diff
@@ -26,7 +25,7 @@ const abandonedBasketSchema = object({
 type AbandonedBasket = InferInput<typeof abandonedBasketSchema>;
 
 export function useAbandonedBasketCookie(
-	product: ProductCheckout,
+	product: string,
```
andrewHEguardian, line 28: ideally I would make this more type safe instead of less, but I couldn't figure out how to simply make the types match

jamesgorrie: We are playing a little fast and loose here. At the moment this is `string` - as seen in [`productCatalog`](https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/helpers/productCatalog.tsx#L22).

If we are going to force it we should probably do it there.

Saying that, I just thought, maybe to avoid having TODOs all over the shot, we could add something like [this](https://github.com/guardian/support-frontend/pull/6133)?

jamesgorrie (approved):
Fab - just a note on if you want to use the quick hack I've put in for `ProductId`?

support-frontend/assets/helpers/storage/abandonedBasketCookies.ts
andrewHEguardian: that's a nice idea! Still a little bit of confusion about whether it has names like eg `NationalDelivery` or just `Paper`. Which we've included in [SDC](https://github.com/guardian/support-dotcom-components/blob/b84199a728184b11dcc7fbeb8de482cd996edb28/packages/shared/src/types/targeting/shared.ts#L47) and could cause issues further down the line

jamesgorrie: I think given that the source is our aimed solution for these things - we should try to keep the `ProductCatalogApi` as the source for the types. 

I am going to have a go at publishing the types to NPM at `@guardian/product-catalog`.

It will require syncing everything up but should avoid this model proliferation we have.

PR #6130, opened 2024-06-26
Update ProductCatalog every minute

jamesgorrie:
To help ensure the new prices are picket up as quickly as possible, we want to refresh the cache of the [Product API](https://product-catalog.guardianapis.com/product-catalog.json) every minute.


paul-daniel-dempsey (approved):
Great even simpler on Monday, no need to RiffRaff deploy to Prod presume.

PR #6131, opened 2024-06-26
T3 emails

jamesgorrie:
- adds `tier-three` and `tier-three-failure` emails
- renames a couple of values from `ThreeTier` => `TierThree`


rupertbates (approved):
Looks good, well done for standardising the names.

PR #6132, opened 2024-06-26
update docs for test user page

johnduffell:
as per this PR https://github.com/guardian/members-data-api/pull/1088, we now use email address not first name to detect test users in MDAPI/manage

This PR updates the bumf on the test users page to illustrate that.

### DIFF
--- a/support-frontend/app/views/testUsers.scala.html
+++ b/support-frontend/app/views/testUsers.scala.html
@@ -22,9 +22,7 @@
             All other fields can be populated as you wish.
             <h2>All recurring product purchases</h2>
             <ul>
-                <li><strong>First Name:</strong> @key</li>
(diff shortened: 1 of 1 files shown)


support-frontend/app/views/testUsers.scala.html
```diff
@@ -22,9 +22,7 @@ <h1>Test User generator.</h1>
             All other fields can be populated as you wish.
             <h2>All recurring product purchases</h2>
             <ul>
-                <li><strong>First Name:</strong> @key</li>
-                <br><strong>(Optional) Recommended Email (if you do not want to receive emails):</strong> @{key}@@thegulocal.com<br>
```
johnduffell, line 26: if you want to access manage, you would need to set a password, which means getting the validation email.  So I've removed the option not to get email.

rupertbates (approved):

PR #6133, opened 2024-06-26
refactor: add pseudo ProductId type

jamesgorrie:
Adds a type alias of `ProductId` => `string`. 

This way features implementing anything to do with the product catalog could use this type and as we make it stricter, all should be good, but we'll pick up on the errors.

The reason to not do it properly now is that [we have a use case](https://github.com/guardian/support-frontend/pull/6129#discussion_r1655120195), and it will take a few minutes / hours to update this as it's used a lot across the checkout.


tomrf1 (approved):
👍 as a temporary solution

PR #6134, opened 2024-06-26
use updated test users library to produce lower case only tokens

johnduffell:
use lower case only test user tokens by pulling in updated library - see https://github.com/guardian/identity-test-users/pull/10

I have tested locally and it generates test users that work with MDAPI locally.

Also the test-users page is updated accordingly 
image


support-frontend/build.sbt
```diff
@@ -25,7 +25,7 @@ libraryDependencies ++= Seq(
   "com.gu.identity" %% "identity-auth-play" % "4.12",
   "com.okta.jwt" % "okta-jwt-verifier" % oktaJwtVerifierVersion,
   "com.okta.jwt" % "okta-jwt-verifier-impl" % oktaJwtVerifierVersion % Runtime,
-  "com.gu" %% "identity-test-users" % "0.8",
+  "com.gu" %% "identity-test-users" % "0.10.2-SNAPSHOT",
```
johnduffell, line 28: tbc when this is released https://github.com/guardian/identity-test-users/pull/10

tjmw (approved):

PR #6135, opened 2024-06-27
Add @guardian/product-catalog as a dependency to supoprt-frontend

jamesgorrie:
- adds `@guardian/support-service-lambdas`
- tidies up `productCatalog` with some left over things from benefits tests etc

We're adding @guardian/product-catalog as a dependency to support-frontend so that we don't need to copy/paste types between repos, but rather use the source types from the source system.

~We use [gitpkg](https://gitpkg.vercel.app/about) as you can't install subfolders from Git in NPM.~ We were going to use this, but after a chat with @sndrs we thought there were
- security risks as they could drop any old code into that proxy
- complexity risks i.e. another cog in the machine
- stability risks i.e. if something goes wrong with there service we have no way of debugging or fixing it

This way we use a [standard path of installing with npm via github](https://docs.npmjs.com/cli/v10/commands/npm-install). It also leaves it in a relatively unrefined state outlining that this is to solve the problem that the code should probably be more closely co-located.

Dependant on https://github.com/guardian/support-service-lambdas/pull/2351


support-frontend/assets/helpers/productCatalog.tsx
jamesgorrie, line 1: [This file is easier to review with whitespace hidden](https://github.com/guardian/support-frontend/pull/6135/files?diff=unified&w=1#diff-aa345beaa3afa99181c298a3321c21403bc832b95b5793f25987b21a2c85c050)

support-frontend/assets/helpers/productCatalog.tsx
```diff
@@ -41,325 +45,278 @@ export function filterBenefitByRegion(
...
-			MonthlyWithGuardianWeeklyInt: {
-				billingPeriod: 'Monthly',
-			},
-			AnnualWithGuardianWeeklyInt: {
-				billingPeriod: 'Annual',
-			},
-		},
-	},
+export const productKeys = Object.keys(typeObject) as ProductKey[];
+export function isProductKey(val: unknown): val is ProductKey {
```
andrewHEguardian, line 49: candidate for [isOneOf](https://github.com/guardian/csnx/blob/32628d3d59087c58d39d8c0d3f6ee6b794e5a642/libs/%40guardian/libs/src/isOneOf/README.md)?

support-frontend/package.json
```diff
@@ -90,6 +90,7 @@
     "@guardian/pasteup": "1.0.0-alpha.7",
     "@guardian/source": "1.0.1",
     "@guardian/source-development-kitchen": "1.0.0",
+    "@guardian/support-service-lambdas": "guardian/support-service-lambdas#3a4a8d7c841b713b736115772a9aadd05995b66a",
```
andrewHEguardian, line 93: what does the hash represent?

support-frontend/tsconfig.json
```diff
@@ -13,7 +13,8 @@
 		"noImplicitAny": true, // Enable error reporting for expressions and declarations with an implied any type
 		"baseUrl": "./assets", // The baseUrl to be used when resolving absolute module paths
 		"paths": {
-			"ophan": ["../node_modules/ophan-tracker-js/build/ophan.support"]
+			"ophan": ["../node_modules/ophan-tracker-js/build/ophan.support"],
```
andrewHEguardian, line 16: side note: what's this ophan thing doing?

yarn.lock
```diff
@@ -0,0 +1,12 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
andrewHEguardian, line 1: unnecessary extra `yarn.lock`?

andrewHEguardian (approved):
Seems like a really cool solution! 👍 😎

support-frontend/tsconfig.json
GHaberis: It's a variant of Ophan's tracker-js module tailored for the support site (https://github.com/guardian/ophan/tree/main/tracker-js), we use the module for running Ophan in the client and tracking different events.

support-frontend/assets/helpers/productCatalog.tsx
jamesgorrie: Looking at [the code](https://github.com/guardian/csnx/blob/32628d3d59087c58d39d8c0d3f6ee6b794e5a642/libs/%40guardian/libs/src/isOneOf/isOneOf.ts), I do find it quite hard to read compared to the above. And probably way harder to change.

What benefit do you think we get from using it?

support-frontend/package.json
jamesgorrie: [It's the commit hash of the repo](https://github.com/guardian/support-service-lambdas/commit/70e330b5c433184981e2c6955025cbaa7666f353).

Strangely I can't find the official documentation for the resolver, but it's documented here: https://www.warp.dev/terminus/npm-install-from-github#remote-package-by-branch

jamesgorrie: I've updated it to be more explicit and use the whole URL to hopefully make this more clear.

yarn.lock
jamesgorrie: Removed

support-frontend/assets/helpers/productCatalog.tsx
andrewHEguardian: In general I think it's worthwhile to use a function if it already exists, and type guards always make me think twice.

In this case I thought maybe it could get rid of an `as` or two that we're using here. But I see your point 👍

PR #6142, opened 2024-07-01
feat: disabled setting country and state

paul-daniel-dempsey:
## What are you doing in this PR?

[**Trello Card**](https://trello.com)

## Accessibility test checklist

-  [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-  [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-  [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-  [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

PR #6143, opened 2024-07-02
Revert "ApplePay/GooglePay: PromoCode supplied is not applicable in this country"

paul-daniel-dempsey:
Reverts guardian/support-frontend#6142

Mandatory Country & State are required for US/CA/AU, even ApplePay/GooglePay users (so these fields need taking from there even if they conflict with a valid PromoCode region).


GHaberis (approved):
👍

PR #6146, opened 2024-07-03
T3 Pricing and Promotions

rupertbates:

This PR 
- updates the three tier landing page to be able to work with default promotions and promotions specified via a query parameter.
- updates the generic checkout to be able to work with default promotions and promotions specified via a query parameter.
- introduces a new AB test `tierThreeFromApi` which uses prices and promotions from the product catalog and product prices object instead of the old hard coded prices on both the landing page and the checkout.
- passes through query parameter containing the promo code to generic checkout
- make sure we only show one set of state and zip code fields for delivery products

# Changes
- add the `window.guardian.productPrices` object containing information about valid promotions to the generic checkout
- change the three tier landing page so that we now have two product prices objects - one for Supporter Plus and one for Tier Three.  Whereas previously we had one `window.guardian.productPrices` object we now have two: 
  - `window.supporterPlusProductPrices`
  - `window.threeTierProductPrices`

## TODO in future PRs
- update the design and on the generic checkout to match Figma designs
- make the thank you page work with T3 purchases

## Testing

3-Tier Landing Page, Tier3 PromoCode & GuardianWeekly Modified Checkout (current PROD) ->
https://support.thegulocal.com/us/contribute

3-Tier Landing Page, Tier3 PromoCode & Tier3 Generic Checkout->
https://support.thegulocal.com/us/contribute#ab-tierThreeFromApi=variant

Generic Checkout ->
https://support.thegulocal.com/uk/checkout?product=TierThree&ratePlan=DomesticMonthly&promoCode=TIER3_UK_MONTHLY


support-frontend/app/controllers/Application.scala
```diff
@@ -274,15 +281,29 @@ class Application(
...
 
   def router(countryGroupId: String): Action[AnyContent] = MaybeAuthenticatedAction { implicit request =>
-    implicit val settings: AllSettings = settingsProvider.getAllSettings()
+    request.queryString
+      .getOrElse("product", Nil)
+      .headOption
+      .flatMap(productString => Product.fromString(productString))
+      .map(routeForProduct(_))
+      .getOrElse(BadRequest("No product name provided"))
+  }
```
jamesgorrie, line 290: [We're doing this validation on the frontend](https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/pages/%5BcountryGroupId%5D/checkout.tsx#L321-L337).

The point of this was to move as much logic to the client as possible to avoid having to context switch. I don't think this is much of an issue, just something to be mindful of.

support-frontend/app/controllers/Application.scala
```diff
@@ -274,15 +281,29 @@ class Application(
...
     val isTestUser = testUserService.isTestUser(request)
     // This will be present if the token has been flashed into the session by the PayPal redirect endpoint
     val guestAccountCreationToken = request.flash.get("guestAccountCreationToken")
     val productCatalog = cachedProductCatalogServiceProvider.fromStage(stage, isTestUser).get()
 
+    val queryPromos =
+      request.queryString
+        .getOrElse("promoCode", Nil)
+        .toList
+    val productPrices = priceSummaryServiceProvider.forUser(isTestUser).getPrices(product, queryPromos)
```
jamesgorrie, line 305: To get your thoughts on this, I was trying to avoid using this model as:
- it is an extra layer of transformation / indirection between the source data and the app (we have traversed through this logic a lot while pairing when working on promotions)
- it needs context switching between the scala app and the JS app
- it's using the old Zuora catalog as source data

Maybe as a follow up we might look at how we pass down promo logic in the same shape as the source data, but validate promoCodes on the server, as I guess this is part of why this service exists?

support-frontend/app/services/pricing/DefaultPromotionService.scala
```diff
@@ -41,14 +43,20 @@ class DefaultPromotionServiceS3(
     extends DefaultPromotionService
     with LazyLogging {
 
+  // There is a DEV version of the default promos file, but CODE is much more use to us in regular use
```
jamesgorrie, line 46: Could we just remove this?

My assumption was we were moving towards:
- `CODE` for everything besides `PROD`
- `PROD` for `PROD``
- If we need a version for testing, we use a local file?

Non-blocking

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -311,7 +313,10 @@ function CheckoutComponent({ geoId }: Props) {
 	const productId = query.product in productCatalog ? query.product : undefined;
 	const product = productId ? productCatalog[query.product] : undefined;
 	const ratePlan = product?.ratePlans[query.ratePlan];
-	const price = query.price ?? ratePlan?.pricing[currencyKey];
+	const priceOriginal = query.price ?? ratePlan?.pricing[currencyKey];
+
+	const fulFilmentOption =
```
Comment (jamesgorrie) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:318]: tiny typo
```suggestion
	const fulfilmentOption =
```

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -744,7 +760,7 @@ function CheckoutComponent({ geoId }: Props) {
 					paymentMethod: paymentMethod,
 				};
 				setThankYouOrder(order);
-				window.location.href = `/${geoId}/thank-you`;
+				window.location.href = `/${geoId}/thank-you?product=${productId}&ratePlan=${query.ratePlan}&promoCode=${promoCode}`;
```
Comment (jamesgorrie) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:763]: This data [is being set](https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/pages/%5BcountryGroupId%5D/checkout.tsx#L746) in the [`localStorage` `order` object](https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/pages/%5BcountryGroupId%5D/thank-you.tsx#L56-L77).

The reasoning is that the thank you page is based on the state of a complete order, which is reset when trying to make a new order.

Do you think this is easier to reason about? If so, it would be good to remove the `localStorage` implementation all together. 

If we want to keep that solution maybe this is a sign we should signpost it better.

jamesgorrie (approved):
This is great and easy to understand. 

I do have a few questions I would really appreciate feedback as it is on thoughts I have been having during the refactor.

support-frontend/app/controllers/Application.scala
rupertbates: Yes I totally agree that we should move beyond using this approach but I didn't want to guess at how you were planning on doing that and this seemed the easiest way to get promotions going for now.

support-frontend/app/services/pricing/DefaultPromotionService.scala
rupertbates: Yes I would be inclined to get rid of it, not sure I can think of a use for it that you couldn't achieve by using code or a unit test.

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
Comment (rupertbates) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:763]: Personally I prefer query strings in general because then the url contains everything it needs to render the page. Otherwise the page can be broken if the correct state is not set in local storage but it won't be obvious why, it's like the difference between a pure function and one with side effects. Having said that, if you think the local storage solution works better in this case I'm happy to change it.

support-frontend/app/services/pricing/DefaultPromotionService.scala
jamesgorrie: You might want to see things locally working for e2e testing - but having some sort of local file could work for that?

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
Comment (jamesgorrie) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:763]: Indeed - I just thought the page was specific to your order, hence the differentiation. I like the idea of keeping the state URL driven - and the worst I can see happening is someone changing the URL params and seeing the page change...

PR #6150, opened 2024-07-08
Sort states in config

andrewHEguardian:

Ordering the states objects by their values, instead of the key 

## Why are you doing this?

The generic checkout states appear out of alphabetical order, they are sorted in code on the other checkouts. It seems like an unnecessary step to do this at runtime.

## How to test

Go to generic checkout in US, see states in alphabetical order.

## Screenshots

Generic checkout
Before
image
After
image


rupertbates (approved):
👍

PR #6136, opened 2024-06-27
Feast: Tooltip & Tracking Link Changes

paul-daniel-dempsey:
## What are you doing in this PR?

Fast Tooltip and Tracking Link Chnages specified below:-

- Tooltip copy ->
`Make a feast out of anything with the Guardian’s new recipe app. Feast has thousands of recipes including quick and budget-friendly weeknight dinners, and showstopping weekend dishes – plus smart app features to make mealtimes inspiring.`

- `new benefits` reads `new`

- Thankyou Page Single Android/Ios Download Link (tracking implied) ->
https://guardian-feast.go.link?adj_t=1dufrlhf&adj_fallback=https%3A%2F%2Fwww.theguardian.com%2Fhelp%2Finsideguardian%2F2024%2Fapr%2F17%2Fintroducing-the-feast-app&adj_redirect_macos=https%3A%2F%2Fwww.theguardian.com%2Fhelp%2Finsideguardian%2F2024%2Fapr%2F17%2Fintroducing-the-feast-app

[**Trello Card**](https://trello.com/c/6NYrax1T/857-wording-and-app-download-changes-update-to-feast-on-lp-ty-page)

## Screenshots
|before|after|
|-----|-----|
|image|image|

Testing
https://support.thegulocal.com/uk/contribute#ab-feast=variant

PR #6151, opened 2024-07-08
Add feast benefit to checkout step 2 benefits list

andrewHEguardian:

Add a Feast benefit to the benefits list of step 2 of the checkout. #6136 updated the copy of the Feast offer on the landing page. This PR adds the benefit on the next step of the checkout.

This change is not behind the `feast` AB test so can only be merged once that test goes live (it's not actually an AB test as there is only one variant)
 
 

## How to test

Go to https://support.thegulocal.com/uk/contribute/checkout?selected-amount=12&selected-contribution-type=monthly&product=SupporterPlus, click "View Details" and see Feast benefit.

## Screenshots

Before
image

After
image


GHaberis (approved):
👍

PR #6152, opened 2024-07-09
Remove non-mandatory zip code validation message on checkout

andrewHEguardian:

Remove the validation on zip code used in two places: the legacy contributions checkout and the new generic checkout.

[**Trello Card**](https://trello.com/c/jpZYNi8q/933-non-mandatory-field-zip-code-incorrectly-labelled-as-mandatory-when-client-side-validation-errors-are-displayed)

## Why are you doing this?

On the old contributions and S+ checkouts, #6099 made the zip code field in personal details non-mandatory when the form is submitted. However the field still shows the validation error for zip code.

On the generic checkout the field was mandatory, which it shouldn't be.

## How to test

Go to any of these checkout flows and hit submit without entering any info - do not see error message beside zip code.

- US RC - Generic
- US S+ - Generic
- US RC - Non-generic
- US S+ - Non-generic

The following checkouts should have a mandatory validation of the post code/zip code in delivery/billing address

- GW
- Paper
- Tier Three - Generic

### Non-generic

Before
image

After
image

### Generic

Before
image

After
image


rupertbates (approved):
Looks good

PR #6154, opened 2024-07-09
Validate zip code on checkout if not empty

andrewHEguardian:

Fix the zip code validation on non-generic checkout by including the "path" property, so it knows which field to attach the validation message to.

## How to test

In non-generic checkout 
- enter a zip code in personal details that doesn't match the pattern. Submit form. See error.
- leave zip code blank. Submit form. See no error.

## Screenshots

Before
image

After
image


GHaberis (approved):
Looks good 👍

PR #6166, opened 2024-07-17
Increase timeout from e2e workflow to 45 mins

GHaberis:

This e2e workflow has been timing out after 30 mins, the chance of it timing out is greater on runs when a test has failed as we retry failed test runs, when the workflow times out no alerts are sent to inform us the test run has failed. I'm increasing to 45 mins to hopefully improve the chances of test runs completing and alerting in the event of failure.

### DIFF
--- a/.github/workflows/playwright.yaml
+++ b/.github/workflows/playwright.yaml
@@ -12,7 +12,7 @@ jobs:
   test:
     if: ${{ github.event.label.name == 'Seen-on-PROD' }}
     name: Playwright
-    timeout-minutes: 30
+    timeout-minutes: 45
     runs-on: ubuntu-latest
     env:
       BROWSERSTACK_USERNAME: ${{ secrets.BROWSERSTACK_USERNAME }}


jamesgorrie (approved):

jamesgorrie (approved):

jamesgorrie (commented):
@GHaberis - it looks like when tests are running well they come in in under 30 minutes.

The only time they take longer is if they're broken, meaning a PR like this would just increase the time to knowing there's an issue.

Do you think this understanding is right, and if it is, we should close this?

GHaberis:
> @GHaberis - it looks like when tests are running well they come in in under 30 minutes.
> 
> The only time they take longer is if they're broken, meaning a PR like this would just increase the time to knowing there's an issue.
> 
> Do you think this understanding is right, and if it is, we should close this?

I think the issue is we're not currently notified when these failed builds occur as the action times out after 30 mins, so we're not increasing the time it takes to notify us, as right now we're not being notified at all.

jamesgorrie:
@GHaberis does [this PR](https://github.com/guardian/support-frontend/pull/6181) solve the same thing?

GHaberis:
> @GHaberis does [this PR](https://github.com/guardian/support-frontend/pull/6181) solve the same thing?

Yep, that's a better solution! Happy to close this PR in that case. @jamesgorrie

PR #6167, opened 2024-07-17
fix: reinstate productPrices on contributions template

jamesgorrie:
`productPrices` is used throughout the codebase, mostly in redux, so is used a lot in the redux state. This reinstates it in the contributions template.

We were trying to be more specific in our naming, but unfortunately this coupling caught us. We started to explore using the more explicit naming, but realised that the coupling is quite deep. 

Given that we are hoping to deprecate this template, and thus this implementation, I have just sign-posted how we use the values. 

## `main`
Screenshot 2024-07-17 at 11 09 55

## `product-prices` playwright
Screenshot 2024-07-17 at 11 25 31

### DIFF
--- a/support-frontend/app/views/contributions.scala.html
+++ b/support-frontend/app/views/contributions.scala.html
@@ -1,6 +1,6 @@
 @import admin.settings.AllSettings
 @import admin.ServersideAbTest.Participation
-@import assets.{AssetsResolver, RefPath, StyleContent}
+@import assets.{AssetsResolver, RefPath}
 @import com.gu.identity.model.{User => IdUser}
 @import views.ReactDiv
 @import models.GeoData
@@ -28,7 +28,7 @@
   shareUrl: String,
   v2recaptchaConfigPublicKey: String,
   serversideTests: Map[String, Participation] = Map(),
-  productPrices: LandingPageProductPrices,
+  landingPageProductPrices: LandingPageProductPrices,
   productCatalog: JsonObject,
 )(implicit assets: AssetsResolver, request: RequestHeader, settings: AllSettings)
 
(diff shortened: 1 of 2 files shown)


support-frontend/app/views/contributions.scala.html
```diff
@@ -55,8 +55,16 @@
       };
     }
 
-    window.guardian.supporterPlusProductPrices = @Html(outputJson(productPrices.supporterPlusProductPrices))
-    window.guardian.tierThreeProductPrices = @Html(outputJson(productPrices.tierThreeProductPrices))
+    /** Both the landing page (`/uk/contribution`) and contributions checkout (`/uk/contribution/checkout`) are rendered from this template.
+     *
+     * We use `supporterPlusProductPrices` and `tierThreeProductPrices` on the landing page as it renders both products with their promotions.
+     *
+     * We use `productPrices` in the checkout as the value is coupled across the codebase via redux state used across multiple checkouts i.e. Guardian Weekly, Paper, Contribution etc
```
rupertbates, line 62: And this is only the T2 checkout right? Because T3 is using GW checkout currently, and generic checkout has something different?

rupertbates (approved):

paul-daniel-dempsey (approved):
Looks great, thanks

support-frontend/app/views/contributions.scala.html
jamesgorrie: Yea, T2 @ `/uk/contribute/checkout` which uses `SupporterPlus` `productPrices`.

T3 @ `/subscribe/weekly/checkout` uses `productPrices` but from the [`GuardianWeekly` product](https://github.com/guardian/support-frontend/blob/main/support-frontend/app/controllers/WeeklySubscriptionFormController.scala#L68).

PR #6168, opened 2024-07-17
fix: add productCatalog to subscriptionCheckout

jamesgorrie:
Currently when trying to purchase [a digital edition](https://support.theguardian.com/uk/subscribe/digitaledition) - the loading screen never makes it to the thank you page as it breaks  when trying to read the `productCatalog`.

The purchase is still made, and emails sent, the user just never get's the feedback from the web UI that the transaction was successful.

As this template is used by all subscription checkouts, we needed to add it to all the controllers. This hopefully creates a little consistency in the short term.

Long term we are going to deprecate these checkout with the generic checkout.


jamesgorrie:
> Just a question around how long this has been broken, is this a result of something we've changed recently or has it been broken for ages?

This has been broken for a while. There would have been some change implementing the S+ into the thank you page that broken it.

I am hoping to get a better understanding of how we can get our e2e testing working more efficiently, and then add more coverage to catch these cases that are easy to forget about.

PR #6169, opened 2024-07-17
Provide a default product match for /p/ endpoint

rupertbates:
support-frontend has some funtionality where a user can put in a url in the form `/p/[promocode]` and they will be redirected to the landing page for the particular product which the promo applies to.

It seems that this functionality has been forgotten about when recent products (supporter plus and tier three) have been added, so any promotions for those products which were sent through this endpoint were causing 500 errors.

This PR provides a default case for any unknown product types which is to redirect to the main three tier landing page. This is correct for supporter plus and tier three, and is better than an error page for any future products that we might develop.


paulbrown1982 (approved):
👍

PR #6181, opened 2024-07-22
feat: alert on e2e test timeouts

jamesgorrie:
Alerts our SR dev team channel when the github action timeouts, or fails for a reason outside of the Browserstack context.

This is a reactive fix due to the fact that if a test suit in Browserstack takes more than the timeout, we abruptly disconnect, but don't get the failure state from browserstack and thus not alert.

We get into this state as some bugs can just leave the browser hanging without failing. We are going to do some more work around mitigating against this in the tests, but this will at least be a catch-all given that these tests are integral to catching system errors.

Screenshot 2024-07-22 at 13 43 52

### DIFF
--- a/.github/workflows/playwright.yaml
+++ b/.github/workflows/playwright.yaml
@@ -12,7 +12,7 @@ jobs:
   test:
     if: ${{ github.event.label.name == 'Seen-on-PROD' }}
     name: Playwright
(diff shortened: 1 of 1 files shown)


tomrf1 (approved):
👏

PR #6182, opened 2024-07-22
Simplify playwright config

jamesgorrie:
- Removes the large, global `expect` timeout in favour of only using long running `expect` assertions for the thank-you page wait
- Removes the `workers` settings as this is set automatically to the best it can be i.e. `"Defaults to half of the number of logical CPU cores"`. We only have 1 in CI, so it runs with 1 worker.

None of this is expected to speed up our playwright tests, but simplifies the config to help make impactful changes.


andrewHEguardian (approved):
nice 👍

PR #6187, opened 2024-07-23
fix: up paypal timeout to 10 seconds in e2e test

jamesgorrie:
Upping the PayPal timeout as it was failing at 5 seconds.

It feels like PayPal should be able to handle this - it might be because it's a sandbox,  but still feels a little slow.


rupertbates (approved):
Hopefully this will be enough. It's timeouts on top of timeouts all the way down isn't it?!

PR #6193, opened 2024-07-29
Add billing_period into T3 email fields

rupertbates:
Adds these fields to the T3 email
- `billing_period`
- `first payment date`
- `subscription_details`

Whilst these fields are available via some of the other fields from the `paperFieldsGenerator` - I've chosen to just add these to align with the S+ fields as that's what we copied from in Braze.

We could probably start consolidating this - although I wonder if that would be part of a larger initiative to do this throughout the pipeline.


tjmw (approved):
This looks good to me 👍

PR #6195, opened 2024-07-29
Remove dependabot exclusions from GHA

andrewHEguardian:

Remove dependabot exclusions in our workflows, we should build dependabot PRs in the same way as any other to give us confidence in the changes.

### DIFF
--- a/.github/workflows/payment-api-build.yml
+++ b/.github/workflows/payment-api-build.yml
@@ -12,9 +12,8 @@ env:
 jobs:
   payment_api_build:
     if: >-
-      (github.actor != 'dependabot[bot]') &&
-        (github.repository_owner == 'guardian' ||
-          github.event_name == 'push')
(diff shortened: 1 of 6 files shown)


jamesgorrie (approved):
Great!

For another PR - why are the other two checks there?

The `github.event_name == 'push'` would stop `workflow_dispatch` and `pull_request` events triggering them?

And I am not sure what the `github.repository_owner == 'guardian'` rule is trying to achieve? Maybe something before the security of actions prevented other users triggering actions from forks?

jamesgorrie:
@andrewHEguardian shall we get this in?

andrewHEguardian:
> @andrewHEguardian shall we get this in?

Can do - but I hesitated when I realised how many Scala projects there are in the repo. Logically dependabot can never affect these so it's wasted compute to execute them right?

PR #6197, opened 2024-07-29
E2E Playwright Testing : Generic Checkout Tier3 

paul-daniel-dempsey:

Playwright : Tier3 testing chnags :-
- Remove repeat 2xcontribute genericTests
- Add 3xTier3 test (per paymentType)
- Add Billing/Delivery address difference test
- Rename tests into Tiers

[**Trello Card**](https://trello.com/c/X65MMrZ8/974-end-2-end-testing-tier-3-generic-checkout)

## How to test

`cd e2e-support`
`yarn test-local`

## Screenshots
image

### DIFF
--- a/support-e2e/tests/utils/testUserDetails.ts
+++ b/support-e2e/tests/utils/testUserDetails.ts
@@ -1,6 +1,7 @@
 import { Page } from '@playwright/test';
+import { TestDetails } from '../tieredCheckout.test';
 
-export const setTestUserDetails = async (
+export const setTestUserRequiredDetails = async (
 	page: Page,
 	firstName: string,
 	lastName: string,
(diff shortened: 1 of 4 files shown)


support-e2e/tests/utils/testUserDetails.ts
```diff
@@ -10,3 +11,64 @@ export const setTestUserDetails = async (
...
+			.getByRole('checkbox', {
+				name: 'Billing address same as delivery address',
+			})
+			.uncheck();
+	}
+
+	if (testDetails.fields.addresses) {
+		let index = 0;
+		// To run in sequence using async/await, for required over forEach
+		for (const address of testDetails.fields.addresses) {
```
andrewHEguardian, line 37: this is clever - but maybe better to just make it really simple and repeat ourselves? We're only going to have two addresses ever right? Delivery and billing

andrewHEguardian (approved):
Nice! 👍

support-e2e/tests/utils/testUserDetails.ts
paul-daniel-dempsey: Yes, its not perfect, but just wanted to keep repeat code to a minimum thus using the loop. Hope that's ok.

PR #6200, opened 2024-07-30
Remove existing payment methods

andrewHEguardian:

Removing the existing payment method functionality from the frontend. This includes existing card and existing direct debit.

[**Trello Card**](https://trello.com/c/YI4wttbD/970-deprecate-and-remove-existing-card-methods)

## Why are you doing this?

Existing card and existing direct debit are feature switches that have been off for a while in the RRCP. These features have not been migrated to the new checkout, and there is no indication that the features should be turned back on.

After this PR there is some extra work to remove the `PreparePaymentMethodForReuse` lambda that forms part of the support workers step function

image

## How to test

Test the following checkouts still work.

- [x] Generic checkout
- [x] Digital edition
- [x] Supporter Plus
- [x] Newspaper
- [x] Guardian Weekly

All playwright tests pass locally.

1. Breaking the checkout! Mitigate by releasing early in the day and monitoring, plus above testing.
2. This feature is requested again...


paul-daniel-dempsey (approved):
Thats a LOT of code removed, difficult to triple check without doing all the removals but  testing fine via Playwright and no linting errors picking up anything. It might be best to wait until approval for this and possibly remove the switches from RRCP at the same time?

GHaberis (approved):
All looks good, a fair amount of complex code removed 🧠 , good to clean out unused features! 🧹

PR #6201, opened 2024-07-31
fix: reorder Checkout and CheckoutComponent for clarity

jamesgorrie:
As we've been working through the file and the PRs, the order in which the code is run, currently bottom to top, is confusing.

This is a no-op, but puts the initially run code at the top.


paul-daniel-dempsey (approved):
Looks good, like it at the top.

PR #6208, opened 2024-08-07
fix: remove selectedContributionType from sessionStorage

jamesgorrie:
Removes `selectedContributionType` from `sessionStorage`. This adds complexity both to the code and to the UX.

If we want to store this information - we should use either URL params (already exist) or `window.location.hash` values if we're controlling the UI.

This was identified as I was looking a the tab interface on the landing page in preparing for the Growth team working on it.

### DIFF
--- a/support-frontend/assets/helpers/redux/checkout/product/contributionsSideEffects.ts
+++ b/support-frontend/assets/helpers/redux/checkout/product/contributionsSideEffects.ts
@@ -67,20 +67,6 @@ export function addProductSideEffects(
 	});
 
 	startListening({
-		actionCreator: setProductType,
-		effect(action) {
-			/**
-			 * selectedContributionType is read from storage in 2 places:
-			 * 1) It is used to set the selected contribution type on return
-			 * visits to the Contributions Landing Page in the same session.
-			 * 2) It is used on the Contributions Thank You page to send data to Quantum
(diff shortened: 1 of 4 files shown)


support-frontend/assets/helpers/redux/checkout/product/contributionsSideEffects.ts
```diff
@@ -66,20 +66,6 @@ export function addProductSideEffects(
...
 	});
 
-	startListening({
-		actionCreator: setProductType,
-		effect(action) {
-			/**
-			 * selectedContributionType is read from storage in 2 places:
-			 * 1) It is used to set the selected contribution type on return
-			 * visits to the Contributions Landing Page in the same session.
-			 * 2) It is used on the Contributions Thank You page to send data to Quantum
```
andrewHEguardian, line 76: How (or are we even) sending the data to QM now? And what about the cookie

jamesgorrie, line 76: I've tested this on the thank-you page, and it works for all three contribution types.

jamesgorrie: I _think_ this might be for if you refresh the page. Reason being we don't store this information in the URL of the thank you page.

PR #6209, opened 2024-08-07
Make feature switches optional in support-frontend

andrewHEguardian:

Making the feature switches optional - a missing feature switch should behave the same as an Off feature switch.

[**Trello Card**](https://trello.com/b/KiMIxAsm/sr-portfolio-platform-team)

## Why are you doing this?

Adding a new feature switch to the code without adding it to the RRCP first or deleting an existing feature switch would cause a runtime error running the frontend. We should handle this better by making it an `Option`

image

This alert does show in the RRCP, but it's unintuitive that it can be deleted through the UI when it's hardcoded in code
image

## How to test

Run the site locally - delete a dev feature switch by running `support-admin-console` locally - no error.

## Next steps:

`support-payment-api` also uses these feature switches, we will do that in a later PR


jamesgorrie (approved):
Fabulous!

PR #6212, opened 2024-08-08
refactor: move data fetching and validation outside of checkout component

jamesgorrie:
Moves _some_ (soon to be more) data fetching and validation outside the component and into the fetcher.

This is a common pattern in [NextJS](https://nextjs.org/docs/app/building-your-application/data-fetching/fetching-caching-and-revalidating), [Astro](https://docs.astro.build/en/guides/data-fetching/) and many others. 

I've done this as I was preparing the way for S+ to make it's way into the checkout and was finding it quite complex, so I hope this simplifies the path.

There is a small challenge that the data fetching is from a JS literal. You might normally have an API that 404s etc, but we are having to do those checks explicitly in the fetching method.

i.e.
- Get the `product` query string param
- Check it exists in the `products` object
- fail if not
- Get the `ratePlan` query string param
- Check it exists in the `product` object
- Do the same for price

The code is a bit verbose, but I think it's clear and easy to change - which feels like a decent payoff.

### DIFF
--- a/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
@@ -89,9 +89,11 @@ import type { AppConfig } from 'helpers/globalsAndSwitches/window';
 import CountryHelper from 'helpers/internationalisation/classes/country';
 import type { IsoCountry } from 'helpers/internationalisation/country';
 import { countryGroups } from 'helpers/internationalisation/countryGroup';
+import type { ProductKey } from 'helpers/productCatalog';
 import {
 	filterBenefitByRegion,
 	isProductKey,
+	productCatalog,
 	productCatalogDescription,
 } from 'helpers/productCatalog';
 import { NoFulfilmentOptions } from 'helpers/productPrice/fulfilmentOptions';
@@ -262,21 +264,6 @@ const processPayment = async (
 	});
 };
 
-/** QueryString - this is setup specifically for the checkout page */
-function isNumeric(str: string) {
-	return !isNaN(parseFloat(str));
-}
-const searchParams = new URLSearchParams(window.location.search);
-const searchParamsPrice = searchParams.get('price');
-const query = {
-	product: searchParams.get('product') ?? '',
-	ratePlan: searchParams.get('ratePlan') ?? '',
-	price:
-		searchParamsPrice && isNumeric(searchParamsPrice)
-			? parseFloat(searchParamsPrice)
-			: undefined,
-};
-
 /** Form Validation */
 /**
  * This uses a Unicode character class escape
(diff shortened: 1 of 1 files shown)


support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -316,10 +303,51 @@ function ChangeButton({ geoId }: ChangeButtonProps) {
...
 	const { currencyKey } = getGeoIdConfig(geoId);
+	const searchParams = new URLSearchParams(window.location.search);
+
+	/** Get and validate product */
+	const productParam = searchParams.get('product');
+	const productKey =
+		productParam && isProductKey(productParam) ? productParam : undefined;
+	const product = productKey && productCatalog[productKey];
+	if (!product) {
+		return <div>Product not found</div>;
```
Comment (jamesgorrie) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:317]: We could probably start returning the actual error pages here.

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -442,125 +461,106 @@ function CheckoutComponent({
 	 * We might be able to defer this to the backend.
 	 */
 	let productFields: RegularPaymentRequest['product'];
-
-	if (productId === 'TierThree') {
```
Comment (jamesgorrie) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:446]: This was swapped out for a `switch` statement as [`switch` statements check for exhaustiveness, not `if` statements](https://github.com/microsoft/TypeScript/issues/38731)... 😿

andrewHEguardian (approved):
Great!

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -316,10 +303,51 @@ function ChangeButton({ geoId }: ChangeButtonProps) {
...
+	const ratePlan = ratePlanKey && product.ratePlans[ratePlanKey];
+	if (!ratePlan) {
+		return <div>Rate plan not found</div>;
+	}
+
+	/** Get and validate price */
+	let price: number | undefined;
+	if (productKey === 'Contribution') {
+		const priceParam = searchParams.get('price');
+		price = priceParam ? parseInt(priceParam, 10) : undefined;
```
Comment (paul-daniel-dempsey) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:341]: Does 10 need to b specified inside parseInt?

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -316,10 +303,51 @@ function ChangeButton({ geoId }: ChangeButtonProps) {
...
+	} else {
+		price =
+			currencyKey in ratePlan.pricing
```
Comment (paul-daniel-dempsey) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:344]: Does this work for international US - do we pressume the US price?

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -371,17 +399,31 @@ export function Checkout({ geoId, appConfig }: Props) {
 			<CheckoutComponent
 				geoId={geoId}
 				appConfig={appConfig}
+				productKey={productKey}
```
Comment (paul-daniel-dempsey) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:402]: Neat!

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
Comment (jamesgorrie) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:341]: [From the docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#description)

> Be careful — this does not always default to 10!

paul-daniel-dempsey (approved):
Looks good, small comment presuming international is usd amount.

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
Comment (jamesgorrie) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:344]: It works if the currency is [in the API](https://product-catalog.guardianapis.com/product-catalog.json) - for example
- `/au/checkout?product=TierThree&ratePlan=RestOfWorldAnnual`
- `/au` converts to `AUD`
- `AUD` is not available on `TierThree.ratePlans.RestOfWorldAnnual.pricing`
- We would error

The way we fork between `us` and `int` is by using different rate plans. I wonder if we actually need to have products only available in certain areas... something to take outside of this PR - great spot.

PR #6213, opened 2024-08-08
Make payment api switches optional

andrewHEguardian:

Making payment api switches optional, after doing the same for the frontend in #6209 

[**Trello Card**](https://trello.com/c/1PkjvuVX/971-deleting-a-switch-in-rrcp-takes-the-whole-site-down)

## Why are you doing this?

We've prevented deleted feature switches crashing the frontend, it makes sense to do this for the payment-api too.

## How to test

Tests pass, will deploy to `CODE` and test a payment

### DIFF
--- a/support-payment-api/src/main/scala/backend/AmazonPayBackend.scala
+++ b/support-payment-api/src/main/scala/backend/AmazonPayBackend.scala
@@ -43,9 +43,7 @@ class AmazonPayBackend(
     with PaymentBackend {
 
   private def amazonPayEnabled =
-    switchService.allSwitches.map(switch =>
-      switch.oneOffPaymentMethods.exists(s => s.switches.amazonPay.exists(_.state.isOn)),
-    )
+    switchService.allSwitches.map(_.oneOffPaymentMethods.exists(_.switches.amazonPay.exists(_.state.isOn)))
 
   def makePayment(
       amazonPayRequest: AmazonPayRequest,
--- a/support-payment-api/src/main/scala/backend/AmazonPayBackend.scala
+++ b/support-payment-api/src/main/scala/backend/AmazonPayBackend.scala
@@ -43,7 +43,9 @@ class AmazonPayBackend(
     with PaymentBackend {
 
   private def amazonPayEnabled =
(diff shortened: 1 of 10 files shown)


support-payment-api/src/main/scala/services/SwitchService.scala
```diff
@@ -46,21 +46,21 @@ case class FeatureSwitches(switches: FeatureSwitchesTypes)
 case class SwitchDetails(state: SwitchState)
 
 case class RecaptchaSwitchTypes(
-    enableRecaptchaBackend: SwitchDetails,
-    enableRecaptchaFrontend: SwitchDetails,
+    enableRecaptchaBackend: Option[SwitchDetails],
```
andrewHEguardian, line 49: I toyed with the idea of trying to do what support-frontend does and flattening it from `{ name: { state: 'On' }}` to just `{ name: 'On' }` but it seemed like too much change

support-payment-api/src/main/scala/backend/AmazonPayBackend.scala
```diff
@@ -43,7 +43,9 @@ class AmazonPayBackend(
     with PaymentBackend {
 
   private def amazonPayEnabled =
-    switchService.allSwitches.map(switch => switch.oneOffPaymentMethods.exists(s => s.switches.amazonPay.state.isOn))
+    switchService.allSwitches.map(switch =>
+      switch.oneOffPaymentMethods.exists(s => s.switches.amazonPay.exists(_.state.isOn)),
+    )
```
rupertbates, line 48:
```suggestion
 switchService.allSwitches.map(_.oneOffPaymentMethods.exists(_.switches.amazonPay.exists(_.state.isOn)))
```
Very minor, maybe this is slightly easier to parse?

support-payment-api/src/main/scala/services/SwitchService.scala
rupertbates: Yeah the flattened version is a lot easier to work with but maybe not worth it.

rupertbates (approved):
Looks good to me

support-payment-api/src/main/scala/services/SwitchService.scala
andrewHEguardian: Was also slightly beyond me in Scala 😀

PR #6217, opened 2024-08-08
fix: set StripeExpress payment postcode and state

jamesgorrie:
Sets a the postcode and state of a person from the [Stripe billingAddress](https://docs.stripe.com/js/custom_checkout/session_object#custom_checkout_session_object-savedPaymentMethods-billingDetails-address) to ensure we pass the state validation.

This has a bug where if a person is accessing the site from `/us/checkout` but has a e.g. UK address and thus no state, we cannot take a payment as state is required.

https://trello.com/c/ZD2BbAod/932-supportworkers-alarm-promocode-country-not-aligning-with-applepay-googlepay-country


GHaberis (approved):
Looks good 👍

andrewHEguardian (approved):
🇺🇸 👍

PR #6219, opened 2024-08-09
Handle Canary Islands country code

tomrf1:
We had an in-app purchase acquisition event fail because the country code, `IC`, is not recognised.
There isn't a valid 3-character code for it, so there's a small change to catch `Locale` lookup failures


support-internationalisation/src/main/scala/com/gu/i18n/CountryGroup.scala
```diff
@@ -335,9 +337,12 @@ object CountryGroup {
 
   val countriesByISO2: Map[String, Country] = countries.map { c => c.alpha2 -> c }.toMap
 
-  val countriesByISO3 = countries.map { country =>
-    val locale = new Locale("", country.alpha2)
-    locale.getISO3Country.toUpperCase -> country
+  val countriesByISO3 = countries.flatMap { country =>
+    Try {
+      // java's locale library may throw an exception if no locale is found
```
johnduffell, line 342: small comment update based on the javadoc - https://docs.oracle.com/javase/8/docs/api/java/util/Locale.html#getISO3Country--
```suggestion
      // getISO3Country will throw a MissingResourceException if the three-letter country abbreviation is not available for this locale
```

As a separate point, (This won't break things), but I noticed it can return empty string(!) in some situations.  This won't affect us apart from leaving junk in the map, which probably wouldn't affect us regardless, but we could always be double safe and filter blanks out!
https://docs.oracle.com/javase/8/docs/api/java/util/Locale.html#getISO3Country--:~:text=If%20the%20locale%20doesn%27t%20specify%20a%20country%2C%20this%20will%20be%20the%20empty%20string.

johnduffell (approved):
good, not too invasive 👍

PR #6221, opened 2024-08-09
feat: use payment abstraction in checkout

jamesgorrie:
Whilst trying to get the right abstraction to allow for both
- `Contribution`: a product, no price, a voluntary contribution, no discount
- `SupporterPlus`: a product, a price, a voluntary contribution, promotional discounts

And having 1 eye on the fact we will have to support
- `OneTime`: not a product, no price, a voluntary contribution, no discounts

It felt that could be abstracted to:
- `originalAmount` the person would pay without a discount or contribution
- `discountedAmount` the amount a person would pay with a discount applied
- `contribution` and amount extra to the `originalAmount`
- `finalAmount` the amount a person will pay

This PR has the naming in such a way that we transform product prices (what something costs) => payment (what someone pays).

While this isn't anything new per say - I feel the naming is a little clearer and should be easier to get `OneTime` payments to fit this abstraction.

We could also write a test for the `product` => `payment` abstraction. 

This also enables `ExpressCheckout` for `SupporterPlus`

### DIFF
--- a/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
@@ -335,25 +335,58 @@ export function Checkout({ geoId, appConfig }: Props) {
 		return <div>Rate plan not found</div>;
 	}
 
-	/** Get and validate price */
-	let price: number | undefined;
+	/**
+	 * Get and validate the amount
+	 *
+	 * For products the amount is based on
+	 * - the product price in the catalog
+	 * - any promotions applied
+	 * - any contributions made
+	 */
+
+	/**
+	 * - `originalAmount` the amount pre any discounts or contributions
+	 * - `discountredAmount` the amount with a discountApplied
+	 * - `finalAmount` is the amount a person will pay
+	 */
+	let amount: {
+		originalAmount: number;
+		discountedAmount?: number;
+		finalAmount: number;
+	};
+
+	const contributionParam = searchParams.get('contribution');
+	const contributionAmount = contributionParam
+		? parseInt(contributionParam, 10)
+		: undefined;
+
+	let promotion;
 	if (productKey === 'Contribution') {
-		const priceParam = searchParams.get('price');
-		price = priceParam ? parseInt(priceParam, 10) : undefined;
+		/**
+		 * Contributions are dynamic amounts, often selected from the `amounts` from RRCP
+		 * @see https://support.gutools.co.uk/amounts
+		 */
+		if (!contributionAmount) {
+			return <div>Contribution not specified</div>;
+		}
+
+		amount = {
+			originalAmount: contributionAmount,
+			finalAmount: contributionAmount,
+		};
 	} else {
-		price =
+		const productPrice =
 			currencyKey in ratePlan.pricing
 				? ratePlan.pricing[currencyKey]
 				: undefined;
-	}
-	if (!price) {
-		return <div>Price not found</div>;
-	}
 
-	/** Get any promotions */
-	const productPrices = appConfig.productPrices;
-	let promotion;
-	if (productPrices) {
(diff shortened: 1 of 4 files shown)


support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -335,25 +335,58 @@ export function Checkout({ geoId, appConfig }: Props) {
...
+	 * For products the amount is based on
+	 * - the product price in the catalog
+	 * - any promotions applied
+	 * - any contributions made
+	 */
+
+	/**
+	 * - `originalAmount` the amount pre any discounts or contributions
+	 * - `discountredAmount` the amount with a discountApplied
+	 * - `finalAmount` is the amount a person will pay
```
Comment (andrewHEguardian) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:350]: When would the final amount be different to the discounted amount? I could see this happening with supporter plus by adding a contribution, but it doesn't seem like a useful use case

support-frontend/assets/pages/supporter-plus-landing/twoStepPages/threeTierLanding.tsx
```diff
@@ -455,7 +455,7 @@ export function ThreeTierLanding(): JSX.Element {
 	const tier1GenericCheckoutUrlParams = new URLSearchParams({
 		product: 'Contribution',
 		ratePlan: selectedContributionRatePlan,
-		price: recurringAmount.toString(),
+		contribution: recurringAmount.toString(),
```
paul-daniel-dempsey, line 458: Does the equivalent `price` on Tier2 Params need to be replaced with a zero `contribution`?

andrewHEguardian (approved):
Feels like a good abstraction! 💡 Just one bit of confusion about discounts

paul-daniel-dempsey (approved):
This change makes amounts easier to understand for sure. Maybe change to Tier2 Params on Landing page?

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
Comment (jamesgorrie) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:350]: The case you explain above is 1 case. It's also to differentiate between the two. If discountedAmount exists, you know a discount has been applied.

I was hoping the naming would capture a universal flow of how a price is calculated and can be used in all cases. 


e.g. `original` - `discounted` + `contribution` = `final` and then we can use those values in different places.

support-frontend/assets/pages/supporter-plus-landing/twoStepPages/threeTierLanding.tsx
jamesgorrie: `undefined` should mean there is no contribution. We shouldn't have to send a price across to `SupportPlus` as that value is derived from the product API.

jamesgorrie:
> Feels like a good abstraction! 💡 Just one bit of confusion about discounts

@andrewHEguardian let me know if the explanation to this makes sense. The idea for this was to be a useful abstraction - so if it's still a little unclear, we could probably try refine it.

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
Comment (andrewHEguardian) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:350]: This does make sense. But could we pass around the amount of the discount instead? i.e. the `discounted` in your equation

Comment (jamesgorrie) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:350]: Do we mean removing `final` and just use `discounted`? You could do this, but we lose the part of the model that indicates that a discount has been applied e.g. `discounted = undefined`. You could check that `original === final` repeatedly - but it felt as though `discountedPrice === undefined` was more explicit.

Part of this is trying to remove as many inferences from the data as I can, while allowing you to render the different types of UI.

PR #6222, opened 2024-08-12
Generic Checkout : Tier2 and Tier3 Promotions Supported

paul-daniel-dempsey:

FulfilmentOptions change dependent on ProductType ->
S+/Contribution :  `NoFulFilmentOption`
TierThree :  `Domestic` / `RestOfWorld`

## How to test

https://support.thegulocal.com/us/checkout?product=SupporterPlus&ratePlan=Monthly&price=15&promoCode=LX4FFRA8

https://support.thegulocal.com/uk/checkout?product=TierThree&ratePlan=DomesticMonthly&promoCode=TIER3_UK_MONTHLY

## Screenshots
|S+|Tier3|
|-----|-----|
|image|image|

### DIFF
--- a/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
@@ -96,6 +96,7 @@ import {
 	productCatalog,
 	productCatalogDescription,
 } from 'helpers/productCatalog';
+import type { FulfilmentOptions } from 'helpers/productPrice/fulfilmentOptions';
 import { NoFulfilmentOptions } from 'helpers/productPrice/fulfilmentOptions';
 import { NoProductOptions } from 'helpers/productPrice/productOptions';
 import type { Promotion } from 'helpers/productPrice/promotions';
@@ -306,7 +307,7 @@ type Props = {
 	appConfig: AppConfig;
 };
 export function Checkout({ geoId, appConfig }: Props) {
-	const { currencyKey } = getGeoIdConfig(geoId);
+	const { currencyKey, countryGroupId } = getGeoIdConfig(geoId);
 	const searchParams = new URLSearchParams(window.location.search);
 
 	/** Get and validate product */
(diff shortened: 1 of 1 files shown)


jamesgorrie (approved):

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -365,15 +366,27 @@ export function Checkout({ geoId, appConfig }: Props) {
 				? 'Monthly'
 				: 'Annual';
 
+		let fulfilmentOption: FulfilmentOptions;
```
Comment (rupertbates) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:369]: I would break the setter out into a separate function so that this can be a const. eg.
```typescript
const fulfilmentOptions = getFulfilmentOptions(productKey);
```

Comment (jamesgorrie) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:369]: Completely agree that mutable state can cause issues.

There is a little tension here in trying not to create early / unnecessary abstractions that might lead to coupling i.e. a `function` that could be used elsewhere. Defining it and not exporting it would work too (which is what I think the suggestion is here).

It's a pattern that's used across JS codebases - as the `switch` / `case` syntax is a little lacking in the sense that you can't run it and assign it to a value as and when. The convention is to not mutate the value outside of the `switch` / `case`.

You can actually run anonymous `switch` statements, but they do look ugly
```ts
const productKey = '...';
const fulfilmentOptions = (() => {
  switch (productKey) {
    case 'SupporterPlus':
    case 'Contribution':
      return 'NoFulfilmentOptions';
    case 'TierThree':
      return countryGroupId === 'International' ? 'RestOfWorld' : 'Domestic';
    default:
      // ToDo: define for every product here
      return 'NoFulfilmentOptions';
  }
})()
```

[Interesting read here](https://github.com/tc39/proposal-pattern-matching) on a better match proposal.

Comment (andrewHEguardian) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:369]: Super interesting discussion, I'd suggest let's merge without it and come back to this - because it's a pattern throughout this checkout

Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx]: Have applied Rupert's immutable suggestion/change (private function) so we can get the merge in. As suggested, perhaps we can examine this pattern more closely afterwards as we are running out of time before tomorrow.

PR #6232, opened 2024-08-14
DRAFT

AndreaDiotallevi:
DRAFT


support-models/src/test/scala/com/gu/support/zuora/api/SerialisationSpec.scala
```diff
@@ -21,14 +21,6 @@ class SerialisationSpec extends AsyncFlatSpec with SerialisationTestHelpers with
     val json = account().asJson
     (json \\ "Currency").head.asString should be(Some("GBP"))
     (json \\ "PaymentGateway").head.asString should be(Some("Stripe Gateway 1"))
-    (json \\ "InvoiceTemplateId").head.asString should be(Some("defaultInvoiceTemplateId"))
-  }
-
-  "AUAccount" should "serialise to correct json" in {
```
rupertbates, line 27: Is it worth retaining this test for the other fields that are being checked?

AndreaDiotallevi: Good point!
I have restored the AUS test.

rupertbates (approved):
This is great, removes a load of code and improves the system 👍

PR #6234, opened 2024-08-14
Fix for CTA showing amount when discount is 100%

LAKSHMIRPILLAI:


support-frontend/assets/components/paymentButton/defaultPaymentButtonContainer.tsx
```diff
@@ -64,6 +64,9 @@ export function DefaultPaymentButtonContainer({
 			  )
 			: getUserSelectedAmount(state),
 	);
+	const { countryId } = useContributionsSelector(
+		(state) => state.common.internationalisation,
+	);
 	const promotion = isSupporterPlus
 		? useContributionsSelector((state) =>
 				getPromotion(
```
LAKSHMIRPILLAI, line 72: On investigation, it was found that somehow the countryId  getting passed to the code here was undefined (due to where it was specified initially), as a result the promotion would always be received as undefined.This in turn make the amount to always fall to the selectedAmount.

jamesgorrie: You could move this into [this line](https://github.com/guardian/support-frontend/pull/6234/files#diff-33b4335cb610a3a6f42c1176b50cd30036f5f65b35559f19495dbcdae41e0f44R54).

I'd really like to understand how we're not getting a `Variable 'countryId' is used before being assigned.ts(2454)`. [I can replicate it being valid](https://www.typescriptlang.org/play/?#code/MYewdgzgLgBAZnGBeGAKOAuNBKZA+GMAVwFsAjAUwCdckCBvAKAEgqKoiqx5VtGBfRqEiwA5snhxUvfDAAWfYdHkSAjAG5GQ8BBAAbCgDo9IUalHYgA) - but it does error.

jamesgorrie (approved):

PR #6235, opened 2024-08-15
fix: send correct amount to /create endpoint for `SupporterPlus`

jamesgorrie:
Sends the correct `amount` to the `/create` endpoint for `SupporterPlus`.

The logic is described in a comment in the code.

Currently `promoCodes` don't work on the generic checkout for `SupporterPlus` because of this.

[This will get us closer to releasing the test for generic checkout for S+](https://github.com/guardian/support-frontend/pull/6236).


GHaberis (approved):
Looks good 👍

PR #6236, opened 2024-08-15
feat: Enable generic checkout 50% test for SupporterPlus

jamesgorrie:
Enabled the 50% test for sending people to the generic checkout for SupporterPlus from Tier 2 of the landing page.

This also adds some test to ensure that we're working and that we don't add any regressions.

Screenshot 2024-08-20 at 11 50 58

~Currently the tests do not pass - so I will be attaching some PRs until they do.~

### DIFF
--- a/support-frontend/assets/pages/supporter-plus-landing/twoStepPages/threeTierLanding.tsx
+++ b/support-frontend/assets/pages/supporter-plus-landing/twoStepPages/threeTierLanding.tsx
@@ -392,7 +392,7 @@ export function ThreeTierLanding(): JSX.Element {
 			/**
 			 * Only Testing CardTier1 wth checkout
 			 */
-			if ((useGenericCheckout && cardTier === 1) || tier2UseGenericCheckout) {
+			if (useGenericCheckout) {
 				/**
 				 * Generic Checkout is not defined in supporterPlusRouter
 				 */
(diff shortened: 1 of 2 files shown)


support-frontend/assets/pages/supporter-plus-landing/twoStepPages/threeTierLanding.tsx
```diff
@@ -392,7 +392,7 @@ export function ThreeTierLanding(): JSX.Element {
 			/**
 			 * Only Testing CardTier1 wth checkout
 			 */
-			if ((useGenericCheckout && cardTier === 1) || tier2UseGenericCheckout) {
+			if (useGenericCheckout) {
```
jamesgorrie, line 395: [`cardTier === 3` is handled above](https://github.com/guardian/support-frontend/pull/6236/files#diff-9e9d19a3470db4ff48cd4c73865be05d061001afc0751360451d6d509f390e9bR363), thus this is only `2 || 3`. 

I would prefer if this was explicit incase we added a tier (unlikely), but eslint complains with a [`no-unnecessary-condition`](https://typescript-eslint.io/rules/no-unnecessary-condition/) error.

andrewHEguardian (approved):
Awesome! 🚀

PR #6237, opened 2024-08-15
Use qs params to power thank-you page

jamesgorrie:
Brings the way we get data into the thank you page more inline with the checkout page (as suggested by @rupertbates in a PR I can't find).

It is a copy/paste job (and labelled as such) - I'm not sure we're at the stage of abstracting it just yet. As we get into the world of working on the one-time checkout I think we will start to find good abstractions.

Fixes some bugs raised in https://github.com/guardian/support-frontend/pull/6236

### DIFF
--- a/support-frontend/assets/pages/[countryGroupId]/thank-you.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/thank-you.tsx
@@ -9,18 +9,24 @@ import {
 } from '@guardian/source/react-components';
 import { FooterWithContents } from '@guardian/source-development-kitchen/react-components';
 import type { InferInput } from 'valibot';
-import { number, object, optional, picklist, safeParse, string } from 'valibot';
+import { object, picklist, safeParse, string } from 'valibot';
 import { Header } from 'components/headers/simpleHeader/simpleHeader';
 import { PageScaffold } from 'components/page/pageScaffold';
 import type { ThankYouModuleType } from 'components/thankYou/thankYouModule';
 import ThankYouModule from 'components/thankYou/thankYouModule';
 import { getThankYouModuleData } from 'components/thankYou/thankYouModuleData';
+import type { AppConfig } from 'helpers/globalsAndSwitches/window';
 import CountryHelper from 'helpers/internationalisation/classes/country';
+import type { ProductKey } from 'helpers/productCatalog';
 import {
 	filterBenefitByRegion,
+	isProductKey,
+	productCatalog,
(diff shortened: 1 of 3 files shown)


support-frontend/assets/pages/[countryGroupId]/thank-you.tsx
```diff
@@ -224,15 +377,14 @@ export function ThankYou({ geoId }: Props) {
...
+							amount={payment.finalAmount}
 							contributionType={contributionType}
 							amountIsAboveThreshold={isSupporterPlus}
 							isTier3={isTier3}
 							isOneOffPayPal={isOneOffPayPal}
 							showDirectDebitMessage={order.paymentMethod === 'DirectDebit'}
 							currency={currencyKey}
-							// TODO - generic checkout support promotions
-							promotion={undefined}
+							promotion={promotion}
```
Comment (jamesgorrie) [inline support-frontend/assets/pages/[countryGroupId]/thank-you.tsx:387]: This is actually the main bug fixed.

support-frontend/assets/pages/[countryGroupId]/thank-you.tsx
```diff
@@ -152,16 +305,16 @@ export function ThankYou({ geoId }: Props) {
 		return <div>Unable to find contribution type {contributionType}</div>;
 	}
 
-	const isOneOff = order.product === 'Contribution';
+	const isOneOff = productKey === 'Contribution';
```
Comment (GHaberis) [inline support-frontend/assets/pages/[countryGroupId]/thank-you.tsx:308]: Is this correct? I thought `productKey === 'Contribution'` for recurring contributions? Do we need this `isOneOff` logic in the generic TY page at that moment?

Comment (jamesgorrie) [reply support-frontend/assets/pages/[countryGroupId]/thank-you.tsx:308]: This is a good point - I'll do that as a secondary fix to make sure it's not conflated.

Comment (GHaberis) [reply support-frontend/assets/pages/[countryGroupId]/thank-you.tsx:308]: Yeah I was worried it may introduce an unexpected behaviour for users paying via paypal when taking out a recurring contribution, it's just used for the `isOneOffPayPal` prop passed to `ThankYouHeader` and it looks like it just alters the header copy - I don't think this PR is introducing a change to that behaviour though. So yes probably something we can look at as a secondary fix.

Comment (jamesgorrie) [reply support-frontend/assets/pages/[countryGroupId]/thank-you.tsx:308]: Yea, so I think we should merge this in and then check current functionality and see we haven't coupled this to anything where what we actually mean is `isContribution`.

GHaberis (approved):
Looks good 👍

PR #6238, opened 2024-08-15
fix: restrict delivery countries to countryGroup

jamesgorrie:
Redirects a person to the correct internationalisation ID (`uk`, `au`, etc) if the country they select on the checkout doesn't match where they currently are. 

The reason is described in the comment below https://github.com/guardian/support-frontend/blob/353856042d71e9b8479dd56f2b894ab7c233972e/support-frontend/assets/helpers/productCatalog.tsx?plain=1#L349-L364


rupertbates (dismissed):
Let's get Helene to look at this idea

jamesgorrie:
### Unused alternatives

Screenshot 2024-09-02 at 12 22 17

Screenshot 2024-08-20 at 15 39 05

support-frontend/assets/components/subscriptionCheckouts/address/addressFields.tsx
```diff
@@ -184,10 +189,47 @@ export function AddressFields({ scope, ...props }: PropTypes): JSX.Element {
...
 				label="Country"
 				value={props.country}
-				onChange={(e) => {
-					const isoCountry = Country.fromString(e.target.value);
-					if (isoCountry) {
-						props.setCountry(isoCountry);
+				onChange={(event) => {
+					const selectedCountry = Country.fromString(event.target.value);
+
+					if (selectedCountry && countryGroupId) {
```
jamesgorrie, line 195: These transforms are a little verbose, but necessary - I am transforming from `CountryGroupId` then looking up the `CountryGroupId` of a `Country` and then redirecting if the `current` and `selected` country groups don't match.

support-frontend/assets/components/subscriptionCheckouts/address/addressFields.tsx
```diff
@@ -184,10 +189,47 @@ export function AddressFields({ scope, ...props }: PropTypes): JSX.Element {
...
+							const { productKey, ratePlanKey } =
+								internationaliseProductAndRatePlan(
+									selectedInternationalisationId,
+									product,
+									ratePlan,
+								);
+							urlSearchParams.set('product', productKey);
+							urlSearchParams.set('ratePlan', ratePlanKey);
+
+							const location = `${redirectPathname}?${urlSearchParams.toString()}`;
```
GHaberis, line 226: Just realised they'll lose the promoCode applied via the current URL when we redirect, hmm need to have a think about that and the best way to handle this.

jamesgorrie: Yea, not 100% sure - currently, for `TierThree` it works as there are default promotions.

We don't have "Groups" of promo codes, although if we gave them all the same name, but only made them valid in certain groups, I think that may work too.

e.g.
- `TIER_THREE_MONTHLY` valid in `uk`
- `TIER_THREE_MONTHLY` valid in `us`
- `TIER_THREE_MONTHLY` valid in `au`
- ...

And then set `TIER_THREE_MONTHLY` as the default promo code - I think this would work.

GHaberis: Nice ok, given the checkout already picks the default promo if it's available regardless of whether the `promoCode` query string is in the URL I don't think this is an issue

support-frontend/assets/helpers/productCatalog.tsx
```diff
@@ -342,3 +345,53 @@ export const productCatalogDescriptionNewBenefits: Record<
...
+	supportInternationalisationId: SupportInternationalisationId,
+	productKey: ProductKey,
+	ratePlanKey: string,
+): { productKey: ProductKey; ratePlanKey: string } {
+	let productKeyToUse = productKey;
+	let ratePlanToUse = ratePlanKey;
+
+	if (productKey === 'SupporterPlus') {
+		if (supportInternationalisationId === 'int') {
+			if (ratePlanKey === 'DomesticAnnual') {
```
andrewHEguardian, line 375: Does there need to be the inverse operation of this?

support-frontend/assets/helpers/productCatalog.tsx
```diff
@@ -342,3 +345,53 @@ export const productCatalogDescriptionNewBenefits: Record<
...
+ * */
+export function internationaliseProductAndRatePlan(
+	supportInternationalisationId: SupportInternationalisationId,
+	productKey: ProductKey,
+	ratePlanKey: string,
+): { productKey: ProductKey; ratePlanKey: string } {
+	let productKeyToUse = productKey;
+	let ratePlanToUse = ratePlanKey;
+
+	if (productKey === 'SupporterPlus') {
```
andrewHEguardian, line 373:
```suggestion
	if (productKey === 'TierThree') {
```
Right?

support-frontend/assets/components/subscriptionCheckouts/address/addressFields.tsx
```diff
@@ -184,10 +189,47 @@ export function AddressFields({ scope, ...props }: PropTypes): JSX.Element {
...
+							window.location.href = location;
+						}
+					}
+
+					if (selectedCountry) {
```
paul-daniel-dempsey, line 231: I think the page reload overrides this selection. ie selecting `Taiwan` from within Uk region causes path to be updated to `int` and the resulting reload replaces Taiwan with `India`.

support-frontend/assets/helpers/productCatalog.tsx
paul-daniel-dempsey: Think this is needed because f you start in region `int` then change to a country like France you get `price not found in product catalog`

andrewHEguardian: It looks like changing from France to int is missing as far as I can tell?

support-frontend/assets/helpers/productCatalog.tsx
jamesgorrie: Indeed - updated.

andrewHEguardian (approved):
Great solution

support-frontend/assets/helpers/productCatalog.tsx
jamesgorrie: Nice catch, I've added the alternative.

support-frontend/assets/helpers/productCatalog.tsx
jamesgorrie: Indeed!

support-frontend/assets/components/subscriptionCheckouts/address/addressFields.tsx
jamesgorrie: I've added the `country` URL parameter and that's fixed it.

GHaberis: Selecting Taiwan will redirect to the `/int` country group checkout, the default country for this is India, if you're checking out out locally (eg. in the UK) on the `/int` checkout it'll set your delivery/billing country to India - if you're in Taiwan and you're on the `/int` checkout it should identify via your geo location that you're in Taiwan the  delivery/billing country will be set to Taiwan.

GHaberis: > Selecting Taiwan will redirect to the `/int` country group checkout, the default country for this is India, if you're checking out out locally (eg. in the UK) on the `/int` checkout it'll set your delivery/billing country to India - if you're in Taiwan and you're on the `/int` checkout it should identify via your geo location that you're in Taiwan the delivery/billing country will be set to Taiwan.

Saying that if you're geo location is Taiwan you should've ended up on the `/int` checkout anyway, forcing the country with the `country` query string param as @jamesgorrie has done seems like a safe way of ensuring the country is correctly set in the reloaded form,

GHaberis (commented):
Looks good 👍

PR #6240, opened 2024-08-15
Delete Supporter Plus Only variant of the old 2-tier checkout

GHaberis:

This variant of the 2-step checkout is no longer in use, so this is a tidy-up PR to remove it from the codebase.

I'm documenting any components removed here: https://github.com/guardian/support-frontend/wiki/Archived-Components


jamesgorrie (approved):
Nice - I wonder if we could drop that Wiki entry into storybook keeping all component stuff together?

PR #6242, opened 2024-08-16
Tidy old campaigns code

tomrf1:
Historically we've used the `CampaignSettings` model to customise parts of the support site for campaigns.
It's not currently in use, and much of the code is very old and not relevant.
We're going to revive the concept for the upcoming US end-of-year moment (and beyond), but first I want to clean up legacy code.


GHaberis (approved):
Looks good, good to tidy this up 🧹

PR #6245, opened 2024-08-19
fix: remove script tag from inside script tag

jamesgorrie:
Currently this renders
```html
<script>
  <script></script>
</script>
```

which is no good. This PR unnests that.


andrewHEguardian (approved):
✓

PR #6246, opened 2024-08-19
Send page view ID to Quantum Metric

GHaberis:

Send Ophan Page View ID to QM. This will help us debug alarms as we can query QM with Page View ID.

[**Trello Card**](https://trello.com/c/ne53rvdo/1020-send-page-view-id-to-qm)

### DIFF
--- a/support-frontend/assets/helpers/tracking/quantumMetric.ts
+++ b/support-frontend/assets/helpers/tracking/quantumMetric.ts
@@ -1,4 +1,5 @@
 import { loadScript } from '@guardian/libs';
+import { viewId } from 'ophan';
 import type { Participations } from 'helpers/abTests/abtest';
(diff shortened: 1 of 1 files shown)


support-frontend/assets/helpers/tracking/quantumMetric.ts
```diff
@@ -19,6 +20,8 @@ import {
 
 type SendEventTestParticipationId = 30;
 
+type SendEventPageViewId = 181;
```
andrewHEguardian, line 23: what is this number?

andrewHEguardian (approved):

support-frontend/assets/helpers/tracking/quantumMetric.ts
GHaberis: It's an event ID assigned by QM

PR #6067, opened 2024-06-05
Prototype: Lazy Load 3-Tier Routes

GHaberis:
10% time prototype

Testing React's `Lazy` / `Suspense`  exports: https://react.dev/reference/react/lazy#suspense-for-code-splitting

I'm interested to see whether we could loading component’s code until they're rendered for the first time, eg. Checkout or Thankyou page. Currently everything's loaded in one bundle. I wanted to see whether deferring loading of components has any descernable improvement to page speed performance on the 3-tier landing page.

PR #6247, opened 2024-08-19
Create empty one time checkout and abstract common checkout scaffolding

andrewHEguardian:

Creating an initial route and page component for a one time checkout. Extract out the container styling that it shares with the generic checkout.

[**Trello Card**](https://trello.com/c/4xsbNGb4/1008-do-we-use-the-generic-checkout-or-have-a-separate-one-time-checkout-spike)

## Why are you doing this?

After some discussions and pairing we've decided it's best to implement the logic of the one time checkout separately to the generic checkout - they do not share enough to make it worthwhile to do in the same place. However we want to copy and abstract the visual elements they share.

## How to test

Go to https://support.thegulocal.com/uk/one-time-checkout?product=SupporterPlus, the product in the query param is still TODO

## Screenshots

image


jamesgorrie:
> Filename	Size	Change
> `./public/compiled-assets/javascripts/[countryGroupId]/router.js`	252 kB	+4.92 kB (+2%)

This is something we'll need to keep an eye on @andrewHEguardian. It's a major pitfall with our current pattern that everytime we add a new page, we add _a lot_ to the bundle size.

I know @GHaberis [did some prototyping with lazy loading](https://github.com/guardian/support-frontend/pull/6067), maybe we can use that here?

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
Comment (jamesgorrie) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:1]: Minor - but losing the crazy indenting here has made this file a lot nicer.

jamesgorrie (approved):
This looks like a great start - I have a question around the naming of these components. We can have the char post-PR or as part of it?

support-frontend/assets/pages/[countryGroupId]/components/checkoutScaffold.tsx
```diff
@@ -0,0 +1,72 @@
...
+	color: ${palette.neutral[7]};
+	${textSans17};
+	padding-top: ${space[2]}px;
+`;
+
+type CheckoutScaffoldProps = {
+	children: ReactNode;
+};
+
+export function CheckoutScaffold({ children }: CheckoutScaffoldProps) {
```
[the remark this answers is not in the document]
Comment (andrewHEguardian) [reply support-frontend/assets/pages/[countryGroupId]/components/checkoutScaffold.tsx:46]: Yep it's not a great name, purely there because I copied it from the `PageScaffold` component. Layout sounds more sensible

support-frontend/assets/pages/[countryGroupId]/components/checkoutScaffold.tsx
Comment (jamesgorrie) [inline support-frontend/assets/pages/[countryGroupId]/components/checkoutScaffold.tsx:1]: Could we rename the file to `checkoutLayout`?

Comment (andrewHEguardian) [reply support-frontend/assets/pages/[countryGroupId]/components/checkoutScaffold.tsx:1]: whoops good spot

PR #6251, opened 2024-08-20
Validate delivery state field on Tier 3 generic checkout

GHaberis:

Currently a user can submit the generic checkout without selecting a delivery state from the state dropdown list when it is shown (US, AUS and CA) checkouts, this results in a server side error as delivery state is required.

On investigation I found we're using this [AddressFields](https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/components/subscriptionCheckouts/address/addressFields.tsx) component, which in turn uses the [Select](https://github.com/guardian/csnx/blob/main/libs/%40guardian/source/src/react-components/select/Select.tsx) component from `@guardian/source`. Unfortunately the Select component does not add a required attribute to the select element rendered, despite taking an optional `boolean` prop, unlike the [Text Input](https://github.com/guardian/csnx/blob/main/libs/%40guardian/source/src/react-components/text-input/TextInput.tsx#L181) component from `@guardian/source` for example. 

The ideal solution would be the Select component in `@guardian/source` adds a required attribute. This is something I can look at. In the meantime I have manually added the required attribute in the [AddressFields](https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/components/subscriptionCheckouts/address/addressFields.tsx) component. Having this attribute present means the browser's native HTML form validation prevents form submission if this field hasn't been selected.

This isn't an issue on the standard Guardian Weekly checkout as we don't use the browser's native HTML form validation, we make use of the [error handling](https://github.com/guardian/csnx/blob/main/libs/%40guardian/source/src/react-components/select/Select.tsx#L118) built into the `@guardian/source` component. 

We're using a mix of native browser validation and that provided by `@guardian/source` components right now, so we may want to look at making it consistent. 

## Screenshots

Native HTML form validation doing it's thing...

https://github.com/user-attachments/assets/3c076d14-c9fe-407f-9a28-0d76fd9fdab3


support-frontend/assets/components/subscriptionCheckouts/address/addressFields.tsx
```diff
@@ -48,6 +48,13 @@ const marginBottom = css`
 	margin-bottom: ${space[6]}px;
 `;
 
+const selectStateStyles = css`
```
GHaberis, line 51: Without this the select field always had a red border as soon as you rendered the page, I couldn't fathom why it was in an invalid state having not submitted the form.

andrewHEguardian (approved):

GHaberis:
Replaced by https://github.com/guardian/support-frontend/pull/6255

PR #6253, opened 2024-08-22
One Time Checkout - Create 'Your details' section of form

andrewHEguardian:

Creating the personal details section of new one time checkout form. Re-using styles from the generic checkout by creating [reusable components](https://emotion.sh/docs/best-practices#method-2-share-styles-via-component-reuse). 

[**Trello Card**](https://trello.com/c/mTo10TC1/1028-one-time-checkout-your-details)

## Screenshots
image


support-frontend/assets/pages/[countryGroupId]/components/formSection.tsx
```diff
@@ -0,0 +1,25 @@
...
+			margin-top: ${space[4]}px;
+		}
+	}
+`;
+
+type FormSectionProps = {
+	children: ReactNode;
+};
+
+export function FormSection({ children }: FormSectionProps) {
```
Comment (andrewHEguardian) [inline support-frontend/assets/pages/[countryGroupId]/components/formSection.tsx:23]: Feels like a potentially unnecessary abstraction but it's quite nice to [colocate component with the styles](https://emotion.sh/docs/best-practices#colocate-styles-with-components)

GHaberis (approved):
Looks good to me 👍

support-frontend/assets/pages/[countryGroupId]/components/formSection.tsx
Comment (jamesgorrie) [reply support-frontend/assets/pages/[countryGroupId]/components/formSection.tsx:23]: If we're going to be having more than 1 for, this makes sense. I might consider having a `form.tsx` that just contains it all - but that's a stylistic thing.

[I quite like the pattern they have here](https://emotion.sh/docs/best-practices#consider-defining-styles-outside-your-components), which includes co-location of components that are coupled (a `FormSection` would never live outside a form) and keeping styles related to their component.

support-frontend/assets/pages/[countryGroupId]/components/legend.tsx
```diff
@@ -0,0 +1,23 @@
...
+	justify-content: space-between;
+`;
+
+type LegendProps = {
+	children: ReactNode;
+};
+
+export function Legend({ children }: LegendProps) {
+	return <legend css={legend}>{children}</legend>;
+}
```
Comment (jamesgorrie) [inline support-frontend/assets/pages/[countryGroupId]/components/legend.tsx:23]: As per the above - a `Legend` would never live outside of a `form`, so I wonder if it's worth containing these styles together.

support-frontend/assets/pages/[countryGroupId]/validation.ts
```diff
@@ -0,0 +1,21 @@
...
+		// 2. setCustomValidity to " " which avoids the browser's default message
+		currentTarget.setCustomValidity(' ');
+	}
+}
+
+/**
+ * This uses a Unicode character class escape
+ * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Unicode_character_class_escape
+ */
+export const doesNotContainEmojiPattern = '^[^\\p{Emoji_Presentation}]+$';
```
Comment (jamesgorrie) [inline support-frontend/assets/pages/[countryGroupId]/validation.ts:21]: Broken record - should this all live in 1 form component?

support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx
```diff
@@ -114,7 +142,91 @@ function OneTimeCheckoutComponent({ geoId }: Props) {
...
+				method="POST"
+				onSubmit={() => {
+					/* ToDo */
+				}}
+			>
+				<Box cssOverrides={shorterBoxMargin}>
+					<BoxContents>
+						<FormSection>
+							<Legend>1. Your details</Legend>
+							<div>
```
Comment (jamesgorrie) [inline support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx:157]: What's this `div` do?

jamesgorrie (approved):
Looks good - main question is would it be useful to couple coupled components e.g. in a `form.tsx` to maintain colocation of related code?

support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx
Comment (andrewHEguardian) [reply support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx:157]: Can't see any reason for it - copy paste from `checkout.tsx`. Let's get rid

support-frontend/assets/pages/[countryGroupId]/components/legend.tsx
Comment (andrewHEguardian) [reply support-frontend/assets/pages/[countryGroupId]/components/legend.tsx:23]: Ideally I would have had one component that took the legend children and the rest of the fieldset's children as props. But alas we re-use the legend styling within a differently styled fieldset

support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx
Comment (andrewHEguardian) [reply support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx:157]: image

The div is for styling purposes - the style in `formSection` is adding top margin to the not first in type, which the `div` is. Maybe we can do this in a more obvious way but let's come back to it

support-frontend/assets/pages/[countryGroupId]/components/formSection.tsx
Comment (andrewHEguardian) [reply support-frontend/assets/pages/[countryGroupId]/components/formSection.tsx:23]: Could be a good option - there are some other shared bits that don't feel deserving of an entire file. Let's come back to this in the next PR

PR #6254, opened 2024-08-22
Allow unknown country in acquisition rows

tomrf1:
Currently the `AcquisitionDataRow` model requires a `country`. This means we cannot write an acquisition row to BigQuery if the country code is not one of the expected countries.
We've had an acquisition with country code of `IC`, which is not supported currently, and would be difficult to support (java locale library doesn't have a 3 character code for it).
Instead, it's better to write a `null` country value for an event like this.


charleycampbell (approved):
I really like this solution! 👍

johnduffell:
apparently they are in CLDR but not the 3 letter locale system, so perhaps we are using an old standard ?
https://docs.oracle.com/javase/8/docs/technotes/guides/intl/enhancements.8.html#cldr

tomrf1:
Closing in favour of https://github.com/guardian/support-frontend/pull/6219/files
with a small adjustment we can avoid the need for the 3 letter code!

PR #6255, opened 2024-08-22
Use Source Components validation error message on generic checkout addresses

GHaberis:

this is a first pass at using the validation error messaging builid into `@guardian/source` components  on the generic checkouts address fields, this is an alternative approach to using native browser validation error messaging, implemented here: https://github.com/guardian/support-frontend/pull/6251

**support-frontend/assets/components/subscriptionCheckouts/address/addressFields.tsx**
- I've set-up an `setErrorsOnInvalid` function that we can attach to our form fields and execute via the `onInvalid` handler. The `onInvalid` event handler  fires when a submittable element has been checked for validity and doesn't satisfy its constraints (see: https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/invalid_event). `onAddressFieldInvalid`  manages adding/removing error message to either the `errors` prop via the `setErrors` prop.
- I've added an `onBlur` handler to the form fields, this revalidates the form fields when a user has entered content and triggers `onInvalid`, if the input is valid any error related to the field will be removed from the `errors` list, and will disappear from the UI.
- I've added some `maxLength` content restrictions to the form fields, which were in place in the non-generic GW checkout, these stop invalid content causing step function failures because the fields content exceeds a limit in Zuora.
- I've added some `maxLength`/`doesNotContainEmojiPattern` content restrictions to the form fields, which were in place in the non-generic GW checkout, these stop invalid content causing step function failures because the fields content exceeds a limit in Zuora or contains invalid characters.
  
## Screenshots

`@guardian/source` component validation doing it's thing...

https://github.com/user-attachments/assets/88ca1ac8-851d-4134-a29b-ece1f2da8f55

### DIFF
--- a/support-frontend/assets/components/subscriptionCheckouts/address/addressFields.tsx
+++ b/support-frontend/assets/components/subscriptionCheckouts/address/addressFields.tsx
@@ -84,48 +84,47 @@ function statesForCountry(country: Option<IsoCountry>): React.ReactNode {
 
 type ValidityStateError = 'valueMissing' | 'patternMismatch';
 
-function getErrorMessage(
-	field: keyof AddressFieldsType,
-	scope: 'delivery' | 'billing',
-	validityStateError: ValidityStateError,
-	country: IsoCountry,
-): string {
-	const patternMismatch = 'Please use only letters, numbers and punctuation.';
-	const errorMessages: Record<string, {
-			[key in ValidityStateError]?: string;
-		}> = {
-		lineOne: {
-			valueMissing: `Please enter a ${scope} address.`,
-			patternMismatch,
-		},
-		lineTwo: {
-			patternMismatch,
-		},
-		city: {
-			valueMissing: `Please enter a ${scope} town/city.`,
-			patternMismatch,
-		},
-		state: {
-			valueMissing:
-				country === 'CA'
-					? `Please select a ${scope} province/territory.`
-					: `Please select a ${scope} state.`,
-			patternMismatch,
-		},
-		postCode: {
-			valueMissing: `Please enter a ${scope} ${
-				country === 'US' ? 'ZIP code' : 'postcode'
-			}`,
-			patternMismatch,
-		},
+export function AddressFields({ scope, ...props }: PropTypes): JSX.Element {
+	const getErrorMessage = (
+		field: keyof AddressFieldsType,
+		validityStateError: ValidityStateError,
+	): string => {
+		const patternMismatch = 'Please use only letters, numbers and punctuation.';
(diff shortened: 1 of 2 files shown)


andrewHEguardian (approved):
The result in the screen record looks really good! 👍  I'm finding the code a little complicated but it seems to do the job

GHaberis:
> The result in the screen record looks really good! 👍 I'm finding the code a little complicated but it seems to do the job

Yep the UX is nice, the code could be improved I'm sure, I'll take sometime today to refine it!

support-frontend/assets/components/subscriptionCheckouts/address/addressFields.tsx
```diff
@@ -75,7 +82,97 @@ function statesForCountry(country: Option<IsoCountry>): React.ReactNode {
...
+				...(possibleValidityStateErrors
+					.map((possibleValidityStateError) => {
+						if (validityState[possibleValidityStateError]) {
+							return {
+								field,
+								message: getErrorMessage(field, possibleValidityStateError),
+							};
+						}
+					})
+					.filter(Boolean) as AddressFormFieldError[]),
```
andrewHEguardian, line 170: What are we filtering out here? Is it empty strings or null/undefined values?

andrewHEguardian: Could we use https://github.com/guardian/csnx/tree/main/libs/@guardian/libs/src/isNonNullable and get rid of the `as` ?

GHaberis: > What are we filtering out here? Is it empty strings or null/undefined values?

Yes, the `.map` which iterates over `const possibleValidityStateErrors: ValidityStateError[] = ['valueMissing','patternMismatch']` can return an array with `undefined` items in it, if the `validityState` doesn't contain a corresponding error. I'm removing all the `undefined` items from that returned array here.

I wasn't happy having to define `possibleValidityStateErrors` tbh, the first thing I attempted was to iterate over `validityState`, which looks like an Object with keys, like below, unfortunately I couldn't iterate over this object, so I introduced `possibleValidityStateErrors` which I could then use to interrogate `validityState`.

Screenshot 2024-08-27 at 12 30 30

GHaberis: > Could we use https://github.com/guardian/csnx/tree/main/libs/@guardian/libs/src/isNonNullable and get rid of the `as` ?

:eyes I'll take a look at this

PR #6257, opened 2024-08-27
refactor: remove geo information from redux on landing page

jamesgorrie:
We would like to remove redux from the landing page as a larger initiative to remove it for simplification of our codebase. 

This starts to make movements in that direction by removing piece by piece as an alternative to something more large scale like https://github.com/guardian/support-frontend/pull/6239

This feels more controllable and should remove the need to have a new route and all the other testing as per the above PR.


andrewHEguardian (approved):
Great stuff 👍

PR #6260, opened 2024-08-27
fix: add oauth callback URL for events domain

jamesgorrie:
This adds the ability to use `support.` _**OR**_ `live.` as a redirect URL for auth.

Currently we use `live.` to serve `/events`. This is because `support.` is considered a in-app-purchase URL from our apps and if we link to it, it's blocked.

An example workflow this supports
* Editor creates a new event
* The event is linked to from the site as `live.theguardian.com/events/{eventId}`
* A person has outdated cookies
* They're re-logged in, and redirected back to live.theguardian.com

- [x] Tests will be added retrospectively as we are on a bit of a deadline.


andrewHEguardian (approved):
👍

PR #6262, opened 2024-08-28
Categorise Playwright tests

jamesgorrie:
This PR creates and categorises some Playwright tests to try and address the speed of the suite, as well as some flakiness issues we've been seeing with third party services timing out e.g. `PayPal`.

The categories we have created are:
- Smoke: tests to be run post-deploy
- Cron: tests that suffer from rate limiting, run on a schedule

Other changes that have been made in this PR:
- moving away from the `GitHub Action => Playwright => Browserstack` pattern to a `GitHub Action => Playwright` model

### DIFF
--- a/.github/workflows/playwright-cron.yml
+++ b/.github/workflows/playwright-cron.yml
@@ -1,0 +1,32 @@
+name: Playwright Cron Tests
+on:
+  workflow_dispatch:  
+  schedule:
+    # every hour, 9-17:00, Mon - Fri
+    - cron: "0 9-17 * * 1-5"
+jobs:
+  test:
+    timeout-minutes: 60
+    runs-on: ubuntu-latest
+    env:
+      PAYPAL_TEST_PASSWORD: ${{ secrets.PAYPAL_TEST_PASSWORD }}
+    steps:
+    - uses: actions/checkout@v4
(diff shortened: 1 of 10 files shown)


.github/workflows/playwright-cron.yml
```diff
@@ -0,0 +1,30 @@
+name: Playwright Tests
+on:
+  workflow_dispatch:  
+  schedule:
+    # every hour, Mon - Fri
+    - cron: "0 * * * 1-5"
```
andrewHEguardian, line 6: How about only 9-5?

andrewHEguardian (approved):
This a very cool idea 💡

.github/workflows/playwright-cron.yml
jamesgorrie: I like this from a "You shouldn't be working outside of hours" perspective - but do you think we should be optimising for compute time? I'm not 100% sure we pay for it (although burning event a tiny amount of power for nothing seems silly).

It's also nice we can call this manually via `workflow_dispatch`. As a secondary PR - I will have a think about how we socialise this stuff.

support-e2e/package.json
jamesgorrie, line 1: This was just linting

PR #6265, opened 2024-08-29
feat: add logging for when we do not get state from stripe

jamesgorrie:
Adds logging so we know how many people are being caught by a bug where we need a state for territories where we require it i.e. US, CA, AU, NZ.


andrewHEguardian (approved):
🗺️

PR #6266, opened 2024-08-29
fix: add consents to event page

jamesgorrie:
Uses the standard `setUpTrackingAndConsents`.
Not 100% sure why we didn't in the first place.

Screenshot 2024-08-29 at 10 50 07

### DIFF
--- a/support-frontend/assets/pages/[countryGroupId]/events/router.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/events/router.tsx
@@ -1,21 +1,10 @@
-import ophan from 'ophan';
 import { createBrowserRouter, RouterProvider } from 'react-router-dom';
-import * as abTest from 'helpers/abTests/abtest';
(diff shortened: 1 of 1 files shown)


andrewHEguardian (approved):
👍

support-frontend/assets/pages/[countryGroupId]/events/router.tsx
```diff
@@ -1,21 +1,10 @@
...
-const countryId = Country.detect();
-const countryGroupId = CountryGroup.detect();
-const participations = abTest.init({
-	countryId,
-	countryGroupId,
-});
-ophan.init();
-trackAbTests(participations);
-
+setUpTrackingAndConsents();
```
Comment (GHaberis) [inline support-frontend/assets/pages/[countryGroupId]/events/router.tsx:7]: This function has a few bits of logic specific to Contributions (such as the call to `getAmountsTestVariant`), this doesn't cause any issues?

PR #6270, opened 2024-08-29
Use data-qm-masking on TY page

GHaberis:

This PR wraps the first name used on the Thank You page in a `<span>` with the attribute `data-qm-masking="blocklist"`, this blocks QM from reading the first name, so it won't be rendered in playbacks.


andrewHEguardian (approved):
****

PR #6271, opened 2024-08-29
feat: add non-deliverable product state select tests

jamesgorrie:
This adds tests to ensure we sure the `StateSelect` for non-deliverable products.

The more important thing I am trying to capture here is the reason why, using comments and tests to explain that. 

This is part of a large set of institutional / team knowledge that it would be good to codify to socialise between ourselves, but also future folks who join the team and fancy changing this at some stage. 

This PR is more for discussion as opposed to a "We should do this".

### DIFF
--- a/support-e2e/tests/smoke/non-delivery-state-selector.test.ts
+++ b/support-e2e/tests/smoke/non-delivery-state-selector.test.ts
@@ -1,0 +1,43 @@
+import { test } from '@playwright/test';
+import { afterEachTasks } from '../utils/afterEachTest';
+import { setTestCookies } from '../utils/cookies';
+
+afterEachTasks(test);
+
+const countryGroupsWithStates = ['us', 'au', 'ca'];
+const nonDeliveryProducts = [
+	{ product: 'SupporterPlus', ratePlan: 'Monthly' },
+	{ product: 'Contribution', ratePlan: 'Monthly', contribution: 10 },
+];
+
+/** We require state for non-deliverable products as we use different taxes within those regions upstream */
+countryGroupsWithStates.forEach((internationalisationId) => {
+	nonDeliveryProducts.forEach(({ product, ratePlan, contribution }) => {
+		test(`${internationalisationId} has a required state selector for a non-delivery ${product}`, async ({
+			context,
+			baseURL,
+		}) => {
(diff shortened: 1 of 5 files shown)


GHaberis (approved):
Looks good 👍

support-e2e/tests/smoke/non-delivery-state-selector.test.ts
```diff
@@ -0,0 +1,43 @@
...
+afterEachTasks(test);
+
+const countryGroupsWithStates = ['us', 'au', 'ca'];
+const nonDeliveryProducts = [
+	{ product: 'SupporterPlus', ratePlan: 'Monthly' },
+	{ product: 'Contribution', ratePlan: 'Monthly', contribution: 10 },
+];
+
+/** We require state for non-deliverable products as we use different taxes within those regions upstream */
+countryGroupsWithStates.forEach((internationalisationId) => {
```
andrewHEguardian, line 14: This is good, just wonder if it's worthy of it's own test. Do we have another that's testing the other form fields validation?

jamesgorrie: If we start adding more fields, we lose the reason for the test, which is to describe **_why_** we collect state on non-deliverable products in certain `countryGroups`.

andrewHEguardian: I guess my question is do we need the test to be isolated or is the comment describing why ("we require state for [...] taxes") enough and it could live within another test?

jamesgorrie: I'm trying to follow the ["isolation of tests" best practise](https://playwright.dev/docs/best-practices#make-tests-as-isolated-as-possible). There can definitely be other tests, but what heuristic do you think we could group them by?

jamesgorrie:
@andrewHEguardian I'm going to merge this in - but happy to continue the conversation outside of this PR.

support-e2e/tests/smoke/non-delivery-state-selector.test.ts
andrewHEguardian: In this case maybe validation rules? But this test is good

PR #6276, opened 2024-09-02
Comment out SEPA from validPaymenMethods on generic checkout

GHaberis:

Currently we show users from the EU "SEPA" as a payment method on the generic checkout. There are 2 issues here, 1) we have not implemented SEPA as a payment method, so selecting the payment method does nothing, and 2) SEPA is currently turned off on the non-generic checkout, so it shouldn't be listed as a valid payment method. The quickest change to resolve the issue we have right now is to comment out SEPA. Next steps would be:

1) Check payment method switch is ON when adding them to `validPaymentMethods`
2) Check whether we intend to reintroduce SEPA, and if so reimplement on the generic checkout.


andrewHEguardian (approved):
👍  good catch

PR #6278, opened 2024-09-02
fix: add error handling from support-workers on checkout

jamesgorrie:
Adds error handling from support-workers onto the checkout.

Screenshot 2024-09-02 at 14 51 40


andrewHEguardian (approved):
Great! 👍

PR #6279, opened 2024-09-03
refactor: remover productType and billingPeriod from landing page redux

jamesgorrie:
Removes productType and billingPeriod from the landing page redux state and puts it into a standard react `useState`.

> [!WARNING]  
> Do not merge as this breaks the integration with the old checkout that relies on this state being dispatched via redux


andrewHEguardian:
Completed in #6344

PR #6284, opened 2024-09-03
CDK upgrade

jamesgorrie:
Dependabot seems to be having a hard time doing this, probably because there's some breaking changes. 

This upgrades `@guardian/cdk` and fixes the related issues.


andrewHEguardian (approved):
Looks good to me 👍

PR #6285, opened 2024-09-04
fix: update peer deps for CDK

jamesgorrie:
- Bumps the `aws-cdk` and `aws-cdk-lib` to match `@guardian/cdk`
- Removes old V1 `aws-apigatewayv2` APIs in favour of V2 `aws-cdk-lib` apis

The result of this is almost a no-op bar that some of our services are now being tagged in a more granular and useful way.

I am checking with DevX to see if the tag `"gu:cdk:version": "TEST"` is correct though as that looks a little fishy.


andrewHEguardian (approved):

cdk/package.json
```diff
@@ -18,15 +18,13 @@
...
-    "@aws-cdk/aws-apigatewayv2": "1.113.0",
-    "@aws-cdk/aws-apigatewayv2-alpha": "^2.14.0-alpha.0",
     "@guardian/cdk": "v59.3.1",
     "@guardian/eslint-config-typescript": "1.0.11",
     "@guardian/prettier": "2.1.5",
     "@types/jest": "^29.5.12",
-    "@types/node": "20.13.0",
-    "aws-cdk": "2.122.0",
-    "aws-cdk-lib": "2.130.0",
+    "@types/node": "22.5.3",
```
andrewHEguardian, line 25: Does this need to match the node version we run?

jamesgorrie: Can you see docs to say this tracks? It looks like it does, but I can't find any evidence that it does, [bar on stackoverflow which suggests it didn't, then it did, then it didn't again](https://stackoverflow.com/questions/42035263/relationship-between-the-version-of-node-js-and-the-version-of-types-node)?

PR #6286, opened 2024-09-04
fix: tidy cdk libs a little more

jamesgorrie:
Going through warnings and tiding them up

- add license fields to `package.json`
- update node version to match `.nvmrc` as per [@andrewHEguardian's suggestion](https://github.com/guardian/support-frontend/pull/6285#discussion_r1743278168) ([from the docs](https://github.com/definitelytyped/definitelytyped?tab=readme-ov-file#how-do-definitely-typed-package-versions-relate-to-versions-of-the-corresponding-library): "If you're updating type declarations for a library, always set the major.minor version in package.json to match the library version that you're documenting!")
- upgrade typescript
- upgrade `@guardian/eslint-config-typescript` to match `support-frontend`

This should become easier if we start to treat this like a true mono-repo. This is helping setup for that.


cdk/package.json
```diff
@@ -19,19 +20,20 @@
...
-    "ts-jest": "^29.1.5",
-    "ts-node": "^10.9.2",
-    "typescript": "~4.9.5"
+    "eslint": "8.57.0",
+    "jest": "29.7.0",
+    "prettier": "2.8.8",
+    "source-map-support": "0.5.20",
+    "ts-jest": "29.1.5",
+    "ts-node": "10.9.2",
+    "tslib": "2.6.2",
```
andrewHEguardian, line 36: What is `tslib`, what does it do?

cdk/package.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "cdk",
-  "version": "0.0.0",
+  "version": "1.0.0",
+  "license": "Apache-2.0",
```
andrewHEguardian, line 4: What's the purpose of the license field?

andrewHEguardian (approved):
Great! Just a couple of questions for my own understanding

cdk/package.json
jamesgorrie: It's required - but besides that - because we code in the open, we should have a license  indicating how people _could_ reuse the code. They probably won't.

It's also part of [our recommendations](https://github.com/guardian/recommendations/blob/654e08854c2c0d3e83616339f5558010830d8fbd/github.md?plain=1#L25) - I might try and add a "Why Apache 2" to the docs.

cdk/package.json
jamesgorrie: It's a peer dependency of `@guardian/eslint-config-typescript`.

I've messed around with adding it to [`peerDependencies`](https://nodejs.org/en/blog/npm/peer-dependencies) which just causes more issues, so I've just left it here for now. 

Not sure how to make it more clear, but if you remove it you get warnings from yarn

PR #6288, opened 2024-09-04
Dynamic oauth redirect uri

jamesgorrie:
- Uses a dynamic oauth `redirect_url` based on if you're on `live.` or `support.` 🗡️ 
- Adds support to test `live.` locally
- Removes the baseURL from redirect URLs as they are no longer static i.e. `support.`

🗡️ this is what was causing the bug

### DIFF
--- a/support-frontend/app/controllers/CanonicalLinks.scala
+++ b/support-frontend/app/controllers/CanonicalLinks.scala
@@ -6,13 +6,13 @@ trait CanonicalLinks {
 
   def buildCanonicalDigitalSubscriptionLink(countryCode: String, orderIsAGift: Boolean): String =
     if (orderIsAGift) s"${supportUrl}/${countryCode}/subscribe/digital/gift"
-    else s"${supportUrl}/${countryCode}/subscribe/digital"
+    else s"/${countryCode}/subscribe/digital"
 
   def buildCanonicalWeeklySubscriptionLink(countryCode: String, orderIsAGift: Boolean): String =
     if (orderIsAGift) s"${supportUrl}/${countryCode}/subscribe/weekly/gift"
-    else s"${supportUrl}/${countryCode}/subscribe/weekly"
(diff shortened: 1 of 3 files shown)


support-frontend/nginx/config.yml
```diff
@@ -0,0 +1,9 @@
+name: support-frontend
+domain-root: thegulocal.com
+mappings:
+  - port: 9211
+    prefix: support
+    websocket: /ws
+  - port: 9211
+    prefix: live
+    websocket: /ws
```
jamesgorrie, line 9: I'll add some docs on this in a bit.

support-frontend/app/controllers/CanonicalLinks.scala
```diff
@@ -6,13 +6,13 @@ trait CanonicalLinks {
 
   def buildCanonicalDigitalSubscriptionLink(countryCode: String, orderIsAGift: Boolean): String =
     if (orderIsAGift) s"${supportUrl}/${countryCode}/subscribe/digital/gift"
-    else s"${supportUrl}/${countryCode}/subscribe/digital"
+    else s"/${countryCode}/subscribe/digital"
```
andrewHEguardian, line 9: Why has this changed?

andrewHEguardian (approved):
Great!

support-frontend/app/controllers/CanonicalLinks.scala
jamesgorrie: `supportUrl` is taken from config - [which is set here](https://github.com/guardian/support-frontend/blob/main/support-frontend/conf/PROD.public.conf). 

This essentially makes the link relative the URL, rather than absolute with it in order to support multiple domains.

andrewHEguardian: But if the `orderIsAGift` we do use the absolute URL?

PR #6290, opened 2024-09-05
Address State Select - remove validation error style on page load

andrewHEguardian:

Add styling to prevent the state select having a red error border on page load. Use the [user-invalid](https://developer.mozilla.org/en-US/docs/Web/CSS/:user-invalid) selector to check whether it was actually the user form submission that caused the error. It seems like a [known issue](https://stackoverflow.com/questions/27021801/inputinvalid-css-rule-is-applied-on-page-load) that the `invalid` selector is applied on page load. Should this change in [source](https://github.com/guardian/csnx/blob/67fbe1c1ed95c043ede73fc48020fb4b1bfc5e4f/libs/%40guardian/source/src/react-components/select/styles.ts#L84)?

## Screenshots
Before (on page load)
image

After (on page load)
image

However there is some strange behaviour now when interacting with the select but not selecting another option
image

After submit it looks as expected
image

### DIFF
--- a/support-frontend/assets/components/subscriptionCheckouts/address/addressFields.tsx
+++ b/support-frontend/assets/components/subscriptionCheckouts/address/addressFields.tsx
@@ -1,6 +1,6 @@
 import { css } from '@emotion/react';
 import { isNonNullable } from '@guardian/libs';
-import { space } from '@guardian/source/foundations';
+import { palette, space } from '@guardian/source/foundations';
 import {
 	Option as OptionForSelect,
 	Select,
@@ -54,6 +54,13 @@ type PropTypes = StatePropTypes & {
 
 const marginBottom = css`
 	margin-bottom: ${space[6]}px;
+`;
+
+const selectStateStyles = css`
+	&:invalid:not(&:user-invalid) {
+		/* Remove styling of invalid select element */
+		border: 1px solid ${palette.neutral[46]};
+	}
 `;
 
 const MaybeSelect = canShow(Select);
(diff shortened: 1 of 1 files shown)


support-frontend/assets/components/subscriptionCheckouts/address/addressFields.tsx
```diff
@@ -56,6 +56,13 @@ const marginBottom = css`
 	margin-bottom: ${space[6]}px;
 `;
 
+const selectStateStyles = css`
+	&:invalid:not(&:user-invalid) {
+		/* Remove styling of invalid select element */
+		border: 1px solid ${palette.neutral[46]};
+	}
+`;
+
```
jamesgorrie, line 65: We do this [here](https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/components/personalDetails/stateSelect.tsx#L60-L72) too.

Maybe worth creating a `Select` of our own applying this - but we should also fix this upstream in Source.

jamesgorrie (approved):

support-frontend/assets/components/subscriptionCheckouts/address/addressFields.tsx
GHaberis: Nice, I've never used this `:user-invalid` pseudo class before

GHaberis (approved):

support-frontend/assets/components/subscriptionCheckouts/address/addressFields.tsx
andrewHEguardian: It's fairly new. It's not [compatible](https://developer.mozilla.org/en-US/docs/Web/CSS/:user-invalid#browser_compatibility) with older browsers but that's probably okay since it's not preventing any usability

PR #6291, opened 2024-09-05
feat: add appconfig endpoint

jamesgorrie:
As per this diagram - a step to work towards having a server side rendered page would be to make the app config accessible via a URL that a node app could read and render from.

This starts us on that journey.

Here's the full diagram with the end goal, the part this is focussing on is labelled "This PR" in purple.
Screenshot 2024-09-05 at 21 22 11

### DIFF
--- a/support-frontend/app/controllers/Application.scala
+++ b/support-frontend/app/controllers/Application.scala
@@ -1,32 +1,174 @@
 package controllers
 
 import actions.AsyncAuthenticatedBuilder.OptionalAuthRequest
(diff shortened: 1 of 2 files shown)


andrewHEguardian (approved):
Love the diagram. Which bits of it does `config.json` represent? Is it everything in Data fetching bar private config?

jamesgorrie:
> Love the diagram. Which bits of it does `config.json` represent? Is it everything in Data fetching bar private config?

It is indeed - basically everything on the `window.guardian` object currently

PR #6293, opened 2024-09-09
fix: move generic checkout to 100%

jamesgorrie:
[After an incredible investigation](https://docs.google.com/spreadsheets/d/19kkkQfEHxWx-hyvgXOaTS3YU9aYs2qDfvzFW8dn7RIM/edit?gid=872405396#gid=872405396)(🔒 Private) from @GHaberis - it looks like we're hitting at least parity with the contributions checkout - so off we pop to 100%

Great work by all.


andrewHEguardian (approved):
🚀💯🚀💯🚀💯🚀💯

PR #6294, opened 2024-09-10
fix: make variant the default for useGenericCheckout test

jamesgorrie:
Changes the default for `useGenericCheckout` test to `variant`.
Oopsie.


andrewHEguardian (approved):
👍

PR #6295, opened 2024-09-10
Remove useGenericCheckout test

jamesgorrie:
Removes references to the useGenericCheckout test. 

This will remain in draft until we're happy with the numbers and sure that the new checkout is good.

It is good to review though.

## All tests 🍏 

Screenshot 2024-09-10 at 10 56 37


support-frontend/assets/pages/supporter-plus-landing/twoStepPages/threeTierLanding.tsx
```diff
@@ -392,65 +387,6 @@ export function ThreeTierLanding({
...
-				contributionCurrency,
-			);
-			/**
-			 * Lower & middle tier track component click fired via redux side effects.
-			 * Top tier accessed via network request to GuardianWeekly landing page
-			 * therefore tracking required
-			 **/
-			trackComponentClick(
-				`npf-contribution-amount-toggle-${countryGroupId}-${contributionType}-${price}`,
-			);
```
jamesgorrie, line 417: @GHaberis should this be fired on the pageview of the checkout?

paul-daniel-dempsey (commented):
Nice simplication for `ThreeTierLanding`, reduces the amount of test too. Dont thnk the Tier3 tests are passing at the moment however.

support-frontend/assets/pages/supporter-plus-landing/twoStepPages/threeTierLanding.tsx
GHaberis: @jamesgorrie `sendEventContributionCartValue` reports to QM the "cart value" so they can attribute financial impacts of issues they pick-up, it should therefore be fired on the pageview of the checkout.

To be honest, I'd look at whether this is easily implementable and if not proceed this PR without implementing this and create a card to revisit `sendEventContributionCartValue`  as it's not a piece of data we're using - so is not a priority.

jamesgorrie: I'd agree that we should should look at writing these methods with the new product model.

GHaberis (approved):
Looks great 👍

PR #6298, opened 2024-09-10
Create Newspaper archive benefit Thank you module structure

andrewHEguardian:

Implementing the structure and logic for the new Thank You module: Newspaper archive benefit. This re-uses the existing module styling but updating the placement based on new designs.

[**Trello Card**](https://trello.com/c/domGTnMJ/1018-ux-design-of-newspaper-archive-component-on-thank-you-page)

## Why are you doing this?

Creating the initial design structure for the new benefit.

## How to test

Activate AB test `newspaperArchiveBenefit` - this module is shown regardless of which variant is in play. Go through to Tier 3 thank you page, see new module with temp imagery, new images not yet provided. 

Storybook should catch that there are no regressions to existing thank you modules. (Although we aren't snapshotting the mobile versions so I've tested manually)

## Screenshots
Desktop
image

Mobile
image


support-frontend/assets/components/thankYou/thankYouModule.tsx
```diff
@@ -253,21 +244,22 @@ function ThankYouModule({
 		trackComponentLoadId && trackComponentLoad(trackComponentLoadId);
 	}, []);
 
-	const isDownloadModule = moduleType === 'appDownload';
+	const hasQrCodes = moduleType === 'appDownload';
 	const isDownloadModules = moduleType === 'appsDownload';
+	const hasImagery = ['appDownload', 'newspaperArchiveBenefit'].includes(
```
andrewHEguardian, line 249: I've reused the imagery layout but I think we should just delete `appDownload` as an option (separate PR to keep it small) since it's unused

paul-daniel-dempsey (approved):
Looks good, like the `hasQRCodes` even if we do remove it later.

PR #6299, opened 2024-09-10
feat: redirect contribution checkout => generic checkout

jamesgorrie:
Redirects traffic from `/contribute/checkout` to the correct product (`Contribution | SupporterPlus`) on `/checkout`.

Given the old model was to calculate whether something was `SupportPlus` or a rucurring `Contribution` - `getProductParamsFromContributionParams` takes the old query string params, and calculates this at the edge of the app, and then transforms that data in the common model we've been using for the generic checkout.

~We could have done this on the server, but I feel trying to push as little to the server as possible is desired as `TypeScript` is the de facto language the the rendering layer is written in.~

We have done this on the server 
- because it means the responses are cached and actually return the correct HTTP code i.e. `302`
- we don't pollute the react app with non-app code
- there were issues with the page having the correct `productPrices` if we just did client side routing

e.g.

- Contribution
  - Monthly
    - `/uk/contribute/checkout?selected-amount=4&selected-contribution-type=monthly`
    - `/uk/contribute/checkout?product=Contribution&ratePlan=Monthly&contribution=4`
  - Annual
    - `/uk/contribute/checkout?selected-amount=4&selected-contribution-type=annual`
    - `/uk/contribute/checkout?product=Contribution&ratePlan=Annual&contribution=4`

- SupporterPlus
  - Monthly
    - `/uk/contribute/checkout?selected-amount=12&selected-contribution-type=monthly`
    - `/uk/contribute/checkout?product=SupporterPlus&ratePlan=Monthly`
  - Annual
    - `/uk/contribute/checkout?selected-amount=120&selected-contribution-type=annual`
    - `/uk/contribute/checkout?product=SupporterPlus&ratePlan=Annual`
  - With contribution (Monthly)
    - `/uk/contribute/checkout?selected-amount=15&selected-contribution-type=monthly`
    - `/uk/contribute/checkout?product=SupporterPlus&ratePlan=Annual&contribution=3`


support-frontend/assets/pages/[countryGroupId]/router.tsx
```diff
@@ -1,17 +1,92 @@
...
+	return (
+		<Navigate
+			to={`/${geoId}/checkout?product=${product}&ratePlan=${ratePlan}${
+				product === 'Contribution' ? `&contribution=${amount}` : ''
+			}`}
+			replace
+		/>
+	);
+}
+
```
Comment (jamesgorrie) [inline support-frontend/assets/pages/[countryGroupId]/router.tsx:80]: I'll probably write a test for this.

support-frontend/app/controllers/Application.scala
```diff
@@ -315,11 +308,13 @@ class Application(
...
-      idUser: Option[IdUser],
       campaignCode: Option[String],
-      guestAccountCreationToken: Option[String],
-  )(implicit request: RequestHeader, settings: AllSettings) = {
+  )(implicit request: OptionalAuthRequest[AnyContent], settings: AllSettings) = {
+    val geoData = request.geoData
+    val idUser = request.user
+
+    // This will be present if the token has been flashed into the session by the PayPal redirect endpoint
+    val guestAccountCreationToken = request.flash.get("guestAccountCreationToken")
```
jamesgorrie, line 317: I can't say I fully understand why we have `contributionsHtml` and `contributionsLanding` but it's helped here - I just needed to shift some of the arguments into the actual method.

support-frontend/assets/pages/[countryGroupId]/components/contributionCheckoutRedirect.tsx
```diff
@@ -0,0 +1,98 @@
...
+import type { AppConfig } from 'helpers/globalsAndSwitches/window';
+import { type GeoId, getGeoIdConfig } from 'pages/geoIdConfig';
+
+const contributionTypes = ['monthly', 'annual'] as const;
+type SelectedContributionType = (typeof contributionTypes)[number];
+function isContributionType(val: unknown): val is SelectedContributionType {
+	return contributionTypes.includes(val as SelectedContributionType);
+}
+
+export function getProductFromContributionParams(
```
Comment (paul-daniel-dempsey) [inline support-frontend/assets/pages/[countryGroupId]/components/contributionCheckoutRedirect.tsx:11]: v.minor, is `export` used outside of this file?

paul-daniel-dempsey (approved):
Great work, re-directs great, wondered whether your switch idea would be useful in case we need to go back in the short term? Presume there is no real need for fallback to the old checkout if we dont supply either  `selected-contribution-type` or `selected-amount` or are missing both paramUrl's?

jamesgorrie:
> wondered whether your switch idea would be useful in case we need to go back in the short term

@paul-daniel-dempsey I think I have gone back on this as too cautious as we have run the a/b test and we're knowing serving it at 100%. 

If something goes wrong short terms we can revert. The intention of this is to start removing the old checkout code and start using the new simplified workflow so we no longer need to develop on both and speed up delivery.

support-frontend/app/controllers/Application.scala
```diff
@@ -422,6 +418,83 @@ class Application(
...
+    val currency = countryGroupId match {
+      case "uk" => "GBP"
+      case "us" => "USD"
+      case "au" => "AUD"
+      case "eu" => "EUR"
+      case "int" => "USD"
+      case "nz" => "NZD"
+      case "ca" => "CAD"
+      case _ => "GBP"
+    }
```
rupertbates, line 446:
```suggestion
val currency = CountryGroup.byId(countryGroupId).getOrElse(CountryGroup.UK).currency.iso
```

support-frontend/app/controllers/Application.scala
```diff
@@ -422,6 +418,83 @@ class Application(
...
+    val isAnnual = maybeSelectedContributionType.contains("annual")
+    val ratePlan = if (isAnnual) "Annual" else "Monthly"
+    val maybeSupporterPlusAmount = productCatalog.asJson.hcursor
```
rupertbates, line 449: Do we not have a `ProductCatalog` case class we can deserialise this to?

jamesgorrie: We've purposefully not defined this in Scala as we don't use it there - we just pass the JSON down from the API into the `window.guardian` object. 

We have the definitions for this API in TypeScript as that's how it's defined.

If we want Scala types - should we look at defining the API in a more generic way like Thrift to generate the models as maintaining a few sets of models if error prone and not great use of dev time?

jamesgorrie: > Cursors are very low level to be working with in application code

@rupertbates not sure I get this - [from the docs](https://circe.github.io/circe/cursors.html) it looks like a general way to traverse JSON?

rupertbates: It is general, that's kind of my point there's a whole load of unnecessary concepts in that code like cursor and downField and we're using strings which give us no type safety. I take your point about not needing it anywhere else currently though, lets leave it for now and we can improve it when we get to using the product catalog elsewhere in the stack.

rupertbates (approved):

PR #6301, opened 2024-09-10
Ticket Tailor: widget border hide

paul-daniel-dempsey:

- Ts&Cs creates new tab
- Tablet resize backdrop upwards to display Ts&Cs clearly
- Remove border shown here ->

|before|after|
|-----|-----|
|image|image|

[**Trello Card**](https://trello.com/c/j6F32kdQ/1064-ticket-tailor-ui-border-tweak)

### DIFF
--- a/support-frontend/assets/pages/[countryGroupId]/events/events.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/events/events.tsx
@@ -91,7 +91,11 @@ export function Events({ geoId }: Props) {
 
 	const params = useParams();
 	const eventId = params.eventId;
-	const termsEvents = <a href={guardianLiveTermsLink}>Terms and Conditions</a>;
+	const termsEvents = (
+		<a href={guardianLiveTermsLink} target="_blank">
+			Terms and Conditions
+		</a>
+	);
 	const privacyPolicy = <a href={privacyLink}>Privacy Policy</a>;
 
 	const pageviewId = getPageViewId();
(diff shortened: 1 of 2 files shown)


support-frontend/assets/pages/[countryGroupId]/events/events.tsx
```diff
@@ -31,10 +31,20 @@ const columns = css`
...
+
+	> div {
+		padding: 0;
+	}
+
+	// ticket tailor widget (class:box-color-background) margin
+	// needs removal, its container we can reduce to hide
+	> div div div {
+		margin: -2px -2px -8px -2px;
+	}
```
Comment (jamesgorrie) [inline support-frontend/assets/pages/[countryGroupId]/events/events.tsx:47]: This is quite a confusing solution - I am assuming it's because we're using Source, and we need to target the `div`s that creates?

Could we explore not using source?

Either way - we might want to change the name `shorterBoxMargin` to something more descriptive.

jamesgorrie (approved):
It works, just a couple of suggestions as it's a little fiddly to understand.

support-frontend/assets/pages/[countryGroupId]/events/events.tsx
Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/events/events.tsx:59]: Have renamed the css to `narrowBoxMarginAndPadding` which is a bit more descriptive.

Struggled to access css on div's within the ticket tailor widget itself (presume does not exist until runtime?) thus could not access the class:`box-color-background` directly to override its css.

Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/events/events.tsx:61]: Will put in as-is and revisit if a better way appears.

PR #6302, opened 2024-09-10
Fix Cover transaction checkbox on mobile

andrewHEguardian:

Cherry picking the latest Checkbox from Source to fix the Cover transaction checkbox styling from misaligning on mobile.

Unfortunately we have quite a few breaking changes in the upgrade path to source v7, which is where this fix is from.

The AB test is meant go live ASAP, so doing this hack should give us some time to upgrade smoothly and remove this temporary measure after

## Screenshots
Before
image

After
image


GHaberis (approved):
Looks good 👍

PR #6312, opened 2024-09-12
refactor: use more basic renderToString setup for SSR content

jamesgorrie:
This adds a simple file that writes an HTML file into existing `ssr-cache` directory that is then referenced by the twirl templates.

We were investigating whether we could just do this with HTML, but I wonder if that adds more complexity by having an other way of rendering.

I also feel this starts to get us a little closer to server rendering more - we can extend this script to host more and more of the HTML and remove it from twirl.

If we think this is a good approach, I can remove all the other pieces.

As much as we can use React - because oif bypassing the webpack config which is convoluted and contains a tonne of stuff we have deprecated but not given the time to sunset - we can't use things like

- SASS
- Not sure what else, but that's a large one as a few components use it, but a lot of them are for the deprecated checkouts, so we might be able to start removing those references soon

https://github.com/user-attachments/assets/70deed7b-f3f0-4ef9-b3a6-2712fd6d9245


support-frontend/assets/components/page/pageScaffold.tsx
```diff
@@ -1,9 +1,10 @@
 import { css, Global } from '@emotion/react';
-import { brand, FocusStyleManager } from '@guardian/source/foundations';
+// import { brand, FocusStyleManager } from '@guardian/source/foundations';
+import { brand } from '@guardian/source/foundations';
 import type { ReactNode } from 'react';
 import CsrBanner from 'components/csr/csrBanner';
 import { SkipLink } from 'components/skipLink/skipLink';
-import { TestUserBanner } from 'components/test-user-banner/testUserBanner';
+// import { TestUserBanner } from 'components/test-user-banner/testUserBanner';
```
jamesgorrie, line 7: This pulls in Ophan which then breaks `renderToString` by referencing `document`, so I am having a think about what we do here.

support-frontend/scripts/build-ssr-holding-content.tsx
```diff
@@ -0,0 +1,107 @@
...
+		flex-direction: column;
+		& > :last-child {
+			flex: 1;
+		}
+	}
+`;
+
+function Loading() {
+	return (
+		<div css={container}>
```
jamesgorrie, line 73: I've avoided using `PageScaffold` here as
- it introduces elements that conflict with `renderToString` by using `window.document` not in a `useEffect`

andrewHEguardian (approved):
This feels a lot simpler 👍 But also still flexible

PR #6314, opened 2024-09-13
Generic checkout - check payment method feature switches

andrewHEguardian:

Updating generic and one-time checkout to check payment method feature switches.

This is a screenshot of the payment method switches. The notion of recurring vs subscriptions isn't a distinction the generic checkout currently makes, so the plan is to use the recurring payment methods for all products on generic checkout and phase out `subscriptionsPaymentMethods`.

image

[**Trello Card**](https://trello.com/c/V5Y5w0xO/1044-check-feature-switches-when-populating-validpaymentmethods-on-generic-checkout)

### DIFF
--- a/support-frontend/assets/helpers/forms/paymentMethods.ts
+++ b/support-frontend/assets/helpers/forms/paymentMethods.ts
@@ -1,3 +1,4 @@
+import type { PaymentMethodSwitch } from './checkouts';
 import type { StripePaymentMethod } from './paymentIntegrations/readerRevenueApis';
 
 const Stripe = 'Stripe';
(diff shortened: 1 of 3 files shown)


support-frontend/assets/helpers/forms/paymentMethods.ts
```diff
@@ -44,4 +45,28 @@ export const isPaymentMethod = (
 	);
 };
 
+export function toPaymentMethodSwitchNaming(
```
paul-daniel-dempsey, line 48: Presume we will eventually retire the same call in `support-frontend/assets/helpers/forms/checkouts.ts`?

paul-daniel-dempsey (approved):
Nice and neat filtering, looks great

PR #6313, opened 2024-09-13
Ticket Tailor: Playwright Cron & Smoke Integration

paul-daniel-dempsey:
## What are you doing in this PR?
Add Cron (hourly) and Smoke (upon build) tests for Ticket Tailor->
1. Check test event container loads

[**Trello Card**](https://trello.com/c/Rfxl2t59/1036-ticket-tailor-alarm-implementation)

PR #6320, opened 2024-09-17
feat: add ticket tailor e2e test

jamesgorrie:
Adds an e2e test that checks the ticket tailor iframe loads correctly.

90% generated with the prompt
```
Write me a playwright test that checks that an iframe with a tickettailor.com URL loads correctly
```

and then refined with 
```
elementHandle doesn't seem to be available on the iframe
```

Which then fixed trying access a method on the `frame` called `elementHandle` which didn't exist.

The main difference between #6313 and this is we use the [`contentFrame`](https://playwright.dev/docs/api/class-elementhandle#element-handle-content-frame) on the `iframe` element.

## Testing

1. `test-smoke` script added.
2. `test-cron` script added.

To run script containing ticket tailor e2e test->
`yarn test-smoke`

## Screenshots
Screenshot 2024-09-17 at 08 24 39

### DIFF
--- a/.github/workflows/playwright-cron.yml
+++ b/.github/workflows/playwright-cron.yml
@@ -24,7 +24,7 @@ jobs:
       run: yarn playwright install --with-deps
     - name: Run Playwright tests
       working-directory: ./support-e2e
-      run: yarn playwright test --project=cron --config=playwright.config.ts
+      run: yarn test-cron
     - uses: actions/upload-artifact@v4
       if: ${{ !cancelled() }}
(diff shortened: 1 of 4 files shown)


paul-daniel-dempsey (approved):
Thanks James for figuring this out. The use of `page.waitForSelector` & `iframe.contentFrame` to get to the frame seems the way to go.

jamesgorrie (commented):
@paul-daniel-dempsey - your changes look good!

PR #6322, opened 2024-09-17
New Plans for T3 archive benefit

rupertbates:
We want to AB test adding the newspaper archive benefit to the tier three product, to see whether it causes an uplift in conversion - [more details in this presentation](https://docs.google.com/presentation/d/18f1uYissTG27WCxDrfV69fBG2tYX90aG_0cE1PkDvWI/edit#slide=id.g2fbf351fd3e_0_102)

To do this we need to introduce new product rate plans with an additional archive charge and sell these plans to users in the cohort who will be getting the new benefit. 
Because currently we only have one rate plan per billing period and fulfilment option for tier three, this PR differentiates the new plans by using a `productOption` value of `NewspaperArchive`, the original non-archive plans have a `productOption` value of `NoProductOptions`

[More details in this doc](https://docs.google.com/document/d/1lL8SCyy-LRrHOxzP9C3HV1UFdQ8AMOuQxQEK77j4ogs/edit?usp=sharing)

### DIFF
--- a/support-frontend/assets/helpers/forms/paymentIntegrations/readerRevenueApis.ts
+++ b/support-frontend/assets/helpers/forms/paymentIntegrations/readerRevenueApis.ts
@@ -67,6 +67,7 @@ export type TierThree = {
 	currency: string;
 	billingPeriod: BillingPeriod;
 	fulfilmentOptions: FulfilmentOptions;
+	productOptions: ProductOptions;
 };
 export type DigitalSubscription = {
 	productType: typeof DigitalPack;
--- a/support-frontend/assets/helpers/productCatalog.tsx
+++ b/support-frontend/assets/helpers/productCatalog.tsx
@@ -103,6 +103,18 @@ export const productCatalogDescription: Record<ProductKey, ProductDescription> =
 				RestOfWorldAnnual: {
 					billingPeriod: 'Annual',
 				},
+				DomesticMonthlyV2: {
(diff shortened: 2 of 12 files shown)


jamesgorrie (commented):
I'm uncertain why we need to specify that the `productOptions` is `NewspaperArchive` - isn't that intrinsic to the V2 ratePlan?

rupertbates:
> I'm uncertain why we need to specify that the `productOptions` is `NewspaperArchive` - isn't that intrinsic to the V2 ratePlan?

We need to be able to differentiate between the new rate plans and the old ones when we are creating the subscription. Currently we just pick the rate plan based on the billing period (monthly or annual) and the fulfilmentOptions (domestic or rest of world), however that is no longer sufficient to pick a unique plan because there is a V1 & a V2 version, so we use productOptions to differentiate between those. 
Users in the V1 & V2 AB test variants get the `NewspaperArchive` productOptions setting and so we create a sub with the V2 rate plan, those in the control get the `NoProductOptions` value so we pick the V1 rate plan.
Not that we have two different uses of V1 & V2 here which is a bit confusing - one is the rate plan versions, the other is the AB test variant names

jamesgorrie (approved):
Thanks for the explanation - this makes sense.

Worth mentioning something we spoke about on chat - if the `ratePlan` percolated through the whole stack i.e. `support-workers` - this would be a little easier.

PR #6323, opened 2024-09-18
refactor: use product qs param to calculate user selected tier

jamesgorrie:
Uses the `product` query string to choose whether the tier is selected.

@paul-daniel-dempsey and I discussed using a `selected-tier` or similar - but figured currently there is a 1-1 mapping between tier and product, so let's keep the simplicity.

Once [this PR](https://github.com/guardian/dotcom-rendering/pull/12403) is merged we can remove the `isCardUserSelected` method.


paul-daniel-dempsey (approved):
Use of the `product` for `isUserSelected` rather than amounts feels more future-proof against price changes.

PR #6324, opened 2024-09-18
fix: only return amount if it is notNan

jamesgorrie:
Only returns the value is it is not NaN.


paul-daniel-dempsey (approved):
Great! The 2nd version with `0` case think is more type-safe would say.

PR #6326, opened 2024-09-18
One Time Checkout - PriceCard Other Amount Validition

paul-daniel-dempsey:

PriceCard `OtherAmount` Validation->

- Starts With Validation Disabled
- No Item entered (after focus lose)
-  Non-numeric entered (after focus lose)

[**Trello Card**](https://trello.com/c/1zyFpDGD/1073-one-time-generic-checkout-form-validation-upon-defaultpaymentbutton-request)

|Initial Startup|Empty|Invalid Entry|
|-----|-----|-----|
|image|image|image|

### DIFF
--- a/support-frontend/assets/components/otherAmount/otherAmount.tsx
+++ b/support-frontend/assets/components/otherAmount/otherAmount.tsx
@@ -1,7 +1,7 @@
 import { css } from '@emotion/react';
 import { space } from '@guardian/source/foundations';
 import { NumericInput } from '@guardian/source-development-kitchen/react-components';
-import type { FocusEventHandler, FormEventHandler } from 'react';
+import type { HTMLAttributes } from 'react';
 import { currencies } from 'helpers/internationalisation/currency';
 import type { IsoCurrency } from 'helpers/internationalisation/currency';
 
@@ -15,10 +15,8 @@ export type OtherAmountProps = {
 	currency: IsoCurrency;
 	minAmount: number;
 	onOtherAmountChange: (newAmount: string) => void;
-	onBlur?: FocusEventHandler<HTMLInputElement>;
-	onInvalid?: FormEventHandler<HTMLInputElement>;
 	errors?: string[];
-};
+} & HTMLAttributes<HTMLInputElement>;
 
 export function OtherAmount({
 	selectedAmount,
--- a/support-frontend/assets/components/otherAmount/otherAmount.tsx
+++ b/support-frontend/assets/components/otherAmount/otherAmount.tsx
@@ -4,7 +4,6 @@ import { NumericInput } from '@guardian/source-development-kitchen/react-compone
 import type { FocusEventHandler, FormEventHandler } from 'react';
 import { currencies } from 'helpers/internationalisation/currency';
 import type { IsoCurrency } from 'helpers/internationalisation/currency';
-import { numericOnlyPattern } from 'pages/[countryGroupId]/validation';
 
 const topSpacing = css`
 	margin-top: ${space[2]}px;
@@ -48,7 +47,7 @@ export function OtherAmount({
 					suffixText={suffix}
 					error={errors?.[0]}
 					value={otherAmount}
-					pattern={numericOnlyPattern}
+					type="number"
 					onBlur={onBlur}
 					onChange={(e: React.ChangeEvent<HTMLInputElement>) =>
 						onOtherAmountChange(e.target.value)
--- a/support-frontend/assets/components/otherAmount/otherAmount.tsx
+++ b/support-frontend/assets/components/otherAmount/otherAmount.tsx
@@ -1,8 +1,10 @@
 import { css } from '@emotion/react';
 import { space } from '@guardian/source/foundations';
 import { NumericInput } from '@guardian/source-development-kitchen/react-components';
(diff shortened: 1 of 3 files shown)


support-frontend/assets/components/otherAmount/otherAmount.tsx
```diff
@@ -14,6 +16,8 @@ export type OtherAmountProps = {
 	currency: IsoCurrency;
 	minAmount: number;
 	onOtherAmountChange: (newAmount: string) => void;
+	onBlur?: FocusEventHandler<HTMLInputElement>;
+	onInvalid?: FormEventHandler<HTMLInputElement>;
```
jamesgorrie, line 20: given this element is actually just an HTMLInputElement - you could use an [intersection](https://www.typescriptlang.org/docs/handbook/2/objects.html#intersection-types) of those objects

```ts
type OtherAmountProps = {/*...*/} & HTMLAttributes<HTMLButtonElement>
```

support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx
```diff
@@ -206,6 +206,25 @@ function OneTimeCheckoutComponent({
 
 	const formRef = useRef<HTMLFormElement>(null);
 
+	const displayValidity = (
```
Comment (jamesgorrie) [inline support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx:209]: Is this clearer than the code it encompasses? The code in this method is standard HTML apis, whereas this is a new custom one. 

I do think it looks a little neater in the hander methods, but obfuscates what's actually happening.

I don't have a preference really, just asking.

support-frontend/assets/pages/[countryGroupId]/validation.ts
```diff
@@ -19,3 +19,4 @@ export function preventDefaultValidityMessage(
  * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Unicode_character_class_escape
  */
 export const doesNotContainEmojiPattern = '^[^\\p{Emoji_Presentation}]+$';
+export const numericOnlyPattern = '^[0-9]*$';
```
Comment (jamesgorrie) [inline support-frontend/assets/pages/[countryGroupId]/validation.ts:22]: Should this not be validated by having [`<input type="number" />`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number)?

jamesgorrie (commented):
It's looking good, not changes requested, just a few questions.

support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx
Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx:209]: Think mainly was reducing duplication of code as its content was repeated  3 times.

Comment (jamesgorrie) [reply support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx:209]: Repetition of code is not necessarily a bad thing - sometimes things happen multiple times. 

There are multiple reasons to abstract something
- encapsulating good practise e.g. `POST`ing to an auth endpoint with creds
- Simplification and understandability - Avoiding having a future person work out how to do something complicated e.g. potentially this?


There's more. Just wondering - from your perspective - what you are achieving with this abstraction?

Comment (jamesgorrie) [inline support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx:209]: As a second point to this, it does not really matter - but would `onInvalid` or `setInvalid` be a more descriptive name? Trying to work out what `display` is referencing?

jamesgorrie (approved):
Definitely not blocking, just some questions I had.

support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx
Comment (jamesgorrie) [reply support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx:209]: To be clear though - if we think this makes it more readable, that's a good reason, especially that's it's encapsulated to this package and not leaky.

Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx:209]: I had it as `setInvalid` but thought was too close in naming to a Set<State> calls. hence used the word `displayValidity` as it checks the validity for a field and then displays the error message. Do you think I shoul dnot worry about using set in functions alongside SetState?

Comment (jamesgorrie) [reply support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx:209]: > checks the validity for a input text field and then displays the error message

This feels like part of it - what it does is check the validity and then conditionally show or hide the error message based on that.

Maybe just `validate` could work?

Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx:209]: Yes thats more concise, done, thanks.

support-frontend/assets/pages/[countryGroupId]/validation.ts
Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/validation.ts]: Ah yes, much better than using un-required regex. Changed.

support-frontend/assets/components/otherAmount/otherAmount.tsx
paul-daniel-dempsey: That's clever, like that, presume it works because `HTMLInputElement` props are subset of `HTMLAttributes` props. Thanks, changed.

PR #6327, opened 2024-09-19
Revert "New Plans for T3 archive benefit"

rupertbates:
Reverts guardian/support-frontend#6322

#6322 causes a problem with promotions being applied to T3. Reverting while I investigate.


paul-daniel-dempsey (approved):
Yep revert

PR #6328, opened 2024-09-19
Reapply "New Plans for T3 archive benefit"

rupertbates:
This PR reapplies the changes in #6322 with a [fix for the bug](https://github.com/guardian/support-frontend/pull/6328/commits/5a2ce32e0f5e2ddb4dcbb229bf50f5f7ca145f1e) in that release


jamesgorrie (approved):
Got it - so the bug was that `abParticipations.newspaperArchiveBenefit` could be `undefined` and thus we set the `productOptions` to `NewspaperArchive`.

PR #6329, opened 2024-09-19
fix: deactivate coverTransactionCost ab test

jamesgorrie:
Deactivates the `coverTransactionCost` test as we have seen a drop in CVR.


paul-daniel-dempsey (approved):
LG

PR #6331, opened 2024-09-19
feat: add local e2e testing

jamesgorrie:
## What
Adds the ability to run the new style of e2e tests locally.

## Why
I'd like to make a change that I would feel way more comfortable knowing works locally - this would allow me to do that.

I've gone with the `.dev` naming as that what we, and other frameworks, tend to use e.g. `make dev` in dcr, `npm run dev` in other places in the ecosystem.

It also means I can keep the `.local` file which will be deleted soon.

I've left this running in `--ui` mode as I feel this is the most common `dev` way of doing it from working with @paul-daniel-dempsey and @andrewHEguardian


paul-daniel-dempsey (approved):
LG

jamesgorrie:
@paul-daniel-dempsey [this was added here](https://github.com/guardian/support-frontend/pull/6331). The command should be `test-smoke-dev`.

PR #6332, opened 2024-09-20
refactor: rename assets.map to assets.json

jamesgorrie:
Renames the map generated by [Webpack Manifest Plugin](https://github.com/shellscape/webpack-manifest-plugin) to JSON.

This is so we can `import` it in NodeJs files without having to build any fancy loader for it as Node supporting JSON imports by default.

Another reason is that this is a JSON file


paul-daniel-dempsey (approved):
Looks good, maybe update package.json with `test-local-dev` if needed.

PR #6333, opened 2024-09-20
Use new T3 archive rate plans for all cohorts in archive AB test

rupertbates:
This PR changes the newspaper archive AB test to use the new archive rate plans for all variants of the test, so the control group will also receive the new benefits. The only time when the old rate plans are used now is when the test is inactive.

More information about this test is [here](https://docs.google.com/document/d/1lL8SCyy-LRrHOxzP9C3HV1UFdQ8AMOuQxQEK77j4ogs/edit?usp=sharing)

## Changes
To support this behaviour I have had to make a change to the `Participations` type in our TS code. This was not accurately describing the possible values of an AB test because it was of type `[string, string]` where the first string is the test name and the second is the variant which the user is in, however when a test is inactive (`isActive: false`) the variant value will be undefined. I have redefined the type as `[string, string | undefined]` to more accurately model the possible values a test participation can be in.

## [Trello Card](https://trello.com/c/gVvJIwvC/1091-all-ab-test-cohorts-should-get-new-archive-rate-plans)

### DIFF
--- a/support-frontend/assets/helpers/abTests/abtest.ts
+++ b/support-frontend/assets/helpers/abTests/abtest.ts
@@ -37,7 +37,7 @@ type BreakpointRange = {
 	maxWidth?: Breakpoint;
 };
 
-export type Participations = Record<string, string>;
+export type Participations = Record<string, string | undefined>;
 
 export type Audience = {
 	offset: number;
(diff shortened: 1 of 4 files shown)


andrewHEguardian (approved):
Makes sense 👍  always good to have a more accurate type

jamesgorrie (approved):

PR #6334, opened 2024-09-20
Rename ophan file to enable testing

rupertbates:
This PR renames the ophan module to trackingOphan. 

The reason for doing this is that when it was named ophan it was not possible to test any functions in it because it is getting mocked by this code here: https://github.com/guardian/support-frontend/blob/08643c2d5aa3875a5934db82758c2d3f73852a06/support-frontend/jestSetup.js#L6-L10

Strangely, changing the module name to eg. ophanTracking still results in the same issue, the name needs to start with something other than ophan to work.

If anyone has a better solution for this please chip in!

### DIFF
--- a/support-frontend/assets/helpers/tracking/__tests__/ophanTest.ts
+++ b/support-frontend/assets/helpers/tracking/__tests__/ophanTest.ts
@@ -1,0 +1,26 @@
+import type { Participations } from '../../abTests/abtest';
+import type { OphanABPayload } from '../trackingOphan';
+import { buildOphanPayload } from '../trackingOphan';
+
+describe('Ophan AB Payload', () => {
+	it('should build a payload from participations', () => {
+		const input: Participations = {
+			test1: 'control',
+			test2: 'variant',
+		};
+		const expected: OphanABPayload = {
+			test1: {
+				variantName: 'control',
+				complete: false,
+				campaignCodes: [],
+			},
+			test2: {
+				variantName: 'variant',
+				complete: false,
+				campaignCodes: [],
+			},
+		};
+		const payload = buildOphanPayload(input);
+		expect(payload).toEqual(expected);
+	});
+});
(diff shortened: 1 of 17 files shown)


support-frontend/assets/helpers/tracking/__tests__/ophanTest.ts
```diff
@@ -0,0 +1,26 @@
+import type { Participations } from '../../abTests/abtest';
+import type { OphanABPayload } from '../trackingOphan';
+import { buildOphanPayload } from '../trackingOphan';
```
rupertbates, line 3: We can now test functions in the trackingOphan module

jamesgorrie (approved):
This is really odd behavior from jest. 

`Jest.mock('ophan')` seems to mock `ophan` in the `node_modules` but also when importing the local `./ophan`.

Even more strangely, on investigation, it seems to also match a module called `ooooooooophan` or pretty much anything similar. 

This solution works - but it would be good to find out why jest is working in this way and if we can make it more explicit so this doesn't sting someone in the future.

The docs don't say that it should work this way, but also doesn't say it shouldn't... 🤔

jamesgorrie (commented):
Actually - re-looking at this - [I think we might be able to have this set more explicitly here](https://github.com/guardian/support-frontend/blob/main/support-frontend/package.json#L63) - seems weird that we're pattern matching

rupertbates:
> Actually - re-looking at this - [I think we might be able to have this set more explicitly here](https://github.com/guardian/support-frontend/blob/main/support-frontend/package.json#L63) - seems weird that we're pattern matching

Oh yes good spot, that is weird

rupertbates:
> Actually - re-looking at this - [I think we might be able to have this set more explicitly here](https://github.com/guardian/support-frontend/blob/main/support-frontend/package.json#L63) - seems weird that we're pattern matching

I've tried making this just ophan with no pattern match and the behaviour is still the same. I'm going to go with this solution for now as it works and allows me to move forward.

PR #6337, opened 2024-09-23
One Time Checkout - Thankyou Page

paul-daniel-dempsey:

- Current `thank-you.tsx` page modified to display One-Off messages based upon: URLParams-
product=Contrubution
{contribution=<amount>} _optional_
- Regular Contributions will require an additional ratePlan.
- Feeback help module requires email, Do we have to assume it equates to signed-in ONLY? Because GPDR does not allow us to not store email in session storage going forwards?

[**Trello Card**](https://trello.com/c/iL56a1mO/1085-generic-checkout-one-time-thank-you-page)

## Screenshots

Original 
|Guest|New|Existing|
|-----|-----|-----|
|image|image|image|

New 
|Guest|New|Existing|
|-----|-----|-----|
|image|image||


support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx
```diff
@@ -286,6 +287,21 @@ function OneTimeCheckoutComponent({
...
+						firstName: '',
+						paymentMethod: paymentMethod,
+					};
+					setThankYouOrder(order);
+
+					/*  Contribution product reused for both
+              OneOff (excl ratePlan)
+              Recurring (incl ratePlan) */
+					const thankYouUrlSearchParams = new URLSearchParams();
+					thankYouUrlSearchParams.set('product', 'Contribution');
```
Comment (jamesgorrie) [inline support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx:301]: This is reserved for when the product is actually `Contribution` as per the [Product API](https://product-catalog.guardianapis.com/product-catalog.json). 

We shouldn't try to hijack this.

support-frontend/assets/pages/[countryGroupId]/thank-you.tsx
```diff
@@ -105,30 +105,34 @@ export function ThankYou({ geoId, appConfig }: Props) {
...
-	const ratePlanKey =
+	const ratePlanKeyOneOff =
+		productKey === 'Contribution' && !ratePlanParam ? 'OneOff' : undefined;
+	const ratePlanKeyRecurring =
 		ratePlanParam && ratePlanParam in product.ratePlans
 			? ratePlanParam
 			: undefined;
-	const ratePlan = ratePlanKey && product.ratePlans[ratePlanKey];
+	const ratePlanKey = ratePlanKeyOneOff ?? ratePlanKeyRecurring;
+	const ratePlan = ratePlanKey && (product.ratePlans[ratePlanKey] ?? 'OneOff');
```
Comment (jamesgorrie) [inline support-frontend/assets/pages/[countryGroupId]/thank-you.tsx:117]: We should avoid hijacking this model to put `OneOff` in - as mentioned above, these are directly relating to the [Product API](https://product-catalog.guardianapis.com/product-catalog.json).

Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/thank-you.tsx:117]: I think there were hints on this approach in the original `thank-you.tsx` hence went down that path.  

Agree it's not ideal but seemed better approach than creating completely new thankyou page simple-copy just for `OneOff`. 

This PR can't be progressed without some sort of assumption on how to get to `OneOff`/`OneTime`/`1-Time`. Possibly I can finish the associated Storybook/Playwright test's (which will need) and leave it at that until our meeting decision - then revisit this PR (unless it needs to change drastically). Or it goes in and we revise afterwards?

support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx
Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx:301]: Agree, thats why a description was placed there to highlight the assumption. Hijack is not ideal. I do not think product=contributiom with no RatePlan combination is used in the code-base hence it was selected for ease of fitting in rather reflecting the setup. Meeting is need to progress this further would say however.

support-frontend/assets/pages/[countryGroupId]/thank-you.tsx
Comment (andrewHEguardian) [reply support-frontend/assets/pages/[countryGroupId]/thank-you.tsx:117]: How about we have two different components that simply work out which thank you modules to show (one which knows about ratePlans etc and one which doesn't) - then both call a shared component which is responsible for rendering them?

Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/thank-you.tsx:117]: Will look at now. To confirm, do you mean two different thankyou components = two separate routes to each from their respective checkout? If so, currently I think as soon as routing is involved there is a requirement to have a (Zuora) product specified in the url however - would need to remove that I think.

Comment (andrewHEguardian) [reply support-frontend/assets/pages/[countryGroupId]/thank-you.tsx:117]: Yes I think we're on the same page. By routing I'm assuming you mean the scala server routes? You're right that expects a product so will have to change

Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/thank-you.tsx:117]: Ok, thats the part will ask George if can pair over, feels like we are lifting a fundamental routing product check however.

PR #6338, opened 2024-09-24
Remove urgent wording from alarms

rupertbates:

Having all alarms starting with the word urgent is making the non-dev members of the team anxious, and I don't think it adds any value as it was just getting prepended to all alarm names.


andrewHEguardian (approved):
URGENT 👍

PR #6342, opened 2024-09-25
Use productPrices consistently across landing/checkout in AppConfig.

jamesgorrie:
Adds the `allProductPrices` object to the `window.guardian` `AppConfig`.

This makes the config less stateful - and relies on the client to calculate which bit of the config it should use.

It also unblocks the work @paul-daniel-dempsey is working on with the one off checkout.

The idea here is to have 1 `AppConfig` that each page can read from consistently. We have already seen on the Landing page that we need both `SupporterPlus` and `TierThree` - so why not just use that model. It's 1 extra lookup, but when you;'re swapping between the pages, you know where to look.

### DIFF
--- a/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
@@ -261,16 +261,10 @@ export function Checkout({ geoId, appConfig }: Props) {
 		}
 
 		/** Get any promotions */
-		const allProductPrices = appConfig.allProductPrices;
-		const productPricesKey =
-			productKey === 'TierThree'
-				? 'tierThreeProductPrices'
-				: productKey === 'SupporterPlus'
-				? 'supporterPlusProductPrices'
+		const productPrices =
+			productKey === 'SupporterPlus' || productKey === 'TierThree'
+				? appConfig.allProductPrices[productKey]
 				: undefined;
-		const productPrices = productPricesKey
-			? allProductPrices[productPricesKey]
-			: undefined;
 
 		/**
 		 * This is some annoying transformation we need from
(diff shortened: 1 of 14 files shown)


support-frontend/assets/pages/supporter-plus-landing/twoStepPages/secondStepCheckout.tsx
```diff
@@ -144,16 +143,6 @@ export function SupporterPlusCheckout({
...
-	/** Promotions on the checkout are for SupporterPlus only for now */
-	const promotion = isSupporterPlus
-		? useContributionsSelector((state) =>
-				getPromotion(
-					state.page.checkoutForm.product.productPrices,
-					countryId,
-					state.page.checkoutForm.product.billingPeriod,
-				),
-		  )
-		: undefined;
```
jamesgorrie, line 156: This has been removed as this checkout is no longer used for anything but one time payments.

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -261,7 +261,10 @@ export function Checkout({ geoId, appConfig }: Props) {
 		}
 
 		/** Get any promotions */
-		const productPrices = appConfig.productPrices;
+		const productPrices =
```
Comment (jamesgorrie) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:264]:
```suggestion
    /** Promos are only available on SupporterPlus and TierThree and we only use this value to determine promotion values */
		const productPrices =
```

andrewHEguardian (approved):
Looks good 👍

paul-daniel-dempsey (approved):
Looks great, productPrice consistency and routing changes will suit my thankyou PR, thanks , cron & smoke tests all pass too.

PR #6344, opened 2024-09-25
Remove redux from landing

jamesgorrie:
This removes all `useContributionsDispatch` and `useContributionsSelector` from the landing page, probably meaning we can implement this page without setting up redux.

I've branched off `remove-server-product-lookup` as that has [a couple changes](https://github.com/guardian/support-frontend/pull/6342/files#diff-9e9d19a3470db4ff48cd4c73865be05d061001afc0751360451d6d509f390e9b) that would be overridden here.

Part of #5722


support-frontend/assets/pages/supporter-plus-landing/twoStepPages/threeTierLanding.tsx
```diff
@@ -308,36 +295,19 @@ export function ThreeTierLanding({
...
-			state.page.checkoutForm.product.allProductPrices.TierThree,
-			countryId,
-			billingPeriod,
-			countryGroupId === 'International' ? 'RestOfWorld' : 'Domestic',
-
-			abParticipations.newspaperArchiveBenefit === undefined
-				? 'NoProductOptions'
-				: 'NewspaperArchive',
-		),
-	);
```
jamesgorrie, line 339: These were just moved down the page closer to where they're being used.

paul-daniel-dempsey (approved):
Nice, removed redux and used contributionType state, great!

PR #6346, opened 2024-09-26
feat: make one time tab selectable

jamesgorrie:
This is an approach, although I have questions around it:
- What if I want the tab selected, but I want the default amount to be used from the `amounts` object?
-


andrewHEguardian:
Completed in #6355

PR #6347, opened 2024-09-26
Only create alarms in PROD

rupertbates:
We don't want to create alarms in code for support-frontend. Previously we had a boolean which meant they were inactive, but there's no point in creating them at all as they just clutter things up, so this PR ensures that alarms are only created in PROD.

### DIFF
--- a/cdk/lib/frontend.ts
+++ b/cdk/lib/frontend.ts
@@ -28,7 +28,7 @@ interface FrontendProps extends GuStackProps {
   membershipSubPromotionsTables: string[];
   domainName: string;
   scaling: GuAsgCapacity;
-  shouldEnableAlarms: boolean;
+  shouldCreateAlarms: boolean;
 }
 
 export class Frontend extends GuStack {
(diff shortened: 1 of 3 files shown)


cdk/lib/frontend.ts
```diff
@@ -124,15 +135,8 @@ export class Frontend extends GuStack {
...
-        http5xxAlarm: {
-          alarmName: alarmName("support-frontend is returning 5XX errors"),
-          alarmDescription: alarmDescription(
-            "Some or all actions on support website are failing"
-          ),
-          actionsEnabled: shouldEnableAlarms,
-          tolerated5xxPercentage: 5,
-        },
-        unhealthyInstancesAlarm: true,
+        http5xxAlarm: http5xxAlarm,
```
andrewHEguardian, line 138: What happens here when the `http5xxAlarm` has value false?

andrewHEguardian (approved):
Nice 👍

cdk/lib/frontend.ts
rupertbates: It just doesn't create one.

johnduffell:
> there's no point in creating them at all as they just clutter things up

...and make it easier to test in CODE

PR #6351, opened 2024-09-27
One-Time Checkout : ApplePay/GooglePay Express Checkout

paul-daniel-dempsey:

Express Checkout Supporting->
- GooglePay
- AmazonPay

[**Trello Card**](https://trello.com/c/xRSG5YAd/1098-generic-checkout-one-time-implement-express-checkout-payment-options-for-one-time-apple-google-pay)

## How to test
- GooglePay : Dev -> Set to `Always`, run Chrome, requires Google account with an attached CCard.
- AmazonPay: Code -> IPhone, requires ApplePay account with an attached CCard.


support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx
```diff
@@ -135,8 +148,20 @@ export function OneTimeCheckout({ geoId, appConfig }: OneTimeCheckoutProps) {
...
 	const stripePromise = loadStripe(stripePublicKey);
 
+	const elementsOptions = {
+		mode: 'payment',
+		/**
+		 * Stripe amounts are in the "smallest currency unit"
+		 * @see https://docs.stripe.com/api/charges/object
+		 * @see https://docs.stripe.com/currencies#zero-decimal
+		 */
+		amount: 100,
```
Comment (andrewHEguardian) [inline support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx:158]: If the user doesn't change their selection will they be charged £1? Maybe can load the default amount here or run one `elements.?update` as soon as the final amount is ready in the component

support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx
```diff
@@ -175,13 +200,40 @@ function OneTimeCheckoutComponent({
...
+			// no onclick?
+			setStripeExpressCheckoutEnable(false);
+		}
+	}, [finalAmount]);
 
 	/** Payment methods: Stripe */
 	const stripe = useStripe();
 	const elements = useElements();
 	const cardElement = elements?.getElement(CardNumberElement);
+	const [, setStripeExpressCheckoutPaymentType] =
```
Comment (andrewHEguardian) [inline support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx:224]: We don't seem to be using this payment type, not sure if we should

andrewHEguardian (approved):
Great!

support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx
Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx]: Agree, its now been set the the minimum at start (but it does immediately use the config default one-time amount). Should finalAmount be undefined (our preferred amount over NaN/0) the ExpressCheckout is disabled in the useEffect (comment update there too). Thanks

support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx
Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx]: I think this might be used to inidicate `ApplePay` amount items within Stripe by checking `stripeExpressCheckoutPaymentType = 'apple-pay` and labelling the stripe paymentMethod as `StripeApplePay`. Annoyingly need to test this on CODE. Perhaps we can my last commit together to see if makes sense tomorrow morning?

PR #6352, opened 2024-09-27
Test the value of S+ partner offers in AU

rupertbates:
We want to test what the value of the 'Exclusive access to partner offers' benefit is in converting AU supporter plus customers, so this PR sets up an AB test - `auPartnerBenefit`:
- `control` - see the copy, as is the case now
- `variant` - do not see the copy
### Landing page in control
Screenshot 2024-09-27 at 14 41 13
### Landing page in variant
Screenshot 2024-09-27 at 14 40 40
### Checkout page in control
Screenshot 2024-09-27 at 14 41 30
### Checkout page in variant
Screenshot 2024-09-27 at 14 41 46

### DIFF
--- a/support-frontend/assets/helpers/productCatalog.tsx
+++ b/support-frontend/assets/helpers/productCatalog.tsx
@@ -84,6 +84,7 @@ export const productCatalogDescription: Record<ProductKey, ProductDescription> =
 					tooltip: `Guardian Weekly is a beautifully concise magazine featuring a handpicked selection of in-depth articles, global news, long reads, opinion and more. Delivered to you every week, wherever you are in the world.`,
 				},
 			],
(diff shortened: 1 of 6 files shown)


support-frontend/assets/helpers/productCatalog.tsx
```diff
@@ -49,6 +51,20 @@ export function filterBenefitByRegion(
 	return true;
 }
 
+export function filterBenefitByABTest(
```
andrewHEguardian, line 54: Very nice! This would be helpful for the newspaper benefits because atm they're bodged in by spreading in the whole product catalog and adding them in

andrewHEguardian (approved):
Great! 🤠 🇦🇺

PR #6354, opened 2024-09-30
Remove redirect to checkout for one-off

tomrf1:
Currently if the landing page url query string contains `selected-contribution-type=ONE_OFF`, we redirect to the checkout.
We no longer want this behaviour as we're about to run A/B tests on this journey.


charleycampbell (approved):
👍

PR #6356, opened 2024-09-30
Keep contribution amount in the redirect for recurring contributions

rupertbates:
For recurring contributions, when we redirect from the old checkout to the new one we want to pass the `selected-amount` query string parameter across to the new checkout renamed to `contribution`. 

This PR does that.


andrewHEguardian (approved):
👍  Looks good to me

PR #6367, opened 2024-10-02
Recaptcha alarm

rupertbates:
This PR does a number of things:
- Adds recaptcha failure to the list of ignored alarms
- Fixes handling of error responses from the `/create` endpoint (CreateSubscriptionController.create) - previously these were not being handled, leaving the user with a never ending spinner
- Improve the logging of Identity errors to make it clear that they are from Identity

### DIFF
--- a/support-frontend/app/actions/CustomActionBuilders.scala
+++ b/support-frontend/app/actions/CustomActionBuilders.scala
@@ -67,19 +67,25 @@ class CustomActionBuilders(
       AwsCloudWatchMetricPut(AwsCloudWatchMetricPut.client)(cloudwatchEvent)
     }
 
+    private def isNon200Result(result: Result) = result.header.status.toString.head != '2'
+
     private def maybePushAlarmMetric(result: Result) = {
-      val silentFailureReasons = Set(
+      val ignoreList = Set(
         emailProviderRejectedCode,
         invalidEmailAddressCode,
         recaptchaFailedCode,
       )
-      if (!silentFailureReasons.contains(result.header.reasonPhrase.getOrElse(""))) {
-        logger.error(
-          scrub"pushing alarm metric - non 2xx response. Http code: ${result.header.status}, reason: ${result.header.reasonPhrase}",
-        )
-        pushAlarmMetric
-      } else {
-        logger.info(s"not pushing alarm metric for ${result.header.status} ${result.header.reasonPhrase}")
+      if (isNon200Result(result)) {
+        if (!ignoreList.contains(result.header.reasonPhrase.getOrElse(""))) {
+          logger.error(
+            scrub"pushing alarm metric - non 2xx response. Http code: ${result.header.status}, reason: ${result.header.reasonPhrase}",
+          )
+          pushAlarmMetric
+        } else {
+          logger.info(
+            s"not pushing alarm metric for ${result.header.status} ${result.header.reasonPhrase} as it is in our ignore list",
+          )
+        }
       }
     }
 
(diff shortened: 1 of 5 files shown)


support-frontend/app/controllers/CreateSubscriptionController.scala
```diff
@@ -159,9 +160,9 @@ class CreateSubscriptionController(
       case InvalidEmailAddress(_) => RequestValidationError(invalidEmailAddressCode)
       case OtherIdentityError(message, description, endpoint) =>
         endpoint match {
-          case Some(GuestEndpoint) => ServerError(s"Error calling /guest: $message; $description")
-          case Some(UserEndpoint) => ServerError(s"Error calling /user: $message; $description")
-          case None => ServerError(s"$message: $description")
+          case Some(GuestEndpoint) => ServerError(s"Identity error calling /guest: $message; $description")
+          case Some(UserEndpoint) => ServerError(s"Identity error calling /user: $message; $description")
+          case None => ServerError(s"Error calling Identity: $message: $description")
```
rupertbates, line 165: Make it clear in the logs that Identity is the source of this error

support-frontend/assets/helpers/forms/errorReasons.ts
```diff
@@ -79,6 +80,9 @@ function appropriateErrorMessage(errorReason: string): string {
 
 			case 'invalid_email_address':
 				return 'Please enter a valid email address';
+
+			case 'recaptcha_validation_failed':
+				return 'Please prove you are not a robot';
```
andrewHEguardian, line 85: 🤖

andrewHEguardian (approved):
Great 👍

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -912,20 +914,27 @@ function CheckoutComponent({
...
-				.then((json) => json as StatusResponse);
+			});
+
+			let processPaymentResponse: ProcessPaymentResponse;
+
+			if (createResponse.ok) {
+				const statusResponse = (await createResponse.json()) as StatusResponse;
+				processPaymentResponse = await processPayment(statusResponse, geoId);
+			} else {
+				const errorReason = (await createResponse.text()) as ErrorReason;
```
Comment (paul-daniel-dempsey) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:931]: Do we always assume the error is text, ie is there a situation where an error returned could be json object instead?

support-frontend/app/actions/CustomActionBuilders.scala
```diff
@@ -66,17 +67,27 @@ class CustomActionBuilders(
...
+      )
+      if (isNon200Result(result)) {
+        if (!ignoreList.contains(result.header.reasonPhrase.getOrElse(""))) {
+          logger.error(
+            scrub"pushing alarm metric - non 2xx response. Http code: ${result.header.status}, reason: ${result.header.reasonPhrase}",
+          )
+          pushAlarmMetric
+        } else {
+          logger.info(
+            s"not pushing alarm metric for ${result.header.status} ${result.header.reasonPhrase} as it is in our ignore list",
```
paul-daniel-dempsey, line 86: Perhaps this might be more inline with the log error above?  `not pushing ignored alarm metric - 2xx response. Http code: ${result.header.status}, reason: ${result.header.reasonPhrase}`

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
Comment (rupertbates) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:931]: It looks like the convention on the server is to return a single string code for errors

support-frontend/app/actions/CustomActionBuilders.scala
rupertbates: This would still be a non 200 response, we're just not alarming on it because it's in the ignored list.

PR #6368, opened 2024-10-02
One Time Checkout: pre-selected amount from URL

andrewHEguardian:

Adds pre-selection of an amount card and "other amount" to the new one time checkout.

[**Trello Card**](https://trello.com/c/uSm7lmme/1108-one-time-checkout-url-params)

## Why are you doing this?

For feature parity with the old one time checkout. And for the use case where a user is given the choice cards on the landing page (or possibly an epic). The pre-filled 'other' amount doesn't seem to be a use case at the moment but for parity with the old checkout it seems worth carrying over.
image

## How to test

Go to `uk/one-time-checkout` with no param. See default amount selected.

Go to `uk/one-time-checkout?contribution=30`. See 30 selected.

Go to `uk/one-time-checkout?contribution=35`. See "Other" amount selected and 35 pre-filled.

## Screenshots

image

image

### DIFF
--- a/support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx
@@ -136,6 +136,34 @@ function paymentMethodIsActive(paymentMethod: PaymentMethod) {
 	);
 }
 
+function getPreSelectedAmount(
+	preSelectedAmountParam: string | null,
+	amountChoices: number[],
+): {
+	preSelectedOtherAmount?: string;
+	preSelectedPriceCard?: number | 'other';
+} {
+	const preSelectedAmount = preSelectedAmountParam
+		? parseInt(preSelectedAmountParam, 10)
+		: undefined;
+
+	if (preSelectedAmount === undefined) {
+		return {
+			preSelectedOtherAmount: undefined,
+			preSelectedPriceCard: undefined,
+		};
+	}
+
+	const preSelectedPriceCard = amountChoices.includes(preSelectedAmount)
+		? preSelectedAmount
+		: 'other';
+
+	return {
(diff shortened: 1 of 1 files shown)


support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx
```diff
@@ -197,17 +228,28 @@ function OneTimeCheckoutComponent({
...
+	const { preSelectedPriceCard, preSelectedOtherAmount } = getPreSelectedAmount(
+		preSelectedAmountParam,
+		amounts,
+	);
+
 	const minAmount = config[countryGroupId]['ONE_OFF'].min;
 
-	const [amount, setAmount] = useState<number | 'other'>(defaultAmount);
-	const [otherAmount, setOtherAmount] = useState<string>('');
+	const [selectedPriceCard, setSelectedPriceCard] = useState<number | 'other'>(
```
Comment (paul-daniel-dempsey) [inline support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx:238]: Great changing this naming from `amount` to `selectedPriceCard`, could we use `priceCardAmount`?

paul-daniel-dempsey (approved):
Looks great, works perfectly. Wonder whether its not just pre-selection we should consider. Perhaps we could keep the ParamUrl `contribution` inline with what the user selects on the one-time-checkout too . Perhaps something for a new PR if we actually need.

support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx
Comment (andrewHEguardian) [reply support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx:238]: I wanted to avoid calling it an amount, because that felt confusing when "other" is an option

andrewHEguardian:
>Perhaps we could keep the ParamUrl `contribution` inline with what the user selects on the one-time-checkout too 

That's quite a nice idea, in case you had to refresh the page for example. Let's discuss after

PR #6369, opened 2024-10-02
One Time Checkout - AB Testing

andrewHEguardian:

Adding 

1.  An AB test for the new one time checkout (inactive)
2. Capability for AB tests on the checkout

[**Trello Card**](https://trello.com/c/1s1OSFLd/1039-generic-checkout-one-time-ab-test-setup)

## Why are you doing this?

Once its ready we'll launch an AB test of the new one time checkout vs the old. We also want other AB testing to be possible on the page.

## How to test

Activate AB test, go to landing page and add `uk/contribute#ab-newOneTimeCheckout=variant` to the URL. Click the support just once CTA and see the new checkout.

## Screenshots

Datalake showing labels and AB test participations 
image

### DIFF
--- a/support-frontend/assets/helpers/abTests/abtestDefinitions.ts
+++ b/support-frontend/assets/helpers/abTests/abtestDefinitions.ts
@@ -8,7 +8,7 @@ import type { Tests } from './abtest';
 export const pageUrlRegexes = {
 	contributions: {
 		allLandingPagesAndThankyouPages:
-			'/checkout|contribute|thankyou|thank-you(/.*)?$',
+			'/checkout|one-time-checkout|contribute|thankyou|thank-you(/.*)?$',
 		notUkLandingPage: '/us|au|eu|int|nz|ca/contribute(/.*)?$',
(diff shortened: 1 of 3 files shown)


support-frontend/assets/helpers/abTests/abtestDefinitions.ts
```diff
@@ -147,4 +147,25 @@ export const tests: Tests = {
...
+			ALL: {
+				offset: 0,
+				size: 1,
+			},
+		},
+		isActive: false,
+		referrerControlled: false, // ab-test name not needed to be in paramURL
+		seed: 4,
+		targetPage: pageUrlRegexes.contributions.allLandingPagesAndThankyouPages,
+		excludeCountriesSubjectToContributionsOnlyAmounts: false,
```
GHaberis, line 169: @andrewHEguardian I think `excludeCountriesSubjectToContributionsOnlyAmounts` should be `true` for this test, users who are subject to contributions only amounts see a different checkout flow (the old 2-step checkout), it's easier to exclude them than include them. Including them means we should update their checkout flow to also use the new one-time-checkout if they're in the variant.

PR #6370, opened 2024-10-02
Log recaptcha error codes

rupertbates:
Up until now we haven't been logging any [error codes returned from the Recaptcha service](https://developers.google.com/recaptcha/docs/verify#error_code_reference). As we are seeing regular checkout failures due to Recaptcha errors it makes sense to log these to see if there is a common theme.

### DIFF
--- a/support-frontend/app/services/RecaptchaService.scala
+++ b/support-frontend/app/services/RecaptchaService.scala
@@ -10,7 +10,7 @@ import play.api.libs.ws.WSClient
 
 import scala.concurrent.{ExecutionContext, Future}
 
-case class RecaptchaResponse(success: Boolean, score: Option[BigDecimal])
+case class RecaptchaResponse(success: Boolean, `error-codes`: Option[List[String]])
 object RecaptchaResponse {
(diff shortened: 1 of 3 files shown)


support-frontend/app/services/RecaptchaService.scala
```diff
@@ -10,7 +10,7 @@ import play.api.libs.ws.WSClient
 
 import scala.concurrent.{ExecutionContext, Future}
 
-case class RecaptchaResponse(success: Boolean, score: Option[BigDecimal])
```
rupertbates, line 13: There is no `score` field on the Json response, I'm not sure why this is here.

support-frontend/app/services/RecaptchaService.scala
```diff
@@ -10,7 +10,7 @@ import play.api.libs.ws.WSClient
...
+case class RecaptchaResponse(success: Boolean, `error-codes`: Option[List[String]])
```
rupertbates, line 13: The `error-codes` field is what we need to log.

paul-daniel-dempsey (approved):
Looks Good, let see what comes through

PR #6372, opened 2024-10-03
Log exception on issues that prevent page rendering

GHaberis:

There are certain pieces of core product info that are required in the querystring params in order to render the generic checkout and thank-you page. If they're not present the client renders an almost entirely blank screen with a simple message, which won't mean much to users but indicates the problem.

This PR introduces some Sentry logging so we can be aware of if these issues are occurring, as right now we don't have this visibility.


andrewHEguardian (approved):
Good addition 🪵

paul-daniel-dempsey (approved):
LG

PR #6374, opened 2024-10-03
AB test Amazon Pay

GHaberis:

Amazon Pay is used in the US one-time checkout only. This PR introduces a US only AB test for Amazon Pay as a payment method.

### DIFF
--- a/support-frontend/assets/helpers/abTests/abtestDefinitions.ts
+++ b/support-frontend/assets/helpers/abTests/abtestDefinitions.ts
@@ -168,4 +168,36 @@ export const tests: Tests = {
 		targetPage: pageUrlRegexes.contributions.allLandingPagesAndThankyouPages,
 		excludeCountriesSubjectToContributionsOnlyAmounts: false,
 	},
+	amazonPay: {
+		variants: [
+			{
+				id: 'control',
+			},
+			{
+				id: 'variant',
+			},
+		],
+		audiences: {
+			US: {
+				offset: 0,
+				size: 1,
+			},
+		},
+		canRun: () => {
+			/**
+			 * Only allocate users on old one-time
+			 * checkout to this AB test.
+			 */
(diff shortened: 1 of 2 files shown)


support-frontend/assets/helpers/abTests/abtestDefinitions.ts
```diff
@@ -168,4 +168,36 @@ export const tests: Tests = {
 		targetPage: pageUrlRegexes.contributions.allLandingPagesAndThankyouPages,
 		excludeCountriesSubjectToContributionsOnlyAmounts: false,
```
andrewHEguardian, line 169: Could I piggyback on this PR and add this change in this PR?
https://github.com/guardian/support-frontend/pull/6369#discussion_r1785992806

GHaberis: @andrewHEguardian, yes I did think about just doing it here! I'll commit the change now

andrewHEguardian (approved):
Nice 👍

support-frontend/assets/helpers/abTests/abtestDefinitions.ts
```diff
@@ -168,4 +168,36 @@ export const tests: Tests = {
...
+				id: 'variant',
+			},
+		],
+		audiences: {
+			US: {
+				offset: 0,
+				size: 1,
+			},
+		},
+		canRun: () => {
```
andrewHEguardian, line 186: Good to see this used - we could probably do things like this more. Although maybe speaks to the fact the targetPage is very non-specific?

GHaberis: Yeah it's a handy function to have. Yeah `targetPage` is non-specific which is sufficient most of the time, we could probably revisit `pageUrlRegexes` at some point though, as maybe it could be updated so we can target users on specific checkouts, eg. Tier 1, 2 or 3.

PR #6376, opened 2024-10-03
Supress ProductCatalogService logging

rupertbates:
The ProductCatalogService polls the product catalog api once a minute to retrieve the latest version. As each request logs the request details and the full response, it quickly fills up the logs with identical messages. This can be particularly annoying when running the site locally as log files can quickly reach multiple megabytes in size making them hard to open and scan in an editor.

This PR allows WebServiceHelper derived classes (which ProductCatalogService is) to switch off verbose logging, and does switch them off for the ProductCatalogService.

### DIFF
--- a/support-frontend/app/services/ProductCatalogService.scala
+++ b/support-frontend/app/services/ProductCatalogService.scala
@@ -21,6 +21,7 @@ trait ProductCatalogService extends WebServiceHelper[ProductCatalogServiceError]
   val client: FutureHttpClient
   override val wsUrl: String
   override val httpClient: FutureHttpClient = client
+  override val verboseLogging: Boolean = false
 
   def get(): Future[JsonObject] = {
     get[JsonObject](endpoint = "product-catalog.json")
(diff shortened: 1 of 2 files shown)


support-frontend/app/services/ProductCatalogService.scala
```diff
@@ -21,6 +21,7 @@ trait ProductCatalogService extends WebServiceHelper[ProductCatalogServiceError]
   val client: FutureHttpClient
   override val wsUrl: String
   override val httpClient: FutureHttpClient = client
+  override val verboseLogging: Boolean = false
```
rupertbates, line 24: Switch off verbose logging for this class

support-modules/rest/src/main/scala/com/gu/rest/WebServiceHelper.scala
```diff
@@ -44,6 +44,7 @@ case class CodeBody(code: String, body: String) {
 trait WebServiceHelper[Error <: Throwable] extends SafeLogging {
   val wsUrl: String
   val httpClient: FutureHttpClient
+  val verboseLogging: Boolean = true
```
rupertbates, line 47: Enable classes which inherit this trait to switch off verbose logging

andrewHEguardian (approved):
👍

PR #6381, opened 2024-10-07
Persist landing page test tracking

tomrf1:
Currently, if an AB test is targeted specifically at the landing page then the tracking doesn't carry through to the checkout. This means it doesn't get included in the acquisition event.

This PR fixes this by changing the landing page to add the AB test participations to session storage.
Subsequent pages will then pick up participations from session storage.
This is limited by the new `persistPage` field on the AB test model. A page will only pick up participations from session storage if `persistPage` matches the current path.

### DIFF
--- a/support-frontend/assets/helpers/abTests/abtest.ts
+++ b/support-frontend/assets/helpers/abTests/abtest.ts
@@ -102,6 +102,7 @@ type ABtestInitalizerData = {
 	abTests?: Tests;
 	mvt?: number;
 	acquisitionDataTests?: AcquisitionABTest[];
+	path?: string;
 };
 
 function init({
@@ -111,6 +112,7 @@ function init({
 	abTests = tests,
 	mvt = getMvtId(),
 	acquisitionDataTests = getTestFromAcquisitionData() ?? [],
+	path = window.location.pathname,
 }: ABtestInitalizerData): Participations {
 	const sessionParticipations = getParticipationsFromSession();
 	const participations = getParticipations(
@@ -118,6 +120,7 @@ function init({
 		mvt,
 		countryId,
 		countryGroupId,
+		path,
 		acquisitionDataTests,
 		selectedAmountsVariant,
 		sessionParticipations,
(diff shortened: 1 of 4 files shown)


rupertbates:
Why would we not just target the landing page and checkout page?

tomrf1:
> Why would we not just target the landing page and checkout page?

we don't want people who go straight to checkout to be included in the test - they must go via the landing page. We're sending quite a lot of users straight to checkout now. The existing setup for AB test targeting doesn't really support the 2-step flow correctly

support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -120,6 +122,7 @@ function init({
 	const urlParticipations = getParticipationsFromUrl();
 	const serverSideParticipations = getServerSideParticipations();
 	return {
+		...sessionParticipations,
```
GHaberis, line 125: `sessionParticipations` could be `undefined`, can you use a spread operator on `undefined`? I would've expected TS to complain here 🤔

GHaberis (approved):
Looks good 👍

support-frontend/assets/helpers/abTests/abtest.ts
tomrf1: Good question. It appears you _can_ spread on undefined - and here `serverSideParticipations` and `urlParticipations` are also potentially undefined.
I guess we could be a bit more deliberate about it with `...(sessionParticipations ?? {})`

tomrf1:
DP have approved the use of the session storage item

GHaberis:
> DP have approved the use of the session storage item

Excellent, this looks good to go for me then

support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -154,8 +163,10 @@ function getParticipations(
 	mvtId: number,
 	country: IsoCountry,
 	countryGroupId: CountryGroupId,
+	path: string,
```
tomrf1, line 166: To make this function easier to test, we now pass in a `path` instead of using `window.location.pathname` directly

rupertbates (approved):
This is really nice, well done for providing tests 👍

PR #6385, opened 2024-10-08
Spike: add lighthouse-ci workflow

andrewHEguardian:

Part of a Spike on how we might measure page speed with reports, and alert against budgets. This PR implements https://github.com/treosh/lighthouse-ci-action to monitor our prod site after every release to assert against a set of performance budget.

[**Trello Card**](https://trello.com/c/rZDwH4l3/1102-spike-page-speed-github-actions)

When the assertions fail we can see the result in the Github action and 

image

in the assertions.json file in the artifact 
image


.github/workflows/lighthouse.yml
```diff
@@ -0,0 +1,18 @@
...
+  Lighthouse:
+    if: ${{ github.event.label.name == 'Seen-on-PROD' }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Audit URLs using Lighthouse
+        uses: treosh/lighthouse-ci-action@v12
+        with:
+          urls: |
+            https://support.theguardian.com/uk/contribute
```
rupertbates, line 16: We should have a think about other urls to add here. The subscribe ones for instance.

lighthouserc.json
```diff
@@ -0,0 +1,19 @@
+{
+  "ci": {
+    "assert": {
+      "assertions": {
```
rupertbates, line 4: How did you decide on these assertions?

rupertbates (approved):
Great, really nice and simple 👍

lighthouserc.json
andrewHEguardian: Some were in examples and some are prominent in the lighthouse report. But we can definitely tweak it to what we want
image

paul-daniel-dempsey (commented):
Like this, great job finding out. A lot of settings ned to get familiar with I guess..

PR #4522, opened 2022-12-07
Remove sixForSixSuppression A/B test

i-hardy:
## What are you doing in this PR?

This removes the `sixForSixSuppression` A/B test and makes excluding the six for six offer the default on Guardian Weekly pages.

[**Trello Card**](https://trello.com/c/QoztCe0Z)

## Why are you doing this?

The test was obsolete as 100% of users were in the variant.

PR #6395, opened 2024-10-10
Remove 6 for 6

rupertbates:
6 for 6 was an introductory offer we used to have for Guardian Weekly, where a user would be able to purchase 6 issues for 6 pounds. It had a few problems though which were that it had a very high churn rate when people reached the end of the introductory period, and it was also possible to just cancel and subscribe again with the same offer - this would cost us money as 6 for 6 was a loss making acquisition mechanism. For these reasons 6 for 6 was [discontinued in 2022](https://github.com/guardian/support-frontend/pull/4522).

6 for 6 also added a lot of technical complexity because it was modelled as a new billing period (SixWeekly), and a new type of promotion (IntroductoryPeriod) which required an additional rate plan to be added to the subscription in Zuora. Now it is no longer on offer, we can remove all that complexity.

As a related note, we have since done 12 for 12 promotions, but these were set up using a regular discount promotion on a quarterly rate plan so were much simpler from a technical point of view.


andrewHEguardian (approved):
👏👏 Wow! 🧹 🧹

support-workers/src/test/scala/com/gu/support/workers/lambdas/SendAcquisitionEventSpec.scala
```diff
@@ -24,10 +24,6 @@ class SendAcquisitionEventSpec extends AsyncLambdaSpec with MockContext {
     sendEvent(sendAcquisitionEventPrintJson)
   }
 
-  "SendAcquisitionEvent" should "work with GW 6 for 6 input" taggedAs IntegrationTest in {
-    sendEvent(sendAcquisitionEventGWJson)
-  }
```
johnduffell, line 29: should we not change this to be non 6 for 6 rather than deleting?  The print one is card and this one is/was paypal.

rupertbates: I'll do this in the next PR which will remove the old promo fields

rupertbates: Added back in https://github.com/guardian/support-frontend/pull/6414

PR #6397, opened 2024-10-11
feat: add payment method validation

andrewHEguardian:

Add some validation for when a user hasn't selected a payment method. However due to the way we're using HTML validation, this will only kick in after the other form fields have been validated (by native validation).

[**Trello Card**](https://trello.com/c/TtB5PxAJ/1135-generic-one-time-checkout-payment-validation)

## Why are you doing this?

This validation existed on the old checkout - and provides a better user experience if someone tries to press the button without having selected a payment radio.

## How to test

To go generic or one time checkout, fill out personal details. Do not select a payment method. Click payment CTA. See "please select a payment method" validation.

## Screenshots
image


support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -1080,8 +1085,6 @@ function CheckoutComponent({
 			</Box>
 			<form
 				ref={formRef}
-				action="/contribute/recurring/create"
```
Comment (andrewHEguardian) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:1083]: This action is not used because we `event.preventDefault` in the onSubmit. So having this (which is actually a different URL to what we're posting to in the `formOnSubmit` function) felt confusing

Comment (GHaberis) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:1083]: Good call 👍

support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx
```diff
@@ -342,6 +343,10 @@ function OneTimeCheckoutComponent({
 		if (finalAmount) {
 			setIsProcessingPayment(true);
 
+			if (paymentMethod === 'None') {
```
Comment (GHaberis) [inline support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx:346]: Out of interest is there a reason why the equivalent check on the generic checkout checking `paymentMethod === undefined` whilst this is `'None'` for `paymentMethod`? 🤔

Comment (andrewHEguardian) [reply support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx:346]: There's a slight inconsistency, the generic makes the default `undefined` in the setState whereas the one-time sets it as `None`

Comment (GHaberis) [reply support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx:347]: Makes sense, possibly another point to add to the improving maintainability of the checkouts list, consistency of common types.

GHaberis (approved):
nice 👍

paul-daniel-dempsey (commented):
Nice and neat paymentMethodError display

PR #6400, opened 2024-10-11
Validate Promotions Using Region (CountryGroup) Not Billing Country

rupertbates:
Up until now we have validated promotions which are applied to a purchases to check that they are valid in the country which the user is from. 

There is a problem with this however because we have been using the country from the billing address of the customer, and when this country is set from a payment provider such as Apple Pay or Google Pay it does not necessarily match to the region where the user is purchasing the product.

The solution to this is to validate promotions against the region (country group) which the user has selected on the support site as this determines the price and currency of the product they are purchasing.

Screenshot 2024-10-14 at 14 52 23

#### Region (country group) on the support site

This PR adds a new field called `appliedPromotion` to the json body posted to the create subscription endpoint, and passes this through the support-workers state to the promo validation code. 

```Typescript
export type AppliedPromotion = {
	promoCode: string;
	countryGroupId: SupportInternationalisationId; 
};
```
This is optional in all places, but where it exists we will have both a promo code and the checkout region which it was used on, allowing us to validate using that rather than the billing address.

### TODO
This PR has left the original `promoCode` field in place to handle compatibility when deploying the front end and back end. After this has been merged `promoCode` can be removed.

### Related Trello cards
- https://trello.com/c/ndSqCsCG/1138-pass-country-group-to-support-workers-and-use-it-to-validate-promo-code
- https://trello.com/c/8CQcIybD/1016-payment-express-button-prevent-billing-country-vs-billing-region-mismatch-with-client-side-validation
- https://trello.com/c/KLYpxRru/1041-promocode-validation-from-client-to-support-workers-apple-google-pay


support-frontend/assets/helpers/internationalisation/classes/countryGroup.ts
```diff
@@ -126,10 +126,6 @@ export default class CountryGroupHelper {
 		);
 	}
 
-	static stringToCountryGroupId(countryGroupId: string): CountryGroupId {
```
rupertbates, line 129: This was unused

support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
```diff
@@ -211,12 +211,6 @@ class DigitalPackEmailFields(
       state: SendThankYouEmailDigitalSubscriptionGiftPurchaseState,
   )(implicit ec: ExecutionContext) = {
     import state._
-
```
rupertbates, line 214: I noticed that the applied promotion is never used by the email fields generator so I removed the code related to it.

andrewHEguardian (approved):
Looks good to me 👍 should clear up some alarms, hopefully this validation doesn't ever fail in practice!

rupertbates:
> Looks good to me 👍 should clear up some alarms, hopefully this validation doesn't ever fail in practice!

Yeah it really shouldn't without something weird happening or intentional fiddling.

PR #6412, opened 2024-10-15
Newspaper archive - SSR custom header 

rupertbates:

Creating a new server side rendered page to eventually be iframed on to theguardian.newspapers.com as the custom header.

[**Trello Card**](https://trello.com/c/pr57zkkZ/1099-customise-newspaper-archivecom-header)

## Why are you doing this?

Currently on there is a header that exists as an interactive. We want to create a new version of this, and have it easier to maintain, so we are hosting it on the support site.

## How to test

Run `yarn build-ssr` if locally 
Go to `/newspaper-archive-header` and see the hello world SSR rendered page at the moment.

## Screenshots
image

## Next steps: 
Combine the two scripts into one, this would make it easier to add SSR rendered content in future without duplicating the `tsx` scripts.
Develop the actual content of the header.


GHaberis (approved):
Look good 🚀

support-frontend/app/views/newspaperArchive.scala.html
```diff
@@ -0,0 +1,23 @@
+@import views.{EmptyDiv, ReactDiv, SSRContent}
+
+@(
+  title: String,
```
GHaberis, line 4: I suppose passing title from the controller in isn't necessary, we could just hard code it to "Newspaper Archive" in the HTML template. Nota. blocking issue though.

support-frontend/app/views/newspaperArchive.scala.html
```diff
@@ -0,0 +1,23 @@
+@import views.{EmptyDiv, ReactDiv, SSRContent}
```
GHaberis, line 1: Do we still need these imports now we've simplified the template somewhat? 🤔

PR #6414, opened 2024-10-15
Remove old promo field

rupertbates:
This PR follows on from #6400 which added a new field `appliedPromotion` to various objects to replace the `promoCode` field which was there before. 

To ensure there were no compatibility issues when releasing various parts of the application, the previous PR left both fields in place, this one removes the old field.


support-workers/src/main/scala/com/gu/zuora/productHandlers/ZuoraDigitalSubscriptionDirectHandler.scala
```diff
@@ -41,7 +41,7 @@ class ZuoraDigitalSubscriptionDirectHandler(
       state.product,
       state.paymentMethod,
       paymentSchedule,
-      state.promoCode,
+      state.appliedPromotion.map(_.promoCode),
```
rupertbates, line 44: Later states in the support-workers step function still have the `promoCode` field. This is because we have already done the necessary validation by this point so we don't need the full `appliedPromotion` fields any more.

andrewHEguardian (approved):
Looks good 👍

PR #6318, opened 2024-09-16
Source and source development kitchen updates

charleycampbell:
## What are you doing in this PR?
Updating Source to V8. This has breaking changes so anything which is broken/deprecated has been updated.
[**Trello Card**](https://trello.com/c/U8uXI3u8/678-upgrade-source-in-support-frontend)

## How to test

- approved chromatic changes, and signed off by design on any small diffs (TBC)
- Deployed onto code and ran the playwright tests which did pass after some minor tweaks to the code locally (like the url)
- Created a handful of subs for different regions and products.

## Have we considered potential risks?
There is a risk that an edge case is missed in testing. If that is the case we will need to roll this change back to investigate.

## Accessibility test checklist

-  [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-  [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-  [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-  [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

PR #6416, opened 2024-10-15
refactor: remove Checkbox workaround

andrewHEguardian:

Removing a workaround introduced in #6302. This prevented a spacing issue on the checkbox, which was had been fixed in the latest Source but we had no clear upgrade path. Thanks to #6318 🎉  we are now up to date so we should remove this hack


paul-daniel-dempsey (approved):
Nice :)

PR #6426, opened 2024-10-17
[Fix Scala Warnings] - Add explicit types to implicits

rupertbates:
This PR is the first of potentially several to fix Scala warnings in the support-frontend codebase, it contains two main changes:
- Add [Scalafix](https://scalacenter.github.io/scalafix/) to the project to enable a number of automated fixes
- Use Scalafix to carry out the [ExplicitResultTypes](https://scalacenter.github.io/scalafix/docs/rules/ExplicitResultTypes.html) fix to add explicit types to all implicit vals in the project

### DIFF
--- a/.scalafix.conf
+++ b/.scalafix.conf
@@ -2,3 +2,4 @@ rules = [
  ExplicitResultTypes,
 ]
 ExplicitResultTypes.onlyImplicits = true
+ExplicitResultTypes.skipSimpleDefinitions = false
--- a/support-payment-api/src/main/scala/MyApplicationLoader.scala
+++ b/support-payment-api/src/main/scala/MyApplicationLoader.scala
@@ -92,7 +92,7 @@ class MyComponents(context: Context)
       .buildRequestBasedProvider(requestEnvironments)
       .valueOr(throw _)
 
-  implicit val allowedCorsUrl = configuration.get[Seq[String]](s"cors.allowedOrigins").toList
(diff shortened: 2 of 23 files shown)


.scalafix.conf
```diff
@@ -2,3 +2,4 @@ rules = [
  ExplicitResultTypes,
 ]
 ExplicitResultTypes.onlyImplicits = true
+ExplicitResultTypes.skipSimpleDefinitions = false
```
rupertbates, line 5: This config tells Scalafix to only carry out the fix on implicits but to do it even for simple val declarations, which it won't do by default.

support-payment-api/src/main/scala/MyApplicationLoader.scala
```diff
@@ -92,7 +92,7 @@ class MyComponents(context: Context)
       .buildRequestBasedProvider(requestEnvironments)
       .valueOr(throw _)
 
-  implicit val allowedCorsUrl = configuration.get[Seq[String]](s"cors.allowedOrigins").toList
+  implicit val allowedCorsUrl: List[String] = configuration.get[Seq[String]](s"cors.allowedOrigins").toList
```
johnduffell, line 95: implicit `List[String]` is a brave one

johnduffell (approved):
looks good, not as much as I expected to see though

rupertbates:
> looks good, not as much as I expected to see though

Yeah I thought there might be more. This is just implicits though, there will be a lot more for public members

PR #6429, opened 2024-10-18
Omit users on the Contributions Only checkout from the coverTransactionCost test

GHaberis:

We have a  "Contributions Only" checkout flow for some countries: eg. https://support.theguardian.com/int/contribute?country=OM

We shouldn't be bucketing users on this "Contributions Only" checkout flow in the `coverTransactionCost` A/B/C test because this checkout doesn't show the one-time choice cards component on the checkout page, which means we users allocated to `variantB` will never see the cover transaction component as it sits in this component.


paul-daniel-dempsey (approved):
Ah, sorry I missed that, thanks George.

GHaberis:
> Ah, sorry I missed that, thanks George.

It's OK, we had it set up like this on the original test too but it wasn't an issue for that test as we'd positioned the cover transaction component at the bottom, so it appeared correctly for the `variant` on this flow too.

PR #6433, opened 2024-10-21
Landing page one-time tab test (v2)

tomrf1:
This is a re-run of a test we tried recently.
Only users on the US landing page enter this test.
Users in the variant see the one-time tab.


support-frontend/assets/helpers/abTests/abtestDefinitions.ts
```diff
@@ -9,10 +9,7 @@ export const pageUrlRegexes = {
 	contributions: {
 		allLandingPagesAndThankyouPages:
 			'/checkout|one-time-checkout|contribute|thankyou|thank-you(/.*)?$',
-		notUkLandingPage: '/us|au|eu|int|nz|ca/contribute(/.*)?$',
```
tomrf1, line 12: I think all these pre-date the 2-step checkout, and so the name "landing page" is now incorrect

rupertbates (approved):
👍

PR #6435, opened 2024-10-22
Only throw missing Billing State error if required by checkout's country group

GHaberis:

We're logging an error in Sentry if a user pays with a wallet payment method, but that payment method does not have a Billing State associated to it. Currently we throw this error in all country group id regions, however it's only an issue we care about in countries that require Billing State, eg. US, CA, AU. 

This PR updates the code that logs this issue, so we only report in the US, CA, AU. This will give us a better idea of how many users are experiencing an issue due to the missing Billing State.

Note: The objective here is to get a clearer picture of how many users are affected by a known issue where they're in the US, CA, AU but the Billing Address associated with their "wallet" payment method does not have a Billing State. Although we send `countryId` to `logException` this info is impossible to find/filer on in Sentry's UI (I've tried!), otherwise I could have used that to help estimate the impact of this issue.


andrewHEguardian (approved):
👍

PR #6437, opened 2024-10-22
Enable Link Express Checkout AB test 

andrewHEguardian:

Creating the proper AB test of the Link payment method in the Stripe express checkout method, labelling the acquisitions which use this method.

[**Trello Card**](https://trello.com/c/aHoB0Ihf/1104-enabling-link-with-the-stripe-express-checkout-element)

## Why are you doing this?

Testing if this new payment method increases conversion. We have tested this in production with a real payment and it worked.

## How to test

Go to `uk/checkout?product=Contribution&ratePlan=Monthly&contribution=3#ab-linkExpressCheckout=variant`. The acquisition is labelled with "express-checkout-link"

CVR

## Risks

This did not work in our Australian Stripe account in test mode - so I have not enabled it there and Link will not be available in AUD territories.

## Screenshots
Control

image

Variant
image

(other variations exist when paired with other express checkout methods available)
image

### DIFF
--- a/support-frontend/assets/helpers/abTests/abtestDefinitions.ts
+++ b/support-frontend/assets/helpers/abTests/abtestDefinitions.ts
@@ -178,10 +178,21 @@ export const tests: Tests = {
 			},
 		],
 		audiences: {
-			ALL: {
+			UnitedStates: {
 				offset: 0,
 				size: 1,
 			},
+			GBPCountries: {
+				offset: 0,
+				size: 1,
+			},
+			EURCountries: {
+				offset: 0,
+				size: 1,
+			},
+			Canada: { offset: 0, size: 1 },
+			NZDCountries: { offset: 0, size: 1 },
+			International: { offset: 0, size: 1 },
 		},
 		isActive: false,
 		referrerControlled: false, // ab-test name not needed to be in paramURL
--- a/support-frontend/assets/helpers/abTests/abtestDefinitions.ts
+++ b/support-frontend/assets/helpers/abTests/abtestDefinitions.ts
@@ -10,6 +10,7 @@ export const pageUrlRegexes = {
 		allLandingPagesAndThankyouPages:
 			'/checkout|one-time-checkout|contribute|thankyou|thank-you(/.*)?$',
 		usLandingPageOnly: '/us/contribute$',
+		genericCheckoutOnly: '/checkout|thank-you(/.*)?$',
 	},
 	subscriptions: {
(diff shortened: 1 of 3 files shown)


support-frontend/assets/helpers/abTests/abtestDefinitions.ts
```diff
@@ -178,9 +178,9 @@ export const tests: Tests = {
 			},
 		],
 		audiences: {
-			GBPCountries: {
+			ALL: {
```
GHaberis, line 181: This will pick-up AUS too, is the Link button working for `GNM Membership AUS` in PROD? If not we should only include regions using `GNM Membership`

GHaberis:
I wonder if we need a way of specifying the `targetPage` should only be the generic checkout / generic thank-you page, `.allLandingPagesAndThankyouPages` will bucket users on the landing pages and old-non generic (one-time) checkout.

support-frontend/assets/helpers/abTests/abtestDefinitions.ts
andrewHEguardian: It's not enabled in the payment method settings so it won't show up

GHaberis: But we'll still be assigning users in Australia to the A/B test despite users in the `variant` never seeing the button (as it's not enabled), so that could skew the results against the `variant`, the impact in reality will probably be negligible however if we know certain users will never experience the variant experience then we should exclude them from the test.

support-frontend/assets/helpers/abTests/abtestDefinitions.ts
```diff
@@ -10,6 +10,7 @@ export const pageUrlRegexes = {
 		allLandingPagesAndThankyouPages:
 			'/checkout|one-time-checkout|contribute|thankyou|thank-you(/.*)?$',
 		usLandingPageOnly: '/us/contribute$',
+		genericCheckoutOnly: '/checkout|thank-you(/.*)?$',
```
GHaberis, line 13: I have a feeling this will also pick up the existing one-time checkout at `/contribute/checkout`, updating to `'(uk|ca|eu|nz|int)/checkout|thank-you(/.*)?$'` should fix that? I believe we have unit tests in `abtestTest.ts` for validating these regexes, could be handy adding some to confirm.

andrewHEguardian: Great spot!

GHaberis (approved):
Looks good to me 👍

PR #6439, opened 2024-10-24
New One Time - use old type for stripe payment method

andrewHEguardian:

Change the stripe payment method passed to the backend to the old (inaccurate) value of 'stripePaymentRequestButton'. Our backend is expecting this
<!--


GHaberis (approved):
Looks good

PR #6259, opened 2024-08-27
Updated M25 postcode list

graham228221:
## What are you doing in this PR?
Aligns the postcodes defined as "inside M25" for the Newspaper Home Delivery product to the list of postcodes provided by Customer Operations.

Also combines the previous two lists into one list, as previously suggested [here](https://github.com/guardian/support-frontend/pull/4117#issuecomment-1214917910). It's a bit difficult to maintain two separate lists, and [the way postcodes are validated](https://github.com/guardian/support-frontend/blob/f3d660a2b7c91c02c3c5f60cb437bf1310cd5af4/support-frontend/app/utils/CheckoutValidationRules.scala#L400) should mean that having every variation of postcode area + subdistrict is not required (and actually is more risky - DA22 does not equal DA2 2!).

## Why are you doing this?

Latest "Inside M25" delivery list from Customer Operations:

Screenshot 2024-08-27 at 14 33 22

## Accessibility test checklist

-  [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-  [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-  [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-  [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

PR #6446, opened 2024-10-28
Remove GU postcodes from "inside M25" delivery area

graham228221:

Removes all GU postcodes from the inside M25 list, as our Head of Supply Chain confirms that they shouldn't be served as "Home Delivery":

> yes, all those GU postcodes listed can be removed.
> 
> GU11, GU12, GU13, GU14, GU21, GU24, GU27, GU29 shouldn't link to the Home Delivery product.


graham228221:
Doing this as a quick PR to unblock a customer in the contact centre - I will finish off the bigger/better version here once we finally get confirmation on the actual list: https://github.com/guardian/support-frontend/pull/6259

tjmw (approved):
👍

PR #6447, opened 2024-10-28
Gh patch gu postcodes

graham228221:
Partial reversion of #6446 - our Supply Chain colleagues have realised that they do want to keep the GU1, GU2, GU21 and GU22 postcodes under the "Inside M25" area.


tjmw (approved):
👍

PR #6449, opened 2024-10-28
Support workers CDK

rupertbates:
This PR replaces the existing home rolled way of generating the cloudformation for support-workers with CDK.

This is more standard and a lot simpler generally as the previous approach involved messing about with handlebar templates. It has been successfully tested on CODE.


cdk/bin/cdk.ts
```diff
@@ -102,3 +103,43 @@ new BigqueryAcquisitionsPublisher(app, "BigqueryAcquisitionsPublisher-PROD", {
...
+
+new SupportWorkers(app, "SupportWorkers-CODE", {
+  stack: "support",
+  stage: "CODE",
+  promotionsDynamoTables: [
+    "arn:aws:dynamodb:*:*:table/MembershipSub-Promotions-CODE",
+  ],
+  s3Files: [
+    "arn:aws:s3:::gu-zuora-catalog/PROD/Zuora-CODE/catalog.json",
+    "arn:aws:s3:::support-workers-private/*",
```
johnduffell, line 115: I assume this is just copied from elsewhere but it does seem a little generous

tomrf1 (approved):
👍 looks good

cdk/lib/support-workers.ts
```diff
@@ -0,0 +1,236 @@
...
+        initialPolicy: [
+          s3Policy,
+          cloudWatchLoggingPolicy,
+          ...additionalPolicies,
+        ],
+      });
+      this.overrideLogicalId(lambda, {
+        logicalId: lambdaId,
+        reason: "Moving existing lambda to CDK",
+      });
```
johnduffell, line 121: that's a clever trick, it's always tricky to get around this kind of auto generated IDs thing normally.  I assume this is a permanent override though?
```suggestion
      this.overrideLogicalId(lambda, {
        logicalId: lambdaId,
        reason: "Retain pre-CDK logical ID",
      });
```

cdk/lib/support-workers.ts
```diff
@@ -0,0 +1,236 @@
...
+      this,
+      "ShouldClonePaymentMethodChoice"
+    );
+    const accountExists = Condition.booleanEquals(
+      "$.requestInfo.accountExists",
+      true
+    );
+    const checkoutSuccess = new Succeed(this, "CheckoutSuccess");
+
+    const parallelSteps = new Parallel(this, "Parallel")
```
johnduffell, line 207:
```suggestion
    const postAcquisitionSteps = new Parallel(this, "Parallel")
```

cdk/lib/support-workers.ts
```diff
@@ -0,0 +1,236 @@
...
+      .branch(sendThankYouEmail)
+      .branch(updateSupporterProductData)
+      .branch(sendAcquisitionEvent)
+      .branch(checkoutSuccess);
+
+    const commonSteps = createZuoraSubscription.next(parallelSteps);
```
johnduffell, line 213: common in what way?

cdk/lib/support-workers.ts
```diff
@@ -0,0 +1,236 @@
...
+
+    const stepsWithExistingAccount =
+      preparePaymentMethodForReuse.next(commonSteps);
+
+    const stepsForNewAccount = createPaymentMethodLambda
+      .next(createSalesforceContactLambda)
+      .next(commonSteps);
+
+    const startPoint = shouldClonePaymentMethodChoice
```
johnduffell, line 222: slightly more mathematic perhaps?
```suggestion
    const initialState = shouldClonePaymentMethodChoice
```

cdk/lib/support-workers.ts
```diff
@@ -0,0 +1,236 @@
...
+      .otherwise(stepsForNewAccount);
+
+    const stateMachine = new StateMachine(this, "SupportWorkers", {
+      stateMachineName: `${app}-${this.stage}`,
+      definitionBody: DefinitionBody.fromChainable(startPoint),
+    });
+
+    this.overrideLogicalId(stateMachine, {
+      logicalId: `SupportWorkers${this.stage}`,
+      reason: "Moving existing step function to CDK",
```
johnduffell, line 233: this is the ID used in support frontend to find the state machine isn't it?

cdk/lib/support-workers.ts
```diff
@@ -0,0 +1,236 @@
...
+    const stepsForNewAccount = createPaymentMethodLambda
+      .next(createSalesforceContactLambda)
+      .next(commonSteps);
+
+    const startPoint = shouldClonePaymentMethodChoice
+      .when(accountExists, stepsWithExistingAccount)
+      .otherwise(stepsForNewAccount);
+
+    const stateMachine = new StateMachine(this, "SupportWorkers", {
+      stateMachineName: `${app}-${this.stage}`,
```
johnduffell, line 227: this is used by support frontend to find the state machine I think?  might be worth a comment, now we're blessed with that ability
```suggestion
      stateMachineName: `${app}-${this.stage}`,// used by support-frontend to find the state machine
```

cdk/lib/support-workers.ts
```diff
@@ -0,0 +1,236 @@
...
+      definitionBody: DefinitionBody.fromChainable(startPoint),
+    });
```
johnduffell, line 229: can't see the overall (24h) timeout anywhere.

cdk/lib/support-workers.ts
```diff
@@ -0,0 +1,236 @@
...
+        .addRetry({
+          errors: [Errors.ALL], // Wildcard to capture any error which originates from outside of our code (e.g. an exception from AWS)
+          maxAttempts: 999999,
+          interval: Duration.seconds(3),
+          backoffRate: 2,
+        });
+    };
+
+    const checkoutFailure = new Pass(this, "CheckoutFailure"); // This is a failed execution we don't want to alert on
+    const failState = new Fail(this, "FailState"); // This is a failed execution we do want to alert on
```
johnduffell, line 151: I assume the alarms are still going to be defined in the existing CFN for now

support-workers/scripts/update-code-cloudformation.sh
```diff
@@ -1,24 +1,17 @@
 #!/bin/bash
 
-./build-cloudformation.sh
```
johnduffell, line 3: do we need to run pnpm build or similar at this point?

cdk/lib/support-workers.ts
rupertbates: Steps that are carried out for all users whether they have an existing account or not

johnduffell: postAccountExistsSteps ?

cdk/lib/support-workers.ts
rupertbates: Great catch, I'd missed the alarms off. Will add them.

cdk/lib/support-workers.ts
```diff
@@ -232,5 +242,304 @@ export class SupportWorkers extends GuStack {
       logicalId: `SupportWorkers${this.stage}`,
       reason: "Moving existing step function to CDK",
     });
+
+    // Alarms
```
johnduffell, line 246: would it make sense to split it out into several files rather than relying on comments? or is that not the done thing for CDK?

cdk/lib/support-workers.ts
```diff
@@ -232,5 +242,304 @@ export class SupportWorkers extends GuStack {
...
+    //  and product in ('RECURRING_CONTRIBUTION', 'PRINT_SUBSCRIPTION', 'SUPPORTER_PLUS')
+    //  )
+    //  select
+    //  product,
+    //  print_product,
+    //  payment_provider,
+    //  max(timestamp_diff(event_timestamp, last_acquisition, HOUR)) max_diff,
+    //  from timegaps
+    //  group by 1,2,3
+    //  order by 1,2,3
```
johnduffell, line 312: can we change this to the /* .... */ comments, then it will be easier to copy and paste into bigquery?

cdk/lib/support-workers.ts
```diff
@@ -232,5 +242,304 @@ export class SupportWorkers extends GuStack {
...
+          Stage: this.stage,
+        },
+        statistic: "Sum",
+        period: Duration.seconds(3600),
+      }),
+      comparisonOperator: ComparisonOperator.LESS_THAN_OR_EQUAL_TO_THRESHOLD,
+      evaluationPeriods: 4,
+      treatMissingData: TreatMissingData.BREACHING,
+      threshold: 0,
+    }).node.addDependency(stateMachine);
```
johnduffell, line 333: why does this depend on stateMachine but the above ones don't?

cdk/lib/support-workers.ts
```diff
@@ -232,5 +242,304 @@ export class SupportWorkers extends GuStack {
...
+          PaymentProvider: "PayPal",
+          ProductType: "Contribution",
+          Stage: this.stage,
+        },
+        statistic: "Sum",
+        period: Duration.seconds(3600),
+      }),
+      comparisonOperator: ComparisonOperator.LESS_THAN_OR_EQUAL_TO_THRESHOLD,
+      evaluationPeriods: 4,
+      treatMissingData: TreatMissingData.BREACHING,
```
johnduffell, line 331: I assume this is how the alarm was already set up rather than a change you made, but actually this will take a fair bit longer than 3600s * 4 (4 hours) to go off, as when there is missing data it looks back to find more data earlier.  The usual solution is to use a calculated metric and use `FILL(m1, 0)` as the expression, where `m1` is the actual PaymentSuccess metric.

johnduffell: see example on line 453 (although that also SUMs several metric which is not needed here)

cdk/lib/support-workers.ts
rupertbates: I think it would make it more difficult to get a full picture of what's going on

support-workers/scripts/update-code-cloudformation.sh
rupertbates: Yes, it's running `yarn synth`

johnduffell: I must be blind it looks like
`./build-cloudformation.sh`
has been replaced with  a comment
`# We need to upload the template to S3 because it is too big to inline`

cdk/lib/support-workers.ts
```diff
@@ -0,0 +1,236 @@
...
+          ...additionalPolicies,
+        ],
+      });
+      this.overrideLogicalId(lambda, {
+        logicalId: lambdaId,
+        reason: "Moving existing lambda to CDK",
+      });
+      return new LambdaInvoke(this, lambdaName, {
+        lambdaFunction: lambda,
+        outputPath: "$.Payload", // Without this, LambdaInvoke wraps the output state as described here: https://github.com/aws/aws-cdk/issues/29473
```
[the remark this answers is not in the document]
rupertbates: I couldn't find a way to do that with the api and this approach seems fine.

support-workers/scripts/update-code-cloudformation.sh
rupertbates: Realised I didn't push this change!

PR #6450, opened 2024-10-29
End 3-tab LP test

tomrf1:
The variant with the one-time tab has performed well.
Now all users in the US EOY campaign will see it.
When the campaign ends, it will go back to normal (no one-time tab)

Screenshot 2024-10-29 at 09 35 15


charleycampbell (approved):
Excellent!

PR #6454, opened 2024-10-29
Newspaper Archive : Thankyou cards re-sort, display missing benefit

paul-daniel-dempsey:

Thankyou cards re-sort, display missing benefit

1.  re-sort
2. missing benefit display

[**Trello Card**](https://trello.com/c/FjKV8faD/1054-update-copy-and-imagery-newspaper-archive-benefit-to-lp)

## How to test

dev => `https://support.thegulocal.com/...`
prod => `https://support.theguardian.com/...`

`...uk/checkout?product=TierThree&ratePlan=DomesticMonthlyV2#ab-newspaperArchiveBenefit=control`
`...uk/thank-you?product=TierThree&ratePlan=DomesticMonthlyV2#ab-newspaperArchiveBenefit=control`

eg prod
`https://support.theguardian.com/uk/thank-you?product=TierThree&ratePlan=DomesticMonthlyV2#ab-newspaperArchiveBenefit=control`

### DIFF
--- a/support-frontend/assets/components/thankYou/thankyouModules.tsx
+++ b/support-frontend/assets/components/thankYou/thankyouModules.tsx
@@ -25,16 +25,20 @@ const firstColumnContainer = css`
 
 export interface ThankYouModulesProps {
 	isSignedIn?: boolean;
+	showNewspaperArchiveBenefit?: boolean;
 	thankYouModules: ThankYouModuleType[];
 	thankYouModulesData: Record<ThankYouModuleType, ThankYouModuleData>;
 }
 
 export function ThankYouModules({
 	isSignedIn,
+	showNewspaperArchiveBenefit,
 	thankYouModules,
 	thankYouModulesData,
 }: ThankYouModulesProps): JSX.Element {
-	const numberOfModulesInFirstColumn = thankYouModules.length >= 6 ? 3 : 2;
+	const maxModules = showNewspaperArchiveBenefit ? 5 : 6;
+	const numberOfModulesInFirstColumn =
+		thankYouModules.length >= maxModules ? 3 : 2;
 	const firstColumn = thankYouModules.slice(0, numberOfModulesInFirstColumn);
 	const secondColumn = thankYouModules.slice(numberOfModulesInFirstColumn);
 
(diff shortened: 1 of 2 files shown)


support-frontend/assets/components/thankYou/thankyouModules.tsx
```diff
@@ -25,16 +25,20 @@ const firstColumnContainer = css`
 
 export interface ThankYouModulesProps {
 	isSignedIn?: boolean;
+	showNewspaperArchiveBenefit?: boolean;
```
tjmw, line 28: Super minor but would it simplify things to have this be required but with a default (of `false`)? That way within the component you know it's only ever going to be true or false and not undefined.

support-frontend/assets/pages/[countryGroupId]/components/thankyou.tsx
```diff
@@ -177,9 +178,16 @@ export function ThankYouComponent({
 	const isNewAccount = userTypeFromIdentityResponse === 'new';
 	const emailExists = !isNewAccount && isSignedIn;
 
+	const abParticipations = abTestInit({ countryId, countryGroupId });
+	const showNewspaperArchiveBenefit = ['v1', 'v2', 'control'].includes(
+		abParticipations.newspaperArchiveBenefit ?? '',
```
Comment (tjmw) [inline support-frontend/assets/pages/[countryGroupId]/components/thankyou.tsx:183]: Is the nullish coalescing required here? My guess would be that TS is okay with `['v1', 'v2', 'control'].includes(undefined)` but I might be missing something!

tjmw (approved):
Looks good! I left some very minor suggestions to take or leave.

support-frontend/assets/components/thankYou/thankyouModules.tsx
paul-daniel-dempsey: Hmmm, your right, this module was called only so it did not need to be optional, have changed to `boolean` so cannot get `undefined`. I think as soon as I declare it as boolean it will default to false (is that ok?) but have made it explicit anyway here. Thanks for tips, perhaps should reduce the use of optional parameters in general?

support-frontend/assets/pages/[countryGroupId]/components/thankyou.tsx
Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/components/thankyou.tsx:212]: Hope ok but I get from the includes using undefined ->  `Argument of type 'string | undefined' is not assignable to parameter of type 'string'. Type 'undefined' is not assignable to type 'string'.ts(2345)`

Comment (tjmw) [reply support-frontend/assets/pages/[countryGroupId]/components/thankyou.tsx:212]: 👍  Yep all good, sorry for the derail!

PR #6455, opened 2024-10-29
QM : Report conversion value from generic thankyou

paul-daniel-dempsey:
What are you doing in this PR?

Fires the new conversion one-time and recurring conversion events.

[Trello](https://trello.com/c/H1okzlsN/1173-qm-events-part-3-conversion-tracking)

This PR does not cover:
- Reporting one-time-checkout value from the 3-tier landing page


andrewHEguardian (approved):
Nice 👍

PR #6468, opened 2024-11-01
Create new digital newspapers archive header

andrewHEguardian:

Creating the new design of the header that lives on theguardian.newspapers.com. 

[**Trello Card**](https://trello.com/c/pr57zkkZ/1099-build-layout-for-newspaper-archivecom-header)
[Figma](https://www.figma.com/design/h53HMhA753GLQ4ycN1uoPc/newspaper-archive?node-id=1606-23957&node-type=canvas&t=MFmOXb7ZliDlqT41-0)

## How to test

Go to `https://support.theguardian.com/newspaper-archive-header`. I've tested this as an iframe by changing the DEV.public.conf to allow another site and embedding the iframe there.

Or see in storybook.

## Screenshots
Mobile
image

Tablet
image

Wide
image


support-frontend/assets/components/serverSideRendered/newspaperArchiveHeader.tsx
```diff
@@ -1,18 +1,175 @@
...
+
+	${until.tablet} {
+		padding: 0;
+	}
+
+	padding-top: ${space[4]}px;
+`;
+
+const description = css`
+	${until990px} {
```
paul-daniel-dempsey, line 42: These are unusual breakpoints compared to our Guardian ones.  Presume specified by newspaper archives themselves?

paul-daniel-dempsey (approved):
Container looks great on storybook, great job.

support-frontend/assets/components/serverSideRendered/newspaperArchiveHeader.tsx
andrewHEguardian: Yep, from Helene "The 991 px is specific to the Newspaper archive website - so if we're able to use it it would be good as the changes of layout will follow what happens on the rest of the page".
So it's not specified by them but it matches their layout

PR #6479, opened 2024-11-04
Migrate securePadlock and securePadlockCircle SVGs to React components

tjmw:

We'd like to remove the `react-svg-loader` package from the project, which is used by webpack to process `.svg` files. This PR starts that process by migrating two svgs which are currently handled by `react-svg-loader` to React components. The other svgs will be migrated in a subsequent PR (as well as removing `react-svg-loader`.

[**Trello Card**](https://trello.com/c/CkEPRszr/1177-dependabot-high-severity-alert-for-nth-check-package)

## Why are you doing this?

The `react-svg-loader` package is bringing in a version of `nth-check` which has a security vulnerability and so we'd like to remove it from the project. There isn't an obvious upgrade path because the `react-svg-loader` package hasn't seen any updates in quite a long time.

There are a small number of `svg` files in the project which are processed by `react-svg-loader`. We're migrating them to tsx components which removes the need for the loader.


support-frontend/assets/components/svgs/securePadlock.tsx
```diff
@@ -0,0 +1,31 @@
+type Props = {
+	colour: string;
+	opacity?: number;
+};
+
+export default function SecurePadlock({
+	colour,
+	opacity = 1,
```
tjmw, line 8: These are now passed as props. This feels more intuitive than controlling via css with `fill: currentColour` (which wasn't working correctly with the additional div). I'm interested in thoughts on this approach!

support-frontend/assets/components/secureTransactionIndicator/secureTransactionIndicator.tsx
```diff
@@ -1,18 +1,22 @@
 import { css } from '@emotion/react';
 import { neutral, textSansBold14 } from '@guardian/source/foundations';
+import SecurePadlock from 'components/svgs/securePadlock';
```
andrewHEguardian, line 3: Just curious if this would be in curly braces normally? Obviously if it's not flagged by prettier/lint it's all good

andrewHEguardian (approved):
Great!

support-frontend/assets/components/secureTransactionIndicator/secureTransactionIndicator.tsx
tjmw: Good question! I made it a default export from the component which is why there are no curly braces. This is consistent with the interface of the old jsx component which was generated by `react-svg-loader` e.g.:

```
import SecurePadlock from './securePadlock.svg';
```

Very happy to switch to a named export though if that's preferable? It looks like that's generally the approach used elsewhere. What do you reckon?

andrewHEguardian: Looks like we have a [todo](https://github.com/guardian/support-frontend/blob/a5f4a1b3c8a3c32d14bbeade9601087f1e43c2ea/support-frontend/.eslintrc.js#L9 
) to remove a lint override related to this. Probably makes sense to switch to named

paul-daniel-dempsey (approved):
Looks great

support-frontend/assets/components/secureTransactionIndicator/secureTransactionIndicator.tsx
tjmw: Fixed in 4431c8f

PR #6480, opened 2024-11-04
Typescript support workers

rupertbates:
This PR is the first step in migrating the support-workers state machine from Scala to Typescript.

There are a lot of changes in this PR, but they broadly break down into the following categories:
- CDK changes to support deploying a Typescript lamba
- boiler plate for a Typescript project - package.json, tsconfig.json etc.
- Typescript schemas for all the state models passed between the various lambdas
- One Typescript lambda - the CreatePaymentMethod one

In retrospect it probably would have been better to do the infrastructure stuff on its own as a separate PR but I did a lot of this over 2 days of a hack day so I didn't have time to make and merge incremental PRs.

Once this PR is merged, the CreatePaymentMethod step of the state machine will run from the Typescript lambda whilst all the others continue to run from the Scala ones. We can then swap the other lambdas out one by one, and delete the old scala code.

I have tested this in code with the following:
- Supporter plus purchase with Visa, Mastercard, Discovery, Google Pay, Apple Pay, PayPal and Direct Debit
- Guardian Weekly gift purchase with credit card.
- Newspaper with Amex
- Ran the smoke tests successfully


cdk/lib/support-workers.ts
```diff
@@ -180,30 +220,31 @@ export class SupportWorkers extends GuStack {
       errors: ["States.TaskFailed"],
     };
 
-    const preparePaymentMethodForReuse = createLambda(
+    const preparePaymentMethodForReuse = createScalaLambda(
       "PreparePaymentMethodForReuse"
     ).addCatch(failureHandler, catchProps);
 
-    const createPaymentMethodLambda = createLambda(
+    const createPaymentMethodLambda = createTypescriptLambda(
```
rupertbates, line 227: This is the first and only (for now) lambda migrated to Typescript

support-workers/src/typescript/model/billingPeriod.ts
```diff
@@ -0,0 +1,8 @@
+import { z } from 'zod';
```
rupertbates, line 1: I'd like to share a lot of these types with the front end eventually

support-workers/deploy-to-code.sh
```diff
@@ -0,0 +1,13 @@
+S3_BUCKET="membership-dist"
```
rupertbates, line 1: This is a shell script to simplify deploying the zipped artifact to CODE for a quicker testing cycle

support-workers/package.json
```diff
@@ -0,0 +1,44 @@
...
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "test": "jest --group=-integration",
+    "it-test": "jest --group=integration",
+    "lint": "eslint src/typescript/**/*.ts",
+    "check-formatting": "prettier --check **.ts",
+    "build": "tsc",
+    "clean-target": "rm -rf target && mkdir -p target",
+    "install-runtime-dependencies": "yarn install --modules-folder target/typescript/node_modules --production",
```
rupertbates, line 13: This line installs only the runtime (non-dev) dependencies into the target directory so they can be zipped up into the deployment artifact

support-workers/src/typescript/lambdas/createPaymentMethodLambda.ts
```diff
@@ -0,0 +1,194 @@
+import { combinedAddressLine } from '../model/address';
```
rupertbates, line 1: Lambda and application code start here

support-workers/src/typescript/test/address.test.ts
```diff
@@ -0,0 +1,94 @@
+import {
```
rupertbates, line 1: Tests and json fixtures from here on

cdk/lib/support-workers.ts
```diff
@@ -369,7 +426,11 @@ export class SupportWorkers extends GuStack {
       actionsEnabled: isProd,
       snsTopicName: `alarms-handler-topic-${this.stage}`,
       alarmName: `support-workers ${this.stage} No successful recurring stripe supporter plus contributions recently.`,
-      metric: this.buildPaymentSuccessMetric("Stripe", "SupporterPlus", Duration.seconds(3600)),
+      metric: this.buildPaymentSuccessMetric(
+        "Stripe",
+        "SupporterPlus",
+        Duration.seconds(3600)
+      ),
```
johnduffell, line 433: there seem to be quite a few formatting changes so far, assuming they are just automated, perhaps you can do them on another PR, and merge into this PR to make it easier to review?

cdk/lib/support-workers.ts
```diff
@@ -109,18 +100,67 @@ export class SupportWorkers extends GuStack {
...
       lambdaName: string,
       additionalPolicies: PolicyStatement[] = []
     ) => {
       const lambdaId = `${lambdaName}Lambda`;
       const lambda = new GuLambdaFunction(this, lambdaId, {
         ...lambdaDefaultConfig,
+        fileName: `support-workers.jar`,
+        runtime: Runtime.JAVA_21,
         handler: `com.gu.support.workers.lambdas.${lambdaName}::handleRequest`,
         functionName: `${this.stack}-${lambdaName}Lambda-${this.stage}`,
```
johnduffell, line 113: assume everything is just a duplicate between scala->ts other than these 4 lines?

cdk/lib/support-workers.ts
```diff
@@ -109,18 +100,67 @@ export class SupportWorkers extends GuStack {
...
-        },
+        initialPolicy: [
+          s3Policy,
+          cloudWatchLoggingPolicy,
+          ...additionalPolicies,
+        ],
+      });
+      this.overrideLogicalId(lambda, {
+        logicalId: lambdaId,
+        reason: "Moving existing lambda to CDK",
```
johnduffell, line 122: this probably doesn't make sense any more does it?

rupertbates: The comment? I think it's roughly accurate, what else would we put?

cdk/lib/support-workers.ts
rupertbates: https://github.com/guardian/support-frontend/pull/6756

cdk/lib/support-workers.ts
rupertbates: The functionName is the same so it's just the three lines difference

cdk/lib/support-workers.ts
johnduffell: so it's a permanent change? (I realise it was not added from your PR)
In my mind I thought it needed to be overridden during the switch over to avoid clashes and then it could be removed?
Maybe if it's needed to match up with log group names or state machine definitions it would be something like
`"Align function name with state machine definition"`?

cdk/lib/support-workers.ts
johnduffell: consider making it a function that only takes those 3 parameters, to make it clearer they are the same

.github/workflows/support-workers-build.yml
```diff
@@ -97,6 +102,13 @@ jobs:
                     - "-PreparePaymentMethodForReuseLambda-"
                   fileName: support-workers.jar
                 dependencies: [cfn]
+              support-workers-typescript:
```
johnduffell, line 105: I'm glad you've distinguished them by their language as it's clearer 👍  (vs the change from product-move-api -> product-switch-api)

support-workers/deploy-to-code.sh
```diff
@@ -0,0 +1,13 @@
...
+S3_KEY="support/CODE/support-workers-typescript/support-workers.zip"
+LAMBDA_FUNCTIONS=(
+  "-CreatePaymentMethodLambda-"
```
johnduffell, line 4: why not put the `-` where it's used rather than in (every) function name?
`support-${function}-CODE`

support-workers/deploy-to-code.sh
```diff
@@ -0,0 +1,13 @@
...
+)
+echo "Uploading target/typescript/support-workers.zip to S3 (to update the zipfile before upload use the yarn deploy-to-code command)"
```
johnduffell, line 6: if we want to make this script require less knowledge it could either call the yarn command for you, or if that's slow, it could check if the file exists and is up to date first.

support-workers/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "build-and-zip": "yarn clean-target && yarn install-runtime-dependencies && yarn build && cd target/typescript && zip -r support-workers.zip ./* -x \".*\" -x \"__MACOSX\"",
```
johnduffell, line 14: why do we need to clean first, is it just trying to be on the safe side in case incremental build causes an issue?

support-workers/src/typescript/lambdas/createPaymentMethodLambda.ts
```diff
@@ -0,0 +1,194 @@
...
+	paymentFields: PaymentFields,
+	user: User,
+): Promise<PaymentMethod> {
+	switch (paymentFields.paymentType) {
+		case 'Stripe':
+			return createStripePaymentMethod(user.isTestUser, paymentFields);
+		case 'PayPal':
+			return createPayPalPaymentMethod(user.isTestUser, paymentFields);
+		case 'DirectDebit':
+			return createDirectDebitPaymentMethod(paymentFields, user);
```
johnduffell, line 66: keep the params a similar way around?
```suggestion
			return createDirectDebitPaymentMethod(user, paymentFields);
```

support-workers/src/typescript/lambdas/createPaymentMethodLambda.ts
```diff
@@ -0,0 +1,194 @@
...
+	const outputState: CreateSalesforceContactState = {
+		...inputState,
+		paymentMethod,
+	};
+
+	return {
+		state: outputState,
+		error: null,
+		requestInfo: {
+			testUser: false,
```
johnduffell, line 90: if you're hard coding to false, have you checked that none of the logic in the state machine uses this field?
(I assume the scala lambdas would use the one from the user?)

johnduffell: https://github.com/guardian/support-frontend/blob/f2bbb0a87cf906276a7a17e12a36a5e5a0d1a0c6/cdk/lib/support-workers.ts#L174

support-workers/src/typescript/services/stripe.ts
```diff
@@ -0,0 +1,82 @@
...
+
+		const customer: Stripe.Customer = await stripe.customers.create(params);
+		console.log(`Created Stripe customer with id ${customer.id}`);
+		return customer;
+	};
+
+	getPaymentMethod = async (publicKey: string, paymentMethodId: string) => {
+		const stripe = this.stripeForPublicKey(publicKey);
+		const paymentMethod = await stripe.paymentMethods.retrieve(paymentMethodId);
+		return paymentMethod;
```
johnduffell, line 80: what's the return type here?

johnduffell, line 80: what's the return type here?

support-workers/src/typescript/lambdas/createPaymentMethodLambda.ts
```diff
@@ -0,0 +1,194 @@
...
+export async function createStripePaymentMethod(
+	isTestUser: boolean,
+	paymentFields: StripePaymentFields,
+): Promise<StripePaymentMethod> {
+	const stripeService = await stripeServiceHandler.getServiceForUser(
+		isTestUser,
+	);
+	const customer = await stripeService.createCustomer(
+		paymentFields.stripePublicKey,
+		paymentFields.paymentMethod,
```
johnduffell, line 126: this is actually paymentMethodId is it?

support-workers/src/typescript/model/address.ts
```diff
@@ -0,0 +1,108 @@
...
+
+type AddressLine = {
+	streetNumber?: string;
+	streetName: string;
+};
+
+export function combinedAddressLine(
+	addressLine1: string | null,
+	addressLine2: string | null,
+): AddressLine | undefined {
```
johnduffell, line 20: is the null/undefined inconsistency here just because we're getting the incoming zod data into usable format in our code?

support-workers/src/typescript/model/address.ts
```diff
@@ -0,0 +1,108 @@
...
+				};
+			}
+		}
+		return undefined;
+	})();
+
+	return combinedLine;
+}
+
+export function clipForZuoraStreetNameLimit(
```
johnduffell, line 86: truncate is a more usual word I think
```suggestion
export function truncateForZuoraStreetNameLimit(
```

support-workers/src/typescript/lambdas/createPaymentMethodLambda.ts
```diff
@@ -0,0 +1,194 @@
...
+	isTestUser: boolean,
+	payPal: PayPalPaymentFields,
+): Promise<PayPalPaymentMethod> {
+	const payPalService = await paypalServiceHandler.getServiceForUser(
+		isTestUser,
+	);
+	const email = await payPalService.retrieveEmail(payPal.baid);
+	return {
+		PaypalBaid: payPal.baid,
+		PaypalEmail: getIfDefined(email, 'Could not retrieve email from PayPal'),
```
johnduffell, line 162: I feel like validation of the response should be down to `payPalService.retrieveEmail` rather than trying to do it later, I've found the getIfDefined calls in the business logic clutter things up a lot.
I guess it's because `Either` is not really a thing in typescript but would throwing an Error/exception be a idiomatic alternative?

support-workers/src/typescript/lambdas/createPaymentMethodLambda.ts
```diff
@@ -0,0 +1,194 @@
...
+	);
+	const customer = await stripeService.createCustomer(
+		paymentFields.stripePublicKey,
+		paymentFields.paymentMethod,
+	);
+	const paymentMethod = await stripeService.getPaymentMethod(
+		paymentFields.stripePublicKey,
+		paymentFields.paymentMethod,
+	);
+	const card = getIfDefined(
```
johnduffell, line 132: could `stripeService.getPaymentMethod` validate this for us and only return a value if it's valid?
it looks like `card` is the only field we use in `paymentMethod`

support-workers/src/typescript/model/address.ts
```diff
@@ -0,0 +1,108 @@
...
+export const addressSchema = z.object({
+	lineOne: z.string().nullable(),
+	lineTwo: z.string().nullable(),
+	city: z.string().nullable(),
+	state: z.string().nullable(),
+	postCode: z.string().nullable(),
+	country: z.string(), //TODO: build a schema for this
+});
+
+type AddressLine = {
```
johnduffell, line 12:
```suggestion
type StreetAddressLine = {
```

johnduffell: or even ZuoraStreetAddress? if that's what this is needed for .  Just we seem to have the general AddressSchema for the support workers' own model, and then we reprocess into a special format for DD/gocardless purposes, so it would be good to differentiate the reason between the two of them.

support-workers/src/typescript/lambdas/createPaymentMethodLambda.ts
```diff
@@ -0,0 +1,194 @@
...
+	const config = await getStripeConfig(stage);
+	return new StripeService(config);
+});
+const paypalServiceHandler = new ServiceHandler(stage, async (stage) => {
+	const config = await getPayPalConfig(stage);
+	return new PayPalService(config);
+});
+
+export const handler = async (
+	state: WrappedState<CreatePaymentMethodState>,
```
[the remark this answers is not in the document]
rupertbates: Yes completely agree, I did look at doing that in this PR but it was non trivial. I'll probably do that as a seperate PR at some point though

support-workers/src/typescript/model/address.ts
```diff
@@ -0,0 +1,108 @@
...
+	streetNumber?: string;
+	streetName: string;
+};
+
+export function combinedAddressLine(
+	addressLine1: string | null,
+	addressLine2: string | null,
+): AddressLine | undefined {
+	const singleAddressLine = (addressLine: string): AddressLine => {
+		const pattern = /([0-9]+) (.+)/;
```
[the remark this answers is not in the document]
johnduffell: do you intend to strip leading non numbers?
```
"Flat 12, Floor 3, 27 Main Street".match(/([0-9]+) (.+)/).slice(1)
(2) ['27', 'Main Street']
```
maybe it would be better to fail if we don't match the whole string
```suggestion
		const pattern = /^([0-9]+) (.+)$/;
```

support-workers/src/typescript/model/address.ts
```diff
@@ -0,0 +1,108 @@
+import { z } from 'zod';
```
johnduffell, line 1: I find it quite hard to follow this code, probably partly because I'm not used to the language, but also it's not really clear what the aim is.  I'm not sure if the naming could be improved, and perhaps some comments?

support-workers/src/typescript/model/address.ts
```diff
@@ -0,0 +1,108 @@
...
+				return {
+					streetName: concatStreetNames(
+						addressLine1MaybeSplit.streetName,
+						addressLine2MaybeSplit.streetName,
+					),
+				};
+			}
+		}
+		return undefined;
+	})();
```
johnduffell, line 81: is the anonymous function called straight away just to physically group the code together or does it change the meaning somehow also?

support-workers/src/typescript/model/address.ts
```diff
@@ -0,0 +1,108 @@
...
+	return combinedLine;
+}
+
+export function clipForZuoraStreetNameLimit(
+	addressLine: AddressLine,
+): AddressLine {
+	if (addressLine.streetName.length > 100) {
+		return { ...addressLine, streetName: addressLine.streetName.slice(0, 100) };
+	}
+	return addressLine;
```
johnduffell, line 92: might simplify it if we just limit it regardless?
```suggestion
	return { ...addressLine, streetName: addressLine.streetName.slice(0, 100) };
```

support-workers/src/typescript/model/stateSchemas.ts
```diff
@@ -0,0 +1,159 @@
...
+export const userSchema = z.object({
+	id: z.string(),
+	primaryEmailAddress: z.string(),
+	title: titleSchema.nullable(),
+	firstName: z.string(),
+	lastName: z.string(),
+	billingAddress: addressSchema,
+	deliveryAddress: addressSchema.nullable(),
+	telephoneNumber: z.string().nullable(),
+	isTestUser: z.boolean().default(false),
```
johnduffell, line 27: I think it's always in the json
```suggestion
	isTestUser: z.boolean(),
```
the original case class is not a json deserialiser, it's just a scala model.
This line would only make it false when constructed from scala code
https://github.com/guardian/support-frontend/blob/f2bbb0a87cf906276a7a17e12a36a5e5a0d1a0c6/support-models/src/main/scala/com/gu/support/workers/User.scala#L16
whereas the line written above would return false if it's not set in the json
https://zod.dev/?id=default

support-workers/src/typescript/model/stateSchemas.ts
```diff
@@ -0,0 +1,159 @@
...
+
+const baseStateSchema = z.object({
+	requestId: z.string(),
+	user: userSchema,
+	giftRecipient: giftRecipientSchema.nullable(),
+	product: productTypeSchema,
+	analyticsInfo: analyticsInfoSchema,
+	//TODO: This should probably be a date but the scala lambdas struggle to deserialise the default date representation
+	// so leave it as a string until all the lambdas are Typescript
+	firstDeliveryDate: z.string().nullable(),
```
johnduffell, line 89: not sure what it is now or what the "default" date representation is, but for consistency it would be nice to be a zuora date format if that's what TS dates are compatible with e.g. "2025-02-03".
I remember for ease of deserialisation somewhere someone used (milli?)seconds since 1970 in a cookie we set, and it was a massive pain when you wanted to debug things.

support-workers/src/typescript/services/config.ts
```diff
@@ -0,0 +1,48 @@
...
+	const response = await ssmClient.send(command);
+	return schema.parse(JSON.parse(response.Parameter?.Value ?? ''));
+}
+
+export class ServiceHandler<T> {
+	defaultService: T | undefined;
+	testService: T | undefined;
+	constructor(
+		private stage: Stage,
+		private serviceCreator: (stage: Stage) => Promise<T>,
```
johnduffell, line 26:
```suggestion
		private createService: (stage: Stage) => Promise<T>,
```

support-workers/src/typescript/services/config.ts
```diff
@@ -0,0 +1,48 @@
...
+	schema: T,
+): Promise<O> {
+	const ssmClient = new SSMClient(awsConfig);
+	const params = {
+		Name: `/${stage}/support/support-workers/${configKeyName}`,
+		WithDecryption: true,
+	};
+	const command = new GetParameterCommand(params);
+	const response = await ssmClient.send(command);
+	return schema.parse(JSON.parse(response.Parameter?.Value ?? ''));
```
johnduffell, line 18: should we throw (or at least log) something if the config is missing rather than assuming empty?

support-workers/src/typescript/services/payPal.ts
```diff
@@ -0,0 +1,74 @@
...
+				console.log(msg('Warning'));
+				break;
+			case 'Failure':
+				console.log(msg('Error'));
+				break;
+			case 'FailureWithWarning':
+				console.log(msg('Error With Warning'));
+				break;
+		}
+	}
```
johnduffell, line 49: we can flip this over I think
```suggestion
	logNVPResponse(response: URLSearchParams) {
		const ack = response.get('ACK');
		let status: string = 'unknown';
		switch (ack) {
			case 'Success':
				status = 'Successful PayPal NVP request';
				break;
			case 'SuccessWithWarning':
				status = 'Warning';
				break;
			case 'Failure':
				status = 'Error';
				break;
			case 'FailureWithWarning':
				status = 'Error With Warning';
				break;
		}
		console.log(`PayPal: ${status} (NVPResponse: ${response.toString()})`);
	}
```

support-workers/src/typescript/services/stripe.ts
```diff
@@ -0,0 +1,82 @@
...
+			Object.entries(config.accounts).map(([, account]) => [
+				account.publicKey,
+				{
+					paymentGateway: account.paymentGateway,
+					stripe: new Stripe(account.secretKey),
+				},
+			]),
+		);
+	}
+	private stripeForPublicKey(publicKey: string) {
```
johnduffell, line 50:
```suggestion
	private stripeClientForPublicKey(publicKey: string) {
```

support-workers/package.json
rupertbates: So that we make sure we only get what's really needed in the zip, not any cruft that's built up during development. Not crucial for pre-prod but that was the thinking

PR #6483, opened 2024-11-06
Remove some flow artefacts

tjmw:

Remove a couple of left-over artefacts from when this project used flow as the JS type system.


GHaberis (approved):
Nice 🧹

PR #6484, opened 2024-11-06
Upgrade Storybook to v8

tjmw:

Upgrade Storybook 1 major version from v7 -> v8.

[**Trello Card**](https://trello.com/c/vXycbysd/1175-dependabot-high-severity-alert-for-ip-package)

## Why are you doing this?

Dependabot has reported a vulnerability for the `ip` package which is brought in by Storybook. Upgrading to v8 removes this package from the project as storybook v8 doesn't rely on it.

After upgrading:

```
$ yarn why ip
yarn why v1.22.22
[1/4] 🤔  Why do we have the module "ip"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
error We couldn't find a match!
✨  Done in 0.29s.
```

## How to test

Build checks pass.

Storybook reports new version:

Screenshot 2024-11-08 at 17 40 32


support-frontend/tsconfig.json
```diff
@@ -20,7 +20,8 @@
 		// These override @guardian/tsconfig as they have introduced a lot of errors.
 		// The number of errors indicated below is at the time of writing to give an idea of the scale of the fix needed.
 		// We should look at removing these as soon as we can.
-		"noUncheckedIndexedAccess": false // 209 errors in 63 files.
+		"noUncheckedIndexedAccess": false, // 209 errors in 63 files.
+		"types": ["@testing-library/jest-dom"]
```
tjmw, line 24: Without this I was getting a large number of errors like this from the TypeScript compiler:

```
error TS2339: Property 'toBeInTheDocument' does not exist on type 'Matchers<void, HTMLElement | null>'.

133     expect(screen.queryByText('Direct debit')).not.toBeInTheDocument();
```

andrewHEguardian: Which change caused this? Was it adding `@types/jest`?

andrewHEguardian (approved):
Great stuff!

support-frontend/tsconfig.json
tjmw: Good question @andrewHEguardian. No it wasn't the addition of `@types/jest`, that was added to fix a whole load of similar type errors, so it was the combination of adding that package and the change on this line (`"types": ["@testing-library/jest-dom"]`) which got TypeScript happy again. I'm not 100% of the cause, only that it happened after upgrading storybook. I'll dig a bit more and see if I can find out.

tjmw: After some digging I think I've gotten to the bottom of this:

## Before

Storybook v7 brought in `@types/testing-library__jest-dom` as a transitive dependency and it was what was providing the types for `.toBeInTheDocument` & friends. That package is now [deprecated](https://www.npmjs.com/package/@types/testing-library__jest-dom) as `@testing-library/js-dom` now includes types. We were getting the types for free as packages under `@types` are [included by default](https://www.typescriptlang.org/tsconfig/#typeRoots) (TIL!)

## After

After upgrading to Storybook v8 we've lost `@types/testing-library__jest-dom` from the bundle and should be getting the types from `@testing-library/js-dom`. But they aren't included for free (as they aren't namespaced under `@types`) so we need to include them by adding this line to our tsconfig.

PR #6485, opened 2024-11-06
Make PaymentFields a discriminated union

rupertbates:

Generally with JSON objects it is a good idea to have the type encoded in the body to assist with mapping back to concrete types. 

We have this with most of the JSON models which we pass around, but not with the PaymentFields type - this PR adds it which will enable us to create a Zod discriminated union for the payment field types.


support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -808,8 +809,8 @@ function CheckoutComponent({
 				console.error(stripeIntentResult.error);
 			} else if (stripeIntentResult.setupIntent.payment_method) {
 				paymentFields = {
+					paymentType: Stripe,
 					stripePublicKey,
-					recaptchaToken: recaptchaToken,
```
Comment (andrewHEguardian) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:812]: Where did the recaptcha token go?

support-workers/cloud-formation/src/templates/cfn-template.yaml
```diff
@@ -13,12 +13,6 @@ Parameters:
     Description: Stack name
     Type: String
     Default: support
-  OphanRole:
```
andrewHEguardian, line 16: What was this doing?

andrewHEguardian (approved):
Makes a lot of sense!

support-frontend/assets/helpers/subscriptionsForms/submit.ts
```diff
@@ -221,10 +222,7 @@ function buildRegularPaymentRequest(
 		telephoneNumber: telephone,
 		product,
 		firstDeliveryDate: state.page.checkoutForm.product.startDate,
-		paymentFields: {
-			...paymentFields,
-			recaptchaToken,
-		},
+		paymentFields,
```
tjmw, line 225: It looks like the recaptcha token is only relevant here for certain payment methods (direct debit), is that right? (Whereas before it was always included in this data).

tjmw (approved):
Looks good to me 👍 

(I'm also interested in Andrew's question about the cloudformation changes).

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
Comment (rupertbates) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:812]: DD is the only payment method which should include this field, it was wrong before.
https://github.com/guardian/support-frontend/blob/65cc4b1a199f07eaefb9ee67f652b96864cd0341/support-frontend/app/controllers/CreateSubscriptionController.scala#L81

support-workers/cloud-formation/src/templates/cfn-template.yaml
rupertbates: This was needed when we used to call a lambda in the Ophan account to calculate annualised value for acquisitions. We haven't done this for ages but this had been missed. 

The main reason I've removed it in this PR was because otherwise I would have had to set a value for it from the AWS console to deploy this over the support-workers cdk deploy which was there where it had been removed - same for the Kinesis parameter which is also no longer used.

support-frontend/assets/helpers/subscriptionsForms/submit.ts
rupertbates: Yes exactly 👍

PR #6488, opened 2024-11-07
One-Time-Checkout: Cover transaction cost

paul-daniel-dempsey:
Add cover transaction cost to generic one-time-checkout.

[**Trello Card**](https://trello.com/c/beKF97Yg/1195-new-one-time-implement-cover-transaction-element)

|before|after|
|------|------|
|image|image|


support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx
```diff
@@ -282,7 +283,23 @@ function OneTimeCheckoutComponent({
...
 
 	const [otherAmountError, setOtherAmountError] = useState<string>();
-	const finalAmount = getFinalAmount(selectedPriceCard, otherAmount, minAmount);
+	const [coverTransactionCost, setCoverTransactionCost] =
+		useState<boolean>(false);
+
+	const amountWithoutCoverCost =
+		getFinalAmount(selectedPriceCard, otherAmount, minAmount, false) ?? 0;
+	const transactionCoverCost = amountWithoutCoverCost * 0.04;
+	const transactionCostCopy = `I’d like to add a further ${simpleFormatAmount(
```
Comment (andrewHEguardian) [inline support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx:292]: We could put the copy inside the component itself

support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx
```diff
@@ -831,6 +848,17 @@ function OneTimeCheckoutComponent({
...
 						</FormSection>
+						<CoverTransactionCost
+							transactionCost={coverTransactionCost}
+							transactionCostCopy={transactionCostCopy}
+							onChecked={(check) => {
+								setCoverTransactionCost(check);
+							}}
+							transactionCostAmount={simpleFormatAmount(
+								currency,
+								finalAmount ? finalAmount : 0,
```
Comment (andrewHEguardian) [inline support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx:859]: Is showing "I’d like to add a further £0" the copy we want? I guess it's an edge case

support-frontend/assets/helpers/forms/checkouts.ts
```diff
@@ -341,6 +341,7 @@ function getAvailablePaymentRequestButtonPaymentMethod(
 
 // ----- Exports ----- //
 export {
+	round,
```
andrewHEguardian, line 344: Why do we need a new export? Can't we use the same rounding function as the old checkout?

andrewHEguardian (approved):
Neat! 👍  Just a couple of small questions

support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx
Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx:859]: Yep, I checked that and the current checkout does this, so presumed was ok.
image

support-frontend/assets/helpers/forms/checkouts.ts
paul-daniel-dempsey: Hmmm, this has been a thorny problem this afternoon! The old checkout does not multiply the price by 1.04 in the same way its done in the one-time-checkout. Oddly am getting something like `60 x 1.04 = 62.40000000001` hence using the round. Will spend a bit more time looking into a workaround without it now.

support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx
Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/oneTimeCheckout.tsx]: Yep good idea, have moved it inside.

support-frontend/assets/helpers/forms/checkouts.ts
paul-daniel-dempsey: Ok have used `roundToDecimalPlaces` as used in old checkout, not needing to expose `round`

PR #6489, opened 2024-11-07
add identity cookies to diagnostic info

johnduffell:
We sometimes have users with signin issues which are hard to debug, this just adds a little more info to see whether support frontend can see those cookies (without revealing their value, only its length)
image

### DIFF
--- a/support-frontend/app/controllers/DiagnosticsController.scala
+++ b/support-frontend/app/controllers/DiagnosticsController.scala
@@ -21,24 +21,30 @@ class DiagnosticsController(
     "gu.contributions.recurring.contrib-timestamp.Annual",
     "gu.contributions.contrib-timestamp",
     "gu_article_count_opt_out",
+    "GO_SO",
+  )
+  val privateCookies = List(
+    "GU_U",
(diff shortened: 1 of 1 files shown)


support-frontend/app/controllers/DiagnosticsController.scala
```diff
@@ -21,24 +21,30 @@ class DiagnosticsController(
...
 
   import actionRefiners._
   def cookies(): Action[AnyContent] = NoCacheAction() { request =>
-    val cookies = request.cookies.filter { cookie =>
-      relevantCookies.contains(cookie.name)
-    }
     val humanReadableCookies =
       "Please make sure you are logged in and have visited https://www.theguardian.com and seen the issue," +
         " immediately before sending the information from this page" ::
-        s"you have ${cookies.size} cookie(s)." ::
```
johnduffell, line 34: probably don't need to know the number of cookies.  Maybe that was just so we knew the reader copied and pasted the entire response correctly?

tomrf1 (approved):
👍

PR #6495, opened 2024-11-11
refactor: change internationalisation/index to named export

andrewHEguardian:

Refactor the index file exporting countryHelper and countryGroupHelper.

[**Trello Card**](https://trello.com/c/K3zOK9zP/1230-orphans-script-index-filename-removal)

## Why are you doing this?

The file was flagged as a false positive of `yarn orphans`. This refactor should make that output more useful. Also removes some default exports, which we have a lint rule exception for

## How to test

Run `yarn orphans` and do not see `internationalision/index.ts`. No eslint errors (concern as this could undo #5500)


tjmw (approved):
Nice 👍

PR #6498, opened 2024-11-11
First steps to support Guardian Light in the generic checkout

tjmw:

First steps to support Guardian Light in the generic checkout.

[**Trello Card**](https://trello.com/c/5Bjvwbvg/1204-checkout-support-for-new-product-investigate-frontend-and-if-it-needs-light-design-if-any)

### DIFF
--- a/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
@@ -496,6 +496,14 @@ function CheckoutComponent({
 	 */
 	let productFields: RegularPaymentRequest['product'];
 	switch (productKey) {
+		case 'GuardianLight':
+			productFields = {
+				productType: 'GuardianLight',
+				currency: currencyKey,
+				billingPeriod: ratePlanDescription.billingPeriod,
+			};
+			break;
+
 		case 'TierThree':
 			productFields = {
 				productType: 'TierThree',
--- a/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
@@ -1129,7 +1129,14 @@ function CheckoutComponent({
 							promotion,
 						)}
 						headerButton={
-							<BackButton path={`/${geoId}/contribute`} buttonText="Change" />
+							productKey === 'GuardianLight' ? (
+								<BackButton
+									path={`/${geoId}/guardianlight`}
+									buttonText="Back"
+								/>
+							) : (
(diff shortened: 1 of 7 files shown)


support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -496,6 +496,14 @@ function CheckoutComponent({
 	 */
 	let productFields: RegularPaymentRequest['product'];
 	switch (productKey) {
+		case 'GuardianLight':
+			productFields = {
+				productType: 'GuardianLight',
```
Comment (andrewHEguardian) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:501]: I know this is just following the established pattern but it does seem to be ready for some genericness here maybe? The product type is just equal to the product key we switched on

andrewHEguardian (approved):
This looks great 🚀  And a small change, woo for generic checkout!

paul-daniel-dempsey (approved):
Looks great, do you know how I update the `support-servicelambda/typeObject.ts` file with the `GuardianLight` productKey? Does a `yarn` pull it through?

tjmw:
> Looks great, do you know how I update the `support-services-lambda/typeObject.ts` file with the `GuardianLight` productKey? Does a `yarn` pull it through?

Yep it's referenced in the [`package.json` for support-frontend](https://github.com/guardian/support-frontend/blob/713bf80302a9216abbe9b4f79ff6e8d93a93456b/support-frontend/package.json#L92) and pulled directly from GitHub (the specific version is controlled by the Git commit SHA).

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
Comment (tjmw) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:501]: Oops I missed this! That's a good spot, yeah maybe there's a refactoring here which could reduce duplication. I'll take a look.

Comment (tjmw) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:501]: > The product type is just equal to the product key we switched on

Looking more closely this is the case for _some_ product keys, but not all. So we could do something like:

```js
productType: productKey,
```

for the cases where it's true, but I wonder if that's got the potential for confusion as we're mixing concepts. Product keys aren't the same as product types, for example the keys `NationalDelivery`, `SubscriptionCard`, `HomeDelivery` all have a product type of `Paper` (so I guess it's a slightly higher level concept), but making this change could lead future readers to think that they _are_ the same thing. What do you think?

Comment (andrewHEguardian) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:501]: ah that makes sense. Keep it explicit 👍

PR #6501, opened 2024-11-11
One Time Checkout : Set Test Live

paul-daniel-dempsey:

Set `One-Time-Checkout` live for single contributions

[**Trello Card**](https://trello.com/c/UCJ9CNMj/1238-set-a-b-test-live-for-generic-checkout)


andrewHEguardian (approved):
🚀  let's go!

PR #6507, opened 2024-11-12
Return user type from create subscription endpoint

rupertbates:

Return the identity user type from the create endpoint and use this information to pass to the generic thank you page.

## How to test

Go through the generic checkout signed out and see the following messaging on the thank you page

- Has a previous guest account : “Continue to your account”
- Has an existing account: “Continue to your account”
- Has no account: “Complete your Guardian account”

## Screenshots
image

## Next steps
Do the same for one time


support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx
```diff
@@ -736,6 +744,8 @@ export function CheckoutComponent({
 				thankYouUrlSearchParams.set('product', productKey);
 				thankYouUrlSearchParams.set('ratePlan', ratePlanKey);
 				promoCode && thankYouUrlSearchParams.set('promoCode', promoCode);
+				userType && thankYouUrlSearchParams.set('userType', userType);
```
Comment (paul-daniel-dempsey) [inline support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx:747]: Will this filter out `userType=undefined` at this point or later?

Comment (andrewHEguardian) [reply support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx:747]: This will filter it out here. But we have to assume it could be null when the search params are read on the next page

PR #6511, opened 2024-11-13
refactor: split checkouts into two files

andrewHEguardian:

Reduce the size of the checkout files by splitting the two components in each file into separate files. 

[**Trello Card**](https://trello.com/c/LPsGE92d/1156-refactor-checkouttsx)

## Why are you doing this?
The files are massive


paul-daniel-dempsey (approved):
A lot of diffs related to chopping code up! Looks good, smoke testing all passed too.

PR #6512, opened 2024-11-13
Add sub-campaign theme to countdown

sookburt:

Adds the ability to set up different colour and background colours for countdowns for each sub-campaign for US EOY.

[**Trello Card**](https://trello.com/c/gECTtqrM)

## Why are you doing this?

The US would like each sub-campaign to have a different colour scheme.

## How to test

Set the datetime for the campaign to be around 'now' then run locally and go to the US 3-tier landing page with the feature flag of ?forceCampaign=usEoy2024 or check storybook

### Default countdown colour scheme:
Screenshot 2024-11-20 at 13 03 21

### Giving Tuesday countdown colour scheme:
Screenshot 2024-11-20 at 13 03 09

### DIFF
--- a/support-frontend/assets/pages/supporter-plus-landing/twoStepPages/threeTierLanding.tsx
+++ b/support-frontend/assets/pages/supporter-plus-landing/twoStepPages/threeTierLanding.tsx
@@ -24,6 +24,8 @@ import {
 	init as abTestInit,
 	getAmountsTestVariant,
 } from 'helpers/abTests/abtest';
+import { defaultInitialCampaign, getCampaignSettings } from 'helpers/campaigns/campaigns';
+import type { CountdownSetting } from 'helpers/campaigns/campaigns';
 import type {
 	ContributionType,
 	RegularContributionType,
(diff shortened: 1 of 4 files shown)


support-frontend/assets/pages/supporter-plus-landing/twoStepPages/threeTierLanding.tsx
```diff
@@ -315,6 +315,10 @@ export function ThreeTierLanding({
 		label: 'testing',
 		countdownStartInMillis: Date.parse('01 Jan 1970 00:00:00 GMT'),
 		countdownDeadlineInMillis: Date.parse('01 Jan 1970 00:00:00 GMT'),
+		theme: {
```
tomrf1, line 318: why do we define some settings here? It can just come from `campaignSettings`

sookburt: It's setting the initial state of the countdown campaign so that it is not visible - I've moved it into campaign.tsx and imported it to reduce the size of the threetierlanding.

tomrf1: I don't think initial config like this should be necessary. If a countdown should be displayed then we can use `campaignSettings?.countdownSettings`

tomrf1 (approved):
great!

PR #6515, opened 2024-11-14
Remove AU partner benefit test (keep control)

andrewHEguardian:
Remove AU partner benefit from landing page.

## Why are you doing this?
This regional benefit has basically no negative neither positive impact on that landing page key metrics, but we'll keep it there.

## Screenshots
image


tjmw (approved):
👍

PR #6277, opened 2024-09-02
Cover cost of transaction cost

andrewHEguardian:
## What are you doing in this PR?

Add functionality for a user to add 4% to a one time contribution to cover the cost of transaction. Implemented as a checkbox but design is TBC.

[**Trello Card**](https://trello.com/c/RshnFVy7/1032-experiment-adding-cover-transaction-costs-tickbox-to-page-2-on-one-time-journey-checkout-a-b-test)

## Why are you doing this?

A small percentage of our acquisition revenue goes towards covering the cost transaction, we are testing letting users have the option to cover the cost of transaction fees to direct more money to our mission when they choose to make a one-time contribution.

## How to test

Go to `uk/contribute/checkout?selected-contribution-type=one_off#ab-coverTransactionCost=variant`. See checkbox to add 4%. Check checkbox. Amount in CTA changes to add 4% (eg choose 100 see 104). Selected choice card remains the same. Pay. Thank you page shows amount + 4%.

## How can we measure success?

No statistically significant difference in landing page conversion rate

At least 30% opt-in for “cover cost”

## Screenshots
CTA amount updates 
image

Thank you page correct
image

Amount correct in MMA
image

PR #6520, opened 2024-11-14
Increase max one-off contribution by 4 percent

andrewHEguardian:

Increasing the maximum allowed one-off contribution by 4% to allow people to use the "cover transaction cost" checkbox on the highest contributions.

[**Trello Card**](https://trello.com/c/y5Mk6jbI/1224-when-user-chooses-max-contr-amount-cover-transaction-fee-amount-4-can-be-added-on-top-on-all-regions)

## Why are you doing this?

We've seen alarms since adding the checkbox (#6277) that people went over the maximum! This is because the client did not validate the maximum after the additional 4% added by the checkbox. This maximum amount is not a legal / tax decision and is just reputational, so adding 4% is OK.

## How to test
I pointed the payment-api to local and ran it there with the changes. Then in the UK made a one off contribution of £1991 plus the checkbox

## Screenshots
Before
image
After
image

### DIFF
--- a/support-payment-api/src/main/scala/model/Currency.scala
+++ b/support-payment-api/src/main/scala/model/Currency.scala
@@ -29,6 +29,6 @@ object Currency extends Enum[Currency] with CirceEnum[Currency] {
       case USD => 10000
       case _ => 2000
     }
-    amount > maxAmount
+    amount > maxAmount * 1.04
   }
 }


support-payment-api/src/main/scala/model/Currency.scala
```diff
@@ -29,6 +29,6 @@ object Currency extends Enum[Currency] with CirceEnum[Currency] {
       case USD => 10000
       case _ => 2000
     }
-    amount > maxAmount
+    amount > maxAmount * 1.04
```
tjmw, line 32: Would it be worth adding a comment here to explain this maths for future readers? Or perhaps give the 0.04 a name?

```suggestion
    val transactionCostPercentage = 0.04
    amount > maxAmount * (1 + transactionCostPercentage)
```

tjmw (approved):
👍  Great! 1 very minor comment/suggestion to take or leave.

support-payment-api/src/test/scala/services/PaypalServiceSpec.scala
```diff
@@ -112,14 +112,14 @@ class PaypalServiceSpec extends AnyFlatSpec with Matchers with MockitoSugar with
   }
 
   it should "return an error if the payment amount exceeds Australia max" in new PaypalServiceTestFixture {
-    val createPaypalPaymentData = CreatePaypalPaymentData(Currency.AUD, 16000.50, "url", "url")
+    val createPaypalPaymentData = CreatePaypalPaymentData(Currency.AUD, 16640.50, "url", "url")
```
paul-daniel-dempsey, line 115: AU& UK the only countries with upper limits?

support-payment-api/src/main/scala/model/Currency.scala
paul-daniel-dempsey: maybe we shoudl use 4% constant somewhere (config) with a comment?

paul-daniel-dempsey (approved):
Looks good, didnt relaise it was the frontend limiting. Quick question if its just AU & UK that are limited?

support-payment-api/src/test/scala/services/PaypalServiceSpec.scala
andrewHEguardian: The existing test only covers AUD and non-AUD for some reason

PR #6523, opened 2024-11-15
Remove `userTypeFromIdentityResponse` thunk and usage

andrewHEguardian:
## What are you doing in this PR?

Remove the `userTypeFromIdentityResponse` redux property, the async thunk to set it and all uses.

This breaks the continue to your account / complete your account messaging on the (old) one time thank you page and digital subscription thank you page. We are migrating one time checkout page soon, and it's not worth the dev effort on the digisub thank you page.

## How to test

- [x] Go through the checkouts.
- [x] Run the playwright tests.

PR #6524, opened 2024-11-15
Remove getUserType from backend

andrewHEguardian:

Remove the "getUserType" endpoint from the support-frontend backend. 

## Why are you doing this?

We're now returning this information in the create endpoint (#6507). This PR needs to go in after #6523


paul-daniel-dempsey (approved):
Great, that sa loophole closed. Put the first PR in first mind!

PR #6535, opened 2024-11-18
Turn off zip code autocomplete

andrewHEguardian:

Turning off autocomplete on the zip code field. We've seen this be filled in with an email address when payment method is PayPal. Our theory is somehow the modal or browser is autofilling this.

## How do we measure success?

No PayPal email address in postcode support worker failures!


paul-daniel-dempsey (approved):
LG

PR #6536, opened 2024-11-18
Use countdown label as headline for duration of countdown visibility

sookburt:

This ensures that the countdown disappears at the end of it's timespan and the heading of the page is altered to pick up the label associated with the sub-campaign while the countdown is on the page.  It then automatically reverts to the default heading when the countdown disappears (or if there is an empty label).

Also sets the ticker colour scheme based on the figma.

Also changes the ticker colours for the Giving Tuesday campaign.

[**Trello Card**](https://trello.com/c/7RhNsW3L)

## Why are you doing this?

As part of the US end of year campaign, we are using countdown components for a limited period of time.  During the time of the countdown, we would like to change the title of the 3 tier landing page to reflected the limited time till the end of the campaign. 

## How to test

Change the label and start/end dates in the campaign.ts file and run locally using the `?forceCampaign=usEoy2024` flag.

### With label and first countdown and it's label as the title
Screenshot 2024-11-21 at 17 40 46

The countdown disappears automatically and the title switches to the default copy while the page remains in the browser.

### With label and second countdown and it's label as the title
After a refresh, this appears a minute later.
Screenshot 2024-11-21 at 17 43 15

### After countdown has finished the title reverts to the default
Screenshot 2024-11-21 at 17 45 12


support-frontend/assets/pages/supporter-plus-landing/twoStepPages/threeTierLanding.tsx
```diff
@@ -334,6 +337,28 @@ export function ThreeTierLanding({
 		}
 	}, [memoizedCurrentCountdownCampaign]);
 
+	useEffect(() => {
```
tomrf1, line 340: what does this useEffect do?

support-frontend/assets/pages/supporter-plus-landing/components/countdown.tsx
```diff
@@ -106,15 +113,17 @@ export default function Countdown({
 			const isActive =
 				campaign.countdownStartInMillis < now &&
 				campaign.countdownDeadlineInMillis > now;
-			setShowCountdown(isActive);
+			setShow(isActive);
```
charleycampbell, line 116: I think `setShowCountdown` might be a clearer name for this function, as it directly reflects its purpose. That said, if there’s a specific reason for the renaming that I might have missed, so feel free to ignore

support-frontend/assets/pages/supporter-plus-landing/components/countdown.tsx
```diff
@@ -88,14 +90,19 @@ const ensureRoundedDoubleDigits = (timeSection: number): string => {
...
 	countdownCampaign: campaign,
 }: CountdownProps): JSX.Element {
 	// one for each timepart to reduce DOM updates where unnecessary.
 	const [seconds, setSeconds] = useState<string>(initialTimePart);
 	const [minutes, setMinutes] = useState<string>(initialTimePart);
 	const [hours, setHours] = useState<string>(initialTimePart);
 	const [days, setDays] = useState<string>(initialTimePart);
-	const [showCountdown, setShowCountdown] = useState<boolean>(false);
+
+	const hideMyself = () => {
```
charleycampbell, line 103: Could this be renamed to something a bit clearer? Maybe something like `hideCountdown`

support-frontend/stories/landingPage/Countdown.stories.tsx
```diff
@@ -34,81 +34,90 @@ function Template(args: CountdownProps) {
 	);
 }
 
+function dummy(b: boolean){
```
charleycampbell, line 37: can this be removed?

charleycampbell: I see further down it's being passed in so maybe it needs  refactored so it uses the countdown?

support-frontend/assets/pages/supporter-plus-landing/components/countdown.tsx
sookburt: I've changed those, thanks!

support-frontend/assets/pages/supporter-plus-landing/components/countdown.tsx
sookburt: I've changed that, thanks!

support-frontend/assets/pages/supporter-plus-landing/twoStepPages/threeTierLanding.tsx
tomrf1: I can't see how this useEffect has any effect though, as `getHeadline` just returns a value

sookburt: Removed, thanks

support-frontend/stories/landingPage/Countdown.stories.tsx
sookburt: I've removed the dummy function, thanks

PR #6544, opened 2024-11-19
Fix stripe payment method tracking on one time

andrewHEguardian:

Fix the payment method tracking of Stripe payments by adding the 'StripeCheckout' case for non Google or Apple pay.

## Why are you doing this?
We noticed in the data that no payments were recorded as 'Stripe' in the variant, which was obviously suspicious.

## How to test
Put through a credit card payment. See 'Stripe' in the data.

## Screenshots
Purchases with credit card on CODE before and after this change:
image

### DIFF
--- a/support-frontend/assets/pages/[countryGroupId]/components/oneTimeCheckoutComponent.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/components/oneTimeCheckoutComponent.tsx
@@ -275,10 +275,6 @@ export function OneTimeCheckoutComponent({
 		stripeExpressCheckoutPaymentType,
 		setStripeExpressCheckoutPaymentType,
 	] = useState<ExpressPaymentType>();
-	const stripePaymentMethod: StripePaymentMethod =
-		stripeExpressCheckoutPaymentType === 'apple_pay'
-			? 'StripeApplePay'
-			: 'StripePaymentRequestButton';
 
 	const [stripeExpressCheckoutSuccessful, setStripeExpressCheckoutSuccessful] =
 		useState(false);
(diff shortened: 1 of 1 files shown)


support-frontend/assets/pages/[countryGroupId]/components/oneTimeCheckoutComponent.tsx
```diff
@@ -423,6 +419,13 @@ export function OneTimeCheckoutComponent({
 						);
 					}
 				} else {
+					const stripePaymentMethod: StripePaymentMethod =
+						paymentMethod === 'StripeExpressCheckoutElement'
+							? stripeExpressCheckoutPaymentType === 'apple_pay'
+								? 'StripeApplePay'
+								: 'StripePaymentRequestButton'
+							: 'StripeCheckout';
```
Comment (paul-daniel-dempsey) [inline support-frontend/assets/pages/[countryGroupId]/components/oneTimeCheckoutComponent.tsx:427]: Should this be `Stripe`? Wonder whether this is leaking down from line 307.

paul-daniel-dempsey (approved):
Looks good, quick question of difference between `Stripe` versus `StripeCheckout` thats all.

support-frontend/assets/pages/[countryGroupId]/components/oneTimeCheckoutComponent.tsx
Comment (andrewHEguardian) [reply support-frontend/assets/pages/[countryGroupId]/components/oneTimeCheckoutComponent.tsx:427]: It's 'StripeCheckout' in the [generic checkout](https://github.com/guardian/support-frontend/blob/cf852b87a6734ed91f85145eaa87973fb8697b48/support-frontend/assets/pages/%5BcountryGroupId%5D/components/checkoutComponent.tsx#L581) and the old code. And it worked when tested in CODE/dev

PR #6546, opened 2024-11-20
Validation: update pattern to check for whitespace and more emoji-ish characters

andrewHEguardian:

Add whitespace to field validation, and extend the emoji matching.

[**Trello Card**](https://trello.com/c/AMNkTmfY/1208-generic-checkout-user-can-submit-whitespace-as-name)

## Why are you doing this?

We had an alarm where a user had submitted whitespace in the first name field. The extended pictograph regex matches more emojis, eg I tested  ✈️ which isn't strictly an emoji

## How to test

On generic checkout, cannot submit whitespace or emoji. 

### Whitespace
Before
image

After
image

### Emoji
Before
image

After
image

### DIFF
--- a/support-frontend/assets/pages/[countryGroupId]/validation.ts
+++ b/support-frontend/assets/pages/[countryGroupId]/validation.ts
@@ -18,4 +18,5 @@ export function preventDefaultValidityMessage(
  * This uses a Unicode character class escape
  * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Unicode_character_class_escape
  */
-export const doesNotContainEmojiPattern = '^[^\\p{Emoji_Presentation}]+$';
+export const doesNotContainEmojiOrWhitespacePattern =
+	'^(?!.*[\\p{Extended_Pictographic}\\s]).*$';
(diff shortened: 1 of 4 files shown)


paul-daniel-dempsey (approved):
Works like a charm, great

support-frontend/assets/pages/[countryGroupId]/validation.ts
```diff
@@ -18,4 +18,5 @@ export function preventDefaultValidityMessage(
  * This uses a Unicode character class escape
  * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Unicode_character_class_escape
  */
-export const doesNotContainEmojiPattern = '^[^\\p{Emoji_Presentation}]+$';
+export const doesNotContainEmojiOrWhitespacePattern =
+	'^(?!.*[\\p{Extended_Pictographic}\\s]).*$';
```
Comment (paul-daniel-dempsey) [inline support-frontend/assets/pages/[countryGroupId]/validation.ts:22]: Perhaps a comment here to explain the move from `Emoji_Presentation` to `Extended_Pictographic`? Looks like it would pickup an Emoji with a number in front of it which perhaps was a loophole previously.

Comment (andrewHEguardian) [reply support-frontend/assets/pages/[countryGroupId]/validation.ts:22]: It's hard to write an authoritative comment on this I think, based on the googling I've done it's a bit confusing. But this definitely catches more

PR #6551, opened 2024-11-20
Remove Amazon pay as a payment method from the server

rupertbates:
We no longer support Amazon Pay as a payment method, so this PR removes the server side code related to it.

Client side code will be removed in a future PR.


andrewHEguardian (approved):
🧹

PR #6552, opened 2024-11-20
Use `concurrently` npm package in `devrun.sh`

tjmw:

In the `devrun.sh` script we run three processes:

* webpack (for the client)
* sbt (for the backend)
* storybook

The output from each is interleaved, and sometimes I find it can be hard to tell which command a line of output is coming from. I also find using `ctrl-c` to kill things doesn't always work and I'm left with orphaned processes running in the background. Then the next time I try to run e.g. storybook, it binds to a different port as its usual port is taken.

This PR introduces an npm package [`concurrently`](https://www.npmjs.com/package/concurrently) to improve the dev x.

## Why are you doing this?

Running our commands through `concurrently` seems to do a better job of managing processes (killing everything when I `ctrl-c`. It also adds a colour coded prefix to each output line, so we can easily know where output is coming from. For example:

Screenshot 2024-11-20 at 17 21 00

### DIFF
--- a/support-frontend/devrun.sh
+++ b/support-frontend/devrun.sh
@@ -17,6 +17,11 @@ source_nvm() {
 source_nvm
 
 nvm use
-yarn devrun &
-yarn storybook &
-cd ..; sbt -mem 2048 "project support-frontend" devrun
+
+yarn concurrently \
+  --prefix-colors auto \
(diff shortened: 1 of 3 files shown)


andrewHEguardian (approved):
This looks really cool!

PR #6555, opened 2024-11-21
Add one-time label to paypal acquisitions

andrewHEguardian:

This label wasn't being added to PayPal acquisitions on the new one time checkout, which means we couldn't see these when querying the data by label


rupertbates (approved):

paul-daniel-dempsey (commented):
LG

PR #6556, opened 2024-11-21
Set expiry to test user cookie

andrewHEguardian:

Adding an expiry of two days to the test_username cookie.

[**Trello Card**](https://trello.com/c/RsJrEweH/1159-test-users-cookie-out-of-sync-with-backend)

## Why are you doing this?

The cookie seems to persist longer than the backend accepts a test user (the old cookie maxAge was session, and modern browsers persist the session for a long time). The red banner stays present for days on end.

image

 This mismatch causes an alarm if you actually try to make a purchase, as the frontend uses the test Stripe key, and the backend isn't expecting it.

I came up with 2 days because I found this [code](https://github.com/guardian/support-frontend/blob/5a7b58ae37d3142239704b38a8a4ea2ffa57b6ed/support-frontend/app/services/TestUserService.scala#L13) -> 
`  val ValidityPeriod = ofDays(2)`

## How to test
I set the maxAge to a few seconds and went to `/test-users`, the red banner then disappeared after the time elapsed.

I will test a test user in prod once merged (slightly hard to do this in dev because the non-test mode would work anyway)

## Screenshots
image

### DIFF
--- a/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
@@ -190,11 +190,6 @@ export function Checkout({ geoId, appConfig }: Props) {
 		}
 	}
 
-	/**
-	 * TODO: We should probaly send this down from the server as
-	 * this cookie is not always an accurate indicator as to
-	 * whether an account is still valid
-	 */
 	const isTestUser = !!cookie.get('_test_username');
 	const stripePublicKey = getStripeKey(
 		'REGULAR',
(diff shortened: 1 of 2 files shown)


support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -190,11 +190,6 @@ export function Checkout({ geoId, appConfig }: Props) {
 		}
 	}
 
-	/**
```
Comment (andrewHEguardian) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:193]: I'm removing this comment as hopefully this change makes the cookie a more accurate indicator. I'm not sure what the effort entailed to "send it down from the server" would be

paul-daniel-dempsey (approved):
Looks good, will let you test in Prod

johnduffell:
I don't think this is working right as it seems to expire much quicker than 2 days 
image

PR #6557, opened 2024-11-22
Revert "Validation: update pattern to check for whitespace and more emoji-ish characters"

andrewHEguardian:
Validation is broken on the delivery address, reverting 
image


tjmw (approved):
👍

PR #6560, opened 2024-11-22
One time checkout - get user type

andrewHEguardian:

Returning the identity user type for one time payments through the API, and using this information on the client to set a URL param on the thank you page.

This client side solution does not work for PayPal, as we redirect them to the site and can't set the URL params

## How to test

Use the new one time checkout and see "userType=x" when redirected to the thank you page. 

I used the [always authenticate](https://docs.stripe.com/testing?locale=en-GB#regulatory-cards) card to simulate 3DS, as that makes an additional API call 

## Screenshots
image

### DIFF
--- a/support-frontend/assets/helpers/forms/paymentIntegrations/oneOffContributions.ts
+++ b/support-frontend/assets/helpers/forms/paymentIntegrations/oneOffContributions.ts
@@ -3,6 +3,7 @@ import { fetchJson, requestOptions } from 'helpers/async/fetch';
 import { logPromise } from 'helpers/async/promise';
 import type { ErrorReason } from 'helpers/forms/errorReasons';
 import type { IsoCurrency } from 'helpers/internationalisation/currency';
+import type { UserType } from 'helpers/redux/checkout/personalDetails/state';
 import * as cookie from 'helpers/storage/cookie';
 import type { PaymentAPIAcquisitionData } from 'helpers/tracking/acquisitions';
 import 'helpers/tracking/acquisitions';
@@ -147,7 +148,10 @@ function paymentResultFromObject(
 		});
 	}
 
-	return Promise.resolve(PaymentSuccess);
+	const response = json as { data?: { userType: UserType } };
+	const maybeUserType = response.data?.userType;
+
+	return Promise.resolve({ ...PaymentSuccess, userType: maybeUserType });
 }
 
 const postToPaymentApi = (
--- a/support-frontend/assets/helpers/forms/paymentIntegrations/oneOffContributions.ts
+++ b/support-frontend/assets/helpers/forms/paymentIntegrations/oneOffContributions.ts
@@ -136,8 +136,6 @@ const errorResponseSchema = z.object({
 const successResponseSchema = z.object({
 	type: z.literal('success'),
 	data: z.object({
-		currency: z.string(),
-		amount: z.number(),
 		userType: userTypeSchema.optional(),
 	}),
 });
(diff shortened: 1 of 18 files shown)


support-frontend/assets/helpers/forms/paymentIntegrations/oneOffContributions.ts
```diff
@@ -147,7 +148,10 @@ function paymentResultFromObject(
 		});
 	}
 
-	return Promise.resolve(PaymentSuccess);
+	const response = json as { data?: { userType: UserType } };
```
andrewHEguardian, line 151: Admittedly a bit of a bodge 🤷‍♂️

tjmw: Would it be worth writing a zod schema to validate this, rather than casting?

support-payment-api/src/main/scala/backend/PaypalBackend.scala
```diff
@@ -215,7 +254,7 @@ class PaypalBackend(
     }
   }
 
-  private def getOrCreateIdentityIdFromEmail(email: String): Future[Option[String]] =
```
tjmw, line 218: Was the removal of the return type here intentional?

support-payment-api/src/main/scala/backend/PaypalBackend.scala
```diff
@@ -97,28 +97,67 @@ class PaypalBackend(
...
+//
+//          maybeEmail.foreach { email =>
+//            getOrCreateIdentityIdFromEmail(email).foreach { identityId =>
+//              postPaymentTasks(payment, email, identityId, capturePaymentData.acquisitionData, clientBrowserInfo)
+//            }
+//          }
+//
+//          EnrichedPaypalPayment(payment, maybeEmail)
+//        },
+//      )
```
tjmw, line 159: Should this commented section be removed before merging?

support-frontend/assets/pages/[countryGroupId]/components/oneTimeCheckoutComponent.tsx
```diff
@@ -463,6 +463,9 @@ export function OneTimeCheckoutComponent({
 				});
 				const thankYouUrlSearchParams = new URLSearchParams();
 				thankYouUrlSearchParams.set('contribution', finalAmount.toString());
+				'paymentStatus' in paymentResult &&
```
Comment (tjmw) [inline support-frontend/assets/pages/[countryGroupId]/components/oneTimeCheckoutComponent.tsx:466]: Why do we need to check this?

tjmw (approved):
This looks great! 👍 

I left a few minor comments/questions but they aren't blockers.

support-frontend/assets/pages/[countryGroupId]/components/oneTimeCheckoutComponent.tsx
Comment (andrewHEguardian) [reply support-frontend/assets/pages/[countryGroupId]/components/oneTimeCheckoutComponent.tsx:466]: paymentResult could be `PaymentResult` | `CreatePayPalPaymentResponse`. `userType` isn't on the payPal type (although possibly could be). So it's a way of type switching that we use further down too

support-frontend/assets/helpers/forms/paymentIntegrations/oneOffContributions.ts
andrewHEguardian: Yeah that would be ideal, there's another cast in this function so definitely a worthwhile refactor

support-frontend/assets/pages/[countryGroupId]/components/oneTimeCheckoutComponent.tsx
Comment (tjmw) [reply support-frontend/assets/pages/[countryGroupId]/components/oneTimeCheckoutComponent.tsx:466]: (Really sorry if I'm missing something obvious!) Should it not be `'userType' in paymentResult` in that case?

Comment (andrewHEguardian) [reply support-frontend/assets/pages/[countryGroupId]/components/oneTimeCheckoutComponent.tsx:466]: Yep that makes more sense, it was a copy paste from below - my initial intent was to add it to PayPalResponse

PR #6561, opened 2024-11-22
Update supporter product data export schedule in CODE

tjmw:

Updating the cron schedule for the supporter-product-data step function in CODE to run Monday - Friday.

## Why are you doing this?

We noticed that the supporter product data export was running Sunday - Thursday in CODE and not Monday - Friday as intented. It turns out the cron expression specified `1-5` for the days of the week, but this is indexed from 1 which corresponds to Sunday, so `1-5` actually means Sunday - Thursday. You can also specify this using abbreviated day names, which I think is easier to understand.

See: https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-scheduled-rule-pattern.html

## Screenshots

From the schedule rule in AWS.

### Before

Screenshot 2024-11-22 at 16 06 55

### After

Screenshot 2024-11-22 at 16 08 32


andrewHEguardian (approved):
So much nicer 👍

PR #6572, opened 2024-11-25
Add a countdown feature switch

sookburt:

Add the ability to remove countdowns in case of problems.

[**Trello Card**](https://trello.com/c/gECTtqrM)

## Why are you doing this?

If there are significant problems with the countdowns, we need the ability to turn them off quickly.

## How to test

Use the feature switch in the RRCP switches dashboard for the appropriate stage.


tomrf1 (approved):
👍

PR #6576, opened 2024-11-26
Requirement to hide the countdown unless there is a promotion

sookburt:

The Black Friday countdown should not appear unless there is a promotion running - here we're using the promoCode URL parameter to determine this. We are not checking for any specific promo code at this point but we may change that in future to be more resilient.

The rest of the campaign should show regardless of the promoCode's existence - such as the title change, any ticker, etc.

[**Trello Card**](https://trello.com/c/wQSXGTnX)

## Why are you doing this?

We have been asked to ensure that the countdown is hidden unless there is a promotion.

## How to test

Run locally with the countdown settings appropriate to show. Ensure the url contains a promoCode and you can force the campaign with ?forceCampaign=usEoy2024

### UK Black Friday set to require a promoCode but without one - countdown hidden
Screenshot 2024-11-26 at 15 32 57

### UK Black Friday set to require a promoCode with one - countdown shows
Screenshot 2024-11-26 at 15 33 22

### US EOY campaign which is not set to require a promoCode - countdown shows
Screenshot 2024-11-26 at 15 33 47

### DIFF
--- a/support-frontend/assets/helpers/campaigns/campaigns.tsx
+++ b/support-frontend/assets/helpers/campaigns/campaigns.tsx
@@ -15,6 +15,7 @@ export type CountdownSetting = {
 		backgroundColor: string;
 		foregroundColor: string;
 	};
+	requiresPromoCode?: boolean;
 };
 
 export function countdownSwitchOn(): boolean {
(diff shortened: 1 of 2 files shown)


LAKSHMIRPILLAI (approved):

sookburt:
Agreed to simplify the process and hide the whole campaign instead of just the countdown (see the trello card).

PR #6582, opened 2024-11-27
Hide campaign without promo code

sookburt:

For the Black Friday 2024 campaign we have been asked to hide the campaign unless there is a promo code in the URL parameters.

[**Trello Card**](https://trello.com/c/wQSXGTnX)

## Why are you doing this?

So that users are not confused by the headline not matching the discount.


tomrf1 (approved):
👍

PR #6584, opened 2024-11-27
Send all traffic from 3-tier LP to new one-time-checkout

GHaberis:

The `newOneTimeCheckout` A/B test has ended, the new one-time-checkout achieved parity with the existing checkout so we will be pushing all traffic to that. This is the first step in that roll-out, where we end the test and direct all traffic from 3-tier LP to new one-time-checkout.


tjmw (approved):

paul-daniel-dempsey (approved):
LG, smoke tests all passed

PR #6587, opened 2024-11-28
Generic Checkout: Retry Poll Support Workers Process Payment

paul-daniel-dempsey:

The old checkout had a timeout when submitting a purchase to the server, where it would poll for a success or failure response for a maximum number of tries, and if the result was still pending, it would take the user to the thank you page any way on the assumption that the it would succeed eventually, and if it didn’t then the user will be emailed.

The reason this is useful is because the support-workers step function which processes subscriptions will retry for up to 24 hrs to create a subscription in the event of Zuora or Salesforce being down.

- Create retry processPayment mechanism
- Apply to processPayment (after SupportWorker `subscribe/create`) when status marked as `pending` (in order to bring-up thankyou page after 10x3000ms)
- Create Jest test cases
  1. Success (no retry)
  2. Failure (no retry)
  3. Pending followed by Success
  4. Pending until retry limit reached (left Pending, converted to Success in checkout to get to Thankyou)
  5. Reject Promise until retry limit reached (left Reject)
  6. Reject Promise followed by Success

image

[**Trello Card**](https://trello.com/c/2JsHHtXd/1283-add-timeout-to-generic-checkout)

## Testing
`jest -t 'retryPaymentStatus'`


support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx
```diff
@@ -134,8 +134,42 @@ function paymentMethodIsActive(paymentMethod: LegacyPaymentMethod) {
...
+// ----- Setup ----- //
+const POLLING_INTERVAL = 3000;
+const MAX_POLLS = 10;
+
+// retry mechanism for polling based upon
+// https://bpaulino.com/entries/retrying-api-calls-with-exponential-backoff
+function timeOut(milliseconds: number | undefined): Promise<void> {
+	return new Promise((resolve) => setTimeout(resolve, milliseconds));
+}
+function retry(
```
Comment (tjmw) [inline support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx:146]: Since this retry code is very generic I wonder if it should be moved to its own module? We could also write some tests for if we pulled it out. I think the only thing which ties it concretely to our use case is the `Promise<Response>` in the args, but if we changed this use TS generics, it would work for any use case (and would make the testing a bit simpler):

```
function retry<A>(
    promise: Promise<A>,
...
```
```

File: support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx [hunk 2]
```diff
@@ -709,14 +745,24 @@ export function CheckoutComponent({
 			};
 			setErrorMessage(undefined);
 			setErrorContext(undefined);
-			const createResponse = await fetch('/subscribe/create', {
+
+			const fetchCreateSupportWorker = fetch('/subscribe/create', {
```
Comment (tjmw) [inline support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx:749]: I _think_ this should be `() => fetch('/subscribe/create', {`, i.e. add an extra layer of function application. Otherwise we'll be retrying the exact same promise each time (which I don't think will make a difference, since a promise can only have one outcome).

support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx
```diff
@@ -709,13 +745,24 @@ export function CheckoutComponent({
...
+				fetch('/subscribe/create', {
+					method: 'POST',
+					body: JSON.stringify(createSupportWorkersRequest),
+					headers: {
+						'Content-Type': 'application/json',
+					},
+				});
+
+			const createResponse = await retry(
+				fetchCreateSupportWorker,
```
Comment (tjmw) [inline support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx:759]: Is creating the support workers request the thing we want to retry? It looks like in the response we get back a `trackingUri` - is that the thing we should be retrying?

support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx
```diff
@@ -134,8 +134,42 @@ function paymentMethodIsActive(paymentMethod: LegacyPaymentMethod) {
 	);
 }
 
+// ----- Setup ----- //
+const POLLING_INTERVAL = 3000;
```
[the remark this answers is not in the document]
Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx]: Yep, makes sense, updated

support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx
Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx]: Agree, we have done this and tests now in files `retryPaymentStatus.ts` and `retryPaymentStatusTest.ts`

support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx
Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx]: Think we now only need the same promise here. After this, the standalone processPaymentWithRetries mechanism will handle the 10 pending retries.

support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx
Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx]: Crating suppo workers is now not retried.  We retry now in  `processPaymentWithRetries` with eth `trackingUri`

support-frontend/assets/pages/[countryGroupId]/components/retryPaymentStatus.ts
```diff
@@ -0,0 +1,51 @@
+import type { StatusResponse } from 'helpers/forms/paymentIntegrations/readerRevenueApis';
+
+// ----- Setup ----- //
+const POLLING_INTERVAL_MILLIS = 3000;
+const MAX_POLLS = 10;
```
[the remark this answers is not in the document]
Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/components/retryPaymentStatus.ts]: Yep, done

support-frontend/assets/pages/[countryGroupId]/components/retryPaymentStatus.ts
```diff
@@ -0,0 +1,51 @@
...
+			return result; // success or failure, exit
+		} catch (error) {
+			if (polls < pollMax) {
+				if (onRetry) {
+					onRetry(polls + 1, (polls + 1) * pollInterval); // optional retry notification
+				}
+				return retryPollAndPromise(polls + 1);
+			} else {
+				if (error instanceof StillPendingError) {
+					return { status: 'pending', trackingUri: '' };
```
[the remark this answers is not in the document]
Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/components/retryPaymentStatus.ts]: Yes that is a good idea rather than leave blank. Coul donly seem to pass Uri through as the message. Is tha acceptable (with only only string needed) or there is a better way?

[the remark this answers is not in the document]
Comment (tjmw) [reply support-frontend/assets/pages/[countryGroupId]/components/retryPaymentStatus.ts]: Heya, yeah I think that works, since we're not using the message for anything else. I think an alternative could be to augment `StillPendingError` with a custom property [like this](https://www.typescriptlang.org/play/?#code/MYGwhgzhAEDKAuBLEIAKBTAdgE0Zg5gKIBOxA9sdOgB7xbYwnmUDeAUNNMevMQJ4BVYogBc0CLzz4A3GzYBIYGUwTiAV2DwKACgC26KGHzoxqqQBouPfkNHjJBAJTR28+RDUAHdMT0GIRuiOsm4A8gBGAFbomgB0EDyo5FrwfN6hAGba8AAWiBCWCMho9FJMFLGeyWSp3sEcnNC5+bHcvILC0AC8Vu22IQC+bENs7S4NueQA7tCY6DNFKBg4ZaQ6AERkZJ4QAITrlus58PA7IgD05zRgup4g6LFKuudt-Ov1A9DAYPDAOdDaILjTiIDIA9DQPASMCYYDoMhgxYlFYEcrEZzsRpfZQQMj3WIgMj4QGtawdRD1ThDIZAA).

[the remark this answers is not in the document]
Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/components/retryPaymentStatus.ts]: Thats perfect, just updated with ath approach, thanks.

support-frontend/assets/pages/[countryGroupId]/components/__tests__/retryPaymentStatusTest.ts
```diff
@@ -0,0 +1,72 @@
...
+			.fn()
+			.mockImplementationOnce(() => Promise.resolve(pendingStatus))
+			.mockImplementationOnce(() => Promise.resolve(pendingStatus))
+			.mockImplementationOnce(() => Promise.resolve(finalStatus));
+
+		const result = await retryPaymentStatus(getPaymentStatus, 2, 100);
+
+		expect(result).toEqual(finalStatus);
+		expect(getPaymentStatus).toHaveBeenCalledTimes(3);
+	});
```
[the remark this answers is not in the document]
Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/components/__tests__/retryPaymentStatusTest.ts:71]: Think we updated this in our pairing session, implementing the reject tests.

support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx
```diff
@@ -159,30 +161,34 @@ type CreateSubscriptionResponse = StatusResponse & {
...
+				'Content-Type': 'application/json',
+			},
+		}).then((response) => response.json() as unknown as StatusResponse);
+
+	return retryPaymentStatus(getTrackingStatus).then((response) =>
+		processPayment(response),
+	);
+};
+
+const processPayment = (
```
Comment (tjmw) [inline support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx:183]: Very nitpicky but I wonder if this should be called `handlePaymentStatus` or similar? It's not really processing anything more just mapping the response to something we can use here.

I also wonder if the return value should just be a `ProcessPaymentResponse`, do we need to wrap both branches of the if/else in a `Promise`?

Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx]: Thats a good spot, have updated this and removed the promise

support-frontend/assets/pages/[countryGroupId]/components/retryPaymentStatus.ts
```diff
@@ -0,0 +1,52 @@
...
+
+function timeOut(milliseconds: number | undefined): Promise<void> {
+	return new Promise((resolve) => setTimeout(resolve, milliseconds));
+}
+
+export function retryPaymentStatus(
+	promiseFunction: () => Promise<StatusResponse>,
+	pollMax: number = DEFAULT_MAX_POLLS,
+	pollInterval: number = DEFAULT_POLLING_INTERVAL_MILLIS,
+	onRetry?: (pollCount: number, pollTimeTotal: number) => void,
```
Comment (rupertbates) [inline support-frontend/assets/pages/[countryGroupId]/components/retryPaymentStatus.ts:25]: I can't see anywhere where this parameter is ever passed in, do we need it?

support-frontend/assets/pages/[countryGroupId]/components/retryPaymentStatus.ts
```diff
@@ -0,0 +1,52 @@
...
+): Promise<StatusResponse> {
+	async function retryPollAndPromise(polls: number): Promise<StatusResponse> {
+		try {
+			if (polls > 0) {
+				await timeOut(pollInterval); // retry
+			}
+			const result = await promiseFunction();
+			if (result.status === 'pending') {
+				throw new StillPendingError('status is pending', result.trackingUri); // retry on pending
```
Comment (rupertbates) [inline support-frontend/assets/pages/[countryGroupId]/components/retryPaymentStatus.ts:34]: Why do we throw here and then run the retry logic in the catch block? Couldn't we move that logic here?

Comment (tjmw) [reply support-frontend/assets/pages/[countryGroupId]/components/retryPaymentStatus.ts:34]: It's a good question! I think Paul and I talked about this and left it like this because it avoids duplicating some logic (both places would have to do the comparison of polls and pollMax to know if we were at the limit or not). Maybe that's better though, what do you think?

support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx
```diff
@@ -148,7 +149,8 @@ function paymentMethodIsActive(paymentMethod: LegacyPaymentMethod) {
 }
 
 /**
- * This method removes the `pending` state by retrying,
+ * This method retry's `pending` a maximum number of polls
```
Comment (rupertbates) [inline support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx:152]:
```suggestion
/**
 * Attempt to submit a payment to the server. The response will be either `success`, `failure` or `pending`.
 * If it is pending, we keep polling until we get either a success or failure response, or we reach the 
 * maximum number of retries. Reaching the maximum number of retries is treated as a success, as we assume 
 * that the job has been delayed, but will complete successfully in the future and if it doesn't, then the 
 * user will be emailed.
 */
```

support-frontend/assets/pages/[countryGroupId]/components/retryPaymentStatus.ts
Comment (rupertbates) [reply support-frontend/assets/pages/[countryGroupId]/components/retryPaymentStatus.ts:34]: I haven't completely thought through the alternative, but I would generally try to avoid using exceptions to control normal execution flow ie. things which are not actually errors

support-frontend/assets/pages/[countryGroupId]/components/retryPaymentStatus.ts
Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/components/retryPaymentStatus.ts]: Good point, was used for testing purposes (notify of a `pending` retry), removed now.

support-frontend/assets/pages/[countryGroupId]/components/retryPaymentStatus.ts
Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/components/retryPaymentStatus.ts]: Ok, have re-factored to  to avoid using exceptions to control normal (pending) execution flow. Does this approach  look ok?

support-frontend/assets/pages/[countryGroupId]/components/retryPaymentStatus.ts
```diff
@@ -0,0 +1,55 @@
...
+function timeOut(milliseconds: number | undefined): Promise<void> {
+	return new Promise((resolve) => setTimeout(resolve, milliseconds));
+}
+
+export function retryPaymentStatus(
+	promiseFunction: () => Promise<StatusResponse>,
+	pollMax: number = DEFAULT_MAX_POLLS,
+	pollInterval: number = DEFAULT_POLLING_INTERVAL_MILLIS,
+): Promise<StatusResponse> {
+	async function retryPollAndPromiseNew2(
```
Comment (rupertbates) [inline support-frontend/assets/pages/[countryGroupId]/components/retryPaymentStatus.ts:26]:
```suggestion
	async function retryPollAndPromise(
```
Not sure what this name was originally but the `New2` suffix looks like testing stuff?

support-frontend/assets/pages/[countryGroupId]/components/retryPaymentStatus.ts
```diff
@@ -0,0 +1,55 @@
...
+		polls: number,
+	): Promise<StatusResponse> {
+		try {
+			if (polls > 0) {
+				await timeOut(pollInterval); // retry
+			}
+			const result = await promiseFunction();
+			if (result.status === 'pending' || result.status === 'failure') {
```
Comment (rupertbates) [inline support-frontend/assets/pages/[countryGroupId]/components/retryPaymentStatus.ts:34]: Do we want to keep retrying on failure? Shouldn't we just return at this point?

support-frontend/assets/pages/[countryGroupId]/components/retryPaymentStatus.ts
```diff
@@ -0,0 +1,55 @@
...
+				if (polls < pollMax - 1) {
```
Comment (rupertbates) [inline support-frontend/assets/pages/[countryGroupId]/components/retryPaymentStatus.ts:35]: Why pollMax -1? Just reduce the value of DEFAULT_MAX_POLLS if we want one less.

support-frontend/assets/pages/[countryGroupId]/components/retryPaymentStatus.ts
```diff
@@ -0,0 +1,55 @@
...
+					return retryPollAndPromiseNew2(polls + 1); // retry on pending
+				} else if (result.status === 'pending') {
+					throw new StillPendingError(
```
Comment (rupertbates) [inline support-frontend/assets/pages/[countryGroupId]/components/retryPaymentStatus.ts:38]: Why throw StillPendingError, why not just `return { status: 'pending', trackingUri: error.trackingUri };`

support-frontend/assets/pages/[countryGroupId]/components/__tests__/retryPaymentStatusTest.ts
```diff
@@ -0,0 +1,99 @@
...
+
+describe('retryPaymentStatus', () => {
+	it('returns immediately if the payment is successful', async () => {
+		const status: StatusResponse = {
+			status: 'success',
+			trackingUri: 'https://support/status',
+		};
+		const getPaymentStatus = jest.fn(() => Promise.resolve(status));
+
+		const result = await retryPaymentStatus(getPaymentStatus, 1, 1);
```
Comment (rupertbates) [inline support-frontend/assets/pages/[countryGroupId]/components/__tests__/retryPaymentStatusTest.ts:12]: Is this actually testing anything? Won't passing in 1 for the maximum number of polls mean that the function will return immediately regardless of the result?

Comment (tjmw) [reply support-frontend/assets/pages/[countryGroupId]/components/__tests__/retryPaymentStatusTest.ts:12]: I think it's testing the behaviour that if the promise immediately succeeds then no retries are necessary and the return value reflects the successful outcome.

support-frontend/assets/pages/[countryGroupId]/components/retryPaymentStatus.ts
Comment (tjmw) [reply support-frontend/assets/pages/[countryGroupId]/components/retryPaymentStatus.ts:38]: Yeah I think that's right, we don't need the special exception any more 👍

support-frontend/assets/pages/[countryGroupId]/components/__tests__/retryPaymentStatusTest.ts
Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/components/__tests__/retryPaymentStatusTest.ts:12]: Yep, want to be assured it can succeed in first attempt

support-frontend/assets/pages/[countryGroupId]/components/retryPaymentStatus.ts
Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/components/retryPaymentStatus.ts:34]: The original behaviour (oneTimeCheckout) allowed a number of failures to be retried so was matched. Am unsure the condition that would need this however it does ask to match behaviour so perhaps re-asses after PR?

support-frontend/assets/pages/[countryGroupId]/components/retryPaymentStatus.ts
Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/components/retryPaymentStatus.ts:35]: Was originally starting at 1, agree better to start at zero retires, changed to do this.

support-frontend/assets/pages/[countryGroupId]/components/retryPaymentStatus.ts
Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/components/retryPaymentStatus.ts:36]: Looks much shorter/cleaner now without, thanks Rupert, will try to ensure normal execution flow does not run through error handler in future. Removed.

support-frontend/assets/pages/[countryGroupId]/components/retryPaymentStatus.ts
Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/components/retryPaymentStatus.ts]: Agree, status `failure` or `success` should drop out straight away.  status `pending` or a rejected promise will retry a number of default or specified attempts (starting from zero).  This has been changed to match this behaviour.

rupertbates (approved):
👍 looks good to me

PR #6605, opened 2024-12-05
Introduce knip depedency for identifying unused code

GHaberis:

"knip" (https://knip.dev) is a dev dependency that can be used to declutter JavaScript & TypeScript projects by identifying unused files, module exports and dependencies.  

We have an "orphan" script introduced 4+ years ago (https://github.com/guardian/support-frontend/pull/2535), that set out to do something similar that used a dependency called "madge" (https://www.npmjs.com/package/madge) to identify unused files, this required a lengthy custom script to be written and didn't identify unused exports, it served a purpose but wasn't ideal.

Knip on the other hand can be zero/low config and offers more features, such as a Github Action we could use on PRs! So I wanted to see if we could use this instead, as I think this will be more useful, particularly when we begin to delete code - such as the 2-step checkout!

### Configuration
I had to go down the low config route, as we have multiple [entry points](https://knip.dev/explanations/entry-files) for our JS bundles, and knip needed to know what these are. The config lives in a `knip.config.js` file that imports our entry points.

Other than defining the JS entry points I have also instructed Knip to:

- I've used the `-include files,exports,nsExports,types,nsTypes` flag to instruct knip to only report on unused files, exports and types. Out of the box it flags unused dependencies too but the results contained dependencies I'm pretty sure we use! So I think this needs further investigation / congig which could come later.
- Only flag unused `.js`, `.jsx`, `.ts`, `.tsx` and `.scss` files.
- Ignore unused exports that are referenced in the file their defined (`ignoreExportsUsedInFile: true`), it turns out we have a lot of exports that aren't used outside their module, but are used inside it. I think we can come back to these at a later date, for now they clutter the results as the most useful thing IMO is the code that's not used anywhere!
- I've defined the webpack entry points as knip was flagging these as unused, their config accounts for this and allowed me to exclude them. 

### Results
Knip flagged a few unused files and plenty of exports that are unused inside and outside their own file. This is something our other linting tools would not catch as unused code, as as soon as you stick an "export" keyword in front of a function they could be feasibly used outside their immediate module. I've been through the top 10 exports in the results and sure enough they weren't referenced inside or outside their files.

Screenshot 2024-12-05 at 11 45 23

### Next steps
- Replace "orphan" script with knip
- Delete unused code
- Delete unused files
- Revisit unused dependency reporting
- Remove unused `export` keyword from exports that are referenced inside the file their defined, using knip's [auto-fix](https://knip.dev/features/auto-fix ) feature: `knip --fix-type exports`
- Set-up a knip Github Action for catching the introduction of unused code! https://knip.dev/reference/plugins/github-actions


GHaberis:
Screenshot 2024-12-05 at 12 22 16

support-frontend/package.json
```diff
@@ -70,6 +71,37 @@
 		"verbose": true,
 		"testEnvironment": "jest-environment-jsdom-global"
 	},
+  "knip": {
+    "entry": [
```
tjmw, line 75: Not necessarily for now, but it'd be great if this was driven off of the same list of entry points we define in `webpack.entryPoints.js` (that way we couldn't forget to add a new entry point in here). It looks like as well as defining in `package.json`, this config can also exist in a separate js config file, from which we _could_ import from `webpack.entryPoints.js`.

tjmw (approved):
Great! Excited to see how well this works for us 👍

support-frontend/package.json
GHaberis: That's a great idea, it was a pain copying them over! I'll set-up a separate knip config and import the `webpack.entryPoints.js` so we don't miss one if something new gets added!

PR #6613, opened 2024-12-09
Remove unused flow-typed files 🧹

GHaberis:

This PR removes some files that were used when we used Flow's type system (pre-Typescript), they should've been deleted when we removed Flow from the codebase.


rupertbates (approved):

support-frontend/assets/helpers/campaigns/campaigns.tsx
```diff
@@ -48,7 +48,8 @@ const campaigns: Record<string, CampaignSettings> = {
 	usEoy2024: {
 		isEligible: (countryGroupId: CountryGroupId) =>
 			countryGroupId === UnitedStates,
-		enableSingleContributions: false,
+		// enableSingleContributions: false,
+		enableSingleContributions: true,
```
tomrf1, line 52: this should be false still

GHaberis: Thanks @tomrf1, sorry I missed that!

PR #6614, opened 2024-12-09
Update final countdown title copy and use display start and end dates

sookburt:

As the US would like to change only the title on the last (ish) day, this adds a display start and end date and checks for that rather than the countdown start and end dates to determine which countdown setting to show.  This way the countdown deadline remains the same but the title switches at a specific point.
This allows the headline copy change to happen without the need for a developer to create and merge PR to make the change at a specific time.

[**Trello Card**](https://trello.com/c/jK383PT4)

## Why are you doing this?

So that the countdown displays the correct headline copy during the specified dates.

## How to test

Clone locally, take a copy of the existing settings and change the dates to sometime in the near future, run locally and use the campaign flag to see the countdown and title.  Watch over the course of the timeframes you've added to see how the title changes but the deadline remains the same. Note, it requires a page refresh to see the change.


sookburt:
It turns out the requirements have been simplified - this is more complex than needed.

PR #6617, opened 2024-12-09
Apply min/max validation to other amount field on one-time checkout

GHaberis:

We had some alarms from the one-time checkout where the other amount field was allowing large numbers, above the configured min/max contribution values, we found on investigation no validation to be in place on this input, so have used the native HTML min/max [attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number) available to `<input type='number />` to manage this validation:

### DIFF
--- a/support-frontend/assets/pages/[countryGroupId]/components/oneTimeCheckoutComponent.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/components/oneTimeCheckoutComponent.tsx
@@ -229,6 +229,7 @@ export function OneTimeCheckoutComponent({
 	);
 
 	const minAmount = config[countryGroupId]['ONE_OFF'].min;
+	const maxAmount = config[countryGroupId]['ONE_OFF'].max;
 
 	const [selectedPriceCard, setSelectedPriceCard] = useState<number | 'other'>(
 		preSelectedPriceCard ?? defaultAmount,
@@ -327,7 +328,7 @@ export function OneTimeCheckoutComponent({
 			if (validityState.valueMissing) {
 				setAction(missing); // required
 			} else {
-				setAction(invalid ?? ' '); // pattern mismatch
+				setAction(invalid ?? 'Invalid input'); // pattern mismatch
 			}
 		}
 	};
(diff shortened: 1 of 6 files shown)


support-frontend/assets/pages/[countryGroupId]/components/oneTimeCheckoutComponent.tsx
```diff
@@ -327,7 +328,7 @@ export function OneTimeCheckoutComponent({
 			if (validityState.valueMissing) {
 				setAction(missing); // required
 			} else {
-				setAction(invalid ?? ' '); // pattern mismatch
+				setAction(invalid ?? 'Invalid input'); // pattern mismatch
```
Comment (GHaberis) [inline support-frontend/assets/pages/[countryGroupId]/components/oneTimeCheckoutComponent.tsx:331]: @paul-daniel-dempsey felt it would be better to fallback to a very general error message, rather than an empty error message here.

rupertbates (approved):
Great 👍

GHaberis:
We've had to make a follow-up change as TS flagged some errors following our introduction of a non-optional `maxAmount` prop on support-frontend/assets/components/otherAmount/otherAmount.tsx.

TS highlighted 4 places where support-frontend/assets/components/otherAmount/otherAmount.tsx was also referenced, which weren't passing `maxAmount`:

1. support-frontend/assets/pages/supporter-plus-landing/components/contributionsPriceCards.tsx
2. support-frontend/assets/pages/supporter-plus-landing/components/oneOffCard.tsx
3. support-frontend/assets/pages/supporter-plus-landing/formSections/amountAndBenefits.tsx
4. support-frontend/assets/pages/supporter-plus-landing/formSections/patronsPriceCards.tsx

 We had 2 options: Update these references to also pass `maxAmount` OR make `maxAmount` optional. 

We decided to make `maxAmount` optional AND make `minAmount` optional. We did this because despite these 4 `otherAmount` references passing `minAmount` already the prop was not actually used, until we added the min/max attributes in this PR.

Making `minAmount` optional, meant we could stop passing it from the 4 references to `otherAmount`, this is a zero side effect approach as up until now these implementations may have passed the `minAmount`  but it was never used.

PR #6625, opened 2024-12-11
Fail e2e test noisily if the `PAYPAL_TEST_PASSWORD` env var isn't set

tjmw:

Fail e2e test noisily if the `PAYPAL_TEST_PASSWORD` env var isn't set.

## Why are you doing this?

Previously we silently allowed this env var to be undefined and then interpolated into the string we build up for the PayPal password, giving an incorrect password. Too many incorrect login attempts causes the account to be locked, which we want to avoid.

## How to test

Run the tests locally with the env var: ✅ 

Run the tests locally without the env var:
Screenshot 2024-12-11 at 09 38 09

### DIFF
--- a/support-e2e/tests/utils/paypal.ts
+++ b/support-e2e/tests/utils/paypal.ts
@@ -7,6 +7,18 @@ const paypalUsernamePrefixes = [
 	'sb-hk3ov30650585',
 ];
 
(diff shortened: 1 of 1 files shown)


rupertbates (approved):
👍

paul-daniel-dempsey (approved):
works like a charm, thanks

PR #6634, opened 2024-12-16
Make the Tier Three checkout alarm less sensitive

tjmw:
## What are you doing in this PR?

Increase the time period for the Tier Three alarm.

[**Trello Card**](https://trello.com)

## Why are you doing this?

Since adding this alarm we've had a few false positives.

## Accessibility test checklist

-  [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-  [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-  [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-  [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

PR #6636, opened 2024-12-17
Update the Tier Three alarm description and increase period

tjmw:

Fix the Tier Three alarm description and also increase the period to 24 hours.

## Why are you doing this?

The Tier Three alarm description is currently out of sync with the alarm config. This is because I forgot to update the description in #6634. Instead of having to remember, this PR changes approach to deriving the description from the alarm config so it's impossible to get out of sync. If we like the approach we can use if for the other alarms in here.

## How to test

`yarn test` in the `cdk` dir. The snapshot contains the updated description.

### DIFF
--- a/cdk/lib/__snapshots__/support-workers.test.ts.snap
+++ b/cdk/lib/__snapshots__/support-workers.test.ts.snap
@@ -1661,9 +1661,9 @@ exports[`The support-workers stack matches the snapshot 1`] = `
             ],
           },
         ],
-        "AlarmName": "URGENT 9-5 - PROD support-workers No successful Tier Three checkouts in 12 hours.",
+        "AlarmName": "URGENT 9-5 - PROD support-workers No successful Tier Three checkouts in 1 day.",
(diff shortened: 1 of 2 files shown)


cdk/lib/__snapshots__/support-workers.test.ts.snap
```diff
@@ -1661,7 +1661,7 @@ exports[`The support-workers stack matches the snapshot 1`] = `
             ],
           },
         ],
-        "AlarmName": "URGENT 9-5 - PROD support-workers No successful Tier Three checkouts in 6h.",
+        "AlarmName": "URGENT 9-5 - PROD support-workers No successful Tier Three checkouts in 12 hours.",
```
tjmw, line 1664: `Duration#toHumanString` works nicely

paul-daniel-dempsey (approved):
Looks clearer, still not sure why it fired under 12hrs last night but hopefully this will work

PR #6640, opened 2024-12-18
Contributions Only Landing - link to generic checkout

andrewHEguardian:

Change the CTAs on the contributions only checkout (shown to users in certain countries) to redirect to the new generic or one time checkouts, instead of the old second step checkout.

Also remove the patrons section.

[**Trello Card**](https://trello.com/c/rhciXIi5/1289-contributions-only-landing-page-for-users-in-vat-affected-country-list-countries-that-dons-sell-subs-phase-1)

## Why are you doing this?

Part one of deprecating the old two step checkout

## How to test

Go to https://support.thegulocal.com/int/contribute?country=lb (LB is an example that works in local), click "Continue to checkout" from any of the tabs and go to the correct checkout with the correct amount and rateplan

## Screenshots
image
Link
image


paul-daniel-dempsey (approved):
Works great, might be worth moving all files over to the `[countryGroupId]` folder so its with everything else we have moved over to the new checkout?

PR #6643, opened 2024-12-19
Set 'hide support messaging' cookie for one time contributors

rupertbates:
We want to stop working out user benefits from what products they have and instead delegate that to the user-benefits api.

As part of this work we no longer want to set a cookie to inform dotcom that a user has made a one time contribution, but instead set the 'hide support messaging' cookie which tells DCR & support-dotcom-components to show fewer support asks.

This PR works in combination with: https://github.com/guardian/dotcom-rendering/pull/13037

### DIFF
--- a/support-frontend/assets/helpers/storage/contributionsCookies.ts
+++ b/support-frontend/assets/helpers/storage/contributionsCookies.ts
@@ -1,14 +1,12 @@
 import { set } from './cookie';
 
-const ONE_OFF_CONTRIBUTION_COOKIE_NAME = 'gu.contributions.contrib-timestamp';
-const ONE_OFF_CONTRIBUTION_COOKIE_NAME_DAYS_TO_LIVE = 365;
+const HIDE_SUPPORT_MESSAGING_COOKIE_NAME = 'gu_hide_support_messaging';
+const HIDE_SUPPORT_MESSAGING_COOKIE_DAYS_TO_LIVE = 90;
 
-export const setOneOffContributionCookie = (): void => {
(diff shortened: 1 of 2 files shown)


tomrf1:
Thinking through some scenarios:
1. a user already has the `gu_hide_support_messaging` cookie from a recurring product. They then make a single contribution. Does it matter that it's overwritten by one with an expiry? I think not, because DCR will overwrite it next time it talk to MDAPI.
2. a user makes a single contribution, then later signs into an account with a recurring product. Again, DCR will overwrite it.
3. a user makes a single contribution, then later takes out a recurring product. The cookie is replaced by one with a 24 hour expiry, and after that the user must sign in.

Does this all sound right?

rupertbates:
> Thinking through some scenarios:
> 
>     1. a user already has the `gu_hide_support_messaging` cookie from a recurring product. They then make a single contribution. Does it matter that it's overwritten by one with an expiry? I think not, because DCR will overwrite it next time it talk to MDAPI.
> 
>     2. a user makes a single contribution, then later signs into an account with a recurring product. Again, DCR will overwrite it.
> 
>     3. a user makes a single contribution, then later takes out a recurring product. The cookie is replaced by one with a 24 hour expiry, and after that the user must sign in.
> 
> 
> Does this all sound right?

Yes, this is all correct I think. The cookie set by support-frontend will cover the 90 days period which we give one time contributors until they sign in, from that point on the mdapi/user-benefits response takes over and they need to remain signed in to receive the benefit.

PR #6650, opened 2024-12-20
Remove utility types package and replace all usages of $Keys

andrewHEguardian:

Remove utility types package and replace all usages of $Keys with keyof

## Why are you doing this?

I noticed we had this dependency which only seemed to serve one purpose, which can be done with native Typescript


GHaberis (approved):
Excellent 🧹

PR #6655, opened 2024-12-23
Configurable landing page

tomrf1:
Begins the process of making the landing page properly configurable.
For now it's just the heading and subheading.

Until now everything was hardcoded into the client, including any AB testing. Currently there's some logic for setting a different subheading for the US.

Instead we will now pass landing page config through to the client from the server in the `guardian.settings` object. This model supports AB testing. Targeting is by country group only for now.

In the longer term we'll define this config in a database and create a tool for configuring it. For now it's hardcoded into the backend in `settings/LandingPageTest.scala`.

Assignment to a test/variant happens in the existing AB test `init` function, but it uses the config from the server instead of the tests defined in `abtestDefinitions.ts`. In order to track participation through to the checkout we set a new session storage item, `landingPageParticipations`.


support-frontend/assets/pages/supporter-plus-landing/twoStepPages/threeTierLanding.tsx
```diff
@@ -142,12 +140,6 @@ const paymentFrequencyButtonsCss = css`
 	}
 `;
 
-const tabletLineBreak = css`
```
tomrf1, line 145: This does mean we lose the line-break in the heading below desktop. We could solve that in the tooling, but we should also reconsider how important it is

support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -219,22 +235,6 @@ function getParticipationsFromUrl(): Participations | undefined {
 	return;
 }
 
-function getParticipationsFromSession(): Participations | undefined {
```
tomrf1, line 222: moved to `sessionStorage.ts` for reuse

support-frontend/assets/helpers/campaigns/campaigns.tsx
```diff
@@ -28,7 +28,7 @@ export type CampaignTickerSettings = Omit<TickerSettings, 'tickerData'> & {
 	id: string;
 };
 
-export type CampaignSettings = {
+type CampaignSettings = {
```
tomrf1, line 31: No longer used by the landing page - we can probably remove this file later

support-frontend/assets/pages/supporter-plus-landing/twoStepPages/threeTierLanding.tsx
```diff
@@ -138,12 +136,6 @@ const paymentFrequencyButtonsCss = css`
...
```
tomrf1, line 141: The css here prevents the line break from appearing on desktop.
If we want to reproduce this then we'll need separate config for mobile+tablet vs desktop.
We should consider how important this feature is - perhaps we can delay making this choice for now and just always have the line break in the heading.

support-frontend/assets/pages/supporter-plus-landing/twoStepPages/threeTierLanding.tsx
```diff
@@ -138,12 +137,6 @@ const paymentFrequencyButtonsCss = css`
...
-	${from.desktop} {
-		display: none;
```
tomrf1, line 143: The need for this css is replaced by the `text-wrap: balance` on the heading. It means the line break still happens after the comma when it wraps:
Screenshot 2025-02-04 at 10 04 22
And it saves us from needing separate config for mobile vs desktop

support-frontend/app/admin/settings/LandingPageTest.scala
```diff
@@ -0,0 +1,92 @@
...
+  implicit val statusEncoder = deriveEnumerationEncoder[Status]
+  implicit val statusDecoder = deriveEnumerationDecoder[Status]
+}
+
+case class LandingPageTestTargeting(
+    countryGroups: List[String],
+)
+
+object LandingPageTestTargeting {
+  implicit val codec: Codec[LandingPageTestTargeting] = deriveCodec
```
charleycampbell, line 23: I don't think seen (or noticed) Codec before - what does it do?

support-frontend/app/admin/settings/LandingPageTest.scala
```diff
@@ -0,0 +1,92 @@
...
+            subheading =
+              "We're not owned by a billionaire or profit-driven corporation: our fiercely independent journalism is funded by our readers. Monthly giving makes the most impact (and you can cancel anytime). Thank you.",
+          ),
+        ),
+      ),
+    ),
+    LandingPageTest(
+      name = "LP_DEFAULT",
+      status = Status.Live,
+      targeting = LandingPageTestTargeting(countryGroups =
```
charleycampbell, line 77: do we need to anything here with the new way of targeting by location?

support-frontend/assets/helpers/abTests/__tests__/landingPageAbTests.spec.ts
```diff
@@ -0,0 +1,125 @@
+import type { LandingPageTest } from '../../globalsAndSwitches/landingPageSettings';
+import {
+	fallBackLandingPageSelection,
+	getLandingPageParticipations,
+	getLandingPageVariant,
+} from '../landingPageAbTests';
+import { LANDING_PAGE_PARTICIPATIONS_KEY } from '../sessionStorage';
+
+const ukTest: LandingPageTest = {
```
charleycampbell, line 9: Could this be renamed to non-US Test as we're passing in all the country groups other than the US here? Or we could just pass in the GBPCountries

support-frontend/assets/helpers/abTests/landingPageAbTests.ts
```diff
@@ -0,0 +1,97 @@
+import seedrandom from 'seedrandom';
```
charleycampbell, line 1: ohh seedrandom! what does this do?

support-frontend/assets/helpers/abTests/landingPageAbTests.ts
```diff
@@ -0,0 +1,97 @@
...
+import {
+	getSessionParticipations,
+	LANDING_PAGE_PARTICIPATIONS_KEY,
+	setSessionParticipations,
+} from './sessionStorage';
+
+export type LandingPageSelection = LandingPageVariant & { testName: string };
+
+export const fallBackLandingPageSelection: LandingPageSelection = {
+	testName: 'FALLBACK_LANDING_PAGE',
```
charleycampbell, line 17: Will the fallback be hardcoded or will we pull config from DynamoDB? I don't mind it being hard coded as it's problaby quite important that it doesn't change

support-frontend/assets/helpers/campaigns/campaigns.tsx
charleycampbell: I've never really understood the campaign feature...are we going to retire it?

support-frontend/assets/pages/supporter-plus-landing/twoStepPages/threeTierLanding.tsx
```diff
@@ -102,6 +99,7 @@ const innerContentContainer = css`
 `;
 
 const heading = css`
+	text-wrap: balance;
```
charleycampbell, line 102: Great addition ❤️

support-frontend/assets/pages/supporter-plus-landing/twoStepPages/threeTierLanding.tsx
```diff
@@ -619,36 +591,13 @@ export function ThreeTierLanding({
 						/>
 					)}
 					<h1 css={heading}>
-						{getHeadline(
-							showCountdown,
-							currentCountdownSettings,
-							campaignSettings,
-							countryGroupId,
-						)}
+						<span dangerouslySetInnerHTML={{ __html: settings.copy.heading }} />
```
charleycampbell, line 594: Such a great name! Why is it dangerous?  😂

support-frontend/assets/helpers/campaigns/campaigns.tsx
tomrf1: yes I think so, it'll eventually be replaced by the configuration from the server

support-frontend/app/admin/settings/LandingPageTest.scala
tomrf1: It's a [class](https://github.com/guardian/support-frontend/blob/main/support-models/src/main/scala/com/gu/support/encoding/Codec.scala#L12) in this repo which groups together Circe's concepts of Encoders + Decoders for handling JSON.
Here we're using implicit magic to generate the necessary JSON logic

support-frontend/app/admin/settings/LandingPageTest.scala
tomrf1: I was thinking not because on the support site we have a very strict set of regions for our pages (`/uk/contribute`, `/us/contribute` etc).
But actually this is something we should discuss first

support-frontend/assets/helpers/abTests/landingPageAbTests.ts
tomrf1: Yes I was thinking we should keep it hardcoded in the client, in case something goes wrong elsewhere

support-frontend/assets/pages/supporter-plus-landing/twoStepPages/threeTierLanding.tsx
tomrf1: Yes it's really not ideal!
This lets us pass in a string containing html, for rendering. For e.g. bold text. Although it does prevent <script> tags, there are other ways of injecting bad code into a page this way. So we need to be very confident that we control whatever is in the string.
I noticed we do [already use it elsewhere](https://github.com/guardian/support-frontend/blob/e3b1f322cb06e8b218bc5c809c4c44ea2b432686/support-frontend/assets/helpers/productPrice/promotions.tsx#L146) on the support site for rendering copy from the promos tool.
But perhaps we should be more careful now! One approach is to sanitize using e.g. https://github.com/cure53/DOMPurify.

tomrf1: And in fact we already have dompurify in this project!
I've [pushed a commit](https://github.com/guardian/support-frontend/pull/6655/commits/b0ef219766c97bfe413e96554b1aeea8f6ad758e) to use it to sanitise the strings before using them

support-frontend/app/admin/settings/LandingPageTest.scala
tomrf1: thinking about this more, we could add `countryCodes` to the `LandingPageTestTargeting` model later on. But I don't think it's necessary now

support-frontend/assets/pages/supporter-plus-landing/twoStepPages/threeTierLanding.tsx
tomrf1: Thanks for the comment!

support-frontend/assets/helpers/abTests/sessionStorage.ts
```diff
@@ -0,0 +1,32 @@
...
+const PARTICIPATIONS_KEY = 'abParticipations';
+// For participation in landing page tests, which are passed through from the server
+const LANDING_PAGE_PARTICIPATIONS_KEY = 'landingPageParticipations';
+type Key = typeof PARTICIPATIONS_KEY | typeof LANDING_PAGE_PARTICIPATIONS_KEY;
+
+function getSessionParticipations(key: Key): Participations | undefined {
+	const participations = storage.getSession(key);
+	if (participations) {
+		try {
+			return JSON.parse(participations) as Participations;
```
paul-daniel-dempsey, line 14: fyi: Not necessary but we recently setup a module for session storage, it uses `Valibot.Is` to check objects from session storage match a schema, might be of interest [shown here](https://github.com/guardian/support-frontend/blob/tf-lp-config/support-frontend/assets/pages/%5BcountryGroupId%5D/checkout/helpers/sessionStorage.ts#L31). This is a

support-frontend/assets/helpers/abTests/landingPageAbTests.ts
```diff
@@ -0,0 +1,97 @@
...
+				);
+
+				return participations;
+			}
+		}
+		// No test assigned
+		return;
+	} else {
+		// This is not a landing page, but check if the session has a landing page test participation
+		return getSessionParticipations(LANDING_PAGE_PARTICIPATIONS_KEY);
```
paul-daniel-dempsey, line 72: fyi: We have been reporting to Data Privacy re: new session storage items. Just in case.

support-frontend/assets/helpers/abTests/landingPageAbTests.ts
charleycampbell: Great idea! We could add a code comment explaining that or some sort of warning to not change the template

support-frontend/assets/pages/supporter-plus-landing/twoStepPages/threeTierLanding.tsx
charleycampbell: Excellent!

charleycampbell (approved):
This looks great Tom! First steps a an exciting feature

support-frontend/assets/helpers/page/page.ts
```diff
@@ -33,8 +33,10 @@ function getAbParticipations(): Participations {
 		countryId,
 		countryGroupId,
 		selectedAmountsVariant,
+		settings,
 	};
 	const participations: Participations = abTest.init(abtestInitalizerData);
+	console.log({ participations });
```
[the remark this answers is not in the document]
tomrf1: Not really, but I've found it really useful for checking what test participations are on each page, particularly when we want it to carry throught to the checkout. I used to use a redux browser extension for this but that's no longer an option!

support-frontend/assets/helpers/abTests/landingPageAbTests.ts
tomrf1: Thanks - I've submitted the DP form for this

PR #6661, opened 2024-12-30
Extend client side input validation to reject all emojis

tjmw:

Extending the checkout client side input validation to reject all emojis.

## Why are you doing this?

We had a case in prod where an emoji made it to the backend (and was correctly rejected). However we expected this to be rejected in the client side validation. It turns out that some emojis don't match `Extended_Pictographic` so the regex which said "anything other than `Extended_Pictographic`" let it through. We now reject `Extended_Pictographic` and `Emoji_Presentation`.

## How to test

I've extended the tests and checked in CODE.

### Before

Emoji character accepted by client side validation:

Screenshot 2024-12-30 at 12 09 19

### After

Emoji character rejected by client side validation:

Screenshot 2024-12-30 at 12 05 45

### DIFF
--- a/support-frontend/assets/helpers/__tests__/validationTest.ts
+++ b/support-frontend/assets/helpers/__tests__/validationTest.ts
@@ -1,6 +1,11 @@
 import { doesNotContainExtendedEmojiOrLeadingSpace } from 'pages/[countryGroupId]/validation';
 
-const regexToValidate = new RegExp(doesNotContainExtendedEmojiOrLeadingSpace);
+const regexToValidate = new RegExp(
+	doesNotContainExtendedEmojiOrLeadingSpace,
+	// The v flag here replicates the way a pattern attribute is used in an input element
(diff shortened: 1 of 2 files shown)


support-frontend/assets/helpers/__tests__/validationTest.ts
```diff
@@ -1,17 +1,38 @@
...
+	'v',
+);
 
 describe('validateRegex', () => {
 	it('leading space', () => {
 		expect(' z').not.toMatch(regexToValidate);
 		expect('z z').toMatch(regexToValidate);
 	});
 	it('pictographic', () => {
-		expect('\\p✈️z').not.toMatch(regexToValidate);
```
tjmw, line 11: I don't think the `\\p` was needed here.

andrewHEguardian (approved):
Great investigation 👍  let's give it a go

PR #6662, opened 2024-12-30
Support Workers Alarm: Extend go-cardless (directdebit) timeframe from 14hrs to 18hrs

paul-daniel-dempsey:

We appear to be gettting this alarm triggering more than once a month. This instigates an investigation every time. 

Looking into this months (December) alarms, an increase from 14 to 18hrs would appear to quieten this alarm whilst still ensuring it fires in good time.


tjmw (approved):
Looks good (but I think the snapshot needs updating!)

paul-daniel-dempsey:
> Looks good (but I think the snapshot needs updating!)

Thanks Tom, thats exactly what was looking for. Am wondering if this is a sufficiently low risk PR to go in or wait until Dec 6th?

PR #6664, opened 2024-12-30
Remove overlay if one time card payment fails server side

tjmw:

Ensure that the overlay is cleared on the one time checkout if a card payment fails server side.

## Why are you doing this?

Previously we noticed that if there was a server side failure on the one time checkout, you'd see an error message but the "processing payment" overlay would remain up, which meant you couldn't interact with the form to fix whatever the problem was.

We've also added some playwright coverage for this scenario.

## How to test

We've added some playwright test coverage and tested in the browser.

I've also deployed this branch to CODE and put through a card and PayPal transaction.


paul-daniel-dempsey (approved):
Looks great

PR #6665, opened 2024-12-30
Add min amount validation to server side one time payments

tjmw:

Add min amount validation to the payment API for one time contributions.

[**Trello Card**](https://trello.com/c/NXeDtm19/1305-min-max-amount-server-side-validation)

## Why are you doing this?

We already have client side validation for this, but we'd like to enforce server side too as this makes us more robust.

## How to test

I've tested this in CODE. In order to do this I had to temporarily remove the client side validation. The error message we show in this case is quite generic, but I think in general we don't expect users to hit this as the client side validation should kick in before the request even makes it to the backend.

Stripe:
Screenshot 2025-01-06 at 13 49 26

PayPal:
Screenshot 2025-01-06 at 13 50 21

If we need to, perhaps we could follow up with a more contextual error message in another PR.

### DIFF
--- a/support-payment-api/src/main/scala/model/Currency.scala
+++ b/support-payment-api/src/main/scala/model/Currency.scala
@@ -4,6 +4,10 @@ import enumeratum.{CirceEnum, Enum, EnumEntry}
 
 import scala.collection.immutable.IndexedSeq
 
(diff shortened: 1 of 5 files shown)


andrewHEguardian (approved):
Nice 👍

PR #6669, opened 2025-01-02
Can we lazy load below the fold?

andrewHEguardian:

### DIFF
--- a/support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx
@@ -1,12 +1,6 @@
 import { css } from '@emotion/react';
 import { palette, space } from '@guardian/source/foundations';
-import {
-	Checkbox,
-	Label,
-	Radio,
-	RadioGroup,
-	TextInput,
-} from '@guardian/source/react-components';
+import { Checkbox, Label, TextInput } from '@guardian/source/react-components';
 import {
 	Divider,
 	ErrorSummary,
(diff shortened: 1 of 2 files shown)


support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx
```diff
@@ -113,27 +96,19 @@ import {
...
-	checkedRadioLabelColour,
-	defaultRadioLabelColour,
-	paymentMethodBody,
-	PaymentMethodRadio,
-	PaymentMethodSelector,
-} from './paymentMethod';
 import { retryPaymentStatus } from './retryPaymentStatus';
 import { setThankYouOrder, unsetThankYouOrder } from './thankYouComponent';
 
+const PaymentSection = lazy(() => import('./paymentSection'));
```
Comment (paul-daniel-dempsey) [inline support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx:110]: Maybe a comment here on why lazy loading used?

paul-daniel-dempsey (approved):
Really nice to pull these componets out of the checkout regardless of lazy loading. Great job. Really good example of how to do it, had to check what `<suspense>` was for, good use of that too. Like this. Cron/Smoke tests also passed.

support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx
Comment (andrewHEguardian) [reply support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx:110]: Yep, if we decided to go with it

GHaberis:
Interesting that the reported results vary so much between runs on the identical build. Have you looked into alternative ways of measuring page speed performance such as Chrome DevTools' Lighthouse panel, [Lighthouse CLI](https://github.com/GoogleChrome/lighthouse) or WebPageTest to see if these offer more consistent results? 

If there's variability could we run several tests for both the baseline and the build with updates, then calculate the average and standard deviation to account for variability?

GHaberis:
Also interested to know if an external factor impacts page speed metrics, eg. I'm thinking the Sourcepoint CMP. Could we we disable the CMP when profiling the page to see whether it produces more consistent results?

PR #6673, opened 2025-01-06
Remove noUncheckedIndexedAccess override and fix all cases 

andrewHEguardian:

Remove the tsconfig override `noUncheckedIndexedAccess` and fix all the errors. This rule basically means that any array item accessed by index could always be undefined. Some of the strategies used to fix are:

- if the array is actually readonly, add `as const`
- check for undefined with an if statement (must watch out for falsy string/number values)
- use nullish coalescing `??` and fallback to an arbitrary default
- use the `as` assertion, although this is not preferred
- use `!` non-null assertion, I've only used this in the tests

[**Trello Card**](https://trello.com/c/Jqyut6Fm/1139-tsconfig-set-nouncheckedindexedaccess-true)

## Why are you doing this?
https://github.com/guardian/support-frontend/issues/5829

> We gain far more security by having noUncheckedIndexedAccess: true in [our ](https://github.com/guardian/support-frontend/blob/main/support-frontend/tsconfig.json#L23)tsconfig. Currently this is an override of the Guardian’s tsconfig rules.

## How to test

- Build passes
- Ran smoke tests

The changes should be no-op, since all the errors presumably should have worked okay at runtime before. But we may want to split this PR up into bite size chunks to merge?


support-frontend/assets/helpers/globalsAndSwitches/globals.test.ts
```diff
@@ -20,7 +20,7 @@ const getSpecifiedRegionAmountsFromGlobal = (
 	if (!testArray.length) {
 		return {};
 	}
-	return testArray[0];
+	return testArray[0] ?? {};
```
rupertbates, line 23: Could move the check for undefined into the if statement above?

andrewHEguardian: Could do, it's a test which is why I did the inelegant solution

support-frontend/assets/helpers/redux/checkout/product/selectors/productType.ts
```diff
@@ -45,7 +45,7 @@ function getSubscriptionTypeFromURL(): SubscriptionProduct {
 	};
 	const [, match] = urlPathRegex.exec(window.location.pathname) ?? [];
 	if (match) {
-		return productsToUrlPath[match];
+		return productsToUrlPath[match] ?? DigitalPack;
```
rupertbates, line 48: again, maybe include. the check for undefined in the if statement

support-frontend/assets/helpers/tracking/acquisitions.ts
```diff
@@ -285,15 +285,20 @@ function getAcquisitionDataFromUtmParams():
 	);
 
 	// All must be present in the URL for them to be accepted
-	if (utmParamNames.every((paramName) => utmParameters[paramName].length > 0)) {
+	if (
```
rupertbates, line 288: Maybe this is a better implemenatation for this function?

```typescript
function getAcquisitionDataFromUtmParams():
	| Record<string, string | AcquisitionABTest | AcquisitionQueryParameters>
	| undefined {
	// Same order of fields as https://reader-revenue-lynx.s3.eu-west-1.amazonaws.com/v3.html
  const utmCampaign = getQueryParameter('utm_campaign');
  const utmContent = getQueryParameter('utm_content');
  const utmTerm = getQueryParameter('utm_term');
  const utmSource = getQueryParameter('utm_source');
  const utmMedium = getQueryParameter('utm_medium');

	// All must be present in the URL for them to be accepted
	if (
    utmCampaign !== '' &&
    utmContent !== '' &&
    utmTerm !== '' &&
    utmSource !== '' &&
    utmMedium !== ''
	) {
		return {
			campaignCode: utmCampaign,
			abTest: {
				name: utmContent,
				variant: utmTerm,
			},
			source: utmSource,
			componentType: utmMedium,
			queryParameters: toAcquisitionQueryParameters(getAllQueryParams()),
		};
	}

	return;
}
```

support-frontend/assets/pages/digital-subscriber-checkout/setup/setUpRedux.ts
```diff
@@ -76,7 +76,7 @@ function getInitialContributionType(
 	);
 	return defaultContributionType
 		? defaultContributionType.contributionType
-		: contributionTypes[countryGroupId][0].contributionType;
+		: contributionTypes[countryGroupId][0]?.contributionType ?? 'ONE_OFF';
```
rupertbates, line 79: Are we sure this isn't changing behaviour?

support-frontend/assets/helpers/tracking/acquisitions.ts
andrewHEguardian: That looks a lot better

support-frontend/assets/pages/digital-subscriber-checkout/setup/setUpRedux.ts
andrewHEguardian: It has some kind of fallback so in practice I think it's always populated, but we could do an `as` instead to definitely not change runtime behaviour

rupertbates: I'm sure it's fine, up to you which you go for.

rupertbates (approved):
Great to get this done 👏👏👏

PR #6675, opened 2025-01-06
Prevent Guardian Ad-Lite purchase by users who have either of `adFree` or `allowRejectAll ` benefits already

tjmw:

If a non signed-in user already has the `adFree` or `allowRejectAll ` benefit, they shouldn't be able to purchase Guardian Ad-Lite. We're assuming that a signed-in user with `adFree` or `allowRejectAll ` wouldn't make it to the checkout as there's no user journey for this.

## Why are you doing this?

It doesn't make sense to buy Guardian Ad-Lite if you get ad free reading or can already reject tracking.

## How to test

I've tested locally and in CODE. When a signed out user already has the ad-free or reject-tracking benefit, an error is shown:

Screenshot 2025-01-10 at 11 55 40

A user without these benefits is able to purchase successfully.

### DIFF
--- a/support-frontend/app/config/UserBenefitsApi.scala
+++ b/support-frontend/app/config/UserBenefitsApi.scala
@@ -1,8 +1,0 @@
-package config
-
-import com.typesafe.config.Config
-
-class UserBenefitsApi(config: Config) {
-  lazy val host = config.getString("host")
-  lazy val apiKey = config.getString("apiKey")
-}
--- a/support-frontend/app/config/UserBenefitsApi.scala
+++ b/support-frontend/app/config/UserBenefitsApi.scala
@@ -1,0 +1,8 @@
+package config
+
+import com.typesafe.config.Config
+
+class UserBenefitsApi(config: Config) {
+  lazy val host = config.getString("host")
+  lazy val apiKey = config.getString("apiKey")
+}
--- a/support-frontend/app/controllers/CreateSubscriptionController.scala
+++ b/support-frontend/app/controllers/CreateSubscriptionController.scala
@@ -152,7 +152,9 @@ class CreateSubscriptionController(
     }
   }
 
-  private def validateBenefits(response: UserBenefitsResponse): EitherT[Future, CreateSubscriptionError, Unit] = {
+  private def validateBenefitsForAdLitePurchase(
+      response: UserBenefitsResponse,
+  ): EitherT[Future, CreateSubscriptionError, Unit] = {
     if (response.benefits.contains("adFree") || response.benefits.contains("allowRejectAll")) {
       // Not eligible
       EitherT.leftT(RequestValidationError("guardian_ad_lite_purchase_not_allowed"))
@@ -178,7 +180,7 @@ class CreateSubscriptionController(
               .forUser(testUsers.isTestUser(request))
               .getUserBenefits(userDetails.userDetails.identityId)
               .leftMap(_ => ServerError("Something went wrong calling the user benefits API"))
-            _ <- validateBenefits(benefits)
+            _ <- validateBenefitsForAdLitePurchase(benefits)
           } yield ()
         }
       }
--- a/support-frontend/app/controllers/CreateSubscriptionController.scala
+++ b/support-frontend/app/controllers/CreateSubscriptionController.scala
@@ -31,6 +31,7 @@ import services.{
   RecaptchaService,
   TestUserService,
   UserBenefitsApiService,
+  UserBenefitsApiServiceProvider,
   UserBenefitsResponse,
   UserDetails,
 }
@@ -71,7 +72,7 @@ class CreateSubscriptionController(
     components: ControllerComponents,
     guardianDomain: GuardianDomain,
     paperRoundServiceProvider: PaperRoundServiceProvider,
-    userBenefitsApiService: UserBenefitsApiService,
+    userBenefitsApiServiceProvider: UserBenefitsApiServiceProvider,
 )(implicit val ec: ExecutionContext, system: ActorSystem)
     extends AbstractController(components)
     with Circe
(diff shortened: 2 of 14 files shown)


support-frontend/app/controllers/CreateSubscriptionController.scala
```diff
@@ -142,13 +151,57 @@ class CreateSubscriptionController(
...
+      case GuardianLight(_) => {
+        if (userDetails.isSignedIn) {
+          // If the user is signed in, we'll assume they're eligible as they shouldn't have got
+          // to this point of the journey.
+          EitherT.rightT(())
+        } else {
+          for {
+            benefits <- userBenefitsApiService
+              .getUserBenefits(userDetails.userDetails.identityId)
+              .leftMap(_ => ServerError("Something went wrong calling the user benefits API"))
```
tjmw, line 178: If there's an issue with calling the user benefits API the purchase will be blocked and we'll show a generic error message to the user.

support-frontend/assets/helpers/forms/errorReasons.ts
```diff
@@ -83,6 +84,9 @@ function appropriateErrorMessage(errorReason: string): string {
 
 			case 'recaptcha_validation_failed':
 				return 'Please prove you are not a robot';
+
+			case 'guardian_light_purchase_not_allowed':
+				return 'Guardian Light purchase not allowed copy TBD';
```
tjmw, line 89: TODO: Update copy.

support-frontend/app/controllers/CreateSubscriptionController.scala
```diff
@@ -142,13 +151,57 @@ class CreateSubscriptionController(
...
+      userDetails: UserDetails,
+      isSignedIn: Boolean,
+  )
+
   private def createSubscription(implicit
       settings: AllSettings,
       request: CreateRequest,
   ): EitherT[Future, CreateSubscriptionError, CreateSubscriptionResponse] = {
 
     val maybeLoggedInUserDetails =
```
andrewHEguardian, line 197: Should this now be "maybeUserDetails" since the signed in status can vary?

tjmw: I had this same thought and decided to leave it as is because in the case of this `val` it's still either:

1) `None` or
2) The details of a user who is signed in.

I.e. `maybeLoggedInUserDetails` won't ever contain information about a user who isn't signed in. Though like you say it's true that the `Option` type `UserDetailsWithSignedInStatus` can represent this state. What do you reckon?

andrewHEguardian: ah, right. That's quite subtle but seems fine to leave as is

rupertbates:
Does this work for test users? We're going to need to query the code version of user-benefits api I think.

tjmw:
TODO: update the benefit checked after https://github.com/guardian/support-service-lambdas/pull/2645 is merged.

support-frontend/app/controllers/CreateSubscriptionController.scala
```diff
@@ -142,13 +152,58 @@ class CreateSubscriptionController(
...
+    request.body.product match {
+      case GuardianAdLite(_) => {
+        if (userDetails.isSignedIn) {
+          // If the user is signed in, we'll assume they're eligible as they shouldn't have got
+          // to this point of the journey.
+          EitherT.rightT(())
+        } else {
+          for {
+            benefits <- userBenefitsApiServiceProvider
+              .forUser(testUsers.isTestUser(request))
```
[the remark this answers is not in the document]
tjmw: Good question! Depending on whether the current user is a test user it'll return a `UserBenefitsApiService` instance configured to talk to the correct environment. The only env this will differ in is prod, where a real user will talk to the prod API and a test users will talk to the code API.

andrewHEguardian (approved):

support-frontend/app/controllers/CreateSubscriptionController.scala
```diff
@@ -157,17 +214,21 @@ class CreateSubscriptionController(
         case Some(userDetails) => EitherT.pure[Future, CreateSubscriptionError](userDetails)
         case None =>
           getOrCreateIdentityUser(request.body, request.headers.get("Referer"))
+            .map(userDetails => UserDetailsWithSignedInStatus(userDetails, isSignedIn = false))
             .leftMap(mapIdentityErrorToCreateSubscriptionError)
       }
-      supportWorkersUser = buildSupportWorkersUser(userDetails, request.body, testUsers.isTestUser(request))
+
+      _ <- validateUserIsEligibleForPurchase(request, userDetails)
```
rupertbates, line 221: Can this validation not be folded into the validate function on line 155 which is already set up to run product specific validation?

support-frontend/app/config/UserBenefitsApi.scala
```diff
@@ -0,0 +1,8 @@
+package config
+
+import com.typesafe.config.Config
+
+class UserBenefitsApi(config: Config) {
```
rupertbates, line 5: What is this class for? It seems to be just the same as the UserBenefitsApiConfig case class.

tjmw: Removed in ec11d92a6ca7d2655d8390e269e0f90adc1d7068.

support-frontend/app/controllers/CreateSubscriptionController.scala
tjmw: I've looked into this. I think it could be done, but would require reordering some things. Right now the validation performed by the existing `validate` method doesn't require any information about the current user, so can happen before the `getOrCreateIdentityUser` call. Validating whether the user is eligible for purchasing Ad-Lite requires an identity ID, so has to happen after that call. I guess this keeps that initial `validate` method quite lightweight and fast to fail. We'd have to add `UserDetails` to the signature of `validate`. So keeping them separate seems okay to me, what do you reckon?

rupertbates (approved):
Looks good 👍

PR #6676, opened 2025-01-07
Knip - Production mode

andrewHEguardian:

Remove unused code found using knip in [production mode](https://knip.dev/features/production-mode), and add our entry points as the production code using the `!` syntax. This reveals code used only by tests, obviously some of these are necessary (eg mocks) so we shouldn't run our GHA in production mode.

## How to test

Run `yarn knip --production` - see files and dependencies used by tests
image


support-frontend/knip.config.js
```diff
@@ -1,11 +1,11 @@
 const entryPoints = require('../webpack.entryPoints').common;
 
 const flattenedEntryPoints = [...Object.keys(entryPoints)].map(
-	(key) => `assets/${entryPoints[key]}`,
+	(key) => `assets/${entryPoints[key]}!`,
```
tjmw, line 4: So the `!` here is a signal to knip that this is production code?

tjmw (approved):
👏  Great!

support-frontend/knip.config.js
andrewHEguardian: Exactly, is it worth a comment or [link](https://knip.dev/features/production-mode#:~:text=and%20project%20patterns-,suffixed%20with%20!,-Only%20production%20entry)?

tjmw: Yeah I think a comment/link would be good 👍

PR #6680, opened 2025-01-08
Initialise ab participations once

tomrf1:
The existing `setUpTrackingAndConsents` function calls `abTest.init` in order to register ab tests with ophan.
We then call `abTest.init` again in various places, to decide what to render based on the ab participations.
This means we're running the same logic twice per pageview, and we call `abTest.init` from an awful lot of places.

With this refactor, we instead call `abTest.init` once at the start, and pass this down into the places it's needed.
This means `setUpTrackingAndConsents` now takes a `participations` parameter.

Also, any pages which are still using Redux already have the ab participations available in the store. So in these cases we now use this.

Context: I'm trying to simplify ab test participations logic ahead of some work on the landing page.

### DIFF
--- a/support-frontend/assets/helpers/page/page.ts
+++ b/support-frontend/assets/helpers/page/page.ts
@@ -14,7 +14,16 @@ import {
 import { getReferrerAcquisitionData } from 'helpers/tracking/acquisitions';
 import { getSettings } from '../globalsAndSwitches/globals';
 
-function setUpTrackingAndConsents(): void {
+function setUpTrackingAndConsents(participations: Participations): void {
+	const countryId: IsoCountry = Country.detect();
+	const acquisitionData = getReferrerAcquisitionData();
+
+	void consentInitialisation(countryId);
+	analyticsInitialisation(participations, acquisitionData);
+	sendConsentToOphan();
+}
+
+function getAbParticipations(): Participations {
 	const settings = getSettings();
 	const countryId: IsoCountry = Country.detect();
 	const countryGroupId: CountryGroupId = CountryGroup.detect();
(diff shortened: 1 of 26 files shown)


support-frontend/assets/helpers/page/page.ts
```diff
@@ -14,29 +14,35 @@ import {
...
+function getAbParticipations(): Participations {
 	const settings = getSettings();
 	const countryId: IsoCountry = Country.detect();
 	const countryGroupId: CountryGroupId = CountryGroup.detect();
 	const abtestInitalizerData = {
 		countryId,
 		countryGroupId,
 	};
+
 	const participations: Participations = abTest.init(abtestInitalizerData);
```
tjmw, line 35: So this is now the only place this is called?

support-frontend/assets/pages/error/error404.tsx
```diff
@@ -4,7 +4,7 @@ import { renderPage } from 'helpers/rendering/render';
 import ErrorPage from './components/errorPage';
 
 // ----- Page Startup ----- //
-setUpTrackingAndConsents();
+setUpTrackingAndConsents({});
```
tjmw, line 7: It's okay to pass an empty object here because presumably we won't be running any A/B tests on this page?

tjmw (approved):
I think this looks good! 👍  I'm not that familiar with the A/B testing setup on support-frontend yet, so it might be worth waiting for a ✅ from someone else on the team before merging.

support-frontend/assets/helpers/page/page.ts
tomrf1: It's now called from 2 places:
1. here (for non-redux pages)
2. the place where the redux store is initialised

support-frontend/assets/pages/error/error404.tsx
tomrf1: yes I've assumed AB testing is pointless on these types of pages!

GHaberis (approved):
Nice work 👍

PR #6684, opened 2025-01-10
PROD payment api needs to be able to write to CODE SupporterProductData dynamo table for test users

rupertbates:
We are seeing regular errors in the payment-api logs whenever a test user makes a one time contribution. This is because we record these contributions in the SupporterProductData DynamoDB table, and for test users these are written to the CODE version of the table. However currently the PROD app doesn't have permissions to write to the CODE table so there is an exception. This PR fixes the issue.


andrewHEguardian (approved):
Nice 👍  is it worth leaving a comment to explain why?

PR #6685, opened 2025-01-10
One-time checkout: prevent large other amount causing error

andrewHEguardian:

Add a max amount check to the 'Other' amount in one time checkout. If it is above the maximum - it does not go into the `finalAmount` variable where it can be used to set the amount in the rest of the page. This stops it reaching Stripe's JS where it could be above the maximum.

[**Trello Card**](https://trello.com/c/HPxFALFh/1360-revisit-client-side-max-validation-on-amount-field-in-one-time-checkout)

## How to test

Type in a _very_ large number in the other amount, the amount is not listed in the payment button and does not crash the page.

## Screenshots

Before:
(Above our max)
image

(Above Stripe's max)
image

After:
image


tjmw (approved):
Nice, this looks great 👍

PR #6628, opened 2024-12-11
Routing: lazy loading CountryGroup Routes

paul-daniel-dempsey:
We tested this branch vs main on `CODE` using [webpage test](https://www.webpagetest.org/) to compare page speed performance metrics when viewing the pages below supplied through the router (simple / median of 3 x Motorola 4G page loads). I tested these entry points->

- [checkout](https://support.code.dev-theguardian.com/uk/checkout?product=SupporterPlus&ratePlan=Monthly)
- [one-time-checkout](https://support.code.dev-theguardian.com/uk/one-time-checkout)
- [guardian-light landing page
](https://support.code.dev-theguardian.com/uk/guardian-light)

**Conclusion**

1. Speed results in-conclusive
2. Revisit this in a new PR after server-side Ab-Testing Available. 
3. This should work out if a 'faster' variant is more succesful.

[(Summary WorkSheet)](https://docs.google.com/spreadsheets/d/1TW4qLQSPhX-RFWOyVD0Tog_HUwXpUxJZoC6_A0VDzn8/edit?usp=sharing)

**Results: Checkout [(16/12/24)](https://www.webpagetest.org/video/compare.php?tests=241216_AiDcFW_97Y,241216_AiDcTS_9K0,241216_BiDcRE_8V0)**

| Metric    | Router | lazyRouter| ghlazyRouter |
| -------- | ------- | ------- | ------- | 
| FCP (How soon did text and images start to appear) | 1.577s | 2.371s | 1.873s |
| Speed Index (Speed Index (How soon did the page look usable) | 10.861s | 11.241s | 11.07s |
| LCP (When did the largest visible content finish loading) | 5.385s | 6.47s | 5.271s |
| Page Weight (How many bytes downloaded) | 2249kb |2310kb | 2294Kb |
| router.js g-zip | 264kb | 93kb+162Kb | 81.5Kb + 162Kb |

**Results: OneTimeCheckout [(16/12/24)](https://www.webpagetest.org/video/compare.php?tests=241216_AiDc3D_98G,241216_BiDcH6_9BC,241216_BiDcC4_8VX)**

| Metric    | Router | lazyRouter| ghlazyRouter |
| -------- | ------- | ------- | ------- | 
| FCP (How soon did text and images start to appear) | 1.652s | 1.761 s | 1.709s |
| Speed Index (Speed Index (How soon did the page look usable) | 11.532s | 10.38s | 10.462 |
[...]

PR #6690, opened 2025-01-14
Create server side test for lazy loading router

andrewHEguardian:

Creating a server side A/B test: the variant is a lazy loading router previously implemented in #6628, the control is the router as normal.

[**Trello Card**](https://trello.com/c/ZWtVE5Xg/1313-spike-server-side-performance-a-b-test)

## Why are you doing this?

We want to test the changes by seeing whether better page performance improves our conversion and business metrics

## How to test

We will probably use the normal A/B test dashboard for results. The participation in the test is rendered as a script tag in the HTML and is passed to the normal participations object.

## Screenshots
image

### DIFF
--- a/support-frontend/app/controllers/Application.scala
+++ b/support-frontend/app/controllers/Application.scala
@@ -496,7 +496,7 @@ class Application(
   def productCheckoutRouter(countryGroupId: String) = MaybeAuthenticatedAction { implicit request =>
     implicit val settings: AllSettings = settingsProvider.getAllSettings()
     val geoData = request.geoData
-    val serversideTests = generateParticipations(Nil)
+    val serversideTests = generateParticipations(List("lazyLoading"))
     val isTestUser = testUserService.isTestUser(request)
     // This will be present if the token has been flashed into the session by the PayPal redirect endpoint
     val guestAccountCreationToken = request.flash.get("guestAccountCreationToken")
--- a/support-frontend/app/views/router.scala.html
+++ b/support-frontend/app/views/router.scala.html
@@ -1,4 +1,4 @@
-@import admin.ServersideAbTest.Participation
+@import admin.ServersideAbTest.{Participation, Variant}
 @import assets.RefPath
 @import assets.AssetsResolver
 @import admin.settings.AllSettings
@@ -23,7 +23,7 @@
 @main(
   title = "Support the Guardian | Checkout",
   mainElement = assets.getSsrCacheContentsAsHtml(divId = "router", file = "ssr-holding-content.html"),
-  mainJsBundle = RefPath("[countryGroupId]/router.js"),
+  mainJsBundle = if (serversideTests.get("lazyLoading").contains(Variant)) RefPath("[countryGroupId]/lazyRouter.js")  else RefPath("[countryGroupId]/router.js"),
   mainStyleBundle = None,
   description = Some("Help us deliver the independent journalism the world needs. Support the Guardian by making a contribution."),
   canonicalLink = Some("https://support.theguardian.com/checkout"),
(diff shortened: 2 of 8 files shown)


support-frontend/assets/pages/[countryGroupId]/lazyRouter.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { HoldingContent } from 'components/serverSideRendered/holdingContent';
+import { parseAppConfig } from 'helpers/globalsAndSwitches/window';
+import {
+	getAbParticipations,
+	setUpTrackingAndConsents,
+} from 'helpers/page/page';
+import { renderPage } from 'helpers/rendering/render';
+import { geoIds } from 'pages/geoIdConfig';
+
+const abParticipations = getAbParticipations();
```
Comment (andrewHEguardian) [inline support-frontend/assets/pages/[countryGroupId]/lazyRouter.tsx:12]: Given we get the participations before creating the router maybe this could be a client side test? But this would probably affect performance

support-frontend/app/controllers/Application.scala
```diff
@@ -496,7 +496,7 @@ class Application(
   def productCheckoutRouter(countryGroupId: String) = MaybeAuthenticatedAction { implicit request =>
     implicit val settings: AllSettings = settingsProvider.getAllSettings()
     val geoData = request.geoData
-    val serversideTests = generateParticipations(Nil)
+    val serversideTests = generateParticipations(List("lazyLoading"))
```
paul-daniel-dempsey, line 499: Presume this function creates the `control` and `variant` for use later.

support-frontend/app/views/router.scala.html
```diff
@@ -23,7 +23,7 @@
 @main(
   title = "Support the Guardian | Checkout",
   mainElement = assets.getSsrCacheContentsAsHtml(divId = "router", file = "ssr-holding-content.html"),
-  mainJsBundle = RefPath("[countryGroupId]/router.js"),
+  mainJsBundle = if (serversideTests.get("lazyLoading").contains(Variant)) RefPath("[countryGroupId]/lazyRouter.js")  else RefPath("[countryGroupId]/router.js"),
```
paul-daniel-dempsey, line 26: Ah, this is where it makes the decision of what router to take. Should I be able to find these in the network tab under `*.js` bundles?

paul-daniel-dempsey (commented):
Looks good, tests perfectly. Just a couple of questions on the comments if you have time.

paul-daniel-dempsey (approved):

support-frontend/app/controllers/Application.scala
andrewHEguardian: Correct yes, it's a 50/50 chance of assigning control or variant

support-frontend/app/views/router.scala.html
andrewHEguardian: Indeed 👍

support-frontend/assets/pages/[countryGroupId]/lazyRouter.tsx
Comment (andrewHEguardian) [reply support-frontend/assets/pages/[countryGroupId]/lazyRouter.tsx:12]: I'm wondering if we could have done this client side - check the test then render the right router. But I'm assuming this would negate the performance benefit completely?

Comment (GHaberis) [reply support-frontend/assets/pages/[countryGroupId]/lazyRouter.tsx:12]: Hmm possibly, we would have to move the routing logic out of this entryb point and into 2 modules eg. a `nonLazyRouter` and  a`lazyRouter` and conditionally lazy load  one of these based on the AB test participation, hard to know how that would impact performance, however this server side approach is testing something closer to what would actually be implemented vs the aforementioned client-side approach so is worth measuring.

GHaberis:
@andrewHEguardian this is great to see, a first attempt at using the server side AB testing functionality that I can recall! Did you confirm this server side AB test is tracked with the Ophan page view (eg. in the Ophan abTestRegister request), and also in the payload sent on form submission (eg. the /create payload for the generic checkout)

andrewHEguardian:
> abTestRegister

Yes it is, the AB test is loaded into the window.guardian and read from there

image

PR #6698, opened 2025-01-16
ab-test: contributions-only tracking passes through landings to checkout, one-time-checkout & thankyou

paul-daniel-dempsey:

Pageview tracking added to generic checkout and one-time checkout

[**Trello Card**](https://trello.com/c/aQtJpr7y/1314-build-contributions-only-landing-page)

## How to test

Set `abDefinition.ts:ab-contributionsOnly` test visibiility to `true`
https://support.thegulocal.com/int/contribute?country=lb#ab-contributionsOnly=control 
https://support.thegulocal.com/int/contribute?country=lb#ab-contributionsOnly=variant 

## Screenshots

|test|landing|one-time-checkout|checkout|
|-----|-----|-----|-----|
|control|image|image|image|
|variant|image|image|image|

### DIFF
--- a/support-frontend/assets/helpers/abTests/abtest.ts
+++ b/support-frontend/assets/helpers/abTests/abtest.ts
@@ -127,7 +127,7 @@ function init({
 		selectedAmountsVariant,
 		sessionParticipations,
 	);
-	console.log('*** init.participations', participations);
+
 	const urlParticipations = getParticipationsFromUrl();
 	const serverSideParticipations = getServerSideParticipations();
 	return {
@@ -171,10 +171,9 @@ function getParticipations(
 	sessionParticipations: Participations = {},
 ): Participations {
 	const participations: Participations = {};
+
 	Object.entries(abTests).forEach(([testId, test]) => {
-		console.log('*** testId:', testId);
 		if (!test.isActive) {
-			console.log('*** getParticipations 0');
 			return;
 		}
 
@@ -224,7 +223,6 @@ function getParticipations(
 			selectedAmountsVariant.testName !== contributionsOnlyAmountsTestName &&
 			includeOnlyContributionsOnlyCountries
 		) {
-			console.log('*** getParticipations 6');
 			return;
 		}
 
@@ -242,20 +240,13 @@ function getParticipations(
 		}
 
 		const variantAssignment = assignUserToVariant(mvtId, test, participation);
-		console.log(
-			'*** getParticipations 9, test, participation',
-			test,
-			participation,
-		);
+
 		if (variantAssignment.type === 'NOT_ASSIGNED') {
 			return;
 		}
-		console.log(
-			'*** getParticipations 9, variantAssignment',
-			variantAssignment,
-		);
+
 		const testVariantId = test.variants[variantAssignment.variantIndex]?.id;
-		console.log('*** getParticipations 9, testVariantId', testVariantId);
+
 		if (testVariantId) {
 			participations[testId] = testVariantId;
 		}
@@ -651,14 +642,9 @@ function assignUserToVariant(
 ): VariantAssignment {
 	// For non-referrrer controlled tests we assign the user randomly
 	if (participation.type === 'ACTIVE_PARTICIPATION') {
-		console.log(
-			'*** assignUserToVariant.ACTIVE_PARTICIPATION (non-referrrer controlled)',
-			randomNumber(mvtId, test.seed) % test.variants.length,
-			assigned(randomNumber(mvtId, test.seed) % test.variants.length),
-		);
 		return assigned(randomNumber(mvtId, test.seed) % test.variants.length);
 	}
-	console.log('*** assignUserToVariant. referrrer controlled ');
+
 	// For referrrer controlled tests the assignment comes from the acquisition data.
 	// If the variant in the acquisition data doesn't match up with any defined in
 	// the test, we log an error and don't assign the user to the test.
--- a/support-frontend/assets/helpers/abTests/abtest.ts
+++ b/support-frontend/assets/helpers/abTests/abtest.ts
@@ -340,8 +340,11 @@ function getAmountsTestVariant(
 			path,
 			'/??/checkout|one-time-checkout|contribute|thankyou(/.*)?$',
 		);
-
-		if (pathMatches && test.variants.length > 1 && test.isLive) {
+		console.log('*** testName: variantName', testName, variantName);
+		console.log('*** pathMatches, path', pathMatches, path);
+		console.log('*** test.variants', test.variants);
+		// if (pathMatches && test.variants.length > 1 && test.isLive) {
+		if (pathMatches && test.isLive) {
 			return {
 				[testName]: variantName,
 			};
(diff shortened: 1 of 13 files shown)


support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -338,7 +338,7 @@ function getAmountsTestVariant(
 		// Check if we actually want to track this test
 		const pathMatches = targetPageMatches(
 			path,
-			'/??/contribute|thankyou(/.*)?$',
```
GHaberis, line 341: @tomrf1 fyi this is the change we discussed yesterday

GHaberis: This will mean we now start tracking participation in multi-variant amounts tests on the generic checkout & one-time-checkout in Ophan page views.

paul-daniel-dempsey:
> @paul-daniel-dempsey do we not need to pass the `selectedAmountsVariant` in the `abtestInitalizerData` here: https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/helpers/page/page.ts#L32 still?
> 
> I believe this is necessary in order for the ab test initialisation to know whether a user is paricipating the "Contributions Only" amounts test on pages that don't use Redux, such as the generic checkout and one-time-checkout?
> 
> Without this change a user in the the "Contributions Only" amounts test won't be allocated the `contributinsOnly` AB test on the generic checkout and one-time-checkout, we can validate this by checking the Ophan `abTestRegister` tracking event and the payload when we submit the form.

I think the way Tom did this, by passing through and saving `abParticipation` in session storage, have done in both Landing pages, this ensures the checkout and one-time-checkout can access  'ab-contributionsOnly' test. Have tested this as you suggest above and it is in the checkout Ophan `abTestRegister` and passed through in the create (and in the thankyou Ophan `abTestRegister`).

GHaberis:
> > @paul-daniel-dempsey do we not need to pass the `selectedAmountsVariant` in the `abtestInitalizerData` here: https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/helpers/page/page.ts#L32 still?
> > I believe this is necessary in order for the ab test initialisation to know whether a user is paricipating the "Contributions Only" amounts test on pages that don't use Redux, such as the generic checkout and one-time-checkout?
> > Without this change a user in the the "Contributions Only" amounts test won't be allocated the `contributinsOnly` AB test on the generic checkout and one-time-checkout, we can validate this by checking the Ophan `abTestRegister` tracking event and the payload when we submit the form.
> 
> @GHaberis I think the way Tom did this, by passing through and saving `abParticipation` in session storage, have done in both Landing pages, this ensures the checkout and one-time-checkout can access 'ab-contributionsOnly' test. Have tested this as you suggest above and it is in the checkout Ophan `abTestRegister` and passed through in the create (and in the thankyou Ophan `abTestRegister`).

Ah yes I see how the session storage would mean the participation is tracked on the checkouts although I have a feeling my PR here will stop it working! https://github.com/guardian/support-frontend/pull/6691 😆, lets discuss after stand-up tomorrow.

paul-daniel-dempsey:
> > > @paul-daniel-dempsey do we not need to pass the `selectedAmountsVariant` in the `abtestInitalizerData` here: https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/helpers/page/page.ts#L32 still?
> > > I believe this is necessary in order for the ab test initialisation to know whether a user is paricipating the "Contributions Only" amounts test on pages that don't use Redux, such as the generic checkout and one-time-checkout?
> > > Without this change a user in the the "Contributions Only" amounts test won't be allocated the `contributinsOnly` AB test on the generic checkout and one-time-checkout, we can validate this by checking the Ophan `abTestRegister` tracking event and the payload when we submit the form.
> > 
> > 
> > @GHaberis I think the way Tom did this, by passing through and saving `abParticipation` in session storage, have done in both Landing pages, this ensures the checkout and one-time-checkout can access 'ab-contributionsOnly' test. Have tested this as you suggest above and it is in the checkout Ophan `abTestRegister` and passed through in the create (and in the thankyou Ophan `abTestRegister`).
> 
> Ah yes I see how the session storage would mean the participation is tracked on the checkouts although I have a feeling my PR here will stop it working! #6691 😆, lets discuss after stand-up tomorrow.

Yes, your right, perhaps not best option to rely on session storage especially if can switch it off it future. I think ensuring the abTest initialization has the `testAmountsVariant` and/or checking for its existance should go back in

support-frontend/assets/helpers/abTests/abtest.ts
```diff
@@ -219,7 +219,8 @@ function getParticipations(
 		 * if the  the ab test definition has includeOnlyContributionsOnlyCountries as true
 		 */
 		if (
-			selectedAmountsVariant?.testName !== contributionsOnlyAmountsTestName &&
+			selectedAmountsVariant &&
```
paul-daniel-dempsey, line 222: `selectedAmountsVariant` is still an optional parameter within this call. This ensures one exists for the not equal check below👇

GHaberis: What if `selectedAmountsVariant` is undefined and  `includeOnlyContributionsOnlyCountries` is `true` for the test? It looks like we'd now include a user in the `includeOnlyContributionsOnlyCountries` AB test in this case? Not sure that's what we'd want to do. 

Before this change we would have not allocated users to a test where `includeOnlyContributionsOnlyCountries` is `true`, if:

- `selectedAmountsVariant` was undefined OR
-  `selectedAmountsVariant.testName` was not `contributionsOnlyAmountsTestName`

Which would be what I'd expect? We can probably right a unit test to confirm this?

paul-daniel-dempsey: Yes thats a case I had not thought of. I have added the test change and the addition to `abtest.ts` to pick this up this scenario. Hopefully looks better now?

GHaberis: OK good to see and AB test to cover this scenario, but I actually don't think you need the extra if condition you've added below to handle it:

```
// Exclude undefined users assigned to the contributions only amounts test
		if (!selectedAmountsVariant && includeOnlyContributionsOnlyCountries) {
			return;
		}
```

What's the issue with the original if check without the `selectedAmountsVariant &&` condition? I think this original condition was enough:

```
		if (
			selectedAmountsVariant?.testName !== contributionsOnlyAmountsTestName &&
			includeOnlyContributionsOnlyCountries
		) {
			return;
		}
```

GHaberis (approved):
Nice 👍

support-frontend/assets/helpers/page/page.ts
```diff
@@ -27,11 +27,16 @@ function getAbParticipations(): Participations {
...
+		settings,
+	);
 	const abtestInitalizerData = {
 		countryId,
 		countryGroupId,
+		selectedAmountsVariant,
 	};
-
 	const participations: Participations = abTest.init(abtestInitalizerData);
 	const { amountsParticipation } = getAmountsTestVariant(
```
GHaberis, line 41: We're now calling `getAmountsTestVariant` twice, you can remove the 2nd call, and just update the first to be:

```
const { amountsParticipation, selectedAmountsVariant } = getAmountsTestVariant(
		countryId,
		countryGroupId,
		settings,
	);
	```

paul-daniel-dempsey: Ah yes, have neatened to this now, done

support-frontend/assets/helpers/abTests/abtest.ts
paul-daniel-dempsey: updated and removed unecessary condition

PR #6707, opened 2025-01-20
Reinstate one-time contribution cookie.

rupertbates:
We set a cookie after a successful one-time contribution on support-fronted so that the contributor will not see any support messaging on dotcom for 90 days.

In https://github.com/guardian/support-frontend/pull/6643 I switched from using the one-time contribution cookie for this to the gu_hide_support_messaging cookie which is used by dotcom to store the value returned from members-data-api/user-benefits api.

However dotcom will delete this cookie if it exists for non signed in users, so it is not suitable without reworking the behaviour in dotcom which I would rather not do at this point.

This PR reverts the earlier PR, taking account of the fact that the cookie is now set in the new one-time checkout thank you page rather than the old thank you page.


andrewHEguardian (approved):
👍

paul-daniel-dempsey (approved):
LG

PR #6710, opened 2025-01-20
Check if hashUrlSearchParams has p[meta_page_view_id]

GHaberis:

An alternate log to https://github.com/guardian/support-frontend/pull/6694, as the original log didn't report anything in ~5 days. I wondered whether `hashUrlSearchParams` contained a `'p[meta_page_view_id]` key as we'd expect, we can check this using https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/has. If not log an expection in Sentry.

### DIFF
--- a/support-frontend/assets/pages/[countryGroupId]/events/events.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/events/events.tsx
@@ -120,7 +120,10 @@ export function Events({ geoId }: Props) {
 		hashUrlSearchParams.toString(),
 	)}`;
 
-	if (!hashUrlSearchParams.has('p[meta_page_view_id]', pageviewId)) {
+	if (
+		!hashUrlSearchParams.has('p[meta_page_view_id]') ||
+		hashUrlSearchParams.get('p[meta_page_view_id]') !== pageviewId
+	) {
 		logException(`hashUrlSearchParams pageviewId mismatch: ${pageviewId}`);
 	}
 
--- a/support-frontend/assets/pages/[countryGroupId]/events/events.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/events/events.tsx
@@ -120,8 +120,8 @@ export function Events({ geoId }: Props) {
 		hashUrlSearchParams.toString(),
 	)}`;
 
-	if (!hashUrlSearchParams.has('p[meta_page_view_id]')) {
-		logException('pageviewId not available in hashUrlSearchParams');
+	if (!hashUrlSearchParams.has('p[meta_page_view_id]', pageviewId)) {
+		logException(`hashUrlSearchParams pageviewId mismatch: ${pageviewId}`);
 	}
 
 	const embedUrl = `${ticketTailorUrl}/${eventId}/book${presetDataUrl}`;
--- a/support-frontend/assets/pages/[countryGroupId]/events/events.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/events/events.tsx
@@ -104,10 +104,6 @@ export function Events({ geoId }: Props) {
 
 	const pageviewId = getPageViewId();
 
-	if (!pageviewId) {
-		logException('pageviewId not available on event listing');
-	}
-
 	const hashUrlSearchParams = new URLSearchParams({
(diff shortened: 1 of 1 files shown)


support-frontend/assets/pages/[countryGroupId]/events/events.tsx
```diff
@@ -124,6 +120,10 @@ export function Events({ geoId }: Props) {
 		hashUrlSearchParams.toString(),
 	)}`;
 
+	if (!hashUrlSearchParams.has('p[meta_page_view_id]')) {
```
Comment (tjmw) [inline support-frontend/assets/pages/[countryGroupId]/events/events.tsx:123]: Not sure if we'd want to use it here, but apparently `.has` takes an optional second arg which represents the expected value. So we could use that to verify not only that the key is set, but also that it has the value we want:

```
if (!hashUrlSearchParams.has('p[meta_page_view_id]', pageViewId)) {
```

https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/has

tjmw (approved):
Looks good!

I left one tiny suggestion, but not sure if it's useful here.

support-frontend/assets/pages/[countryGroupId]/events/events.tsx
Comment (GHaberis) [reply support-frontend/assets/pages/[countryGroupId]/events/events.tsx:123]: Nice idea! I'll do this 👍

Comment (GHaberis) [reply support-frontend/assets/pages/[countryGroupId]/events/events.tsx]: I've added this 2nd parameter to my `has()` call but the type checking is now returning an error: `Expected 1 arguments, but got 2`, very odd as the type definition for `has()` includes an optional 2nd `value` parameter.

Comment (GHaberis) [reply support-frontend/assets/pages/[countryGroupId]/events/events.tsx]: Screenshot 2025-01-21 at 10 07 31

Comment (GHaberis) [reply support-frontend/assets/pages/[countryGroupId]/events/events.tsx]: This discussion highlights the same issue: https://github.com/microsoft/TypeScript/issues/55569, the conclusion is `has()` is not adequately supported, in this situation the 'algorithm' can decide that a feature is not well supported by browsers and thus removes it: https://github.com/microsoft/TypeScript-DOM-lib-generator#when-the-type-is-missing. However the article suggests this happens if a feature is not supported by two or more browser engines, however the MDB for `has()` shows it is supported in all browsers latest versions: https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/has so I wouldn't expect the algorithm to flag it as unsupported.

A temporary workaround could be to this to our any.d.ts file:

```
global {
  interface URLSearchParams {
    has(name: string, value?: any): boolean;
  }
}
```

@tjmw what do you think?

Comment (tjmw) [reply support-frontend/assets/pages/[countryGroupId]/events/events.tsx]: Weird! Yeah for me VSCode is happy, it looks like its using it's own dom types, but running `yarn validate` I'm seeing the same error. I can't see anything in our tsconfig which would determine the dom types used, maybe it's something we're getting from `@guardian/tsconfig`?

Comment (GHaberis) [reply support-frontend/assets/pages/[countryGroupId]/events/events.tsx]: > Weird! Yeah for me VSCode is happy, it looks like its using it's own dom types, but running `yarn validate` I'm seeing the same error. I can't see anything in our tsconfig which would determine the dom types used, maybe it's something we're getting from `@guardian/tsconfig`?

Yeah I wondered whether there's a compatibility issue between our minimum supported browsers and the feature, I couldn't find where in the Typescript config we define minimum supported browser versions. I'll check `@guardian/tsconfig` now!

Comment (GHaberis) [reply support-frontend/assets/pages/[countryGroupId]/events/events.tsx]: Nothing in `support-frontend/tsconfig.json` the `@guardian/tsconfig` to indicate minumum browser versions supported.

Comment (tjmw) [reply support-frontend/assets/pages/[countryGroupId]/events/events.tsx]: I seem to remember reading that the `lib` types are somehow derived from the `target` if not specified explicitly, which in our case is `esnext` (coming from `@guardian/tsconfig`). Esnext means the laces ecma script version supported by the version of TypeScript you're on, which for us is 5.1.6. I _think_ the latest target supported by that version is `ES2020` but I'm finding it hard to prove that.

PR #6718, opened 2025-01-22
Exclude subscribe pages from landing page abtests

paul-daniel-dempsey:

The `contributionsOnly` ab-test appeared to be  allocating  users into the test when they were on `\subscribe` pages as well as `landing/checkout/thankyou` routes.

It appears tests referencing `targetPage: ... allLandingPagesAndThankyouPages`  were at fault and this regex has been 
 updated to remove any `subscribe` (& `digitaledition`) pages.

The `abtestTests.tsx` have been widened to test for these scenario's now too.


andrewHEguardian (approved):
Good tests to give confidence here 👍

PR #6721, opened 2025-01-23
Change Knip `ignoreExportsUsedInFile` to false and fix

andrewHEguardian:
## What are you doing in this PR?

Changing the Knip config to not ignore unused exports if they're used within the same file. The main motivation for doing this is that there is a Knip [issue](https://github.com/webpro-nl/knip/issues/708) where combining `ignoreExportsUsedInFile: true` and `export { foo }` notation doesn't detect unused exports. This PR fixes the unnecessary `export` statements and removes the completely unused exports (and subsequent unused code)

PR #6723, opened 2025-01-23
Knip: remove exports used only in test files

andrewHEguardian:

Further to #6721 but running in `--production` mode. Because it now finds exports using the `export { foo }` style, the result is a lot noisier.

Production mode uses the entrypoints to find exports which are only ever used by tests.


GHaberis (approved):
Nice 🧽 ✨

paul-daniel-dempsey (approved):
LG 👍

PR #6725, opened 2025-01-24
Update browserslist packages

tjmw:

This updates the `@guardian/browserslist` package version and also the `caniuse-lite` db which contains browser stats. These are used in combination with `preset-env` to determine the code to be generated by babel. When I ran the upgrade I saw the following log line: `No target browser changes`. So I don't think this actually changes anything regarding the browsers we're targeting.

Update: I actually don't think we use caniuse-lite at all since `@guardian/browserslist-config` contains its own stats pulled from our GA data. We can see that this PR doesn't change any bundle sizes at all, so I think that confirms this upgrade is essentially a no-op.

## Why are you doing this?

We see the following warnings in local dev and CI:

```
Browserslist: caniuse-lite is outdated. Please run:
  npx update-browserslist-db@latest
  Why you should do it regularly: https://github.com/browserslist/update-db#readme
```

### DIFF
--- a/support-frontend/package.json
+++ b/support-frontend/package.json
@@ -120,7 +120,7 @@
 		"@babel/preset-react": "^7.23.3",
 		"@babel/preset-typescript": "^7.26.0",
 		"@emotion/eslint-plugin": "^11.11.0",
-		"@guardian/browserslist-config": "^6.1.1",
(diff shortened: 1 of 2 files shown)


andrewHEguardian (approved):
👍  Nice

PR #6726, opened 2025-01-24
Create knip production mode specific config

andrewHEguardian:

Creating a production mode config file for knip and a new script to run it in production mode with that config.

## Why are you doing this?

Production mode is a useful feature because it identifies code only used in tests and not by production code. However in #6721 it became a lot noisier because `ignoreExportsUsedInFile` was set to false. I've set this to true in the new config

## How to test

Run `yarn knip --production` and compare to `yarn knip:production`. 

There are much fewer errors in the latter, most of which are mocks and are easier to ignore

## Screenshots
 `yarn knip --production`
 
image

 `yarn knip:production`
 
 
image

### DIFF
--- a/support-frontend/knip.config.js
+++ b/support-frontend/knip.config.js
@@ -9,6 +9,7 @@ module.exports = {
 	entry: [...flattenedEntryPoints, 'scripts/build-ssr-content.tsx!'],
 	project: ['**/*.{js,jsx,ts,tsx,scss}!'],
 	ignoreExportsUsedInFile: false,
+	ignore: ['**/knip.*.js'],
 	ignoreDependencies: [
 		// used in package.json
 		'@guardian/browserslist-config',
--- a/support-frontend/knip.production.config.js
+++ b/support-frontend/knip.production.config.js
@@ -1,0 +1,5 @@
+const normalConfig = require('./knip.config');
+
+module.exports = Object.assign({}, normalConfig, {
+	ignoreExportsUsedInFile: true,
+});
--- a/support-frontend/package.json
+++ b/support-frontend/package.json
@@ -19,6 +19,7 @@
 		"test": "NODE_ENV=test echo 'Running TS tests' && jest",
 		"test-watch": "NODE_ENV=test echo 'Running TS tests in watch mode' && jest --watch",
 		"knip": "knip --include files,exports,nsExports,types,nsTypes,dependencies",
+		"knip:production": "yarn knip --production --config knip.production.config.js",
 		"jest-update-snapshot": "echo 'Updating TS tests' && jest -u",
 		"prepare": "cd .. && husky install ./support-frontend/.husky",
 		"prettier:check": "prettier --check ./assets",


GHaberis:
Just trying to understand in which context we'd run the new `knip:production` script vs  the original`knip` script?

From what I understand:

1. The `--production` flag means `knip` will flag code only used in tests. A benefit of using this vs not is that without it would not flag exports if they were used in unit tests, even if they weren't used elsewhere?

2. The [knip.production.config.js](https://github.com/guardian/support-frontend/pull/6726/files#diff-9bddf8629db0e31693955f9328be9e7dd629915b8a34be8da4c06c4ab8bda445) instructs knip to `ignoreExportsUsedInFile`, eg. Knip will not flag unused exports if they are used within the file? What is the benefit of keeping an export if the function exported is not used outside it's own module?  I'd assume we'd always want `ignoreExportsUsedInFile` to be `false`?

andrewHEguardian:
> Just trying to understand in which context we'd run the new `knip:production` script vs the original`knip` script?
> 
> From what I understand:
> 
> 1. The `--production` flag means `knip` will flag code only used in tests. A benefit of using this vs not is that without it would not flag exports if they were used in unit tests, even if they weren't used elsewhere?
> 2. The [knip.production.config.js](https://github.com/guardian/support-frontend/pull/6726/files#diff-9bddf8629db0e31693955f9328be9e7dd629915b8a34be8da4c06c4ab8bda445) instructs knip to `ignoreExportsUsedInFile`, eg. Knip will not flag unused exports if they are used within the file? What is the benefit of keeping an export if the function exported is not used outside it's own module?  I'd assume we'd always want `ignoreExportsUsedInFile` to be `false`?

1. 100% 👍 
2. An example of why we would use this setting is a component tested in storybook, we export the props type - but we only use the type in the file the component is defined in and the storybook test. Because the story is not production code, it gets flagged by knip.

GHaberis:
> > Just trying to understand in which context we'd run the new `knip:production` script vs the original`knip` script?
> > From what I understand:
> > 
> > 1. The `--production` flag means `knip` will flag code only used in tests. A benefit of using this vs not is that without it would not flag exports if they were used in unit tests, even if they weren't used elsewhere?
> > 2. The [knip.production.config.js](https://github.com/guardian/support-frontend/pull/6726/files#diff-9bddf8629db0e31693955f9328be9e7dd629915b8a34be8da4c06c4ab8bda445) instructs knip to `ignoreExportsUsedInFile`, eg. Knip will not flag unused exports if they are used within the file? What is the benefit of keeping an export if the function exported is not used outside it's own module?  I'd assume we'd always want `ignoreExportsUsedInFile` to be `false`?
> 
> 1. 100% 👍
> 2. An example of why we would use this setting is a component tested in storybook, we export the props type - but we only use the type in the file the component is defined in and the storybook test. Because the story is not production code, it gets flagged by knip.

Ok excellent thanks for explaining point 2, so we don't want `knip` flagging exports as unused if they're used in Storybook, which is a perfectly valid use case for. 

So would we want to update the knip action to use the `knip:production` script?

GHaberis (approved):

andrewHEguardian:
> So would we want to update the knip action to use the `knip:production` script?

We will use the default run in the GHA, to only flag completely unused code. This avoids having to maintain a list of code to ignore or code around knip

PR #6727, opened 2025-01-24
Make lazy loading default on checkouts

andrewHEguardian:

Make the lazy loading router the default router for our checkouts

[**Trello Card**](https://trello.com/c/FbcvCLda/1288-move-threetierlanding-route-out-of-supporterplusroutertsx-and-into-countrygroupid-routertsx)

## Why are you doing this?
Checklist item two of the above card. The goal is to move other routes into this router, without bundling more code.

## How to test

Playwright tests


GHaberis (approved):
Looks good to me 👍

PR #6728, opened 2025-01-24
Remove sentry logs from events page

GHaberis:

We didn't capture any issues with this logging, the problem we were troubleshooting was caused by something different


tjmw (approved):
👍

PR #6683, opened 2025-01-09
Contributions Only Landing Page

paul-daniel-dempsey:
## What are you doing in this PR?

No redux contributions only landing page

[**Trello Card**](https://trello.com/c/aQtJpr7y/1314-build-contributions-only-landing-page)

## How to test
https://support.thegulocal.com/int/contribute?country=lb {default=control}
https://support.thegulocal.com/int/contribute?country=lb#ab-contributionsOnly=control
https://support.thegulocal.com/int/contribute?country=lb#ab-contributionsOnly=variant

## Screenshots

|FROM default/control (existing)|TO contributionsOnly|
|-----|------|
|image|image|

PR #6730, opened 2025-01-24
Remove contributions only A/B test and page

andrewHEguardian:
Ends the A/B test of the new contributions only landing page (#6683) - it hasn't reached significance but the performance looks the same.

[**Trello Card**](https://trello.com/c/FbcvCLda/1288-move-threetierlanding-route-out-of-supporterplusroutertsx-and-into-countrygroupid-routertsx)

## Why are you doing this?
Item 3 in trello card. Part of removing the old redux-ful code for the landing page.

## How to test
Go to https://support.thegulocal.com/int/contribute?country=lb and see new contributions only page

## Before merging
This branch is branched off #6727, merge that and rebase before this PR


support-frontend/assets/helpers/storage/abandonedBasketCookies.ts
```diff
@@ -59,36 +39,6 @@ function parseAmount(amount: number): number | 'other' {
 	return isNaN(amount) ? 'other' : amount;
 }
 
-export function updateAbandonedBasketCookie(amount: string) {
```
andrewHEguardian, line 62: This is now unused, should we be setting and updating the value in the one time checkout? If so it was probably missed in the migration to the new one time checkout. @tomrf1 maybe you have some thoughts

support-frontend/assets/helpers/tracking/trackingOphan.ts
```diff
@@ -136,40 +127,9 @@ const setReferrerDataInLocalStorage = (
 
 const getPageViewId = (): string => ophan.viewId;
 
-/**
- * To be used when navigating with React Router
- * in order to generate new Page View event in Ophan
- */
-const navigateWithPageView = (
```
andrewHEguardian, line 143: Would we need this if we go back to navigating with React Router? @GHaberis

GHaberis:
👀 📉  interesting that removing the 2-step checkout brings the `[countryGroupId]/router.js` bundle size down by 63%! This router wasn't even rendering the 2-step checkout, but it's bundle must've somehow included some weighty code related to it that's been deleted here!

Screenshot 2025-01-24 at 17 53 29

support-frontend/assets/helpers/tracking/trackingOphan.ts
GHaberis: Hypothetically it might be useful in the future, but unless we're using this I don't think we need to keep it tbh.

support-frontend/assets/helpers/storage/abandonedBasketCookies.ts
tomrf1: Yes the abandoned basket cookie is still in use! Does it still work for other checkouts?

andrewHEguardian: Yes, it's used in the generic checkout (so RC, S+, Tier 3)

paul-daniel-dempsey (approved):
Looks good, smoke/cron e2e tests pass too.

PR #6736, opened 2025-01-27
Remove Digital Subscription gifting

rupertbates:
We no longer sell gift versions of the digital subscription so this PR removes all the code related to it.

Digital Subscription gift subs were a two stage process, the gifter would purchase them and a unique gift code would be sent to the giftee. The giftee would then need to activate their subscription by redeeming this gift code.

This PR removes both parts of this.

[**Trello Card**](https://trello.com/c/mg6AipPP/1381-remove-digi-pack-gift-code-server-side)

Tested by running e2e tests locally with this branch deployed to CODE support-workers


support-frontend/app/controllers/CreateSubscriptionController.scala
```diff
@@ -103,7 +103,7 @@ class CreateSubscriptionController(
         // For Stripe payments, recaptcha validation is done earlier in the flow
         val recaptchaToken =
           request.body.paymentFields match {
-            case Left(dd: DirectDebitPaymentFields) => Some(dd.recaptchaToken)
+            case dd: DirectDebitPaymentFields => Some(dd.recaptchaToken)
```
rupertbates, line 106: Previously the `paymentFields` value was an `Either` which could be payment field data or redemption data. Now it will always be of type `PaymentFields` which simplifies the code significantly.

support-frontend/app/services/stepfunctions/SupportWorkersClient.scala
```diff
@@ -117,24 +115,13 @@ class SupportWorkersClient(
     product match {
       case _: GuardianWeekly =>
         Right(
-          GiftRecipient.WeeklyGiftRecipient(
```
rupertbates, line 120: There used to be two types of gift recipient - one for Digital Subscription giftees and one for Guardian Weekly giftees. Now Guardian Weekly is the only product which still supports gifting, so we only need one case class

support-frontend/app/views/subscriptionRedemptionForm.scala.html
rupertbates, line 1: This was the Twirl template for the redemption form

support-models/src/main/scala/com/gu/support/catalog/Product.scala
```diff
@@ -135,23 +135,18 @@ case object DigitalPack extends Product {
       id: String,
       billingPeriod: BillingPeriod,
       description: String,
-      readerType: ReaderType = Direct,
   ) =
-    ProductRatePlan(id, billingPeriod, NoFulfilmentOptions, NoProductOptions, description, readerType = readerType)
```
rupertbates, line 140: We used to distinguish gifts from regular subs using the readerType field

support-models/src/main/scala/com/gu/support/redemptions/RedemptionCode.scala
rupertbates, line 1: Code related to the gift codes that we would send out to users

support-models/src/main/scala/com/gu/support/workers/PaymentProvider.scala
```diff
@@ -43,17 +43,16 @@ object PaymentProvider {
 
   implicit val encoder: Encoder[PaymentProvider] = Encoder.encodeString.contramap[PaymentProvider](_.toString)
 
-  def fromPaymentFields(paymentFields: Option[PaymentFields]): PaymentProvider = paymentFields match {
```
rupertbates, line 46: In the past this could have been None for gift redemptions, now it will always be present

support-models/src/main/scala/com/gu/support/workers/states/CreateZuoraSubscriptionState.scala
```diff
@@ -66,29 +64,14 @@ object CreateZuoraSubscriptionProductState {
       salesForceContact: SalesforceContactRecord,
   ) extends CreateZuoraSubscriptionProductState
 
-  case class DigitalSubscriptionDirectPurchaseState(
```
rupertbates, line 69: This unwieldy name was to distinguish the state needed for a regular DS purchase from a gift purchase or a gift redemption. Now we only have regular purchases, I have simplified it.

support-models/src/main/scala/com/gu/support/zuora/api/SubscriptionData.scala
```diff
@@ -214,7 +213,6 @@ case class Subscription(
     termType: String = "TERMED",
     readerType: ReaderType = ReaderType.Direct,
     promoCode: Option[PromoCode] = None,
-    redemptionCode: Option[RawRedemptionCode] = None,
```
rupertbates, line 217: The gift code used by the giftee to redeem their sub

support-frontend/app/services/stepfunctions/SupportWorkersClient.scala
```diff
@@ -173,11 +160,11 @@ class SupportWorkersClient(
         ipAddress =
           request.headers.get("X-Forwarded-For").flatMap(_.split(',').headOption).getOrElse(request.remoteAddress),
       )
-      isExistingAccount = createPaymentMethodState.paymentFields.left.exists(_.isInstanceOf[ExistingPaymentFields])
+      isExistingAccount = createPaymentMethodState.paymentFields.isInstanceOf[ExistingPaymentFields]
```
johnduffell, line 163: this code probaly predates any of this, but annoyingly the simplest way - `isInstanceOf` - is not type safe as it only gives a warning if the types are not compatible.  You would have to either use a pattern match or a partial function to get the type safety.
https://scastie.scala-lang.org/xG1JBBG0QcW9ehKDG05jMA

Partial function is probably less verbose than pattern match
```scala
PartialFunction.cond(createPaymentMethodState.paymentFields) { case _: ExistingPaymentFields => true }
```

support-models/src/main/scala/com/gu/support/workers/GiftRecipient.scala
```diff
@@ -3,55 +3,18 @@ package com.gu.support.workers
...
-      firstName: String,
-      lastName: String,
-      email: String,
-      message: Option[String],
-      deliveryDate: LocalDate,
-  ) extends GiftRecipient
-
-  val discriminatedType = new DiscriminatedType[GiftRecipient]("giftRecipientType")
-  implicit val weeklyCodec: discriminatedType.VariantCodec[WeeklyGiftRecipient] =
-    discriminatedType.variant[WeeklyGiftRecipient]("Weekly")
```
johnduffell, line 37: just thinking it through
if we deploy this while a GW gift execution is processing, then the newly deployed lambdas would not check the disciminator any more so it would deserialise successfully.
And noone could be checking out a DS gift, so that wouldn't be an issue.

support-models/src/main/scala/com/gu/support/zuora/api/SubscribeRequest.scala
```diff
@@ -29,33 +29,3 @@ object SubscribeRequest {
...
-
-case class UpdateRedemptionDataRequest(
-    requestId: String,
-    gifteeIdentityId: String,
-    giftRedemptionDate: LocalDate,
-    currentTerm: Int,
-    currentTermPeriodType: PeriodType,
-)
-
-object DistributeRevenueRequest {
```
johnduffell, line 51: I think we also have a lambda somewhere that distributes revenue, because we found that if we tried to do it immediately, the invoice + revenue schedules had not been created by zuora, causing a support workers failure.

support-models/src/main/scala/com/gu/support/workers/states/ExecutionTypeDiscriminators.scala
```diff
@@ -7,9 +7,7 @@ object ExecutionTypeDiscriminators {
   val contribution = "Contribution"
   val supporterPlus = "SupporterPlus"
   val tierThree = "TierThree"
-  val digitalSubscriptionDirectPurchase = "DigitalSubscriptionDirectPurchase"
-  val digitalSubscriptionGiftPurchase = "DigitalSubscriptionGiftPurchase"
-  val digitalSubscriptionGiftRedemption = "DigitalSubscriptionGiftRedemption"
+  val digitalSubscription = "DigitalSubscription"
```
johnduffell, line 10: if there is a `DigitalSubscriptionDirectPurchase` in progress when it's deployed, it would fail to deserialise if the discriminator is changed to `DigitalSubscription`
Not a showstopper, but worth avoiding merging it if we're pushing people to digi subs for any reason.

support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
```diff
@@ -144,19 +144,7 @@ class DigitalPackEmailFields(
   )
 
   def build(digi: SendThankYouEmailDigitalSubscriptionState)(implicit ec: ExecutionContext): Future[List[EmailFields]] =
```
johnduffell, line 146: don't think this needs to be a list any more, at least at the time, this was the only product that sent multiple emails 
```suggestion
  def build(digi: SendThankYouEmailDigitalSubscriptionState)(implicit ec: ExecutionContext): Future[EmailFields] =
```

support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
```diff
@@ -173,93 +161,8 @@ class DigitalPackEmailFields(
     EmailFields(attributePairs, user, dataExtensionName, userAttributes)
```
johnduffell, line 161: I don't think `userAttributes` is needed any more
I think we also did work in membership-workflow to push unmanaged attributes to braze
we could take a view on whether it's worth keeping these things?  I remember this route to getting things into emails was frowned upon heavily at the time though.

johnduffell (approved):
great work to get all that complexity out, I've added a few comments but the main thing is that everything still works 👏

support-workers/src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
rupertbates: I'm not going to change that in this PR because I don't really understand what it's doing. It looks like all products do pass something into the userAttributes, mostly just a string of the product name

support-models/src/main/scala/com/gu/support/zuora/api/SubscribeRequest.scala
rupertbates: Well remembered, I'll do a PR in support-service-lambdas for that.

rupertbates: https://github.com/guardian/support-service-lambdas/pull/2654

PR #6733, opened 2025-01-27
Move landing page to lazy loading router

andrewHEguardian:
## What are you doing in this PR?

Moving the three tier landing (and contributions only landing page) to the lazy loading router used to serve the generic checkout

[**Trello Card**](https://trello.com/c/FbcvCLda/1288-move-threetierlanding-route-out-of-supporterplusroutertsx-and-into-countrygroupid-routertsx)

## Why are you doing this?
Part of the migration away from redux

## How to test
Go to `/uk/contribute` and see the three tier landing. 
Go to `/int/contribute?country=lb` and see the contributions only landing page.

PR #6737, opened 2025-01-27
Lighthouse - remove third party resource count

andrewHEguardian:

Removing the assertion of third party resource count from our lighthouse CI action, after refactoring the router in #6733 it's been bumped up to 33. But this doesn't seem like a metric worth tracking

## Screenshots
Example of what these are from the generated report
image


GHaberis (approved):
Makes sense 👍

PR #6745, opened 2025-01-29
Don't alarm on "Email in use" response from Identity `/guest` endpoint

tjmw:

Don't alarm on `Email in use` response from Identity `/guest` endpoint.

## Why are you doing this?

Sometimes `POST /guest` returns `Email in use` even though we already checked whether the user existed using the `GET /user` endpoint. According to identity occasionally users get into a weird state and this is a result of that. I think this should still result in a 500 (it's not related to something the user has done) but we don't want to alarm on it.

## How to test

I've tested locally by forcing the error, in the logs I see:

```
not pushing alarm metric for 500 Some(email_address_already_taken) as it is in our ignore list
```

### DIFF
--- a/support-frontend/app/controllers/CreateSubscriptionController.scala
+++ b/support-frontend/app/controllers/CreateSubscriptionController.scala
@@ -233,10 +233,14 @@ class CreateSubscriptionController(
     )
   }
 
-  private def mapIdentityErrorToCreateSubscriptionError(identityError: IdentityError) =
+  private def mapIdentityErrorToCreateSubscriptionError(identityError: IdentityError): CreateSubscriptionError =
     identityError match {
       case EmailProviderRejected(_) => RequestValidationError(emailProviderRejectedCode)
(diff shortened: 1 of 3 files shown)


support-frontend/app/controllers/CreateSubscriptionController.scala
```diff
@@ -345,6 +349,14 @@ class CreateSubscriptionController(
                 ),
                 body = writeable.toEntity(err.message),
               )
+            case ServerError(code) if code == emailAddressAlreadyTakenCode =>
+              Result(
+                header = new ResponseHeader(
+                  status = INTERNAL_SERVER_ERROR,
+                  reasonPhrase = Some(emailAddressAlreadyTakenCode),
+                ),
+                body = writeable.toEntity(emailAddressAlreadyTakenCode),
```
tjmw, line 358: This would make it very easy for us to show a more relevant error message to the user on the client if we wanted.

rupertbates (approved):
Looks good, thanks for doing this.

PR #6742, opened 2025-01-28
Add A/B test for confirm email form field

andrewHEguardian:
## What are you doing in this PR?

Adding a form field to the generic checkout for a user to confirm their email address. 

[**Trello Card**](https://trello.com/c/mCYE3GTi/1373-add-confirm-email-field-across-products-generic-checkout-a-b-test)

## Why are you doing this?
This feature is present on the old subscription checkout, and should stop cases where a user might have made a mistake and tied their subscription to the wrong email address. We want to test if the additional friction of having to confirm the email will harm the CVR.

### Signed out

Go to `https://support.thegulocal.com/uk/checkout?product=Contribution&ratePlan=Monthly&contribution=3#ab-confirmEmail=variant`. See the confirm email field. Enter an email address in the email field. Then enter a different email in the confirm email field. See error message and cannot submit the form. Make the two fields match. Can submit the form.

### Signed in
Go to `https://support.thegulocal.com/uk/checkout?product=Contribution&ratePlan=Monthly&contribution=3#ab-confirmEmail=variant`. Do not see the confirm email field.

### E2E

We need an exception to stop this affecting our playwright e2e tests.

### Signed out
image

### Signed in
image

PR #6751, opened 2025-01-31
Confirm email - use pattern validation

andrewHEguardian:

Implement the confirm email field but using HTML form validation - a pattern regex field that updates based on whatever is in the email field. 

This PR also enables it as an A/B test

[**Trello Card**](https://trello.com/c/mCYE3GTi/1373-add-confirm-email-field-across-products-generic-checkout-a-b-test)

## Why are you doing this?

This is a branch off of #6742 but using a different mechanism.

Comparison of this approach (HTML validation) vs the other approach (validate in JS)

Pros: 

- Fits with our other validation of the form
- Doesn't require a check based on the sign-in status in the form's submit function

Cons:

- Adds an additional dependency (we wanted to use [RegExp.escape](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/escape) but it has limited availability)
- ~Slightly clunky useRef~

## How to test

See #6742

### DIFF
--- a/support-frontend/assets/pages/[countryGroupId]/checkout/components/PersonalDetailsFields.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/checkout/components/PersonalDetailsFields.tsx
@@ -14,6 +14,9 @@ type PersonalDetailsFieldsProps = {
 	email: string;
 	setEmail: (value: string) => void;
 	isEmailAddressReadOnly: boolean;
+	requireConfirmedEmail: boolean;
+	confirmedEmail: string;
+	setConfirmedEmail: (value: string) => void;
 };
 
 export function PersonalDetailsFields({
@@ -25,10 +28,19 @@ export function PersonalDetailsFields({
 	email,
 	setEmail,
 	isEmailAddressReadOnly,
+	requireConfirmedEmail,
+	confirmedEmail,
+	setConfirmedEmail,
 }: PersonalDetailsFieldsProps) {
 	const [firstNameError, setFirstNameError] = useState<string>();
 	const [lastNameError, setLastNameError] = useState<string>();
 	const [emailError, setEmailError] = useState<string>();
+
+	const emailAddressDoesNotMatch =
+		confirmedEmail.length && email !== confirmedEmail;
+	const confirmedEmailError = emailAddressDoesNotMatch
+		? 'The email addresses do not match.'
+		: undefined;
 
 	return (
 		<>
@@ -66,6 +78,25 @@ export function PersonalDetailsFields({
 					}}
 				/>
 			</div>
+			{requireConfirmedEmail && !isEmailAddressReadOnly && (
+				<div>
+					<TextInput
+						id="confirm-email"
+						data-qm-masking="blocklist"
+						label="Confirm email address"
+						value={confirmedEmail}
+						type="email"
+						autoComplete="email"
+						onChange={(event) => {
+							setConfirmedEmail(event.currentTarget.value);
+						}}
+						name="confirm-email"
+						required
+						maxLength={80}
+						error={confirmedEmailError}
+					/>
+				</div>
+			)}
 			{children}
 			<div>
 				<TextInput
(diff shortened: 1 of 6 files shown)


support-frontend/assets/pages/[countryGroupId]/checkout/components/PersonalDetailsFields.tsx
```diff
@@ -66,6 +77,42 @@ export function PersonalDetailsFields({
 					}}
 				/>
 			</div>
+			{requireConfirmedEmail && !isEmailAddressReadOnly && (
+				<div ref={confirmEmailRef}>
```
Comment (andrewHEguardian) [inline support-frontend/assets/pages/[countryGroupId]/checkout/components/PersonalDetailsFields.tsx:81]: We can't put the ref on the source component see this [discussion](https://github.com/guardian/source/pull/1106#issuecomment-951761234)

support-frontend/assets/pages/[countryGroupId]/checkout/components/PersonalDetailsFields.tsx
```diff
@@ -45,6 +55,7 @@ export function PersonalDetailsFields({
 					}}
 					onBlur={(event) => {
 						event.target.checkValidity();
+						confirmEmailRef.current?.querySelector('input')?.checkValidity();
```
Comment (GHaberis) [inline support-frontend/assets/pages/[countryGroupId]/checkout/components/PersonalDetailsFields.tsx:58]: I assume the UX would be that as soon as a user tabs out of the email field we'll invalidate the confirm email field assuming the values don't match at this point, which is most likely as they complete email first before confirm email? Would it be better UX to only check the validity of the input if it has a `.value.length > 0` ?

support-frontend/assets/pages/[countryGroupId]/checkout/components/PersonalDetailsFields.tsx
```diff
@@ -45,6 +55,9 @@ export function PersonalDetailsFields({
 					}}
 					onBlur={(event) => {
 						event.target.checkValidity();
+						if (email.length > 0) {
```
Comment (GHaberis) [inline support-frontend/assets/pages/[countryGroupId]/checkout/components/PersonalDetailsFields.tsx:58]: Sorry my comment here might not have been clear https://github.com/guardian/support-frontend/pull/6751#discussion_r1936994386.

I was suggesting we check `confirmEmailRef.current?.querySelector('input')` has a value before checking it's validity, as I suspected we could have this scenario:

- Non-signed in user enters their email
- User tabs out of the email input and into the confirm email input
- The email onBlur immediately triggers a validation error on the empty confirm email input 

Does this happen?

Comment (andrewHEguardian) [reply support-frontend/assets/pages/[countryGroupId]/checkout/components/PersonalDetailsFields.tsx:58]: yes, sorry you're right that should be `confirmEmail.length`. Although that might be slightly confusing to mix together the react state variable and the event target

support-frontend/assets/pages/[countryGroupId]/checkout/components/PersonalDetailsFields.tsx
```diff
@@ -45,6 +55,9 @@ export function PersonalDetailsFields({
 					}}
 					onBlur={(event) => {
 						event.target.checkValidity();
+						if (confirmedEmail.length > 0) {
+							confirmEmailRef.current?.querySelector('input')?.checkValidity();
```
Comment (GHaberis) [inline support-frontend/assets/pages/[countryGroupId]/checkout/components/PersonalDetailsFields.tsx:59]: Just a question with this logic, we added it to guard against a user dong the following:

- Enter a value email `george@theguardian.com`
- Enter a value in confirm email `george@theguardian.com` 
- Change value in to email `andrew@theguardian.com`

With this is ref the validation on `confirmEmailRef` will alert there's a mismatch immediately after the user enters `andrew@theguardian.com`. 

If we hadn't implemented this `ref` logic would this mismatch have been picked up on form submit and prevented submission?

Comment (andrewHEguardian) [reply support-frontend/assets/pages/[countryGroupId]/checkout/components/PersonalDetailsFields.tsx:59]: Just tested - yes that's correct. This is just doing the validation sooner on the email fields onBlur

Comment (andrewHEguardian) [reply support-frontend/assets/pages/[countryGroupId]/checkout/components/PersonalDetailsFields.tsx:59]: yeah it's a fairly edgecase situation, that simplification would be better 👍

PR #6764, opened 2025-02-07
Guardian Ad-Lite : Checkout Ts&Cs Update 

paul-daniel-dempsey:

1. TsCs: Footer should read `By proceeding, you are agreeing to the Guardian Ad-Lite Terms` and link to new Ts&Cs for Guardain-AdLite only.
2. TsCs: All Card payments links need capitalisation.

## Screenshots

|Item|From|To|
|-----|------|-----|
|1/2|image|image|

### DIFF
--- a/support-frontend/assets/pages/supporter-plus-landing/components/paymentTsAndCs.tsx
+++ b/support-frontend/assets/pages/supporter-plus-landing/components/paymentTsAndCs.tsx
@@ -127,12 +127,11 @@ export function TsAndCsFooterLinks({
 		: inAdLite
 		? termsGuardianAdLite('Terms')
 		: termsContributions;
-	const productNameSummary = inAdLite ? ' the Guardian Ad-Lite ' : ' our ';
+	const productNameSummary = inAdLite ? 'the Guardian Ad-Lite' : 'our';
 
 	return (
 		<div css={marginTop}>
-			By proceeding, you are agreeing to{productNameSummary}
-			{terms}.{' '}
+			By proceeding, you are agreeing to {productNameSummary} {terms}.{' '}
 			<p css={marginTop}>
 				To find out what personal data we collect and how we use it, please
 				visit our {privacy}.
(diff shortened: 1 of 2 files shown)


support-frontend/assets/pages/supporter-plus-landing/components/paymentTsAndCs.tsx
```diff
@@ -121,11 +124,15 @@ export function TsAndCsFooterLinks({
 
 	const terms = amountIsAboveThreshold
 		? termsSupporterPlus('Terms and Conditions')
+		: inAdLite
+		? termsGuardianAdLite('Terms')
 		: termsContributions;
```
GHaberis, line 129: Interesting, I've never used [conditional chains](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_operator#conditional_chains) in a ternary before!

support-frontend/assets/pages/supporter-plus-landing/components/paymentTsAndCs.tsx
```diff
@@ -121,11 +124,15 @@ export function TsAndCsFooterLinks({
...
+	const productNameSummary = inAdLite ? ' the Guardian Ad-Lite ' : ' our ';
 
 	return (
 		<div css={marginTop}>
-			By proceeding, you are agreeing to our {terms}.{' '}
+			By proceeding, you are agreeing to{productNameSummary}
```
GHaberis, line 134: why not put the leading space in here rather than in the `productNameSummary` string?

tjmw (approved):

support-frontend/assets/pages/supporter-plus-landing/components/paymentTsAndCs.tsx
paul-daniel-dempsey: Is that ok to do if its simple or better to breakout into multiple line?

tjmw: I _thought_ about suggesting an alternative here but actually I think it's quite elegant! I think the alternatives (`if/else if` or a `switch`) would require defining a `let` variable and then setting. It's quite nice here just assigning the result straight to a `const`.

support-frontend/assets/pages/supporter-plus-landing/components/paymentTsAndCs.tsx
paul-daniel-dempsey: Yes, will avoid spaces in string consts going forwards. Updated.

PR #6766, opened 2025-02-10
Force the recaptcha to re-initialise after a server side failure

tjmw:

Reinitialise the recaptcha each time server side validation fails. This causes a new Stripe payment intent to be created.

[**Trello Card**](https://trello.com/c/YDTXTfHT/1420-fix-issue-with-server-side-validation-and-stripe-state)

## Why are you doing this?

Before this change, if the reader was paying with Stripe, and server side validation rejected the call to `/subscribe/create`, then even if the user then fixed the issue which caused the validation failure, future form submissions fail with the following error from Stripe: `You cannot confirm this SetupIntent because it has already succeeded.` ([setup_intent_unexpected_state]).

[setup_intent_unexpected_state]:
https://stripe.com/docs/error-codes/setup-intent-unexpected-state

## How to test

On the generic checkout, submit the payment form with data which will cause server side validation to fail. Fix the issue and re-submit, the payment should now go through.

I've run the smoke test against my branch and all tests passed.

### DIFF
--- a/support-frontend/assets/components/recaptcha/recaptcha.tsx
+++ b/support-frontend/assets/components/recaptcha/recaptcha.tsx
@@ -31,14 +31,14 @@ export type RecaptchaProps = {
 	id?: string;
 	onRecaptchaCompleted: (token: string) => void;
 	onRecaptchaExpired?: () => void;
-	attemptNumber: number;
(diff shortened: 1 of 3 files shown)


tjmw:
Going to use a different approach for this.

PR #6759, opened 2025-02-05
Add server side validation for billing address postcode

tjmw:
## What are you doing in this PR?

Add server side validation for billing address postcode.

[**Trello Card**](https://trello.com/c/h2sFEVzL/1399-paypal-server-side-validation-placeholder)

## Why are you doing this?

We've had an ongoing issue where requests where the payment method is PayPal make it to support workers with an email address in the billing postcode field. We're not sure why this happens but are hoping some server side validation will prevent it and alert users that they need to fix something. It primarily seems to affect USD Supporter Plus and recurring contributions (we collect the postcode for these) so I've focused on fixing those cases, in the PaidProductValidation object used by both products.

I've introduced a concept of a preservable error message, i.e. something we can map to copy to show the user. It'd be nice to extend this but for now it's a one off for this field. As it stands, it's not perfect - if there are other validation errors we'll lose the message. I'd like to fix that in a future PR.

~~**Note: The error message copy needs updating.**~~ Copy has been updated.

## How to test

Entering something which looks like an email address (contains an `@` character) triggers the validation:

Screenshot 2025-02-06 at 09 32 31

## Accessibility test checklist

-  [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-  [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-  [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-  [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

PR #6768, opened 2025-02-10
Create Stripe setup intent on form submit, not on recaptcha check

tjmw:

Before this change, if the reader was paying with Stripe, and server side validation rejected the call to `/subscribe/create`, then even if the user fixed the issue which caused the validation failure, future form submissions (until the recaptcha expires) fail with the following error from Stripe: `You cannot confirm this SetupIntent because it has already succeeded` (`setup_intent_unexpected_state`).

This was happening because we were previously taking the recaptcha token, verifying it and retrieving a Stripe client secret all at the point where the recaptcha box was checked. This only happened once in the above scenario, and calling `stripe.confirmCardSetup` twice with the same client secret results in the above error.

Now, instead of getting the client secret (using the recaptcha token) when the recaptcha is checked, we use the recaptcha token from state to get the client secret at the point the form is submitted, so we get a new client secret each time to pass to `stripe.confirmCardSetup`.

Supersedes #6766.

[**Trello Card**](https://trello.com/c/YDTXTfHT/1420-fix-issue-with-server-side-validation-and-stripe-state)

## Why are you doing this?

I noticed this issue while testing #6759, but I think it affects all Stripe payments on the generic checkout where server side validation prevents the purchase from going through.

## How to test

On the generic checkout, submit the payment form with data which will cause server side validation to fail. Fix the issue and re-submit, the payment should now go through.

I've run the smoke tests against my branch and they all passed.

### DIFF
--- a/support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx
@@ -1230,8 +1230,6 @@ export function CheckoutComponent({
 														errors={{}}
 														recaptcha={
 															<Recaptcha
-																// We could change the parents type to Promise and use await here, but that has
-																// a lot of refactoring with not too much gain
 																onRecaptchaCompleted={(token) => {
 																	setRecaptchaToken(token);
 																}}
--- a/support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx
@@ -332,7 +332,6 @@ export function CheckoutComponent({
 	const stripe = useStripe();
 	const elements = useElements();
 	const cardElement = elements?.getElement(CardNumberElement);
-	const [stripeClientSecret, setStripeClientSecret] = useState<string>();
 	const [
 		stripeExpressCheckoutPaymentType,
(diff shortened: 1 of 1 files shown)


support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx
```diff
@@ -347,13 +346,6 @@ export function CheckoutComponent({
 		}
 	}, [stripeExpressCheckoutSuccessful]);
 
-	/**
-	 * flag that disables the submission of the form until the stripeClientSecret is
-	 * fetched from the Stripe API with using the reCaptcha token
-	 */
-	const [stripeClientSecretInProgress, setStripeClientSecretInProgress] =
```
Comment (tjmw) [inline support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx:354]: No longer needed as this now happens when the form is submitted, not when the recaptcha is checked.

support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx
```diff
@@ -1240,16 +1233,7 @@ export function CheckoutComponent({
 																// We could change the parents type to Promise and use await here, but that has
```
Comment (rupertbates) [inline support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx:1233]: Maybe remove this comment?

rupertbates (approved):
With one suggestion

PR #6770, opened 2025-02-11
Adds back the abandon basket cookie to one time checkout

charleycampbell:
Adds the abandon basket cookie back into the one time checkout

## Why are you doing this?
The abandon basket cookie was removed from the one time checkout during a refactor and this PR adds it back in. 

## How to test
One the time checkout, open the dev tools -> Application tab -> Cookies -> filter by GU_ and you should see....

image

when you change the amount on the choice card the cookie should update to reflect the selected amount, unless you select Other which just displays 'other'

### DIFF
--- a/support-frontend/assets/helpers/storage/abandonedBasketCookies.ts
+++ b/support-frontend/assets/helpers/storage/abandonedBasketCookies.ts
@@ -1,13 +1,13 @@
 import { useEffect } from 'react';
 import type { InferInput } from 'valibot';
 import {
-  flatten,
-  literal,
-  number,
-  object,
-  safeParse,
-  string,
-  union,
+	flatten,
+	literal,
+	number,
+	object,
+	safeParse,
+	string,
+	union,
 } from 'valibot';
 import type { ActiveProductKey } from 'helpers/productCatalog';
 import * as cookie from 'helpers/storage/cookie';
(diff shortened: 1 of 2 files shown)


support-frontend/assets/helpers/storage/abandonedBasketCookies.ts
```diff
@@ -19,13 +39,8 @@ export function useAbandonedBasketCookie(
 		region,
 	};
 
-	// support-dotcom-components can only return a user to the checkout for these products
-	// don't drop the cookie if they came from a different checkout
-	const isSupportedProduct =
-		product === 'Contribution' || product === 'SupporterPlus';
-
```
charleycampbell, line 26: I've removed this function but wondering whether are any reasons not too... @tomrf1? we could just add in an or statement to include one time contribution as well

tomrf1: So `'Contribution'` is for recurring contributions?
It depends if `useAbandonedBasketCookie` is ever called from a checkout that isn't supported, I don't know the answer to this!

tomrf1 (approved):
nice

support-frontend/assets/helpers/storage/abandonedBasketCookies.ts
charleycampbell: Yes `Contribution` is for recurring contributions.

I've re-added this with OneTimeContribution instead as it's such a small addition!

PR #6772, opened 2025-02-12
Fetch landing page config from dynamodb

tomrf1:
Previously we [hardcoded](https://github.com/guardian/support-frontend/pull/6655) the landing page tests config into the backend.
Now that we have a dynamodb table with this config, we can remove this.

This PR introduces the `LandingPageTestService` for polling and caching the dynamodb data.
It also adds a new alarm (based on a cloudwatch metric) to tell us when the server fails to fetch or parse a landing page test from Dynamodb. Note that even if only one test config is invalid, and the rest are successfully returned to the client, the alarm will still trigger.


support-frontend/app/services/LandingPageTestService.scala
```diff
@@ -0,0 +1,140 @@
...
+import scala.jdk.FutureConverters._
+import scala.util.control.NonFatal
+import scala.concurrent.duration._
+
+trait LandingPageTestService {
+  def getTests(): List[LandingPageTest]
+}
+
+object LandingPageTestService {
+  // Converts Dynamodb Attributes to Circe Json
```
tomrf1, line 34: based on https://github.com/guardian/support-admin-console/blob/main/app/utils/Circe.scala#L34

support-frontend/app/admin/settings/LandingPageTest.scala
```diff
@@ -53,43 +53,3 @@ object LandingPageTest {
   implicit val encoder: Encoder[LandingPageTest] = deriveEncoder
   implicit val decoder: Decoder[LandingPageTest] = deriveDecoder
 }
-
-// TODO - fetch config from dynamodb instead of hardcoding here
-object LandingPageTestsProvider extends SettingsProvider[List[LandingPageTest]] {
```
tomrf1, line 58: there's still a hardcoded fallback in the client, in case something goes wrong

LAKSHMIRPILLAI (approved):
LGTM 👍🏼

PR #6774, opened 2025-02-12
More cookie changes

rupertbates:
Some additional work on the cookies which support-frontend sets:
- Renames the gu_user_features_expiry to gu_user_benefits_expiry to match the same rename in DCR made [here](https://github.com/guardian/dotcom-rendering/pull/13277)
- Sets the gu_hide_support_messaging cookie on the succsessful completion of a one time contribution rather than the gu.contributions.contrib-timestamp cookie. This is to work with the new benefits mechanism introduced in the PR above
- Updates the DiagnosticController class which serves the /cookies page to only show currently relevant cookies

### DIFF
--- a/support-frontend/app/controllers/DiagnosticsController.scala
+++ b/support-frontend/app/controllers/DiagnosticsController.scala
@@ -2,28 +2,23 @@ package controllers
 
 import actions.CustomActionBuilders
 import play.api.mvc._
-import Results.Ok
+import play.api.mvc.Results.Ok
 
 class DiagnosticsController(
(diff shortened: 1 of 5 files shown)


tjmw (approved):
Looks good. Do we need to be careful about synchronising the release of this and the changes in DCR?

rupertbates:
> Looks good. Do we need to be careful about synchronising the release of this and the changes in DCR?

I think the only concern is that we need to keep respecting the one time contribution cookie on DCR for 90 days so that everyone that made a contribution before this PR continues to receive their hide support messaging benefit.

PR #6775, opened 2025-02-12
Digital Editions: Subscribe Landing Page ab-test simplification

paul-daniel-dempsey:

ABTest: digitalEditionCheckout ->
- Remove non-required use of `canRun` 
- Replace with `persistsPage` instead.

### DIFF
--- a/support-frontend/assets/helpers/abTests/abtestDefinitions.ts
+++ b/support-frontend/assets/helpers/abTests/abtestDefinitions.ts
@@ -131,13 +131,10 @@ export const tests: Tests = {
 		isActive: false,
 		referrerControlled: false, // ab-test name not needed to be in paramURL
 		seed: 7,
-		targetPage:
-			'(/uk/)(subscribe$|subscribe/digitaledition$|subscribe/digitaledition/thankyou$|checkout|thank-you)', // one-off test using canRun to exclude all products except DigitalSubscription
+		persistPage:
+			// uk will ensure we match the generic checkout
+			'(/uk/)(subscribe/digitaledition$|subscribe/digitaledition/thankyou$|checkout|thank-you)',
+		targetPage: '/subscribe$',
 		excludeContributionsOnlyCountries: true,
-		canRun: () => {
-			const urlSearchParams = new URLSearchParams(window.location.search);
-			const productParam = urlSearchParams.get('product');
-			return !productParam || productParam === 'DigitalSubscription';
-		},
 	},
 };
--- a/support-frontend/assets/helpers/abTests/abtestDefinitions.ts
+++ b/support-frontend/assets/helpers/abTests/abtestDefinitions.ts
@@ -133,7 +133,7 @@ export const tests: Tests = {
 		seed: 7,
 		persistPage:
 			// uk will ensure we match the generic checkout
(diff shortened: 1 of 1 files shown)


support-frontend/assets/helpers/abTests/abtestDefinitions.ts
```diff
@@ -131,13 +131,10 @@ export const tests: Tests = {
 		isActive: false,
 		referrerControlled: false, // ab-test name not needed to be in paramURL
 		seed: 7,
-		targetPage:
-			'(/uk/)(subscribe$|subscribe/digitaledition$|subscribe/digitaledition/thankyou$|checkout|thank-you)', // one-off test using canRun to exclude all products except DigitalSubscription
+		persistPage:
+			// uk will ensure we match the generic checkout
+			'(/uk/)(subscribe/digitaledition$|subscribe/digitaledition/thankyou$|checkout|thank-you)',
```
tjmw, line 136: This looks good. I don't think we need the brackets around `/uk/` here, so they could be removed for a slight simplification but definitely doesn't do any harm keeping them.

Similarly, is it weird that we're anchoring some paths to the end of the string but not others? I think we could also use `$` after `checkout` and `thank-you`. I don't think it'll affect things too much, but if I was reading this without knowing the detail I might wonder why we're anchoring in some cases but not others (which might in turn make me think `checkout` was matching more than just `/uk/checkout`.

support-frontend/assets/helpers/abTests/abtestDefinitions.ts
```diff
@@ -131,13 +131,10 @@ export const tests: Tests = {
...
+		targetPage: '/subscribe$',
 		excludeContributionsOnlyCountries: true,
-		canRun: () => {
-			const urlSearchParams = new URLSearchParams(window.location.search);
-			const productParam = urlSearchParams.get('product');
-			return !productParam || productParam === 'DigitalSubscription';
-		},
```
tjmw, line 141: 🎉  Nice

tjmw (approved):
Looks good!

I left some really nitpicky comments on the regex to take or leave.

support-frontend/assets/helpers/abTests/abtestDefinitions.ts
paul-daniel-dempsey: I thought `checkout` and `thank-you` could have UrlParams (ie `product` or `ratePlan`) and the `$` anchor would cause this not to be recognised? Similarly there are none with  those that end in `$`? Yep will remove the UK brackets.

paul-daniel-dempsey: Hmmm, must be stripping out the urlParams because if I regex test it it apears to need the $ removed. eg below for checkout -> 
image

tjmw: Oh sorry yeah I think at the time the regex is checked in `getParticipations` the path will be `/uk/checkout` and not `/uk/checkout?product=DigitalSubscription&ratePlan=Monthly`.

paul-daniel-dempsey: Think if someone was to test the regex might be ok to stick with missing `$` unless you think its just cleaner without. Happy to do either, what do you think?

PR #6782, opened 2025-02-13
GuardianAdLite: Add Landing Page poster image

paul-daniel-dempsey:

Inserts (guim optimised) poster image onto Guardian Ad-Lite Landing Page

[**Trello Card**](https://trello.com/c/mocDMfVV/1405-guardian-ad-lite-landing-page-image-update)

## Screenshots

|before|after|
|-----|-----|
|image|image|


tjmw (approved):
👍 

The image looks a little blocky on a high pixel density screen to me. I can't tell if that's an intentional part of the design or if a higher res image would be better. The original in the Grid is quite small, do you know where that came from?

paul-daniel-dempsey:
> 👍
> 
> The image looks a little blocky on a high pixel density screen to me. I can't tell if that's an intentional part of the design or if a higher res image would be better. The original in the Grid is quite small, do you know where that came from?

Agree, think its by design, its the original size PNG from the figma design link Shraddha pointed me too [here](https://www.figma.com/design/daOtcjwsPKfCc1GvSpuBn7/Consent-or-Pay---user-flows?node-id=2406-5188&t=xdZdASRyL36YDDlf-1). Can revisit if necessary would guess.

PR #6783, opened 2025-02-14
round print saving vs retail % down

GHaberis:

`getDiscountVsRetail` returns the "saving vs retail" % to show on a pricing card on the print subscriptions pages when a promotion is applied. Marketing have stated that we should never overstate a discount, even by a fraction of a %, so a saving vs retail of 5.75% should be advertised as 5%, not 6% as it is now. Therefore we always round down to the nearest whole number.


tjmw (approved):
👍

paul-daniel-dempsey (approved):
Rounding down looks good, new test passes too. Great. Smoke and Cron tests pass too.

PR #6795, opened 2025-02-19
Pd/one time checkout email verification

paul-daniel-dempsey:

### DIFF
--- a/support-frontend/assets/helpers/abTests/abtestDefinitions.ts
+++ b/support-frontend/assets/helpers/abTests/abtestDefinitions.ts
@@ -15,6 +15,8 @@ export const pageUrlRegexes = {
 			'^(?!(?:/subscribe/(paper|weekly)/checkout$))(?:/(uk|us|ca|eu|nz|int))?/(checkout|one-time-checkout|contribute|thankyou|thank-you)(/.*)?$',
 		usLandingPageOnly: '/us/contribute$',
 		genericCheckoutOnly: '(uk|us|au|ca|eu|nz|int)/checkout|thank-you(/.*)?$',
+		oneTimeCheckoutOnly:
+			'(uk|us|au|ca|eu|nz|int)/one-time-checkout|thank-you(/.*)?$',
 	},
 	subscriptions: {
 		paper: {
@@ -91,6 +93,27 @@ export const tests: Tests = {
 		seed: 2,
 		targetPage: pageUrlRegexes.contributions.allLandingPagesAndThankyouPages,
 		excludeContributionsOnlyCountries: true,
+	},
+	oneTimeConfirmEmail: {
+		variants: [
+			{
+				id: 'control',
+			},
+			{
+				id: 'variant',
+			},
+		],
+		audiences: {
+			ALL: {
+				offset: 0,
+				size: 1,
+			},
+		},
+		isActive: false,
+		referrerControlled: false, // ab-test name not needed to be in paramURL
+		seed: 5,
+		targetPage: pageUrlRegexes.contributions.oneTimeCheckoutOnly,
+		excludeContributionsOnlyCountries: false,
 	},
 	digitalEditionCheckout: {
 		variants: [
--- a/support-frontend/assets/pages/[countryGroupId]/checkout/components/PersonalDetailsFields.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/checkout/components/PersonalDetailsFields.tsx
@@ -17,6 +17,7 @@ type PersonalDetailsFieldsProps = {
 	isEmailAddressReadOnly: boolean;
 	confirmedEmail: string;
 	setConfirmedEmail: (value: string) => void;
+	requireConfirmedEmail: boolean;
 };
 
 export function PersonalDetailsFields({
@@ -30,6 +31,7 @@ export function PersonalDetailsFields({
 	isEmailAddressReadOnly,
 	confirmedEmail,
 	setConfirmedEmail,
+	requireConfirmedEmail,
 }: PersonalDetailsFieldsProps) {
 	const [firstNameError, setFirstNameError] = useState<string>();
 	const [lastNameError, setLastNameError] = useState<string>();
@@ -72,7 +74,7 @@ export function PersonalDetailsFields({
 					}}
 				/>
 			</div>
-			{!isEmailAddressReadOnly && (
+			{!isEmailAddressReadOnly && requireConfirmedEmail && (
 				<div>
 					<TextInput
 						id="confirm-email"
--- a/support-frontend/assets/pages/[countryGroupId]/checkout/components/PersonalDetailsFields.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/checkout/components/PersonalDetailsFields.tsx
@@ -8,10 +8,10 @@ import {
 
 type PersonalDetailsFieldsProps = {
 	children: React.ReactNode;
-	firstName: string;
-	setFirstName: (value: string) => void;
-	lastName: string;
-	setLastName: (value: string) => void;
+	firstName?: string;
+	setFirstName?: (value: string) => void;
+	lastName?: string;
+	setLastName?: (value: string) => void;
 	email: string;
 	setEmail: (value: string) => void;
 	isEmailAddressReadOnly: boolean;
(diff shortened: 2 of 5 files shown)


support-frontend/assets/pages/[countryGroupId]/checkout/components/PersonalDetailsFields.tsx
```diff
@@ -111,74 +113,78 @@ export function PersonalDetailsFields({
...
-					pattern={doesNotContainExtendedEmojiOrLeadingSpace}
-					onInvalid={(event) => {
-						preventDefaultValidityMessage(event.currentTarget);
-						const validityState = event.currentTarget.validity;
-						if (validityState.valid) {
-							setFirstNameError(undefined);
-						} else {
-							if (validityState.valueMissing) {
-								setFirstNameError('Please enter your first name.');
+			{firstName !== undefined && (
```
Comment (GHaberis) [inline support-frontend/assets/pages/[countryGroupId]/checkout/components/PersonalDetailsFields.tsx:116]: Should we also check setFirstName has been passed in here too? eg.

```{firstName && setFirstName && <div> ... </div>}```

The reason I suggest this is do we really see a use case where we'd render the `firstName` field but not need a `setFirstName` handler in place too?

Comment (GHaberis) [reply support-frontend/assets/pages/[countryGroupId]/checkout/components/PersonalDetailsFields.tsx:116]: Lets discuss whether reusing the `PersonalDetailsFields.tsx` component is the best approach here, it introduces complexity to the component, and a few new potential failure points, it might be simpler to stick with the old approach of using an inline `<TextInput>` for email fields in the one-time-checkout?

Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/checkout/components/PersonalDetailsFields.tsx:116]: Maybe `PersonalDetailsFields.tsx` component could contain a `PersonalEmailFields.tsx` component that could also be called directly from the `oneTimeCheckoutComponent` instead? Would that be a better solution? Then no need for additional `setFirstName` checks you mention above too?

Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/checkout/components/PersonalDetailsFields.tsx]: Have done the refactor listed above. It avoids using uneccesary `firstName`/`setFirstName` checks and is componentised better/simpler now too. Hope thats what you were suggesting.

File: support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx [hunk 2]
```diff
@@ -939,6 +939,7 @@ export function CheckoutComponent({
 								setConfirmedEmail={(confirmedEmail) =>
 									setConfirmedEmail(confirmedEmail)
 								}
+								requireConfirmedEmail={true}
 							>
 								<Signout isSignedIn={isSignedIn} />
```
Comment (GHaberis) [inline support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx:944]: The`<Signout />` component, is passed as a child to `<PersonalDetailsFields />` which in turn passes it on as a child to `<PersonalEmailFields />`, I'd like to understand if there's a good reason for doing this vs importing  `<Signout />` in `<PersonalEmailFields />` and including it as inline component there? It needs access to the `isSignedIn` prop, but we're passing that down to `<PersonalDetailsFields />` already, which could easily pass it on to `<PersonalEmailFields />` .

The only reason I can think of for passing `<Signout />` as a child is we have a use case where we want to render `<PersonalEmailFields />` without a `<Signout />` link? But I can't think of any instances of that? And even then I'd suggest we pass it from whatever component includes `<PersonalEmailFields />` rather than down from the `checkoutComponent`.

If we always have a `<Signout />` link with the email, then maybe we should just import `<Signout />` in `<PersonalEmailFields />` and include it as inline component there?

support-frontend/assets/pages/[countryGroupId]/checkout/components/PersonalDetailsFields.tsx
```diff
@@ -30,87 +31,23 @@ export function PersonalDetailsFields({
...
-								} else {
-									setConfirmedEmailError('Please enter a valid email address.');
-								}
-							}
-						}}
-					/>
-				</div>
-			)}
-			{children}
+			<PersonalEmailFields
```
Comment (GHaberis) [inline support-frontend/assets/pages/[countryGroupId]/checkout/components/PersonalDetailsFields.tsx:41]: Nice, I like the idea to extract the Email and Confirm Email inputs into a separate component that encapsulates the logic, which can be shared between `checkoutComponent` and `omeTimeCheckoutComponent` 👏

support-frontend/assets/pages/[countryGroupId]/checkout/components/PersonalDetailsFields.tsx
Comment (GHaberis) [reply support-frontend/assets/pages/[countryGroupId]/checkout/components/PersonalDetailsFields.tsx]: @paul-daniel-dempsey https://github.com/guardian/support-frontend/pull/6795#discussion_r1971410896, looks neat 👍

GHaberis:
Something I think that will need adding: When a user pays via the Express Checkout Element we'll need to call `setConfirmEmail` with the email address provided by the payment provider, otherwise the the confirm email field will be in an invalid state and the form will fail to submit (it could be worth validating this happens before implementing the fix), the same fix was applied on the `checkoutComponent` here: https://github.com/guardian/support-frontend/blob/cc1340077017220049243bd2b7c8aaf3cc04e462/support-frontend/assets/pages/%5BcountryGroupId%5D/components/checkoutComponent.tsx#L874

support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx
Comment (paul-daniel-dempsey) [reply support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx]: Yes understand, because its always there, so why pass a <Signout /> down when you can just pass a smaller  the `signedIn` boolean prop instead and load the component down inside <PersonalEmailFields />. Updated to do that.

paul-daniel-dempsey:
> Something that will need addressing: When a user pays via the Express Checkout Element we'll need to call `setConfirmEmail` with the email address provided by the payment provider, otherwise the the confirm email input field will be in an invalid state and the form will fail to submit (it could be worth validating this happens before implementing the fix), the same fix was applied on the `checkoutComponent` here:
> 
> https://github.com/guardian/support-frontend/blob/cc1340077017220049243bd2b7c8aaf3cc04e462/support-frontend/assets/pages/%5BcountryGroupId%5D/components/checkoutComponent.tsx#L874

Yes your right, this failed as shown -> 
image

support-frontend/assets/helpers/abTests/abtestDefinitions.ts
```diff
@@ -15,6 +15,8 @@ export const pageUrlRegexes = {
 			'^(?!(?:/subscribe/(paper|weekly)/checkout$))(?:/(uk|us|ca|eu|nz|int))?/(checkout|one-time-checkout|contribute|thankyou|thank-you)(/.*)?$',
 		usLandingPageOnly: '/us/contribute$',
 		genericCheckoutOnly: '(uk|us|au|ca|eu|nz|int)/checkout|thank-you(/.*)?$',
```
GHaberis, line 17: I don't think there's any impact here, but maybe we should tighten the `genericCheckoutOnly`/`oneTimeCheckoutOnly` regex as the `thank-you` route is the same. EG. I'm pretty sure a user going through the generic checkout and landing on the TY page would then get bucketed in any test targeting `oneTimeCheckoutOnly` on the TY page view.

GHaberis: We'd probably have to add something to the `oneTimeCheckoutOnly` regex to look for the absence of a `product=` query string param, and the opposite for the `genericCheckoutOnly` regex, looking for the presence of the `product=` query string param.

GHaberis: @paul-daniel-dempsey I also think we can do this in a follow-up PR

GHaberis (approved):
Nice, looks good 👏

support-frontend/assets/helpers/abTests/abtestDefinitions.ts
paul-daniel-dempsey: Have updated the last commit to attempt this regx change (looks for `?contribution` or `?product` after `thank-you` in url). would be worth going through on Monday whether its good enough  if you have time. Will not put PR in until we have checked this out this approach.

support-frontend/assets/helpers/abTests/__tests__/abtestTest.ts
```diff
@@ -635,8 +638,20 @@ it('targetPage matching', () => {
 		targetPageMatches('/uk/contribute/checkout', genericCheckoutOnly),
 	).toEqual(false);
 	expect(targetPageMatches('/uk/checkout', genericCheckoutOnly)).toEqual(true);
-	expect(targetPageMatches('/uk/thank-you', genericCheckoutOnly)).toEqual(true);
+	expect(
+		targetPageMatches(
+			'/uk/thank-you?product=SupporterPlus&ratePlan=Monthly&userType=current',
+			genericCheckoutOnly,
+		),
+	).toEqual(true);
```
GHaberis, line 646: Good to add the test 👍 , we could also add a test for a `false` condition, eg. testing the `genericCheckoutOnly` regex with the `thank-you` page URL without the `product` query string param, `targetPageMatches` should return `false` in this scenario.

paul-daniel-dempsey: Good idea, added `genericCheckoutOnly` & `oneTimeCheckoutOnly` regex with the `thank-you `page URL only. Thanks.

PR #6805, opened 2025-02-26
Extract form submission code out of the checkout component

rupertbates:
This PR moves the logic related to submitting the generic checkout form to the server and handling the response, out of checkoutComponent.tsx and into separate files.

## Why are you doing this?
checkoutComponent.tsx currently contains several unrelated concerns from presentation through to the server interaction code mentioned above. By separating them out the aim is to apply the single responsibility principle - that "A module should be responsible to one, and only one, actor".

The extracted functions are fairly different to the original version because the original used React hooks extensively, so to move the logic out of the component and to introduce a separation of concerns, I have had to refactor that dependency on React out of the code.

## Testing
I have deployed this to CODE and run the E2E smoke tests against it successfully


support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx
```diff
@@ -248,7 +168,7 @@ export function CheckoutComponent({
 	abParticipations,
 }: CheckoutComponentProps) {
 	/** we unset any previous orders that have been made */
-	unsetThankYouOrder();
+	// unsetThankYouOrder(); TODO: after refactor this clears the session state after the submit is finished but before we get to the thank you page which causes an error
```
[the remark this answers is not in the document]
Comment (GHaberis) [reply support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx:171]: I'm not sure why this was originally added tbh, we use the `thankYouOrder` item in session storage to retrieve info about the order to be rendered on the TY page such as `firstName`, `email` and `paymentMethod`. The only reason to unset AFAICT would be to prevent rendering incorrect info on the TY page if they order two products in the same session? However I can't see how this scenario would be possibke as we call setThankYouOrder on form submission, so it should always have the latest correct details by the time a user lands on the TY page: `https://github.com/guardian/support-frontend/blob/pd/one-time-checkout-email-verification/support-frontend/assets/pages/%5BcountryGroupId%5D/components/checkoutComponent.tsx#L638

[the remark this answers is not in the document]
Comment (rupertbates) [reply support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx:171]: Yes that was the conclusion I came to as well

support-frontend/assets/pages/[countryGroupId]/components/formOnSubmit.ts
```diff
@@ -0,0 +1,204 @@
...
+		throw new FormSubmissionError(
+			'Sorry, something went wrong.',
+			appropriateErrorMessage(
+				processPaymentResponse.failureReason ?? 'unknown',
+			),
+		);
+	}
+};
+
+const createNewSupportWorkersExecution = async (
```
Comment (tjmw) [inline support-frontend/assets/pages/[countryGroupId]/components/formOnSubmit.ts:138]: Would it make sense to move these methods around creating the support workers request and polling for the status to a separate file?

support-frontend/assets/pages/[countryGroupId]/components/formOnSubmit.ts
```diff
@@ -0,0 +1,204 @@
...
+			userType: undefined,
+			processPaymentResponse: {
+				status: 'failure',
+				failureReason: errorReason,
+			},
+		};
+	}
+};
+
+const goToThankYouPage = (
```
Comment (tjmw) [inline support-frontend/assets/pages/[countryGroupId]/components/formOnSubmit.ts:178]: Not necessarily for now, but perhaps in future the return value of `submitForm` could be used to represent success or failure to the caller? The `success` status could even include the URL to redirect to, e.g.

```
{
  status: "success",
  redirectPath: "/uk/thank-you..."
}
```

Then we'd decouple this code from needing to know about the `window`.

tjmw (approved):
Nice, a definite improvement 👍

support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx
```diff
@@ -429,244 +348,53 @@ export function CheckoutComponent({
...
-					processPaymentResponse.failureReason,
-				);
-				setErrorMessage('Sorry, something went wrong.');
-				setErrorContext(
-					appropriateErrorMessage(
-						processPaymentResponse.failureReason ?? 'unknown',
-					),
-				);
-				setIsProcessingPayment(false);
+				setErrorMessage('Sorry, something went wrong');
```
Comment (GHaberis) [inline support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx:391]: I wonder whether we should log errors that aren't `FormSubmissionError` in Sentry? I don't think we'd want to do that for `FormSubmissionError` errors as they're known errors that could occur and are handled with meaningful messages displayed.

support-frontend/assets/pages/[countryGroupId]/components/createSubscription.ts
```diff
@@ -0,0 +1,131 @@
+import type { ErrorReason } from 'helpers/forms/errorReasons';
```
Comment (tjmw) [inline support-frontend/assets/pages/[countryGroupId]/components/createSubscription.ts:1]: 👍  Thanks for doing this. I think it's nice to see all of this functionality grouped into one file.

support-frontend/assets/pages/[countryGroupId]/components/paymentFields.ts
```diff
@@ -0,0 +1,178 @@
...
+			paymentType: 'Stripe',
+			paymentMethod: setupIntent.payment_method as string,
+			stripePaymentType,
+			stripePublicKey,
+		};
+	}
+	return;
+};
+
+export const getPaymentFieldsForPaymentMethod = async (
```
Comment (paul-daniel-dempsey) [inline support-frontend/assets/pages/[countryGroupId]/components/paymentFields.ts:135]: very nicely pulled out of the `checkoutComponent`, am wondering a similar approach for `one-time-checkout` in future

support-frontend/assets/pages/[countryGroupId]/components/createSubscription.ts
```diff
@@ -0,0 +1,131 @@
...
+		);
+
+	const finalResponse = await pollUntilCompleteOrTimeout(getCurrentStatus);
+	return {
+		userType,
+		...finalResponse,
+	};
+};
+
+export function pollUntilCompleteOrTimeout(
```
Comment (paul-daniel-dempsey) [inline support-frontend/assets/pages/[countryGroupId]/components/createSubscription.ts:98]: Keeping this retry call alongside the create above  is good idea

paul-daniel-dempsey (approved):
A lot of breaking out and simplifcations, nice. Cron/Smoke tests all pass perfectly. Great job.

PR #6806, opened 2025-02-26
Migrate support-frontend from yarn to pnpm

tjmw:

Migrating the support-frontend client side project from yarn to pnpm.

[**Trello Card**](https://trello.com/c/vomKKRpT/621-migrate-to-pnpm-support-frontend)

## Why are you doing this?

We need to move off of yarn v1 as it's deprecated. Using `pnpm` is generally our preference for JS/TS projects as it has a number of DevX benefits (better performance, improvements in monorepos).

## Notes

The compressed-size-action build check doesn't work on this PR, as the current branch uses `pnpm` and the target is still on `yarn`, which is a scenario it (understandably) doesn't support.

Using `pnpm` seems to offer much better behaviour around the `@guardian/support-service-lambdas` package compared to yarn 1 which made it hard to update, requiring us to add a script to do this. It was non-obvious what was happening when things weren't working for developers, especially those outside the team.

The `package.json` has changed in a few places. `pnpm` is much stricter around "phantom dependencies" i.e. those which you rely on/import but which aren't explicitly added to the package.json. Yarn would allow these if they were available as transitive dependencies, i.e. a dependency of another package which we _do_ have in our package.json.

I also locked down to specific versions of react and react-dom (and their definitely typed counterparts). This was because I wanted to keep these versions identical as we migrated (I ended up getting mismatches which caused type errors).

I've commented out the caching config for dependencies in the workflow config. This was causing confusing behaviour as I worked on migrating. Once we're moved over to `pnpm` I think we can add it back. Interestingly it doesn't seem to have made the build any slower on this branch compared to recent builds on main.

## How to test

I've deployed to code and run the smoke and cron playwright tests.


.github/workflows/chromatic.yml
```diff
@@ -12,19 +12,25 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
+
+      - run: npm install --global corepack@0.31.0
```
tjmw, line 16: This is needed as there's an issue with pnpm signatures on the default corepack version. We also had to fix this in support-service-lambdas previously: guardian/support-service-lambdas#2666.

rupertbates (approved):
👍

PR #6811, opened 2025-02-28
Unify ActivePaperProducts and ActivePaperProductTypes

rupertbates:
Looking at the product options defined for newspaper we have a type `ActivePaperProducts` and a list of values `ActivePaperProductTypes`, however these are out of sync with each other.

This PR derives the type from the list so that they will always be in sync.


tjmw (commented):
Looks good 👍 

Do we still need to define inactive options like `Sixday`?

tjmw (approved):

rupertbates:
> Looks good 👍
> 
> Do we still need to define inactive options like `Sixday`?

Probably not no? That's a larger refactor though so I'll just go with this for now.

PR #6809, opened 2025-02-28
Put a switch around Stripe Express on the one time checkout

tjmw:
**TODO: Enable the switch in prod before merging** ✅ 

## What are you doing in this PR?

Adding support for a switch around Stripe Express (which powers Apple Pay and Google Pay) on the one time checkout.

[**Trello Card**](https://trello.com/c/lq9WEyx0/1444-one-time-payment-method-switches)

## Why are you doing this?

We recently discovered that although a switch exists in the RRCP it isn't respected in the one time checkout. So if we ever wanted to disable Stripe Express for any reason, the switch wouldn't have worked.

## How to test

Deployed to CODE and toggled the switch in the CODE RRCP. When enabled the Apple Pay button is rendered on the one time checkout. When disabled, the Apple Pay button is not rendered (it takes a moment to take effect due to the polling mechanism on the backend).

## Accessibility test checklist

-  [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-  [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-  [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-  [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

PR #6820, opened 2025-03-03
Revert "Put a switch around Stripe Express on the one time checkout"

tjmw:
Reverts guardian/support-frontend#6809

I think there may have been a problem with this, reverting while we investigate.


GHaberis (approved):
👍

PR #6825, opened 2025-03-04
Support National Delivery on the generic checkout

rupertbates:
We want to make the generic checkout support National Delivery subscriptions. To do this we need to fetch a list of valid delivery agents for users whose postcode is not in our list of 'inside the M25' postcodes.
We also need to send a different `FulfilmentOptions` value through to the back end when creating the sub - `NationalDelivery` rather than `HomeDelivery`.

I've chosen to make this work without changing the productKey query string parameter as this would cause a full page refesh and be a worse user experience. Instead we send a different fulfilment option for postcodes that are outside the M25 and change the product key which is passed through to the thank you page.

### DIFF
--- a/support-frontend/assets/helpers/redux/checkout/addressMeta/state.ts
+++ b/support-frontend/assets/helpers/redux/checkout/addressMeta/state.ts
@@ -1,3 +1,5 @@
+import type { DeliveryAgentsResponse } from '../../../../pages/[countryGroupId]/checkout/helpers/getDeliveryAgents';
+
 export type AddressMetaState = {
 	billingAddressMatchesDelivery: boolean;
 	deliveryInstructions?: string;
@@ -14,24 +16,4 @@ export type DeliveryAgentState = {
 export const initialState: AddressMetaState = {
 	billingAddressMatchesDelivery: true,
 	deliveryAgent: { isLoading: false, response: undefined },
-};
-
-export type DeliveryAgentsResponse = {
-	type:
-		| 'Covered'
-		| 'NotCovered'
-		| 'UnknownPostcode'
-		| 'ProblemWithInput'
-		| 'PaperRoundError';
-	agents?: DeliveryAgentOption[];
-};
-
-export type DeliveryAgentOption = {
-	agentId: number;
-	agentName: string;
-	deliveryMethod: string;
-	nbrDeliveryDays: number;
-	postcode: string;
-	refGroupId: number;
-	summary: string;
 };
--- a/support-frontend/assets/helpers/redux/checkout/addressMeta/thunks.ts
+++ b/support-frontend/assets/helpers/redux/checkout/addressMeta/thunks.ts
@@ -1,15 +1,8 @@
 import { createAsyncThunk } from '@reduxjs/toolkit';
-import type { DeliveryAgentsResponse } from './state';
+import type { DeliveryAgentsResponse } from '../../../../pages/[countryGroupId]/checkout/helpers/getDeliveryAgents';
+import { getDeliveryAgents } from '../../../../pages/[countryGroupId]/checkout/helpers/getDeliveryAgents';
 
 export const getDeliveryAgentsThunk = createAsyncThunk<
 	DeliveryAgentsResponse,
 	string
 >(`addressMeta/getDeliveryAgents`, getDeliveryAgents);
-
-async function getDeliveryAgents(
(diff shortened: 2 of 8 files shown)


support-frontend/assets/helpers/redux/checkout/addressMeta/state.ts
```diff
@@ -15,23 +17,3 @@ export const initialState: AddressMetaState = {
 	billingAddressMatchesDelivery: true,
 	deliveryAgent: { isLoading: false, response: undefined },
 };
-
```
rupertbates, line 18: I moved some code related to fetching the delivery agents from the server into a separate file so it could be used by both checkouts without all the other Redux related stuff

support-frontend/assets/helpers/redux/checkout/addressMeta/thunks.ts
```diff
@@ -1,15 +1,8 @@
 import { createAsyncThunk } from '@reduxjs/toolkit';
-import type { DeliveryAgentsResponse } from './state';
+import type { DeliveryAgentsResponse } from '../../../../pages/[countryGroupId]/checkout/helpers/getDeliveryAgents';
+import { getDeliveryAgents } from '../../../../pages/[countryGroupId]/checkout/helpers/getDeliveryAgents';
 
 export const getDeliveryAgentsThunk = createAsyncThunk<
 	DeliveryAgentsResponse,
 	string
 >(`addressMeta/getDeliveryAgents`, getDeliveryAgents);
-
```
rupertbates, line 8: This as well

support-frontend/assets/pages/[countryGroupId]/checkout/helpers/getDeliveryAgents.ts
```diff
@@ -0,0 +1,26 @@
+export type DeliveryAgentsResponse = {
```
Comment (rupertbates) [inline support-frontend/assets/pages/[countryGroupId]/checkout/helpers/getDeliveryAgents.ts:1]: This is the file it moved to

support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx
```diff
@@ -304,6 +318,20 @@ export function CheckoutComponent({
 	const [deliveryAddressErrors, setDeliveryAddressErrors] = useState<
 		AddressFormFieldError[]
 	>([]);
+	useEffect(() => {
```
Comment (rupertbates) [inline support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx:321]: When we have a valid delivery postcode, if it is outside the M25 then request delivery agents for it.

support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx
```diff
@@ -890,10 +926,39 @@ export function CheckoutComponent({
...
 
 						<FormSection>
 							<Legend>
-								{productDescription.deliverableTo ? '3' : '2'}. Payment method
+								{productDescription.deliverableTo
+									? deliveryPostcodeIsOutsideM25
+										? '4'
+										: '3'
+									: '2'}
+								. Payment method
```
Comment (rupertbates) [inline support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx:961]: The numbering logic gets a bit messy with two optional sections, I think it's readable enough though?

support-services/src/main/scala/com/gu/support/paperround/PaperRoundService.scala
```diff
@@ -1,18 +1,17 @@
...
 import com.gu.support.touchpoint.{TouchpointService, TouchpointServiceProvider}
 
+import scala.concurrent.duration.DurationInt
 import scala.concurrent.{ExecutionContext, Future}
 
 class PaperRoundServiceProvider(configProvider: PaperRoundConfigProvider)(implicit ec: ExecutionContext)
     extends TouchpointServiceProvider[PaperRoundService, PaperRoundConfig](configProvider) {
   override protected def createService(config: PaperRoundConfig) = {
-    new PaperRoundService(config, futureRunner)
+    new PaperRoundService(config, configurableFutureRunner(20.seconds))
```
rupertbates, line 14: This was the thing we got an alarm on a while back. It was timing out for me locally as well so I increased it.

rupertbates:
> @rupertbates : I noticed this side effect during my testing with a London postcode, not sure if the `homeDeliveryPostcodes` are attached to the `window.guardian` ? image

Well spotted, it shouldn't be showing at all for T3

PR #6854, opened 2025-03-14
Checkout: Ts&Cs Components Testing Storybook Approach

paul-daniel-dempsey:

Create Storybook of the Terms and Conditions components shown below:-
- paymentsTsAndTcs component
- summaryTsAndCs component

[**Trello Card**](https://trello.com/c/hUjNytsY/1495-onetimecheckout-checkout-tscs-components-snapshot-test-setup)

## Why are you doing this?
The ensure ALL changes affecting Ts&Cs are flagged and accepted.

## How to run
Visual regressions will be flagged for accpetance with the github action `UI Checks`

## Screenshot

|summaryTs&Cs|paymentTs&Cs|
|----|----|
|image|image|

### DIFF
--- a/support-frontend/assets/helpers/legalCopy.ts
+++ b/support-frontend/assets/helpers/legalCopy.ts
@@ -1,37 +1,26 @@
 import type { RegularContributionType } from 'helpers/contributions';
 import type { CountryGroupId } from 'helpers/internationalisation/countryGroup';
-import {
-	currencies,
-	detect,
-	fromCountryGroupId,
-} from 'helpers/internationalisation/currency';
+import { currencies, detect } from 'helpers/internationalisation/currency';
 import type { Promotion } from 'helpers/productPrice/promotions';
-import { getLowerProductBenefitThreshold } from 'helpers/supporterPlus/benefitsThreshold';
 import { simpleFormatAmount } from './forms/checkouts';
-import type { ActiveProductKey } from './productCatalog';
 
 export const productLegal = (
 	countryGroupId: CountryGroupId,
 	contributionType: RegularContributionType,
 	divider: string,
-	product: ActiveProductKey,
+	thresholdAmount: number,
 	promotion?: Promotion,
 ) => {
 	const isoCurrency = detect(countryGroupId);
 	const currency = currencies[isoCurrency];
-	const amount = getLowerProductBenefitThreshold(
-		contributionType,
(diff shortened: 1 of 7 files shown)


support-frontend/assets/helpers/legalCopy.ts
```diff
@@ -1,37 +1,26 @@
...
-import { getLowerProductBenefitThreshold } from 'helpers/supporterPlus/benefitsThreshold';
 import { simpleFormatAmount } from './forms/checkouts';
-import type { ActiveProductKey } from './productCatalog';
 
 export const productLegal = (
 	countryGroupId: CountryGroupId,
 	contributionType: RegularContributionType,
 	divider: string,
-	product: ActiveProductKey,
+	thresholdAmount: number,
```
GHaberis, line 11: @paul-daniel-dempsey I'm interested in what `thresholdAmount` represents and why it's required for the T&Cs? Is it the product price from the product catalog? 

If so, it feels like the "threshold" terminology is a bit if a hangover from the pre-catalog approach to getting prices (eg. when S+ was just a recurring contribution over a certain amount), so maybe we should rename this?  What do you think?

support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx
```diff
@@ -467,6 +469,20 @@ export function CheckoutComponent({
 	const returnToLandingPage = `/${geoId}${productLanding(productKey)}`;
 	const isAdLite = productKey === 'GuardianAdLite';
 
+	const contributionType =
```
Comment (GHaberis) [inline support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx:472]: Maybe we should reconsider this variable name? Referring to the billing period as `contributionType` feels like a hangover of when we were just selling different types of contribution?  It's not really a "contribution type" at Tiers 2 and 3 anymore, so doesn't seem appropriate terminology? It feels like `paymentFrequency` or just `billingPeriod` seems more appropriate? 

Also is this mapping of 'Monthly' to 'MONTHLY', 'Annual' to 'ANNUAL', 'OneTime' to 'ONE_OFF' something we do a lot? If it's something we often do maybe we should centralise into a helper?

Looking at it maybe this is a bigger refactor throughout the code to revisit the `ContributionType` type, and something to look at in the future? what do you think @paul-daniel-dempsey?

GHaberis (approved):
Nice 👍

PR #6849, opened 2025-03-13
Checkout: Ts&Cs tidy, tests and snapshots

paul-daniel-dempsey:
## What are you doing in this PR?

Re-organised all terms and conditions so its predominantly organised by 
props: `productKey`/`ContributionType`/`CountryGroupId`
optional props: `Promotion`/`ThresholdAmount` (currently remaining unchanged as used in tests)

These components have been amended->
1. paymentTsAndCs
2. summaryTsAndCs
3. orderSummaryTsAndCs (previously`contributionsOrderSummaryContainer.tsx`)

Storybook scenarios created for components above
React Testing Library snapshots created for components above

[**Trello Card**](https://trello.com/c/l7aUrNob/1421-onetimecheckout-checkout-tscs-components-improvements)

## Why are you doing this?

To make the terms and conditions easer to maintain, traverse and view.

## How to test
OneTimeCheckout: https://support.thegulocal.com/uk/one-time-checkout
AdLite: https://support.thegulocal.com/uk/checkout?product=GuardianAdLite&ratePlan=Monthly
DigitalEdition: https://support.thegulocal.com/uk/checkout?product=DigitalSubscription&ratePlan=Monthly
Contribution: https://support.thegulocal.com/uk/checkout?product=Contribution&ratePlan=Monthly&contribution=3
SupporterPlus: https://support.thegulocal.com/uk/checkout?product=SupporterPlus&ratePlan=Monthly
Digital+Print: https://support.thegulocal.com/uk/checkout?product=TierThree&ratePlan=DomesticMonthly

PR #6855, opened 2025-03-17
Checkout Thankyou: Subscription Start Component

paul-daniel-dempsey:

Re-use existing `subscriptionStart` component. 

Apply copy (shown below) for
- Guardian Weekly (portion of TierThree)
- Newpaper (delivery)
- Newspaper (subscription Card)

Dates
- NewpaperHome/National we need delivery date.
- GuardianWeekly we need the publication date.

NewsPaper Header Copy Update (complete /pending)
- Newspaper Pending : `Look out for an email from us confirming your subscription. It has everything you need to know about how to manage it in the future. As well as future communications on how to make the most of your subscription and weekly newsletters written by the editors.`
- Newspaper Complete : <above> + `You can opt out any time via your account`
Add Storybook tests
Add Component tests

[**Trello Card**](https://trello.com/c/EJkcUEb2/1483-ty-component-modify-existing-what-happens-next-subscriptionstart-component-for-all-print-products)

## Screenshots

|TierThree|GuardianWeekly|NewspaperDelivered|NewspaperSubsCard|
|----|----|----|----|
|image|image|image|image|

Note: Paper Content Based Upon
image

## How to test
Storybook : http://localhost:9001/?path=/story/checkouts-thank-you-module--subscription-start-paper-delivery

### DIFF
--- a/support-frontend/assets/components/thankYou/subscriptionStart/subscriptionStartItems.tsx
+++ b/support-frontend/assets/components/thankYou/subscriptionStart/subscriptionStartItems.tsx
@@ -83,9 +83,12 @@ export function SubscriptionStartBodyCopy({
 			break;
 		}
 		case 'GuardianWeeklyDomestic':
-		case 'GuardianWeeklyRestOfWorld':
-		default: {
+		case 'GuardianWeeklyRestOfWorld': {
 			copy = [guardianWeeklyBodyHeading];
+			break;
+		}
+		default: {
+			copy = [{ heading: '', body: [''] }];
 		}
 	}
 	return (
--- a/support-frontend/assets/components/thankYou/subscriptionStart/subscriptionStartItems.tsx
+++ b/support-frontend/assets/components/thankYou/subscriptionStart/subscriptionStartItems.tsx
@@ -16,6 +16,9 @@ const subscriptionItems = css`
 const boldText = css`
 	font-weight: bold;
 `;
+const paragraphSpacing = css`
+	margin-bottom: ${space[1]}px;
+`;
 
 export const benefitsHeader = 'What’s included?';
 export function BenefitsBodyCopy(): JSX.Element {
@@ -40,19 +43,17 @@ export function SubscriptionStartItems({
 					css={boldText}
 				>{`You will receive your newspaper from ${startDate}`}</span>
 			</p>
-			<div>
-				{productKey === 'SubscriptionCard' && (
-					<p>
-						You will receive your Subscription Card in your subscriber pack in
-						the post, along with your home delivery letter.
-					</p>
-				)}
-				<p>
-					Visit your chosen participating newsagent to pick up your newspaper
-					using your Subscription Card, or arrange a home delivery using your
-					delivery letter.
+			{productKey === 'SubscriptionCard' && (
+				<p css={paragraphSpacing}>
+					You will receive your Subscription Card in your subscriber pack in the
+					post, along with your home delivery letter.
 				</p>
-			</div>
+			)}
+			<p>
+				Visit your chosen participating newsagent to pick up your newspaper
+				using your Subscription Card, or arrange a home delivery using your
+				delivery letter.
+			</p>
 		</div>
 	);
 	const guardianWeeklyCopy = (
@@ -62,12 +63,10 @@ export function SubscriptionStartItems({
 					css={boldText}
 				>{`Your first issue of Guardian Weekly will be published on ${startDate}`}</span>
 			</p>
-			<div>
-				<p>
-					Please allow one to seven days after the publication date for your
-					copy to be delivered to your door, depending on postal services.
-				</p>
-			</div>
+			<p>
+				Please allow one to seven days after the publication date for your copy
+				to be delivered to your door, depending on postal services.
+			</p>
 		</div>
 	);
 	const copyContent: Partial<Record<ActiveProductKey, JSX.Element>> = {
@@ -77,18 +76,19 @@ export function SubscriptionStartItems({
 		GuardianWeeklyRestOfWorld: guardianWeeklyCopy,
 		GuardianWeeklyDomestic: guardianWeeklyCopy,
 		TierThree: (
-			<div>
-				<p>
-					<span css={boldText}>Your digital benefits start today.</span>
-				</p>
+			<>
+				{guardianWeeklyCopy}
 				<div>
+					<p>
+						<span css={boldText}>Your digital benefits start today.</span>
+					</p>
 					<p>
 						Please ensure you are signed in on all your devices to enjoy all
 						your benefits, including unlimited app access and uninterrupted
 						ad-free reading..
 					</p>
 				</div>
-			</div>
+			</>
 		),
 	};
 	return (
--- a/support-frontend/assets/components/thankYou/subscriptionStart/subscriptionStartItems.tsx
+++ b/support-frontend/assets/components/thankYou/subscriptionStart/subscriptionStartItems.tsx
@@ -85,7 +85,7 @@ export function SubscriptionStartItems({
 					<p>
 						Please ensure you are signed in on all your devices to enjoy all
 						your benefits, including unlimited app access and uninterrupted
-						ad-free reading..
+						ad-free reading.
 					</p>
 				</div>
 			</>
--- a/support-frontend/assets/components/thankYou/subscriptionStart/subscriptionStartItems.tsx
+++ b/support-frontend/assets/components/thankYou/subscriptionStart/subscriptionStartItems.tsx
@@ -104,7 +104,7 @@ export function SubscriptionStartBodyCopy({
 			break;
 		}
 		default: {
-			copy = () => <div></div>;
+			copy = () => <></>;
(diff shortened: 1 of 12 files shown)


support-frontend/assets/components/thankYou/thankYouModuleData.tsx
```diff
@@ -111,8 +112,9 @@ const defaultSupportReminder = {
 const defaultFeedbackSurveyHasBeenCompleted = false;
 
 export const getThankYouModuleData = (
-	countryId: IsoCountry,
+	productKey: ActiveProductKey,
```
paul-daniel-dempsey, line 115: Might be good to refactor a number of these props so they use this `productKey` (and `ratePlan`/`countryGroupId`) in future. Think outside scope of this PR but for a future cleanup PR would say.

support-frontend/assets/pages/[countryGroupId]/thankYou.tsx
```diff
@@ -157,7 +158,7 @@ export function ThankYou({
 	return (
 		<ThankYouComponent
 			geoId={geoId}
-			appConfig={appConfig}
+			csrf={csrf}
```
Comment (paul-daniel-dempsey) [inline support-frontend/assets/pages/[countryGroupId]/thankYou.tsx:161]: This was unavailable in the thankyou component tests so has been drilled down from here instead

support-frontend/assets/helpers/productCatalog.ts
```diff
@@ -166,6 +166,8 @@ const guardianAdLiteBenefits = [
 	{ copy: 'You can cancel at any time.' },
 ];
 
+const paperThankyouMessage = `Look out for an email from us confirming your subscription. It has everything you need to know about how to manage it in the future. As well as future communications on how to make the most of your subscription and weekly newsletters written by the editors. `;
+
```
maloky, line 170: could this copy live in the same file as the other? or are they not related?

support-frontend/assets/components/thankYou/subscriptionStart/subscriptionStartItems.tsx
```diff
@@ -29,36 +30,79 @@ export function BenefitsBodyCopy(): JSX.Element {
...
+			{copy.map(({ heading, body }) => (
+				<div>
+					<p>
+						<span css={boldText}>{heading}</span>
+					</p>
+					{body.map((body) => (
+						<p>{body}</p>
+					))}
+				</div>
+			))}
```
GHaberis, line 105: Would you consider using JSX directly instead of defining the copy inside an array of mixed strings and objects? I think this could make the code more readable and maintainable? The `copy` array contains both strings and objects, which could make iteration tricky. Also, the code assumes every element in `copy` has a `heading` and `body`, if this ever changes (eg. you need a list) you'll have to refactor everything. I think an alternative approach would be to define the content using JSX directly:

```jsx
const Copy = () => (
  <>
    <p>You will receive your newspaper from ${startDate}</p>
    <div>
      <p>
        <span css={boldText}>Your digital benefits start today.</span>
      </p>
      <p>
        Visit your chosen participating newsagent to pick up your newspaper using your Subscription Card, or arrange a home delivery using your delivery letter.
      </p>
    </div>
  </>
);
```

This would eliminate the need for data structures and avoids potential mapping issues.

support-frontend/assets/helpers/productCatalog.ts
paul-daniel-dempsey: We usually store entirely static copy here if its directly linked to the `productKey` (like this). Other coy thats dynamically changing would generally be in the relevent component.

support-frontend/assets/components/thankYou/subscriptionStart/subscriptionStartItems.tsx
```diff
@@ -29,36 +30,79 @@ export function BenefitsBodyCopy(): JSX.Element {
...
+	const paperHeading = `You will receive your newspaper from ${startDate}`;
+	const paperBody = `Visit your chosen participating newsagent to pick up your newspaper using your Subscription Card, or arrange a home delivery using your delivery letter.`;
+	const guardianWeeklyBodyHeading = {
+		heading: `Your first issue of Guardian Weekly will be published on ${startDate}`,
+		body: [
+			`Please allow one to seven days after the publication date for your copy to be delivered to your door, depending on postal services.`,
+		],
+	};
+	let copy;
+	switch (productKey) {
```
GHaberis, line 50: I wonder whether we should consider using an object lookup instead of a switch, we have what is essentially a simple mapping between product key and copy, eg. the example below (which also uses JSX per my other comment)...  

```jsx
const copyContent: Record<ProductKey, JSX.Element> = {
  HomeDelivery: (
    <div>
      <p>
        <span css={boldText}>Your digital benefits start today.</span>
      </p>
      <p>
        Please ensure you are signed in on all your devices to enjoy all your
        benefits, including unlimited app access and uninterrupted ad-free
        reading.
      </p>
    </div>
  ),
};
```
Object lookups are considered more performant for these kind of mappings as 
switch statements require sequential checks each condition till we get a match or default, the performance difference will probably be of negligible here, but it's something to consider. 

Also with a `switch` there's a little more boilerplate code, we have to define the `let copy`,  and then and the break for each case.

paul-daniel-dempsey:
> @paul-daniel-dempsey, nice to see the `thankYouComponent.test.tsx` tests for these components, very comprehensive! 👏 Did you find them useful when validating your changes to this component?

Yes I did, running it manually  keeps tight control of component changs am finding  on each build. Looking at Helens WIP  Observer Figma we need to apply a few more styling changes to match what she has shown. This should be, in turn, flagged here when they are done.

support-frontend/assets/components/thankYou/subscriptionStart/subscriptionStartItems.tsx
paul-daniel-dempsey: Yes, have refactored all copy in this component and can definitely see its easier to read and simpler to maintain. Good recommend, will try this approach in future. Updated

support-frontend/assets/helpers/productCatalog.ts
maloky: I think I may have a solution for these cases. Maybe we can go trough ot tomorrow and se what the team thinks about it.

GHaberis: > I think I may have a solution for these cases. Maybe we can go trough ot tomorrow and se what the team thinks about it.

Yes definitely! Lets discuss it in the Friday dev catch-up at 11am?

support-frontend/assets/components/thankYou/subscriptionStart/subscriptionStartItems.tsx
paul-daniel-dempsey: @GHaberis : Have refactored this once more, using the object lookup approach. Used `Partial<Record<ActiveProductKey, JSX.Element>>` to avoid having to explicitly define all cases.  Like the way it skips having to use a `let` before the `switch` especially. Would be good if you could check it looks optimal?

GHaberis: @paul-daniel-dempsey yes the refactor looks good! Using the `Partial` for `ActiveProductKey` makes sense too 👍 , I just had one last comment following the refactor: https://github.com/guardian/support-frontend/pull/6855#issuecomment-2740298431

support-frontend/assets/components/thankYou/subscriptionStart/subscriptionStartItems.tsx
```diff
@@ -8,104 +8,92 @@ const downloadCopy = css`
...
-			copy = () => <></>;
-		}
-	}
-	return <span css={[downloadCopy, subscriptionItems]}>{copy()}</span>;
+			</div>
+		),
+	};
+	return (
+		<span css={[downloadCopy, subscriptionItems]}>
+			{copyContent[productKey]}
```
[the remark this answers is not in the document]
paul-daniel-dempsey: Empty Fragment, have added an `thankyouModule/SubscriptionStartEmpty` test storybook case now.
image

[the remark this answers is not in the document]
GHaberis: > Empty fragment I think is ok , have added an `thankyouModule/SubscriptionStartEmpty` test storybook case now. image {You can specify the header text so have left in displaying for consistency}

Do we want to render an empty subscription start component if there's no matching `copyContent`? I think it'd be better to render nothing if there's no available content?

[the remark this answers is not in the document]
paul-daniel-dempsey: @GHaberis : Ah yes, can see this similar recommend from Jorges Ts&Tc's PR comment [here](https://github.com/guardian/support-frontend/pull/6849#discussion_r2007279924). Have now used this recommended `null` approach on this `SubscriptionStartItems.tsx` component and removed the unnecessary storybook test mentioned previously in this thread. Updated.

GHaberis (approved):
Nice, looks good to me! 👍

PR #6856, opened 2025-03-17
Don't show National Delivery error for London postcodes

rupertbates:

A bug was introduced in #6825 which means that all visitors to the newspaper checkout would see an error message until they entered a valid postcode which is outside of the M25:
Screenshot 2025-03-17 at 17 32 29

For customers who live inside the M25 they would continue to see this error message even after entering their valid postcode, however they WOULD still be able to purchase successfully as the validation was working correctly.

This PR fixes that error.


paul-daniel-dempsey (approved):
Works perfectly, LG

PR #6862, opened 2025-03-19
Bump the resource-summary:document:size threshold

tjmw:

Bump the resource-summary:document:size threshold (lighthouse metric).

## Why are you doing this?

This lighthouse check is failing on every PR merge which is spamming the SR Dev chat channel. We've got an agenda item to discuss what to do with this check later in the week but for now lets increase the threshold slightly so we stop spamming the chat channel.

### DIFF
--- a/lighthouserc.json
+++ b/lighthouserc.json
@@ -7,7 +7,7 @@
         "interactive": ["error", { "maxNumericValue": 9500 }],
         "resource-summary:document:size": [
           "error",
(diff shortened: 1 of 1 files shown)


paul-daniel-dempsey (approved):
LG

johnduffell (approved):
nice one should make it quieter!  might be worth deleting the unneeded checklists from the PR description to reduce clutter

PR #6863, opened 2025-03-20
Stripe Hosted Checkout: Validate form data and redirect to Stripe

tjmw:

This PR implements the first part of the Stripe hosted checkout flow. For HomeDelivery/Sunday selecting the `Credit/Debit card` payment option and clicking `Continue to payment` will:

* Validate the user form data
* Create the checkout session with Stripe
* Redirect the user to Stripe (client side redirect)
* If the Stripe checkout is successful, redirect back to the checkout with the checkout session ID set as a URL fragment

This is not a complete flow. In prod the Stripe hosted checkout option will be hidden by a switch until ready. We also won't be showing any cards directing the user to the checkout for Sunday.

Missing elements:

* Persisting the user's details on the form before they are redirected to Stripe
* Processing the sub when the user returns to the support site

~~**Note: I'm currently using our default Stripe account (there's a secretKey value which needs to be set for prod BEFORE MERGING, it's already set in code). In a subsequent PR I'll migrate to using a different Stripe account.**~~ Done ✅ 

[**Trello Card**](https://trello.com/c/udzEIxlr/1519-stripe-checkout-validate-form-data-and-transfer-user-to-stripe-hosted-checkout-sunday-homedelivery)

## Why are you doing this?

Incrementally working towards Sunday only subs going through Stripe Checkout.

## How to test

* Deployed to CODE and run the smoke tests to ensure I've not broken any existing flows.
* Manually verified going through the flow for Sunday/HomeDelivery and successfully being transferred to Stripe.
* Verified that turning off the `stripeHostedCheckout` switch turns off `Credit/Debit card` for Sunday only.

### DIFF
--- a/support-frontend/app/services/StripeCheckoutSessionService.scala
+++ b/support-frontend/app/services/StripeCheckoutSessionService.scala
@@ -1,0 +1,100 @@
+package services
+
+import cats.data.EitherT
+import cats.implicits.{catsSyntaxApplicativeError, toBifunctorOps}
+import com.gu.i18n.Currency
+import com.gu.monitoring.SafeLogging
+import com.gu.support.config.StripeConfigProvider
+import io.circe.Decoder
+import io.circe.generic.semiauto.deriveDecoder
+import play.api.libs.ws.{WSAuthScheme, WSClient, WSResponse}
+import io.circe.parser.decode
+
+import scala.concurrent.{ExecutionContext, Future}
+
+sealed trait ResponseError
+object ResponseError {
+  case class DecodingError(error: io.circe.Error) extends ResponseError
+  case class ExecuteError(error: Throwable) extends ResponseError
+}
+
+case class CreateCheckoutSessionResponseSuccess(url: String, id: String)
+object CreateCheckoutSessionResponseSuccess {
+  implicit val decoder: Decoder[CreateCheckoutSessionResponseSuccess] = deriveDecoder
+}
+
+case class CheckoutPaymentMethod(id: String)
+object CheckoutPaymentMethod {
+  implicit val decoder: Decoder[CheckoutPaymentMethod] = deriveDecoder
+}
+case class CheckoutSetupIntent(id: String, payment_method: CheckoutPaymentMethod)
+object CheckoutSetupIntent {
+  implicit val decoder: Decoder[CheckoutSetupIntent] = deriveDecoder
+}
+case class RetrieveCheckoutSessionResponseSuccess(setup_intent: CheckoutSetupIntent)
+object RetrieveCheckoutSessionResponseSuccess {
+  implicit val decoder: Decoder[RetrieveCheckoutSessionResponseSuccess] = deriveDecoder
+}
+
+class StripeCheckoutSessionService(
+    configProvider: StripeConfigProvider,
+    client: WSClient,
+)(implicit ec: ExecutionContext)
+    extends SafeLogging {
+  val baseUrl: String = "https://api.stripe.com/v1"
+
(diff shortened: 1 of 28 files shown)


support-frontend/app/services/StripeCheckoutSessionService.scala
```diff
@@ -0,0 +1,128 @@
...
+      isTestUser: Boolean,
+      successUrl: String,
+  ): EitherT[Future, String, CreateCheckoutSessionResponseSuccess] = {
+    val privateKey = getPrivateKey(isTestUser)
+
+    // We use the default expiration of 24 hours
+    val data = Map(
+      "mode" -> Seq("setup"),
+      "success_url" -> Seq(successUrl),
+      "currency" -> Seq(currency.toString.toLowerCase), // Is this the format Stripe expect?
```
tjmw, line 58: Should we be using `iso` or similar on the Currency class?

support-frontend/app/services/StripeCheckoutSessionService.scala
```diff
@@ -0,0 +1,128 @@
...
+      "payment_method_types[]" -> Seq("card"),
+      "customer_email" -> Seq(email),
+    )
+
+    client
+      .url(s"$baseUrl/checkout/sessions")
+      .withAuth(privateKey, "", WSAuthScheme.BASIC)
```
tjmw, line 65: Comment about the usage of basic auth here?

support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx
```diff
@@ -151,6 +152,11 @@ type CheckoutComponentProps = {
 	abParticipations: Participations;
 };
 
+const shouldUseStripeHostedCheckout = (
+	productKey: ProductKey,
+	ratePlanKey: string,
+) => productKey === 'HomeDelivery' && ratePlanKey === 'Sunday';
```
Comment (tjmw) [inline support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx:158]: Need to include `SubscriptionCard` product key here too.

support-frontend/assets/helpers/productPrice/productOptions.ts
```diff
@@ -35,7 +35,7 @@ export type PaperProductOptions =
 	| typeof Sixday
 	| typeof Everyday;
 
-const ActivePaperProductTypes = [Everyday, Weekend, Saturday] as const;
+const ActivePaperProductTypes = [Everyday, Weekend, Saturday, Sunday] as const;
```
tjmw, line 38: Is it okay to add `Sunday` here? Any other effects I should be aware of?

rupertbates: I think adding that will put Sunday back on sale as soon as it's merged, so make sure that's what you want!

tjmw: Ah okay thanks. I'll dig into exactly what that means in terms of how this is used.

tjmw: This was a good spot - I don't think it's what I want right now. `Sunday` shouldn't go back on sale when this PR is merged. I wonder if I can differentiate by stage or something, I'd like to be testing this in CODE at least and if `Sunday` isn't active I can't complete the checkout flow.

tjmw: I've made a change - we'll now skip over Sunday when rendering the cards on the paper landing page, but other than that Sunday is an active product type. I've added a story for `PaperProductPrices` which should catch any change which accidentally makes the Sunday card visible.

PR #6787, opened 2025-02-14
Lambda: Typescript conversion bigquery-acquisitions-publisher

paul-daniel-dempsey:
## What are you doing in this PR?

Re-implementing the bigquery-acquisitions-publisher lambda in TypeScript.

[**Trello Card**](https://trello.com/c/iQqk6Nrl/1275-migrate-bigquery-acquisitions-publisher-to-typescript)

Note: we haven't deleted all the old Scala code in this PR as it's been useful to have it side by side with the new TypeScript. We'll raise a follow up PR to do this once this PR is merged.

## Why are you doing this?

There’s an issue with the compressed jar file size which is close to the limit of S3 based lambda deployments enforced by AWS. When we tried to update the google-cloud-bigquery Scala library this pushed us over the size limit. We could migrate to deploying using ECR images instead of S3, which has a much higher size limit. But we’d like to migrate to TypeScript anyway, and this will probably sidestep the size issue due to the smaller size of TypeScript/JS code artefacts.

## How to test

We've tested by going through the checkout process in CODE and inspecting the resulting row in `fact_acquisition_events`.

## How can we measure success?

Consumers of the `fact_acquisition_events` table shouldn't notice any differences.

PR #6867, opened 2025-03-24
Make schema more lenient in bigquery-acquisitions-publisher

tjmw:

Switch from nullable to nullish for a number of fields. This allows undefined or null. We saw a few errors in prod where a nullable field was omitted (seemed to be Feast IAPs).

[**Trello Card**](https://trello.com/c/iQqk6Nrl/1275-migrate-bigquery-acquisitions-publisher-to-typescript)

## Why are you doing this?

We had a few errors in prod after merging #6787.


paul-daniel-dempsey (approved):
LG

PR #6873, opened 2025-03-25
Update conditionals for checkout messages

GHaberis:

There are 2 components on the generic checkout that need the conditions in which they're rendered updating 

1) The "Guardian Patrons" message and 
2) The "charitable donation" message that sits directly beneath it.

These components are currently conditionally rendered using these rules:

- We hide them both for the Ad-Lite product, eg. https://support.theguardian.com/uk/checkout?product=GuardianAdLite&ratePlan=Monthly
- We hide the "Guardian Patrons" container for all products in the US, eg. https://support.theguardian.com/us/checkout?product=Contribution&ratePlan=Monthly&contribution=5

Screenshot 2025-03-24 at 15 21 01

### Changes

Both messages are "support" orientated in the messaging, which isn't appropriate for non-Contributions product checkouts, in this PR I'm therefore making the following changes

- Rename `GuardianTsAndCs` to  `ContributionCheckoutFinePrint`, we have a few "T&Cs" components, I think the new name is clearer in conveying when and where this component is used.
-  Only show `ContributionCheckoutFinePrint` and `PatronsMessage` on the generic checkout if the product `isRecurringContribution`

### DIFF
--- a/support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx
@@ -535,8 +535,6 @@ export function CheckoutComponent({
 	const isAdLite = productKey === 'GuardianAdLite';
 	const isRecurringContribution = productKey === 'Contribution';
 
-	console.log('*** isRecurringContribution ***', isRecurringContribution);
-
 	const contributionType =
 		productFields.billingPeriod === 'Monthly'
 			? 'MONTHLY'
--- a/support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx
@@ -624,7 +624,7 @@ export function CheckoutComponent({
 				<Box
 					cssOverrides={[
 						shorterBoxMargin,
-						isAdLite || !isRecurringContribution ? lengthenBoxMargin : css``,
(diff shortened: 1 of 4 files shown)


support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx
```diff
@@ -622,7 +624,7 @@ export function CheckoutComponent({
 				<Box
 					cssOverrides={[
 						shorterBoxMargin,
-						isAdLite ? lengthenBoxMargin : css``,
+						isAdLite || !isRecurringContribution ? lengthenBoxMargin : css``,
```
Comment (tjmw) [inline support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx:627]: Could this be simplified?

```suggestion
						!isRecurringContribution ? lengthenBoxMargin : css``,
```

I think `!isRecurringContribution` will be true for Ad-Lite so probably no need to cover that case separately. And if the `||` is removed the ternary can be un-negated/swapped to:

```
    isRecurringContribution ? css`` : lengthenBoxMargin,
```

tjmw (approved):
Looks great, really nice simplification 👏 

I left a minor comment to take or leave, definitely not a blocker.

support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx
Comment (GHaberis) [reply support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx:627]: Good spot, the suggestion is neater so I'll commit it 👍

paul-daniel-dempsey (commented):
LG

PR #6881, opened 2025-03-26
Remove overridable amount from Digital Subscriptions

rupertbates:
A long time ago in a galaxy far far away, we ran a test into the efficacy of giving digital benefits to increase ARPU. 

This involved supporting a variable price on Digital Subscriptions but this functionality is no longer needed so this PR removes it.


tjmw (approved):
Looks good (great PR stats: `+6 −167` 🎉 )

Looks like there's a failure in the build though, hopefully nothing major!

PR #6884, opened 2025-03-26
Remove optionality of stripe public key on the server

tjmw:

In the `StripePaymentFields` case class convert the `stripePublicKey` field's type from `Option[StripePublicKey]` to `StripePublicKey`. The `paymentFields` object is passed by the client to the server and the public key is used to map to a secret key. In the past, there were scenarios where the public key wasn't passed, so we had to fall back on deriving the key from the currency. This is no longer the case - we've tested a number of scenarios and the public key is always provided. This is backed up the the TypeScript type on the client `RegularStripePaymentIntentFields` in which the `stripePublicKey` field is required/not optional.

## Why are you doing this?

This makes the code simpler as we don't have do deal with the case where the field is a `None`. This also means we can remove the code the fall back on a public key based on the currency.

## How to test

We checked the digi subs checkout, which was the place we thought _might_ be omitting the public key from the request, but saw it was present.

I've deployed to CODE and run the smoke tests.

I've checked the last few days of logs for the warning in the (now removed) method to fallback on currency (`DEPRECATED FALLBACK - using currency`) and don't see any instances of it.


rupertbates (approved):
LGTM

PR #6885, opened 2025-03-27
Fix 100% discounts on the generic checkout

rupertbates:
Currently 100% discounts don't work correctly on the generic checkout - the order summary shows the correct price, but the CTA doesn't. This PR fixes that.

## Changes
- The reason for the issue was a falsy boolean where a zero price is equivalent to no discount. This has been changed to be more explicit
- After the first change, the CTA showed the correct amount but it was no longer possible to checkout out with Stripe, the fields on the form didn't appear:
Screenshot 2025-03-27 at 15 17 08
This is because we pass the amount in to the Stripe Elements component and the minimum amount which it supports is $0.50.
- The fix to this issue is to change the `mode` setting of the Elements component from `payment` to `subscription`. As [the documentation](https://docs.stripe.com/js/elements_object/create_without_intent) says 'The minimum amount is 0.50 USD or [equivalent in charge currency](https://docs.stripe.com/currencies#minimum-and-maximum-charge-amounts). **If mode is subscription, the value supports 0 to account for coupons and free trials, but any amount above 0 is subject to the minimum.**'

##  Testing
I have tested:
- regular S+ checkout with no promotion and Stripe checkout payment method
- S+ checkout with 100% promotion and Stripe checkout payment method
- S+ checkout with 100% promotion and Google Pay

[**Trello Card**](https://trello.com/c/ksxvvod9/1542-fix-100-discounts-on-generic-checkout)

### DIFF
--- a/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
@@ -207,6 +207,7 @@ export function Checkout({ geoId, appConfig, abParticipations }: Props) {
 		window.guardian.settings.switches.recurringPaymentMethods
 			.stripeExpressCheckout === 'On';
 
+	let elementsOptions = {};
 	let useStripeExpressCheckout = false;
 	if (stripeExpressCheckoutSwitch) {
 		/**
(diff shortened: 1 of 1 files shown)


support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -171,9 +171,8 @@ export function Checkout({ geoId, appConfig, abParticipations }: Props) {
 			billingPeriod,
 		);
 
-		const discountedPrice = promotion?.discountedPrice
-			? promotion.discountedPrice
-			: undefined;
+		const discountedPrice =
+			promotion !== undefined ? promotion.discountedPrice : undefined;
```
Comment (tjmw) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:175]: Would `const discountedPrice = promotion?.discountedPrice ?? undefined` be clearer here? I think that's the same.

tjmw (approved):
Looks good to me! (With one very minor suggestion to take or leave)

PR #6887, opened 2025-03-27
Rename StripeServiceForCurrency to StripeServiceForAccount

tjmw:

Following on from #6884, rename `StripeServiceForCurrency` to `StripeServiceForAccount` as it isn't really related to the currency any longer.


paul-daniel-dempsey (approved):
lg

PR #6891, opened 2025-03-31
Persist and retrieve persisted form fields on checkout

tjmw:

For the Stripe hosted checkout flow the checkout process looks like this:

```
+----------------------+     +------------------------+     +----------------------+
| Normal checkout form | --> | Stripe hosted checkout | --> | Back to support site |
+----------------------+     +------------------------+     +----------------------+
```

Validation of form fields etc takes place in the first step, before transferring to Stripe. When the user returns to our site, we don't want them to have to fill in their details again, but these details are needed to submit the create subscription request to support-workers. It's also possible that something may go wrong with the support-workers execution, in which case we'd want to place the user on the checkout form, with an error message, but their form fields still filled.

In this PR we persist the filled form fields to session storage under the key `checkoutSessionFormData`, with an expiration of 24 hours (managed by `@guardian/libs`). The reason we use 24 hours is to align with the expiration of the checkout session on Stripe's side. Since we're using session storage this data is tied to the current browser tab so the expiration will be the shorter of 24 hours and the lifespan of the current browser tab. After the user successfully returns from Stripe, we retrieve the details if there's a `checkoutSessionId` query string arg. If the checkout session ID doesn't match, the persisted form details aren't used. If there is a match, and the data is valid, we fill the form with the persisted details.

[**Trello Card**](https://trello.com/c/YU3EaFAw/1544-when-a-user-clicks-to-pay-by-credit-debit-card-on-the-generic-checkout-and-the-product-is-homedelivery-or-subscriptioncard-and-t)

## Why are you doing this?

Following on from #6863 this is the next step of implementing the complete flow.

## How to test

In CODE go to the Sunday HomeDelivery/Subs card checkout and pick debit/credit card payment option. After filling out the form with valid data and submitting you'll be transferred to Stripe to collect payment details. On success you will be taken back to the generic checkout with the form fields filled in to reflect the data you'd previously entered.

I've also run the smoke tests against this branch to ensure no existing flows are broken.

### DIFF
--- a/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
@@ -23,6 +23,7 @@ import type { Participations } from '../../helpers/abTests/models';
 import type { LandingPageVariant } from '../../helpers/globalsAndSwitches/landingPageSettings';
 import type { LegacyProductType } from '../../helpers/legacyTypeConversions';
 import { getLegacyProductType } from '../../helpers/legacyTypeConversions';
+import type { PersistableFormFields } from './checkout/helpers/stripeCheckoutSession';
 import { getFormDetails } from './checkout/helpers/stripeCheckoutSession';
 import { CheckoutComponent } from './components/checkoutComponent';
 
@@ -67,6 +68,28 @@ const getPromotionFromProductPrices = (
 		fulfilmentOption,
 		productOptions,
 	);
+};
+
+const attemptToRetrievePersistedFormData = (
+	urlSearchParams: URLSearchParams,
+): PersistableFormFields | undefined => {
+	const checkoutSessionIdUrlParam = 'checkoutSessionId';
+	const maybeCheckoutSessionId = urlSearchParams.get(checkoutSessionIdUrlParam);
+
+	if (maybeCheckoutSessionId) {
+		const persistedFormData = getFormDetails(maybeCheckoutSessionId);
+
+		if (persistedFormData) {
+			return persistedFormData;
+		}
+
+		// If there's no persisted data, remove the checkoutSessionId from the URL
+		const url = new URL(window.location.href);
+		url.searchParams.delete(checkoutSessionIdUrlParam);
+		window.location.href = url.toString();
+	}
+
+	return undefined;
 };
 
 export function Checkout({
@@ -261,20 +284,7 @@ export function Checkout({
 		);
 	}, []);
 
-	const checkoutSessionIdUrlParam = 'checkoutSessionId';
-	const maybeCheckoutSessionId = urlSearchParams.get(checkoutSessionIdUrlParam);
-
-	let persistedFormData;
-	if (maybeCheckoutSessionId) {
-		persistedFormData = getFormDetails(maybeCheckoutSessionId);
-
-		// If there's no persisted data, remove the checkoutSessionId from the URL
-		if (!persistedFormData) {
-			const url = new URL(window.location.href);
-			url.searchParams.delete(checkoutSessionIdUrlParam);
-			window.location.href = url.toString();
-		}
-	}
+	const persistedFormData = attemptToRetrievePersistedFormData(urlSearchParams);
 
 	return (
 		<Elements stripe={stripePromise} options={elementsOptions}>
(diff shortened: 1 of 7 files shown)


support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -255,6 +256,11 @@ export function Checkout({ geoId, appConfig, abParticipations }: Props) {
 		);
 	}, []);
 
+	const checkoutSessionId = window.location.hash;
+	const persistedFormData = checkoutSessionId
```
Comment (tjmw) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:260]: If there's a value in the hash param and we don't manage to retrieve persisted form data (because if fails to parse of it doesn't match the checkout session ID) should we update `window.location` to remove the hash param?

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -255,6 +256,21 @@ export function Checkout({ geoId, appConfig, abParticipations }: Props) {
...
+	if (maybeCheckoutSessionId) {
+		persistedFormData = getFormDetails(maybeCheckoutSessionId);
+
+		// If there's no persisted data, remove the checkoutSessionId from the URL
+		if (!persistedFormData) {
+			const url = new URL(window.location.href);
+			url.searchParams.delete(checkoutSessionIdUrlParam);
+			window.location.href = url.toString();
+		}
+	}
```
Comment (rupertbates) [inline support-frontend/assets/pages/[countryGroupId]/checkout.tsx:272]: I would extract this to a function which returns `PersistedFormData | undefined`

support-frontend/assets/pages/[countryGroupId]/checkout/helpers/formDataExtractors.ts
```diff
@@ -14,21 +14,21 @@ export const extractPersonalDataFromForm = (
...
-		lineTwo?: string;
-		city?: string;
-		state: string;
-		postCode: string;
+	deliveryAddress?: {
+		lineOne?: string | null;
+		lineTwo?: string | null;
+		city?: string | null;
+		state?: string | null;
+		postCode?: string | null;
```
Comment (rupertbates) [inline support-frontend/assets/pages/[countryGroupId]/checkout/helpers/formDataExtractors.ts:31]: I think if there is a delivery address then line one, city and postcode will always be present won't they?

support-frontend/assets/pages/[countryGroupId]/checkout/helpers/stripeCheckoutSession.ts
```diff
@@ -0,0 +1,94 @@
...
+	safeParse,
+	string,
+} from 'valibot';
+import { isoCountries } from 'helpers/internationalisation/country';
+import type {
+	FormAddressFields,
+	FormPersonalFields,
+} from './formDataExtractors';
+
+export type PersistableFormFields = {
```
Comment (rupertbates) [inline support-frontend/assets/pages/[countryGroupId]/checkout/helpers/stripeCheckoutSession.ts:18]: Should we derive this type from the schema?

Comment (tjmw) [reply support-frontend/assets/pages/[countryGroupId]/checkout/helpers/stripeCheckoutSession.ts:18]: Yeah that's a good shout

support-frontend/assets/pages/[countryGroupId]/components/formOnSubmit.ts
```diff
@@ -94,29 +95,39 @@ export const submitForm = async ({
...
-				personalData,
-				appliedPromotion,
-				productKey,
-				ratePlanKey,
-				contributionAmount,
-				paymentMethod,
-				geoId,
-				paymentRequest,
-			});
+	if (paymentMethod === 'StripeHostedCheckout') {
```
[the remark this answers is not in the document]
Comment (tjmw) [reply support-frontend/assets/pages/[countryGroupId]/components/formOnSubmit.ts:98]: It doesn't currently - in this PR there isn't a concept of the second time around. That will be done in a subsequent PR.

support-frontend/assets/pages/[countryGroupId]/checkout/helpers/formDataExtractors.ts
Comment (tjmw) [reply support-frontend/assets/pages/[countryGroupId]/checkout/helpers/formDataExtractors.ts:31]: The changes here were to make these consistent with `RegularPaymentRequestAddress` defined here: https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/helpers/forms/paymentIntegrations/readerRevenueApis.ts#L145

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
Comment (tjmw) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx]: I've made this change.

rupertbates (approved):

PR #6892, opened 2025-03-31
Connect Stripe hosted checkout frontend and backend changes

rupertbates:

Following on from #6863 and #6891 this PR adds code to connect the stripe hosted checkout to the backend and support-workers.

Now after the user returns from Stripe we can submit a request to the `/subscribe/create` endpoint containing the checkout session ID, which will be used in the support-workers execution to retrieve a setup intent and payment method.

## Why are you doing this?

Incrementally working toward the complete Stripe hosted checkout flow for Sunday newspaper subscriptions.

## How to test

Visit the generic checkout for Sunday HomeDelivery or SubscriptionCard and select Credit/Debit card as the payment method. Use test card details when transferred to Stripe. When you return back to the generic checkout select Credit/Debit card again and submit the form. This will create the support workers execution which should successfully complete.

### DIFF
--- a/support-frontend/assets/pages/[countryGroupId]/components/paymentFields.ts
+++ b/support-frontend/assets/pages/[countryGroupId]/components/paymentFields.ts
@@ -145,6 +145,7 @@ export const getPaymentFieldsForPaymentMethod = async (
 	stripePublicKey: string,
 	recaptchaToken: string | undefined,
 	formData: FormData,
+	checkoutSessionId: string | undefined,
 ): Promise<RegularPaymentFields | undefined> => {
 	if (paymentMethod === 'Stripe') {
 		return getStripePaymentFields(
@@ -181,6 +182,7 @@ export const getPaymentFieldsForPaymentMethod = async (
 	if (paymentMethod === 'StripeHostedCheckout') {
 		return {
 			paymentType: StripeHostedCheckout,
+			checkoutSessionId,
 		};
 	}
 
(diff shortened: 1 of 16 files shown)


support-frontend/assets/pages/[countryGroupId]/components/paymentFields.ts
```diff
@@ -181,6 +182,8 @@ export const getPaymentFieldsForPaymentMethod = async (
 	if (paymentMethod === 'StripeHostedCheckout') {
 		return {
 			paymentType: StripeHostedCheckout,
+			checkoutSessionId: checkoutSessionId ?? '',
```
Comment (tjmw) [inline support-frontend/assets/pages/[countryGroupId]/components/paymentFields.ts:185]: I wonder if there's anything we can do to avoid having to pass the empty string here. I think we did this because there are two versions of the payment fields - one when we first submit the form to the server where we haven't got a checkout session ID yet - we're about to create it but it doesn't exist when the submit happens, and then another when the user has been through the Stripe flow and we have got a checkout session ID, which is used by support-workers to get the payment method.

tjmw (approved):
Looks good, with one question about the empty string checkout session ID case (I think it would be okay to think about/resolve this in a future PR)

support-frontend/assets/pages/[countryGroupId]/components/paymentFields.ts
Comment (rupertbates) [reply support-frontend/assets/pages/[countryGroupId]/components/paymentFields.ts:185]: I'll have a look at refactoring to an option

Comment (rupertbates) [reply support-frontend/assets/pages/[countryGroupId]/components/paymentFields.ts]: That was pretty easy!

PR #6897, opened 2025-04-01
Refactor landing page test participations

tomrf1:
### Existing behaviour:
The AB tests [initialisation function](https://github.com/guardian/support-frontend/blob/main/support-frontend/assets/helpers/abTests/abtest.ts#L70) calls `getLandingPageParticipations`, which uses the settings from the landing page tool. If the user is on the landing page, or has a landing page participation in session storage, then it is added here to the set of test participations.
Later, we resolve the test participation to the actual landing page settings using `getLandingPageVariant`, for use in the components. The checkout and thankyou components also need these settings, as they may need the product benefits.

If a user arrives directly onto the checkout (not via the landing page) then the user is not put into a landing page test, and we use the `fallBackLandingPageSelection` settings. In this case we'd instead like to use the correct benefits settings from the landing page tool, even though we do not track the user as having been in the landing page test.

### New behaviour:
Move the call to `getLandingPageParticipations` out of the AB tests initialisation function.
Now we call it from the router.tsx.
`getLandingPageParticipations` now returns 2 fields:
```
interface LandingPageParticipationsResult {
	variant: LandingPageVariant;
	participations: Participations;
}
```
It will always return a variant, but only include the participation if the user is (or has been) on the landing page.
This way we ensure that the correct benefits copy is always available, but only track landing page test participation where appropriate.

### DIFF
--- a/support-frontend/assets/helpers/abTests/landingPageAbTests.ts
+++ b/support-frontend/assets/helpers/abTests/landingPageAbTests.ts
@@ -13,8 +13,6 @@ import {
 	setSessionParticipations,
 } from './sessionStorage';
 
-export type LandingPageSelection = LandingPageVariant & { testName: string };
-
 // Fallback config in case there's an issue getting it from the server
 export const fallBackLandingPageSelection: LandingPageVariant = {
 	name: 'CONTROL',
--- a/support-frontend/assets/helpers/abTests/landingPageAbTests.ts
+++ b/support-frontend/assets/helpers/abTests/landingPageAbTests.ts
@@ -103,7 +103,7 @@ interface LandingPageParticipationsResult {
 export function getLandingPageParticipations(
 	countryGroupId: CountryGroupId = CountryGroup.detect(),
 	path: string = window.location.pathname,
-	tests: LandingPageTest[] = getSettings().landingPageTests,
+	tests: LandingPageTest[] = getSettings().landingPageTests ?? [],
 	mvtId: number = getMvtId(),
 ): LandingPageParticipationsResult {
 	// Is there already a participation in session storage?
--- a/support-frontend/assets/helpers/abTests/landingPageAbTests.ts
+++ b/support-frontend/assets/helpers/abTests/landingPageAbTests.ts
@@ -3,8 +3,10 @@ import type {
 	LandingPageTest,
 	LandingPageVariant,
 } from '../globalsAndSwitches/landingPageSettings';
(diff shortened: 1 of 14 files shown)


support-frontend/assets/helpers/abTests/mvt.ts
```diff
@@ -0,0 +1,22 @@
+import * as cookie from '../storage/cookie';
+
+const MVT_COOKIE = 'GU_mvt_id';
+const MVT_MAX = 1_000_000;
+
+// Attempts to retrieve the MVT id from a cookie, or sets it.
+export function getMvtId(): number {
```
tomrf1, line 7: moved out of abtest.ts for reuse

support-frontend/assets/helpers/page/page.ts
```diff
@@ -15,6 +15,7 @@ import { getReferrerAcquisitionData } from 'helpers/tracking/acquisitions';
 import { getSettings } from '../globalsAndSwitches/globals';
 
 function setUpTrackingAndConsents(participations: Participations): void {
+	console.log({ participations });
```
tomrf1, line 18: moving this log to here because otherwise it would not include the landing page tests

support-frontend/assets/helpers/abTests/landingPageAbTests.ts
```diff
@@ -84,17 +84,43 @@ function randomNumber(mvtId: number, seed: string): number {
...
 }
 
+/**
+ * getLandingPageParticipations will always return a landing page variant, regardless of which
+ * page the user is on. We sometimes need these settings on other pages as well.
+ *
+ * If the user is on the landing page, or session storage contains a landing page participation,
+ * then it will also return the participations data for tracking.
+ * Otherwise we assume the user has not arrived via the landing page, and the participations
+ * object will be empty because we do not need to track it.
```
tjmw, line 97: 👍 Great comment

support-frontend/assets/helpers/abTests/mvt.ts
tjmw: It looks like this is still in `abtest.ts` - should it be removed?

tjmw (approved):
This looks good - nice work! 👍

support-frontend/assets/helpers/abTests/mvt.ts
tomrf1: good spot - removed now

PR #6888, opened 2025-03-27
Observer: Thankyou paper Sunday

paul-daniel-dempsey:
## What are you doing in this PR?

Update the Observer Thankyou Page including components 
- `SignIn`
- `SignUp`
- `WhatNext`

Applies to Sunday Paper Only.

Changes to HomeDelivery/NationalDelivery/Subscription.

[**Trello Card**](https://trello.com/c/AglZdDpe/1451-update-ty-page-for-sunday-only)

[**Figma**](https://www.figma.com/design/LF4FSOzeusjX057R1l7AUe/Observer---Q4---Q1-2025?node-id=2308-16741&t=14EVRy0vUWY3DuGw-0)

## How to test

Paper
Signed-In: {as below but logged-In}
Non-SignedIn : https://support.thegulocal.com/uk/thank-you?product=HomeDelivery&ratePlan=Sunday&userType=current
Guest :  https://support.thegulocal.com/uk/thank-you?product=HomeDelivery&ratePlan=Sunday&userType=new

Subscription
SignedIn: {as below but logged-In}
Non-SignedIn: https://support.thegulocal.com/uk/thank-you?product=SubscriptionCard&ratePlan=Sunday&userType=current
Guest :  https://support.thegulocal.com/uk/thank-you?product=SubscriptionCard&ratePlan=Sunday&userType=new

## Screenshots

Paper
|SignedIn|Non-SignedIn|Guest|
|----|----|----|
||image|image|

SubscriptionCard
|SignedIn|Non-SignedIn|Guest|
|----|----|----|
||image|image|

PR #6902, opened 2025-04-02
Checkout Thankyou: Subscription Start Component paper date correction

paul-daniel-dempsey:

Thankyou component updates ->
- `SubscriptionStartItems` : Currently display's Weekly Start Date not Paper Start Date.
- `whatNext` : tidy to use new call from fix above.

[**Trello Card**](https://trello.com/c/CXqJlXXP/1549-ty-component-subscriptionstart-date-component-fix)

## How to test

1. Checkout via -> https://support.thegulocal.com/uk/checkout?product=NationalDelivery&ratePlan=Everyday&userType=current
2. Checkout via -> https://support.thegulocal.com/uk/checkout?product=HomelDelivery&ratePlan=Everyday&userType=current
3. Checkout via -> https://support.thegulocal.com/uk/checkout?product=SubscriptionCard&ratePlan=Everyday&userType=current

## Screenshots

|before|after|
|----|----|
|image|image|

### DIFF
--- a/support-frontend/assets/components/thankYou/guardianAdLite/whatNext.tsx
+++ b/support-frontend/assets/components/thankYou/guardianAdLite/whatNext.tsx
@@ -3,7 +3,7 @@ import BulletPointedList from '../utilityComponents/BulletPointedList';
 
 type WhatNextProps = {
 	amount: string;
-	startDate: string;
+	startDate?: string;
 	isSignedIn?: boolean;
 };
 
@@ -12,17 +12,24 @@ export function WhatNext({
 	startDate,
 	isSignedIn = false,
 }: WhatNextProps): JSX.Element {
-	const bulletItems = [
+	const bulletAllItems = [
 		'You will receive an email confirming the details of your subscription',
 		`Your payment of £${amount}/month will be taken on ${startDate}`,
-	];
-	const bulletPointSignedIn = bulletItems.concat([
 		'You can now start reading the Guardian website on all your devices without personalised advertising',
-	]);
+	];
+	const displayBulletItems: string[] = bulletAllItems.filter((item, index) => {
+		switch (index) {
+			case 0:
+				return item;
+			case 1:
+				return !!startDate;
+			case 2:
+				return isSignedIn;
+			default:
+				return false;
+		}
+	});
 	return (
-		<BulletPointedList
-			items={isSignedIn ? bulletPointSignedIn : bulletItems}
-			color={palette.neutral[7]}
-		/>
+		<BulletPointedList items={displayBulletItems} color={palette.neutral[7]} />
 	);
 }
--- a/support-frontend/assets/components/thankYou/subscriptionStart/subscriptionStartItems.tsx
+++ b/support-frontend/assets/components/thankYou/subscriptionStart/subscriptionStartItems.tsx
@@ -30,7 +30,7 @@ export const subscriptionStartHeader = 'When will your subscription start?';
 
 type SubscriptionStartProps = {
 	productKey: ActiveProductKey;
-	startDate: string;
+	startDate?: string;
 };
 export function SubscriptionStartItems({
 	productKey,
@@ -39,11 +39,13 @@ export function SubscriptionStartItems({
 	const paperCopy = (
 		<span css={[downloadCopy, subscriptionItems]}>
 			<div>
-				<p>
-					<span
-						css={boldText}
-					>{`You will receive your newspaper from ${startDate}`}</span>
-				</p>
+				{startDate && (
+					<p>
+						<span
+							css={boldText}
+						>{`You will receive your newspaper from ${startDate}`}</span>
+					</p>
+				)}
 				{productKey === 'SubscriptionCard' && (
 					<p css={paragraphSpacing}>
 						You will receive your Subscription Card in your subscriber pack in
--- a/support-frontend/assets/components/thankYou/thankYouModuleData.tsx
+++ b/support-frontend/assets/components/thankYou/thankYouModuleData.tsx
@@ -113,7 +113,7 @@ export const getThankYouModuleData = (
 	csrf: CsrfState,
 	isOneOff: boolean,
 	amountIsAboveThreshold: boolean,
-	startDate: string,
+	startDate?: string,
 	email?: string,
 	campaignCode?: string,
 	isTier3?: boolean,
--- a/support-frontend/assets/components/thankYou/thankYouModuleData.tsx
+++ b/support-frontend/assets/components/thankYou/thankYouModuleData.tsx
@@ -28,11 +28,6 @@ import {
 	OPHAN_COMPONENT_ID_SURVEY,
 } from 'helpers/thankYouPages/utils/ophan';
 import { manageSubsUrl } from 'helpers/urls/externalLinks';
-import {
-	formatMachineDate,
-	formatUserDate,
-} from 'helpers/utilities/dateConversions';
-import { getWeeklyDays } from 'pages/weekly-subscription-checkout/helpers/deliveryDays';
 import AppDownloadBadges, {
 	AppDownloadBadgesEditions,
 } from './appDownload/AppDownloadBadges';
@@ -118,6 +113,7 @@ export const getThankYouModuleData = (
 	csrf: CsrfState,
 	isOneOff: boolean,
 	amountIsAboveThreshold: boolean,
+	startDate: string,
 	email?: string,
 	campaignCode?: string,
 	isTier3?: boolean,
@@ -125,7 +121,6 @@ export const getThankYouModuleData = (
 	supportReminder?: ThankYouSupportReminderState,
 	feedbackSurveyHasBeenCompleted?: boolean,
 	finalAmount?: number,
-	startDate?: string,
 	returnAddress?: string,
 	isSignedIn?: boolean,
 ): Record<ThankYouModuleType, ThankYouModuleData> => {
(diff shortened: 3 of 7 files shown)


support-frontend/assets/components/thankYou/thankYouModuleData.tsx
```diff
@@ -118,14 +113,14 @@ export const getThankYouModuleData = (
 	csrf: CsrfState,
 	isOneOff: boolean,
 	amountIsAboveThreshold: boolean,
+	startDate: string,
```
tjmw, line 116: I noticed that we sometimes pass in an empty string here, presumably to represent no `startDate`, which I think has allowed us to tighten the type here to just `string` (before it was optional). I wonder if using empty string is desirable, and whether `null` or `undefined` would be better? To me those more clearly represent the lack of start date. What do you reckon?

support-frontend/assets/pages/weekly-subscription-checkout/helpers/deliveryDays.ts
```diff
@@ -59,14 +64,54 @@ function getTierThreeDeliveryDate(today?: number) {
...
-	return firstValidDate;
-}
+const productDeliveryDate = (
+	productKey: ActiveProductKey,
+	paperProductOptions?: ActivePaperProductOptions,
+): Date | undefined => {
+	console.log(paperProductOptions);
+	switch (productKey) {
+		case 'GuardianAdLite':
+			return addDays(new Date(), 15);
```
tjmw, line 74: What does the delivery date mean for Ad Lite?

support-frontend/assets/pages/weekly-subscription-checkout/helpers/deliveryDays.ts
```diff
@@ -59,14 +64,54 @@ function getTierThreeDeliveryDate(today?: number) {
...
 
-function getGuardianAdLiteDate(today?: number) {
-	const firstValidDate = addDays(new Date(today ?? new Date()), 15);
-	return firstValidDate;
-}
+const productDeliveryDate = (
+	productKey: ActiveProductKey,
+	paperProductOptions?: ActivePaperProductOptions,
+): Date | undefined => {
+	console.log(paperProductOptions);
```
tjmw, line 71: Should this `console.log` be removed before merging?

support-frontend/assets/pages/weekly-subscription-checkout/helpers/deliveryDays.ts
```diff
@@ -59,14 +64,54 @@ function getTierThreeDeliveryDate(today?: number) {
...
+		case 'SubscriptionCard': {
+			if (paperProductOptions === undefined) {
+				throw new Error(`ratePlan not found for ${productKey}`);
+			}
+			const paperDeliveryDate =
+				productKey === 'SubscriptionCard'
+					? getPaymentStartDate(Date.now(), paperProductOptions)
+					: getHomeDeliveryDays(Date.now(), paperProductOptions)[0];
+			if (paperDeliveryDate === undefined) {
+				throw new Error('delivery date not found for Home Delivery');
```
tjmw, line 88: Is `Home Delivery` in this error accurate (looks like this code path could be reached for national delivery or subs card too?

support-frontend/assets/pages/weekly-subscription-checkout/helpers/deliveryDays.ts
```diff
@@ -59,14 +64,54 @@ function getTierThreeDeliveryDate(today?: number) {
 	return result;
 }
 
-function getGuardianAdLiteDate(today?: number) {
-	const firstValidDate = addDays(new Date(today ?? new Date()), 15);
-	return firstValidDate;
-}
+const productDeliveryDate = (
```
tjmw, line 67: 👍  For this method and switching on the product key, nice to see the logic all located together.

support-frontend/assets/pages/weekly-subscription-checkout/helpers/deliveryDays.ts
tjmw: What happens if this error gets thrown? Is it caught anywhere? Do we expect this to ever happen?

support-frontend/assets/pages/weekly-subscription-checkout/helpers/deliveryDays.ts
paul-daniel-dempsey: Yes, should of spotted that, removed.

support-frontend/assets/pages/weekly-subscription-checkout/helpers/deliveryDays.ts
paul-daniel-dempsey: Its a date used with the 'whatsNext' component, which will need specialist copy for the Observer Sunday in next PR. Hence putting it here (as we will need a special case for Observer Sunday Paper too). Does renaming function to `productDeliveryOrStartDate` work? Updated

support-frontend/assets/pages/weekly-subscription-checkout/helpers/deliveryDays.ts
paul-daniel-dempsey: > Is `Home Delivery` in this error accurate (looks like this code path could be reached for national delivery or subs card too?

I think you theoretically can only get an undefined from the HomeDelivery array, hence error message containing `Home Delivery`.

paul-daniel-dempsey: > What happens if this error gets thrown? Is it caught anywhere? Do we expect this to ever happen?

Looking at how to use throw, I don't think its necessary to throw these errors at all. Have opted for returning `undefined` for these theoretical possibilities (with an associated comment). Does that sound like a better approach? Updated

support-frontend/assets/pages/weekly-subscription-checkout/helpers/deliveryDays.ts
tjmw: Ah yes that rename sounds great, thanks for explaining.

support-frontend/assets/pages/weekly-subscription-checkout/helpers/deliveryDays.ts
paul-daniel-dempsey: > Is `Home Delivery` in this error accurate (looks like this code path could be reached for national delivery or subs card too?

That is a superb spot, just checked and should be using `getVoucherDays` for `NationalDelivery`. Thankyou, updated

support-frontend/assets/components/thankYou/thankYouModuleData.tsx
paul-daniel-dempsey: Previously had this but can see the advantage of `undefined` theoretically not displaying the whole line containing the date at all - a better situation. Also included some stories for the `whatNext` component directly affected. Updated.

support-frontend/assets/pages/weekly-subscription-checkout/helpers/deliveryDays.ts
tjmw: > Does that sound like a better approach? Updated

Great, yes I think that sounds good 👍

support-frontend/assets/pages/weekly-subscription-checkout/helpers/deliveryDays.ts
```diff
@@ -59,14 +65,58 @@ function getTierThreeDeliveryDate(today?: number) {
...
+				productKey === 'SubscriptionCard'
+					? getPaymentStartDate(Date.now(), paperProductOptions)
+					: productKey === 'HomeDelivery'
+					? getHomeDeliveryDays(Date.now(), paperProductOptions)[0]
+					: getVoucherDays(Date.now(), paperProductOptions)[0];
+			// paper home and voucher delivery date array empty check
+			if (paperDeliveryDate === undefined) {
+				return undefined;
+			}
+			return paperDeliveryDate;
```
tjmw, line 94: I think this could be simplified to just `return paperDeliveryDate` since if it's undefined, we return undefined :-)

support-frontend/assets/pages/weekly-subscription-checkout/helpers/deliveryDays.ts
```diff
@@ -59,14 +65,58 @@ function getTierThreeDeliveryDate(today?: number) {
...
+				const date = formatMachineDate(day);
+				return !invalidPublicationDates.some((dateSuffix) =>
+					date.endsWith(dateSuffix),
+				);
+			});
+			// guardian weekly delivery date array empty check
+			if (publicationStartDays[0] === undefined) {
+				return undefined;
+			}
+			return publicationStartDays[0];
```
tjmw, line 110: As above, I think we could just return `publicationStartDays[0]` here.

tjmw (approved):
👏  Great!

support-frontend/assets/components/thankYou/subscriptionStart/subscriptionStartItems.tsx
```diff
@@ -39,11 +39,13 @@ export function SubscriptionStartItems({
...
 		<span css={[downloadCopy, subscriptionItems]}>
 			<div>
-				<p>
-					<span
-						css={boldText}
-					>{`You will receive your newspaper from ${startDate}`}</span>
-				</p>
+				{startDate && (
+					<p>
+						<span
```
maloky, line 44: I think we can apply the bold style to the P directly and remove the span.

support-frontend/assets/components/thankYou/guardianAdLite/whatNext.tsx
```diff
@@ -12,17 +12,24 @@ export function WhatNext({
...
+			case 0:
+				return item;
+			case 1:
+				return !!startDate;
+			case 2:
+				return isSignedIn;
+			default:
+				return false;
+		}
+	});
```
maloky, line 31: I think we could build this better by conditionally adding this string to the array directly. no need to check for indexes and we always return strings.

paul-daniel-dempsey: Will re-asses this in the next PR in our meeting. I liked the approach but might be a simpler way.

support-frontend/assets/components/thankYou/subscriptionStart/subscriptionStartItems.tsx
paul-daniel-dempsey: Will re-asses this in the next [PR](https://github.com/guardian/support-frontend/pull/6888) in our meeting.

PR #6904, opened 2025-04-03
Use Tortoise Stripe account for Sunday newspaper payments

rupertbates:
This PR add key config and a new payment gateway for the Tortoise media Stripe account. This is so that we can process payment for Sunday only newspaper sales using their Stripe account not ours.

## Changes
- Add new Stripe public keys to the touchpoint.conf files which we can then use to look up the secret keys
- Create all the relevant secret keys in config - S3 for support-workers and parameter store for support-frontend
- Add a new payment gateway for the Tortoise media Stripe account
- Use the new config and Stripe account throughout the stack for Sunday only newspaper subscription. Blended subscriptions (those which contain both a Guardian and an Observer element) are unchanged.

[**Trello Card**](https://trello.com/c/P8mvEADx/1507-use-sunday-only-payment-gateway-support-workers)

### DIFF
--- a/support-frontend/app/services/StripeCheckoutSessionService.scala
+++ b/support-frontend/app/services/StripeCheckoutSessionService.scala
@@ -71,7 +71,10 @@ class StripeCheckoutSessionService(
       .withBody(data)
       .execute()
       .attemptT
-      .leftMap(error => error.getMessage)
+      .leftMap(error => {
+        logger.warn(s"Error creating Stripe checkout session: ${error.getMessage}")
+        "Failed to create Stripe checkout session"
+      })
       .subflatMap(decodeResponse[CreateCheckoutSessionResponseSuccess])
   }
 
--- a/support-frontend/app/services/StripeCheckoutSessionService.scala
+++ b/support-frontend/app/services/StripeCheckoutSessionService.scala
@@ -32,10 +32,6 @@ case class CheckoutSetupIntent(id: String, payment_method: CheckoutPaymentMethod
 object CheckoutSetupIntent {
   implicit val decoder: Decoder[CheckoutSetupIntent] = deriveDecoder
 }
-case class RetrieveCheckoutSessionResponseSuccess(setup_intent: CheckoutSetupIntent)
-object RetrieveCheckoutSessionResponseSuccess {
-  implicit val decoder: Decoder[RetrieveCheckoutSessionResponseSuccess] = deriveDecoder
-}
 
 class StripeCheckoutSessionService(
     configProvider: StripeConfigProvider,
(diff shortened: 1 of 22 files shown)


support-frontend/app/services/StripeCheckoutSessionService.scala
```diff
@@ -65,16 +71,16 @@ class StripeCheckoutSessionService(
       .withBody(data)
       .execute()
       .attemptT
-      .leftMap(error => {
-        logger.warn(s"Error creating Stripe checkout session: ${error.getMessage}")
-        "Failed to create Stripe checkout session"
-      })
+      .leftMap(error => error.getMessage)
```
tjmw, line 74: I had this previously, but wanted to be sure that no detail about the key could make it back to the controller in the error message. Hence logging the actual error message but returning something hard coded - what do you think?

support-frontend/app/views/windowGuardianPaymentConfig.scala.html
```diff
@@ -58,6 +58,17 @@
     REGULAR: window.guardian.stripeKeyDefaultCurrencies.REGULAR
   };
 
+  window.guardian.stripeKeyTortoiseMedia = {
+    ONE_OFF: {
+      default: "@paymentMethodConfigs.oneOffDefaultStripeConfig.tortoiseMediaAccount.rawPublicKey",
+      test: "@paymentMethodConfigs.oneOffTestStripeConfig.tortoiseMediaAccount.rawPublicKey"
```
tjmw, line 64: Since these won't (/shouldn't!) ever be used, should we set them to something unusable?

tjmw: Oh good question, I guess they both achieve the same thing. Hmm I'm not sure which one I prefer.

support-frontend/assets/helpers/forms/stripe.ts
```diff
@@ -1,50 +1,69 @@
+import type { ActiveProductKey } from '@guardian/support-service-lambdas/modules/product-catalog/src/productCatalog';
 import type { Stripe as StripeJs } from '@stripe/stripe-js';
 import { loadStripe } from '@stripe/stripe-js/pure';
 import { useEffect, useState } from 'react';
 import type { ContributionType } from 'helpers/contributions';
 import type { IsoCountry } from 'helpers/internationalisation/country';
 import type { IsoCurrency } from '../internationalisation/currency';
 
-export type StripeAccount = 'ONE_OFF' | 'REGULAR';
+export type StripeAccountType = 'ONE_OFF' | 'REGULAR';
```
rupertbates, line 9: I just renamed this to make it clearer what it represents. Quite a few of the changes in this PR are in files which reference this type.

support-frontend/app/views/windowGuardianPaymentConfig.scala.html
```diff
@@ -58,6 +58,17 @@
     REGULAR: window.guardian.stripeKeyDefaultCurrencies.REGULAR
   };
 
+  window.guardian.stripeKeyTortoiseMedia = {
+    ONE_OFF: {
+      default: "unused", // We will never need one-off payments for Tortoise Media
```
tjmw, line 63: 👍

tjmw (approved):
Nice work, this looks great.

PR #6909, opened 2025-04-04
Don't notify on Playwright cron failures

rupertbates:
This PR prevents the Playwright cron job from notifying on the SR Dev chat channel. 

This is to stop that channel getting spammed until we can make the tests less flakey.


paul-daniel-dempsey (approved):
LG

PR #6913, opened 2025-04-04
Add Tortoise Direct Debit Payment Gateway

rupertbates:
This PR configures support-frontend to use a different Zuora payment gateway for Observer (Sunday only) newspaper subscriptions. 
This is required as we now want to process Observer payments through Tortoise media's Go Cardless account, not GNMs.

~~Because of a difference in the way that GNM and Tortoise Media's Go Cardless accounts are configured, we need to pass an email address in the payment method object for the Tortoise account. This is not needed for our account and may be removed in future if the Tortoise account is updated to match the GNM one.~~ The Tortoise account has now been reconfigured to no longer require an email address so this additional code was removed from this PR.

[**Trello Card**](https://trello.com/c/lrrRQHUb/1551-use-different-gateway-for-gocardless-sunday-subs-in-support-workers)

### DIFF
--- a/support-workers/src/main/scala/com/gu/support/workers/lambdas/CreatePaymentMethod.scala
+++ b/support-workers/src/main/scala/com/gu/support/workers/lambdas/CreatePaymentMethod.scala
@@ -6,9 +6,11 @@ import com.gu.paypal.PayPalService
 import com.gu.salesforce.AddressLineTransformer
 import com.gu.services.{ServiceProvider, Services}
 import com.gu.stripe.StripeService
+import com.gu.support.catalog.Sunday
 import com.gu.support.workers._
 import com.gu.support.workers.lambdas.PaymentMethodExtensions.PaymentMethodExtension
 import com.gu.support.workers.states.{CreatePaymentMethodState, CreateSalesforceContactState}
+import com.gu.support.zuora.api.{DirectDebitGateway, DirectDebitTortoiseMediaGateway, PaymentGateway}
 
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future
@@ -26,11 +28,8 @@ class CreatePaymentMethod(servicesProvider: ServiceProvider = ServiceProvider)
   ): FutureHandlerResult = {
     logger.debug(s"CreatePaymentMethod state: $state")
     createPaymentMethod(
-      state.paymentFields,
-      state.user,
+      state,
       services,
-      state.ipAddress,
-      state.userAgent,
     )
       .map(paymentMethod =>
         HandlerResult(
@@ -43,13 +42,10 @@ class CreatePaymentMethod(servicesProvider: ServiceProvider = ServiceProvider)
   }
 
   private def createPaymentMethod(
-      paymentFields: PaymentFields,
-      user: User,
+      state: CreatePaymentMethodState,
       services: Services,
-      ipAddress: String,
-      userAgent: String,
   ): Future[PaymentMethod] =
-    paymentFields match {
+    state.paymentFields match {
       case stripeHosted: StripeHostedPaymentFields =>
         createStripeHostedPaymentMethod(stripeHosted, services.stripeService)
       case stripe: StripePaymentFields =>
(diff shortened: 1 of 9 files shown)


tomrf1 (approved):

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreatePaymentMethod.scala
```diff
@@ -154,7 +150,18 @@ class CreatePaymentMethod(servicesProvider: ServiceProvider = ServiceProvider)
...
+  def createDirectDebitPaymentMethod(
+      dd: DirectDebitPaymentFields,
+      state: CreatePaymentMethodState,
+  ): Future[DirectDebitPaymentMethod] = {
+    import state.user
+
+    // We only want to send the email address to the Tortoise Media gateway, the GNM gateway doesn't need it
+    val (email, paymentGateway) = state.product match {
+      case paper: Paper if paper.productOptions == Sunday =>
+        (Some(state.user.primaryEmailAddress), DirectDebitTortoiseMediaGateway)
```
johnduffell, line 162: minor but since you're importing state.user you can just refer to this as user
```suggestion
        (Some(user.primaryEmailAddress), DirectDebitTortoiseMediaGateway)
```

johnduffell: @andresilva-guardian is this email address something we also need to consider in MDAPI or will it get migrated automatically?

support-workers/src/test/scala/com/gu/support/workers/JsonFixtures.scala
```diff
@@ -246,6 +253,17 @@ object JsonFixtures {
...
     """
 
+  val sundayPaperJson =
+    """
+      {
+        "productType": "Paper",
+        "currency": "GBP",
+        "billingPeriod" : "Monthly",
+        "fulfilmentOptions" : "HomeDelivery",
+        "productOptions" : "Sunday"
```
johnduffell, line 263:
```suggestion
        "billingPeriod": "Monthly",
        "fulfilmentOptions": "HomeDelivery",
        "productOptions": "Sunday"
```
slightly odd formatting here

support-workers/src/test/scala/com/gu/support/workers/lambdas/CreatePaymentMethodSpec.scala
```diff
@@ -70,6 +75,53 @@ class CreatePaymentMethodSpec extends AsyncLambdaSpec with MockContext {
...
+        // Check the output
+        val createSalesforceContactState = Encoding.in[CreateSalesforceContactState](outStream.toInputStream)
+
+        createSalesforceContactState.isSuccess should be(true)
+        createSalesforceContactState.get._1.paymentMethod match {
+          case dd: DirectDebitPaymentMethod =>
+            withClue("direct debit: " + dd) {
+              dd.PaymentGateway should be(DirectDebitGateway)
+            }
+            dd.Email should be(None)
```
johnduffell, line 96: minor but if the clue is useful we can wrap the email check in it too.  same in the below test
```suggestion
            withClue("direct debit: " + dd) {
              dd.PaymentGateway should be(DirectDebitGateway)
              dd.Email should be(None)
            }
```

support-workers/src/test/scala/com/gu/support/workers/lambdas/CreatePaymentMethodSpec.scala
```diff
@@ -70,6 +75,53 @@ class CreatePaymentMethodSpec extends AsyncLambdaSpec with MockContext {
...
+          case _ => fail()
```
johnduffell, line 97: will it report if we do this?
```suggestion
          case _ => fail("Payment method: " + createSalesforceContactState.get._1.paymentMethod)
```
or do we have to do this?
```suggestion
          case _ => withClue("Payment method: " + createSalesforceContactState.get._1.paymentMethod)(fail())
```

johnduffell (approved):
looks good, however my one question is regarding "we need to pass an email address in the payment method object for the Tortoise account"
Do we have to keep this in sync with identity?  At the moment IIRC membership-workflow only syncs email address updates to SF and zuora.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreatePaymentMethod.scala
rupertbates: This code is no longer needed as the Tortoise account has been reconfigured.

rupertbates:
> looks good, however my one question is regarding "we need to pass an email address in the payment method object for the Tortoise account" Do we have to keep this in sync with identity? At the moment IIRC membership-workflow only syncs email address updates to SF and zuora.

We don't need to pass the email address to GoCardless any more so no need to worry about this.

PR #6927, opened 2025-04-10
Fail on any TRANSACTION_FAILED message from Zuora

rupertbates:
We have noticed that some failures from Zuora are not being alarmed on. This is because if the FailureHandler doesn't requestInfo.failed = true in it's output JSON the failure will be silent and the alarm will not be triggered.

I think that any TRANSACTION_FAILED message should probably be alarmed on. I'm not sure why most weren't being, there's probably a reason lost in the mists of time somewhere, but let's make sure they are for now and we can take it from there.

[**Trello Card**](https://trello.com/c/zexoCkvb/1574-check-why-dd-support-workers-failures-are-not-alarming)


GHaberis (approved):
Looks good 👍

PR #6931, opened 2025-04-10
Don't alarm for card declined errors

rupertbates:
Following on from https://github.com/guardian/support-frontend/pull/6927 we don't want to alarm on card declined errors, as these are common and the user should be able to retry.


GHaberis (approved):
Nice 👍

PR #6938, opened 2025-04-14
Track StripePaymentRequestButton metrics distinctly

tjmw:

When recording Cloudwatch metrics for `PaymentSuccess` track `StripePaymentRequestButton` in the payment provider dimension distinctly.

## Why are you doing this?

It looks like StripePaymentRequestButton payments (any payments via Stripe Express which aren't Apple Pay) appear in the metrics with a PaymentProvider of Stripe. We'd like to alarm on these so first we need to track them distinctly.

## How to test


paul-daniel-dempsey (approved):
LG

PR #6943, opened 2025-04-15
Add alarms for period with no Stripe Apple Pay/Google Pay payments

tjmw:

Adding two new Cloudwatch alarm for going a given period with no Stripe Apple Pay/Google Pay payments for any recurring product.

[**Trello Card**](https://trello.com/c/hPyvKKvX/1582-alarms-to-monitor-payment-methods)

## Why are you doing this?

The Stripe Express integration was broken on the generic checkout for a period and this made us realise there wasn't an alarm to pro-actively alert us.

## How to test

I tested by deploying support-workers in CODE and tested the Apple Pay alarm. Initially it was in an alarm state. Then I put through an Apple Pay transaction (S+) and it changed to OK:

Screenshot 2025-04-16 at 12 54 02

### DIFF
--- a/cdk/lib/support-workers.ts
+++ b/cdk/lib/support-workers.ts
@@ -27,15 +27,31 @@ import {
 } from "aws-cdk-lib/aws-stepfunctions";
 import { LambdaInvoke } from "aws-cdk-lib/aws-stepfunctions-tasks";
 
-type PaymentProvider = "Stripe" | "DirectDebit" | "PayPal" | "StripeApplePay" | "StripePaymentRequestButton";
+const Contribution = "Contribution";
+const Paper = "Paper";
+const GuardianWeekly = "GuardianWeekly";
+const SupporterPlus = "SupporterPlus";
+const TierThree = "TierThree";
+const GuardianAdLite = "GuardianAdLite";
 
 type ProductType =
-  | "Contribution"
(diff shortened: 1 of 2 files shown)


rupertbates:
Would this alarm have caught the issue? Was Apple pay broken as well?

tjmw:
> Would this alarm have caught the issue? Was Apple pay broken as well?

@rupertbates yeah I think it was both Stripe Express payment methods (i.e. Apple and Google). Annoyingly the wildcard product dimension in this PR isn't working how I wanted so there's still work to do here. I was trying to avoid having to enumerate all products specifically, but maybe that can't be avoided.

cdk/lib/support-workers.ts
```diff
@@ -27,15 +27,36 @@ import {
 } from "aws-cdk-lib/aws-stepfunctions";
 import { LambdaInvoke } from "aws-cdk-lib/aws-stepfunctions-tasks";
 
-type PaymentProvider = "Stripe" | "DirectDebit" | "PayPal";
+const Contribution = "Contribution";
```
rupertbates, line 30: It would be good to share these types with the frontend at some point

tjmw: 💯% that'd be great

PR #6945, opened 2025-04-15
Reintroduce the changes to fix 100% discounts

rupertbates:
https://github.com/guardian/support-frontend/pull/6885 introduced a fix for 100% discounts but it also broke checkouts via Apple and Google Pay by removing the `paymentMethodCreation: 'manual'` parameter from the Stripe configuration.
It was then reverted in #6934 when this was discovered.

This PR reintroduces the fix from that PR while leaving the `paymentMethodCreation` parameter in place so that GPay and Apple Pay work correctly.


paul-daniel-dempsey (approved):
Looks good, checked out fine with ApplePay on Code.

PR #6946, opened 2025-04-15
Three new transaction declined messages to ignore

rupertbates:
#6927 flipped the alarm behaviour for transaction failed error messages from Zuora so that we would alarm by default and only not do so for particular specified error messages.

This PR follows on from #6931 which added an exception to the alarm exclusion list by adding ~~two~~ three more error messages which it is safe to ignore.

[**Trello Card**](https://trello.com/c/lFopxdf3/1589-add-more-excluded-messages-to-zuora-transaction-failed-alarm)

### DIFF
--- a/support-workers/src/main/scala/com/gu/support/workers/lambdas/FailureHandler.scala
+++ b/support-workers/src/main/scala/com/gu/support/workers/lambdas/FailureHandler.scala
@@ -64,6 +64,7 @@ class FailureHandler(emailService: EmailService) extends Handler[FailureHandlerS
       "Transaction declined.402 - [card_error/card_declined/do_not_honor] Your card was declined.",
       "Transaction declined.402 - [card_error/card_declined/insufficient_funds] Your card has insufficient funds.",
       "Transaction declined.402 - [card_error/card_declined/try_again_later] Your card was declined.",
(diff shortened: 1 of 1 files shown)


support-workers/src/main/scala/com/gu/support/workers/lambdas/FailureHandler.scala
```diff
@@ -60,8 +60,11 @@ class FailureHandler(emailService: EmailService) extends Handler[FailureHandlerS
     logger.info(s"Attempting to handle error $error")
     val pattern =
       "No such token: (.*); a similar object exists in test mode, but a live mode key was used to make this request.".r
-    val cardDeclinedMessage =
-      "Transaction declined.402 - [card_error/card_declined/do_not_honor] Your card was declined."
+    val cardDeclinedMessages = List(
+      "Transaction declined.402 - [card_error/card_declined/do_not_honor] Your card was declined.",
+      "Transaction declined.402 - [card_error/card_declined/insufficient_funds] Your card has insufficient funds.",
+      "Transaction declined.402 - [card_error/card_declined/try_again_later] Your card was declined.",
+    )
```
rupertbates, line 67: Potentially we could just look for messages that start with 'Transaction declined.402' but I thought I'd stick with the current approach for a bit longer.

maloky (approved):
LGTM 👍

PR #6950, opened 2025-04-16
More transaction declined messages

rupertbates:
This PR adds more transaction declined messages which we want to ignore for now.


tjmw (approved):
Nice 👍

PR #6951, opened 2025-04-16
Fix dependabot action

tjmw:

Fix the GH workflow which runs on dependabot PRs.

## Why are you doing this?

I noticed that this action was failing for dependabot PRs, I think this must have started after merging #6806 to switch to pnpm.

## How to test

I temporarily commented out the condition in the dependbot workflow, and saw the dependabot workflow pass on this PR.


tjmw:
Screenshot 2025-04-16 at 15 01 36

Passes on this PR with the dependabot condition removed.

PR #6969, opened 2025-04-22
Another payment declined message

rupertbates:
Follows on from #6956 and others


paul-daniel-dempsey (approved):
LG

PR #6971, opened 2025-04-23
Remove the digital subscription gift redemption form

rupertbates:

https://github.com/guardian/support-frontend/pull/6736 removed all server side code related to Digital Subscription gift subscriptions. This PR removes the client side code, specifically that related to gift redemptions. 

I think the code relating to selling gift subs has already been removed, but if not it will be by the work to move DS to the generic checkout.

[**Trello Card**](https://trello.com/c/rPpgdwaW/1599-remove-ds-gift-redemption-form)


support-frontend/package.json
```diff
@@ -197,5 +197,5 @@
 	"lint-staged": {
 		"./assets/**/**/*.{js,jsx,ts,tsx}": "eslint"
 	},
-	"packageManager": "pnpm@10.5.0"
+	"packageManager": "pnpm@10.9.0+sha256.ea53ab747ac8b7921de63b42eab1adf021e1fe6ded05f62f00885cc33a46118d"
```
tjmw, line 200: Was this change intentional? I think it should be fine as corepack will handle it, but in case of issues it might be surprising that it happened in this PR!

tjmw (approved):
👏  Great!

support-frontend/package.json
rupertbates: Yes fair point, I did when prompted without really thinking about it!

PR #6974, opened 2025-04-23
Re-instate playwright cron test failure alerts

tjmw:

Re-instating playwright cron test failure alerts

## Why are you doing this?

The alerting part was commented out a few weeks ago (see #6909) due to flakiness, we think on the PayPal sandbox end. The tests have continued running and it looks like they've been much more reliable recently so let's turn the alerting back on so we find out about any genuine failures.


maloky (approved):
LGTM 👍

paul-daniel-dempsey (approved):
lg

PR #6976, opened 2025-04-24
Don't render the CMP banner in lighthouse tests

tjmw:

Hide the CMP banner when running lighthouse tests.

[**Trello Card**](https://trello.com/c/pFHFaLRq/1463-disable-cmp-from-lighthouse-tests)

## Why are you doing this?

We think loading the CMP banner affects metrics like largest-contentful-paint. This isn't something we can control so we want to remove this variability from our tests. Typically the user will have already interacted with the CMP before hitting the landing page.


paul-daniel-dempsey (approved):
Great to isolate this out

PR #6980, opened 2025-04-25
Run bigquery-acquisitions-publisher tests for dependabot PRs

tjmw:

Run bigquery-acquisitions-publisher tests for dependabot PRs

## Why are you doing this?

Dependabot is opening PRs against the bigquery-acquisitions-publisher, so we should test it.


paul-daniel-dempsey (approved):
Looks good to include.

PR #6983, opened 2025-04-28
Add another two Stripe errors to the list we don't want to alarm on

tjmw:

Following on from #6969, #6966 and others, add more Stripe errors to the list we don't want to alarm on.

[**Trello Card**](https://trello.com/c/lFopxdf3/1589-add-more-excluded-messages-to-zuora-transaction-failed-alarm)

## Why are you doing this?

These types of errors aren't anything we can do to fix. We should show informative messages to the user to allow them to fix themselves where possible.

### DIFF
--- a/support-workers/src/main/scala/com/gu/support/workers/exceptions/CardDeclinedMessages.scala
+++ b/support-workers/src/main/scala/com/gu/support/workers/exceptions/CardDeclinedMessages.scala
@@ -12,6 +12,8 @@ object CardDeclinedMessages {
     "Transaction declined.402 - [card_error/card_declined/card_velocity_exceeded] Your card was declined for making repeated attempts too frequently or exceeding its amount limit.",
     "Transaction declined.402 - [card_error/card_declined/revocation_of_authorization] Your card was declined.",
     "Transaction declined.402 - [card_error/authentication_required/authentication_required] Your card was declined. This transaction requires authentication.",
+    "Transaction declined.402 - [card_error/card_declined/fraudulent] Your card was declined.",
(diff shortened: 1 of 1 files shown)


support-workers/src/main/scala/com/gu/support/workers/exceptions/CardDeclinedMessages.scala
```diff
@@ -12,6 +12,8 @@ object CardDeclinedMessages {
     "Transaction declined.402 - [card_error/card_declined/card_velocity_exceeded] Your card was declined for making repeated attempts too frequently or exceeding its amount limit.",
     "Transaction declined.402 - [card_error/card_declined/revocation_of_authorization] Your card was declined.",
     "Transaction declined.402 - [card_error/authentication_required/authentication_required] Your card was declined. This transaction requires authentication.",
+    "Transaction declined.402 - [card_error/card_declined/fraudulent] Your card was declined.",
+    "Transaction declined.402 - [card_error/expired_card/expired_card] Your card has expired.",
```
maloky, line 16: i wonder if there is any 402 code that we want to raise alarm?! sounds like a good candidate to exclude all those by code.

maloky (approved):
Small comment for future iterations. otherwise looks good 👍

support-workers/src/main/scala/com/gu/support/workers/exceptions/CardDeclinedMessages.scala
tjmw: Yeah you may be right. For now I think having all these enumerated individually is quite useful as there's a future step to map these to user friendly messages (in some cases, Stripe's documentation makes it clear which should be shown to users and which shouldn't).

PR #6986, opened 2025-04-28
Upgrade support-frontend to TypeScript 5.5.4

tjmw:

This PR upgrades the version of TypeScript used by support-frontend from `5.1.6` to `5.5.4`. This supercedes the dependabot PR #6986 which attempted to go to `5.8.3`. The jump to `5.8.3` caused some peer dependency warnings across a number of guardian packages. Initially I followed these by upgrading things but this lead to needing to do a major upgrade of eslint to v9, which I don't think we're ready for just yet (we're using other packages which don't work with eslint 9 yet). 

## Why are you doing this?

Stepping stone to get us closer to the latest TypeScript version.

## How to test

I've run the smoke tests against a local instance and against CODE ✅


rupertbates (approved):
Nice 👍

PR #6988, opened 2025-04-29
Ophan component event for the similar products checkbox

rupertbates:

This PR adds an Ophan component event for when a user checks or unchecks the similar products consent checkbox. It records the new value of the checkbox, ie. when the checkbox is checked and the user clicks to uncheck it a `false` value is sent. 

## How to test
- Go to the generic checkout making sure that you are in an AB test variant which sees the checkbox eg. https://support.thegulocal.com/uk/checkout?product=SupporterPlus&ratePlan=Monthly#ab-similarProductsConsent=VariantB
- Open the network tab
- Uncheck the similar product consents checkbox
- Check that there is a new request in the network tab similar to: https://ophan.theguardian.com/img/2?viewId=ma2eewtv3cj00qe9yjy2&componentEvent={"component":{"componentType":"ACQUISITIONS_OTHER","id":"similar-products-consent-checkbox"},"action":"CLICK","value":"false"}

### DIFF
--- a/support-frontend/assets/helpers/tracking/behaviour.ts
+++ b/support-frontend/assets/helpers/tracking/behaviour.ts
@@ -34,13 +34,14 @@ const trackThankYouPageLoaded = (
 	}
 };
 
-const trackComponentClick = (componentId: string): void => {
+const trackComponentClick = (componentId: string, value?: string): void => {
 	trackComponentEvents({
 		component: {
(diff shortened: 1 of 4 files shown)


paul-daniel-dempsey (approved):
Looks perfect, followed instructions to check component event fired and worked fine.

PR #6991, opened 2025-04-29
Ensure acquisition events are passed to consent setter lambda queue

michaelbjacobson:
This PR makes two changes to facilitate user-set consents being sent to Identity:
- Defines a new type, `ProductName`, used to send a format of the product name in-keeping with the capitalisation and spacing used for Zuora product names, eg. `"Guardian Weekly"`, as-opposed to the existing `"GUARDIAN_WEEKLY"` that is included already in the acquisition event.
- Via the CDK stack for the `bigquery-acquisitions-publisher` lambda, defines a new EventBridge rule on the existing `acquisitions-bus` which tells the bus to enqueue acquisition events to the existing `soft-opt-in-consent-setter-queue`, for consumption by the `soft-opt-in-consent-setter-IAP` lambda.

So far, the only consent for which we're rolling out explicit, user-set opt-ins is `similar-guardian-products`.


cdk/bin/cdk.ts
```diff
@@ -97,11 +97,13 @@ new AcquisitionEventsApi(app, "Acquisition-Events-API-PROD", {
 new BigqueryAcquisitionsPublisher(app, "BigqueryAcquisitionsPublisher-CODE", {
   stack: "support",
   stage: "CODE",
+  softOptInConsentSetterQueueArn: "arn:aws:sqs:eu-west-1:865473395570:soft-opt-in-consent-setter-queue-CODE",
```
johnduffell, line 100: I think you can use something like ${aws::accountid} to avoid hard coding it here

cdk/lib/bigquery-acquisitions-publisher.ts
```diff
@@ -66,7 +72,34 @@ export class BigqueryAcquisitionsPublisher extends GuStack {
...
       eventBus: eventBus,
-      targets: [new SqsQueue(queue)],
+      targets: [new SqsQueue(acquisitionsQueue)],
+    });
+
+    const softOptInConsentSetterQueue =
+      aws_sqs.Queue.fromQueueArn(this, "SoftOptInConsentSetterQueue", props.softOptInConsentSetterQueueArn)
+
+    // Rule which passes events from support-workers on to soft opt-in queue
+    new Rule(this, "SoftOptInToSQSRule", {
```
johnduffell, line 82:
```suggestion
    new Rule(this, "SoftOptInQueueRule", {
```

johnduffell: Maybe it's changed since, but I thought the plan was to put this rule in one of the support-service-lambdas CDKs, perhaps the soft opt in one? And you converted that over to CDK in order to make that possible?

support-payment-api/src/main/scala/model/acquisition/AcquisitionDataRowBuilder.scala
```diff
@@ -44,7 +46,7 @@ object AcquisitionDataRowBuilder {
       paymentProvider = Some(paymentProvider),
       printOptions = None,
       browserId = acquisitionData.browserId,
-      identityId = acquisition.identityId.map(_.toString),
+      identityId = acquisition.identityId,
```
johnduffell, line 49: this was a warning, as identityId already contains a String (I vaguely remember that it was incorrectly modelled as a Long in some places)

support-workers/src/main/scala/com/gu/acquisitions/AcquisitionDataRowBuilder.scala
```diff
@@ -54,14 +54,15 @@ import com.gu.support.zuora.api.ReaderType
...
+    val productInformation = getProductInformation(commonState.product)
     val acquisitionTypeDetails = getAcquisitionTypeDetails(commonState)
     val printOptions = printOptionsFromProduct(commonState.product, commonState.user.deliveryAddress.map(_.country))
 
     AcquisitionDataRow(
       eventTimeStamp = DateTime.now(DateTimeZone.UTC),
-      product = acquisitionProduct,
-      amount = amount,
+      product = productInformation.product,
+      productName = productInformation.productName,
```
johnduffell, line 64: it's a bit odd that we now have product and productName, where product is the UPPER_CASE ophan thrift enum that lumps all print/GW together, and productName is the more nuanced zuora catalog style one that separates all products out.

I think this is a side effect of the fact that this bus was originally just added as a layer before the old ophan call, so the format is pretty much identical to that?

I'd rather rename product to ophanProduct but that would be too much work.

PR #6993, opened 2025-04-30
Refactor benefits checklist logic into a separate file

rupertbates:
This PR refactors the code responsible for creating the benefits checklist on the three tier landing page.

The reason for the refactor is that we are trying to break code out of checkoutComponent.tsx to make it easier to work with. 

The PR also adds a test for the output of the function which loads benefits from the landing page configuration to ensure that nothing was broken in the refactor.


tomrf1 (approved):
nice

PR #6957, opened 2025-04-17
ABC test 'Similar Products' marketing consent on the generic checkout

rupertbates:
## What are you doing in this PR?
We want to add a checkbox to all of our checkouts to explicitly capture a user's consent to 'similar Guardian products' marketing.

This PR adds an AB test to measure the impact of adding this checkbox to the conversion rate/LTV of the checkout.

## Changes
- Add an AB test to control the placement of the checkbox (see screenshots). This AB test is currently inactive so the PR can be merged before we put the test live.
- Pass the value of the checkbox to the backend as a new boolean property named `similarProductsConsent`.
- Pass the `similarProductsConsent` boolean through the support-workers state classes to the SendAcquisitionEvent lambda where it will be added to the json which is sent to the `acquisitions-bus` event bus. This can then be picked up and processed in the work being done by the value team.

Note that the similarProductsConsent boolean is actually an Option[Boolean] (scala) or boolean | undefined (typescript). This is because some products will never have this checkbox (Guardian Ad-Lite and Observer). 

From the point of view of the code which picks up this value and actually sets a user's consent status, a value of true or false means set the status to that value whereas a missing value means leave the user's consent status unchanged. See [this discussion](https://chat.google.com/room/AAAAFNJwTZs/Ff0hZzdC1Vo/Ff0hZzdC1Vo?cls=10) for more information.

## TODO
- [ ] Update Playwright tests to check the checkbox (when AB test is active) - follow up PR
- [ ] Don't show similar consents checkbox for Guardian Ad-Lite or Observer checkouts  - follow up PR

[**Trello Card**](https://trello.com/c/x5lNpUFx/1575-a-b-test-on-adding-soft-opt-in-tickbox-to-all-recurring-checkouts-except-ad-lite-guardian-weekly)

## How to test
https://support.code.dev-theguardian.com/uk/checkout?product=HomeDelivery&ratePlan=Everyday#ab-similarProductsConsent=VariantB
https://support.code.dev-theguardian.com/uk/checkout?product=HomeDelivery&ratePlan=Everyday#ab-similarProductsConsent=VariantA

## Test Variants
*Control:*
No checkbox will be added to the checkout, this matches the existing situation.

*Variant A:*
Screenshot 2025-04-22 at 16 40 38
*Variant B:*
Screenshot 2025-04-23 at 15 20 34

## JSON which appears on the Acquisitions event bus
*User in one of the variants who opted in to receiving marketing:*
``` json
{
    "version": "0",
[...]

PR #6994, opened 2025-04-30
Don't show the similar products consent checkbox for some products

rupertbates:
Now that we are explicitly asking users to consent to 'Similar Guardian products' marketing by adding a checkbox to the checkout (see #6957), we need to make sure that we are not showing the checkbox for products where we do not want marketing consent - Guardian Ad-Lite and The Observer.

This PR adds a new boolean named `hideSimilarProductsConsent` onto the `ProductDescription.ratePlan` object in the product catalog. This has to be at the rate plan level rather than the product level because Observer is a rate plan of the home delivery and subscription card products, not a product in its own right.

### DIFF
--- a/support-frontend/assets/helpers/productCatalog.ts
+++ b/support-frontend/assets/helpers/productCatalog.ts
@@ -37,9 +37,15 @@ export type ProductDescription = {
 		{
 			billingPeriod: 'Annual' | 'Monthly' | 'Quarterly';
 			label?: string;
+			hideSimilarProductsConsent?: boolean;
 		}
 	>;
 };
+
+export const showSimilarProductsConsentForRatePlan = (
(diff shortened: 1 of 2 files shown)


support-frontend/assets/helpers/productCatalog.ts
```diff
@@ -37,10 +37,16 @@ export type ProductDescription = {
 		{
 			billingPeriod: 'Annual' | 'Monthly' | 'Quarterly';
 			label?: string;
+			hideSimilarProductsConsent?: boolean;
```
rupertbates, line 40: I have defined this as hide products and then negated it below so that it only needs to be added to three rate plans, rather than adding it to every rate plan in all the other products.

tjmw (approved):
Great 👍

PR #7002, opened 2025-05-05
Hosted checkout: verify that user entered form fields match the checkout session

tjmw:

When the checkout session is created store the SHA256 hash of the user entered form fields in the `metadata` field. When retrieving the checkout session in support-workers, compare the hash with the hash of the form fields provided. If they don't match, ~~return an error~~ send a CloudWatch metric event.

[**Trello Card**](https://trello.com/c/88hS6qF5/1571-ensure-that-the-checkout-session-matches-the-form-fields-persisted-hash-these-and-add-the-hash-to-the-checkout-session-metadata)

## Why are you doing this?

There's no scenario where the form fields (which we stash in session storage) should change between redirecting the user to Stripe and coming back to the support site to complete the transaction, so let's make sure it doesn't. Before rolling this out in a way which actually prevents a purchase, let's emit a metric event so we can see if it ever happens first.

## How to test

I've verified that I can still put a Sunday only subscription through. I've also run the smoke tests against this branch.

I deliberately broke the hash check and can see the metric in CODE:

Screenshot 2025-05-12 at 13 53 55

### DIFF
--- a/support-frontend/app/controllers/CreateSubscriptionController.scala
+++ b/support-frontend/app/controllers/CreateSubscriptionController.scala
@@ -150,7 +150,7 @@ class CreateSubscriptionController(
 
     urls match {
       case Some((validSuccessUrl, validCancelUrl)) =>
-        val fieldsHash = FormFieldsHash.create(requestBody)
+        val fieldsHash = FormFieldsHash.createFromSupportWorkersRequest(requestBody)
         stripeCheckoutSessionService
           .createCheckoutSession(
             stripePublicKey,
--- a/support-frontend/app/controllers/CreateSubscriptionController.scala
+++ b/support-frontend/app/controllers/CreateSubscriptionController.scala
@@ -18,6 +18,7 @@ import io.circe.Encoder
 import io.circe.generic.semiauto._
 import io.circe.syntax._
 import lib.PlayImplicits._
+import models.FormFieldsHash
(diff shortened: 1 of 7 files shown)


support-frontend/app/controllers/CreateSubscriptionController.scala
```diff
@@ -32,7 +32,6 @@ import services.{
   RecaptchaService,
   StripeCheckoutSessionService,
   TestUserService,
-  UserBenefitsApiService,
```
tjmw, line 35: Unrelated - this was unused.

support-workers/src/main/scala/com/gu/support/workers/lambdas/CreatePaymentMethod.scala
```diff
@@ -47,7 +47,7 @@ class CreatePaymentMethod(servicesProvider: ServiceProvider = ServiceProvider)
   ): Future[PaymentMethod] =
     state.paymentFields match {
       case stripeHosted: StripeHostedPaymentFields =>
-        createStripeHostedPaymentMethod(stripeHosted, services.stripeService)
+        createStripeHostedPaymentMethod(stripeHosted, services.stripeService, state.user)
```
tjmw, line 50: The user is passed here because it contains everything we need to generate the hash.

rupertbates (approved):
Looks good 👍

PR #7006, opened 2025-05-07
Use defaultChecked instead of checked prop in `SimilarProductsConsent.tsx`

tjmw:

Update `SimilarProductsConsent` to specify `defaultChecked` instead of `checked`.

## Why are you doing this?

Previously the checkbox reverted to a checked state any time the component re-rendered. This way we specify the initial state but let the browser handle the state going forward (this is an uncontrolled component).

### DIFF
--- a/support-frontend/assets/pages/[countryGroupId]/components/formOnSubmit.ts
+++ b/support-frontend/assets/pages/[countryGroupId]/components/formOnSubmit.ts
@@ -31,6 +31,24 @@ import { createSubscription } from './createSubscription';
 import type { PaymentMethod } from './paymentFields';
 import { FormSubmissionError } from './paymentFields';
 
+const getConsentValue = (
+	formData: FormData,
+	abParticipations: Participations,
+) => {
+	const similarProductsCheckbox = formData.get('similarProductsConsent');
+	if (similarProductsCheckbox) {
+		return similarProductsCheckbox === 'on';
+	}
+
+	if (
+		['VariantA', 'VariantB'].includes(
+			abParticipations.similarProductsConsent ?? '',
+		)
+	) {
+		return false;
+	}
+	return;
+};
 export const submitForm = async ({
 	geoId,
 	productKey,
@@ -72,6 +90,8 @@ export const submitForm = async ({
 		productFields,
 	);
 
+	const similarProductsConsent = getConsentValue(formData, abParticipations);
+
 	const promoCode = promotion?.promoCode;
 	const appliedPromotion =
 		promoCode !== undefined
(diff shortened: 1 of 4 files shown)


rupertbates (approved):
👍

support-frontend/assets/pages/[countryGroupId]/components/formOnSubmit.ts
```diff
@@ -72,6 +84,11 @@ export const submitForm = async ({
 		productFields,
 	);
 
+	const similarProductsConsent = getConsentValue(
+		formData,
+		userWasShownCheckbox,
+	);
+
```
Comment (maloky) [inline support-frontend/assets/pages/[countryGroupId]/components/formOnSubmit.ts:91]: what do you think about skip passing a boolean into the function and evaluate it right away.
```suggestion
	const similarProductsConsent = !userWasShownCheckbox && getConsentValue(formData);

```

support-frontend/assets/pages/[countryGroupId]/components/formOnSubmit.ts
```diff
@@ -31,6 +31,16 @@ import { createSubscription } from './createSubscription';
 import type { PaymentMethod } from './paymentFields';
 import { FormSubmissionError } from './paymentFields';
 
+const getConsentValue = (formData: FormData, userWasShownCheckbox: boolean) => {
+	const similarProductsCheckbox = formData.get('similarProductsConsent');
+	if (similarProductsCheckbox) {
+		return similarProductsCheckbox === 'on';
+	}
```
Comment (maloky) [inline support-frontend/assets/pages/[countryGroupId]/components/formOnSubmit.ts:38]: i guess this will always be true if `similarProductsCheckbox`is `on` because unchecked never shows up.
we could 
```suggestion
	    const similarProductsCheckbox = formData.get('similarProductsConsent');
      return similarProductsCheckbox === 'on';
	}
```

maloky:
Just to understand this better @tjmw @rupertbates , similarProductConsent type is a boolean only or can be undefined?

maloky (approved):
minor comments that are not mandatory for this fix. great work guys

tjmw:
> Just to understand this better @tjmw @rupertbates , similarProductConsent type is a boolean only or can be undefined?

Yeah it can be `true`, `false` or `undefined` (undefined doesn't set the consent value explicitly, instead it triggers the old/existing behaviour.

maloky:
> > Just to understand this better @tjmw @rupertbates , similarProductConsent type is a boolean only or can be undefined?
> 
> Yeah it can be `true`, `false` or `undefined` (undefined doesn't set the consent value explicitly, instead it triggers the old/existing behaviour.

makes sense, so my suggestions are not valid anymore as they only return `true` or `false`.

rupertbates:
> Just to understand this better @tjmw @rupertbates , similarProductConsent type is a boolean only or can be undefined?

It can be undefined if a product doesn't need to ask for consent, for instance Guardian Ad-lite

PR #7008, opened 2025-05-08
Typescript support-workers infrastructure

rupertbates:
I previously created a PR https://github.com/guardian/support-frontend/pull/6480 which started the work of moving support-workers to Typescript from Scala. However I included in that PR the migration of the first lambda (CreatePaymentMethod) which made the PR very big.

Due to the number of comments on that PR I thought it would be easier to break it down into separate PRs - this one which will set up the basic CDK, build and deployment stuff required to move to TS support-workers, and then a later one which contains the code to actually migrate the first lambda.

There is some dummy code in this PR (dummyTypescriptLambda.ts & dummy.test.ts) which are just placeholders for code which will come later, but they need to be there to check that all the foundational stuff works correctly.


cdk/lib/support-workers.ts
```diff
@@ -98,22 +95,16 @@ export class SupportWorkers extends GuStack {
         Fn.importValue(table)
       ),
     });
+    const parameterStorePolicy = new PolicyStatement({
```
rupertbates, line 98: For the new TS lambdas I am going to store config in parameter store rather than S3 as the Scala lambdas do.

.github/workflows/support-workers-build.yml
```diff
@@ -36,12 +36,18 @@ jobs:
...
-          cache-dependency-path: cdk/yarn.lock
+          cache-dependency-path: '**/yarn.lock'
 
       - name: Build CFN from CDK
         run: ./script/ci
         working-directory: cdk
 
+      - name: Build TypeScript lambdas
+        run: |
+          yarn install --frozen-lockfile
```
tjmw, line 47: Should we be using `pnpm` for a new TypeScript project?

cdk/lib/support-workers.ts
```diff
@@ -178,6 +169,58 @@ export class SupportWorkers extends GuStack {
...
+      lambdaName: string,
+      additionalPolicies: PolicyStatement[] = []
+    ) => {
+      const lambdaId = `${lambdaName}Lambda`;
+      const lambdaTSFile = lambdaId.charAt(0).toLowerCase() + lambdaId.slice(1);
+      const lambda = new GuLambdaFunction(this, lambdaId, {
+        ...lambdaDefaultConfig,
+        app: "support-workers-typescript",
+        fileName: `support-workers.zip`,
+        runtime: Runtime.NODEJS_20_X,
```
tjmw, line 182: Maybe we should use Node 22, which is active LTS currently? (Node 20 is in maintenance mode).

cdk/lib/support-workers.ts
```diff
@@ -178,6 +169,58 @@ export class SupportWorkers extends GuStack {
...
+        handler: `lambdas/${lambdaTSFile}.handler`,
+        functionName: `${this.stack}-${lambdaName}Lambda-${this.stage}`,
+        initialPolicy: [
+          s3Policy,
+          cloudWatchLoggingPolicy,
+          parameterStorePolicy,
+          ...additionalPolicies,
+        ],
+      });
+      this.overrideLogicalId(lambda, {
```
tjmw, line 192: Is this needed? Does the lambda already exist?

support-workers/.nvmrc
```diff
@@ -0,0 +1 @@
+20.11.0
```
tjmw, line 1: As above, shall we go for Node 22?

tjmw (approved):
Looks great 👍 Good idea to break this down into smaller PRs.

cdk/lib/support-workers.ts
rupertbates: Yes I'm going to do that actually, I needed to upgrade CDK first though which I did on Friday.

.github/workflows/support-workers-build.yml
rupertbates: Yes good idea

cdk/lib/support-workers.ts
rupertbates: It's not for the dummy lambda but it will be once we start adding the real ones in.

PR #7018, opened 2025-05-13
Changes to use countdown configuration from landing page tests

LAKSHMIRPILLAI:
This PR has changes to use the countdown settings  from the Landing Page Tests
The related support-admin -console changes are in the [SAC PR](https://github.com/guardian/support-admin-console/pull/729)
[**Trello Card**](https://trello.com/c/ZSSk3jdO/1061-landing-page-tool-countdown-config)
## Why are you doing this?
This is to support Landing Page Tooling and add countdown to it 
## How to test
Deploy both the SAC and Support Frontend branches in CODE and test

## Screenshots

For RRCP Countdown Config

image

image

### Note: We can change the countdown timer colour as well, but needs design approval

image

image

### Note: We can have different counter values for different regions in separate tests

image

image

###  NOTE:

If we have multiple countries selected for the same landing page with the counter , all the countries will show the same timer values irrespective of  the time zone.So in that case , need to create separate landing page tests for different countries

Also for all regions other than GB , use start and deadline timestamp in UTC


support-frontend/assets/helpers/globalsAndSwitches/landingPageSettings.ts
```diff
@@ -17,6 +17,28 @@ export interface TickerSettings {
...
+	foregroundColor: string;
+}
+interface CountdownSettings {
+	label: string;
+	countdownStartInMillis: string;
+	countdownDeadlineInMillis: string;
+	theme: CountdownTheme;
+}
+
+export type CountdownSetting = {
```
tomrf1, line 31: The model name here is quite confusing.
Since this model is almost the same, and the timestamp fields in `CountdownSettings` need to be renamed, how about:
```
interface CountdownSettings {
	label: string;
	countdownStartTimestamp: string;
	countdownDeadlineTimestamp: string;
	theme: CountdownTheme;
}

type ParsedCountDownSettings = CountdownSettings & {
	countdownStartInMillis: number;
	countdownDeadlineInMillis: number;
}
```

tomrf1 (approved):

PR #7019, opened 2025-05-13
Another card declined error message to ignore

rupertbates:
We have been adding card declined error messages to our list of errors not to alarm on for a while now. This is the latest one.


paul-daniel-dempsey (approved):
lg - perhaps check the build however.

PR #7009, opened 2025-05-08
create AB test Logic on one time Contribution checkout for Soft Opt-In

maloky:
## What are you doing in this PR?

1. Create a new ab test participation to track soft opt-in (SOI) consent on the on time contribution checkout.
2. Add backend support for `similarProductsConsent` flag for both Stripe and Paypal payment methods to the Acquisition Event Bus.
**Stripe**:  Pass the value directly on the request payload for the payment-api
**PayPal**:  Set the value as cookie so it can be passed by the return endpoint after PayPal redirect. 
3. Refactor Implementation for getting the `similarProductsConsent` value from the Form.

Refactor main points:
- use a hidden input element to track consent value when displayed
- remove unneeded prop that evaluates component visibility since the input will provide us that value
- move getConsentValue to own module, and use hidden input value as flag.
  

[**Trello Card**](https://trello.com/c/nez975uB/1576-build-a-b-c-test-on-adding-soft-opt-in-tick-box-to-one-time-journey-checkout)

## Why are you doing this?
We are rolling out soft in consent to one time contribution checkout page.

### Locally
| SimilarProductsConsent | OneTimeContributionConsent |
| -- | -- | 
| [Control](https://support.thegulocal.com/uk/checkout?product=SupporterPlus&ratePlan=Monthly#ab-similarProductsConsent=Control) | [Control](https://support.thegulocal.com/uk/one-time-checkout#ab-oneTimeContributionConsent=Control) | 
| [VariantA](https://support.thegulocal.com/uk/checkout?product=SupporterPlus&ratePlan=Monthly#ab-similarProductsConsent=VariantA) | [VariantA](https://support.thegulocal.com/uk/one-time-checkout#ab-oneTimeContributionConsent=VariantA) | 
| [VariantB](https://support.thegulocal.com/uk/checkout?product=SupporterPlus&ratePlan=Monthly#ab-similarProductsConsent=VariantB) | [VariantB](https://support.thegulocal.com/uk/one-time-checkout#ab-oneTimeContributionConsent=VariantB) | 
### Similar Products

1. Open the browser dev tools in the network tab and filter for `create`.
2. On the payload tab for the request check the sent object.

Screenshot 2025-05-08 at 12 59 38

### One Time Contribution

1. Open the browser dev tools in the network tab and filter for `create-payment`.
2. On the payload tab for the request check the sent object.

Screenshot 2025-05-08 at 16 42 57

## Accessibility test checklist

[...]

PR #7026, opened 2025-05-15
Remove SoftOptInsService from payment-api

rupertbates:
This PR removes the SoftOptinService from the payment api because after https://github.com/guardian/support-service-lambdas/pull/2821 and  https://github.com/guardian/support-frontend/pull/7009 it is no longer needed. 

The consent value is now sent via the acquisitions event bus.


maloky (approved):

johnduffell (approved):
thanks for remaking the PR to be clearer.  I think the old one was based off another PR that was force pushed, meaning there were identical commits in main with a different commit id.

PR #7027, opened 2025-05-16
Simplify countdown

tomrf1:
Removes some of the complexity from the landing page countdown logic.
In particular, it removes much of the code from `ThreeTierLanding`, which largely duplicates the checks done inside the `Countdown` component.

- all timestamp parsing/checking happens inside the `Countdown` component
- the state used to decide whether to display the countdown is now inside the `Countdown` component
- a `setHeadingOverride` callback is used to set the heading in the `ThreeTierLanding` component
- the `Countdown` component now takes the same `CountdownSettings` model that RRCP publishes

One advantage is that the storybook stories no longer take a `showCountdown` boolean, and instead decide to render the countdown based on the actual timestamps.


LAKSHMIRPILLAI (approved):
Great!!

PR #7031, opened 2025-05-22
Handle discount rate plans when fetching contribution amount in ProcessSupporterProductDataLambda

rupertbates:
I noticed that there were a lot of errors in the ProcessSupporterRatePlanItem lambda. This is the one which is responsible for syncing subscriptions between Zuora and the SupporterProductData Dynamo store which drives the user benefits api.

This PR fixes an error which is caused by the way we were trying to find out the contribution amount for recurring contribution records - we assumed that there would only be one rate plan on the subscription, and at one point that was true, but now we are using discounts for cancellation saves in MMA, and this means that the subscription will have two rate plans - one for the contribution and one for the discount.

The error was actually caused by trying to deserialise a discount rate plan which doesn't have a price to a case class model which insists that it does, so this PR makes the model more permissive and then takes account of that optionality when working out the contribution amount.

### DIFF
--- a/supporter-product-data/src/main/scala/com/gu/lambdas/ProcessSupporterRatePlanItemLambda.scala
+++ b/supporter-product-data/src/main/scala/com/gu/lambdas/ProcessSupporterRatePlanItemLambda.scala
@@ -2,14 +2,16 @@ package com.gu.lambdas
 
 import com.amazonaws.services.lambda.runtime.Context
 import com.gu.conf.ZuoraQuerierConfig
+import com.gu.lambdas.ProcessSupporterRatePlanItemLambda.{contributionIds, discountIds}
 import com.gu.model.StageConstructors
 import com.gu.model.dynamo.SupporterRatePlanItemCodecs._
 import com.gu.model.sqs.SqsEvent
+import com.gu.model.zuora.response.MinimalZuoraSubscription
 import com.gu.monitoring.SafeLogging
 import com.gu.okhttp.RequestRunners.configurableFutureRunner
 import com.gu.services.{AlarmService, ConfigService, DiscountService, ZuoraSubscriptionService}
 import com.gu.supporterdata.model.Stage.{CODE, PROD}
-import com.gu.supporterdata.model.{Stage, SupporterRatePlanItem}
+import com.gu.supporterdata.model.{ContributionAmount, Stage, SupporterRatePlanItem}
 import com.gu.supporterdata.services.SupporterDataDynamoService
 import io.circe.parser._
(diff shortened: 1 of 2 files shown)


supporter-product-data/src/main/scala/com/gu/model/zuora/response/MinimalZuoraSubscription.scala
```diff
@@ -1,19 +1,13 @@
...
-    charges <- ratePlans.ratePlanCharges.headOption
-  } yield ContributionAmount(charges.price, charges.currency)
-}
+case class MinimalZuoraSubscription(ratePlans: List[RatePlan])
 
-case class RatePlan(ratePlanCharges: List[RatePlanCharge])
+case class RatePlan(id: String, ratePlanCharges: List[RatePlanCharge])
 
-case class RatePlanCharge(price: BigDecimal, currency: String)
+case class RatePlanCharge(price: Option[BigDecimal], currency: String)
```
rupertbates, line 10: Discount rate plans don't have a price set

supporter-product-data/src/main/scala/com/gu/lambdas/ProcessSupporterRatePlanItemLambda.scala
```diff
@@ -91,14 +95,22 @@ object ProcessSupporterRatePlanItemLambda extends SafeLogging {
 class ContributionAmountFetcher(config: ZuoraQuerierConfig) extends SafeLogging {
   lazy val zuoraService = new ZuoraSubscriptionService(config, configurableFutureRunner(60.seconds))
 
-  def fetchContributionAmountFromZuora(supporterRatePlanItem: SupporterRatePlanItem) =
+  private def contributionAmountFromZuoraSubscription(subscription: MinimalZuoraSubscription) =
+    for {
+      contributionRatePlan <- subscription.ratePlans.find(ratePlan => contributionIds.contains(ratePlan.id))
+      charges <- contributionRatePlan.ratePlanCharges.headOption
```
tjmw, line 101: Just to confirm - it's always the first rate plan charge we care about here?

tjmw (approved):
Not very familiar with this code but I think it looks good!

supporter-product-data/src/main/scala/com/gu/lambdas/ProcessSupporterRatePlanItemLambda.scala
rupertbates: Yes there's only ever one charge on a contribution rate plan

PR #7032, opened 2025-05-23
Handle errors from our backend after returning from the Stripe hosted checkout

tjmw:

If a user returns from the Stripe hosted checkout and receives an error from our backend, clear the form fields and start the checkout process over again.

As part of this I've migrated the checkout session retrieval logic to a custom hook which wraps useState and useEffect. The reason I'm doing this is so that we can manipulate/clear the checkout session state and cause a re-render as expected.

[**Trello Card**](https://trello.com/c/9YD5ZbkO/1647-handle-server-side-errors-when-user-returns-from-stripe-hosted-checkout)

TODO:

- [x] Show an error at the top of the page when the user returns to the checkout
- [x] Payment method selector should be returned to default state
- [x] No error at the bottom of the page (currently showing generic transaction declined message)
- [x] Ensure top error message appears in viewport

## Why are you doing this?

When a user returns from the Stripe hosted checkout to the generic checkout we automatically retrieve their form details and submit the form. It's theoretically possible (though we haven't seen it) that we may return an error from our backend at this point. If this happened the user would be left on the form with an error. Resubmitting at this stage will always just resubmit to the backend and use their Stripe hosted checkout details. Depending on the type of error this may not be sufficient to fix the problem. Because this is such an unlikely scenario, we decided the best thing to do is reset the form fields and effectively start over, which is what this PR adds.

## How to test

- [x] Smoke tests pass

Locally I can simulate this scenario by breaking the backend request after the user returns to our checkout. Screenshot of how this error state looks (generic error anchored to the top of the viewport/form fields cleared):

Screenshot 2025-06-04 at 10 10 10

### DIFF
--- a/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/checkout.tsx
@@ -33,8 +33,7 @@ import type { Participations } from '../../helpers/abTests/models';
 import type { LandingPageVariant } from '../../helpers/globalsAndSwitches/landingPageSettings';
 import type { LegacyProductType } from '../../helpers/legacyTypeConversions';
 import { getLegacyProductType } from '../../helpers/legacyTypeConversions';
-import type { CheckoutSession } from './checkout/helpers/stripeCheckoutSession';
-import { getFormDetails } from './checkout/helpers/stripeCheckoutSession';
+import { useStripeHostedCheckoutSession } from './checkout/hooks/useStripeHostedCheckoutSession';
 import { CheckoutComponent } from './components/checkoutComponent';
 
 type Props = {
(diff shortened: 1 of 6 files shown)


support-frontend/assets/pages/[countryGroupId]/checkout/hooks/useStripeHostedCheckoutSession.ts
```diff
@@ -0,0 +1,36 @@
...
+export function useStripeHostedCheckoutSession(
+	checkoutSessionId: string | null,
+): [CheckoutSession | undefined, () => void] {
+	const [checkoutSession, setCheckoutSession] = useState<
+		CheckoutSession | undefined
+	>();
+
+	useEffect(() => {
+		if (checkoutSessionId) {
+			const persistedFormData = getFormDetails(checkoutSessionId);
```
Comment (maloky) [inline support-frontend/assets/pages/[countryGroupId]/checkout/hooks/useStripeHostedCheckoutSession.ts:17]: I wonder since `checkoutSessionId` will never change in the component lifecycle, if we should initialise the state with the `getFormDetails` function so we only call it once regardless of component re-renders.

Comment (tjmw) [reply support-frontend/assets/pages/[countryGroupId]/checkout/hooks/useStripeHostedCheckoutSession.ts:17]: (Although I think because of the way `checkoutSessionId` is read, and the fact that it won't change, the useEffect will only be run once anyway)

Comment (tjmw) [reply support-frontend/assets/pages/[countryGroupId]/checkout/hooks/useStripeHostedCheckoutSession.ts:17]: I've tried out this change locally and it works. I'm _slightly_ hesitant about going this direction because right now this hook will handle a scenario if it were possible for `checkoutSessionId` to change within the component lifecycle. I think that's unlikely, but if feels like the current approach might be slightly more robust in case we ever need to handle that happening. I'm going to mull it over...

support-frontend/assets/pages/[countryGroupId]/checkout.tsx
```diff
@@ -276,7 +253,11 @@ export function Checkout({
 		);
 	}, []);
 
-	const checkoutSession = attemptToRetrievePersistedFormData(urlSearchParams);
+	const checkoutSessionIdUrlParam = 'checkoutSessionId';
+	const maybeCheckoutSessionId = urlSearchParams.get(checkoutSessionIdUrlParam);
+
+	const [checkoutSession, clearCheckoutSession] =
+		useStripeHostedCheckoutSession(maybeCheckoutSessionId);
```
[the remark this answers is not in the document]
Comment (maloky) [reply support-frontend/assets/pages/[countryGroupId]/checkout.tsx:260]: let's go for consistency 👍

maloky (approved):
Nice work 👍

PR #7040, opened 2025-05-29
add LastPlanAddedDate__c value to all new subs to allow accurate cooling off period

johnduffell:
We refund supporter plus subscriptions if they are cancelled within 14 days of taking out (or switching)

This means that we need to determine the date they are taken out or last switched accurately.

At the moment, we use the charge Effective start date, however that gets updated every time there's any amendment (switch, renewal, price rise, etc etc)

Having done some research, it doesn't seem easy to find the date of the first amendment that added a given rate plan (not to mention that price rises and migrations don't count).  It would be possible to follow the chain of amendments, or query all the subscriptions with the same name and check the charges, or even run an export zoql query asynchronously.

However, another solution it just to store the value on the subscription or charge as a custom field, and update it at any point where they become eligiible for any kind of refund window.

I have added a field in zuora sandbox on Subscription called LastPlanAddedDate__c
image

I didn't put it on plan in the end, because if we do a price rise or migration, we should ideally copy it over.  And we only put one plan on a subscription, so we can get away with putting it at that level.

This PR makes main site acqusisition add the field to all subscriptions when they are taken out.  It's only needed on S+ at present, but it makes enough sense for all subscriptions to include by default.  It is the date the agreement is made, rather than the first payment date.

Everything is open to question at this stage including the field name and even the whole approach, so please comment anything and everything.

Tested in sandbox locally
https://apisandbox.zuora.com/platform/subscriptions/71a1431927d970b055171cb4b74b1069
image

Custom field has been added in PROD (9th Sept 2025)

Tested in CODE and working 
image
https://apisandbox.zuora.com/platform/subscriptions/71a140ce7e99928fbd892e9622d9d157

Followup work
- add similar logic to switch api (RC to S+ mainly, but membership to RC should ideally support it) https://github.com/guardian/support-service-lambdas/pull/2875
- add similar logic to new product api https://github.com/guardian/support-service-lambdas/pull/2950
- wait 14 days and change suporter plus cancellation api look at the field https://github.com/guardian/support-service-lambdas/pull/2951

### DIFF
--- a/support-workers/src/test/scala/com/gu/zuora/subscriptionBuilders/PaperSubscriptionBuilderSpec.scala
+++ b/support-workers/src/test/scala/com/gu/zuora/subscriptionBuilders/PaperSubscriptionBuilderSpec.scala
@@ -4,7 +4,12 @@ import com.gu.i18n.Currency.GBP
 import com.gu.support.catalog.{EverydayPlus, HomeDelivery, NationalDelivery, Sunday}
 import com.gu.support.config.TouchPointEnvironments.CODE
 import com.gu.support.promotions.PromotionService
-import com.gu.support.workers.JsonFixtures.{salesforceContact, stripePaymentMethodObj, userJsonWithDeliveryAddress, userJsonWithDeliveryAddressOutsideLondon}
+import com.gu.support.workers.JsonFixtures.{
+  salesforceContact,
+  stripePaymentMethodObj,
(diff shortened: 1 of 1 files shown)


andresilva-guardian (approved):
Thanks @johnduffell , the `LastPlanAddedDate__c` seems like a great idea, good job 💪

PR #7054, opened 2025-06-02
Guardian Weekly : Generic Checkout AB-Test Enable

paul-daniel-dempsey:

Enable `guardianWeeklyGenericCheckout` ab-test, set `control` and `variant` to cover 100% of audience.

Ensure these (non-gifting) GuardianWeekly paths are part of the test->
`.../<countryGroupId>/subscribe/weekly`
`.../<countryGroupId>/subscribe/weekly/checkout`
`.../subscribe/weekly/checkout`
`.../<countryGroupId>/thank-you`
`.../<countryGroupId>/thankyou`

## Screenshots

Checkouts->
|control|variant|
|----|----|
|image|image|

### DIFF
--- a/support-frontend/assets/helpers/abTests/abtestDefinitions.ts
+++ b/support-frontend/assets/helpers/abTests/abtestDefinitions.ts
@@ -29,7 +29,7 @@ export const pageUrlRegexes = {
 		},
 		// Includes landing, original & generic checkout/thankyou pages
 		subsWeeklyPages:
-			'(/subscribe/weekly/checkout$)(/.*)?$|(?:/(uk|us|ca|eu|nz|int))(?:/(subscribe/weekly|checkout|one-time-checkout|contribute|thankyou|thank-you))(/.*)?$',
+			'(/subscribe/weekly/checkout$)(/.*)?$|(?:/(uk|us|ca|eu|nz|int))(?:/(subscribe/weekly|checkout|thankyou|thank-you))(/.*)?$',
 	},
 };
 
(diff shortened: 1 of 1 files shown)


rupertbates (approved):
👍

support-frontend/assets/helpers/abTests/abtestDefinitions.ts
```diff
@@ -27,8 +27,9 @@ export const pageUrlRegexes = {
 			paperLandingPage: /^\/uk\/subscribe\/paper?(\?.*)?$/,
 			weeklyLandingPage: /\/subscribe\/weekly\/checkout?(\?.*)?$/,
 		},
+		// Includes landing, original & generic checkout/thankyou pages
 		subsWeeklyPages:
-			'(/??/subscribe(\\?.*)?$|/??/subscribe/weekly(\\/checkout)?(\\?.*)?$)',
+			'(/subscribe/weekly/checkout$)(/.*)?$|(?:/(uk|us|ca|eu|nz|int))(?:/(subscribe/weekly|checkout|thankyou|thank-you))(/.*)?$',
```
tjmw, line 32: Is the $ in the weekly checkout path doing the right thing here? I think that anchors to the end, but we're then specifying that their might be some optional characters after (which can't happen if we already anchored to the end). What characters might appear after `/subscribe/weekly/checkout`? If none then we could lose the `(/.*)?`, or if it's possible we should probably lose the `$` right after checkout, e.g.

```suggestion
			'(/subscribe/weekly/checkout)(/.*)?$|(?:/(uk|us|ca|eu|nz|int))(?:/(subscribe/weekly|checkout|thankyou|thank-you))(/.*)?$',
```

PR #7069, opened 2025-06-10
Remove two redundant payment-api alarms

tjmw:

Removing two payment-api alarms which I think are already covered by other alarms with a lower threshold:

`NoPaypalPaymentsInTwoHours247Alarm` is covered by `NoPaypalPaymentsInOneHourAlarm`.
`NoStripePaymentsInThreeHours247Alarm` is covered by `NoStripePaymentsInOneHourAlarm`.

## Why are you doing this?

I don't think there's a reason to have these - they all go to the same channel so it'll be the same people looking at these as those with a lower threshold.


rupertbates (approved):
Makes sense

PR #7074, opened 2025-06-11
Fix ThankYou header when there is no amount

maloky:

### DIFF
--- a/support-frontend/assets/pages/supporter-plus-thank-you/components/thankYouHeader/GenericHeading.tsx
+++ b/support-frontend/assets/pages/supporter-plus-thank-you/components/thankYouHeader/GenericHeading.tsx
@@ -1,0 +1,14 @@
+import { headerTitleText } from './headingStyles';
+
+export default function GenericHeading({
+	contributorName,
+}: {
+	contributorName: string;
+}) {
+	return (
+		<h1 css={headerTitleText}>
+			Thank you <span data-qm-masking="blocklist">{contributorName}</span> for
+			your valuable contribution
+		</h1>
+	);
+}
--- a/support-frontend/assets/pages/supporter-plus-thank-you/components/thankYouHeader/heading.tsx
+++ b/support-frontend/assets/pages/supporter-plus-thank-you/components/thankYouHeader/heading.tsx
@@ -1,5 +1,3 @@
-import { css } from '@emotion/react';
-import { from, titlepiece42 } from '@guardian/source/foundations';
 import type { PaymentStatus } from 'helpers/forms/paymentMethods';
 import type { IsoCurrency } from 'helpers/internationalisation/currency';
 import type {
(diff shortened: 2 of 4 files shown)


support-frontend/assets/pages/supporter-plus-thank-you/components/thankYouHeader/heading.tsx
```diff
@@ -8,37 +6,21 @@ import type {
...
-		font-size: 28px;
-	}
-`;
-
 type HeadingProps = {
 	name: string | null;
 	productKey: ActiveProductKey;
 	isOneOffPayPal: boolean;
-	amount: number;
+	amount: number | undefined;
```
paul-daniel-dempsey, line 23: can this be optional instead? ie `amount?: number`

support-frontend/assets/pages/supporter-plus-thank-you/components/thankYouHeader/heading.tsx
```diff
@@ -63,6 +45,10 @@ function Heading({
 
 	const contributorName = name && name.length < 10 ? name : '';
 
+	if (!amount) {
```
paul-daniel-dempsey, line 48: Why not reverse this and put `if (amount)`? Then only need the single dropout ` <GenericHeading` at the bottom?

support-frontend/assets/pages/supporter-plus-thank-you/components/thankYouHeader/GenericHeading.tsx
```diff
@@ -0,0 +1,14 @@
+import { headerTitleText } from './headingStyles';
+
+export default function GenericHeading({
+	contributorName,
```
tjmw, line 4: Super minor point but I think this can be an empty string, would it be worth making that clear by naming this something like `maybeContributorName`? Happy to leave this for a future PR if we want to get this one merged quickly.

support-frontend/assets/pages/supporter-plus-thank-you/components/thankYouHeader/heading.tsx
tjmw: I quite like the explicitness of this, it makes it really clear how we're handing the case where we don't have an amount 👍

tjmw (approved):
Thanks for getting this fixed Jorge!

support-frontend/assets/pages/supporter-plus-thank-you/components/thankYouHeader/GenericHeading.tsx
maloky: wouldn't we prefer to say it's optional prop than inferring  possible value from the name convention? like: 
`const contributorName = name && name.length < 10 ? name : undefined;`

PR #7076, opened 2025-06-13
Increase threshold payment api stripe express alarm

tjmw:

Following on from #7075 this PR slightly increases the threshold of the payment-api Stripe Express period alarm.

I've also refactored the alarm slightly to derive the description from the parameters, so it's impossible for the description to get out of sync.

## Why are you doing this?

This alarm has fired occassionally recently, so make it a little less sensitive.

### DIFF
--- a/cdk/lib/payment-api.ts
+++ b/cdk/lib/payment-api.ts
@@ -323,7 +323,7 @@ export class PaymentApi extends GuStack {
     });
 
     const stripeExpressMetricDuration = Duration.minutes(5);
-    const stripeExpressEvaluationPeriods = 12; // The number of 5 minute periods in 1 hour
+    const stripeExpressEvaluationPeriods = 18; // The number of 5 minute periods in 90 minutes
     const stripeExpressAlarmPeriod = Duration.minutes(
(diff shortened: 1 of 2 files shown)


cdk/lib/payment-api.ts
```diff
@@ -322,6 +322,11 @@ export class PaymentApi extends GuStack {
       snsTopicName: `alarms-handler-topic-${this.stage}`,
     });
 
+    const stripeExpressMetricDuration = Duration.minutes(5);
+    const stripeExpressEvaluationPeriods = 18; // The number of 5 minute periods in 90 minutes
+    const stripeExpressAlarmPeriod = Duration.minutes(
+      stripeExpressMetricDuration.toMinutes() * stripeExpressEvaluationPeriods
+    );
```
maloky, line 329: just food for though, but i tend to prefer grouping these values under an object. Are there any preferences on this independent variables vs grouped ones? 
```suggestion
        const stripeExpressMetric = {
        duration: Duration.minutes(5);
        evaluationPeriods: 18; // The number of 5 minute periods in 90 minutes
        alarmPeriod = Duration.minutes(stripeExpressMetricDuration.toMinutes() * stripeExpressEvaluationPeriods
     }
```

PR #7083, opened 2025-06-16
Make generic checkout the only checkout for GW

rupertbates:
We have been running an AB test between the generic checkout and the old Guardian Weekly checkout for over a week now and there is no difference in performance so we are switching to using the generic checkout exclusively and removing the old checkout.

This PR just switches all traffic to the generic checkout, there will be a follow up PR to remove old code.


maloky (approved):
LGTM 👍

PR #7087, opened 2025-06-17
Remove valibot

rupertbates:
Currently we have a mix of Valibot and Zod schemas in support-frontend. We have decided to standardise on Zod for now, mostly because it is more of a standard across our code bases (support-service-lambdas uses it extensively for example).

This PR removes Valibot and migrates the schemas which used it to Zod.


support-frontend/assets/helpers/globalsAndSwitches/window.ts
```diff
@@ -220,6 +211,7 @@ const ProductCatalogSchema = object({
 });
 
 /**
+ * TODO: try making this schema stricter no that we are no using valibot
```
rupertbates, line 214: I am going to do this in a follow up PR

PR #7099, opened 2025-06-23
Make missing Stripe card payment fields on the checkout more obvious to the user

tjmw:

When a user selects the Credit/Debit card payment method and submits the form without entering one or more of the required card fields, show specific errors next to the field and scroll to them.

[**Trello Card**](https://trello.com/c/sXWIyRpu/1676-handle-missing-payment-fields-on-the-generic-checkout)

## Why are you doing this?

Currently if the user performs the actions above, it's not very clear what's gone wrong. A very generic error message is show below the pay/submit button and we don't make it clear what the user can do to fix things. There's no indication of the actual problem:

Screenshot 2025-07-01 at 12 37 48

Currently this causes an exception which is reported to Sentry. The volumes of this error suggests this situation is actually happening to users.

Now, when this happens we'll show an error message next to the relevant field(s) and scroll the viewport to the payment section of the form to make sure these errors are visible. See screenshots below.

Note: This PR contains a small styling change to ensure the expiry and cvc fields are always equal width. Without this an error on only one of the fields rendered next to each other (expiry/CVC) sometimes causes the other field to be weirdly narrow. Originally I included another change to ensure they are always vertically aligned with each other but this created a detachment from the label which we felt could be confusing for users so I rolled this back.

## How to test

* Visit the generic checkout
* Fill in personal details (email etc)
* Select Credit/Debit card but leave fields blank
* Submit the form
* See new contextual inline errors

I'd expect the Sentry error mentioned above to decrease in volume.

## Screenshots

All fields missing:

Screenshot 2025-07-01 at 12 19 38

Some fields missing:

Screenshot 2025-07-02 at 11 27 34

### DIFF
--- a/support-frontend/assets/components/stripeCardForm/elementDecorator.tsx
+++ b/support-frontend/assets/components/stripeCardForm/elementDecorator.tsx
@@ -72,6 +72,10 @@ const mobileCreditDebitIcons = css`
 	}
 `;
 
+const stripeFieldsContainer = css`
+	flex: 1;
+`;
+
 type ElementRenderProps = {
 	id: string;
 	options: StripeCardElementOptions;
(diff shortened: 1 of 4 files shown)


rupertbates (approved):
Looks good to me

PR #7101, opened 2025-06-25
Revert "Merge pull request #7044 from guardian/dependabot/npm_and_yarn/cdk/guardian/cdk-61.5.1"

tjmw:
## What are you doing in this PR?

This reverts commit 9616bbe3f42a48580d1952a0affec9eae86212e9, reversing changes made to 22f951c815340db9dbb8a4df11f1c1467ee3c44d. (Added in #7044).

## Why are you doing this?

The gu/cdk release 61.5.0 (which the reverted PR includes) needs to be released in two stages because of a breaking change. Currently deploys are failing. I'm going to revert and then use the two step approach.

PR #7103, opened 2025-06-25
Move forward with @guardian/cdk 61.8.1 upgrade

tjmw:

This upgrade originally happened in #7044. However this caused deploys to fail so it was rolled back (in #7101). Now we're moving forward again using the strategy documented here: https://github.com/guardian/cdk/blob/main/CHANGELOG.md#minor-changes-3

## Why are you doing this?

We found that after we reverted #7044 typechecking started failing. We're not 100% sure why but fixing forward seems to get things working.


paul-daniel-dempsey (approved):
LG , feels like we will need to revert this temporary security change in future, but its a requirement so need to do

PR #7105, opened 2025-06-25
Add tests for the CreatePaymentMethod lambda handler

tjmw:

Building on top of #7104 add some tests for the lambda handler, specifically testing the thing which #7104 touches (which Direct Debit payment gateway to use).

## Why are you doing this?

Build confidence in the code by increasing test coverage, and make it easier to add to these tests in future.

## How to test

`pnpm --filter support-workers test`

### DIFF
--- a/support-workers/src/typescript/test/createPaymentMethodLambda.test.ts
+++ b/support-workers/src/typescript/test/createPaymentMethodLambda.test.ts
@@ -1,0 +1,57 @@
+import { handler } from '../lambdas/createPaymentMethodLambda';
+import createPaymentDirectDebitPaper from './fixtures/createPaymentMethod/paperDirectDebit.json';
+import {
+	CreatePaymentMethodState,
+	createPaymentMethodStateSchema,
+} from '../model/stateSchemas';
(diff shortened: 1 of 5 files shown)


support-workers/src/typescript/test/createPaymentMethodLambda.test.ts
```diff
@@ -0,0 +1,57 @@
...
+				productType: 'Paper',
+				currency: 'GBP',
+				billingPeriod: BillingPeriod.Monthly,
+				fulfilmentOptions: 'HomeDelivery',
+				productOptions: 'Everyday',
+			};
+
+			const result = await handler(wrapPayload(payload));
+
+			expect(result.state.paymentMethod.PaymentGateway).toBe('GoCardless');
```
maloky, line 33: we could just import these gateways as const  implementation, so we have a single source of truth.

tjmw: Done in 3869a716f24a901e185c97db30e45d5f01b99581.

PR #7116, opened 2025-07-01
Make dependabot run against the root pnpm workspace

rupertbates:
This PR configures dependabot to run against the root of the pnpm workspace rather than on individual projects. 

This is because the root is now where the pnpm-lock.yml file lives, and should hopefully fix the issue of dependabot PRs not updating the lockfile.


maloky (approved):
Looks Good 👍 how do we test this?

PR #7122, opened 2025-07-02
Prevent flash of checkout without overlay before redirect to thank you page

tjmw:

In the checkoutComponent move the `setIsProcessingPayment(false)` call to catch block of the form submit handler because it has the effect of removing the processing overlay. Previously it was below/outside of the catch and this had the effect of removing the overlay even in the case where the submitForm was successful which meant there was a flash of the checkout with no overlap before the redirect to the thank you page. There's no need to remove the overlay unless we're dropping the user back on the checkout with errors in case something went wrong.

## Why are you doing this?

Prevent the flash of checkout without overlay when payment is successful before the (client side) redirect to the thank you page.

## How to test

Put through a transaction and watch the redirect to the thank you page.

### DIFF
--- a/support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx
+++ b/support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx
@@ -617,11 +617,12 @@ export function CheckoutComponent({
 					`An error occurred in checkoutComponent.tsx while trying to submit the form: ${errorMessage}`,
 				);
 			}
-			// This state update is in the catch block because it has the effect of
-			// removing the processing overlay. Previously it was outside of the catch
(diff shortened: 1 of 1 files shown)


support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx
```diff
@@ -617,8 +617,13 @@ export function CheckoutComponent({
 					`An error occurred in checkoutComponent.tsx while trying to submit the form: ${errorMessage}`,
 				);
 			}
+			// This state update is in the catch block because it has the effect of
```
Comment (paul-daniel-dempsey) [inline support-frontend/assets/pages/[countryGroupId]/components/checkoutComponent.tsx:620]: The description is perfect, do you need previously in the comments if its in the PR though?

PR #7126, opened 2025-07-04
Extract Ad-Lite e2e tests out to start from landing page

tjmw:

Extracting Ad-Lite e2e tests out to start from landing page instead of the going directly to the checkout.

This establishes a new pattern where checkout tests start on the landing page for the relevant product. Eventually this will be applied to all the current tests defined in `checkout.test.ts`, which will be grouped by the landing page. I've extracted a method `completeCheckout` which can be used to fill in the checkout and submit. This is shared by the new tests which start on the landing page and the existing checkout tests.

[**Trello Card**](https://trello.com/c/lnE2F0us/1690-update-playwright-e2e-tests-to-begin-checkout-flows-on-the-relevant-landing-page)

## Why are you doing this?

We'd like to start tests on the landing page as this is closer to how real users will interact with the site.

### DIFF
--- a/support-e2e/tests/test/adLiteCheckout.ts
+++ b/support-e2e/tests/test/adLiteCheckout.ts
@@ -1,0 +1,33 @@
+import test, { expect } from '@playwright/test';
+import { setupPage } from '../utils/page';
+import { completeCheckout } from './checkout';
+
+export type TestDetails = {
+	paymentType: string;
+	ratePlan: string;
+};
+
+export const testAdLiteCheckout = (testDetails: TestDetails) =>
+	test(`Ad-Lite - ${testDetails.paymentType}`, async ({ context, baseURL }) => {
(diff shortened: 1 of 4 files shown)


support-e2e/tests/test/adLiteCheckout.ts
```diff
@@ -0,0 +1,33 @@
+import test, { expect } from '@playwright/test';
+import { setupPage } from '../utils/page';
+import { completeCheckout } from './checkout';
+
+export type TestDetails = {
+	paymentType: string;
+	ratePlan: string;
+};
+
+export const testAdLiteCheckout = (testDetails: TestDetails) =>
```
tjmw, line 10: Is including the word `Checkout` in this function misleading, since we're testing more than just the checkout? We're testing the journey from the landing page. Would `testAdLitePurchase` be better? Or something else?

PR #7130, opened 2025-07-07
Add /UTS to the student landing page path

tjmw:

Adds /UTS to the student landing page path.

Previously for non au region paths we redirected to the 3 tier landing page. This was kind of a placeholder until we broadened the student landing page to other regions. Now that the university is part of the URL I think that changes things, and that visiting /student/UTS when not in Australia doesn't really make sense (and won't in future). So I've removed the other regions and the non-region path from the routes completely. Interested in thoughts on this.

## Why are you doing this?

Feedback from MRR.


maloky (approved):
super clean and clear 👍

PR #7017, opened 2025-05-12
Typescript CreatePaymentMethodLambda

rupertbates:
## What are you doing in this PR?
This PR migrates the CreatePaymentMethod lambda to Typescript it follows on from #7008 and is part of the work to fully migrate support-workers from Scala to Typescript.

[**Trello Card**](https://trello.com/c/GOjj1Ewj/1675-createpaymentmethod-lambda)

## Things to test
- [x] Make sure FailureHandler understands the errors thrown from TS
- [x] Stripe payment in US, UK & AU
- [x] Stripe hosted payment
- [x] PayPal payment
- [x] GoCardless payment

PR #7140, opened 2025-07-10
Migrate the CreateSalesforceContact Lambda to Typescript

rupertbates:

This PR follows on from https://github.com/guardian/support-frontend/pull/7017 and is the next step in moving support-workers from Scala to Typescript.

This is a simpler lambda than the CreatePaymentMethod lambda and most of the schemas and model classes were already in place so this PR is smaller than previous ones. 

TODO:
- [x] Check error handling, particularly readonly mode

[**Trello Card**](https://trello.com/c/UPZDpNiB/1745-createsalesforcecontact-lambda)

## How to test

- Deploy this branch of support-workers to CODE
- Run smoke tests against CODE
- All the resulting support-workers executions should succeed

### DIFF
--- a/support-workers/src/typescript/lambdas/createSalesforceContactLambda.ts
+++ b/support-workers/src/typescript/lambdas/createSalesforceContactLambda.ts
@@ -37,7 +37,10 @@ export const handler = async (
 			createSalesforceContactState.user,
 			createSalesforceContactState.giftRecipient,
 		);
-		return createNextState(createSalesforceContactState, contactRecords);
+		return {
+			...state,
+			state: createNextState(createSalesforceContactState, contactRecords),
+		};
 	} catch (error) {
 		throw asRetryError(error);
 	}
@@ -51,6 +54,7 @@ const createNextState = (
 		case 'Contribution':
 			return {
 				productSpecificState: {
+					productType: 'Contribution',
 					product: state.product,
 					paymentMethod: state.paymentMethod,
 					salesForceContact: contactRecords.buyer,
@@ -61,6 +65,7 @@ const createNextState = (
 		case 'SupporterPlus':
 			return {
 				productSpecificState: {
+					productType: 'SupporterPlus',
 					billingCountry: user.billingAddress.country,
 					product: state.product,
 					paymentMethod: state.paymentMethod,
@@ -73,6 +78,7 @@ const createNextState = (
 		case 'TierThree':
 			return {
 				productSpecificState: {
+					productType: 'TierThree',
 					user: state.user,
 					product: state.product,
 					paymentMethod: state.paymentMethod,
@@ -89,6 +95,7 @@ const createNextState = (
 		case 'GuardianAdLite':
 			return {
 				productSpecificState: {
+					productType: 'GuardianAdLite',
 					product: state.product,
 					paymentMethod: state.paymentMethod,
 					salesForceContact: contactRecords.buyer,
@@ -98,6 +105,7 @@ const createNextState = (
 		case 'GuardianWeekly':
 			return {
 				productSpecificState: {
+					productType: 'GuardianWeekly',
 					user: state.user,
 					giftRecipient: state.giftRecipient,
 					product: state.product,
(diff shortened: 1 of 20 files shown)


support-workers/src/typescript/lambdas/createSalesforceContactLambda.ts
```diff
@@ -0,0 +1,154 @@
...
+
+const createNextState = (
+	state: CreateSalesforceContactState,
+	contactRecords: SalesforceContactRecords,
+): CreateZuoraSubscriptionState => {
+	switch (state.product.productType) {
+		case 'Contribution':
+			return {
+				productSpecificState: {
+					productType: 'Contribution',
```
tjmw, line 57: Would it be worth using `state.product.productType` to avoid duplicating the value? Applies to each branch of this switch statement, and is consistent with how we're retrieving other fields from `state`.

support-workers/src/typescript/lambdas/createSalesforceContactLambda.ts
```diff
@@ -0,0 +1,154 @@
...
+					product: state.product,
+					paymentMethod: state.paymentMethod,
+					salesForceContact: contactRecords.buyer,
+					similarProductsConsent: state.similarProductsConsent,
+				},
+				...state,
```
tjmw, line 63: There are a couple of pieces which are common to all branches of this switch - nesting the product specific state under `productSpecificState` and merging with `state` at the top level. Would it be worth building up the product specific state separately, and then doing the common pieces in one place which applies to all products?

support-workers/src/typescript/model/address.ts
```diff
@@ -5,21 +5,23 @@ export const countrySchema = z.enum(isoCountries);
 
 export const addressSchema = z.object({
 	lineOne: z.string().nullable(),
-	lineTwo: z.string().nullable(),
+	lineTwo: z.string().nullish(),
```
tjmw, line 8: So this can be null or undefined, unlike the others which can only be null?

support-workers/src/typescript/test/errorHandler.test.ts
```diff
@@ -0,0 +1,26 @@
+import { SalesforceError, salesforceErrorCodes } from '../services/salesforce';
+import { asRetryError } from '../util/errorHandler';
+
+describe('errorHandler', () => {
+	test('should throw a the correct retry error for Salesforce errors', () => {
```
tjmw, line 5:
```suggestion
	test('should throw the correct retry error for Salesforce errors', () => {
```

support-workers/src/typescript/test/salesforce.it.test.ts
```diff
@@ -0,0 +1,166 @@
...
+
+	test('should upsert a gift SalesforceContactRecord', async () => {
+		const inputState: WrappedState<CreateSalesforceContactState> =
+			wrapperSchemaForState(createSalesforceContactStateSchema).parse(
+				createGiftSubscription,
+			);
+		const result = await handler(inputState);
+
+		expect(result.state.product.productType).toBe('GuardianWeekly');
+		if (result.state.productSpecificState.productType === 'GuardianWeekly') {
```
tjmw, line 156: When would this not be true? Can we control this in the tests? Or is this necessary to keep TypeScript happy somehow?

rupertbates: Yes it's just to keep Typescript happy, it would be nice if it could infer it from the expect statement but it can't

support-workers/src/typescript/lambdas/createSalesforceContactLambda.ts
tjmw: Ohh sorry I didn't spot that, yeah sounds like a future thing 👍

support-workers/src/typescript/test/salesforce.it.test.ts
tjmw: 👍 Would it be worth adding a comment mentioning this?

support-workers/src/typescript/lambdas/createSalesforceContactLambda.ts
rupertbates: I had a go at doing this but ran into difficulties convincing TS that the product returned from the sub function was the same type as the one in the calling function. It didn't save much code anyway because we still need the specific productType as mentioned in a previous comment so I've left it as is.

tjmw (approved):

PR #7152, opened 2025-07-16
Remove google analytics domains

rupertbates:
We don't use Google Analytics any more so we don't need the domains in our CSP.

### [Trello Card](https://trello.com/c/9hmxduoU/1493-remove-references-to-google-analytics-from-the-support-site-csp)


maloky:
Not that they are doing too much, but just for consistency we could remove also from the DEV conf. 

- https://github.com/guardian/support-frontend/blob/21e0295e36e18abe4b653073f8128a8102823992/support-frontend/conf/DEV.public.conf#L39
- https://github.com/guardian/support-frontend/blob/21e0295e36e18abe4b653073f8128a8102823992/support-frontend/conf/DEV.public.conf#L65

maloky (commented):
LGTM 👍

